PR #161, opened 2016-01-31
Implement `diesel database reset`

mcasper:
This has a lot of changes because it does a few things at the same time:
* Changed most functions to take a generic Connection. This gets us a
little closer to making the CLI generic over the backend, and allows for
easier testing, as it gives us control over the connection it uses.

* Document more functions and add some messages to the user when
running setup/reset.

* Add the `diesel database reset` command, and move `diesel setup` to be
a subcommand of `diesel database`.

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -13,7 +13,8 @@ diesel setup --database-url='postgres://localhost/my_db'
 diesel migration generate create_users_table
 ```
 
-You'll see that two files were generated for you,
+You'll see that a `migrations/` directory was generated for you (by the setup
+command), and two sql files were generated,
 `migrations/{current_timestamp}_create_users_table/up.sql` and
 `migrations/{current_timestamp}_create_users_table/down.sql`. You should edit
 these files to show how to update your schema, and how to undo that change.
@@ -32,18 +33,18 @@ CREATE TABLE users (
 DROP TABLE USERS;
 ```
 
-You can then run your new migration by running `diesel migration run`. Make sure
-that you set the `DATABASE_URL` environment variable first, or pass it directly
-by doing `diesel migration run --database-url="postgres://localhost/your_database"`
-Alternatively, you can call
-[`diesel::migrations::run_pending_migrations`][pending-migrations] from
+You can then run your new migration by running `diesel migration run`. Make
+sure that you set the `DATABASE_URL` environment variable first, or pass it
+directly by doing `diesel migration run
+--database-url="postgres://localhost/your_database"` Alternatively, you can
+call [`diesel::migrations::run_pending_migrations`][pending-migrations] from
 `build.rs`.
 
 Diesel will automatically keep track of which migrations have already been run,
 ensuring that they're never run twice.
 
-### Commands
-#### `setup`
+## Commands
+## `diesel setup`
 Searches for a `migrations/` directory, and if it can't find one, creates one
 in the same directory as the first `Cargo.toml` it finds.  It then tries to
 connect to the provided DATABASE_URL, and will create the given database if it
@@ -51,6 +52,18 @@ cannot connect to it. Finally it will create diesel's internal table for
 tracking which migrations have been run, and run any existing migrations if the
 internal table did not previously exist.
 
+## `diesel database`
+#### `database setup`
+Tries to connect to the provided DATABASE_URL, and will create the given
+database if it cannot connect to it.  It then creates diesel's internal
+migrations tracking table if it needs to be created, and runs any pending
+migrations if it created the internal table.
+
+#### `database reset`
+Drops the database specified in your DATABASE_URL if it can, and then runs
+`database database setup`.
+
+## `diesel migration`
 #### `migration generate`
 Takes the name of your migration as an argument, and will create a migration
 directory with `migrations/` in the format of
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -3,13 +3,13 @@ extern crate chrono;
 extern crate clap;
 extern crate diesel;
 
-mod setup_error;
+mod database_error;
 
 use chrono::*;
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
 use diesel::{migrations, Connection};
 use diesel::connection::PgConnection;
-use self::setup_error::SetupError;
+use self::database_error::DatabaseError;
 use std::{env, fs};
 use std::error::Error;
 use std::path::{PathBuf, Path};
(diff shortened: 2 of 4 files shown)


sgrif:
Given that `diesel setup` does stuff that has nothing to do with the database, are we sure that moving the namespace is the right thing to do?

sgrif:
Does it make sense to split out `diesel setup` from `diesel database setup`? I suppose `diesel setup` in that case would be `diesel database setup` + `mkdir migrations`?

diesel_cli/README.md
```diff
@@ -51,6 +51,10 @@ cannot connect to it. Finally it will create diesel's internal table for
 tracking which migrations have been run, and run any existing migrations if the
 internal table did not previously exist.
 
+#### `database reset`
+Drops the database specified in your DATABASE_URL if it can, and then runs
+`database setup`.
```
sgrif: The fact that this is in backticks makes me feel like this should be `diesel database setup`. Thoughts?

diesel_cli/src/main.rs
```diff
@@ -103,69 +112,121 @@ fn run_migration_command(matches: &ArgMatches) {
...
+/// Drops the database specified in the connection url. It returns a
+/// `DatabaseError::ConnectionError` if it can't connect to the postgres
+/// connection url, and returns a `DatabaseError::QueryError` if it is unable
+/// to drop the database.
+fn drop_database<Conn: Connection>(conn: &Conn, database: &String)
+    -> Result<(), DatabaseError>
+{
+    println!("Dropping database: {}", database);
+    try!(conn.silence_notices(|| {
+             conn.execute(&format!("DROP DATABASE IF EXISTS {};", database))
```
sgrif: Indentation

diesel_cli/src/main.rs
```diff
@@ -103,69 +112,121 @@ fn run_migration_command(matches: &ArgMatches) {
...
-        migrations::create_schema_migrations_table_if_needed(&connection)
-            .map_err(handle_error).unwrap();
-        migrations::run_pending_migrations(&connection)
-            .map_err(handle_error).unwrap();
-    }
+/// Creates the __diesel_schema_migrations table if it doesn't exist. If the
+/// table didn't exist, it also runs any pending migrations. Returns a
+/// `DatabaseError::ConnectionError` if it can't create the table, and exits
+/// with a migration error if it can't run migrations.
+fn setup_schema_table<Conn: Connection>(conn: &Conn) -> Result<(), DatabaseError> {
```
sgrif: This function name feels like it describes much less than the documentation does.

diesel_cli/src/main.rs
```diff
@@ -103,69 +112,121 @@ fn run_migration_command(matches: &ArgMatches) {
...
+    if !schema_table_exists(conn).map_err(handle_error).unwrap() {
+        try!(migrations::create_schema_migrations_table_if_needed(conn));
+        migrations::run_pending_migrations(conn).unwrap_or_else(handle_error);
+    };
+    Ok(())
+}
+
+/// Drops the database specified in the connection url. It returns a
+/// `DatabaseError::ConnectionError` if it can't connect to the postgres
```
sgrif: We're going to need to abstract this soon. Can we not mention postgres here?

sgrif:
On cursory review this looks pretty reasonable, but I'm a bit fried tonight and the diff is rendering rather silly for me. I'll take a more in depth look tomorrow. Thanks for tackling this so quickly.

mcasper:
I think having `diesel setup` and `diesel database setup` be separate things makes sense, as you're right, setup does at least one thing that has nothing to do with the database. That also makes it a little nicer for `database reset`, as it only has to call the database specific portions of setup.

diesel_cli/README.md
mcasper: Agreed

diesel_cli/src/main.rs
mcasper: To inside the `silence_notices` block?

diesel_cli/src/main.rs
mcasper: Agreed, though I've been trying to come up with a good generic name for this. `system_url`? `admin_url`? `url_i_connect_to_to_drop_tables`?

sgrif: `database url` is fine.

sgrif: Also for SQLite this method is literally just "delete a file". Which I find kinda funny

diesel_cli/src/main.rs
sgrif: You indented 2 levels, you should only indent 1.

diesel_cli/src/main.rs
mcasper: Yeah but how do we want to represent it in functions that deal with both a specific database's url, and the generic url?

sgrif: Actually let's just be more vague. "This will return an error if it is unable to drop the database."

diesel_cli/src/main.rs
```diff
@@ -110,62 +126,120 @@ fn run_setup_command(matches: &ArgMatches) {
...
+        ("setup", Some(args)) => setup_database(args).unwrap_or_else(handle_error),
+        ("reset", Some(args)) => reset_database(args).unwrap_or_else(handle_error),
+        _ => unreachable!("The cli parser should prevent reaching here"),
+    };
+}
+
+fn reset_database(args: &ArgMatches) -> Result<(), DatabaseError> {
+    let (database, postgres_url) = split_pg_connection_string(&database_url(args));
+    let connection = try!(PgConnection::establish(&postgres_url));
+    try!(drop_database(&connection, &database));
```
sgrif: I don't think `drop_database` should take a connection. The process of doing this will vary too much by backend.

mcasper: In that case should `create_database` and `drop_database` be methods on Connection? In that case the `split_pg_connection_string` will move to a more logical place, as that will also be backend specific (and as you pointed out, Sqlite just `rm -rf`s a directory)

sgrif: I think it's too early to make that determination.

mcasper:
Alright, I think everything has been resolved. Final review @sgrif @mfpiccolo ?

sgrif:
I'm about to head to bed but assuming all the comments I left previously
have been addressed this is fine.

sgrif:
I need committers who aren't on Pacific time. :P

mcasper:
:smile:

mfpiccolo:
Nicely done Matt :+1:

Issue #157, opened 2016-01-31
[SQLite] SQLite has no support for the `DEFAULT` keyword


sgrif:
### Note: PRs addressing this issue should target the `diesel-sqlite-support` branch, not `master`.

I don't see a good way to work around this and support batch inserts for the time being. I think the short term solution comes in 3 steps.

- Make `Insertable` generic over the backend, so we can change the behavior for SQLite.
- On SQLite, when inserting a single record, instead of generating `INSERT INTO users (name, hair_color) VALUES ('Tess', DEFAULT)`, instead generate `INSERT INTO users (name) VALUES ('Tess')`.
- On SQLite, disallow batch inserts.

We can revisit batch inserts in the future. Single inserts are the overwhelmingly common case, so let's focus on that for now.

sgrif:
Fixed by #166

PR #164, opened 2016-01-31
Disallow batch insert on SQLite

sgrif:
Related to #157. Whether this functionality is eventually added back is
out of scope for 0.5. What is clear is that we need to fix the single
record case on SQLite, and that we cannot safely insert multiple records
in a single query on SQLite.

This does not fix the single record case, and does not resolve the
issue. However, the actual solution to that will be much more complex,
and this is comparatively simple -- and will be a pre-requisite for any
solution to the actual issue.

### DIFF
--- a/diesel_tests/tests/insert.rs
+++ b/diesel_tests/tests/insert.rs
@@ -10,7 +10,7 @@ fn insert_records() {
         NewUser::new("Tess", None),
     ];
 
-    insert(new_users).into(users).execute(&connection).unwrap();
+    batch_insert(new_users, users, &connection);
     let actual_users = users.load(&connection).unwrap().collect::<Vec<User>>();
 
     let expected_users = vec![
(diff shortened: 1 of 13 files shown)


sgrif:

diesel_tests/tests/insert.rs
```diff
@@ -66,28 +66,20 @@ fn insert_with_defaults() {
 }
 
 #[test]
-#[should_panic] // FIXME: SQLite has no DEFAULT keyword. We need to work around this.
-#[cfg(feature = "sqlite")] // FIXME: This test should run on everything, the only difference is create table syntax.
+#[cfg(feature = "postgres")] // FIXME: This test should run on everything, the only difference is create table syntax.
```
sgrif: The diff here rendered in a not terribly helpful way. The change was that I renamed our existing `insert_with_defaults` test to `batch_insert_with_defaults`, and set it to run on everything except SQLite. I then changed the `insert_with_defaults` test on both backends to operate on a single record instead of a batch.

PR #165, opened 2016-02-01
Adding seed command

Virviil:
Adding simple command
```sh
$ diesel seed
```
analog from rails db:seed command.

It bootstraps database from `./migrations/seeds.sql` file.

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -11,6 +11,7 @@ use diesel::{migrations, Connection};
 use diesel::connection::PgConnection;
 use self::setup_error::SetupError;
 use std::{env, fs};
+use std::io::Read;
 use std::error::Error;
 use std::path::{PathBuf, Path};
 
@@ -50,18 +51,24 @@ fn main() {
         .about("Creates the migrations directory, creates the database \
                 specified in your DATABASE_URL, and runs existing migrations.")
         .arg(database_arg());
+        
+    let seed_subcommand = SubCommand::with_name("seed")
+        .about("Adding initial data after database is created.")
+        .arg(database_arg());
 
     let matches = App::new("diesel")
         .version(env!("CARGO_PKG_VERSION"))
         .setting(AppSettings::VersionlessSubcommands)
         .subcommand(migration_subcommand)
         .subcommand(setup_subcommand)
+        .subcommand(seed_subcommand)
         .setting(AppSettings::SubcommandRequiredElseHelp)
         .get_matches();
 
     match matches.subcommand() {
         ("migration", Some(matches)) => run_migration_command(matches),
         ("setup", Some(matches)) => run_setup_command(matches),
+        ("seed", Some(matches)) => run_seed_command(matches),
         _ => unreachable!(),
     }
 }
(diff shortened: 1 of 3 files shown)


sgrif:
So there's a couple of things here.

- I don't want to merge this until we merge #161.
  - After we merge that, this command should be `diesel database seed` not `diesel seed`
  - `diesel database setup` should call this automatically.
- The tests are failing with this change, as the current setup doesn't like `migrations/seeds.sql`. This should probably go somewhere other than the migrations directory, but I'm not sure where.

I'm unsure how I feel about this feature. Can you give some details on what the use case for this is? Is there a reason we need this to be separate from migrations?

@mfpiccolo @mcasper What are your thoughts?

Virviil:
I took the idea from here:
http://www.xyzpub.com/en/ruby-on-rails/4.0/seed_rb.html

Dont think, that seeds.sql is the best way to make the bootstrap, but migrations for now are made without orm but with sql raw syntax, so I made seeds also in row sql. This is "first" iteration, just not to start conversation about new feature without some "working" solution.

Also, I think that seeds should be called directly with `setup`, but didn't want to change `run_setup_command` code, so this is probably up to maintainers.

Also, some instrument to generate seed from existing database should be performed. Mb 
`seed export --table_name` to add/replace full table data into seed.{ext}

mcasper:
I can see the need for separate seeding functionality, in case you want to be able to populate your development environment with a bunch of fake data, but never want that to run in production (which might/will end up happening if you make it a migration).

I do agree that `migrations/seeds.(rs|sql)` feels off, maybe we should have a `db/` directory that holds migrations, seeds, and others, a la ActiveRecord? Or if we don't think we'll run into any other files we related to the database that need to exist, seeds could live at the root of the project.

Either way it should end up being `diesel database seed`, and the setup command should call it, so I'll try to push that PR through tonight

sgrif:
> Dont think, that seeds.sql is the best way to make the bootstrap

> This is "first" iteration, just not to start conversation about new feature without some "working" solution.

I think if we are going to have that conversation, we should identify the problem we're trying to solve. I don't think that "Rails does it that way" is a good enough reason.

Don't get me wrong, it's a useful tool in Rails. But we're working with a clean slate and I'd prefer to avoid extraneous features if we can avoid it. All I'm looking for is a concrete use case that is sufficiently painful without us providing this functionality out of the box.

Virviil:
> Dont think, that seeds.sql is the best way to make the bootstrap

I thought about seeds.toml file with such structure:
```toml
[[table_name]]
col1 = "val1"
...
coln = "valn"

[[table_name]]
...

[[table2_name]]
...


```

or mb seeds.rs which would `use foo::schema::*`and working with ORM types directly.

> concrete use case
Schema and data should be separated, because for testing, dev databases you can need some mock data, and for some test this data should be not random, but in deployment you dont need this mock data.

> we should have a db/ directory that holds migrations, seeds, and others, a la ActiveRecord
In ActiveRecord migration is single file, which store `up`, `down`, and `change` description. but here each migration probably should be stored in a directory, because have 2 sql files, so what the difference of the name of root dir: `migrations` or `db`? The only benefit of having `db` dir, as I understand, is having there gitignored `db.toml` file, which can automaticly give connection string for diesel_cli or even for the entire app.

mcasper:
@Virviil I suggested `db/` only because a true seeds file doesn't really have anything to do with migrations. So in rails land it's
```
- db
  - migrations
    - 12345_create_user.rb
    - 67890_create_post.rb
  - seeds.rb
```

And if we decided to follow a similar structure, it would be the same, but with migration folders instead of single migration files.

But again, I don't think I would be opposed to keeping our current `migrations/` dir and just searching for a `seeds.(sql|rs)` file at the root of the project (defined by wherever the migrations folder is) either.

sgrif:
I've given this a bit more thought, and am going to go ahead and close this. It's already easy enough to add `bin/seeds.rs` to your application without any additions to Diesel.

PR #166, opened 2016-02-01
Fix inserting default values on SQLite

### LINKED ISSUE #157: [SQLite] SQLite has no support for the `DEFAULT` keyword

sgrif:
Our previous infrastructure relied on the ability to use the `DEFAULT`
keyword for values which were not present. This keyword does not exist
on SQLite. This means that the query:

```sql
INSERT INTO users (name, hair_color) VALUES ('Tess', DEFAULT)
```

needs to instead be written as

```sql
INSERT INTO users (name) VALUES ('Tess')
```

This does not work for batch inserts, where some of the rows provide
values, but others want the defaults. There is no way to express this in
a single query using SQLite. Because of this, we disallowed batch
inserts for SQLite entirely in 10caf03dd8566977bc63f18ebc784f6b61b99eb4.
This issue will be revisted after 0.5 is released.

There were several changes included in this commit which are
tangentially related to this change.

We were able to remove the "context" concept from our query builder
entirely. This was only there to insert the `"DEFAULT"` keyword instead
of a bind param on nulls during insert. It was never something I was
super happy with, and it resulted in allocations that the compiler would
never be able to optimize away. With this change, we have a better place
to handle this.

I've had to alter codegen to specifically reference `Bound` instead of
using the `AsExpr` helper which is technically corrent. Without that
change, we get some very obscure lifetime errors that I honestly don't
fully understand. This is either a bug in Rust itself, or some *very*
obscure interaction with
https://github.com/sgrif/diesel/blob/8a33029/diesel/src/types/impls/mod.rs#L23-L29
that I do not understand (I believe it is a bug in Rust, given that it
only manifests in `where` clauses. I have been unable to isolate it in a
playground link, but I will follow up with the core team tomorrow on
this.)

The changes that were required to the `migrations` module are
unfortunate, but necessary. Even though SQLite is the *only* backend
which would not implement `SupportsDefaultKeyword`, I have no way to
represent this in the type system. As such, I can provide no evidence
that a tuple of `ColumnInsertValue`s always implements
`InsertValues<DB>` for all possible `DB`. This can go away once
specialization lands in stable.

Ultimately we can likely re-implement batch inserts for SQLite in the
future by having it execute one query per row under the hood. As has
been pointed out to me numerous times in other places, since SQLite is
an embedded databbase, there is no downside to doing 10 queries instead
of 1 (assuming we're in a transaction so there is no IO cost).

Fixes #157

### DIFF
--- a/diesel/src/query_builder/debug.rs
+++ b/diesel/src/query_builder/debug.rs
@@ -1,12 +1,11 @@
 use backend::Debug;
-use super::{QueryBuilder, BuildQueryResult, Context};
+use super::{QueryBuilder, BuildQueryResult};
 use types::HasSqlType;
 
 #[doc(hidden)]
 pub struct DebugQueryBuilder {
     pub sql: String,
     pub bind_types: Vec<u32>,
-    context_stack: Vec<Context>,
 }
 
 impl DebugQueryBuilder {
@@ -14,7 +13,6 @@ impl DebugQueryBuilder {
         DebugQueryBuilder {
             sql: String::new(),
             bind_types: Vec::new(),
-            context_stack: Vec::new(),
         }
     }
 }
@@ -31,20 +29,10 @@ impl QueryBuilder<Debug> for DebugQueryBuilder {
         Ok(())
     }
 
+    #[allow(unused_variables)]
     fn push_bound_value<T>(&mut self, bind: Option<Vec<u8>>) where
         Debug: HasSqlType<T>,
     {
-        match (self.context_stack.first(), bind) {
-            (Some(&Context::Insert), None) => self.push_sql("DEFAULT"),
-            _ => self.push_sql("?"),
-        }
-    }
-
-    fn push_context(&mut self, context: Context) {
-        self.context_stack.push(context);
-    }
-
-    fn pop_context(&mut self) {
-        self.context_stack.pop();
+        self.push_sql("?");
     }
 }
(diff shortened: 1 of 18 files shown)


sgrif:

sgrif:
@mcasper @mfpiccolo ping

diesel/src/query_builder/debug.rs
```diff
@@ -31,20 +29,10 @@ impl QueryBuilder<Debug> for DebugQueryBuilder {
         Ok(())
     }
 
+    #[allow(unused_variables)]
     fn push_bound_value<T>(&mut self, bind: Option<Vec<u8>>) where
```
mcasper: Are we keeping the `bind` variable around for clarity?

sgrif: I'm just tired of adding and removing underscores.

mcasper:
Looks good, my only comment is just for clarification

PR #168, opened 2016-02-01
Display generated migration paths as relative for diesel CLI

mcasper:
I would love some comments on how to make this function more readable, as I'm currently all up in my head about it.

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -103,16 +103,20 @@ fn run_migration_command(matches: &ArgMatches) {
             let migration_name = args.value_of("MIGRATION_NAME").unwrap();
             let timestamp = Local::now().format("%Y%m%d%H%M%S");
             let versioned_name = format!("{}_{}", &timestamp, migration_name);
-            let migration_dir = migrations::find_migrations_directory()
+            let mut migration_dir = migrations::find_migrations_directory()
                 .map_err(handle_error).unwrap().join(versioned_name);
             fs::create_dir(&migration_dir).unwrap();
 
-            // FIXME: It would be nice to print these as relative paths
+            let migration_dir_relative = convert_absolute_path_to_relative(
+                &mut migration_dir,
+                &mut env::current_dir().unwrap()
+            );
+
             let up_path = migration_dir.join("up.sql");
-            println!("Creating {}", up_path.display());
+            println!("Creating {}", migration_dir_relative.join("up.sql").display());
             fs::File::create(up_path).unwrap();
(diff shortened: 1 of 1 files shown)


diesel_cli/src/main.rs
```diff
@@ -259,6 +263,41 @@ fn connection(database_url: &str) -> PgConnection {
...
+// target path must be in the same directory or above the current path. We
+// iterate over both the target and the current path, doing nothing if the
+// path parts are the same. The first time the parts are different, we know we
+// just passed the last common ancestor, so everything left in the current
+// path needs to get converted to "../", and everything in the target path
+// gets appended after the up path.
+fn convert_absolute_path_to_relative(target_path: &PathBuf, current_path: &PathBuf)
+    -> PathBuf
+{
+    let mut target_path_iter = target_path.iter();
```
sgrif: Maybe use https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.relative_from here?

mcasper: That could be used to generate the "down" part of the path after we've figured out the common ancestors, yeah. I think it would just replace line 294-296

sgrif: Two questions:

- Does this become significantly easier implementation-wise if we make the paths relative to the project root and not `$CWD`?
- Do we actually care about it being relative to `$CWD` instead of project root?

sgrif: Never mind, the definition of "project root" is too ambiguous.

sgrif: Could this implementation be simplified to something like:

```rust
let mut result = PathBuf::new();

while !target_path.starts_with(current_path) {
    result.push("..");
    current_path = current_path.parent();
}

result.join(target_path.relative_from(current_path))
```

mcasper: According to the tests, yes! But the only way to use `relative_from`/`strip_prefix` is to be on nightly :/

sgrif: Ugh. Let's just copy paste the body of it as a bare function, with a `FIXME: Remove when 1.7 is stable`

mcasper:
Had to import `iter_after` as well, and lost the `StripPrefixError`. I figured that rather than continuing down the rabbit hole and adding our own error type to fill the gap, we can just `unwrap()`, as our previous checks should guarantee that it won't fail.

sgrif:
Yeah, that's fine for now.

sgrif:
This is literally just a shim for like 4 weeks

Issue #167, opened 2016-02-01
[SQLite] Travis is failing for... reasons...


sgrif:
### Note: PRs addressing this issue should target the `diesel-sqlite-support` branch, not `master`.

This has been absurdly painful to debug. We need to fix it before we can merge into master, but I really am at my wits end about this. We've had to change our build infrastructure, because travis doesn't have a "not fucking ancient" option for the SQLite version. By default they ship a version from 2011, and we rely on `sqlite3_errstr`, which was added in 2012.

The new infrastructure *should* be using a more recent version, but for some reason we fail to compile when we get to codgen. There is absolutely no reason I can think of that this would fail at codegen, but the main crate would continue to compile.

We need to figure out some way to get the build green, either by getting travis working with the code as is, or by removing our dependency on `sqlite3_errstr` and switching back to the container infrastructure.

sgrif:
Oh also, somehow nightly on PG passes even though the failure is at compile time not runtime. Because... yeah... reasons...

SoftwareWarlock:
Not sure if you came across this in your searching, but apparently Travis CI provided a [workaround](https://launchpad.net/~travis-ci/+archive/ubuntu/sqlite3) for upgrading sqlite3, but unfortunately it only seems to work on their legacy infrastructure.

SoftwareWarlock:
Actually, scratch that. I was able to get the build to pass by manually installing the work around and setting the distro to [precise instead of trusty](https://github.com/SoftwareWarlock/diesel/blob/diesel-sqlite-support/.travis.yml#L2).

sgrif:
I'd prefer to avoid using apt-get for this if we can. Trusty *should* ship
with 3.8 by default.

SoftwareWarlock:
Agreed. Maybe Travis is manually changing the version they have installed? Would you be willing to consider switching to a different CI system? I've had a pretty pleasant experience using CircleCI.

SoftwareWarlock:
Just did a sanity check, and the version really is 3.8, so maybe the codegen is compatible with 3.7.15, but not 3.7.9 nor 3.8?

sgrif:
It's definitely compatible with 3.8. If you look at the failure message
appearing on Travis, it's due to `sqlite3_errstr` not being found, which
means it's an ancient 3.7 version. Also note that the problem isn't that
*codegen* is failing to compile, but that Diesel as a dependency of codegen
is failing

sgrif:
[...]

PR #170, opened 2016-02-01
Fix failing build due to old sqlite3 version

SoftwareWarlock:
It seems that the original post in #167 was correct, the sqlite support branch was failing due to an ancient version of sqlite.

Tavis CI has provided a workaround for this, but it seems to only work on Ubuntu Precise. I haven't been able to tell for sure if this slows down the build, but it doesn't seem like it at the moment. One foreseeable issue is that this prevents moving to Travis' new container infrastructure, which offers features like caching.

Any thoughts on possible alternatives?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: rust
-dist: trusty
+dist: precise
 rust:
   - stable
   - beta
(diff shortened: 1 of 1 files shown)


sgrif:
Closing in favor of a9117a3.

PR #171, opened 2016-02-01
Allow backend agnostic table creation in tests

### LINKED ISSUE #158: [SQLite] Add a backend agnostic way to create a table

We have a lot of tests which are duplicated based on the backend, simply because there's no backend agnostic way to define an auto incrementing primary key. Let's just bite the bullet and add an (internal only) DSL for this.

sgrif:
I have not added this DSL to the public API, as it's unsuitable for
public use at this time. It handles nothing other than the exact cases
that we need in our test suite, and requires you to manually quote
default values.

Fixes #158

### DIFF
--- a/diesel_tests/tests/insert.rs
+++ b/diesel_tests/tests/insert.rs
@@ -66,68 +66,21 @@ fn batch_insert_with_defaults() {
 }
 
 #[test]
-#[cfg(feature = "postgres")] // FIXME: This test should run on everything, the only difference is create table syntax.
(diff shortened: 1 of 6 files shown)


sgrif:
Review? @mfpiccolo @mcasper

mfpiccolo:
This looks good to me.  Is this a preliminary DSL for migrations or is this only going to live in test?

sgrif:
Only in tests for now. It might eventually turn into something that becomes more general.

PR #172, opened 2016-02-02
Run significantly more of our test suite against SQLite

### LINKED ISSUE #155: [SQLite] Our tests are compiling against a PG database, even when run against SQLite

The short version for this is because codegen is tightly coupled to PG. We either need to decouple it, or stop using `infer_schema!` when our tests are used w/ SQLite.

sgrif:
This removes the usage of the `infer_schema!` macro, which was
previously compiling against PG. We may eventually be able to use it
again on SQLite, but I think making that macro backend agnostic is out
of scope for 0.5. The `Post` struct on SQLite also does not contain the
`tags` field, which would make no sense on SQLite.

I've opted to leave the `annotations` module out of SQLite, as it'll be
a pain to stop using batch insert and returning there, and really it's
only testing that codegen generates valid Rust in various edge cases,
which isn't something that would change for SQLite.

Fixes #155 (more or less)

### DIFF
--- a/diesel_tests/tests/sqlite_specific_schema.rs
+++ b/diesel_tests/tests/sqlite_specific_schema.rs
@@ -1,0 +1,64 @@
+use diesel::*;
+use super::User;
+
+//FIXME: We can go back to `infer_schema!` when codegen becomes generic
+table! {
+    users {
+        id -> Integer,
+        name -> VarChar,
+        hair_color -> Nullable<VarChar>,
+    }
+}
+
+table! {
+    posts {
+        id -> Integer,
+        user_id -> Integer,
+        title -> VarChar,
+        body -> Nullable<Text>,
+    }
+}
+
+table! {
+    comments {
+        id -> Integer,
+        post_id -> Integer,
+        text -> Text,
+    }
+}
+
+table! {
+    special_posts {
+        id -> Integer,
+        user_id -> Integer,
+        title -> VarChar,
+    }
+}
+
+table! {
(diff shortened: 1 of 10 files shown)


sgrif:
Review? @mcasper @mfpiccolo

diesel_tests/tests/sqlite_specific_schema.rs
```diff
@@ -0,0 +1,64 @@
+use diesel::*;
+use super::User;
+
+//FIXME: We can go back to `infer_schema!` when codegen becomes generic
+table! {
+    users {
+        id -> Integer,
```
mcasper: Should these ids all be type `Serial`?

sgrif: Serial is just an alias right now. If we ever turned it back into a proper type, it would be invalid on SQLite as it's the PG specific name for an auto incrementing integer.

mcasper: Gotcha, didn't know if we were going to be using it to represent any auto incrementing integer

mcasper:
Nice work! Looks good

mcasper:
How do we plan on divvying up the test suite as we add specific support for more databases? Whatever the most recent database is gets a bunch of tests converted and thrown at it? Or do you think we'll eventually want to build out full test suites for each one we specifically support?

sgrif:
I have no clue. We'll do whatever is required as specific needs come up. :)

PR #175, opened 2016-02-02
Remove hard dependencies on `libpq` and `libsqlite3`

### LINKED ISSUE #159: [SQLite] Diesel should not have a hard dependency on libpq and libsqlite

Right now we currently always compile against both. We should only compile against whatever the user wants to use. They should both be able to co-exist in the same app if desired. This can either be done via features, or by moving SQLite and PG into separate crates. I have no preference as to which solution we do, but we should isolate the backend specific things to a single top level module in either case so we don't end up littering `#[cfg(feature="whatever")]` into our code.

sgrif:
All code which *directly* depends on these has been moved to a separate
module and placed behind a cfg attr. There is still a ton of PG specific
stuff hidden in the expressions module that simply isn't marked as such.

I had to rework how we do our boilerplate type impls, as `sqlite` and
`postgres` need to be able to co-exist in the same compile pass (even
just for generating docs).

While this PR is pretty large, it ultimately mostly comes down to
clerical work.

Fixes #159.


sgrif:
Review? @mcasper @mfpiccolo

mcasper:
I've started and stopped this a few times and made it most of the way through, but what I've read looks good to me. Would love someone else to read through as well

sgrif:
Well I guess that leaves it to @mcasper

sgrif:
Er @mfpiccolo rather

mfpiccolo:
:eyes:

diesel/src/pg/types/integers.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl ToSql<types::Oid, Pg> for u32 {
+    fn to_sql<W: Write>(&self, out: &mut W) -> Result<IsNull, Box<Error>> {
+        out.write_u32::<BigEndian>(*self)
+            .map(|_| IsNull::No)
+            .map_err(|e| e.into())
+    }
+}
+
+#[test]
```
mfpiccolo: Do we want these in this file?

mfpiccolo:
That was beefy.  Looks good.  :+1:

PR #178, opened 2016-02-02
Add helpers for managing `updated_at TIMESTAMP` columns

### LINKED ISSUE #91: Add helpers for dealing with timestamps
`created_at` doesn't need anything from us, we can just have `DEFAULT current_timestamp`. `updated_at` requires a trigger, but we shouldn't require our users to write that. We can use the same procedure for pretty much every table:

```
CREATE FUNCTION set_updated_at() RETURNS trigger AS $$
BEGIN
    IF (
        NEW IS DISTINCT FROM OLD AND
        NEW.updated_at IS NOT DISTINCT FROM OLD.updated_at
    ) THEN
        NEW.updated_at := current_timestamp;
    END IF;
    RETURN NEW;
END
$$ LANGUAGE plpgsql;
```

We should provide that so users don't have to write it repeatedly

robertmaloney:
Fixes #91 

###### TODO
* [x] Migrate functions setup around `__diesel_schema_migrations` setup
* [x] Remove CLI

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -93,4 +93,6 @@ pub trait Connection: SimpleConnection + Sized {
     #[doc(hidden)] fn rollback_transaction(&self) -> QueryResult<()>;
     #[doc(hidden)] fn commit_transaction(&self) -> QueryResult<()>;
     #[doc(hidden)] fn get_transaction_depth(&self) -> i32;
+
+    #[doc(hidden)] fn setup_helper_functions(&self);
 }
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -114,6 +114,12 @@ impl Connection for PgConnection {
     fn get_transaction_depth(&self) -> i32 {
         self.transaction_depth.get()
     }
+
+    fn setup_helper_functions(&self) {
+        self.batch_execute(
+            include_str!("setup/timestamp_helpers.sql")
+        ).expect("Error creating timestamp helper functions for Pg");
+    }
 }
 
 impl PgConnection {
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -107,6 +107,10 @@ impl Connection for SqliteConnection {
     fn get_transaction_depth(&self) -> i32 {
         self.transaction_depth.get()
     }
+
+    fn setup_helper_functions(&self) {
+        // this will be implemented at least when timestamps are supported in SQLite
+    }
 }
 
 impl SqliteConnection {
--- a/diesel_tests/tests/connection.rs
+++ b/diesel_tests/tests/connection.rs
@@ -1,0 +1,49 @@
+use schema::connection_without_transaction;
+use diesel::*;
+use diesel::expression::dsl::sql;
+
+table! {
+    auto_time {
+        id -> Integer,
+        n -> Integer,
+        updated_at -> Timestamp,
+    }
+}
+
+#[test]
+#[cfg(feature = "postgres")]
+fn managing_updated_at_for_table() {
+    use self::auto_time::columns::*;
+    use self::auto_time::table as auto_time;
+    use diesel::pg::types::date_and_time::PgTimestamp;
+
+    // transactions have frozen time, so we can't use them
+    let connection = connection_without_transaction();
+    connection.execute("CREATE TABLE auto_time (
+        id SERIAL PRIMARY KEY,
+        n INTEGER,
+        updated_at TIMESTAMP
+    );").unwrap();
+    connection.execute("SELECT diesel_manage_updated_at('auto_time');").unwrap();
+
+    connection.execute("INSERT INTO auto_time (n) VALUES (2), (1), (5);").unwrap();
+    let result = connection.query_one::<_, i64>(
+        select(sql("COUNT(*) FROM auto_time WHERE updated_at IS NULL"))
+    );
+    assert_eq!(Ok(3), result);
+
+    connection.execute("UPDATE auto_time SET n = n + 1 WHERE true;").unwrap();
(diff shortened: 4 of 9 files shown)


diesel/src/connection/mod.rs
```diff
@@ -97,4 +97,6 @@ pub trait Connection: SimpleConnection + Sized {
     #[doc(hidden)] fn rollback_transaction(&self) -> QueryResult<()>;
     #[doc(hidden)] fn commit_transaction(&self) -> QueryResult<()>;
     #[doc(hidden)] fn get_transaction_depth(&self) -> i32;
+    #[doc(hidden)] fn manage_updated_at_for_table(&self, &str) -> QueryResult<()>;
+    #[doc(hidden)] fn neglect_updated_at_for_table(&self, &str) -> QueryResult<()>;
```
sgrif: This name is hilarious

sgrif: It's probably not the most self-explanatory name if it doesn't appear next to the `manage` form. I'm actually not sure why these are Rust functions. Hopefully it'll become clear as I read on.

sgrif:
I'm a bit confused why we added the timestamp stuff to CLI. What I was imagining here is that we'd automatically create this function at the same point that we create the `__diesel_schema_migrations` table (which I need to better generalize to "general metadata management" before we can merge this), and then there was a SQL function you could call with a table name for no reason other than that I can never remember the trigger syntax. If we can't do the second half, I think we should just drop that and just provide this trigger automatically with no other behavior.

diesel_cli/src/main.rs
```diff
@@ -122,6 +139,19 @@ fn run_migration_command(matches: &ArgMatches) {
             println!("Creating {}", migration_dir_relative.join("down.sql").display());
             fs::File::create(down_path).unwrap();
         }
+        ("manage-timestamps", Some(args)) => {
+            let table_name = args.value_of("TABLE_NAME").unwrap();
+            let connection = connection(&database_url(args));
+            match connection.manage_updated_at_for_table(&table_name) {
```
mcasper: Could this be written as `connection.manage_updated_at_for_table(&table_name).unwrap_or_else(handle_error)`?

robertmaloney:
Ah, yeah I was overthinking it. In thinking of a better name, what are your thoughts on `enable/disable_automatic_updated_at_for_table`.

sgrif:
I think `diesel_manage_updated_at` is fine.

sgrif:
Just a heads up, this looks like it still has a good bit of work left to do. At this point I'm calling the cutoff point for 0.5 features. This can make it into 0.6 once it's updated.

robertmaloney:
Thats fine, I'm holding off on this anyways until the DB setup flow is refactored. Thank you for letting me know.

diesel/src/pg/connection/mod.rs
```diff
@@ -114,6 +114,30 @@ impl Connection for PgConnection {
...
+            END;
+            $$ LANGUAGE plpgsql;"
+        ).expect("Error creating function `diesel_set_updated_at()`");
+        self.execute(
+            "CREATE OR REPLACE FUNCTION diesel_manage_updated_at(_tbl regclass) RETURNS VOID AS $$
+            BEGIN
+                EXECUTE format('CREATE TRIGGER set_updated_at BEFORE UPDATE ON %s
+                FOR EACH ROW EXECUTE PROCEDURE diesel_set_updated_at()', _tbl);
+            END;
+            $$ LANGUAGE plpgsql;"
```
sgrif: Two requests on this. First, can you put this in a `.sql` file and use `include_str!` here? Second, can we do this in a single query with `batch_execute` or whatever the function is called? The one on `SimpleConnection`

diesel/src/sqlite/connection/mod.rs
```diff
@@ -107,6 +107,10 @@ impl Connection for SqliteConnection {
     fn get_transaction_depth(&self) -> i32 {
         self.transaction_depth.get()
     }
+
+    fn setup_helper_functions(&self) {
+        // this will be implemented at least when timestamps are supported in SQLite
```
sgrif: Does this need to be on connection given that it's PG specific right now? Should this at least have `unimplemented!()`?

robertmaloney: Implementing in `Connection` will have to happen eventually, so I figured I might as well add the skeleton for it. I left out the `unimplemented!()` just because it would crash whenever someone wanted to setup their database in SQLite. Would you rather I make a temporary function in `migrations` that's called only when the `postgres` feature is applied?

sgrif: No, this is fine.

diesel/src/pg/connection/mod.rs
robertmaloney: Done. I put the sql in a `setup` folder.

diesel_tests/tests/connection.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[test]
+#[cfg(feature = "postgres")]
+fn managing_updated_at_for_table() {
+    let connection = connection();
+    connection.execute("CREATE TABLE yestest (n INTEGER, updated_at TIMESTAMP);").unwrap();
+    connection.execute("SELECT diesel_manage_updated_at('yestest');").unwrap();
+
+    connection.execute("INSERT INTO yestest (n) VALUES (2), (1), (5);").unwrap();
+    connection.execute("UPDATE yestest SET n = n + 1 WHERE true;").unwrap();
```
sgrif: While I know we can't test the exact value, can you test that this is setting the timestamp to a later time than before?

sgrif:
This is perfect. One last comment. This also needs a changelog entry, and docs.

sgrif:
@robertmaloney `SystemTime` is unstable. You need to write your test against `PgTimestamp`

robertmaloney:
Fixed :+1:

PR #181, opened 2016-02-03
Add a new getting started guide

sgrif:
[Rendered](https://github.com/sgrif/diesel/blob/sg-new-guide/guide.md)

This is not actually meant for this repo, this will end up on the new
website. However, I wanted to write out all the content in markdown and
review it here before I go through all the formatting, etc over there.

After the website is live, the README will get slimmed down by quite a
lot, mostly pointing to this guide. One thing that I couldn't figure out
how to fit in naturally was passing a struct to `update().set`, or using
methods like `save_changes`. Maybe it's fine to skip them, as this isn't
intended to be an exhaustive guide?

I also want to follow this up with an "advanced" guide that covers joins
and associations in 0.6.

### DIFF
--- a/guide.md
+++ b/guide.md
@@ -1,0 +1,388 @@
+Getting Started 2.0
+===
+
+For this guide, we're going to walk through some simple examples for each of the
+pieces of CRUD. Each step in this guide will build on the previous, and is meant
+to be followed along. This guide assumes that you're using PostgreSQL on
+nightly. We'll talk about how to use Diesel on stable in the final chapter.
+Before we start, make sure you have PostgreSQL installed and running.
+
+The first thing we need to do is generate our project.
+
+```shell
+cargo new diesel_demo
+cd diesel_demo
+```
+
+We're going to use a [`.env`][dotenv-rust] to manage our environment variables
+for us, and [`diesel_cli`][diesel-cli] to help set up the database.
+
+[dotenv-rust]: https://github.com/slapresta/rust-dotenv
+[diesel-cli]: https://github.com/sgrif/diesel/diesel_cli
+
+```shell
+echo DATABASE_URL=postgres://localhost/diesel_demo > .env
+cargo install diesel_cli
+diesel setup
+```
+
+This will create our database for us (if it didn't already exist), and create an
+empty migrations directory that we can use to manage our schema. More on that
+later. Let's add Diesel to our `Cargo.toml`. We'll also add `dotenv` so we can
+use that same `.env` file for our environment variables.
+
+```toml
+[dependencies]
+diesel = "^0.5.0"
+diesel_codegen = { version = "^0.5.0", default-features = false, features = ["nightly", "postgres"] }
+dotenv = "^0.8.0"
+dotenv_macros = "^0.8.0"
+```
+
+We'll also add this to the top of `src/lib.rs`, which will allow us to use
+features from `diesel_codegen`.
+
+```rust
+#![feature(custom_derive, custom_attribute, plugin)]
+#![plugin(diesel_codegen, dotenv_macros)]
+```
+
+Now we're going to write a small CLI that lets us manage a blog (ignoring the
+fact that we can only access the database from this CLI...). The first thing
+we're going to need is a table to store our posts. Let's create a migration for
+that:
+
+```shell
+diesel migration generate create_posts_table
+```
+
+You should see output which looks something like this:
+
+```
+Creating migrations/20160202154039_create_posts/up.sql
+Creating migrations/20160202154039_create_posts/down.sql
+```
+
+Every migration consists of a way to run it, and a way to undo it. `down.sql`
+should always leave your database in the state it was in before running `up.sql`
+when possible. We'll edit our migration to contain the following:
+
+```sql
+-- up.sql
+CREATE TABLE posts (
+  id SERIAL PRIMARY KEY,
+  title VARCHAR NOT NULL,
+  body TEXT NOT NULL,
+  published BOOLEAN NOT NULL DEFAULT 'f'
+)
+
+-- down.sql
+DROP TABLE posts
+```
+
+We can run our new migration with `diesel migration run`. It's usually a good
+idea to make sure that `down.sql` is correct. You can do a quick sanity check by
+doing `diesel migration redo`.
+
+OK enough SQL, let's write some Rust. We'll start by writing some code to show
+us the last 5 published posts. The first thing we need to do is establish a
+database connection.
+
+```rust
+// src/lib.rs
+#[macro_use]
+extern crate diesel;
+extern crate dotenv;
+
+use diesel::prelude::*;
+use diesel::pg::PgConnection;
+use dotenv::dotenv;
+use std::env;
+
+pub fn establish_connection() -> PgConnection {
+    dotenv().ok();
+
+    let database_url = env::var("DATABASE_URL")
+        .expect("DATABASE_URL must be set");
+    PgConnection::establish(&database_url)
+        .expect(&format!("Error connecting to {}", database_url))
+}
+```
+
+We'll also want to create a `Post` struct that we can read our data into, and
+have diesel generate the names we'll use to reference our tables and columns in
+our queries.
+
+We'll add the following two lines to `src/lib.rs`
+
+```rust
+pub mod schema;
+pub mod models;
+```
+
+and then create those modules.
+
+```rust
+// src/models.rs
+#[derive(Queryable)]
+pub struct Post {
+    pub id: i32,
+    pub title: String,
+    pub body: String,
+    pub published: bool,
+}
+
+// src/schema.rs
+infer_schema!(dotenv!("DATABASE_URL"));
+```
+
+The `#[derive(Queryable)]` will generate all of the code needed to load a `Post`
+struct from a SQL query. The `infer_schema!` macro connects to the database URL
+given to it, and creates a bunch of code based on the database schema to
+represent all of the tables and columns. We'll see exactly what that looks like
+next. Let's write the code to actually show us our posts.
+
+```rust
+// src/bin/show_posts.rs
+extern crate diesel_demo;
+extern crate diesel;
+
+use self::diesel_demo::*;
+use self::diesel_demo::models::*;
+use self::diesel::prelude::*;
+
+fn main() {
+    use diesel_demo::schema::posts::dsl::*;
+
+    let connection = establish_connection();
+    let results = posts.filter(published.eq(true))
+        .limit(5)
+        .load(&connection)
+        .expect("Error loading posts")
+        .collect::<Vec<Post>>();
+
+    println!("Displaying {} posts", results.len());
+    for post in results {
+        println!("{}", post.title);
+        println!("----------\n");
+        println!("{}", post.body);
+    }
+}
+```
+
+The `use posts::dsl::*` line imports a bunch of aliases so that we can say
+`posts` instead of `posts::table`, and `published` instead of
+`posts::published`. It's useful when we're only dealing with a single table, but
+that's not always what we want.
+
+We can run our script with `cargo run --bin show_posts`. Unfortunately, the results
+won't be terribly interesting, as we don't actually have any posts in the
+database. Still, we've written a decent amount of code, so let's commit.
+
+The full code for the demo at this point can be found [here][commit-no-1].
+
+Next, let's write some code to create a new post. We'll want a struct to use for
+inserting a new record.
+
+```rust
+// src/models.rs
+use super::schema::posts;
+
+#[insertable_into(posts)]
+pub struct NewPost<'a> {
+    pub title: &'a str,
+    pub body: &'a str,
+}
+```
+
+Now let's add a function to save a new post.
+
+```rust
+// src/lib.rs
+use self::models::{Post, NewPost};
+
+pub fn create_post<'a>(conn: &PgConnection, title: &'a str, body: &'a str) -> Post {
+    use schema::posts;
+
+    let new_post = NewPost {
+        title: title,
+        body: body,
+    };
+
+    diesel::insert(&new_post).into(posts::table)
+        .get_result(conn)
+        .expect("Error saving new post")
+}
+```
+
+When we call `.get_result` on an insert or update statement, it automatically
+adds `RETURNING *` to the end of the query, and lets us load it into any struct
+that implements `Queryable` for the right types. Neat!
+
+#### Protip (This would be a sidebar or something)
+
+Diesel can insert more than one record in a single query. Just pass a `Vec` or
+slice to `insert`, and then call `get_results` instead of `get_result`. If you
+don't actually want to do anything with the row that was just inserted, call
+`.execute` instead. The compiler won't complain at you, that way. :)
+
+Now that we've got everything set up, we can create a little script to write a
+new post.
+
+```rust
+// src/bin/write_post.rs
+extern crate diesel_demo;
+extern crate diesel;
+
+use self::diesel_demo::*;
+use std::io::{stdin, Read};
+
+fn main() {
+    let connection = establish_connection();
+
+    println!("What would you like your title to be?");
+    let mut title = String::new();
+    stdin().read_line(&mut title).unwrap();
+    let title = &title[..(title.len() - 1)]; // Drop the newline character
+    println!("\nOk! Let's write {} (Press {} when finished)\n", title, EOF);
+    let mut body = String::new();
+    stdin().read_to_string(&mut body).unwrap();
+
+    let post = create_post(&connection, title, &body);
+    println!("\nSaved draft {} with id {}", title, post.id);
+}
+
+#[cfg(not(windows))]
+const EOF: &'static str = "CTRL+D";
+
+#[cfg(windows)]
+const EOF: &'static str = "CTRL+Z";
+```
+
+We can run our new script with `cargo run --bin write_post`. Go ahead and write
+a blog post. Get creative! Here was mine:
+
+```
+   Compiling diesel_demo v0.1.0 (file:///Users/sean/Documents/Projects/open-source/diesel_demo)
+     Running `target/debug/write_post`
+
+What would you like your title to be?
+Diesel demo
+
+Ok! Let's write Diesel demo (Press CTRL+D when finished)
+
+You know, a CLI application probably isn't the best interface for a blog demo.
+But really I just wanted a semi-simple example, where I could focus on Diesel.
+I didn't want to get bogged down in some web framework here.
+Plus I don't really like the Rust web frameworks out there. We might make a
+new one, soon.
+
+Saved draft Diesel demo with id 1
+```
+
+Unfortunately, running `show_posts` still won't display our new post, because we
+saved it as a draft. We need to publish it! But in order to do that, we'll need
+to look at how to update an existing record. First, let's commit. The code for
+this demo at this point can be found [here][commit-no-2].
+
+Now that we've got create and read out of the way, update is actually relatively
+simple. Let's jump write into the script:
+
+```rust
+// src/bin/publish_post.rs
+extern crate diesel_demo;
+extern crate diesel;
+
+use self::diesel::prelude::*;
+use self::diesel_demo::*;
+use self::diesel_demo::models::Post;
+use std::env::args;
+
+fn main() {
+    use diesel_demo::schema::posts::dsl::{posts, published};
+
+    let id = args().nth(1).expect("publish_post requires a post id")
+        .parse::<i32>().expect("Invalid ID");
+    let connection = establish_connection();
+
+    let post = diesel::update(posts.find(id))
+        .set(published.eq(true))
+        .get_result::<Post>(&connection)
+        .expect(&format!("Unable to find post {}", id));
+    println!("Published post {}", post.title);
+}
+```
+
+That's it! Let's try it out with `cargo run --bin publish_post 1`.
+
+```
+   Compiling diesel_demo v0.1.0 (file:///Users/sean/Documents/Projects/open-source/diesel_demo)
+     Running `target/debug/publish_post 1`
+Published post Diesel demo
+```
+
+And now, finally, we can see our post with `cargo run --bin show_posts`.
+
+```
+     Running `target/debug/show_posts`
+Displaying 1 posts
+Diesel demo
+----------
+
+You know, a CLI application probably isn't the best interface for a blog demo.
+But really I just wanted a semi-simple example, where I could focus on Diesel.
+I didn't want to get bogged down in some web framework here.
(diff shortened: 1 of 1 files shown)


sgrif:
Review? @samphippen @mfpiccolo @mcasper

sgrif:
Also @derekprior @tessgriffin @steveklabnik will likely have feedback on this. Please rip it to shreds. <3

guide.md
```diff
@@ -0,0 +1,388 @@
+Getting Started 2.0
+===
+
+For this guide, we're going to walk through some simple examples for each of the
+pieces of CRUD. Each step in this guide will build on the previous, and is meant
+to be followed along. This guide assumes that you're using PostgreSQL on
+nightly. We'll talk about how to use Diesel on stable in the final chapter.
```
derekprior: "PostgreSQL and Rust nightly" or something similar? At first read I thought "nightly" applied to Postgres.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+Before we start, make sure you have PostgreSQL installed and running.
```
derekprior: If this needs to be said then does it maybe deserve a link?

guide.md
```diff
@@ -0,0 +1,388 @@
...
+
+The first thing we need to do is generate our project.
+
+```shell
+cargo new diesel_demo
+cd diesel_demo
+```
+
+We're going to use a [`.env`][dotenv-rust] to manage our environment variables
```
derekprior: "a `.env` file..."

mcasper: Do you pronounce it `dotenv` or `env`?

guide.md
sgrif: Do you think it needs to be said?

guide.md
mcasper: Ignore me. Misread the comment

sgrif: "a dotenv file"

guide.md
```diff
@@ -0,0 +1,388 @@
...
+for us, and [`diesel_cli`][diesel-cli] to help set up the database.
+
+[dotenv-rust]: https://github.com/slapresta/rust-dotenv
+[diesel-cli]: https://github.com/sgrif/diesel/diesel_cli
+
+```shell
+echo DATABASE_URL=postgres://localhost/diesel_demo > .env
+cargo install diesel_cli
```
mcasper: The user also has to add `$CARGO_PATH/bin` to their PATH, worth mentioning? Or leave that to learning rust and cargo?

guide.md
```diff
@@ -0,0 +1,388 @@
...
+diesel setup
+```
+
+This will create our database for us (if it didn't already exist), and create an
+empty migrations directory that we can use to manage our schema. More on that
+later. Let's add Diesel to our `Cargo.toml`. We'll also add `dotenv` so we can
```
derekprior: The ordering of this all took me a bit by surprise because we did something related to `dotenv` already (albeit, we just added it's file) and only now are we declaring the dependency.

I gather that's possibly because `diesel_cli` depends on .env internally?

guide.md
```diff
@@ -0,0 +1,388 @@
...
+    stdin().read_to_string(&mut body).unwrap();
+
+    let post = create_post(&connection, title, &body);
+    println!("\nSaved draft {} with id {}", title, post.id);
+}
+
+#[cfg(not(windows))]
+const EOF: &'static str = "CTRL+D";
+
+#[cfg(windows)]
```
sgrif: Yes, I'm being "that guy"

guide.md
```diff
@@ -0,0 +1,388 @@
...
+
+```
+Creating migrations/20160202154039_create_posts/up.sql
+Creating migrations/20160202154039_create_posts/down.sql
+```
+
+Every migration consists of a way to run it, and a way to undo it. `down.sql`
+should always leave your database in the state it was in before running `up.sql`
+when possible. We'll edit our migration to contain the following:
+
```
tessgriffin: When you run the diesel migration, does it create just one file or two? Lines 62 and 63 imply 2 to me. If that's the case, you should say on line 68 "We'll edit our migrations".

guide.md
sgrif: Yes, `diesel_cli` depends directly on `dotenv`. Since it's a separate crate, there's no reason for it not to

derekprior: Having read a bit more, I'd consider re-ordering everything after generating the project. Maybe:

1. set up your project dependencies in the `cargo.toml` file.
2. introduce diesel cli. What's it for? Why isn't it a project dependency? How do I install it?
3. Use .env and diesel cli to create our database.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+```shell
+cargo new diesel_demo
+cd diesel_demo
+```
+
+We're going to use a [`.env`][dotenv-rust] to manage our environment variables
+for us, and [`diesel_cli`][diesel-cli] to help set up the database.
+
+[dotenv-rust]: https://github.com/slapresta/rust-dotenv
+[diesel-cli]: https://github.com/sgrif/diesel/diesel_cli
```
derekprior: This link does not work

guide.md
sgrif: It creates two files. However, I think that `up.sql` and `down.sql` can be collectively referred to as "a migration". 2 files != 2 migrations.

guide.md
sgrif: I opted to leave it out, as cargo outputs this for me:

```
be sure to add `/Users/sean/.multirust/toolchains/nightly/cargo/bin` to your PATH to be able to run the installed binaries
```

I'm fine with calling it out if you think we should, though.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+```toml
+[dependencies]
+diesel = "^0.5.0"
+diesel_codegen = { version = "^0.5.0", default-features = false, features = ["nightly", "postgres"] }
+dotenv = "^0.8.0"
+dotenv_macros = "^0.8.0"
+```
+
+We'll also add this to the top of `src/lib.rs`, which will allow us to use
+features from `diesel_codegen`.
```
derekprior: Can you give a quick summary of what `diesel_codegen` does or at least let me know we're going to cover that later if indeed we do?

sgrif: I'm open to suggestions on this one. codegen being a separate crate might be fairly alien to devs from other languages. I don't think I ever really cover what it does, and I don't have a great "quick summary" other than "it generates code".

sgrif: I could maybe say something like "which gives us several useful class annotations, as well as the `infer_schema!` macro. We'll see both of those soon."

guide.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+You should see output which looks something like this:
+
+```
+Creating migrations/20160202154039_create_posts/up.sql
+Creating migrations/20160202154039_create_posts/down.sql
+```
+
+Every migration consists of a way to run it, and a way to undo it. `down.sql`
```
derekprior: Maybe something like: "Migrations allow us to evolve the database schema over time. Each migration can be applied (`up.sql`) or rolled-back (`down.sql`). Applying and immediately rolling-back a migration should leave your database schema unchanged."

My chief worry was the term `run`. You `run` your down migration too, really...

sgrif: Good point. I picked `run` because that's the CLI command, but I like your description better (though we say `revert` instead of `rollback`)

guide.md
tessgriffin: I think it would make sense then to visually separate the two. Even just a white line. Something to indicate that there are two files (even though yes, it's one migration)

guide.md
```diff
@@ -0,0 +1,388 @@
...
+  published BOOLEAN NOT NULL DEFAULT 'f'
+)
+
+-- down.sql
+DROP TABLE posts
+```
+
+We can run our new migration with `diesel migration run`. It's usually a good
+idea to make sure that `down.sql` is correct. You can do a quick sanity check by
+doing `diesel migration redo`.
```
derekprior: which will roll back and then apply your migration again.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+  title VARCHAR NOT NULL,
+  body TEXT NOT NULL,
+  published BOOLEAN NOT NULL DEFAULT 'f'
+)
+
+-- down.sql
+DROP TABLE posts
+```
+
+We can run our new migration with `diesel migration run`. It's usually a good
```
derekprior: `run` troubles me here for the previously specified reason. Maybe I'm overthinking it.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+database connection.
+
+```rust
+// src/lib.rs
+#[macro_use]
+extern crate diesel;
+extern crate dotenv;
+
+use diesel::prelude::*;
+use diesel::pg::PgConnection;
```
mcasper: A few things here, most of which I assume were intentional and will be true in 0.5
* PgConnection lives in `diesel::connection::pg::PgConnection` right now
* "postgres" isn't currently a valid feature (listed above in the Cargo.toml)
* `diesel::Connection` has to be imported for the `Connection.establish` to be available

Again, I assume most of these things will be true by 0.5, just bringing them up

guide.md
sgrif: Yes, they will be in separate panes on the website.

guide.md
sgrif: This code all runs when used on the code as of https://github.com/sgrif/diesel/pull/175 (which I'm still waiting for a review on. :P)

guide.md
sgrif: I don't have strong feelings about it, though `apply` feels weird to me. @mfpiccolo @mcasper do either of you have an opinion here?

sgrif: Wait are you saying "We can run our new migration", or are you saying you think we should rename the CLI command `diesel migration run`?

guide.md
```diff
@@ -0,0 +1,388 @@
...
+new one, soon.
+
+Saved draft Diesel demo with id 1
+```
+
+Unfortunately, running `show_posts` still won't display our new post, because we
+saved it as a draft. We need to publish it! But in order to do that, we'll need
+to look at how to update an existing record. First, let's commit. The code for
+this demo at this point can be found [here][commit-no-2].
+
```
tessgriffin: So at this point, we've saved a "draft"? So is it not actually written to the database yet? The next section we create the update functionality. So does it only get saved to Postgresql at that point?

sgrif: It's in the database with `published = false`. Maybe I should expand on that more?

tessgriffin: Yeah, I think so. I can't tell if the published part is cheeky because you're working with posts, or if that's an actual thing? I'm just confused as to where this is actually being saved to the database

sgrif: I guess I assumed that it was implied that "drafts" are still saved somewhere

tessgriffin: You should expand on it more. Is a "draft" something on the diesel side? At what exact point does the record get inserted into the database?

sgrif: It's just a word for a post that isn't published (e.g. `published` is set to `false` in the database.) I set up the example this way to demonstrate that

a. Diesel doesn't fuck with your database defaults (we defaulted `published` to `false` in the migration)
b. To give me a bullshit reason to demonstrate how `update` works.

Clearly this is too implicit, though.

tessgriffin: I think the terms are just confusing. You say show posts won't render because it's not 'published'. To me that implies that it's not saved in the database? I guess I don't relate 'publishing' to calling update. I might be clearer to have say a 'rough draft' when it's first saved into the database. And you're updating the post to be a 'final draft'

tessgriffin: That could be partially my fault to. I didn't even realized there was a published column on the database. I think this example would probably be fine as long as you say "hey, we set this to be false by default, now let's update it to be true". That would make sense if it's more explicit.

sgrif: Maybe I need to better call out the `published.eq(true)` section of https://github.com/sgrif/diesel/pull/181/files#diff-8c27d0d0b409c1de1867fdc9a128c72cR158 ?

sgrif: FWIW I'm not trying to be confrontational or anything, just trying to identify *where* I need to improve things.

tessgriffin: Yeah, that'd be a great thing to point out. And then to point out in your update that that's the column that you're changing.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+You know, a CLI application probably isn't the best interface for a blog demo.
+But really I just wanted a semi-simple example, where I could focus on Diesel.
+I didn't want to get bogged down in some web framework here.
+Plus I don't really like the Rust web frameworks out there. We might make a
+new one, soon.
+
+Saved draft Diesel demo with id 1
+```
+
+Unfortunately, running `show_posts` still won't display our new post, because we
```
tessgriffin: I think even here on this line. You could say "Unfortunately, running `show_posts` still won't display our new post because we published to default to false in the database earlier. We'll have to 'update' our post to publish it!" or something like that.

guide.md
derekprior: I think the use of the term `run`, both in the cli and the documentation bothers me because regardless of the direction of the migration, you are `running` something.

sgrif: :runner:

guide.md
mfpiccolo: I think something along those lines could be sufficient but I agree with Derek that it should have at least a quick explanation.

guide.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+This will create our database for us (if it didn't already exist), and create an
+empty migrations directory that we can use to manage our schema. More on that
+later. Let's add Diesel to our `Cargo.toml`. We'll also add `dotenv` so we can
+use that same `.env` file for our environment variables.
+
+```toml
+[dependencies]
+diesel = "^0.5.0"
```
steveklabnik: `^` is already the default with Cargo, so this and the others should just be "0.5.0"

guide.md
```diff
@@ -0,0 +1,388 @@
...
+  body TEXT NOT NULL,
+  published BOOLEAN NOT NULL DEFAULT 'f'
+)
+
+-- down.sql
+DROP TABLE posts
+```
+
+We can run our new migration with `diesel migration run`. It's usually a good
+idea to make sure that `down.sql` is correct. You can do a quick sanity check by
```
steveklabnik: ... this raises the question, when is it a good idea not _not_ make sure something is correct?

sgrif:
I've addressed the feedback given so far. Can anyone take another look?

guide.md
```diff
@@ -0,0 +1,406 @@
+Getting Started 2.0
+===
+
+For this guide, we're going to walk through some simple examples for each of the
+pieces of CRUD. Each step in this guide will build on the previous, and is meant
```
samphippen: Is it worth explaining the acronym `CRUD` here. Generally, in guides like this, it's a good idea to define each term the first time it's used.

guide.md
```diff
@@ -0,0 +1,406 @@
...
+cargo new diesel_demo
+cd diesel_demo
+```
+
+First, let's add Diesel to our dependencies. We're also going to use a tool
+called [`.env`][dotenv-rust] to manage our environment variables for us. We'll
+add it to our dependencies as well.
+
+[dotenv-rust]: https://github.com/slapresta/rust-dotenv
+
```
samphippen: is it worth adding a thing that points out this is cargo.toml here?

sgrif: Yes, this will go into those little terminal windows with the file name in the top bar

guide.md
```diff
@@ -0,0 +1,406 @@
...
+echo DATABASE_URL=postgres://localhost/diesel_demo > .env
+```
+
+Now Diesel CLI can set everything up for us.
+
+```shell
+diesel setup
+```
+
+This will create our database for us (if it didn't already exist), and create an
```
samphippen: the `for us` here is redundant.

guide.md
```diff
@@ -0,0 +1,406 @@
...
+```
+
+You should see output which looks something like this:
+
+```
+Creating migrations/20160202154039_create_posts/up.sql
+Creating migrations/20160202154039_create_posts/down.sql
+```
+
+Migrations allow us to evolve the database schema over time. Each migration can
```
samphippen: this is a beautiful sentence.

guide.md
```diff
@@ -0,0 +1,406 @@
...
+be applied (`up.sql`) or reverted (`down.sql`). Applying and immediately
```
samphippen: Applied and reverted are not antonyms. I wonder if there are two words here we could use that are true antonyms of each other.

guide.md
sgrif: screen shot 2016-02-03 at 10 48 38 am

guide.md
```diff
@@ -0,0 +1,406 @@
...
+-- down.sql
+DROP TABLE posts
+```
+
+We can apply our new migration with `diesel migration run`. It's a good idea to
+make sure that `down.sql` is correct. You can do a quick sanity check by doing
+`diesel migration redo`, which will revert and reapply the latest migration.
+
+OK enough SQL, let's write some Rust. We'll start by writing some code to show
+us the last 5 published posts. The first thing we need to do is establish a
```
samphippen: I might delete the "us" here.

guide.md
```diff
@@ -0,0 +1,406 @@
...
+    dotenv().ok();
+
+    let database_url = env::var("DATABASE_URL")
+        .expect("DATABASE_URL must be set");
+    PgConnection::establish(&database_url)
+        .expect(&format!("Error connecting to {}", database_url))
+}
+```
+
+We'll also want to create a `Post` struct that we can read our data into, and
```
samphippen: The dangling preposition on this clause makes me unhappy, can this read "into which we can read our data,"?

guide.md
```diff
@@ -0,0 +1,406 @@
...
+have diesel generate the names we'll use to reference our tables and columns in
```
samphippen: I would drop the "our" from "our tables and columns"

guide.md
```diff
@@ -0,0 +1,406 @@
...
+#![feature(custom_derive, custom_attribute, plugin)]
+#![plugin(diesel_codegen, dotenv_macros)]
+
+pub mod schema;
+pub mod models;
+```
+
+The first two lines tell Rust that we want to use some special compiler plugins
+provided by Diesel and Dotenv. These will add various useful class annotations
+that we can use, as well as the `infer_schema!` macro which we'll see in just a
```
samphippen: this sentence is very complex, is there any way of breaking it apart.

guide.md
```diff
@@ -0,0 +1,406 @@
...
+    pub published: bool,
+}
+
+// src/schema.rs
+infer_schema!(dotenv!("DATABASE_URL"));
+```
+
+The `#[derive(Queryable)]` will generate all of the code needed to load a `Post`
+struct from a SQL query. The `infer_schema!` macro connects to the database URL
+given to it, and creates a bunch of code based on the database schema to
```
samphippen: "The `infer_schema!` macro connects to the database URL which it is given. The macro also creates a bunch of code based on the database schema to represent all of the tables and columns."

guide.md
sgrif: screen shot 2016-02-03 at 10 53 45 am

guide.md
```diff
@@ -0,0 +1,406 @@
...
+The `use posts::dsl::*` line imports a bunch of aliases so that we can say
+`posts` instead of `posts::table`, and `published` instead of
+`posts::published`. It's useful when we're only dealing with a single table, but
+that's not always what we want.
+
+We can run our script with `cargo run --bin show_posts`. Unfortunately, the results
+won't be terribly interesting, as we don't actually have any posts in the
+database. Still, we've written a decent amount of code, so let's commit.
+
+The full code for the demo at this point can be found [here][commit-no-1].
```
samphippen: worth emboldening?

guide.md
```diff
@@ -0,0 +1,406 @@
...
+----------
+
+You know, a CLI application probably isn't the best interface for a blog demo.
+But really I just wanted a semi-simple example, where I could focus on Diesel.
+I didn't want to get bogged down in some web framework here.
+Plus I don't really like the Rust web frameworks out there. We might make a
+new one, soon.
+```
+
+We've still only covered 3 of the 4 letters of CRUD though. Let's show how to
```
samphippen: "let's show how to delete things" is an awkward sentence.

guide.md
```diff
@@ -0,0 +1,406 @@
+Getting Started 2.0
+===
+
+For this guide, we're going to walk through some simple examples for each of the
+pieces of CRUD. Each step in this guide will build on the previous, and is meant
+to be followed along. This guide assumes that you're using PostgreSQL on
```
mcasper: `is meant to be follow along.` feels a little awkward. Maybe reword to `is meant to be followed in order.`?

guide.md
```diff
@@ -0,0 +1,406 @@
...
+
+The first thing we need to do is generate our project.
+
+```shell
+cargo new diesel_demo
+cd diesel_demo
+```
+
+First, let's add Diesel to our dependencies. We're also going to use a tool
+called [`.env`][dotenv-rust] to manage our environment variables for us. We'll
```
mcasper: Nitpicking: The tool is called `dotenv`, should we list it as such? Especially since we have to put it in the toml as that

guide.md
```diff
@@ -0,0 +1,406 @@
...
+
+Unfortunately, running `show_posts` still won't display our new post, because we
+saved it as a draft. If we look back to the code in `show_posts`, we added
+`.filter(published.eq(true))`, and we had `published` default to false in our
+migration. We need to publish it! But in order to do that, we'll need to look at
+how to update an existing record. First, let's commit. The code for this demo at
+this point can be found [here][commit-no-2].
+
+Now that we've got create and read out of the way, update is actually relatively
+simple. Let's jump write into the script:
```
mcasper: `Let's jump right`

guide.md
```diff
@@ -0,0 +1,406 @@
...
+    let connection = establish_connection();
+    let num_deleted = diesel::delete(posts.filter(title.like(pattern)))
+        .execute(&connection)
+        .expect("Error deleting posts");
+
+    println!("Deleted {} posts", num_deleted);
+}
+```
+
+We can run the script with `cargo run --bin delete_post demo` (at least with the
```
mcasper: The paren on this line should probably be closed off on the next line, like
```
(at least with the title I chose. Your output should look something like):
```

sgrif: I just like leaving unbalanced parens to make programmers feel uncomfortable (for the rest of the day

guide.md
sgrif: Open to suggestions.

guide.md
sgrif: This is the most British suggestion ever.

sgrif:
Thanks everyone. I've addressed the last round of feedback in the implemented version. https://github.com/sgrif/diesel.rs-website/blob/master/source/views/guides/getting-started.html.slim

PR #187, opened 2016-02-03
Move `compile-fail` to a separate suite

sgrif:
I want to write some tests confirming that PG expressions don't compile
on SQLite. These are only meaningful when tested against Diesel with
both SQLite *and* PG enabled, which we explictly don't want to do in the
main test suite.


sgrif:
Review? @mcasper @mfpiccolo

PR #188, opened 2016-02-03
Move PG specific expressions out of the global namespace

### LINKED ISSUE #153: Audit the expression module for PG specific expressions
Right now there's nothing to stop something like `foo.is_distinct_from(bar)` from compiling when used against SQLite. We need to audit our codebase for cases where something is PG specific, and make sure it only compiles for PG. We also need to look for cases where the syntax is different on SQLite. For example, SQLite has no `NOW` function, using the string literal `'now'` instead.

sgrif:
We've also changed them to only compile when passed to a `PgConnection`,
and added tests to ensure this continues to be true. The
`WithQuerySource` is PG specific syntax, but interestingly, I didn't
have a good place to mark it as such. We've done a hacky workaround for
now.

Fixes #153


diesel/src/expression/expression_methods/mod.rs
```diff
@@ -7,11 +7,10 @@
...
-pub mod timestamp_expression_methods;
 
 pub use self::global_expression_methods::ExpressionMethods;
 pub use self::bool_expression_methods::BoolExpressionMethods;
 pub use self::text_expression_methods::{TextExpressionMethods, VarCharExpressionMethods};
-#[doc(hidden)]
-pub use self::timestamp_expression_methods::TimestampExpressionMethods;
+
+#[cfg(feature = "postgres")]
+pub use pg::expression::expression_methods::*;
```
mcasper: Do we still want this to be hidden?

sgrif: We want `TimestampExpressionMethods` to be hidden. We don't want the fact that we're glob importing PG's expressions to be hidden.

mcasper: Ah, true

mcasper:
:+1:

PR #189, opened 2016-02-04
Implement Sqlite support for the CLI.

mcasper:
Because `Connection` isn't object safe (and probably shouldn't be),
switching based on the desired backend had to be achieved through
branching, handled with a macro wherever possible. This will obviously
grow more and more burdensome as more backends are added, and a better
solution will have to be figured out in the future.

I also split out database related functions into their own module, as it
was starting to feel like too much to juggle in one file.

The tests in the `database` mod each use their own database, as for many
of them we're dropping and creating the database, which can't be wrapped
in a transaction. Using a separate database for each allows us to still
run the tests multi threaded, but does add the extra cost of some setup
and teardown in the code.

The Sqlite portions don't use the `tempdir` crate for generating their
database path because I couldn't figure out an ergonomic way of keeping
the tests able to run for both backends, while keeping the TempDir in
scope (returning a path from a function causes the TempDir to go out of
scope, which removes the directory in its destructor).

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,7 +2,8 @@
 set -e
 
 (cd diesel && cargo test --features "unstable chrono sqlite")
-(cd diesel_cli && cargo test)
+(cd diesel_cli && cargo test --features "postgres" --no-default-features)
+(cd diesel_cli && cargo test --features "sqlite" --no-default-features)
 (cd diesel_codegen && cargo test --no-default-features --features nightly)
 (cd diesel_tests && cargo test --features "unstable postgres" --no-default-features)
 (cd diesel_tests && cargo test --features "unstable sqlite" --no-default-features)
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -15,8 +15,12 @@ name = "diesel"
 [dependencies]
 chrono = "^0.2.17"
 clap = "^1.5.5"
-diesel = "^0.4.0"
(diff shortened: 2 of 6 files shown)


sgrif:
The tests for this fail to compile.

bin/test
```diff
@@ -2,7 +2,8 @@
 set -e
 
 (cd diesel && cargo test --features "unstable chrono sqlite")
-(cd diesel_cli && cargo test)
+(cd diesel_cli && cargo test --features "postgres" --no-default-features)
+(cd diesel_cli && cargo test --features "sqlite" --no-default-features)
```
sgrif: We need to tell travis to run both of these, too.

diesel_cli/Cargo.toml
```diff
@@ -15,8 +15,12 @@ name = "diesel"
 [dependencies]
 chrono = "^0.2.17"
 clap = "^1.5.5"
-diesel = "^0.4.0"
+diesel = { path = "../diesel", features = ["postgres", "sqlite"] }
```
sgrif: This should just be `default-features = false`

sgrif:
Can you break out the stuff that moved into `database.rs` without changing in a separate commit to make it easier to review?

mcasper:
Yeah sorry about that, the first commit should be a little easier to diff now

sgrif:
One comment. Please squash before merging.

PR #191, opened 2016-02-04
Clean up the migration signatures to something reasonable

sgrif:
The where clauses that were needed when we moved to a generic backend
structure are leaking into the docs, making these functions seem way
more scary than they actually are (this could just as easily be
`&Connection` if connection were object safe).

screen shot 2016-02-04 at 10 30 01 am

I've moved everything onto another trait so that the signatures of these
functions look a bit more reasonable.

Review? @mcasper @mfpiccolo

### DIFF
--- a/diesel/src/migrations/connection.rs
+++ b/diesel/src/migrations/connection.rs
@@ -1,0 +1,41 @@
+use std::collections::HashSet;
+
+use prelude::*;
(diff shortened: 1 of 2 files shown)


mcasper:
This is great, making the documentation more approachable is a huge win IMO.

PR #192, opened 2016-02-04
Panic if we failed to drop SQLite resources

sgrif:
SQLite is unique in that we can actually fail to clean up our resources
in `Drop`. The original implementation would *always* panic if we failed
there. However, the only time these `drop` calls would fail is if we're
in an inconsistent state. The most likely reason is that we are dropping
as the result of another `panic!`, which means that panicking will cause
Rust to abort with the unhelpful message "panicked while panicking".

However, I've had cases where the tests for SQLite have error output,
but did not fail, because this *isn't* panicking. This change causes us
to panic if not already panicking, and write to STDOUT otherwise. (Same
as before, we just crash if writing to stdout fails. If we can't clean
up resources, *and* we can't write to STDOUT, the computer is on fire.)

### DIFF
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -61,10 +61,16 @@ impl RawConnection {
 
 impl Drop for RawConnection {
     fn drop(&mut self) {
+        use std::thread::panicking;
+
         let close_result = unsafe { ffi::sqlite3_close(self.internal_connection) };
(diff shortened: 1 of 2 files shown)


sgrif:
@mfpiccolo @mcasper Review?

mcasper:
Neat. Should we standardize on `Display` vs `Debug` in these error messages?

sgrif:
Yes, I need to re-use that `ensure_sqlite_ok` function somewhere.

PR #193, opened 2016-02-05
Add support for `IN` statements.

### LINKED ISSUE #186: Add support for `IN` statements
Prior to 1.5 there was no need for it, as we could use `eq(any())` instead, which is more friendly to prepared statements. However, the `any` operator is PG specific. We need to be careful with this, as it could cause unbounded growth of prepared statements.

sgrif:
We didn't have a use for this previously, as PG supports `= ANY` which
is more friendly to prepared statements. Since SQLite does not support
this operator, we need to fall back to `IN`. Eventually I'd like to have
this just "do the right thing" on PG, but that's out of scope for 0.5.

The name `eq_any` was chosen both to allow me to change the behavior on
PG in the future, and also because `in` is a keyword in Rust and cannot
be used as a function name.

Fixes #186

### DIFF
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -1,0 +1,57 @@
+use backend::Backend;
+use expression::*;
+use query_builder::{QueryBuilder, QueryFragment, BuildQueryResult};
+use types::Bool;
+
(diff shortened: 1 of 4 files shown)


mfpiccolo:
Nicely done. :+1:

PR #194, opened 2016-02-05
Support subselects for IN statements.

sgrif:
This is a feature that was supported by `eq(any())`, but not currently
by the in clause. The `Subselect` type was introduced to normalize the
differences between the SQL type of `SelectStatement` depending on the
backend. That implementation is a bit of a wart, regardless. We can
probably revisit this later and factor out some kind of a `Many` SQL
type which is independent of `Array`. I need to better analyze the
impact of that change though, (specifically making sure that it
continues to work with `eq(any())`), and I think it's out of scope for
0.5.

### DIFF
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -5,11 +5,11 @@ use types::Bool;
 
 pub struct In<T, U> {
     left: T,
-    values: Vec<U>,
+    values: U,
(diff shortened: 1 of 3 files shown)


mcasper:
`Many(expressions)` such expressions

Looks good to me :+1:

mfpiccolo:
Agreed :+1:

Issue #179, opened 2016-02-03
We should be able to pass a model to `update`


sgrif:
I'd like to be able to write something like this:

``` rust
let post = try!(posts::table.find(id).get_result::<Post>(&connection));

update(post).set(posts::published.eq(true)).execute(&connection)
```

`post` in that example should be a shorthand for `posts::table.find(id)`. Why not just write that? Because I want this code to return `Err(DatabaseError::NotFound)` if the id doesn't exist. This will probably require a new trait, instead of tacking it onto an exiting one. I've been mulling over `#[derive(Identifiable)]` for a while, which I think will eventually be required for associations anyway.

sgrif:
This was implemented by https://github.com/diesel-rs/diesel/commit/a7373087d7c18f7b594435def83697098511a249

PR #195, opened 2016-02-05
Fix remaining FIXMEs that are going to be fixed in 0.5

### LINKED ISSUE #173: Remove any remaining `FIXME`s related to SQLite support
This is a blanket reminder issue, as I'm not sure I have an individual issue for each of these. Any that can be resolved for 0.5 should. Any that we are not going to fix in 0.5 should still be removed, and have a separate issue opened for them.

If you're reading this and you've found a specific comment which doesn't have an issue, please open one.

sgrif:
This adds two new missing features to `SQLite`. A `ToSql` impl for
`Bool`, and an implementation of `save_changes`. I wanted to keep the
blanket impl for `SupportsReturningKeyword`, but it looks like I can't
rely on `Sqlite` not implementing it from the user's crate. I think this
will just get moved in crate once we solve #179 anyway.

There are still some remaining things to be addressed, such as the lack
of support for numeric and date types, but those are out of scope for
0.5.

Fixes #173.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -8,6 +8,7 @@ pub mod load_dsl;
 pub mod select_dsl;
 #[doc(hidden)]
 pub mod filter_dsl;
+mod save_changes_dsl;
 mod offset_dsl;
 mod order_dsl;
 mod with_dsl;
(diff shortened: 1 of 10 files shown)


mcasper:
Nice work! :+1:

PR #197, opened 2016-02-07
Run rustfmt on a truckload of files

killercup:
> Let's run [rustfmt](https://github.com/rust-lang-nursery/rustfmt) on diesel. What could go wrong?

Well, it reformatted about a gazillion `where`s and re-indented every second method call… rustfmt tends to add a lot of 'right-shift' to the code (which is somewhat mitigated by setting chained method calls to be indented by 4 spaces instead of visually). This also helped me discover rust-lang-nursery/rustfmt#811 (whose occurrences I manually fixed here).

I tried to split the changes into commits that make sense on their own (and in case Github struggles with the size of this diff). I omitted some changes that seemed non-sensical (mostly reformatting of manually aligned arrays).

This only as a suggestion—feel free to pick and choose what you want from this PR! :smile:


killercup:
By the way, I wrote [this shell script](https://gist.github.com/killercup/d2455f122157a93c2bdc) to quickly check if everything still compiled (without running the tests, it takes ages on my machine). Maybe it's useful to you.

sgrif:
Last time I'd checked `rustfmt`, it still didn't support a few options that we'd need to use it. The big one that stands out immediately is moving the `where` to the next line. Generally our style is chosen to reduce git churn as much as possible, as I rely on `git blame` pretty heavily for context. The `where` change would mean the line with the second predicate would have to change if I want to delete the first.

Still, I'll fiddle with it a bit and see if it's at a reasonable place for us to use. I'm willing to make some concessions if it means being able to use this in place of a style guide.

killercup:
> The where change would mean the line with the second predicate would have to change if I want to delete the first.

That's my issue with the current style as well. I really enjoy the visual style of function arguments (it's quite nice to read once one gets used to it), but it's pretty bad for diffs as well. At least rustfmt now allows setting more trailing commas! :smile:

sgrif:
Just so I'm clear, when you say "the current style", you mean the style proposed by this PR?

killercup:
> Just so I'm clear, when you say "the current style", you mean the style proposed by this PR?

Yes. I meant the style that rustfmt currently outputs by default.

I'm not sure if you saw this (or remember it, it was some time ago), but this problem was also discussed in <https://github.com/rust-lang/rfcs/pull/1190>.

sgrif:
Looks like there is no option for the placement of `where`. Unfortunately, this would be just too much churn without it. :( https://github.com/rust-lang-nursery/rustfmt/issues/815

PR #199, opened 2016-02-09
Allow to run migrations on application start up

weiznich:
* Allow to run migrations on application start up without depending on migrations directory
* Add functionality to generate the needed code from the current migrations directory


diesel/src/migrations/migration.rs
```diff
@@ -1,18 +1,13 @@
...
 use std::path::{Path, PathBuf};
 
-pub trait Migration {
-    fn version(&self) -> &str;
-    fn run(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
-    fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
-}
 
-pub fn migration_from(path: PathBuf) -> Result<Box<Migration>, MigrationError> {
+pub fn migration_from(path: PathBuf) -> Result<SqlFileMigration, MigrationError> {
```
sgrif: The signature of this shouldn't change. We keep these boxed for a reason.

diesel/src/migrations/migration.rs
```diff
@@ -49,7 +44,7 @@ fn version_from_path(path: &Path) -> Result<String, MigrationError> {
 use std::fs::File;
 use std::io::Read;
 
-struct SqlFileMigration(PathBuf, String);
+pub struct SqlFileMigration(PathBuf, String);
```
sgrif: This shouldn't need to change.

diesel/src/migrations/migration.rs
```diff
@@ -65,10 +60,27 @@ impl Migration for SqlFileMigration {
...
 }
 
-fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigrationsError> {
+
+pub fn programmatic_migration_from(path: PathBuf) -> Result<String, MigrationError> {
+    if valid_sql_migration_directory(&path) {
+        let version = try!(version_from_path(&path));
+        let up = try!(sql_from_file(&path.join("up.sql")));
+        let down = try!(sql_from_file(&path.join("down.sql")));
+        Ok(format!("ProgrammaticMigration{{ version:\"{}\", up:\"{}\", down: \"{}\"}}",version.trim(),up.trim(),down.trim()))
```
sgrif: Code shouldn't get constructed by string munging. Can you move this to an annotation in codegen?

diesel/src/migrations/mod.rs
```diff
@@ -91,27 +93,128 @@ use std::path::{PathBuf, Path};
...
+/// include!(concat!(env!("OUT_DIR"), "/migrations.rs"));
+/// 
+/// use self::migrations::run;
+/// let connection = establish_connection();
+/// run(&conn).unwrap();
+/// ```
+/// See the [module level documentation](index.html) for information on how migrations should be
+/// structured, and where Diesel will look for them by default.
+
+pub fn generate_migrations_module(dst: &Path) -> Result<(), RunMigrationsError> {
```
sgrif: Again, code generation doesn't belong in the main Diesel crate, nor should we be writing to files. This should happen inside of codegen.

diesel/src/migrations/mod.rs
```diff
@@ -91,27 +93,128 @@ use std::path::{PathBuf, Path};
...
+    let mut f = try!(File::create(dst));
+    try!(f.write_all(&migration_module.as_bytes()));
+    Ok(())
+}
+
+
 #[doc(hidden)]
-pub fn run_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path, output: &mut Write)
-    -> Result<(), RunMigrationsError> where
+pub fn run_pending_migrations_form_iter<Conn, T, M>(conn: &Conn, all_migrations:T, output: &mut Write)-> Result<Vec<String>, RunMigrationsError> where
```
sgrif: This method has a typo.

sgrif: Please make sure the code adheres to the [style guidelines](https://github.com/sgrif/diesel/issues/120#issuecomment-174222639)

diesel/src/migrations/mod.rs
```diff
@@ -91,27 +93,128 @@ use std::path::{PathBuf, Path};
...
+    run_pending_migrations_form_iter(conn,all_migrations.into_iter(),output)
+}
+
+fn build_mirgations_module() -> Result<String, MigrationError>{
+
+    let migrations_dir = try!(find_migrations_directory());
+    build_migrations_module_form_directory(&migrations_dir)
+}
+
+fn join<S:ToString> (l: Vec<S>, sep: &str) -> String{
```
sgrif: Is there a reason we can't use join from the standard library?

diesel/src/migrations/mod.rs
```diff
@@ -91,27 +93,128 @@ use std::path::{PathBuf, Path};
...
+                  |a,b| if a.len()>0 {a+sep} else {a} + &b.to_string()
+                  )
+}
+
+fn build_migrations_module_form_directory(migrations_dir: &Path) -> Result<String, MigrationError>
+{
+    use self::migration::programmatic_migration_from;
+
+    let migrations = try!(migrations_in_directory(migrations_dir,&programmatic_migration_from));
+    Ok(format!("pub mod migrations {{
```
sgrif: Same thing here. We shouldn't be creating code using string munging, this needs to go in codegen.

diesel/src/migrations/mod.rs
```diff
@@ -238,6 +346,39 @@ fn search_for_migrations_directory(path: &Path) -> Result<PathBuf, MigrationErro
...
 
+#[doc(hidden)]
+pub trait Migration {
+    fn version(&self) -> &str;
+    fn run(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
+    fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
+}
+
+#[doc(hidden)]
+pub struct ProgrammaticMigration {
```
sgrif: There's no reason that this needs to be in the main crate, we can stick this in the module generated for the users.

diesel_compile_tests/tests/compile_tests.rs
```diff
@@ -23,4 +23,5 @@ fn run_mode(mode: &'static str) {
 #[test]
 fn compile_test() {
     run_mode("compile-fail");
+    run_mode("run-pass");
```
sgrif: Is there a reason these tests can't go in the main integration test suite?

sgrif:
Also worth noting that all of the documentation examples you've added fails to run. That will need to be addressed as well before this can be merged. We have a script to help you run all of the test suites locally at `bin/test`, usually good to double check it before pushing.

weiznich:
Thanks for the comments. I will rework this to use codegen instead.

weiznich:
I've updated the pull request according to your comments.

sgrif:
Sorry for the delay in responding to this. I've been on hiatus for a few weeks. I'll take another look at this soon -- just wanted to let you know I haven't forgotten

weiznich:
Is there anything preventing this from being merged, except of time?

diesel_codegen/Cargo.toml
```diff
@@ -17,7 +17,7 @@ syntex_syntax = { version = "^0.28.0", optional = true }
 [dependencies]
 syntex = { version = "^0.28.0", optional = true }
 syntex_syntax = { version = "^0.28.0", optional = true }
-diesel = { version = "0.5.0", default-features = false }
+diesel = { path="../diesel/", default-features = false }
```
sgrif: This shouldn't be required, it's overloaded in `.cargo/config`

diesel_tests/build.rs
```diff
@@ -39,7 +39,7 @@ mod outer {
         let database_url = ::std::env::var("DATABASE_URL")
             .expect("DATABASE_URL must be set to run tests");
         let connection = PgConnection::establish(&database_url).unwrap();
-        let migrations_dir = migrations::find_migrations_directory().unwrap().join("postgresql");
+        let migrations_dir = migrations::find_migrations_directory().unwrap().join("../../migrations/postgresql");
```
sgrif: Why did this change?

diesel_tests/tests/migrations.rs
```diff
@@ -0,0 +1,65 @@
...
+use diesel::{select, LoadDsl};
+use diesel::result::QueryResult;
+use diesel::connection::Connection;
+use diesel::expression::dsl::sql;
+use diesel::migrations::revert_migration;
+use schema::*;
+
+
+// FIXME: Remove the duplicates of this function once expression level attributes
+// are stable (I believe this is in 1.7)
```
sgrif: Sorry this has taken so long to review, but 1.7 is stable now.

sgrif:
There's a lot of style nitpicks as well, but I can just fix those manually. Will take another look today or tomorrow when I can go a bit more in depth.

diesel_tests/build.rs
weiznich: Without this change the dummy migrations from diesel_test are used(They are added in this commit)

sgrif: Ah, makes sense. Honestly, with this feature added, I'm thinking we might just be able to cut most of the existing "runtime" migration functionality, and move it into `diesel_cli`. Packaging the migrations in the binary feels more rust-like. That doesn't need to be part of this PR though, this is fine for now.

sgrif: Actually, can you move the migrations that you added in this PR to a directory name other than `migrations` so this doesn't need to change?

diesel_tests/tests/migrations.rs
mcasper: Unfortunately expression level attributes weren't stabilized in 1.7 :/ 
https://github.com/rust-lang/rust/blob/1.7.0/src/libsyntax/feature_gate.rs#L232

sgrif: For match arms it was.

weiznich: We don't even need this here. Tests are only build for postgres or sqlite, not for both at the same time.

weiznich:
Hopefully all style nitpicks are fixed. If not please leave a comment.

sgrif:
Heading out of town, but I'll take a look when I have a chance.

diesel/src/migrations/mod.rs
```diff
@@ -91,35 +92,47 @@ use std::path::{PathBuf, Path};
...
-pub fn run_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path, output: &mut Write)
-    -> Result<(), RunMigrationsError> where
-        Conn: MigrationConnection,
-{
+pub fn run_pending_migrations_from_iter<Conn, T>(conn: &Conn,
+                                                 all_migrations: T,
+                                                 output: &mut Write)
+    -> Result<Vec<String>, RunMigrationsError>
+    where Conn: MigrationConnection,
+          T: Iterator<Item = Box<Migration>>, {
```
sgrif: Format for this should be:

```rust

pub fn run_pending_migrations_from_iter<Conn, T>(
    conn: &Conn,
    all_migrations: T,
    output: &mut Write,
) -> Result<Vec<String>, RunMigrationsError> where
    Conn: MigrationConnection,
    T: Iterator<Item = Box<Migration>>,
{
```

weiznich: Fixed

diesel/src/migrations/migration.rs
```diff
@@ -1,13 +1,9 @@
 use connection::SimpleConnection;
-use super::{MigrationError, RunMigrationsError};
+use super::{Migration, MigrationError, RunMigrationsError, valid_sql_migration_directory,
+            version_from_path};
 
 use std::path::{Path, PathBuf};
 
-pub trait Migration {
```
sgrif: Why move this?

diesel/src/migrations/mod.rs
```diff
@@ -91,66 +92,89 @@ use std::path::{PathBuf, Path};
...
+    Conn: MigrationConnection,
 {
     migration_with_version(ver)
         .map_err(|e| e.into())
-        .and_then(|m| revert_migration(conn, m, output))
+        .and_then(|ref m| revert_migration(conn, m, output))
 }
 
 #[doc(hidden)]
-pub fn run_migration_with_version<Conn>(conn: &Conn, ver: &str, output: &mut Write)
```
sgrif: Please don't reformat existing code that isn't being changed.

sgrif:
It's unclear to me why any code in the migrations module should need to be run. It seems like this change needs to:

- Generate a struct with the up and down string, along with the version
- Implement the `Migration` trait
- Generate a constant with all the instances of this struct
- Generate a single function to which runs from that constant

diesel/src/migrations/migration.rs
weiznich: this module is not public, so I can't use this trait in codegen

weiznich:
Most changes in the migration module are needed to reuse the code in codegen (migration::migrations is not public). Furthermore I add a function to run migrations from a Iterator instead of running them from a directory.
The codegen part does more or less what you describe.

sgrif:
Can we do some pub exports or make the module public instead of moving so
much code around?

weiznich:
I revisited and minimized the changes in the migration module.

sgrif:
Thank you. I will review/merge soon.

sgrif:
Thank you for working on this. I've gone ahead with a separate implementation, though this PR was used as a reference. I'll make sure you're credited for the feature in the next release announcement.

PR #200, opened 2016-02-10
Add integration tests for diesel_cli

mcasper:
I only added tests for `diesel setup` and `diesel migration generate` for now, wanted to start getting feedback on the design and shared code approach.

Shared code lives in its own crate called `diesel_test_helpers` (name very open to suggestions) and provides utilities for setting up a test environment/database/connection.

I'll go through and add the rest of the integration tests if we like this approach.


sgrif:
What is the `diesel_test_helpers` code shared with? It's only used by the tests you added?

mcasper:
Yeah - because diesel_cli is a binary package, I couldn't find a great way to share the code easily other than making it its own crate

sgrif:
What do you mean by share code? What is it being shared with?

mcasper:
The `TestEnvironment/TestDatabase/TestConnection` structs and associated methods

mcasper:
Sorry that's not actually a clear answer. Those structs setup a clean environment for each test to run in, so they're needed in each `tests` file. I would have loved to put them all in just a module inside of `diesel_cli`, but I don't think the tests directory supports that, and we can't import the package into the tests since it's a binary

sgrif:
You don't need to move it to another crate for that, take a look at cargo's test suite for example.

mcasper:
That's what I tried at first, but I don't know what black magic they do in their Makefile to get that all to work

sgrif:
https://github.com/rust-lang/cargo/blob/master/Cargo.toml#L57-L58 no black magic involved

mcasper:
Of course... Alright, off to a support module it goes!

diesel_cli/tests/migration_generate.rs
```diff
@@ -0,0 +1,29 @@
...
+    let diesel_exe = env::current_exe().unwrap().parent().unwrap().join("diesel");
+    let mut command = Command::new(diesel_exe);
+    command.arg("migration")
+        .arg("generate")
+        .arg("create_posts_table")
+        .current_dir(&test_environment.root_path());
+    command.output().unwrap();
+
+    let create_posts_dir = fs::read_dir(&migrations_dir).unwrap().nth(0).unwrap().unwrap().path();
+    let filenames: Vec<String> = create_posts_dir.read_dir()
```
sgrif: The style of a lot of these tests bugged me. I just pushed some stuff up to master to demonstrate what I'd like the style of these tests to look like. Mainly I want the signal to noise ratio to be a bit better.

mcasper: Fair enough, I'll close this for now

sgrif: Sorry. :(

mcasper: No worries :p

PR #203, opened 2016-02-10
Allow the migration version to be overridden at the command line

sgrif:
This will make our lives much easier for testing purposes, as we don't
have to deal with timestamps (unless we're tesing something that
involves timestamps)

Review? @mcasper


mcasper:
Looks good to me :+1:

PR #204, opened 2016-02-10
Allow the migration directory to be overridden

sgrif:
This currently only affects the generator, as we don't have tests for the other commands yet. This is to allow us to use CLI on our own codebase again, as we have a non-standard migration directory structure.

This relies on #203.

Review? @mcasper

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -31,7 +31,14 @@ fn main() {
 
     let migration_subcommand = SubCommand::with_name("migration")
         .setting(AppSettings::VersionlessSubcommands)
-        .subcommand(
+        .arg(Arg::with_name("MIGRATION_DIRECTORY")
+            .long("migration-dir")
+            .help("The location of your migration directory. By default this \
+                   will look for a directory called `migrations` in the \
+                   current directory and its parents.")
+            .takes_value(true)
+            .global(true)
+        ).subcommand(
             SubCommand::with_name("run")
                 .about("Runs all pending migrations")
         ).subcommand(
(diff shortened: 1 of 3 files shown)


diesel_cli/src/main.rs
```diff
@@ -191,22 +221,23 @@ fn redo_latest_migration<Conn>(conn: &Conn) where
     }).unwrap_or_else(handle_error);
 }
 
-fn handle_error<E: Error>(error: E) {
+fn handle_error<E: Error, T>(error: E) -> T {
```
mcasper: Doesn't this need to return something?

sgrif: No, this function never returns. I just noticed it was being used in a strange way on some lines affected by this change, and modified the signature so it can be used with `unwrap_or_else` instead of `map_err.unwrap`

mcasper: Oh neat, I didn't know the compiler would let this slide. Does this emit an `unreachable_code` warning?

sgrif: No, there's no code that is unreachable.

reem: A better way to indicate this is to return `!`, which is the type of `panic!()`, which should have less inference errors and actually indicate that code following this is unreachable.

sgrif: `-> !` does not type check in these cases.

sgrif: `src/main.rs:175:71: 175:85 error: the trait `core::ops::FnOnce<(database_error::DatabaseError,)>` is not implemented for the type `fn(_) -> ! {handle_error}` [E0277]`

PR #207, opened 2016-02-11
Remove the hard dependencies on SQLite and Postgres from CLI

### LINKED ISSUE #205: cargo install diesel_cli does not work if you haven't configured sqlite
If you haven't installed sqlite  then (on windows) the out of the box `cargo install diesel_cli ` will not work, due to not finidng the SQLITE3_LIB_DIR. 

I'm not sure if this matters but it means that if you are just targeting PostGres you have to install an additional dependency.

sgrif:
Right now it's impossible to install Diesel CLI on your system if you
don't have both of these dependencies installed. This continues to
support both by default, but allows you to opt out of one of them if you
specifically need to.

This differs from our other crates where we only target Postgres by
default. However, I want `cargo install diesel_cli` to "just work", and
this case doesn't affect the binary size or memory usage of your actual
application.

Fixes #205

Review? @mcasper

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -15,7 +15,7 @@ name = "diesel"
 [dependencies]
 chrono = "0.2.17"
 clap = "1.5.5"
-diesel = { version = "0.5.0", features = ["postgres", "sqlite"] }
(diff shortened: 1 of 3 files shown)


mcasper:
Ouch. CLI Please

:+1:

PR #213, opened 2016-02-16
Add more CLI integration tests

mcasper:
Add integration tests for all of the current CLI commands, including the hidden `database drop` command. There are a couple small fixes that the CLI needs, but I wanted to get these tests in first.

I did each test file as a separate commit, in case you prefer reviewing that way. I'll squash if we like this


sgrif:
:shipit:

mcasper:
Squashing!

PR #214, opened 2016-02-16
Fix create_database_if_needed output

mcasper:
We should only notify that we're creating a database when we actually
are. Sqlite has always been missing this, and it seems to have been
introduced to Postgres unintentionally.

Relies on #213 for testing infrastructure


sgrif:
Woo tests!

:shipit:

mcasper:
I'm not sure why Travis decided to ignore that last commit :confused: Hopefully the rebased commit will get picked up

PR #219, opened 2016-02-17
Update bin/test to work for all channels

mcasper:
While trying to figure out why our master build is failing, I realized `bin/test` only works on nightly, so I updated it to behave more like our travis build. Still no clue why travis is sad though :(


sgrif:
I don't think we need to add that overhead here, as this is usually just meant to be the "locally run smoke test". When dealing with stable failures, just running the integration test suite is almost always enough.

PR #223, opened 2016-02-20
Remove database tests and replace with integration tests

### LINKED ISSUE #222: Diesel CLI's tests leave files laying around if tests fail
I'm working on some test failures that are occuring on Windows for Diesel CLI, and noticed that the tests in the database module aren't structured properly. They rely on calling a teardown function at the end, but that function is never called if the tests fail. These tests need to operate inside of a tempdir instead so that cleanup happens automatically. /cc @mcasper

mcasper:
These database tests aren't providing much more than our integration
tests are now, and the integration tests have a much cleaner and nicer
test structure to run in, so there doesn't seem like much of a point in
keeping these around. All removed tests have a mirror test or assertion
in the integration test suite, and the one lacking has been added.

This would also fix the problem in which files are created and left behind by failing database tests.
Resolves #222


mcasper:
@sgrif The CLI test suite also passes on a (my) Windows machine on this branch.

PR #226, opened 2016-02-25
Implement infer_schema for sqlite

weiznich:
* refactor postgres implementation to be more generic
* add implementation for sqlite

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -132,6 +132,16 @@ impl SqliteConnection {
         }
         query.map(|_| ())
     }
+
+    #[doc(hidden)]
+    pub fn execute_pragma<ST, U>(&self, source: &str) -> QueryResult<Vec<U>>
+        where U: Queryable<ST, Sqlite>,
+              <SqliteConnection as Connection>::Backend: HasSqlType<ST>
+    {
+        Statement::prepare(&self.raw_connection, source)
+            .map(StatementIterator::new)
+            .and_then(Iterator::collect)
+    }
 }
 
 fn error_message(err_code: libc::c_int) -> &'static str {
--- a/diesel_codegen/src/lib.in.rs
+++ b/diesel_codegen/src/lib.in.rs
@@ -3,10 +3,6 @@ mod attr;
 mod insertable;
 mod model;
 mod queryable;
-#[cfg(feature = "postgres")]
-mod schema_inference;
-#[cfg(not(feature = "postgres"))]
-#[path="dummy_schema_inference.rs"]
 mod schema_inference;
 
 mod update;
--- a/diesel_codegen/src/schema_inference/data_structures.rs
+++ b/diesel_codegen/src/schema_inference/data_structures.rs
@@ -1,54 +1,61 @@
 use diesel::*;
+#[cfg(feature = "postgres")]
 use diesel::pg::Pg;
+#[cfg(feature = "postgres")]
+use diesel::pg::PgConnection;
+#[cfg(feature = "sqlite")]
+use diesel::sqlite::Sqlite;
+#[cfg(feature = "sqlite")]
+use diesel::sqlite::SqliteConnection;
 use diesel::types::{HasSqlType, FromSqlRow};
 
-table! {
-    pg_attribute (attrelid) {
-        attrelid -> Oid,
-        attname -> VarChar,
-        atttypid -> Oid,
-        attnotnull -> Bool,
-        attnum -> SmallInt,
-        attisdropped -> Bool,
-    }
-}
-
-table! {
-    pg_type (oid) {
-        oid -> Oid,
-        typname -> VarChar,
-    }
-}
-
-joinable!(pg_attribute -> pg_type (atttypid));
-select_column_workaround!(pg_attribute -> pg_type (attrelid, attname, atttypid, attnotnull, attnum, attisdropped));
-select_column_workaround!(pg_type -> pg_attribute (oid, typname));
-
-table! {
-    pg_class (oid) {
-        oid -> Oid,
-        relname -> VarChar,
-    }
-}
-
 #[derive(Debug, Clone)]
-pub struct PgAttr {
+pub struct ColumnInformation {
     pub column_name: String,
     pub type_name: String,
     pub nullable: bool,
 }
 
-impl<ST> Queryable<ST, Pg> for PgAttr where
+#[cfg(feature = "postgres")]
+impl<ST> Queryable<ST, Pg> for ColumnInformation where
     Pg: HasSqlType<ST>,
     (String, String, bool): FromSqlRow<ST, Pg>,
 {
     type Row = (String, String, bool);
 
     fn build(row: Self::Row) -> Self {
-        PgAttr {
+        ColumnInformation {
             column_name: row.0,
             type_name: row.1,
             nullable: !row.2,
         }
     }
+}
+
+#[cfg(feature = "sqlite")]
+impl<ST> Queryable<ST, Sqlite> for ColumnInformation where
+    Sqlite: HasSqlType<ST>,
+    (i32, String, String, bool, Option<String>, i32): FromSqlRow<ST, Sqlite>,
+{
+    type Row = (i32, String, String, bool, Option<String>, i32);
+
+    fn build(row: Self::Row) -> Self {
+        ColumnInformation {
+            column_name: row.1,
+            type_name: row.2,
+            nullable: !row.3,
+        }
+    }
+}
+
+#[cfg(all(feature = "sqlite", not(feature = "postgres")))]
+pub type InferConnection = SqliteConnection;
+
+#[cfg(all(feature = "postgres", not(feature = "sqlite")))]
+pub type InferConnection = PgConnection;
+
+#[cfg(all(feature = "sqlite", feature = "postgres"))]
+pub enum InferConnection{
+    Sqlite(SqliteConnection),
+    Pg(PgConnection),
 }
--- a/diesel_codegen/src/schema_inference/mod.rs
+++ b/diesel_codegen/src/schema_inference/mod.rs
@@ -1,7 +1,10 @@
 mod data_structures;
+#[cfg(feature = "postgres")]
+mod pg;
+#[cfg(feature = "sqlite")]
+mod sqlite;
 
-use diesel::*;
-use diesel::pg::PgConnection;
+use diesel::{QueryResult, Connection};
 use syntax::ast;
 use syntax::codemap::Span;
 use syntax::ext::base::*;
@@ -15,7 +18,8 @@ pub fn expand_load_table<'cx>(
     cx: &'cx mut ExtCtxt,
     sp: Span,
     tts: &[ast::TokenTree]
-) -> Box<MacResult+'cx> {
+) -> Box<MacResult + 'cx>
+{
     let mut exprs = match get_exprs_from_tts(cx, sp, tts) {
         Some(ref exprs) if exprs.is_empty() => {
             cx.span_err(sp, "load_table_from_schema! takes 2 arguments");
@@ -35,7 +39,8 @@ pub fn load_table_body<T: Iterator<Item=P<ast::Expr>>>(
     cx: &mut ExtCtxt,
     sp: Span,
     exprs: &mut T,
-) -> Result<Box<MacResult>, Box<MacResult>> {
+) -> Result<Box<MacResult>, Box<MacResult>>
+{
     let database_url = try!(next_str_lit(cx, sp, exprs));
     let table_name = try!(next_str_lit(cx, sp, exprs));
     let connection = try!(establish_connection(cx, sp, &database_url));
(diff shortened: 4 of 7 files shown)


diesel/src/sqlite/connection/mod.rs
```diff
@@ -132,6 +132,16 @@ impl SqliteConnection {
...
+
+    #[doc(hidden)]
+    pub fn execute_pragma<ST, U>(&self, source: &str) -> QueryResult<Vec<U>>
+        where U: Queryable<ST, Sqlite>,
+              <SqliteConnection as Connection>::Backend: HasSqlType<ST>
+    {
+        Statement::prepare(&self.raw_connection, source)
+            .map(StatementIterator::new)
+            .and_then(Iterator::collect)
+    }
```
sgrif: Does this differ in any meaningful way from `sql::<ST>("query").load::<U>(&connection)`?

diesel_codegen/src/lib.in.rs
```diff
@@ -3,9 +3,9 @@ mod attr;
 mod insertable;
 mod model;
 mod queryable;
-#[cfg(feature = "postgres")]
+#[cfg(any(feature = "postgres", feature = "sqlite"))]
```
sgrif: You can remove this config, we don't support any other backends.

diesel_codegen/src/lib.in.rs
```diff
@@ -3,9 +3,9 @@ mod attr;
...
 mod schema_inference;
-#[cfg(not(feature = "postgres"))]
+#[cfg(all(not(feature = "postgres"), not(feature = "sqlite")))]
```
sgrif: You can just remove this module entirely.

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -1,7 +1,12 @@
 mod data_structures;
+#[cfg(feature = "postgres")]
+#[path="pg.rs"]
+mod implementation;
+#[cfg(feature = "sqlite")]
```
sgrif: We need to support using both PG and SQLite in the same code base. We should do the same thing that we do in CLI here.

diesel_codegen/src/schema_inference/pg.rs
```diff
@@ -0,0 +1,104 @@
...
+    database_url: &str,
+) -> Result<PgConnection, Box<MacResult>> {
+    PgConnection::establish(database_url).map_err(|_| {
+        cx.span_err(sp, "failed to establish a database connection");
+        DummyResult::any(sp)
+    })
+}
+
+pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<ast::Ty> {
+    let tpe;
```
sgrif: Can you merge this with the line below? I dislike uninitialized variables.

diesel_codegen/src/schema_inference/data_structures.rs
```diff
@@ -1,54 +1,45 @@
...
+    Sqlite: HasSqlType<ST>,
+    (i32, String, String, i32, Option<String>, i32): FromSqlRow<ST, Sqlite>,
+{
+    type Row = (i32, String, String, i32, Option<String>, i32);
+
+    fn build(row: Self::Row) -> Self {
+        ColumnInformation {
+            column_name: row.1,
+            type_name: row.2,
+            nullable: row.3 == 0,
```
sgrif: Shouldn't this field just be `bool`?

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -0,0 +1,76 @@
...
+    conn.execute_pragma::<pragma_table_info::SqlType, ColumnInformation>(
+        &format!("PRAGMA TABLE_INFO('{}')", table_name))
+}
+
+pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<ast::Ty>
+{
+    let type_name=attr.type_name.to_lowercase();
+    //see https://www.sqlite.org/datatype3.html
+    let tpe = if type_name.contains("int"){
+        quote_ty!(cx, ::diesel::types::Int8)
```
sgrif: `BigInteger` is the proper name for this type.

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -0,0 +1,76 @@
...
+    }else if type_name.contains("char")||
+        type_name.contains("clob")||
+        type_name.contains("text")
+    {
+        quote_ty!(cx, ::diesel::types::Text)
+    }else if type_name.contains("blob") || type_name.is_empty(){
+        quote_ty!(cx, ::diesel::types::Binary)
+    }else{
+        quote_ty!(cx, ::diesel::types::Float8)
```
sgrif: `Double` is the proper name for this type.

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -0,0 +1,76 @@
...
+    let type_name=attr.type_name.to_lowercase();
+    //see https://www.sqlite.org/datatype3.html
+    let tpe = if type_name.contains("int"){
+        quote_ty!(cx, ::diesel::types::Int8)
+    }else if type_name.contains("char")||
+        type_name.contains("clob")||
+        type_name.contains("text")
+    {
+        quote_ty!(cx, ::diesel::types::Text)
+    }else if type_name.contains("blob") || type_name.is_empty(){
```
sgrif: The style of this whole section is really wonky. Please add spaces around your braces, and avoid multiline predicates. If something is too long to fit on a line it should be moved to a function.

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -0,0 +1,76 @@
...
+pub fn get_table_data(conn: &SqliteConnection, table_name: &str) -> QueryResult<Vec<ColumnInformation>>
+{
+    conn.execute_pragma::<pragma_table_info::SqlType, ColumnInformation>(
+        &format!("PRAGMA TABLE_INFO('{}')", table_name))
+}
+
+pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<ast::Ty>
+{
+    let type_name=attr.type_name.to_lowercase();
+    //see https://www.sqlite.org/datatype3.html
```
sgrif: :scissors:

diesel_codegen/src/schema_inference/data_structures.rs
weiznich: If I try to use a bool here the compiler is complaining that `the trait diesel::types::FromSqlRow<diesel::types::Integer, diesel::sqlite::backend::Sqlite> is not implemented for the type bool`

diesel/src/sqlite/connection/mod.rs
weiznich: sql does not implement Query or AsQuery so I cannot call load on it as far as I understood.

weiznich:
Thanks for the comments. 
I've updated the pull request according to your comments.

diesel/src/sqlite/connection/mod.rs
sgrif: Sorry, my previous example should have been `select(sql::<ST>("query")).load::<U>(&connection)`.

weiznich: select(...) will add a select statement  in front of the summited query. so this seems not to be a option here. One possibility would be to introduce a new pragma function, which adds pragma instead of select to the beginning of the query.

diesel_codegen/src/schema_inference/data_structures.rs
sgrif: Change the type to `Bool` in your `table!` call

diesel/src/sqlite/connection/mod.rs
sgrif: Good point.

sgrif: Can you move this out of the main Diesel crate, at least? I prefer not to add code there if it's only needed for codegen or CLI.

diesel_codegen/src/schema_inference/data_structures.rs
```diff
@@ -1,54 +1,65 @@
...
+#[cfg(all(feature = "postgres", not(feature = "sqlite")))]
+pub enum InferConnection{
+    Pg(PgConnection),
+}
+
+#[cfg(all(feature = "sqlite", feature = "postgres"))]
+pub enum InferConnection{
+    Sqlite(SqliteConnection),
+    Pg(PgConnection),
+}
```
sgrif: This is interesting. I wonder if we can pull this into a crate...

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -73,25 +76,16 @@ pub fn infer_schema_body<T: Iterator<Item=P<ast::Expr>>>(
...
-    cx: &mut ExtCtxt,
-    sp: Span,
-    database_url: &str,
-) -> Result<PgConnection, Box<MacResult>> {
-    PgConnection::establish(database_url).map_err(|_| {
-        cx.span_err(sp, "failed to establish a database connection");
-        DummyResult::any(sp)
-    })
-}
+
```
sgrif: :scissors:

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -73,25 +76,16 @@ pub fn infer_schema_body<T: Iterator<Item=P<ast::Expr>>>(
...
 fn table_macro_call(
     cx: &mut ExtCtxt,
     sp: Span,
-    connection: &PgConnection,
+    connection: &InferConnection,
     table_name: &str,
 ) -> Result<P<ast::Item>, Box<MacResult>> {
     match get_table_data(connection, table_name) {
-        Err(NotFound) => {
+        Err(::diesel::result::Error::NotFound) => {
```
sgrif: Why change this?

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -124,60 +118,143 @@ fn next_str_lit<T: Iterator<Item=P<ast::Expr>>>(
...
-    pg_class.select(oid).filter(relname.eq(table_name)).first(conn)
+
+// FIXME: Remove the duplicates of this function once expression level attributes
+// are stable (I believe this is in 1.7)
+#[cfg(all(feature = "sqlite", not(feature = "postgres")))]
+fn establish_connection(
+    cx: &mut ExtCtxt,
+    sp: Span,
+    database_url: &str,
+) -> Result<InferConnection, Box<MacResult>>{
```
sgrif: Please ensure that your code matches the style guide (spaces before all your braces)

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -124,60 +118,143 @@ fn next_str_lit<T: Iterator<Item=P<ast::Expr>>>(
...
 
-    if attr.nullable {
-        quote_ty!(cx, Nullable<$tpe>)
-    } else {
-        tpe
+
+#[cfg(all(feature = "sqlite", not(feature = "postgres")))]
+fn get_table_data(conn: &InferConnection, table_name: &str) -> QueryResult<Vec<ColumnInformation>>
+{
+    match *conn{
```
sgrif: I don't like having the enum change to single variant in these cases. We should just have it be a type alias.

diesel/src/sqlite/connection/mod.rs
weiznich: To move this out of the Diesel main crate I will need to expose raw_connection. This seems like a bad idea for me.
The alternative implementation of this would be to add a pragma function, that act like select but add `Pragma` instead (without any from part).

diesel_codegen/src/schema_inference/data_structures.rs
weiznich: Should be possible, but why? This are only a few lines of code. And it's not easily extensible for new database implementations from outside the crate.

diesel_codegen/src/schema_inference/mod.rs
weiznich: Because the compiler is telling me to do this. (Don't know where the other NotFound is coming from)
```
src/schema_inference/mod.rs:93:13: 93:21 warning: pattern binding `NotFound` is named the same as one of the variants of the type `diesel::result::Error` [E0170]
src/schema_inference/mod.rs:93         Err(NotFound) => {
                                           ^~~~~~~~
src/lib.rs:22:1: 22:23 note: in this expansion of include!
src/schema_inference/mod.rs:93:13: 93:21 help: run `rustc --explain E0170` to see a detailed explanation
src/schema_inference/mod.rs:93:13: 93:21 help: if you meant to match on a variant, consider making the path in the pattern qualified: `diesel::result::Error::NotFound`
src/schema_inference/mod.rs:97:9: 97:15 error: unreachable pattern [E0001]
src/schema_inference/mod.rs:97         Err(_) => {
```

weiznich:
Is there anything preventing this from being merged, except of time?

sgrif:
Been out of the country and still getting caught up. Haven't had a chance
to re-review yet. Will do today.

sgrif:
There were a lot of style changes that needed to be fixed, but I've gone ahead and updated manually. Normally I'd squash it into the previous commits to avoid git churn, but they were numerous enough to warrant keeping them separate. *Please* make an effort to avoid these sorts of changes in the future, we take git churn pretty seriously.

weiznich:
Thanks for the style fixes, but does the style guide not explicitly saying that brackets after multiline function definition needs to be  on the next line? Did I miss read something or should the guide being updated? Maybe someone should try to write a appropriate config for rustfmt?

Issue #183, opened 2016-02-03
Remove internal copy of Path#strip_prefix when 1.7 is stable


mcasper:
Specific issue for a FIXME per #173 

As of b1a215bd906039e2f4fa7e1af6259acd0f06129a, this FIXME is located here: https://github.com/sgrif/diesel/blob/master/diesel_cli/src/main.rs#L289

sgrif:
Permalink: https://github.com/sgrif/diesel/blob/b1a215b/diesel_cli/src/main.rs#L289

SoftwareWarlock:
Just to be clear, [`relative_from`](https://doc.rust-lang.org/std/path/struct.Path.html#method.relative_from) isn't being used because it's marked unstable (rust-lang/rust#23284), so once 1.7 is stable we should replace `strip_prefix` with [`Path::strip_prefix`](http://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.strip_prefix)?

mcasper:
Resolved by #230

PR #230, opened 2016-03-03
Removed internal copy of Path#strip_prefix

SoftwareWarlock:
This should pass after beta gets bumped to stable sometime today. Refer to #183


mcasper:
:+1: Thanks! We can hit the rebuild button and merge later today, like you said

SoftwareWarlock:
Good to go :+1:

PR #234, opened 2016-03-13
Add `#into_boxed` to all queries, for conditional changes

sgrif:
This is the first pass of what will be a much larger feature. The
intention is to support cases such as this:

```rust
let mut source = users::table.into_boxed();

if let Some(search) = params.search {
    source = source.filter(name.like(format!("%{}%", search)));
}

if let Some(order) = column_with_name(params.order) {
    source = source.order(order.desc());
}

let users = source.load(&connection);
```

Right now, this is basically impossible, as `source` needs to have the
same type, and each query in diesel has a unique type. This *will*
result in a slowdown of the query builder when used (though I think we
can still work to eliminate this cost through prepared statement
caching). However, this will not affect the performance of anything that
isn't using it.

As of this commit, we do not achieve the above use case, as
`BoxedSelectStatement` doesn't implement any of our query DSLs. The
intent is to have it implement all of them, but this felt like a good
place to cut a first pass.

This will fail to compile due to
https://github.com/rust-lang/rust/issues/32222. We can resolve this by
pointing at `nightly-2016-03-05` if needed, but hopefully that can be
resolved upstream in the next day or two.

### DIFF
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -18,7 +18,7 @@ pub mod insert_statement;
 pub mod update_statement;
 
 #[doc(hidden)]
-pub use self::select_statement::SelectStatement;
+pub use self::select_statement::{SelectStatement, BoxedSelectStatement};
 #[doc(inline)]
 pub use self::update_statement::{IncompleteUpdateStatement, AsChangeset, Changeset, UpdateTarget};
 #[doc(inline)]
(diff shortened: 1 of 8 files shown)


sgrif:
@mcasper @mfpiccolo Review?

@mcasper When finished, this feature should cover the use case we discussed.

mfpiccolo:
Nice feature @sgrif!  :+1:

mcasper:
Thanks so much for working on this @sgrif :clap:

PR #236, opened 2016-03-13
Print error and exit 1 instead of panicking in CLI

mcasper:
This was removed earlier because if a diesel_cli test started failing,
it would abort the entire test suite because it exited with a status
code of 1, which rust interprets as a problem. Now, because we shell out
all our tests that ever hit this part of the code, we can go back to
printing errors and exiting 1. This presentation is much nicer than a
panic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,10 @@ script:
 env:
   matrix:
     - BACKEND=sqlite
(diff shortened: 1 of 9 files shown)


sgrif:
:+1:

PR #242, opened 2016-03-22
Bump syntex and syntex_syntax to 0.30.0

### LINKED ISSUE #241: Increment Syntex Version to 0.30.0
Would it be possible to increment the syntex version to 0.30.0 so that I can use Diesel with serde_codegen?

mcasper:
Resolves #241


sgrif:
Should this be `>= 28.0.0, < 30.1.0`?

mcasper:
Yeah I think that'd work, I'll update this real quick

sgrif:
Cool. Was about to push the same, but if you can update this I'd like to see a green CI run. I've updated dotenv master

mcasper:
@sgrif We're good to go

PR #244, opened 2016-03-28
Allow omitting the return type from `sql_function!`

sgrif:
I've been debugging the build failures, and realized that I wanted this
functionality. Seems like an oversight not to have it. Forcing `-> ()`
is a pain.


sgrif:
Review? @mfpiccolo @mcasper

mcasper:
:+1:

PR #252, opened 2016-04-01
Fix #211: Allow other primary keys than id for infer_schema

### LINKED ISSUE #211: `infer_schema!` blows up on tables with primary keys other than `id`
Not only that, but the error message gives absolutely nothing useful.

```
<diesel macros>:22:19: 16:16 error: use of undeclared type name `columns::id` [E0412]
(internal compiler error: unprintable span)
<diesel macros>:15:1: 17:58 note: in this expansion of table_body! (defined in <diesel macros>)
<diesel macros>:5:1: 6:2 note: in this expansion of table! (defined in <diesel macros>)
<diesel macros>:2:1: 2:64 note: in this expansion of table! (defined in <diesel macros>)
src/schema.rs:2:1: 2:40 note: in this expansion of table! (defined in <diesel macros>)
```

We can't *fully* solve this without adding composite primary key support, but we should try to handle singular primary keys other than id, and either skip tables with composite primary keys or at least give a more useful warning/error.

weiznich:
* Implement detection of primary keys
* Print an error if there is a composite primary key
* If there is only one primary key, use this one

### DIFF
--- a/diesel_codegen/src/schema_inference/mod.rs
+++ b/diesel_codegen/src/schema_inference/mod.rs
@@ -92,15 +92,30 @@ fn table_macro_call(
             Err(DummyResult::any(sp))
         }
         Ok(data) => {
-            let tokens = data.iter().map(|a| column_def_tokens(cx, a, &connection))
-                .collect::<Vec<_>>();
-            let table_name = str_to_ident(table_name);
-            let item = quote_item!(cx, table! {
-                $table_name {
-                    $tokens
+            let primary_keys = match get_primary_keys(connection, table_name) {
+                Ok(keys) => keys,
+                Err(_) =>{
+                    cx.span_err(sp, "error loading schema");
+                    return Err(DummyResult::any(sp));
                 }
-            }).unwrap();
-            Ok(item)
+            };
+            if primary_keys.len() != 1 {
+                cx.span_err(sp,
+                    &format!("table {} has {} primary keys, only one is currently supported",
+                        table_name, primary_keys.len()));
+                Err(DummyResult::any(sp))
+            } else {
+                let tokens = data.iter().map(|a| column_def_tokens(cx, a, &connection))
+                    .collect::<Vec<_>>();
+                let table_name = str_to_ident(table_name);
+                let primary_key = str_to_ident(&primary_keys[0]);
+                let item = quote_item!(cx, table! {
+                    $table_name ($primary_key) {
+                        $tokens
+                    }
+                }).unwrap();
+                Ok(item)
+            }
         }
     }
 }
@@ -199,5 +214,14 @@ fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation, conn: &Infe
         InferConnection::Sqlite(_) => sqlite::determine_column_type(cx, attr),
         #[cfg(feature = "postgres")]
         InferConnection::Pg(_) => pg::determine_column_type(cx, attr),
+    }
+}
+
+fn get_primary_keys(conn: &InferConnection, table_name: &str) -> QueryResult<Vec<String>> {
+    match *conn {
+        #[cfg(feature = "sqlite")]
+        InferConnection::Sqlite(ref c) => sqlite::get_primary_keys(c, table_name),
+        #[cfg(feature = "postgres")]
+        InferConnection::Pg(ref c) => pg::get_primary_keys(c, table_name),
     }
 }
--- a/diesel_codegen/src/schema_inference/pg.rs
+++ b/diesel_codegen/src/schema_inference/pg.rs
@@ -31,6 +31,15 @@ select_column_workaround!(pg_attribute -> pg_type (attrelid, attname, atttypid, 
 select_column_workaround!(pg_type -> pg_attribute (oid, typname));
 
 table! {
+    pg_index (indrelid) {
+        indrelid -> Oid,
+        indexrelid -> Oid,
+        indkey -> Array<SmallInt>,
+        indisprimary -> Bool,
+    }
+}
+
+table! {
     pg_class (oid) {
         oid -> Oid,
         relname -> VarChar,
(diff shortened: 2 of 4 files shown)


diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -149,11 +165,14 @@ fn establish_connection(
 ) -> Result<InferConnection, Box<MacResult>> {
     match database_url {
         #[cfg(feature = "postgres")]
-        _ if database_url.starts_with("postgres://") || database_url.starts_with("postgresql://") => {
-            establish_real_connection(cx, sp, database_url).map(|c| InferConnection::Pg(c))
+        _ if database_url.starts_with("postgres://") ||
+            database_url.starts_with("postgresql://") => {
+                establish_real_connection(cx, sp, database_url)
+                    .map(|c| InferConnection::Pg(c))
```
sgrif: Why did you change the format of these lines?

diesel_codegen/src/schema_inference/pg.rs
```diff
@@ -16,19 +16,21 @@ table! {
...
         attisdropped -> Bool,
+        atttypmod -> Integer,
     }
 }
-
 table! {
-    pg_type (oid) {
-        oid -> Oid,
-        typname -> VarChar,
+    pg_index(indrelid){
```
sgrif: Please follow the style that was used before.

diesel_codegen/src/schema_inference/pg.rs
```diff
@@ -38,11 +40,11 @@ table! {
 }
 
 pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<ast::Ty> {
-    let tpe = if attr.type_name.starts_with("_") {
-        let subtype = str_to_ident(&capitalize(&attr.type_name[1..]));
+    let tpe = if attr.type_name.ends_with("[]") {
```
sgrif: What was the purpose of this change? There are array types that don't end with `[]`

weiznich: I've changed the query to get the column informations. Resulting of this the type names changed a bit. One of this changes results in a rename of the array type names. They don't start with _ anymore, but they seems to end with [] now.
Check this query
```sql
select oid::regtype, typname from pg_type where typname like '\_%';
```

sgrif: Is there a reason this needed to change?

weiznich: Informations relating to primary keys living in pg_index, informations about datatypes in pg_types and column related stuff in pg_attributes.
I couldn't get more than one join working, so I decided to rewrite the data type part of the query to use a cast, instead a join to get name of the data type. Unfortunately this results in this changes.

sgrif: Seems like we could just do a separate query to get the primary key name? (Which is what we ultimately need). This would result in much less code changing to support it.

```rust
fn primary_key_name(conn: &PgConnection, table_oid: u32) -> Vec<String> {
    pg_index
        .inner_join(pg_attribute)
        .select(attname)
        .filter(indisprimary.eq(true))
        .filter(attrelid.eq(table_oid))
        .load(&connection)
}
```

weiznich: I considered first to implement this using 2 queries, but decided against it. My concerns with this solution are, that you will need to build the ColumnInformation struct by hand.

If you don't like the current implementation, I can change this back.
Maybe the best solution would be to get this working with two joins.

sgrif: I don't think this needs to be on the `ColumnInformation` struct. Ultimately we pass the primary key name independently of where that struct is used.

Two joins is high on the list, but it would be incredibly painful to use until some changes land in Rust or I figure out a workaround.

weiznich: Returning the primary keys independently would be a way, but then bigger changes on sqlite side would be required. I will look into this tomorrow.

diesel_codegen/src/schema_inference/mod.rs
```diff
@@ -92,15 +92,24 @@ fn table_macro_call(
...
+                    &format!("table {} has {} primary keys, only one is currently supported",
+                        table_name, primary_keys.len()));
+                Err(DummyResult::any(sp))
+            } else {
+                let tokens = data.iter().map(|a| column_def_tokens(cx, a, &connection))
+                    .collect::<Vec<_>>();
+                let table_name = str_to_ident(table_name);
+                let primary_key = str_to_ident(&primary_keys[0]);
+                let item = quote_item!(cx, table! {
+                    $table_name($primary_key) {
```
sgrif: `$table_name ($primary_key)` (missing space)

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -62,3 +62,12 @@ pub fn load_table_names(
         .filter(sql::<types::Bool>("type='table' AND name NOT LIKE '\\_\\_%'"));
     query.load(connection)
 }
+
+pub fn get_primary_keys(conn: &SqliteConnection, table_name: &str) -> Vec<String>{
+    conn.execute_pragma::<pragma_table_info::SqlType, (i32, String, String, bool, Option<String>, bool)>(
+        &format!("PRAGMA TABLE_INFO('{}')", table_name))
+        .unwrap()
+        .iter()
+        .filter_map(|i| if i.5 { Some(i.1.clone()) } else { None })
```
sgrif: Maybe put this into a struct?

diesel_codegen/src/schema_inference/pg.rs
```diff
@@ -74,17 +83,28 @@ pub fn load_table_names(
...
+
+    let pk_query = pg_index.select(indexrelid)
+        .filter(sql(&format!("pg_index.indrelid = '{}'::regclass",table_name)))
+        .filter(indisprimary.eq(true));
+
+    pg_attribute.select(attname)
+        .filter(attrelid.eq_any(pk_query))
+        .order(attnum)
+        .load(conn)
+        .unwrap()
```
sgrif: Can we avoid using `unwrap` here?

diesel_codegen/src/schema_inference/pg.rs
```diff
@@ -74,17 +83,28 @@ pub fn load_table_names(
...
-fn table_oid(conn: &PgConnection, table_name: &str) -> QueryResult<u32> {
-    use self::pg_class::dsl::*;
-    pg_class.select(oid).filter(relname.eq(table_name)).first(conn)
+pub fn get_primary_keys(conn: &PgConnection, table_name: &str) -> Vec<String> {
+    use self::pg_attribute::dsl::*;
+    use self::pg_index::dsl::{pg_index, indisprimary, indexrelid};
+    use diesel::expression::sql_literal::sql;
+
+    let pk_query = pg_index.select(indexrelid)
+        .filter(sql(&format!("pg_index.indrelid = '{}'::regclass",table_name)))
```
sgrif: Why the cast instead of the `table_oid` function that was there before? (I'd prefer to avoid literal SQL as it's unchecked by the compiler)

weiznich: I removed the raw sql, but inlined the table_oid query as subquery.

sgrif:
@weiznich Is this ready to go?

weiznich:
Yes

sgrif:
Thank you for working on this!

PR #255, opened 2016-04-02
Add support for the `DISTINCT` keyword

### LINKED ISSUE #253: `DISTINCT` support
There isn't currently support for creating SELECT DISTINCT queries.

sgrif:
A few things to note about the implementation. I've opted for 2 unit
structs instead of something like an enum or boolean so that the code
gets inlined via monomorphisation instead of containing a branch (e.g.
if you don't use distinct, it's free). As always, it's enforced that our
query builder doesn't do "real work" by
https://github.com/sgrif/diesel/blob/c8127d408f759faa056503014cd00e92479215e0/diesel_tests/tests/perf_details.rs#L8-L13

This ended up churning a lot of `select_dsl`. I've reformatted the code
surrounding what was touched to reduce churn in the future.

This doesn't yet add support for `DISTINCT ON(expr)`, as that is a PG
specific feature and can be added as a separate PR.

An alternative API for this would be something like
`users.select(distinct(name))`. I actually think I might prefer that
API, but actually enforcing correctness with it (making sure `distinct`
can only be passed to `select` and not something like `filter` --
disallowing things like `max(distinct(id))`, etc) would be pretty much
impossible.

Fixes #253.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * The maximum table size can be increased to 52 by enabling the `huge-tables`
   feature. This feature will substantially increase compile times.
 
+* The `DISTINCT` keyword can now be added to queries via the `distinct()`
+  method.
+
 ### Changed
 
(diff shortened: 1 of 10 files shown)


sgrif:
Review? @mfpiccolo @mcasper

PR #256, opened 2016-04-02
Implement `FilterDsl` for BoxedSelectStatement

sgrif:
We had to switch the type of the where clause to an option, and handle
whether or not to put `WHERE` in the query explicitly to handle the
`AND` case when filter is called twice. I had tried to instead box up
`WhereAnd`, but ultimately if you try and write
`impl WhereAnd<Box<QueryFragment<DB>>>`, the output type ends up being
infinitely recursive, as the output type would be a `Box<WhereAnd>`, and
would need the output type specified, etc.

### DIFF
--- a/diesel/src/query_builder/select_statement/boxed.rs
+++ b/diesel/src/query_builder/select_statement/boxed.rs
@@ -1,16 +1,16 @@
 use std::marker::PhantomData;
 
 use backend::Backend;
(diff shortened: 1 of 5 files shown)


sgrif:
Review? @mfpiccolo @mcasper

mcasper:
Looks good!

PR #257, opened 2016-04-02
Remove `no select` from `table!`

sgrif:
This was something that I added to support crates.io, as their crates
table has a column that's only used for querying but not ever actually
selected. Ultimately this just doesn't fit with Diesel's overall
philosophy. It was poorly documented and untested. Just set up an `all`
function that calls `select` if you need this.


sgrif:
Review? @mcasper @mfpiccolo

mcasper:
:+1:

PR #258, opened 2016-04-03
Allow users to opt into usage of specialization

sgrif:
This is a change that I want to make eventually just to clean up our
code. However, there is a concrete case that can't wait for
specialization to land. There is not way for a user to implement
`FromSqlRow` for a nullable field of a custom type (e.g. a PG enum).

Using diesel with the `unstable` feature now uses specialization, which
we can start using elsewhere. We unfortunately still won't be able to
clean up the code base until specialization is stable, but we can at
least start to look at what the code *might* look like.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -2,6 +2,7 @@
 //! interactions. [A getting started guide](https://github.com/sgrif/diesel#getting-started) can be
 //! found in the README.
 #![deny(warnings)]
+#![cfg_attr(feature = "unstable", feature(specialization))]
 pub mod backend;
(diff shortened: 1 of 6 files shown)


sgrif:
Review? @mcasper @mfpiccolo 

/cc @LordPython

sgrif:
Ping @mcasper @mfpiccolo I would like to roll a release soon with this PR in it.

mcasper:
👍

mfpiccolo:
This looks good to me.  Way to be on top of Rust rolling out new features!  Looking forward to 1.9.

👍

PR #259, opened 2016-04-06
Implement `Send` for `SqliteConnection`

### LINKED ISSUE #228: implement `Send` for SqliteConnection
Why is this not the case? It seems like it'd be ok given the reasons outlined in #190, but I admit that I'm not that familiar with the matter.

Use case: I'm working on a web application where I'd like to be able to share a SqliteConnection between multiple requests using a combination of [iron's](ironframework.io) [persistent](https://github.com/iron/persistent) crate and r2d2-diesel.

sgrif:
After spending some time poking at it, I can't find any reason this
wouldn't be safe. There's a legitimate use case for this where you want
to share a connection to an in memory database (wrapped in a mutex of
course)

Fixes #228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * The `DISTINCT` keyword can now be added to queries via the `distinct()`
   method.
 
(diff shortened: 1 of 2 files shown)


sgrif:
Review? @mfpiccolo @mcasper

mcasper:
👍

PR #262, opened 2016-04-09
Properly generate a single element tuple for 1 column tables

### LINKED ISSUE #261: Tables with a single column aren't queryable
Reported via gitter

```
table! {
    users {
        id -> Integer,
    }
}

#[derive(Queryable)]
pub struct User {
    pub id: i32,
}

users.load::<User>(&connection).expect("Error loading users");
```

sgrif:
Our code was generating `(id)` and not `(id,)`. Without the trailing
comma, the parenthesis are meaningless. This meant that you couldn't
actually load the query into a struct without manually selecting the
tuple.

While this fixes the bug, I think it probably also makes sense for us to
have a `FromSqlRow` and `Queryable` impl for the non-tuple type into a
single element tuple/struct.

Fixes #261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * `#[insertable_into]` can now be used with structs that have lifetimes with
   names other than `'a'`.
 
+* Tables with a single column now properly return a single element tuple. E.g.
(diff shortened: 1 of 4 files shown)


sgrif:
Review? @mcasper @mfpiccolo

sgrif:
Build will fail as master still needs a syntex update.

mfpiccolo:
👍 

Agree with the single element tuple/struct as well.

PR #266, opened 2016-04-13
Change bind collection to be a 2-pass process

sgrif:
This is primarily a structural change, separated out from some larger changes that I'd like to make. However, I wanted to keep the PR size reasonable.

This separates out construction of bind parameters from the construction of the SQL. This will ease implementation of prepared statement caching, and will be required for any implementation that uses the type identifier instead of constructing the SQL.

This will also enable future refactorings to bring `ToSql` closer to the actual process occurring, letting us eliminate the hacky code we have in SQLite right now, and lower the number of allocations for all backends.

Finally, this will all potentially enable us to actually show the values of the bind params in the debug query builder as long as the value of the bind param is `Debug`.

The bind param collector is generic over the backend because I plan to use a different struct for SQLite (which will eventually have a different interface). Additionally, the signature is `QueryResult<()>` instead of our usual `Box<Error>` because my eventual intention is to have the sqlite3 bind methods get called in this same function, which can fail in ways that we'd like to represent.

I do expect that the `RawBytesBindCollector` will be useable for every type except for SQLite, and have structured it so that code can be shared by other backends in the future

Walking the AST twice does not introduce any performance regressions, as the actual cost of this gets eliminated by monomorphisation and inlining. In fact, our benchmarks actually show a moderate improvement with this change. It's possible that the removal of the bind param lines from the query construction is allowing the optimizer to better operate on `to_sql`.

Through implementing this, `Changeset` and `InsertValues` stood out to me as places where I'd like to remove the "special `to_sql`". I explored this, but once again ran into the fact that if I had them return a `QueryFragment`, it wouldn't work as the type needs to change conditionally for both of those. I will continue to explore potential improvements in the future, as I didn't like that those had to change as a result of the signature of a (semi-)unrelated trait changing.

### DIFF
--- a/diesel/src/backend.rs
+++ b/diesel/src/backend.rs
@@ -1,4 +1,4 @@
-use query_builder::QueryBuilder;
+use query_builder::{QueryBuilder, BindCollector};
 use query_builder::debug::DebugQueryBuilder;
 use types::{self, HasSqlType};
 
@@ -17,6 +17,7 @@ pub trait Backend where
     Self: HasSqlType<types::Timestamp>,
 {
     type QueryBuilder: QueryBuilder<Self>;
+    type BindCollector: BindCollector<Self>;
     type RawValue: ?Sized;
 }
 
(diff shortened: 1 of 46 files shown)


sgrif:

sgrif:
Any objections to merging this?

samphippen:
I like the idea but I feel like I don't fully get it. Happy to do a screen-share if you'd like to walk me through it.

samphippen:
I mean, I have no objections, but I feel like I'm not the best judge.

PR #267, opened 2016-04-13
Replace the implementation specific NOW with something more generic

sgrif:
`CURRENT_TIMESTAMP` is part of the ANSI SQL standard. We can use this on
all backends, instead of our current behavior of `NOW()` on everything
but SQLite and `'now'` on SQLite


sgrif:

samphippen:
:+1:

PR #269, opened 2016-04-15
Add the `escape` method to `Like` and `NotLike`

sgrif:
This is functionality I need to fix a bug in diesel_codegen related to
`infer_schema!` on SQLite. It felt appropriate to expose to the public
API. `name.like("foo\\%").escape('\')` will generate `name LIKE 'foo\%'
ESCAPE '\'` (with identifiers quoted and bind params for the strings).

While `ESCAPE` isn't part of the ANSI standard, it appears to be
supported by every backend, so I've left it unconstrained.


sgrif:

samphippen:
LGTM

mfpiccolo:
`.filter(name.like("%😀%%").escape('😀'))`

Awesome!

mcasper:
Finally, emoji support!

PR #271, opened 2016-04-15
Implement `infer_schema!` that actually works for SQLite

sgrif:
I really dropped the ball on this one. I never should have merged in a
PR with such poor test coverage. As it turns out, this feature doesn't
even come close to working.

The `load_table_names` function did not specify the escape character for
the `LIKE` clause, meaning it was including Diesel's metadata tables.
Additionally, SQLite has several internal tables that it creates, which
also need to be excluded from schema inference.

Additionally, the types that it mapped to were far too broad, and would
lead to columns being unable to perform certain actions. Even if SQLite
treats `Text` and `VarChar` the same under the hood, we treat them
semantically differently and need to respect the users' schema.

I also learned that `PRIMARY KEY AUTOINCREMENT` doesn't imply `NOT NULL`
on SQLite. So that's fun.

I have changed our integration tests to make use of this feature, to
ensure it is working in the future. I had hoped to eliminate the
`backend_specific_schema` module entirely, but was unable to due to
https://github.com/rust-lang/rust/issues/31369


sgrif:
Review? @mfpiccolo @mcasper @samphippen 

I'd like to roll a release with this fix ASAP

PR #272, opened 2016-04-15
Add a logging Connection

weiznich:
- Add a Connectionwrapper that logs each executed sql statement
- Use the native Querybuilder to build the sql queries
- Make logging optional even if DebugConnection is used

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -17,6 +17,7 @@ byteorder = "0.3.*"
 quickcheck = { version = "0.2.25", optional = true }
 chrono = { version = "^0.2.17", optional = true }
 uuid = { version = "^0.2.0", optional = true, features = ["use_std"] }
+log = { version = "0.3", optional = true }
 
 [dev-dependencies]
 quickcheck = "0.2.25"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+
 language: rust
 sudo: required
 dist: trusty
@@ -28,8 +29,10 @@ script:
   fi &&
   if [[ "$TRAVIS_RUST_VERSION" == nightly* ]]; then
     (cd diesel_tests && travis-cargo test -- --no-default-features --features "unstable_$BACKEND")
+    (cd diesel_tests && RUST_TEST_THREADS=1 travis-cargo test -- --no-default-features --features "unstable_$BACKEND debug_$BACKEND")
   else
     (cd diesel_tests && travis-cargo test -- --no-default-features --features "stable_$BACKEND")
+    (cd diesel_tests && RUST_TEST_THREADS=1 travis-cargo test -- --no-default-features --features "stable_$BACKEND debug_$BACKEND")
   fi &&
   if [[ "$TRAVIS_RUST_VERSION" == nightly* ]]; then
     (cd diesel_compile_tests && travis-cargo test)
--- a/bin/test
+++ b/bin/test
@@ -13,6 +13,8 @@ else
   (cd diesel_cli && cargo test --features "sqlite" --no-default-features)
   (cd diesel_codegen_syntex && cargo test --no-default-features --features "postgres")
   (cd diesel_tests && cargo test --features "unstable_postgres" --no-default-features)
+  (cd diesel_tests && RUST_TEST_THREADS=1 cargo test --features "unstable_postgres debug_postgres" --no-default-features)
   (cd diesel_tests && DATABASE_URL=/tmp/test.db cargo test --features "unstable_sqlite" --no-default-features)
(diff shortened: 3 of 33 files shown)


sgrif:
This breaks the build. We need this to be pretty extensively tested before we can review it further. In addition to tests that demonstrate logging works at all, we should have tests making sure that backend specific features can be used with the logging connection, and compile-fail tests to show that features for a different backend can't be used (e.g. if this is wrapping a SqliteConnection, we need to ensure that you can't use PG specific features)

diesel/Cargo.toml
```diff
@@ -17,6 +17,7 @@ byteorder = "0.3.*"
 quickcheck = { version = "0.2.25", optional = true }
 chrono = { version = "^0.2.17", optional = true }
 uuid = { version = "^0.2.0", optional = true, features = ["use_std"] }
+log = { version = "0.3", optional = true }
```
sgrif: I don't think we should use the `log` crate here. It looks like it only works with a global logger, which isn't a route I'd like to go down. Since everything would be the same log level anyway, I don't see much reason to take anything other than a `W: Write`.

weiznich: You should be able to filter the logging statement on crate basis. For example if someone uses [env_logger](https://github.com/rust-lang/log) as logging backend one could set RUST_LOG=diesel=debug to output all debugging statements from diesel with a log level of debug or higher. Also you could implement your own backend to filter on custom rules.
If you don't like this approach I will change this as suggested. I've choosen the `log` crate because it seems to be the standard for such things and seems to be greatly configurable form the users of this library.

weiznich:
I will fix the testbuild as soon as I have some spare time. I've manly opened this pull request to start talking about this. (Maybe I should  have mention this, when I've opened this request). Also this is missing a strategy to log the binded parameters.

What do you exactly mean with backend specific features? Queries using parts that are only available with a specific backend? Or custom functions implemented on a specific backend. The former should work with the current implementation, as it passing everything down to the specific backend in the end. The later wont work in the current state. (Is this needed at some point?)

Regarding to compile tests: There is already an doctest, testing the usage of PgConnection or SqliteConnection as inner connection. So if this one compiles I think there are only tests needed to check the functionality (Are there are some tests for PgConnection/SqliteConnection?)

diesel/Cargo.toml
sgrif: @mfpiccolo @mcasper Thoughts?

sgrif:
> Queries using parts that are only available with a specific backend?

Yes. For example, `DebugConnection<SqliteConnection>` should not be able to use `Any` for example, while `DebugConnection<PgConnection>` should.

> The former should work with the current implementation, as it passing everything down to the specific backend in the end

Cool. Need tests for it though, and most importantly we need tests that trying to pass a query containing an `Any` or similar to `DebugConnection<SqliteConnection>` fails to compile.

> I think there are only tests needed to check the functionality

Yes we should test the functionality. ;)

> Are there are some tests for PgConnection/SqliteConnection?

It's not so much about testing the connection as it is about testing the behavior of the entire system. That is what our integration tests are for. We prefer to be fairly exhaustive in our testing.

diesel/Cargo.toml
sgrif: If we do go forward with using the log crate for this, I wonder if it would make sense for the other connections just to call those macros directly instead of a separate struct.

weiznich:
I've updated the pull request to work with the newly introduced caching mechanisms. 

Also I added some tests:
* All tests in diesel_tests are run against the normal connection and the debug connection
* I added 2 test to check if the logged query is right. To verify those results tests with the debugging connection cannot run in parallel, because in that case there might be also a log of a other test case in there. (If more tests are needed here, it should be fairly simple to add them) 
* pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs also checks now against the debug_connection 

What`s left:
* Add some way to log the query parameters
* check the code style

weiznich:
Any comments on this?

sgrif:
I'm in the middle of moving to another country for the next few weeks. I'm
pretty unavailable, but I'll see if anyone else in the core team has
comments.

weiznich:
Take your time. There is no need to have this merged in the next few days.

weiznich:
Any news?

sgrif:
Probably going to include this in 0.8, will re-review soon once 0.7 ships

sgrif:
I'm going to go ahead and close this, as it would need quite a bit of work to bring up to date. I really wanted to have logging be a separated concern from everything else, but I don't think that's feasible as we've encouraged people to write `&PgConnection` and not `C: Connection<Backend=Pg>(_: &C)`. So ultimately this solution will probably need to have a `set_logger` method on connection. It'll also need to integrate with prepared statement caching, since we generally don't actually generate the SQL string anymore.

PR #274, opened 2016-04-15
Make diesel_codegen 0.6 depend on diesel 0.6.

### LINKED ISSUE #273: diesel_codegen depends on old version of diesel
I think somebody forgot to bump the version of the `diesel` dependency in `diesel_codegen` when releasing version 0.6, since it currently depends on 0.5.0. I'm not sure what the effects of such a thing would be, but probably nothing good and possibly confusing errors.

tari:
Fixes #273. I don't think this is a breaking change, since codegen doesn't export any types from diesel that are used at runtime.

This allows `infer_schema!` to work with `bytea` columns at the least, and might fix other issues.

### DIFF
--- a/diesel_codegen/Cargo.toml
+++ b/diesel_codegen/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diesel_codegen"
-version = "0.6.0"
+version = "0.6.1"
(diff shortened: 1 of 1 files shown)


sgrif:
Nice catch. Thanks!

sgrif:
Also usually you shouldn't bump crate versions in PRs. We will bump when actually releasing.

PR #276, opened 2016-04-16
Allow migrations to be embedded in the resulting binary

sgrif:
This is an alternate implementation of #199. That PR was on the
right track, and the work done so far is very much appreciated (and
provided some basis for this PR). However, the lack of tests,
documentation, and changelog entry as well as the remaining issues with
both implementation and code style ultimately meant it was just faster
to write out an implementation I was happier with.

The goal of this feature is to better enable usage of Diesel in embedded
applications. If you're distributing the binary to an environment you
don't control, you can't guarantee that the migrations will actually be
at the right location in the file system.

The `embed_migrations` macro will look for migrations in the directory
specified (or default to the same search CLI does if left unspecified),
read in the migration files, and exposes a module to allow you to run
them later. The embedded migrations are entirely constant, and won't
allocate when run any more than executing the SQL string directly would
have.

The generated code as it appears in our test suite can be found at
https://gist.github.com/sgrif/975bfd8418cde8766ab82ac001f1cd84

We do not read from `down.sql`, nor does the resulting struct store it,
as reverting a migration doesn't make sense for the use case that this
is targeting. Outside of development, you would always make a new
migration to change schema. The `unimplemented!()` would never result in
a runtime error, as the struct and constant are not public, and could
never be accessed other than through the `run` functions that we expose.

The generated module is private and called `embedded_migrations` for
considerations with anybody who is doing `use diesel::*` and/or `use
schema::*`. The idea is that this module would be generated close to
where it's being used. For those who do want a public module, it's quite
easy to write `pub use embedded_migrations as migrations`.

While this is redundant with our current "runtime" migration support
that exists in `run_pending_migrations`, I don't think it eliminates
their use case entirely. You can't use codegen to build a build script,
and I think automatically running migrations automatically in `build.rs`
makes complete sense, especially for tests, to ensure that the schema is
up to date before `infer_schema!` is evaluated. I could be convinced
that the CLI handles this enough, though.

Close #199

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,7 +1,9 @@
 #!/bin/sh
 set -e
 
-if [ "$1" == "integration" ]; then
+if [ "$1" == "integration" ] && [ "$2" == "sqlite" ]; then
+  (cd diesel_tests && DATABASE_URL=/tmp/test.db cargo test --features "unstable sqlite" --no-default-features)
+elif [ "$1" == "integration" ]; then
   (cd diesel_tests && cargo test --features "unstable postgres" --no-default-features)
 elif [ "$1" == "compile" ]; then
   (cd diesel_compile_tests && cargo test)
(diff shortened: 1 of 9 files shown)


sgrif:

samphippen:
LGTM.

PR #277, opened 2016-04-17
Reform types inferred by `infer_schema!` on SQLite

sgrif:
`infer_schema!` is woefully undertested. Really our tests for it at the
moment are "we use it for our test suite so the cases our suite covers
work". However, even though I defined `tinyint` == `Bool`, I wanted to
make sure that `TINYINT(1)` was treated as bool as well, as I wasn't
certain how SQLite handles limit/precision/scale.

The answer is that it doesn't, and it's way looser about allowed type
names than I had thought. The process listed at
https://www.sqlite.org/datatype3.html is a literal description, and any
possible string is a valid type.

This adds tests for every example given on that page, plus a few extras.
We create a table with as many of these fields as possible, and do a
trivial roundtrip to make sure that it *actually* infers something we
can deserialize from, and that we're not doing anything dumb.

The new logic for type inference matches pretty closely to how SQLite
handles things with a few exceptions:

- "boolean" or types containing "tiny" and "int" are treated as bool
- smallint and bigint are separated from int
- float is separated from double
- varchar is separated from text
- We do not accept random unrecognized type names as numeric

Unresolved Questions
--------------------

This actually starts to make me a bit more nervous about our semantics
with SQLite. If you're just using Diesel, everything is fine. However,
you can definitely insert values that would fail to deserialize with so
little constraints on the backend. I'm starting to wonder if we should
truly embrace SQLite's definitions and map exactly to that, allowing
only the following types:

- BigInt
- VarChar (yes, it's the ANSI text type but we treat VarChar as the
  "default" string type)
- Binary
- Double

We're omitting numeric, as there's no observable difference in SQLite
between the real affinity and the numeric affinity.

This would have several *major* implications. Aside from not being able
to use basic things like an `i32`, it would also mean that there is no
boolean type, and no dates/times/datetimes. Functions for those do exist
on the SQLite side though, so some of the interactions might get super
janky.

That said, Diesel's goal is not to abstract away the backend. These are
the semantics of the backend chosen, and maybe we should go whole hog
and embrace them.

I'm still unsure. In the meantime, with our current semantics, this
should improve the reliability of `infer_schema!`

### DIFF
--- a/diesel_codegen/src/schema_inference/sqlite.rs
+++ b/diesel_codegen/src/schema_inference/sqlite.rs
@@ -26,29 +26,29 @@ pub fn get_table_data(conn: &SqliteConnection, table_name: &str)
     sql::<pragma_table_info::SqlType>(&query).load(conn)
 }
 
-fn is_text(type_name: &str) -> bool {
-    type_name.contains("clob") ||
-        type_name.contains("text")
-}
-
 pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<ast::Ty> {
     let type_name = attr.type_name.to_lowercase();
-    let tpe = match &*type_name {
-        "tinyint" => quote_ty!(cx, ::diesel::types::Bool),
-        "smallint" | "int2" => quote_ty!(cx, ::diesel::types::SmallInt),
-        "int" | "integer" | "int4" => quote_ty!(cx, ::diesel::types::Integer),
-        "bigint" => quote_ty!(cx, ::diesel::types::BigInt),
-        _ if type_name.contains("char") => quote_ty!(cx, ::diesel::types::VarChar),
-        _ if is_text(&type_name) => quote_ty!(cx, ::diesel::types::Text),
-        _ if type_name.contains("blob") || type_name.is_empty() => {
-            quote_ty!(cx, ::diesel::types::Binary)
-        }
-        "float" => quote_ty!(cx, ::diesel::types::Float),
-        "double" | "real" | "double precision" => quote_ty!(cx, ::diesel::types::Double),
-        _ => {
-            cx.span_err(cx.original_span(), &format!("Unsupported type: {}", type_name));
-            quote_ty!(cx, ())
-        }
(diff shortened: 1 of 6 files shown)


sgrif:
Review? @diesel-rs/core

sgrif:
Review? @diesel-rs/core

sgrif:
@samphippen @mfpiccolo @mcasper I have no clue if that last comment pinged

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -26,29 +26,29 @@ pub fn get_table_data(conn: &SqliteConnection, table_name: &str)
...
+        quote_ty!(cx, ::diesel::types::Text)
+    } else if type_name.contains("blob") || type_name.is_empty() {
+        quote_ty!(cx, ::diesel::types::Binary)
+    } else if is_float(&type_name) {
+        quote_ty!(cx, ::diesel::types::Float)
+    } else if is_double(&type_name) {
+        quote_ty!(cx, ::diesel::types::Double)
+    } else {
+        cx.span_err(cx.original_span(), &format!("Unsupported type: {}", type_name));
+        quote_ty!(cx, ())
```
samphippen: this structure is much more clear than the previous structure. I like it.

diesel_codegen/src/schema_inference/sqlite.rs
```diff
@@ -58,6 +58,40 @@ pub fn determine_column_type(cx: &mut ExtCtxt, attr: &ColumnInformation) -> P<as
...
 }
 
+fn is_text(type_name: &str) -> bool {
+    type_name.contains("clob") ||
+        type_name.contains("text")
+}
+
+fn is_bool(type_name: &str) -> bool {
+    type_name == "boolean" ||
+        type_name.contains("tiny") &&
```
samphippen: I am bad at remembering boolean binding rules, do you think wrapping the ampersand expressions in parentheses would aid readability here?

samphippen:
Few comments. Would like thoughts, but they're not desperate merge blockers.

PR #279, opened 2016-04-17
Add prepared statement caching for SQLite

sgrif:
This is the first pass at adding prepared statement caching to Diesel.
The query builder will still run, as we are using the SQL string as the
key. Eventually this will be replaced with a type identifier, but we're
not there yet.

Most of the groundwork is laid to apply this to PG as well, but to start
it only applies to SQLite. The cache is placed at the top of the struct
instead of the end to ensure that the statements are dropped before the
raw connection. I'm still trying to determine if the order that fields
are dropped is something we can rely on. We may need to come up with an
alternate implementation for that part of this.

The prepared statements are stored in a hash map, and not some sort of
LRU cache. This means that the memory growth here could be unbounded if
left unchecked. In practice though, there is an upper bound as long as
we exclude certain types of queries from the cache. In particular, we
need to exclude queries that can generate different SQL without the type
changing. Therefore, we apply the following rules:

- ASTs containing the `Many` node are not cached (`WHERE foo IN (?, ?,
  ?)` type queries)
- ASTs containing a `SqlLiteral` node are not cached
- Insert statements are not cached
- Update statements are not cached

There's an argument for allowing `SqlLiteral` nodes to be cached, but I
am erring on the side of caution here. We can't tell if you're
dynamically generating the SQL in that node or not, and allowing them to
be cached would require exposing controls to care about internals that I
want to keep internal for the time being.

This results in a healthy performance improvement across the board.

Before
------

```
test bench_selecting_0_rows_with_medium_complex_query         ... bench:      43,125 ns/iter (+/- 12,545)
test bench_selecting_0_rows_with_trivial_query                ... bench:       5,755 ns/iter (+/- 2,063)
test bench_selecting_10k_rows_with_medium_complex_query       ... bench:  22,090,236 ns/iter (+/- 7,078,394)
test bench_selecting_10k_rows_with_medium_complex_query_boxed ... bench:  20,322,178 ns/iter (+/- 5,829,031)
test bench_selecting_10k_rows_with_trivial_query              ... bench:   2,980,795 ns/iter (+/- 1,046,410)
test bench_selecting_10k_rows_with_trivial_query_boxed        ... bench:   2,894,526 ns/iter (+/- 829,044)
```

After
-----

```
test bench_selecting_0_rows_with_medium_complex_query         ... bench:      16,414 ns/iter (+/- 4,357)
test bench_selecting_0_rows_with_trivial_query                ... bench:       1,187 ns/iter (+/- 296)
test bench_selecting_10k_rows_with_medium_complex_query       ... bench:  19,490,094 ns/iter (+/- 4,403,771)
test bench_selecting_10k_rows_with_medium_complex_query_boxed ... bench:  18,980,924 ns/iter (+/- 3,537,891)
test bench_selecting_10k_rows_with_trivial_query              ... bench:   2,873,268 ns/iter (+/- 1,174,266)
test bench_selecting_10k_rows_with_trivial_query_boxed        ... bench:   2,891,690 ns/iter (+/- 1,012,551)
```

We can probably structure this a little bit more "unsafely" to eliminate
the overhead of `Rc` and `RefCell`. A statement will never have more
than one reference at a time (and this is enforced by `RefCell`, since
basically every method on `Statement` requires `&mut self`). We could in
theory just "copy" the `Statement` struct instead. The code is
encapsulated to allow this, but the overhead of `Rc<RefCell<>>` isn't
high enough to justify it.

### DIFF
--- a/diesel/src/query_builder/delete_statement.rs
+++ b/diesel/src/query_builder/delete_statement.rs
@@ -34,5 +34,10 @@ impl<T, DB> QueryFragment<DB> for DeleteStatement<T> where
         }
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.0.from_clause().is_safe_to_cache_prepared() &&
+            self.0.where_clause().map(|w| w.is_safe_to_cache_prepared()).unwrap_or(true)
+    }
 }
 
--- a/diesel/src/query_builder/select_statement/boxed.rs
+++ b/diesel/src/query_builder/select_statement/boxed.rs
@@ -95,6 +95,17 @@ impl<'a, ST, QS, DB> QueryFragment<DB> for BoxedSelectStatement<'a, ST, QS, DB> 
         try!(self.offset.collect_binds(out));
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.distinct.is_safe_to_cache_prepared() &&
+            self.select.is_safe_to_cache_prepared() &&
+            self.from.from_clause().is_safe_to_cache_prepared() &&
+            self.where_clause.as_ref().map(|w| w.is_safe_to_cache_prepared())
+                .unwrap_or(true) &&
+            self.order.is_safe_to_cache_prepared() &&
+            self.limit.is_safe_to_cache_prepared() &&
+            self.offset.is_safe_to_cache_prepared()
+    }
 }
 
 impl<'a, ST, QS, DB, Type, Selection> SelectDsl<Selection, Type>
--- a/diesel/src/expression/aliased.rs
+++ b/diesel/src/expression/aliased.rs
@@ -39,6 +39,10 @@ impl<'a, T, DB> QueryFragment<DB> for Aliased<'a, T> where
     fn collect_binds(&self, _out: &mut DB::BindCollector) -> QueryResult<()> {
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        true
+    }
 }
 
 // FIXME This is incorrect, should only be selectable from WithQuerySource
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -55,6 +55,11 @@ impl<T, U, DB> QueryFragment<DB> for In<T, U> where
         try!(self.values.collect_binds(out));
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.left.is_safe_to_cache_prepared() &&
+            self.values.is_safe_to_cache_prepared()
+    }
 }
 
 use std::marker::PhantomData;
@@ -121,6 +126,10 @@ impl<T, DB> QueryFragment<DB> for Many<T> where
         }
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        false
+    }
 }
 
 pub struct Subselect<T, ST> {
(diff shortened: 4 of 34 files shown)


sgrif:

diesel/src/query_builder/delete_statement.rs
```diff
@@ -34,5 +34,10 @@ impl<T, DB> QueryFragment<DB> for DeleteStatement<T> where
         }
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.0.from_clause().is_safe_to_cache_prepared() &&
+            self.0.where_clause().map(|w| w.is_safe_to_cache_prepared()).unwrap_or(true)
```
samphippen: This stands out to me as a little odd, I would expect that if we fail to unwrap, we want false, but I suspect I'm just ignorant?

diesel/src/query_builder/select_statement/boxed.rs
```diff
@@ -95,6 +95,17 @@ impl<'a, ST, QS, DB> QueryFragment<DB> for BoxedSelectStatement<'a, ST, QS, DB>
         try!(self.offset.collect_binds(out));
         Ok(())
     }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.distinct.is_safe_to_cache_prepared() &&
+            self.select.is_safe_to_cache_prepared() &&
+            self.from.from_clause().is_safe_to_cache_prepared() &&
+            self.where_clause.as_ref().map(|w| w.is_safe_to_cache_prepared())
```
samphippen: this isn't single level of abstraction, but it looks like it could be with a few changes? Is it worth it?

samphippen:
Your tests here seem legit, but you've got like a billion fiddly little boolean cases, are there any important ones that are missing?

diesel/src/query_builder/delete_statement.rs
sgrif: The lack of a where clause is considered safe to prepare.

samphippen: :+1:

sgrif:
I don't think so. The only important cases are the ones that return the literal `false`. The tests there were more to drive implementation than enforce anything specific. Perhaps I need to change the implementation to make this more clear, but all the "fiddly" cases are basically saying "I'm safe to cache if all of my components are safe to cache"

diesel/src/query_builder/select_statement/boxed.rs
sgrif: Can you elaborate? To make this code uniform we'd need to implement `QueryFragment` for `Option<T>` which I'm not sure is something we want to do for this one case.

sgrif: Ah I thought we were in a different module. Yeah, `where_clause` is absolutely at a different level of abstraction in this case. It's treated specially because it's the only case where the equivalent query builder method is append and not replace (e.g. `filter(foo).filter(bar)` is the same as `filter(foo.and(bar))`, while `order(foo).order(bar)` is equivalent to `order(bar)`).

The fact that we're boxed makes things a little tricky. Actually making it uniform would require doing something like `Box<QueryFragment<DB> + WhereAnd<Box<QueryFragment<DB> + WhereAnd<Box<QueryFragment<DB> ...`, and infinitely recursive types like that are illegal. I'm not super happy with it, but don't have a great answer. It's a problem with this module as a whole though, not just this specific method.

sgrif: I elaborated a bit on the reasoning in 5fe68e03

PR #280, opened 2016-04-18
Implement prepared statement caching for PG

sgrif:
The overall structure is somewhat similar to that for SQLite. I have a
few issues with the implementation, but all of them point to some larger
issues with the pg::connection module and I feel that they're outside of
the scope of this PR. In particular, I don't like how much
`RawConnection` needed to leak for this to be factored well. That struct
lives at a really weird layer of abstraction, to the point that I'm
unsure it should exist at all. I need to give the pg module an update,
as it's full of warts from before `Connection` was generic. That will be
done as a separate PR.

I think there's a common ground between how we implemented this for
SQLite and how we implemented this for PG that I want to eventually
refactor both towards. This may even result in sufficiently generic code
that we can make it part of the `Connection` trait to make this easier
for third party adapters to hook into.

Once again, this results in a healthy performance boost across the
board.

```
name                                                      old ns/iter  new ns/iter    diff ns/iter   diff %
bench_selecting_0_rows_with_medium_complex_query          223,674      119,463            -104,211  -46.59%
bench_selecting_0_rows_with_trivial_query                 77,281       51,273              -26,008  -33.65%
bench_selecting_10k_rows_with_medium_complex_query        57,555,386   55,233,745       -2,321,641   -4.03%
bench_selecting_10k_rows_with_medium_complex_query_boxed  58,937,340   54,434,717       -4,502,623   -7.64%
bench_selecting_10k_rows_with_trivial_query               4,739,991    4,559,375          -180,616   -3.81%
bench_selecting_10k_rows_with_trivial_query_boxed         4,610,084    4,466,172          -143,912   -3.12%
```

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -3,7 +3,13 @@ extern crate libc;
 mod cursor;
 pub mod raw;
 mod row;
-mod result;
+#[doc(hidden)]
+pub mod result;
+mod stmt;
+
+use std::cell::Cell;
+use std::ffi::{CString, CStr};
+use std::rc::Rc;
 
 use connection::{SimpleConnection, Connection};
 use pg::{Pg, PgQueryBuilder};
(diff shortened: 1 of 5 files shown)


sgrif:
Review? @mfpiccolo @mcasper @samphippen

mcasper:
Yay caching!

I also agree that those refactors belong in a separate PR, so looks good to me

mfpiccolo:
👍 Like it!

PR #283, opened 2016-04-19
Make our query builder faster than a SQL string

sgrif:
We've added a new trait, `QueryId`, which must be implemented for all
AST nodes, and optionally returns a type which conforms to
`std::any::Any`. Nodes which have their SQL change without the type
changing (such as boxed nodes) will return `None` here. If we're able to
uniquely identify the query on the type alone, we use that for the cache
key rather than the SQL string. This allows us to skip the query
construction entirely. Additionally, a type identifier can be used for a
faster hash lookup than a string could, allowing us to actually beat the
string literal case.

My intent is that all types which implement `QueryFragment` also
implement `QueryId`. However, we can't make it a supertrait, as that
would require specifying the `QueryId` associated type when creating a
boxed trait object, which defeats the purpose entirely. Ironcially, we
actually implement `QueryId` for the trait object directly, so in theory
we should be able to separate them. I don't know if there's a way to
actually do this, though.

There's a handful of types which implement the "noop" form of `TypeId`
that could potentially be safe to implement for real. The main one that
comes to mind is `Aliased`. While the SQL from its implementation of
`QueryFragment` implies that the SQL can change without the type, I
think the type encapsulates enough information. It's a niche feature
though so I erred on the side of caution.

Here are the benchmark results:

```
name                                                                            old ns/iter  new ns/iter    diff ns/iter   diff %
pg::connection::tests::benchmarks::prepared_statement_lookup_query_builder      1,048        191                    -857  -81.77%
pg::connection::tests::benchmarks::prepared_statement_lookup_raw_sql            319          318                      -1   -0.31%
sqlite::connection::tests::benchmarks::prepared_statement_lookup_query_builder  743          123                    -620  -83.45%
sqlite::connection::tests::benchmarks::prepared_statement_lookup_raw_sql        183          180                      -3   -1.64%
```

These benchmarks measured creating/fetch the prepared statement, without
executing the query (calling `prepare_query` on both backends). I opted
to benchmark this as the inherent noise of a database round trip on PG
made it difficult to demonstrate the difference. The benchmarks are not
included in this PR as the code is messy and these aren't benchmarks
that I care to keep in the long term. For those interested in the seeing the
code or running it for themselves, you can apply this patch:
https://gist.github.com/sgrif/24ed31f66246d71a2f7356a806ebc57a

We could potentially gain a little bit more performance here. A type
identifier is a u64 internally, so can be used with a perfect hashing
function. This would change the lookup to be constant time. I believe
this could save us another 100 ns or so, bringing the query builder
close to 0.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -3,6 +3,12 @@
 //! found in the README.
 #![deny(warnings)]
 #![cfg_attr(feature = "unstable", feature(specialization))]
+
+#[macro_use]
+mod macros;
+#[macro_use]
+pub mod query_builder;
+
 pub mod backend;
 pub mod connection;
 #[macro_use]
@@ -17,11 +23,7 @@ pub mod pg;
 #[cfg(feature = "sqlite")]
 pub mod sqlite;
 
-#[macro_use]
-mod macros;
-
 pub mod migrations;
-pub mod query_builder;
 mod query_dsl;
 pub mod query_source;
 pub mod result;
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -1,7 +1,7 @@
 extern crate libc;
 
 use backend::Backend;
-use query_builder::{AsQuery, QueryFragment};
+use query_builder::{AsQuery, QueryFragment, QueryId};
 use query_source::Queryable;
 use result::*;
 use types::HasSqlType;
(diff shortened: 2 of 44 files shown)


sgrif:
Review? @diesel-rs/core

killercup:
Before going all the way to prefect hashing, I think you could quite easily change the default hashing function (SipHash IIRC) to one better suited to hashing small things like FNV for another improvement of a few ns.

sgrif:
Yeah, definitely. Since we still have the string keys in some cases though, it'd probably be best to split out two different caches, and perform the lookup based on which variant we have. Since that's a bit of a larger change, I wanted to keep it as a separate PR (if we do it at all -- I'm not sure the benefits are worth it)

diesel/src/lib.rs
```diff
@@ -3,6 +3,12 @@
 //! found in the README.
 #![deny(warnings)]
 #![cfg_attr(feature = "unstable", feature(specialization))]
+
+#[macro_use]
+mod macros;
```
mfpiccolo: This got me earlier.  Took me a min to figure out why debug_sql! was not defined inside connection.  👍

mfpiccolo:
LGTM.  I even skimmed the twitch.  Intense!   💯

PR #288, opened 2016-04-20
Change `VarChar` to be an alias to `Text`

### LINKED ISSUE #287: Should `types::Text` be an alias for `types::VarChar`?
Right now these are different types, with separate bindings for all related things. As things stand right now, functions can only take arguments of a single type. For example, you can pass an expression of type `VarChar` to `lower`, but not one of type `Text`. We could certainly implement it in such a way that we use a trait to mark valid argument types, but that would end up with ambiguity when you do `lower("STRING LITERAL")` since that could be an expression of either type, and I'm not sure that type inference will always be able to save us.

All backends seem to treat these as pretty much identical. The only case that I can see where they differ is in things like limits, which we cannot enforce on the Diesel side<sup>1</sup>. Are there backends which treat them differently, either for valid arguments to functions/operators or for serialization/deserialization? If not, I think we should just make these aliases for one another.

<sup>1</sup>Ok, we technically could enforce this, and it would be more correct. But this is the same as how technically we shouldn't provide `impl ToSql<VarChar, Pg> for String` because PG doesn't allow nul bytes so we should only allow `CString` technically. While we do open up one case for runtime errors, the useability trade off is more than worth it for me.

sgrif:
This removes the `VarChar` type entirely, unifying it with text.
This simplifies the interaction with things like SQL functions, and
removes gotchas like `lower` not working with text columns. SQLite will
be unaffected by this change. PG is mostly unaffected. However, it
doesn't treat `varchar[]` arrays as equivalent to `text[]`. INSERT and
UPDATE will continue to work, but other operators/functions will error
out. Since they have identical representations in the database, I'm fine
with introducing the weird ideosynchracy of "use `text[]` instead of
`varchar[]`, as we remove some much larger warts.

There is some more in depth research/reasoning at #287.

One other visible effect here is that `VarChar` can no longer be
referenced in expression positions, as it is not a struct. This can be
seen in the change made to `diesel_tests::schema_dsl`. Since the structs
have no size and no method, there's never any reason to use them in
expression positions, and I've simply replaced that field entirely.

Close #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * `infer_schema!` on SQLite now accepts a larger range of type names
 
+* `types::VarChar` is now an alias for `types::Text`. Most code should be
+  unaffected by this. PG array columns are treated slightly differently,
+  however. If you are using `varchar[]`, you should switch to `text[]` instead.
+
 ## [0.6.1] 2016-04-14
 
 ### Added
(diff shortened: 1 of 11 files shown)


sgrif:
Review? @diesel-rs/core

samphippen:
LGTM.

mfpiccolo:
👍

PR #293, opened 2016-04-21
Add the ability to load and group a child association for many parents

sgrif:
This is somewhat analogous to the `includes` method from Active Record.
However, it does not allow arbitrarily deep nesting. That means you can
load as many child associations for the parent as you want, but not
grandchildren or great grandchildren (you can certainly use this
infrastructure to do that, we just don't provide helpers for it out of
the box)

There's still several enhancements I'd like to make. For instance,
having to write `#[derive(Identifiable)]` feels funky, as you'd never
use that on its own. We could maybe do it on `#[derive(Queryable)]` if
the model has a field with the same name as the primary key, but that
feels really gross. Probably the best place for us to do it
automatically would be in `#[has_many]`, but that can be a separate PR.

This also does not generate the optimal query for PG, as it creates an
`IN` statement instead of `= ANY`. We eventually want to change the node
returned by `eq_any` to do the right thing for PG, but it needs
specialization to be done cleanly. Perhaps for 0.7 we can do that
optimization if `feature = "unstable"`.

This does not handle cases where the foreign key is nullable. We skip
implementing `BelongsTo` in that case, meaning the associations can be
loaded but not grouped. I have some ideas on how we can handle this in
the future, but that will be a separate PR.

I'd like to find a way to eliminate the need to specify the struct type
for the association twice. This could either be by moving
`BelongingToDsl` over to the table, or by somehow carrying the type that
the method was called on.

Finally, I think we can add an additional layer on top of this to handle
the common case of "load and group the associations without additional
filtering" into something like `users.load_associated::<Post>(&conn)`

An example of what this might look like in the wild:
```rust
// Get the first 10 users, all the posts they've published, and all the comments they've written
fn load_stuff(conn: &Connection) -> QueryResult<Vec<(User, Vec<Post>, Vec<Comment>)>> {
    let users = users::table.limit(10).load::<User>(conn)?;
    let posts = Post::belonging_to(&users).load::<Post>(conn)?.grouped_by(&users);
    let comments = Comment::belonging_to(&users).load::<Comment>(conn)?.grouped_by(&users);
    Ok(users.into_iter().zip3(posts, comments).collect())
}

### DIFF
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -1,0 +1,32 @@
+use std::hash::Hash;
+
+pub trait Identifiable {
+    type Id: Hash + Eq + Copy;
+
+    fn id(&self) -> Self::Id;
+}
+
+pub trait BelongsTo<Parent: Identifiable> {
+    fn foreign_key(&self) -> Parent::Id;
+}
+
+pub trait GroupedBy<Parent>: IntoIterator + Sized {
+    fn grouped_by(self, parents: &[Parent]) -> Vec<Vec<Self::Item>>;
+}
+
+impl<Parent, Child> GroupedBy<Parent> for Vec<Child> where
+    Child: BelongsTo<Parent>,
+    Parent: Identifiable,
+{
+    fn grouped_by(self, parents: &[Parent]) -> Vec<Vec<Child>> {
+        use std::collections::HashMap;
+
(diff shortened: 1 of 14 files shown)


sgrif:
Review? @diesel-rs/core 

Docs will be a separate PR, as we need to document the existing behavior as well.

Issue #196, opened 2016-02-06
Add a story for upsert


sgrif:
This is currently a gap in our API. Both SQLite and PG support it, though I'm not sure we can do a backend agnostic abstraction. I'm fine with providing two backend specific abstractions for this case.

sgrif:
Bumping to the 0.7 milestone. SQLite has extremely simple semantics, and we should at least be able to implement it for at least that backend. PG might get bumped to 0.8.

I don't think we can abstract over this in a database agnostic way. Instead we will provide backend specific APIs. SQLite will probably be `insert(&new_record).into(table).on_conflict(replace)`

sgrif:
SQLite provides 5 options for alternate behaviors. `REPLACE`, `ROLLBACK`, `ABORT`, `FAIL`, and `IGNORE`. We won't provide support for `INSERT OR ROLLBACK`, as our handling of transactions is based on lexical scope and we can't properly handle this case.

Eijebong:
Any ETA for postgres support ? :)

sgrif:
Is this a blocker for your application? I can definitely get it done for 0.8

Eijebong:
Yes, it's a blocker.

sgrif:
Tentative release date for 0.8 is the 28th (if enough little things like this get completed I might release sooner though)

Eijebong:
Cool, thank you !

drusellers:
:+1:

norcalli:
This doesn't seem to have made it into the 0.8 release for Postgres, is it possible to prioritize a backend specific method for the next release? If I have the time, I may be able to send a PR (over the next few weeks)

sgrif:
It will be in the next release. 0.8 was featureless as we needed to get macros 1.1 support out.

sgrif:
I'm going to release 0.9 today. Unfortunately, I wasn't able to complete this in time despite my best efforts. PG's upsert is an extremely complex feature, and it's just going to take some more time for me to implement proper support with an API that is up to our standards. I can't push back the 0.9 release any longer, as I am going on vacation on Tuesday and I want to release with enough time to fix any issues that may occur. So I will move this to the 0.10 release, which is slated for the end of January.

norcalli:
I understand.

norcalli:
I didn't mean to submit that comment, dang mobile. I understand, as I've had some dealings with trying to implement it myself in Scala/Slick, and I appreciate waiting to implement it properly.

[...]

PR #297, opened 2016-04-23
Add support for `INSERT OR REPLACE` on SQLite

sgrif:
First piece of #196. SQLite has extremely simple semantics here. `INSERT
OR REPLACE` will literally take what the new row would have been had the
insert succeeded, and will replace the existing row with it. For this to
be most useful, we'll probably want to make it easier to use non-literal
expressions in insert at some point in the future. We'll need to see
what other usage looks like.

This also lays the groundwork for supporting the other modifiers on
SQLite. I had originally planned on doing an API that looks more like
what PG will look like. However, since the semantics are so different,
we'll probably want to keep the APIs from stepping on each others toes.

Another alternative I had leaned towards was `insert.or(replace)`.
However, as I started thinking about the compile fail tests that'd be
required there, and saw the implementation, I shied away from it. Since
there's a fixed set of 4 modifiers that we will support, I think one
function each is fine.


sgrif:
Review? @diesel-rs/core 

/cc @llogiq

llogiq:
@sgrif Thanks for the ping and the fast implementation. I'm very impressed with what I see so far. In fact I'm so impressed, I may blog about it.

sgrif:
@llogiq :heart: !!! Let me know if there's anything else you're missing.

mfpiccolo:
LGTM.

llogiq:
Cool! :+1:

PR #301, opened 2016-04-23
Change the format of `#[column_name]`

sgrif:
This is entirely to support the move from syntax extensions to normal
macros on stable. We cannot match the RHS of `#[column_name="name"]` and
use the result in an `ident` position. We can, however, handle
`#[column_name(name)]`.

I prefer the old form, but it would require changes to the rust compiler
to support without syntax extensions, which defeats the purpose of
moving to stable.


sgrif:
Review? @diesel-rs/core

mfpiccolo:
LGTM

Issue #99, opened 2016-01-15
Let's consider going back to macros for stable


sgrif:
Syntex has proven to be pretty painful to track for supporting our compiler plugins on Stable. I think it's probably worth keeping support for, but we've been having issues with it causing released versions to break after the fact. I'd like to keep it as an option, but remove it by default in favor of our old "put your struct definition in these macros" solution as the "will definitely work on this version of Rust until the end of time" option.

bjz:
The reliance on Syntex scares me, as somebody who wished to introduce Rust for one of our data analytics projects. I would have definitely been more confident if there were macro alternatives available, in lieu of the stabilization of syntax extensions :(

sgrif:
> The reliance on Syntex scares me

I wouldn't call it reliance. Every single thing that we use syntax extensions for is completely reasonable to write by hand (which I often do when I don't want to use syntex, or can't use codegen).

sgrif:
Putting this on the 0.7 milestone. I've been working on this for a few days now. Still need to audit a few more cases, but I'm fairly certain that every single one of our syntax extensions (with the exception of `infer_schema!` and `infer_table_from_schema!`) can be handled with a normal macro.

The overall plan is to add macros for each of our custom derive cases (all of our annotations are basically custom derive. The ones that aren't explicitly a `#[derive]` annotation are because they need arguments.) I'm thinking I'd like to structure the macros so that they are compatible with the [`custom-derive` crate](https://github.com/DanielKeep/rust-custom-derive), but I see no reason to enforce using that crate.

I think it makes sense to keep the syntax extension form, but simply have them call the macro from the main Diesel crate. Aside from the "this eventually becomes stable" argument, we can also provide *much* better error messages from a procedural macro. For example, if `Insertable` is derived on a tuple struct, all fields must be annotated with the name of the column. We simply can't communicate that effectively from a normal macro.

I plan on opening a PR moving `#[insertable_into]` over to a normal macro this weekend, and I think we should declare this as a blocker for 1.0.

sgrif:
These have been implemented wherever possible.

PR #303, opened 2016-04-24
Provide a syntax extension free alternative to `#[insertable_into]`

sgrif:
This provides a pure stable alternative to the `#[insertable_into]`
annotation. The intention is to change the annotation to call this
macro, rather than `impl Insertable` directly. However, there are some
unaddressed issues for that, and I will submit that as a separate PR to
attempt to keep the PR size reasonable.

The tests for this are a bit messy, as the actual test body doesn't
change much -- Just the struct definition. I moved the most common case
out into a macro, but I opted to just leave the duplication for the
remaining 4-5 cases that didn't fit, instead of trying to make it so dry
it chafes.

We will continue to support syntex as an option on stable, as we can
provide much better error messages from a procedural macro. I would like
to improve the error messages in some cases if possible though (in
particular, we want to handle the case where a unit struct is passed or
where a tuple struct has unannotated fields).

The structure of the macro is intended to be compatible with the
`custom_derive` crate. This is untested, but will be fully tested once
I've moved all our annotations to stable macros. The goal is for any
struct definition to be copy pasted into this macro, and the macro
parses the struct body to create the proper implementation. For
sufficiently large structs, we can hit the recursion limit, but there's
really no way around that. People will just need to bump the limit.

One case that this macro *doesn't* handle is when there are annotations
on struct fields other than `#[column_name]`. I had originally planned
to handle these, but I realized that the only recognized annotation that
could be there on stable is `#[cfg]`, and we are *not* handling cfg
attributes. We might handle that in the future, but it'd look *really*
ugly.

Related to #99

### DIFF
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -210,6 +210,7 @@ macro_rules! Insertable_column_expr {
 }
 
 #[cfg(test)]
+#[allow(missing_debug_implementations, missing_copy_implementations)]
 mod tests {
     use prelude::*;
 
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -1,0 +1,421 @@
+/// Implements the [`Insertable`][insertable] trait for a given struct. This
+/// macro should be called with the name of the table you wish to use the struct
+/// with, followed by the entire struct body.
+///
+/// [insertable]: prelude/trait.Insertable.html
+///
+/// # Example
+///
+/// ```no_run
+/// # #[macro_use] extern crate diesel;
+/// # table! { users { id -> Integer, name -> VarChar, hair_color -> Nullable<VarChar>, } }
+/// struct NewUser<'a> {
+///     name: &'a str,
+///     hair_color: &'a str,
+/// }
+///
+/// Insertable! {
+///     (users)
+///     struct NewUser<'a> {
+///         name: &'a str,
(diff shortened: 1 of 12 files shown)


sgrif:
Review? @diesel-rs/core 

/cc @aturon @nikomatsakis Seeing some cases of this sort of thing out in the wild is likely relevant to the lang team.

sgrif:
The build failures are because the code assumes https://github.com/rust-lang/rust/issues/31776. I need to address that, but the changes are minor and shouldn't need to affect review.

diesel/src/macros/insertable.rs
```diff
@@ -0,0 +1,422 @@
...
+        $(#[$ignore:meta])*
+        $(pub)* struct $($body:tt)*
+    ) => {
+        Insertable! {
+            ($table_name)
+            $($body)*
+        }
+    };
+
+    // Handle struct with lifetimes
```
mfpiccolo: One of the few places where I really appreciate in-code comments. 👍

sgrif: Yup. Would normally restructure this code to not need it, but that's basically impossible with this type of macro.

mfpiccolo:
Macros for days!   Nice work Sean.   Once travis is happy I am happy. 👍

tessgriffin:
Documentation seems very reasonable. Looks good to me!

PR #304, opened 2016-04-25
Fix Doc Comment Markup

killercup:


killercup:
cf. <https://www.reddit.com/r/rust/comments/4gbl2j/whats_everyone_working_on_this_week_172016/d2gaxp3?context=5>

killercup:
(Assuming the code still worked was a bit far fetched. I'll look into the failure tonight and will try to fix it)

sgrif:
Nice catch. The fix is pretty easy, just change `Serial` to `Integer`

killercup:
Should be good to go now.

sgrif:
Thanks!

PR #305, opened 2016-04-25
Add Clippy

killercup:
Add [clippy](https://github.com/Manishearth/rust-clippy) and fix all of its warnings. This is

- adding the latest clippy 0.0.103 as an optional dependency with the `lint` feature
- improving a whole bunch of stuff in the various crates to make clippy real happy
- green with `#[deny(warnings)]`!
- adding a new entry in Travis' build matrix (which compiles quite fast as it is using `no-trans`)

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,10 @@
+cyclomatic-complexity-threshold = 30
+doc-valid-idents = [
+  "MiB", "GiB", "TiB", "PiB", "EiB",
+  "DirectX", "OpenGL", "TrueType",
+  "GPLv2", "GPLv3",
+  "GitHub",
+  "IPv4", "IPv6",
+  "JavaScript", "NaN", "OAuth",
+  "SQLite", "PostgreSQL"
+]
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -21,6 +21,7 @@ quickcheck = { version = "0.3.1", optional = true }
 serde_json = { version = ">=0.8.0, <0.10.0", optional = true }
 uuid = { version = ">=0.2.0, <0.5.0", optional = true, features = ["use_std"] }
 url = { version = "1.4.0", optional = true }
+clippy = {optional = true, version = "0.0.114"}
 
 [dev-dependencies]
 cfg-if = "0.1.0"
@@ -33,6 +34,7 @@ tempdir = "^0.3.4"
 default = ["with-deprecated"]
 extras = ["chrono", "serde_json", "uuid"]
 unstable = []
+lint = ["clippy"]
 large-tables = []
 huge-tables = ["large-tables"]
 postgres = ["pq-sys"]
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -21,7 +21,7 @@ quickcheck = { version = "0.3.1", optional = true }
 serde_json = { version = ">=0.8.0, <0.10.0", optional = true }
 uuid = { version = ">=0.2.0, <0.5.0", optional = true, features = ["use_std"] }
 url = { version = "1.4.0", optional = true }
-clippy = {optional = true, version = "0.0.114"}
+clippy = { optional = true, version = "=0.0.114" }
 
 [dev-dependencies]
 cfg-if = "0.1.0"
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -31,18 +31,24 @@ pub trait TransactionManager<Conn: Connection> {
 
 use std::cell::Cell;
 
-/// An implementation of TransactionManager which can be used for backends
+/// An implementation of `TransactionManager` which can be used for backends
 /// which use ANSI standard syntax for savepoints such as SQLite and PostgreSQL.
 #[allow(missing_debug_implementations)]
 pub struct AnsiTransactionManager {
     transaction_depth: Cell<i32>,
 }
 
-impl AnsiTransactionManager {
-    pub fn new() -> Self {
+impl Default for AnsiTransactionManager {
+    fn default() -> Self {
         AnsiTransactionManager {
             transaction_depth: Cell::new(0),
         }
+    }
+}
+
+impl AnsiTransactionManager {
+    pub fn new() -> Self {
+        AnsiTransactionManager::default()
     }
 
     fn change_transaction_depth(&self, by: i32, query: QueryResult<()>) -> QueryResult<()> {
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -34,16 +34,9 @@ use std::cell::Cell;
 /// An implementation of `TransactionManager` which can be used for backends
 /// which use ANSI standard syntax for savepoints such as SQLite and PostgreSQL.
 #[allow(missing_debug_implementations)]
+#[derive(Default)]
 pub struct AnsiTransactionManager {
     transaction_depth: Cell<i32>,
-}
-
-impl Default for AnsiTransactionManager {
-    fn default() -> Self {
-        AnsiTransactionManager {
-            transaction_depth: Cell::new(0),
-        }
-    }
 }
 
 impl AnsiTransactionManager {
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -1,8 +1,25 @@
 //! Diesel is an ORM and query builder designed to reduce the boilerplate for database
 //! interactions. [A getting started guide](http://diesel.rs/guides/getting-started/) can be
 //! found on our website.
-#![deny(warnings, missing_debug_implementations, missing_copy_implementations)]
 #![cfg_attr(feature = "unstable", feature(specialization))]
+
+// Built-in Lints
+#![deny(warnings, missing_debug_implementations, missing_copy_implementations)]
+
+// Clippy lints
+#![cfg_attr(feature = "clippy", allow(unstable_features))]
+#![cfg_attr(feature = "clippy", feature(plugin))]
+#![cfg_attr(feature = "clippy", plugin(clippy(conf_file="../clippy.toml")))]
+#![cfg_attr(feature = "clippy", allow(
+    option_map_unwrap_or_else, option_map_unwrap_or,
+    match_same_arms, type_complexity,
+))]
+#![cfg_attr(feature = "clippy", warn(
+    option_unwrap_used, result_unwrap_used, print_stdout, wrong_pub_self_convention,
+    mut_mut, non_ascii_literal, similar_names, unicode_not_nfc,
+    enum_glob_use, if_not_else, items_after_statements, used_underscore_binding,
+))]
+#![cfg_attr(all(test, feature = "clippy"), allow(result_unwrap_used))]
 
 extern crate byteorder;
 
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -231,7 +231,6 @@ macro_rules! Insertable_column_expr {
 
 #[cfg(test)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
-#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
 mod tests {
     use prelude::*;
 
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -231,6 +231,7 @@ macro_rules! Insertable_column_expr {
 
 #[cfg(test)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
 mod tests {
     use prelude::*;
 
--- a/diesel/src/migrations/migration.rs
+++ b/diesel/src/migrations/migration.rs
@@ -53,19 +53,27 @@ fn valid_sql_migration_directory(path: &Path) -> bool {
 }
 
 fn file_names(path: &Path) -> Result<Vec<String>, MigrationError> {
-    try!(path.read_dir()).map(|e| {
-        Ok(try!(e).file_name().into_string().unwrap())
-    }).filter(|name| match name {
-        &Ok(ref n) => !n.starts_with("."),
+    try!(path.read_dir()).map(|entry| {
+        let file_name = try!(entry).file_name();
+
+        // FIXME(killercup): Decide whether to add MigrationError variant for this
+        match file_name.into_string() {
+            Ok(utf8_file_name) => Ok(utf8_file_name),
+            Err(original_os_string) =>
+                panic!("Can't convert file name `{:?}` into UTF8 string", original_os_string)
+        }
+    }).filter(|file_name| match *file_name {
+        Ok(ref name) => !name.starts_with('.'),
         _ => true
     }).collect()
 }
 
 #[doc(hidden)]
 pub fn version_from_path(path: &Path) -> Result<String, MigrationError> {
-    path.file_name().unwrap()
+    path.file_name()
+        .expect(&format!("Can't get file name from path `{:?}`", path))
         .to_string_lossy()
-        .split("_")
+        .split('_')
         .nth(0)
         .map(|s| Ok(s.into()))
         .unwrap_or_else(|| {
@@ -101,6 +109,7 @@ fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigr
 }
 
 #[cfg(test)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
 mod tests {
     extern crate tempdir;
 
--- a/diesel/src/migrations/migration.rs
+++ b/diesel/src/migrations/migration.rs
@@ -109,7 +109,6 @@ fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigr
 }
 
 #[cfg(test)]
-#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
 mod tests {
     extern crate tempdir;
 
--- a/diesel/src/migrations/migration_error.rs
+++ b/diesel/src/migrations/migration_error.rs
@@ -5,8 +5,6 @@ use std::{fmt, io};
 use std::path::PathBuf;
 use std::error::Error;
 
-use self::MigrationError::*;
-
 #[derive(Debug)]
 pub enum MigrationError {
     MigrationDirectoryNotFound,
@@ -18,10 +16,15 @@ pub enum MigrationError {
 impl Error for MigrationError {
     fn description(&self) -> &str {
         match *self {
-            MigrationDirectoryNotFound => "Unable to find migrations directory in this directory or any parent directories.",
-            UnknownMigrationFormat(_) => "Invalid migration directory, the directory's name should be <timestamp>_<name_of_migration>, and it should only contain up.sql and down.sql.",
-            IoError(ref error) => error.description(),
-            UnknownMigrationVersion(_) => "Unable to find migration version to revert in the migrations directory.",
+            MigrationError::MigrationDirectoryNotFound =>
+                "Unable to find migrations directory in this directory or any parent directories.",
+            MigrationError::UnknownMigrationFormat(_) =>
+                "Invalid migration directory, the directory's name should be \
+                <timestamp>_<name_of_migration>, and it should only contain up.sql and down.sql.",
+            MigrationError::IoError(ref error) =>
+                error.description(),
+            MigrationError::UnknownMigrationVersion(_) =>
+                "Unable to find migration version to revert in the migrations directory.",
         }
     }
 }
@@ -55,6 +58,7 @@ impl From<io::Error> for MigrationError {
 }
 
 #[derive(Debug, PartialEq)]
+#[cfg_attr(feature = "clippy", allow(enum_variant_names))]
 pub enum RunMigrationsError {
     MigrationError(MigrationError),
     QueryError(result::Error),
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -263,7 +263,6 @@ pub fn search_for_migrations_directory(path: &Path) -> Result<PathBuf, Migration
 }
 
 #[cfg(test)]
-#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
 mod tests {
     extern crate tempdir;
 
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -73,7 +73,6 @@ use std::io::{stdout, Write};
 
 use expression::expression_methods::*;
 use query_dsl::*;
-use self::migration_error::MigrationError::*;
 use self::schema::__diesel_schema_migrations::dsl::*;
 use {Connection, QueryResult};
 
@@ -150,7 +149,7 @@ fn migration_with_version(ver: &str) -> Result<Box<Migration>, MigrationError> {
     });
     match migration {
         Some(m) => Ok(m),
-        None => Err(UnknownMigrationVersion(ver.into())),
+        None => Err(MigrationError::UnknownMigrationVersion(ver.into())),
     }
 }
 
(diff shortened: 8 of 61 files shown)


killercup:
Currently, [these](https://gist.github.com/killercup/2f921016c158a051b01ebd3cd5bf4e25) warnings remain.

diesel/src/lib.rs
```diff
@@ -1,9 +1,13 @@
 //! Diesel is an ORM and query builder designed to reduce the boilerplate for database
 //! interactions. [A getting started guide](http://diesel.rs/guides/getting-started/) can be
 //! found on our website.
-#![deny(warnings, missing_debug_implementations, missing_copy_implementations)]
+#![deny(missing_debug_implementations, missing_copy_implementations)]
```
killercup: This change should be reverted before merging (but it is really when checking for warnings)

clippy.toml
```diff
@@ -0,0 +1,2 @@
+cyclomatic-complexity-threshold = 30
+doc-valid-idents = ["MiB", "GiB", "TiB", "PiB", "EiB", "SQLite", "PostgreSQL"]
```
sgrif: What is `MiB` and friends here? Sizes? Where did they come from?

killercup: Those are clippy's default values, copied from the docs. There are actually defined [here](https://github.com/Manishearth/rust-clippy/blob/c66e90303fe7cd0df6f3191d157e65f4d58b5de5/src/utils/conf.rs#L148-L161) and I just noticed they also included 'GitHub'. I guess we could just leave them and add diesel's own CamelCase-names-that-are-not-code when we see an invalid warning?

sgrif: Seems good

diesel/src/lib.rs
```diff
@@ -1,9 +1,13 @@
...
 #![cfg_attr(feature = "unstable", feature(specialization))]
 
+#![cfg_attr(feature = "dev", allow(unstable_features))]
+#![cfg_attr(feature = "dev", feature(plugin))]
+#![cfg_attr(feature = "dev", plugin(clippy(conf_file="../clippy.toml")))]
```
sgrif: What do you think about adding clippy to `dev-dependencies` instead, and changing this to `cfg_attr(all(test, feature = "unstable")))`

diesel/src/query_dsl/boxed_dsl.rs
```diff
@@ -57,14 +57,14 @@ impl<'a, T, DB> InternalBoxedDsl<'a, DB> for T where
...
+/// Diesel queries also have a similar problem to [`Iterator`][iterator], where
 /// returning them from a function requires exposing the implementation of that
-/// function. The [helper_types][helper_types] module exists to help with this,
+/// function. The [`helper_types`][helpertypes] module exists to help with this,
 /// but you might want to hide the return type or have it conditionally change.
 /// Boxing can achieve both.
 ///
 /// [iterator]: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html
-/// [helper_types]: ../helper_types/index.html
+/// [helpertypes]: ../helper_types/index.html
```
sgrif: Lol what

killercup: Oh, I tried making sense of the doc_markdown lint which wanted me to put
helper_types in back ticks. I didn't mean to leave that in. (Is there a
clippy issue for that? I should probably create one otherwise)
Sean Griffin <notifications@github.com> schrieb am Mo., 25. Apr. 2016 um
23:24:

killercup:
(Updated to split into two commits and remove change for doc_markdown false positive.)

diesel/src/query_dsl/boxed_dsl.rs
killercup: cf. https://github.com/Manishearth/rust-clippy/issues/883

diesel/src/lib.rs
killercup: I created a new feature so we can control if/when CI includes clippy. Do you think we should just include it every time unstable is tested?

sgrif: I would imagine we'd want it whenever possible since it's a lint pass. I
don't feel terribly strongly about it though

killercup:
I just updated the PR to update clippy and deny warnings _unless_ clippy is used. I also added a clippy 'step' to travis. You can see some of the remaining warnings here: https://travis-ci.org/diesel-rs/diesel/jobs/126730544#L1039

diesel/src/migrations/migration_error.rs
```diff
@@ -55,6 +55,7 @@ impl From<io::Error> for MigrationError {
 }
 
 #[derive(Debug, PartialEq)]
+#[cfg_attr(feature = "dev", allow(enum_variant_names))]
```
sgrif: Do you hold the opinions that the current names are better than what clippy wants, or is this just to avoid the API change?

sgrif: By which I mean I don't feel strongly about this enum, and I'd be fine with changing the names for this one.

diesel/src/pg/types/floats/mod.rs
```diff
@@ -43,12 +43,12 @@ impl Error for InvalidNumericSign {
 impl FromSql<types::Numeric, Pg> for PgNumeric {
     fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
         let mut bytes = not_none!(bytes);
-        let ndigits = try!(bytes.read_u16::<BigEndian>());
-        let mut digits = Vec::with_capacity(ndigits as usize);
+        let num_digits = try!(bytes.read_u16::<BigEndian>());
```
sgrif: This is objectively a better name. It's worth noting that I named these local variables for parity with PG internally though, which is what this is based off of. https://github.com/postgres/postgres/blob/6928484bda454f9ab2456d385b2d317f18b6bf1a/src/backend/utils/adt/numeric.c#L822-L825

I don't feel strongly about whether that matters or what this should be named, but I just wanted to mention it. We should go with whichever you think is best.

diesel/src/query_builder/debug.rs
```diff
@@ -2,7 +2,7 @@ use backend::Debug;
 use super::{QueryBuilder, BuildQueryResult};
 
 #[doc(hidden)]
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
sgrif: I wish `Default` had `fn new() -> Self` as an alias for `Default`. Since we can derive this, and the derived fn will be identical to our explicit `new` function. Since this struct is purely internal, I wonder if it's worth just deleting our `new` function entirely and changing the call site to use `default`. It reads slightly worse, but if we're adding the derive we may as well have less code as a result.

diesel/Cargo.toml
```diff
@@ -17,6 +17,7 @@ byteorder = "0.3.*"
 quickcheck = { version = "0.2.25", optional = true }
 chrono = { version = "^0.2.17", optional = true }
 uuid = { version = "^0.2.0", optional = true, features = ["use_std"] }
+clippy = {optional = true, version = "0.0.64"}
```
sgrif: This version number makes me chuckle.

diesel/src/migrations/migration_error.rs
killercup: I wasn't sure if it was a breaking change, so I left it as is. I think "QueryError" reads nicely, but if it's only used as "RunMigrationsError::QueryError" I would remove the suffix.

diesel/src/pg/types/floats/mod.rs
killercup: Clippy found this because the names `ndigits` and `digits` were to similar, I was actually glad that it did. I had to think about it for a bit to understand what was meant here and I initially thought this was `n` represented as bytes (but there was no `n`). I think `num_*` is still a bit weird (but was just 3 chars to add), I would probably write `digit_count`.

diesel/src/query_builder/debug.rs
killercup: Interesting. I agree, I would remove that `fn new() -> Self`.

I rarely have `fn new() -> Self` (i.e. with no parameter) but often use `fn new(stuff) -> Self` as a constructor (which may internally do stuff like `StructName { field1: stuff, ... StructName::default() }`).

diesel/Cargo.toml
killercup: We should plan to from that directly to 0.0.128, obviously 😄

killercup:
I just rebased this on master and updated to the latest clippy version (maybe that needs a newer nightly -- Travis will tell us).

I've also added some _allow_s for lints which I don't think are useful here ([option_map_unwrap_or_else](https://github.com/Manishearth/rust-clippy/wiki#option_map_unwrap_or_else) and [option_map_unwrap_or](https://github.com/Manishearth/rust-clippy/wiki#option_map_unwrap_or) or for specific parts of the code I didn't want to change. `git grep -i 'feature = "dev", allow'` should show all instances.

killercup:
Travis confirms that this clippy version does not compile with the nightly from May 9. @sgrif, I think you mentioned that in the future you wanted to use the nightly from the date the latest stable was released, right? According to http://rusty-dash.com that'd currently be May 27 and the next one will be Jul 8.

I'm okay with downgrading clippy to get this merged, btw.

killercup:
I finally updated this PR to a newer clippy version and to fix all the lints (I also enabled a few optional ones). This passes locally with a nightly from 2016-09-29. Once we bump nightly (probably along with the macros 1.1 stuff), I'd like to try to land this.

After that, I'll add clippy to other crates.

sgrif:
Sounds good. After Macros 1.1 I'm going to finish composite primary keys and roll a release, we can merge after that

sgrif:
Want to rebase so we can merge?

killercup:
Rebased, bumped Clippy to version 0.0.103, and allowed/fixed some new lints. Since Clippy fails on current nightly, I made Travis stick to the one from clippy's last release.

Let's see if they manage to release 0.0.104 soon and merge _after that_.

killercup:
Rebased, again :D

The most recent nightly is from the 19th (6 days ago), which is kinda unfortunate. Since this is open for 3/4 of a year now, I'll not squash this all right now but wait for one more nightly, and maybe for @sgrif to have quick look.

Ideally, I'd like to add another travis env with a fixed nightly that runs clippy (and is an allowed failure), until we can install clippy with rustup and have a guarantee that it compiles at all times.

killercup:
Finally! This is

- adding the latest clippy 0.0.103 as an optional dependency with the `lint` feature
- improving a whole bunch of stuff in the various crates to make clippy real happy
- green with `#[deny(warnings)]`!
- adding a new entry in Travis' build matrix (which compiles quite fast as it is using `no-trans`)

clippy.toml
```diff
@@ -0,0 +1,10 @@
+cyclomatic-complexity-threshold = 30
+doc-valid-idents = [
+  "MiB", "GiB", "TiB", "PiB", "EiB",
+  "DirectX", "OpenGL", "TrueType",
+  "GPLv2", "GPLv3",
+  "GitHub",
+  "IPv4", "IPv6",
+  "JavaScript", "NaN", "OAuth",
+  "SQLite", "PostgreSQL"
```
sgrif: What's this array actually doing? Are we sure we want to allow this? For example I named everything `Mysql` instead of `MySQL` to be ideomatic. I would assume we want the same to apply to other backends.

diesel/src/lib.rs
```diff
@@ -1,9 +1,26 @@
...
+
+// Clippy lints
+#![cfg_attr(feature = "clippy", allow(unstable_features))]
+#![cfg_attr(feature = "clippy", feature(plugin))]
+#![cfg_attr(feature = "clippy", plugin(clippy(conf_file="../clippy.toml")))]
+#![cfg_attr(feature = "clippy", allow(
+    option_map_unwrap_or_else, option_map_unwrap_or,
+    match_same_arms, type_complexity,
+))]
+#![cfg_attr(feature = "clippy", warn(
```
sgrif: This still errors since we have `deny(warnings)` right?

diesel/src/macros/insertable.rs
```diff
@@ -231,6 +231,7 @@ macro_rules! Insertable_column_expr {
 
 #[cfg(test)]
 #[allow(missing_debug_implementations, missing_copy_implementations)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
```
sgrif: Isn't this redundant with https://github.com/diesel-rs/diesel/pull/305/files#diff-7ab3c70a207718b2fce878a8f925a415R22?

diesel/src/migrations/migration.rs
```diff
@@ -101,6 +109,7 @@ fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigr
 }
 
 #[cfg(test)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
```
sgrif: Isn't this redundant with https://github.com/diesel-rs/diesel/pull/305/files#diff-7ab3c70a207718b2fce878a8f925a415R22?

diesel/src/migrations/mod.rs
```diff
@@ -264,6 +263,7 @@ pub fn search_for_migrations_directory(path: &Path) -> Result<PathBuf, Migration
 }
 
 #[cfg(test)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
```
sgrif: Isn't this redundant with https://github.com/diesel-rs/diesel/pull/305/files#diff-7ab3c70a207718b2fce878a8f925a415R22?

diesel/src/mysql/connection/raw.rs
```diff
@@ -46,10 +46,10 @@ impl RawConnection {
             // Make sure you don't use the fake one!
             ffi::mysql_real_connect(
                 self.0,
-                host.map(|x| x.as_ptr()).unwrap_or(ptr::null_mut()),
+                host.map(CStr::as_ptr).unwrap_or_else(|| ptr::null_mut()),
```
sgrif: What's the lint that led to the `unwrap_or_else` change? `ptr::null_mut` is a `const fn` so it probably shouldn't be complaining here.

diesel/src/pg/connection/mod.rs
```diff
@@ -107,11 +107,12 @@ impl Connection for PgConnection {
 }
 
 impl PgConnection {
+    #[cfg_attr(feature = "clippy", allow(type_complexity))]
```
sgrif: Heh. I need to refactor this.

diesel/src/pg/connection/mod.rs
```diff
@@ -142,7 +143,9 @@ extern "C" fn noop_notice_processor(_: *mut libc::c_void, _message: *const libc:
 extern "C" fn default_notice_processor(_: *mut libc::c_void, message: *const libc::c_char) {
     use std::io::Write;
     let c_str = unsafe { CStr::from_ptr(message) };
-    ::std::io::stderr().write(c_str.to_bytes()).unwrap();
+    ::std::io::stderr()
+        .write_all(c_str.to_bytes())
+        .expect("Error writing to `stderr`");
```
sgrif: The computer is on fire. :)

diesel/src/mysql/connection/raw.rs
sgrif: https://github.com/Manishearth/rust-clippy/issues/1527

diesel/src/pg/connection/raw.rs
```diff
@@ -45,6 +45,7 @@ impl RawConnection {
         RawResult::new(PQexec(self.internal_connection, query), self)
     }
 
+    #[cfg_attr(feature = "clippy", allow(too_many_arguments))]
```
sgrif: We should probably enable this for the whole file

diesel/src/pg/connection/result.rs
```diff
@@ -40,7 +40,8 @@ impl PgResult {
             let count_str = str::from_utf8_unchecked(count_bytes);
             match count_str {
                 "" => 0,
-                _ => count_str.parse().unwrap()
+                _ => count_str.parse()
+                    .expect("Error parsing `rows_affected` as integer value")
```
sgrif: Now that I'm more familiar with the wire protocol, I know why libpq is giving us this as a c string. And it's really dumb.

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -24,6 +24,7 @@ pub enum Query {
 }
 
 impl Query {
+    #[cfg_attr(feature = "clippy", allow(ptr_arg))]
```
sgrif: We should probably enable this for the whole module.

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -34,8 +35,10 @@ impl Query {
                  .unwrap_or(ptr::null()))
```
sgrif: Wait did clippy not break on this one?

diesel/src/pg/types/floats/mod.rs
```diff
@@ -41,12 +41,12 @@ impl Error for InvalidNumericSign {
 impl FromSql<types::Numeric, Pg> for PgNumeric {
     fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
         let mut bytes = not_none!(bytes);
-        let ndigits = try!(bytes.read_u16::<NetworkEndian>());
```
sgrif: I named this variable for parity with the pg code which is the "documentation" for this structure. I don't feel strongly on whether we need to keep it or not.

diesel/src/connection/transaction_manager.rs
```diff
@@ -31,19 +31,25 @@ pub trait TransactionManager<Conn: Connection> {
...
 /// which use ANSI standard syntax for savepoints such as SQLite and PostgreSQL.
 #[allow(missing_debug_implementations)]
 pub struct AnsiTransactionManager {
     transaction_depth: Cell<i32>,
 }
 
-impl AnsiTransactionManager {
-    pub fn new() -> Self {
+impl Default for AnsiTransactionManager {
+    fn default() -> Self {
```
sgrif: I think we can derive this.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -182,6 +182,7 @@ fn error_message(err_code: libc::c_int) -> &'static str {
 }
 
 #[cfg(test)]
+#[cfg_attr(feature = "clippy", allow(result_unwrap_used))]
```
sgrif: Isn't this redundant with the line in `lib.rs`?

diesel/src/sqlite/query_builder/mod.rs
```diff
@@ -10,14 +10,20 @@ pub struct SqliteQueryBuilder {
     pub sql: String,
 }
 
-impl SqliteQueryBuilder {
-    pub fn new() -> Self {
+impl Default for SqliteQueryBuilder {
+    fn default() -> Self {
```
sgrif: I think we can derive this

diesel/src/test_helpers.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg_attr(feature = "clippy", allow(result_unwrap_used))]
```
sgrif: Isn't this redundant with the line in `lib.rs`?

diesel/src/types/impls/option.rs
```diff
@@ -110,6 +110,7 @@ use pg::Pg;
 
 #[test]
 #[cfg(feature = "postgres")]
+#[cfg_attr(feature = "clippy", allow(string_lit_as_bytes))]
```
sgrif: Eh. We may as well just change that line to `b"string lit"`. Also there's a typo in `expected` on that line, too

diesel/src/types/impls/tuples.rs
```diff
@@ -19,8 +19,8 @@ macro_rules! tuple_impls {
             impl<$($T),+, DB> HasSqlType<($($T,)+)> for DB where
                 $(DB: HasSqlType<$T>),+,
                 DB: Backend,
-            {
```
sgrif: I don't think you meant to make this change.

diesel/src/types/impls/tuples.rs
```diff
@@ -71,6 +71,7 @@ macro_rules! tuple_impls {
                 type SqlType = ($(<$T as Expression>::SqlType,)+);
             }
 
+            #[cfg_attr(feature = "clippy", allow(eq_op))] // Clippy doesn't like the trivial case for 1-tuples
```
sgrif: Should we enable this for the whole file?

diesel_cli/Cargo.toml
```diff
@@ -18,13 +18,15 @@ clap = "2.20.3"
 diesel = { version = "0.10.0", default-features = false }
 dotenv = "0.8.0"
 diesel_infer_schema = { version = "0.1.0", default-features = false, path = "../diesel_infer_schema" }
+clippy = {optional = true, version = "0.0.113"}
```
sgrif: Spaces?

diesel_cli/src/main.rs
```diff
@@ -1,5 +1,20 @@
+// Built-in Lints
+#![deny(warnings, missing_copy_implementations)]
+
+// Clippy lints
+#![cfg_attr(feature = "clippy", allow(unstable_features))]
```
sgrif: Should we stick this in a file that we can `include!` in all the crates?

sgrif:
Can you add the `lint` feature to the integration test lines in `bin/test`, too? (Maybe hold off for a few on rebasing I'm about to push up something that changes that file for MySQL)

clippy.toml
killercup: This is only about doc comments, and which CamelCasel-looking things to allow to be used without putting them in back ticks.

diesel/src/lib.rs
killercup: Indeed. This is to enable additional lints which are allow by default.

diesel/src/pg/connection/mod.rs
killercup: Only the screen that shows errors! 🤣

diesel/src/pg/connection/stmt/mod.rs
killercup: Huh. Does it have a special case for null but not null_mut?

diesel/src/pg/types/floats/mod.rs
killercup: That's a good reason. I'll allow the lint here (and add a link to the pg code)

diesel_cli/Cargo.toml
killercup: [Sure](https://i.ytimg.com/vi/DHibe7XxZAQ/maxresdefault.jpg)

diesel_cli/src/main.rs
killercup: Good idea. Where should I put that?

clippy.toml
sgrif: Ah, I see. We should probably add `MySQL` then. Do we actually need the two lines above this one? `NaN` probably should be in backticks when it appears actually.

diesel_cli/src/main.rs
killercup: Actually, if we put that in the root directory, how can we make sure the `include!`d file gets, well, included in the `cargo package` artifact?

diesel/Cargo.toml
```diff
@@ -21,6 +21,7 @@ quickcheck = { version = "0.3.1", optional = true }
 serde_json = { version = ">=0.8.0, <0.10.0", optional = true }
 uuid = { version = ">=0.2.0, <0.5.0", optional = true, features = ["use_std"] }
 url = { version = "1.4.0", optional = true }
+clippy = {optional = true, version = "0.0.114"}
```
sgrif: Does this need to be locked down to `=0.0.114`?

diesel_cli/src/main.rs
sgrif: Not sure. Let's just try it and see what happens. I'd put it in the project root directory and call it `make_crate_pedantic.rs`

sgrif: Though CLI should probably continue to allow things in release builds so that random lints added in future versions of Rust don't cause the binary to stop compiling.

diesel/Cargo.toml
killercup: Hm, ^0.0.X should be equivalent to =0.0.X IIRC. But better safe than sorry! 😅

killercup:
🎉

PR #306, opened 2016-04-25
Add a `debug_assert!` for integer size

sgrif:
If a user ever does have to drop down to using the `sql` function
directly, it's easy to accidentally use the wrong type for an integer
field. A common mistake I see is trying to decode `COUNT(*)` into an
`i32` (The type of that expression should be `BigInt`, so it needs to be
decoded into an `i64`.) When this occurs, `0` will be read as the encoding is big endian.
This isn't a problem with the standard query builder, but for people trying to use
something like `group_by` they'll be writing the SQL manually.

Since this is a common mistake I've seen multiple times now, we should
explicitly check for it.

The assertion is checking `<=` instead of `==`, as if we get too few
bytes we'll fail to decode the field anyway, and I wanted the error
message to focus on the too many bytes case.


sgrif:
Review? @diesel-rs/core 

/cc @dikaiosune

mfpiccolo:
👍

PR #311, opened 2016-04-27
Give an error message when an invalid PG url is given to CLI

### LINKED ISSUE #310: Installing diesel_cli without e.g. feature `sqlite` can lead to cryptic errors
I have installed `diesel_cli` without SQLite support (as the SQLite libraries are not installed on my system):

```powershell
cargo install diesel_cli --no-default-features --features "postgres"
```

Invoking `diesel` with an invalid database url generates a cryptic error message:

```
> diesel database setup --database-url=https://localhost/test-db
thread '<main>' panicked at 'internal error: entered unreachable code: The backend function should ensure we never get here.', C:\Users\benjaminf\.cargo\registry\src\github.com-88ac128001ac3a9a\diesel_cli-0.6.1\src\database.rs:73
```

While the fix is simple (use the correct protocol), I did have to examine the source code to puzzle out the error message. Perhaps an improved error message could be implemented?

```rust
panic!(format!("The {0} backend is not supported. You must recompile with feature `{0}` enabled", other))
```

sgrif:
Right now we assume URLs that aren't for PG are for SQLite instead.
However, the CLI crate can be compiled without SQLite support. In this
case, we end up hitting the `unreachable` branch in our match statement.
I'd like to eventually change this code to use an enum, similar to
codegen. In the short term though, we just need to improve the error
message.

Fixes #310

### DIFF
--- a/diesel_cli/src/database.rs
+++ b/diesel_cli/src/database.rs
@@ -153,10 +153,13 @@ pub fn database_url(matches: &ArgMatches) -> String {
 /// Returns a &str representing the type of backend being used, determined
 /// by the format of the database url.
 pub fn backend(database_url: &String) -> &str {
-    if database_url.starts_with("postgres://") || database_url.starts_with("postgresql://") {
+    if database_url.starts_with("postgres://") || database_url.starts_with("postgresql://") && cfg!(feature = "postgres") {
         "postgres"
-    } else {
+    } else if cfg!(feature = "sqlite") {
         "sqlite"
+    } else {
+        panic!("{:?} is not a valid PostgreSQL URL. It should start with\
+                `postgres://` or `postgresql://`", database_url);
     }
 }
 


sgrif:
Review? @mcasper

deadalusai:
One note - it is also possible to compile `diesel_cli` without Postgres support. Is it reasonable to add an error message for that edge case as well?

sgrif:
If CLI is compiled without PG support, everything should be passed unconditionally to SQLite.

mcasper:
Thanks for taking care of the @sgrif, LGTM 👍

deadalusai:
The issue I see is that `fn backend` may return "postgres", and the code which switches on that value will again hit the `unreachable!` branch.

sgrif:
Ah you're right. We used to have an alternate version of that function, but it looks like it was lost. Fixed.

PR #312, opened 2016-04-28
Follow convention for new AST pointers

brookst:
I had to change these in order to use diesel_codegen on `rustc 1.10.0-nightly (bd938166d 2016-04-25)`.
The diesel_compile_tests all pass with this change but I'm not sure if there's anything else to check, I'm only just starting out with diesel_demo at the moment.

Old empty() method removed in
https://github.com/rust-lang/rust/commit/9108fb7bae11f18715d971eeae1e5ca84662e1ee

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ dist: trusty
 rust:
   - stable
   - beta
-  - nightly-2016-04-08
+  - nightly-2016-04-25
   - nightly
 addons:
   postgresql: '9.4'
--- a/diesel_codegen/README.md
+++ b/diesel_codegen/README.md
@@ -9,7 +9,7 @@ boilerplate needing to be written. It can be used through `rustc_plugin`, or
 Using on nightly
 ----------------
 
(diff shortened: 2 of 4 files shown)


sgrif:
Can you bump the date in .travis.yml and all the READMEs?

brookst:
Which READMEs need updating? diesel_codegen/README.md seems to be the only one that lists a nightly. Should that be the same as the one run in Travis? (It says 2016-03-11 while Travis is currently pinned to 2016-04-07)
The change was introduced for rust-1.7.0 so I guess it breaks for versions earlier than that. Is that worth mentioning?

sgrif:
Yes, the codegen readme is the one I'm referring to.

> The change was introduced for rust-1.7.0 so I guess it breaks for versions earlier than that. Is that worth mentioning?

This only affects nightly and we don't work on nightlies from over a month ago anyway.

brookst:
Ok, I've bumped Travis and the README. Tests reran fine.

PR #315, opened 2016-05-01
Add Bpchar type aliased to VarChar

derekdreery:
I ran into errors using `char` fields (these are padded fixed length char strings). They are rarely used but can be useful e.g. country codes (2-char). Externally they behave like varchars but internally they are stored differently.

Because they appear the same as varchar externally, I believe it's safe just to alias VarChar

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -16,5 +16,6 @@ pub mod sql_types {
     #[cfg(feature = "uuid")]
     #[derive(Debug, Clone, Copy, Default)] pub struct Uuid;
     pub type Bytea = ::types::Binary;
+    #[doc(hidden)]
     pub type Bpchar = ::types::VarChar;
 }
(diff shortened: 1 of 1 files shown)


diesel/src/pg/types/mod.rs
```diff
@@ -16,4 +16,5 @@ pub mod sql_types {
     #[cfg(feature = "uuid")]
     #[derive(Debug, Clone, Copy, Default)] pub struct Uuid;
     pub type Bytea = ::types::Binary;
+    pub type Bpchar = ::types::VarChar;
```
sgrif: Can you make this `#[doc(hidden)]`? In the long run what we need is a better story for using types that aren't provided by diesel from `infer_schema!` (both for apps, but also for third party crates adding support for things like full text search). However, I'm not fully sure how that's going to be done yet, and I acknowledge that there's no way to work around this in the short term. I'm fine with adding this temporarily, but want to be able to remove it once we have a better story for other types besides "use `table!` instead of `infer_schema!`

derekdreery:
Hi - I might be being silly, but isn't the whole mod doc_hidden already.

sgrif:
It gets pub rexported

Issue #33, opened 2015-11-29
Support union / intersect / except


Meyermagic:
http://www.postgresql.org/docs/current/static/queries-union.html

These are quite useful for set-like queries.

sgrif:
Definitely reasonable to support. Not going to assign it a milestone just yet, but if anyone wants to tackle it, feel free. Shouldn't be hard to implement, just need to make sure that `union`, `intersect`, and `except` can be called multiple times, but the other query DSL methods cannot be called after one of these methods.

Meyermagic:
I'm going to start working on this soon (those livestreams are good advertising). I'm not familiar with your codebase, though, and want to make sure I have the right idea of how this will work.

I haven't yet started coding, but I see two main issues right now:
1. Union/Intersect/Except only work with SELECT in at least PG 9.4, and not with UPDATE ... RETURNING even with a compatible type. So I can't just use AsQuery as is while relying on matching SqlType.
2. Union/Intersect/Except can be called on a simple select query or a compound one, but most DSL methods can only be called on simple queries.

My high-level plan is then:

Introduce a `CompoundSelectStatement` enum with Union, Intersect, and Except variants. This should satisfy AsQuery when T: AsQuery, U: AsQuery with compatible types, but with an additional bound to satisfy 1.

Introduce three new DSL methods UnionDsl, IntersectDsl, and ExceptDsl which also require the trait added to satisfy 1.

Add a new bound to the other DSL methods, something like NotCompoundQuery, to satisfy 2.

Is there a Contribution Guide or architecture document or similar? I feel like a higher-level perspective might be helpful.

Disclaimer: I haven't put down any code for this yet, only poking around on GitHub. This might be way off-base.

sgrif:
There's some random scattered thoughts with regards to contributing here. https://github.com/diesel-rs/diesel/issues/120

I'd avoid having the types returned by these methods implement any of the DSLs, really, and instead require a fully built query. We want to avoid things like enums if possible, as they carry runtime information that can't always be inlined, and can't be reasoned about at the type level.

Go ahead and rely on `AsQuery`, we can add marker traits if needed to filter out the problem cases.

Keep in mind that anything adding support for something like this would need to be thoroughly tested.

sgrif:
Gave this a little thought. It actually should be extremely straightforward, just needs a lot of tests.

```rust
trait UnionDsl<Other: AsQuery<SqlType=Self::SqlType>>: AsQuery {
[...]

PR #316, opened 2016-05-02
Implement `UNION` and `UNION ALL` combination queries

robertmaloney:
First part and foundation of #33.

`CombinableQuery` is the marker trait that allows unions of `SelectStatements` and `UnionStatements`. This trait can be used in the future PRs for `EXCEPT` and `INTERSECT`.

Per [this comment](https://github.com/diesel-rs/diesel/issues/33#issuecomment-215920493), I decided to not write compile-fail tests for other query DSL methods yet.

Still left to do:
- [ ] Support `ORDER BY`, `LIMIT`, and `OFFSET` DSLs (per [SQLite Spec](https://www.sqlite.org/lang_select.html) and [Pg Spec](http://www.postgresql.org/docs/current/static/queries-order.html))
- [ ] Docs (and tests)
- [ ] Split `UnionQuery` by `all` flag
- [ ] More compile-fails (`GROUP BY`, `WHERE`, etc.)

### DIFF
--- a/diesel_tests/tests/union.rs
+++ b/diesel_tests/tests/union.rs
@@ -1,0 +1,85 @@
+use super::schema::*;
+use diesel::*;
+use schema_dsl::*;
+
+#[test]
+fn union_on_selects() {
+    use schema::users::dsl::*;
+
+    let connection = connection();
+    connection.execute("INSERT INTO users (name) VALUES ('Sean'), ('Tess')")
+        .unwrap();
+
+    let expected_data =
+        if cfg!(feature = "postgres") {
+            vec![
+                "Tess".to_string(),
+                "Sean".to_string(),
+            ]
+        } else if cfg!(feature = "sqlite") {
+            vec![
+                "Sean".to_string(),
+                "Tess".to_string(),
+            ]
+        } else {
+            vec![]
+        };
+    let query = users.select(name);
+    let union = query.union(query);
+    let actual_data: Vec<String> = union
+        .load(&connection)
+        .unwrap();
+    assert_eq!(expected_data, actual_data);
+}
+
+#[test]
+fn union_all_on_selects() {
+    use schema::users::dsl::*;
+
+    let connection = connection();
+    connection.execute("INSERT INTO users (name) VALUES ('Sean'), ('Tess')")
+        .unwrap();
+
+    let expected_data = vec![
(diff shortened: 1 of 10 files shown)


Meyermagic:
Might be better to keep Union and UnionAll as separate structs to avoid the 1 byte of overhead and potentially allow us to track "distinctness" with the type system?

robertmaloney:
Because combination queries support `ORDER BY`, `LIMIT`, and `OFFSET`, it might be better to extract these fields away from `SelectStatement` into another struct, in order to better reflect the [SQLite diagram][0] (which I suspect is a similar syntax tree to Pg). However that might mean a preliminary PR. Thoughts @Meyermagic @sgrif?

[0]: https://www.sqlite.org/lang_select.html

diesel_tests/tests/union.rs
```diff
@@ -0,0 +1,85 @@
...
+                "Sean".to_string(),
+            ]
+        } else if cfg!(feature = "sqlite") {
+            vec![
+                "Sean".to_string(),
+                "Tess".to_string(),
+            ]
+        } else {
+            vec![]
+        };
```
sgrif: Maybe we can just add an order by clause instead? ;)

diesel_tests/tests/union.rs
```diff
@@ -0,0 +1,85 @@
...
```
sgrif: I'd like to see a more "real" test here. Maybe `users.select(name).union(posts.select(title)).union(comments.select(text))`?

diesel_tests/tests/union.rs
```diff
@@ -0,0 +1,85 @@
...
+    let query = users.select(name);
+    let union = query.union_all(query);
+    let actual_data: Vec<String> = union
+        .load(&connection)
+        .unwrap();
+    assert_eq!(expected_data, actual_data);
+}
+
+#[test]
+fn union_on_unions() {
```
sgrif: What about unions on unions on unions on unions? :trollface: #UnionsForDays

sgrif:
@robertmaloney Are you still interested in working on this?

robertmaloney:
@sgrif Yeah, I still want to finish this up, but I have to a lot more work for it, and I need to relearn the codebase 😄 . I'll close it for now and reopen when I get to a good point.

PR #319, opened 2016-05-06
Implement `RETURNING` clause for `DELETE` statements

robertmaloney:
In addition to inserts and updates, Pg has support for returning columns of deleted rows. The only difference between this and the others was that `SelectableExpression<FilteredQuerySource, ...>` had to be implemented for the column macro. This might also mean that `RETURNING` clauses on filtered updates were considered invalid up until now.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -45,6 +45,13 @@ macro_rules! column {
         {
         }
 
+        impl<ST, Source, Predicate> SelectableExpression<
+            $crate::query_source::filter::FilteredQuerySource<Source, Predicate>, ST>
+            for $column_name where
+                $column_name: SelectableExpression<Source, ST>
+        {
+        }
+
         impl $crate::expression::NonAggregate for $column_name {}
 
         impl $crate::query_source::Column for $column_name {
--- a/diesel/src/query_builder/delete_statement.rs
+++ b/diesel/src/query_builder/delete_statement.rs
@@ -1,4 +1,5 @@
-use backend::Backend;
+use backend::{Backend, SupportsReturningClause};
+use expression::{Expression, SelectableExpression, NonAggregate};
 use query_builder::*;
 use query_source::Table;
 use result::QueryResult;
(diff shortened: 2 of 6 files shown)


diesel/src/macros/mod.rs
```diff
@@ -45,6 +45,13 @@ macro_rules! column {
         {
         }
 
+        impl<ST, Source, Predicate> SelectableExpression<
+            $crate::query_source::filter::FilteredQuerySource<Source, Predicate>, ST>
+            for $column_name where
+            $column_name: SelectableExpression<Source, ST>
```
sgrif: Can you indent this once and add a trailing comma?

sgrif: Hm. This shouldn't be necessary actually, there should be a blanket `SelectableExpression` impl on `FilteredQuerySource`

sgrif:
Can you add a compile fail test to assert that `.returning` must take a selectable expression?

diesel/src/macros/mod.rs
robertmaloney: Couldn't find such an impl, so I tried making it myself:
```rust
impl<T, Source, Predicate> SelectableExpression<FilteredQuerySource<Source, Predicate>> for T where
    T: Expression + SelectableExpression<Source>
{
}
```

but it apparently conflicts with [this](https://github.com/robertmaloney/diesel/blob/c124253f50a94727bd64f5b8cc09c2f4e7171449/diesel/src/expression/mod.rs#L120-L124):
```rust
impl<'a, T: ?Sized, ST, QS> SelectableExpression<QS, ST> for &'a T where
    T: SelectableExpression<QS, ST>,
    &'a T: Expression,
{
}
```
Am I missing something?

diesel/src/query_builder/delete_statement.rs
```diff
@@ -44,3 +45,98 @@ impl<T, U, DB> QueryFragment<DB> for DeleteStatement<T, U> where
...
+        DeleteQuery {
+            returning: T::all_columns(),
+            statement: self,
+        }
+    }
+}
+
+impl<T, U> DeleteStatement<T, U> {
+    /// Specify what expression is returned after execution of the `delete`.
+    /// # Examples
```
killercup: Can you add an empty (doc) line before the examples headline?

killercup:
I think you can disregard the travis error on beta as this _beta_ is newer than the latest _nightly_ we support for codegen! (It's `rustc 1.12.0-beta.1 (822166b84 2016-08-16)`)

killercup:
@robertmaloney There is a weird error with compiling compile_test on Travis; but I just looked and after rebasing on master can't reproduce this locally. Can you maybe rebase this and see how it goes? Otherwise I can do a new PR to do that.

robertmaloney:
@killercup That worked, thanks!

diesel_compile_tests/tests/compile-fail/returning_clause_requires_selectable_expression.rs
```diff
@@ -0,0 +1,44 @@
...
+        noname -> VarChar,
+    }
+}
+
+fn main() {
+    let connection = PgConnection::establish("").unwrap();
+
+    delete(users::table.filter(users::columns::name.eq("Bill")))
+        .returning(non_users::columns::noname);
+    //~^ ERROR: SelectableExpression
```
sgrif: What's with the `:`?

robertmaloney: Woops, that's a typo. Interesting that it still passed! I'll remove.

sgrif: I'm curious if that means that there happens to be a `: ` in the right spot or if the test isn't running. XD

robertmaloney: [It passed](https://travis-ci.org/diesel-rs/diesel/jobs/159457293#L1241)

Issue #34, opened 2015-11-30
Remove all uses of `unwrap` from our docs


sgrif:
It's a terrible habit, and I agree that we shouldn't have it in our docs. I've gotten rid of a decent number recently, but I'd like to eliminate the rest in favor of `try!` (this will take some annoying boilerplate in our doctests, but it's probably worth it).

masklinn:
How do you figure that'd work for examples using `main` e.g. second example of the readme?

defyrlt:
I think that `.unwrap()` is a great thing for docs/examples. You don't have to teach your users how to handle error cases - there's a chapter in the Book for it.

sgrif:
The example using `main` is probably fine. I think many of our examples could change to use `try!`, or change their assertion to compare against `Ok` though

sgrif:
I don't think we need to teach users how to handle error cases, but if we can avoid `unwrap` without actually making the example less clear, I'd like to.

mfpiccolo:
If this repo is going to be a popular place for people who newer to Rust (which I think it will be) it is important to show how to do it the right way.  Or don't show it at all.  

Perhaps saving the Result as a variable and then a comment underneath about dealing with the result?

hiimtaylorjones:
Shouldn't this be closed because of the above PR? If not, I'm more than willing to finish what work wasn't covered by #45!

sgrif:
I think that there are still a few places left but I can't confirm at the moment. Feel free to let me know if that's the case

hiimtaylorjones:
Ah, I see what you're talking about.

Upon doing a global project search for `/// assert_eq!(`, I've found that cases that use `.unwrap()` in the assert are all using `Ok([statement])` now. Is this the desired beahvior? 

Anyways, I'm closing this PR and recommending that #34 be closed if everything is as it should be. Sorry for the misunderstanding!

sgrif:
Thanks!

PR #320, opened 2016-05-10
[Issue #34] - removing more instances of .unwrap() from documentation in favor of try!()

hiimtaylorjones:
As per #34, there were still a few instances of .unwrap() left in documentation. So, this (hopefully) covers some more ground towards eliminating these instances. 

Please note that `try!()` is a new Rust idiom to me personally, but I think I was able to get the hang of it pretty quickly. There's definitely a chance I messed something up though!

### DIFF
--- a/diesel/src/expression/expression_methods/escape_expression_methods.rs
+++ b/diesel/src/expression/expression_methods/escape_expression_methods.rs
@@ -23,7 +23,7 @@ use types::VarChar;
 /// #     use diesel::insert;
 /// #     let connection = establish_connection();
 /// #     try!(insert(&NewUser { name: "Ha%%0r".into() }).into(users)
-/// #         .execute(&connection));
+/// #         .execute(&connection)xt);
 /// let users_with_percent = users.select(name)
 ///     .filter(name.like("%😀%%").escape('😀'))
 ///     .load(&connection);
--- a/diesel/src/expression/expression_methods/escape_expression_methods.rs
+++ b/diesel/src/expression/expression_methods/escape_expression_methods.rs
@@ -22,8 +22,8 @@ use types::VarChar;
 /// #     use self::users::dsl::*;
 /// #     use diesel::insert;
 /// #     let connection = establish_connection();
-/// #     insert(&NewUser { name: "Ha%%0r".into() }).into(users)
-/// #         .try!(execute(&connection));
+/// #     try!(insert(&NewUser { name: "Ha%%0r".into() }).into(users)
+/// #         .execute(&connection));
 /// let users_with_percent = users.select(name)
 ///     .filter(name.like("%😀%%").escape('😀'))
 ///     .load(&connection);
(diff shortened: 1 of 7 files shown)


diesel/src/expression/expression_methods/escape_expression_methods.rs
```diff
@@ -23,7 +23,7 @@ use types::VarChar;
 /// #     use diesel::insert;
 /// #     let connection = establish_connection();
 /// #     insert(&NewUser { name: "Ha%%0r".into() }).into(users)
-/// #         .execute(&connection).unwrap();
+/// #         .try!(execute(&connection));
```
Ryman: Sadly, macros (which `try!` is) can't be called like methods, so the entire expression needs to be wrapped. You can try...
```rust
try!(insert(..).into(..).execute(..));
```

diesel/src/expression/expression_methods/global_expression_methods.rs
```diff
@@ -97,8 +97,8 @@ pub trait ExpressionMethods: Expression + Sized {
     /// # fn main() {
     /// #     use self::users::dsl::*;
     /// #     let connection = establish_connection();
-    /// #     connection.execute("INSERT INTO users (name) VALUES
-    /// #         ('Jim')").unwrap();
+    /// #     try!(connection.execute("INSERT INTO users (name) VALUES
```
Ryman: Slightly different problem with these, as `main` doesn't expect to return a `Result`, so the `try!` will fail. Not sure if there's any benefit to switching to `try!` here as this line is hidden from the output (It's prefixed with '#')

diesel/src/expression/expression_methods/escape_expression_methods.rs
hiimtaylorjones: 😢 Well that's a bummer. 

Thanks for the pointer! Pushing up the change in a few.

diesel/src/expression/expression_methods/global_expression_methods.rs
hiimtaylorjones: Hmmmmmm.

Maybe we should leave unwrap() in there then. Whatcha think @sgrif?

sgrif: We shouldn't remove it anywhere except documentation. This line isn't displayed

hiimtaylorjones: Oh! I see how the docs are being picked up and interpreted now. Will do.

sgrif:
I don't think any of these cases really need to be changed. We can't use `try!` for the most part in the main functions, what we're more looking for is anywhere that we're doing `assert_eq!(foo, bar.unwrap())` where we could instead make a change like `assert_eq!(Ok(foo), bar)`. I'm not sure if there are any of those cases left or not.

PR #324, opened 2016-05-11
Implement `Queryable` as a standard non-procedural macro

sgrif:
Queryable is fairly simple to derive compared ot some of the other
cases, so the macro definition is fairly uninteresting. I needed to make
some changes to the parse code to support this additional case, but I've
kept those as separate commits which have more context around each
change.

The most surprising piece here is that we were able to make this work
without requiring the `#[column_name(name)]` annotation. Due to the way
that we parse the struct, the `$field_kind` variable actually acts as a
fresh ident, due to macro hygiene. So even if the resulting code *looks*
like it's using the same name more than once, the parser treats them as
separate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   those wishing to avoid syntax extensions from `diesel_codegen`. See
   http://docs.diesel.rs/diesel/macro.Insertable!.html for details.
 
+* The `Queryable!` macro can now be used instead of `#[derive(Queryable)]` for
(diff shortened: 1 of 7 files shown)


sgrif:
Review? @diesel-rs/core @diesel-rs/contributors 

I recommend looking at each commit individually

killercup:
The first 3 commits LGTM, that last one I have to re-read in the morning 😄 

Do you want to maybe add some more documentation to the `Queryable` macro? The docs for [Insertable](https://github.com/diesel-rs/diesel/blob/e471837eb65e46cea5cb340dfcaa4ebd2f083854/diesel/src/macros/insertable.rs) are quite extensive and mention how the macro can be used with the `custom_derive` crate, for example.

sgrif:
Ah yes I need to add docs before this is merged

PR #327, opened 2016-05-13
Implement `#[derive(Identifiable)]` as a stable macro

sgrif:
There are several additional cases that I'd like to handle in the
future, but for this first pass it only handles cases that the
procedural macro handles. It does not handle generics, tuple structs, or
structs with a primary key other than ID.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   those wishing to avoid syntax extensions from `diesel_codegen`. See
   http://docs.diesel.rs/diesel/macro.Queryable!.html for details.
 
+* The `Identifiable!` macro can now be used instead of `#[derive(Identifiable)]` for
+  those wishing to avoid syntax extensions from `diesel_codegen`. See
+  http://docs.diesel.rs/diesel/macro.Identifiable!.html for details.
+
 ### Changed
 
 * `infer_schema!` on SQLite now accepts a larger range of type names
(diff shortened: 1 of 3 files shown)


sgrif:
Review? @diesel-rs/contributors @diesel-rs/core

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   those wishing to avoid syntax extensions from `diesel_codegen`. See
   http://docs.diesel.rs/diesel/macro.Queryable!.html for details.
 
+* The `Identifiable!` macro can now be used instead of `#[derive(Identifiable)]` for
+  those wishing to avoid syntax extensions from `diesel_codegen`. See
+  http://docs.diesel.rs/diesel/macro.Identifiable!.html for details.
```
killercup: I had to look up whether you could use an exclamation point as part of an URL. Turns out you can ([RFC 1738, page 3](http://www.ietf.org/rfc/rfc1738.txt)) and I don't need to file a bug with rustdoc :)

Oh, and I would probably turn this into the nicer

> See [the documentation](http://docs.diesel.rs/diesel/macro.Identifiable!.html) for details.

But that's totally not important.

sgrif: Good point. I'll probably give the whole changelog a pass before the next release.

killercup:
LGTM!

(After comparing the diff with that from #324 and looking at the code for 5min, I believe I can actually understand your macros now. Took me a while to parse the order in which the macro arms were called; GitHub's syntax highlighting of _only_ the `$` doesn't help either… 😄 )

samphippen:
👍

sgrif:
Yeah, it's rather annoying that the order has to be this way, but unfortunately it does.

PR #328, opened 2016-05-13
Implement `#[changeset_for]` as a standard non-procedural macro

sgrif:
This is a full implementation of the current behavior of the procedural macro, and also adds support for tuple structs, and lifetimes with names other than `'a` which we didn't have previously. There are no unit tests for it, as they would be extremely redundant with the integration test suite, and I'd prefer to just eventually add a third case in our test suite to test using `custom_derive!`.

The implementation gets a little funky. This is because macros cannot be used in type positions, but we need to parse each field individually to determine the type of `Changeset`. This will vary per field based on whether the field is an option type, and whether `treat_none_as_null` is true or not.

This code in theory would be ready to merge + docs, but the fact that this macro is so complicated makes me concerned that the feature itself is hard to reason about for users. I'm going to look into refactoring to make the underlying concepts a little simpler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   those wishing to avoid syntax extensions from `diesel_codegen`. See
   http://docs.diesel.rs/diesel/macro.Identifiable!.html for details.
 
+* The `AsChangeset!` macro can now be used instead of `#[changeset_for(table)]`
+  for those wishing to avoid syntax extensions from `diesel_codegen`. See
+  http://docs.diesel.rs/diesel/macro.AsChangeset!.html for details.
+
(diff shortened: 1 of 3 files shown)


sgrif:
I want to rethink the overall structure of the update module before merging this, but I figured you might have some feedback anyway @diesel-rs/contributors

sgrif:
Updated and ready for review @diesel-rs/contributors

killercup:
Sorry, just a drive-by comment: This is meant to be used with custom_derive, right? Can you add examples/docs? I tried following the flow of the macros, and all I can say is that they appear to make sense. 😄  (I don't know what edge cases to look for though)

sgrif:
Ah yes, I thought I had done that on the first pass. I will make it so.

sgrif:
Hm... Travis doesn't seem to want to run this branch anymore.

sgrif:
Tests passed locally. YOLO merging.

PR #344, opened 2016-05-26
Add sqlite support for chrono's Naive{DateTime,Date,Time} types

kardeiz:
Adds initial `chrono` support for sqlite. Uses the text representations of dates/times/datetimes as apparently preferred by sqlite (e.g., those created by `DATETIME('now')`).

### DIFF
--- a/diesel/src/sqlite/types/date_and_time.rs
+++ b/diesel/src/sqlite/types/date_and_time.rs
@@ -18,3 +18,79 @@ impl types::HasSqlType<types::Timestamp> for Sqlite {
         SqliteType::Text
     }
 }
+
+#[cfg(feature = "chrono")]
+mod chrono {
+    extern crate chrono;
+
+    use std::error::Error;
+    use std::io::Write;
+    use self::chrono::{NaiveDateTime, NaiveDate, NaiveTime};
+
+    use sqlite::Sqlite;
+    use sqlite::connection::SqliteValue;
+    use types::{self, Date, FromSql, IsNull, Time, Timestamp, ToSql, Text};
+
+    const SQLITE_DATETIME_FMT: &'static str = "%Y-%m-%d %H:%M:%S";
+    const SQLITE_DATE_FMT: &'static str = "%Y-%m-%d";
+    const SQLITE_TIME_FMT: &'static str = "%H:%M:%S";
+
+    expression_impls! {
+        Date -> NaiveDate,
+        Time -> NaiveTime,
+        Timestamp -> NaiveDateTime,
+    }
+
+    queryable_impls! {
+        Date -> NaiveDate,
+        Time -> NaiveTime,
+        Timestamp -> NaiveDateTime,
+    }
+
+    impl FromSql<Date, Sqlite> for NaiveDate {
+        fn from_sql(value: Option<&SqliteValue>) -> Result<Self, Box<Error+Send+Sync>> {
+            let text = not_none!(value).read_text();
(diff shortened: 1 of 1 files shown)


sgrif:
Does SQLite give something reasonable back if you read the field as something other than text? I would prefer to avoid date parsing if we can

diesel/src/sqlite/types/date_and_time.rs
```diff
@@ -18,3 +18,85 @@ impl types::HasSqlType<types::Timestamp> for Sqlite {
...
+    const SQLITE_TIME_FMT: &'static str = "%H:%M:%S";
+
+    #[cfg(not(feature = "postgres"))]
+    expression_impls! {
+        Date -> NaiveDate,
+        Time -> NaiveTime,
+        Timestamp -> NaiveDateTime,
+    }
+
+    #[cfg(not(feature = "postgres"))]
```
sgrif: Can we move these up somewhere outside of the adapter specific modules so there's one clear place it occurs regardless of the backend as long as `#[cfg(feature = "chrono")]` is true?

diesel/src/sqlite/types/date_and_time.rs
```diff
@@ -18,3 +18,85 @@ impl types::HasSqlType<types::Timestamp> for Sqlite {
         SqliteType::Text
     }
 }
+
+#[cfg(feature = "chrono")]
+mod chrono {
```
sgrif: I'd prefer to move this to a separate file

kardeiz:
@sgrif Right, I'm not sure the best way to handle this&mdash;I was mostly going by how `rusqlite` gets datetimes out of sqlite. I'm not a sqlite expert, but it appears sqlite can store [datetimes as epoch-second ints](http://stackoverflow.com/questions/17227110/what-is-the-best-way-to-save-datetime-type-in-sqlite).

The only issue with this is that in sqlite, `CURRENT_TIMESTAMP` always creates a **string** in the format above. You can work around this like `strftime('%s', CURRENT_TIMESTAMP)` to get seconds, but if you happen to insert the default `CURRENT_TIMESTAMP` and then try and extract it as a `numeric`, you will get just the year component (for example, `select cast(CURRENT_TIMESTAMP as numeric)` gives `2016`). I believe `expression::dsl::now` would have to be specialized for sqlite with a workaround in order to use integers.

diesel/src/sqlite/types/date_and_time.rs
kardeiz: Yeah, I realized how hacky this was after I wrote it. I didn't think about moving them out of the backend-specific modules, but I also don't know the appropriate place for them (`diesel/types/mod.rs` ?).

sgrif:
I don't mean that we should cast it in the SQL, I mean that the deseriailzation function we call can be anything, SQLite doesn't care. If you call `read_double`, it'll give back what it thinks that should be.

kardeiz:
I'm not sure if I understand completely, but when I do:

    let as_i32 = not_none!(value).read_long();
    println!("{:?}", &as_i32);

inside the `from_sql` function for `NaiveDateTime`, where the sqlite value is one inserted by `CURRENT_TIMESTAMP`, it prints `2016`. It seems to be doing the same thing as `select cast(CURRENT_TIMESTAMP as numeric)`, which makes it seem like sqlite is actually just storing the formatted datetime string.

Eijebong:
This has been finished in #887. Closing

PR #351, opened 2016-06-10
Feature/desc nulls last

kaj:
Add support for ordering with NULLS FIRST or NULLS LAST in postgres.

Methods nulls_first() and nulls_last() can be applied to the result of the asc() and desc() sorting order methods.

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -211,3 +211,95 @@ impl<T, ST> ArrayExpressionMethods<ST> for T where
     T: Expression<SqlType=Array<ST>>,
 {
 }
+
+use expression::predicates::{Asc, Desc};
+
+pub trait SortExpressionMethods : Sized {
+    /// Specify that nulls should come before other values in this ordering.
+    /// Normally, nulls come last when sorting in ascending order and first
+    /// when sorting in descending order.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("src/doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     users {
+    /// #         id -> Integer,
+    /// #         name -> VarChar,
+    /// #     }
+    /// # }
+    /// #
+    /// # table! {
+    /// #     foos {
+    /// #         id -> Integer,
+    /// #         foo -> Nullable<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     let connection = connection_no_data();
+    /// #     connection.execute("DROP TABLE IF EXISTS foos").unwrap();
+    /// connection.execute("CREATE TABLE foos (id SERIAL PRIMARY KEY, foo INTEGER)").unwrap();
+    /// connection.execute("INSERT INTO foos (foo) VALUES (NULL), (1), (2)").unwrap();
+    ///
+    /// #     use self::foos::dsl::*;
+    /// assert_eq!(Ok(vec![Some(1), Some(2), None]),
+    ///            foos.select(foo).order(foo.asc()).load(&connection));
+    /// assert_eq!(Ok(vec![None, Some(1), Some(2)]),
+    ///            foos.select(foo).order(foo.asc().nulls_first()).load(&connection));
+    /// #     connection.execute("DROP TABLE foos").unwrap();
+    /// # }
+    /// ```
+    fn nulls_first(self) -> NullsFirst<Self> {
+        NullsFirst::new(self)
+    }
+
+    /// Specify that nulls should come after other values in this ordering.
+    /// Normally, nulls come last when sorting in ascending order and first
+    /// when sorting in descending order.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("src/doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     users {
+    /// #         id -> Integer,
+    /// #         name -> VarChar,
+    /// #     }
+    /// # }
+    /// #
+    /// # table! {
+    /// #     foos {
+    /// #         id -> Integer,
+    /// #         foo -> Nullable<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     let connection = connection_no_data();
+    /// #     connection.execute("DROP TABLE IF EXISTS foos").unwrap();
+    /// connection.execute("CREATE TABLE foos (id SERIAL PRIMARY KEY, foo INTEGER)").unwrap();
+    /// connection.execute("INSERT INTO foos (foo) VALUES (NULL), (1), (2)").unwrap();
+    ///
+    /// #     use self::foos::dsl::*;
+    /// assert_eq!(Ok(vec![None, Some(2), Some(1)]),
+    ///            foos.select(foo).order(foo.desc()).load(&connection));
+    /// assert_eq!(Ok(vec![Some(2), Some(1), None]),
+    ///            foos.select(foo).order(foo.desc().nulls_last()).load(&connection));
+    /// #     connection.execute("DROP TABLE foos").unwrap();
+    /// # }
+    /// ```
+    fn nulls_last(self) -> NullsLast<Self> {
+        NullsLast::new(self)
+    }
+}
+
+impl<T> SortExpressionMethods for Asc<T> {}
+
+impl<T> SortExpressionMethods for Desc<T> {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## Unreleased
 
+* Added support for PostgreSQL `NULLS FIRST` and `NULLS LAST` when sorting.
+  See http://docs.diesel.rs/diesel/prelude/trait.SortExpressionMethods.html
+  for details.
+
 ## [0.7.0] - 2016-08-01
 
 ### Added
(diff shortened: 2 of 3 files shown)


diesel/src/expression/expression_methods/sort_expression_methods.rs
```diff
@@ -0,0 +1,44 @@
+use expression::predicates::*;
+
+#[cfg(feature = "postgres")]
+pub trait Ordering : Sized {
+    /// Specify that nulls should come before other values in this ordering.
+    /// Normally, nulls come last when sorting in ascending order and first
+    /// when sorting in descending order.
```
killercup: Could you add basically the same example as below here as well? (It's really nice to use doc tests for this, btw.)

diesel/src/expression/expression_methods/sort_expression_methods.rs
```diff
@@ -0,0 +1,44 @@
+use expression::predicates::*;
+
+#[cfg(feature = "postgres")]
+pub trait Ordering : Sized {
```
killercup: Is the name 'ordering' a good choice? It only set the ordering of nulls, right? Maybe call it `NullsOrdering`? (I don't have a preference.) Maybe also add a short doc comment.

diesel/src/expression/expression_methods/sort_expression_methods.rs
kaj: Sure, will do!

diesel/src/expression/expression_methods/sort_expression_methods.rs
kaj: I don't know.  I thought about it as a trait _on_ ordering, providing methods that work on orderings (it is implemented for Asc and Desc).  But maybe it should be thought of as a trait _for_ where to put the nulls, and in that case `NullsOrdering` (or something similar) is a better name.

killercup: FYI: Searching the docs for "trait:ord" shows that there is a `diesel::types::SqlOrd`.

kaj: Ah, so maybe I should put these methods in that trait instead?  Hm.  On second thought, I guess not, since that trait is implemented for column types, and not for `Asc` and `Desc`.  So if I put `nulls_last()` there it would be possible to `order(foo.nulls_last())` but not `order(foo.desc().nulls_last())`, and that would make no sense at all ...

kaj:
So, should I rename the trait, or is it fine as is?

Anything else? @killercup ? @sgrif ?

killercup:
I would at least add a doc comment for the trait :)

I don't have a preference for the name, though. That's up to @sgrif ;)

diesel/src/expression/expression_methods/mod.rs
```diff
@@ -7,11 +7,15 @@
...
 pub mod global_expression_methods;
+#[cfg(feature = "postgres")]
+pub mod sort_expression_methods;
 pub mod text_expression_methods;
 
 pub use self::bool_expression_methods::BoolExpressionMethods;
 pub use self::escape_expression_methods::EscapeExpressionMethods;
 pub use self::global_expression_methods::ExpressionMethods;
+#[cfg(feature = "postgres")]
+pub use self::sort_expression_methods::Ordering;
```
sgrif: Can we put this trait in the `pg` module? I think we can implement this for sqlite as well by generating `expr is not null, expr` for nulls first and `expr is null, expr` for nulls last.

diesel/src/expression/expression_methods/sort_expression_methods.rs
sgrif: `SqlOrd` is just a marker trait for SQL types which have a defined order, meaning that `max` and `min` should be called. It shouldn't affect this feature. A trait for this is fine, but I'd prefer `OrderingMethods` or something similar that matches the naming convention of everything else here. (I know it's redundant with the module name, but these traits are exported in prelude so I prefer to have them use non-generic names. It's the same reason that everything in `query_dsl` has `Dsl` in the name.)

diesel/src/expression/expression_methods/sort_expression_methods.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn nulls_last(self) -> NullsLast<Self> {
+        NullsLast::new(self)
+    }
+}
+
+#[cfg(feature = "postgres")]
+impl<T> Ordering for Asc<T> {}
+
+#[cfg(feature = "postgres")]
+impl<T> Ordering for Desc<T> {}
```
sgrif: Should this be implemented for any `T: Expression`? It's completely valid to specify this without a defined order.

diesel/src/expression/expression_methods/mod.rs
sgrif: The more I think about it, the more I think that implementing this for SQLite is a bad idea. It should be moved into the appropriate submodule in `pg` though.

diesel/src/expression/expression_methods/sort_expression_methods.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// #         id -> Integer,
+    /// #         name -> VarChar,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// let query = users.select(id).order(name.asc().nulls_first());
+    /// assert_eq!("SELECT `users`.`id` FROM `users` ORDER BY `users`.`name` ASC NULLS FIRST",
+    ///            debug_sql!(query));
```
sgrif: I'd prefer to avoid testing the generated SQL, and instead demonstrate the results in a data set.

kaj:
I tried moving the added code into to pg submodule, as @sgrif suggested 25 days ago.  It turns out a bit harder than I anticipated. As far as I understand, the problem is that the `postfix_expression!` macro don't allow me to specify a backend implementation, as the `infix_predicate!` macro does.

Any help fixing this would be much appreciated.

PS.  Sorry for being away for so long.

sgrif:
No worries. I will take a look at it as soon as I can. I am in Europe right
now for work so it will probably be mid next week.

kaj:
After merging master into the branch I read the compiler errors more carefully and managed to fix the problem (which actually was rather simple).  I also changed the name of the trait for consistency with similar traits and changed the doctests to test actual results rather than generated SQL.

I used `src/doctest_setup.rs` to get a connection, but a local table specific for the test.  I had to include a `table!` definition for `users` even though I don't use it.

diesel/src/pg/expression/sort_expression_methods.rs
```diff
@@ -0,0 +1,89 @@
...
+    ///            foos.select(foo).order(foo.asc()).load(&connection));
+    /// assert_eq!(Ok(vec!(None, Some(1), Some(2))),
+    ///            foos.select(foo).order(foo.asc().nulls_first()).load(&connection));
+    /// #     connection.execute("drop table foos").unwrap();
+    /// # }
+    /// ```
+    fn nulls_first(self) -> NullsFirst<Self> {
+        NullsFirst::new(self)
+    }
+    /// Specify that nulls should come after other values in this ordering.
```
sgrif: Can you add a blank line above this one?

sgrif:
LGTM. Can you add a changelog entry and squash? Thanks for working on this.

kaj:
Done.  The documentation link that most changelog entries have will be a 404, but I hope it will be the correct url if / when this PR is merged / released.

sgrif:
Hah I'm glad you mentioned something, as I noticed an issue with your structure. This should not get re-exported from `expression`, this should go in `diesel::expression::expression_methods`, which is already glob-reexported in `prelude` (sorry, I know this is nitpicky but we end up with a *lot* of these modules so I want it to be consistent)

kaj:
Ok. Are you talking about just the "use" structure, or moving the actual code into expression_methods.rs?

sgrif:
Moving the actual code into another module nested under
`expression_methods`, similarly to the other 5 or so modules in there.

kaj:
Ok, done.  Anything else, @sgrif ?

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -211,3 +211,93 @@ impl<T, ST> ArrayExpressionMethods<ST> for T where
...
+    /// # fn main() {
+    /// #     let connection = connection_no_data();
+    /// #     connection.execute("drop table if exists foos").unwrap();
+    /// #     connection.execute("create table foos ( id SERIAL PRIMARY KEY, foo Integer )").unwrap();
+    /// #     connection.execute("insert into foos (foo) values (null), (1), (2)").unwrap();
+    /// #     use self::foos::dsl::*;
+    /// assert_eq!(Ok(vec!(None, Some(2), Some(1))),
+    ///            foos.select(foo).order(foo.desc()).load(&connection));
+    /// assert_eq!(Ok(vec!(Some(2), Some(1), None)),
+    ///            foos.select(foo).order(foo.desc().nulls_last()).load(&connection));
```
killercup: dito for this example

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -211,3 +211,93 @@ impl<T, ST> ArrayExpressionMethods<ST> for T where
...
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     let connection = connection_no_data();
+    /// #     connection.execute("drop table if exists foos").unwrap();
+    /// #     connection.execute("create table foos ( id SERIAL PRIMARY KEY, foo Integer )").unwrap();
+    /// #     connection.execute("insert into foos (foo) values (null), (1), (2)").unwrap();
+    /// #     use self::foos::dsl::*;
+    /// assert_eq!(Ok(vec!(Some(1), Some(2), None)),
+    ///            foos.select(foo).order(foo.asc()).load(&connection));
```
kaj: The thing I really want to show is that the order of the elements, in this case `vec![Some(3), Some(17), Some(4711), None, None]` would have been equally correct, but `vec![Some(2), None, Some(1)]` would have been obviously wrong.  Right?

But actually showing the table definition and contents might be a good thing.  I'll take a look at it.

The bracket thing: Yes, will fix, thank you!

kaj:
Updated again after suggestions from @killercup .  Ok?

Issue #120, opened 2016-01-20
Document our overall design principles, and guidelines for contribution


sgrif:
There's a lot of implicit knowledge in much of our internals. While I try to give a lot of context to decision making in my commit messages, and you can probably figure out why a line is a certain way by git blaming it (and maybe looking at the parents of that line), but we should probably write down the most important bits. I will likely comment on this periodically with bits I think are important as I think of them, but I really don't know where we should document them

sgrif:
Type Constraints
==

Due to the nature of what we're doing, we often have data structures that need to constrain their types in several different ways. Let's take `Eq<T, U>` as an example.

`Eq<T, U>` as a data structure can reasonably exist for any values of `T` and `U`. Therefore, the actual struct definition of `T` and `U` should not have any constraints on the struct definition itself, or its inherent impl.

`Eq<T, U>` is only a valid expression if `T` and `U` are the same SQL type. Therefore, in the `impl Expression`, we should constrain that `T: Expression, U: Expression<SqlType=T::SqlType>`. This should be loosened as needed based on the various usages of the expression. Traits like `AsExpression` should never appear on the struct itself. If we want to have a DSL which takes `U: AsExpression<T::SqlType>`, those sorts of constraints should appear on the DSL itself (the `GlobalExpressionMethods#eq` method in this case)

Constraints required for conversion to SQL should appear on the impl for `QueryFragment`. The impl for `QueryFragment` *should not* care about type safety, as that has been handled at higher levels. `Eq<T, U>` would likely only have the constraints that `T: QueryFragment` and `U: QueryFragment`.

Any type that implements `Expression` should likely implement `NonAggregate` and `SelectableExpression`, based on the implementations of its members. See the body of the `infix_predicate!` macro for common examples.

sgrif:
Some Style Guidelines
====

One of the biggest lessons I've learned from maintaining Rails is that git history is *really* important, particularly the ability to get context on a line of code quickly via `git blame`. As such, much of the code is styled in such a way that lines are unlikely to change in order to support unrelated changes. Here's a few points that have come out of it:

**If a function, struct, trait, or any other definition spans more than one line, the `{` goes on its own line.**

Example:

```rust
// GOOD
fn foo(some_args: OmgVerboseTypeName, more_args: OmgVerboseTypeTwo)
    -> OurReturnType
{
    // ...
}

// BAD
fn foo(some_args: OmgVerboseTypeName, more_args: OmgVerboseTypeTwo)
[...]

PR #352, opened 2016-06-13
Move the trait bounds for `FindDsl` onto the impl

sgrif:
Since Rust does not infer the where clasue of traits when using that
trait as a bound, you effectively have to restate the entire where
clause every time you want to reference that trait. This makes `FindDsl`
basically impossible to actually abstract over in practice.

This promotes it to the same place that the rest of the query DSL is,
where it has an abstract output and a helper type to follow the return
type chain through multiple levels.

This is part of a larger audit of this module to ensure consistency and
ease of abstraction throughout.

### DIFF
--- a/diesel/src/query_dsl/filter_dsl.rs
+++ b/diesel/src/query_dsl/filter_dsl.rs
@@ -32,7 +32,7 @@ use types::Bool;
 /// assert_eq!(Ok(2), tess_id);
 /// # }
 /// ```
-pub trait FilterDsl<Predicate: Expression<SqlType=Bool> + NonAggregate> {
+pub trait FilterDsl<Predicate> {
     type Output: AsQuery;
 
     fn filter(self, predicate: Predicate) -> Self::Output;
@@ -57,7 +57,7 @@ impl<T: Table> NotFiltered for T {}
 impl<Left, Right> NotFiltered for InnerJoinSource<Left, Right> {}
 impl<Left, Right> NotFiltered for LeftOuterJoinSource<Left, Right> {}
 
-use expression::{AsExpression, SelectableExpression};
+use expression::AsExpression;
 use expression::expression_methods::*;
 use expression::helper_types::Eq;
 use helper_types::FindBy;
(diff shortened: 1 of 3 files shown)


sgrif:

diesel/src/query_dsl/filter_dsl.rs
```diff
@@ -88,19 +88,20 @@ use helper_types::FindBy;
...
+    fn find(self, id: PK) -> Self::Output;
 }
 
 impl<T, PK> FindDsl<PK> for T where
     T: Table + FilterDsl<Eq<<T as Table>::PrimaryKey, PK>>,
     PK: AsExpression<<T::PrimaryKey as Expression>::SqlType>,
-    Eq<T::PrimaryKey, PK>: SelectableExpression<T, SqlType=Bool> + NonAggregate,
-{}
+{
+    type Output = FindBy<Self, T::PrimaryKey, PK>;
```
killercup: Did you mean `Find` instead of `FindBy`?

sgrif: No, that would expand to `type Output = <Self as FindDsl<PK>>::Output`, which would be infinitely recursive.

killercup:
Totally off topic, but: Do you have a file with info on diesel internals somewhere? That one diff is full of `PK` and I'm note sure everyone can see that's mean to be say 'primary key' 😄  (That's what it means, right? I didn't just make this up, did I?)

diesel/src/query_dsl/filter_dsl.rs
sgrif: `FindBy` is just a helper for `Filter<Source, Eq<Column, Value>>` which was common enough to justify a slightly more concise alias.

sgrif:
Yes, `PK` means `PrimaryKey`. I'm fine with changing the code base to use the longer form everywhere (as a separate commit)

diesel/src/query_dsl/filter_dsl.rs
killercup: Ah, I see. I got that backwards somehow. 👍

killercup:
I actually like `PK` because it's concise. I was just wondering if you documented the meaning somewhere (e.g. regex has a nice `Hacking.md` file).

sgrif:
That's a good idea. I have some random thoughts (with a section for abbreviations) written down in https://github.com/diesel-rs/diesel/issues/120 but I haven't organized them into a coherent document yet. `PK` wasn't documented there but I've just added it.

PR #353, opened 2016-06-13
Improve the consistency of our query DSL trait bounds.

sgrif:
There were two goals here:

- Consistently put the constraint `AsQuery` on the output type. Many
  traits were using no bound or `Query` as the bound
- Enforce that the SQL type can't change for DSLs which shouldn't change
  the output type (which is basically everything except select and
  joins)

We shouldn't have to specify `SqlType=ST` in so many places, but we have
to do it here to work around
https://github.com/rust-lang/rust/issues/34257.

As part of this, I've also done the same cleanup of `LoadDsl` that we
did to `FindDsl` in #352, moving as many where clause constraints as
possible off of the trait definition and into the impl. This let us
implement `first` purely in terms of `limit` and methods from `LoadDsl`,
with a more reasonable where clause on that method. `first` was the only
place that referenced `LimitDsl` generically in our code base, and was
the original motivation for this change.

### DIFF
--- a/diesel/src/query_dsl/load_dsl.rs
+++ b/diesel/src/query_dsl/load_dsl.rs
@@ -1,6 +1,6 @@
 use connection::Connection;
 use helper_types::Limit;
-use query_builder::{Query, QueryFragment, AsQuery, QueryId};
+use query_builder::{QueryFragment, AsQuery, QueryId};
 use query_source::Queryable;
 use result::QueryResult;
 use super::LimitDsl;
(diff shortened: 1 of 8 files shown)


diesel/src/query_dsl/load_dsl.rs
```diff
@@ -9,39 +9,31 @@ use types::HasSqlType;
...
     Conn::Backend: HasSqlType<Self::SqlType>,
 {
     /// Executes the given query, returning an `Iterator` over the returned
     /// rows.
     fn load<'a, U>(self, conn: &Conn) -> QueryResult<Vec<U>> where
-        U: Queryable<Self::SqlType, Conn::Backend> + 'a,
-    {
-        conn.query_all(self)
-    }
+        U: Queryable<Self::SqlType, Conn::Backend> + 'a;
```
sgrif: This git churn irks me to no end, but there is no way to write this in a way that doesn't churn. I would have added a random semicolon on a separate line if it would have let me keep this comma. YES I AM THAT DRUNK WITH STYLE GUIDE POWER

killercup: Looking forward to your RFC to allow

```rust
fn foo<T: Bar>(bar: T) where
    T: Copy,
; // <- this guy right here
```

killercup: Or this the point where you switch to

```rust
where T: Copy
,     U: Send
```

😧

killercup:
LGTM! (Or, well, consistent ;))

diesel/src/query_dsl/load_dsl.rs
sgrif: Remove the `T: Copy` constraint and BAM CHURN

killercup: Fair enough. The easiest thing after that I can come up with right now: change that RFC to allow to put where clauses in braces.

Issue #360, opened 2016-06-23
Cannot detect unique constraint violation in a backend-agnostic way


jooert:
When trying to insert a new row with an already existing primary key, Diesel returns the generic `DatabaseError(String)`. Therefore it is impossible to tell this case apart from any other database error that might occur without parsing the (backend-specific) error message.

(This came up while implementing ruma/ruma#77.)

sgrif:
Agreed that we should include more information about the error when possible. I need to think through the API though, as I'm not sure I want to make `QueryResult` generic over the backend.

sgrif:
I guess rather than try to implement an exhaustive enum that is backend specific, a more useful API would be to identify the "common" errors that are likely to be handled programmatically, and abstract over them.

jooert:
Sounds great!

sgrif:
Anyone have opinions on

```rust
enum Error {
    DatabaseError(Box<DatabaseErrorInformation>),
    UniqueConstraintViolation(Box<DatabaseErrorInformation>),
    ForeignKeyViolation(Box<DatabaseErrorInformation>),
    InvalidTransactionState(Box<DatabaseErrorInformation>),
    // Rest of error variants, which is not related to the database specifically
}
```

vs.

```
enum Error {
    DatabaseError(DatabaseErrorKind, Box<DatabaseErrorInformation>),
    // Rest of error variants, which is not related to the database specifically
}

enum DatabaseErrorKind {
    UniqueConstraintViolation,
    ForeignKeyViolation,
    InvalidTransactionState,
    #[doc(hidden)]
    __Unknown, // Match against _ instead, more variants may be added in the future
}
```

vs.

```
enum Error {
    DatabaseError(DatabaseErrorKind, Box<DatabaseErrorInformation>),
    // Rest of error variants, which is not related to the database specifically
}

enum DatabaseErrorKind {
    ConstraintViolation(ConstraintKind),
    InvalidTransactionState,
    #[doc(hidden)]
    __Unknown, // Match against _ instead, more variants may be added in the future
}

enum ConstraintKind {
    Integrity,
    Restrict,
    NotNull,
    ForeignKey,
    Unique,
    Check,
    Exclusion,
}
```

[...]

PR #361, opened 2016-06-27
Don't leak PGResult pointers when queries failed

sgrif:
Noticed this memory leak while exploring #360. We're never freeing the
result in the error case. Ideally I'd like to have a test for this, but
I'm not entirely sure if we actually can.

### DIFF
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -22,7 +22,11 @@ impl PgResult {
                     internal_result: internal_result,
                 })
             },
-            _ => Err(Error::DatabaseError(conn.last_error_message())),
+            _ => {
+                let error_message = conn.last_error_message();
+                unsafe { PQclear(internal_result) };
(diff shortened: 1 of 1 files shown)


samphippen:
how ridiculous would it be to write a test for this?

sgrif:
The setup seems like it would be pretty trivial. We call the relevant libpq functions and pass in a result pointer here manually. I'm not sure how we actually *verify* that the result was cleared though

samphippen:
Is there a c function to check whether an address is owned by the program? Then you could check the pointer address

sgrif:
As far as I know there is no way to determine if memory has been freed or not.

samphippen:
I'd say we don't need a test here then, and this is 👍

sgrif:
I'm leaning towards having a proper type to represent the error case anyway, which would own the `PGResult` pointer and implement drop, which should more or less eliminate this case (we likely have the same leak for SQLite but I haven't checked yet)

PR #362, opened 2016-06-27
Provide richer information from `DatabaseError`

sgrif:
This is part of the restructuring required to give a reasonable solution
to #360. In the case of PG, we hold onto the result pointer, and expose
all of the most commonly relevant fields that can be returned from
`PQresultErrorField`. Of the fields we've chosen to expose, only the
primary message is guaranteed to be present.

As per libpq's documentation, no assumptions can be made about when any
other fields will or will not be present, so they unconditionally return
an option, instead of exposing a separate API for different error types
where we expect certain fields to always be present.

This works out with SQLite pretty well as well, since SQLite doesn't
actually give any additional information beyond the error code and
"extended error code". I think it might be possible for us to improve
the error message by using `sqlite3_errmsg` instead of `sqlite3_errstr`,
but that needs further exploration. Either way, short of parsing the
error message, we won't be able to populate any of the other fields so
we continue to return a string.

This restructuring has the side effect of encoding the fix for the leak
resolved by #361 in the type system, which should help to ensure that
it causes no further problems in the future.

### DIFF
--- a/diesel/src/pg/query_builder.rs
+++ b/diesel/src/pg/query_builder.rs
@@ -27,7 +27,10 @@ impl QueryBuilder<Pg> for PgQueryBuilder {
     }
 
     fn push_identifier(&mut self, identifier: &str) -> BuildQueryResult {
-        let escaped_identifier = try!(self.conn.escape_identifier(identifier));
+        let escaped_identifier = match self.conn.escape_identifier(identifier) {
+            Ok(v) => v,
+            Err(e) => return Err(Box::new(e)),
+        };
         Ok(self.push_sql(&escaped_identifier))
     }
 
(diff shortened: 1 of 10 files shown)


diesel/src/pg/query_builder.rs
```diff
@@ -27,7 +27,10 @@ impl QueryBuilder<Pg> for PgQueryBuilder {
     }
 
     fn push_identifier(&mut self, identifier: &str) -> BuildQueryResult {
-        let escaped_identifier = try!(self.conn.escape_identifier(identifier));
+        let escaped_identifier = match self.conn.escape_identifier(identifier) {
+            Ok(v) => v,
+            Err(e) => return Err(Box::new(e)),
+        };
```
sgrif: This was expanded because the relevant `From` impls that we were relying on are provided for `Box<Error>` and `Box<Error+Send+Sync>`, but not `Box<Error+Send>`. This seemed like the simplest path forward.

samphippen:
👍

PR #366, opened 2016-07-02
Provide the ability to handle unique constraint violations

### LINKED ISSUE #360: Cannot detect unique constraint violation in a backend-agnostic way

sgrif:
This lays the groundwork for a more robust error handling story, but to
start we are only handling `UniqueViolation` specifically. SQLite
separates primary key violations from unique constraints, but I've opted
to group them together.

Due to how PG handles errors, we will only know one of the constraints
that was violated (if more than one was violated), and you can only get
the constraint name, not the column name (in the single column case).
SQLite provides none of this information, and the error message would
have to be parsed to determine which constraint failed.

Ultimately I think all of this is fine, as you rarely are expecting more
than exactly one constraint to fail in a given query.

Fixes #360

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * Struct fields annotated with `#[column_name="name"]` should be changed to
   `#[column_name(name)]`.
 
+* The structure of `DatabaseError` has changed to hold more information. See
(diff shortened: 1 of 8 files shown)


killercup:
LGTM

This is a good start and probably covers 80% of the db errors in my apps 😄

PR #367, opened 2016-07-02
Simplify the structure and codegen of `UpdateStatement`

sgrif:
The main goal of this refactoring was to remove the implementations of
`SaveChangesDsl` from codegen and provide them generically instead.
Since I want the code outside of codegen to be as approachable as
possible, I made a few other changes to reduce the where clause of these
implementations. I still would like to reduce the where clause further,
but this is at a decent place. The following tree of changes were in
support of this goal:

- Identifiable needs to have the table. This is so we can generically do
  `table.find(id)` for `T: Identifiable`.
- Structs can now be passed directly to `update` and `delete`
  - This caused us to change `UpdateTarget` to be a struct, and
    introduce an `AsUpdateTarget` trait so that we can abstract over
    whether the fields on that struct are a reference or not

All of this will drastically simplify the non-procedural-macro form of
`#[derive(AsChangeset)]`, and hopefully lead to more refactoring in the
future.

It should be noted that this changes the query on SQLite from

```
UPDATE users SET ... WHERE id = 1;
SELECT * FROM users WHERE id = 1 LIMIT 1;
```

to

```
UPDATE users SET ... WHERE id = 1;
SELECT * FROM users WHERE id = 1;
```

I can get the `LIMIT 1` back in there if needed, but it makes the where clause more complex, and I don't think it matters to keep that here. I couldn't simplify it in the way I wanted to by pulling `first` into a separate trait because of https://github.com/rust-lang/rust/issues/34260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Changed
 
+* Most structs that implement `Queryable` will now also need
+  `#[derive(Identifiable)]`.
+
 * `infer_schema!` on SQLite now accepts a larger range of type names
 
 * `types::VarChar` is now an alias for `types::Text`. Most code should be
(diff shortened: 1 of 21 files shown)


samphippen:
LGTM

PR #372, opened 2016-07-04
Bash completion

davidszotten:
clap can now generate bash completion. not sure what the best way to ship this is, but at least we can generate them

```
$ diesel<TAB>
--database-url  --version       -h              help            setup
--help          -V              database        migration

$ diesel migration generate<TAB>
--help            --version         -h                <MIGRATION_NAME>
```

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -89,3 +89,23 @@ Runs the `down.sql` and then the `up.sql` for the most recent migration.
 
 [pending-migrations]: http://docs.diesel.rs/diesel/migrations/fn.run_pending_migrations.html
 [rust-dotenv]: https://github.com/slapresta/rust-dotenv#examples
+
+
+Bash completion
+---------------
+
+Diesel can generate a bash completion script for itself:
+
+#### linux
+
+```shell
+$ diesel bash-completion > /etc/bash_completion.d/diesel
+```
+
+
+#### os x (homebrew)
+
+```shell
+$ brew install bash-completion  # you may already have this installed
+$ diesel bash-completion > $(brew --prefix)/etc/bash_completion.d/diesel
+```
--- a/diesel_cli/Cargo.lock
+++ b/diesel_cli/Cargo.lock
@@ -3,7 +3,7 @@ name = "diesel_cli"
 version = "0.7.0"
 dependencies = [
  "chrono 0.2.25 (registry+https://github.com/rust-lang/crates.io-index)",
- "clap 1.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "clap 2.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "diesel 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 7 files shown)


davidszotten:
i'm pretty new to rust. is there no way for e.g. `cargo install` to ship the completion script as well?

killercup:
AFAIK cargo install only installs the binary. There is https://github.com/rust-lang/cargo/issues/2729 open requesting to install other files.

While I would like to have this, I'd probably just (a) want to use a system package manager (brew, apt; long term thing) or (b) install diesel-cli myself from the repository (short term and while developing dieselt itself of course).

davidszotten:
as a user of diesel (and by definition a rust developer), `cargo install` seems like a perfectly fine way to install the cli to me. it's unfortunate that we can't ship the generated completion script, and, it seems the only entry point to `gen_completions` is to give it an output dir. this seems unintuitive to me (e.g. "please run `diesel generate-bash-completion /etc/bash_completion.d` to have it write a file in that directory)

thoughts on such an api? might it be possible for the `build.rs` in this pr to read the generated bash code and then compile the contents into the app (to support something like `diesel generate-bash-completion >  /etc/bash_completion.d/diesel`

diesel_cli/README.md
```diff
@@ -89,3 +89,23 @@ Runs the `down.sql` and then the `up.sql` for the most recent migration.
 
 [pending-migrations]: http://docs.diesel.rs/diesel/migrations/fn.run_pending_migrations.html
 [rust-dotenv]: https://github.com/slapresta/rust-dotenv#examples
+
+
+Bash completion
+---------------
+
+Diesel can generate a bash completion script for the itself:
```
killercup: probably: s/for the itself/for the CLI

davidszotten: yup. thanks

davidszotten:
don't love `generate-bash-completion` as a name so open to suggestions. also, should it be "hidden" under some sub-command?

davidszotten:
looks like rustup failed on one of the travis jobs. think you need more access than i have to re-trigger the build

davidszotten:
clap can now generate completions to a string; no need for codegen

davidszotten:
just to follow up on this. if this feature isn't suitable for this project, please let me know and i'll close the pr. if this is interesting, but there are issues with the implementation, i'll happily work with you to fix/improve

mcasper:
Sorry for the long absence, I think I agree that this is the best approach we can support within Diesel right now.

Maybe shorten the command name to just `bash-completion`? To me it just seems a little friendlier to type, and most users should be familiar with what it is.

Also, looks like there are some conflicts to deal with now :/

davidszotten:
no worries. command shortened and commits rebased+squashed

sgrif:
Thanks. I manually merged this in https://github.com/diesel-rs/diesel/commit/17326dd26ab4a57862784b4c4d2dfffb0325b4a7. There were some style issues that needed to be addressed which I did by amending your commit to avoid git churn. As such Github won't automatically pick up on me merging it, but it has been accepted. Thank you for working on this.

davidszotten:
thanks!

Issue #373, opened 2016-07-05
infer_schema! is treating PostgreSQL views as tables.


cascalheira:
First of all, thank you all for your great work on Diesel!

I am using PostGIS in a Postgres database and i was just starting to try to use diesel to access it. The main problem so far is if i use infer_schema! i get this error:

`error: table geography_columns has 0 primary keys, only one is currently supported`

It seems that infer_schema! is treating views as if they were tables. Is there any work around for this? 
Unfortunately my rust-fu is not yet strong enough for me to help with a proper solution :)

Edit: Just submitted a pull request that solves this issue: https://github.com/diesel-rs/diesel/pull/374

PR #374, opened 2016-07-05
infer_schema! will no longer treat views as tables

cascalheira:
When using PostgreSQL infer_schema! treats views as tables. This will produce errors like:
`error: table geography_columns has 0 primary keys, only one is currently supported`

On this pull request i just added `AND table_type NOT LIKE 'VIEW'` to the SQL query that fetches all the table names.


killercup:
Wow, that was quick! It might be nice to make this configurable in the future but if this breaks your current setup I think this is good as-is.

Can you mention "Fix #373" somewhere in the commit message? This will automatically close that issue when this is merged.

It would be great if you could also add a test for this. I think there are already some teste for `infer_schema!`, so you could probably just add a view to the db schema.

cascalheira:
@killercup Thank you for the fast reply. I just changed the commit message and added "Fix #373" to it.

About the tests, i will try to do that later today or tomorrow. I am quite new to rust and need to figure how to write that test (from what i saw so far, it shouldn't be hard, but i still need the time to figure it out).

cascalheira:
@killercup As making it optional, i don't really think that is a good idea because views don't actually have primary keys. In my opinion, views should be dealt in some other way.

sgrif:
I'm fine with this until we add better support for nonstandard primary keys. Can you add a test for this and a changelog entry?

cascalheira:
@sgrif I just added a changelog entry and modified the SQL query so that instead of excluding views from the query, it only fetches 'BASE TABLES'.

As diesel doesn't have any kind of views support i don't really see the point of adding another test. This will just enforce propper behaviour when fetching tables (views aren't really tables).

Let me know if you need me to change something else.

sgrif:
Well one argument for adding a test: I have no clue if we need to make an equivalent change for SQLite or not.

cascalheira:
Seems like SQLite also supports views and also fetches them in the same way, alongside with normal tables: [Stack Overflow](http://stackoverflow.com/questions/82875/how-to-list-the-tables-in-an-sqlite-database-file-that-was-opened-with-attach). I don't use SQLite but next week i can do a new pull request with the SQLite fix.

As it currently is, diesel crashes if we have a view (views are just queries, not tables and as such have no primary keys) so the current tests would already be what we need to test this situation. If you still want me to add something to the tests just for the sake of it, i will just add the SQL to create a view and check if it wasn't fetched (If it was, it would crash diesel anyway when using `infer_schema!`).

sgrif:
Yes, I would prefer if a minimal test case could be added for this as
you've described. I'll write the patch to fix this for SQLite as well.

killercup:
I think this already works for SQLite because of [this line](https://github.com/diesel-rs/diesel/blob/bb1587768bb0dc7c18e8752848f3c9123bea70bb/diesel_codegen_syntex/src/schema_inference/sqlite.rs#L107).

I'm going to rebase this and write a small test. Or, actually, I'm just going to add views to the diesel_test migrations, as there don't seem to be any tests for infer_schema itself.

sgrif:
I think just a test for `load_table_names` that creates a view should be sufficient.

killercup:
@cascalheira I rebased your branch on master and added some tests. #434 includes your commits (with you as author, of course), so I'll close this PR. Feel free to comment over there :)

cascalheira:
@killercup Sorry for the late answer i've been quite busy lately :)
Thank you for including that as it will make possible for me to use the master branch on my projects.

PR #382, opened 2016-07-25
Implement `BelongingToDsl` generically

sgrif:
I'm working on the non-procedural macro implementation of associations,
and this thing would have been a pain to do from a normal macro. The
only thing we needed that we didn't already have on other traits was the
foreign key. I've thrown the appropriate method on the `BelongsTo`
trait.

The where clause is a touch hairier than I'd like. I really need a
concise way to express "this is a table, and this is a column on this
table, everything that stems from just these two will be valid"


sgrif:

PR #383, opened 2016-07-27
Implement a stable version of `BelongsTo`

sgrif:
This requires a bit more information to be explicitly given in the
options, as we can no longer infer the table name or the foreign key
from the relevant struct names. This differs from the procedural form,
as you state the name of the struct that you belong to, not a lower
cased singular form of the table (I'm considering changing the
procedural form to match)

Since we need the table name potentially available in multiple places,
I've modified the only other place that is related and explititly needs
it (Identifiable) to match with a new annotation. As usual, I've had to
do `#[table_name(value)]` instead of `#[table_name="value"]`, as there's
no way for me to match the second form as an ident.

Due to the proposed changes in [RFC 1681][rfc-1681], I'm considering
eventually having the "public API" form of this use annotations simlar
to the current procedural macros, and then an `Associations!` macro
which walks over each of them, which would look something like this:

    #[derive(associations)]
    #[has_many(Comment, foreign_key = post_id)]
    #[belongs_to(User, foreign_key = user_id)]
    #[table_name(posts)]
    pub struct Comment {
        id: i32,
        user_id: i32,
    }

As I write this, I am realizing that the way I'm passing the foreign key
isn't actually a valid meta item once this goes to stable, and I have
no way to pattern match on it. I'll probably need to change it to
`foreign_key(post_id)` in the future...

Anyway the change to that form can still call this macro under the hood,
it would only need to be a documentation change, so I think we can move
forward with this interface for now.

[rfc-1681]: https://github.com/rust-lang/rfcs/pull/1681

### DIFF
--- a/diesel/src/macros/associations/belongs_to.rs
+++ b/diesel/src/macros/associations/belongs_to.rs
@@ -128,11 +128,13 @@ macro_rules! BelongsTo {
             }
         }
 
-        __diesel_belongs_to_joinable_impl!(
-            parent_table_ty = <$parent_struct as $crate::associations::Identifiable>::Table,
-            parent_table_expr = <$parent_struct as $crate::associations::Identifiable>::table(),
-            child_table = $child_table_name::table,
+        joinable_inner!(
+            left_table_ty = $child_table_name::table,
+            right_table_ty = <$parent_struct as $crate::associations::Identifiable>::Table,
+            right_table_expr = <$parent_struct as $crate::associations::Identifiable>::table(),
             foreign_key = $child_table_name::$foreign_key_name,
+            primary_key_ty = <<$parent_struct as $crate::associations::Identifiable>::Table as $crate::Table>::PrimaryKey,
+            primary_key_expr = $crate::Table::primary_key(&<$parent_struct as $crate::associations::Identifiable>::table()),
         );
 
         $(select_column_inner!(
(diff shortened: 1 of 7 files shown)


sgrif:

diesel/src/macros/associations/belongs_to.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// [identifiable]: prelude/trait.Identifiable.html
+///
+/// # Options
+///
+/// ## foreign_key
+///
+/// Required. The name of the foreing key column for this association.
+///
+/// # Example
```
killercup: convention in std is to always write Examples

diesel/src/macros/associations/belongs_to.rs
```diff
@@ -0,0 +1,167 @@
...
+/// }
+/// # fn main() {}
+/// ```
+///
+/// To avoid copying your struct definition, you can use the
+/// [custom_derive crate][custom_derive].
+///
+/// [custom_derive]: https://crates.io/crates/custom_derive
+///
+/// ```ignore
```
killercup: You could make custom_derive a dev-dependency and doc test this as well

diesel/src/macros/associations/belongs_to.rs
```diff
@@ -0,0 +1,167 @@
...
+        __diesel_parse_struct_body! {
+            (
+                struct_name = $struct_name,
+                $($args)*
+            ),
+            callback = BelongsTo,
+            body = $body,
+        }
+    };
+}
```
killercup: Macro looks good as far as I can tell. Nicely commented :)

killercup:
LGTM, left a few inline comments.

Btw, I think `#[table_name(posts)]` is pretty weird, because it looks like a function call – of a setter method –, but doesn't read like one. I would really like to see `#[table_name="posts"]`, but I understand that current macros don't support that. `#[set_table_name("posts")]` would be a bit better (yes that's a string not an ident that comes from god knows where), but not enough to justify the change. Thanks for reading this rant 😄

PR #385, opened 2016-07-29
Allow inserting empty slices

### LINKED ISSUE #384: Insert is crashing while trying to insert an empty vector

Backtrace:

```
stack backtrace:
...
```

killercup:
Handle the case where an empty slice is to be inserted into a table. The expected behaviour is to just skip the execution of the query. This crashed previously when trying to get the column names of the fields.

Fixes #384

### DIFF
--- a/diesel/src/persistable.rs
+++ b/diesel/src/persistable.rs
@@ -17,10 +17,21 @@ pub trait Insertable<T: Table, DB: Backend> {
     fn values(self) -> Self::Values;
 }
 
+/// Values that can be inserted into a row
 pub trait InsertValues<DB: Backend> {
+    /// Add column names to `out` query
     fn column_names(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult;
+
+    /// Add list of values to `out` query
     fn values_clause(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult;
+
+    /// Add value bind params to `out` bind collector
     fn values_bind_params(&self, out: &mut DB::BindCollector) -> QueryResult<()>;
+
+    /// Determine if this list of values is actually empty
+    ///
+    /// This will be used to determine whether the `INSERT` can be skipped
+    fn is_empty(&self) -> bool { false }
 }
 
 #[derive(Debug, Copy, Clone)]
@@ -60,6 +71,7 @@ impl<'a, T, U, DB> Insertable<T, DB> for &'a Vec<U> where
     }
 }
 
+/// Represents multiple rows that will be inserted at the same time
 #[derive(Debug, Copy, Clone)]
 pub struct BatchInsertValues<'a, T, U: 'a, DB> {
     values: &'a [U],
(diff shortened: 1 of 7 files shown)


killercup:
I'm not sure if this implementation is elegant. I'm especially wondering whether the "insert/update is empty" state can be more cleverly encoded in the type system. The methods I added should, to a large extend, be constant evaluable and will probably be inlined anyway. Are there any other edge cases we need to deal with here?

sgrif:
> I'm especially wondering whether the "insert/update is empty" state can be more cleverly encoded in the type system.

Having `column_names` return an option seems like a reasonable place to do that, since that's the place that we're doing something which can panic. I'm not sure if that would be too implicit though, with the structure that we have. If column names returned something rather than operating on the query builder, I'd feel better about it but we'd need `const fn` to do that efficiently.

For now I think this is fine. However, I don't like having an additional on `QueryFragment`. Let's just have `to_sql` on `InsertStatement` and `InsertQuery` do nothing if `InsertValues#is_empty` returns `true`. One other option would be to do something like `Err(AbortQuery)`, but have the connection handle it.

killercup:
> For now I think this is fine. However, I don't like having an additional on QueryFragment. Let's just have to_sql on InsertStatement and InsertQuery do nothing if InsertValues#is_empty returns true.

I tried that at first (changing `<InsertStatement as QueryFragment>::to_sql` to do nothing), but that would mean that the QueryBuilder is still used to execute a query late, right? And that query is invalid (incomplete or empty).

Or maybe I just forgot something. (What are your reasons for not wanting an `is_empty` method, by the way?)

> One other option would be to do something like Err(AbortQuery), but have the connection handle it.

Interesting idea; this could potentially be re-used for other circumstances where queries should be aborted for some reason (though I can't come up with another case right now). The overhead of doing this is potentially the same as checking for `is_empty` but could be done higher up (in the connection handler).

Let me know what you want to do. (Also, feel free to change is in some other way and skip this PR entirely, I'm okay with that.)

sgrif:
An empty query *should* be valid IIRC, but we can always just have the connection do nothing if the query builder constructs an empty string.

> (What are your reasons for not wanting an `is_empty` method, by the way?)

It only affects this one case (and I guess potentially update). Don't need to pollute the entire AST with it. Especially since we can just see if the query builder constructs an empty string.

> Let me know what you want to do.

For now let's just have it construct an empty query, and then check in the connection for an empty string if needed. We can always change to the abort idea later on if we feel like there's a good use case.

killercup:
@sgrif updated to return a (new) `Error::EmptyQuery` and handle that in `load_dsl`.

sgrif:
`get_result` and `get_results` will need to handle it as well.

killercup:
I discussed this with @sgrif and we want to reactor the QueryBuilder to not return a `Box<std::error::Error>` but a `QueryBuilderError` that has an `AbortQuery` variant. I'll write a PR that refactors how errors are handled here (and elsewhere, and probably introduce the quick-error to make this more concise), and try to land it for 0.8 (it's a breaking change).

PR #386, opened 2016-07-30
Implement `HasMany` as a stable non-procedural macro

sgrif:
This follows the same approach as belongs to. I've also changed the
procedural form to more or less share code with the non-procedural, so
that we get access to `$crate`.

When I started this process, I was really hoping to have the procedural
version just parse the options and give better error messages if they
are invalid, then call the stable form. Unfortunately, because of how
macro expansion works, I can't just "insert the struct body into the
macro call", as at that point the struct is a single token and can't be
matched on. I'll probably end up still doing this eventually, and have
the procedural form just call the entry point that occurs after
`__diesel_parse_struct_body`.

### DIFF
--- a/diesel/src/macros/associations/has_many.rs
+++ b/diesel/src/macros/associations/has_many.rs
@@ -1,0 +1,151 @@
+/// Defines a one-to-many association for the parent table. This macro is only required if you need
+/// to join between the two tables. This macro should be called with the name of the child table,
+/// followed by any options, followed by the entire struct body. The struct *must* be annotated with
+/// `#[table_name(name_of_table)]`. Both the parent and child structs must implement
+/// [`Identifiable`][identifiable].
+///
+/// [identifiable]: prelude/trait.Identifiable.html
+///
+/// # Options
+///
+/// ## foreign_key
+///
+/// Required. The name of the foreign key column for this association.
+///
+/// # Examples
+///
+/// ```no_run
+/// # #[macro_use] extern crate diesel;
+/// # table! { users { id -> Integer, } }
+/// # table! { posts { id -> Integer, user_id -> Integer, } }
+/// pub struct User {
+///     id: i32,
+/// }
+/// # Identifiable! { #[table_name(users)] struct User { id: i32, } }
+///
+/// pub struct Post {
+///     id: i32,
+///     user_id: i32,
+/// }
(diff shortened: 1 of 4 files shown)


sgrif:

diesel/src/macros/associations/has_many.rs
```diff
@@ -0,0 +1,151 @@
+/// Defines a one-to-many association for the parent table. This macro is only requried if you need
```
killercup: s/requried/required

diesel/src/macros/associations/has_many.rs
```diff
@@ -0,0 +1,151 @@
...
+/// to join between the two tables. This macro should be called iwth the name of the child table,
```
killercup: s/iwth/with

diesel/src/macros/associations/has_many.rs
```diff
@@ -0,0 +1,151 @@
...
+/// followed by any options, follwed by the entire struct body. The struct *must* be annotated with
```
killercup: s/follwed/followed

diesel/src/macros/associations/has_many.rs
```diff
@@ -0,0 +1,151 @@
...
+/// `#[table_name(name_of_table)]`. Both the parent and child structs must implement
+/// [`Identifiable`][identifiable].
+///
+/// [identifiable]: prelude/trait.Identifiable.html
+///
+/// # Options
+///
+/// ## foreign_key
+///
+/// Required. The name of the foreing key column for this association.
```
killercup: s/foreing/foreign

diesel/src/macros/associations/has_many.rs
```diff
@@ -0,0 +1,151 @@
...
+    };
+
+    // Receive parsed fields of normal struct from `__diesel_parse_struct_body`
+    // These patterns must appear above those which start with an ident to compile
+    (
+        (
+            struct_name = $struct_name:ident,
+            parent_table_name = $parent_table_name:ident,
+            child_table = $child_table:path,
+            foreign_key = $foreign_key:path,
```
killercup: Are these the only fields ever given here? Would it make sense/this more future-compatible if you add a generic token tree match at the end? Like `$($other_headers:tt)*`

killercup:
Nice work, I like that you try to reuse this in the procedural macro :)

Left a few comments inline, LGTM all in all.

PR #387, opened 2016-07-30
Rejigger procedural macros to simply call the stable macros

sgrif:
This implements a procedural version of `__diesel_parse_struct_body`,
and calls into the stable equivalent of each macro at the entry point
after that step in the non-procedural process. This affects all
procedural macros that have stable equivalents except `Queryable`, which
appears to have some divergence in terms of what is supported that I
need to address, and `AsChangeset`, which isn't yet merged.

### DIFF
--- a/diesel/src/macros/associations/belongs_to.rs
+++ b/diesel/src/macros/associations/belongs_to.rs
@@ -68,6 +68,7 @@ macro_rules! BelongsTo {
             (
                 parent_struct = $parent_struct,
                 foreign_key_name = $foreign_key_name,
+                optional_foreign_key = "false",
             )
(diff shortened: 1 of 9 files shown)


sgrif:

killercup:
Wow, that's quite a bit of code that you managed to get rid of! I guess `to_stable_macro_tokens` is the main addition here. I can't pretend to understand everything that goes on here but it looks consistent and useful, so 👍 from me :)

Travis fails in `diesel_tests/target/debug/build/diesel_tests-10b02b5ec417d2a0/out/lib.rs:206:9` with "expected type `i32`, found type `std::option::Option<i32>`".

PR #390, opened 2016-08-01
Make the API of procedural `belongs_to` mirror that of non-procedural

sgrif:
The stable macro had to take the struct name as the argument, and this
is the API I want to go with going forward. I'm hoping to converge the
procedural and non-procedural forms as much as possible.

### DIFF
--- a/diesel_codegen/src/associations/belongs_to.rs
+++ b/diesel_codegen/src/associations/belongs_to.rs
@@ -1,10 +1,6 @@
-use syntax::ast::{
-    self,
-    MetaItem,
-};
+use syntax::ast::MetaItem;
 use syntax::codemap::Span;
 use syntax::ext::base::{Annotatable, ExtCtxt};
-use syntax::parse::token::str_to_ident;
 
 use super::{parse_association_options, to_foreign_key};
 use util::is_option_ty;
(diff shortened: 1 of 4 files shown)


sgrif:

diesel_codegen/src/associations/belongs_to.rs
```diff
@@ -19,10 +15,10 @@ pub fn expand_belongs_to(
 ) {
     let options = parse_association_options("belongs_to", cx, span, meta_item, annotatable);
     if let Some((model, options)) = options {
-        let association_name = options.name.name.as_str();
+        let parent_struct = options.name;
         let struct_name = model.name;
 
-        let foreign_key_name = to_foreign_key(&association_name);
+        let foreign_key_name = to_foreign_key(&parent_struct.name.as_str());
```
killercup: &....as_str()? You might be able to get rid of that & IIRC

killercup:
Lgtm!

diesel_codegen/src/associations/belongs_to.rs
sgrif: `as_str()` returns an `InternedString` so it needs the additional deref to get to `&str`

sgrif: (There is almost certainly a better way than this to go from ident to str)

PR #391, opened 2016-08-01
Allow tuples to be used with `grouped_by`

sgrif:
The current implementation of associations basically ignores multiple
level nesting. It lets us get `Vec<(User, Vec<Post>)>`, but not
`Vec<(User, Vec<(Post, Vec<Comment>)>)>`. This allows a tuple of
arbitrary size to be used with `grouped_by`, as long as the first
element in the tuple is the actual child used in the association. This
means that `Vec<(Post, Vec<Comment>)>` implements `GroupedBy<User>`.
I've also moved the impl off of `Vec` and onto `T: IntoIterator`, to get
rid of some unneccesary `collect` calls in tests.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -16,7 +16,8 @@ pub trait GroupedBy<Parent>: IntoIterator + Sized {
     fn grouped_by(self, parents: &[Parent]) -> Vec<Vec<Self::Item>>;
 }
 
-impl<Parent, Child> GroupedBy<Parent> for Vec<Child> where
+impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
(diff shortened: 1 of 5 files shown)


killercup:
Lgtm.

Are there any other places where you used Vec speicifically? (Though it's not a breaking change to change to IntoIter, right? Should strictly allow more stuff.)

Also, that tuple macro is crazy but still getting crazier. I like it 😄

sgrif:
> Though it's not a breaking change to change to IntoIter, right?

Given this changes a feature that hasn't been released yet, it's definitely not a breaking change. ;)

> Also, that tuple macro is crazy but still getting crazier.

Yeah, I hesitated to make this change for that reason. Everything we add to that macro has a non-trivial hit to compile times. This one appears to be less bad since it has only 2 constraints.

PR #394, opened 2016-08-02
Remove the `Copy` constraint from `Identifiable`

sgrif:
I had originally wanted to returned owned values here because I'm
thinking that when we eventually add support for composite primary keys,
I'll need to return a tuple of two fields from this method. I've gone
back and forth on a bunch of different options, but ultimately I think
I'm going to change this to return `Cow<'a, Self::Id>`. However, I don't
want to make that change pre-emptively, as I'm still not sure exactly
what composite primary keys will look like.

Other alternatives:

- Implement `Identifiable` on `&'a T` -- makes reasoning about it a pain
- Change `Identifiable` to be `Identifiable<'a>` -- makes referencing it
  a pain
- Change `id` to `with_id`, and have it take a `FnOnce(&Self::Id)`. This
  would work, but would be really weird and just returning `Cow<'a,
  Self::Id>` makes way more sense.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -8,7 +8,7 @@ use super::Identifiable;
 pub trait BelongsTo<Parent: Identifiable> {
     type ForeignKeyColumn: Column;
 
-    fn foreign_key(&self) -> Parent::Id;
+    fn foreign_key(&self) -> &Parent::Id;
     fn foreign_key_column() -> Self::ForeignKeyColumn;
 }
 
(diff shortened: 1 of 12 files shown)


killercup:
Okay, so this change adds a bunch of lifetimes and references all over the place – and not enough according to travis 😄 

Composite indices will probably be exception, so I wouldn't want to make life hard for "regular" users just to make composite indices really nice. But on the other hand: I really like `Cow` and think it should get way more exposure (and some more default impls).

Another option would be a custom wrapper around `Cow<'a, Id>` (that de-refs to Cow) but has some custom index-specific behaviour (if that makes sense).

PR #397, opened 2016-08-08
WIP Error refactoring

killercup:
Work in progress – I began working on this last week when it became apparent this would be required for #385.

Todo:
- [x] Reduce boilerplate code by using quick-error
- [ ] Get rid of `Box<Error>` in `QueryBuilder` by writing specific new error types


beatgammit:
Any progress on this? I'm excited about using diesel with the new macros 1.1 support, and this looks like the last major blocker.

killercup:
Sorry, I haven't had much time to work on this. It should actually not be a block for a macros 1.1 release and can later than 0.8. (Though it will be a breaking change for people depending on these error types.)

weiznich:
I would like to look into this. Is there any specific reason why quick_error and not error_chain is used?

killercup:
Just that I'm more familiar with quick_error. What advantages does error_chain give us?

weiznich:
I'm not that familiar with quick_error, but I'm using error_chain in an other project. 
As far as I understand error_chain wraps quick_error and adds some stuff on top of this. For example it let you easily link other error_chain generated errors into your chain. Also there is some code around the actual error-enum generated (Implementation of Display and std::error::Error). Furthermore the error-type contains an optional backtrace. Take a look at the [documentation](https://docs.rs/error-chain/0.5.0/error_chain/) for more details.

What are the next steps for this pull-request? Changing `BuildQueryResult` form `Result<(), Box<Error+Send>>` to something generated exact error and then fixing all compiler errors?

weiznich:
I've opened a [pull-request](https://github.com/killercup/diesel/pulls) to @killercup repo to implement the remaining point.

killercup:
I'm closing this as it's just a bunch of conflicts right now. I'd really like to spend some time doing this right, with a good forward-compatibility story, before we go 1.0.

Issue #388, opened 2016-07-30
Rust 1.10 stabilised {Occupied,Vacant,}Entry::key


killercup:
A comment [here](https://github.com/diesel-rs/diesel/blob/e2490c38d5c4cc38a38d31c5ff9d4272febfa8fb/diesel/src/pg/connection/stmt/cache.rs#L60) says:

> ```rust
> // FIXME: This can be cleaned up once https://github.com/rust-lang/rust/issues/32281
> // is stable
> ```

This has been stabilised in Rust 1.10 (cf. https://github.com/rust-lang/rust/pull/33699).

cengizIO:
Hello!

I can work on this, if there's nobody already assigned.

sgrif:
Have at it!

cengizIO:
Hello @killercup, @sgrif 

I replaced old if block with new matchers but currently having borrowing/ownership issues.

I brought a Postgres docker instance up and fixing tests one by one.

Will submit a PR in following days.

Thanks

gsquire:
This could help too: https://gist.github.com/sgrif/68cbba9a87084da3a2e8c42a12e7d6cf

cengizIO:
Sorry it's been a very busy week. 

I'm back to working on this.

PR #400, opened 2016-08-11
Cleanup prepared statement caching #388

cengizIO:
Hello there @killercup @sgrif @gsquire!

This is my first PR to Diesel project.

As stated in the #388 rust 1.10 now has stabilized  `std::collections::hash_map::Entry::{Occupied, Vacant}` so there's no need for an if block.

Any feedback is more than welcome.

Cheers!

### DIFF
--- a/diesel/src/pg/connection/stmt/cache.rs
+++ b/diesel/src/pg/connection/stmt/cache.rs
@@ -54,32 +54,36 @@ impl StatementCache {
         bind_types: Vec<u32>,
     ) -> QueryResult<Rc<Query>> {
         use std::borrow::Cow;
+        use std::collections::hash_map::Entry::{Occupied, Vacant};
 
+        let cache_suffix = self.cache.borrow().len() + 1;
         let mut cache = self.cache.borrow_mut();
         let (cache_key, maybe_binds) = try!(cache_key(conn, source, bind_types));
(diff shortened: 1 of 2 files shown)


killercup:
LGTM!

Do we have a place where we say which Rust versions we support? Maybe just mention it again in the changelog?

cengizIO:
@killercup Should we open an issue for it?

killercup:
@cengizIO, FYI, @sgrif just [said](https://gitter.im/diesel-rs/diesel?at=57ac797ff7f11a7936b59fbb):

> Until 1.0, I think it's reasonable that we only support the latest version of Rust. Once we hit 1.0 we can evaluate if we need to do something different.

It'd just add this as a note to the Changelog to remind people to update to Rust 1.10 if they haven't yet.

sgrif:
Can you do the same to the SQLite connection?

cengizIO:
@sgrif Oops. I missed your comment.

cengizIO:
@killercup I guess changelog is only updated while tagging releases. Am I right?

@sgrif I just added the sqlite part.

sgrif:
No the change log is kept up to date. This doesn't need a change log entry though as it's invisible to users.

killercup:
Sorry, I seem to have forgotten to check up on this PR. Looks like this is good to go! Thanks again for working on this, @cengizIO!

PR #401, opened 2016-08-11
Support nullable foreign keys in associations

sgrif:
The `BelongsTo!` macro now does crazy hacks to pull out the foreign key field
from the list given, checks if it's an option type, and will generate different
code as a result. The types that `BelongingToDsl` and `GroupedBy` get
implemented for are the same as if the foreign key was not nullable.
`GroupedBy` will simply discard any records with `None` for the foreign key
(which it is unlikely to receive in the first place since the result of
`BelongingToDsl` implicitly never returns records with a null value for the
foreign key.)

This differs from the direction I had previously discussed, where we would
implement the various traits for `Option<Parent>`. There are two main reasons
for this:

- The impl of `BelongingToDsl` would be gross, as we now have to care about
  `NULL` appearing anywhere and give it special handling.
- In the real world, you're never dealing with "records belonging to these 3
  parents, and records which have no parent".

For simplication in the code, I've set up `BelongsTo` to always return an
option for the foreign key. This causes an unneccesary conditional to appear in
the body of `GroupedBy`. However, I'm fairly certain that conditional will get
optimized away after inlining. I have not checked the generated ASM or LLVM IR
to confirm this.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -8,7 +8,7 @@ use super::Identifiable;
 pub trait BelongsTo<Parent: Identifiable> {
     type ForeignKeyColumn: Column;
 
-    fn foreign_key(&self) -> &Parent::Id;
+    fn foreign_key(&self) -> Option<&Parent::Id>;
     fn foreign_key_column() -> Self::ForeignKeyColumn;
 }
 
@@ -27,8 +27,9 @@ impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
         let id_indices: HashMap<_, _> = parents.iter().enumerate().map(|(i, u)| (u.id(), i)).collect();
         let mut result = parents.iter().map(|_| Vec::new()).collect::<Vec<_>>();
         for child in self {
-            let index = id_indices[child.foreign_key()];
-            result[index].push(child);
+            if let Some(index) = child.foreign_key().map(|i| id_indices[i]) {
+                result[index].push(child);
+            }
         }
         result
     }
--- a/diesel/src/macros/associations/belongs_to.rs
+++ b/diesel/src/macros/associations/belongs_to.rs
@@ -68,7 +68,6 @@ macro_rules! BelongsTo {
             (
                 parent_struct = $parent_struct,
                 foreign_key_name = $foreign_key_name,
-                optional_foreign_key = "false",
             )
             $($rest)*
         }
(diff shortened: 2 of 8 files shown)


sgrif:

diesel/src/associations/belongs_to.rs
```diff
@@ -8,7 +8,7 @@ use super::Identifiable;
 pub trait BelongsTo<Parent: Identifiable> {
     type ForeignKeyColumn: Column;
 
-    fn foreign_key(&self) -> &Parent::Id;
+    fn foreign_key(&self) -> Option<&Parent::Id>;
```
killercup: This is a private method on a public trait. Thus, not a breaking change, right?

sgrif: All methods on traits are public. This is a breaking change.

diesel/src/associations/belongs_to.rs
```diff
@@ -27,8 +27,9 @@ impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
         let id_indices: HashMap<_, _> = parents.iter().enumerate().map(|(i, u)| (u.id(), i)).collect();
         let mut result = parents.iter().map(|_| Vec::new()).collect::<Vec<_>>();
         for child in self {
-            let index = id_indices[child.foreign_key()];
-            result[index].push(child);
+            if let Some(index) = child.foreign_key().map(|i| id_indices[i]) {
+                result[index].push(child);
+            }
         }
```
killercup: I think there is a way to write that a little cleaner/functional, with flat_map and collect/fold. I'm just reading through this quickly though, and maybe come back later with an actual refactor suggestion.

diesel/src/associations/belongs_to.rs
```diff
@@ -27,8 +27,9 @@ impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
         let id_indices: HashMap<_, _> = parents.iter().enumerate().map(|(i, u)| (u.id(), i)).collect();
```
killercup: `parents` is a slice you immediately call `iter` on (two times)—why not take an `IntoIter<Item=Parent>` instead? Might be cheaper to clone/work with. Totally unimportant though, I guess you always deal with `Vec`s here anyway.

diesel/src/associations/belongs_to.rs
sgrif: It could probably be changed to

```rust
for (index, child) in self.into_iter().filter_map(|child| child.foreign_key().map(|k| (id_indices[k], child))
```

I can't turn the whole thing into a functional statement, as there's no method out there for this. The equivalent in Ruby would be something along the lines of `group_by { |child| id_indices[foreign_key] }`, but neither Rust nor itertools provide that method (itertools has a method with that name that does something extremely different)

diesel/src/macros/associations/belongs_to.rs
```diff
@@ -102,22 +101,105 @@ macro_rules! BelongsTo {
...
+        (
+            fields = $fields:tt,
+            $($remaining_args:tt)*
+        ),
+        found_field_with_column_name = $ignore:ident,
+        field = {
+            field_name: $ignore2:ident,
+            column_name: $ignore3:ident,
+            field_ty: $ignore4:ty,
+            field_kind: $foreign_key_kind:ident,
```
killercup: took me a while to see the magic to determine if this is an option is already part of `__diesel_parse_struct_body` :)

diesel/src/associations/belongs_to.rs
killercup: Actually, what put me off at first about this was the HashMap—is this really necessary? I guess so, and I don't see a better way. Maybe use a hasher like Fnv which is optimised for smaller keys? Downsides might be that it's (a) not worth the effort (though I think this gets called quite a lot for each association), and (b) that this is not a cryptographic hasher, so it might be a security risk if users can influence the `id`s to do a collision attack.

Yes, I'm totally overthinking this.

diesel/src/associations/belongs_to.rs
```diff
@@ -27,8 +27,9 @@ impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
...
         let mut result = parents.iter().map(|_| Vec::new()).collect::<Vec<_>>();
```
killercup: Wait, are you actually just doing `let mut result = vec![vec![]; parents.len()]`; here?

killercup:
I left a bunch of totally unrelated comments, *but* i also followed the macro callbacks, which LGTM!

diesel/src/associations/belongs_to.rs
sgrif: I mean the keys aren't necessarily always small. String keys are a thing and the user can do literally anything here. That said, I'm open to using a different hasher if there's a good reason. I'm not sure anything can end up being a security risk here, the worst it could be is a DOS vector since collisions would mean slower execution.

diesel/src/associations/belongs_to.rs
sgrif: Yes. I have to arbitrarily do `Child: Clone` to write it that way though, since even though it's an empty vector we still only have `impl<T: Clone> Clone for Vec<T>`

PR #407, opened 2016-08-17
Better document the `types` module

### LINKED ISSUE #251: Better document which type mappings exist, as well as optional features
We don't really have any docs on what actually implements `ToSql` and `FromSql`. The implementors lists on those traits are incomplete, and we don't really mention the `chrono` option anywhere. We should document these things at the module level for the types module.

sgrif:
I've seen a lot of people making the mistake of trying to put things
like `diesel::types::Timestamp` in their structs. This attempts to
better document the role of the `types` module, and documents any
`ToSql` and `FromSql` impls that are provided. I've un-inlined the pg
types module, as stating "this is a postgres specific type" in the
documentation over and over again didn't feel right. I've also made sure
that the UUID feature is turned on when generating docs.

Fixes #251.

### DIFF
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -1,5 +1,17 @@
 //! Types which represent a native SQL data type, and the conversions between
-//! them and Rust primitives. Additional types can be added by other crates.
+//! them and Rust primitives. The structs in this module are *only* used as
+//! markers to represent a SQL type, and shouldn't be used in your structs. See
+//! the documentation for each type to see the Rust types that can be used with
+//! a corresponding SQL type. Additional types can be added by other crates.
+//!
+//! To see which Rust types can be used with a given SQL type, see the
+//! "Implementors" section of the [`ToSql`][ToSql] and [`FromSql`][FromSql]
+//! traits, or see the documentation for that SQL type.
+//!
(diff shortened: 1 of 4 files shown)


sgrif:

diesel/src/types/mod.rs
```diff
@@ -1,5 +1,17 @@
 //! Types which represent a native SQL data type, and the conversions between
-//! them and Rust primitives. Additional types can be added by other crates.
+//! them and Rust primitives. The structs in this module are *only* used as
+//! markers to represent a SQL type, and shouldn't be used in your structs. See
+//! the documentation for each type to see the Rust types that can be used with
+//! a corresponding SQL type. Additional types can be added by other crates.
```
killercup: This might be a good place to mention one other crate that does this, as people reading this might want to implement types themselves and having example code to look at might be nice.

killercup:
More documentation is always good :)

I'm not sure how wise it is to mention database specifics, though (like representing `bool` as `int` on SQLite). Assuming diesel gets adapters for other DBMSs we could end up needing to mention `n` specific cases for each SQL type. Let's keep it for now, though.

Might be a good idea to mention the correct usage of these types (i.e., don't use the in structs, use Rust's types) in the guide as well.

PR #409, opened 2016-08-18
Add support for timestamp with time zone

### LINKED ISSUE #106: Support `Timestamptz`
This should be separate from `Timestamp`. To start I would like to be incredibly conservative about any comparisons with these types, and not treat `timestamp` and `timestamptz` as interchangeable. I'm not super well versed in the semantics of this type -- it does sound like it mostly ends up being UTC on the backend, so we might be able to loosen this later.

### LINKED ISSUE #295: Handle `DateTime` in addition to `NaiveDateTime`
People trying to use `DateTime<UTC>` is really common. Perhaps we should write an `impl<TZ> ToSql<Timestamp, Pg> for DateTime<TZ>` and `impl FromSql<Timestamp, Pg> for DateTime<UTC>`, where the `ToSql` impl converts to UTC. I'm concerned this implies more guarantees than exist though, as the database does not contain TZ information on a normal `timestamp` type column.

### LINKED ISSUE #402: infer_schema!() with postgres Timestamptz fails
Using `infer_schema!()` macro on a postgres database with Timestamptz fails with the following error:

```
rc/lib.rs:1:1: 1:1 error: type name `Timestamptz` is undefined or not in scope [E0412]
src/lib.rs:1 #![feature(custom_derive, custom_attribute, plugin)]
             ^
<diesel macros>:5:1: 5:71 note: in this expansion of table_body! (defined in <diesel macros>)
src/schema.rs:1:1: 1:40 note: in this expansion of table! (defined in <diesel macros>)
src/schema.rs:1:1: 1:40 note: in this expansion of infer_schema! (defined in src/lib.rs)
src/lib.rs:8:1: 8:23 note: in this expansion of include!
src/lib.rs:1:1: 1:1 help: run `rustc --explain E0412` to see a detailed explanation
src/lib.rs:1:1: 1:1 help: no candidates by the name of `Timestamptz` found in your project; maybe you misspelled the name or forgot to import an external crate?
src/lib.rs:1:1: 1:1 error: type name `Timestamptz` is undefined or not in scope [E0412]
src/lib.rs:1 #![feature(custom_derive, custom_attribute, plugin)]
             ^
<diesel macros>:45:5: 45:51 note: in this expansion of column! (defined in <diesel macros>)
<diesel macros>:5:1: 5:71 note: in this expansion of table_body! (defined in <diesel macros>)
src/schema.rs:1:1: 1:40 note: in this expansion of table! (defined in <diesel macros>)
src/schema.rs:1:1: 1:40 note: in this expansion of infer_schema! (defined in src/lib.rs)
src/lib.rs:8:1: 8:23 note: in this expansion of include!
src/lib.rs:1:1: 1:1 help: run `rustc --explain E0412` to see a detailed explanation
src/lib.rs:1:1: 1:1 help: no candidates by the name of `Timestamptz` found in your project; maybe you misspelled the name or forgot to import an external crate?
error: aborting due to 2 previous errors
error: Could not compile `diesel_demo`.
```

An example of such can be found at https://github.com/strawlab/diesel_demo/tree/timestamptz . The specific issue is uncovered by adding this to the database

```
CREATE TABLE mytable (
  id SERIAL PRIMARY KEY,
  when_utc TIMESTAMPTZ
);
```

sgrif:
This adds support for the `timestamp with time zone` type. This is part
of the SQL standard, but not supported explicitly by SQLite as far as I
can tell. As such I've made it a PG specific type for now.

timestamp with time zone does not actually mean that a time zone is
stored. It instead means that when dealing with strings, Postgres will
no longer ignore the time zone portion. It will convert the time zone to
UTC for storage. When transmitted as text, it will be converted to the
database's local time zone. When transmitted as binary, it will be sent
as UTC.

As such, I've provided `ToSql` implementations for basically all flavors
of `DateTime`, but I've only provided a `FromSql` implementation for
`DateTime<UTC>` and `NaiveDateTime`. I have not provided any
implementation for `std::time::SystemTime`, as it implies local time
zone for the machine and we do not have the tools to handle the
conversion in the standard library.

Fixes #106.
Fixes #295.
Fixes #402.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   See http://docs.diesel.rs/diesel/prelude/trait.SortExpressionMethods.html
   for details.
 
+* Added support for the `timestamp with time zone` type in PostgreSQL (referred
+  to as `diesel::types::Timestamptz`)
+
 ## [0.7.1] - 2016-08-11
 
 ### Changed
--- a/diesel/src/pg/expression/date_and_time.rs
+++ b/diesel/src/pg/expression/date_and_time.rs
@@ -3,7 +3,13 @@ use expression::{Expression, SelectableExpression, NonAggregate};
 use pg::{Pg, PgQueryBuilder};
 use query_builder::*;
 use result::QueryResult;
-use types::{Timestamp, VarChar};
+use types::{Timestamp, Timestamptz, Date, VarChar};
+
+/// Marker trait for types which are valid in `AT TIME ZONE` expressions
+pub trait DateTimeLike {}
+impl DateTimeLike for Date {}
+impl DateTimeLike for Timestamp {}
+impl DateTimeLike for Timestamptz {}
 
 #[derive(Debug, Copy, Clone)]
 pub struct AtTimeZone<Ts, Tz> {
@@ -21,10 +27,10 @@ impl<Ts, Tz> AtTimeZone<Ts, Tz> {
 }
 
 impl<Ts, Tz> Expression for AtTimeZone<Ts, Tz> where
-    Ts: Expression<SqlType=Timestamp>,
+    Ts: Expression,
+    Ts::SqlType: DateTimeLike,
     Tz: Expression<SqlType=VarChar>,
 {
-    // FIXME: This should be Timestamptz when we support that type
     type SqlType = Timestamp;
 }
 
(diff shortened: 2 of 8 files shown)


sgrif:

killercup:
LGTM, though I'd probably add a few tests for edge cases like inserting a datetime in one time zone and trying to get it out in another. Is this a feature that is regularly used? Do you want to add some docs for that (even if you copying the ones from `timestamp`)?

sgrif:
You mean documentation for the SQL type itself? Yeah I'll update that PR adding those docs after I merge this.

PR #415, opened 2016-08-20
Upgrade quickcheck and enable unstable roundtrip test

### LINKED ISSUE #410: Upgrade Quickcheck
Currently, there is this FIXME in [`diesel_tests/tests/types_roundtrip.rs`](https://github.com/diesel-rs/diesel/blob/bfbe86bd60204b1904804fbe6d8e0e26befebb7f/diesel_tests/tests/types_roundtrip.rs#L125-L144):

> FIXME: Needs https://github.com/BurntSushi/quickcheck/pull/128

That PR has been merged and is part of quickcheck 0.2.27. In fact, version 0.3 of quickcheck has been released in the meantime, though I don't know what the changes are.

shssoichiro:
Closes #410

### DIFF
--- a/diesel_tests/tests/types_roundtrip.rs
+++ b/diesel_tests/tests/types_roundtrip.rs
@@ -127,23 +127,17 @@ pub fn mk_naive_date(days: u32) -> NaiveDate {
     earliest_pg_date + Duration::days(days as i64 % num_days_representable)
 }
 
-// FIXME: Needs https://github.com/BurntSushi/quickcheck/pull/128
-// #[cfg(all(feature = "unstable", feature = "postgres"))]
-// mod unstable_types {
-//     use super::*;
-//     use std::time::*;
+#[cfg(feature = "postgres")]
+mod unstable_types {
+    use super::*;
+    use std::time::*;
 
-//     fn strip_nanosecond_precision(time: SystemTime) -> SystemTime {
-//         let res = match time.duration_from_earlier(UNIX_EPOCH) {
-//             Ok(duration) => time - Duration::new(0, duration.subsec_nanos() % 1000),
-//             Err(e) => time + Duration::new(0, e.duration().subsec_nanos() % 1000),
-//         };
-//         work_around_rust_lang_30173(res)
-//     }
+    fn strip_nanosecond_precision(time: SystemTime) -> SystemTime {
+        match time.duration_since(UNIX_EPOCH) {
+            Ok(duration) => time - Duration::new(0, duration.subsec_nanos() % 1000),
+            Err(e) => time + Duration::new(0, e.duration().subsec_nanos() % 1000),
+        }
+    }
 
-//     fn work_around_rust_lang_30173(time: SystemTime) -> SystemTime {
-//         time + Duration::new(0, 1) - Duration::new(0, 1)
-//     }
-
-//     test_round_trip!(systemtime_roundtrips, Timestamp, SystemTime, strip_nanosecond_precision);
-// }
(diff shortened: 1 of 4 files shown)


diesel_tests/tests/types_roundtrip.rs
```diff
@@ -127,23 +127,22 @@ pub fn mk_naive_date(days: u32) -> NaiveDate {
...
+
+    fn strip_nanosecond_precision(time: SystemTime) -> SystemTime {
+        let res = match time.duration_since(UNIX_EPOCH) {
+            Ok(duration) => time - Duration::new(0, duration.subsec_nanos() % 1000),
+            Err(e) => time + Duration::new(0, e.duration().subsec_nanos() % 1000),
+        };
+        work_around_rust_lang_30173(res)
+    }
+
+    fn work_around_rust_lang_30173(time: SystemTime) -> SystemTime {
```
killercup: This should no longer be necessary: https://github.com/rust-lang/rust/pull/30173

killercup:
Thanks, LGTM!

I think we can get rid of one more fixme (see inline comment) :)

diesel_tests/tests/types_roundtrip.rs
```diff
@@ -127,23 +127,22 @@ pub fn mk_naive_date(days: u32) -> NaiveDate {
...
-//         work_around_rust_lang_30173(res)
-//     }
-
-//     fn work_around_rust_lang_30173(time: SystemTime) -> SystemTime {
-//         time + Duration::new(0, 1) - Duration::new(0, 1)
-//     }
-
-//     test_round_trip!(systemtime_roundtrips, Timestamp, SystemTime, strip_nanosecond_precision);
-// }
+#[cfg(all(feature = "unstable", feature = "postgres"))]
```
sgrif: I think we can just remove `feature = "unstable"`. Quickcheck still hides this behind a flag because they support older versions of Rust, but we don't.

shssoichiro: Would you suggest I move the `"quickcheck/unstable"` feature in `diesel_tests/Cargo.toml` from our `unstable` feature to the `postgres` feature? The tests fail to compile without that flag, since quickcheck only has impls for `SystemTime` behind their `unstable` flag.

diesel_tests/tests/types_roundtrip.rs
shssoichiro: Fixed!

diesel_tests/tests/types_roundtrip.rs
sgrif: If they only have `SystemTime` behind the unstable flag I would just always have it on.

shssoichiro: Looks like it passes on stable with `quickcheck/unstable` turned on. Should be good to go now.

shssoichiro: Sorry, looks like I was wrong. If I take out this test from behind our "unstable" feature flag, the test suite fails to compile on stable.

```the trait bound `std::time::SystemTime: diesel::Queryable<diesel::types::Timestamp, diesel::pg::Pg>` is not satisfied``` and several simliar errors.

shssoichiro: Looks like this is because `diesel::std_time` is behind the `unstable` flag. If we remove it, the tests pass on the latest stable. Should I go ahead and remove that flag from `std_time`?

sgrif: Yes

shssoichiro: Pushed. Should be good to go.

Issue #42, opened 2015-12-02
Support composite primary keys


sgrif:
I believe this should be straightforward. There's a couple of things to consider

Edit: The thoughts in this comment are super outdated, and likely no longer relevant.
- Right now `Table::PrimaryKey` has the constraint `Column<Table=Self>`. Unsure if we want to relax that to `SelectableExpression`, or if we want to introduce the `Columns` trait. We could even implement `Column` for tuples now if we really want to, as there's no longer any automatically added behaviors.
- With the way our type system is set up, `conn.find(&posts_tags, &(1, 1))` would be more likely to generate `SELECT * FROM posts_tags WHERE (post_id, tag_id) = $1`, instead of `SELECT * FROM posts_tags WHERE post_id = $1 AND tag_id = $2`. Might be fine for an initial implementation, but we do eventually need to solve this to support other databases.

sgrif:
This will likely be much easier once we implement the `Identifiable` trait that we're looking to add to support associations as part of 0.6

PR #417, opened 2016-08-25
Add initial support for composite primary keys

sgrif:
This pull request allows the `table!` macro and codegen to handle tables
which use more than one column for their primary key. At the moment they
are fairly useless, but they will compile.

The methods in question are primarily used for `FindDsl` and
associations. Things like update are defined in terms of `FindDsl` and
`Identifiable`. Since the generators for those traits don't handle
composite primary keys, none of that can be used (yet).

Working on this also made me realize a bit of a hole in our type system.
We implement `Expression` for a tuple of expressions as a comma
delimited list. However, a comma delimited list isn't always valid. In
particular a composite table *does* actually implement `FindDsl` for an
expression that matches it's primary key -- and would generate invalid
SQL like `pk1, pk2 = $1, $2`. Luckily, we're saved by the fact that
tuples don't implement `AsExpression`, so trying to do `.find((1, 2))`
won't compile.

I'm not sure if we should fix this or not, but the fix would basically
mean separating out "arbitrary expression" from "expression valid as a
select or returning clause", and potentially other places where tuples
appear such as set clauses and in expressions. For now though, as long
as the inproper code doesn't compile for the reasonable cases, I'm not
super worried. We should fix this eventually though, as I don't like
something that we can statically determine is invalid SQL compiling.

Ref #42

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Added
 
+* Added support for composite primary keys.
+
 * Added support for PostgreSQL `NULLS FIRST` and `NULLS LAST` when sorting.
   See http://docs.diesel.rs/diesel/prelude/trait.SortExpressionMethods.html
   for details.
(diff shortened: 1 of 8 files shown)


sgrif:

killercup:
LGTM. Do you maybe want to add some docs on how to use this? (Mentioning that the `table!` macro allows a comma separated list of columns for composite primary keys should suffice.)

sgrif:
(I added docs before merging)

PR #418, opened 2016-08-25
Respect overridden migration directories in `migration run`

sgrif:
`revert` and `redo` are still not respecting them, but I will fix those
as a separate PR since they require code changes on the Diesel side.


sgrif:

/cc @mcasper

killercup:
LGTM.

Is there a way to test this? (I don't think it's really necessary.)

mcasper:
Looks good to me too!

Should be pretty simple to add the test to `migration_run.rs`, happy to do so if you want @sgrif

PR #422, opened 2016-08-31
add time type

kardeiz:
Adds support for the `time::Timespec` `Timestamp` type. Modeled after the implementation for `std_time`.

Please note: while the `time` crate lives under the `rust-lang-deprecated` org, it is not really "deprecated" per se (as the authors point out) and is still by far the most popular time handling crate according to crates.io.


killercup:
Hey, thanks for the pull request!

I'm totally inexperience in how to deal with date/time types in Rust, but I know that diesel already supports `chrono` (the impl for Postgres lives [here](https://github.com/diesel-rs/diesel/blob/0c7693cd67f3b2852bdc30c53cbc04fa1886b097/diesel/src/pg/types/date_and_time/chrono.rs)). Sorry if this is a dumb question, but: Why not use this? IIRC, chrono internally uses the time crate as well and seems to support more use cases while having a more powerful API.

kardeiz:
I prefer the `time` API and don't need the additional features of `chrono`. I also depend on other crates that depend on `time` and I don't really want to carry around more than one time library.

sgrif:
Can we call the feature `deprecated_time` instead of `time`?

killercup:
`deprecated_time` sound so… degrading. I propose calling it `good_old_time`! 😉

kardeiz:
@sgrif Is that really necessary? I understand wanting to not give the impression that `time` is the authoritative datetime type, but I think that's the nature of crate names, and I think calling it `deprecated_time` is more misleading (and degrading!) than leaving it as is.

sgrif:
I'm certainly open to alternative suggestions. But you hit the nail on the head, `time` as a feature name sounds too authoritative for something in the `rust-lang-deprecated` org.

sgrif:
Sorry for the delay. Can you add a changelog entry, squash and rebase? Thank you for working on this!

PR #426, opened 2016-09-04
Update codegen to latest nightly.

SergioBenitez:
This PR brings Diesel's codegen up to date with the latest nightly.

I've noticed that the nightly policy has been updated to track the Rust release schedule. I believe this to be a mistake that will result in far fewer people using Diesel. This is because most major codegen libraries (Serde, in particular) track nightly directly, and so users will be forced to choose between Diesel and other libraries. Given that some of those libraries are critically important, I imagine this will result in users switching away from Diesel.

A more amenable policy might be to track a once-a-week nightly. I'd be happy to lend a hand in this capacity.

P.S: The nightly tests will fail since they're on the 8/18 release.


killercup:
I can see you pain, @SergioBenitez, but also understand that @sgrif doesn't want to spend his time tracking nightly (with a delay of X days anyway).

Is there a compromise we could make? E.g., bump codegen version more often, but use semver's beta flag/build numbers to specify the nightly targeted? Something like `diesel_codegen 0.7.3+20160904`?

SergioBenitez:
I understand the desire to not want to track nightly with a rigid schedule. Not doing so is okay with me. I like where your idea of a compromise is heading. My main issue with the policy is that it seems that it specifically does not allow changes to support newer nightlies than those released on the Rust schedule.

What if the policy is to simply work with _some_ recent nightly version without a schedule? That is, the README can simply say: "last tested nightly release: 9/05/2016". This way, contributors can keep Diesel up to date between Rust releases without @sgrif worrying about it. I don't have much of a preference as to how version numbers fit into this idea.

ElijahCaine:
Forgive my ignorance, I am new to the world of Rust and have much to learn, but is there a reason Diesel doesn't just track Rust Beta?

It seems like Diesel verifies it works on Rust's six weeks release cycle (which is how often Beta is updated), but instead of running `rustup default beta` a developer has to `rustup default nightly && rustup override set nightly-XXXX-YY-ZZ`.  Just seems like unnecessary hassle and an easy way for newbies to make mistakes.

Since Diesel has become the defacto & most stable Database ORM it might be worth considering tracking something more stable than an *outdated nightly*.

sgrif:
@ElijahCaine We do work on beta as well if you use syntex. Beta allows the same things that stable does. If you want to use `#![feature]` attributes (opting into an unstable feature), you have to use nightly. We switched to the six week tracking cycle because keeping up with every breaking change in nightly was a huge maintenance burden and was taking up enough time that it was preventing feature development.

Stable Rust is always an option on syntex, but we point to nightly because it's a bit simpler.

A lot of the discussion around this is moot now that the procedural macros 1.1 RFC has been accepted. That RFC was specifically designed to be something easy to stabilize, and will likely end up being the only thing we target once it becomes stable (which we're hoping is by the end of the year)

SergioBenitez:
@sgrif What is your decision on this PR? I'm proposing that you accept changes to bring Diesel up to date with the latest nightly, not that _you_ keep Diesel up to date with the latest nightly. This means that those of us that want to use Diesel on the latest nightly can do so without burdening you with the maintenance required to keep it up to date.

diesel_codegen/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_codegen"
 keywords = ["orm", "database", "postgres", "sql", "codegen"]
 
 [dependencies]
-diesel_codegen_syntex = { version = "0.7.0", default-features = false }
+diesel_codegen_syntex = { path = "../diesel_codegen_syntex", version = "0.7.3", default-features = false }
```
sgrif: This shouldn't be required, we have a local override.

diesel_codegen_syntex/Cargo.toml
```diff
@@ -11,12 +11,12 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_codegen"
 keywords = ["orm", "database", "postgres", "sql", "codegen"]
 
 [build-dependencies]
-syntex = { version = ">= 0.37.0, < 0.43.0", optional = true }
-syntex_syntax = { version = ">= 0.37.0, < 0.43.0", optional = true }
+syntex = { version = ">= 0.43.0", optional = true }
```
sgrif: This needs to be bounded. Just `0.43.0` is fine.

diesel_tests/Cargo.toml
```diff
@@ -6,11 +6,11 @@ license = "MIT OR Apache-2.0"
 build = "build.rs"
 
 [build-dependencies]
-syntex = { version = "0.42.0", optional = true }
+syntex = { version = ">= 0.43.0", optional = true }
```
sgrif: This needs to be bounded.

diesel_tests/Cargo.toml
```diff
@@ -6,11 +6,11 @@ license = "MIT OR Apache-2.0"
...
 diesel_codegen_syntex = { path = "../diesel_codegen_syntex", optional = true }
-dotenv_codegen = { version = "0.9.0",  optional = true }
+dotenv_codegen = { git = "https://github.com/slapresta/rust-dotenv", version = "0.9.3",  optional = true }
```
sgrif: Has this version been released? Do we need to point at git?

SergioBenitez: It has [not been released](https://crates.io/search?q=dotenv). As such, we'll need to point to Git for the time being.

diesel_codegen/Cargo.toml
SergioBenitez: Even with the local override, this is required because version 0.7.3 hasn't been published to crates.io. From the [Cargo documentation](http://doc.crates.io/specifying-dependencies.html#overriding-with-local-dependencies):

> Note: using a local configuration to override paths will only work for crates that have been published to crates.io. You cannot use this feature to tell Cargo how to find local unpublished crates.

sgrif: Where are you getting 0.7.3 from? The version numbers haven't changed.

sgrif:
> What is your decision on this PR?

I appreciate the work on this, and will merge it when the time is appropriate. I'm in the process of migrating our nightly story to macros 1.1. I'm not planning on releasing sooner than we would have otherwise. Since macros 1.1 have landed, the entire discussion around our nightly version support is now moot, and I'll be moving nightly off of allowed failures in the next release.

diesel_codegen/Cargo.toml
SergioBenitez: I changed the version numbers to 0.7.3 given that this is a breaking change.

killercup: If this is a breaking change, the next version should be 0.8.0.

SergioBenitez: It's not a breaking change in the semver sense, I believe. I bumped the version since 0.7.3 will work with the latest nightly while 0.7.2 will not.

sgrif: Dropping support for Syntex versions is definitely a breaking change. This would need to be in 0.8.0. However, pull requests shouldn't bump the version we'll do it when it's time to release.

PR #429, opened 2016-09-06
Simplify the support of custom types

weiznich:


emk:
This (or something like it) would be lovely! I would love to add Diesel support for a couple of types in https://github.com/faradayio/bigml-rs, but right now, it's prohibitively difficult to implement the full set of traits.

sgrif:
So ultimately there are two cases where you need to implement traits around this. The first is when you have a custom type in Rust that is represented by an existing database type, and the second is when you are adding support for a new database type that isn't already supported. This PR doesn't address the second case (which ultimately will need to care about the exact bytes sent over the wire, and thus needs more traits).

For the first case, ultimately the only two traits that you need to implement are `AsExpression<OtherDbType>` and `FromSqlRow<OtherDbType, Backend>`. So comparing the code with/without this PR:

### Without PR
```rust
impl FromSqlRow<SmallInt, Pg> for Color {
    fn build_from_row<R: Row<Pg>>(row: &mut R) -> Result<Self, Box<Error+Send+Sync>> {
        match i16::build_from_row(row)? {
            1 => Ok(Color::Red),
            2 => Ok(Color::Green),
            3 => Ok(Color::Blue),
            v => panic!("Unknown value {} for Color found", v),
        }
    }
}

impl<'a> AsExpression<Nullable<SmallInt>> for &'a Color {
    type Expression = AsExprOf<i16, Nullable<SmallInt>>;

    fn as_expression(self) -> Self::Expression {
        AsExpression::<Nullable<SmallInt>>::as_expression(*self as i16)
    }
}
```

### With PR

```rust
impl CustomSqlType for Color {
    type DataBaseType = SmallInt;
    type RawType = i16;

    fn to_database_type(&self) -> i16 {
        *self as i16
    }

    fn from_database_type(v: &i16) -> Result<Self, Box<Error + Send + Sync>> {
        match *v {
            1 => Ok(Color::Red),
            2 => Ok(Color::Green),
            3 => Ok(Color::Blue),
            v => panic!("Unknown value {} for Color found", v),
        }
    }
}

// Add all needed implements for diesel
CustomSqlType!(Color);
```

I think we can definitely improve the ergonomics of the before code. `FromSqlRow` could use a better name. The signature of `build_from_row` could be simplified with some type aliases. The `AsExpression` impl could be `impl AsExpression<SmallInt> for Color`. I'm open to all of those, but I don't think we need to introduce a new trait that effectively mirrors the existing one. Especially when it involves macros. For code likely to be introduced by users (not libraries), I'd prefer to keep as much of it out of codegen and in Rust as we can, so that understanding Rust is enough to understand what it's doing. Most importantly, these are the error messages I get if the impls are missing:

### `#[derive(Queryable)]` with missing `FromSqlRow`

```
error[E0277]: the trait bound `Color: diesel::types::FromSqlRow<diesel::types::SmallInt, _>` is not satisfied
  --> src/main.rs:64:11
   |
64 |     users.load::<User>(&connection).unwrap();
   |           ^^^^ the trait `diesel::types::FromSqlRow<diesel::types::SmallInt, _>` is not implemented for `Color`
   |
```

### `#[derive(Insertable)]` without `AsExpression`:

```
error[E0277]: the trait bound `Color: diesel::Expression` is not satisfied
  --> src/main.rs:54:1
   |
54 | #[derive(Insertable)]
   | ^^^^^^^^^^^^^^^^^^^^^ the trait `diesel::Expression` is not implemented for `Color`
   |
   = note: required because of the requirements on the impl of `diesel::Expression` for `&'insert Color`
   = note: required because of the requirements on the impl of `diesel::expression::AsExpression<diesel::types::Nullable<diesel::types::SmallInt>>` for `&'insert Color`
```

The second one could definitely be improved, and I wish it mentioned the actual bound it's looking for more prominently, which I've been pushing on the Rust team to do. But both error messages provide the exact trait impl that needs to be written.

As such, I think this is ultimately a documentation problem. We should better document how to do this in a prominent location, and we should improve the ergonomics where we can. But I don't think this PR improves the situation at the end of the day.

weiznich:
@sgrif unfortunately it seems to be not that simple.

Using the following code: 
```rust
#[derive(Clone, Copy)]
#[repr(i16)]
enum Color {
    Red = 1,
    Green = 2,
    Blue = 3,
}

// Use the type like every other type provided by diesel
table!{
     users{
         id -> Integer,
         name -> Text,
         hair_color -> Nullable<SmallInt>,
     }
 }

#[derive(Queryable, Debug, Clone)]
struct User {
    name: String,
    hair_color: Option<Color>,
}

#[derive(Debug, Clone, Insertable)]
#[table_name = "users"]
struct NewUser<'a> {
    name: &'a str,
    hair_color: Option<Color>,
}

use diesel::types::*;
use diesel::expression::AsExpression;
use diesel::expression::helper_types::AsExprOf;
use diesel::pg::Pg;
use diesel::row::Row;
use std::error::Error;

impl FromSqlRow<SmallInt, Pg> for Color {
    fn build_from_row<R: Row<Pg>>(row: &mut R) -> Result<Self, Box<Error+Send+Sync>> {
        match i16::build_from_row(row)? {
            1 => Ok(Color::Red),
            2 => Ok(Color::Green),
            3 => Ok(Color::Blue),
            v => panic!("Unknown value {} for Color found", v),
        }
    }
}

impl<'a> AsExpression<Nullable<SmallInt>> for &'a Color {
    type Expression = AsExprOf<i16, Nullable<SmallInt>>;

    fn as_expression(self) -> Self::Expression {
        AsExpression::<Nullable<SmallInt>>::as_expression(*self as i16)
    }
}

fn test() {
    use self::users::dsl::*;
    use diesel::*;

    let p = PgConnection::establish("").unwrap();

    let u = NewUser{
        name: "",
        hair_color: None,
    };

    let u: User = insert(&u).into(users).get_result(p).unwrap();
}
```

produces the following error about missing implementations:
```
error: no method named `get_result` found for type `diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>>` in the current scope
  --> src/lib.rs:76:42
   |
76 |     let u: User = insert(&u).into(users).get_result(p).unwrap();
   |                                          ^^^^^^^^^^
   |
   = note: the method `get_result` exists but the following trait bounds were not satisfied: `diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>, diesel::query_builder::insert_statement::Insert, diesel::query_builder::returning_clause::ReturningClause<(users::columns::id, users::columns::name, users::columns::hair_color)>> : diesel::query_builder::QueryFragment<_>`, `&diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::AsQuery`, `_ : diesel::query_builder::QueryFragment<_>`, `&diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `&diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `&mut diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::AsQuery`, `_ : diesel::query_builder::QueryFragment<_>`, `_ : diesel::query_builder::QueryId`, `&mut diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `&mut diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query`, `&mut diesel::query_builder::insert_statement::InsertStatement<users::table, &NewUser<'_>> : diesel::query_builder::Query
```

In an ideal world one would only have to implement `ToSql` and `FromSql`  for any custom type.

emk:
> In an ideal world one would only have to implement ToSql and FromSql for any custom type.

Or if that fails because of trait coherence rules, then implementing `ToSql` and `FromSql`, and then calling one or more macros might also work.

sgrif:
There's a blanket impl for `Option` that violates coherence without specialization. It is present with `features = "unstable"` though.

weiznich:
@sgrif Finally I've got a working implementation against diesel 0.10.

The following traits needs to be implemented: (Implementation omitted)

```rust
struct MyCustomType;

impl<DB> ToSql<DatabaseType, DB> for MyCustomType;
impl<DB> ToSql<Nullable<DatabaseType>, DB> for MyCustomType;

impl<DB> FromSql<DatabaseType, DB> for MyCustomType;
impl<DB> FromSqlRow<DatabaseType, DB> for MyCustomType;

impl AsExpression<DatabaseType> for MyCustomType;
impl<'a> AsExpression<DatabaseType> for &'a MyCustomType;
impl<'a> AsExpression<Nullable<DatabaseType>> for &'a MyCustomType;
```

Compared to this 6 implementation the implementation of the CustomSqlType trait is rather short and easy. (I'm not saying we should go that specific way, but this needs to be improved in some way)

sgrif:
You can implement `AsExpression` in a way that doesn't require a `ToSql` impl, and you can implement `FromSqlRow` in a way that doesn't require a `FromSql` impl, as in my example above.

weiznich:
@sgrif As shown above this won't work if you start to use the type with `LoadDsl`. (Maybe I'm missing something?)

sgrif:
Works fine for me. https://gist.github.com/sgrif/299a94c7fa818801d2604ba97972c6a9

TruputiGalvotas:
@sgrif I've tried writing AsExpression implementation for my custom Enum type which should map to SmallInt using the sample of code you've provided here in the comments

Using this sample:
```
impl<'a> AsExpression<Nullable<SmallInt>> for &'a Color {
    type Expression = AsExprOf<i16, Nullable<SmallInt>>;

    fn as_expression(self) -> Self::Expression {
        AsExpression::<Nullable<SmallInt>>::as_expression(*self as i16)
    }
}
```

minus Nullable (since my enum is never a null) I get this error from compiler:

> cannot move out of borrowed content

it happens at `*self` in the provided code sample.. Any ideas?

SOF3:
Is the documentation for the first case (support of newtypes wrapping already-supported types) written anywhere?

PR #431, opened 2016-09-07
Fix #430

### LINKED ISSUE #430: Identifiable!, BelongsTo!, HasMany! are brocken when used with custom_derive
Using one of those 3 macros with custom_derive will result in the following error:

```
error: expected ident, found ,
  --> <diesel macros>:21:25
   |
21 | Identifiable ! { $ args , fields = [ $ ( $ fields ) * ] , } } ; (
   |                         ^
```

Example Code:
```rust
custom_derive! {
     #[derive(Identifiable)]
     #[table_name(users)]
     struct User {
         id: i32,
         name: String,
     }
}
```

This is most likely caused by custom_derive not passing custom attributes down to the macro. The documentation mentions clearly:

"Secondly, the macro is passed the entire item, ***sans attributes***. It is the derivation macro's job to parse the item correctly."

We should change the syntax here to something like:

```rust
custom_derive! {
     #[derive(Identifiable(table_name=users))]
     struct User {
         id: i32,
         name: String,
     }
}
```

weiznich:


diesel/src/macros/associations/belongs_to.rs
```diff
@@ -45,15 +45,25 @@
...
+/// # custom_derive!{
+/// #     #[derive(Identifiable(users))]
+/// #     pub struct User {
+/// #         id: i32,
+/// #     }
+/// # }
 /// custom_derive! {
-///     #[derive(BelongsTo(User, foreign_key = user_id)]
-///     #[table_name(posts)]
+///     #[derive(BelongsTo(User, foreign_key = user_id, table = posts), Identifiable(posts))]
```
sgrif: Isn't the table name specified as `#[table_name(posts)]` on the struct?

diesel/src/macros/identifiable.rs
```diff
@@ -31,15 +31,18 @@
...
 /// [custom_derive]: https://crates.io/crates/custom_derive
 ///
-/// ```ignore
+/// ```no_run
+/// # #[macro_use] extern crate diesel;
+/// # #[macro_use] extern crate custom_derive;
+/// # table! { users { id -> Integer, } }
 /// custom_derive! {
-///     #[derive(Identifiable)]
-///     #[table_name(users)]
```
sgrif: This is the intended API, not something I want to change.

diesel/src/macros/associations/belongs_to.rs
weiznich: In theory yes.
In practice custom_derive removes every attribute from the item, so `#[table_name(posts)]` is not passed to the underlying macro.

Secondly, the macro is passed the entire item, **sans attributes**. It is the derivation macro's job to parse the item correctly.

[Documentation](https://danielkeep.github.io/rust-custom-derive/doc/custom_derive/index.html) of custom_derive, below the example code.

sgrif: Wait what? That was not true before.

weiznich: The file defining  the custom_derive [macro](https://github.com/DanielKeep/rust-custom-derive/blob/master/src/lib.rs) didn't change in the last 5 month according to the github history.
Also testing the code provided as documentation results in a compiler error. So this definitively needs to change.
Do you have any better ideas?

sgrif: No. I'm inclined to just drop `custom-derive` from the documentation in this case. I'm not happy with this change to the API. This push was also before the macros 1.1 RFC which changes things a lot now that it's landed. Core team is discussing what we want to do.

weiznich: This commit doesn't change anything for other methods to invoke this macro, it only fixes the usage with custom_derive. 
It should also be possible to provide this outside of diesel, but I think it belongs into the main crate, because all other macros suggests the usage of custom_derive.

Obviously things will change as soon as macros 1.1 will become usable on stable, but up to this date custom_derive seems like the best method to provide support for stable.

sgrif: The goal is to provide a single consistent API and for the macro to be useable with or without custom_derive

sgrif:
Since Macros 1.1 has landed, and is on a fast path to stabilization, I think the right call here is to just remove all mentions of custom-derive from the docs instead.

PR #432, opened 2016-09-09
Upgrade uuid to version 0.3

shssoichiro:
Unblocks users of diesel's `uuid` feature from using version 0.3 of the `uuid` crate, which also unblocks `serde` 0.8 for `uuid` users.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -16,7 +16,7 @@ libsqlite3-sys = { version = "^0.4.0", optional = true }
 byteorder = "0.3.*"
 quickcheck = { version = "0.3.1", optional = true }
 chrono = { version = "^0.2.17", optional = true }
-uuid = { version = "^0.2.0", optional = true, features = ["use_std"] }
+uuid = { version = ">=0.2.0, <0.4.0", optional = true, features = ["use_std"] }
 
 [dev-dependencies]
 quickcheck = "0.3.1"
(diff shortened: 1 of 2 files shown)


killercup:
Thanks. If the UUID crate's API as used by diesel did not change it, would it make sense to allow `^0.2, ^0.3`?

shssoichiro:
This is what I had thought as well, that I would need to make code changes to satisfy the tests passing with uuid 0.3. However, with diesel 0.7.2 which requires uuid ^0.2, if I have in my app's Cargo.toml that I want uuid ^0.3, the app will fail to build, saying that `diesel::expression::Expression` is not satisfied for `uuid::Uuid`. If I change my application's Cargo.toml to require uuid ^0.2, the build succeeds. So for whatever reason, if there's a version mismatch in the uuid crate between diesel and my Cargo.toml, the diesel traits don't get implement for my app's version of `uuid::Uuid`. 

Note that this is with the 2016-08-18 rustc nightly that is recommended for diesel.

killercup:
@shssoichiro Since 0.2 and 0.3 are not SemVer compatible according to cargo, it is my understanding that cargo will build both versions—0.3 because you want it, and 0.2 because diesel wants it. And diesel's trait will only get implemented for the 0.2 Uuid type.

If you change diesel to allow more version of uuid, say `>=0.2.0, <0.4.0`, cargo can give diesel the same version of uuid (0.3) that you get in your own code with `extern crate uuid`.

shssoichiro:
Awesome. I've updated my PR to allow `version = ">=0.2.0, <0.4.0"` for diesel's uuid.

killercup:
Sweet. Thank you!

As uuid's bump from 0.2 to 0.3 was just to allow newer versions of serde\*, this should not give us any trouble\*\* :)

\* if you are using serde: please note that newer versions of serde-codegen are depending too-recent versions of syntex/nightlies
\** famous last words

PR #404, opened 2016-08-15
Add the getting started guide to examples

sgrif:
This code will differ slightly from the existing code in `diesel_demo` and on the website, as I rewrote it by following the guide instead of just copying the code over. This is intended to replace the demo repo, as I want this to be part of our test suite, and it gives me more stable URLs to link to from the guide.

I'm planning on changing the guide on the website to pull code snippets from this file automatically.

PR #433, opened 2016-09-09
Add the getting started guide to examples (take 2)

killercup:
@sgrif in #404:

> This code will differ slightly from the existing code in `diesel_demo` and on the website, as I rewrote it by following the guide instead of just copying the code over. This is intended to replace the demo repo, as I want this to be part of our test suite, and it gives me more stable URLs to link to from the guide.
> 
> I'm planning on changing the guide on the website to pull code snippets from this file automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,13 @@ script:
   else
     (cd diesel && travis-cargo test -- --no-default-features --features "chrono $BACKEND")
   fi &&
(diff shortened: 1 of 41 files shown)


sgrif:
Looks like nightly got moved out of allowed failures is the only problem.

killercup:
And it's green! 💚

PR #434, opened 2016-09-09
Make infer_schema! skip views

killercup:
Rebased version of #374. Added some test migrations, will add actual tests later on.


killercup:
@sgrif I'm not sure how to elegantly test the Postgres version. For SQLite we can just open a new `:memory:` database, create tables and views, and see that it works (i.e. ignore the views). For Postgres this is more tricky. Would it make sense to write compile-tests for this? Or do we just trust that it works based on the fact that we call `infer_schema!` in `diesel_test` and the migrations now include views?

sgrif:
You should be able to just create the view in a transaction and it'll be fine.

killercup:
Added a simple test for postgres. Tell me if you want to have more.

migrations/postgresql/20160909174741_create_highlight_post_view/up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE stuff (
+  id SERIAL PRIMARY KEY,
+  name VARCHAR NOT NULL,
+  important BOOLEAN NOT NULL DEFAULT 'f'
+);
+
+CREATE VIEW important_stuff AS
```
sgrif: Do you think we need this since we have the test on `load_table_names`?

killercup: Good question. I think it'd be fine to leave it in but currently it doesn't give us anything, so I'll remove it. If we were to support views later, we can just re-add these migrations.

killercup:
Feel free to merge or squash-merge this as soon as travis likes it. I just pushed a revert for the migrations.

PR #435, opened 2016-09-11
Implement `#[changeset_for]` in terms of `AsChangeset!`

sgrif:
This required two changes which stand out as odd. The first is the
addition of `, ..` to the pattern. This is because the stable macro does
not perform filtering of the `id` column like the proc macro does. We
can just not pass that field, but that meant we were generating an
invalid pattern as it didn't match all columns. This works around that.
This does mean that a tuple struct with a field mapping to the primary
key will blow up here, but that's not a case I'm terribly worried about.

The second weird thing we had to do is the specific token representation
of `treat_none_as_null`. Unfortunately just using
`options.treat_none_as_null.to_string()` (when the field was a boolean)
doesn't work here. The pattern `"false"` doesn't match the resulting
tokens on the stable macro side. This is because interpolating that
results in a non terminal interpolated expression token, rather than a
string literal token, and because of obscure details of how proc macros
communicate back it doesn't match up.

### DIFF
--- a/diesel_codegen_syntex/src/update.rs
+++ b/diesel_codegen_syntex/src/update.rs
@@ -1,14 +1,13 @@
-use syntax::ast::{self, MetaItem, MetaItemKind, TyKind};
+use syntax::ast::{self, MetaItem, MetaItemKind};
 use syntax::attr::AttrMetaMethods;
 use syntax::codemap::Span;
 use syntax::ext::base::{Annotatable, ExtCtxt};
-use syntax::ext::build::AstBuilder;
 use syntax::ptr::P;
 use syntax::parse::token::{InternedString, str_to_ident};
+use syntax::tokenstream::TokenTree;
 
-use attr::Attr;
 use model::Model;
-use util::{ty_param_of_option, is_option_ty};
+use util::lifetime_list_tokens;
 
 pub fn expand_changeset_for(
     cx: &mut ExtCtxt,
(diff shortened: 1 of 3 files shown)


diesel_codegen_syntex/src/update.rs
```diff
@@ -53,13 +51,14 @@ fn table_name(cx: &mut ExtCtxt, meta_item: &MetaItem) -> Result<InternedString,
     }
 }
 
+#[allow(unused_imports)] // quote_tokens! generates warnings
 fn boolean_option(cx: &mut ExtCtxt, meta_items: &[P<MetaItem>], option_name: &str)
-    -> Result<Option<bool>, ()>
+    -> Result<Vec<TokenTree>, ()>
```
killercup: Why did you change this fn to return a token tree vec? IIUC this is given the same context as `changeset_impl` so you could just return a bool here and match on it to create the quoted version in `changeset_impl` below.

diesel_codegen_syntex/src/update.rs
```diff
@@ -84,76 +83,26 @@ fn changeset_impl(
...
-                <$column as ::diesel::expression::Expression>::SqlType,
-                &'update $field_ty,
-            >,
-        >
-    )
-}
+    let pk = model.primary_key_name();
+    let fields = model.attrs.iter()
+        .filter(|a| a.column_name.name != pk.name)
+        .map(|a| a.to_stable_macro_tokens(cx))
```
killercup: I almost missed that line! This is where some magic happens :)

killercup:
LGTM. Left a comment about creating "true"/"false" token tree later, but is not important.

Can you fix the spelling of 'applied' [here](https://github.com/diesel-rs/diesel/blob/7137e1d1e80d9e3ed75511a05f18b2b849ee6e73/diesel_codegen_syntex/src/update.rs#L24) while you are at it? :)

By the way, you could probably update some of the calls to `cx.span_err` with some of the new inline span annotations that came error message styles. Though diesel's code gen the error messages are okay as is.

diesel_codegen_syntex/src/update.rs
sgrif: The code just ended up flowing better handling it up here.

diesel_codegen_syntex/src/update.rs
sgrif: This line is way better w/ Macros 1.1 because the quoting library I'm using supports repeations so I can just impl `ToTokens` on `Attr` directly and do `$($fields)*` like I would in a normal macro

sgrif: Also lifetimes are way less of a pain

PR #442, opened 2016-09-13
Change associations to a Macros 1.1 compatible API

sgrif:
This does not re-implement associations to use Macros 1.1, but does include the
API change that will be required for them eventually. Macros 1.1 *only*
supports custom derive. We can do something crazy and dumb to make bang macros
work, but we cannot represent non-derive annotations. `#[as_changeset]` and
`#[insertable_into]` can both be represented nicely as derives, but
associations cannot individually be represented that way. So as a workaround
we'll just leave the annotations separate and require `#[derive(Associations)]`
to kick things off.

I have made this change for both syntex and nightly even though only nightly
requires it, as I want to keep those APIs in sync and deprecate syntex once
Macros 1.1 is on a path to stabilization.

The actual implementation is pretty straightforward, as we're just looping
through the attributes and calling out to what the macro system would have
called out to previously.

r? @diesel-rs/core


killercup:
LGTM. Looking forward to the actual macros 1.1 implementation :)

PR #443, opened 2016-09-13
Replace `#[changeset_for]` with `#[derive(AsChangeset)]`

sgrif:
This is required to make the API compatible with macros 1.1. While syntex could
theoretically keep the old API, I'd like to keep it in sync with nightly.
Macros 1.1 requires that we represent this as a custom derive. This does mean
that the same struct can no longer be used as a changeset for more than one
table (at least not with codegen) but that doesn't feel like something we need
to be concerned about. The stable macro is still an option for that case.


diesel_codegen_syntex/src/update.rs
```diff
@@ -3,80 +3,86 @@ use syntax::attr::AttrMetaMethods;
...
-            })
+fn changeset_options(
+    cx: &mut ExtCtxt,
+    span: Span,
+    attributes: &[ast::Attribute]
+) -> Result<ChangesetOptions, ()> {
+    let changeset_options_attr = attributes.iter().find(|a| a.check_name("changeset_options"));
+    let treat_none_as_null = try!(changeset_options_attr
+        .map(|a| extract_treat_none_as_null(cx, a))
+        .unwrap_or(Ok(false)));
```
killercup: That's a weird way to write this (at least it looks weird on my too-small phone display ;)). What does the `try!` accomplish when you always `unwrap_or`?

killercup: Oh wait the .map caller returns a result? Did you mean to use and_then? Or unwrap_or before the map and then replace the map with an new let binding with `try!(extract_treat…)`?

tl;dr I think the type is `Option<Result<_, _>>` but it can be simpler :)

sgrif: This will probably clarify:

```rust
let x: Option<&ast::Attribute> = x;
let y: Option<Result<bool, ()>> = y.map(|a| extract_treat_none_as_null(cx, a));
let z: Result<bool, ()> = z.unwrap_or(Ok(false));
let w: bool = try!(z);
```

sgrif: I'm not really sure there's any way to reasonably simplify it. It's an operation which can fail on a value which may not be present. We can't `unwrap_or` before the map because we'd have to synthesize an attribute from nowhere, and I don't think we'd want to do that anyway.

killercup: I was just confides by the nested try/map+unwrap. I think I understand it now, and I'd probably write it myself like

```rust
let treat_none_as_null = if let Some(a) = changeset_options_attr {
    try!(extract_treat_none_as_null(cx, a))
} else { false };
```

(I hate typing code on an iPhone.)

killercup:
LGTM. You can ignore the nit with the maybe-simplifyable code.

diesel_codegen_syntex/src/update.rs
sgrif: Ah I see what you mean.

sgrif: Question mark would make it way less difficult to parse. XD

killercup:
Oh, on travis, stable and beta (correctly) fail with an unknown attribute error. Is it correct that these new attributes are visible to the stable libsyntax?

A workaround is replacing

    #[changeset_options(treat_none_as_null = "true")]

with something like

    #[cfg_attr(diesel_derive, changeset_options(treat_none_as_null = "true"))]

sgrif:
No I need to rebase and then add changeset options to the list of attributes for Syntex to strip

PR #453, opened 2016-09-30
WIP Macros 1.1 port

sgrif:
This pull request is incomplete (unfortunately it will need to be all one pull request as I don't want a ton of git churn from crate renaming, and I won't be able to get the build in a passing state otherwise.

However, enough is done that I'd like some feedback on the overall structure before I finish this. /cc @killercup


sgrif:
Gah the insertable commit got pulled in here and I can't make it go away

diesel_codegen_new/src/model.rs
```diff
@@ -1,31 +1,76 @@
...
+        if character.is_uppercase() {
+            result.push('_');
+            for lowercase in character.to_lowercase() {
+                result.push(lowercase);
+            }
+        } else {
+            result.push(character);
+        }
+    }
+    result
```
sgrif: This function is just a straight copy paste from the old crate. I'd like to avoid having an inflector beyond the absolute simplest cases. It's been a major source of pain in Rails, as it's brittle, will never get everything right, and is impossible to change without breaking people's apps.

sgrif: Also admit it you just want to have a function called `to_pascal_case`

killercup: Interesting. I didn't know Rails had problems with that, but I can totally see that happening. Makes sense to keep this really simple; if it doesn't work you can just set the table name explicitly anyway.

> Also admit it you just want to have a function called to_pascal_case

Haha, you got me 😉 (I usually call it UpperCamelCase because it just sounds weird to me otherwise… "Huh, is anyone talking to me? Aww, no, they're still discussion string casing…")

diesel_codegen_new/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[rustc_macro_derive(AsChangeset)]
+pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
+    expand_derive(input, as_changeset::derive_as_changeset)
+}
+
+fn expand_derive(input: TokenStream, f: fn(syn::Item) -> quote::Tokens) -> TokenStream {
+    let input = input.to_string();
+    // FIXME: https://github.com/rust-lang/rust/issues/35900#issuecomment-245971366
+    let input = input.replace("#[structural_match]", "");
```
dtolnay: This was fixed upstream in https://github.com/rust-lang/rust/pull/36782 and the workaround is no longer necessary.

diesel_codegen_new/src/as_changeset.rs
```diff
@@ -0,0 +1,62 @@
...
+        `#[changeset_options(treat_none_as_null = "true")]`"#);
+
+    match options_attr.value {
+        syn::MetaItem::List(_, ref values) => {
+            if values.len() != 1 {
+                usage_err();
+            }
+            match values[0] {
+                syn::MetaItem::NameValue(ref name, ref value)
+                    if name.as_ref() == "treat_none_as_null" => value == "true",
```
dtolnay: I added an `impl<T: ?Sized> PartialEq<T> for Ident where T: AsRef<str>` in syn 0.5.0 so the explicit `as_ref()` will no longer be necessary.

diesel_codegen_new/src/as_changeset.rs
```diff
@@ -0,0 +1,62 @@
...
+fn treat_none_as_null(attrs: &[syn::Attribute]) -> bool {
+    let options_attr = match attr_with_name(attrs, "changeset_options") {
+        Some(attr) => attr,
+        None => return false,
+    };
+
+    let usage_err = || panic!(r#"`#[changeset_options]` must be in the form \
+        `#[changeset_options(treat_none_as_null = "true")]`"#);
+
+    match options_attr.value {
```
dtolnay: I filed https://github.com/dtolnay/syn/issues/25 to make attribute parsing less annoying.

diesel_codegen_new/src/as_changeset.rs
```diff
@@ -0,0 +1,62 @@
...
+    let struct_name = &model.name;
+    let table_name = model.table_name();
+    let struct_ty = &model.ty;
+    let mut lifetimes = item.generics.lifetimes;
+    let attrs = model.attrs.into_iter()
+        .filter(|a| a.column_name != Some(syn::Ident::new("id")))
+        .collect::<Vec<_>>();
+
+    if lifetimes.is_empty() {
+        lifetimes.push(syn::LifetimeDef {
```
dtolnay: I added `syn::LifetimeDef::new("'a")` in syn 0.8.5 to simplify this a bit.

diesel_codegen_new/src/ast_builder.rs
```diff
@@ -0,0 +1,16 @@
+use syn::{Ident, Ty, Path, PathSegment};
+
+pub fn ty_ident(ident: Ident) -> Ty {
```
dtolnay: Do you think it makes sense to add these upstream as constructors? `syn::Ty::ident(i)`, `syn::Ty::path(p)`, `syn::Path::ident(i)` or something like that. And should the ident ones take `Ident` or `Into<Ident>` so they can be called with strings?

diesel_codegen_new/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    let mut item = parse_item(&input);
```
dtolnay: I changed this to parse_macro_input in syn 0.6.0 which handles only structs and enums, as opposed to the entire array of items (functions, modules, typedefs, statics, extern crates, traits, impls, etc). The parse_item is behind a feature flag because it compiles a bit slower and most people don't need it.

diesel_codegen_new/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+            .any(|trait_name| KNOWN_CUSTOM_DERIVES.contains(&trait_name.as_ref()))
+    };
+
+    if finished_deriving_diesel_traits {
+        item.attrs = strip_attributes(item.attrs, KNOWN_CUSTOM_ATTRIBUTES);
+        strip_field_attributes(&mut item, KNOWN_FIELD_ATTRIBUTES);
+    }
+    let input = quote!(#item);
+
+    TokenStream::from_str(&format!("{} {}", input, output)).unwrap()
```
dtolnay: Or equivalently, `quote!(#item #output).to_string().parse().unwrap()`.

diesel_codegen_new/src/util.rs
```diff
@@ -0,0 +1,128 @@
...
+pub fn attr_with_name<'a>(
+    attrs: &'a [Attribute],
+    name: &str,
+) -> Option<&'a Attribute> {
+    let ident_name = Ident::new(name);
+    attrs.into_iter().find(|attr| *attr_name(attr) == ident_name)
+}
+
+pub fn attr_name(attr: &Attribute) -> &Ident {
+    match attr.value {
```
dtolnay: I added `attr.value.name() -> &str` in syn 0.7.1 and `attr.name() -> &str` in 0.8.5.

diesel_codegen_new/src/ast_builder.rs
sgrif: I think it makes sense for syn to only provide the AST definitions and parsing, and having a separate crate for the AST builder. I definitely think that abstraction needs to exist. I'm not confident enough in the usage patterns that have emerged to really find the right abstractions yet though.

PR #459, opened 2016-10-03
Re-implement `#[derive(Identifiable)]` using Macros 1.1

sgrif:
This was extracted from #453.

Going forward Macros 1.1 is the intended path of stabilization for
procedural macros, so `diesel_codegen` will need to be rewritten to use
it.

Much of the helper code around this is a direct port of the libsyntax
version of our code, rewritten to use `syn` instead.


sgrif:
Merging when CI is green again since this has been extracted from a reviewed PR

PR #461, opened 2016-10-03
Re-implement `#[derive(Associations)]` using Macros 1.1

sgrif:
Going forward Macros 1.1 is the intended path of stabilization for
procedural macros, so `diesel_codegen` will need to be rewritten to use
it.

Much of the helper code around this is a direct port of the libsyntax
version of our code, rewritten to use `syn` instead.

We're cloning the attributes, as `quote!` only works for repetition with
`T where for<'a> &'a T: IntoIterator`. I think this is a bug and will
address upstream.

### DIFF
--- a/diesel_codegen/src/associations.rs
+++ b/diesel_codegen/src/associations.rs
@@ -1,0 +1,103 @@
+use syn;
+use quote;
+
+use model::{Model, infer_association_name};
+use util::str_value_of_meta_item;
+
+pub fn derive_associations(input: syn::MacroInput) -> quote::Tokens {
+    let mut derived_associations = Vec::new();
+    let model = t!(Model::from_item(&input, "Associations"));
+
+    for attr in &input.attrs {
+        if attr.name() == "has_many" {
+            let options = t!(build_association_options(attr, "has_many"));
+            derived_associations.push(expand_has_many(&model, options))
+        }
+
+        if attr.name() == "belongs_to" {
+            let options = t!(build_association_options(attr, "belongs_to"));
+            derived_associations.push(expand_belongs_to(&model, options))
+        }
+    }
+
+    quote!(#(derived_associations)*)
+}
+
+fn expand_belongs_to(model: &Model, options: AssociationOptions) -> quote::Tokens {
+    let parent_struct = options.name;
+    let struct_name = &model.name;
+
+    let foreign_key_name = options.foreign_key_name.unwrap_or_else(||
+        to_foreign_key(&parent_struct.as_ref()));
+    let child_table_name = model.table_name();
+    let fields = &model.attrs;
+
+    quote!(BelongsTo! {
+        (
+            struct_name = #struct_name,
+            parent_struct = #parent_struct,
+            foreign_key_name = #foreign_key_name,
+            child_table_name = #child_table_name,
+        ),
+        fields = [#(fields)*],
+    })
+}
+
+fn expand_has_many(model: &Model, options: AssociationOptions) -> quote::Tokens {
+    let parent_table_name = model.table_name();
+    let child_table_name = options.name;
+    let foreign_key_name = options.foreign_key_name.unwrap_or_else(||
+        to_foreign_key(&model.name.as_ref()));
+    let fields = &model.attrs;
+
+    quote!(HasMany! {
+        (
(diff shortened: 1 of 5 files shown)


sgrif:

diesel_codegen/src/associations.rs
```diff
@@ -0,0 +1,103 @@
...
+        fields = [#(fields)*],
+    })
+}
+
+fn expand_has_many(model: &Model, options: AssociationOptions) -> quote::Tokens {
+    let parent_table_name = model.table_name();
+    let child_table_name = options.name;
+    let foreign_key_name = options.foreign_key_name.unwrap_or_else(||
+        to_foreign_key(&model.name.as_ref()));
+    let fields = model.attrs.clone();
```
sgrif: /cc @dtolnay It looks like this is an issue in quote. I can't actually use borrowed data here. Since you're always putting at addr-of in front of the value before iterating over it, it's effectively `T where for<'a> &'a T: IntoIterator`, which is only true for `Vec`. It looks like *with* a separator you are taking ownership of the value though. Worth always taking ownership of the iterator? Do you have a better suggestion?

dtolnay: You're right, let's always take ownership of the iterator. What do you think of adding `#(&fields)*` as a way to not take ownership equivalent to what happens currently?

sgrif: Seems fine. Although since it needs to be assigned to a local variable anyway, the same could always be achieved by sticking a `ref` or `&` somewhere.

sgrif: I think the use case for quasi quoting a vector and then using it again afterwards is significantly less than the reasoning behind `println!` taking a reference though.

diesel_codegen/src/associations.rs
```diff
@@ -0,0 +1,103 @@
+use syn;
+use quote;
+
+use model::{Model, infer_association_name};
+use util::str_value_of_meta_item;
+
+pub fn derive_associations(input: syn::MacroInput) -> quote::Tokens {
+    let mut derived_associations = Vec::new();
+    let model = t!(Model::from_item(&input, "Associations"));
```
killercup: Instead of using strings here (and const lists of strings in lib), you could use a set of modules with constants for derives and attributes [like this](https://play.rust-lang.org/?gist=6a7fea0458a4d3fbd8cbf9b3fbc5c513&version=stable&backtrace=0). (Doesn't need to be in this PR. I just hate doing stringly typed stuff in a language like Rust 😉 )

diesel_codegen/src/associations.rs
```diff
@@ -0,0 +1,103 @@
...
+
+fn build_association_options(
+    attr: &syn::Attribute,
+    association_kind: &str,
+) -> Result<AssociationOptions, String> {
+    let usage_error = Err(format!(
+            "`#[{}]` must be in the form `#[{}(table_name, option=value)]`",
+            association_kind, association_kind));
+    match attr.value {
+        syn::MetaItem::List(_, ref options) if options.len() >= 1 => {
```
killercup: `options.len() >= 1` can probably be `!options.is_empty()`

diesel_codegen/src/associations.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use model::{Model, infer_association_name};
+use util::str_value_of_meta_item;
+
+pub fn derive_associations(input: syn::MacroInput) -> quote::Tokens {
+    let mut derived_associations = Vec::new();
+    let model = t!(Model::from_item(&input, "Associations"));
+
+    for attr in &input.attrs {
+        if attr.name() == "has_many" {
```
killercup: You could also match on `attr.name()` and combine the `if`s

sgrif: I avoided that because of the additional rightward drift, and the need for a no-op `_` arm.

diesel_codegen/src/associations.rs
sgrif: Good poitn

diesel_codegen/src/associations.rs
sgrif: Done and released. Thanks.

PR #462, opened 2016-10-03
Pin the nightly version again

sgrif:
screen shot 2016-10-03 at 2 55 33 pm

3 hours later, there was a breaking change in nightly...


sgrif:

samphippen:
LGTM

sgrif:
Once the macros 1.1 port is finished we won't be affected by any breakage in nightly, so it'll be fine.

PR #463, opened 2016-10-03
Allow strings starting with `env:` and `dotenv:` to be used

sgrif:
Due to implementation details of the macros 1.1 port, we will no longer
be able to take arbitrary macros as arguments to `infer_schema!` and
`infer_table_from_schema!`. We can only take literal values. In
practice, the only two useful things to take here are `env!` and
`dotenv!`. We can special case those and represent them as strings
instead.

We cannot detect whether people are attempting to use `env!` here, as it
is opaque to us by the time we reach our code in the old procedural
macros. However, the string form will be the only form that works in the
macros 1.1 implementation, and it will be deprecated along with the
syntex crate when macros 1.1 is stable.


sgrif:

samphippen:
LGTM

PR #465, opened 2016-10-03
Implement `infer_schema!` with Macros 1.1

sgrif:
What's that RFC authors? You want to specifically limit the RFC to
stabilize only custom-derive and exclude bang macros? WE'LL SEE ABOUT
THAT!

I had to comment out one test from the syntex crate, as cargo won't let
me have a feature called dotenv while dotenv is a dev dependency. That
code will move to the shared crate soon, so it shouldn't be a problem.


sgrif:

diesel/src/macros/mod.rs
```diff
@@ -551,6 +551,19 @@ macro_rules! print_sql {
...
+#[macro_export]
+/// This macro can only be used in combination with the `diesel_codegen` or
+/// `diesel_codegen_syntex` crates. It will not work on its own.
+///
+/// FIXME: Oh look we have a place to actually document this now.
+macro_rules! infer_schema {
+    ($database_url: expr) => {
+        #[derive(InferSchema)]
+        #[options(database_url=$database_url)]
+        struct _Dummy;
```
steveklabnik: I wonder if there's any better name here, even though it is just an implementation detail.

diesel_codegen/src/schema_inference.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `infer_schema!");
+    }
+
+    let options = input.attrs.iter().find(|a| a.name() == "options").map(|a| &a.value);
+    let options = match options {
+        Some(&syn::MetaItem::List(_, ref options)) => options,
+        _ => bug()
```
steveklabnik: do you do trailing commas in diesel? this one is missing one

diesel_codegen/src/schema_inference.rs
```diff
@@ -0,0 +1,29 @@
...
+    };
+    let database_url = options.iter().find(|a| a.name() == "database_url")
+        .map(|a| str_value_of_meta_item(a, "database_url"))
+        .unwrap_or_else(|| bug());
```
steveklabnik: I think this can just be `unwrap_or_else(bug)`?

diesel_codegen_shared/src/schema_inference/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+        #[cfg(feature = "postgres")]
+        InferConnection::Pg(c) => pg::load_table_names(&c),
+    }
+}
+
+fn establish_connection(database_url: &str) -> Result<InferConnection, Box<Error>> {
+    let database_url = try!(extract_database_url(database_url));
+    match database_url {
+        #[cfg(feature = "postgres")]
+        _ if database_url.starts_with("postgres://") || database_url.starts_with("postgresql://") => {
```
steveklabnik: imho, `match` with `_ if` is a code smell, but maybe it's easier to do the `cfg`s this way?

diesel_codegen/src/schema_inference.rs
sgrif: I tried that, but the compiler didn't figure out to that `fn() -> !` can coerce to ,`Fn() -> ExpectedType` without this.

diesel_codegen_shared/src/schema_inference/mod.rs
sgrif: Yeah. cfg can't be used on arbitrary items yet and this code is gross with separate functions for each variant.

steveklabnik: :+1:

diesel_codegen/src/schema_inference.rs
steveklabnik: seems reasonable then

diesel_codegen/src/schema_inference.rs
```diff
@@ -0,0 +1,29 @@
...
+use quote;
+
+use diesel_codegen_shared::load_table_names;
+
+use util::str_value_of_meta_item;
+
+pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `infer_schema!");
```
killercup: Maybe add a link to <https://github.com/diesel-rs/diesel/issues/new>?

diesel/src/macros/mod.rs
killercup: Yeah, having a `_Dummy` type in user code (which might be observable if you tell rustdoc to render private types as well) might not be as nice as haven, say, `_DieselSchemaInferenceHelper`.

diesel_codegen_shared/src/schema_inference/sqlite.rs
```diff
@@ -0,0 +1,20 @@
+use diesel::*;
+use diesel::sqlite::SqliteConnection;
+use std::error::Error;
+
+table! {
```
killercup: Travis says `table!` is undefined here :(

diesel/src/macros/mod.rs
sgrif: The `_Dummy` struct will never appear, it's eliminated entirely during macro expansion

sgrif:
Travis is refusing to re-run but I believe I've fixed the failures

PR #467, opened 2016-10-04
Implement `infer_table_from_schema!` as a Macros 1.1 macro

sgrif:
This was a bit tougher to pull out since the code responsible for
loading the schema information was pretty tightly coupled to libsyntax.
I've attempted to return as much relevant information as possible for
multiple AST frontends in an agnostic way.

This should fix the failures on master as well.


diesel_codegen/src/schema_inference.rs
```diff
@@ -11,19 +11,79 @@ pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
...
-    };
-    let database_url = options.iter().find(|a| a.name() == "database_url")
-        .map(|a| str_value_of_meta_item(a, "database_url"))
-        .unwrap_or_else(|| bug());
+    let options = get_options_from_input(&input.attrs, bug);
+    let database_url = get_option(&options, "database_url", bug);
 
     let table_names = load_table_names(&database_url).unwrap();
     let schema_inferences = table_names.into_iter().map(|table_name| {
         quote!(infer_table_from_schema!(#database_url, #table_name);)
```
killercup: I love how this tiny line right here is sets of a huge awesome piece of rust-macro-derive-inception as it (1) is part of a macro (2) that is contained a custom derive (3) which expands to a (set of) macro calls (4) that themselves include custom derives (5) where each one actually expands to calls to the `tables!` macros. You could probably write a few pages of internal docs just about that 😄

diesel_codegen/src/schema_inference.rs
```diff
@@ -11,19 +11,79 @@ pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
...
+               with your invocation of `infer_table_from_schema!");
+    }
+
+    let options = get_options_from_input(&input.attrs, bug);
+    let database_url = get_option(&options, "database_url", bug);
+    let table_name = get_option(&options, "table_name", bug);
+
+    let connection = establish_connection(&database_url).unwrap();
+    let data = get_table_data(&connection, &table_name).unwrap();
+    let primary_keys = get_primary_keys(&connection, &table_name).unwrap()
```
killercup: I'm fairly certain you can omit the `&` in `&table_name` as it's already a `&str`. (Same for `&options` and `&database_url` above.)

diesel_codegen_shared/src/schema_inference/mod.rs
```diff
@@ -52,3 +55,49 @@ fn establish_real_connection<Conn>(database_url: &str) -> Result<Conn, Box<Error
...
+    conn: &InferConnection,
+    table_name: &str,
+) -> Result<Vec<String>, Box<Error>> {
+    let primary_keys = try!(match *conn {
+        #[cfg(feature = "sqlite")]
+        InferConnection::Sqlite(ref c) => sqlite::get_primary_keys(c, table_name),
+        #[cfg(feature = "postgres")]
+        InferConnection::Pg(ref c) => pg::get_primary_keys(c, table_name),
+    });
+    if primary_keys.len() == 0 {
```
killercup: `primary_keys.is_empty()`

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -2,6 +2,98 @@ use diesel::*;
 use diesel::pg::PgConnection;
 use std::error::Error;
 
+use super::data_structures::*;
+
+table! {
+    pg_attribute (attrelid) {
```
killercup: maybe add a link to <https://www.postgresql.org/docs/9.5/static/catalog-pg-attribute.html> since the field names are pretty weird (same for the next 3 comments or so)

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -2,6 +2,98 @@ use diesel::*;
...
+        attname -> VarChar,
+        atttypid -> Oid,
+        attnotnull -> Bool,
+        attnum -> SmallInt,
+        attisdropped -> Bool,
+    }
+}
+
+table! {
+    pg_type (oid) {
```
killercup: maybe add a link to <https://www.postgresql.org/docs/9.5/static/catalog-pg-type.html>

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -2,6 +2,98 @@ use diesel::*;
...
+        typname -> VarChar,
+    }
+}
+
+joinable!(pg_attribute -> pg_type (atttypid));
+select_column_workaround!(pg_attribute -> pg_type (attrelid, attname, atttypid, attnotnull, attnum, attisdropped));
+select_column_workaround!(pg_type -> pg_attribute (oid, typname));
+
+table! {
+    pg_index (indrelid) {
```
killercup: https://www.postgresql.org/docs/9.5/static/catalog-pg-index.html

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -2,6 +2,98 @@ use diesel::*;
...
+        indrelid -> Oid,
+        indexrelid -> Oid,
+        indkey -> Array<SmallInt>,
+        indisprimary -> Bool,
+    }
+}
+
+table! {
+    pg_class (oid) {
```
killercup: https://www.postgresql.org/docs/9.5/static/catalog-pg-class.html

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -2,6 +2,98 @@ use diesel::*;
...
+        oid -> Oid,
+        relname -> VarChar,
+    }
+}
+
+pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box<Error>> {
+    let is_array = attr.type_name.starts_with("_");
```
killercup: Why do array types start with an underscore? This ist just the values of `pg_types.typname`, right? (The [docs](https://www.postgresql.org/docs/9.5/static/catalog-pg-type.html) don't say much about the values in there)

sgrif:
Tests should *finally* be green. @diesel-rs/contributors review?

examples/getting_started_step_1/Cargo.toml
```diff
@@ -5,7 +5,6 @@ authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
 
 [dependencies]
 diesel = "0.7.1"
-diesel_codegen = "0.7.2"
+diesel_codegen = { version = "0.7.2", features = ["postgres"] }
 diesel_codegen_old = { path = "../../diesel_codegen_old", features = ["postgres"] }
```
killercup: wrong pr, but: you need specify this with a version to publish it to crates.io, right?

diesel_codegen_shared/src/schema_inference/pg.rs
sgrif: No clue. It's probably not safe to rely on, but it was the simplest path forward at the time. This is maintaining the existing behavior.

examples/getting_started_step_1/Cargo.toml
sgrif: This crate will never be published

killercup: Ah, of course! And this is a good way to ensure this :)

diesel_codegen/src/schema_inference.rs
sgrif: Added like 6 more `&`s because I can

killercup: What good is a borrow checker if you don't borrow stuff, right?

PR #469, opened 2016-10-06
Remove `embed_migrations!` temporarily

sgrif:
Unfortunately it will be impossible to write this macro using the Macros
1.1 API. While this is an important feature that I want to keep, I more
importantly want to keep the main repository working on stable. It will
have to live as a separate crate, and possibly have an API change to
work around build scripts.


sgrif:

killercup:
So, the 0.8 announcement should probably mention that people using `embed_migrations!` should not update just yet.

CI failure is because of `extern crate dotenv_codegen` in `diesel_test/build.rs` and breaking changes in syntax. (Did you mean to remove the specific nightly date?)

sgrif:
Fixed those failures, but now seeing a really cryptic message from Cargo about path overrides doing things incorrectly. We've also got the rustc-macro -> proc-macro rename landing tonight so I'm going to wait until tomorrow to figure this out

sgrif:
And there's more dead code to remove

weiznich:
@sgrif If we change the root location of the relative path used in this macro it is possible to implement this using macros 1.1. See #470 .

PR #470, opened 2016-10-09
Implement embed_migration! using macros 1.1

weiznich:
* port embed_migration to macros 1.1
* remove codegen_old
* some minor fixes

This is a breaking change, because the root path passed to `embed_migrations!` have changed. It's now CARGO_MANIFEST_HOME, but was before the location of the file that called that macro.

PR #471, opened 2016-10-10
Resolve paths given to `embed_migrations` as relative to Cargo.toml

sgrif:
Macros 1.1 does not give us any way to access the file the macro was
invoked from. In order to make the API compatible with this new system,
we instead resolve paths given as relative to the only stable path we
have access to, which is the location of Cargo.toml. Since it is
somewhat common for people to place their migrations in
`src/migrations`, we will continue to search that path if no specific
path is given. However, we will no longer find paths such as
`src/foo/migrations`.

This is based on code from #470, and I've committed as the author of that PR to ensure @weiznich is credited.


sgrif:
@killercup Got time for a review?

PR #472, opened 2016-10-10
Reimplement `embed_migrations!` using Macros 1.1

sgrif:
There wasn't a ton of code that could be shared between the old crate
and the new one, as the majority of the lines of code are just the setup
of the structs and functions used, which need to go into the respective
quote macro. In order to share that code, it'd have to be moved into
Diesel proper and out of the macro output, which I want to avoid. Since
the syntex crate will be removed once Macros 1.1 is stable, we can live
with the duplication for now.

This is based on code from #470, and I've committed as the author of that PR.

### DIFF
--- a/diesel_codegen/src/embed_migrations.rs
+++ b/diesel_codegen/src/embed_migrations.rs
@@ -1,0 +1,91 @@
+use syn;
+use quote;
+
+use diesel::migrations::{migration_paths_in_directory, version_from_path};
+use diesel_codegen_shared::migration_directory_from_given_path;
+use std::error::Error;
+use std::path::Path;
+
+use util::{get_options_from_input, get_option};
+
+pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `embed_migrations!");
+    }
+
+    let options = get_options_from_input(&input.attrs, bug);
+    let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
+    let migrations_expr = migration_directory_from_given_path(migrations_path_opt)
+        .and_then(|path| migration_literals_from_path(&path));
+    let migrations_expr = match migrations_expr {
+        Ok(v) => v,
+        Err(e) => panic!("Error reading migrations: {}", e),
+    };
+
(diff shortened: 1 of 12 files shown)


sgrif:
@killercup Review?

diesel_codegen/src/embed_migrations.rs
```diff
@@ -0,0 +1,91 @@
...
+use diesel_codegen_shared::migration_directory_from_given_path;
+use std::error::Error;
+use std::path::Path;
+
+use util::{get_options_from_input, get_option};
+
+pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `embed_migrations!");
```
killercup: Maybe add a link to <https://github.com/diesel-rs/diesel/issues/new>?

diesel_codegen/src/embed_migrations.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    let options = get_options_from_input(&input.attrs, bug);
+    let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
+    let migrations_expr = migration_directory_from_given_path(migrations_path_opt)
+        .and_then(|path| migration_literals_from_path(&path));
+    let migrations_expr = match migrations_expr {
+        Ok(v) => v,
+        Err(e) => panic!("Error reading migrations: {}", e),
```
killercup: Isn't that `match` the same as `migrations_expr.expect("Error reading migrations")` (also print the error afterwards)?

diesel_codegen/src/lib.rs
```diff
@@ -94,6 +96,13 @@ pub fn derive_infer_table_from_schema(input: TokenStream) -> TokenStream {
         .to_string().parse().unwrap()
 }
 
+#[rustc_macro_derive(EmbedMigrations)]
+pub fn derive_embed_migratoins(input: TokenStream) -> TokenStream {
```
killercup: "migratoins"

diesel_codegen/src/embed_migrations.rs
sgrif: `expect` will use the `Debug` implementation of the error, not `Display`

killercup: Oh, right!

PR #473, opened 2016-10-10
Remove `diesel_codegen_old`, unpin nightly

sgrif:
This removes the dead code that was laying around as the result of the
now defunct pre-macros 1.1 implementation of codegen. I've also unpinned
the nightly version, as we should no longer break due to changes in
libsyntax. Finally, I've made sure the newly added crates are run as
part of the test suite.

The removal of `.cargo/config` and the addition of `[replace]` all over
the place is to work around
https://github.com/rust-lang/cargo/commit/fc0e64262a3c3b461f004d6fa75732ca656e73bd.
I'm not sure what our long term plan is yet, as juggling that doesn't
sound fun to me.


sgrif:
Review? @killercup

sgrif:
Uh I guess I did this in the wrong order but I merged after squashing

sgrif:
Oh I forgot to push is why

PR #474, opened 2016-10-10
Documentation pass for `diesel_codegen` [ci skip]

sgrif:
This adds proper documentation for everything provided by `codegen`.
With the old crate, it was just a giant blob of text in the README. Now
the procedural macros all appear in the proper documentation, so we can
document everything inline.


sgrif:
Review? @killercup

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -2,7 +2,11 @@
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
 ///
-/// FIXME: Oh look we have a place to actually document this now.
+/// Queries the database for the names of all tables, and calls
```
killercup: This sentence should be the first paragraph, so the rustdoc summary shows it instead of "This macro can only be used in combination …". Same for the doc comments below.

diesel/src/associations/mod.rs
```diff
@@ -122,6 +122,19 @@ use query_source::Table;
...
+/// database. This must be implemented to use associations, and some features of
+/// updating.
+///
+/// ### Deriving
+///
+/// This trait can be automatically derived using `diesel_codegen` by adding
+/// `#[derive(Identifiable)]` to your struct. The primary key will be assumed to
+/// be a field and column called `id`. By default the table will be
+/// assumed to be the plural form of the struct name (using *very* dumb
+/// pluralization). If your table name differs from that convention, or requires
```
killercup: If the convention is so dumb you can probably describe it half a sentence, e.g.

> using *very* dumb pluralization -- it just adds an 's' at the end

I'm also wondering if this is something we want to guarantee any kind of stability of (which we should mention regardless of yes or no).

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -15,7 +19,15 @@ macro_rules! infer_schema {
...
-/// FIXME: Oh look we have a place to actually document this now.
+/// Establishes a database connection at compile time, loads the schema information about a table's
+/// columns, and invokes [`table!`](macro.table!.html) for you automatically. Attempting to use the
+/// `env!` or `dotenv!` macros here will not work due to limitations of the Macros 1.1 system, but
+/// you can pass a string in the form `"env:SOME_ENV_VAR"` or `"dotenv:SOME_ENV_VAR"` to achieve
+/// the same effect.
+///
+/// At this time, the schema inference macros do not support types from third party crates, and
+/// having any columns with a type not already supported will result in a compiler error (please
+/// open an issue if this happens unexpectedly for a type listed in our docs.)
```
killercup: Another good place to link to <https://github.com/diesel-rs/diesel/issues/new>

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -15,7 +19,15 @@ macro_rules! infer_schema {
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
 ///
-/// FIXME: Oh look we have a place to actually document this now.
+/// Establishes a database connection at compile time, loads the schema information about a table's
+/// columns, and invokes [`table!`](macro.table!.html) for you automatically. Attempting to use the
```
killercup: Is `macro.table!.html` the correct URL? I saw a rustdoc PR that go rid of the bangs in URLs but I don't know if it got merged (yet).

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -15,7 +19,15 @@ macro_rules! infer_schema {
...
+/// `env!` or `dotenv!` macros here will not work due to limitations of the Macros 1.1 system, but
+/// you can pass a string in the form `"env:SOME_ENV_VAR"` or `"dotenv:SOME_ENV_VAR"` to achieve
+/// the same effect.
+///
+/// At this time, the schema inference macros do not support types from third party crates, and
+/// having any columns with a type not already supported will result in a compiler error (please
```
killercup: "not already supported" → "not supported by the core diesel crate"

diesel/src/associations/mod.rs
```diff
@@ -122,6 +122,19 @@ use query_source::Table;
 
 pub use self::belongs_to::{BelongsTo, GroupedBy};
 
+/// Represents a struct which can be identified on a single table in the
+/// database. This must be implemented to use associations, and some features of
+/// updating.
+///
+/// ### Deriving
+///
+/// This trait can be automatically derived using `diesel_codegen` by adding
```
killercup: Maybe link `diesel_codegen` to its readme/crate docs/page with explanation how to use it?

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -4,13 +4,20 @@ use query_source::QuerySource;
...
-/// `#[changeset_options(treat_none_as_null = "true")]`
+/// ### Deriving
+///
+/// This trait can be automatically derived using `diesel_codegen` by adding
+/// `#[derive(AsChangeset)]` to your struct.  Structs which derive this trait
+/// must be annotated with `#[table_name = "something"]`. If the field name of
+/// your struct differs from the name of the column, you can annotate the field
+/// with `#[column_name = "some_column_name"]`.
+///
+/// By default, any option fields on the struct are skipped if their value is
```
killercup: "option" → "Option" or "optional"

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -4,13 +4,20 @@ use query_source::QuerySource;
...
+/// `None`. If you would like to assign `NULL` to the field instead, you can
+/// annotate your struct with `#[changeset_options(treat_none_as_null =
+/// "true")]`
```
killercup: period at the end

diesel/src/macros/macros_from_codegen.rs
sgrif: It's the URL currently on the website. I'd expect that they will make old URLs continue to work.

diesel/src/macros/macros_from_codegen.rs
sgrif: I think it makes more sense here than in the library itslef.

diesel/src/associations/mod.rs
sgrif: Can you expand on what you mean by stability and how we should elaborate on it?

killercup: Imagine you have a struct called `Mouse` and diesel 0.8 assumes your table name is `Mouses` – which it for some reason it actually is. Now, diesel 0.8.2 comes with an awesome new pluralization and infers your table name is `Mice`.  You are also not using `infer_schema` so your code continues to compile. Sadly, your application doesn't work any longer.

We should say whether we guarantee this to never happen.

sgrif: Yes.

For the record, this is exactly why I am against any "real" pluralization logic. It is never going to be correct, and we can't "fix" it without breaking people's apps.

sgrif: How about `The inferred table name is considered public API and will never change in without a major version bump.`

killercup: I think there's an "in" too many in "…will never change in without …"

PR #476, opened 2016-10-12
Make database URL tests work in parallel.

### LINKED ISSUE #475: A few of our tests are quite flaky
Specifically the tests in https://github.com/diesel-rs/diesel/blob/master/diesel_codegen_shared/src/database_url.rs are failing intermittently. The fix is likely just to make sure each test has a uniquely named environment variable that it's working with.

cyplo:
There was a collision on the name of the environment variable "foo"
between two tests. One would set it (extract_database_url_returns_env_vars) and other (extract_database_url_errors_if_env_var_is_unset) would unset it.
As Cargo runs tests in parallel this would lead to this set of test
having intermittent failures.

Changed the names of the environment variables used in the tests to make
them unique between tests, hence able to run in parallel.

Should close #475

### DIFF
--- a/diesel_codegen_shared/src/database_url.rs
+++ b/diesel_codegen_shared/src/database_url.rs
@@ -41,14 +41,14 @@ fn extract_database_url_returns_the_given_string() {
 
 #[test]
 fn extract_database_url_returns_env_vars() {
-    env::set_var("foo", "lololol");
-    env::set_var("bar", "trolololol");
-    assert_eq!("lololol", extract_database_url("env:foo").unwrap());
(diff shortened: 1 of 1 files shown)


cyplo:
These were the only vars used in tests for that file. Haven't search for all vars in all tests though, maybe worth making a separate issue to do just that  - not sure here.

killercup:
Thanks again for doing this!

(And so quickly! When I say "I'll look into it on the weekend" I always end up doing it two weeks later…)

cyplo:
Thanks ! Same here, I just had a bit of unexpected "can't go to work today" time, so why not spend it pushing a PR ? :D

PR #479, opened 2016-10-13
Allow SQLite date/times to be used with strings

sgrif:
Here's the part where we start to fully embrace the dynamic bullshit
that SQLite forces on us. Unfortunately, dates and times aren't
recognized as separate types. Things like `CURRENT_TIMESTAMP` and
`DATETIME` return strings. I had hoped SQLite would give us some way to
get the underlying unix timestamp so we could avoid date parsing, but
that appears to not be the case. That means that while we can build off
of types like `PgTimestamp` which represent purely the structure that
the database exposes, the equivalent primitive for SQLite is `String`.
That also means that we will not be able to provide impls for
`std::time::SystemTime` with SQLite, as that conversion would require
date parsing.

We can build chrono support on top of this, but I wanted to get the
primitive representation in first.


sgrif:

sgrif:
Yeah, I'll address the docs after I add chrono support. I'm not super worried about adding specific tests for `&str` (we don't have explicit tests for `&[u8]` vs `Vec<u8>` either)

PR #480, opened 2016-10-14
Fix #384 (Invalid query on empty insert/select in)

### LINKED ISSUE #384: Insert is crashing while trying to insert an empty vector

weiznich:
Check if a queryresult would be empty
- do not execute such queries, because they are invalid, instead simply
  return nothing

Fix #384

### DIFF
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -64,6 +64,10 @@ impl<T, U, DB> QueryFragment<DB> for In<T, U> where
         self.left.is_safe_to_cache_prepared() &&
             self.values.is_safe_to_cache_prepared()
     }
+
+    fn is_empty(&self) -> bool {
+        self.values.is_empty()
+    }
 }
 
 impl_query_id!(In<T, U>);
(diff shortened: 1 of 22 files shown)


killercup:
Oh, nice! Do we handle empty arrays anywhere else? Can you add a test for this?

weiznich:
I realized that this fixes the issue only partially. I will try to add tests and provide a better fix for this in the next few days.

weiznich:
I've updated the implementation to handle more cases. Also I realized that this could also fix #384.
I'm not sure if I've implemented `is_empty` in all places where it is needed.

killercup:
@weiznich I just remembered that I tried fixing that and @sgrif and I discussed an implementation in <https://github.com/diesel-rs/diesel/pull/385>, which resulted in our idea to refactor the error handling instead of introducing an `is_empty` method for this one case.

weiznich:
I think using a custom error for this would be problematic, because I think a user expect such queries to succeed. Furthermore with using an `is_empty` method we can save a call into the database, because we can reason about situations where the query would be illegal.

weiznich:
Any news on this? It's annoying and error prone to check in every place where you use eq_any or insert for an empty input.

sgrif:
@weiznich As mentioned, I'd prefer an implementation that looked something like https://github.com/diesel-rs/diesel/pull/385#issuecomment-236690138. The query would still succeed from the user's point of view, as we'd handle the error in the connection itself. See the discussion on the linked thread.

sgrif:
Going to tackle this tomorrow

weiznich:
@sgrif I've opened a pullrequst to @killercup repo which rebases and improves his pull request.

PR #489, opened 2016-11-01
Fix failures on latest nightlies

### LINKED ISSUE #485: Macros 1.1 broken on nightly-2016-10-20
If I try to compile a program which uses diesel 0.8.0, using rustc nightly-2016-10-20, the compilation fails in my crate with the following errors:

```
error: `Identifiable` is a derive mode
  --> src/models/account.rs:16:1
   |
16 | pub struct Account {
   | ^

error: `Queryable` is a derive mode
  --> src/models/account.rs:16:1
   |
16 | pub struct Account {
   | ^

error: `Insertable` is a derive mode
  --> src/models/account.rs:45:1
   |
45 | pub struct NewAccount {
   | ^
```

The program compiles and runs successfully on nightly-2016-10-19. Unless this is a bug in the latest nightly, the assumption that macros 1.1 would keep diesel from breaking with nightly updates may have been incorrect.

sgrif:
Macros 1.1 was changed so that bang macros and custom derive macros
share the same namespace. This causes problems for us, since we expose
and use a bang macro with the same name as the custom derive macro. I
was hoping we could have this changed upstream, but
https://github.com/rust-lang/rust/issues/35900#issuecomment-257190405
makes it seem like that's not going to happen. I'm not sure what my long
term plan for the bang macros are, but in the short term we can just
expose them under another name internally. Those who are using the bang
macros are likely not using them in conjunction with `diesel_codegen`.

Fixes #485.

### DIFF
--- a/diesel/src/macros/as_changeset.rs
+++ b/diesel/src/macros/as_changeset.rs
@@ -71,12 +71,20 @@
 /// ```
 #[macro_export]
 macro_rules! AsChangeset {
+    ($($args:tt)*) => {
+        _AsChangeset!($($args)*);
+    };
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! _AsChangeset {
     // Provide a default value for treat_none_as_null if not provided
     (
         ($table_name:ident)
         $($body:tt)*
     ) => {
-        AsChangeset! {
+        _AsChangeset! {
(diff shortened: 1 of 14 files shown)


sgrif:

sgrif:
This was merged.

Issue #437, opened 2016-09-12
CLI: Add a subcommand to print db schema as `table!` calls


killercup:
I've been thinking about adding a `diesel_cli` subcommand to use `infer_schema` to print out the current schema as calls to `table!`.

The motivation is that for a large application I'd want to have the current schema somewhere in the repo and not just be the result of a bunch of migrations. (This might make it easier to join an existing diesel project as well.)

First draft for the CLI usage:

```
diesel-schema
Commands for operating on the database's schema

USAGE:
    diesel schema [FLAGS] [OPTIONS] [SUBCOMMAND]

FLAGS:
    -h, --help    Prints help information

OPTIONS:
        --database-url <DATABASE_URL>    Specifies the database URL to connect to. Falls back to the DATABASE_URL environment variable if unspecified.

SUBCOMMANDS:
    help     Prints this message
    print    Print the database's schema as code that can be used in a diesel-based project.
```

iamsebastian:
Would be a great addition, as I write a new Postgresql-rust/Nickel project, based on an existent MySQL-PHP project with dozens of tables.

sgrif:
@iamsebastian Is there a reason you're unable to use `infer_schema!` for that?

iamsebastian:
Am I able *to print out* the schema with the `infer_schema!` macro, @sgrif ?

sgrif:
No, what I mean is that I'm curious why you need to print out the schema instead of inferring it.

iamsebastian:
Because sometimes, as I started to use rust & diesel.rs, I struggled around if I had chosen the correct types, that match the types defined in the different postgres-tables (I migrated freshly to a postgres database from our mysql-production database).

bronson:
I also hope for the ability to print the schema.  Rails's `schema.rb` has saved my bacon a number of times when migrations diverged for whatever reason ( a dev ran a migration on a branch but it changed when hitting master, etc).

Committing `schema.rb` to the source tree makes it really easy to see if the user's database matches what the code expects.  Without it, database issues become more of a poke-and-pray operation.

marcusball:
Regarding a reason why one might not want to use `infer_schema!`: I believe this subcommand would solve an issue I encountered a few weeks ago—correct me if this wouldn't. 

[...]

PR #493, opened 2016-11-12
Fix breakage in latest nightlies

sgrif:
Note: This is not currently expected to pass CI, nor is it ready to be
merged. Nightly builds have been failing since a few days before this
change was made upstream. This is a PR that I *think* will fix it based
on testing from a local build, and I want to have this ready to go once
nightlies are building again.

The latest nightlies have added built-in support for custom attributes,
but have removed the ability to modify the input token stream. This
means that our bang macros now need to worry about namespace collisions.
For `embed_migrations` this doesn't matter, since the output is a module
with a known name. `infer_schema` is ok as well, since it's not expected
to be called more than once. However, `infer_table_from_schema` outputs
a module which differs, and the API currently takes a string which we
can't turn into a module name. Even if we changed the API to take an
ident instead of an expr, we can't then pass that ident into the
procedural macros. We can work around this in our invocation of it from
`infer_schema!`, but this will force this API change onto our users.

Since `infer_table_from_schema!` isn't highly used, I think this is OK,
but we might want to start looking at alternatives long term. There have
been murmurs of a proper bang macros 1.1, which is the best solution. We
could also go back to only allowing `infer_schema!` on nightly, and
change to a solution that involves Diesel CLI or build scripts, but both
of those will increase the barrier to entry for newcomers.

Either way this fixes the issue short term, and assuming all works when
nightlies are building again can be released immediately afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## Unreleased
+
+### Changed
+
+* Fixed support for nightlies later than ???
+
+* Removed support for nightlies earlier than ???
+
+* Calls to `infer_table_from_schema!` will need to be wrapped in a module if
+  called more than once. This change is to work around further limitations of
+  the Macros 1.1 system. Example:
+
+  ```rust
+  mod infer_users {
+      infer_table_from_schema!("dotenv:DATABASE_URL", "users");
+  }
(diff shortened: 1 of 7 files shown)


sgrif:

killercup:
Interesting change. The code looks good, but I'm not familiar enough with the macros 1.1 changes to actually determine if it _is_ good :)

> Since infer_table_from_schema! isn't highly used, I think this is OK, but we might want to start looking at alternatives long term.

I agree. Also, this is another reminder to implement #437.

sgrif:
Seeing ICEs on the latest nightlies so I'm holding off on merging this for now. https://github.com/rust-lang/rust/issues/35900#issuecomment-261091872

sgrif:
Rust master should be fixed. Will check again tomorrow

PR #494, opened 2016-11-12
Refactor `AsChangeset!` now that macros in type positions are stable

sgrif:
Cleaner code, less recursion. We can just return the type now!

### DIFF
--- a/diesel_tests/Cargo.toml
+++ b/diesel_tests/Cargo.toml
@@ -31,6 +31,10 @@ stable_sqlite = ["with-syntex", "sqlite", "diesel_codegen_syntex/sqlite"]
 unstable_postgres = ["unstable", "postgres", "diesel_codegen/postgres"]
 unstable_sqlite = ["unstable", "sqlite", "diesel_codegen/sqlite"]
 
+[lib]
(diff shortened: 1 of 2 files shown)


sgrif:

diesel_tests/Cargo.toml
```diff
@@ -31,6 +31,10 @@ stable_sqlite = ["with-syntex", "sqlite", "diesel_codegen_syntex/sqlite"]
 unstable_postgres = ["unstable", "postgres", "diesel_codegen/postgres"]
 unstable_sqlite = ["unstable", "sqlite", "diesel_codegen/sqlite"]
 
+[lib]
+name = "integration_tests"
+path = "tests/lib.rs"
+
```
killercup: Huh? Why is this now necessary when it wasn't before?

sgrif: Oh whoops that wasn't supposed to get committed. I do that when I need to do `-Ztrace-macros`

sgrif:
I don't have a list, no. I usually add a FIXME comment if something is significantly worse off without a feature.

PR #504, opened 2016-12-01
update syn + quote

nerdrew:
I'm not sure on lots of this, but the tests pass. I had multiple versions of syn in one of my projects because of diesel. Figured I'd just update it. It was more involved than I expected.

### DIFF
--- a/diesel_codegen/src/associations.rs
+++ b/diesel_codegen/src/associations.rs
@@ -79,11 +79,9 @@ fn build_association_options(
                 _ => return usage_error,
             };
             let foreign_key_name = options.iter()
-                .filter_map(|o| {
-                    match o {
-                        &syn::NestedMetaItem::MetaItem(ref mi) => Some(mi),
-                        _ => None
-                    }
+                .filter_map(|o| match o {
+                    &syn::NestedMetaItem::MetaItem(ref mi) => Some(mi),
+                    _ => None
                 })
                 .find(|a| a.name() == "foreign_key")
                 .map(|a| str_value_of_meta_item(a, "foreign_key"))
--- a/diesel_codegen/src/associations.rs
+++ b/diesel_codegen/src/associations.rs
@@ -20,7 +20,7 @@ pub fn derive_associations(input: syn::MacroInput) -> quote::Tokens {
         }
     }
 
-    quote!(#(derived_associations)*)
+    quote!(#(#derived_associations)*)
 }
 
 fn expand_belongs_to(model: &Model, options: AssociationOptions) -> quote::Tokens {
@@ -39,7 +39,7 @@ fn expand_belongs_to(model: &Model, options: AssociationOptions) -> quote::Token
             foreign_key_name = #foreign_key_name,
             child_table_name = #child_table_name,
         ),
-        fields = [#(fields)*],
+        fields = [#(#fields)*],
     })
 }
 
@@ -56,7 +56,7 @@ fn expand_has_many(model: &Model, options: AssociationOptions) -> quote::Tokens 
             child_table = #child_table_name::table,
             foreign_key = #child_table_name::#foreign_key_name,
         ),
-        fields = [#(fields)*],
+        fields = [#(#fields)*],
     })
 }
 
@@ -75,10 +75,17 @@ fn build_association_options(
     match attr.value {
         syn::MetaItem::List(_, ref options) if options.len() >= 1 => {
             let association_name = match options[0] {
-                syn::MetaItem::Word(ref name) => name.clone(),
+                syn::NestedMetaItem::MetaItem(syn::MetaItem::Word(ref name)) => name.clone(),
                 _ => return usage_error,
             };
-            let foreign_key_name = options.iter().find(|a| a.name() == "foreign_key")
+            let foreign_key_name = options.iter()
+                .filter_map(|o| {
+                    match o {
+                        &syn::NestedMetaItem::MetaItem(ref mi) => Some(mi),
+                        _ => None
+                    }
+                })
+                .find(|a| a.name() == "foreign_key")
                 .map(|a| str_value_of_meta_item(a, "foreign_key"))
                 .map(syn::Ident::new);
 
--- a/diesel_codegen/src/ast_builder.rs
+++ b/diesel_codegen/src/ast_builder.rs
@@ -11,6 +11,6 @@ pub fn ty_path(path: Path) -> Ty {
 pub fn path_ident(ident: Ident) -> Path {
     Path {
         global: false,
-        segments: vec![PathSegment::ident(ident)],
+        segments: vec![PathSegment::from(ident)],
     }
 }
--- a/diesel_codegen/src/ast_builder.rs
+++ b/diesel_codegen/src/ast_builder.rs
@@ -1,16 +1,9 @@
-use syn::{Ident, Ty, Path, PathSegment};
+use syn::{Ident, Ty, Path};
 
 pub fn ty_ident(ident: Ident) -> Ty {
-    ty_path(path_ident(ident))
+    ty_path(ident.into())
 }
 
 pub fn ty_path(path: Path) -> Ty {
     Ty::Path(None, path)
-}
-
-pub fn path_ident(ident: Ident) -> Path {
-    Path {
-        global: false,
-        segments: vec![PathSegment::from(ident)],
-    }
 }
--- a/diesel_codegen/src/embed_migrations.rs
+++ b/diesel_codegen/src/embed_migrations.rs
@@ -15,8 +15,8 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
     }
 
     let options = get_options_from_input(&input.attrs, bug);
-    let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
-    let migrations_expr = migration_directory_from_given_path(migrations_path_opt.as_ref().map(|s| s.as_str()))
+    let migrations_path_opt = options.as_ref().map(|o| get_option(&o, "migrations_path", bug));
+    let migrations_expr = migration_directory_from_given_path(migrations_path_opt)
         .and_then(|path| migration_literals_from_path(&path));
     let migrations_expr = match migrations_expr {
         Ok(v) => v,
--- a/diesel_codegen/src/embed_migrations.rs
+++ b/diesel_codegen/src/embed_migrations.rs
@@ -16,7 +16,7 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
 
     let options = get_options_from_input(&input.attrs, bug);
     let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
-    let migrations_expr = migration_directory_from_given_path(migrations_path_opt)
+    let migrations_expr = migration_directory_from_given_path(migrations_path_opt.as_ref().map(|s| s.as_str()))
         .and_then(|path| migration_literals_from_path(&path));
     let migrations_expr = match migrations_expr {
         Ok(v) => v,
@@ -64,7 +64,7 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
         use self::diesel::connection::SimpleConnection;
         use std::io;
 
-        const ALL_MIGRATIONS: &'static [&'static Migration] = &[#(migrations_expr),*];
+        const ALL_MIGRATIONS: &'static [&'static Migration] = &[#(#migrations_expr),*];
 
         #embedded_migration_def
 
@@ -82,7 +82,7 @@ fn migration_literals_from_path(path: &Path) -> Result<Vec<quote::Tokens>, Box<E
 fn migration_literal_from_path(path: &Path) -> Result<quote::Tokens, Box<Error>> {
     let version = try!(version_from_path(path));
     let sql_file = path.join("up.sql");
-    let sql_file_path = sql_file.to_string_lossy();
+    let sql_file_path = sql_file.to_str();
 
     Ok(quote!(&EmbeddedMigration {
         version: #version,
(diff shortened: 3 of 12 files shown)


diesel_codegen/src/util.rs
```diff
@@ -101,22 +101,29 @@ pub fn inner_of_option_ty(ty: &Ty) -> Option<&Ty> {
...
     match options {
-        Some(&MetaItem::List(_, ref options)) => Some(options),
+        Some(&MetaItem::List(_, ref options)) => {
+            Some(options.iter().map(|o| {
+                match o {
+                   &NestedMetaItem::MetaItem(ref m) => m.clone(),
+                   _ => on_bug(),
+                }
+            }).collect())
+        }
```
killercup: This looks pretty complicated, cf. <https://github.com/diesel-rs/diesel/pull/508/files#r90534159>

diesel_tests/tests/expressions/ops.rs
```diff
@@ -91,6 +91,6 @@ fn mix_and_match_all_numeric_ops() {
         (3, 'Jim'), (4, 'Bob')").unwrap();
 
     let expected_data = vec![4, 6, 7, 9];
-    let data = users.select(id * 3 / 2 + 4 - 1).load(&connection);
+    let data = users.select(id * 3 / 2 + 4 - 1).order(id).load(&connection);
```
sgrif: We have a bunch of intermittent failures from needing an order added, would definitely prefer in a separate PR. It affects most of our tests. Actually making me contemplate default ordering.

diesel_codegen/src/schema_inference.rs
```diff
@@ -35,20 +35,20 @@ pub fn derive_infer_table_from_schema(input: syn::MacroInput) -> quote::Tokens {
...
-    let table_name = get_option(options, "table_name", bug);
+    let database_url = get_option(&options, "database_url", bug);
+    let table_name = get_option(&options, "table_name", bug);
 
-    let connection = establish_connection(database_url).unwrap();
-    let data = get_table_data(&connection, table_name).unwrap();
-    let primary_keys = get_primary_keys(&connection, table_name).unwrap()
+    let connection = establish_connection(database_url.as_str()).unwrap();
+    let data = get_table_data(&connection, table_name.as_str()).unwrap();
+    let primary_keys = get_primary_keys(&connection, table_name.as_str()).unwrap()
```
nerdrew: No reason. What would you prefer?

diesel_tests/tests/types_roundtrip.rs
```diff
@@ -138,12 +138,12 @@ pub fn mk_naive_date(days: u32) -> NaiveDate {
 #[cfg(feature = "postgres")]
 mod unstable_types {
     use super::*;
-    use std::time::*;
+    use std::time::{SystemTime, UNIX_EPOCH};
```
nerdrew: `std::time::Duration` conflicts with `super::Duration`.

diesel_tests/tests/expressions/ops.rs
nerdrew: Should I break out the ordering changes to the tests into a separate PR or do you want to fix it more holistically?

sgrif: If you want to break them out into a separate PR, that's fine. I want to evaluate the issue a bit more before I would merge it though. Either way it should be left out of this PR

diesel_codegen/src/util.rs
```diff
@@ -101,22 +101,29 @@ pub fn inner_of_option_ty(ty: &Ty) -> Option<&Ty> {
...
         Some(_) => on_bug(),
         None => None,
     }
 }
 
-pub fn get_option<'a>(
-    options: &'a [MetaItem],
+pub fn get_option(
+    options: &Vec<MetaItem>,
```
nerdrew: It isn't. I just couldn't get it to compile using references :(

diesel_tests/tests/expressions/ops.rs
sgrif: As a general rule of thumb we try to keep things small, focused, and easy to review. For example, #508, #509, and #510 could have reasonably been a single PR, but it helps the reviewer a ton to keep things more granular.

diesel_codegen/src/util.rs
nerdrew: I think this is the change that prompted the `&str` to `String` and `&[..]` to `Vec` changes.

`MetaItem::List` needs to be transformed into a list of `MetaItem`'s. I couldn't figure out how to return a slice of these. Ideas?

diesel_codegen/src/schema_inference.rs
sgrif: `&x` is preferred to `x.as_str()`

diesel_codegen/src/util.rs
sgrif: This still shouldn't need to affect the input type.

diesel_codegen/src/util.rs
sgrif: I could see the return types needing to change. Unclear on why the input type should have needed to change. There is no case where `&Vec<MetaItem>` wouldn't deref to `&[MetaItem]`

diesel_codegen/src/embed_migrations.rs
```diff
@@ -16,7 +16,7 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
 
     let options = get_options_from_input(&input.attrs, bug);
     let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
```
nerdrew: Is there a way to pass the lifetime of `options` to `get_option` and use that as the lifetime on the returned `str`? I couldn't get that to work, which is why I switch the return value of `get_option` to `String` (and all the associated changes `str` to `String`).

nerdrew:
General workflow question: do you prefer "fixup" commits and then squash before merging, or should I just amend this commit?

diesel_codegen/src/schema_inference.rs
nerdrew: Will fix.

sgrif:
Separate commits during code review. I will often request a squash after everything looks good

diesel_codegen/src/util.rs
nerdrew: Yup. Fixed.

diesel_tests/tests/expressions/ops.rs
nerdrew: Will remove the `order` changes. The rest seems to be atomic though, yes?

diesel_codegen/src/embed_migrations.rs
```diff
@@ -16,7 +16,7 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
...
-    let migrations_expr = migration_directory_from_given_path(migrations_path_opt)
+    let migrations_expr = migration_directory_from_given_path(migrations_path_opt.as_ref().map(|s| s.as_str()))
```
nerdrew: Is there a cleaner way to go from `Option<String>` to `Option<&str>`?

diesel_tests/tests/expressions/ops.rs
sgrif: Yes, I think so

sgrif:
Thanks for working on this. I have to go make dinner, but I will finish reviewing tomorrow

diesel_codegen/src/associations.rs
```diff
@@ -75,10 +75,17 @@ fn build_association_options(
...
                 _ => return usage_error,
             };
-            let foreign_key_name = options.iter().find(|a| a.name() == "foreign_key")
+            let foreign_key_name = options.iter()
+                .filter_map(|o| {
+                    match o {
+                        &syn::NestedMetaItem::MetaItem(ref mi) => Some(mi),
+                        _ => None
+                    }
+                })
```
killercup: You can make this a bit more concise by the way:

```rust
.filter_map(|o| match o {
    &syn::NestedMetaItem::MetaItem(ref mi) => Some(mi),
    _ => None
})
```

diesel_codegen/src/embed_migrations.rs
killercup: Not that I know of. I'd probably replace `.map(|s| s.as_str())` with `.map(String::as_ref)` though.

Edit: There even is a [SO question](http://stackoverflow.com/q/31233938/1254484) about that with basically the same answer.

diesel_codegen/src/embed_migrations.rs
killercup: I think the problem is that `.map` creates a reference that is only valid in its closure, but you want to return a value that lives as long as `option` does (and not the reference to option). Try this:

```diff
- let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
+ let migrations_path_opt = options.as_ref().map(|o| get_option(o, "migrations_path", bug));
```

diesel_codegen/src/associations.rs
nerdrew: Done

diesel_codegen/src/embed_migrations.rs
nerdrew: Done.

diesel_codegen/src/embed_migrations.rs
nerdrew: Good to know; unnecessary in this PR now though ;)

diesel_codegen/src/schema_inference.rs
```diff
@@ -35,20 +35,20 @@ pub fn derive_infer_table_from_schema(input: syn::MacroInput) -> quote::Tokens {
...
-    let table_name = get_option(options, "table_name", bug);
+    let database_url = get_option(&options, "database_url", bug);
+    let table_name = get_option(&options, "table_name", bug);
 
-    let connection = establish_connection(database_url).unwrap();
-    let data = get_table_data(&connection, table_name).unwrap();
-    let primary_keys = get_primary_keys(&connection, table_name).unwrap()
+    let connection = establish_connection(&database_url).unwrap();
+    let data = get_table_data(&connection, &table_name).unwrap();
+    let primary_keys = get_primary_keys(&connection, &table_name).unwrap()
```
killercup: Just for a fun exercise: I'm pretty sure the `&`s you added here are no longer necessary – the bindings are already references. (Deref coercion allows rustc to accept `&&&foo` where `&foo` was wanted.)

diesel_codegen/src/embed_migrations.rs
```diff
@@ -15,7 +15,7 @@ pub fn derive_embed_migrations(input: syn::MacroInput) -> quote::Tokens {
     }
 
     let options = get_options_from_input(&input.attrs, bug);
-    let migrations_path_opt = options.map(|o| get_option(&o, "migrations_path", bug));
+    let migrations_path_opt = options.as_ref().map(|o| get_option(&o, "migrations_path", bug));
```
killercup: Nice. I think you can just give it `o` instead of `&o` here (same as other comment).

diesel_codegen/src/ast_builder.rs
```diff
@@ -11,6 +11,6 @@ pub fn ty_path(path: Path) -> Ty {
 pub fn path_ident(ident: Ident) -> Path {
     Path {
         global: false,
-        segments: vec![PathSegment::ident(ident)],
+        segments: vec![PathSegment::from(ident)],
     }
```
killercup: FYI this whole function can be replaced with `ident.into()`

diesel_codegen/src/schema_inference.rs
```diff
@@ -62,7 +62,7 @@ fn column_def_tokens(
     let path_segments = column_type.path
         .into_iter()
         .map(syn::Ident::new)
-        .map(syn::PathSegment::ident)
+        .map(syn::PathSegment::from)
```
killercup: I added a bunch of `From` impls to syn a while back. I don't think we need the `.map(syn::Ident::new)` here anymore.

sgrif:
@killercup feel free to merge if you think it's good to go

killercup:
@nerdrew, can you also have a look at the conflict and the CI failure? If you don't have time, that's okay, I can do it tomorrow.

Travis says:

```
error: unused variable: `lifetimes`
  --> /home/travis/build/diesel-rs/diesel/diesel_codegen/src/identifiable.rs:10:9
   |
10 |     let lifetimes = model.generics.lifetimes;
   |         ^^^^^^^^^
   |
```

which surprises me because

0. it's not a bug in cargo/rustc (yay, CI works again!),
1. this line is not in `identifiable.rs:10` but in `insertable.rs:21` (maybe proc macros messing up the source of the error),
2. `lifetimes` is passed to the `_Insertable!` as `(#(#lifetimes),*)`,
3. and it compiles fine locally with my nightly from 2016-09-28.

You can find the code in its current form [here](https://github.com/diesel-rs/diesel/blob/9d90d6d70dd1dc295231963854a0593aa4f95d8b/diesel_codegen/src/insertable.rs#L21-L29). Maybe adding `#[allow(unused_assignments)]` (or similar, with a comment) as a temporary solution is fine.

sgrif:
I think the reason it's unused is that this branch was made before we had lifetimes there, but the file merged fine, so the "usage" is using the wrong syntax and therefore not actually being used.

diesel_codegen/src/schema_inference.rs
nerdrew: Just the `&`s for the `&str`s right? I still need them for the `options: Option<Vec<MetaItem>>` returned from `get_options_from_input`?

diesel_codegen/src/schema_inference.rs
nerdrew: Removed.

diesel_codegen/src/schema_inference.rs
nerdrew: TIL.

diesel_codegen/src/embed_migrations.rs
nerdrew: Done.

diesel_codegen/src/ast_builder.rs
nerdrew: The whole `path_ident` function? I.e. `path_ident(ident)` => `ident.into()`?

nerdrew: The method is public, does that matter in this case?

nerdrew:
@sgrif ya, those "used variable" errors are caused by `quote` changes. The errors are real. Fixing now.

sgrif:
Thanks!

PR #505, opened 2016-12-01
Remove the `FindDsl` constraint from `Identifiable`

sgrif:
Turns out we don't rely on it in as many places as we used to, and its
presence is making other refactorings difficult. I think I originally
added it to make the where clause on the impl of `SaveChangesDsl` less
gnarly, but it turns out it's more or less exactly the same anyway.


sgrif:
Review? @diesel-rs/contributors Not sure what's up with Cargo & SSL on CI, but the failure is unrelated

PR #506, opened 2016-12-01
Separate the `Table` type and function from `Identifiable`

sgrif:
I'm trying to figure out the best way to get composite primary keys to
work with the `Identifiable` trait. The issue is that for single primary
keys, you'll want to return `&'a i32` for example, but for composite
primary keys you'll want `(&'a i32, &'a i32)`. Restricting primary keys
to `Copy` types is not an option, so we have to assume references.
Without some form of HKT or associated type constructors (see
https://github.com/rust-lang/rfcs/pull/1598), we'll need to change the
return type of `Identifiable#id` from `&Self::Id` to `Self::Id` and
reference the lifetime in the trait impl.

I'm not yet sure whether that means we want the lifetime on the trait
itself, or if we want to implement the trait on references, but in both
cases we want this change. In the case of putting the lifetime on the
trait, there are a few cases where we just don't want to reason about
the lifetime because we just need the table. In the case of implementing
the trait on references, we will end up having places where `T::Table`
is ambiguous between `Identifiable` and `IntoUpdateTarget`, so unifying
them makes sense.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -3,7 +3,7 @@ use expression::helper_types::{Eq, EqAny};
 use expression::array_comparison::AsInExpression;
 use helper_types::{FindBy, Filter};
 use prelude::*;
-use super::Identifiable;
+use super::{Identifiable, HasTable};
 
 pub trait BelongsTo<Parent: Identifiable> {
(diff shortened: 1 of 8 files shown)


sgrif:

sgrif:
This was :+1:'d in gitter

PR #507, opened 2016-12-01
Refactor macros to always pass the "inner ty" of fields

sgrif:
Sometimes we need to differ in behavior based on whether we have an
option, but don't need the inner type. Sometimes we need to differ in
behavior, and we do need the inner type. I've been running into more of
the latter recently. The reason we do the `field_kind` dance in the
first place is because as soon as we match `Option<i32>` as a `:ty`, we
can no longer match on any of its individual tokens. Right now if we
want that inside type, we can't pass a field very far. This makes it
easier for future macros to do so, and future-proofs our existing macros
against additional information from Diesel's "shitty but probably good
enough struct parser"™


sgrif:

sgrif:
Script gone rogue. Reopening

PR #508, opened 2016-12-01
Refactor macros to always pass the "inner ty" of fields

sgrif:
Sometimes we need to differ in behavior based on whether we have an
option, but don't need the inner type. Sometimes we need to differ in
behavior, and we do need the inner type. I've been running into more of
the latter recently. The reason we do the `field_kind` dance in the
first place is because as soon as we match `Option<i32>` as a `:ty`, we
can no longer match on any of its individual tokens. Right now if we
want that inside type, we can't pass a field very far. This makes it
easier for future macros to do so, and future-proofs our existing macros
against additional information from Diesel's "shitty but probably good
enough struct parser"™

### DIFF
--- a/diesel/src/macros/as_changeset.rs
+++ b/diesel/src/macros/as_changeset.rs
@@ -173,6 +173,7 @@ macro_rules! _AsChangeset {
                 column_name: $column_name:ident,
                 field_ty: $field_ty:ty,
                 field_kind: $field_kind:ident,
+                $($rest:tt)*
             })+],
             struct_name = $struct_name:ident,
             $($headers:tt)*
@@ -196,6 +197,7 @@ macro_rules! _AsChangeset {
                 column_name: $column_name:ident,
                 field_ty: $field_ty:ty,
                 field_kind: $field_kind:ident,
+                $($rest:tt)*
             })+],
             struct_name = $struct_name:ident,
             $($headers:tt)*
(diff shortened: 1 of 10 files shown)


diesel_codegen/src/util.rs
```diff
@@ -84,6 +84,23 @@ pub fn is_option_ty(ty: &Ty) -> bool {
...
+
+    match *ty {
+        Ty::Path(_, Path { ref segments, .. }) =>
+            match segments[0].parameters {
+                AngleBracketed(ref data) => data.types.first(),
+                _ => None,
+            },
+        _ => None,
+    }
+}
```
killercup: This basically extracts the `T` from `std::option::Option<T>`, right? (Might be interesting to see if syn can offer a simpler API here.)

diesel/src/macros/as_changeset.rs
```diff
@@ -259,8 +261,10 @@ macro_rules! AsChangeset_construct_changeset_ty {
         fields = [{
             $(field_name: $field_name:ident,)*
             column_name: $column_name:ident,
-            field_ty: Option<$field_ty:ty>,
+            field_ty: $ignore:ty,
             field_kind: option,
+            inner_field_ty: $field_ty:ty,
+            $($rest:tt)*
```
killercup: FYI you commented above this

> FIXME: This can be cleaned up significantly when https://github.com/rust-lang/rust/issues/27245 is stable

Well, <https://github.com/rust-lang/rust/issues/27245> is stable now!

sgrif: I know! That's what https://github.com/diesel-rs/diesel/pull/494 was!

sgrif: I should merge it

PR #509, opened 2016-12-01
Separate the types of foreign and primary keys from belongs_to

sgrif:
As part of the attempt to allow composite primary keys in
`Identifiable`, we either need ATCs or reference the lifetime at the
trait impl level either by adding it to the `Identifiable` trait or by
implementing the trait on a reference. However, no matter what I did I
would always end up with a lifetime error in [`grouped_by`][grouped-by]
about the child being borrowed for too long.

I've finally realized that the reason this was occurring was that we
were [insisting that the foreign key be the same type as the primary
key][fk-type]. When we bring an additional lifetime into the picture,
what we were effectively doing is "downcasting" the foreign key to be of
the type of the primary key. This ended up implying that the child had
to live at least as long as the hash map containing the primary keys.
What we're doing now is effectively upcasting the primary key when we do
the comparison, allowing the foreign key to live for a short as it needs
to.

This still does not do the actual lifting of the reference of the
primary key type, that will come following this piece, assuming I don't
run into any more walls.

[grouped-by]: https://github.com/diesel-rs/diesel/blob/7d2c9fa1c11e870065d5cbb4f5cb91b51d7ed5fc/diesel/src/associations/belongs_to.rs#L30-L32
[fk-type]: https://github.com/diesel-rs/diesel/blob/7d2c9fa1c11e870065d5cbb4f5cb91b51d7ed5fc/diesel/src/associations/belongs_to.rs#L11

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -5,10 +5,14 @@ use helper_types::{FindBy, Filter};
 use prelude::*;
 use super::{Identifiable, HasTable};
 
-pub trait BelongsTo<Parent: Identifiable> {
+use std::borrow::Borrow;
+use std::hash::Hash;
+
+pub trait BelongsTo<Parent> {
+    type ForeignKey: Hash + ::std::cmp::Eq;
     type ForeignKeyColumn: Column;
 
-    fn foreign_key(&self) -> Option<&Parent::Id>;
+    fn foreign_key(&self) -> Option<&Self::ForeignKey>;
     fn foreign_key_column() -> Self::ForeignKeyColumn;
 }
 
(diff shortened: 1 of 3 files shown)


sgrif:

diesel/src/associations/belongs_to.rs
```diff
@@ -5,10 +5,14 @@ use helper_types::{FindBy, Filter};
...
-pub trait BelongsTo<Parent: Identifiable> {
+use std::borrow::Borrow;
+use std::hash::Hash;
+
+pub trait BelongsTo<Parent> {
+    type ForeignKey: Hash + ::std::cmp::Eq;
     type ForeignKeyColumn: Column;
 
-    fn foreign_key(&self) -> Option<&Parent::Id>;
+    fn foreign_key(&self) -> Option<&Self::ForeignKey>;
```
killercup: These two additions make so much sense. It's much clearer how this is used in grouped_by now.

sgrif:
> I'm not quite sure why you needed to explicitly state for<'a> &'a Parent::Id: Borrow<Child::ForeignKey> for that, but okay. :)

It's because the lifetime of `Parent` is unnamed, but I change that in the final commit in this series.

PR #510, opened 2016-12-02
Allow structs with composite primary keys to implement `Identifiable`

sgrif:
The fundamental problem is that right now we're assuming the primary key
is something that you can return a reference to. However, for composite
keys you'll likely want to return a tuple of references. In a perfect
world we would use an associated type constructor for this, but we don't
have those. Instead we're going through these contortions to make it
work.

These signatures are going to be significantly less clear to newcomers
now, and I'm not looking forward to answering "why am I seeing that
`User` doesn't implement `Identifiable` in gitter, but this is our only
option at the moment.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -16,17 +16,19 @@ pub trait BelongsTo<Parent> {
     fn foreign_key_column() -> Self::ForeignKeyColumn;
 }
 
-pub trait GroupedBy<Parent>: IntoIterator + Sized {
-    fn grouped_by(self, parents: &[Parent]) -> Vec<Vec<Self::Item>>;
+pub trait GroupedBy<'a, Parent>: IntoIterator + Sized {
+    fn grouped_by(self, parents: &'a [Parent]) -> Vec<Vec<Self::Item>>;
 }
 
-impl<Parent, Child, Iter> GroupedBy<Parent> for Iter where
+type Id<T> = <T as Identifiable>::Id;
+
+impl<'a, Parent: 'a, Child, Iter> GroupedBy<'a, Parent> for Iter where
     Iter: IntoIterator<Item=Child>,
     Child: BelongsTo<Parent>,
-    Parent: Identifiable,
-    for<'a> &'a Parent::Id: Borrow<Child::ForeignKey>,
+    &'a Parent: Identifiable,
+    Id<&'a Parent>: Borrow<Child::ForeignKey>,
 {
-    fn grouped_by(self, parents: &[Parent]) -> Vec<Vec<Child>> {
+    fn grouped_by(self, parents: &'a [Parent]) -> Vec<Vec<Child>> {
         use std::collections::HashMap;
 
         let id_indices: HashMap<_, _> = parents.iter().enumerate().map(|(i, u)| (u.id(), i)).collect();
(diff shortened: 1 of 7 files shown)


sgrif:
@killercup Last one. <3

diesel/src/macros/identifiable.rs
```diff
@@ -146,6 +147,7 @@ table! {
 fn derive_identifiable_on_simple_struct() {
     use associations::Identifiable;
 
+    #[allow(missing_debug_implementations, missing_copy_implementations)]
```
sgrif: I have no idea why this commit made these lints start failing here.

diesel/src/associations/belongs_to.rs
```diff
@@ -58,21 +60,21 @@ impl<'a, Parent, Child> BelongingToDsl<&'a Parent> for Child where
 }
 
 impl<'a, Parent, Child> BelongingToDsl<&'a [Parent]> for Child where
-    Parent: Identifiable,
+    &'a Parent: Identifiable,
     Child: HasTable + BelongsTo<Parent>,
-    Vec<&'a Parent::Id>: AsInExpression<<Child::ForeignKeyColumn as Expression>::SqlType>,
-    <Child as HasTable>::Table: FilterDsl<EqAny<Child::ForeignKeyColumn, Vec<&'a Parent::Id>>>,
+    Vec<Id<&'a Parent>>: AsInExpression<<Child::ForeignKeyColumn as Expression>::SqlType>,
+    <Child as HasTable>::Table: FilterDsl<EqAny<Child::ForeignKeyColumn, Vec<Id<&'a Parent>>>>,
```
killercup: Funny idea for teaching: Tell people how to read a clause like this as an English sentence. E.g.:

> A [reference to a] Parent is Identifiable, i.e. we can find the corresponding row.

> We know about Child that it has a table and BelongsTo Parent.

> The [List] of [references to] Parent Ids can be turned into an `IN` expression based on the SqlType of the Child's ForeignKeyColumn.

> The Child's Table can be Filtered [using our Dsl] to get the rows where the Child's ForeignKeyColumn is Eq to Any of the entries in the [List] of Parent Ids.

diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -32,15 +31,17 @@ impl<'a, T, ST, Conn> SaveChangesDsl<Conn, ST> for &'a T where
...
-    T: Identifiable,
-    T::Table: FindDsl<&'a T::Id, SqlType=ST>,
-    &'a T: AsChangeset<Target=T::Table> + IntoUpdateTarget<Table=T::Table>,
-    Update<&'a T, &'a T>: ExecuteDsl<SqliteConnection>,
-    Find<T::Table, &'a T::Id>: LoadDsl<SqliteConnection, SqlType=ST>,
+    T: Copy + Identifiable,
+    T: AsChangeset<Target=<T as HasTable>::Table> + IntoUpdateTarget,
+    T::Table: FindDsl<T::Id, SqlType=ST>,
+    Update<T, T>: ExecuteDsl<SqliteConnection>,
+    Find<T::Table, T::Id>: LoadDsl<SqliteConnection, SqlType=ST>,
```
killercup: That looks confusing, but I guess if it compiles it's fine (rustc is better at this than I am) 😄 What does it mean for T to be required to be Copy here? (Though I guess this was implicitly required before as well.)

sgrif:
> IIRC in the ATC RFC they were asking for concrete use cases (I haven't followed that discussion). Do you have a vision for how ATCs can make this API and the type signatures involved nicer?

With ATCs we would be able to remove pretty much every place that we are implementing traits for `&'a T` but not `T`. (Which is this trait, and every trait touched by this PR). Since we could just get the ID for a given lifetime at that point, life would be much easier.

PR #512, opened 2016-12-02
Add support for `NOT IN`

### LINKED ISSUE #511: `IN` condition in Postgres
Is there a way to create a query like the one below with the query builder?

```sql
SELECT id FROM foo WHERE id NOT IN (1, 3, 5);
```

sgrif:
I will probably eventually re-implement this with an arbitrary `not`
function, but I didn't want to go there quite yet as I'm not sure if
we'll need to do anything with parenthesis for that.

This will add another case we have to deal with empty arrays for, but I
don't think this will affect the implementation other than requiring an
additional test. What I'm thinking we'll do is have the `In` expression
write out `1=0` when it sees an empty array. With this implementation
it'll be `NOT 1=0`, which is exactly what we want. (In empty set = no
rows, not in empty set = all rows)

Fixes #511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## [0.8.2] - 2016-11-22
 
+### Added
+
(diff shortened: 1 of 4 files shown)


sgrif:

PR #515, opened 2016-12-02
Allow custom schema names in the `table!` macro

### LINKED ISSUE #447: Add custom schema support
As discussed on gitter, I want to use custom schemas with diesel. I am using postgresql. My use case is that the tables I want to query on aren't in the `public` schema. For example, I want to query `schema_1.table_1` and `schema_1.table_2`, however currently, `infer_schema!` doesn't find them.

sgrif:
This does not add support for `infer_schema!` with tables on schemas
other than public, but it does lay the ground work by changing the
macros to support it. I removed the `name` function from the `Table`
trait, as any code that was using that function would be broken on
tables with custom schemas.

Partially fixes #447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added support for SQL `NOT IN` using the `ne_any` method.
 
+* The `table!` macro now allows custom schemas to be specified. Example:
+
+  ```rust
+  table! {
+    schema_1.table_1 {
+      id -> Integer,
+    }
+  }
+  ```
+
+  The generated module will still be called `table_1`.
+
 ### Fixed
 
 * `#[derive(Identifiable)]` now works on structs with lifetimes
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -19,9 +19,10 @@ macro_rules! __diesel_column {
 
         impl<DB> $crate::query_builder::QueryFragment<DB> for $column_name where
             DB: $crate::backend::Backend,
+            <$($table)::* as QuerySource>::FromClause: QueryFragment<DB>,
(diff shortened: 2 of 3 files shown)


sgrif:

CHANGELOG.md
```diff
@@ -6,6 +6,20 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## [0.8.2] - 2016-11-22
 
+### Added
```
killercup: Did you mean to add this under a new "unreleased" h2?

diesel/src/macros/mod.rs
```diff
@@ -19,9 +19,10 @@ macro_rules! __diesel_column {
 
         impl<DB> $crate::query_builder::QueryFragment<DB> for $column_name where
             DB: $crate::backend::Backend,
+            <$($table)::* as QuerySource>::FromClause: QueryFragment<DB>,
         {
             fn to_sql(&self, out: &mut DB::QueryBuilder) -> $crate::query_builder::BuildQueryResult {
-                try!(out.push_identifier($($table)::*::name()));
+                try!($($table)::*.from_clause().to_sql(out));
```
killercup: Changing from just pushing the table name to using `QuerySource` is a nice refactoring.

diesel/src/macros/mod.rs
```diff
@@ -174,36 +175,44 @@ macro_rules! __diesel_column {
 #[macro_export]
 macro_rules! table {
     (
-        $name:ident {
-            $($column_name:ident -> $Type:ty,)+
+        $name:ident $body:tt
+    ) => {
+        table! {
+            public . $name (id) $body
```
killercup: ~~TIL~~ TIR (remembered): In macro token trees, whitespace around `.` and tokens don't matter (the same way they don't matter around `+`). Still reads weird though ;)

PR #516, opened 2016-12-03
Give a more helpful error on no-op updates

sgrif:
Prior to this commit, an update with no changes would generate invalid
SQL which would result in a database error. Even if the resulting error
were handled, the transaction would be aborted at that point, which may
not be desired. With this change, we provide a more useful error message
and don't risk aborting the current transaction.

### DIFF
--- a/diesel/src/query_builder/update_statement/mod.rs
+++ b/diesel/src/query_builder/update_statement/mod.rs
@@ -51,6 +51,11 @@ impl<T, U, V, DB> QueryFragment<DB> for UpdateStatement<T, U, V> where
     V: changeset::Changeset<DB>,
 {
     fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
+        if self.values.is_noop() {
+            return Err("There are no changes to save. This \
+                       query cannot be built".into())
+        }
+
         out.push_sql("UPDATE ");
         try!(self.table.from_clause().to_sql(out));
         out.push_sql(" SET ");
--- a/diesel_tests/tests/update.rs
+++ b/diesel_tests/tests/update.rs
@@ -1,3 +1,5 @@
+use std::error::Error;
+
 use schema::*;
 use diesel::*;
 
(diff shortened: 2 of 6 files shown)


sgrif:

diesel_tests/tests/update.rs
```diff
@@ -257,3 +259,22 @@ fn struct_with_option_fields_treated_as_null() {
...
+    }
+
+    let connection = connection();
+    let changes = Changes { name: None, hair_color: None, };
+    let updated_rows = update(users::table).set(&changes).execute(&connection);
+
+    match updated_rows {
+        Ok(_) => panic!("noop update was expected to return an error"),
+        Err(e) => assert!(e.description().starts_with("There are no changes to save.")),
+    }
```
killercup: You can replace this match with an `unwrap` and adding `#[should_panic(expected = "There are no changes to save.")]` to the fn

diesel_tests/tests/update.rs
```diff
@@ -1,3 +1,5 @@
+use std::error::Error;
```
killercup: You probably don't need this anymore then

diesel/src/query_builder/update_statement/mod.rs
```diff
@@ -51,6 +51,11 @@ impl<T, U, V, DB> QueryFragment<DB> for UpdateStatement<T, U, V> where
     V: changeset::Changeset<DB>,
 {
     fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
+        if self.values.is_noop() {
+            return Err("There are no changes to save. This \
+                       query cannot be built".into())
```
killercup: As this should be a cold path anyway (is `#[cold]` a thing?), you could create a more fancy error message like ```format!("There are no changes to be written to `{table}`. This query cannot be built.", table=self.table.from_clause())```

sgrif: Interesting idea. We'd need to run it through a query builder, but we have the debug query builder for that. I'm going to do it as a future refactoring.

sgrif:
CI failure is due to what appears to be a bug in the latest nightly

PR #518, opened 2016-12-05
Unify `InsertStatement` and `InsertQuery` into a single struct

sgrif:
I'm working on adding support for batch insert on SQLite, and fixing the
insert empty slice bug (#384). My approach is to override `ExecuteDsl`
and `LoadDsl` at the point where `InsertStatement` currently lives.
I expect that we'll have `into` perform some polymorphic dispatch to get
back either a `SingleInsertStatement` or `BatchInsertStatement`. If we
also have `InsertQuery` on top of that, the amount of redundant structs
and impls will get quite large. This change should make that refactoring
much easier.

I also took this opportunity to make sure our compile tests were up to
snuff around this, and made sure that the tests failed in the expected
fashion when I introduced some mistakes as part of this refactor.

### DIFF
--- a/diesel/src/query_builder/insert_statement.rs
+++ b/diesel/src/query_builder/insert_statement.rs
@@ -1,9 +1,10 @@
-use backend::{Backend, SupportsReturningClause};
+use backend::Backend;
 use expression::{Expression, SelectableExpression, NonAggregate};
 use persistable::{Insertable, InsertValues};
 use query_builder::*;
 use query_source::Table;
 use result::QueryResult;
+use super::returning_clause::*;
 
 /// The structure returned by [`insert`](fn.insert.html). The only thing that can be done with it
 /// is call `into`.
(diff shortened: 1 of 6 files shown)


sgrif:

diesel/src/query_builder/insert_statement.rs
```diff
@@ -73,23 +79,26 @@ impl<T, U, Op, DB> QueryFragment<DB> for InsertStatement<T, U, Op> where
...
     }
 }
 
+impl<T, U, Op, Ret> Query for InsertStatement<T, U, Op, ReturningClause<Ret>> where
+    Ret: Expression + SelectableExpression<T> + NonAggregate,
+{
+    type SqlType = Ret::SqlType;
+}
+
 impl<T, U, Op> InsertStatement<T, U, Op> {
```
killercup: This is the magic right here, isn't it? As this assumes the default type param for `Ret` it is only implemented for `NoReturningClause`, right? And thus not implemented once you called `returning`. Do you want to explicitly write `NoReturningClause` here and maybe add a short doc comment before the impl to explain this?

sgrif: Yeah, that's why this works. Though the reason I added the default type parameter is because `NoDefaultClause` is a voldemort type from outside this crate, and we have some test code that references `InsertStatement` that needed to keep working (at least temporarily). I'll probably add a default for `Op` as well

killercup: Yeah I understand using the default type param you don't need to expose it publicly. I was just asking to make it more explicit in this impl :)

sgrif: Yeah, I will in a few once I'm done with this other branch

PR #519, opened 2016-12-05
Support batch insert on SQLite, fix inserting empty slices

### LINKED ISSUE #384: Insert is crashing while trying to insert an empty vector

sgrif:
Those who watched my recent stream will notice that this has exactly
none of the code from that stream. Once I decided that empty errors
should be an error case, I realized that the "noop query that is
successful but should not execute" only applies to inserts, so I wanted
to take an approach that only affected inserts. I realized that the
plumbing required for batch inserts on SQLite would allow me to fix
this, since if we have a hook to execute N queries, N can be 0.

So this adds support for batch insert on SQLite. When `.into()` is
called, the return type will now either be `InsertStatement` or
`BatchInsertStatement`. The latter hooks into `ExecuteDsl` and `LoadDsl`
to do separate queries on SQLite.

While there should techincally be a separate queries implementation of
`LoadDsl` as well, we don't support any backend that supports returning
clauses but not the default keyword. I don't think such a backend
exists, so I've left out that code path as it would be untested.

We can probably partially revert this commit to leave the constraints on
`ExecuteDsl` and `LoadDsl` alone once specialization is stable, but the
addition of the second parameter on `ExecuteDsl` will still be required
unless [RFC 1672](https://github.com/rust-lang/rfcs/pull/1672) is
accepted.

The default type parameters on `IntoInsertStatement` are because
`NoReturningClause` is a voldemort type outside of the diesel crate, and
we're still needing to reference it in the `batch_insert` function in
our tests. That function can be removed with this feature, but it'll be
a large amount of churn that I want to commit separately. Once that's
done, the default types can be removed.

Fixes #384


sgrif:

diesel/src/query_builder/insert_statement.rs
```diff
@@ -148,6 +189,140 @@ impl<T, U, Op> InsertStatement<T, U, Op, NoReturningClause> {
...
+#[cfg(feature="sqlite")]
+impl<'a, T, U, Op, Ret> ExecuteDsl<::sqlite::SqliteConnection>
+    for BatchInsertStatement<T, &'a [U], Op, Ret> where
+        InsertStatement<T, &'a U, Op, Ret>: ExecuteDsl<::sqlite::SqliteConnection>,
+        T: Copy,
+        Op: Copy,
+        Ret: Copy,
+{
+    fn execute(self, conn: &::sqlite::SqliteConnection) -> QueryResult<usize> {
+        let mut result = 0;
```
sgrif: I would love to use `Iterator::sum` for this, but I don't see a good way to do it without first using `Iterator::collect` to put it into a `Result<Vec<T>>`

killercup: Well, you could do

```rust
let inserts = self.records.map(|r| InsertStatement::new(self.target, r, self.operator, self.returning).execute(conn));
let insert_sum = inserts
    .take_while(Result::is_ok)
    .map(Result::unwrap)
    .sum();

// next element is always an error or None
if let Some(Err(err)) = inserts.next() { Err(err) }
else { Ok(insert_sum) }
```

or something like… but it's quite long and not really more readable IMHO :/

diesel/src/query_builder/insert_statement.rs
```diff
@@ -24,26 +27,64 @@ impl<T, Op> IncompleteInsertStatement<T, Op> {
...
 
+impl<'a, T, Tab, Op, Ret> IntoInsertStatement<Tab, Op, Ret> for &'a Vec<T> {
+    type InsertStatement = <&'a [T] as IntoInsertStatement<Tab, Op, Ret>>::InsertStatement;
+
+    fn into_insert_statement(self, target: Tab, operator: Op, returning: Ret)
+        -> Self::InsertStatement
+    {
+        (&**self).into_insert_statement(target, operator, returning)
+    }
+}
```
killercup: Do you actually need to impl this for `&Vec` if you already have `&[]`? I mean, it doesn't hurt, but deref coercions should take care of that, right? Then again, you apparently need to write `&**self`?

diesel_tests/tests/insert.rs
```diff
@@ -64,16 +64,18 @@ fn insert_records_with_custom_returning_clause() {
 }
 
 #[test]
-#[cfg(not(feature = "sqlite"))]
```
killercup: Yay!

diesel_tests/tests/insert.rs
```diff
@@ -182,3 +184,30 @@ fn insert_on_conflict_replace() {
     let names = users.select(name).order(id).load::<String>(&connection);
     assert_eq!(Ok(expected_names), names);
 }
+
+#[test]
+fn insert_empty_slice() {
```
killercup: Yay²!

diesel/src/query_builder/insert_statement.rs
sgrif: Yeah I'm just going to submit a PR to make `sum` work out of the box

diesel/src/macros/insertable.rs
```diff
@@ -199,6 +199,24 @@ macro_rules! _Insertable {
                 ,)+)
             }
         }
+
+    } __diesel_parse_as_item! {
+        impl<$($lifetime,)* 'insert, Op, Ret> $crate::query_builder::insert_statement::IntoInsertStatement<$table_name::table, Op, Ret>
+            for &'insert $struct_ty where
+                $('insert: $lifetime,)*
+                $crate::query_builder::insert_statement::InsertStatement<$table_name::table, &'insert $struct_ty, Op, Ret>: $crate::query_builder::AsQuery,
```
sgrif: Doesn't `'insert: $lifetime` mean those lifetimes outlive `'insert` not the other way around? A reference can't outlive its data!

diesel/src/query_builder/insert_statement.rs
sgrif: Deref coercions don't apply to trait impls

diesel/src/query_builder/insert_statement.rs
```diff
@@ -148,6 +189,140 @@ impl<T, U, Op> InsertStatement<T, U, Op, NoReturningClause> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+/// The result of calling `insert(records).into(some_table)` when `records` is
+/// a slice or a `Vec`. When calling methods from `ExecuteDsl` or `LoadDsl`.
```
sgrif: I'm going to do a doc pass on the whole module before we ship. The `.returning` function is basically undiscoverable right now.

diesel/src/macros/insertable.rs
killercup: Uhm, no? `'a: 'b` means that `'a` lives at least as long as `'b`. IIRC, this not really intuitive. But [the nomicon](https://doc.rust-lang.org/nomicon/subtyping.html) and [Niko](http://smallcultfollowing.com/babysteps/blog/2016/05/09/non-lexical-lifetimes-adding-the-outlives-relation/#the-outlives-relation-in-light-of-control-flow) seem to agree with me.

sgrif: I have observed that the compiler often doesn't care if I do `$lifetime: 'insert` or `'insert: $lifetime` though.

PR #524, opened 2016-12-05
Update example tables to be compatible with CockroachDB

jethrogb:
This syntax is supported on both [Postgres](https://www.postgresql.org/docs/9.5/static/datatype-boolean.html) and [CockroachDB](https://www.cockroachlabs.com/docs/bool.html#format). I also tested it on SQLite and I'm assuming the client will handle the string `"FALSE"` appropriately.


killercup:
Thanks!

SQLite doesn't have these literals, but the examples are Postgres specific anyway. So I see no reason not do merge this!

PR #525, opened 2016-12-06
Unused import

jethrogb:


killercup:
LGTM. Cleanups are always welcome.

(I'm not sure if it helps you, though. `pg/connection/{mod,raw,result}.rs` and others, incl. most sqlite stuff, still use libc.)

PR #526, opened 2016-12-06
Don't panic when calling `eq_any` or `ne_any` with an empty array

### LINKED ISSUE #520: `eq_any(&[])` and `ne_any(&[])` generate invalid SQL

sgrif:
I had hoped to avoid introducing a new trait or methods for this.
However, the earliest point in the query builder that we can handle this
without introducing some new method is after the opening parenthesis has
been emitted. I had attempted to do something like emit `SELECT 1 WHERE
1=0` or `SELECT NULL WHERE 1=0` but both cases can cause type errors.

Still, due to the nature of how `In` is implemented, we can at least add
the required methods in an extremely contained fashion.

I have also added CHANGELOG entries for other "empty array causes
crashes" commits that had preceded this one, as I had forgotten to do so
earlier and this is the last case that I'm aware of.

Fixes #520.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,27 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
   The generated module will still be called `table_1`.
 
+* Added support for batch insert on SQLite. This means that you can now pass a
+  slice or vector to [`diesel::insert`][insert] on all backends.
+
+[insert]: http://docs.diesel.rs/diesel/fn.insert.html
+
 ### Fixed
 
 * `#[derive(Identifiable)]` now works on structs with lifetimes
+
+* Attempting to insert an empty slice will no longer panic. It does not execute
+  any queries, but the result will indicate that we successfully inserted 0
+  rows.
+
+* Attempting to update a record with no changes will no longer generate invalid
+  SQL. The result of attempting to execute the query will still be an error, but
+  but it will be a `Error::QueryBuilderError`, rather than a database error.
+  This means that it will not abort the current transaction, and can be handled
+  by applications.
+
+* Calling `eq_any` or `ne_any` with an empty array no longer panics.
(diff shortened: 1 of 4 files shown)


sgrif:
Review? @killercup

CHANGELOG.md
```diff
@@ -22,10 +22,27 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
...
+* Added support for batch insert on SQLite. This means that you can now pass a
+  slice or vector to [`diesel::insert`][insert] on all backends.
+
+[insert]: http://docs.diesel.rs/diesel/fn.insert.html
+
 ### Fixed
 
 * `#[derive(Identifiable)]` now works on structs with lifetimes
 
+* Attempting to insert an empty slice will no longer panic. It not execute any
```
killercup: s/It not/It does not

diesel/src/expression/array_comparison.rs
```diff
@@ -73,13 +73,17 @@ impl<T, U> NonAggregate for NotIn<T, U> where
...
-    U: QueryFragment<DB>,
+    U: QueryFragment<DB> + MaybeEmpty,
 {
     fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
-        try!(self.left.to_sql(out));
-        out.push_sql(" IN (");
-        try!(self.values.to_sql(out));
-        out.push_sql(")");
+        if self.values.is_empty() {
+            out.push_sql("1=0");
```
killercup: Maybe add a comment to say this is just to return a valid sql expression that always evals to false?

CHANGELOG.md
sgrif: Me not grammar

diesel/src/expression/array_comparison.rs
sgrif: That's what I wrote originally, but I felt like it was clearer to just actually say `1=0`

PR #527, opened 2016-12-06
Generate `foo NOT IN ...` not `NOT foo IN ...`

sgrif:
This means the code is a bit more copy-pasta, but not to a level that
concerns me. This approach felt better (and simpler) to me than unifying
these two AST nodes into something with an operator, since the operator
would need special methods for what to generate when empty.

In addition to generating `foo NOT IN` instead of `NOT foo IN`, we now
also generate `1=1` on an empty array, instead of `NOT 1=0`.

This change has no tests, as the two are functionally equivalent (in
fact, both SQLite and PG will generate identical query plans for both
forms). I prefer to avoid directly testing generated SQL unless it's
absolutely necessary.

### DIFF
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -13,7 +13,8 @@ pub struct In<T, U> {
 
 #[derive(Debug, Copy, Clone)]
 pub struct NotIn<T, U> {
-    in_expr: In<T, U>
+    left: T,
+    values: U,
 }
 
 pub type EqAny<T, U> = In<T, <U as AsInExpression<SqlTypeOf<T>>>::InExpression>;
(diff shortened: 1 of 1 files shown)


sgrif:

diesel/src/expression/array_comparison.rs
```diff
@@ -13,7 +13,8 @@ pub struct In<T, U> {
 
 #[derive(Debug, Copy, Clone)]
 pub struct NotIn<T, U> {
-    in_expr: In<T, U>
+    left: T,
```
killercup: I'm not sure if left is a good name for this but I don't have a better idea right now

sgrif: I think it originally was because this code was based on the general infix expression code where the members are called `left` and `right`. I agree I should come up with a better name for this context.

PR #528, opened 2016-12-06
Use the query DSL in `load_table_names` on PG

sgrif:
The `table!` macro allows custom schema names, so we no longer need to
use raw SQL for this query.

### DIFF
--- a/diesel_codegen_shared/src/schema_inference/pg.rs
+++ b/diesel_codegen_shared/src/schema_inference/pg.rs
@@ -46,6 +46,17 @@ table! {
     }
 }
 
+mod information_schema {
+    table! {
+        information_schema.tables (table_catalog, table_schema, table_name) {
+            table_catalog -> VarChar,
+            table_schema -> VarChar,
+            table_name -> VarChar,
+            table_type -> VarChar,
+        }
+    }
+}
+
 pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box<Error>> {
     let is_array = attr.type_name.starts_with("_");
     let tpe = if is_array {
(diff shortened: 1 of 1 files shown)


sgrif:

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -99,14 +110,12 @@ pub fn get_primary_keys(conn: &PgConnection, table_name: &str) -> QueryResult<Ve
...
-        .filter(sql::<types::Bool>("\
-            table_schema = 'public' AND \
-            table_name NOT LIKE '\\_\\_%' AND \
-            table_type LIKE 'BASE TABLE'\
-        "));
+    use self::information_schema::tables::dsl::*;
+
+    let query = tables.select(table_name)
+        .filter(table_schema.eq("public"))
+        .filter(table_name.not_like("\\_\\_%"))
```
killercup: What kind of escaping is this? The first of the slashes is for rust, the second for SQL? What is `\_` in SQL? Could this be written as `r"\_\_%"`?

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -46,6 +46,17 @@ table! {
     }
 }
 
+mod information_schema {
+    table! {
+        information_schema.tables (table_catalog, table_schema, table_name) {
```
killercup: Makes sense. (Though I have no idea what `table_catalog` is here 😄)

diesel_codegen_shared/src/schema_inference/pg.rs
weiznich: It's used to hide the `__diesel_schema_migrations` from the inferring method. Maybe this should be done more explicitly with `.filter(table_name.ne("__diesel_schema_migrations")`? Also the escaping of "_" should in theory happen inside of diesel.

sgrif: The `_` character in a SQL `LIKE` expression is equivalent to `.` in a regular expression. `\\_` is matching a literal `_` character in the resulting string.

killercup: Good to know! Might still read better as a raw string.

@weiznich I'm not sure I would expect diesel to escape this if you explicitly use `fn not_link(String)`. I found a method for stuff like `name.not_like("%a%%").escape('a')` though ([docs](http://docs.diesel.rs/diesel/expression/expression_methods/escape_expression_methods/trait.EscapeExpressionMethods.html)).

PR #529, opened 2016-12-06
Load custom schemas from `infer_schema!`

### LINKED ISSUE #447: Add custom schema support

sgrif:
This changes `infer_table_from_schema!` to optionally take a custom
schema in the form `schema_name.table_name`, and generate different code
accordingly. `infer_schema!` will now load the names of all custom
schemas, and generate the code accordingly. Any schemas other than the
default will be wrapped in a module.

Unresolved Questions
--------------------

On basically every backend other than PG, the term "schema" is
interchangeable with "database". We almost certainly never want to load
the schemas for other databases by default. Instead of automatically
picking up the schema names, should we instead require that the schema
name be passed to `infer_schema!`? That would mean it looks something
like:

```rust
infer_schema!("dotenv:DATABASE_URL", "custom_schema");
```

Fixes #447.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
   The generated module will still be called `table_1`.
 
+* The `infer_table_from_schema!` macro now allows custom schemas to be
+  specified. Example:
+
+  ```rust
+  infer_table_from_schema!("dotenv:DATABASE_URL", "schema_1.table_1");
+  ```
+
+* The `infer_schema!` optionally allows a schema name as the second argument. Any
+  schemas other than `public` will be wrapped in a module with the same name as
+  the schema. For example, `schema_1.table_1` would be referenced as
+  `schema_1::table_1`.
+
 * Added support for batch insert on SQLite. This means that you can now pass a
   slice or vector to [`diesel::insert`][insert] on all backends.
 
--- a/diesel_codegen/src/schema_inference.rs
+++ b/diesel_codegen/src/schema_inference.rs
@@ -3,7 +3,7 @@ use quote;
 
 use diesel_codegen_shared::*;
 
-use util::{get_options_from_input, get_option};
+use util::{get_options_from_input, get_option, get_optional_option};
 
 pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
     fn bug() -> ! {
@@ -13,19 +13,9 @@ pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
 
     let options = get_options_from_input(&input.attrs, bug).unwrap_or_else(|| bug());
     let database_url = get_option(&options, "database_url", bug);
+    let schema_name = get_optional_option(&options, "schema_name");
 
-    let table_names = load_table_names(&database_url).unwrap();
-    let schema_inferences = table_names.into_iter().map(|table_name| {
-        let mod_ident = syn::Ident::new(format!("infer_{}", table_name));
-        quote! {
-            mod #mod_ident {
-                infer_table_from_schema!(#database_url, #table_name);
-            }
-            pub use self::#mod_ident::*;
-        }
-    });
-
-    quote!(#(#schema_inferences)*)
+    infer_schema_for_schema_name(&database_url, schema_name.as_ref().map(|s| &**s))
 }
 
 pub fn derive_infer_table_from_schema(input: syn::MacroInput) -> quote::Tokens {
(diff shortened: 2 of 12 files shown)


sgrif:

CHANGELOG.md
```diff
@@ -22,6 +22,18 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
...
+  infer_table_from_schema!("dotenv:DATABASE_URL", "schema_1.table_1");
+  ```
+
+* The `infer_schema!` macro will now load tables from custom schemas. Any
+  schemas other than `public` will be wrapped in a module with the same name as
+  the schema. For example, `schema_1.table_1` would be referenced as
+  `schema_1::table_1`.
+
 * Added support for batch insert on SQLite. This means that you can now pass a
   slice or vector to [`diesel::insert`][insert] on all backends.
```
killercup: This link 404s. Should be <http://docs.diesel.rs/diesel/fn.insert.html>

diesel_codegen/src/schema_inference.rs
```diff
@@ -14,15 +14,12 @@ pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
...
-        quote! {
-            mod #mod_ident {
-                infer_table_from_schema!(#database_url, #table_name);
-            }
-            pub use self::#mod_ident::*;
-        }
+    let schema_names = load_schema_names(&database_url).unwrap()
+        .into_iter()
+        .map(Some)
+        .chain(vec![None]);
```
killercup: The last item is `None` to default to loading the `public` schema, right? Might be nice to add a comment :)

diesel_codegen/src/schema_inference.rs
```diff
@@ -53,6 +50,31 @@ pub fn derive_infer_table_from_schema(input: syn::MacroInput) -> quote::Tokens {
...
+        }
+    });
+
+    match schema_name {
+        Some(name) => {
+            let schema_ident = syn::Ident::new(name);
+            quote! { pub mod #schema_ident { #(schema_inferences)* } }
+        }
+        None => quote!(#(schema_inferences)*),
+    }
```
killercup: FTR/For the docs pass: For all schemas≠`public` this wraps the inferred tables etc in a `pub mod $schema`, but the inferred tables for `public` are inserted in the scope as-is. This may lead to collisions/uncompilable code with ugly error messages.

sgrif:
@killercup Do you have any thoughts on this approach vs the alternative mentioned?

diesel_codegen/src/schema_inference.rs
sgrif: Added it to the docs. I don't think we need to specifically call out the error case there, as I don't think having a table with the same name as a schema is a common thing to do, especially for tables in `public`. If the error did occur I think it's pretty self-explanatory what happened.

CHANGELOG.md
sgrif: Oh markdown is fun. I also had a link with the `[insert]` label in the release notes for 0.2.0, and that link is overriding this one.

sgrif: Fixed on master

killercup:
The alternative being writing the schema name explicitly? I'm okay with being explicit here. infer_schema does so much magic already! It might also be nice to write the schema names as an ident, so it's not surprising we generate a module with that name.

I'm a bit confused about the story with the other backends. Using PG, "schema" obviously means what PG calls "(custom) schema". In SQLite, it means "attached database", right? (What you get when you do `ATTACH DATABASE foo as foo" which lets you query it as `… FROM foo.table_name`, like a PG custom schema). I have no idea if MySQL has something like that.

> give a decent opt-in for MySQL if you actually needed to compile with multiple databases

What do you mean by "compile with multiple databases"? All three backends could, in theory, be used with multiple connections to multiple databases, couldn't they? (Assuming some trickery with build scripts maybe even using infer_schema.) What makes MySQL special here?

sgrif:
Unfortunately, if we want to be able to use macros 1.1, we can only take string literals as arguments.

sgrif:
> What makes MySQL special here?

It uses the term "schema" and "database" interchangeably. If you query `information_schema.tables` (which is ANSI standard by the way) on MySQL, tables from databases other than the one you're connected to would appear, and the database name appears where the schema name appears in PG.

killercup:
> and the database name appears where the schema name appears in PG

Damn, that *is* unfortunate. So that means implementing schema inference with custom schemas gives up… multiple dbs in MySQL? Crazy.

So, the options are to (a) not try to load all the custom schemas automatially or (b) special-case this for pg and tell people why that is. I say, go with option (a) and make people type their custom schemas' names so they know they will get them.

> Unfortunately, if we want to be able to use macros 1.1, we can only take string literals as arguments.

Can you do `($database_url:expr, $schema:indent)` and expand to `#[options(database_url=strigify!($schema))]` so users can only supply something that can become a module name?

sgrif:
> Can you do ($database_url:expr, $schema:indent) and expand to #[options(database_url=strigify!($schema))] so users can only supply something that can become a module name?

No, this is the same reason we had to add the `"env:ENV_VAR"` magic string form. Macros can't appear in this position.

compressed:
Thanks for working on this, really excited to see this!

I'm okay providing the schema names by hand, I don't see that as an issue. I agree that it will make it clearer this way also, especially if you have a database with tons of schemas, but only plan on using a few.

Just to clarify though, can we provide _multiple_ schemas, e.g. `infer_schema!("dotenv:DATABASE_URL", "custom_schema1, custom_schema2");`?

sgrif:
It'd just be one `infer_schema!` call per schema name

compressed:
Oh I see, that works too!

sgrif:
Updated

sgrif:
The CI failures are because syntex is trying to compile the PG specific module, even if we're using the SQLite feature. I'm just going to move it out of the set of modules evaluated by syntex and restrict that test to nightly only since we're quite close to removing syntex entirely.

PR #530, opened 2016-12-06
Rename all bang derive macros to `impl_`

### LINKED ISSUE #495: Diesel clobbers the macros from `diesel_codegen`
If you have

```rust
```

The macro definitions from diesel will completely clobber the ones from diesel codegen. We should probably just remove them (or rename them to impl_) at this point.

sgrif:
Since all macro types share a single namespace now, even just declaring
these macros can cause problems even if they're never used.

Fixes #495.


sgrif:

sgrif:
Should be green now

sgrif:
Looks like doctests for macros aren't running on the latest nightly

killercup:
Yay, all green!

PR #532, opened 2016-12-08
Allow custom primary keys to be specified for `Identifiable`

sgrif:
This is the first half of implementing `#[derive(Identifiable)]` with
composite primary keys. The second half will be to allow keys with more
than one name. The bulk of the code is actually for the "copy your
struct definition into this macro" code path, which we will probably get
rid of soon. Still, until macros 1.1 is actually stable, I want to keep
it around, which means we should keep feature parity.

### DIFF
--- a/diesel/src/macros/parse.rs
+++ b/diesel/src/macros/parse.rs
@@ -332,3 +332,67 @@ macro_rules! __diesel_field_with_column_name {
         )*
     };
 }
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __diesel_find_option_with_name {
+    // When no default is given, just a simple search will work
+    (
+        $headers:tt,
+        option_name = $target_option_name:ident,
+        args = [$(#[$option_name:ident $option_value:tt])*],
+        callback = $callback:ident,
+    ) => { $(
+        static_cond! {
+            if $target_option_name == $option_name {
+                $callback! {
+                    $headers,
+                    found_option_with_name = $target_option_name,
+                    value = $option_value,
+                }
+            }
+        }
+    )* };
+
+    // When a default is given, we need to recurse so we know when we're done
+    (
+        $headers:tt,
+        option_name = $target_option_name:ident,
+        args = [#[$option_name:ident $option_value:tt] $($rest:tt)*],
+        callback = $callback:ident,
+        default = $default:tt,
+    ) => {
+        static_cond! {
+            if $target_option_name == $option_name {
+                $callback! {
+                    $headers,
+                    found_option_with_name = $target_option_name,
+                    value = $option_value,
+                }
+            } else {
+                __diesel_find_option_with_name! {
+                    $headers,
+                    option_name = $target_option_name,
+                    args = [$($rest)*],
+                    callback = $callback,
+                    default = $default,
+                }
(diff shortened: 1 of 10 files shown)


diesel_codegen/src/model.rs
```diff
@@ -22,6 +23,9 @@ impl Model {
         let ty = struct_ty(item.ident.clone(), &item.generics);
         let name = item.ident.clone();
         let generics = item.generics.clone();
+        let primary_key_name = ident_value_of_attr_with_name(&item.attrs, "primary_key")
+            .map(Clone::clone)
+            .unwrap_or(syn::Ident::new("id"));
```
killercup: Can probably be `.unwrap_or_else(|| "id".into())`

diesel_codegen/src/model.rs
```diff
@@ -22,6 +23,9 @@ impl Model {
         let ty = struct_ty(item.ident.clone(), &item.generics);
         let name = item.ident.clone();
         let generics = item.generics.clone();
+        let primary_key_name = ident_value_of_attr_with_name(&item.attrs, "primary_key")
+            .map(Clone::clone)
```
killercup: `.cloned()`?

diesel_codegen_syntex/src/model.rs
```diff
@@ -25,13 +26,17 @@ impl Model {
         if let Annotatable::Item(ref item) = *annotatable {
             let table_name_from_annotation =
                 str_value_of_attr_with_name(cx, &item.attrs, "table_name");
+            let primary_key_name =
+                ident_value_of_attr_with_name(cx, &item.attrs, "primary_key")
+                    .unwrap_or(str_to_ident("id"));
```
killercup: I think we can get rid of str_to_ident with .into()

diesel/src/macros/parse.rs
```diff
@@ -332,3 +332,67 @@ macro_rules! __diesel_field_with_column_name {
...
+        callback = $callback:ident,
+        default = $default:tt,
+    ) => {
+        $callback! {
+            $headers,
+            found_option_with_name = $target_option_name,
+            value = $default,
+        }
+    };
+}
```
killercup: Very nice. I will be missing these macros when macros 1.1 is stable

diesel_tests/tests/annotations.rs
```diff
@@ -172,3 +172,25 @@ mod derive_identifiable_with_lifetime {
...
+    struct Foo<'a> {
+        id: i32,
+        foo_id: &'a str,
+        foo: i32,
+    }
+
+    let foo1 = Foo { id: 1, foo_id: "hi", foo: 2 };
+    let foo2 = Foo { id: 2, foo_id: "there", foo: 3 };
+    assert_eq!(&"hi", foo1.id());
+    assert_eq!(&"there", foo2.id());
```
killercup: I hope it's clear to people reading this in the future that these are `&&str`s

diesel_codegen_syntex/src/model.rs
sgrif: You're right on the syn side. libsyntax *definitely* doesn't have that impl though.

PR #534, opened 2016-12-08
Add a function for SQL `EXISTS` expressions.

### LINKED ISSUE #414: Add `exists` to expression DSL functions
This will allow the user to check for the existence of a record without loading it.

sgrif:
While this is useful for some cases where you don't want to load the
rows, this won't fill every use case for the expression, as right now
you wouldn't be able to build a query that references the outer table.
For us to do that and have it be type safe we'd need overlapping impls
for `SelectableExpression` (story of my life), which requires
https://github.com/rust-lang/rust/issues/29864 being implemented.

Fixes #414.

### DIFF
--- a/diesel/src/expression/exists.rs
+++ b/diesel/src/expression/exists.rs
@@ -1,0 +1,81 @@
+use backend::Backend;
+use expression::{Expression, SelectableExpression, NonAggregate};
+use query_builder::*;
+use result::QueryResult;
+use types::Bool;
+
+/// Creates a SQL `EXISTS` expression.
+///
+/// The argument must be a complete SQL query. The result of this could in
+/// theory be passed to `.filter`, but since the query cannot reference columns
+/// from the outer query, this is of limited usefulness.
+///
+/// # Example
+///
+/// ```rust
+/// # #[macro_use] extern crate diesel;
+/// # include!("src/doctest_setup.rs");
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+/// #
+/// # fn main() {
+/// #     use self::users::dsl::*;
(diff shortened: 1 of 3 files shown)


diesel/src/expression/exists.rs
```diff
@@ -0,0 +1,79 @@
+use backend::Backend;
+use expression::{Expression, SelectableExpression, NonAggregate};
+use query_builder::*;
+use result::QueryResult;
+use types::Bool;
+
+/// Creates a SQL `EXISTS` expression. The argument must be a complete SQL
```
killercup: Start new paragraph after first sentence for nice looking docs :)

diesel/src/expression/exists.rs
```diff
@@ -0,0 +1,79 @@
...
+/// query. The result of this could in theory be passed to `.filter`, but since
+/// the query cannot reference columns from the outer query, this is of limited
+/// usefulness.
+///
+/// # Example
+/// ```rust
```
killercup: Add newline after headline

diesel/src/expression/exists.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
+        out.push_sql("EXISTS (");
+        try!(self.0.to_sql(out));
+        out.push_sql(")");
+        Ok(())
+    }
+
+    fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()> {
+        try!(self.0.collect_binds(out));
+        Ok(())
```
killercup: Doesn't the inner `self.0.collect_binds(out)` suffice here without the try and Ok unit stuff?

killercup: Good point. I appreciate les diff noise :)

PR #538, opened 2016-12-09
Remove unused macro

killercup:
I was gonna write a nice issue to fix the FIXME in that code, but then saw that it wasn't used anymore anyway.


killercup:
Good point, will remove some more stuff!

killercup:
More stuff removed ✅

PR #541, opened 2016-12-10
Construct error messages properly when libpq returns `NULL` result

### LINKED ISSUE #446: Passing too much data results in a fatal error
The largest number of bind parameters a single query can contain is 34464 (1/2 of the value of a short, as it uses a signed integer for the length). Attempting to pass more than that will result in a fatal error, which gives no meaningful information back. However, since we know this case will always occur, we can check for it and bubble up that error.

On SQLite it's a bit more complex, as it's based on the SQLITE_MAX_VARIABLE_NUMBER variable, which is 999 by default, but can change at runtime. I'm fine with just assuming it's 999 for now.

sgrif:
I had originally thought that this was some weird condition that was the
result of the server getting a query with more than `34464` (`i16::MAX`)
bind parameters, where we got back an error but it had no message. We
panic in that case, as PG is documented that all errors have a message.
It turns out I was wrong for two reasons.

First, the number is 65535 (`u16::MAX`), not `34464`. (This might have
been something changed on the PG side in 9.6 which was released since I
last looked at this). Second, we were not getting back an error with no
message, we were getting a `NULL` return value from libpq. From [their
docs][]:

[their docs]: https://www.postgresql.org/docs/9.6/static/libpq-exec.html#LIBPQ-PQPREPARE

> A null result indicates out-of-memory or inability to send the command
> at all. Use PQerrorMessage to get more information about such errors.

`PQerrorMessage` means the error is on the connection. The reason that
we had no result was that the PG wire protocol uses a `u16` for the
number of bind parameters, so there's no way for the driver to have even
sent the message to the server.

Once we realized the error condition that we were seeing is more general
than just the bind parameter case, I've refactored the code to perform a
not null check as early as possible, and hide the `*mut PGresult`.

The test itself is written in a somewhat strange fashion. We're
constructing a query that has 70_000 bind parameters, as this is the
only case that I know of which will always cause libpq to return null. I
don't care what the error message is, just that it's not an empty
string. I've also written the test to explicitly match instead of unwrap
and use `#[should_panic]` because I want to be sure that it *didn't*
panic.

Ideally we would never see `*mut PGresult` anywhere in the code, and use
`Unique<PGresult>` instead. However, it is currently unstable. I had
originally implemented this to use `Unique` if `feature = "unstable"`
was set, but deleted it as the `#[cfg]` branches made the code harder to
comprehend. The only real benefit `Unique` gives us here is clarity of
intent. For the not-null and non-aliased hints to result in any
optimization by the compiler would require libpq to be statically
linked, and LTO to be happening at the LLVM layer, neither of which are
commonly true.

We know that `Send` and `Sync` are safe to implement on our new struct,
because the pointer is unaliased. We can delete those impls if we switch
to using `Unique` in the future, as it already implements those traits.

Fixes #446.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -41,7 +41,7 @@ impl SimpleConnection for PgConnection {
         let inner_result = unsafe {
             self.raw_connection.exec(query.as_ptr())
         };
-        try!(PgResult::new(inner_result));
+        try!(PgResult::new(inner_result?));
         Ok(())
     }
 }
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -57,8 +57,8 @@ impl RawConnection {
         }
     }
 
-    pub unsafe fn exec(&self, query: *const libc::c_char) -> *mut PGresult {
-        PQexec(self.internal_connection, query)
+    pub unsafe fn exec(&self, query: *const libc::c_char) -> QueryResult<RawResult> {
+        RawResult::new(PQexec(self.internal_connection, query), self)
     }
 
     pub unsafe fn exec_params(
@@ -70,8 +70,8 @@ impl RawConnection {
         param_lengths: *const libc::c_int,
         param_formats: *const libc::c_int,
         result_format: libc::c_int,
-    ) -> *mut PGresult {
-        PQexecParams(
+    ) -> QueryResult<RawResult> {
+        let ptr = PQexecParams(
             self.internal_connection,
             query,
             param_count,
@@ -80,7 +80,8 @@ impl RawConnection {
             param_lengths,
             param_formats,
             result_format,
-        )
+        );
+        RawResult::new(ptr, self)
     }
 
     pub unsafe fn exec_prepared(
@@ -91,8 +92,8 @@ impl RawConnection {
         param_lengths: *const libc::c_int,
         param_formats: *const libc::c_int,
         result_format: libc::c_int,
-    ) -> *mut PGresult {
-        PQexecPrepared(
+    ) -> QueryResult<RawResult> {
+        let ptr = PQexecPrepared(
             self.internal_connection,
             stmt_name,
             param_count,
@@ -100,7 +101,8 @@ impl RawConnection {
             param_lengths,
             param_formats,
             result_format,
-        )
+        );
+        RawResult::new(ptr, self)
     }
 
     pub unsafe fn prepare(
@@ -109,14 +111,15 @@ impl RawConnection {
         query: *const libc::c_char,
         param_count: libc::c_int,
         param_types: *const Oid,
-    ) -> *mut PGresult {
-        PQprepare(
+    ) -> QueryResult<RawResult> {
+        let ptr = PQprepare(
             self.internal_connection,
             stmt_name,
             query,
             param_count,
             param_types,
-        )
+        );
+        RawResult::new(ptr, self)
     }
 }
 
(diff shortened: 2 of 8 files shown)


diesel/src/pg/connection/raw.rs
```diff
@@ -167,3 +170,37 @@ impl Drop for PgString {
...
+///
+/// If `Unique` is ever stabilized, we should use it here.
+#[allow(missing_debug_implementations)]
+pub struct RawResult(*mut PGresult);
+
+unsafe impl Send for RawResult {}
+unsafe impl Sync for RawResult {}
+
+impl RawResult {
+    fn new(ptr: *mut PGresult, conn: &RawConnection) -> QueryResult<Self> {
```
killercup: Reading the code above, it took me a while to figure out that `RawResult::new` returned a `QueryResult`, because it really surprised me to see a `new` method that returns a `Result`.

I don't think this is very idiomatic, but it's also what `std::ffi::CString` does, so I guess it's okay. I would probably call this method `try_new`, or `try_form` myself (sadly, `std::convert::TryFrom` is still unstable).

diesel/src/pg/connection/result.rs
```diff
@@ -3,38 +3,39 @@ extern crate libc;
...
 
 pub struct PgResult {
-    internal_result: *mut PGresult,
+    internal_result: RawResult,
 }
 
 impl PgResult {
-    pub fn new(internal_result: *mut PGresult) -> QueryResult<Self> {
-        let result_status = unsafe { PQresultStatus(internal_result) };
+    pub fn new(internal_result: RawResult) -> QueryResult<Self> {
```
killercup: Huh, another `new` method that returns a result. Not as rare as I thought then. Carry on.

diesel_tests/tests/internal_details.rs
```diff
@@ -11,3 +11,22 @@ fn bind_params_are_passed_for_null_when_not_inserting() {
...
+fn query_which_cannot_be_transmitted_gives_proper_error_message() {
+    use schema::comments::dsl::*;
+    use diesel::result::Error::DatabaseError;
+    use diesel::result::DatabaseErrorKind::UnableToSendCommand;
+
+    let data: &[NewComment<'static>] = &[NewComment(1, "hi"); 35_000];
+    let query_with_to_many_binds = insert(data).into(comments);
+
+    match query_with_to_many_binds.execute(&connection()) {
+        Ok(_) => panic!("We successfully executed a query with 40k binds. \
```
killercup: > 40k binds

You "only" send an slice of 35k NewComments actually ;)

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -90,7 +90,7 @@ impl Query {
                 param_types_to_ptr(Some(&param_types)),
             )
         };
-        try!(PgResult::new(internal_result));
+        try!(PgResult::new(internal_result?));
```
killercup: same as above

diesel/src/pg/connection/mod.rs
```diff
@@ -41,7 +41,7 @@ impl SimpleConnection for PgConnection {
         let inner_result = unsafe {
             self.raw_connection.exec(query.as_ptr())
         };
-        try!(PgResult::new(inner_result));
+        try!(PgResult::new(inner_result?));
```
sgrif: Nah, I've been using it in a few places when it feels better than `try!`.

diesel_tests/tests/internal_details.rs
sgrif: And actually it's 70k binds. XD

PR #543, opened 2016-12-10
Fix broken documentation links

mujx:
The changes were tested locally with [linkchecker](https://github.com/wummel/linkchecker).


diesel/src/expression/mod.rs
```diff
@@ -1,14 +1,14 @@
 //! AST types representing various typed SQL expressions. Almost all types
-//! implement either [`Expression`](trait.Expression.html) or
-//! [`AsExpression`](trait.AsExpression.html).
+//! implement either [`Expression`](/diesel/expression/trait.Expression.html) or
```
sgrif: Can you change this to avoid absolute URLs? This will eventually break when we start hosting past versions.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -4,7 +4,7 @@ use query_source::QuerySource;
 use result::QueryResult;
 
 /// Types which can be passed to
-/// [`update.set`](struct.IncompleteUpdateStatement.html#method.set).
+/// [`update.set`](/diesel/query_builder/struct.IncompleteUpdateStatement.html#method.set).
```
sgrif: Same nit here on absolute links.

diesel/src/query_builder/update_statement/mod.rs
```diff
@@ -10,7 +10,7 @@ use query_builder::{Query, AsQuery, QueryFragment, QueryBuilder, BuildQueryResul
 use query_source::Table;
 use result::QueryResult;
 
-/// The type returned by [`update`](fn.update.html). The only thing you can do
+/// The type returned by [`update`](/diesel/fn.update.html). The only thing you can do
```
sgrif: Same nit here on absolute links. Although this definitely makes me think we need a way to say "root of the docs for the current version" easily.

sgrif:
I'm going to merge as-is because I'm going to release a point version and having working links would be nice. If you could follow up with another PR to fix the usage of absolute URLs that would be awesome though.

diesel/src/expression/mod.rs
mujx: The relative links don't work when the links appears on multiple pages with different roots. For example with the current setup `Expression` and `AsExpression` work [here](https://docs.rs/diesel/0.9.0/diesel/expression/index.html) but break [here](https://docs.rs/diesel/0.9.0/diesel/). The opposite will happen if the links were `expression/trait.Expression.html` and `expression/trait.AsExpression.html`.

So you are right that there is a need to refer to the root of the package, without hard coding it.

PR #547, opened 2016-12-19
Fix broken links

SShrike:
I was just browsing the code, and I noticed that a couple of the links in the `diesel_codegen` README seemed broken/missing. Hopefully I've pointed them towards the right things now :)


killercup:
Thank you!

We should probably have a check for that in CI (same with links in docs, we had a PR for that as well recently).

PR #548, opened 2016-12-19
Various things I worked on while on the plane

sgrif:
Normally I'd pull these into separate PRs but given that a few of these commits are dependent and I have limited access to internet and am trying to not work too much from vacation, they're all still together. Each commit message should have the context and reasoning you'd normally expect from my PRs.

The main goal that I'm working towards is unifying `InsertStatement` and `BatchInsertStatement` so that I have a smaller number of cases to worry about when introducing PG upsert, similar to how we unified `InsertStatement` and `InsertQuery` to make it easier to introduce batch insert for SQLite.


diesel/src/insertable.rs
```diff
@@ -36,33 +38,41 @@ pub enum ColumnInsertValue<Col, Expr> where
     Default(Col),
 }
 
+type ValuesFn<Item, T, DB> = fn(Item) -> <Item as Insertable<T, DB>>::Values;
```
killercup: Thank god for type aliases!

diesel/src/insertable.rs
```diff
@@ -36,33 +38,41 @@ pub enum ColumnInsertValue<Col, Expr> where
...
-    type Values = BatchInsertValues<<Iter::Item as Insertable<T, DB>>::Values>;
+    type Values = BatchInsertValues<iter::Map<
+        Iter::IntoIter,
+        ValuesFn<Iter::Item, T, DB>,
+    >>;
 
     fn values(self) -> Self::Values {
-        let values = self.into_iter().map(Insertable::values).collect();
+        let values = self.into_iter()
+            .map(Insertable::values as ValuesFn<Iter::Item, T, DB>);
```
killercup: Nice. (Here was a longer comment musing about fn pointers. Shouldn't have read this commit-by-commit.)

diesel/src/insertable.rs
```diff
@@ -36,33 +38,41 @@ pub enum ColumnInsertValue<Col, Expr> where
...
 
 impl<T, DB> InsertValues<DB> for BatchInsertValues<T> where
-    T: InsertValues<DB>,
+    T: Iterator + Clone,
+    T::Item: InsertValues<DB>,
     DB: Backend,
 {
     fn column_names(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
-        self.0[0].column_names(out)
+        self.0.clone().next().unwrap().column_names(out)
```
killercup: Clippy will not like this unwrap and I will probably make it a `….ok_or(WeirdStuffHappend).and_then(|val| val.column_names(out))` somewhere along the line ;)

llogiq:
I'm already proud :smile:

PR #549, opened 2016-12-21
Make travis like Sean's changes

killercup:
#548 + fixes for tests


killercup:
@sgrif The errors I mentioned in #548 seem to only turn up with recent nightlies. While 2016-12-15 compiled it fine, 2016-12-18 noticed the unused type alias.

I'll merge these fixes right now. Setting `#![allow(dead_code)]` in a macro-generated module seems okay to me. This could potentially also be an issue for users using `#![deny(warnings)]`. Maybe we can look into this more deeply at a later date.

sgrif:
I think that allow on the entire generated module is fine. Agreed that all generated code must always be warning free.

killercup:
Oh — would it make sense to add an `#[allow(warnings)]`? Does that even work?

sgrif:
Also thanks for fixing this. :)

PR #556, opened 2016-12-29
Make CI scripts nicer

killercup:
Just had the epiphany that bash scripts can become pretty readable if I
abuse `$@` enough times... So, of course I needed to try this out!

The idea is to make this nice to look at in travis (using a bit of
logging and hopefully also making travis group the scripts). Changing
the travis config to have multiple script entries will make this execute
all steps regardless of whether some failed. It does _not_ fail fast
anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,31 +12,19 @@ before_script:
   - pip install 'travis-cargo<0.2' --user
   - export PATH=$HOME/.local/bin:$PATH
   - psql -c 'create database diesel_schema;' -U postgres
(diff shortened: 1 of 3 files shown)


killercup:
Oh, travis-cargo doesnt seem to like generating and uploading docs using Cargo workspaces. Which is also a problem [on master](https://travis-ci.org/diesel-rs/diesel/jobs/185805033).

sgrif:
This has gotten pretty stale. Feel free to re-open if you still want to work on this

PR #564, opened 2017-01-02
Use DATETIME('now')^W^W^W CURRENT_TIMESTAMP instead of relying on implicit conversion

aperezdc:
In the `__diesel_schema_migrations` tables, the `run_on` column was defined as:

```sql
  run_on TIMESTAMP NOT NULL DEFAULT 'now'
```

Instead, this patch uses ~~`DATETIME('now')`~~ `CURRENT_TIMESTAMP`, which works both for SQLite and PostgreSQL by explicitly building a `TIMESTAMP` value, without relying on the the database making an implicit cast.

While PostgreSQL worked fine, the old definition did now work in SQLite as expected because it does not really store typed data in the columns (everything is  internally `TEXT`, the type being a hint on how to interpret it), and it was actually storing the `'now'` string value:

```
  % sqlite3 test.db
  sqlite> CREATE TABLE foo (v INTEGER, t DATETIME DEFAULT 'now');
  sqlite> INSERT INTO foo (v) VALUES (42);
  sqlite> SELECT * FROM foo;
  42|'now'
  sqlite>
```

On the other hand, with the updated code:

```
  sqlite> CREATE TABLE bar (v INTEGER, t DATETIME DEFAULT CURRENT_TIMESTAMP);
  sqlite> INSERT INTO bar (v) VALUES (10);
  sqlite> SELECT * FROM bar;
  10|2017-01-02 15:01:14
  sqlite>
```

### DIFF
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -164,7 +164,7 @@ fn create_schema_migrations_table_if_needed<Conn: Connection>(conn: &Conn) -> Qu
     conn.silence_notices(|| {
         conn.execute("CREATE TABLE IF NOT EXISTS __diesel_schema_migrations (
             version VARCHAR PRIMARY KEY NOT NULL,
-            run_on TIMESTAMP NOT NULL DEFAULT 'now'
+            run_on TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
         )")
     })
 }
--- a/diesel_cli/tests/migration_run.rs
+++ b/diesel_cli/tests/migration_run.rs
@@ -1,4 +1,7 @@
 use support::{database, project};
+use diesel::{select, LoadDsl};
+use diesel::expression::sql;
+use diesel::types::Bool;
 
 #[test]
 fn migration_run_runs_pending_migrations() {
@@ -24,4 +27,56 @@ fn migration_run_runs_pending_migrations() {
     assert!(result.stdout().contains("Running migration 12345"),
         "Unexpected stdout {}", result.stdout());
     assert!(db.table_exists("users"));
+}
+
+#[test]
+fn migration_run_inserts_run_on_timestamps() {
+    let p = project("migration_run_on_timestamps")
+        .folder("migrations")
+        .build();
+    let db = database(&p.database_url());
+
+    // Make sure the project is setup.
+    p.command("setup").run();
+
+    p.create_migration("12345_create_users_table",
+                       "CREATE TABLE users ( id INTEGER )",
+                       "DROP TABLE users");
+
+    let migrations_done: bool = select(sql::<Bool>(
+            "EXISTS (SELECT * FROM __diesel_schema_migrations)"))
+        .get_result(&db.conn())
+        .unwrap();
+    assert!(!migrations_done, "Migrations table should be empty");
+
+    let result = p.command("migration")
+        .arg("run")
+        .run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(db.table_exists("users"));
+
+    // By running a query that compares timestamps, we are also checking
+    // that the auto-inserted values for the "run_on" column are valid.
+
+    #[cfg(feature = "sqlite")]
+    fn valid_run_on_timestamp(db: &database::Database) -> bool {
+        select(sql::<Bool>("EXISTS (SELECT run_on < DATETIME('now', '-1 hour') \
+                                    FROM __diesel_schema_migrations)"))
+            .get_result(&db.conn())
+            .unwrap()
+    }
+
+    #[cfg(feature = "postgres")]
+    fn valid_run_on_timestamp(db: &database::Database) -> bool {
+        select(sql::<Bool>("EXISTS (SELECT \
+                              run_on < (CAST('now' AS TIMESTAMP) - \
+                                        CAST('1 hour' AS INTERVAL)) \
+                              FROM __diesel_schema_migrations)"))
+            .get_result(&db.conn())
+            .unwrap()
+    }
+
+    assert!(valid_run_on_timestamp(&db),
+            "Running a migration did not insert an updated run_on value");
 }


aperezdc:
Oops, the PostgreSQL running in Travis-CI are choking on `DATETIME('now')` when inserting values:

```
Creating database: diesel_getting_started_step_4
function datetime(unknown) does not exist
```

Another option is using `DEFAULT CURRENT_TIMESTAMP`, which [it is also supported in PostgreSQL](https://www.postgresql.org/docs/8.1/static/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT) — as I have just learnt. I'll push an update with that, and see how it goes. Hopefully it'll work just fine, otherwise we'll need to have different SQL for each database.

aperezdc:
There are two limitations with `CURRENT_TIMESTAMP`:

* Inside a transaction, it will yield always the same value: the timestamp at which the transaction was started.
*  In order to support SQLite, we cannot use the parameter to specify the number of decimal digits (e.g. `CURRENT_TIMESTAMP(2)` as supported by PostgreSQL), so subsecond timestamp resolutions are not possible.

My (very) limited understanding is that it should be fine, as casting `'now'` to a `TIMESTAMP` value was giving a seconds value already... Is it okay to use `CURRENT_TIMESTAMP` with Diesel given these limitations?

killercup:
Oh, good catch!

Can you add a test to see that this is actually a sane value? Like, add a test that runs a migration and check that `run_on` is set to something greater than 1 hour go. (Huh, I actually have no idea if/where we have tests for this behavior. If you can't find anything, let me know.)

killercup:
> Is it okay to use CURRENT_TIMESTAMP with Diesel given these limitations?

I think that is fine for the migrations table. If it turns out to be a problem wen can still add backend specific code later. Or, I don't know, add a `sleep(1)` somewhere ;)

aperezdc:
@killercup: Sure thing! I'll update with a test case. I'm at work right now but I'll totally going to give it a try later — also I am learning Rust in the process so please bear with me :innocent:

aperezdc:
Okay, I keep being tripped by PostgreSQL not having a `DATETIME()` function... the best cross-database snippet I have been able to come up with for the check is something like this for PostgreSQL:

```
foo=# SELECT t,v, CAST(t AS text) < CAST(current_timestamp AS text) FROM foo;
             t              |  v  | ?column? 
----------------------------+-----+---------
 2017-01-02 18:35:36.832241 | 100 | t
(1 row)
```

and the following for SQLite:

```
sqlite> SELECT t,v, CAST(t AS text) < CAST(current_timestamp AS text) FROM foo;
2017-01-02 17:20:43|100|1
```

It looks a bit brittle to do a `CAST` and then comparing strings lexicographically... which coincidentally works because both databases format the timestamps as ISO8601 (or alike), which have a fixed width and can be compared this way.

@killercup What do you think of using the query above in the test case? Is it okay, or is it too crappy and would it be better to use different queries depending on the database being used?

killercup:
Wait—why don't we specialize this for the specific backends? I would like to not try to be clever here and do what is best practice for the backend in question. Sorry for thinking about this earlier.

(I'm thinking of introducing a new `MigrationsSetup` trait (or similar) that is implemented for both `PgConnection` and `SqliteConnection` and contains a `fn setup_migration_table`, and replacing the implementation of `create_schema_migrations_table_if_needed` with a call to that function.)

aperezdc:
@killercup: That sounds very sensible, and certainly would be future-proof if later on the `__diesel_schema_migrations` table needs a more complex schema. One question, though: If both implementations of `Connection` are going to have a `fn setup_migration_table()`, wouldn't it make more sense to have the function directly in the `Connection` trait instead of adding a new trait? (For example `fn setup_helper_functions()` is there).

killercup:
Well, if setup_helper_functions is already there, sure. I was just thinking
of being a bit more explicit, as it is not really a concern for the
connection itself, but the backend. I'm all for keeping consistency though,
and we can always refactor this later.


Adrian Perez <notifications@github.com> schrieb am Di. 3. Jan. 2017 um
01:08:

sgrif:
I think using `CURRENT_TIMESTAMP` here. I'd prefer to avoid specializing this per-backend if possible. The constraints around transactions and lack of subsecond precision are acceptable for migrations.

Also I like your sense of humor with the title. :)

aperezdc:
@sgrif: Let's go with `CURRENT_TIMESTAMP`, then. I'll see if I can find a non-crappy way of writing the test case. Maybe one option is to write *two* test cases: one for PostgreSQL and anothe one for SQLite. How does that sound?

sgrif:
I'm fine with that.

aperezdc:
Funnily enough, running the PostgreSQL version of the query locally with `psql` works, trying to run “cargo test --no-default-features --features postgres` ends up in `SIGILL` (a panic while panicing), and in Travis-CI the test doesn't pass... It's a bit odd, if somebody has any idea, I would like to hear it. I'll keep trying to solve the issue with the test case anyway.

killercup:
On my machine with Postgres 9.6, the tests pass. Travis is [using](https://travis-ci.org/diesel-rs/diesel/jobs/190829768#L237) Postgres 9.4, though, and it fails with `syntax error at or near "SELECT"`. Is this a newly supported thing?

aperezdc:
@killercup: Mmmh, it could be. Thanks for the suggestion, I'll check the differences among versions from the PostgreSQL manuals. If that's the case, most likkely using explicit `CAST` expressions (`CAST('now' as TIMESTAMP)`) instead of the type names as type prefixes (`TIMESTAMP 'now'`) would work.

aperezdc:
and-it-works

sgrif:
Sorry I haven't been following this super closely, but what was the issue with `CURRENT_TIMESTAMP`? It is ANSII standard and should work on all backends.

aperezdc:
@sgrif: Using `CURRENT_TIMESTAMP` is working fine for both SQLite and PostgreSQL, and the same SQL snippet is used with both databases to create the `__diesel_schema_migrations` table.

The remaining issue writing the an unit test in such a way that the query won't return `true` when the inserted value for `run_on` has an invalid value. Syntax to work with timestamps/intervals and operate with them is different in both databases, and that's what I fixed with the last push :wink:

sgrif:
Ah, I see. Carry on then. :)

aperezdc:
As per @sgrif's comment, probably this can be merged now.

@killercup: WDYT?

killercup:
Very nice! Thanks again!

aperezdc:
You're welcome! Thanks for your patience with me, and for merging the fix :slightly_smiling_face:

PR #565, opened 2017-01-03
Scope 'options' derive attributes under trait name.

### LINKED ISSUE #558: Rename `options` decorator
Rocket uses `options` as a decorator to declare an HTTP `options` route. Diesel uses it to declare options on schemas. This causes a collision when the two libraries are used together, resulting in a compilation error.

It seems that the `options` decorator isn't used frequently by end-users in Diesel, while `options` in Rocket is one of the core decorators. Would you consider renaming `options` to `schema_options` or `table_options`, or something else, so that the two can coexist before macro renaming is allowed?

SergioBenitez:
Resolves #558.

### DIFF
--- a/diesel/src/macros/macros_from_codegen.rs
+++ b/diesel/src/macros/macros_from_codegen.rs
@@ -15,7 +15,7 @@ macro_rules! infer_schema {
     ($database_url: expr) => {
         mod __diesel_infer_schema {
             #[derive(InferSchema)]
-            #[options(database_url=$database_url)]
+            #[infer_schema_options(database_url=$database_url)]
             struct _Dummy;
         }
         pub use self::__diesel_infer_schema::*;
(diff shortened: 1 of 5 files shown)


sgrif:
I think we could change a bit less code if we use a single name like `diesel_codegen_options` instead.

SergioBenitez:
I found it confusing that different "options" parameters had the same name. I think having different names makes the code easier to read and grep later.

sgrif:
I honestly don't have that strong an opinion. Can you changes which are formatting-only? I will merge after that

SergioBenitez:
Sure. That's rustfmt, by the way.

sgrif:
It's not that I disagree with the changes, just that we prefer to avoid git-churn as much as possible.

SergioBenitez:
Failures are unrelated to these changes.

killercup:
Travis failures are from proc_macro becoming stable and deny(warnings). Will make a PR to remove the feature attrs later.

killercup:
@SergioBenitez can you rebase this on master to see if the tests pass? I'll merge it after that.

SergioBenitez:
Done.

killercup:
Sweet! Merging!

PR #567, opened 2017-01-04
Allow inserting `Option<T>` into columns where the type is not nullable

### LINKED ISSUE #554: Insertable should allow Option on non null fields with default
The `Insertable` trait currently does not allow fields that are specified as `NOT NULL` in the database but with a `DEFAULT` to have `Option` defined in the model. 

For example, this model here would fail when using a postgres backend even though the database will automatically fill in the field. 

```rust
#[table_name = "todos"]
#[derive(Queryable, Insertable)]
pub struct Todo {
    id: Option<i32>,
    pub description: String,
}
```

With the SQL definition of this

```sql
CREATE TABLE todos (
  id SERIAL PRIMARY KEY,
  description VARCHAR
);
```

The above SQL generates the following schema with the field being `NOT NULL` and having a `DEFAULT`.

```
                                                     Table "public.todos"
   Column    |       Type        |                     Modifiers                      | Storage  | Stats target | Description
-------------+-------------------+----------------------------------------------------+----------+--------------+-------------
 id          | integer           | not null default nextval('todos_id_seq'::regclass) | plain    |              |
 description | character varying |                                                    | extended |              |
Indexes:
    "todos_pkey" PRIMARY KEY, btree (id)
```

**This should work not only for PKs but also any field that has a default in the schema.**

sgrif:
This is important to allow structs with types like `Option<i32>` for
`id` to be used with `Insertable`, since the column has a default.
Ideally we would only allow this to work with columns that have a
default value. I may make that change in future, but at the moment we do
not have a way of knowing which columns have a default.

I am OK with the additional incorrect queries that this allows, as they
were already possible. There's no reason that `struct NewUser { name:
Option<String>, hair_color: Option<String>, }` shouldn't work if `struct
NewUser { hair_color: Option<String> }` works, as they both generate
incorrect queries (name is not null and has no default in this example).

I've documented the change as only affecting columns with a default, as
we reserve the right to disallow using `Option<T>` with columns that
have no default value in the future. Ideally before 1.0 I'd like to
replace this with a solution that doesn't leak the
`IntoNullable::SqlType` constraint in the public API.

Fixes #554.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * The `persistable` module has been renamed to `insertable`.
 
+### Fixed
+
+* `#[derive(Insertable)]` allows fields of type `Option<T>` to be used with
+  columns that are not null if they have a default value.
+
 ## [0.9.1] - 2016-12-09
 
 ### Fixed
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -40,3 +40,7 @@ pub type NotBetween<Lhs, Rhs> = super::predicates::NotBetween<Lhs,
 pub use super::predicates::{IsNull, IsNotNull, Asc, Desc};
 #[doc(inline)]
 pub use super::array_comparison::EqAny;
+
+#[doc(hidden)]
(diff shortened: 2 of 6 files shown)


sgrif:
Probably a marker trait like `HasDefault` for making `Option<T>` invalid and eventually some trait to represent if a given set of columns is a superset of the columns with no default. `Nullable` would go out of the picture for inserts since a nullable column is technically default null.

PR #569, opened 2017-01-09
Don't use libpq for identifier escaping

sgrif:
The implementation of `PQescapeIdentifier` actually does very little,
especially if we can assume that the connection's client encoding is
UTF-8, and that the identifier is valid UTF-8. Since I'd like to share
as much code between the current PG adapter, and the new async one, we
can just re-implement this function ourselves.

This was also the only place that required an `Rc` on the
`RawConnection`, which isn't a source of any major overhead but I'm glad
to see it go.

For some reason I thought that replace returned a `Cow<str>`, not
`String`. I may add our own version since this function could avoid
allocating the majority of the time.

### DIFF
--- a/diesel/src/pg/query_builder.rs
+++ b/diesel/src/pg/query_builder.rs
@@ -1,20 +1,15 @@
-use std::rc::Rc;
-
 use super::backend::Pg;
-use super::connection::raw::RawConnection;
 use query_builder::{QueryBuilder, BuildQueryResult};
 
 #[allow(missing_debug_implementations)]
 pub struct PgQueryBuilder {
-    conn: Rc<RawConnection>,
     pub sql: String,
     bind_idx: u32,
 }
 
 impl PgQueryBuilder {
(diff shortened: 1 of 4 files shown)


diesel/src/pg/query_builder.rs
```diff
@@ -27,8 +22,10 @@ impl QueryBuilder<Pg> for PgQueryBuilder {
     }
 
     fn push_identifier(&mut self, identifier: &str) -> BuildQueryResult {
-        let escaped_identifier = try!(self.conn.escape_identifier(identifier));
-        Ok(self.push_sql(&escaped_identifier))
+        self.push_sql("\"");
+        self.push_sql(&identifier.replace('"', "\"\""));
+        self.push_sql("\"");
```
killercup: (This broke the first meme generator I tested)

sgrif: `r#"""#` is more characters. ;)

PR #572, opened 2017-01-11
Clean up use of /bin/test in testing script.

### LINKED ISSUE #571: Test script at bin/test may not work correctly on some systems
The testing script uses `==` to compare strings, which is not supported by `/bin/test` on my system (Debian GNU/Linux unstable, running GNU coreutils 8.26). This should be an easy fix.

dstu:
This should fix #571.

### DIFF
--- a/bin/bench
+++ b/bin/bench
@@ -1,9 +1,9 @@
 #!/bin/sh
 set -e
 
-if [ "$1" == "sqlite" ]; then
+if [ "$1" = "sqlite" ]; then
   (cd diesel_tests && DATABASE_URL=/tmp/test.db cargo bench --no-default-features --features "unstable sqlite")
-elif [ "$1" == "postgres" ]; then
+elif [ "$1" = "postgres" ]; then
   (cd diesel_tests && cargo bench --no-default-features --features "unstable postgres")
 else
   (cd diesel_tests && DATABASE_URL=/tmp/test.db cargo bench --no-default-features --features "unstable sqlite")
--- a/bin/test
+++ b/bin/test
@@ -1,11 +1,11 @@
 #!/bin/sh
 set -e
 
-if [ "$1" == "integration" ] && [ "$2" == "sqlite" ]; then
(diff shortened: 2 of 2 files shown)


killercup:
Oh, interesting, I didn't know that `=` was the POSIX version. Do you know if there are any differences between the two, and if they could be problematic?

dstu:
I can't find a reference to `/bin/test` that uses `==` for string comparison.m, so I see no reason to use `==` at all. What platform are you on? Could built-in shell behavior that overrides calling `test` be at fault? This might be some version of a "/bin/sh isn't actually sh, it's Bash or some other shell, and using an idiom from that shell breaks for people who actually use sh."

For reference: 

* [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html): `s1 = s2`
* [GNU test](https://linux.die.net/man/1/test): `STRING1 = STRING2`
* [Apple's "legacy" doc for an old BSD test](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/test.1.html): `s1 = s2`
* [FreeBSD](https://www.freebsd.org/cgi/man.cgi?test): `s1 = s2`

I note that my review of the subject shows that the way I pass more than 4 arguments to `test` is itself [not strictly correct](http://unix.stackexchange.com/questions/57840/posix-test-and-a), so I'll revert that change when I'm able to.

killercup:
Nice research, TIL! Locally, I'm probably using Apple's BSD adaption, though I have gnu coreutils installed as well. I'd say this is a "sh is assumed to be bash" thing as well, since this is what it defaults to on most machine, incl. mine and Travis CI (even though I'm an avid fish user).

I'll merge this as soon as you pushed the `&&` changes :)

dstu:
This should be set now. Sorry about the merge commit. I'm not certain of a good way to get rid of it without opening a new PR. If there's something you'd like me to do about it, let me know.

killercup:
Thanks for sticking with this, @dstu! The commit churn is no problem, Github allows to squash before merging :)

Issue #562, opened 2016-12-31
Implementing `diesel::Expression`, etc., for custom types outside diesel?


emk:
As a followup to #561, I'd like to try defining a custom `ToSql` and `FromSql` implementation for a type in a crate besides `diesel`. I'm going to try this with https://github.com/faradayio/bigml-rs, which a client for the commercial BigML machine learning service.

This crate contains a type `Id`, which represents a string of the form `source/1234` or `execution/5678`. Because these strings refer to different types of resources, the `Id` type takes parameters: `Id<Source>` or `Id<Execution>`. So far, no problems.

```rust
#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
pub struct Id<R: Resource> {
    /// The ID of the resource.
    id: String,
    /// A special 0-byte field which exists just to mention the type `R`
    /// inside the struct, and thus avoid compiler errors about unused type
    /// parameters.
    _phantom: PhantomData<R>,
}
```

I can declare `ToSql` and `FromSql` instances as follows:

```rust
    impl<R: Resource, DB: Backend<RawValue=[u8]>> FromSql<types::Text, DB> for Id<R> {
        fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
            (FromSql::<types::Text, DB>::from_sql(bytes) as Result<String, _>)
                .and_then(|s| {
                    Id::from_str(&s).map_err(|e| format!("{}", e).into())
                })
        }
    }

    impl<R: Resource, DB: Backend> ToSql<types::Text, DB> for Id<R>
        where for<'a> &'a str: ToSql<types::Text, DB>
    {
        fn to_sql<W: Write>(&self, out: &mut W)
                            -> Result<IsNull, Box<Error+Send+Sync>> {
            self.as_str().to_sql(out)
        }
    }
```

But if I try to use these impls, I run into problems. This code:

```
    table! {
        example_structs {
            id -> Integer,
            source_id -> Text,
        }
    }

    // This should compile without errors.
    #[derive(Queryable, Insertable)]
    #[table_name="example_structs"]
    struct ExampleStruct {
        id: i32,
        source_id: Id<Source>,
    }
```

...gives me the errors:

```
error[E0277]: the trait bound `resource::id::Id<resource::source::Source>: diesel::Expression` is not satisfied
   --> src/resource/id.rs:170:5
    |
170 |     #[derive(Queryable, Insertable)]
[...]

PR #574, opened 2017-01-12
Make macros public to allow easily implementing traits outside of diesel

aperezdc:
This makes the `expression_impls!`, `queryable_impls!`, and `primitive_impls!` macros public, so programs (or other crates) can more easily implement the `Expression`, `Queriable`, and `QueryId` traits without duplicating the code from the macros.

These may change before diesel 1.0, so as discussed in #562, they are hidden from the documentation for now.

### DIFF
--- a/diesel/src/types/impls/mod.rs
+++ b/diesel/src/types/impls/mod.rs
@@ -9,6 +9,8 @@ macro_rules! not_none {
     }
 }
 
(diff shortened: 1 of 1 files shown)


sgrif:
Thanks!

aperezdc:
Yay! Thanks for merging :smiley:

PR #577, opened 2017-01-13
Initial version of infer-schema as cli subcommand

### LINKED ISSUE #437: CLI: Add a subcommand to print db schema as `table!` calls

weiznich:
* Factor infer schema related code into own crate
* Implement cli subcommand

Fix #437 

The current version should be seen as proof of concept to start discussion about the implementation. I think before merging

- [x] CLI integration tests
- [x] support for non default schemas
- [x] filtering tables

should be added


Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     # FIXME: getting "multiple matching crates for `diesel`" because of the shared build directory
     # "diesel_compile_tests",
     "diesel_tests",
+    "diesel_infer_schema",
```
killercup: Moving this into a separate, public crate is interesting. I'm not sure I agree, but I don't see many reasons against it.

diesel_cli/src/cli.rs
```diff
@@ -62,6 +62,10 @@ pub fn build_cli() -> App<'static, 'static> {
     let generate_bash_completion_subcommand = SubCommand::with_name("bash-completion")
         .about("Generate bash completion script for the diesel command.");
 
+    let infer_schema_subcommand = SubCommand::with_name("infer-schema")
+        .setting(AppSettings::VersionlessSubcommands)
+        .about("Print table definitions for database schema");
```
killercup: Nit: All other descriptions end with a period

diesel_cli/src/main.rs
```diff
@@ -186,6 +188,59 @@ fn convert_absolute_path_to_relative(target_path: &Path, mut current_path: &Path
...
+    let mut skip_space = false;
+    let mut last_char = ' ';
+    
+
+    for c in schema.as_str().chars() {
+        if c == '!' || c == '(' || c == ';' || c == ','  || c == '<'
+            || (c == ':' && last_char != ':') ||
+            (c == '>' && last_char != '-') {
+            out.pop();
+        }
```
killercup: Whew, this looks complicated. Can add a comment describing what is happening here? Is it possible to rewrite this whole block as an iterator chain that might be easier to read?

diesel_codegen/src/schema_inference.rs
```diff
@@ -1,7 +1,8 @@
 use syn;
```
killercup: Will continue here after lunch

diesel_cli/src/cli.rs
```diff
@@ -62,6 +62,10 @@ pub fn build_cli() -> App<'static, 'static> {
     let generate_bash_completion_subcommand = SubCommand::with_name("bash-completion")
         .about("Generate bash completion script for the diesel command.");
 
+    let infer_schema_subcommand = SubCommand::with_name("infer-schema")
```
killercup: I think I'd call the subcommand "print-schema" instead.

diesel_cli/src/main.rs
```diff
@@ -186,6 +188,59 @@ fn convert_absolute_path_to_relative(target_path: &Path, mut current_path: &Path
...
+    let database_url = database::database_url(matches);
+    let schema_name = None;
+    let schema = diesel_infer_schema::infer_schema_for_schema_name(&database_url, schema_name);
+    let mut out = String::new();
+    let mut indent = String::new();
+    let mut skip_space = false;
+    let mut last_char = ' ';
+    
+
+    for c in schema.as_str().chars() {
```
killercup: Okay, IIUC this whole for-loop is to pretty print the tokens, right? Great job! In my tests you got about 80% of the formatting right in just 40 lines of code! Sadly it will look weird unless we get to 99% :)

Nevertheless, I'd either (a) put this into a separate function that takes `quote::Token` and a `W: io::Write` and make that _perfect_ for using with our `table!` macro, or (b) skip this completely and use a proven tool we don't need to maintain! :)

syn has a "pretty" feature, but that seems to use libsyntax' pretty printing… which is not that great. Can we maybe pipe our string output into rustfmt? That way it might even respect the users' project's rustfmt.toml.

Maybe @dtolnay, the great author of quote and syn, has some suggestions? :)

diesel_cli/src/main.rs
```diff
@@ -186,6 +188,59 @@ fn convert_absolute_path_to_relative(target_path: &Path, mut current_path: &Path
     result.join(target_path.strip_prefix(current_path).unwrap())
 }
 
+fn run_infer_schema(matches: &ArgMatches) {
+    let database_url = database::database_url(matches);
+    let schema_name = None;
+    let schema = diesel_infer_schema::infer_schema_for_schema_name(&database_url, schema_name);
```
killercup: I assume most of the code in that function and its call stack are basically what there were as part of diesel_codegen? Of not, can you point me to important differences?

Also, we should think about error handling here. The codegen code contains a lot of `unwrap` (and the diff suggests you added some more :)), but I think it won't be too difficult to move it to `Result`. Having done that, I'd really like for the function called here to return an `Iterator<Item=(TableName, Result<Tokens, _>)>` so we can emit as many `table!` definitions as possible and not crash at the first error. This doesn't need to be in the first version though.

diesel_infer_schema/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "diesel_infer_schema"
+version = "0.1.0"
```
killercup: Memo to @diesel-rs/core: Need to publish this crate.

FYI, its API surface is quite small. Just two function with the signature `(database_url: &str, table_name: &str) -> Tokens`.

diesel_cli/Cargo.toml
```diff
@@ -17,15 +17,16 @@ chrono = "0.2.17"
 clap = "2.11"
 diesel = { version = "0.9.0", default-features = false }
 dotenv = "0.8.0"
+diesel_infer_schema = { version = "0.1.0", default-features = false, path = "../diesel_infer_schema" }
```
killercup: Can you skip the path and instead use an override in the workspace Cargo.toml?

diesel_cli/src/main.rs
dtolnay: Just like you, I don't want to be in the business of maintaining a pretty-printer either :smiling_imp:. If you care about formatting, run it through [`rustfmt::run`](https://docs.rs/rustfmt/0.6.3/rustfmt/fn.run.html) which writes formatted code to stdout. Or the approach I took in cargo-expand which is check whether rustfmt is available on the path and pipe to it if found.

killercup: Yeah, pretty printing is hard and full of people telling you about their preferred code style 🙈 

I'm actually fine with shipping a version that depends on having rustfmt in `$PATH` and just piping to it. @weiznich, what do you think?

Cargo.toml
weiznich: diesel_codegen_syntex and diesel_codegen_shared are dead, as stated in #568. It is not possible to use a proc-macro lib as depenency without using the proc-macro version. Therefore the infer schema related code needs to live somewhere else. A new crate is in my opinion a clearer solution, as putting it into diesel itself.

diesel_cli/src/main.rs
weiznich: I've implement the error managment in a other way to have all infer_schema related code in the corresponding crate. Otherwise the wrapping with the schema module would be done seperatly in codegen and cli. So everything is shared between codegen and cli. The only difference is what happens in case of an error. Codegen panics, cli will print a message and then tries to infer the other tables.

diesel_cli/src/main.rs
weiznich: The problem with rustfmt is that it fails on formating the macros.
Running rustfmt on the output of `quote::Token::as_str()` will produce something like

```
mod infer_users {
    table ! { users ( id ) { id -> :: diesel :: types :: Int4 , username -> :: diesel :: types :: Varchar , password -> :: diesel :: types :: Varchar , access_level -> :: diesel :: types :: Int2 , } }
}
;
```

In my opinion this is far from optimal. (There seems to be a rustfmt [issue](https://github.com/rust-lang-nursery/rustfmt/issues/438)) 

If rustfmt would work on the generated code, running rustfmt would be the better solution. So I'm not realy sure how to handle this.

diesel_cli/Cargo.toml
weiznich: As stated in [here](http://doc.crates.io/specifying-dependencies.html#overriding-with-local-dependencies) this is not possible as long as the crate is not published to crates.io

> Note: using a local configuration to override paths will only work for crates that have been published to crates.io. You cannot use this feature to tell Cargo how to find local unpublished crates.

Cargo.toml
killercup: > It is not possible to use a proc-macro lib as depenency without using the proc-macro version. 

Oh, right. This seems like the best solution, then!

diesel_cli/Cargo.toml
killercup: TIL! Thanks.

killercup:
Moving this to a top-level comment so it doesn't get lost. In reply to <https://github.com/diesel-rs/diesel/pull/577#discussion_r96125450>:

Orrrr, rustfmt! 😞 Well, I guess we need to define our own formatting, then. You encoded some rules implicitly. Can you document them? I'd start with:

1. Seeing `{` increases indentation level
2. Seeing `}` decreases indentation level
3. Insert newline after `{`, `}`, `,`, and `;`
4. Don't put spaces:
    - between ident and `!`,
    - between path segments and `::`
    - after `(` and before `)`
    - before `,`
5. Insert spaces
    - around arrows (`->`, and probably also `=>`)
    - after keywords like `mod`, and `use`

I'd really like to put this whole formatting business into its own module, encode the rules as well documented as possible and add a bunch of tests.

diesel_cli/src/pretty_printing.rs
```diff
@@ -0,0 +1,194 @@
...
+            '{' => {
+                skip_space = true;
+                out += "\n";
+                indent += "\t";
+                out += &indent;
+            }
+            _=>{}
+        }        
+    }
+    write!(&mut output, "{}", out).unwrap();
```
killercup: Can you make this function return `std::io::Result<()>` and omit the unwrap?

And now for something really profound: Tabs vs. spaces. I see you used `\t` for indentation, probably because it's exactly one character. Idiomatic Rust code uses four spaces, though. Can you add a `.replace("\t", "    ")` here?

diesel_cli/src/pretty_printing.rs
```diff
@@ -0,0 +1,194 @@
...
+/// Rules:
+///   1. Seeing { increases indentation level
+///   2. Seeing } decreases indentation level
+///   3. Insert newline after {, }, ,, and ;
+///   4. Don't put spaces:
+///      between ident and !,
+///      between path segments and ::
+///      after ( and before )
+///      before ,
+pub fn format_schema<W>(schema: &str, mut output: W) where W: ::std::io::Write {
```
killercup: I think it'd look nicer if you `use std::io::Write` at the top of the file and use `<W: Write>` here instead of a where clause with a fully qualified path.

diesel_cli/src/pretty_printing.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+    write!(&mut output, "{}", out).unwrap();
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::format_schema;
+    use std::io::Cursor;
+
```
killercup: All these tests follow the same pattern. I think it'd be nice to reduce the boilerplate code a bit by adding a simple macro here:

```rust
macro_rules! test_format_output {
    ($test_name:ident, $input:expr => $output:expr) => {
        #[test]
        fn $test_name() {
            let out = Vec::<u8>::new();
            let mut c = Cursor::new(out);
            let input = $input;

            format_schema(input, &mut c);
            let format_output = String::from_utf8(c.into_inner()).unwrap();

            assert_eq!($output, format_output);
        }
    }
}
```

which you can then call like this:

```rust
test_format_output!{
    test_remove_whitespace_colon,
    ":: diesel :: types :: Text" => "::diesel::types::Text"
}
```

diesel_infer_schema/src/codegen.rs
```diff
@@ -0,0 +1,104 @@
...
+    }))
+}
+
+pub fn infer_schema_for_schema_name<F1, F2>(
+    database_url: &str,
+    schema_name: Option<&str>,
+    error_handler: F1,
+    filter: F2
+) -> Result<quote::Tokens, Box<Error>>
+    where for<'a> F1: Fn(&'a str, Box<Error>),
```
sgrif: You can just write `Fn(&str, Box<Error>)` here

diesel_infer_schema/src/codegen.rs
```diff
@@ -0,0 +1,104 @@
...
+        #table_name (#(#primary_keys),*) {
+            #(#tokens),*,
+        }
+    }))
+}
+
+pub fn infer_schema_for_schema_name<F1, F2>(
+    database_url: &str,
+    schema_name: Option<&str>,
+    error_handler: F1,
```
sgrif: Taking an error handler function seems unideomatic compared to returning an error.

weiznich: The problem here is that you cannot simply return a `Vec<Result<Tokens, Box<Error>>` because the generated code could be wrapped in another module if a custom schema is set. 
The other solution would be to split this function into two:
 * One generating  `Vec<Result<Tokens, Box<Error>>`
 * One taking `Vec<Tokens>` and returning Tokens (basically wrapping everything if needed into a custom module for custom schemas)

weiznich:
@killercup I've updated the pull request. Now all planed features are implemented. Any idea why the [Travis](https://travis-ci.org/diesel-rs/diesel/jobs/193549575#L703) is failing?

diesel_cli/src/main.rs
```diff
@@ -186,6 +190,33 @@ fn convert_absolute_path_to_relative(target_path: &Path, mut current_path: &Path
...
+            println!("Failed to infer schema for table {}: {}", table_name, error);
+        }, |table_name| {
+            if is_whitelist {
+                return tables.contains(table_name);
+            }
+            if is_blacklist {
+                return !tables.contains(table_name);
+            }
+            true
+        })
```
killercup: I know why you did it this way, but I'm not too happy with the structure of this. I'll try to think of something.

diesel_cli/src/pretty_printing.rs
```diff
@@ -0,0 +1,163 @@
...
+    Ok(out.replace("\t", "    "))
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::format_schema;
+
+    fn run_test(input: &str, expected: &str){
```
killercup: `run_test` is the most generic name ever… 😄 how about `test_formatting`? I still think this would be nicer as a macro, btw ;) Also, please add space before opening bracket.

diesel_codegen/src/schema_inference.rs
```diff
@@ -13,10 +14,14 @@ pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
...
-    let database_url = get_option(&options, "database_url", bug);
+    let database_url = extract_database_url(get_option(&options, "database_url", bug)).unwrap();
     let schema_name = get_optional_option(&options, "schema_name");
 
-    infer_schema_for_schema_name(&database_url, schema_name.as_ref().map(|s| &**s))
+    diesel_infer_schema::infer_schema_for_schema_name(&database_url,
+        schema_name.as_ref().map(|s| &**s),
+        |table_name, error|{
+             panic!("Could not infer table {}: {}", table_name, error)
+        }, |_| true).expect("Could not load tables from database")
```
killercup: Same as above

diesel_infer_schema/src/codegen.rs
```diff
@@ -0,0 +1,104 @@
+use quote;
+use syn;
+
+use std::error::Error;
+use data_structures::ColumnInformation;
+use inference::{establish_connection, get_table_data, determine_column_type,
+                get_primary_keys, load_table_names, InferConnection};
+
+
```
killercup: One empty line too many

diesel_infer_schema/src/codegen.rs
```diff
@@ -0,0 +1,104 @@
...
+    let column_name = syn::Ident::new(&*column.column_name);
+    let column_type = try!(determine_column_type(column, connection));
+    let tpe = if column_type.path[0] == "diesel" && column_type.path[1] == "types" {
+        let path_segments = column_type.path
+            .into_iter()
+            .skip(2)
+            .map(syn::PathSegment::from)
+            .collect();
+        syn::Path { global: false, segments: path_segments }
+    }else{
```
killercup: Add spaces around keyword

diesel_infer_schema/src/inference.rs
```diff
@@ -1,16 +1,21 @@
...
 use diesel::result::Error::NotFound;
 use std::error::Error;
+#[cfg(feature = "postgres")]
+use diesel::pg::PgConnection;
+#[cfg(feature = "sqlite")]
+use diesel::sqlite::SqliteConnection;
+use data_structures::{ColumnInformation, ColumnType};
+
+
+
```
killercup: Two empty lines too many

Mingun:
> thread panicked while panicking. aborting.

Very likely that this happen because you `unwrap` error in `Drop`. I encounter with exactly such error some days ago.

weiznich:
@killercup I've implemented a other variant for the error handling.

weiznich:
All tests are passing on beta now. There are failures on nightly, but they are seaming unrelated to this pull request.
If there are no more issues regarding the code style I think this should be ready to be merged.

killercup:
Thanks, @weiznich!

The nightly failures are from compile-tests, and we need to fix them. It seems to me that either some compiler messages changed because of changes by use (unlikely) or by changes in a new nightly (with nicer/other error messages). I'll have a look this evening.

killercup:
As promised, I made a few more changes to the print-schema/inference code: https://github.com/weiznich/diesel/pull/2

The most important abstraction is are the new `TableData` and `TableDataWithTokens` types that are passed around. I think this makes the code a bit clearer and easier to handle. What do you think?

I'm not sure if this gets run on Travis, but I didn't want to open a new PR on the main repo to not fork the discussion. The tests pass locally though, except for that weird `thread panicked while panicking. aborting.` failure I haven't investigated further.

weiznich:
Thanks for opening the pull request. I'll try to have a look at it tomorrow.

weiznich:
@killercup I've reviewed and merged your changes (Thanks again). Also I added some small tweaks. I think this should be ready for merge now.

killercup:
Thanks, @weiznich. I like your changes, very nice! Travis reports some failures in some unit tests, can you have a look? There are some known problems with recent nightlies; I'll try to get it to run on a known good version.

diesel_infer_schema/src/pg.rs
```diff
@@ -175,6 +177,7 @@ fn skip_views() {
 
     let table_names = load_table_names(&connection, None).unwrap();
 
-    assert!(table_names.contains(&"a_regular_table".to_string()));
-    assert!(!table_names.contains(&"a_view".to_string()));
+    println!("{:?}", table_names);
```
killercup: I think you forgot a println here :)

diesel_cli/src/main.rs
```diff
@@ -194,7 +194,19 @@ fn run_infer_schema(matches: &ArgMatches) {
     let database_url = database::database_url(matches);
     let schema_name = matches.value_of("schema");
 
-    let filtering_tables = matches.values_of("table-name").map(|v| v.collect())
+    let filtering_tables = matches.values_of("table-name").map(|v| {
```
killercup: What does `v` stand for here and in the outer scope? Value? I [agree](https://scribbles.pascalhertleif.de/programming-wisdom.html) that it should have a short name, but it's not immediately obvious what this is. Even after looking at the [clap docs](https://docs.rs/clap/2.20.0/clap/struct.ArgMatches.html#method.values_of) I only got the vague `clap::Values` type, which I assume can be seen as an `Iterator<Item=String>` here.

Can you maybe swap out that `.map.unwrap_or` for a `if let … else`? This way it's a bit clearer we are dealing with an `Option` here and, hopefully, that the inner `map` is on an Iterator.

diesel_cli/src/main.rs
```diff
@@ -194,7 +194,19 @@ fn run_infer_schema(matches: &ArgMatches) {
...
+        if cfg!(feature = "postgres") {
```
killercup: Can you move that `if` into the inner `map`? This way, you can combine the `if` there to something like `if !cfg!(feature = "postgres") || v.contains(".")`?

killercup:
Oh, and if you merge master into this (or rebase on master if like), the tests should pass again on nightly (#596).

weiznich:
@killercup I've fixed the last failing test and rebased on master. All tests are passing now. I also addressed those small style nitpicks.

killercup:
Awesome! Merged! 🎉

PR #579, opened 2017-01-14
Don't use `impl_` macros in doctests anymore.

### LINKED ISSUE #576: Migrate doc tests to use macros 1.1
Since macros 1.1 is _the_ way to use diesel from 0.10/rust 1.15 forward, we should migrate to using custom derives in our API documentation, instead of the macros. This will introduce a (dev) dependency on diesel_codegen, but in general it shouldn't be a problem.

With a quick `rg '///(.*?)impl_(\w+)!' -g '!src/macros'` I could only find 4 usages of the impl_* macros.

dstu:
Change use of `impl_Insertable`, `impl_Queryable`, etc., to use procedural macros now that Macros 1.1 is just around the corner. This should fix #576.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -19,7 +19,7 @@ chrono = { version = "^0.2.17", optional = true }
 uuid = { version = ">=0.2.0, <0.4.0", optional = true, features = ["use_std"] }
 
 [dev-dependencies]
-diesel_codegen = { path = "../diesel_codegen" }
+diesel_codegen = "0.9.0"
 quickcheck = "0.3.1"
 dotenv = "0.8.0"
 tempdir = "^0.3.4"
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -19,6 +19,7 @@ chrono = { version = "^0.2.17", optional = true }
 uuid = { version = ">=0.2.0, <0.4.0", optional = true, features = ["use_std"] }
 
 [dev-dependencies]
+diesel_codegen = { path = "../diesel_codegen" }
 quickcheck = "0.3.1"
 dotenv = "0.8.0"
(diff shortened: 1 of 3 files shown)


diesel/Cargo.toml
```diff
@@ -19,6 +19,7 @@ chrono = { version = "^0.2.17", optional = true }
 uuid = { version = ">=0.2.0, <0.4.0", optional = true, features = ["use_std"] }
 
 [dev-dependencies]
+diesel_codegen = { path = "../diesel_codegen" }
```
killercup: Can you add this with the current version and add an override in the root-level Cargo.toml? Otherwise crates.io won't let us publish this (IIRC).

diesel/src/sqlite/query_builder/functions.rs
```diff
@@ -17,18 +17,13 @@ use super::nodes::Replace;
 /// #     }
 /// # }
 /// #
+/// # #[derive(Insertable)]
+/// # #[table_name="users"]
```
killercup: I think this doc test needs a `#[macro_use] extern crate diesel_codegen;`

diesel/Cargo.toml
dstu: It looks like there's already an override in `/Cargo.toml`:

```
[replace]
"diesel_codegen:0.9.0" = { path = "diesel_codegen" }
```

diesel/src/sqlite/query_builder/functions.rs
dstu: Done.

killercup:
Thanks for the quick fixes, @dstu!

(By the way, Github says the mail address in your commits is not associated with your Github user, but it's just a matter of adding it to the list in your account settings.)

Issue #343, opened 2016-05-24
Support enumerated types


mkroman:
Diesel does not currently support enumerated types created with PostgreSQL.

Documentation on enumerated types [can be found here](http://www.postgresql.org/docs/current/static/datatype-enum.html).

More documentation on `CREATE TYPE` [can be found here](http://www.postgresql.org/docs/current/static/sql-createtype.html).
#### Example

``` sql
CREATE TYPE track_type AS ENUM ('download', 'stream');

CREATE TABLE track (
  id SERIAL,
  type_ track_type
);
```

When compiling, the generated code expects a type named like the custom data type.

```
error: type name `Track_type` is undefined or not in scope [--explain E0412]
 --> src/lib.rs:1:1
1 |> #![feature(custom_derive, custom_attribute, plugin)]
  |> ^ undefined or not in scope
<diesel macros>:5:1: 5:71: note: in this expansion of table_body! (defined in <diesel macros>)
src/schema.rs:4:1: 4:37: note: in this expansion of table! (defined in <diesel macros>)
src/schema.rs:4:1: 4:37: note: in this expansion of infer_schema! (defined in src/lib.rs)
help: no candidates by the name of `Track_type` found in your project; maybe you misspelled the name or forgot to import an external crate?
```

killercup:
FYI in SQLite you could try to do something similar with a text column and `CHECK`:

```sql
CREATE TABLE track (
  id INTEGER PRIMARY KEY,
  track TEXT CHECK(track IN ('download', 'stream')) NOT NULL
);
```

carsonmyers:
Is there currently a good way around this? Like a custom de/serializer?

dstu:
As an outsider who looked through the code to figure out if there is a way to do this, it appears that supporting a user-defined Postgres type requires knowing the object identifier that is in use for that type. In the case of enums, this requires extracting information from `pg_type` and `pg_enum` tables. All told, you're looking at:

* Getting your type information out of relevant [postgres system catalogs](https://www.postgresql.org/docs/current/static/catalogs.html). You'll need to refer to pg_type and pg_enum, along the lines of:
```sql
select pt.typname, pt.typtype, pt.typcategory, pe.enumlabel, pt.oid
  from pg_type pt
  inner join pg_enum pe on pe.enumtypid = pt.oid
  where pt.typcategory = 'E';
```
* Creating peer/placeholder types, analogous to `diesel::types::Date`, for each type so discovered.
[...]

PR #580, opened 2017-01-15
Support for Postgres enums

dstu:
This PR provides the `infer_enums!` macro, which creates Rust enums for each enum discovered in a database, and extends `infer_schema!` so that you can name columns .

See #343 for discussion about how this has been developed up until now.

Items to be addressed before this is ready for merge:
- [ ] Correctly implement all necessary traits for newly declared types.
- [ ] Clean up changes to `infer_schema!` and related macros. (Argument-passing to indicate where to look up non-builtin types is rather unslightly.)
- [ ] Add more Rust documentation describing how the new features work.
- [ ] Fix migrations and tests so that tests pass on SQLite again. (SQLite does not support custom enum types the way Postgres does.)
- [ ] Get things working within custom schemas.
- [ ] Implement and test arrays of custom types.

### DIFF
--- a/diesel/src/macros/macros_from_codegen.rs
+++ b/diesel/src/macros/macros_from_codegen.rs
@@ -88,3 +88,33 @@ macro_rules! embed_migrations {
         }
     }
 }
+
+#[macro_export]
+/// Queries the database for the names of all enum types and creates a Rust enum
+/// type for each one in the current namespace. The enum type name and variants
+/// of the database enum type will be translated from snake case to camel case
+/// (so enum_type::variant becomes EnumType::Variant). For types in a schema
+/// other than the default, the type will be created in a submodule with the
+/// schema name.
+///
+/// This macro can only be used in combination with the `diesel_codegen` or
+/// `diesel_codegen_syntex` crates. It will not work on its own.
+macro_rules! infer_enums {
+    ($database_url: expr) => {
+        mod __diesel_infer_enums {
+            #[derive(InferEnums)]
+            #[options(database_url=$database_url)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_enums::*;
+    };
+
+    ($database_url: expr, $schema_name: expr) => {
+        mod __diesel_infer_enums {
+            #[derive(InferEnums)]
+            #[options(databsae_url=$database_url, schema_name=$schema_name)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_enums::*;
+    };
+}
--- a/diesel/src/macros/macros_from_codegen.rs
+++ b/diesel/src/macros/macros_from_codegen.rs
@@ -12,6 +12,22 @@
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
 macro_rules! infer_schema {
+    (extra_types_module = $extra_types_module: expr, $database_url: expr) => {
+        mod __diesel_infer_schema {
+            #[derive(InferSchema)]
+            #[infer_schema_options(database_url=$database_url, extra_types_module=$extra_types_module)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_schema::*;
+    };
+    (extra_types_module = $extra_types_module: expr, $database_url: expr, $schema_name: expr) => {
+        mod __diesel_infer_schema {
+            #[derive(InferSchema)]
+            #[infer_schema_options(database_url=$database_url, schema_name=$schema_name, extra_types_module=$extra_types_module)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_schema::*;
+    };
     ($database_url: expr) => {
         mod __diesel_infer_schema {
             #[derive(InferSchema)]
@@ -51,11 +67,16 @@ macro_rules! infer_schema {
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
 macro_rules! infer_table_from_schema {
+    (extra_types_module = $extra_types_module: expr, $database_url: expr, $table_name: expr) => {
+        #[derive(InferTableFromSchema)]
+        #[infer_table_from_schema_options(database_url=$database_url, table_name=$table_name, extra_types_module=$extra_types_module)]
+        struct __DieselInferTableFromSchema;
+    };
     ($database_url: expr, $table_name: expr) => {
         #[derive(InferTableFromSchema)]
         #[infer_table_from_schema_options(database_url=$database_url, table_name=$table_name)]
         struct __DieselInferTableFromSchema;
-    }
+    };
 }
 
 #[macro_export]
@@ -103,7 +124,7 @@ macro_rules! infer_enums {
     ($database_url: expr) => {
         mod __diesel_infer_enums {
             #[derive(InferEnums)]
-            #[options(database_url=$database_url)]
+            #[infer_enum_options(database_url=$database_url)]
             struct _Dummy;
         }
         pub use self::__diesel_infer_enums::*;
@@ -112,7 +133,7 @@ macro_rules! infer_enums {
     ($database_url: expr, $schema_name: expr) => {
         mod __diesel_infer_enums {
             #[derive(InferEnums)]
-            #[options(databsae_url=$database_url, schema_name=$schema_name)]
+            #[infer_enum_options(databsae_url=$database_url, schema_name=$schema_name)]
             struct _Dummy;
         }
         pub use self::__diesel_infer_enums::*;
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -249,8 +249,18 @@ pub type VarChar = Text;
 #[cfg(feature = "postgres")]
 pub use pg::types::sql_types::*;
 
+pub trait ProvidesSqlTypeFor<DB> where DB: TypeMetadata {
+    fn self_metadata() -> DB::TypeMetadata;
+}
+
 pub trait HasSqlType<ST>: TypeMetadata {
     fn metadata() -> Self::TypeMetadata;
+}
+
+impl<ST, DB> HasSqlType<ST> for DB where DB: TypeMetadata, ST: ProvidesSqlTypeFor<DB> {
+    fn metadata() -> Self::TypeMetadata {
+        <ST as ProvidesSqlTypeFor<DB>>::self_metadata()
+    }
 }
 
 pub trait NotNull {
--- a/diesel_codegen/Cargo.toml
+++ b/diesel_codegen/Cargo.toml
@@ -10,11 +10,14 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_codegen"
 keywords = ["orm", "database", "postgres", "sql", "codegen"]
 
 [dependencies]
-syn = "0.10.3"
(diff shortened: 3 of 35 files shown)


killercup:
Nice! I saw your branch a while ago, and am glad to see this PR. This is quite a lot of code! I'll try to read through it this evening, or tomorrow afternoon. Feel free to bug me if I forget about it though :)

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -74,19 +88,51 @@ mod information_schema {
     }
 }
 
-pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box<Error>> {
+pub fn camel_cased(snake_case: &str) -> String {
```
killercup: Yes! Finally! PascalCase! Errrr, I mean, "upper camel case". (On a more serious note, I think this `fn` should be called `snake_to_type_case`.)

Please note that all other case conversions we do are trivial conversions by choice, to not introduce surprisingly complex behavior.

The snake case version we get here are valid Postgres identifiers, right? What properties do they have in comparison to Rusts type identifiers?

diesel_codegen_shared/src/schema_inference/pg.rs
```diff
@@ -171,3 +228,39 @@ fn skip_views() {
...
+        assert_eq!("CamelCased", camel_cased("camel_cased_"));
+        assert_eq!("CamelCased", camel_cased("_camel_cased"));
+        assert_eq!("CamelCased", camel_cased("_camel_cased_"));
+        assert_eq!("CamelCased", camel_cased("_camel__cased_"));
+    }
+
+    #[test]
+    fn camel_cased_i18n() {
+        assert_eq!("Außerdem", camel_cased("außerdem"));
+        assert_eq!("AuSSerdem", camel_cased("au_ßerdem"));
```
killercup: N̵͡i̛c͘e ̀e͜d̨͘͢g͜͡e͠͞ ̴̛c̶̨͞a͢͜s͞ȩ̵s.͟͠

Can enum variant names contain umlauts, though?

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -88,3 +109,33 @@ macro_rules! embed_migrations {
...
+#[macro_export]
+/// Queries the database for the names of all enum types and creates a Rust enum
+/// type for each one in the current namespace. The enum type name and variants
+/// of the database enum type will be translated from snake case to camel case
+/// (so enum_type::variant becomes EnumType::Variant). For types in a schema
+/// other than the default, the type will be created in a submodule with the
+/// schema name.
+///
+/// This macro can only be used in combination with the `diesel_codegen` or
+/// `diesel_codegen_syntex` crates. It will not work on its own.
```
killercup: Just a heads up: diesel_codegen_syntex will be removed in 0.10

diesel_codegen/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(warnings)]
+#![recursion_limit = "512"]
```
killercup: What is this needed for? (It's fine, just maybe add a comment what requires this, so we can remove it when it's no longer needed.)

diesel/src/types/mod.rs
```diff
@@ -249,10 +249,20 @@ pub type VarChar = Text;
 #[cfg(feature = "postgres")]
 pub use pg::types::sql_types::*;
 
+pub trait ProvidesSqlTypeFor<DB> where DB: TypeMetadata {
```
killercup: I would love to see a doc comment here telling me what this type is all about. I assume it abstracts over structs and enums providing SQL types? Maybe also mention `HasSqlType` is implemented for everything that implements this trait.

diesel_codegen/Cargo.toml
```diff
@@ -10,11 +10,15 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_codegen"
...
 [dependencies]
-syn = "0.10.3"
+syntex_syntax = "0.52.0"
 quote = "0.3.10"
 diesel = { version = "0.9.0", default-features = false }
 diesel_codegen_shared = { version = "0.9.0", default-features = false }
 
+[dependencies.syn]
+version = "0.10.3"
+features = ["full"]
```
killercup: Can you use the same inline table syntax as above?

```toml
syn = { version = "0.10.3", features = ["full"] }
```

diesel_codegen/Cargo.toml
```diff
@@ -10,11 +10,15 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_codegen"
 keywords = ["orm", "database", "postgres", "sql", "codegen"]
 
 [dependencies]
-syn = "0.10.3"
+syntex_syntax = "0.52.0"
```
killercup: I hope we can not introduce this dependency. Syntex takes ages to compile.

diesel_codegen/src/schema_inference.rs
```diff
@@ -1,4 +1,5 @@
 use syn;
+// use syntex_syntax::symbol::keywords;
```
dstu: Yes, I was trying to use syntax to work around some issues but don't need it right now. I'll kill the dependency if it's not needed (which it probably won't be) before any merge.

diesel_codegen_shared/src/schema_inference/pg.rs
dstu: OK on changing the method name.

Right now, this PR converts a Postgres type name or variant `foo_bar` into `FooBar` on the Rust side. There are bool arguments to the macro expansion methods to control whether this happens, but they aren't currently used because there is no macro-level interface for toggling them.

I think it makes sense to map to Rust-style naming, but I am willing to be convinced otherwise. If any name manging could happen, clearly documenting default behavior and providing an option to turn it off are also necessary.

diesel_codegen_shared/src/schema_inference/pg.rs
dstu: In Rust, it's [an open issue](https://github.com/rust-lang/rust/issues/28979).

In Postgres, it appears perfectly valid to have both types and variants contain umlauts. I'll have to dig deeper into the documentation to see if there are any surprises with regards to other unusual characters.

It seems like eventual convergence (arbitrary whitespace-free Unicode allowed) is an eventual goal, but it would probably be better to check explicitly that we're generating a legal type name and die with a reasonable explanation rather than letting rustc barf.

diesel_codegen/src/lib.rs
dstu: The `quote!` macro makes use of (potentially deep) recursion, and my large invocations of it are running into the default limit. This number should be tuned to fit whatever this PR's eventual needs are; it may be that we wind up needing something smaller at first.

diesel_codegen/Cargo.toml
dstu: Agreed. I don't think this will be needed.

dstu:
This can be broken up fairly easily; it is quite feasible to implement support simply for generating enum types to match what's in a database.

I'll make changes to the PR that address the comments so far when I have more time this week.

diesel_codegen/src/type_inference.rs
```diff
@@ -0,0 +1,254 @@
...
+fn read_type_names(csl: &str) -> HashSet<String> {
+    csl.split(",")
+        .map(|t| t.trim())
+        .map(canonicalize_pg_type_name)
+        .filter(|t| !t.is_empty()).collect()
+}
+
+fn infer_enums_for_schema_name(database_url: &str, schema_name: Option<&str>,
+                               type_list: Option<&str>, camel_case_types: bool,
+                               camel_case_variants: bool) -> quote::Tokens {
```
dstu: Formatting is mostly a combination of what rust-mode does in Emacs and my attempts at respecting cosmetic issues like line length as per a hazy recollection of [the recommendations aturon compiled](https://aturon.github.io/). If you prefer, I'll run rustfmt over things before declaring this ready for a merge.

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -12,6 +12,22 @@
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
 macro_rules! infer_schema {
+    (extra_types_module = $extra_types_module: expr, $database_url: expr) => {
```
dstu: I really don't like the solution of the user manually specifying `extra_types_module`.

From a usability standpoint, it would be nicer to be able to `infer_everything!` and just get tables, types, and (eventually) views from the database.

From a design standpoint, this is just kicking the problem down the field a short distance. As-is, all peers to SQL types live in `diesel::types`, and library internals freely assume this. Looking up types that are not on a whitelist on a single user-specified alternate path works for now, but a few problems are readily evident:

* All DB-derived types are assumed to live in one namespace. In Postgres, you have the potential for multiple namespaces (which naturally arise from each distinct database connection string, or even from schemas nested inside of a database). This could lead to name clashes if, for example, identically named types are declared in different databases or schemas. (It seems like a good solution is to recommend strongly that different invocations of `infer_schema!` use different `extra_types_module` settings, but I'd prefer a cleaner design where remembering to do that sort of thing isn't necessary.)
* The binding between a DB type and a Rust type is implicit (based on type name alone). This loose coupling could lead to bugs whenever there are name clashes.
* It is quite conceivable that a user will want to use types that come from disparate Rust namespaces, at which point they will have to do a lot of tedious and potentially confusing re-exporting with `pub use` (so that everything is in one namespace), or we will have to extend this mechanism to support having multiple extra type paths.

This could be solved by creating an intermediate representation of types in a database schema and mapping DB types to Rust types through that. This would require threading some sort of global table state through most of the code generation process, so the effects of adopting this design would be felt throughout a lot of Diesel. (Perhaps simply having a global shared structure would be adequate, since it will only be used at compilation time.)

Having a central, explicit set of data structures that describe types, tables, views, procedures, etc., could be quite useful. I suspect that some sort of compile-time metadata model would be very helpful for validating what is read from the database ("table has a primary key structure we don't support"), generating good error messages ("typename foobaz is not a valid Rust identifier"), debugging/testing (write tests against the intermediate representation code), streamlining code generation (all necessary state is in one obvious place), and error reporting (have richer local state from which to produce meaningful error messages).

This approach could go very wrong by being too top-heavy and prematurely putting straightjacketing Diesel. The project has gotten really far without building a top-down intermediate model of database components, and maybe it's best to keep it that way for now. It would be unpleasant to have to refactor a big metadata layer every time you want to support a new DB feature.

If it seems a good idea, I suggest working towards merging just the enum declaration part of this PR and opening a separate issue to discuss whether a metadata model is necessary or a good idea. That discussion may yield solutions to the problems I've described here which can be used to add DB type support to `infer_schema!` without having to commit to a great deal more overhead.

gsollazzo:
Hi! 
When can we expect to see this PR merged in master?
Thanks

dstu:
I apologize for not having been on top of this. I may be able to put some more time into it this weekend. I still need to rebase on top of the latest Diesel release and restrict this PR's scope to consist only of generating types for Postgres enums.

dstu:
After looking at how far diesel has progressed since I wrote this PR, I have concluded that it must be rewritten. The basic approach (query Postgres catalog tables to get enum definitions) is still perfectly valid, but diesel's switch to using SQL schema standard means that the way this is done has to change. I am closing this PR and will move discussion back to #343.

PR #592, opened 2017-01-26
Adding documentation to connection module.

ericho:
Signed-off-by: Erich Cordoba <erich.cm@yandex.com>


diesel/src/connection/mod.rs
```diff
@@ -4,12 +4,16 @@ use query_source::Queryable;
...
 use types::HasSqlType;
 
+/// Trait to perform a simple connection with the associated backend.
 pub trait SimpleConnection {
     #[doc(hidden)]
     fn batch_execute(&self, query: &str) -> QueryResult<()>;
 }
 
+/// Adds `establish`, `transaction`, `test_transaction` and `begin_test_transaction`
+/// to all types that implements `SimpleConnection` trait.
```
killercup: While technically right, this is the same information that rustdoc shows you :)

Mentioning that it extend tre SimpleConnection trait is a good idea though.

Maybe it helps to instead answer questions like "When do I use this?", or "Where does this fit in my application?", or "What problem does this solve?". Or, as this is a trait, "What does this abstract over?"

ericho:
Thanks @killercup I'll address the changes :smile:

ericho:
Looking more deeply in the code (and trying to learn rust :smile:) it looks to me that `SimpleConnection` is only used on migrations as you pointed. And in migrations, only the `batch_execute` function is used.

For any new implementation of `Connection` the `SimpleConnection` trait shall be implemented as well as `batch_execute` in order to support migrations. 

So, I have a couple questions:
1. Should `batch_execute` be documented as well? Currently is marked as `#[doc[hidden]]`
2. Why not using `Connection` instead of `SimpleConnection` in migrations?

diesel/src/connection/mod.rs
```diff
@@ -4,12 +4,18 @@ use query_source::Queryable;
 use result::*;
 use types::HasSqlType;
 
+/// To perform simple operations to the backed that implements this trait.
+/// `SimpleConnection` will be used for migrations. Every backend shall implement
+/// their own way to run migrations scripts.
 pub trait SimpleConnection {
-    #[doc(hidden)]
+    /// Migrations will use this function to run the `up.sql` and `down.sql` files.
+    /// New backends shall provide their own mechanism to execute the migration SQL scripts
```
sgrif: This function is expected to be able to handle strings containing multiple SQL statements.

diesel/src/connection/mod.rs
```diff
@@ -4,12 +4,18 @@ use query_source::Queryable;
...
     fn batch_execute(&self, query: &str) -> QueryResult<()>;
 }
 
+/// Every backend shall implement this trait to perform connections to the database.
```
sgrif: Can this be phrased as a description of what the trait does, rather than a command?

diesel/src/connection/mod.rs
```diff
@@ -4,12 +4,18 @@ use query_source::Queryable;
...
 pub trait Connection: SimpleConnection + Sized {
+    /// A trait that provides specific operations for the backed.
```
sgrif: How about "The backend this connection represents"?

diesel/src/connection/mod.rs
```diff
@@ -8,12 +8,19 @@ use types::HasSqlType;
 
 pub use self::transaction_manager::{TransactionManager, AnsiTransactionManager};
 
+/// To perform simple operations to the backed that implements this trait.
+/// `SimpleConnection` will be used for migrations. Every backend shall implement
+/// their own way to run migrations scripts.
 pub trait SimpleConnection {
-    #[doc(hidden)]
+    /// Brings the ability to execute multiple SQL statements within the same string.
+    /// This function is tipically used on migrations where the statements to upgrade or
```
killercup: "typically" with y. I think it's also "in migrations" (instead of "on").

diesel/src/connection/mod.rs
```diff
@@ -8,12 +8,19 @@ use types::HasSqlType;
 
 pub use self::transaction_manager::{TransactionManager, AnsiTransactionManager};
 
+/// To perform simple operations to the backed that implements this trait.
+/// `SimpleConnection` will be used for migrations. Every backend shall implement
+/// their own way to run migrations scripts.
 pub trait SimpleConnection {
-    #[doc(hidden)]
+    /// Brings the ability to execute multiple SQL statements within the same string.
```
killercup: "Brings the ability to" doesn't really add anything here, and removing this part of the sentence make the rest a nice headline. I'd put the sentences in separate paragraphs too, as the first one can serve as summary and second one is an example.

diesel/src/connection/mod.rs
```diff
@@ -8,12 +8,19 @@ use types::HasSqlType;
 
 pub use self::transaction_manager::{TransactionManager, AnsiTransactionManager};
 
+/// To perform simple operations to the backed that implements this trait.
+/// `SimpleConnection` will be used for migrations. Every backend shall implement
+/// their own way to run migrations scripts.
```
killercup: This is better, but still too many words for too little information. No need to make it longer than necessary – technical documentation is typically very terse. How about

> Perform simple operations on a backed.
>
> This is used for migrations.

diesel/src/connection/mod.rs
```diff
@@ -8,12 +8,19 @@ use types::HasSqlType;
...
 pub trait SimpleConnection {
-    #[doc(hidden)]
+    /// Brings the ability to execute multiple SQL statements within the same string.
+    /// This function is tipically used on migrations where the statements to upgrade or
+    /// downgrade the database are stored in SQL batch files.
     fn batch_execute(&self, query: &str) -> QueryResult<()>;
 }
 
+/// Provides functionality to perform connections to the backend.
```
killercup: Same here, you can leave out "Provides functionality to"

ericho:
I just realize that if I delete my branch the PR is closed as well :( I created the #766 

Sorry for the mess..

PR #604, opened 2017-02-02
Clean up output of `diesel print-schema`

sgrif:
Prior to this commit, the output would include the `mod infer_table {
table! { ... }; pub use self::infer_table::*;` wrapper that we generate.
This is only needed due to how the macros work with Macros 1.1. When
calling the various functions directly, we should generate code that is
closer to what the user would write themselves.

In addition to not including that wrapper code, I've also inserted a
newline between each table! macro, and trimmed all trailing whitespace,
making it nearly exactly what the user would write themselves.

### DIFF
--- a/diesel_cli/src/pretty_printing.rs
+++ b/diesel_cli/src/pretty_printing.rs
@@ -23,9 +23,9 @@ pub fn format_schema(schema: &str) -> Result<String, FmtError> {
         // The `quote!` macro inserts whitespaces at some strange location,
         // let's remove them!
         match c {
-            '!' | ';' | ',' | '<' | ')' | '>' if last_char.is_whitespace() => {
+            '!' | ',' | '<' | ')' | '>' if last_char.is_whitespace() => {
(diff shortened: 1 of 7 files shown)


diesel_cli/src/pretty_printing.rs
```diff
@@ -82,7 +82,7 @@ pub fn format_schema(schema: &str) -> Result<String, FmtError> {
         }
     }
 
-    Ok(out.replace("\t", "    "))
+    Ok(out.replace("\t", "    ").replace("table!", "\ntable!").trim().to_string())
```
killercup: I wonder if there's a pull-based string processor crate. Like an iterator, but with replace and trim and such

sgrif: I want to completely rewrite this at some point. I seriously doubt that this function should be operating on token streams at all. But for now I just wanted to get the minimum possible change for the user visible changes.

PR #608, opened 2017-02-02
[WIP] Remove codegen shared

weiznich:
Before merging the following macro-modules need to be ported to codegen:

- [x] queryable
- [ ] as_changeset
- [ ] associations
- [ ] identifiable
- [ ] insertable


killercup:
@sgrif, I think we talked about moving these macros to pure custom derives a while ago. Do you want to do that right now?

Anyway, it should be one PR per macro IMHO.

sgrif:
The main reason those macros are in the main Diesel crate is to provide access to `$crate`, and avoid hard coding `::diesel`.

killercup:
Ohhhh right, `$crate`, I remember. That's obviously not possible in a separate derive crate. You can't rename an extern crate, can you? So the full path may be ugly, but not inherently worse.

sgrif:
You can rename an extern crate, and you also do not have to import it at the root of your project.

weiznich:
Serde does something like [this](https://github.com/weiznich/diesel/blob/remove_codegen_shared/diesel_codegen/src/queryable.rs#L42-L57) to prevent such issues. This should solve the `$crate` issue.

I assume the first commit should be submitted as own pull request? It only moves the remaining modules from diesel_codegen_shared to diesel_codegen.

sgrif:
Hm. I did like having the macros in the main crate since it made it possible to unit test a lot of the codegen stuff (especially weird combinations of things that need to be enumerated) without relying on actual codegen, but I should reconsider that. I can't reopen this PR so you'll need to open a new one.

weiznich:
I opened a new pull request that only removes diesel_codegen_shared for now. (See #613 )
After the new pull request was merged I could open followup pull requests to move the codegen from the macro implementation to diesel_codegen. 
(In my opinion the generation by syn/quote is much more readable and shorter as the current macro implementation)

PR #611, opened 2017-02-03
Add initial structure for MySQL support

sgrif:
This represents the bare minimum required to define all the basic
structures MySQL will be needing, and have the tests compile. I've opted
not to include this as a travis allowed failure or in `bin/test` until
at least the core crate is mostly complete.

A few things that stood out to me while going through all of this:

It's called `Mysql` and not `MySql` because the module and feature are
called `mysql` not `my_sql`. The struct would never be called `MySQL` in
ideomatic Rust.

I'm not sure I generated bindings correctly since there's a lot of
different header files, so I haven't published yet. I was rather
surprised nobody has written a wrapper around their C API yet. Just the
pure Rust client (which mirrors rust-postgres's API so will be
impossible to use for all the same reasons)

The C API has global initialization. I believe that wrapping it in a
`std::sync::Once` and calling that from `MysqlConnection::establish`.

Lol if your enum variant for a 32 bit integer is called `Long`

The bind type variants for input vs output are asymmentrical. We are
mirroring the input, as we would only ever care about the output for
debug purposes. There are a lot of types missing from the input list.
Notably, decimal is missing. Unsure how we transmit in that case
(granted, it's not like we have any Rust type to work with anyway)

SqliteQueryBuilder, DebugQueryBuilder, and MysqlQueryBuilder are all
currently identical. I'm not sure that I ever expect them to diverge,
and will consider unifying them.

### DIFF
--- a/diesel/src/doctest_setup.rs
+++ b/diesel/src/doctest_setup.rs
@@ -1,60 +1,91 @@
 extern crate dotenv;
+#[macro_use] extern crate cfg_if;
 
 use diesel::prelude::*;
 use self::dotenv::dotenv;
 
-#[cfg(feature = "postgres")]
-#[allow(dead_code)]
-type DB = diesel::pg::Pg;
+cfg_if! {
+    if #[cfg(feature = "postgres")] {
+        #[allow(dead_code)]
+        type DB = diesel::pg::Pg;
 
-#[cfg(feature = "postgres")]
-fn connection_no_data() -> diesel::pg::PgConnection {
+        fn connection_no_data() -> diesel::pg::PgConnection {
+            let connection_url = database_url_from_env();
+            let connection = diesel::pg::PgConnection::establish(&connection_url).unwrap();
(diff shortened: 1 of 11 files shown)


diesel/src/doctest_setup.rs
```diff
@@ -3,58 +3,87 @@ extern crate dotenv;
...
+            )").unwrap();
+            connection.execute("INSERT INTO users (name) VALUES ('Sean'), ('Tess')").unwrap();
 
-#[cfg(all(not(feature = "postgres"), feature = "sqlite"))]
-#[allow(dead_code)]
-type DB = diesel::sqlite::Sqlite;
+            connection
+        }
+    } else if #[cfg(feature = "mysql")] {
+        type DB = diesel::mysql::Mysql;
```
killercup: `#[allow(dead_code)]` for consistency

sgrif:
I'll fix the doctests before merging, just need an extern crate in the setup file

PR #612, opened 2017-02-03
Upgrade libsqlite3-sys to 0.6

### LINKED ISSUE #610: Update libsqlite3-sys dependency
I'd like to use the latest version of `libsqlite3-sys` alongside Diesel in a workspace. Cargo requires that only a single version of a native library be linked to at once. Because Diesel depends on `0.4` and the latest version is `0.6.2`, the workspace does not test successfully with `cargo test --all`.

Any chance Diesel could upgrade to the latest version of `libsqlite3-sys`?

killercup:
0.6.2 works fine locally, haven't tested any version in between though.

Fixes #610

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["database"]
 [dependencies]
 libc = "0.2.*"
 pq-sys = { version = "^0.2.0", optional = true }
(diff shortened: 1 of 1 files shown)


sgrif:
LGTM pending CI

PR #614, opened 2017-02-03
Connect to a database in `MysqlConnection::establish`

sgrif:
MySQL does handle DNS resolution, but doesn't handle URL parsing for us.
However, other than the `database` field, the arguments to
`mysql_real_connect` are basically just URI fragments, so I don't think
this warrants moving away from always taking a connection URL.

I have never felt less safe using unsafe than calling a function called
`mysql_real_connect`.

### DIFF
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -1,0 +1,103 @@
+extern crate mysqlclient_sys as ffi;
+
+use std::ffi::CStr;
+use std::os::{raw as libc};
+use std::ptr;
+use std::sync::{Once, ONCE_INIT};
+
+use result::{ConnectionResult, ConnectionError};
+use super::url::ConnectionOptions;
+
+pub struct RawConnection(*mut ffi::MYSQL);
+
+impl RawConnection {
+    pub fn new() -> Self {
+        perform_thread_unsafe_library_initialization();
+        let raw_connection = unsafe { ffi::mysql_init(ptr::null_mut()) };
+        if raw_connection.is_null() {
+            // We're trusting https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html
+            // that null return always means OOM
+            panic!("Insufficient memory to allocate connection");
+        }
+        let result = RawConnection(raw_connection);
+
+        // This is only non-zero for unrecognized options, which should never happen.
+        let charset_result = unsafe { ffi::mysql_options(
+            result.0,
+            ffi::mysql_option::MYSQL_SET_CHARSET_NAME,
+            b"utf8mb4\0".as_ptr() as *const libc::c_void,
+        ) };
+        assert_eq!(0, charset_result, "MYSQL_SET_CHARSET_NAME was not \
+                   recognized as an option by MySQL. This should never \
+                   happen.");
+
+        result
+    }
+
+    pub fn connect(&self, connection_options: ConnectionOptions) -> ConnectionResult<()> {
+        let host = try!(connection_options.host());
+        let user = try!(connection_options.user());
+        let password = try!(connection_options.password());
+        let database = try!(connection_options.database());
+        let port = connection_options.port();
+
+        unsafe {
+            // Make sure you don't use the fake one!
+            ffi::mysql_real_connect(
+                self.0,
+                host.map(|x| x.as_ptr()).unwrap_or(ptr::null_mut()),
+                user.as_ptr(),
+                password.map(|x| x.as_ptr()).unwrap_or(ptr::null_mut()),
+                database.map(|x| x.as_ptr()).unwrap_or(ptr::null_mut()),
+                port.unwrap_or(0) as u32,
+                ptr::null_mut(),
+                0,
+            )
+        };
+
+        let last_error_message = self.last_error_message();
+        if last_error_message.is_empty() {
+            Ok(())
+        } else {
+            Err(ConnectionError::BadConnection(last_error_message))
+        }
+    }
+
+    pub fn last_error_message(&self) -> String {
+        unsafe { CStr::from_ptr(ffi::mysql_error(self.0)) }
+            .to_string_lossy()
+            .into_owned()
+    }
+}
+
+impl Drop for RawConnection {
+    fn drop(&mut self) {
+        unsafe { ffi::mysql_close(self.0); }
+    }
+}
+
+/// In a nonmulti-threaded environment, mysql_init() invokes mysql_library_init() automatically as
+/// necessary. However, mysql_library_init() is not thread-safe in a multi-threaded environment,
+/// and thus neither is mysql_init(). Before calling mysql_init(), either call mysql_library_init()
+/// prior to spawning any threads, or use a mutex to protect the mysql_library_init() call. This
+/// should be done prior to any other client library call.
+///
+/// https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html
+static MYSQL_THREAD_UNSAFE_INIT: Once = ONCE_INIT;
+
+fn perform_thread_unsafe_library_initialization() {
+    MYSQL_THREAD_UNSAFE_INIT.call_once(|| {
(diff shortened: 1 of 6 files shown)


diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+        if raw_connection.is_null() {
+            panic!("Insufficient memory to allocate connection");
+        }
+        let result = RawConnection(raw_connection);
+
+        // This is only non-zero for unrecognized options, which should never happen.
+        let charset_result = unsafe { ffi::mysql_options(
+            result.0,
+            ffi::mysql_option::MYSQL_SET_CHARSET_NAME,
+            b"utf8mb4\0".as_ptr() as *const libc::c_void,
```
sgrif: `utf8mb4`. Because why would you have `utf8` be the thing that adheres to the unicode standard?

killercup: Don't get me started on mysql utf8 support… a coworker recently had to migrate their Redmine instance to utf8mb4 (that's "UTF-8, much better, fourth try", right?), because I used an emoji in a commit message.

The docs [say](https://dev.mysql.com/doc/refman/5.7/en/mysql-options.html),

> The name of the character set to use as the default character set.

What does setting the default charset actually do? Does mysql use that when you don't specify the charset in a `create table` or `create database` call? Or some more magic?

diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+        // which isn't picked up by bindgen
+        let error_code = unsafe { ffi::mysql_server_init(0, ptr::null_mut(), ptr::null_mut()) };
+        if error_code != 0 {
+            // FIXME: This is documented as Nonzero if an error occurred.
+            // Presumably the value has some sort of meaning that we should
+            // reflect in this message. We are going to panic instead of return
+            // an error here, since the documentation does not indicate whether
+            // it is safe to call this function twice if the first call failed,
+            // so I will assume it is not.
+            panic!("Unable to perform MySQL global initialization");
```
killercup: Maybe also give the error_code to avoid making this branch less of a potential heisenbug

diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn drop(&mut self) {
+        unsafe { ffi::mysql_close(self.0); }
+    }
+}
+
+/// In a nonmulti-threaded environment, mysql_init() invokes mysql_library_init() automatically as
+/// necessary. However, mysql_library_init() is not thread-safe in a multi-threaded environment,
+/// and thus neither is mysql_init(). Before calling mysql_init(), either call mysql_library_init()
+/// prior to spawning any threads, or use a mutex to protect the mysql_library_init() call. This
+/// should be done prior to any other client library call.
```
killercup: This is a quote from https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html :)

diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+        let result = RawConnection(raw_connection);
+
+        // This is only non-zero for unrecognized options, which should never happen.
+        let charset_result = unsafe { ffi::mysql_options(
+            result.0,
+            ffi::mysql_option::MYSQL_SET_CHARSET_NAME,
+            b"utf8mb4\0".as_ptr() as *const libc::c_void,
+        ) };
+        assert_eq!(0, charset_result);
```
killercup: Might be user facing, I'd add a short description (as 3rd param). (Or turn into debug_assert instead)

diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+use super::url::ConnectionOptions;
+
+pub struct RawConnection(*mut ffi::MYSQL);
+
+impl RawConnection {
+    pub fn new() -> Self {
+        perform_thread_unsafe_library_initialization();
+        let raw_connection = unsafe { ffi::mysql_init(ptr::null_mut()) };
+        if raw_connection.is_null() {
+            panic!("Insufficient memory to allocate connection");
```
killercup: I'd add a comment that this is the error case that https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html documents, trusting it's actually the only case where NULL is returned.

diesel/src/mysql/connection/raw.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    pub fn connect(&self, connection_options: ConnectionOptions) -> ConnectionResult<()> {
+        let host = try!(connection_options.host());
+        let user = try!(connection_options.user());
+        let password = try!(connection_options.password());
+        let database = try!(connection_options.database());
+        let port = connection_options.port();
+
+        unsafe {
+            // Make sure you don't use the fake one!
```
killercup: :trollface:

diesel/src/mysql/connection/url.rs
```diff
@@ -0,0 +1,97 @@
...
+    let msg = "MySQL connection URLs must be in the form \
+        `mysql://[[user]:[password]@]host[:port][/database]`";
+    ConnectionError::InvalidConnectionUrl(msg.into())
+}
+
+#[test]
+fn urls_with_schemes_other_than_mysql_are_errors() {
+    assert!(ConnectionOptions::parse("postgres://localhost").is_err());
+    assert!(ConnectionOptions::parse("http://localhost").is_err());
+    assert!(ConnectionOptions::parse("mysql://localhost").is_ok());
```
killercup: Maybe also

    assert!(ConnectionOptions::parse("/lol_no_schema/at_all").is_err());

Actually—how do you connect to a socket?

diesel/src/mysql/connection/raw.rs
sgrif: It's the equivalent of the client encoding setting in PG as best I can tell. It's the encoding used for non-query-result data. Specifically I'm setting it here because it's the encoding used for all of the parameters to `mysql_real_connect`.

sgrif:
> Do you want to add more to this to include a test that actually connects to a MySQL server?

All of our existing tests do that. They now fail on the `unimplemented!()` call that is the result of `begin_test_transaction()`

diesel/src/mysql/connection/raw.rs
sgrif: As mentioned in the comment above, this should *never* happen unless there was a linker error. I'll add a description though.

diesel/src/mysql/connection/raw.rs
sgrif: Ugh. You and your links.

diesel/src/mysql/connection/raw.rs
killercup: Ok, cool. I wasn't suggesting not setting it; always make sure to get utf-8—sorry, i mean utf_real_8—if you can 😄

diesel/src/mysql/connection/raw.rs
sgrif: The error code is useless as best I can tell ATM. The FIXME comment is to figure out what to actually do here.

diesel/src/mysql/connection/url.rs
sgrif: `"/lol_no_schema/at_all"` is not a valid URL. We don't support connecting to unix sockets at the moment.

killercup: Yeah it was meant as a sanity check. If you feel like it, you could add `socket://` and `file://` as `Err` cases.

sgrif: :+1:

diesel/src/mysql/connection/raw.rs
sgrif: Added a message. I'm leaving at as an actual assert and not just a debug assertion though since failing to successfully set this option could cause safety issues.

Issue #521, opened 2016-12-05
Refactorings using macros in type position


killercup:
Since [Rust 1.13](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1130-2016-11-10) we can use macros in type position ([RFC #873](https://github.com/rust-lang/rfcs/blob/master/text/0873-type-macros.md)). With those, we should be able to refactor a bunch of thing, e.g. the impls of [EqAll](https://github.com/diesel-rs/diesel/blob/49e3cdc079ec80eb6962bb5e6ca4e6471653327d/diesel/src/expression/expression_methods/eq_all.rs#L16-L17).

phungleson:
Hey mate, I am trying to tackle this but it seems more complicated than I initially thought.

Do you have any other examples rather than the one in RFC873? or any simple ideas that I can refer to, like how the macro type might look like?

sgrif:
This was implemented by #617 and we didn't like the code. Variadics will eventually make this pointless anyway.

PR #617, opened 2017-02-04
Use macro to impl EqAll for tuples

phungleson:
Part of https://github.com/diesel-rs/diesel/issues/521

@killercup how is this implementation, any suggestions?


sgrif:
Going to close this for now, as I think the need for this is going to end up going away one way or another. If we continue to only implement for sizes up to 4 (which seems to be all we need, really), I think the code is much more accessible without the macros. I'm going to remove the comment for now. Thanks for working on this, and sorry it got caught in the confusion.

PR #618, opened 2017-02-04
Refactor duplicated transaction management code

sgrif:
The code for transactions is duplicated between SQLite and PostgreSQL.
MySQL would have also used identical code. However, the SQL being
executed is not universal across all backends. Oracle appears to use the
same SQL, but SQL Server has its own special syntax for this. As such,
I'm not comfortable promoting this to a default impl on the trait.
Instead I've moved the code out into a shared trait/struct, and operate
on that instead.

I had wanted to make `TransactionManager` be generic over the backend,
not the connection itself, since constraints for it will always be about
the backend, but I ran into
https://github.com/rust-lang/rust/issues/39532 when attempting to do so.

### DIFF
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -1,0 +1,92 @@
+use backend::UsesAnsiSavepointSyntax;
+use connection::Connection;
+use result::QueryResult;
+
+/// Manages the internal transaction state for a connection. You should not
+/// interface with this trait unless you are implementing a new connection
+/// adapter. You should use [`Connection::transaction`][transaction],
+/// [`Connection::test_transaction`][test_transaction], or
+/// [`Connection::begin_test_transaction`][begin_test_transaction] instead.
+pub trait TransactionManager<Conn: Connection> {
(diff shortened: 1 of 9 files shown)


diesel/src/connection/transaction_manager.rs
```diff
@@ -0,0 +1,92 @@
+use backend::UsesAnsiSavepointSyntax;
+use connection::Connection;
+use result::QueryResult;
+
+/// Manages the internal transaction state for a connection. You should not
+/// interface with this trait unless you are implementing a new connection
+/// adapter. You should use [`Connection::transaction`][transaction],
+/// [`Connection::test_transaction`][test_transaction], or
+/// [`Connection::begin_test_transaction`][begin_test_transaction] instead.
```
sgrif: But you just said that *you* would link links. ;)

killercup: Sorry, what I meant to say was that I link _likes_! :D

sgrif: I'm going to do it as a separate commit. I need to rebase something off this branch to open a PR

PR #622, opened 2017-02-04
Execute queries in `execute` and `batch_execute`

sgrif:
Neither of these methods are public API. `batch_execute` is used by
migrations, and `execute`.... Actually can probably be deleted at this
point. They are both methods that are intended to run raw SQL queries
(batch runs more than one, execute returns affected rows) which do not
involve prepared statements or any results, so they make good low
targets for this piece of the implementation.

Setting `MYSQL_OPTION_MULTI_STATEMENTS_ON` appears to have all kinds of
nasty side effects that may blow everything up. However, our test suite
does not appear to be blowing up yet. I suspect that a migration which
contains an insert/update statement with a returning clause or a select
statement (which there's no reason to do, but let's check anyway) will
blow up.

Things that are terrible about MySQL that I had forgotten about before
today:

- Columns of type `BLOB`, `TEXT`, `GEOMETRY`, and `JSON` cannot have
  default values. Because reasons I guess.
- `VARCHAR` is not valid syntax, even though the documentation claims
  that you can have a varchar column of unspecified length.
- Any modifications of schema will implicitly commit the current
  transaction, meaning our test suite will need major modifications to
  support MySQL.
- Nobody is fucking consistent in how the declare auto incrementing
  columns.

### DIFF
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -12,13 +12,15 @@ use types::HasSqlType;
 
 #[allow(missing_debug_implementations, missing_copy_implementations)]
 pub struct MysqlConnection {
-    _raw_connection: RawConnection,
+    raw_connection: RawConnection,
     transaction_manager: AnsiTransactionManager,
 }
 
 impl SimpleConnection for MysqlConnection {
-    fn batch_execute(&self, _query: &str) -> QueryResult<()> {
-        unimplemented!()
+    fn batch_execute(&self, query: &str) -> QueryResult<()> {
+        self.raw_connection.enable_multi_statements(|| {
+            self.raw_connection.execute(query)
+        })
     }
 }
 
@@ -31,14 +33,15 @@ impl Connection for MysqlConnection {
         let connection_options = try!(ConnectionOptions::parse(database_url));
         try!(raw_connection.connect(&connection_options));
         Ok(MysqlConnection {
-            _raw_connection: raw_connection,
+            raw_connection: raw_connection,
             transaction_manager: AnsiTransactionManager::new(),
         })
     }
 
     #[doc(hidden)]
-    fn execute(&self, _query: &str) -> QueryResult<usize> {
-        unimplemented!()
+    fn execute(&self, query: &str) -> QueryResult<usize> {
+        self.raw_connection.execute(query)
+            .map(|_| self.raw_connection.affected_rows())
     }
 
     #[doc(hidden)]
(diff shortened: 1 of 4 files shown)


diesel/src/mysql/connection/raw.rs
```diff
@@ -68,6 +68,62 @@ impl RawConnection {
...
+            ffi::mysql_real_query(
+                self.0,
+                query.as_ptr() as *const libc::c_char,
+                query.len() as libc::c_ulong,
+            );
+        }
+        self.did_an_error_occur()
+    }
+
+    pub fn enable_multi_statements<T, F>(&self, f: F) -> QueryResult<T> where
```
killercup: This does some weird things. I think this is the place where I write

> Can you add a link to <https://dev.mysql.com/doc/refman/5.7/en/mysql-set-server-option.html>?

diesel/src/mysql/connection/mod.rs
```diff
@@ -12,13 +12,15 @@ use types::HasSqlType;
...
     transaction_manager: AnsiTransactionManager,
 }
 
 impl SimpleConnection for MysqlConnection {
-    fn batch_execute(&self, _query: &str) -> QueryResult<()> {
-        unimplemented!()
+    fn batch_execute(&self, query: &str) -> QueryResult<()> {
+        self.raw_connection.enable_multi_statements(|| {
+            self.raw_connection.execute(query)
+        })
```
killercup: In the section starting with "The following procedure outlines a suggested strategy for handling multiple statements", <https://dev.mysql.com/doc/refman/5.7/en/c-api-multiple-queries.html> talks about how you need to write a loop that calls `mysql_next_result` to actually get the results for multiple queries. Especially, it says

> For each iteration of the loop, handle the current statement result, retrieving either a result set or an affected-rows count. If an error occurs, exit the loop.

How would you interpret this? Do you maybe need to iterate the results to get all possible errors? (Currently, you are only calling `did_an_error_occur` after the initial execute and when setting the options.)

sgrif: Second paragraph of the commit message is addressing that case. I don't want to go further than the naive approach until I better understand the case it's describing and can test it.

sgrif:
> but I have not tested the code locally or have enough knowledge of MySQL to really be sure.

Well it seems to at least get to calls to `query_all` and `execute_returning_count` before blowing up now, so I think we're on the right track.

diesel/src/mysql/connection/mod.rs
killercup: Sounds good. Wasn't sure if "not getting all errors" is part of "all kinds of nasty side effects that may blow everything up" ;)

PR #625, opened 2017-02-05
Implement `MysqlConnection::execute_returning_count`

sgrif:
This adds the initial implementation of prepared statement support for
MySQL. This does not handle prepared statements which return values, nor
does it implement any form of prepared statement caching.

Prepared statements in MySQL operate on a structure called `MYSQL_BIND`
both for input and output. This thing is a fucking nightmare to work
with. We need to give it the structs in an array, which is behind a
pointer. We are responsible for allocating, managing, keeping alive, and
cleaning up the `MYSQL_BIND` structures. Additionally, it has several
fields which are also behind pointers, and we are also responsible for
keeping those values alive. Looking through their API docs, I see no
reasonable cause to have any of the fields except the buffer behind a
pointer. Still, that's the API we have to work with, so we have a bunch
of random vecs on the struct that only really exist so that we keep
these pointers alive.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -1,0 +1,72 @@
+extern crate mysqlclient_sys as ffi;
+
+use mysql::MysqlType;
+use std::mem;
+use std::os::{raw as libc};
+
+pub struct Binds {
+    data: Vec<Vec<u8>>,
+    lengths: Vec<libc::c_ulong>,
+    is_nulls: Vec<ffi::my_bool>,
+    mysql_binds: Vec<ffi::MYSQL_BIND>,
+}
+
+impl Binds {
+    pub fn from_input_data(input: Vec<(MysqlType, Option<Vec<u8>>)>) -> Self {
+        let is_nulls = input.iter()
+            .map(|&(_, ref data)| if data.is_none() { 1 } else { 0 })
+            .collect();
+        let (types, data): (Vec<_>, Vec<_>) = input.into_iter()
+            .map(|(t, data)| (t, data.unwrap_or(Vec::new()))).unzip();
+        let lengths = data.iter().map(|x| x.len() as libc::c_ulong).collect();
+
+        let mysql_binds = types.into_iter().map(|tpe| {
+            let mut bind: ffi::MYSQL_BIND = unsafe { mem::zeroed() };
+            bind.buffer_type = mysql_type_to_ffi_type(tpe);
+            bind
+        }).collect();
+
+        let mut res = Binds {
+            data: data,
+            lengths: lengths,
+            is_nulls: is_nulls,
+            mysql_binds: mysql_binds,
+        };
+        unsafe { res.link_mysql_bind_pointers(); }
+        res
+    }
+
+    pub fn mysql_binds(&mut self) -> &mut [ffi::MYSQL_BIND] {
+        &mut self.mysql_binds
+    }
+
+    // This function relies on the invariant that no further mutations to this
+    // struct will occur after this function has been called.
+    unsafe fn link_mysql_bind_pointers(&mut self) {
+        for (i, data) in self.data.iter_mut().enumerate() {
+            self.mysql_binds[i].buffer = data.as_mut_ptr() as *mut libc::c_void;
+            self.mysql_binds[i].buffer_length = data.capacity() as libc::c_ulong;
+            self.mysql_binds[i].length = &mut self.lengths[i];
+            self.mysql_binds[i].is_null = &mut self.is_nulls[i];
+        }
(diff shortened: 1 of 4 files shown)


diesel/src/mysql/connection/bind.rs
```diff
@@ -0,0 +1,74 @@
...
+impl Binds {
+    pub fn from_input_data(input: Vec<(MysqlType, Option<Vec<u8>>)>) -> Self {
+        let lengths = input.iter().map(|x| match *x {
+            (_, Some(ref data)) => data.len() as libc::c_ulong,
+            (_, None) => 0,
+        }).collect();
+        let is_nulls = input.iter()
+            .map(|&(_, ref data)| if data.is_none() { 1 } else { 0 })
+            .collect();
+        let (types, data) = input.into_iter().unzip::<_, _, Vec<_>, _>();
```
killercup: Might be easier to read like `let (types, data): (Vec<MysqlType>, Vec<Option<Vec<u8>>>) = …` instead of turbofish

diesel/src/mysql/connection/bind.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    // This function relies on the invariant that no further mutations to this
+    // struct will occur after this function has been called.
+    unsafe fn link_mysql_bind_pointers(&mut self) {
+        for (i, bind_data) in self.data.iter_mut().enumerate() {
+            if let Some(ref mut data) = bind_data.as_mut() {
+                self.mysql_binds[i].buffer = data.as_mut_ptr() as *mut libc::c_void;
+                self.mysql_binds[i].buffer_length = data.capacity() as libc::c_ulong;
+            }
```
killercup: What is the semantic meaning of the `None` case here? <https://dev.mysql.com/doc/refman/5.7/en/c-api-prepared-statement-data-structures.html> doesn't mention what a Null pointer means here. Is the None case only relevant when calling `mysql_stmt_bind_result()` and getting a SQL `NULL` result?

diesel/src/mysql/connection/mod.rs
```diff
@@ -60,8 +64,17 @@ impl Connection for MysqlConnection {
...
+    fn execute_returning_count<T>(&self, source: &T) -> QueryResult<usize> where
+        T: QueryFragment<Self::Backend> + QueryId,
+    {
+        let mut query_builder = MysqlQueryBuilder::new();
+        try!(source.to_sql(&mut query_builder).map_err(Error::QueryBuilderError));
+        let mut bind_collector = RawBytesBindCollector::<Mysql>::new();
+        try!(source.collect_binds(&mut bind_collector));
+        let mut stmt = try!(self.raw_connection.prepare(&query_builder.sql));
+        try!(stmt.bind(bind_collector.binds));
+        try!(stmt.execute());
```
killercup: I think this is one of those functions that look far nicer using `?`

diesel/src/mysql/connection/bind.rs
sgrif: `None` means that the bind parameter is going to be populated with SQL `NULL`. We're not dealing with `mysql_stmt_bind_result` yet, just `mysql_stmt_bind_param`.

Since the value of the buffer is `None`, `self.is_nulls[i]` will be `1`, and `self.lengths[i]` will be `0`. The value of `buffer` and `buffer_length` are both 0 since we initialized the struct with zeroed memory (I don't think any of that matters though as long as `self.is_nulls[i]` is non-zero)

sgrif: I think I may change the type of `data` to just be `Vec<Vec<u8>>` though, since we'll always want a buffer for the `mysql_stmt_bind_result` case. Since `Vec::new` doesn't allocate, it should be fine. (This will also make it easier to do an eventual refactoring that I'd like to do for `RawBytesBindCollector` where it puts everything in a flat `Vec<u8>` and then has a `Vec<Option<usize>>`)

Either way I won't make that change until we're dealing with `msyql_stmt_bind_result`.

diesel/src/mysql/connection/bind.rs
```diff
@@ -0,0 +1,74 @@
...
+            self.mysql_binds[i].length = &mut self.lengths[i];
+            self.mysql_binds[i].is_null = &mut self.is_nulls[i];
+        }
+    }
+}
+
+fn mysql_type_to_ffi_type(tpe: MysqlType) -> ffi::enum_field_types {
+    match tpe {
```
sgrif: Ugh, fine...

diesel/src/mysql/connection/bind.rs
sgrif: Actually I like the code better with that change now

PR #626, opened 2017-02-06
impl Sync for diesel::result::Error

quodlibetor:
This adds a Sync bound for the DatabaseErrorInformation trait object in Error.
Which seems fine since right now DEI is only implemented for String, but I
suppose I could imagine it being intended to hold a reference to some sort of
thread-local thing?

Looking at blame it seems like the lack of Sync was just an oversigh, but if
Sync isn't something you want to commit to and you have other ideas for how to
make result::Error more compatible with error-chain I'm happy to put in some
work to make it so.

Also, I can't make the tests compile locally for me either with or without this change, so I'm not super confident that I didn't miss anything.

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -9,7 +9,7 @@ use std::ffi::NulError;
 /// future without a major version bump.
 pub enum Error {
     InvalidCString(NulError),
-    DatabaseError(DatabaseErrorKind, Box<DatabaseErrorInformation+Send>),
+    DatabaseError(DatabaseErrorKind, Box<DatabaseErrorInformation+Send+Sync>),
     NotFound,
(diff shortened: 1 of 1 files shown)


sgrif:
> Which seems fine since right now DEI is only implemented for String

Actually, it's also implemented for `PgErrorInformation`, which is the main reason this trait exists in the first place. Still I think you're right that it's fine to require `Sync` here. The original implementation of `PgErrorInformation` was not `Sync`, but that changed in 7e378c48, where I specifically mention that the `unsafe impl Sync` is fine. I've double checked that there's no issue sending a `*mut PGresult` to another thread or sharing it, so this seems fine.

quodlibetor:
awesome, thanks!

PR #629, opened 2017-02-06
Implement prepared statements which return data for MySQL

sgrif:
After this change, the core crate's unit test suite is 100% green for
MySQL (`cd diesel && cargo test --features mysql`)

So here's how loading data works. Output values use the same data
structure as input values. For types with a fixed size, we can allocate
the small buffer needed, and everything will just work. For things that
we don't know the size of until we fetch the row, we set the buffer to
size 0.

When mysql hits a field where the buffer isn't large enough, the return
value of `mysql_stmt_fetch` will be `MYSQL_DATA_TRUNCATED`. For each
field that was truncated, the value of `*error` will be set to non-zero
(it's a `my_bool`, so presumably that value is 1. This appears to be the
only error. I do not know why they didn't call it `is_truncated`). It
will also set `*length` to the number of bytes required to hold the
value. At this point we can reallocate the buffer, and try to load that
column directly.

Aside from the general horribleness of this API that I've ranted about
elsewhere, two things in particular greatly bother me here:

- `MYSQL_DATA_TRUNCATED` and all other values mentioned as return values
  for `mysql_stmt_fetch` are `unsigned int`. `mysql_stmt_fetch` returns
  `int`.
- The MySQL client C API is clearly buffering this value on its own. Why
  the fuck can it not just give me that pointer, a byte count, and tell
  me to `memcpy` it before calling `mysql_stmt_fetch` again if I need
  it?!

Probably the worst part of this code is the implementation of
`StatementIterator::next`. Since we're actually just mutating the same
buffers over and over, I need to yield the same value each time. Of
course I can't just return `Self`, so I need to return a reference to
the appropriate value. There's not really any way to do this without
unsafe code. In particular what I need is a way to guarantee that the
value is unused by the next call to `next`, but I have no way to express
that. So I'm transmuting the reference to `'static`, and never exposing
this API. I know it is safe in the place this is currently used. In
theory we could at least ensure no use after free by implementing
iterator on a reference instead of the struct, so we have a smaller
lifetime. Even if we did that though, we're still mutating immutable
data, which is UB. So given that any code which holds onto that
reference past the next call to `next` is invoking UB no matter what, I
went with the smallest possible solution. I am open to recommendations
here.

### DIFF
--- a/diesel/src/mysql/backend.rs
+++ b/diesel/src/mysql/backend.rs
@@ -64,3 +64,24 @@ impl HasSqlType<::types::Timestamp> for Mysql {
         MysqlType::Timestamp
     }
 }
+
+use types::{ToSql, IsNull, FromSql};
+use std::error::Error as StdError;
+use std::io::Write;
+
+impl ToSql<::types::Bool, Mysql> for bool {
+    fn to_sql<W: Write>(&self, out: &mut W) -> Result<IsNull, Box<StdError+Send+Sync>> {
+        let int_value = if *self {
+            1
+        } else {
+            0
+        };
+        <i32 as ToSql<::types::Integer, Mysql>>::to_sql(&int_value, out)
+    }
+}
+
+impl FromSql<::types::Bool, Mysql> for bool {
+    fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<StdError+Send+Sync>> {
+        Ok(not_none!(bytes).iter().any(|x| *x != 0))
+    }
+}
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -9,6 +9,7 @@ pub struct Binds {
     lengths: Vec<libc::c_ulong>,
     is_nulls: Vec<ffi::my_bool>,
     mysql_binds: Vec<ffi::MYSQL_BIND>,
+    errors: Option<Vec<ffi::my_bool>>,
 }
 
 impl Binds {
@@ -31,6 +32,36 @@ impl Binds {
             lengths: lengths,
             is_nulls: is_nulls,
             mysql_binds: mysql_binds,
+            errors: None,
+        };
+        unsafe { res.link_mysql_bind_pointers(); }
+        res
+    }
+
+    pub fn from_output_types<Iter>(types: Iter) -> Self where
+        Iter: IntoIterator<Item=ffi::enum_field_types>,
+    {
+        let mysql_binds = types.into_iter().map(|tpe| {
+            let mut bind: ffi::MYSQL_BIND = unsafe { mem::zeroed() };
+            bind.buffer_type = tpe;
+            bind
+        }).collect::<Vec<_>>();
+        let is_nulls = vec![0; mysql_binds.len()];
+        let errors = is_nulls.clone();
+        let data = mysql_binds.iter().map(|bind| {
+            match known_buffer_size_for_ffi_type(bind.buffer_type) {
+                Some(size) => vec![0; size as usize],
+                None => Vec::new(),
+            }
+        }).collect::<Vec<_>>();
+        let lengths = data.iter().map(|x| x.len() as libc::c_ulong).collect();
+
+        let mut res = Binds {
+            data: data,
+            lengths: lengths,
+            is_nulls: is_nulls,
+            mysql_binds: mysql_binds,
+            errors: Some(errors),
         };
         unsafe { res.link_mysql_bind_pointers(); }
         res
@@ -38,6 +69,56 @@ impl Binds {
 
     pub fn mysql_binds(&mut self) -> &mut [ffi::MYSQL_BIND] {
         &mut self.mysql_binds
+    }
+
+    /// The caller of this function must immediately check for errors
+    /// after return
+    pub fn populate_dynamic_buffers(&mut self, stmt: *mut ffi::MYSQL_STMT) {
+        use std::cmp::min;
+
+        for (i, buffer) in self.data.iter_mut().enumerate() {
+            if self.errors.as_ref().unwrap()[i] != 0 {
+                let written_bytes = buffer.len();
+                let needed_capacity = self.lengths[i] as usize;
+                buffer.reserve(needed_capacity - written_bytes);
+                self.mysql_binds[i].buffer = buffer.as_mut_ptr() as *mut libc::c_void;
+                mem::swap(&mut self.lengths[i], &mut self.mysql_binds[i].buffer_length);
+                let load_result = unsafe {
+                    ffi::mysql_stmt_fetch_column(
+                        stmt,
+                        &mut self.mysql_binds[i],
+                        i as libc::c_uint,
+                        written_bytes as libc::c_ulong,
+                    )
+                };
+                if load_result != 0 {
+                    return;
+                }
+            }
+
+            let actual_bytes_in_buffer = min(
+                self.mysql_binds[i].buffer_length,
+                self.lengths[i],
+            );
+            unsafe { buffer.set_len(actual_bytes_in_buffer as usize) }
+        }
+    }
+
+    pub fn reset_dynamic_buffers(&mut self) {
+        for (i, bind) in self.mysql_binds.iter().enumerate() {
+            if known_buffer_size_for_ffi_type(bind.buffer_type).is_none() {
+                self.lengths[i] = 0;
+                unsafe { self.data[i].set_len(0) };
+            }
+        }
+    }
+
+    pub fn field_data(&self, idx: usize) -> Option<&[u8]> {
+        if self.is_nulls[idx] == 0 {
+            Some(&*self.data[idx])
+        } else {
+            None
+        }
     }
 
     // This function relies on the invariant that no further mutations to this
@@ -48,6 +129,10 @@ impl Binds {
             self.mysql_binds[i].buffer_length = data.capacity() as libc::c_ulong;
             self.mysql_binds[i].length = &mut self.lengths[i];
             self.mysql_binds[i].is_null = &mut self.is_nulls[i];
+
+            if let Some(ref mut errors) = self.errors {
+                self.mysql_binds[i].error = &mut errors[i];
+            }
         }
     }
 }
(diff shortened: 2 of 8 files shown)


diesel/src/mysql/connection/bind.rs
```diff
@@ -40,6 +71,56 @@ impl Binds {
...
+                if load_result != 0 {
+                    return;
+                }
+            }
+
+            let actual_bytes_in_buffer = min(
+                self.mysql_binds[i].buffer_length,
+                self.lengths[i],
+            );
+            unsafe { buffer.set_len(actual_bytes_in_buffer as usize) }
```
killercup: Okay, I'm probably too tired to see this. Can _this_ `set_len` leak memory? Can you use `truncate` instead? Is this what you meant earlier? (The docs don't say much about `truncate` but it's source uses `drop_in_place` which may call destructors we don't want to run.)

diesel/src/mysql/connection/bind.rs
```diff
@@ -40,6 +71,56 @@ impl Binds {
...
+        }
+    }
+
+    pub fn reset_dynamic_buffers(&mut self) {
+        for (i, bind) in self.mysql_binds.iter().enumerate() {
+            if known_buffer_size_for_ffi_type(bind.buffer_type).is_none() {
+                self.lengths[i] = 0;
+                unsafe { self.data[i].set_len(0) };
```
killercup: Same with truncate as above

diesel/src/mysql/connection/bind.rs
```diff
@@ -70,3 +155,25 @@ fn mysql_type_to_ffi_type(tpe: MysqlType) -> ffi::enum_field_types {
...
+        enum_field_types::MYSQL_TYPE_LONGLONG |
+        enum_field_types::MYSQL_TYPE_DOUBLE => Some(8),
+        enum_field_types::MYSQL_TYPE_TIME |
+        enum_field_types::MYSQL_TYPE_DATE |
+        enum_field_types::MYSQL_TYPE_DATETIME |
+        enum_field_types::MYSQL_TYPE_TIMESTAMP =>
+            Some(size_of::<ffi::MYSQL_TIME>() as libc::c_ulong),
+        _ => None,
+    }
+}
```
killercup: Memo to self: Continue from here

diesel/src/mysql/backend.rs
```diff
@@ -64,3 +64,24 @@ impl HasSqlType<::types::Timestamp> for Mysql {
...
+        } else {
+            0
+        };
+        <i32 as ToSql<::types::Integer, Mysql>>::to_sql(&int_value, out)
+    }
+}
+
+impl FromSql<::types::Bool, Mysql> for bool {
+    fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<StdError+Send+Sync>> {
+        Ok(not_none!(bytes)[0] != 0)
```
sgrif: Yeah, boolean expressions in MySQL are transmitted as `MYSQL_TYPE_TINY` which corresponds to `signed char`. For codegen I will infer `TINYINT(1)` to mean bool, yes. I don't really want to define/care about the behavior of a > 1 byte type being used here, but I think your code is actually easier to read anyway

diesel/src/mysql/connection/bind.rs
```diff
@@ -9,6 +9,7 @@ pub struct Binds {
     lengths: Vec<libc::c_ulong>,
     is_nulls: Vec<ffi::my_bool>,
     mysql_binds: Vec<ffi::MYSQL_BIND>,
+    errors: Option<Vec<ffi::my_bool>>,
```
sgrif: Hm interesting point. If nothing else, `mysql_binds` will always need to be in its own vec.

diesel/src/mysql/connection/bind.rs
```diff
@@ -31,6 +32,36 @@ impl Binds {
...
+
+    pub fn from_output_types<Iter>(types: Iter) -> Self where
+        Iter: IntoIterator<Item=ffi::enum_field_types>,
+    {
+        let mysql_binds = types.into_iter().map(|tpe| {
+            let mut bind: ffi::MYSQL_BIND = unsafe { mem::zeroed() };
+            bind.buffer_type = tpe;
+            bind
+        }).collect::<Vec<_>>();
+        let is_nulls = vec![0; mysql_binds.len()];
```
sgrif: Type inference is doing that.

diesel/src/mysql/connection/bind.rs
```diff
@@ -40,6 +71,56 @@ impl Binds {
         &mut self.mysql_binds
     }
 
+    /// The caller of this function must immediately check for errors
+    /// after return
+    pub fn populate_dynamic_buffers(&mut self, stmt: *mut ffi::MYSQL_STMT) {
+        use std::cmp::min;
+
+        for (i, buffer) in self.data.iter_mut().enumerate() {
+            if self.errors.as_ref().unwrap()[i] != 0 {
```
sgrif: Errors is none if this is an input bind, some if it's an output bind. This method makes no sense in the context of an output bind. Honestly, I should probably just separate these into two structs. There's basically no shared code other than that one function to link up the pointers.

diesel/src/mysql/connection/bind.rs
```diff
@@ -40,6 +71,56 @@ impl Binds {
...
+                let written_bytes = buffer.len();
+                let needed_capacity = self.lengths[i] as usize;
+                buffer.reserve(needed_capacity - written_bytes);
```
sgrif: The only time we'd be in this branch where that wasn't true was a bug in the MySQL C library. Since there's no memory safety violation if that bug occurred (it'd probably just abort with OOM), I don't think we need to defend against it here.

diesel/src/mysql/connection/bind.rs
```diff
@@ -40,6 +71,56 @@ impl Binds {
...
+                self.mysql_binds[i].buffer = buffer.as_mut_ptr() as *mut libc::c_void;
```
sgrif: Was true before this as well. This is needed because the call to reserve above is invalidating the previous pointer (Forgetting this line was the cause of the segfault I mentioned in gitter)

diesel/src/mysql/connection/bind.rs
killercup: That's it, I'm going to bed 😆

diesel/src/mysql/connection/bind.rs
sgrif: `set_len` can never leak memory. It only affects whether destructors are run for the values in the vec. In this case, the values are `u8` so it's meaningless. The amount of memory used is defined by capacity, which we don't touch. I don't want to use `truncate` or `shrink_to_fit`, because we'll likely just have to re-allocate on the next row. This code re-uses the buffer for each query run (deallocating after the last row is processed)

diesel/src/mysql/connection/bind.rs
sgrif: Same answer

diesel/src/mysql/connection/stmt/iterator.rs
```diff
@@ -0,0 +1,88 @@
...
+            );
+        }
+        stmt.did_an_error_occur()?;
+
+        Ok(StatementIterator {
+            stmt: stmt,
+            output_binds: output_binds,
+        })
+    }
+}
```
killercup: LGTM

diesel/src/mysql/connection/bind.rs
```diff
@@ -70,3 +155,25 @@ fn mysql_type_to_ffi_type(tpe: MysqlType) -> ffi::enum_field_types {
         MysqlType::Blob => MYSQL_TYPE_BLOB,
     }
 }
+
+fn known_buffer_size_for_ffi_type(tpe: ffi::enum_field_types) -> Option<libc::c_ulong> {
+    use std::mem::size_of;
+    use self::ffi::enum_field_types;
```
sgrif: Not for this one. `enum_field_types::MYSQL_TYPE_TINY => stuff()` will run stuff if the value is equal to `MYSQL_TYPE_TINY`. `MYSQL_TYPE_TINY => stuff()` will run stuff for all possible values (and also fail to compile since we have warnings disabled and using a catch-all ident which matches an imported name in a pattern warns)

diesel/src/mysql/connection/result.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub struct ResultFields<'a>(&'a mut RawResult);
+
+impl<'a> Iterator for ResultFields<'a> {
+    type Item = &'a ffi::MYSQL_FIELD;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        unsafe {
+            ffi::mysql_fetch_field((self.0).0)
+                .as_ref()
```
sgrif: The lifetime is the `'a` referenced in the `Self::Item` type, which is the lifetime of the `RawResult` reference we are holding onto, which is the actual correct lifetime. The fields are fully disjoint, so we don't need any sort of streaming iterator abstraction here.

diesel/src/mysql/connection/bind.rs
killercup: Fine. You win this ti— wait! Not so fast! You can still `use self::ffi::enum_field_types as t;` to reduce all the visual noise ;)

diesel/src/mysql/connection/stmt/iterator.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Row<Mysql> for MysqlRow {
+    fn take(&mut self) -> Option<&[u8]> {
+        let current_idx = self.col_idx;
+        self.col_idx += 1;
+        self.binds.field_data(current_idx)
+    }
+
+    fn next_is_null(&self, count: usize) -> bool {
+        (0..count).all(|i| self.binds.field_data(self.col_idx + i).is_none())
```
sgrif: I personally think this way is easier to read, but I don't feel strongly. I should change the impls of `SqliteRow` and `PgRow` to use that form as well, so I'll do it in a separate commit.

diesel/src/mysql/connection/bind.rs
sgrif: Fine. :P

diesel/src/mysql/connection/bind.rs
sgrif: I'm going to explore this, but if I end up going that way it'll be a separate PR as this one is large enough as it is.

diesel/src/mysql/connection/stmt/iterator.rs
```diff
@@ -36,14 +33,18 @@ impl<'a> StatementIterator<'a> {
...
-        use std::mem::transmute;
+    pub fn map<F, T>(mut self, mut f: F) -> QueryResult<Vec<T>> where
+        F: FnMut(MysqlRow) -> QueryResult<T>,
+    {
+        let mut results = Vec::new();
+        while let Some(row) = self.next() {
+            results.push(f(row?)?);
+        }
+        Ok(results)
+    }
```
killercup: Heh, life can be so simple

sgrif:
> Let's hope we can deal with MySQl on a higher level of abstraction from now on ;)

Other than adding some `ToSql`/`FromSql` impls, yeah the rest of the interaction should be through the query builder.

PR #630, opened 2017-02-06
Update to chrono 0.3

Eijebong:
I just followed the changelog here https://github.com/chronotope/chrono/blob/master/CHANGELOG.md and fixed every compile error.

I can't run all tests locally because I get a SIGILL before the end but hopefully this doesn't break anything.


sgrif:

Eijebong:
Sure

PR #633, opened 2017-02-07
Refactor MySQL bind params to use vec-of-structs

sgrif:
This idea was raised in https://github.com/diesel-rs/diesel/pull/629#discussion_r99691942.
The answer appears to be yes, the code is much better this way. It
allows me to much more easily break things into small functions in
places where I couldn't before because the borrow checker couldn't see
the disjointness through function boundaries.

As part of writing this I wrote a few bugs (which are fixed), which gave
me a better understanding of what MySQL is actually doing. It turns out
our changes to the binds after the call to `mysql_stmt_bind_result`
aren't seen because MySQL memcpys the binds. We'd actually have a memory
safety violation if we ever had a case where the buffer was non-zero in
initially, but not large enough to hold the value. Luckily that cannot
be the case as we've written this, but I will refactor the code further
with this new understanding in a separate commit.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -5,34 +5,22 @@ use std::mem;
 use std::os::{raw as libc};
 
 pub struct Binds {
-    data: Vec<Vec<u8>>,
-    lengths: Vec<libc::c_ulong>,
-    is_nulls: Vec<ffi::my_bool>,
+    data: Vec<BindData>,
     mysql_binds: Vec<ffi::MYSQL_BIND>,
-    errors: Option<Vec<ffi::my_bool>>,
 }
 
 impl Binds {
     pub fn from_input_data(input: Vec<(MysqlType, Option<Vec<u8>>)>) -> Self {
-        let is_nulls = input.iter()
-            .map(|&(_, ref data)| if data.is_none() { 1 } else { 0 })
-            .collect();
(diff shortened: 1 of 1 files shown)


diesel/src/mysql/connection/bind.rs
```diff
@@ -74,65 +51,124 @@ impl Binds {
...
-
-        for (i, buffer) in self.data.iter_mut().enumerate() {
-            if self.errors.as_ref().unwrap()[i] != 0 {
-                let written_bytes = buffer.len();
-                let needed_capacity = self.lengths[i] as usize;
-                buffer.reserve(needed_capacity - written_bytes);
-                self.mysql_binds[i].buffer = buffer.as_mut_ptr() as *mut libc::c_void;
-                mem::swap(&mut self.lengths[i], &mut self.mysql_binds[i].buffer_length);
+        for (i, data) in self.data.iter_mut().enumerate() {
+            let bind = &mut self.mysql_binds[i];
```
sgrif: This could also be written as `for (i, (data, bind))` in self.data.iter_mut().zip(&mut self.mysql_binds).enumerate()`. Thoughts on which is nicer?

sgrif: Doesn't matter, I end up deleting this in the next commit.

sgrif:
Going to merge this as the follow up is more in need of code review than this is.

PR #635, opened 2017-02-07
Let MySQL try to reuse dynamic buffers

sgrif:
With the previous code, MySQL thought that all dynamic buffers were 0
sized, forcing us to re-fetch the value on every row. This is because
the changes we were making to the `MYSQL_BIND` structure were never
seen, as MySQL is `memcpy`ing the values when passed to
`mysql_stmt_bind_result`.

With this knowledge, I've changed the code to not bother keeping those
structs alive, and re-bind after any of the old binds are invalidated.
This also means that for data that is partially truncated, we are able
to do a smaller fetch which only gives back the truncated data, rather
than re-copying the entire field.

This code looks more unsafe than what was there before, but the old code
was actually doing things that were fundamentally wrong WRT memory
safety, avoiding segfaults only because the initial buffer size was 0.
This code isn't any more unsafe, it just makes the unsafety more clear.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -1,97 +1,75 @@
 extern crate mysqlclient_sys as ffi;
 
-use mysql::MysqlType;
 use std::mem;
 use std::os::{raw as libc};
 
+use mysql::MysqlType;
+use result::QueryResult;
+use super::stmt::Statement;
+
 pub struct Binds {
     data: Vec<BindData>,
-    mysql_binds: Vec<ffi::MYSQL_BIND>,
 }
 
 impl Binds {
     pub fn from_input_data(input: Vec<(MysqlType, Option<Vec<u8>>)>) -> Self {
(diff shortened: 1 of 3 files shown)


diesel/src/mysql/connection/bind.rs
```diff
@@ -170,6 +136,33 @@ impl BindData {
...
+    ///
+    /// This invalidates the bind previously returned by `mysql_bind`. Calling
+    /// this function is unsafe unless the binds are immediately rebound.
+    unsafe fn bind_for_truncated_data(&mut self) -> Option<(ffi::MYSQL_BIND, usize)> {
+        if self.is_truncated() {
+            let offset = self.bytes.capacity();
+            let truncated_amount = self.length as usize - offset;
+
+            debug_assert!(truncated_amount > 0, "output buffers were invalidated \
+                without calling `mysql_stmt_bind_result`");
```
killercup: (This catches `truncated_amount == 0`, underflow already triggers an assert in debug)

PR #636, opened 2017-02-07
Properly load tables from `infer_table_from_schema!`

sgrif:
This bug was introduced by #577, which completely ignores any table
names passed to it which specify a custom schema. This bug slipped
through because that same PR also changed the implementation of
`infer_schema!` to not call `infer_table_from_schema!`. We didn't have
an explicit test for both macros in this case, since we usually
structure our code to reuse itself as much as possible, and fully
exercise as many code paths as we can so we don't have to consider every
combination in every place all the time.

This reverts to the original behavior of that macro, and adds a more
explicit test so it does not happen again.


sgrif:
I rebased poorly. https://github.com/diesel-rs/diesel/commit/e0272a4398444666ea7a62b2cf1e6a98003fd17e was the commit this added.

PR #637, opened 2017-02-07
Refactor PG's schema inference code to be (almost) ANSI

sgrif:
`information_schema` is in fact part of the ANSI standard, and most
backends conform to it, including MySQL. This will allow us to reuse
this code for MySQL and most likely any other future backends we add.
SQLite does not provide an equivalent of `information_schema`. The on
thing that we cannot use the standard for is column type. While there is
an `information_schema.columns.data_type` column, it doesn't provide
what we need. Both PG and MySQL provide a nonstandard column that we can
use instead.


sgrif:
https://en.wikipedia.org/wiki/Information_schema

This code should work with minimal changes for everyone except Oracle, DB2, and SQLite.

PR #639, opened 2017-02-07
Add support for MySQL to diesel_codegen

sgrif:
This provides the minimum amount of code required for migrations to
successfully run when compiling the integration tests. I had a few bugs
in output buffer code as well as `batch_execute` which have been fixed
here as well.


diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -50,3 +50,25 @@ impl<T, ST> SaveChangesDsl<SqliteConnection, ST> for T where
...
     }
 }
+
+#[cfg(feature = "mysql")]
+use mysql::{MysqlConnection, Mysql};
+#[cfg(feature = "mysql")]
+use associations::Identifiable;
+
+#[cfg(feature = "mysql")]
+impl<T, ST> SaveChangesDsl<MysqlConnection, ST> for T where
```
sgrif: This is literally the same code as the sqlite impl, but I don't think it's worth making a macro for 2 cases.

diesel_infer_schema/src/information_schema.rs
```diff
@@ -77,6 +89,12 @@ pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box
         &attr.type_name
     };
 
+    let tpe = if let Some(idx) = tpe.find('(') {
```
sgrif: I'm going to pull out separate `determine_column_type` functions for MySQL and PG, as there's several cases I want to handle. This is required to turn things like `int(11)` into `int` for now.

diesel_tests/build.rs
```diff
@@ -24,13 +24,26 @@ fn connection() -> SqliteConnection {
     SqliteConnection::establish(&database_url).unwrap()
 }
 
+#[cfg(feature = "mysql")]
+use self::diesel::mysql::MysqlConnection;
+#[cfg(feature = "mysql")]
+fn connection() -> MysqlConnection {
+    dotenv().ok();
+    let database_url = ::std::env::var("MYSQL_DATABASE_URL")
```
sgrif: I'm going to apply this pattern to the entire codebase soon. One env var is great for CI, terrible for local development.

diesel_infer_schema/src/information_schema.rs
killercup: This whole `determine_column_type` function got even weirder just now. Pulling it put might be a good idea. While you're at it, can you also add a test or two for its behavior (detecting arrays, things like`int(11)`, and the combination of the two if applicable)?

migrations/mysql/20170207193805_initial_schema/up.sql
```diff
@@ -0,0 +1,36 @@
...
+CREATE TABLE comments (
+  id INTEGER PRIMARY KEY AUTO_INCREMENT,
+  post_id INTEGER NOT NULL,
+  text TEXT NOT NULL
+) CHARACTER SET utf8mb4;
+
+CREATE TABLE special_posts (
+  id INTEGER PRIMARY KEY AUTO_INCREMENT,
+  user_id INTEGER NOT NULL,
+  title TEXT NOT NULL
```
killercup: Make that a MEDIUMTEXT and add a test which adds exactly 64k+1 bytes of text :trollface:

diesel/src/migrations/mod.rs
```diff
@@ -163,7 +163,7 @@ pub fn setup_database<Conn: Connection>(conn: &Conn) -> QueryResult<usize> {
 fn create_schema_migrations_table_if_needed<Conn: Connection>(conn: &Conn) -> QueryResult<usize> {
     conn.silence_notices(|| {
         conn.execute("CREATE TABLE IF NOT EXISTS __diesel_schema_migrations (
-            version VARCHAR PRIMARY KEY NOT NULL,
+            version VARCHAR(255) PRIMARY KEY NOT NULL,
```
killercup: I hope no one ever uses a name with ≥256 chars and sees a weird error.

diesel/src/mysql/connection/bind.rs
```diff
@@ -48,6 +48,8 @@ impl Binds {
             unsafe {
                 if let Some((mut bind, offset)) = data.bind_for_truncated_data() {
                     stmt.fetch_column(&mut bind, i, offset)?
+                } else {
+                    data.update_buffer_length()
```
killercup: Sanity restored. ✔️

diesel_infer_schema/src/inference.rs
```diff
@@ -36,16 +42,28 @@ pub fn establish_connection(database_url: &str) -> Result<InferConnection, Box<E
...
             establish_real_connection(database_url).map(InferConnection::Pg)
         }
+        #[cfg(feature = "mysql")]
+        _ if database_url.starts_with("mysql://") => {
+            establish_real_connection(database_url).map(InferConnection::Mysql)
+        }
         #[cfg(feature = "sqlite")]
         _ => establish_real_connection(database_url).map(InferConnection::Sqlite),
-        #[cfg(not(feature = "sqlite"))]
+        #[cfg(all(feature = "postgres", not(feature = "sqlite")))]
```
sgrif: We still need one to win if both are enabled. I can probably make the error message more generic and just have it be `not(cfg(feature = "sqlite"))` in the future.

diesel/src/migrations/mod.rs
sgrif: I have to put something here. MySQL doesn't allow `TEXT` PKs (because reasons I guess?). 255 has always been the go-to size since the length fits in one byte. Are you suggesting we go up a power of two?

sgrif:
Not quite yet. Still need to get the test suite fully passing locally, and a bit more cleanup. Soon though.

diesel/src/migrations/mod.rs
killercup: Nah, just have a test and a noncryptic error message. Maybe. Later. It's not critical.

sgrif: :+1:

sgrif: Fun fact: By default, it won't even error in MySQL. It'll just silently truncate the data with no indication that anything is wrong.

sgrif: (The `(255)` does nothing in SQLite and PG.)

PR #641, opened 2017-02-08
Don't panic while panicking in CLI's tests

sgrif:
Diesel CLI's tests will abort with no output if the tests for `diesel
database create` fail to drop the database. (This is actually failing
for many people because the code to drop it has diverged from CLI's
behavior, which will be fixed separately).

This change ensures that none of the `Drop` impls in our test suite will
abort the process.

### DIFF
--- a/diesel_cli/tests/support/command.rs
+++ b/diesel_cli/tests/support/command.rs
@@ -66,6 +66,15 @@ impl CommandResult {
     pub fn code(&self) -> i32 {
         self.output.status.code().unwrap()
     }
+
+    #[allow(dead_code)]
+    pub fn result(self) -> Result<Self, Self> {
+        if self.is_success() {
+            Ok(self)
+        } else {
+            Err(self)
+        }
+    }
 }
 
 fn path_to_diesel_cli() -> PathBuf {
--- a/diesel_cli/tests/support/mod.rs
+++ b/diesel_cli/tests/support/mod.rs
@@ -1,3 +1,18 @@
+macro_rules! try_drop {
+    ($e:expr, $msg:expr) => { match $e {
+        Ok(x) => x,
+        Err(e) => {
+            use ::std::io::{Write, stderr};
+            if ::std::thread::panicking() {
+                write!(stderr(), "{}: {:?}", $msg, e);
+                return;
+            } else {
+                panic!("{}: {:?}", $msg, e);
+            }
+        }
+    }}
+}
+
 mod command;
 mod project_builder;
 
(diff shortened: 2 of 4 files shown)


sgrif:
Thanks @dnagir for making me aware of this issue.

diesel_cli/tests/support/mod.rs
```diff
@@ -1,3 +1,18 @@
+macro_rules! try_drop {
+    ($e:expr, $msg:expr) => { match $e {
+        Ok(x) => x,
+        Err(e) => {
+            use ::std::io::{Write, stderr};
+            if ::std::thread::panicking() {
```
killercup: FTR, don't compile these tests with `panic=abort`.

diesel_cli/tests/support/command.rs
```diff
@@ -66,6 +66,15 @@ impl CommandResult {
...
     }
+
+    #[allow(dead_code)]
+    pub fn result(self) -> Result<Self, Self> {
+        if self.is_success() {
+            Ok(self)
+        } else {
+            Err(self)
+        }
+    }
```
killercup: I've seen a few lines like these in the last few days. Maybe it is time to add the glorious [boolinator](https://crates.io/crates/boolinator) as a dependency?

diesel_cli/tests/support/postgres_database.rs
```diff
@@ -54,9 +54,9 @@ impl Database {
 impl Drop for Database {
     fn drop(&mut self) {
         let (database, postgres_url) = self.split_url();
-        let conn = PgConnection::establish(&postgres_url).unwrap();
+        let conn = try_drop!(PgConnection::establish(&postgres_url), "Couldn't connect to database");
         conn.silence_notices(|| {
-            conn.execute(&format!("DROP DATABASE IF EXISTS {}", database)).unwrap();
+            try_drop!(conn.execute(&format!("DROP DATABASE IF EXISTS {}", database)), "Couldn't drop database");
         });
```
killercup: Memo to self: Open clippy issue for denying unwrap (and other panic-able fns) in drop

killercup:
Travis [fails](https://travis-ci.org/diesel-rs/diesel/jobs/199591525#L747-L749) in `information_schema::tests::load_table_names_loads_from_custom_schema` because the tables are not in the expected order.

diesel_cli/tests/support/mod.rs
sgrif: Since tests fail by panicking, that would be bad in any project (I'll bet `cargo test` sets `panic=unwind` even if the project is configured with `panic=abort`)

sgrif:
Rerunning. Will make it deterministic in a separate PR.

PR #642, opened 2017-02-08
Properly drop databases in CLI's tests

sgrif:
The behavior of `diesel cli drop` is to connect to the `postgres`
database, which is known to exist. The test suite for `diesel database
setup` is not specifying a database, which connects to a database with
the same name as the PG user (which deafults to the local username).
This brings the behavior of the two back in line. In theory these two
should be code between them at least for the database URL munging, but
the tests live in a separate crate out of necessity and I don't want to
deal with exposing CLI as a library crate to fix this minor test issue.


sgrif:
Thanks @dnagir for making me aware of this issue.

PR #644, opened 2017-02-08
Use backend specific env vars if present for tests

sgrif:
Having a single `DATABASE_URL` is really nice in contexts where you're
likely only using one backend. It makes a lot of sense for Travis, which
runs against one backend at a time. It also makes sense for CLI to
continue to default to it by convention, as it has to look at the URL to
figure out which backend to use.

However, for development on Diesel itself, we are switching between
different backends all the time. This didn't matter as much for SQLite,
since we can just give it some random file path and have it work (and
we're usually connecting to `:memory:` anyway). For MySQL though, we
need to be more concrete. Where applicable, our test suite will now look
for `PG_DATABASE_URL`, before falling back to just `DATABASE_URL`. This
does not apply to our examples, as they assume PG and set the
`DATABASE_URL` manually in `examples/test_all`. This also doesn't apply
to CLI's test suite, which sets `DATABASE_URL` manually as needed, since
it can't run in a transaction and thus needs a different database
per-test.

So developers who frequently run our test suite against more than one
backend will likely have a `.env` file which looks like:

```
PG_DATABASE_URL=postgres://localhost/diesel_test
MYSQL_DATABASE_URL=mysql://localhost/diesel_test
SQLITE_DATABASE_URL=/tmp/test.db
```

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,7 +2,7 @@
 set -e
 
 if [ "$1" = "integration" ] && [ "$2" = "sqlite" ]; then
-  (cd diesel_tests && DATABASE_URL=/tmp/test.db cargo test --features "sqlite" --no-default-features)
+  (cd diesel_tests && cargo test --features "sqlite" --no-default-features)
 elif [ "$1" = "integration" ]; then
   (cd diesel_tests && cargo test --features "postgres" --no-default-features)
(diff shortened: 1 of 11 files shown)


sgrif:
Agreed. I'll add it to the todo list.

PR #645, opened 2017-02-08
Ensure MySQL schema inference loads tables when no schema given

sgrif:
Reminder: MySQL has no concept of schemas, and calls databases schemas
instead. All connections have access to all databases on that server
that the user has permission to access.

On PG, the default schema is always simply called `public`. For MySQL,
it'll be based on the connection URL. In theory we could actually get
this information without executing a query, but that would require
adding a method to `Connection` explicitly for this which I don't want
to do. I'm fine with just executing a query.

As an aside, I really want to find a way to require `String:
FromSql<types::Text, Self>` as a constraint of `Backend` without forcing
that to be specified *every* time I do `DB: Backend`.

I haven't written a test case for this, since I still need to figure out
what we're going to do about tests which need to mutate schema without
fucking with each other.

### DIFF
--- a/diesel_infer_schema/src/information_schema.rs
+++ b/diesel_infer_schema/src/information_schema.rs
@@ -20,6 +20,9 @@ pub trait UsesInformationSchema: Backend {
     > + NonAggregate + QueryId + QueryFragment<Self>;
 
     fn type_column() -> Self::TypeColumn;
+    fn default_schema<C>(conn: &C) -> QueryResult<String> where
+        C: Connection,
+        String: FromSql<types::Text, C::Backend>;
 }
 
 #[cfg(feature="postgres")]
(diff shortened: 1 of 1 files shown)


diesel_infer_schema/src/information_schema.rs
```diff
@@ -29,6 +32,10 @@ impl UsesInformationSchema for Pg {
     fn type_column() -> Self::TypeColumn {
         self::information_schema::columns::udt_name
     }
+
+    fn default_schema<C>(_conn: &C) -> QueryResult<String> {
+        Ok("public".into())
```
killercup: I'd love to see a `Cow` here. Not really important for the overall picture though.

sgrif: If this were public API or if it were anywhere that we cared the slightest bit about performance I would do that.

PR #649, opened 2017-02-09
cfg out of tests that use `get_result` for MySQL

sgrif:
MySQL does not support returning clauses.


sgrif:
Good idea. I'll do that as a future improvement.

PR #652, opened 2017-02-09
Have MySQL convert values to the expected type when possible

sgrif:
This addresses some of the main differences in behavior that we've found
between MySQL and other backends so far

- Integer literals are of type `BigInteger` not `Integer`
- Addition, subtraction, and multiplication between integer types always
  returns `BigInteger`, not the type being operated on
- Division between integer types returns `Numeric`, not the type being
  operated on
- Mathmatical operations between float types always returns `Numeric`,
  not the type being operated on

While integer literals only really matter for our tests, the difference
in mathematical operations is a problem for us. The setup of
`Expression` and `AsExpression` don't vary by backend, nor do we want
them to. Type checking of queries occurs before we know what backend
we're using, and should continue to remain entirely independent.

So we need to coerce MySQL into behaving similarly to other databases.
By setting the value of `buffer_type` to what we expected to get back,
rather than what MySQL is sending over the wire, we cause the C API to
perform some amount of conversion automatically under the hood. If the
conversion is invalid, we will end up getting an error on the call to
`mysql_bind_result`. This shouldn't result in an error for any code
using the query builder which compiles successfully.

That said, there are three caveats to this change:

- If overflow occurs, we'll end up returning an error to the user. This
  mirrors the behavior of PG, but it's worth noting that we won't
  silently lose data.
- We lose the benefits of the `debug_assert!`s involving the number of
  bytes and integers. This was added to help catch common mistakes from
  misuse of `sql` and `sql_function!`. Users of MySQL will now only see
  an error if the actual return values overflow. However, if we're ok
  with this for numeric operators, I think we should be fine with it in
  general.
- While integer division will behave the same as other backends with
  regards to types, the actual value might differ if it's used in a more
  complex expression. For example `select(1.as_expression() / 2 * 2)`
  will return 0 on PG, but 1 on MySQL. We could change the
  implementation to use the `DIV` operator on MySQL, but I think that
  behavior would be even more confusing.


diesel_tests/tests/expressions/ops.rs
```diff
@@ -17,6 +17,16 @@ fn adding_literal_to_column() {
...
 
 #[test]
+#[cfg(not(feature="sqlite"))] // FIXME: Does SQLite provide a way to detect overflow?
+fn overflow_returns_an_error_but_does_not_panic() {
+    use schema::users::dsl::*;
+
+    let connection = connection_with_sean_and_tess_in_users_table();
+    let query_result = users.select(id + i32::max_value()).load::<i32>(&connection);
+    assert!(query_result.is_err(), "Integer overflow should have returned an error");
+}
```
killercup: For consistency, you might also want a test for detecting integer underflow

PR #654, opened 2017-02-09
Detect unique constraint violations for MySQL

sgrif:
MySQL doesn't expose a specific API to get information about the error
other than the code, message, and SQLSTATE value. We could potentially
get the constraint name by parsing the error message, but that is
going to be fragile at best. If someone wants to rely on that fragile
method of determining the constraint name, they should do so in their
application code directly.


sgrif:
As long as there's a reasonable use case for recovering from the introduced error kind, and it's sufficiently generic, sure

PR #657, opened 2017-02-09
Clean up type lookup for MySQL and PG

sgrif:
This moves the PG type lookup code to its own function, and removes the
mysql specific parenthesis handling.

For MySQL we have a few special cases to handle

- tinyint(1) is treated as a boolean
- integer is usually called int
- types often have limit/precision/scale specified in parenthesis after
  the name

Other than those cases, our lookup behavior is more or less the same as
PG. I added unit tests for the MySQL behavior, but not the PG -- This is
because for both of them I want to test the full output of
`determine_column_type`, but I also don't want that function to care
about paths, which is a larger refactoring I'm going to do for 0.11 to
support `infer_schema!` with application specific sql types or types
which come from a third party crate.

### DIFF
--- a/diesel_infer_schema/src/mysql.rs
+++ b/diesel_infer_schema/src/mysql.rs
@@ -1,0 +1,55 @@
+use std::error::Error;
+
+use data_structures::*;
+
+pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box<Error>> {
+    let tpe = determine_type_name(&attr.type_name);
+
+    Ok(ColumnType {
+        path: vec!["diesel".into(), "types".into(), capitalize(tpe)],
+        is_array: false,
+        is_nullable: attr.nullable,
+    })
+}
+
+fn determine_type_name(sql_type_name: &str) -> &str {
+    if sql_type_name == "tinyint(1)" {
+        "bool"
+    } else if sql_type_name.starts_with("int") {
+        "integer"
+    } else if let Some(idx) = sql_type_name.find('(') {
+        &sql_type_name[..idx]
+    } else {
+        &sql_type_name
+    }
+}
+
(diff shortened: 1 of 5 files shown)


diesel_infer_schema/src/mysql.rs
```diff
@@ -0,0 +1,55 @@
...
+        is_nullable: attr.nullable,
+    })
+}
+
+fn determine_type_name(sql_type_name: &str) -> &str {
+    if sql_type_name == "tinyint(1)" {
+        "bool"
+    } else if sql_type_name.starts_with("int") {
+        "integer"
+    } else if let Some(idx) = sql_type_name.find('(') {
```
killercup: I just assume that the actual format is always `(?P<tpe>\w)+\((?P<len>\d)+\)`, i.e., there is a closing bracket, and `len` is always a non-empty string. Validating that should not be our tasks here though. Let's just deal with what MySQL gives us. (So, this is good as-is.)

diesel_infer_schema/src/pg.rs
```diff
@@ -0,0 +1,22 @@
...
+    Ok(ColumnType {
+        path: vec!["diesel".into(), "types".into(), capitalize(tpe)],
+        is_array: is_array,
+        is_nullable: attr.nullable,
+    })
+}
+
+fn capitalize(name: &str) -> String {
+    name[..1].to_uppercase() + &name[1..]
+}
```
killercup: For symmetry, I'd add at least copy the tests `values_which_already_map_to_type_are_returned_unchanged` test from mysql and add one for detecting arrays.

sgrif: Yeah, I'll test it more thoroughly after I refactor a bit more.

PR #660, opened 2017-02-09
Fix buffer overrun issue with MySQL

sgrif:
`Vec::reserve` ensures that `self.capacity() >= self.len() +
additional`, not `self.capacity() >= original_capacity + additional`. If
we don't set the length before this call, we aren't necessarily actually
allocating to be the size we want to be, and risk a buffer overrun.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -155,6 +155,7 @@ impl BindData {
 
             debug_assert!(truncated_amount > 0, "output buffers were invalidated \
                 without calling `mysql_stmt_bind_result`");
+            self.bytes.set_len(offset);
             self.bytes.reserve(truncated_amount);
             self.bytes.set_len(self.length as usize);
 


sgrif:
Somewhat humorously, this bug was unlikely to ever cause a segfault. As long as the allocator is laying things out as compactly as possible, there will always be allocated memory in the address after the end of the vector by the time the overrun could occur.

diesel/src/mysql/connection/bind.rs
```diff
@@ -155,6 +155,7 @@ impl BindData {
 
             debug_assert!(truncated_amount > 0, "output buffers were invalidated \
                 without calling `mysql_stmt_bind_result`");
+            self.bytes.set_len(offset);
             self.bytes.reserve(truncated_amount);
             self.bytes.set_len(self.length as usize);
```
killercup: Do you want to add a comment to this? Calling set_len twice might be confusing.

sgrif: Will do on master

PR #662, opened 2017-02-09
Add support for MySQL to Diesel CLI

sgrif:
MySQL is relatively similar to PG for CLI's purposes. The main
differences are:

- `information_schema.tables` aren't scoped to the current database
- The default databases that are present are `mysql`,
  `information_schema`, and sometimes `test`. I chcose the second
  because it seemed most likely to have permissions granted
- `migration redo` can't run inside of a transaction because the
  transaction is implicitly committed when the schema is modified.
  `COMMIT` when not inside of a transaction is not an error, but
  wrapping the outer calls in a transaction cause us to use savepoints
  which error when we try to release them.

Running the tests for CLI requires GRANT ALL ON `diesel_%`.* TO
''@'localhost'. I need to see if I can make the URLs we use more
configurable.

The way we check to run a transaction in `migration_redo` is a little
hacky, but was overall less code than restructuring it to pass the enum
through further.

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -17,6 +17,7 @@ use clap::{ArgMatches,Shell};
 use diesel::migrations::schema::*;
 use diesel::types::{FromSql, VarChar};
 use diesel::{migrations, Connection, Insertable};
+use std::any::Any;
 use std::error::Error;
 use std::io::stdout;
 use std::path::{PathBuf, Path};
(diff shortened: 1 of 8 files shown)


diesel_cli/src/main.rs
```diff
@@ -153,15 +154,30 @@ fn search_for_cargo_toml_directory(path: &Path) -> DatabaseResult<PathBuf> {
...
+    }
+}
+
+#[cfg(feature="mysql")]
+fn should_redo_migration_in_transaction(t: &Any) -> bool {
+    !t.is::<::diesel::mysql::MysqlConnection>()
+}
+
+#[cfg(not(feature="mysql"))]
+fn should_redo_migration_in_transaction(t: &Any) -> bool {
```
killercup: `s/t/_t` fixes ci

killercup:
Fixed the lint, now errors with

> thread 'database_drop::database_drop_drops_database' panicked at 'Unexpected stdout ', tests/database_drop.rs:15

which looks like a legit failure.

PR #663, opened 2017-02-09
Change a bunch of tests to not mutate schema in the test

sgrif:
We can't run these tests on MySQL because MySQL can't modify schema
inside of a transaction. None of these are making modifications that
can't just be make ahead of time, though. I do have to ensure that they
all have a primary key, as `infer_schema!` will pick up on them.


sgrif:
I left the `table!` definitions local for documentation purposes

PR #666, opened 2017-02-10
Don't print `Dropping database` to STDOUT when db doesn't exist

sgrif:
There was an interesting failure when running the CLI tests for PG
locally after adding MySQL support. The failure only happened on the
second run of a test, so wouldn't have been caught by CI. The failure
occurred because the test creates a database, then creates a new role,
and runs `diesel database reset` as that role. On the second run it
would fail to drop the role because there was an existing database
depending on it.

The issue was that I had changed the name of the database created by the
project, but that test constructed its own database url and was out of
sync. The fix was [this][fix], but the fact that CI wouldn't have caught
this bothered me. There was no change I could make in just the test to
ensure all my assumptions were correct, since we would always say
"dropping database:" in the CLI.

[fix]: https://github.com/diesel-rs/diesel/commit/a56eae379c37bd9fe39f21ad7d11d2d1f9364f05#diff-6c048bb08a0433845cbca8b7536be378R58,

With this change, we now only print "Dropping database:" if the database
already exists, mimicking the behavior of `diesel database setup`. If
that same bug existed in that test today, it would be caught on the
first test run rather than the second.

### DIFF
--- a/diesel_cli/src/database.rs
+++ b/diesel_cli/src/database.rs
@@ -7,7 +7,7 @@ use diesel::sqlite::SqliteConnection;
 #[cfg(feature="mysql")]
 use diesel::mysql::MysqlConnection;
 use diesel::types::Bool;
-use diesel::{migrations, Connection, select, LoadDsl};
+use diesel::*;
 
 use database_error::{DatabaseError, DatabaseResult};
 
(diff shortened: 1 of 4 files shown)


sgrif:
Oh man I wish I'd realized this was going to be number 666. I would have made it a PR that enabled MySQL on CI or something.

killercup:
> Oh man I wish I'd realized this was going to be number 666. I would have made it a PR that enabled MySQL on CI or something.

Still have that chance! Rebase it ~~to hell~~ to another branch ;)

sgrif:
Lol I don't think the joke is worth that much work. ;)

diesel_cli/src/database.rs
```diff
@@ -162,28 +162,73 @@ fn drop_database(database_url: &str) -> DatabaseResult<()> {
...
+#[cfg(feature="postgres")]
+fn pg_database_exists(conn: &PgConnection, database_name: &str) -> QueryResult<bool> {
+    use self::pg_database::dsl::*;
+
+    pg_database.select(datname) // here come dat name!!!! o shit waddup!!!!
+        .filter(datname.eq(database_name))
+        .filter(datistemplate.eq(false))
+        .get_result::<String>(conn)
+        .optional()
+        .map(|x| x.is_some())
```
killercup: i think `.map(Result::is_some)` reads nicer

diesel_cli/src/database.rs
```diff
@@ -162,28 +162,73 @@ fn drop_database(database_url: &str) -> DatabaseResult<()> {
...
+
+#[cfg(feature="mysql")]
+fn mysql_database_exists(conn: &MysqlConnection, database_name: &str) -> QueryResult<bool> {
+    use self::schemata::dsl::*;
+
+    schemata.select(schema_name)
+        .filter(schema_name.eq(database_name))
+        .get_result::<String>(conn)
+        .optional()
+        .map(|x| x.is_some())
```
killercup: dito

diesel_cli/src/database.rs
```diff
@@ -162,28 +162,73 @@ fn drop_database(database_url: &str) -> DatabaseResult<()> {
...
         Backend::Pg => {
             let (database, postgres_url) = change_database_of_url(database_url, "postgres");
-            println!("Dropping database: {}", database);
             let conn = try!(PgConnection::establish(&postgres_url));
-            try!(conn.silence_notices(|| {
-                conn.execute(&format!("DROP DATABASE IF EXISTS {}", database))
-            }));
+            if try!(pg_database_exists(&conn, &database)) {
+                println!("Dropping database: {}", database);
+                try!(conn.execute(&format!("DROP DATABASE IF EXISTS {}", database)));
```
killercup: [droppin da bass](http://kimjongildroppingthebass-blog.tumblr.com)

diesel_cli/src/database.rs
sgrif: Does `Result::<String, Error>::is_some` still read nicer to you?

diesel_cli/src/database.rs
sgrif: ditto

diesel_cli/src/database.rs
killercup: rustc y u no infer me happiness

PR #667, opened 2017-02-10
Remove impl_Queryable macro

weiznich:
As discussed in #608 this moves the implementation of `#[dervie(Queryable)]` to a pure proc-macro based implementation.
Using a similar approach as `serde` the issue with `$crate` should also be resolved.


diesel_codegen/tests/test_helpers/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+            let database_url = env::var("MYSQL_DATABASE_URL")
+                .or_else(|_| env::var("DATABASE_URL"))
+                .expect("DATABASE_URL must be set in order to run tests");
+            MysqlConnection::establish(&database_url).unwrap()
+        }
+    } else {
+
+        pub type TestConnection = ();
+        pub fn connection() -> TestConnection {
+            panic!("At least one backend must be enabled to run tests")
```
sgrif: Why add this? Nothing is going to compile with `TestConnection = ()` anyway.

diesel_infer_schema/src/sqlite.rs
```diff
@@ -60,17 +60,26 @@ struct FullTableInfo {
...
+            name: row.1,
+            _type_name: row.2,
+            _not_null: row.3,
+            _dflt_value: row.4,
+            primary_key: row.5,
+        }
     }
 }
 
+
```
sgrif: Can you please make more of an effort to not add random whitespace and empty lines in your pull requests?

sgrif:
I cleaned this up manually and merged in e160921c1a1d4e62e1aae480ae405d918fe6bc22

PR #668, opened 2017-02-10
Add MySQL to the local test script

sgrif:
The best I could come up with for enabling the unit test suite is to
make sure that it's using a different database than the integration test
suite. I'm only looking for the `UNIT_TEST` env var and not falling back
to `MYSQL_DATABASE_URL` because running it against the same database as
the integration tests will mess things up (less of a problem for CI
since it runs unit, then integration, then throws everything away)

I'm holding off on adding it to CI until I tweak a few more things.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -31,6 +31,7 @@ tempdir = "^0.3.4"
 
 [features]
 default = ["with-deprecated"]
+extras = ["chrono", "serde_json", "uuid"]
 unstable = []
 large-tables = []
 huge-tables = ["large-tables"]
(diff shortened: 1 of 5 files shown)


diesel/Cargo.toml
```diff
@@ -31,6 +31,7 @@ tempdir = "^0.3.4"
 
 [features]
 default = ["with-deprecated"]
+all = ["chrono", "quickcheck", "serde_json", "uuid"]
```
killercup: Hm. I'd probably call this 'extras'. Is Quickcheck useful for anything but _our_ tests?

sgrif: If anyone were using `PgDateTime`, `PgNumeric` and friends or the interval DSL and wanted to test stuff involving it, yes. But I agree it doesn't warrant inclusion here.

PR #674, opened 2017-02-11
Add some missing tests for PG prepared statement caching

sgrif:
I'm looking into extracting the prepared statement caching for SQLite
and PG into a shared module (so I don't have to re-implement it *again*
for MySQL). While I was making sure I understood the requirements of the
existing implementations, I noticed that we had no tests for PG where
the statement is cached, but not on the type id.

The only cases where `QueryFragment::is_safe_to_cache_prepared` returns
`true`, but `QueryId::has_static_query_id` returns `false` is using
`Aliased`/`with` on PG, and boxed queries. Since these tests run with no
schema, boxing is easier to test.

When I wrote these tests, I noticed that attempting to run a boxed query
with no query source failed to compile. I've added the missing impl
required for it. It's copypasta from the other impl, with the from lines
deleted, but I liked this code better than with `to_sql_before_from`,
`to_sql_after_from`, `collect_binds_before_from`, etc pulled out. We can
always refactor if these impls end up having to change a lot.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -200,6 +200,34 @@ mod tests {
     }
 
     #[test]
+    fn queries_with_identical_types_and_sql_but_different_bind_types_are_cached_separately() {
+        let connection = connection();
+
+        let query = ::select(AsExpression::<Integer>::as_expression(1)).into_boxed::<Pg>();
+        let query2 = ::select(AsExpression::<VarChar>::as_expression("hi")).into_boxed::<Pg>();
+
+        assert_eq!(0, connection.statement_cache.len());
+        assert_eq!(Ok(1), query.get_result(&connection));
+        assert_eq!(Ok("hi".to_string()), query2.get_result(&connection));
+        assert_eq!(2, connection.statement_cache.len());
+    }
+
+    #[test]
+    fn queries_with_identical_types_and_binds_but_different_sql_are_cached_separately() {
+        let connection = connection();
+
+        sql_function!(lower, lower_t, (x: VarChar) -> VarChar);
+        let hi = AsExpression::<VarChar>::as_expression("HI");
+        let query = ::select(hi).into_boxed::<Pg>();
+        let query2 = ::select(lower(hi)).into_boxed::<Pg>();
+
+        assert_eq!(0, connection.statement_cache.len());
+        assert_eq!(Ok("HI".to_string()), query.get_result(&connection));
+        assert_eq!(Ok("hi".to_string()), query2.get_result(&connection));
+        assert_eq!(2, connection.statement_cache.len());
+    }
+
+    #[test]
     fn queries_with_sql_literal_nodes_are_not_cached() {
         let connection = connection();
         let query = ::select(sql::<Integer>("1"));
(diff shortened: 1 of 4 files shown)


diesel/src/pg/connection/mod.rs
```diff
@@ -200,6 +200,32 @@ mod tests {
...
     #[test]
+    fn queries_with_identical_types_and_sql_but_different_bind_types_are_cached_separately() {
+        let connection = connection();
+
+        let query = ::select(AsExpression::<Integer>::as_expression(1)).into_boxed::<Pg>();
+        let query2 = ::select(AsExpression::<VarChar>::as_expression("hi")).into_boxed::<Pg>();
+
+        assert_eq!(Ok(1), query.get_result(&connection));
+        assert_eq!(Ok("hi".to_string()), query2.get_result(&connection));
+        assert_eq!(2, connection.statement_cache.len());
```
killercup: Just as a sanity check, I'd always put a `assert_eq!(0, connection.statement_cache.len());` (or actually `assert!(connection.statement_cache.is_empty());`) at the top (same below).

diesel/src/query_builder/select_statement/boxed.rs
```diff
@@ -102,6 +102,49 @@ impl<'a, ST, QS, DB> QueryFragment<DB> for BoxedSelectStatement<'a, ST, QS, DB>
...
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        self.distinct.is_safe_to_cache_prepared() &&
+            self.select.is_safe_to_cache_prepared() &&
+            self.where_clause.as_ref().map(|w| w.is_safe_to_cache_prepared()).unwrap_or(true) &&
+            self.order.is_safe_to_cache_prepared() &&
+            self.limit.is_safe_to_cache_prepared() &&
+            self.offset.is_safe_to_cache_prepared()
+    }
+}
+
```
killercup: FTR, this is the "real" diff:

```diff
--- impl<'a, ST, QS, DB> QueryFragment<DB> for BoxedSe
+++ untitled (Current)
@@ -1,14 +1,10 @@
-impl<'a, ST, QS, DB> QueryFragment<DB> for BoxedSelectStatement<'a, ST, QS, DB> where
+impl<'a, ST, DB> QueryFragment<DB> for BoxedSelectStatement<'a, ST, (), DB> where
     DB: Backend,
-    QS: QuerySource,
-    QS::FromClause: QueryFragment<DB>,
 {
     fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
         out.push_sql("SELECT ");
         try!(self.distinct.to_sql(out));
         try!(self.select.to_sql(out));
-        out.push_sql(" FROM ");
-        try!(self.from.from_clause().to_sql(out));
 
         if let Some(ref where_clause) = self.where_clause {
             out.push_sql(" WHERE ");
@@ -24,7 +20,6 @@
     fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()> {
         try!(self.distinct.collect_binds(out));
         try!(self.select.collect_binds(out));
-        try!(self.from.from_clause().collect_binds(out));
 
         if let Some(ref where_clause) = self.where_clause {
             try!(where_clause.collect_binds(out));
@@ -39,7 +34,6 @@
     fn is_safe_to_cache_prepared(&self) -> bool {
         self.distinct.is_safe_to_cache_prepared() &&
             self.select.is_safe_to_cache_prepared() &&
-            self.from.from_clause().is_safe_to_cache_prepared() &&
             self.where_clause.as_ref().map(|w| w.is_safe_to_cache_prepared()).unwrap_or(true) &&
             self.order.is_safe_to_cache_prepared() &&
             self.limit.is_safe_to_cache_prepared() &&
```

PR #678, opened 2017-02-11
Implement a shared `StatementCacheKey`

sgrif:
This is part of a refactoring to have a shared `StatementCache` between
all three backends because I don't want to write this code a third time
for MySQL.

A refresher on the situation with prepared statements. For a given
query, it will fall into one of three categories:

- Unsafe to cache. This means that the query could potentially generate
  an unbounded number of prepared statements. Only insert queries
  (SQL varies based on number of records inserted), queries containing
  `IN`/`NOT IN` (SQL varies based on number of items on the right side),
  and queries containing `SqlLiteral` (SQL varies based on who the fuck
  knows because we didn't generate) fall into this category.
- Safe to cache, but can't be identified by type. These are queries
  where the SQL can change without the Rust type changing, but there is
  an upper bound on the number of queries it could generate. Generally
  only boxed select statements fall into this category.
- Safe to cache, uniquely identified by type. The majority of queries
  will fall into this category. These are also the happy path, as we can
  skip the query builder entirely for subsequent executions.

The PG code also goes through a completely separate code path for
queries which aren't safe to cache. It doesn't need to do this though,
and I'm going to refactor it to not do it in the future.

For queries which are cached by SQL instead of type, the types of the
bind parameters also matter for PG. They don't matter for SQLite or
MySQL, but we could potentially save some allocations by holding onto
that vector anyway. The wins from doing this would be questionable at
best.

A few notes on things that might be funky in the implementation:

I opted for a `where DB::QueryBuilder: Default` bound instead of adding
a `query_buidler()` function to `Backend` because I do not want to rule
out future backends requiring additional constructor parameters to their
query builders (those backends just won't be able to use this prepared
statement cache).

I had to derive `Hash`, `PartialEq`, and `Eq` on the structs themselves
because `#[derive]` generates unnecessary bounds. This felt less bad
than writing manual impls.

There's a handful of additional optimizations I could do here (not allocating on the "cache on SQL" case, implementing `PgTypeMetadata` to only hash on oid instead of all fields, etc) but I don't think any of them will result in meaningful performance wins.

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -1,3 +1,4 @@
+mod statement_cache;
 mod transaction_manager;
 
 use backend::Backend;
@@ -7,6 +8,8 @@ use result::*;
 use types::HasSqlType;
 
 pub use self::transaction_manager::{TransactionManager, AnsiTransactionManager};
+#[doc(hidden)]
+pub use self::statement_cache::StatementCacheKey;
 
 pub trait SimpleConnection {
     #[doc(hidden)]
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -1,0 +1,54 @@
+use std::any::TypeId;
+use std::borrow::Cow;
+
+use backend::Backend;
+use query_builder::*;
+use result::QueryResult;
+
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+#[derive(Hash, PartialEq, Eq)]
+pub enum StatementCacheKey<DB: Backend> {
+    Type(TypeId),
+    Sql {
+        sql: String,
+        bind_types: Vec<DB::TypeMetadata>,
+    }
+}
+
+impl<DB> StatementCacheKey<DB> where
+    DB: Backend,
+    DB::QueryBuilder: Default,
+    DB::TypeMetadata: Clone,
+{
+    pub fn for_source<T>(source: &T, bind_types: &[DB::TypeMetadata])
+        -> QueryResult<Self> where
+            T: QueryFragment<DB> + QueryId,
+    {
+        match T::query_id() {
+            Some(id) => Ok(StatementCacheKey::Type(id)),
+            None => {
+                let sql = try!(Self::construct_sql(source));
+                Ok(StatementCacheKey::Sql {
+                    sql: sql,
+                    bind_types: bind_types.into(),
+                })
+            }
+        }
+    }
+
+    pub fn sql<T: QueryFragment<DB>>(&self, source: &T) -> QueryResult<Cow<str>> {
+        match *self {
+            StatementCacheKey::Type(_) => Self::construct_sql(source).map(Cow::Owned),
+            StatementCacheKey::Sql { ref sql, .. } => Ok(Cow::Borrowed(sql)),
+        }
+    }
+
+    fn construct_sql<T: QueryFragment<DB>>(source: &T) -> QueryResult<String> {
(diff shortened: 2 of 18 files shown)


diesel/src/connection/statement_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+#[derive(Hash, PartialEq, Eq)]
+pub enum StatementCacheKey<DB: Backend> {
+    Type(TypeId),
+    Sql {
+        sql: String,
+        bind_types: Vec<DB::TypeMetadata>,
+    }
+}
```
killercup: `TypeId` is as large as one `u64`, while `Sql` is 6 `usize`s. I'd be surprised if it matters perf-wise even in large apps but you mentioned `Type` should be the hot path anyway.

diesel/src/query_builder/mod.rs
```diff
@@ -57,6 +57,7 @@ pub trait QueryBuilder<DB: Backend> {
     fn push_sql(&mut self, sql: &str);
     fn push_identifier(&mut self, identifier: &str) -> BuildQueryResult;
     fn push_bind_param(&mut self);
+    fn finish(self) -> String;
```
killercup: Good idea 👍

diesel/src/connection/mod.rs
```diff
@@ -7,6 +8,8 @@ use result::*;
 use types::HasSqlType;
 
 pub use self::transaction_manager::{TransactionManager, AnsiTransactionManager};
+#[doc(hidden)]
+pub use self::statement_cache::StatementCacheKey;
```
killercup: Is this one of the things where we'd like to have <https://github.com/rust-lang/rust/issues/32409>?

diesel/src/query_builder/bind_collector.rs
```diff
@@ -7,20 +7,22 @@ pub trait BindCollector<DB: Backend> {
 
 #[derive(Debug)]
 pub struct RawBytesBindCollector<DB: Backend + TypeMetadata> {
-    pub binds: Vec<(DB::TypeMetadata, Option<Vec<u8>>)>,
+    pub metadata: Vec<DB::TypeMetadata>,
+    pub binds: Vec<Option<Vec<u8>>>,
 }
 
 impl<DB: Backend + TypeMetadata> RawBytesBindCollector<DB> {
     pub fn new() -> Self {
```
killercup: Did clippy not suggest to derive Default here? Maybe because of the type bounds?

diesel/src/connection/statement_cache.rs
sgrif: Yeah, I'm really not worried about the 40 byte overhead there. I do want to split the two caches into separate maps though since the `TypeId` can work with a phf which will reflect perf-wise

diesel/src/connection/mod.rs
sgrif: Probably, yes. (I actually don't mind third party connection adapters using this, I just don't want to commit to stability with its API)

diesel/src/query_builder/bind_collector.rs
sgrif: Yeah, probably because of the bounds. `#[derive(Default)]` will require adding it to every value of `DB` that we want as well (which is nonsense and not required, but rustc's derives are kinda dumb)

sgrif:
> Still a complicated piece of diesel

Yeah, when you're using `TypeId` as a key in a hashmap as only one possible code path, it's pretty much always going to be complex. XD

PR #681, opened 2017-02-12
Extract a shared `StatementCache` between SQLite and PG

sgrif:
The goal is to re-use this code for MySQL as well in the future, but
this PR does not yet implement this. In the short term, this addition
should make it much easier to remove the `Clone` bound and better
reflect the invariants behind SQLite and MySQL -- That the statement is
only manipulated by one execution at a time. This requires removing the
`Rc` wrapping them. This too will come in a future PR. The goal of this
commit is simply to pull out the common code, and set up the type bounds
required.


diesel/src/connection/statement_cache.rs
```diff
@@ -1,12 +1,70 @@
...
+{
+    pub fn new() -> Self {
+        StatementCache {
+            cache: RefCell::new(HashMap::new())
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.cache.borrow().len()
+    }
```
killercup: Thanks for testing clippy! :trollface: 

```
error: item `connection::statement_cache::StatementCache<DB, Statement>` has a public `len` method but no corresponding `is_empty` method
  --> src/connection/statement_cache.rs:29:5
   |
29 |       pub fn len(&self) -> usize {
   |  _____^ starting here...
30 | |         self.cache.borrow().len()
31 | |     }
   | |_____^ ...ending here
   |
   = note: #[deny(len_without_is_empty)] implied by #[deny(warnings)]
note: lint level defined here
  --> src/lib.rs:7:9
   |
7  | #![deny(warnings, missing_debug_implementations, missing_copy_implementations)]
   |         ^^^^^^^^
   = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#len_without_is_empty
```

(This was okay before as it was `#[cfg(test)]`)

sgrif: Oh I thought you had made `bin/test` do clippy as well. I'll stick an allow on here, no point in adding an unused method.

killercup: I hadn't as you refactored bin/test for mysql at the same time. Gimme a sec to add it

sgrif:
Right. The next step is to have this drop the clone assumption and return a `RefMut` instead.

PR #684, opened 2017-02-12
Contributing.md

killercup:


CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+Thanks for wanting to contribute to Diesel! We are very much looking forward to your suggestions, bug reports, and pull requests.
+
+## Setting up diesel locally
+
+1. Install Rust using [rustup], which allows you to easily switch between Rust versions.
+2. Install the system libraries needed to interface with the database systems you which to use.
+
+    These are the same as when compiling diesel. In general, it is a good idea to have _all_ drivers installed so you can run all tests locally.
+
+    *Shortcut:* On macOS, you don't need to install anything to work with Postgres and SQLite.
```
sgrif: macOS ships with libpq, but it does not ship with a postgres database. Most people will need to install postgres to get it running.

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+3. Clone this repository and open it in your favorite editor.
+4. Create a `.env` file in the `diesel/` directory, and add the connection details for your databases.
+
+    For example:
+
+    ```
+    PG_DATABASE_URL=postgresql://pascal@localhost/diesel_test
+    MYSQL_DATABASE_URL=mysql://root@localhost/diesel_test
```
sgrif: They'll also need to set `MYSQL_UNIT_TEST_DATABASE_URL` to a different database.

sgrif: Not sure how I feel about recommending that they run as root. I think we should probably mention how to not run as root even if we continue to say root in the `.env` example. Something like:

> If you didn't specify the MySQL user to be one with elevated permissions, you'll want to a command like `mysql -c "GRANT ALL ON \`diesel_%\`.* TO ''@'localhost';" -uroot`, or something similar for the user that you've specified.

killercup:
What else should be in this file?

- [x] local setup
- [ ] link to issue tracker, and what information to supply
- [ ] link to CoC
- [ ] general walkthrough through the test infrastructure
- [ ] ???
- [ ] PROFIT!!

CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+Thanks for wanting to contribute to Diesel! We are very much looking forward to your suggestions, bug reports, and pull requests.
+
+## Setting up diesel locally
+
+1. Install Rust using [rustup], which allows you to easily switch between Rust versions.
+2. Install the system libraries needed to interface with the database systems you which to use.
+
+    These are the same as when compiling diesel. In general, it is a good idea to have _all_ drivers installed so you can run all tests locally.
+
+    *Shortcut:* On macOS, you don't need to install anything to work with SQLite and for PostgreSQL you'll only the server (`libpq` is installed by default). So, to get started, `brew install postgresql mysql` and follow the instructions shown to set up the database servers.
```
Eijebong: `you'll only the` => `you'll only need the`

PR #686, opened 2017-02-12
Remove the separate code path for non-cached statements on PG

sgrif:
In the case where a statement isn't safe to cache, we go through a
separate code path to use `PQexecParams` instead of `PQprepare` and
`PQexecPrepared`. `PQexecParams` does effectively the same thing that
we're doing now (preparing the statement with no name, and immediately
executing it).

The old code was slightly more efficient since `PQexecParams` avoided
sending a `Sync` message between the `Parse` and `Bind` messages. The
new code ends up doing one extra round trip, and one extra check for
errors. However, the completely-uncached case is the least used code
path, and the performance difference is minimal even when we hit it.
For that reason, I don't think it's worth the additional code.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -11,7 +11,7 @@ use std::ffi::{CString, CStr};
 use std::rc::Rc;
 
 use connection::*;
-use pg::{Pg, PgQueryBuilder};
+use pg::Pg;
 use query_builder::*;
 use query_builder::bind_collector::RawBytesBindCollector;
 use query_source::Queryable;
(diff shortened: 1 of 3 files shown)


diesel/src/pg/connection/mod.rs
```diff
@@ -109,35 +109,33 @@ impl Connection for PgConnection {
...
-                    &query_name,
-                    &metadata,
-                ).map(Rc::new)
-            }))
-        } else {
-            let mut query_builder = PgQueryBuilder::new();
-            try!(source.to_sql(&mut query_builder).map_err(Error::QueryBuilderError));
-            Rc::new(try!(Query::sql(&query_builder.finish(), Some(metadata))))
-        };
+        let cache_len = self.statement_cache.len();
```
killercup: Why'd you move that out of the closure? Does borrowck not let you access this otherwise?

sgrif: In the closure we have a life mutable reference (which came from a refcell). Trying to get an additional reference would panic. I could pass a reference to the cache to the closure but this felt better.

PR #687, opened 2017-02-12
Allow `now` to be used as a timestamptz

### LINKED ISSUE #685: diesel::expression::dsl::now cannot be used with diesel::pg::types::sql_types::Timestamptz
I'm trying to execute following query:

```
        diesel::update(session.filter(key.eq(new_session.key)))
            .set((
                data.eq(new_session.data),
                accessed.eq(now),
            ))
            .execute(&connection)
            .unwrap();
```

Here `accessed` is Timestamptz field.

It doesn't compile and I receive
```
expected struct `diesel::types::Timestamp`, found struct `diesel::types::Timestamptz`
```

As a workaround I copied-and-pasted `now` and redefined Expression
```
pub struct NowTz;

impl Expression for NowTz {
    type SqlType = Timestamptz;
}

...
```

but it would be nice to have built-in solution.

sgrif:
There are two things wrong with this PR, but I still think this is the
best short term solution.

The first and most glaringly obviously wrong thing about it is that the
type of `NOW()` **IS** `timestamptz`. However, since we currently have
that defined as a PG specific type, we can't set the type of the
expression properly. Timestamp with/without timezone is a thing in the
SQL standard, (it's just ignored by SQLite). I need to examine how MySQL
handles it and figure out how to promote it to a more general type.
However, I don't want to deal with that right now (and probably not
before 0.11). I do still want to fix this issue, so here we are.

We also have a more general problem of type equivalence that we need to
address. `Json` expressions can always be used where a `Jsonb` is
expected. `Text` with `VarChar` (though we've skirted that one by making
them type aliases). Again, this is a deeper problem that I don't
currently have a solution in mind for.

So in the short term, this gives us a way to punt on the problem for a
while. For the most common cases that crop up, we can write
`AsExpression` impls for those cases when we notice them.

Fixes #685.

### DIFF
--- a/diesel/src/expression/coerce.rs
+++ b/diesel/src/expression/coerce.rs
@@ -1,0 +1,75 @@
+use std::marker::PhantomData;
+
+use backend::Backend;
+use expression::{Expression, SelectableExpression, NonAggregate};
+use query_builder::*;
+use result::QueryResult;
+
+#[derive(Debug, Copy, Clone)]
+#[doc(hidden)]
+/// Coerces an expression to be another type. No checks are performed to ensure
+/// that the new type is valid in all positions that the previous type was.
+/// This does not perform an actual cast, it just lies to our type system.
+///
+/// This is used for a few expressions where we know that the types are actually
+/// always interchangeable. (Examples of this include `Timestamp` vs
+/// `Timestamptz`, `VarChar` vs `Text`, and `Json` vs `Jsonb`).
+///
+/// This struct should not be considered a general solution to equivalent types.
+/// It is a short term workaround for expressions which are known to be commonly
+/// used.
+pub struct Coerce<T, ST> {
+    expr: T,
+    _marker: PhantomData<ST>,
+}
+
+impl<T, ST> Coerce<T, ST> {
+    pub fn new(expr: T) -> Self {
+        Coerce {
+            expr: expr,
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<T, ST> Expression for Coerce<T, ST> where
+    T: Expression,
+{
+    type SqlType = ST;
+}
+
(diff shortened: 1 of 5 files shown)


diesel/src/expression/functions/date_and_time.rs
```diff
@@ -42,3 +42,19 @@ operator_allowed!(now, Sub, sub);
...
+use types::Timestamptz;
+
+#[cfg(feature="postgres")]
+impl AsExpression<Timestamptz> for now {
+    type Expression = Coerce<now, Timestamptz>;
+
+    fn as_expression(self) -> Self::Expression {
+        Coerce::new(self)
+    }
+}
```
killercup: Maybe put that in a macro `unsafely_coerce_types!(now, Timestamptz)`?

diesel/src/expression/coerce.rs
```diff
@@ -0,0 +1,75 @@
...
+/// that the new type is valid in all positions that the previous type was.
+/// This does not perform an actual cast, it just lies to our type system.
+///
+/// This is used for a few expressions where we know that the types are actually
+/// always interchangeable. (Examples of this include `Timestamp` vs
+/// `Timestamptz`, `VarChar` vs `Text`, and `Json` vs `Jsonb`).
+///
+/// This struct should not be considered a general solution to equivalent types.
+/// It is a short term workaround for expressions which are known to be commonly
+/// used.
```
sgrif: :+1:

diesel/src/expression/functions/date_and_time.rs
sgrif: It's only 6 lines of code. I'd rather keep the normal Rust code which is easier to understand.

PR #697, opened 2017-02-13
Provide a more helpful error message when to few bytes were received

### LINKED ISSUE #696: Type mismatch errors can be almost impossible to find through error messages
Have a look at the following (almost) minimal example program:

https://gist.github.com/mbr/78fabb54480549ac89ef8c98d9c186c6

The issue here is that a BigInt is used in the `table!` macro call where there should be an `Integer` (due to how the SQL migration is setup. Also, `User.id` needs to be corrected to i32).

The resulting error message (UnexpectedEOF) makes this not that uncommon error very cryptic. Returning a better error might improved usability by quite a bit.

sgrif:
We have a debug check for the semi-common error of mislabeling a
field/expression to a type which is too small. We originally added these
because the value would actually deserialize successfully, but to the
wrong value. We didn't add checks for the "too few bytes" case, as it
would have errored anyway and seemed less common.

However, the error message received when one does make this mistake is
not terribly helpful. We can add an additional check in debug mode to
give a bit more information when someone does make this mistake.

We should see if we can provide some more general improvements to an
error generated by `DeserializationError` in the future. Including the
column name would go a long way, but is a bit larger in scope.

Fixes #696.

### DIFF
--- a/diesel/src/types/impls/integers.rs
+++ b/diesel/src/types/impls/integers.rs
@@ -10,6 +10,9 @@ impl<DB: Backend<RawValue=[u8]>> FromSql<types::SmallInt, DB> for i16 {
         let mut bytes = not_none!(bytes);
         debug_assert!(bytes.len() <= 2, "Received more than 2 bytes decoding i16. \
                       Was an Integer expression accidentally identified as SmallInt?");
+        debug_assert!(bytes.len() == 2, "Received fewer than 2 bytes decoding i16. \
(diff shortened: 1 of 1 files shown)


Eijebong:
You have a typo in the commit message, it should be "too few", not "to few" ;-)

Not sure if you really care about that but eh...

sgrif:
I do. I'll amend when I merge. :)

Issue #695, opened 2017-02-13
Ensure we are resilient across PG versions


sgrif:
We currently always use the binary format for PG. However, the only type which explicitly documents its binary format is integer types, which is simply documented as using network byte order. For all other types we literally have to read the PG source code to find the binary representations. This appears to have been reasonably stable for at least the last couple versions, but we shouldn't assume that will always be the case. In fact the documentation explicitly states:

> Keep in mind that binary representations for complex data types might change across server versions; the text format is usually the more portable choice.

I do want to continue using binary whenever possible. For some types, the difference in performance is several orders of magnitude in terms of the work being done to handle the binary form vs what we'd have to do to parse the text form. However, we should at minimum do the following:

- Test against as many PG versions as we can and is reasonable
- Check the server's version in the connection, error if it's too old, and at least warn if it's newer than the most recent version we've tested against.
- Potentially fall back to the text form if the version isn't one that we know uses the expected binary format

wt:
This seems to have been reported some time ago. Is this still an active concern?

weiznich:
@wt This is still not addressed as we didn't implement any of this. We still look for someone to work at least on some of the points, possibly not the last one.

ayarotsky:
@weiznich do you mind if I take this one?

PR #698, opened 2017-02-13
Allow arrays containing null to be serialized

sgrif:
Deserialization already worked, so there's no reason we shouldn't just
allow this. It's worth noting that there is no way to represent whether
an array is allowed to contain nulls or not within postgres's type
system. As such, the type `Array<Nullable<ST>>` will never be generated
by `infer_schema!` (I do want to make `FromSql<Array<ST>> for
`Vec<Option<T>>` just work though). Since we never provide a
`ToSql<Array<ST>> for Vec<Option<T>>`, Diesel will prevent you from
accidentally inserting nulls at compile time. Ultimately the user does
just have to make sure they represent whether the array could contain
null in the output type, however.

One alternative would be to represent the fact that all arrays can
contain null within the type system, and only provide `impl
FromSql<Array<ST>> for Vec<Option<T>>`. I suspect this would hinder
ergonomics pretty heavily, but we should consider it before 1.0.

One note about the implementation -- Technically we should be setting
`flags` to `1` if the array contains any nulls, as that's what happens
on the data sent to us. However, setting it would require us to perform
an additional level of intermediate buffering on all the values to
determine whether the flag should be set, and that flag is currently
ignored by PG. "But Sean, that's just relying on an implementation
detail!" this is true, but our entire usage of the binary format is
relying on an implementation detail (See #695).

### DIFF
--- a/diesel/src/pg/types/array.rs
+++ b/diesel/src/pg/types/array.rs
@@ -134,10 +134,14 @@ impl<'a, ST, T> ToSql<Array<ST>, Pg> for &'a [T] where
         let mut buffer = Vec::new();
         for elem in self.iter() {
             let is_null = try!(elem.to_sql(&mut buffer));
-            assert!(is_null == IsNull::No, "Arrays containing null are not supported");
-            try!(out.write_i32::<NetworkEndian>(buffer.len() as i32));
-            try!(out.write_all(&buffer));
-            buffer.clear();
+            if let IsNull::No = is_null {
+                try!(out.write_i32::<NetworkEndian>(buffer.len() as i32));
+                try!(out.write_all(&buffer));
+                buffer.clear();
+            } else {
+                // https://github.com/postgres/postgres/blob/82f8107b92c9104ec9d9465f3f6a4c6dab4c124a/src/backend/utils/adt/arrayfuncs.c#L1461
+                try!(out.write_i32::<NetworkEndian>(-1));
+            }
         }
 
         Ok(IsNull::No)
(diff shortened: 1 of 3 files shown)


diesel/src/pg/types/array.rs
```diff
@@ -134,10 +134,13 @@ impl<'a, ST, T> ToSql<Array<ST>, Pg> for &'a [T] where
...
-            assert!(is_null == IsNull::No, "Arrays containing null are not supported");
-            try!(out.write_i32::<NetworkEndian>(buffer.len() as i32));
-            try!(out.write_all(&buffer));
-            buffer.clear();
+            if let IsNull::No = is_null {
+                try!(out.write_i32::<NetworkEndian>(buffer.len() as i32));
+                try!(out.write_all(&buffer));
+                buffer.clear();
+            } else {
+                try!(out.write_i32::<NetworkEndian>(-1));
```
killercup: I'm not familiar with Postgres' binary interface, so this is one of the places where I'd really like to have a link to some semi-official Postgres document that says "-1 is null here".

sgrif: I can link to the line in the source code? There is no documentation on this. But we also have https://github.com/diesel-rs/diesel/blob/3d985d717b0caaad931464e8a658c1f73bc9a57e/diesel/src/pg/types/array.rs#L54-L56 ;)

sgrif: Would you like me to link to https://github.com/postgres/postgres/blob/82f8107b92c9104ec9d9465f3f6a4c6dab4c124a/src/backend/utils/adt/arrayfuncs.c#L1461 ?

killercup: Yes! It even points to something with postgres in the name! Perfect! 😄

No, but seriously, it helps see where this comes from (and where to look when something changed).

PR #699, opened 2017-02-13
rm TransactionError

sgrif:
This was always a fairly clunky API. Its main upside was that type
inference was able to figure out the error type in more cases than the
current implementation does. It has much larger downsides though. The
most common case is that you're just returning `diesel::result::Error`
anyway, so this enum was just an annoyance. If you weren't returning
that, you were probably returning an application specific error which
has a variant for `diesel::result::Error`, and now also had to handle
this additional type. It also turns out that
[`TransactionError<Box<::std::error::Error>>` doesn't implement
`::std::error::Error`](https://github.com/rust-lang/rust/pull/39792).

The new API might require explicitly stating the error type in a few
more places (in particular `try!(conn.transaction(|| try!(query);
Ok(()))); Ok(...)` stopped inferring), but I think that this will end up
being more ergonomic for most uses.

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -34,8 +34,9 @@ pub trait Connection: SimpleConnection + Sized {
     /// function returns an `Err`,
     /// [`TransactionError::UserReturnedError`](../result/enum.TransactionError.html#variant.UserReturnedError)
     /// will be returned wrapping that value.
-    fn transaction<T, E, F>(&self, f: F) -> TransactionResult<T, E> where
+    fn transaction<T, E, F>(&self, f: F) -> Result<T, E> where
         F: FnOnce() -> Result<T, E>,
+        E: From<Error>,
     {
         let transaction_manager = self.transaction_manager();
         try!(transaction_manager.begin_transaction(self));
@@ -46,7 +47,7 @@ pub trait Connection: SimpleConnection + Sized {
             },
             Err(e) => {
                 try!(transaction_manager.rollback_transaction(self));
-                Err(TransactionError::UserReturnedError(e))
+                Err(e)
             },
         }
     }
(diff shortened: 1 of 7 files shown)


diesel_tests/tests/transactions.rs
```diff
@@ -48,9 +40,9 @@ fn transaction_is_rolled_back_when_returned_an_error() {
     setup_test_table(&connection, test_name);
     let get_count = || count_test_table(&connection, test_name);
 
-    let _ = connection.transaction::<(), (), _>(|| {
+    let _ = connection.transaction::<(), _, _>(|| {
         connection.execute(&format!("INSERT INTO {} DEFAULT VALUES", test_name)).unwrap();
-        Err(())
+        Err(Error::__Nonexhaustive)
```
killercup: It's super weird to use this variant. It's in a test though, so I can see why you did it instead of introducing an arbitrary new variant.

sgrif: Actually this just reminded me that I wanted to add `Error::RollbackTransaction` for those times that you don't have an actual error but you want to rollback the transaction. I may as well do that now.

sgrif:
@killercup re-review?

diesel/src/connection/mod.rs
```diff
@@ -67,7 +68,7 @@ pub trait Connection: SimpleConnection + Sized {
         let mut user_result = None;
         let _ = self.transaction::<(), _, _>(|| {
             user_result = f().ok();
-            Err(())
+            Err(Error::__Nonexhaustive)
```
killercup: This test helper fn should probably also use `RollbackTransaction`

diesel/src/result.rs
```diff
@@ -15,6 +15,11 @@ pub enum Error {
     QueryBuilderError(Box<StdError+Send+Sync>),
     DeserializationError(Box<StdError+Send+Sync>),
     SerializationError(Box<StdError+Send+Sync>),
+    /// You can return this variant inside of a transaction when you want to
+    /// roll it back, but have no actual error to return. Diesel will never
+    /// return this variant unless you gave it to us, and it can be safely
+    /// ignored in error handling.
+    RollbackTransaction,
```
killercup: Nice docs. While you're at it you could also add a sentence or two to the other variants (even if just to say when this error can occur) :)

Issue #621, opened 2017-02-04
Predicates that reference a column on the right side of a left outer join cannot be used


sgrif:
A minimal script to reproduce with Diesel 0.10.0

```rust
#[macro_use] extern crate diesel;
#[macro_use] extern crate diesel_codegen;

use diesel::prelude::*;

table! {
    users {
        id -> Integer,
    }
}

table! {
    posts {
        id -> Integer,
        user_id -> Integer,
    }
}

#[derive(Identifiable, Associations)]
#[has_many(posts)]
struct User {
    id: i32,
}

#[derive(Identifiable, Associations)]
#[belongs_to(User)]
struct Post {
    id: i32,
    user_id: i32,
}

fn main() {
    users::table.left_outer_join(posts::table)
        .filter(posts::id.is_null());
}
```

This affects the usage of any predicate in the query DSL. Ultimately the problem comes down to the fact that we are writing impls like:

```rust
impl<T, QS> SelectableExpression<QS> for NotNull<T> where
    T: SelectableExpression<QS>
```

Since the second parameter of `SelectableExpression` defaults to the SQL type of the expression, and that changes to be nullable if we're on the right side of a left outer join, the where clause resolves to the wrong SQL type.

The fix here is to change the second parameter of `SelectableExpression` to be an associated type, not a type parameter. Either by going forward with #6, or with a smaller change.

PR #706, opened 2017-02-15
Ensure aggregate functions enforce the column is from the right table

sgrif:
While working on #621, I noticed that these impls were incorrect and
could be used to compile an incorrect query. I've corrected the impls
and added the appropriate compile-fail test.

I'm not sure if this was just an oversight or if I intentionally did
this to avoid nullability somewhere. The latter is no longer relevant
since we always make these expressions nullable now.


sgrif:
Blame points to 58b59e93, so I'm guessing it was just an oversight. I should have pressed for a compile-fail test.

sgrif:
Rolled up into #709.

PR #707, opened 2017-02-15
Remove unused bounds and `PhantomData` from `any` and `all`

sgrif:
The phantom data was just plain unneccessary. It was either an oversight
or a bug in older rust versions where `SqlType=Array<ST>` didn't count
as the type being constrained. The `HasSqlType` constraints are
sufficiently covered elsewhere (and frankly, I'm fairly certain that
trait is useless and can be removed). It should be noted that we don't
have a compile-test covering that case though, as pg is the only backend
with additional types.


sgrif:
Rolled up into #709

Issue #104, opened 2016-01-17
Our type system can panic when comparing nullable values


sgrif:
I should note that this is relatively hard to actually demonstrate in normal code. The issue here is that `x = NULL` in SQL returns `NULL`. I've managed to distill it into the following test case:

```rust
#[test]
fn test_comparing_null() {
    use diesel::prelude::*;
    use diesel::select;
    use diesel::expression::AsExpression;
    use diesel::types::{Nullable, Integer};

    let query = select(AsExpression::<Nullable<Integer>>::as_expression(None::<i32>).eq(Some(1)));
    query.get_result::<bool>(&connection()).unwrap();
}
```

The query here is ultimately

```sql
SELECT NULL = 1;
```

The problem is that the sql type of `Eq<T, U>` is always `Bool`, and so we panic on an "unexpected null value" when trying to read it. We need to have the type of `Eq<Nullable<T>, Nullable<U>>` to be `Nullable<Bool>`.

I should also note that this *only* applies to these sorts of expressions appearing in a select clause. It does not affect us in places like `where` clauses. Also as an aside, in places like where clauses, `NULL` is effectively the same as `false` which is part of why I'm leaning towards the second answer.

There are basically two answers here:

- We change the type of the expression to be `Nullable<Bool>` where appropriate. This will almost certainly require specialization, and I'm not actually sure that specialization as proposed today will suffice.
- We have the `impl FromSql<Bool> for bool` treat `NULL` as `false`. I'm actually fine with this, if we can be sure that infix predicates which would otherwise return a boolean are the **only** case where null can bubble up like this. It should be noted that this problem applies to all infix predicates, not just `=`. This also applies to `>` and `AND` for example.

sgrif:
I've raised questions on the specialization RFC related to this. I think the *real* answer under specialization might involve adding a `NotNull` trait and `IsNullable` trait (redundant with `Nullable<T>`), and having the following:

```rust
impl<T, U> Expression for Eq<T, U> where
    T: Expression,
    U: Expression<SqlType=T::SqlType>,
    T::SqlType: NotNull,
{
    type SqlType = Bool;

    fn to_sql(...) { ... }
}

impl<T, U> Expression for Eq<T, U> where
    T: Expression,
    U: Expression<SqlType=T::SqlType>,
    T::SqlType: IsNullable,
{
    type SqlType = Nullable<Bool>;

    fn to_sql(...) { // identical to previous }
}

[...]

PR #709, opened 2017-02-15
Use associated types for `SelectableExpression`

### LINKED ISSUE #621: Predicates that reference a column on the right side of a left outer join cannot be used

sgrif:
Note: This PR rolls up commits from a few others. Review might be easier if it's done one commit at a time.

The `SelectableExpression` trait serves two purposes for us. The first and most important role it fills is to ensure that columns from tables that aren't in the from clause cannot be used. The second way that we use it to make columns which are on the right side of a left outer join be nullable.

There were two reasons that we used a type parameter instead of an associated type. The first was to make it so that `(Nullable<X>, Nullable<Y>)` could be treated as `Nullable<(X, Y)>`. We did this because the return type of `users.left_outer_join(posts)` should be `(User, Option<Post>)`, not `(User, Post)` where every field of `Post` is an `Option`.

Since we now provide a `.nullable()` method in the core DSL, I think we can simply require calling that method explicitly if you want that tuple conversion to occur. I think that the most common time that conversion will even be used is when the default select clause is used, where we can just handle it for our users automatically.

The other reason that we went with a type parameter originally was that it was easier, since we can provide a default value for a type parameter but not an associated type. This turned out to actually be a drawback, as it led to #104. This PR actually brings back aspects of that issue, which I'll get to in a moment.

It's expected that any expression which implements `SelectableExpression<QS>` have a `T: SelectableExpression<QS>` bound for each of its parts. The problem is, the missing second parameter is defaulting to `T::SqlType`, which means we are implicitly saying that this bound only applies for `QS` which does not change the SQL type (anything except a left outer join). This ultimately led to #621.

However, with our current structure, it is impossible to fix #621 without re-introducing at least some aspects of #104. In https://github.com/diesel-rs/diesel/issues/104#issuecomment-172281522 I said that we didn't need to worry about `1 + NULL`, because we didn't implement add for any nullable types. However, I'm not sure I considered joins when I made that statement. The statement applied to joins previously because of that implicit "sql type doesn't change" constraint. This commit removes that constraint, meaning #104 will be back at least when the nullability comes from being on the right side of a left join.

I don't think this is a serious enough issue that we need to immediately address it, as the types of queries which would cause the issue still just don't happen in practice. We should come up with a long term plan for it, though. Ultimately the nullability of a field really only matters in the select clause. Since any operation on null returns null, and you basically want null to act as false in the where clasue, it doesn't matter there.

So one partial step we could take is to break this out into two separate traits. One for the "make sure this is valid given the from clause", and one for the "make this nullable sometimes" case and only constrain on the first one in the where clause. We could then re-add the "sql type doesn't change" constraint on the problem cases, which will bring back aspects of #621, but only for select clauses which is a smaller problem.

I'm not sure if I ultimately want to go the two traits route or not. If nothing else, the problem cases are much more obvious with this commit.  Anywhere that has `type SqlTypeForSelect = Self::SqlType` is likely a problem case when joins are involved. This will make it easier to find all the places to apply a solution when I come up with one that I'm happy with.

One last note about this is that it makes `BoxableExpression` that much less ergonomic to use, since now two associated types have to be specified instead of one. There's not much we can do about that (it might be helped by the two trait path), and I'm really not sure if that type is even useful since we now have the boxed queries DSL.

Fixes #621.

### DIFF
--- a/diesel/src/expression/aliased.rs
+++ b/diesel/src/expression/aliased.rs
@@ -20,9 +20,6 @@ impl<'a, Expr> Aliased<'a, Expr> {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
-pub struct FromEverywhere;
-
 impl<'a, T> Expression for Aliased<'a, T> where
     T: Expression,
 {
@@ -57,7 +54,9 @@ impl<'a, T> QueryId for Aliased<'a, T> {
 // FIXME This is incorrect, should only be selectable from WithQuerySource
 impl<'a, T, QS> SelectableExpression<QS> for Aliased<'a, T> where
     Aliased<'a, T>: Expression,
+    T: SelectableExpression<QS>,
 {
+    type SqlTypeForSelect = T::SqlTypeForSelect;
 }
 
 impl<'a, T: Expression + Copy> QuerySource for Aliased<'a, T> {
--- a/diesel_tests/tests/joins.rs
+++ b/diesel_tests/tests/joins.rs
@@ -128,7 +128,27 @@ fn columns_on_right_side_of_left_outer_joins_are_nullable() {
 }
 
 #[test]
-fn select_multiple_from_right_side_returns_optional_tuple() {
+fn columns_on_right_side_of_left_outer_joins_can_be_used_in_filter() {
+    let connection = connection_with_sean_and_tess_in_users_table();
+
+    connection.execute("INSERT INTO posts (user_id, title) VALUES
+        (1, 'Hello'),
+        (1, 'World')
+    ").unwrap();
+
+    let expected_data = vec![
+        ("Sean".to_string(), Some("Hello".to_string())),
+    ];
+    let source = users::table.left_outer_join(posts::table)
+        .select((users::name, posts::title))
+        .filter(posts::title.eq("Hello"));
+    let actual_data: Vec<_> = source.load(&connection).unwrap();
+
+    assert_eq!(expected_data, actual_data);
+}
+
+#[test]
+fn select_multiple_from_right_side_returns_optional_tuple_when_nullable_is_called() {
     let connection = connection_with_sean_and_tess_in_users_table();
 
     connection.execute("INSERT INTO posts (user_id, title, body) VALUES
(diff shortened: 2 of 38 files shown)


diesel_tests/tests/joins.rs
```diff
@@ -128,7 +128,27 @@ fn columns_on_right_side_of_left_outer_joins_are_nullable() {
...
+    connection.execute("INSERT INTO posts (user_id, title) VALUES
+        (1, 'Hello'),
+        (1, 'World')
+    ").unwrap();
+
+    let expected_data = vec![
+        ("Sean".to_string(), Some("Hello".to_string())),
+    ];
+    let source = users::table.left_outer_join(posts::table)
+        .select((users::name, posts::title))
```
killercup: 🎉

diesel/src/expression/aliased.rs
```diff
@@ -20,9 +20,6 @@ impl<'a, Expr> Aliased<'a, Expr> {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
-pub struct FromEverywhere;
```
sgrif: It used to be, yeah. I don't know when it stopped being used.

diesel/src/expression/aliased.rs
```diff
@@ -57,7 +54,9 @@ impl<'a, T> QueryId for Aliased<'a, T> {
 // FIXME This is incorrect, should only be selectable from WithQuerySource
```
sgrif: Still true.

sgrif:
> By the way, do you think it'd make sense to actually add tests to document/assert the expected, broken behavior?

An example broken case is `users::table.left_outer_join(posts::table).select(users::id + posts::id)`. I'm not sure that it's worth adding a case to the repo for it.

PR #711, opened 2017-02-15
Remove the `Clone` requirement from `StatementCache`

sgrif:
While there wasn't any performance concern from the reference counting,
SQLite and MySQL both use a struct that gets mutated as part of
execution. We want to ensure only one execution is occurring at a time,
and representing that with a unique reference better reflects (and
better enforces) that invariant.

### DIFF
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -1,8 +1,9 @@
 use std::any::TypeId;
 use std::borrow::Cow;
-use std::cell::RefCell;
+use std::cell::{RefCell, RefMut};
 use std::collections::HashMap;
 use std::hash::Hash;
+use std::ops::{Deref, DerefMut};
 
 use backend::Backend;
 use query_builder::*;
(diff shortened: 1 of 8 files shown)


diesel/src/connection/statement_cache.rs
```diff
@@ -111,3 +140,23 @@ impl<DB> StatementCacheKey<DB> where
...
+        Err(e) => {
+            error = Some(e);
+            unsafe { mem::uninitialized() }
+        }
+    });
+    match error {
+        Some(e) => Err(e),
+        None => Ok(result),
+    }
+}
```
killercup: I know we talked about this at length but I' still looking for a version that is just one bit nicer. Good for now, though. (I think `std` should have more `try_` methods. `try_into` is a thing, and `try_map`, `try_unwrap_or_else`, et.al. should be too)

diesel/src/connection/statement_cache.rs
```diff
@@ -111,3 +140,23 @@ impl<DB> StatementCacheKey<DB> where
...
 }
+
+fn refmut_map_result<T, U, F>(refmut: RefMut<T>, f: F)
+    -> QueryResult<RefMut<U>> where
+        F: FnOnce(&mut T) -> QueryResult<&mut U>,
+{
+    use std::mem;
+
+    let mut error = None;
+    let result = RefMut::map(refmut, |mutref| match f(mutref) {
```
killercup: I'd probably rename `f` to `mapper`

diesel/src/connection/statement_cache.rs
```diff
@@ -36,30 +37,58 @@ impl<DB, Statement> StatementCache<DB, Statement> where
...
+    Cached(RefMut<'a, T>),
+}
+
+impl<'a, T> Deref for MaybeCached<'a, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        match *self {
+            MaybeCached::CannotCache(ref x) => x,
+            MaybeCached::Cached(ref x) => &**x,
```
killercup: This is how I program in C, too! Add `&` and `*` until it works :trollface: 

FTR, this deref's the RefMut, then its content, and returns a reference to that.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -67,17 +67,19 @@ impl Connection for SqliteConnection {
         Self::Backend: HasSqlType<T::SqlType>,
         U: Queryable<T::SqlType, Self::Backend>,
     {
-        let statement = try!(self.prepare_query(&source.as_query()));
-        let mut statement_ref = statement.borrow_mut();
-        StatementIterator::new(&mut statement_ref).collect()
+        let mut statement = try!(self.prepare_query(&source.as_query()));
+        let statement_use = StatementUse::new(&mut statement);
+        let iter = StatementIterator::new(statement_use);
+        iter.collect()
```
killercup: I assume Rust doen't let you to continue to use `StatementIterator::new(statement_use).collect()` because temporary vars with too short a scope?

diesel/src/mysql/backend.rs
```diff
@@ -4,16 +4,17 @@ use backend::*;
 use query_builder::bind_collector::RawBytesBindCollector;
 use super::query_builder::MysqlQueryBuilder;
 
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub struct Mysql;
 
-#[allow(missing_debug_implementations, missing_copy_implementations)]
+#[allow(missing_debug_implementations)]
```
killercup: Why put that attr above and the derive below the doc string? No, wait, let me guess: Because doc strings are in fact `#[doc=…]` and you wanted it sorted alphabetically? 😄

diesel/src/connection/statement_cache.rs
sgrif: I think there's a more fundamental issue with how types like `Option` and `Result` are able to compose. Adding more `try_` methods is a bandaid IMO. I think this is most apparent with deref coercions. If I have a `String` and want to call a method that takes `&str` I just write `&s`. If I have an `Option<String>` and want to call a method that takes an `Option<&str>` I have to write `s.as_ref().map(|x| &**x)`. Pattern matching also makes the issue quite apparent.

diesel/src/sqlite/connection/mod.rs
sgrif: Yeah, it claims that `statement` is dropped with a life reference. Which is wrong, borrowck should know is wrong (even without non-lexical lifetimes), and I think is probably a bug in rustc. FWIW this also compiles:

```rust
let x = StatementIterator::new(statement_use).collect();
x
```

But clippy doesn't like that one.

diesel/src/mysql/backend.rs
sgrif: Honestly, no real reason. `#[derive()]` feels more attached to the item, while `#[allow]` feels more attached to the block of code.

diesel/src/connection/statement_cache.rs
sgrif: Actually it derefs the reference to the `RefMut`, and then the `RefMut` itself.

```
x: &RefMut<T>
*x: RefMut<T>
**x: T
&**x: &T
```

PR #713, opened 2017-02-15
Add chrono support for MySQL

sgrif:
Beyond the fact that the `ToSql` impl for `ffi::MYSQL_TIME` is
incredibly hacky, there's not a ton to say about this. I have no clue
what the fuck a negative date is, and the documentation doesn't state it
anywhere, so I'm just erroring if we get one. The range of dates that
MySQL accepts are quite small, so we're not worried about chrono
overflowing this time.


sgrif:
> short of adding quickcheck for this

This PR includes tests which use quickcheck for this. :) https://github.com/diesel-rs/diesel/pull/713/files#diff-0d44c4b35c16133e20e082de4f8fc12cR114

killercup:
> This PR includes tests which use quickcheck for this.

Heh, I even looked for this but misremembered what the macro did

PR #714, opened 2017-02-15
Don't panic when running empty migrations.

Eijebong:
PGRES_EMPTY_QUERY is not an error case.

https://github.com/postgres/postgres/blob/93e6e40574bccf9c6f33c520a4189d3e98e2fd1f/src/interfaces/libpq/fe-exec.c#L177

I'm not sure if we should add all the non error cases here (I have no idea how to test the others)

### DIFF
--- a/diesel_cli/tests/migration_run.rs
+++ b/diesel_cli/tests/migration_run.rs
@@ -86,3 +86,25 @@ fn migration_run_inserts_run_on_timestamps() {
     assert!(valid_run_on_timestamp(&db),
             "Running a migration did not insert an updated run_on value");
 }
+
+#[test]
+fn empty_migrations_are_not_valid() {
+    let p = project("migration_run_empty")
+        .folder("migrations")
+        .build();
+
+    p.command("setup").run();
+
+    p.create_migration(
+        "12345_empty_migration",
+        "",
+        ""
+    );
+
+    let result = p.command("migration")
+        .arg("run")
+        .run();
+
+    assert!(!result.is_success());
+    assert!(result.stdout().contains("empty migration"));
+}
--- a/diesel/src/migrations/migration.rs
+++ b/diesel/src/migrations/migration.rs
@@ -104,6 +104,11 @@ fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigr
     let mut sql = String::new();
     let mut file = try!(File::open(path));
     try!(file.read_to_string(&mut sql));
+
+    if sql.is_empty() {
+        return Err(RunMigrationsError::EmptyMigration);
+    }
+
     try!(conn.batch_execute(&sql));
     Ok(())
 }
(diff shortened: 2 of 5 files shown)


Eijebong:
Hu... Should I just add a #[feature(postgres)] on the test ? Or transform this case into an error ?

sgrif:
I think we should special case this in the migrations code. It should still be an error (but one where the CLI gives a reasonable message)

Eijebong:
I didn't think about it but an empty query on pg will panic event outside of migrations.

sgrif:
Diesel should never generate an empty query outside of migrations. It shouldn't panic, but it should be an error.

sgrif:
So there's two parts to this. The first is that we're panicking when `PGRES_EMPTY_QUERY` because we don't have an error message. I agree that we need to avoid this. However, receiving an empty query is pretty much always an error case. Empty migrations are the only way to trigger it through the public API. If it were to happen in any other case, it'd be a bug in Diesel. So I think we should add

```rust
PGRES_EMPTY_QUERY => {
    let error_message = "Received an empty query".to_string();
    Err(Error::DatabaseError(DatabaseErrorKind::__Unknown, Box::new(error_message)))
}
```

Additionally, I think we should add an `EmptyMigration` variant to `MigrationError`, and give a slightly more specific error message like "Attempted to run an empty migration.". Eventually I want to revamp how we handle errors there, and the message can eventually be improved to something like "Attempted to apply migration 348975234957_name, but `up.sql` was empty.", but that's out of scope for now.

diesel_cli/tests/migration_run.rs
```diff
@@ -86,3 +86,24 @@ fn migration_run_inserts_run_on_timestamps() {
...
             "Running a migration did not insert an updated run_on value");
 }
+
+#[test]
+fn empty_migrations_are_not_valid() {
+    let p = project("migration_run_empty")
+        .folder("migrations")
+        .build();
+
+    // Make sure the project is setup.
```
killercup: I don't think this comment adds much to `command("setup").run`

diesel_cli/tests/migration_run.rs
```diff
@@ -86,3 +86,24 @@ fn migration_run_inserts_run_on_timestamps() {
...
+    p.command("setup").run();
+
+    p.create_migration("12345_empty_migration",
+                       "",
+                       "");
```
killercup: Can you write it like this?

```rust
p.create_migration(
    "12345_empty_migration",
    "",
    "",
);
```

diesel_cli/tests/migration_run.rs
Eijebong: Yeah, It's a copy/paste from the test above, i'll remove it ^^

diesel_cli/tests/migration_run.rs
Eijebong: Sure

Eijebong:
Eh, when you add your changes before commiting it's better...

killercup:
Thanks again!

PR #717, opened 2017-02-16
Nicer error when CLI can't parse database URL

killercup:


diesel_cli/src/database.rs
```diff
@@ -34,11 +34,21 @@ impl Backend {
                 Backend::Mysql,
             #[cfg(feature="sqlite")]
             _ => Backend::Sqlite,
-            #[cfg(not(feature="sqlite"))]
+            #[cfg(all(not(feature="sqlite")), all(feature="postgres", feature="mysql"))]
             _ => {
-                panic!("{:?} is not a valid database URL. It should start with \
+                panic!("`{}` is not a valid database URL. It should start with \
```
sgrif: Oh you

diesel_cli/src/database.rs
```diff
@@ -34,11 +34,21 @@ impl Backend {
...
                         `postgres://` or `mysql://`", database_url);
+            },
+            #[cfg(all(not(feature="sqlite")), feature="postgres")]
```
sgrif: I wonder if we can use `cfg_if!` here? If not, you'll need this to be `all(feature="postgres", not(feature="mysql"))` to avoid warnings.

killercup:
No idea I went with the most complicated way possible and tried to play human SAT solver. The code is much simpler now.

PR #718, opened 2017-02-16
Add mysql examples

Eijebong:
They're the same as the postgres ones but adapted to mysql.


examples/mysql/getting_started_step_3/src/bin/publish_post.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    let id = args().nth(1).expect("publish_post requires a post id")
+        .parse::<i32>().expect("Invalid ID");
+    let connection = establish_connection();
+
+    let post: Post = posts.find(id)
+        .first(&connection)
+        .expect(&format!("Unable to find post {}", id));
+
+
```
sgrif: ✂️

examples/mysql/getting_started_step_3/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+    let new_post = NewPost {
+        title: title,
+        body: body,
+    };
+
+    diesel::insert(&new_post).into(posts::table)
+        .execute(conn)
+        .expect("Error saving new post");
+
+    posts::table.order(posts::id.desc()).first(conn).unwrap()
```
sgrif: Probably worth doing `use schema::posts::dsl::*;` for this

Eijebong: If we `use schema::posts::dsl::*` then we need to change the `body` and `title` variable names;

I added `use schema::posts::dsl::{posts, id};` instead

examples/mysql/getting_started_step_1/src/bin/show_posts.rs
```diff
@@ -0,0 +1,23 @@
+extern crate diesel_demo_step_1_mysql;
+extern crate diesel;
+
+use diesel_demo_step_1_mysql::*;
+use diesel_demo_step_1_mysql::models::*;
```
sgrif: Let's replace `diesel_demo_step_x_mysql` with `self` on all of the lines other than `use diesel_demo_step_x_mysql::*`

examples/mysql/test_all
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -e
+
+export DATABASE_URL="mysql://travis@localhost/diesel_test"
```
sgrif: We shouldn't hard-code the `travis` user here. I'd leave the user off and see if CI complains. If it does, let's only set this if `TRAVIS_RUST_VERSION` is unset.

sgrif:
Still need to add this to travis.

Eijebong:
Ok, I had to change the database name to `diesel_example`, otherwise another test failed later but now tests are passing :)

PR #721, opened 2017-02-17
Ensure deterministic ordering of inferred SQLite tables

Mrmaxmeier:
The output generated by `diesel print-schema` is not stable across database environments.
I guess it's sensible to sort the tables alphabetically to be able to diff schemas.

### DIFF
--- a/diesel_cli/tests/print_schema.rs
+++ b/diesel_cli/tests/print_schema.rs
@@ -119,3 +119,65 @@ r"table! {
 ");
     }
 }
+
+#[test]
+fn run_infer_schema_order() {
+    let p = project("print_schema_order").build();
+    let db = database(&p.database_url());
+
+    // Make sure the project is setup
+    p.command("setup").run();
+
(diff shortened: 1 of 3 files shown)


sgrif:
Can we do the sorting in the actual queries that we execute instead?

Mrmaxmeier:
Sure, that makes sense and is consistent with [the code that does this ordering on Postgres/MySQL already](https://github.com/diesel-rs/diesel/commit/9ed311f97b2b6d5de0808c57aaec735c56d058e5).
I've added the same behavior to the SQLite variant. Should I keep the tests and move them to `diesel_tests`?

killercup:
Yeah, that's way better.

@Mrmaxmeier, I'd keep the cli test and add a small unit test to `diesel_infer_schema/src/sqlite.rs` as well. (Do the others have tests for that? If not, maybe you can add them for PG and MySQL too).

Mrmaxmeier:
I've added unit tests for SQLite, PG and MySQL and Travis seems happy with them.

killercup:
Thanks, great work!

PR #723, opened 2017-02-17
Ensure `on_conflict_do_nothing` works with slices

sgrif:
Due to the various signatures, the end result of
`slice.on_conflict_do_nothing()` was `&OnConflictDoNothing<&&[T]>`,
which doesn't implement the various traits. The solution is to remove
the sized contraint from the extension method so that `[T]` implements
it instead of `&[T]`.


sgrif:
Not that I'm aware of, but I'll keep an eye out.

PR #724, opened 2017-02-17
Remove the final type parameter from `IntoInsertStatement`

### LINKED ISSUE #590: Can't write IntoInsertStatement in constraints
Direct cause is that the types in query_builder::returning_clause are not exported

sgrif:
The type `NoReturningClause` is actually a voldemort type outside of our
crate. We could fix this by adding the default back to the type
parameter, but the parameter is never really used so we can just remove
it.

Fixes #590.


sgrif:
lol

PR #727, opened 2017-02-17
Ensure that `MysqlConnection` is `Send`

### LINKED ISSUE #726: MysqlConnection doesn't work with r2d2-diesel
`MysqlConnection` can't be shared between threads, which makes it impossible to use in `r2d2-diesel` (or presumably any multi-thread environment). This is in diesel `0.11.0`.

My code looks like:

```
fn connection_pool() -> ::err::Result<Pool> {
    let url = &::utils::config::CONFIG.db.url;
    let config = ::r2d2::Config::default();
    let manager = ConnectionManager::<MysqlConnection>::new(url);
    Ok(::r2d2::Pool::new(config, manager)?)
}
```

which fails to compile with (truncated):

```
within `diesel::mysql::connection::stmt::Statement`, 
  the trait `std::marker::Send` is not implemented for `*mut mysqlclient_sys::st_mysql_stmt`
note: `*mut mysqlclient_sys::st_mysql_stmt` cannot be sent between threads safely
```

I don't know if just adding `unsafe impl Send for MysqlConnection {}` would fix it (if might not be safe or even compile).

sgrif:
`mysql::Statement` is not send due to the raw pointer. The `MYSQL_STMT`
type is very much unsafe to be sent across threads, since it contains a
pointer to the connection which is not thread safe. However, we never
expose statements directly, and sending the whole connection is safe.

I've also just made it a supertrait of `Connection` so I don't make this
mistake again.

Fixes #726.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * `.on_conflict_do_nothing()` now interacts with slices properly.
 
+* `MysqlConnection` now implements `Send`, which is required for connection
+  pooling.
+
 ## [0.11.0] - 2017-02-16
 
 ### Added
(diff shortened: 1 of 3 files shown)


sgrif:
File descriptors and sockets are fine across threads. They're not fine across forks.

PR #731, opened 2017-02-18
Order by id in filter tests

martijndeh:
These tests would sometimes fail on my machine. It's because we're comparing the return values of the queries without specifying an explicit order. If no order is specified, Postgres simply returns the order it finds the fastest to produce. In practice, this would often work, but sometimes the order would be different and the tests would fail.


sgrif:
Can you rebase onto master? I've fixed the build there (releasing 0.11.1 broke it lol)

martijndeh:
I rebased, awaiting Travis results...

killercup:
Thanks again! Sorry it took a while to merge this :)

Issue #609, opened 2017-02-03
Add #[must_use] for warnings


durango:
@sgrif: `I should add a #[must_use] on queries to prevent that mistake. Anyone mind opening an issue?`

@sgrif: `Making query.filter(name.like(user.name)); warn`

alexcameron89:
I think this is fixed and good to close.

sgrif:
Yup. Thanks!

TimNN:
Is this really fixed? The one PR linking this issue only mentions `select` and `filter` and also states:

> But we should probably also add this to/think about insert, update and delete before we close that issue.

(I also just spend some time trying to figure out why my `update` wasn't executing until I noticed I was missing an `.execute` when adding an `diesel::debug_sql`).

sgrif:
Looks like we need an annotation on `UpdateStatement` and friends. Feel free to open a PR

AndrewSpeed:
I'm happy to put a PR up for this, if no one else is working on it

weiznich:
@AndrewSpeed Feel free to take it. If there is any problem just ask here or (preferable) in our gitter channel.

PR #732, opened 2017-02-18
Add `must_use` to select and filter

killercup:
See #609. But we should probably also add this to/think about insert, update and delete before we close that issue.


sgrif:
Rebase for CI?

killercup:
Rebased. All green.

PR #738, opened 2017-02-19
Bump pq-sys and mysqlclient-sys

sgrif:
I've released new versions of these libraries that have pre-generated
bindings rather than generating them at build time. Requiring a bleeding
edge libclang caused far too many issues.


sgrif:
Once the build is green I'm going to backport this to 0.11.x and release a new version.

killercup:
> error: package `pq-sys v0.4.0` specifies that it links to `pq` but does not have a custom build script

PR #740, opened 2017-02-19
Adds ForeignKeyViolation for DatabaseErrorKind

durango:


sgrif:
Can you add a PG specific test which verifies that all the error fields are correct as well?

migrations/mysql/20170219200159_add_foreign_key/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE fk_inits (
+    id INTEGER PRIMARY KEY
+);
+
+CREATE TABLE fk_tests (
+    id INTEGER PRIMARY KEY,
+    fk_id INTEGER NOT NULL REFERENCES fk_inits(id)
```
sgrif: MySQL accepts this as valid syntax (since it's part of the SQL standard), but it's ignored. The proper way to create a foreign key in MySQL is

```
fk_id INTEGER NOT NULL,
FOREIGN KEY (fk_id) REFERENCES fk_inits (id)
```

PR #742, opened 2017-02-20
cli: Add "db" as alias for "database" subcommand

theduke:
cli: Add "db" as alias for "database" subcommand.

It can be quite annoying to type out "database", just "diesel db" is more convenient.

### DIFF
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -46,6 +46,7 @@ pub fn build_cli() -> App<'static, 'static> {
                 specified in your DATABASE_URL, and runs existing migrations.");
 
     let database_subcommand = SubCommand::with_name("database")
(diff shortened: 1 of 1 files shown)


sgrif:
Can you add tests and a changelog entry?

theduke:
Will do.

theduke:
@sgrif regarding the changelog, should I add it to `0.11.3` and leave the date empty?

sgrif:
No, just add an `## Unreleased` header if there isn't one already.

sgrif:
(This won't go into 0.11.3, it's a feature not a bugfix).

killercup:
Merged as part of #804

theduke:
@killercup  thanks for making up for my laziness!

Issue #650, opened 2017-02-09
Provide a better "raw sql" escape hatch for full queries that we don't support in the query builder


sgrif:
Today we have some form of an escape hatch in the form of the `sql` function. However, it's meant for use with small fragments, not full queries. It's limited because it requires specifying the full SQL type of the output, and cannot handle bind parameters. I'd like to introduce a new function that is a more ergonomic API for raw SQL as an escape hatch for queries we don't support in the query builder yet.

Ultimately I've resisted introducing this, as we want to support all possible safe queries in the query builder eventually, but I think we need a proper escape hatch for us to go 1.0.

I'd like this API to be fully untyped, and named to make it clear that no verification of the query is occurring. For deserialization, it should not go through `Queryable`, as I do not think enforcing column order is ergonomic when writing raw SQL (and `Queryable` needs a SQL type to work). Instead I'd like to introduce an `UntypedQueryable`, which takes a row directly, and is able to fetch fields by name. I'm fine with the API requiring specifying the SQL type on this call if type inference can't pick it up. e.g. `let updated_at: NaiveDateTime = row.get::<Timestamp>("updated_at")?`. I suspect for most types which only implement `FromSql` for a single DB type type inference will handle it, but we should explore our options here and see what feels best.

I'm less clear what we should do with bind parameters. I do know that I don't want to go the route that the postgres crate went where we take `&[&UntypedToSql]`, as constructing that value is painful, it forces dynamic dispatch (which then usually also forces boxing), and requires `&[]` for queries with no bind parameters. We could potentially still pass tuples here as long as type inference does what we want. I also think `conn.raw_query("SELECT * FROM users WHERE id = $1").bind(id)` could work (as an aside, I think we will likely need to pass the connection in before the bind parameters for this to work, which will lead to a semi-weird api of `conn.raw_query().bind().execute()`). I'm fine with also sometimes requiring giving a SQL type for the bind here as long as type inference is able to handle the most common cases.

andy128k:
I'm new to Rust and Diesel. but what I love is type safety and all great things LoadDsl does. That's why I'd prefer "better `sql`" instead of new untyped API.
Nice ORMs like ActiveRecord and Django's ORM allows to put bits of SQL but keep the rest of code working with typed models.

I'm playing with Diesel and see a lot of limitations (like inability to make multiple joins or multiple belongs_to) and that's ok. Even with raw SQL Diesel is still great.

I created couple workarounds to solve issues with raw sql queries.
[...]

PR #747, opened 2017-02-22
Remove the usage of tuples in argument positions from our API

sgrif:
# We are seeking feedback

The core team hasn't decided if we actually want to go forward with this proposal. We are seeking feedback from the community on their thoughts. The purpose of this PR is to provide a place to centralize discussion, and provide the ability for people to try out the proposal against their code bases. We would like to hear from you on how much this change impacts your codebase, how painful migrating to this would be, and whether you'd find this confusing as a newcomer.

## Overview

This pull request changes any functions which could previously take a tuple as an argument to take an hlist instead. The TL;DR is that the rule of thumb for "things that are separated by a comma in SQL are represented as _ in diesel" changes from "a tuple" to "an hlist". This means that `(x, y)` changes to `hlist!(x, y)` (we may land on a different name, see the *Unanswered Questions* section below).

For those unfamiliar, "hlist" is short for "heterogeneous list". It is essentially a singly linked list where each element is able to be of a different type. The list encodes this in its own type. It is effectively the same as a bunch of nested pairs. e.g. `Cons(1, Cons(2, Cons(3, Nil)))` is more-or-less the same as `(1, (2, 3))` or `(1, (2, (3, ())))`.

The specific functions that change are [`query.select`](http://docs.diesel.rs/diesel/prelude/trait.SelectDsl.html#tymethod.select), [`query.order`](http://docs.diesel.rs/diesel/prelude/trait.OrderDsl.html#tymethod.order), and [`update(table).set(...)`](http://docs.diesel.rs/diesel/query_builder/update_statement/struct.IncompleteUpdateStatement.html#method.set) when updating multiple columns but not using a struct that has `#[derive(AsChangeset)]`.

A less visible API that is changed is the type argument to the [`sql`](http://docs.diesel.rs/diesel/expression/dsl/fn.sql.html) function. However, this function may be deprecated and removed before 1.0 in favor of an escape hatch API that is untyped (see #650).

This also affects implementations of [`Queryable`](http://docs.diesel.rs/diesel/query_source/trait.Queryable.html), [`AsChangeset`](http://docs.diesel.rs/diesel/query_builder/trait.AsChangeset.html), and [`Insertable`](http://docs.diesel.rs/diesel/prelude/trait.Insertable.html). However, these are usually derived rather than implemented manually (with the possible exception of `Queryable`).

This does not affect the ability to use tuples in a *return type* position. This means that doing `.load::<(A, B)>` will still work, and associations still work in terms of tuples. This is purely due to how likely this change is to affect the average user. An application can be doing extremely complex queries, and still never call `.select` or `.order` explicitly with more than one argument. However, any time a join is done, or associations are loaded, tuples get involved in a much more visible fashion. Associations in particular are designed to work around `Iterator::zip` and `Iterator::unzip`, which is only feasible if they work on tuples.

Additionally, while code that is passing a tuple can be changed to pass an hlist simply by putting `hlist!` in front of it, the story is more involved for code working with tuples. Hlists cannot simply be indexed like tuples can, they have to instead be destructured. This is acceptable for us to do internally, but not something I'd like to impose on every user of the library.

## What this gets us

First and foremost, the code is simpler, and our compile time goes down. To implement a trait for an arbitrarily sized hlist, you need to provide two impls: `Cons<T, U> where T: SomeTrait, U: SomeTrait` and `Nil`. (In some places we do `Cons<T, Cons<U, Tail>>` and `Cons<T, Nil>` instead when we have things like comma separation to worry about).

Implementing code for an arbitrary sized tuple is much harder. You basically have to write the impls as a macro, and invoke it for whatever size you want to support. This is what the standard library does for traits like `PartialEq` (which aren't implemented on tuples larger than 12 elements). This change removes *half* of the code in the main crate (as measured by `cargo rustc -- -Zunstable-options --pretty=expanded | wc -l`), and decreases compile times by 70% (measured by `touch src/lib.rs && cargo build`). Moving this code out of macros and into regular rust will also make it much easier for new contributors to understand.

This may also open the door for several future improvements that would not be possible with tuples. Several blanket impls that we want to write are impossible because they overlap with the impls for tuples. At the moment most of them overlap with hlists as well, but specialization and/or future improvements to specialization are easier to apply to hlists than tuples.

In particular, I think that this change will eventually allow us to remove the requirements that struct order match the column order (in cases where your struct is one-to-one with a table), and also remove the requirement that you must have a field for your struct for every member of the select statement. Right now we would be unable to write the impls required to make that happen, but several of the possibilities being discussed for specialization would allow us to make that happen. If you're interested in the details of that, I've written them up [here](https://gist.github.com/sgrif/213e7453bc5518596c52cf99b1b80fd7).

You may notice that some of these gains are all assuming speculative future enhancements to the language. Why not wait until they actually happen to make this change? Ultimately we want to release 1.0 before those changes are likely to happen. If we want to avoid breaking backwards compatibility, we need to leave the door open to take advantage of those changes when they happen. Even if we limit the "wins" to the more accessible code and improved compile times, I do think it's still worth it.

## Why didn't we use an existing library for this?

Ultimately if we make this change, hlists become a very fundamental part of people interacting with Diesel, which makes me uncomfortable leaving that in the hands of another crate unless it's already well established within the community. There also isn't a crate that is maintained and *just* does hlists. Having a type which is local to Diesel is useful for us from a coherence point of view.

## Drawbacks

The most obvious drawback is that we take tuples instead of hlists. Tuples are in the standard library. hlists aren't. That said, I don't know that people are actually discovering that they can pass tuples on their own. Two common questions in gitter are "how do I order by multiple things", and "how do I update multiple columns without using a struct". We can fix that by giving some more up front examples of those cases, but if people aren't finding tuples without being shown, hlists aren't really any worse. They're just equally bad. Ultimately what we really want is a variadic function, but that's not going to happen.

Additionally, this will make some of our scary type signatures more scary. `diesel::hlist::Cons<Foo, diesel::hlist::Cons<Bar, diesel::hlist::Nil>>` is harder to read than `(Foo, Bar)`. That said, the type would really only show up in errors that were already incomprehensible. Nobody is actually making sense of `SelectStatement<five, terminal, windows, of, types>` other than the core team when people paste them in gitter asking for help. Our goal has been to structure things so that the types don't get dumped in error messages, and will continue to be.

Some of our error messages got... well I'm not sure if it's worse, or better. For example, the error for the table having one more column than the struct has changed from `(i32, String, String): FromSqlRow<(Integer, Text, Text, Double)> failed` to `Nil: FromSqlRow<Cons<Double, Nil>> failed`. One could argue that without reading the documentation, a person might actually figure out what's going on from the first error message, but not from the second. In practice I'm not sure that's true though, and it's a lot easier to *document* that `Nil: FromSqlRow` means you have more columns on the table than the struct than it is for us to say "hey make sure you count the tuple elements if you see an error about a tuple and another tuple".

## Alternatives

We could also keep this change entirely internal, provide an `IntoHlist` trait, and implement that for tuples of arbitrary sizes. This is a bad alternative for 2 reasons. First, it will bring back the "we need to implement things for *really* large tuples" problem, and hurt compile times. Second, since these functions can also take just one argument, we will need to brute force that trait for a lot of types, including some which are potentially inconvenient to do so. Specialization may improve that situation when it stabilizes.

## Unanswered Questions

Naming. `hlist!` could conflict with other libraries, so we'd want to have the canonical name be `diesel_hlist!` (though I want to keep the short variant. People can choose not to import the short version if they don't want to). However, do we actually want to make the fact that we're using hlists apparent? Would a name like `args!` or `columns!` make more sense?

## How we teach this

We should almost certainly add an example in the getting started guide which demonstrates how hlists work.


hiimtaylorjones:
I think restrictions and drawbacks are worth it for what we gain. It'll definitely require a lot of work to redo a lot of guide documentation, but it'll hopefully make Diesel better in the long run!

sgrif:
> It'll definitely require a lot of work to redo a lot of guide documentation

Ironically the guide is 100% unaffected by this change. We never call `.select` or `.order` in the guide (and even if we did it would probably be with 1 argument since that's the most common case)

Drakulix:
A quick grep on my two largest code bases using diesel (both deployed at work and the only ones I really care about, because the rest is hobby and I would always happily accept using better code, even if it breaks compatibility) showed that I am not using tuples anywhere.

I barely use `select`, I mostly use `filter`, `update().set(...)` always with the full object (or use `.save()` instead) and order never with more then one argument.

This is probably a result from my programming background, I was mostly used to OOP before using Rust, which I would not necessarily call object orientated. I am not a big fan of databases and try to avoid them where not favorable in terms of performance and that probably results in trying to use common code patterns, when dealing with DBs.

That's why I keep most operations simple (get & set) and only use queries where I need to.
And diesel makes this very comfortable.

Regarding documentation I would definitely say the examples can be improved not only in regards to `hlist!`, but simply to provide more examples. I would maybe use more complex queries, if it wouldn't be so difficult and annoying to browse all those traits to find out, what I can use in combination with what else. If it does not provide an obvious performance penalty I mostly don't use, what else diesel has to offer.

diesel/src/hlist.rs
```diff
@@ -0,0 +1,88 @@
+use std::fmt::{Debug, Formatter, Error as FmtError};
+
```
killercup: We should probably add some (module-level) docs here. At least for contributors looking to understand diesel's inner workings. [Like this](https://github.com/lloydmeta/frunk/blob/4d688a460115474d915be7a7d351b508c478ab96/core/src/hlist.rs) maybe.

diesel/src/insertable/hlist_impls.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl<Col, Expr, Tail, DB> InsertValuesRecursive<DB>
+    for Cons<ColumnInsertValue<Col, Expr>, Tail> where
+        DB: Backend + SupportsDefaultKeyword,
+        Col: Column,
+        Col::SqlType: IntoNullable,
+        Expr: Expression<SqlType=<Col::SqlType as IntoNullable>::Nullable> + QueryFragment<DB>,
+        Tail: InsertValuesRecursive<DB>,
+{
+    fn column_names(&self, comma_needed: bool, out: &mut DB::QueryBuilder) -> BuildQueryResult {
```
sgrif: Nice

diesel/src/hlist.rs
sgrif: Yes, the whole thing will need a lot of documentation if merged.

Fiedzia:
I am planning to use diesel, but I haven't yet, so +1 for changing it now. Compile time matters to me.
Is the columns! as intuitive as I think (not being familiar with diesel)? In that case I am definitely for it.

cramertj:
I'd prefer Diesel continue to use tuples, but that there was an easier way to impl traits for tuples. As I commented on [the RFC](https://github.com/rust-lang/rfcs/pull/1921), I want to see `Split` and `Cons` traits added to the language and implemented for all tuples through some sort of intrinsic. @eddyb has a working prototype [here](https://play.rust-lang.org/?gist=030b0b74b3a72eecd09d567368a592bb&version=nightly) (this implementation uses macros, but that would be unnecessary if an intrinsic of some kind was provided).

I believe this solution mitigates many of the drawbacks you listed above, namely:
- "Tuples are in the standard library. hlists aren't."
This approach would allow users to keep using tuples, a familiar type, and to avoid introducing `hlist!` macros all over the place.
- "`diesel::hlist::Cons<Foo, diesel::hlist::Cons<Bar, diesel::hlist::Nil>>` is harder to read than `(Foo, Bar)`"
Since this approach allows users to use tuples directly, they don't have to try and parse new cons-cell types.

Overall, I think the addition of `Split` and `Cons` traits offers the best of both worlds: familiar std types, Rust syntax, smaller generated code, faster compile times, and a standard way to handle heterogeneous lists in Rust.

PlasmaPower:
> Ultimately if we make this change, hlists become a very fundamental part of people interacting with Diesel, which makes me uncomfortable leaving that in the hands of another crate unless it's already well established within the community. There also isn't a crate that is maintained and just does hlists. Having a type which is local to Diesel is useful for us from a coherence point of view.

Maybe we should create a crate `hlists` under the `diesel-rs` GitHub user, for both diesel and whoever else wants `hlists`.

tupshin:
I am hugely in favor of moving to HLists in general. Regarding an internal HList vs using an existing crate, I partially agree with your reasoning. However, it is worth noting that [frunk-core](https://crates.io/crates/frunk_core) actually [consists](https://docs.rs/frunk_core/0.0.1/frunk_core/) of not much more than HList.

So while it might not be the right time to standardize on an external one, worth keeping on eye on that one.

PlasmaPower:
I agree that frunk-core would be a reasonable solution, but I do have my concerns. First, frunk isn't currently all that popular. It only has about 4% of the downloads of diesel. Therefore, it's a lot less likely to be properly maintained. Second, at the moment frunk-core exists solely for frunk. It is subject to change, and later on it may expand its scope, adding unnecessary baggage to diesel projects.

While these might not seen like huge concerns, I am worried because changing the hlists backend would be a breaking change for pretty much all diesel projects (which we would want to avoid if possible). I guess we could re-export hlists to avoid a lot of the breakage, but it's still far from optimal.

sgrif:
Yeah, @PlasmaPower that sums up my concerns with that crate pretty nicely.

tupshin:
Totally fair points.

withoutboats:
One thing that's interesting about hlists (but possibly not relevant to diesel) is the difficulty in pulling a particular type out of an hlist. Today, you need a trait that's indexed by peano numbers.

With intersection impls, if you don't care about being able to pull multiple values of the same type out of an HList, it becomes possible without that (but, again, it has the restriction that you essentially 'leak' any existing `T` if you put another `T` in the list). Here's a code sample of how it works:

```rust
trait Hlist { }

struct Cons<T, U: Hlist>(T, U);

impl Hlist for () { }
impl<T, U> Hlist for Cons<T, U> { }

// Base case
impl<T, U> AsRef<T> for Cons<T, U> where U: Hlist {
    default fn as_ref(&self) -> &T {
        &self.0
    }
}

// Recursive case
impl<X, T, U> AsRef<X> for Cons<T, U> where U: Hlist + AsRef<X> {
    default fn as_ref(&self) -> &X {
        self.1.as_ref()
    }
}

// Intersection impl
impl<T, U> AsRef<T> for Cons<T, U> where U: Hlist + AsRef<T> {
    fn as_ref(&self) -> &T {
        self.1.as_ref()
    }
}
```

sgrif:
[I did link to an explanation on that too](https://gist.github.com/sgrif/213e7453bc5518596c52cf99b1b80fd7) Your explanation is much shorter though. :wink:

J-F-Liu:
In favor of this change, the main concern is how much impact of runtime performance.
For naming use a small word is better, like `cols!`, `fields!`.

lloydmeta:
@all Just chiming in because I was summoned.

Regarding frunk (and frunk-core): I'm open to any suggestions of re-organising the repository so that it can be made useful. That also includes keeping the scope of it small.

I also intend to maintain it for the foreseeable future, but more maintainers are always welcome.

blakepettersson:
Is this change a prerequisite for being able to arbitrarily construct selects and order bys based upon run-time parameters (without having to directly fall back to SQL)? It's not currently clear to me how this works in Diesel [today](https://stackoverflow.com/questions/42557528/querying-a-diesel-table-with-dynamic-parameters).

sgrif:
@blakepettersson If you want to order something dynamically, you need to either [box the order clause](http://docs.diesel.rs/diesel/expression/expression_methods/global_expression_methods/trait.ExpressionMethods.html#method.asc), or [box the query](http://docs.diesel.rs/diesel/prelude/trait.BoxedDsl.html). The latter is more common. The same is true for varying select clauses (though keep in mind that the select clause for each branch has to have the same SQL type)

theduke:
Any movement on this?

Has it been decided to go forward with hlists?

Also, @sgrif , as I understand it, this would "fix" the restrictions on number of columns in a table, right?

sgrif:
This is unlikely to go forward until we at least know which direction the variadic generics RFC goes.

Issue #680, opened 2017-02-11
Support PG money


sgrif:
We get enough questions/bug reports where people are storing a column called `price` as float or numeric to justify supporting this type. It should be extremely simple to add, the storage format is just price in cents as a signed 64 bit integer. We should only support conversion to/from a struct like `struct Cents(i64);`

PR #748, opened 2017-02-22
WIP: implementation of postgres Money type

pfernie:
This is initial support for postgres' Money type for #680. As mentioned in the issue, the implementation is straightforward, however I had some questions:

* The meaning/precision of the `MONEY` type depends on the [`lc_monetary` ](https://www.postgresql.org/docs/9.1/static/datatype-money.html) setting. However, I do not know if it would be useful or necessary to try and expose this to clients.
* Reading the docs for `diesel::types`, `diesel::sql_types`, and `diesel::data_types`, it seems that the first 2 are not to be used for fields in end user structs; this is what `diesel::data_types` should do (although it seems these are typically re-exports of specific types?). I was unsure if it made sense to introduce `diesel::data_types::Money`, especially since I don't believe any other backend (`mysql`, `sqlite`) have a specific `MONEY` type. Basically, I am unclear on the intended structure and relation of "generic" types and backend-specific types...
* There is currently no implementation of various ops (`Add`, `Mul`, etc.); it seems this can be sensibly be implemented here, but was unsure if it should be implemented for `PgMoney` or some `types::Money`

Also, is there a rustfmt setting/style used in this project? I use default rustfmt & autofmt in vim, which ended up creating a bunch of formatting-noise edits (not included here).


sgrif:
My response to the PR description without having read the code:

> I was unsure if it made sense to introduce diesel::data_types::Money

Yes, it does. I would call it `Cents` though, not `Money`

sgrif:
> especially since I don't believe any other backend (mysql, sqlite) have a specific MONEY type. Basically, I am unclear on the intended structure and relation of "generic" types and backend-specific types...

Basically we put it in a backend specific module, and pub re-export it from the generic module. I have not decided what we'll do when there's a namespace collision there

sgrif:
> There is currently no implementation of various ops (Add, Mul, etc.); it seems this can be sensibly be implemented here, but was unsure if it should be implemented for PgMoney or some types::Money

Doesn't need to be handled by this PR.

sgrif:
> Also, is there a rustfmt setting/style used in this project? I use default rustfmt & autofmt in vim, which ended up creating a bunch of formatting-noise edits (not included here).

No, rustfmt doesn't have settings for a few things that we really want (mainly having `where` on the line above the first predicate). We take git churn really seriously and right now the settings we could make with rustfmt would have too much churn. We do want to add one though. I appreciate you making sure that you excluded the formatting-noise changes.

diesel/src/pg/types/money.rs
```diff
@@ -0,0 +1,32 @@
...
+use std::error::Error;
+use std::io::prelude::*;
+
+use byteorder::{ReadBytesExt, WriteBytesExt, NetworkEndian};
+
+/// Money is reprsented in Postgres as a 64 bit signed integer. The fractional precision of the
+/// value is determined by the [lc_monetary setting of the database](https://www.postgresql.org/docs/9.1/static/datatype-money.html).
+/// This struct is a dumb wrapper type, meant only to indicate the integer's meaning.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PgMoney(pub i64);
```
sgrif: Yeah, this is the thing that should be called `Cents`

diesel/src/pg/types/money.rs
```diff
@@ -0,0 +1,32 @@
+//! Support for Money values under PostgreSQL.
+use std::error::Error;
+use std::io::prelude::*;
+
+use byteorder::{ReadBytesExt, WriteBytesExt, NetworkEndian};
+
+/// Money is reprsented in Postgres as a 64 bit signed integer. The fractional precision of the
+/// value is determined by the [lc_monetary setting of the database](https://www.postgresql.org/docs/9.1/static/datatype-money.html).
```
killercup: Clippy wants you to put `lc_monetary` in back ticks (see travis ci)

killercup: Also, any reason you link to the 9.1 docs instead of more recent ones? Was this added in 9.1?

pfernie:
I updated per the comments, although I have not put in doc tests yet.

I may still be getting the naming wrong here; I renamed `pg::types::money::PgMoney` to `pg::types::money::PgCents`, sort of in line with that is done with e.g. `PgTimestamp`, etc. `pg::types::sql_types::Money`, however, I left named as-is to mean "some representation of Money". But, in client code I end up using `PgCents` in my structs, and it is good to know in client code that I'm dealing w/ cents... seems like I am making this too complicated. Should there only be a single struct `Cents` defined, instead of both `pg::types::sql_types::Money` and `pg::types::money::PgCents`? If so, should it be defined as `pg::types::money::Cents` and then re-exported in `pg::types::sql_types`? Sorry if this should be obvious...

RE: using the name `Cents`; I associate cents specifically with 1/100ths, so I wonder if it's the appropriate name here since it actually depends on the `lc_monetary` setting (could be 1/1000th, etc.)? I understand wanting to make explicit what the struct is encoding as opposed to `Money` which could be any representation. I'm not sure I have a better suggestion; `MoneyMinorUnits` maybe?

Once I understand/tidy up naming I can clean up the commits on this branch and put in the doc tests as well.

sgrif:
I'm fine with having it be called `PgMoney` and then aliasing it as `Cents`, since that's the most common case. I don't want `PgCents` though -- Usually that prefix is "this is a type that is the raw representation sent over the wire, not one that you should use". `Cents` is a type that we want to encourage people to use. (Literally the main reason for wanting to support this type is so when people ask about what to do with their `DECIMAL(10, 2)` columns I can stop adding "but also you should be using an integer as cents for monetary data" to the end of my answer)

pfernie:
Updated with a doctest and the suggested naming. This ended up being:
* `pg::types::money::PgMoney` exported as `pg::data_types::Cents` and `diesel::data_types::Cents`.
* `pg::sql_types::Money`, re-exported as `diesel::types::Money`

I don't feel strongly about the choice of the `Cents` name, but just curious if there were thoughts on my suggestion of using a name that isn't specifically associated w/ "1/100th" units. Again, not sure I have great alternate suggestions; `MoneyMinorUnits`, `SmallestMoneyUnits`...? `Cents` maybe is just the best as it's intuitive(?) even if it's not universally accurate...

Otherwise, let me know if I should clean up the commits on this PR to remove the various rename noise.

diesel/src/pg/mod.rs
```diff
@@ -16,5 +16,5 @@ pub mod data_types {
     #[doc(inline)]
     pub use super::types::floats::PgNumeric;
     #[doc(inline)]
-    pub use super::types::money::PgCents;
+    pub use super::types::money::PgMoney as Cents;
```
sgrif: We'll want to export this as both.

sgrif:
`Cents` is not universally correct, but it is correct for the majority of currencies. I think it's fine to have the alternative `PgMoney` spelling as well for users who are working with currencies that are not cents.

pfernie:
Makes sense; straight `PgMoney` export added in addition to the `Cents` alias.

diesel/src/pg/types/mod.rs
```diff
@@ -204,4 +205,76 @@ pub mod sql_types {
     /// # }
     /// ```
     #[derive(Debug, Clone, Copy, Default)] pub struct Jsonb;
+
+    /// The PostgreSQL [Money](https://www.postgresql.org/docs/9.1/static/datatype-money.html) type.
+    ///
+    /// ### [`ToSql`](/diesel/types/trait.ToSql.html) impls
+    ///
+    /// - [`PgMoney`][PgMoney]
```
sgrif: Can we change this to "`Cents` (also aliased as `PgMoney`)"?

diesel/src/pg/types/money.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl ToSql<types::Money, Pg> for PgMoney {
+    fn to_sql<W: Write>(&self, out: &mut W) -> Result<IsNull, Box<Error + Send + Sync>> {
+        out.write_i64::<NetworkEndian>(self.0)
+            .map(|_| IsNull::No)
+            .map_err(|e| e.into())
+    }
+}
+
+#[cfg(feature = "quickcheck")]
```
sgrif: Heh. I wouldn't have even thought to do this. Thanks. ;)

diesel/src/pg/types/money.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    impl Arbitrary for PgMoney {
+        fn arbitrary<G: Gen>(g: &mut G) -> Self {
+            PgMoney(i64::arbitrary(g))
+        }
+    }
+}
+
+#[test]
+fn cents_to_sql() {
```
sgrif: The impl is trivial enough that I don't think this is necessary, as long as we have an entry in `diesel_tests/tests/types_roundtrip.rs`

diesel_tests/tests/types.rs
```diff
@@ -422,6 +422,25 @@ fn pg_jsonb_to_sql_jsonb() {
 
 #[test]
 #[cfg(feature = "postgres")]
+fn pg_cents_from_sql() {
```
sgrif: Same here, we don't need this test.

diesel_tests/tests/types_roundtrip.rs
```diff
@@ -93,6 +93,7 @@ mod pg_types {
     test_round_trip!(timestamp_roundtrips, Timestamp, PgTimestamp);
     test_round_trip!(interval_roundtrips, Interval, PgInterval);
     test_round_trip!(numeric_roundtrips, Numeric, PgNumeric);
+    test_round_trip!(money_roundtrips, Money, PgMoney);
```
sgrif: We do need this one. :)

sgrif:

pfernie:
@sgrif just wanted to see if the latest changes addressed any outstanding comments. Not trying to be pushy on the PR, just unsure what sort of visibility updates have if I don't post a comment or mention...

killercup:
@sgrif, I think all your comments were addressed. This looks good to me now. The conflict in `CHANGELOG.md` is easy to resolve, and can be done while merging.

killercup:
Thanks for working on this, @pfernie!

Issue #719, opened 2017-02-17
Feature Request: Add network address support for Postgresql


KeenS:
Hi, diesel developers.
I'm using diesel at work and need to handle network addresses for better functionalities of our product.
You know PostgreSQL supports network address types. CF [PostgreSQL: Documentation: 9.6: Network Address Types](https://www.postgresql.org/docs/current/static/datatype-net-types.html).
Is it possible to support network address types on diesel? It doesn't seem MySQL and SQLite support it, so it will be a Pg only feature.

As it seems there aren't address range types neither in rust std nor crates.io, you may need to define enums such as:

```rust
use std::net::{Ipv4Addr,IpV6Addr}
enum Inet {
  Ipv4(Ipv4Addr, [u8; 4]),
  Ipv6(Ipvv6Addr, [u8; 16]),
}
```

MacAddress exits in crates.io CF https://docs.rs/eui48/0.1.0/eui48/struct.MacAddress.html .

KeenS:
Er.. subnetmask can be represented in `u8`, no need to have the same bits as addresses.

Eijebong:
I think everything has been implemented now :)

Feel free to reopen if we missed something

KeenS:
No problem at all. Thank you.

PR #749, opened 2017-02-23
implement network address (only cidr and inet)

KeenS:
partially implementing #719 . I looked at uuid for reference implementation.
Are there anything left to do? I'm not confident this is all.

My `rustfmt` touched some code. Must I fix them? (I hope there are a rustfmt.toml to avoid creating meaningless diffs)

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -15,6 +15,7 @@ categories = ["database"]
 byteorder = "1.0"
 chrono = { version = "0.3", optional = true }
 clippy = { optional = true, version = "=0.0.126" }
+libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.8.0, <0.9.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
 mysqlclient-sys = { version = ">=0.1.0, <0.3.0", optional = true }
 pq-sys = { version = ">=0.3.0, <0.5.0", optional = true }
@@ -23,6 +24,7 @@ serde_json = { version = ">=0.8.0, <2.0", optional = true }
 time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.6.0", optional = true, features = ["use_std"] }
+ipnetwork = { version = "0.12.2", optional = true }
 
 [dev-dependencies]
 cfg-if = "0.1.0"
@@ -33,7 +35,7 @@ tempdir = "^0.3.4"
 
 [features]
 default = ["with-deprecated"]
-extras = ["chrono", "serde_json", "uuid", "deprecated-time"]
+extras = ["chrono", "serde_json", "uuid", "deprecated-time", "network-address"]
 unstable = []
 lint = ["clippy"]
 large-tables = []
@@ -43,6 +45,7 @@ sqlite = ["libsqlite3-sys"]
 mysql = ["mysqlclient-sys", "url"]
 with-deprecated = []
 deprecated-time = ["time"]
+network-address = ["ipnetwork", "libc"]
 
 [badges]
 travis-ci = { repository = "diesel-rs/diesel" }
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -1,6 +1,8 @@
 mod array;
 pub mod date_and_time;
 pub mod floats;
+#[cfg(feature = "network-address")]
+mod network_address;
 mod integers;
 mod primitives;
 #[cfg(feature = "uuid")]
@@ -277,4 +279,30 @@ pub mod sql_types {
     /// # }
     /// ```
     #[derive(Debug, Clone, Copy, Default)] pub struct Money;
+
+    /// The [`CIDR`](https://www.postgresql.org/docs/9.6/static/datatype-net-types.html) SQL type. This type can only be used with `feature = "network-address"`
+    ///
+    /// ### [`ToSql`](/diesel/types/trait.ToSql.html) impls
+    ///
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// ### [`FromSql`](/diesel/types/trait.FromSql.html) impls
+    ///
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// [IpNetwork]: https://docs.rs/ipnetwork/0.12.2/ipnetwork/enum.IpNetwork.html
+    #[derive(Debug, Clone, Copy, Default)] pub struct Cidr;
+
+    /// The [`INET`](https://www.postgresql.org/docs/9.6/static/datatype-net-types.html) SQL type. This type can only be used with `feature = "network-address"`
+    ///
+    /// ### [`ToSql`](/diesel/types/trait.ToSql.html) impls
+    ///
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// ### [`FromSql`](/diesel/types/trait.FromSql.html) impls
+    ///
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// [IpNetwork]: https://docs.rs/ipnetwork/0.12.2/ipnetwork/enum.IpNetwork.html
+    #[derive(Debug, Clone, Copy, Default)] pub struct Inet;
 }
--- a/diesel/src/pg/types/network_address.rs
+++ b/diesel/src/pg/types/network_address.rs
@@ -1,0 +1,208 @@
+extern crate ipnetwork;
+extern crate libc;
+
+use std::io::prelude::*;
+use std::error::Error;
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+use pg::Pg;
+use types::{self, ToSql, IsNull, FromSql};
+use self::ipnetwork::{IpNetwork,Ipv4Network,Ipv6Network};
+
+#[cfg(windows)]
+const AF_INET: u8 = 2;
+// Maybe not used, but defining to follow Rust's libstd/net/sys
+#[cfg(redox)]
+const AF_INET: u8 = 1;
+#[cfg(not(any(windows, redox)))]
+const AF_INET: u8 = libc::AF_INET as u8;
+
+
+const PGSQL_AF_INET: u8 = AF_INET + 0;
+const PGSQL_AF_INET6: u8 = AF_INET + 1;
+
+primitive_impls!(Cidr -> (IpNetwork, pg: (650, 651)));
+primitive_impls!(Inet -> (IpNetwork, pg: (869, 1041)));
+
+macro_rules! err {
+    () => (return Err("invalid network address format".into()));
+    ($msg: expr) => (return Err(format!("invalid network address format. {}", $msg).into()));
+}
+
+macro_rules! assert_or_error {
+    ($cond: expr) => {
+        if ! $cond { err!() }
+    };
+
+    ($cond: expr, $msg: expr) => {
+        if ! $cond { err!($msg) }
+    };
+}
+
+macro_rules! impl_Sql {
+    ($ty: ty, $net_type: expr) => {
+        impl FromSql<$ty, Pg> for IpNetwork {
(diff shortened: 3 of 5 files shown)


sgrif:
Is there a reason you used a third party crate rather than implementing it for `IpAddr` from the standard library?

diesel/Cargo.toml
```diff
@@ -39,7 +40,7 @@ unstable = []
 lint = ["clippy"]
 large-tables = []
 huge-tables = ["large-tables"]
-postgres = ["pq-sys"]
+postgres = ["pq-sys", "ipnetwork"]
```
killercup: If you really need the external dependency, please use it as a feature flag itself instead of adding in to the postgres feature (not everyone will need to use these types).

diesel/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde_json = { version = ">=0.8.0, <0.10.0", optional = true }
 time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.5.0", optional = true, features = ["use_std"] }
+ipnetwork = { version = "0.12", optional = true }
```
killercup: What does this offer that std::net doesn't provide?

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,194 @@
...
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+use pg::Pg;
+use types::{self, ToSql, IsNull, FromSql};
+use self::ipnetwork::{IpNetwork,Ipv4Network,Ipv6Network};
+
+// const PGSQL_AF_INET: u8 = libc::AF_INET + 0;
+// const PGSQL_AF_INET6: u8 = libc::AF_INET + 1;
+const PGSQL_AF_INET: u8 = 2 + 0;
+const PGSQL_AF_INET6: u8 = 2 + 1;
```
killercup: Is `AF_INET` _actually_ always `2` on all architectures we support? Or does Postgres define these constants as literal integers as well? (Please add this as comment as well)

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,194 @@
...
+                } else if af == PGSQL_AF_INET6 {
+                    assert_or_error!(bytes.len() == 20);
+                    assert_or_error!(len == 16);
+                    let b = &bytes[4..];
+                    fn c(b1: u8, b2: u8) -> u16 {(b1 as u16) << 8 | (b2 as u16) }
+                    let addr = Ipv6Addr::new(
+                        c(b[0],  b[1]),  c(b[2],  b[3]),
+                        c(b[4],  b[5]),  c(b[6],  b[7]),
+                        c(b[8],  b[9]),  c(b[10], b[11]),
+                        c(b[12], b[13]), c(b[14], b[15]));
```
killercup: The v4 code above I can follow, but I'm not sure what happens here. What's the `c` fn for? What format is `bytes` in? Is this the serialized [inet_struct](https://github.com/postgres/postgres/blob/55c3391d1e6a201b5b891781d21fe682a8c64fe6/src/include/utils/inet.h#L23-L28)? Given that this is an char array, there's no need to care about endianess. And `std::net::IpAddr` implements `From<[u8; 16]>` as well as `From<[u8; 4]>` (for Ipv4).

KeenS:
> Is there a reason you used a third party crate rather than implementing it for IpAddr from the standard library? 

IpAddr in std only provides data for IP addresses. The ipnetwork crate provides IP addresses and netmasks. It uses std::net::IpAddr for IP address part.

diesel/Cargo.toml
KeenS: I see.

diesel/src/pg/types/network_address.rs
KeenS: > And std::net::IpAddr implements From<[u8; 16]> as well as From<[u8; 4]> (for Ipv4)

I didn't notice that impl. I'll use it.

diesel/src/pg/types/network_address.rs
KeenS: To tell the truth, I'm not confident. To err on the side of caution I want to use libc crate. Is it possible to depend on libc only for using a constant?

3541:
Any progress on this issue? I also have need of this feature. If there's anything I can do to help with the implementation, I can try to do so.

KeenS:
As you can see, I'm asking if a can use libc crate. If possible, this patch will complete soon.

diesel/src/pg/types/network_address.rs
YetAnotherMinion: So I went and looked up **libc**'s implementation. https://doc.rust-lang.org/1.7.0/libc/constant.AF_INET.html. It is always 2. Now if you can point me to a list of all the architectures diesel supports I will happily go and verify the values of AF_INET for all of the architectures

YetAnotherMinion: from postgres source code https://doxygen.postgresql.org/utils_2inet_8h.html#a8ba3e5fe500d587d3eb8699968450b18, they rely on AF_INET from <sys/socket.h>

killercup: Sorry for taking so long to answer. I think it's okay to depend on libc here (as an optional dependency). We already have libc as a transitive dependency when using the time/chrono feature.

KeenS:
Now I resolved the requested change. As libc doesn't define `AF_INET` on Windows, I defined it manually (which is also the way rust's std::net does).

KeenS:
A test which isn't related to this patch is failing. How can I fix it?

Eijebong:
Urgh, it's that deadlock again... See #828 
Just ignore it.

KeenS:
@Eijebong I see, thank you.
@killercup then, I finished resolved your comment and ready to merge.

diesel/src/pg/types/mod.rs
```diff
@@ -277,4 +279,29 @@ pub mod sql_types {
     /// # }
     /// ```
     #[derive(Debug, Clone, Copy, Default)] pub struct Money;
+    /// The `CIDR` SQL type. This type can only be used with `feature = "network-address"`
```
killercup: I had no idea what the `CIDR` type is. Maybe add a link to <https://www.postgresql.org/docs/9.6/static/datatype-net-types.html> or a similar page?

diesel/src/pg/types/mod.rs
```diff
@@ -277,4 +279,29 @@ pub mod sql_types {
...
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// ### [`FromSql`](/diesel/types/trait.FromSql.html) impls
+    ///
+    /// - [`ipnetwork::IpNetwork`][IpNetwork]
+    ///
+    /// [IpNetwork]: https://docs.rs/ipnetwork/0.12.2/ipnetwork/enum.IpNetwork.html
+    #[derive(Debug, Clone, Copy, Default)] pub struct Cidr;
+
+    /// The `INET` SQL type. This type can only be used with `feature = "network-address"`
```
killercup: (Also add a link to INET docs for consistency)

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,204 @@
...
+                    let inet = Ipv4Network::new(addr, prefix)?;
+                    Ok(IpNetwork::V4(inet))
+                } else if af == PGSQL_AF_INET6 {
+                    assert_or_error!(bytes.len() == 20);
+                    assert_or_error!(len == 16);
+                    let b = &bytes[4..];
+                    let addr = Ipv6Addr::from([b[0],  b[1],  b[2],  b[3],
+                                               b[4],  b[5],  b[6],  b[7],
+                                               b[8],  b[9],  b[10], b[11],
+                                               b[12], b[13], b[14], b[15]]);
```
killercup: Why do you need to explicitly list all the bytes? Wouldn't `Ipv6Addr::from(&bytes[4..20])` do the same thing?

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,204 @@
...
+                let prefix = bytes[1];
+                let net_type = bytes[2];
+                let len = bytes[3];
+                assert_or_error!(net_type == $net_type);
+                println!("{:?}", bytes);
+                if af == PGSQL_AF_INET {
+                    assert_or_error!(bytes.len() == 8);
+                    assert_or_error!(len == 4);
+                    let b = &bytes[4..];
+                    let addr = Ipv4Addr::new(b[0], b[1], b[2], b[3]);
```
killercup: Why do you need to explicitly list all the bytes? Wouldn't `Ipv6Addr::from(&bytes[4..8])` do the same thing?

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,204 @@
...
+            fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error + Send + Sync>> {
+                // https://github.com/postgres/postgres/blob/55c3391d1e6a201b5b891781d21fe682a8c64fe6/src/include/utils/inet.h#L23-L28
+                let bytes = not_none!(bytes);
+                assert_or_error!(4 <= bytes.len());
+                let af = bytes[0];
+                let prefix = bytes[1];
+                let net_type = bytes[2];
+                let len = bytes[3];
+                assert_or_error!(net_type == $net_type);
+                println!("{:?}", bytes);
```
killercup: I assume this was for left over from debugging?

diesel/src/pg/types/network_address.rs
```diff
@@ -0,0 +1,204 @@
...
+
+primitive_impls!(Cidr -> (IpNetwork, pg: (650, 651)));
+primitive_impls!(Inet -> (IpNetwork, pg: (869, 1041)));
+
+macro_rules! err {
+    () => (return Err("invalid network address format".into()))
+}
+
+macro_rules! assert_or_error {
+    ($cond: expr) => {
```
killercup: Maybe extend this (and `err!`) to also take an optional message. This way, you could make the user facing errors as well as the asserts below more descriptive.

diesel/src/pg/types/mod.rs
KeenS: `CIDR` is a common word in network area. Anyway I'll add a link to the page.

diesel/src/pg/types/network_address.rs
KeenS: Yes, it's a debugging code. It's my mistake.

diesel/src/pg/types/network_address.rs
KeenS: It's less cumbersome. I'll adopt it. Thank you.

KeenS: Ops, you cannot do `Ipv6Addr::from(&bytes[4..8])` because `bytes[4..8]` returns `[u8]` and `Ip4Addr::from` is implemented only for `u32` or `[u8; 4]`.

diesel/src/pg/types/network_address.rs
KeenS: The same above. `From<&[u8]>` isn't implemented for Ipv6Addr.

KeenS:
yes, I'm using rustfmt. Is there any formatting style guide (like rustfmt.toml)? Or, if isn't, I'll revert the diff by hand.

killercup:
Sorry, this totally fell through the cracks of my Github notifications. We don't currently have a fixed code style than can be enforced by rustfmt (patiently waiting for the fmt rfcs). Can you can remove the change introduced by formatting? (Alternatively, reset the tests/types file and only add your new tests.)

There's a small conflict in Cargo.toml. Might be a good chance to rebase and squash some of these commits. (That's not required though, or I can do it if you want.)

KeenS:
Here, I reverted the format and rebased to origin/master(b17bb9cb92fd034ba39c84829f4a53fbaddc7036).

killercup:
Thanks! I say: Let's merge this!

Eijebong:
I think this is missing a CHANGELOG entry

killercup:
I knew I forgot something! I'll add it in a sec.

PR #752, opened 2017-02-23
Add support for `ON CONFLICT conflict_target DO NOTHING`

sgrif:
This adds support for the next third of #196, adding the ability to
specify the conflict target. This does not yet add support for `DO
UPDATE`, that will come in a later PR. This does not support all
possible conflict targets. They're described in the pg syntax as

```
one of:
    ( { index_column_name | ( index_expression ) } [ COLLATE collation ] [ opclass ] [, ...] ) [ WHERE index_predicate ]
    ON CONSTRAINT constraint_name
```

The grammar supported by this PR is:

```
one of:
    ( index_column_name [, ...] )
    ON CONSTRAINT constraint_name
```

Supporting expression indexes is a bit more work, since column names
have to appear without the table name here. We'd either need to add a
new trait which is implemented for everything in the query builder, or
modify `QueryFragment#to_sql` to have an additional parameter of whether
to include the table name or not.

It's something I want to support eventually, but it's low priority (and
may even have to wait until 2.0). I also need to verify whether bind
parameters can appear in expression position here or not. If the answer
is no, we'll just need to provide a raw SQL API for expression indexes
no matter what.

### DIFF
--- a/diesel/src/pg/upsert/on_conflict_clause.rs
+++ b/diesel/src/pg/upsert/on_conflict_clause.rs
@@ -5,6 +5,8 @@ use query_builder::*;
 use query_builder::insert_statement::*;
 use query_source::Table;
 use result::QueryResult;
+use super::on_conflict_actions::*;
+use super::on_conflict_target::*;
 
 #[derive(Debug, Clone, Copy)]
 pub struct OnConflictDoNothing<T>(T);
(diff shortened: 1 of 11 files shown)


diesel/src/pg/upsert/on_conflict_clause.rs
```diff
@@ -5,6 +5,8 @@ use query_builder::*;
 use query_builder::insert_statement::*;
 use query_source::Table;
 use result::QueryResult;
+use super::on_conflict_actions::*;
+use super::on_conflict_target::*;
 
 #[derive(Debug, Clone, Copy)]
 pub struct OnConflictDoNothing<T>(T);
```
killercup: (I think I forgot to add `#[must_use]` to this earlier)

diesel/src/pg/upsert/on_conflict_clause.rs
```diff
@@ -15,6 +17,23 @@ impl<T> OnConflictDoNothing<T> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct OnConflict<Records, Target, Action> {
```
killercup: Probably wants to have `#[must_use]`

sgrif:
> For potential symmetry with future fns like do_update which might have params? (I think it's fine.)

Yes

diesel/src/pg/upsert/on_conflict_extension.rs
```diff
@@ -86,12 +59,86 @@ pub trait OnConflictExtension {
...
+    /// `Target` can be one of:
+    ///
+    /// - A column
+    /// - A tuple of columns
+    /// - [`on_constraint("constraint_name")`](fn.on_constraint.html)
+    ///
+    /// `Action` can be one of:
+    ///
+    /// - [`do_nothing()`](fn.do_nothing.html)
+    /// - [`do_update()`]
```
sgrif: I figured I'd just lay out the docs now

diesel/src/pg/upsert/on_conflict_clause.rs
sgrif: It really only matters for full queries, not for little pieces like this.

PR #753, opened 2017-02-23
Fix result set ordering in filter tests

pfernie:
Fixes random test failures due to result set ordering differences when testing `.filter()` calls. Adds
an `assert_sets_eq` macro to verify two sets contains the same elements without concern for ordering.

Alternatively, could be implemented with a `.order(..)` call on queries, but did not want to
obscure the fact these are testing `.filter(..)`.


diesel_tests/tests/filter.rs
```diff
@@ -1,6 +1,14 @@
 use schema::*;
 use diesel::*;
 
+macro_rules! assert_sets_eq {
+    ($set1:expr, $set2:expr) => {
+        let set1 = {$set1};
+        let set2 = {$set2};
+        assert!(set1.iter().all(|si| set2.contains(si)) && set2.iter().all(|si| set1.contains(si)));
```
killercup: Might be easier to convert them to HashSets and call `.difference().len() == 0`.

pfernie: The error messages weren't that great; I added a commit to make the error case more explicit by identifying the extra items in either set.

Using a `HashSet`  would require that the returned objects impl `Hash`, which for example `schema::User` in the tests currently does not. Doesn't seem like a big deal to simply add, but OTOH we're talking about small test sets here, so the less constrained (although perhaps less efficient) `Vec` seems reasonable? Don't feel strongly about it. That being said, the `Vec` approach also does allow for the slightly clearer error messages...

If/when this gets merged, happy to submit a later PR to perhaps move the macro up higher up in the modules and using in other tests where appropriate when I have time.

PR #756, opened 2017-02-24
(Mostly) allow tests to pass locally

tschottdorf:
Tried to run the local tests as indicated in `CONTRIBUTING.md` but ran into one
problem after another. Most of the trouble came from ad-hoc database URLs that
were used in various locations; those have been replaced and a repo-root `.env`
file should now be used correctly for all databases.

Added a docker-incantation to spin up throwaway Postgres and MySQL compatible
with the default `.env` file.

Some unintended cross-pollination caused further problems: The examples left
state behind that causes the `diesel_tests` to fail as it tried to migrate the
overlapping but incompatible schema. As a workaround, let the example tests
clean up after themselves.

There are still issues left; sometimes I can get a clear pass, but mostly
`diesel_tests` fail with a message such as (this one being from the `sqlite`
invocation):

```
error[E0432]: unresolved import `schema::users::dsl::*`
[hundreds more elided]
```

I suspect that this has something to do with a stale schema somewhere because
it goes away as soon as I touch `build.rs` (after which I suspect it re-ingests
the schema and then everything works).

It's been a bit frustrating getting set up and hopefully this PR can make it
a bit easier for future contributors. Very likely I didn't solve all these
problems satisfactorily, but I'm happy to take feedback to get there.

### DIFF
--- a/diesel_cli/tests/database_reset.rs
+++ b/diesel_cli/tests/database_reset.rs
@@ -1,3 +1,5 @@
+extern crate url;
+
 use support::{database, project};
 
 #[test]
@@ -55,7 +57,13 @@ fn reset_handles_postgres_urls_with_username_and_password() {
     db.execute("DROP ROLE IF EXISTS foo");
     db.execute("CREATE ROLE foo WITH LOGIN SUPERUSER PASSWORD 'password'");
 
-    let database_url = format!("postgres://foo:password@localhost/diesel_{}", p.name);
+    let database_url = {
+        let mut new_url = url::Url::parse(&p.database_url()).expect("invalid url");
+        new_url.set_username("foo").expect("could not set username");
+        new_url.set_password(Some("password")).expect("could not set password");
+        new_url.to_string()
+    };
+
     let result = p.command("database")
         .arg("reset")
         .env("DATABASE_URL", &database_url)
--- a/diesel_cli/tests/support/project_builder.rs
+++ b/diesel_cli/tests/support/project_builder.rs
@@ -1,3 +1,6 @@
+extern crate url;
+extern crate dotenv;
+
 use std::fs::{self, File};
 use std::path::{Path, PathBuf};
 use tempdir::TempDir;
@@ -67,21 +70,33 @@ impl Project {
             .collect()
     }
 
+    #[cfg(any(feature="postgres", feature="mysql"))]
+    fn database_url_from_env(&self, var: &str) -> url::Url {
+        use self::dotenv::dotenv;
+        use std::env;
+        dotenv().ok();
+
+        let mut db_url = url::Url::parse(&env::var_os(var).unwrap().into_string().unwrap())
+            .unwrap();
+        db_url.set_path(&format!("diesel_{}", &self.name));
+        db_url
+    }
+
+
     #[cfg(feature="postgres")]
     pub fn database_url(&self) -> String {
-        format!("postgres://localhost/diesel_{}", self.name)
+        self.database_url_from_env("PG_DATABASE_URL").to_string()
     }
 
     #[cfg(feature="mysql")]
     pub fn database_url(&self) -> String {
         use std::env;
 
+        let mut db_url = self.database_url_from_env("MYSQL_DATABASE_URL");
         if env::var_os("APPVEYOR").is_some() {
-            let password = env::var("MYSQL_PWD").unwrap();
-            format!("mysql://root:{}@localhost/diesel_{}", password, self.name)
-        } else {
-            format!("mysql://localhost/diesel_{}", self.name)
+            db_url.set_password(Some(&env::var("MYSQL_PWD").unwrap())).unwrap();
         }
+        db_url.to_string()
     }
 
     #[cfg(feature="sqlite")]
--- a/examples/postgres/test_all
+++ b/examples/postgres/test_all
@@ -1,7 +1,10 @@
 #!/bin/sh
 set -e
 
-export DATABASE_URL="postgres://localhost/diesel_examples"
+set -a
+if [ -f ../../.env ]; then . ../../.env; fi
+DATABASE_URL=${PG_EXAMPLE_DATABASE_URL}
+set +a
 
 for dir in $(find . -maxdepth 1 -mindepth 1 -type d); do
   cd $dir
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -54,9 +54,12 @@ environment:
     PGPASSWORD: Password12!
     MYSQL_PWD: Password12!
 
+  # TODO: get rid of DATABASE_URL throughout.
   matrix:
     - target: x86_64-pc-windows-msvc
       backend: postgres
+      PG_DATABASE_URL: postgres://postgres:Password12!@localhost/diesel_test
+      PG_EXAMPLE_DATABASE_URL: postgres://postgres:Password12!@localhost/diesel_example
       DATABASE_URL: postgres://postgres:Password12!@localhost/diesel_test
     # SQLite doesn't distribute a static lib
     # - target: x86_64-pc-windows-msvc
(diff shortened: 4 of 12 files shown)


diesel_cli/src/cli.rs
```diff
@@ -57,6 +57,9 @@ pub fn build_cli() -> App<'static, 'static> {
         ).subcommand(SubCommand::with_name("drop")
             .about("Drops the database specified in your DATABASE_URL.")
             .setting(AppSettings::Hidden)
+        ).subcommand(SubCommand::with_name("touch")
```
sgrif: `setup` already does this. We don't need to add a new command for it.

examples/postgres/test_all
```diff
@@ -1,7 +1,22 @@
...
+trap finish EXIT
+
+function finish {
+  # Make sure when we're done, we leave behind an empty database. Otherwise,
+  # at the time of writing, running diesel_tests stumbles as it tries to run
+  # conflicting migrations on an overlapping schema.
+  #
+  # TODO: Find a better way to avoid mixing of resources across tests.
+  ../../bin/diesel database drop
+  ../../bin/diesel database touch
```
sgrif: If there's something going on with our test suite that requires recreating the database every run, we absolutely need to fix that. This is not an acceptable solution.

diesel_cli/src/cli.rs
tschottdorf: But `setup` also runs the migrations, which I don't want - the state left behind by this by the example can't be handled by those in `diesel_tests` (which have a different migration history, but overlapping table name `posts`).

examples/postgres/test_all
tschottdorf: See above, the examples leave `diesel_test.posts` behind which clashes which the migration history use in the `diesel_tests`. Or, rather, that's what it looked like to me.

> This is not an acceptable solution.

I agree, and it's definitely going on. After all, the database is shared across all tests...

There are also two more issues I haven't even looked into yet (well, I've tried, but didn't manage):

* the `sqlite` diesel tests are sometimes built in a way that the inferred schema is empty (see the PR description), and since nothing invalidates the binary when the sqlite database changes, you need to force rebuild of the test so that it can pass
* not a schema problem, but sometimes the postgres `times_relative_to_now_encode_correctly() ` fails (i.e. the assertion fails). Again, this disappears immediately when I force a rebuild (as happens when I expand the assertion to print what's going on). This usually happens after I've had the laptop in sleep mode or when I run it the first time from a clean slate.

diesel_cli/src/cli.rs
tschottdorf: I suppose it could work out better if all tests assumed that their database may or may not exist. At least some of them require it preexisting (I'd wager most), for example dropping the `touch` in the `test_all` scripts for the examples gives

```
+ cd diesel_infer_schema
+ cargo test --features 'lint postgres'
warning: package replacement is not used: https://github.com/rust-lang/crates.io-index#diesel_infer_schema:0.11.0
warning: package replacement is not used: https://github.com/rust-lang/crates.io-index#diesel_codegen:0.11.0
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running /Users/tschottdorf/rust/diesel/target/debug/deps/diesel_infer_schema-a9de714f71691e1f

running 6 tests
test information_schema::tests::load_table_names_loads_from_public_schema_if_none_given ... FAILED
test information_schema::tests::get_table_data_loads_column_information ... FAILED
test information_schema::tests::load_table_names_loads_from_custom_schema ... FAILED
test information_schema::tests::get_primary_keys_only_includes_primary_key ... FAILED
test information_schema::tests::load_table_names_output_is_ordered ... FAILED
test information_schema::tests::skip_views ... FAILED
```

sgrif: The examples should be dropping and recreating their databases, since "create a migration and run it" is part of the example. The test suite however should not leave behind any state. I don't think that requiring a single command to be run to create the database before running the test suites is a major issue, and it means that most of the test suites do not need to assume that they are running as a user with elevated permissions.

examples/postgres/test_all
sgrif: The examples aren't meant to run against the same database as the test suite. We should probably change either `CONTRIBUTING.md` or the database used by the examples to not conflict.

tschottdorf: Unfortunately, all of that juggling environment variables makes that not-so-easy. We could introduce yet another set of env vars (`{PG,MYSQL}_EXAMPLE_DATABASE_URL`?) and use that. Or try to change the database ad-hoc, but trying to do that in bash is going to be frail.
Or, not run those examples normally. However, once you did so manually, you'd run into the same problems again, so I'd rather avoid that option. Any preference on your end?

Also, any pointers on how I can investigate the result of `infer_schema!("dotenv:SQLITE_DATABASE_URL");` to track down the other spurious failure I mentioned?

diesel_cli/src/cli.rs
tschottdorf: Yeah, the issue will hopefully disappear when the examples stop overlapping (the other thread).

examples/postgres/test_all
sgrif: Actually I'm not sure I fully understand the issue you ran into. We hard code the database for examples [here](https://github.com/diesel-rs/diesel/blob/master/examples/postgres/test_all#L4). My point was that we just need to make sure that it doesn't conflict with the example URL given in the CONTRIBUTING document, but it already doesn't. Something else is wrong if the examples were running against your test database.

tschottdorf: Removing this hard-coding was part of why I opened this PR (I thought that was clear from the description and the commit messages). The instructions in `CONTRIBUTING.md` suggest that you get to decide what you're running the tests against, and the hard-coding doesn't work for everyone (even with trivial local servers, for example I would have to add `postgres@` to the hardcoded string).

So, removing the hard-coding, I naturally let them run against what we have, which is `{MYSQL,PG}_DATABASE_URL`, but as you pointed out, that's not a good idea due to the interaction with the tests. That's why I remove that state and so it works, but I agree that it's not a good solution (for example, if the example fails half-way through).

One could argue that it would be easiest to just hardcode it all and force folks to run things at the right location. I don't think that's a great solution though, and certainly it isn't fun for CI, (for example, you don't run the example in appveyor, probably because there you require user and password) but it could work.

My preference so far (and one that should hopefully also reflect your position) is an extra set of `.env` vars to be used for running the examples.

When I cloned the project, I wasn't planning on spending so much time on running the tests, but now that I have I think there's some work to be done to lower the barrier of entry, and ideally that would mean opening `CONTRIBUTING.md`, starting the databases *anywhere I want*, creating an appropriate `.env` and then be able to run `./bin/test` (i.e. what I'm trying to achieve here).

sgrif: Ah sorry, I guess I missed that. I'm fine with adding another set of env vars, but let's have it fall back to the defaults set today if the env var is unset, since the number of environment variables required is getting quite large.

tschottdorf:
Ok, added the env var, looks like CI passes. I'm still getting transient failures locally that are difficult to debug, but I suppose I don't need to solve that in this PR. I'll look into addressing some of the new `TODO`s but let me know if the rest looks good.

tschottdorf:
@sgrif ping - I think I'm done here, please let me know if that's not the case.

diesel_cli/tests/database_reset.rs
```diff
@@ -55,7 +57,13 @@ fn reset_handles_postgres_urls_with_username_and_password() {
     db.execute("DROP ROLE IF EXISTS foo");
     db.execute("CREATE ROLE foo WITH LOGIN SUPERUSER PASSWORD 'password'");
 
-    let database_url = format!("postgres://foo:password@localhost/diesel_{}", p.name);
+    let database_url = {
+      let mut new_url = url::Url::parse(&p.database_url()).expect("invalid url");
```
sgrif: Indentation is funky here.

diesel_cli/tests/database_reset.rs
```diff
@@ -55,7 +57,13 @@ fn reset_handles_postgres_urls_with_username_and_password() {
...
+      new_url.set_username("foo").expect("could not set username");
+      new_url.set_password(Some("password")).expect("could not set password");
+      format!("{}", new_url)
```
sgrif: `new_url.to_string()`

diesel_cli/tests/support/project_builder.rs
```diff
@@ -67,21 +70,34 @@ impl Project {
             .collect()
     }
 
+    #[cfg(any(feature="postgres", feature="mysql"))]
+    fn database_url_from_env(&self, var: &str) -> url::Url {
+      use self::dotenv::dotenv;
```
sgrif: Indentation is off

examples/postgres/test_all
```diff
@@ -1,7 +1,18 @@
...
 
-export DATABASE_URL="postgres://localhost/diesel_examples"
+set -a
+if [ -f ../../.env ]; then . ../../.env; fi
+DATABASE_URL=${PG_EXAMPLE_DATABASE_URL}
+set +a
+
+finish() {
+  # Drop the database again so that the next test run can practice creating it
+  # again. Not required.
```
sgrif: If it's not required why do we need it at all?

diesel_cli/tests/database_reset.rs
tschottdorf: Done.

diesel_cli/tests/support/project_builder.rs
tschottdorf: Done.

examples/postgres/test_all
tschottdorf: We don't. I removed it.

diesel_cli/tests/database_reset.rs
tschottdorf: Done.

tschottdorf:
I rebased again since it sat for a while. Still ready as far as I'm concerned.

diesel_cli/tests/database_reset.rs
tamird: (in the wrong commit)

diesel_cli/tests/database_reset.rs
tamird: (in the wrong commit)

diesel_cli/tests/support/project_builder.rs
tamird: (in the wrong commit)

sgrif:
@killercup Do you mind deleting either `.env.sample` or `.env.example` and then merging this?

killercup:
Merged and added https://github.com/diesel-rs/diesel/commit/71cb54ff8b54986428c280e9f33fc1ddc02e5421

PR #757, opened 2017-02-24
Give a more specific error message for unsigned types

### LINKED ISSUE #754: infer_schema! generates error: no rules expected the token `)`
I'm working on integrating Diesel into one of my projects at work instead of using the MySQL drivers directly. As soon as I add the line `infer_schema!("dotenv:DATABASE_URL");` anywhere in my project, I get the error mentioned in the title: `error: no rules expected the token ')'`. I can't for the life of me figure out what it means or what causes it. 

My .env file exists and is fine, as the diesel command line tool has no objections with it. I'm importing diesel and diesel_codegen in the correct order in my lib.rs, as I've seen that sometimes causes problems. I have also tried varying import statements, though none are required in any of the examples.

I have tried building the examples, and they work fine. That leads me to believe it's a problem in my project, however, I have no idea where to start looking, as this error is pretty cryptic.

Rust version: rustc 1.15.1 (021bd294c 2017-02-08)
Diesel version: 0.11.4
Diesel features: mysql

sgrif:
We don't yet support these types in MySQL, but the error a user will
recieve is not terribly helpful, since the type will contain a space so
the `table!` macro invocation will just have bad tokens.

This gives a more specific error message for unsigned types, as well as
ensuring that we get a reasonable error for other types would result in
a bad macro invocation. (I'm not actually aware of any other types which
have a space, but I'd rather not dig through the entire manual right
now)

In theory we could exclude unsigned floats from this, since unsigned
floats aren't actually a thing and that "type" in MySQL is basically
just a check constraint, but I'd rather hold off and actually represent
in Diesel that the type is different when we add support for unsigned
integer types.

Fixes #754.

### DIFF
--- a/diesel_infer_schema/src/codegen.rs
+++ b/diesel_infer_schema/src/codegen.rs
@@ -21,7 +21,7 @@ pub fn expand_infer_table_from_schema(database_url: &str, table: &TableData)
     let mut tokens = Vec::with_capacity(data.len());
 
     for a in data {
-        tokens.push(column_def_tokens(&a, &connection)?);
+        tokens.push(column_def_tokens(table, &a, &connection)?);
     }
     let default_schema = default_schema(&connection);
     if table.schema != default_schema {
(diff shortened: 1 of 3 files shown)


diesel_infer_schema/src/codegen.rs
```diff
@@ -54,11 +54,20 @@ pub fn handle_schema<I>(tables: I, schema_name: Option<&str>) -> quote::Tokens
...
+    table: &TableData,
     column: &ColumnInformation,
     connection: &InferConnection,
 ) -> Result<quote::Tokens, Box<Error>> {
     let column_name = syn::Ident::new(&*column.column_name);
-    let column_type = try!(determine_column_type(column, connection));
+    let column_type = match determine_column_type(column, connection) {
+        Ok(t) => t,
+        Err(e) => return Err(format!(
+            "Error determining type of {}.{}: {}",
```
killercup: Add some back ticks?

PR #763, opened 2017-02-25
Unify `FilteredQuerySource` and `SelectStatement`

### LINKED ISSUE #760: Unify `FilteredQuerySource` and `SelectStatement`
The reason `FilteredQuerySource` exists is basically to say "select was never called explicitly". This is because we want `update(users.filter(id.eq(1)))` to work, but not `update(users.select(id))`, since that's nonsense.

Right now, `SelectStatement` first comes into play at the `AsQuery` level. A `table` will implement `AsQuery` by doing `SelectStatement::simple(all_columns, self)`. Rather than passing `all_columns` there, I'd like to introduce some sort of "default select clause" marker, and then only implement `UpdateTarget` for `SelectStatement` when the select clause is that marker.

sgrif:
`FilteredQuerySource` only existed to make sure you couldn't pass a
query with a select clause or any other clause besides `WHERE` as a
target for `update` or `delete`. Instead of introducing a parallel
query type, we can instead just add the notion of a "default" select
clause.

I intentionally wanted this to be a voldemort type, so I needed to
rearrange the parameters on `SelectStatement` which caused a bit more
churn than I would have liked.

Close #760.


diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -196,3 +200,58 @@ impl<'a, S, F, D, W, O, L, Of, G, DB> InternalBoxedDsl<'a, DB>
         )
     }
 }
+
+impl<'a, F, D, W, O, L, Of, G, DB> InternalBoxedDsl<'a, DB>
```
sgrif: So I'm assuming I wrote this before you joined because we've not discussed this before. I'm curious on your thoughts though. I do write the types out explicitly [in the struct definition](https://github.com/diesel-rs/diesel/blob/6dec748273b29e95806ccc6a429c93dc4b2bf36d/diesel/src/query_builder/select_statement/mod.rs#L21-L30) because I think it's important to give them meaningful names. All of my usual principals say to use the full names, here, but this is what the code looks like.

```rust

impl<
    'a,
    Select,
    From,
    Distinct,
    Where,
    Order,
    Limit,
    Offset,
    Groupby,
    DB,
> InternalBoxedDsl<'a, DB>
    for SelectStatement<
        Select,
        From,
        Distinct,
        Where,
        Order,
        Limit,
        Offset,
        GroupBy,
    > where
        DB: Backend,
        Select: QueryFragment<DB> + SelectableExpression<From> + 'a,
        Distinct: QueryFragment<DB> + 'a,
        Where: Into<Option<Box<QueryFragment<DB> + 'a>>>,
        Order: QueryFragment<DB> + 'a,
        Limit: QueryFragment<DB> + 'a,
        Offset: QueryFragment<DB> + 'a,
{
    type Output = BoxedSelectStatement<'a, Select::SqlTypeForSelect, From, DB>;

    fn internal_into_boxed(self) -> Self::Output {
        BoxedSelectStatement::new(
            Box::new(self.select),
            self.from,
            Box::new(self.distinct),
            self.where_clause.into(),
            Box::new(self.order),
            Box::new(self.limit),
            Box::new(self.offset),
        )
    }
}
```

diesel_cli/Cargo.toml
```diff
@@ -27,7 +27,7 @@ regex = "0.1.48"
 
 [features]
 default = ["postgres", "sqlite", "mysql"]
-lint = ["clippy"]
+lint = ["clippy", "diesel/lint", "diesel_infer_schema/lint"]
```
sgrif: Whoops, I didn't mean to add this here. I was trying to figure out why `bin/test` twice no longer runs the second time with 0 recompilations and I think it's because crates are recompiled with/without clippy, but the doctest dependency on `diesel_codegen` in `diesel` breaks everything (as usual. What else is new)

diesel/src/query_builder/select_statement/dsl_impls.rs
sgrif: Also this struct *really* makes me want associated types for structs.

sgrif:
Yeah, the diff noise got a bit amplified since I decided to switch the parameters *everywhere*, but it would have been misleading to leave them in the old order in every impl

diesel/src/query_builder/select_statement/dsl_impls.rs
sgrif: Also @nrc if you want a pathological case for rustfmt discussions, see the code 2 comments up.

sgrif:
This was merged I just forgot to push before deleting the branch

diesel/src/query_builder/select_statement/dsl_impls.rs
killercup: Yeah, I think that code has been there the first time I looked at diesel. I'd absolutely continue to use the abbreviations. `F, S, W` by themselves are not that intuitive, but if you have ever seen a SQL query, they totally make sense, and it's not worth the code bloat.

_Maybe_, once I got to `Of`, I'd've switched everything to two-letter names, though (`Se, Fr, Di, Wh, Or, Li, Of, Gr`).

sgrif: `OfCourseYouWouldHave`

PR #765, opened 2017-02-26
Add support for the SQL concat operator `||`

sgrif:
The integration test was actually the main reason I wanted this. I was
originally going to do `sql_function!(concat)`, but on PG that function
ignores null arguments rather than returning null, and SQLite doesn't
even have it. The pipe operator is ANSI standard SQL, but it's behind a
configuration option for MySQL. I set up a bit of groundwork for more
"always set configuration settings", since I think we will probably want
to at minimum ensure client encoding is UTF-8 and that the timezone is
UTC for all backends eventually.

This operator is notably not supported by MS SQL Server, but I don't
plan on supporting that backend in tree, and a single hypothetical third
party backend isn't enough to warrant adding a `SupportsPipeOperator`
marker.

### DIFF
--- a/diesel/src/expression/expression_methods/text_expression_methods.rs
+++ b/diesel/src/expression/expression_methods/text_expression_methods.rs
@@ -1,8 +1,39 @@
 use expression::{Expression, AsExpression};
-use expression::predicates::{Like, NotLike};
+use expression::predicates::{Concat, Like, NotLike};
 use types::Text;
 
 pub trait TextExpressionMethods: Expression<SqlType=Text> + Sized {
+    /// Concatenates two strings using the `||` operator.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("src/doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     users {
+    /// #         id -> Integer,
+    /// #         name -> VarChar,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// #     let connection = establish_connection();
+    /// #
+    /// let names = users.select(name.concat(" the Greatest")).load(&connection);
+    /// let expected_names = vec![
+    ///     "Sean the Greatest".to_string(),
+    ///     "Tess the Greatest".to_string(),
+    /// ];
+    /// assert_eq!(Ok(expected_names), names);
+    /// # }
+    /// ```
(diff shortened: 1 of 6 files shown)


diesel/src/expression/expression_methods/text_expression_methods.rs
```diff
@@ -1,8 +1,39 @@
...
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// #     let connection = establish_connection();
+    /// #
+    /// let names = users.select(name.concat(" the greatest")).load(&connection);
+    /// let expected_names = vec![
+    ///     "Sean the greatest".to_string(),
+    ///     "Tess the greatest".to_string(),
```
killercup: The most important stuff first: In this case "greatest" should be capitalized as it's part of the title "{Sean,Tess} the Greatest" (same as in "Alexander the Great") 😉

diesel/src/expression/expression_methods/text_expression_methods.rs
```diff
@@ -1,8 +1,39 @@
 use expression::{Expression, AsExpression};
-use expression::predicates::{Like, NotLike};
+use expression::predicates::{Concat, Like, NotLike};
 use types::Text;
 
 pub trait TextExpressionMethods: Expression<SqlType=Text> + Sized {
+    /// Concatenates two strings using the `||` operator.
+    ///
+    /// # Example
```
killercup: Convention in std is to always write "Examples" (plural). (Not important and we should probably define some doc comment rules before changing stuff like this. Also, I'll write a Rust RFC on that… soonish)

diesel/src/mysql/connection/mod.rs
```diff
@@ -114,4 +118,9 @@ impl MysqlConnection {
         try!(stmt.bind(metadata.into_iter().zip(binds)));
         Ok(stmt)
     }
+
+    fn set_config_options(&self) -> QueryResult<()> {
+        try!(self.execute("SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT'))"));
```
killercup: Obligatory docs link: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_pipes_as_concat

sgrif:
We could probably make `+` work, but I'm not sure how I feel about having it not map to `+` in SQL.

sgrif:
We could probably make `+` work, but I'm not sure how I feel about having it not map to `+` in SQL.

PR #778, opened 2017-03-04
add Add/Sub impls for PgMoney

pfernie:
Implements `ops::{Add, AddAssign, Sub, SubAssign}` for `PgMoney`. I have explicitly opted not to implement `Mul` and `Div`, as I think end users should deal with rounding/truncation issues.

### DIFF
--- a/diesel/src/pg/types/money.rs
+++ b/diesel/src/pg/types/money.rs
@@ -1,6 +1,6 @@
 //! Support for Money values under PostgreSQL.
 use std::error::Error;
-use std::ops::{Add, Sub};
+use std::ops::{Add, AddAssign, Sub, SubAssign};
 use std::io::prelude::*;
 
 use byteorder::{ReadBytesExt, WriteBytesExt, NetworkEndian};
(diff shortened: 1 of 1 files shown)


sgrif:
Agreed that checked is the only reasonable option when dealing with currency.

sgrif:
Agreed that checked is the only reasonable option for dealing with currency. However, I think we should panic rather than return an option. If you take the combined wealth of the world and convert it to various currencies, the only one which doesn't fit in `i64::MAX` is the Zimbabwe Dollar which hasn't existed since 2009. (I'm not terribly concerned about the Zimbabwe Dollar even historically, since the value of 1USD to 1ZWD in 2009 would overflow, it's unlikely to be stored in this data type).

sgrif:
Actually I lied, since this is cents, it would be `1.458418e+19` which would overflow. Still, I think it's fine to say that you need to operate on the inner value manually if you want to work with amounts which approach the combined wealth of the earth in extremely weak currencies.

pfernie:
Very well; the ergonomics w/o the Option are definitely nicer. I implemented as such; I don't know convention for libraries and panics, I used `expect()` to provide some context as well as a brief "# Panics" doc comment.

diesel/src/pg/types/money.rs
```diff
@@ -39,6 +40,40 @@ impl ToSql<types::Money, Pg> for PgMoney {
...
+    fn add(self, rhs: PgMoney) -> Self::Output {
+        self.0.checked_add(rhs.0).map(PgMoney).expect("overflow adding money amounts")
+    }
+}
+
+/// # Panics
+/// will `panic!` on overflow
+impl AddAssign for PgMoney {
+    fn add_assign(&mut self, rhs: PgMoney) {
+        self.0 += rhs.0
```
killercup: Wait, does this _actually_ panic? It calls `i64::add_assign`, right? I don't think this is a checked add in release mode!

diesel/src/pg/types/money.rs
```diff
@@ -52,3 +87,33 @@ mod quickcheck_impls {
...
+
+#[test]
+fn add_money() {
+    let c1 = PgMoney(123);
+    let c2 = PgMoney(456);
+    assert_eq!(PgMoney(579), c1 + c2);
+}
+
+#[test]
+#[should_panic]
```
sgrif: Should we specify the message here?

diesel/src/pg/types/money.rs
```diff
@@ -39,6 +40,40 @@ impl ToSql<types::Money, Pg> for PgMoney {
     }
 }
 
+/// # Panics
+/// will `panic!` on overflow
```
sgrif: I think we should specify "including release mode", since that's where this differs from the default.

pfernie:
_sigh_, you're right, `{add,sub}_assign` were not doing it checked; that's what I get for skipping "trivial" tests. Corrected now + proper tests; also improved the documentation. Wasn't aware of the `#[should_panic(expected = ".."]` parameter, added that as well.

sgrif:
Fine with me if you want to merge when Travis is green, @killercup

killercup:
Thanks again, @pfernie!

PR #779, opened 2017-03-05
Allow types from other crates with `table!`

sgrif:
This doesn't affect `infer_schema!` yet (which is where this is most
important), but it lays the groundwork required. Additional modules to
import types from can be specified by writing `using types from
list::of::paths`. Due to how `use` works here, these will always need to
be relative to the crate root, and `self` or `super` will be
inconsistent.

This shouldn't matter, since the most common case is going to be to
import types from a third party crate, which means there will usually be
`extern crate whatever` at the root.

For the `table!` macro to finish, we now need to know all of the
following:

- table name
- schema name
- table body
- primary key
- type modules

Only two of those things are mandatory, so we basically have to have
`3!` branches in the `table!` macro to handle every combination of
information that could be given. I'm not sure if there's a way that we
can just have one branch per optional piece, but I don't believe we can.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Diesel CLI: Added `db` as an alias for `database`, so you can now write `diesel db setup` (which is almost 40% faster!).
 
+* The `table!` macro now allows you to use types from crates outside of Diesel.
+  You can specify where types should be imported from by doing: `table! { use
+  some_modules::*; foo { columns... }`. Not specifying any any modules is
+  equivalent to `use diesel::types::*;`.
+
 ### Fixed
 
 * `diesel_codegen` will provide a more useful error message when it encounters
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -177,6 +177,32 @@ macro_rules! __diesel_column {
 /// # }
 /// ```
 ///
+/// If you are using types that aren't from Diesel's core types, you can specify
+/// which types to import. Note that the path given has to be an absolute path
+/// relative to the crate root. You cannot use `self` or `super`.
+///
+/// ```
+/// #[macro_use] extern crate diesel;
+/// # /*
+/// extern crate diesel_full_text_search;
+/// # */
+/// # mod diesel_full_text_search {
+/// #     pub struct TsVector;
+/// # }
+///
+/// table! {
+///     use diesel::types::*;
+///     use diesel_full_text_search::*;
+///
+///     posts {
+///         id -> Integer,
+///         title -> Text,
+///         keywords -> TsVector,
+///     }
+/// }
+/// # fn main() {}
+/// ```
+///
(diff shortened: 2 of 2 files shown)


diesel/src/macros/mod.rs
```diff
@@ -223,46 +240,80 @@ macro_rules! __diesel_column {
...
+            $schema_name . $name ($pk) $($body)*
+        }
+    };
+
+    (
+        $schema_name:ident . $name:ident ($pk:ident, $($composite_pk:ident),+)
+        $($body:tt)*
+    ) => {
+        table_body! {
+            $schema_name . $name ($pk, $($composite_pk,)+) $($body)*
```
killercup: Macro changes lgtm

diesel/src/macros/mod.rs
```diff
@@ -177,6 +177,23 @@ macro_rules! __diesel_column {
 /// # }
 /// ```
 ///
+/// If you are using types that aren't from Diesel's core types, you can specify
+/// which types to import. Note that the path given has to be an absolute path
+/// relative to the crate root. You cannot use `self` or `super`.
+///
+/// ```ignore
+/// #[macro_use] extern crate diesel;
+/// extern crate diesel_full_text_search;
```
sgrif: Probably, but I want to document a real use case here. The doctest wouldn't add anything useful over the existing tests (other than ensuring the example works of course, but meh...)

diesel/src/macros/mod.rs
```diff
@@ -177,6 +177,23 @@ macro_rules! __diesel_column {
...
+///
+/// table! {
+///     posts using types from diesel::types, diesel_full_text_search::types {
```
sgrif: I'm open to suggestions. When it was just table name and optionally the primary key, the macro rules themselves were pretty self documenting. Agreed it's obscure now. Do you think there's another syntax which is more obvious? Or are you just saying we need to document the full syntax better?

killercup: I think the `use` syntax (https://github.com/diesel-rs/diesel/pull/779#pullrequestreview-25265816) is nicer. But in any case, it'd be great to document the full syntax; at least with an example that uses everything.

diesel/src/macros/mod.rs
killercup: True, that's why I'd mock the "real" diesel_full_text_search

sgrif:
@killercup What are you imagining your syntax would look like for `infer_schema!`?

sgrif:
Updated, needs a re-review.

CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Diesel CLI: Added `db` as an alias for `database`, so you can now write `diesel db setup` (which is almost 40% faster!).
 
+* The `table!` macro now allows you to use types from crates outside of Diesel.
+  You can specify where types should be imported from by doing: `table! { use
+  some_modules::*; foo { columns... }`. Not specifying any any modules is
```
PeterW-LWL: any any?

Issue #776, opened 2017-03-04
Split SQLite example into same steps as Postgres/MySQL


killercup:
In #751, we added a SQLite version of the code from the Getting Started guide. But as described in <https://github.com/diesel-rs/diesel/pull/751#issuecomment-282142148>, it's only the final code you end up with, while the other backends have examples with two in-between steps. Comparing the code, it should be easy to extract steps one and two from the current code.

PR #780, opened 2017-03-06
Split sqlite getting started example

ericho:
Split the sqlite example into three parts as the same way that MySQL and PostgreSQL examples does.

Address #776


ericho:
:open_mouth: I forgot to add that file. 
Now it's there.

sgrif:
Test failures seem related.

ericho:
I make a mistake in the last commit, now the build passed 😄

Phrohdoh:
LGTM (would suggest a squash, personally)

killercup:
Thanks again!

PR #782, opened 2017-03-07
Unsigned types support for mysql.

Eijebong:
It needs more tests, some documentation, support for floats (yay, mysql, unsigned floats \o/) but I just pushed to get some feedback.

I'm not convinced by the codegen changes, the only alternative I see is `determine_type_name`  returning `(String, bool)` to indicate if the type is unsigned or not, but then the name isn't right anymore + you've got a bool returned that doesn't really mean anything...

### DIFF
--- a/diesel/src/mysql/types/mod.rs
+++ b/diesel/src/mysql/types/mod.rs
@@ -1,10 +1,12 @@
 #[cfg(feature = "chrono")]
 mod date_and_time;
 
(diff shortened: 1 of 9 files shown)


Eijebong:
Is there anything else to do here ?

sgrif:
Haven't had time for code review lately, we'll get to it.

PR #783, opened 2017-03-07
First draft of an update guide

sgrif:
This isn't quite finished yet, but it got to a place where I want to
take a break from it, and there's enough there to give feedback on the
existing content. I figure for new guides we can start the discussion
here and then move to the website once we're good with the content. It's
easier to review without markup.

[Rendered](https://github.com/diesel-rs/diesel/blob/sg-update-guide-draft/guide_drafts/all-about-updates.md)

### DIFF
--- a/guide_drafts/all-about-updates.md
+++ b/guide_drafts/all-about-updates.md
@@ -1,0 +1,168 @@
+All About Updates
+=================
+
+Most applications fall into a category called "CRUD" apps. CRUD stands for Create Read Update Delete. Diesel provides support for all four pieces, but in this guide we're going to look at all the different ways to go about updating records with Diesel.
+
+An update statement is constructed by calling `diesel::update(target).set(changes)`. The resulting statement is then run be calling either `execute`, `get_result`, or `get_results`.
+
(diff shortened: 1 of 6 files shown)


guide_drafts/all-about-updates.md
```diff
@@ -0,0 +1,168 @@
...
+That would generate the following SQL:
+
+```sql
+UPDATE `posts` SET `body` = ?
+```
+
+If you wanted to assign `NULL` instead, you can either specify `#[changeset_options(treat_none_as_null="true")]` on the struct, or you can have the field be of type `Option<Option<T>>`. Diesel doesn't currently provide a way to explicitly assign a field to its default value, though it may be provided in the future.
+
+TODO: Cover `execute` vs `get_result` vs `get_results` vs `save_changes`
+TODO: Briefly mention upsert exists, point to docs
```
killercup: Would also add a more complex example that uses all features together.

guide_drafts/all-about-updates.md
```diff
@@ -0,0 +1,168 @@
+All About Updates
+=================
+
+Most applications fall into a category called "CRUD" apps. CRUD stands for Create Read Update Delete. Diesel provides support for all four pieces, but in this guide we're going to look at all the different ways to go about updating records with Diesel.
```
sgrif: You're right. I should mention that Diesel provides diesel as well. And also the things that Diesel provides.

killercup: Oil crisis: Resolved!

PR #788, opened 2017-03-09
Allow `Cow` to be used as an expression

sgrif:
I tried to use `Cow<'a, str>` in a stuct which had
`#[derive(Insertable)]` expecting it to work and it didn't.


diesel/src/types/impls/primitives.rs
```diff
@@ -109,3 +109,27 @@ impl <'a, T: ?Sized, ST, DB> ::types::FromSqlRow<ST, DB> for Cow<'a, T> where
         FromSql::<ST, DB>::from_sql(row.take())
     }
 }
+
+use expression::bound::Bound;
+use expression::{AsExpression, Expression};
+impl <'a, T: ?Sized, ST> ::expression::AsExpression<ST> for Cow<'a, T> where
```
killercup: empty line above? Also, `AsExpression` is in scope, no need to write absolute path

PR #791, opened 2017-03-09
Add some internal documentation about prepared statement caching

sgrif:
The way we go about this is fairly complex, and the code is probably a
bit opaque if you don't know the overall structure (though I do think
it's probably at a point where it represents my mental model pretty
well)

This documentation is not public facing (so we don't need the one
sentence module description for pretty docs, @killercup :P), but I do
think it's worth having a primer on how it all works that lives
*somewhere*.

We should maybe consider having `feature = "plugin_docs"` and `feature =
"contributor_docs"` which causes more docs to get rendered.

### DIFF
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -1,3 +1,95 @@
+//! A primer on prepared statement caching in Diesel
+//! ------------------------------------------------
+//!
+//! Diesel uses prepared statements for virtually all queries. This is most
+//! visible in our lack of any sort of "quoting" API. Values must always be
+//! transmitted as bind parameters, we do not support direct interpolation. The
+//! only method in the public API that doesn't require the use of prepared
+//! statements is `Connection#batch_execute`.
+//!
+//! In order to avoid the cost of re-parsing and planning subsequent queries,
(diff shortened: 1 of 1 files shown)


diesel/src/connection/statement_cache.rs
```diff
@@ -1,3 +1,95 @@
+//! A primer on prepared statement caching in Diesel
+//! ------------------------------------------------
```
killercup: This is an h2. For h1, go with `=`.

sgrif: If this were every actually rendered, it'd be the third h1 on the page and my 8 years out of date SEO skills say that there should only be one h1 per page

sgrif:
Yeah, I'm thinking of maybe generating with different commands and hosting them at multiple subdomains so that would do the "contributor docs" one. We'd still want a feature flag for "plugin docs" though since that'd mostly be removing `#[doc(hidden)]`s

diesel/src/connection/statement_cache.rs
killercup: Well, with my 4 year out of date HTML5 knowledge I'd expect rustdoc to put that in an `<article>` :D

PR #792, opened 2017-03-11
Update clippy

sgrif:
There's a few new lints which we had to account for. One is checking for
`assert!(x == y)`. In the case of integer deserialization, I changed the
condition because I actually don't want the error message that
`assert_eq!` would give.

The other new lint is for functions which are pass by value, but only
call methods that take a reference. This was pretty much only false
positives for us. In the case of `diesel_codegen`, that lint triggered
for every `#[proc_macro_derive]` function, where we don't even control
the signature, so I've disabled it crate-wide.

### DIFF
--- a/diesel/src/types/impls/integers.rs
+++ b/diesel/src/types/impls/integers.rs
@@ -10,7 +10,7 @@ impl<DB: Backend<RawValue=[u8]>> FromSql<types::SmallInt, DB> for i16 {
         let mut bytes = not_none!(bytes);
         debug_assert!(bytes.len() <= 2, "Received more than 2 bytes decoding i16. \
                       Was an Integer expression accidentally identified as SmallInt?");
-        debug_assert!(bytes.len() == 2, "Received fewer than 2 bytes decoding i16. \
+        debug_assert!(bytes.len() >= 2, "Received fewer than 2 bytes decoding i16. \
                       Was an expression of a different type accidentally identified \
                       as SmallInt?");
         bytes.read_i16::<DB::ByteOrder>().map_err(|e| Box::new(e) as Box<Error+Send+Sync>)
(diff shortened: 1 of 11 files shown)


diesel/src/types/impls/integers.rs
```diff
@@ -10,7 +10,7 @@ impl<DB: Backend<RawValue=[u8]>> FromSql<types::SmallInt, DB> for i16 {
         let mut bytes = not_none!(bytes);
         debug_assert!(bytes.len() <= 2, "Received more than 2 bytes decoding i16. \
                       Was an Integer expression accidentally identified as SmallInt?");
-        debug_assert!(bytes.len() == 2, "Received fewer than 2 bytes decoding i16. \
+        debug_assert!(bytes.len() >= 2, "Received fewer than 2 bytes decoding i16. \
```
killercup: Oh, were these genuine bugs that clippy found? (Not really bugs in production user-facing code but still)

diesel_codegen/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![deny(warnings, missing_copy_implementations)]
 
 // Clippy lints
-#![cfg_attr(feature = "clippy", allow(unstable_features))]
```
killercup: rustc says:

> enabling unstable features (deprecated. do not use)

So we might as well remove it everywhere.

diesel/src/types/impls/integers.rs
sgrif: No, because the line above is an assertion on `<=`. `x <= y && x >= y`  implies `x == y`

killercup: Oh, I'm stupid, it's an assert.

sgrif: It's technically more correct, but my reasoning was actually "this is less typing than `#[allow]`"

PR #801, opened 2017-03-15
Allow `SqlLiteral` to be used as an on conflict target

sgrif:
We won't have first class support for expression indexes or indexes with
where clauses in 0.12. We will add support eventually, but it's fairly
low priority. Generally speaking, you can just do two queries using
`.on_conflict_do_nothing` and then an update if no rows were returned,
and have the same semantics.

However, in the rare case that you really really do need to specifically
only handle conflicts for one unique index, and error on others, there's
no workaround. So for those cases, we can allow arbitrary SQL until we
have proper support for them (if we allow expression indexes, we'll end
up allowing arbitrary SQL anyway since arbitrary SQL is an expression).


sgrif:
Lol I figure we'll add that to the docs once #650 is resolved. I actually would prefer not to explicitly document this for the time being, since it's such a niche case and I try to avoid encouraging the `sql` function that exists today in our documentation (since it's a pretty shitty API)

PR #803, opened 2017-03-15
Respect custom primary keys in `#[derive(AsChangeset)]`

### LINKED ISSUE #802: #[derive(AsChangeset)] ignores custom primary keys

sgrif:
AsChangeset is supposed to skip assigning the primary key, but right now
we're assuming it's a single column called id. This corrects the
behavior when the annotation is present.

Fixes #802.


sgrif:
It helps if I mark the tests as `#[test]`, too. >_>

PR #807, opened 2017-03-16
Add DateTime<Local> support for FromSql traits of postgresql

genya0407:
Added `DateTime<Local>` support for `FromSql` traits of postgresql.
This PR enables the code like below:

```rust
#[derive(Queryable)]
pub struct Hoge {
    pub id: i32,
    pub start_at: DateTime<Local>,
    pub end_at: DateTime<Local>,
}
```

```rust
fn main() {
    use self::schema::hoges::dsl::*;

    let results = hoges.limit(5)
        .load::<Hoge>(&connection)
        .expect("Error loading hoges");
}
```

### DIFF
--- a/diesel/src/pg/types/date_and_time/chrono.rs
+++ b/diesel/src/pg/types/date_and_time/chrono.rs
@@ -21,6 +21,7 @@ expression_impls! {
 queryable_impls! {
     Timestamptz -> NaiveDateTime,
     Timestamptz -> DateTime<UTC>,
+    Timestamptz -> DateTime<Local>,
 }
 
 // Postgres timestamps start from January 1st 2000.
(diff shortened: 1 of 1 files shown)


sgrif:
This would allow for an incorrect representation of the data being sent, which is why we do not implement this for `DateTime<Local>` or `DateTime<FixedOffset>`. The data is sent as UTC (at least we try to ensure it is). If you've overridden the session time zone setting to be something other than UTC, you can take it as `NaiveDateTime` and convert it manually, but it's up to you to ensure that the time zone has actually been set to that of the client.

sgrif:
Hm. Perhaps I misread the PR the first time. It looks like you're doing a proper conversion. Still, I would prefer to force that conversion to be explicit in the users of Diesel, since allowing this to happen automatically makes it unclear whether we're interpreting the date being sent back as `Local`, or if a conversion occurred.

genya0407:
@sgrif I understarnd your opinion, and I have no objection to it.
However, then, I think the `DateTime<Local>` support for `Insertable` is lack of consistency.

If Diesel's position is 'conversion should be explicit in the users of Diesel', `DateTime<Local>` support for `Insertable` also should be duplicated.

PR #809, opened 2017-03-17
Unify `InnerJoinSource` and `LeftOuterJoinSource` to a single struct

sgrif:
Eventually I'd like to separate out an additional `JoinOn` struct which
represents a join and the `ON` portion, which will let me factor this to
make a little more sense than it does right now. `Join` will actually be
the thing that is visible in the public API, and eventually I want to
have the `Kind` parameter default to `Cross`, and use it to represent
two arbitrary tables for things like determining whether a subselect is
valid

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -37,56 +37,47 @@ macro_rules! __diesel_column {
         impl AppearsOnTable<$($table)::*> for $column_name {
         }
 
-        impl<Right> SelectableExpression<
-            $crate::query_source::joins::InnerJoinSource<$($table)::*, Right>,
+        impl<Right, Kind> SelectableExpression<
+            Join<$($table)::*, Right, Kind>,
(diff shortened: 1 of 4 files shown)


diesel/src/macros/mod.rs
```diff
@@ -446,6 +437,7 @@ macro_rules! table_body {
                 use $crate::{Table, Expression, SelectableExpression, AppearsOnTable, QuerySource};
                 use $crate::backend::Backend;
                 use $crate::query_builder::{QueryBuilder, BuildQueryResult, QueryFragment};
+                use $crate::query_source::joins::{Join, Inner, LeftOuter};
```
killercup: (At first I was wondering why you didn't continue to use the full paths here, but then I saw this macro was only ever called here where they are imported anyway.)

sgrif:
Will look into the CI failure if it continues.

PR #816, opened 2017-03-21
Add function to check for pending migrations

### LINKED ISSUE #811: Add a function to determine if there are pending migrations
Probably something like `migrations::any_pending_migrations(conn: MigrationConnection) -> bool`. Should just return `true` or `false` based on if there are any migrations that haven't been run.

Ref https://github.com/rust-lang/crates.io/issues/638

DorianListens:
This change introduces the function any_pending_migrations, which
returns a `Result` wrapping a `bool`, which will be true if there are
any pending migrations, and false if there are none. `Err` states
indicate problems with the migration setup.

Currently, it introduces some moderate code duplication in
migrations/mod.rs, as it shares logic with the code to actually run the
migrations, but the APIs don't match nicely, and I don't know enough
about the system to make changes to exisiting public API.

Fixes #811 

This code compiles, but currently clippy-lints [doesn't](https://github.com/Manishearth/rust-clippy/issues/1633), so I haven't been able to run the bin/test script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added support for adding two nullable columns.
 
-* Addded support for unsigned types in MySQL.
+* Added support for unsigned types in MySQL.
+
+* Added the `any_pending_migrations` function to determine if there are any pending migrations in the migrations directory.
+
+* Added the `migration pending` command to the Diesel CLI to run the `any_pending_migrations` function and report the output.
 
 
 ## [0.12.0] - 2017-03-16
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -127,6 +127,24 @@ pub fn mark_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path)
     Ok(migrations)
 }
 
+// Returns true if there are outstanding migrations in the migrations directory, otherwise
+// returns false. Returns an `Err` if there are problems with migration setup.
+///
+/// See the [module level documentation](index.html) for information on how migrations should be
+/// structured, and where Diesel will look for them by default.
+pub fn any_pending_migrations<Conn>(conn: &Conn) -> Result<bool, RunMigrationsError> where
+    Conn: MigrationConnection
+{
+    let migrations_dir = find_migrations_directory()?;
+    let all_migrations = migrations_in_directory(&migrations_dir)?;
+    let already_run = conn.previously_run_migration_versions()?;
+
+    let pending = all_migrations.into_iter()
+        .any(|m| !already_run.contains(&m.version().to_string()));
+
+    Ok(pending)
+}
+
 /// Reverts the last migration that was run. Returns the version that was reverted. Returns an
 /// `Err` if no migrations have ever been run.
 ///
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -28,6 +28,8 @@ pub fn build_cli() -> App<'static, 'static> {
                     for testing that a migration can in fact be reverted.")
         ).subcommand(SubCommand::with_name("list")
             .about("Lists all available migrations, marking those that have been applied.")
+        ).subcommand(SubCommand::with_name("pending")
+             .about("Returns true if there are any pending migrations.")
         ).subcommand(SubCommand::with_name("generate")
             .about("Generate a new migration with the given name, and \
                     the current timestamp as the version"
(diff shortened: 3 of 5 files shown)


diesel/src/migrations/mod.rs
```diff
@@ -108,6 +108,34 @@ pub fn run_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &P
...
+///
+/// See the [module level documentation](index.html) for information on how migrations should be
+/// structured, and where Diesel will look for them by default.
+pub fn any_pending_migrations<Conn>(conn: &Conn) -> Result<bool, RunMigrationsError> where
+    Conn: MigrationConnection
+{
+    let migrations_dir = find_migrations_directory()?;
+    let pending_migrations = pending_migrations_in_directory(conn, &migrations_dir)?;
+
+    Ok(!pending_migrations.is_empty())
```
killercup: Your code is nice, I think you can write it a bit more concisely, though. What you are checking is that there is any migration that was not run, right? You don't ever need the list of pending migrations. So, let's use `.any`:

```rust
let all_migrations = migrations_in_directory(migrations_dir)?;
let already_run = conn.previously_run_migration_versions()?;

Ok(all_migrations.into_iter()
    .any(|m| !already_run.contains(&m.version().to_string()))
```

DorianListens: Great suggestion! I wasn't aware of the `.any` method!

DorianListens:
I've updated the code, but I'm not exactly sure how to test this behaviour. I've looked at the migration tests for diesel_cli, but they do all their testing through the CLI, which makes sense.

I experimented with adding a new migration subcommand for this, but that doesn't seem very useful. I can push that here if you think it would be a worthwhile addition to the CLI, but otherwise I'd appreciate further guidance on adding tests at this level.

DorianListens:
I've pushed up a commit that adds a subcommand here and some tests. As I said previously, I don't know that this is necessary, but I figured I'd put it up so others can comment. I'd be happy to remove this commit if you think thats the right thing to do.

DorianListens:
The AppVeyor failure seems to have been caused by a database deadlock, I can't imagine it was introduced by this change.

DorianListens:
I've rebased this on master, but looking at #868 it might be better to wait until the suggested refactoring lands and then include this method in the new `Migrations` struct. Let me know what you think!

DorianListens:
Also: the `Migrations` struct would be a perfect place to add some more tests for this without needing to add it to the CLI.

diesel/src/migrations/mod.rs
```diff
@@ -127,6 +127,22 @@ pub fn mark_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path)
...
+/// structured, and where Diesel will look for them by default.
+pub fn any_pending_migrations<Conn>(conn: &Conn) -> Result<bool, RunMigrationsError> where
+    Conn: MigrationConnection
+{
+    let migrations_dir = find_migrations_directory()?;
+    let all_migrations = migrations_in_directory(&migrations_dir)?;
+    let already_run = conn.previously_run_migration_versions()?;
+
+    Ok(all_migrations.into_iter()
+        .any(|m| { !already_run.contains(&m.version().to_string()) }))
```
killercup: FYI, no need for the inner braces around the closure. I'd probably also split off the `Ok` wrapping, i.e., `let pending = all…; \n Ok(pending)`

Eijebong:
This seems fine to me, the only thing is that it would be great to return a different code when we have pending migrations so it's easier to use from shell scripts but apparently we always return 0 for every command right now.

This can be merged as is I think, and we have tests so it will not be forgotten during the `Migrations` rework if it ever happens :)

diesel/src/migrations/mod.rs
Eijebong: :+1:

DorianListens:
Updated the implementation to remove the unnecessary braces. Anything else, or is this ready to be merged?

killercup:
@DorianListens looks good. Can you add a change log entry real quick? I'll merge it this evening.

CHANGELOG.md
```diff
@@ -13,7 +13,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added support for adding two nullable columns.
 
-* Addded support for unsigned types in MySQL.
+* Added support for unsigned types in MySQL.
```
killercup: Nice 👍

Eijebong: Haha, you don't like addding things ? :(

killercup: It took me literally half a minute to see what has changed in this line xD

Eijebong: Glad I'm not the only one. I need to automagically enable spell checking for .md files to avoid things like that in the future ^^

killercup: Spell checking sounds like a hard problem to get right in an editor plugin. Just write one that allows you to quickly create Github issues "check speling" tagged as "easy for newcomers" instead! :trollface:

PR #817, opened 2017-03-21
Upgrade Clippy to 0.121

DorianListens:
Previous versions of clippy no longer compile with nightly rust after 2017-03-17.
See rust-clippy#1633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-  - rust: nightly-2017-03-04
+  - rust: nightly-2017-03-20
     env: CLIPPY=YESPLEASE
     script:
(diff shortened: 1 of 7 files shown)


killercup:
Thanks! You'll also need to bump the nightly version in the `.travis.yml` file.

DorianListens:
Oh, of course! Thanks!

DorianListens:
My change to the `.travis.yml` file didn't seem to actually get travis to upgrade it's nightly rust version. Is there something else I need to do here?

killercup:
That is weird! Can you push a new comment (maybe just edit/add a comment or something like that) to trigger travis again and see if it's just a caching problem?

DorianListens:
Yep, that seemed to do the trick!

killercup:
Cool, thank's for that. Can you amend without that comment now? 😄 I'll merge it immediately.

DorianListens:
Whoops!

killercup:
Thanks again!

PR #821, opened 2017-03-22
Update dotenv to 0.9

Eijebong:
maybe It should be >= 0.8, < 0.10 ?


killercup:
Yeah, if you can use `>= 0.8, < 0.10`, it will not be breaking change for our users :)

Eijebong:
I'll change it tonight. What about examples ? Should I loosen those dependencies too?

killercup:
No need to change the examples, just the library core crates.

felipesere:
Nudge... Can this be merged? 😸

Issue #358, opened 2016-06-17
(slightly) more readable migration filenames?


davidszotten:
What would you think about inserting an underscore between the date and time in the generated migration filenames? 

for the odd occasion when typing (tab completing) in a terminal, i found it hard to scan the list of migrations to figure out which number to add to disambiguate and continue tab completing:

```
$ rm migrations/2016061<TAB>
20160612132127_add_foo/      20160614173422_add_bar/       20160615123204_add_baz/
20160614153823_add_quu/      20160615122959_add_quux/
```

vs something like?

```
20160612_132127_add_foo/      20160614_173422_add_bar/       20160615_123204_add_baz/
20160614_153823_add_quu/      20160615_122959_add_quux/
```

killercup:
I'm pretty sure you can write it like `20160612-132127_add_foo` or even `2016-06-12T13-21-27_add_foo`. The 'version' is determined by the filename part before the first `_`.

- - -

Browsing through the migration code, it looks like for all entries in the `migrations/` directory, [`migration_from(path)`](https://github.com/diesel-rs/diesel/blob/f0a75bbcb56e2764da6ec5fd12dd950726b837d4/diesel/src/migrations/migration.rs#L12) is called, which tries to build a tuple struct with `(path, version)`, where the 'version' is determined by [`version_from(path)`](https://github.com/diesel-rs/diesel/blob/f0a75bbcb56e2764da6ec5fd12dd950726b837d4/diesel/src/migrations/migration.rs#L36). This is where it gets interesting: It just splits the string at the first `_` and returns the first part:

```rust
.split("_")
.nth(0)
```

davidszotten:
the dates are prepended by the cli when calling e.g. `diesel migration generate foo`. 

indeed, as you've discovered, internally diesel seems to only use the first part:
```
table __diesel_schema_migrations
+----------------+----------------------------+
|        version | run_on                     |
|----------------+----------------------------|
| 20160612132127 | 2016-06-15 11:38:36.130818 |
| 20160614153823 | 2016-06-15 11:38:36.130818 |
| 20160614173422 | 2016-06-15 11:38:36.130818 |
| 20160615122959 | 2016-06-15 11:38:36.130818 |
| 20160615123204 | 2016-06-15 11:38:36.130818 |
+----------------+----------------------------+
SELECT 5
Time: 0.002s
```

hence this issue

killercup:
[...]

PR #827, opened 2017-03-24
Generate more human-readable migration dir names

ggrochow:
Seperate Year/Month/Day with - when generating migration directory names, results in the following: 
```
$ diesel migration generate add_table_users

Creating migrations/2017-03-24-061420_add_table_users/up.sql
Creating migrations/2017-03-24-061420_add_table_users/down.sql
```
Format I went with  `YYYY-MM-DD-HHMMSS_migration_name` ( my preferred middle ground from #358 )

I haven't created any new-tests, as I'd like to see how others feel about the `.replace('-', "")` before I do.
I did this so the resulting migrations will still generate the same 'version' slug as the old-ones would. Though im not sure if we care about that at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   tables. You will need to explicitly invoke `joinable!` instead, unless you are
   using `infer_schema!`
 
+* Changed the migration directory name format to `%Y-%m-%d-%H%M%S`.
+
 ### Removed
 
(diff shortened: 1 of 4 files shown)


sgrif:
@ggrochow Sorry that this never got reviewed. Are you still interested in working on this? This change looks fine, it'll just need a changelog entry (and a rebase)

Eijebong:
I took the liberty of rebasing and adding a CHANGELOG entry. Thanks for your work on this @ggrochow

killercup:
@Eijebong CI failure seems legit

killercup:
Thanks, CI seems happy, merging

PR #829, opened 2017-03-24
add sqlite-bundled feature

king6cong:
use a bundled version of sqlite3 seems to be
a common option for Windows/iOS/Android


sgrif:
You can accomplish this by adding `libsqlite3` to your `Cargo.toml` directly with the features you'd like

vitiral:
For others seeing this, those features are documented here:

https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/Cargo.toml

PR #831, opened 2017-03-30
Add cli option `migration list`

jaemk:
- Move migration-version timestamp format to a global static
  `TIMESTAMP_FORMAT`
  (^^ may need to take this into account for  #827 )
- Added function `mark_migrations_in_directory` that compares available
  & applied migrations and returns a `Vec` of `(Option<PathBuf>, bool)`
  where `true` indicates the migration has been applied.
- Added cli arg `migration list` to invoke `mark_migrations_in_directory`
  and display the status of each migration found in `migrations_dir` in
  chronological order.
  e.g.
```
    Migrations:
      [X] longtimestamp_initial
      [ ] longtimestamp_alter-something
```

### DIFF
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -79,6 +79,9 @@ use {Connection, QueryResult};
 use std::env;
 use std::path::{PathBuf, Path};
 
+
+pub static TIMESTAMP_FORMAT: &'static str ="%Y%m%d%H%M%S";
+
 /// Runs all migrations that have not yet been run. This function will print all progress to
 /// stdout. This function will return an `Err` if some error occurs reading the migrations, or if
 /// any migration fails to run. Each migration is run in its own transaction, so some migrations
@@ -106,6 +109,22 @@ pub fn run_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &P
 {
     let all_migrations = try!(migrations_in_directory(migrations_dir));
     run_migrations(conn, all_migrations, output)
+}
+
+/// Compares migrations found in `migrations_dir` to those that have been applied.
+/// Returns a list of pathbufs and whether they have been applied.
+pub fn mark_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path)
+    -> Result<Vec<(Option<PathBuf>, bool)>, RunMigrationsError> where
+        Conn: MigrationConnection,
+{
+    let migrations = migrations_in_directory(migrations_dir)?;
+    setup_database(conn)?;
+    let already_run = conn.previously_run_migration_versions()?;
+    let migrations = migrations.into_iter().map(|m| {
+        let applied = already_run.contains(&m.version().to_string());
+        (m.file_path().map(|p| p.to_path_buf()), applied)
+    }).collect();
+    Ok(migrations)
 }
 
 /// Reverts the last migration that was run. Returns the version that was reverted. Returns an
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -39,6 +39,8 @@ use std::path::{PathBuf, Path};
 use std::{env, fs};
 
 use self::database_error::{DatabaseError, DatabaseResult};
+use diesel::migrations::TIMESTAMP_FORMAT;
+
 
 fn main() {
     use self::dotenv::dotenv;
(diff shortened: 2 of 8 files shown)


diesel_cli/src/main.rs
```diff
@@ -73,6 +76,48 @@ fn run_migration_command(matches: &ArgMatches) {
...
+            let migrations = call_with_conn!(database_url, migrations::mark_migrations_in_directory(&dir, &mut stdout()))
+                .unwrap_or_else(handle_error);
+
+            // Since our migrations came from `mark_migrations_in_directory` they should all have valid file names.
+            let mut sorted = migrations.into_iter().map(|(path_buf, applied)| {
+                let file_name = path_buf.clone()
+                    .and_then(|p| p.file_name()
+                              .and_then(|os_str| os_str.to_str()
+                                        .map(|s| s.to_string())))
+                    .expect(&format!("Error getting file-name from {:?}", path_buf));
```
killercup: This looks more complicated than it needs to be. You can make the chaining a bit easier, e.g. by

- using UFCS (`str::to_string` instead of `|s| s.to_string` which tells us what type/trait we are dealing with),
- and/or by using the "`Option::iter` trick" that treats the Option as an iterator (yielding exactly one value) so you can call `flat_map` on it (which consumes stuff that can be turned into an Iterator – including Option and Result!):

    ```rust
    let file_name = path_buf.iter()
        .flat_map(|p| p.file_name())
        .flat_map(::std::ffi::OsStr::to_str)
        .map(str::to_string)
        .next()
        .expect(&format!("Error getting file-name from {:?}", path_buf));
    ```

    Don't overdo this :)

Once thing I missed at first was that `path_buf` is actually an `Option<PathBuf>` and the expect at the end also panics when this is `None`. I'd make that explicit, which allows us to go back to regular chaining without `.iter()`:

```rust
let path_buf = path_buf.expect(&format!("Migration without valid file name"));
let file_name = path_buf.file_name()
    .and_then(::std::ffi::OsStr::to_str)
    .map(str::to_string)
    .expect(&format!("Error getting file name from {:?}", path_buf));
```

diesel/src/migrations/mod.rs
```diff
@@ -108,6 +108,21 @@ pub fn run_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &P
...
+    -> Result<Vec<(Option<PathBuf>, bool)>, RunMigrationsError> where
+        Conn: MigrationConnection,
+{
+    let migrations = migrations_in_directory(migrations_dir)?;
+    setup_database(conn)?;
+    let already_run = conn.previously_run_migration_versions()?;
+    Ok(migrations.into_iter().map(|m| {
+        let applied = already_run.contains(&m.version().to_string());
+        (m.file_path().map(|p| p.to_path_buf()), applied)
+    }).collect::<Vec<_>>())
```
killercup: Can you reorder this a bit?

```rust
let already_run = …;
let migrations = ….into_iter().map(…).collect(); // Can probably infer type from function return type
Ok(migrations)
```

Or, if you want to fail fast when `migrations_in_directory` fails, just leave that on top and shadow the binding with `let migrations = migrations.into_iter…`.

jaemk:
Thanks for reviewing! I've updated with your suggestions and I'll add the changelog entry and new tests shortly!

jaemk:
Changelog entry and tests have been added!

diesel_cli/tests/migration_list.rs
```diff
@@ -0,0 +1,113 @@
...
+    let tag2 = format!("{}_alter", UTC::now().format(TIMESTAMP_FORMAT));
+    p.create_migration(&tag2, "", "");
+
+    let result = p.command("migration")
+        .arg("list")
+        .run();
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    let pattern = format!(r"(\s|.)*\[ \] {}(\s|.)*\[ \] {}(\s|.)*", &tag1, &tag2);
+    let expected = Regex::new(&pattern).unwrap();
+    assert!(expected.is_match(result.stdout()));
```
killercup: I see what you are doing with the regex, but I think `stdout.find(&tag1) < stdout.find(&tag2)` is easier to read :)

diesel_cli/tests/migration_list.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    let result = p.command("migration")
+        .arg("list")
+        .run();
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    let pattern = format!(
+        r"(\s|.)*\[ \] {}(\s|.)*\[ \] {}(\s|.)*\[ \] {}(\s|.)*\[ \] {}(\s|.)*\[ \] {}(\s|.)*",
+        &tag1, &tag2, &tag3, &tag4, &tag5);
+    let expected = Regex::new(&pattern).unwrap();
+    assert!(expected.is_match(result.stdout()));
```
killercup: Similarly here. Actually, if you use specific datetimes instead of `now`, you could compare the whole string (and it'd be more deterministic).

diesel_cli/tests/migration_list.rs
jaemk: Yes, using `.find` looks much better and makes more sense for the test with a bunch of `tags`! I'll change that up later today.

diesel_cli/tests/migration_list.rs
jaemk: I left explicitly formatted datetimes out of the expected output in case the `migrations::TIMESTAMP_FORMAT` ever changes.

killercup: Yeah, good point. Btw, is the output of that `TIMESTAMP_FORMAT` tested anywhere?

jaemk: It looks like one of the `diesel_cli` test mods has a `migration_generate_creates_a_migration_with_the_proper_name` test, but it uses a regex that only verifies the expected length of the timestamp. I'll update that so it asserts the timestamp matches `migrations::TIMESTAMP_FORMAT`.

jaemk:
Tests are updated!

diesel_cli/tests/migration_list.rs
```diff
@@ -43,6 +43,22 @@ fn migration_list_lists_pending_applied_migrations() {
...
+            .find(&output).expect(&format!("Expected output to contain timestamp: {:?}", tag));
+        match prev {
+            None => prev = Some(mat.0),
+            Some(prev_index) => {
+                let cur_index = mat.0;
+                assert!(prev_index < cur_index);
+                prev = Some(cur_index);
+            }
+        }
+    }
```
jaemk: Using some form of `windows`/`tuple_windows` we can also have a better assertion error message. I think I'll just use `std::slice::windows` for now so we don't need to pull in itertools ([playpen](https://play.integer32.com/?gist=af34bf9c9c509427b413847cb6d3d81a&version=undefined)).

> Is `tag` ever a Regex

Woops, you're right, we can just use `str::find`

> `RegexSet`

That's really useful!

jaemk:
`migration_list` tests are cleaned up. Any way to re-run the appveyor tests without pushing another commit? Looks like it deadlocked on an unrelated test.

Eijebong:
Yeah that deadlock is annoying :(

I think commits in this PR need to be squashed anyway so just add one or amend the last one.

jaemk:
Alright, tests are updated and commits are cleaned up.

PR #834, opened 2017-04-03
Fix `Connection.transaction` rustdoc

dbrgn:
The `TransactionError` type does not exist anymore since v0.11.


killercup:
Thanks!

PR #840, opened 2017-04-05
Expression examples

dbrgn:
Add more docs regarding expressions.

I added a new table called `animals`, hope that's ok. It contains an integer column (for testing aggregation functions) and a nullable column (for testing things like `count(<col>)`.

I could not get the avg example to work, since it return a `Numeric` when folding integers, which doesn't seem to be processable form rust at the moment.

### DIFF
--- a/diesel/src/expression/count.rs
+++ b/diesel/src/expression/count.rs
@@ -23,7 +23,7 @@ use types::BigInt;
 /// #         name -> VarChar,
 /// #     }
 /// # }
-///
+/// #
 /// # fn main() {
 /// #     use self::animals::dsl::*;
 /// #     let connection = establish_connection();
--- a/diesel/src/expression/count.rs
+++ b/diesel/src/expression/count.rs
@@ -10,7 +10,7 @@ use types::BigInt;
 /// it specifically as `diesel::expression::count`, or glob import
 /// `diesel::expression::dsl::*`
 ///
-/// # Example
+/// # Examples
 ///
 /// ```rust
 /// # #[macro_use] extern crate diesel;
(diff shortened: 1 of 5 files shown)


diesel/src/expression/count.rs
```diff
@@ -9,6 +9,27 @@ use types::BigInt;
...
+/// # include!("src/doctest_setup.rs");
+/// # use diesel::expression::dsl::*;
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+///
```
killercup: This will be rendered as an empty line at the beginning of the code block

diesel/src/expression/count.rs
```diff
@@ -9,6 +9,27 @@ use types::BigInt;
...
+/// # #[macro_use] extern crate diesel;
+/// # include!("src/doctest_setup.rs");
+/// # use diesel::expression::dsl::*;
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
```
dbrgn: No, it is required. Without it:

```
error[E0433]: failed to resolve. Use of undeclared type or module `users`
  --> src/doctest_setup.rs:141:6
   |
141|     (users)
   |      ^ Use of undeclared type or module `users`
```

I thought about moving the table!  macro into the `doctest_setup.rs` module, but in some places the invocation is being displayed in the docs, so I didn't touch it and instead copied it around.

dbrgn:
@killercup did the requested changes and fixed the tests.

Regarding the user table, see my comment above.

Let me know if I should rebase / squash the commits in this PR.

diesel/src/expression/count.rs
killercup: Ah, interesting. Let's keep it like this for now then and refactor it later.

killercup:
Thanks! Squashed and merged :)

Issue #842, opened 2017-04-05
Arithmetic on optionals


dbrgn:
This does not work...

    update(posts.filter(id.eq(7)).set(read_count.eq(read_count + 1))

...if `read_count` is nullable.

According to my tests, Pg, MySQL and SQLite all return null for `SELECT NULL + 1`.

This page seems to agree: http://www-cs-students.stanford.edu/~wlam/compsci/sqlnulls

> An arithmetic operation involving a NULL returns NULL. For example, NULL minus NULL yields NULL, not zero.

I didn't check the SQL ISO standard though.

Note also that aggregation function like `SUM()` seem to ignore null values.

Maybe addition and other arithmetic operations could be implemented on `Nullable<Integer>` and others?

sgrif:
The ISO standard is that any operation on a null value returns null. I'm fine with adding ops for nullable values (preferably as a single blanket impl if possible)

PR #846, opened 2017-04-06
Allow numeric operations on nullable columns.

Eijebong:
For now you can only do operations between two nullable columns as rust
doesn't allow us to implement a trait twice for the same type without
specialization.

### DIFF
--- a/diesel_tests/tests/expressions/ops.rs
+++ b/diesel_tests/tests/expressions/ops.rs
@@ -93,6 +93,77 @@ fn dividing_column() {
 }
 
 #[test]
+fn test_adding_nullables() {
+    use schema::nullable_table::dsl::*;
+    let connection = connection_with_nullable_table_data();
+
+    let expected_data = vec![None, None, Some(2), Some(3), Some(2)];
+    let data = nullable_table.select(value + Some(1)).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data: Vec<Option<i32>> = vec![None; 5];
+    let data = nullable_table.select(value + None as Option<i32>).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data = vec![None, None, Some(2), Some(4), Some(2)];
+    let data = nullable_table.select(value + value).load(&connection);
+    assert_eq!(Ok(expected_data), data);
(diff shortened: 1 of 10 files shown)


dbrgn:
Refs #842.

diesel_tests/tests/associations.rs
```diff
@@ -137,25 +137,7 @@ fn associations_can_be_grouped_multiple_levels_deep() {
...
-    let test_data = vec![
-        Tree { id: 1, parent_id: None },
-        Tree { id: 2, parent_id: None },
-        Tree { id: 3, parent_id: Some(1) },
-        Tree { id: 4, parent_id: Some(2) },
-        Tree { id: 5, parent_id: Some(1) },
-    ];
-    insert(&test_data).into(trees::table)
-        .execute(&conn).unwrap();
+    let (conn, test_data) = connection_with_trees();
```
sgrif: This makes the test significantly harder to read for me.

diesel_tests/tests/expressions/ops.rs
```diff
@@ -93,6 +93,77 @@ fn dividing_column() {
 }
 
 #[test]
+fn test_adding_nullables() {
+    use schema::trees::dsl::*;
+    let (connection, _) = connection_with_trees();
+
+    let expected_data = vec![None, None, Some(2), Some(3), Some(2)];
+    let data = trees.select(parent_id + Some(1)).load(&connection);
```
sgrif: I'd rather we just create a table specifically for testing this.

diesel_tests/tests/associations.rs
Eijebong: Ok, i'll create a table for that then :p

Eijebong:
The nested Nullable thing was just me not understanding something.
A nullable + a non nullable column will not compile though. Should I add a compile-test for that ?

killercup:
> A nullable + a non nullable column will not compile though. Should I add a compile-test for that ?

Hm, that's not really a feature, though, is it?

Is there anything that stands in the way of merging this? I didn't really follow the discussion that lead to the implementation, so I'm not sure which edge cases to look for. Maybe open an issue for (a discussion about) unlocking more possibilities with specialization?

PR #848, opened 2017-04-06
add ability to skip timestamp helpers

### LINKED ISSUE #522: Migrations not supported on CockroachDB
When running `diesel setup` from the getting started example with CockroachDB beta-20161201, I get this error:

```
thread 'main' panicked at 'Error creating timestamp helper functions for Pg: DatabaseError(__Unknown, "syntax error at or near \"OR\"\nCREATE OR REPLACE FUNCTION diesel_set_updated_at() RETURNS trigger AS $$\n       ^\n")', ../src/libcore/result.rs:788
```

SecurityInsanity:
fixes #522

this adds the ability to skip timestamp helpers for postgres which
should allow cockroach migrations to pass so you can start running
normal migrations.

I still generate the `setup_helper_functions` function incase there ever is more helpers that need to be added to postgres. If you'd prefer I'd axe that function altogether, and add in `if cfg!` elsewhere, let me know and I'll be happy to do that.


SecurityInsanity:
Seems that appveyor build failed to download rustup, and not something I did:
```appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
Error downloading remote file: One or more errors occurred.
Inner Exception: The request was aborted: Could not create SSL/TLS secure channel.
Command exited with code 2```

Is there a way I can retrigger that?

**Comment by SecurityInsanity** [comment]
> I'm not sure this is the right way to do this, though. First off, cargo features are meant to be additive, but this basically disables a feature. In this case I don't think it's much of a problem.

Yea It's admittedly not the best solution. Really I'd love to just do the work necessary to make CockroachDB a fully supported backend, but:

  - That's a pretty big change.
  - Has side effects for everyone maintaining this.
  - Means we have to add it to the testing setup, and there isn't a good way to spin one up really without something like docker (a dependency I don't think is worth adding for just cockroach).

> I'm generally not very happy with these timestamp helpers. Not because of the helpers themselves, they are quite nice, in fact, but that they are an implicitly available feature. I've talked with @sgrif about these helpers in February (not in public chat, sorry) and one idea was to introduce a "default migration" (when calling diesel setup) that adds them. This would mean that we need to make this work with existing code bases/migrations though, and it was not worth the effort to revisit this until now.

A default migration being introduced could be a nice change (assuming it could be removed, for say this purpose here (mashing one db to work with another dbs settings)). I'd still think it'd be a worthwhile change so it's easier to tell what's being added by diesel. I'd be interested in taking a look into it at the very least.

**Comment by SecurityInsanity** [comment]
After digging through default migrations for awhile now. I've run into roadblocks with how that'd work for existing consumers. Any solution I seem to come up with seems subpar.

However that being said I'd _*really*_ love to get the ability to skip these in. Do you have some ideas about working with existing code bases I could pick your mind on, and/or a way you can think to do this the "right way" :tm: ?

**Comment by jwadsager** [comment]
What roadblocks did you encounter?

File: diesel/Cargo.toml [hunk 1]
```diff
@@ -39,6 +39,7 @@ lint = ["clippy"]
 large-tables = []
 huge-tables = ["large-tables"]
 postgres = ["pq-sys"]
+skip-timestamp-helpers = []
```
abaumhauer: Adding a cockroach feature would be ideal.  But in this case having a feature the enables the use of triggers in the database is also a good idea. This is business logic being defined in the database. It violates the rules of clean architecture by creating coupling between the app and the db. It's also not portable to the sqlite3 and mysql/mariadb back ends.

PR #849, opened 2017-04-06
Add ability to document items of table! macro

### LINKED ISSUE #378: missing documentation through diesel macros
As I deny missing docs in my actual project, is it possible, to exclude this behaviour from the diesel macros?

My `schema.rs` looks like this:

``` rust
#[allow(missing_docs)]
infer_schema!(dotenv!("DATABASE_URL"));
```

... but this keeps throwing errors:

``` sh
<diesel macros>:27:33: 38:53 error: missing documentation for a module
<diesel macros>:27 ( $ ( $ column_name , ) + ) } } pub mod columns {
                                                   ^
<diesel macros>:5:1: 5:71 note: in this expansion of table_body! (defined in <diesel macros>)
src/schema.rs:2:1: 2:40 note: in this expansion of table! (defined in <diesel macros>)
src/schema.rs:2:1: 2:40 note: in this expansion of infer_schema! (defined in src/lib.rs)
src/lib.rs:1:9: 1:21 note: lint level defined here
src/lib.rs:1 #![deny(missing_docs)]
                     ^~~~~~~~~~~~
```

weiznich:
This commit adds optinal syntax to the table! dsl to document the
generated code. It is now possible to add an doc-comment for the
generated module and for each column struct.

This should partially fix #378

cc @killercup

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -188,7 +188,7 @@ macro_rules! __diesel_column {
 /// ```
 ///
 ///
-/// If you want to add documentation for the generated code you could use the
+/// If you want to add documentation to the generated code you can use the
 /// following syntax:
 ///
 /// ```
@@ -196,11 +196,11 @@ macro_rules! __diesel_column {
 ///
 /// table! {
 ///
-///     /// Table documentation
+///     /// The table containing all blog posts
 ///     posts {
-///         /// Id column documentation
+///         /// The post's unique id
 ///         id -> Integer,
-///         /// Title column documentation
+///         /// The post's title
 ///         title -> Text,
 ///     }
 /// }
@@ -349,7 +349,6 @@ macro_rules! table {
             doc = [$($doc)*];
         }
     };
-
     // Terminal with composite pk (add a trailing comma)
     (
         @parse_body
@@ -364,13 +363,19 @@ macro_rules! table {
         }
     };
     // Terminal with invalid syntax
+    // This is needed to prevent unbounded recursion on for example
+    // table! {
+    //     something strange
+    // }
     (
         @parse_body
         import = [$(use $($import:tt)::+;)*];
         doc = [$($doc:meta,)*];
         $($rest:tt)*
     ) => {
-        const ERROR:i32 = "invalid table! syntax";
+        // Remove this workaround as soon as rust issue 40872 is implemented
+        // https://github.com/rust-lang/rust/issues/40872
+        const ERROR: i32 = env!("invalid table! syntax");
     };
     // Put a parse annotation and empty fields for imports and documentation on top
     // This is the entry point for parsing the table dsl
@@ -387,6 +392,9 @@ macro_rules! table {
 #[macro_export]
 #[doc(hidden)]
 macro_rules! table_body {
+    // Parse a table column definition with documentation
+    // Forward any remaining table column to further instances
+    // of this macro
     (
         schema_name = $schema_name:ident,
         table_name = $name:ident,
@@ -397,7 +405,7 @@ macro_rules! table_body {
         table_doc = [$($table_doc:meta)*],
         #[$new_doc:meta] $new_column_name:ident -> $new_type:ty,
         $($body:tt)*
-    )=> {
+    ) => {
         table_body! {
             schema_name = $schema_name,
             table_name = $name,
@@ -410,6 +418,9 @@ macro_rules! table_body {
             $($body)*
         }
     };
+    // Parse a table column definition without documentation
+    // Forward any remaining table column to further instances
+    // of this macro
     (
         schema_name = $schema_name:ident,
         table_name = $name:ident,
@@ -420,7 +431,7 @@ macro_rules! table_body {
         table_doc = [$($table_doc:meta)*],
         $new_column_name:ident -> $new_type:ty,
         $($body:tt)*
-    )=> {
+    ) => {
         table_body! {
             schema_name = $schema_name,
             table_name = $name,
@@ -433,6 +444,8 @@ macro_rules! table_body {
             $($body)*
         }
     };
+    // Parse the table name  and the primary keys
+    // Forward the table body to further parsing layers
     (
         $schema_name:ident . $name:ident ($pk:ident) {
             $($body:tt)+
@@ -799,6 +812,18 @@ mod tests {
 
         table_with_custom_types {
             id -> Integer,
+            my_type -> MyCustomType,
+        }
+    }
+
+    table! {
+        use types::*;
+        use macros::tests::my_types::*;
+
+        /// Table documentation
+        table_with_custom_type_and_id (a){
+            /// Column documentation
+            a -> Integer,
             my_type -> MyCustomType,
         }
     }
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -2,7 +2,6 @@
 #[doc(hidden)]
 macro_rules! __diesel_column {
     ($($table:ident)::*, $column_name:ident -> $Type:ty, $($doc:expr),*) => {
-
         $(
             #[doc=$doc]
         )*
@@ -256,83 +255,93 @@ macro_rules! table {
     (
         @parse
         import = [$(use $($import:tt)::+;)*];
-        doc = [$($doc:expr,)*];
+        table_doc = [];
         use $($new_import:tt)::+; $($rest:tt)+
     ) => {
         table! {
             @parse
             import = [$(use $($import)::+;)* use $($new_import)::+;];
-            doc = [$($doc,)*];
+            table_doc = [];
             $($rest)+
         }
     };
+
     // Put doc annotation into the doc field
     (
         @parse
         import = [$(use $($import:tt)::+;)*];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         #[doc=$new_doc:expr] $($rest:tt)+
     ) => {
         table! {
             @parse
             import = [$(use $($import)::+;)*];
-            doc = [$($doc,)*$new_doc,];
+            table_doc = [$($doc,)*$new_doc,];
             $($rest)+
         }
     };
-    // Forward to next parsing step
+
+    // We are finished parsing the import list and the table documentation
+    // Now we forward the remaining tokens to parse the body of the table
+    // definition
     (
         @parse
         import = [$(use $($import:tt)::+;)+];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $($rest:tt)+
     ) => {
         table! {
             @parse_body
             import = [$(use $($import)::+;)*];
-            doc = [$($doc,)*];
+            table_doc = [$($doc,)*];
             $($rest)+
         }
     };
-    // Forward to next parsing step and add default import (diesel::types::*)
+
+    // We are finished parsing the import list and the table documentation
+    // Because the import list is empty we add a default import (diesel::types::*)
+    // After that we forward the remaining tokens to parse the body of the table
+    // definition
     (
         @parse
         import = [];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $($rest:tt)+
     ) => {
         table! {
             @parse_body
             import = [use $crate::types::*;];
-            doc = [$($doc,)*];
+            table_doc = [$($doc,)*];
             $($rest)+
         }
     };
+
     // Add the primary key if it's not present
     (
         @parse_body
         import = [$(use $($import:tt)::+;)+];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $($table_name:ident).+ {$($body:tt)*}
     ) => {
         table! {
             @parse_body
             import = [$(use $($import)::+;)+];
-            doc = [$($doc,)*];
+            table_doc = [$($doc,)*];
             $($table_name).+ (id) {$($body)*}
         }
     };
+
     // Add the schema name if it's not present
     (
         @parse_body
         import = [$(use $($import:tt)::+;)+];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $name:ident $(($($pk:ident),+))* {$($body:tt)*}
     ) => {
         table! {
             @parse_body
             import = [$(use $($import)::+;)+];
-            doc = [$($doc,)*];
+            table_doc = [$($doc,)*];
             public . $name $(($($pk),+))* {$($body)*}
         }
     };
@@ -340,28 +349,30 @@ macro_rules! table {
     (
         @parse_body
         import = [$(use $($import:tt)::+;)+];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $schema_name:ident . $name:ident ($pk:ident) $body:tt
     ) => {
         table_body! {
             $schema_name . $name ($pk) $body
             import = [$(use $($import)::+;)+];
-            doc = [$($doc)*];
+            table_doc = [$($doc)*];
         }
     };
+
     // Terminal with composite pk (add a trailing comma)
     (
         @parse_body
         import = [$(use $($import:tt)::+;)+];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $schema_name:ident . $name:ident ($pk:ident, $($composite_pk:ident),+) $body:tt
     ) => {
         table_body! {
             $schema_name . $name ($pk, $($composite_pk,)+) $body
             import = [$(use $($import)::+;)+];
-            doc = [$($doc)*];
+            table_doc = [$($doc)*];
         }
     };
+
     // Terminal with invalid syntax
     // This is needed to prevent unbounded recursion on for example
     // table! {
@@ -370,20 +381,21 @@ macro_rules! table {
     (
         @parse_body
         import = [$(use $($import:tt)::+;)*];
-        doc = [$($doc:expr,)*];
+        table_doc = [$($doc:expr,)*];
         $($rest:tt)*
     ) => {
         // Remove this workaround as soon as rust issue 40872 is implemented
         // https://github.com/rust-lang/rust/issues/40872
         const ERROR: i32 = env!("invalid table! syntax");
     };
+
     // Put a parse annotation and empty fields for imports and documentation on top
     // This is the entry point for parsing the table dsl
     ($($rest:tt)+) => {
         table! {
             @parse
             import = [];
-            doc = [];
+            table_doc = [];
             $($rest)+
         }
     }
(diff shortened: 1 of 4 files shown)


diesel/src/macros/mod.rs
```diff
@@ -194,6 +198,24 @@ macro_rules! __diesel_column {
...
+///
+/// ```
+/// #[macro_use] extern crate diesel;
+///
+/// table! {
+///
+///     /// Table documentation
+///     posts {
+///         id -> Integer => "Id column documentation",
+///         title -> Text => "Title column documentation",
```
killercup: Can you change this to expect a doc attr as well? I.e.,

```rust
/// Title column documentation
title -> Text,
```

(I'm not sure where exactly attrs are allowed, but here it seems to work: [playground](https://play.rust-lang.org/?gist=a22c8d3556c6e51efe91e88a86004771&version=stable&backtrace=0))

weiznich: In theory this could work, but when I tried it it resulted in parsing ambiguous.

weiznich:
> haven't read much of the code, but it currently fails on CI for this case:
```
error: no rules expected the token `{`
   --> src/macros/mod.rs:752:33
    |
748 |       table! {
    |  _____- starting here...
749 | |         use types::*;
750 | |         use macros::tests::my_types::*;
751 | |
752 | |         table_with_custom_types {
    | |                                 ^
753 | |             id -> Integer,
754 | |             my_type -> MyCustomType,
755 | |         }
756 | |     }
    | |_____- ...ending here: in this macro invocation
```

This is caused by the second rewriting rule that is used to parse the `#[doc]` attribute and place the content to the back. A similar error would pop up if we use the same syntax for the columns. (From a short view seems like it is very hard or even impossible to provide the `///` syntax with the current implementation of the `table!` macro)



I'm a few weeks abroad, so I'm not able to work in this time on this. I will continue the development as soon as I returned. (Feel free to close the pull request in the mean time.)

diesel/src/macros/mod.rs
weiznich: This is now working.

weiznich:
I've changed the syntax to that one proposed by @killercup and fixed the failing test cases. This should be now ready to be merged.

killercup:
Awesome! Thanks for sticking with this.

The diff looks pretty wild! Haven't read that much macro code in a while. I'll try to review it this evening. Maybe @Eijebong wants to have a look as well and play a round of "human substitution machine"? ;)

@sgrif did we ever talk about this? You might also have some thoughts on this (and the syntax I suggested).

diesel/src/macros/mod.rs
```diff
@@ -194,6 +198,26 @@ macro_rules! __diesel_column {
 /// # fn main() {}
 /// ```
 ///
+///
+/// If you want to add documentation for the generated code you could use the
```
killercup: s/for/to
s/could/can

diesel/src/macros/mod.rs
```diff
@@ -194,6 +198,26 @@ macro_rules! __diesel_column {
...
+/// ```
+/// #[macro_use] extern crate diesel;
+///
+/// table! {
+///
+///     /// Table documentation
+///     posts {
+///         /// Id column documentation
+///         id -> Integer,
+///         /// Title column documentation
```
killercup: Super unimportant, but can you change this to be actually helpful docs? E.g., "The table containing all blog posts", "The post's unique ID", "The post's title"

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
...
         }
     };
+    // Terminal with invalid syntax
+    (
+        @parse_body
+        import = [$(use $($import:tt)::+;)*];
+        doc = [$($doc:meta,)*];
+        $($rest:tt)*
+    ) => {
+        const ERROR:i32 = "invalid table! syntax";
```
killercup: Huh, is that new? Cool idea! But I can't really see when this happens. Can you add a compile-fail test and a comment?

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
...
         table! {
-            $($table_name).+ $(($($pk),+))* {$($body)*}
-            use $crate::types::*;
+            @parse_body
+            import = [use $crate::types::*;];
+            doc = [$($doc,)*];
+            $($rest)+
+        }
+    };
+    // Add the primary key if it's not present
```
killercup: Okay I can recognize the comments/sections but the code is quite a lot different. This will take me a while.

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
 /// ```
 #[macro_export]
 macro_rules! table {
-    // Put `use` statements at the end because macro_rules! cannot figure out
-    // if `use` is an ident or not (hint: It's not)
```
killercup: Is this still valid? Seems like an important comment.

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
...
+            $($rest)+
+        }
+    }
 }
 
 #[macro_export]
 #[doc(hidden)]
 macro_rules! table_body {
     (
+        schema_name = $schema_name:ident,
```
killercup: Can you add comments what these blocks do, like above? (I know there weren't any before but I guess you know what's happening so it'd be awesome to write that down)

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
...
+        table_name = $name:ident,
+        primary_key_ty = $primary_key_ty:ty,
+        primary_key_expr = $primary_key_expr:expr,
+        columns = [$($column_name:ident -> $Type:ty; doc = [$($doc:meta)*],)*],
+        imports = ($($($import:tt)::+),+),
+        table_doc = [$($table_doc:meta)*],
+        #[$new_doc:meta] $new_column_name:ident -> $new_type:ty,
+        $($body:tt)*
+    )=> {
```
killercup: Please put a space between the `)` and the `=` :)

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,235 @@ macro_rules! __diesel_column {
...
+        schema_name = $schema_name:ident,
+        table_name = $name:ident,
+        primary_key_ty = $primary_key_ty:ty,
+        primary_key_expr = $primary_key_expr:expr,
+        columns = [$($column_name:ident -> $Type:ty; doc = [$($doc:meta)*],)*],
+        imports = ($($($import:tt)::+),+),
+        table_doc = [$($table_doc:meta)*],
+        $new_column_name:ident -> $new_type:ty,
+        $($body:tt)*
+    )=> {
```
killercup: Please put a space between the `)` and the `=` :)

diesel/src/macros/mod.rs
weiznich: This is still valid (the part about macro_rules and use).
But now this is not the main cause for moving import to the import field. Now this is done because it is easier to distinguish between import, doc and the table body in this way. So I think this comment is not needed anymore here.

weiznich:
> Do you think you could add some compile-tests for valid and invalid syntax combinations? I'm not sure we have a comprehensive test suite for table!. It's a pretty complicated beast by now, and we should probably at least add tests for new features.

I've tried to add a compile fail macro for invalid syntax, but compile-test complains `error: 0 unexpected errors found, 1 expected errors not found`. It seems like that compile-test does not recognize error that originates outside of the tested crate.

killercup:
This compile fail test works:

```rust

table! {
    /// User's table
    users {
        /// A user's ID
        id -> Integer,
    }
}

table! {
    12
    //~^ ERROR expected ident
}

fn main() {}
```

diesel/src/macros/mod.rs
```diff
@@ -239,108 +263,248 @@ macro_rules! __diesel_column {
...
+    //     something strange
+    // }
+    (
+        @parse_body
+        import = [$(use $($import:tt)::+;)*];
+        doc = [$($doc:meta,)*];
+        $($rest:tt)*
+    ) => {
+        // Remove this workaround as soon as rust issue 40872 is implemented
+        // https://github.com/rust-lang/rust/issues/40872
```
killercup: cc https://github.com/rust-lang/rust/issues/40872 (for cross linking magic)

weiznich:
> Maybe you can add this one with one more case for the explicit error thing you introduced. We can always add more tests later. (I'd add "compile-success" tests as more doc tests.)

For example the following code should trigger the explicit error I've introduced:
```rust
table! {
     some wrong syntax
}
```

Rustc prints the following error:
```
error: environment variable `invalid table! syntax` not defined
  --> src/lib.rs:59:1
   |
59 | / table!{
60 | |     some wrong syntax
61 | | }
   | |_^
   |
```

If I add this `table!` call to your compile-fail-test without an error annotation this test will still pass :(

(For the new syntax there should be at least 2 compile pass test. On doc test and one at the end of macros/mod.rs)

killercup:
Okay, it seems compiletest is a bit fuzzy in what macro errors it finds or even looks for. This works for me:

- one table! call per tests
- exactly one "error-pattern" matcher below

For example:

```rust

table! {
     some wrong syntax
}
// error-pattern: invalid table! syntax

fn main() {}
```

and 

```rust

table! {
    12
}
// error-pattern: expected ident

fn main() {}
```

Also, I just remembered something: You macro syntax allows _all_ attributes, not just doc ones. Is this on purpose? Do we want that?

```rust

table! {
     #[foobar]
     posts {
         id -> Integer,
     }
}
// error-pattern: invalid table! syntax

fn main() {}
```

(This fails with "attribute `foobar` is currently unknown to the compiler" by the way.)

weiznich:
> Okay, it seems compiletest is a bit fuzzy in what macro errors it finds or even looks for. This works for me:
> * one table! call per tests
> * exactly one "error-pattern" matcher below

This did not work for me. (I've added the first two as test cases, but the second one will fail:()


> Also, I just remembered something: You macro syntax allows all attributes, not just doc ones. Is this on purpose? Do we want that?

This could be quite handy, because so you can also add things like `#[allow(missing_docs)]` to the generated code. (Maybe this should also be documented?)

killercup:
> This did not work for me.

Weird. It worked locally, but I'm using a newer nightly than the clippy CI target. Will check again later.

> This could be quite handy, because so you can also add things like #[allow(missing_docs)] to the generated code. (Maybe this should also be documented?)

I'm not convinced we actually want that. It's not clear to the user where the attributes end up, and, more important, this macro should not be seen as something to generate Rust code (IMO), but as something that creates a mapping to "the SQL side".

weiznich:
> I'm not convinced we actually want that. It's not clear to the user where the attributes end up, and, more important, this macro should not be seen as something to generate Rust code (IMO), but as something that creates a mapping to "the SQL side".

On the other hand this macro generates Rust code internally. So there is a possibility that someone manages to set warnings in such a way that the generated code does not compile(For example `#[deny(missing_docs)]` before this pull request). 
There are two ways to handle this:
1. Allow to add those attributes to the generated code. Downside: Expose some underlying implementation details.
2. If such a warning is triggered fix that warning in the generated code. Downside: To fix an eventually occurring issue some time is needed (Implement the "fix", open a pull request, wait for review, wait till a new version containing those change is published).

(It should be no big problem to remove this, but I think this would be a quite helpful feature in some cases)

weiznich:
What is the way forward with this pull request? How do we want to handle the failing compile-fail test?

weiznich:
@killercup ping

killercup:
Ah, thanks for the ping. So, from my side, nothing changed. Personally, I'd like to add this in a small, isolated increment. I.e, this PR should only allow doc comments in `table!` (and nothing else), with all the tests that come with that. I still don't think we should allow arbitrary attributes in this macro.

But this is just my opinion. What do you think, @Eijebong, @sgrif?

@weiznich, can you easily restrict this to only allow doc comments or is that much work? I'll try to look at the compiletest and hope I can get it to work.

weiznich:
Restricting this to doc comments should only require some small change. I think I must change the `$doc:meta` variables to `$doc:expr` and modify the parser from `#[$doc]` to `#[doc=$doc]`.

weiznich:
@killercup Any news on the failing compile test?

killercup:
Just had a look at the compile tests. They work for me.

Just like Travis, I'm running compiletest-rs 0.2.5 with

```
rustc 1.18.0-nightly (63c77214c 2017-04-24)
binary: rustc
commit-hash: 63c77214c1d38789652b465694b254205d1886e0
commit-date: 2017-04-24
host: x86_64-apple-darwin
release: 1.18.0-nightly
LLVM version: 3.9
```

Changing your second test to what I actually wrote above (with `error-pattern`) works:

```rust

table! {
     some wrong syntax
}
// error-pattern: invalid table! syntax

fn main() {}
```

The third one works as well with your latest changes and with this error matcher:

```rust

table! {
     #[foobar]
     //~^ ERROR expected ident, found #
     posts {
         id -> Integer,
     }
}

fn main() {}
```

Other than that, I still don't like the "environment variable `invalid table! syntax` not defined" message, and it's weird how big this PR is for this simple change. But, well…

weiznich:
> Other than that, I still don't like the "environment variable invalid table! syntax not defined" message, 

If there is a better way to produce an error message there, I'm happy to change this. But it seems like as long as [`compile_error`](https://github.com/rust-lang/rust/issues/40872) is not implemented/stable that there is no better solution to achieve this. Without this message the macro will recuse endless.

> it's weird how big this PR is for this simple change

To implement this change some bigger parts of the `table!`-macro parsing logic needed to be restructured. This results in the big diff :disappointed: 

(Compile tests are fixed now, so this should be ready to merge)

diesel/src/macros/mod.rs
```diff
@@ -1,7 +1,11 @@
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __diesel_column {
-    ($($table:ident)::*, $column_name:ident -> $Type:ty) => {
+    ($($table:ident)::*, $column_name:ident -> $Type:ty, $($doc:expr),*) => {
+
```
sgrif: ✂️

diesel/src/macros/mod.rs
```diff
@@ -228,108 +252,248 @@ macro_rules! __diesel_column {
...
+        table! {
+            @parse
+            import = [$(use $($import)::+;)* use $($new_import)::+;];
+            doc = [$($doc,)*];
+            $($rest)+
+        }
     };
-
-    // Add the primary key if it's not present
+    // Put doc annotation into the doc field
```
sgrif: Can you add a blank line in between each branch here?

diesel/src/macros/mod.rs
```diff
@@ -228,108 +252,248 @@ macro_rules! __diesel_column {
...
 #[macro_export]
 macro_rules! table {
-    // Put `use` statements at the end because macro_rules! cannot figure out
-    // if `use` is an ident or not (hint: It's not)
+    // Put imports into the import field
     (
-        use $($import:tt)::+; $($rest:tt)+
+        @parse
+        import = [$(use $($import:tt)::+;)*];
+        doc = [$($doc:expr,)*];
```
sgrif: Do we need to match on doc this early? If nothing else it seems like this could just be `doc = $doc:tt` -- but it seems like we could just not stick the `doc = []` until we get to that point. Also not sure why we're collecting `doc` into a repeater, but I guess I'll find out.

diesel/src/macros/mod.rs
```diff
@@ -228,108 +252,248 @@ macro_rules! __diesel_column {
...
-
-    // Add the primary key if it's not present
+    // Put doc annotation into the doc field
     (
-        $($table_name:ident).+ {$($body:tt)*}
-        $($imports:tt)*
+        @parse
+        import = [$(use $($import:tt)::+;)*];
+        doc = [$($doc:expr,)*];
+        #[doc=$new_doc:expr] $($rest:tt)+
```
sgrif: This is specifically the table level documentation, right? Is it even possible to have more than one of these?

sgrif: Can we make it clear in the name that this is table level documentation?

diesel/src/macros/mod.rs
```diff
@@ -228,108 +252,248 @@ macro_rules! __diesel_column {
...
-            $($table_name).+ (id) {$($body)*} $($imports)*
+            @parse
+            import = [$(use $($import)::+;)*];
+            doc = [$($doc,)*$new_doc,];
+            $($rest)+
         }
     };
-
-    // Add the schema name if it's not present
+    // Forward to next parsing step
```
sgrif: What is "the next parsing step"? What is the current parsing step?

diesel/src/macros/mod.rs
```diff
@@ -228,108 +252,248 @@ macro_rules! __diesel_column {
...
-            public . $name $(($($pk),+))* {$($body)*} $($imports)*
+            @parse_body
+            import = [$(use $($import)::+;)*];
+            doc = [$($doc,)*];
+            $($rest)+
         }
     };
-
-    // Import `diesel::types::*` if no imports were given
+    // Forward to next parsing step and add default import (diesel::types::*)
```
sgrif: Same thing here. What is "the next parsing step"? What did we just finish?

killercup:
One thing we need to ensure here is that we can extend the macro later to allow renames (yes, we want to have something #424 after all).

diesel/src/macros/mod.rs
weiznich: In this case you are right, we do not need to parse doc here (It's removed in my last changes).

In general:
We need to match those `*_doc =[$($doc:expr)*]` things in a repeater because:
 - There may be none
 - A comment may contain multiple lines:
```rust 
/// First line
/// Second line
 ```
   Rustc desugers this into:
```rust
#[doc=r"First line"]
#[doc=r"Second line"]
```
Therefore we must match on multiple doc attributes

weiznich:
@killercup I've updated the pull request to address @sgrif comments. 
As far as I could tell it should be quite easy to plug something like #424 into the "new" parsing structure, but I've not tried it…

weiznich:
Also I've found a small bug involving multi line doc comments on columns. This is fixed by the second commit.

diesel/src/macros/mod.rs
```diff
@@ -833,8 +845,12 @@ mod tests {
         use macros::tests::my_types::*;
 
         /// Table documentation
-        table_with_custom_type_and_id (a){
+        ///
+        /// some in detail documentation
```
killercup: Good test case, but this can only check if the macro parses doc comment, not if it gets output again, right? Should be good for now, though.

weiznich: As far as I know there is no way to check if the macro expands the doc comments correctly. One could only run rustdoc on some example code and check this output manually.

killercup: Yeah, either running rustdoc and checking the html, or running something like cargo-expand and comparing its output.

I'm actually a bit surprised I couldn't find a crate for testing macros. Something like piping `cargo rustc -- --pretty=expanded` into syn (at least its tokenizer) and comparing that to some `quote!`ed expressions. That may be an interesting project in itself.

PR #854, opened 2017-04-10
Add an example for the embed_migrations macro.

Eijebong:


dbrgn:
This doesn't seem to work:

```
  --> src/db/mod.rs:69:77
   |
69 |         Ok(connection) => embedded_migrations::run_with_output(&connection, io::stdout())
   |                                                                             ^^^^^^^^^^^^ expected mutable reference, found struct `std::io::Stdout`
   |
   = note: expected type `&mut std::io::Write`
              found type `std::io::Stdout`
```

Instead, you need to pass a mut reference:

```
embedded_migrations::run_with_output(&connection, &mut std::io::stdout());
```

How come this wasn't caught by the test suite?

Also, maybe replace the `Example:` headline with `# Examples`? See https://github.com/diesel-rs/diesel/pull/840#pullrequestreview-31039462

dbrgn:
Something else I noted: When using `run_with_output` instead of `run`, rust complains about the `run` function not being used:

```
warning: function is never used: `run`, #[warn(dead_code)] on by default
  --> src/db/mod.rs:62:1
   |
62 | embed_migrations!();
   | ^^^^^^^^^^^^^^^^^^^^
   |
```

dbrgn:
Something else I noted: When using `run_with_output` instead of `run`, rust complains about the `run` function not being used:

```
warning: function is never used: `run`, #[warn(dead_code)] on by default
  --> src/db/mod.rs:62:1
   |
62 | embed_migrations!();
   | ^^^^^^^^^^^^^^^^^^^^
   |
```

Maybe @weiznich knows how to deal with the warning?

Eijebong:
Yeah the unused function makes sense, maybe we should add an `#[allow(dead_code)]` on it.

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -72,6 +72,18 @@ macro_rules! infer_table_from_schema {
 ///
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
+///
+/// Example:
+/// ```
```
killercup: Please add an empty line above code blocks, and below headlines. Bonus points for adding lang annotations to all code blocks, even though `rust` is the default (just my personal preference) :)

diesel/src/macros/macros_from_codegen.rs
```diff
@@ -72,6 +72,18 @@ macro_rules! infer_table_from_schema {
 ///
 /// This macro can only be used in combination with the `diesel_codegen` or
 /// `diesel_codegen_syntex` crates. It will not work on its own.
+///
+/// Example:
```
killercup: Common style is

    # Examples

Eijebong: Oh yeah, this looks way better :)

killercup:
🎉

PR #861, opened 2017-04-16
[bugfix][diesel_cli] parse dbname from connection string in a way that still leaves bugs

YetAnotherMinion:
I corrected to the parsing logic for database name in the connection string

When connecting to a new server, the database specified by the user
in the connection string may not exist. We instead make the first connection
to the default database which will always exist. We accomplish this by
editing the provided connection string to replace the database name with the
default database for the specific database backend.

MySQL and Postgres both have the same connection string format where the database
name appears after the third '/' character. In Postgres the database names continues
until a '?' or until the end of string. In MySQL the database name continues until
the end of the string. As a dirty hack, we parse all connection strings as if they
were Postgres connections, treating the dbname as any characters that occur between
the third '?' and the next '?'. This prevents us from using the '?' character
in the name of MySQL databases.

### DIFF
--- a/diesel_cli/src/database.rs
+++ b/diesel_cli/src/database.rs
@@ -1,4 +1,5 @@
 use clap::ArgMatches;
+use url::Url;
 use diesel::expression::sql;
 #[cfg(feature="postgres")]
 use diesel::pg::PgConnection;
@@ -285,61 +286,20 @@ pub fn database_url(matches: &ArgMatches) -> String {
 
 #[cfg(any(feature="postgres", feature="mysql"))]
 fn change_database_of_url(database_url: &str, default_database: &str) -> (String, String) {
-    // This method accepts a valid MySQL or Postgres connection string returns the
-    // name of the database and a new connection string to the database specified
-    // by the default_database argument. If the database name is not specified in
-    // the connection string then we default to an empty string. NOTE: This function
-    // does not validate connection strings (look at the signature, no Option or
-    // Result is there).
-    //
-    // Connection string formats
-    //         mysql://[user[:password]@]host/database_name
-    //    postgresql://[user[:password]@][netloc][:port][/dbname][?param1=value1&...]
-    //
-    // By inspection we see that database name always directly follows the third '/'.
-    // The database name is terminated either by the end of the string, or in the case
-    // with Postgres only, by the next occurance of '?'.
-    // We can sacrfice correctness by prohibiting MySQL database names containing a
-    // '?' character and use the same parsing rule for both formats.
+    // This method accepts a MySQL or Postgres connection string that conforms to URL BNF grammar
+    // only (and not the database connection string grammars) and returns the name of the database
+    // and a new connection string to the database specified by the default_database argument. If
+    // the database name is not specified in the connection string then we default to an empty
+    // string. NOTE: This function does not validate connection strings (look at the signature, no
+    // Option or Result is there).
 
-    let mut split: Vec<&str> = database_url.splitn(4, '/').collect();
-    // These are the only possible results
-    //   MySQL: ["mysql:", "", "[user[:password]@]host"]
-    //   MySQL: ["mysql:", "", "[user[:password]@]host", "database_name"]
-    //   Postgres: ["postgresql:", "", "[user[:password]@][netloc][:port]"]
-    //   Postgres: ["postgresql:", "", "[user[:password]@][netloc][:port]", "dbname[?param1=value1&...]"]
+    // TODO: handle unix connection strings with percent encoded paths such as:
+    //  postgres://%2Fvar%2Flib/dbname
+    let mut url = Url::parse(database_url).expect("failed to parse connection string");
+    let database: String = url.path().trim_left_matches('/').to_string();
+    url.set_path(default_database);
 
-    if split.len() == 4 {
-        // If there are four elements, then by definition the tail must contain the database
-        // name for both Postgres or MySQL. Now perform the hacky step of assuming that the
-        // first '?' character appearing in the tail terminates the database name.
-        let tail = split.pop().unwrap();
-        let mut tail_searcher = tail.splitn(2, '?');
-        // we will always return at least one element from the split iterator. This is the
-        // database name even if the database name is an empty string
-        let database = match tail_searcher.next() {
-            Some(dbname) => dbname,
-            None => panic!("String::splitn(2, '/') returned zero fragments"),
-        };
-        // If this is a postgres connection string we need to separate the parameters slug
-        // from the database name.
-        let new_url = match tail_searcher.next() {
-            // Now that we have destructured the connection string completely, we reassemble
-            // a connection string for the default_database.
-            Some(parameters) => format!("{}/{}?{}", split.join("/"), default_database, parameters),
-            None => format!("{}/{}", split.join("/"), default_database),
-        };
-        (database.to_owned(), new_url)
-    } else {
-        // There were only 3 elements so the user failed to provide a database name.
-        // We return an empty string for the database name so the caller can handle
-        // this logic error. We can still generate a valid connection string for the
-        // default_database though, so we append the default_database onto the end of
-        // the connection string
-        split.push(default_database);
-        ("".to_owned(), split.join("/"))
-    }
-
+    (database, url.into_string())
 }
 
 #[cfg_attr(feature="clippy", allow(needless_pass_by_value))]
@@ -355,7 +315,7 @@ mod tests {
     #[test]
     fn split_pg_connection_string_returns_postgres_url_and_database() {
         // format: (original_string, dbname, expected_changed_string)
-        let test_cases: [(&'static str, &'static str, &'static str); 11] = [
+        let test_cases: [(&'static str, &'static str, &'static str); 10] = [
             ("postgresql://", "", "postgresql:///postgres"),
             ("postgresql://localhost", "" ,"postgresql://localhost/postgres"),
             ("postgresql://localhost:5433", "", "postgresql://localhost:5433/postgres"),
@@ -366,7 +326,8 @@ mod tests {
             ("postgresql:///mydb?host=localhost&port=5433", "mydb", "postgresql:///postgres?host=localhost&port=5433"),
             ("postgresql://[2001:db8::1234]/database", "database", "postgresql://[2001:db8::1234]/postgres"),
             ("postgresql:///dbname?host=/var/lib/postgresql", "dbname", "postgresql:///postgres?host=/var/lib/postgresql"),
-            ("postgresql://%2Fvar%2Flib%2Fpostgresql/dbname", "dbname", "postgresql://%2Fvar%2Flib%2Fpostgresql/postgres"),
+            // TODO: Support opaque origins
+            //("postgresql://%2Fvar%2Flib%2Fpostgresql/dbname", "dbname", "postgresql://%2Fvar%2Flib%2Fpostgresql/postgres"),
         ];
         for &(database_url, database, postgres_url) in &test_cases {
             assert_eq!((database.to_owned(), postgres_url.to_owned()), change_database_of_url(database_url, "postgres"));
--- a/diesel_cli/src/database.rs
+++ b/diesel_cli/src/database.rs
@@ -285,10 +285,27 @@ pub fn database_url(matches: &ArgMatches) -> String {
 
 #[cfg(any(feature="postgres", feature="mysql"))]
 fn change_database_of_url(database_url: &str, default_database: &str) -> (String, String) {
-    let mut split: Vec<&str> = database_url.split('/').collect();
-    let database = split.pop().unwrap();
-    let new_url = format!("{}/{}", split.join("/"), default_database);
-    (database.to_owned(), new_url)
+    // the database name will always start with the 3rd forward slash and end with
+    // the end of the string (postgres only: or the next '?').
+
+    let mut split: Vec<&str> = database_url.splitn(4, '/').collect();
+    if split.len() == 4 {
+        let tail = split.pop().unwrap();
+        let mut tail_searcher = tail.splitn(2, '?');
+        let database = match tail_searcher.next() {
+            Some(dbname) => dbname,
+            None => panic!("String::splitn(2, '/') returned zero fragments"),
+        };
+        let new_url = match tail_searcher.next() {
+            Some(parameters) => format!("{}/{}?{}", split.join("/"), default_database, parameters),
+            None => format!("{}/{}", split.join("/"), default_database),
+        };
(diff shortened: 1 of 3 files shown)


diesel_cli/src/database.rs
```diff
@@ -303,19 +320,22 @@ mod tests {
...
+            , ("postgresql://localhost:5433", "", "postgresql://localhost:5433/postgres")
+            , ("postgresql://localhost/mydb", "mydb", "postgresql://localhost/postgres")
+            , ("postgresql://user@localhost", "","postgresql://user@localhost/postgres")
+            , ("postgresql://user:secret@localhost", "", "postgresql://user:secret@localhost/postgres")
+            , ("postgresql://other@localhost/otherdb?connect_timeout=10&application_name=myapp", "otherdb", "postgresql://other@localhost/postgres?connect_timeout=10&application_name=myapp" )
+            , ("postgresql:///mydb?host=localhost&port=5433", "mydb", "postgresql:///postgres?host=localhost&port=5433")
+            , ("postgresql://[2001:db8::1234]/database", "database", "postgresql://[2001:db8::1234]/postgres")
+            , ("postgresql:///dbname?host=/var/lib/postgresql", "dbname", "postgresql:///postgres?host=/var/lib/postgresql")
+            , ("postgresql://%2Fvar%2Flib%2Fpostgresql/dbname", "dbname", "postgresql://%2Fvar%2Flib%2Fpostgresql/postgres")
+            ];
```
killercup: I think that's the only place this array formatting is used in this code base. Can you change it to trailing commas and the opening bracket on the same line as the `=`?

diesel_cli/src/database.rs
```diff
@@ -303,19 +320,22 @@ mod tests {
...
+        for &(database_url, database, postgres_url) in test_cases.iter() {
```
killercup: s/test_cases.iter()/&test_cases

diesel_cli/src/database.rs
```diff
@@ -303,19 +320,22 @@ mod tests {
 
     #[test]
     fn split_pg_connection_string_returns_postgres_url_and_database() {
```
killercup: As this is also used for mysql connections, can you add a test with examples for that as well?

diesel_cli/src/database.rs
```diff
@@ -285,10 +285,27 @@ pub fn database_url(matches: &ArgMatches) -> String {
 
 #[cfg(any(feature="postgres", feature="mysql"))]
 fn change_database_of_url(database_url: &str, default_database: &str) -> (String, String) {
-    let mut split: Vec<&str> = database_url.split('/').collect();
-    let database = split.pop().unwrap();
-    let new_url = format!("{}/{}", split.join("/"), default_database);
-    (database.to_owned(), new_url)
+    // the database name will always start with the 3rd forward slash and end with
+    // the end of the string (postgres only: or the next '?').
```
killercup: Can you expand on the intention of this splitting a bit (connection strings using sockets instead of host name by using the query param form)? I'm not sure I can follow this when I see it again in a few month. Maybe also add the "template" version of a connection string, and explicitly mention the two forms, one using the user@host/db form, and the other using query params.

diesel_cli/src/database.rs
YetAnotherMinion: New unit test added. This function should every valid Postgres connection string. On the other side of the fence, not all valid MySQL connection strings are supported. I could not find official documentation of MySQL connection string grammar (found the JDBC and ODBC, but not the c library), so I just used the grammar documented in **diesel** itself (http://docs.diesel.rs/diesel/mysql/struct.MysqlConnection.html).

sgrif:
Can we use the URL crate here instead?

YetAnotherMinion:
@sgrif Good idea. I tried it and found the URL crate is unable to parse the postgres percent encoded unix domain socket format (`postgresql://%2Fvar%2Flib%2Fpostgresql/dbname`), which is the one connection string format I actually need. Other than that the code is really short. 


When you go to handling the error case, it looks like you are back to manually implementing regex.
```rust
fn change_database_of_url(database_url: &str, default_database: &str) -> (String, String) {
    let mut connection_string = match Url::parse(database_url) {                
        Ok(string) => string,                                                   
        Err(_) => {println!("{}", database_url); panic!()},                     
    };                                                                          
    let database: String = connection_string.path().trim_left_matches('/').to_string();
    connection_string.set_path(default_database);                               
                                                                                
    (database, connection_string.into_string())
}
```

YetAnotherMinion:
Note that using URL crate will not solve the problem of being unable to use question mark character in the name of MySQL databases.

jaemk:
Since you're doing the work of parsing the connection strings, maybe it would be worthwhile to move it to a separate function and actually pull out all the info. Going off of the following cases:
```
    // These are the only possible results
    //   MySQL: ["mysql:", "", "[user[:password]@]host"]
    //   MySQL: ["mysql:", "", "[user[:password]@]host", "database_name"]
    //   Postgres: ["postgresql:", "", "[user[:password]@][netloc][:port]"]
    //   Postgres: ["postgresql:", "", "[user[:password]@][netloc][:port]", "dbname[?param1=value1&...]"]
```
you could pull everything out with a regex. This would also allow parsing mysql database names that have `?` and passwords with non `a-z` characters (see issue #871 ). See [playpen](http://play.integer32.com/?gist=55011070b844b87d9296c2fd7af5cf94&version=undefined)

YetAnotherMinion:
@jaemk provided regex panics on `r"postgresql://user@host:3200/dbname?param1=value1&p2=v2"`

I think that before any more effort to implement connection string parsing is expended, we need to determine exactly what the grammar for Postgres and MySQL connection strings are. We have tried the route of "string split here and there" or run a quick regex which satisfies 95% of users but now the long tail has dragged itself out of the woods.

Aka: We need the specification for connection strings beyond just the cutesy "these are the parts of the connection string" provided by database online documentation. What characters are allowed in each part? Are there escapes? Once we have a complete specification we can proceed with firm footing. Otherwise this piece of code will be rewritten half a dozen times over coming years as each new technically correct connection strings slams into the "works for me" code merged by the previous voyager.

YetAnotherMinion:
after reading the libpq source I can speak authoritatively that for Postgres that

- the user cannot contain the characters '@' and '/' but can contain the ':' character
- the password cannot contain the '@' or ':' character but can contain the '/' character
- the netloc cannot contain the ':', '/', '?', or ',' characters
- the dbname cannot contain the '?' character

Given the context dependent requirements for parsing, I doubt that a readable regex can be constructed to parse this string. It would be simpler to wire up rust C-bindings to the already functional postgres C code where the automata has been constructed by hand over several hundred lines.

jaemk:
@YetAnotherMinion If the full connection string spec needs to be met, then yes it may be beneficial to write up some C-bindings. On the otherhand, if connection strings are restricted to a reasonable subset, I think a regex would work fine. Here's an updated version for optional passwords [playpen](http://play.integer32.com/?gist=d478eb0ca3a1b994888c508f80ef1a4e&version=undefined). The tricky part would be the `:` being allowed in the username.

YetAnotherMinion:
@jaemk Obviously it is possible by definition to write a regex because that C parser is a finite automata. However there are two problems I forsee with regex based solution
1. Readability. Consider the email specification -- looks like an even simpler grammar than connection strings -- which has an ungodly long regex.
2. The full connection spec needs to be met. This simple pull request (which greatly expands subset of connection strings that will now be parsed correctly) has brought every #!@$ nitpicker out of the woodwork complaining how it is not perfect. It is apparent to me that only a perfectly standards compliant implementation will be allowed to merge.

I just want the correct functionality for non standard directory unix sockets to be available from upstream so I can stop screwing around with symlinks. The whole point of using a non standard directory was so I did not have to learn where each different distro's postgres package considers "default" .

The code sitting in front of you today satisfies:
>  if connection strings are restricted to a reasonable subset,

and does not pull in an extra dependency on the regex crate for the extremely simple use case of _change the database name_. That's it. The entire business requirement is only to parse the database name, nothing else.

sgrif:
> This simple pull request (which greatly expands subset of connection strings that will now be parsed correctly) has brought every nitpicker out of the woodwork complaining how it is not perfect

Please be polite. There is no need to be offended at one additional person reviewing the code. I've been traveling since I left my previous comment, and haven't had a chance to look at the problem more in depth.

sgrif:
> we need to determine exactly what the grammar for Postgres and MySQL connection strings are

Ok let's try to lay this out as explicitly as possible then. A valid connection string that we accept for MySQL or PostgreSQL is a subset of URLs which meet the following requirements:

- The scheme MUST be one of the expected schemes for the adapter (`postgres://` or `postgresql://` for PG, `mysql://` for MySQL).
- The path MUST have exactly 0 or 1 segments
- The fragment MUST NOT be present
- The query MUST either not be present, or be represented as `application/x-www-form-urlencoded`, where all keys given are additional options expected by that backend

This means that the definition of how to extract the database name should authoritatively be:

- Parse the URL
- Validate the scheme
- If the path length is greater than one, error
- If the path length is zero, return the default database as determined by the scheme
- Otherwise, return the result of percent_decoding the single path segment

> I submit that the database C driver code is the final authority.

Since this applies to drivers other than libpq (and even our PG driver won't use libpq forever), I'd prefer to keep our specification separate. That said, I suspect that they will align more often than not.

> Note that using URL crate will not solve the problem of being unable to use question mark character in the name of MySQL databases.

We should be respecting percent encoded question-marks. That is a separate bug, which should have a separate issue opened.

> The full connection spec needs to be met.

No, we can fix one issue at a time, and open issues for each bug as they come in. My main concern is implementing our own pseudo-url-parser when there's already a crate out there for this.

>  I tried it and found the URL crate is unable to parse the postgres percent encoded unix domain socket format (postgresql://%2Fvar%2Flib%2Fpostgresql/dbname)

As best I can tell, that's a bug in the URL crate. Based on the URL spec, that host should be valid since `postgresql` is not a "special" host, meaning that the host should be parsed as an [opaque host string](https://url.spec.whatwg.org/#concept-opaque-host-parser), which does not percent decode its input. I'd open an issue against the URL crate. As a workaround, you can use the `host` query param instead.

YetAnotherMinion:
@sgrif 
> Since this applies to drivers other than libpq (and even our PG driver won't use libpq forever), I'd prefer to keep our specification separate. That said, I suspect that they will align more often than not.

I do not understand your reasoning for deciding to introduce a new [competing universal standard](https://xkcd.com/927/) here, could you elaborate?

I understand that it is very tempting to change the specification to cover the case of shoving a potentially square peg into the off the shelf round hole by just banning square pegs. 
> No, we can fix one issue at a time, and open issues for each bug as they come in.

This is exactly what I do at work because my boss's incentive compensation is tied to average age of tickets. That line of thinking is what we all do 5 days a week at work, and it is exactly that approach that drive me nuts.  It is the weekend, it is my time, let cost-benefit of complete resolution be damned. I -- personally -- would strongly prefer to implement **The Right Thing That Completely Addresses The Underlying Problem (tm)**.

I understand the aversion to writing lots of code. The best code is the lines you never write. However there is no business pressure to be quick right now. No manager pressuring you to just hack something together so you can close the ticket before lunch (to avoid that extra 30 minutes of time to close). We can take as long as we want to implement a robust, simple, and sound solution that completely solves the issue. No more connection string related issues is a glorious future that I can look forward to. 



-------------------------------
If the consensus is to use the URL crate then I can go to the URL repository, spend time to understand their code, submit PR, wait weeks to get it merged, wait more weeks to get a point release, come back here and introduce the fixed URL crate.

At this point I don't care about turnaround. My original hope was that I could fix the problem I encountered that Friday at work on my own time over the weekend so that crates.io would have the fixed version when I came into Monday. I was encouraged by the active Gitter channel, that I got a response to my questions, and even PR comments the next day. But then nothing happened and I had to go into work that Monday and spend time inserting the patched version of diesel into build farm and development images. That is just another sunk cost now.

The reason I am still here and willing to work on this is because I care about the elegant and correct. It would please me to be part of the "it just works" foundation of the world, where anyone in the future can switch to diesel, use their existing connection strings, and it always works. No chasing rabbit holes where you think the database was set up incorrectly (because coincidentally you just upgraded the OS for the developer image).

My argument against your proposal to require a URL grammar for connection strings is that it stands against the principle of least surprise. The least surprise for me would be to take an existing connection string from a working application, add it to diesel, and it always works. I don't think introducing a new proprietary format is the best tradeoff to make; against having a free internet volunteer spend an hour porting some while loops from C to Rust.

YetAnotherMinion:
Appveyor cannot reach crates.io
```
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: failed to load source for a dependency on `time`
Caused by:
  Unable to update registry https://github.com/rust-lang/crates.io-index
To learn more, run the command again with --verbose.
```

YetAnotherMinion:
After examining URL RFC and the URL crate, I do not thing that getting percent encoding for the host will be free. out of 

```

url            = httpurl | ftpurl | newsurl |
                 nntpurl | telneturl | gopherurl |
                 waisurl | mailtourl | fileurl |
                 prosperourl | otherurl
```

You are correct that the database connection strings are defined within the URL RFC
```
3.1. Common Internet Scheme Syntax. How
//<user>:<password>@<host>:<port>/<url-path>

   Some or all of the parts "<user>:<password>@", ":<password>",
   ":<port>", and "/<url-path>" may be excluded.  The scheme specific
   data start with a double slash "//" to indicate that it complies with
   the common Internet scheme syntax. The different components obey the
   following rules:

    user
        An optional user name. Some schemes (e.g., ftp) allow the
        specification of a user name.

    password
        An optional password. If present, it follows the user
        name separated from it by a colon.

   The user name (and password), if present, are followed by a
   commercial at-sign "@". Within the user and password field, any ":",
   "@", or "/" must be encoded.
```
However it looks like the URL crate only implements a subset of the grammar, specifically the subset useful for a browser. They do not have a generic parser for Common Internet Scheme Syntax. For web urls (http, https, ftp, gopher), percent encoding is not permitted in the host part. To use the URL crate it looks like you would need to introduce a new abstraction for  Common Internet Scheme Syntax and I am not sure how well that would be received as it is pretty orthogonal to the purpose of Servo's needs.

sgrif:
I'm going to go ahead and close this. Nobody on the team wants to deal with the rudeness on this PR, and it has merge conflicts that will require additional work from someone. If anyone wants to take this over and work on this without the attitude, feel free as there are still bugs to be addressed.

YetAnotherMinion:
@sgrif  That is really unfortunate given that I gave you every change you asked for. You wanted to use the URL crate and I put that into the PR, you wanted to not handle every case, I put a panic! in for the todo.

I registered my **strong** objections to your design choices but in the patch I submitted over a month ago I put in exactly what you asked for and nothing more.

PR #862, opened 2017-04-17
Add a simple example for raw sql requests.

Eijebong:


diesel/src/expression/sql_literal.rs
```diff
@@ -10,6 +10,43 @@ use types::HasSqlType;
...
+/// # fn main() {
+/// #     let connection = establish_connection();
+/// let query = sql::<Bool>("INSERT INTO users(name) VALUES('Ruby')");
+/// query.execute(&connection).expect("Can't insert in users");
+///
+/// let query = sql::<(Integer, Text)>("SELECT id, name FROM users WHERE name='Ruby';");
+/// let users = query.load::<User>(&connection).expect("Can't query users");
+/// assert_eq!(users, vec![User{id: 3, name: "Ruby".to_owned()}]);
+///
+/// let query = sql::<(Integer, Text)>("SELECT id, name FROM users WHERE name='Ruby';");
```
killercup: Did you duplicate this line on purpose to show it's the same query? (Maybe add a comment like "Same query as above")

Eijebong: Yeah, I was not really inspired, but wanted to show that you can use sql literals as filters because I didn't know that myself before you told me ^^

PR #863, opened 2017-04-17
Add a mention to the foreign_key argument of has_many and belongs_to.

Eijebong:
I'm not really happy with how the sentence turned out but at least now it's documented ^^

### DIFF
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -43,6 +43,10 @@
 //!
 //! [Identifiable]: trait.Identifiable.html
 //!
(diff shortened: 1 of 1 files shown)


diesel/src/associations/mod.rs
```diff
@@ -43,6 +43,10 @@
 //!
 //! [Identifiable]: trait.Identifiable.html
 //!
+//! If your foreign key doesn't follow the convention `tablename_id`, you can specify a custom
```
killercup: Maybe "If [the name of] your"?

Eijebong: 👍

PR #866, opened 2017-04-20
Allow dead code in embedded migrations

### LINKED ISSUE #865: Embedded migrations: Dead code warning
As mentioned here https://github.com/diesel-rs/diesel/pull/854#issuecomment-293825480, using embedded migrations with `run_with_output` results in a dead code warning.

```
warning: function is never used: `run`, #[warn(dead_code)] on by default
  --> src/db/mod.rs:62:1
   |
62 | embed_migrations!();
   | ^^^^^^^^^^^^^^^^^^^^
   |
```

Would it be possible to add an exception for that? I didn't easily find out where to add it.

CC @Eijebong @weiznich

dbrgn:
This gets rid of a warning when using `run_with_output`.

Fixes #865

### DIFF
--- a/diesel/src/macros/macros_from_codegen.rs
+++ b/diesel/src/macros/macros_from_codegen.rs
@@ -106,6 +106,7 @@ macro_rules! infer_table_from_schema {
 /// ```
 macro_rules! embed_migrations {
     () => {
+        #[allow(dead_code)]
         mod embedded_migrations {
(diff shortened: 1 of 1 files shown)


killercup:
Thanks!

I'm merging this right now without a lot of thinking about other possible lints that we could enable because this prevents most-likely-unnecessary warnings without the user activating any lints.

PR #867, opened 2017-04-20
Loosen the serde dependency to allow using serde 1.0 with diesel.

Eijebong:
I'm not sure this is the right way to do this ?

Maybe it should be `>=0.8.0, < 1.1` but minor versions shouldn't be breaking changes so...


killercup:
`<2` should be good. If anything breaks, we'll just remind they serde developers ;)

PR #873, opened 2017-04-26
Update clippy to 0.0.126

DorianListens:
This change updates clippy, and updates the travis nightly version to
2017-04-25.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-  - rust: nightly-2017-04-19
+  - rust: nightly-2017-04-25
     env: CLIPPY=YESPLEASE
     script:
     - (cd diesel && cargo rustc --no-default-features --features "lint unstable chrono serde_json uuid sqlite postgres mysql" -- -Zno-trans)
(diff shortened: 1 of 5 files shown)


killercup:
There are literally no feature changes in clippy between 0.0.125 and 0.0.126. Did anything cool land in nightly since April 19?

You can always `rustup overwrite add nightly-2017-04-19` in your diesel directory.

DorianListens:
Sure, I could, but we'd need to make these changes eventually, wouldn't we? Is there a reason to lag behind nightly?

killercup:
Laziness :) I'm fine with merging this as you already put the work in
basically but it's more fun to update clippy when there are new lints that
help us :)

Dorian Scheidt <notifications@github.com> schrieb am Mi. 26. Apr. 2017 um
18:36:

sgrif:
In general I'd rather not chase the latest versions on these unless we have a specific reason to update, or there's been a Rust release.

DorianListens:
Entirely reasonable. I was operating under the principle that "rustup update && bin/test" should "just work", but I'm happy to be a little less enthusiastic about nightly releases and instead pin myself to the version in .travis.yml.

Eijebong:
Plus clippy is always behind nightly sometimes by two or three days.

sgrif:
Either way this PR more or less coincided with a Rust release, so it's fine

PR #875, opened 2017-04-30
Refactor join methods into a trait

sgrif:
This change is part of an ongoing refactoring to enable multi-table
joins, explicit on clauses, and other improvements. Unfortunately, this
completely ruins the error message when attempting to join between two
tables that know nothing about each other. The reason is that the
missing link is `Join<Left, Right, Kind>: QuerySource`, which is missing
*both* `Left: JoinTo<Right>` and `(Left::AllColumns, Right::AllColumns):
SelectableExpression<Self>`. Ironically, this missing link in the second
case is actually `Left: JoinTo<Right>`, but that has to go through so
many different impls I wouldn't expect Rust to be able to unify them.

Anyway, I'm planning on eventually applying the `on` segment earlier,
which should result in a missing `Left: JoinTo<Right>` constraint much
sooner and will hopefully bring back that error message. Honestly, I may
even be over thinking the error message here since in reality a missing
`JoinTo` constraint isn't conveying more information than
`InternalJoinDsl` since neither mention the associations API

I want to keep `Join<Left, Right, Kind>` in the public API for
`SelectableExpression` constraints, but it no longer needs to implement
`AsQuery` since you can no longer obtain an instance of it.

### DIFF
--- a/diesel/src/query_builder/select_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_statement/dsl_impls.rs
@@ -13,8 +13,29 @@ use query_builder::{AsQuery, Query, QueryFragment, SelectStatement};
 use query_dsl::*;
 use query_dsl::boxed_dsl::InternalBoxedDsl;
 use query_source::QuerySource;
+use query_source::joins::Join;
 use super::BoxedSelectStatement;
 use types::{self, Bool};
+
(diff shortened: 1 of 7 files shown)


sgrif:
Follow up PR will bring back the errors we want

PR #879, opened 2017-04-30
Refactor join construction to be much more composeable

sgrif:
This moves most of the knowledge of join construction out of `JoinTo`
and into the join methods themselves. Aside from exposing less of our
internals, this is useful to us because it's much easier to implement
`JoinTo` on joins themselves than it is to implement the specific join
methods.

The downside to this approach is that it gives us less specific control
over certain types, and means that
`users.inner_join(posts).inner_join(settings)` will likely have the type
`(User, (Post, Setting))` instead of `(User, Post, Setting)` which will
make it unfortunately difficult to differentiate
`user.inner_join(posts).inner_join(comments)` from
`user.inner_join(posts.inner_join(comments))`.

Still, I think this is the path that will get us to *some* form of
multi-table joins, and we can always refine later. This structure should
make it fairly simple to add all of:

- Joining to subselects
- Cross joins
- Join with explicit on-clauses
- Self-referential joins (though not through the associations API)

Additionally, the only real blocker remaining for multi-table joins is
implementing `SelectableExpression` which should hopefully be unblocked
by Rust soon.

### DIFF
--- a/diesel/src/query_source/joins.rs
+++ b/diesel/src/query_source/joins.rs
@@ -13,6 +13,16 @@ pub struct Join<Left, Right, Kind> {
     kind: Kind,
 }
 
+#[derive(Debug, Clone, Copy)]
+#[doc(hidden)]
+/// A query source representing the join between two tables with an explicit
+/// `ON` given. `Join` should usually be referenced instead, as all "type
+/// safety" traits are implemented in terms of `Join` implementing them.
+pub struct JoinOn<Join, On> {
+    join: Join,
+    on: On,
+}
+
 impl<Left, Right, Kind> Join<Left, Right, Kind> {
     pub fn new(left: Left, right: Right, kind: Kind) -> Self {
         Join {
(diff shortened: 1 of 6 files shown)


sgrif:
Ping for review -- AppVeyor failure is unrelated

diesel/src/query_source/joins.rs
```diff
@@ -21,22 +31,30 @@ impl<Left, Right, Kind> Join<Left, Right, Kind> {
             kind: kind,
         }
     }
+
+    #[doc(hidden)]
+    pub fn on<On>(self, on: On) -> JoinOn<Self, On> {
+        JoinOn {
+            join: self,
+            on: on,
```
killercup: Just wondering: On has no constraints, but it will probably always need to be `On: Expression`. Could adding this here (or in `fn join`) give better/earlier errors?

sgrif:
> I also fear that having joins build up from Join/JoinOn/InternalJoinStuff will make it harder to debug.

I agree. The goal is to have `Join` be the canonical thing you target if you need to care about more than one table (in particular, `Join<T, excluded, Cross>` is a thing I want to target to fix the issue with `excluded` being allowed in wrong places). I *think* we mostly have that with this PR, but overlapping marker impls will make it much easier to be sure that's the only impl you need to care about.

diesel/src/query_source/joins.rs
sgrif: I've been shying away from adding those constraints earlier than needed unless there's a concrete case where it improves errors. I'm not sure that it would do so here.

diesel/src/query_source/joins.rs
```diff
@@ -45,39 +63,103 @@ impl<Left, Right> QuerySource for Join<Left, Right, Inner> where
...
+    Left::FromClause: QueryFragment<DB>,
+    Right: QuerySource,
+    Right::FromClause: QueryFragment<DB>,
+    Kind: QueryFragment<DB>,
+{
+    fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
+        try!(self.left.from_clause().to_sql(out));
+        try!(self.kind.to_sql(out));
+        out.push_sql(" JOIN ");
+        try!(self.right.from_clause().to_sql(out));
```
sgrif: Merge conflicts

sgrif: merge conflicts...

Issue #795, opened 2017-03-11
Appveyor has intermittent failures


sgrif:
Occasionally our builds on appveyor fail before they can even download Rust.

> appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
> Error downloading remote file: One or more errors occurred.
> Inner Exception: The request was aborted: Could not create SSL/TLS secure channel.

Would be great to figure out a way to fix this, since those builds are slow enough that "just restarting it when it happens" is pretty painful.

cyplo:
Would caching the binary in appveyor's cache help ?

PR #883, opened 2017-05-02
Switch to curl for downloading rustup on appveyor as it allows retries

markcatley:


markcatley:
Attempting to fix #795. Not 100% sure this will completely solve the problem, but it has it's a good shot at it.

killercup:
Seems better than before to me, but I have no idea what `appveyor DownloadFile` does under the hood. On the other hand, [other repos](https://github.com/alexcrichton/futures-rs/blob/8cf6291f260a3b2ef2a59980285359e0ddc571dd/appveyor.yml) use curl here as well, so it can't be that bad :D

I'll merge this. Let's see if this reduces some of the spurious failures.

PR #888, opened 2017-05-07
We are returning vectors here

killercup:
Is this an artifact of a diesel that once returned actual iterators here?

[ci skip]


Eijebong:
24c48b1d83c8970a9e1ad6d68987e060911d99ee Yup

killercup:
Ah, yeah, that was the change I was thinking of. I'll merge this then!

PR #891, opened 2017-05-08
 Add ilike method for postgres

### LINKED ISSUE #784: Add ilike method for postgres
Add ilike method for PostgreSql db.

Eijebong:
It's mostly a copy/paste job from like and not_like.
I just had to export the gen_helper_type! macro to use it in the pg
module.

Fixes #784

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Added
 
+* Added support for `ILIKE` in PostgreSQL.
+
 * Added the `migration list` command to Diesel CLI for listing all available migrations and marking those that have been applied.
 
 * Added support for adding two nullable columns.
--- a/diesel/src/pg/expression/predicates.rs
+++ b/diesel/src/pg/expression/predicates.rs
@@ -5,5 +5,7 @@ infix_predicate!(IsNotDistinctFrom, " IS NOT DISTINCT FROM ", backend: Pg);
 infix_predicate!(OverlapsWith, " && ", backend: Pg);
 infix_predicate!(Contains, " @> ", backend: Pg);
 infix_predicate!(IsContainedBy, " <@ ", backend: Pg);
+infix_predicate!(ILike, " ILIKE ", backend: Pg);
+infix_predicate!(NotILike, " NOT ILIKE ", backend: Pg);
 postfix_expression!(NullsFirst, " NULLS FIRST", ());
 postfix_expression!(NullsLast, " NULLS LAST", ());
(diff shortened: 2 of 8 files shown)


diesel/src/expression/helper_types.rs
```diff
@@ -8,6 +8,8 @@ pub type SqlTypeOf<Expr> = <Expr as Expression>::SqlType;
 pub type AsExpr<Item, TargetExpr> = AsExprOf<Item, SqlTypeOf<TargetExpr>>;
 pub type AsExprOf<Item, Type> = <Item as AsExpression<Type>>::Expression;
 
+#[macro_export]
```
sgrif: This name is too generic for us to export. You shouldn't need to export it to use it from another module.

Eijebong: Right, I had some troubles using the macro outside and the `#[macro_export]` stayed

sgrif: Probably just due to module ordering. I'd just move the definitions here instead of the PG module and add another cfg attr

diesel/src/pg/expression/predicates.rs
```diff
@@ -5,5 +5,7 @@ infix_predicate!(IsNotDistinctFrom, " IS NOT DISTINCT FROM ", backend: Pg);
 infix_predicate!(OverlapsWith, " && ", backend: Pg);
 infix_predicate!(Contains, " @> ", backend: Pg);
 infix_predicate!(IsContainedBy, " <@ ", backend: Pg);
+infix_predicate!(ILike, " ILIKE ");
```
sgrif: These should be specifying the backend, shouldn't they?

sgrif:
This needs a compile fail test to assert that using it with SQLite fails.

diesel/src/pg/expression/predicates.rs
Eijebong: This is in `pg/expression/predicate.rs`

diesel/src/expression/helper_types.rs
Eijebong: It was just missing the `#[macro_use]`

Do you want me to move the pg/expression/helper_types.rs in there ?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Added
 
+* Added support for `ILIKE` in postgreSQL.
```
killercup: PostgreSQL
    ^

Eijebong:
There was no doctest for like and not_like hence why there isn't any for ilike and not_ilike. I can add some though if you want ^^

killercup:
Thanks. Let's add docs for (i)like in another PR and merge this as-is.

PR #892, opened 2017-05-09
Give `Debug` a real bind collector

sgrif:
If we want to include the bind parameters in the debug output, we may as
well have this thing use the same implementation as everything else. We
already require a `ToSql<Debug>` impl on everything, so this doesn't
make anything *more* restricted


sgrif:
Yeah this is in preparation for logging. It's slower than the no-op but it should get optimized away if it ever mattered. Either way I'm looking at refactoring `QueryFragment` to use more traditional AST visitors, which will eliminate the problem.

PR #893, opened 2017-05-09
Move the responsibility of serializing bind params

sgrif:
This is in support of a bunch of different ongoing refactorings:

- I'd like to start logging the bind parameters (both when debugging and
  when logging), meaning we'll need access to the full value much later
  in the process.
- I'd like to change `QueryFragment` to be a more traditional AST
  visitor structure, so this method will be merged with
  `QueryBuilder::push_bind_param`. We need to not do the actual
  serialization work for other AST passes
- I'd like to rewrite SQLite's `ToSql` implementations to not do the
  serialize/deserialize hack we have now, so we need to be able to be
  polymorphic higher up

### DIFF
--- a/diesel/src/expression/bound.rs
+++ b/diesel/src/expression/bound.rs
@@ -2,10 +2,9 @@ use std::marker::PhantomData;
 
 use backend::Backend;
 use query_builder::*;
-use result::Error::SerializationError;
 use result::QueryResult;
 use super::*;
-use types::{HasSqlType, ToSql, IsNull};
+use types::{HasSqlType, ToSql};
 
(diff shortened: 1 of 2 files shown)


Eijebong:
Ok github, next time tell me somebody already reviewed it and said the exact same thing please !

killercup:
classy-octocat

sgrif:
It's a breaking change, but one that would only affect third party connection adapters, and I'm not aware of any that exist at the moment. I intend for these traits to be public API, but while we're still 0.x I don't think changes to the query builder traits warrant mentions in the changelog or anything.

PR #894, opened 2017-05-09
Refactor the query builder to make it easier to add new AST passes

sgrif:
I had initially wanted to move us towards a more of a visitor pattern,
where the `AstVisitor` trait would have most of the methods from
`QueryBuilder`, with `BindCollector` and `fn
unsafe_to_cache_prepared(&mut self)` tacked on. However, I quickly hit a
wall there when I realized that `BindCollector` cannot be made object
safe, which would prevent me from making `QueryFragment` object safe, so
that won't work.

The benefit of using the visitor pattern here would have been that it
would be easier for code *outside* of Diesel to add new AST passes if it
wanted to (within limits, I don't want to go so far as to have
`visit_specific_type` as that would be even more painful than what we
have now). This form doesn't bring that benefit.

However, using an enum instead does make it easier for us to add new
passes within Diesel itself, since we now have a single function which
will apply to all future AST passes for the majority of nodes (most
nodes don't care about any pass other than `to_sql`, they just call
their children).

The plan is to follow this up by moving `to_sql` into the `walk_ast`
method as well, and eliminate `Changeset`, and `InsertValues` since
it'll now be easier to add the new "this is like that other pass, but
slightly different for this one case" type passes.

The signature is a little funky here, since we're passing a mutable
reference to an enum which just wraps a mutable reference. I really only
want the inner `&mut`, since I want implementors to be able to mutate
the `QueryBuilder` or `BindCollector`, but not change the variant. The
reason for the outer `&mut` is because reborrowing doesn't apply to
structs. We can't pass `AstPass<'a>` multiple times, as it's not `Copy`.
When we pass a mutable reference directly, Rust creates a new reference
with a narrowed lifetime. We can't do that here. The reason I kept the
inner `&mut` instead of just having `AstPass` own the value is that we'd
have no way to cleanly get the value being mutated back out of it when
we were done.

### DIFF
--- a/diesel/src/expression/functions/date_and_time.rs
+++ b/diesel/src/expression/functions/date_and_time.rs
@@ -23,12 +23,8 @@ impl<DB: Backend> QueryFragment<DB> for now {
         Ok(())
     }
 
-    fn collect_binds(&self, _out: &mut DB::BindCollector) -> QueryResult<()> {
+    fn walk_ast(&self, _: &mut AstPass<DB>) -> QueryResult<()> {
         Ok(())
-    }
-
-    fn is_safe_to_cache_prepared(&self) -> bool {
-        true
     }
 }
 
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -73,6 +73,16 @@ impl<'a, T: Query> Query for &'a T {
     type SqlType = T::SqlType;
 }
 
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub enum AstPass<'a, DB> where
+    DB: Backend,
+    DB::BindCollector: 'a,
+{
+    CollectBinds(&'a mut DB::BindCollector),
+    IsSafeToCachePrepared(&'a mut bool),
+}
+
 /// An untyped fragment of SQL. This may be a complete SQL command (such as
 /// an update statement without a `RETURNING` clause), or a subsection (such as
 /// our internal types used to represent a `WHERE` clause). All methods on
@@ -80,8 +90,17 @@ impl<'a, T: Query> Query for &'a T {
 /// trait to be implemented.
 pub trait QueryFragment<DB: Backend> {
     fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult;
-    fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()>;
-    fn is_safe_to_cache_prepared(&self) -> bool;
+    fn walk_ast(&self, pass: &mut AstPass<DB>) -> QueryResult<()>;
+
+    fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()> {
+        self.walk_ast(&mut AstPass::CollectBinds(out))
+    }
+
+    fn is_safe_to_cache_prepared(&self) -> QueryResult<bool> {
+        let mut result = true;
+        self.walk_ast(&mut AstPass::IsSafeToCachePrepared(&mut result))?;
+        Ok(result)
+    }
 }
 
 impl<T: ?Sized, DB> QueryFragment<DB> for Box<T> where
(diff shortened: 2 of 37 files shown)


diesel/src/expression/functions/date_and_time.rs
```diff
@@ -23,13 +23,9 @@ impl<DB: Backend> QueryFragment<DB> for now {
         Ok(())
     }
 
-    fn collect_binds(&self, _out: &mut DB::BindCollector) -> QueryResult<()> {
+    fn walk_ast(&self, _pass: &mut AstPass<DB>) -> QueryResult<()> {
```
Eijebong: Why use `_pass` here and `_` above ?

diesel/src/query_builder/mod.rs
```diff
@@ -73,15 +73,37 @@ impl<'a, T: Query> Query for &'a T {
...
+    fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()> {
+        self.walk_ast(&mut AstPass::CollectBinds(out))
+    }
+
+    fn is_safe_to_cache_prepared(&self) -> bool {
+        let mut result = true;
+        match self.walk_ast(&mut AstPass::IsSafeToCachePrepared(&mut result)) {
+            Ok(()) => result,
+            // This should never fail, but let's avoid panicking if we somehow do
+            Err(_) => false,
```
Eijebong: Maybe we should still do something if it fails ? If it fails, it's probably a bug and it should be fixed.

diesel/src/query_builder/select_statement/mod.rs
```diff
@@ -176,26 +165,16 @@ impl<S, D, W, O, L, Of, G, DB> QueryFragment<DB>
...
 
-    fn collect_binds(&self, out: &mut DB::BindCollector) -> QueryResult<()> {
-        try!(self.distinct.collect_binds(out));
-        try!(self.select.collect_binds(&(), out));
-        try!(self.where_clause.collect_binds(out));
-        try!(self.group_by.collect_binds(out));
-        try!(self.order.collect_binds(out));
-        try!(self.limit.collect_binds(out));
-        try!(self.offset.collect_binds(out));
+    fn walk_ast(&self, pass: &mut AstPass<DB>) -> QueryResult<()> {
```
Eijebong: What happens here for the is_safe_to_cache_prepared pass ?
If distinct is false, and then select is true ?

Ultimately, only the last one means something right ?
Maybe you should use `**result &= true/false` everywhere ?

sgrif: `&& true` is a no-op in all cases. We are only assigning when the value is `false` (which is the only time the value could change)

Eijebong: Oh, yeah, I didn't notice that was the case ^^' Then it works

diesel/src/expression/functions/date_and_time.rs
sgrif: Because I wrote the first third of this, stashed it, wrote three other prototypes which all turned out to be dead ends, rewrote a sixth of this, remembered that I had a stashed partial impl of this version, and had merge conflicts because I switched styles in the second one. I'll fix it lol

diesel/src/query_builder/mod.rs
sgrif: What do you suggest? The only other options I could come up with are printing to stderr or panicking. The first is untenable, and I've become more hesitant towards the second. Ultimately in this particular case the worst that could happen is we skip prepared statement caching, which is why I went this path. Interested to hear your thoughts.

Eijebong: I understand that this is not the end of the world if we hit that false but as you said it should never happen. The problem is that if it ever happens, it'll go unnoticed for a long time (forever is still a long time !).
Printing "this is a diesel bug, please report" to stderr might be a good solution.

PR #895, opened 2017-05-09
Document the common abbreviations in `select_statement`

sgrif:
I should do this for `DB`, `ST`, and `QS` as well, but it's less clear
where to put those. These docs are not rendered on the site, and are
purely to help ease in new contributors.


killercup:
Nice. I'd put stuff like this in `Readme` files somewhere in `src/`, but this is good as well.

PR #868, opened 2017-04-20
Add support for custom migration_dir to disel migration revert and redo

Eijebong:
This fixes a part of #857 
The rest will follow once this one has been merged :)

Issue #150, opened 2016-01-30
Let's figure out logging


sgrif:
As I'm adding SQLite support, I'm finding it annoying to track down cases like ```called `Result::unwrap()` on an `Err` value: DatabaseError("SQL logic error or missing database")```.

This should definitely be injectable, and I'd prefer that it be decoupled from the individual connection classes, but I'm not sure that's actually possible.

I'd originally imagined something like `LoggingConnection` being a struct, which wrapped another `T: Connection`, and ran everything through the `DebugQueryBuilder` (and eventually cached it through the same mechanisms that we add for prepared statement caching). This should also log the values for the bind params. I am fine with adding the constraint `ToSql<ST, DB>: Debug` to accommodate this, as basically everything should implement it.

However, given that `DebugQueryBuilder` can have output which potentially differs from the actual SQL executed (for example, we're currently figuring out how to deal with SQLite's lack of a `DEFAULT` keyword, and I don't see how what we end up doing can actually be reflected in `DebugQueryBuilder`). So perhaps the correct answer is to have a `set_logger` method added to `Connection`.

Ultimately I'm undecided on this, and am open to either solution, if presented as a working PR. Discussion about how to go about this, or use cases that we think we need to support is certainly welcomed.

sgrif:
I'm going to pull this off the 1.0 milestone. I don't see any way for this to happen in the immediate future. The ecosystem isn't there yet. From my point of view there are two main requirements here:

- Doesn't require the use of a global logger.
  - Doing something as simple as "log crate1 to stdout, crate2 to a file" or "temporarily disable logging for a single crate" should not be difficult.
  - It would be impossible to test our implementation of logging if a global logger is required.
- Has a reasonably simple "getting started" story. e.g. "I just want to log to STDOUT or a file and don't care about anything else" shouldn't require much more than `connection.set_logger(Logger::new(stdout()))`
  - Happy to see more complex configuration available, but this is a thing that people will see pretty early on. It needs to be possible without stalling getting started.

The two options today seem to be `log` and `slog`. `log` fails the first requirement (the `Log` trait is present, but impossible to interact with from outside the crate). `slog` fails the second (Printing to stdout requires 4 lines and 3 different crates. Most of the examples in their docs that I encountered don't compile, a lot of bit-rot over there it seems). Arguably `log` fails the second requirement as well.

I'll be interested to see what happens with https://github.com/SergioBenitez/Rocket/issues/21, as I assume they have similar concerns to mine.

[...]

PR #901, opened 2017-05-12
Respect --migration-dir and the MIGRATION_DIRECTORY env variable everywhere

### LINKED ISSUE #857: diesel_cli doesn't fully respect MIGRATION_DIRECTORY
When creating and running migrations through `diesel migration generate` and `diesel migration run`, diesel_cli looks at the `MIGRATION_DIRECTORY` environment variable and uses that as shown [here](https://github.com/diesel-rs/diesel/blob/897fac76bcdd1da2596ab74e9dd627b083381b70/diesel_cli/src/main.rs#L114).

However when you run `diesel setup`, it will only [search for](https://github.com/diesel-rs/diesel/blob/368674c8f25de2b986af062e1ff0a7d42ea88501/diesel/src/migrations/mod.rs#L255) and [create](https://github.com/diesel-rs/diesel/blob/897fac76bcdd1da2596ab74e9dd627b083381b70/diesel_cli/src/main.rs#L148) the default migrations directory.

Additionally running `diesel database setup` uses the `run_pending_migrations` function which also only [looks at](https://github.com/diesel-rs/diesel/blob/368674c8f25de2b986af062e1ff0a7d42ea88501/diesel/src/migrations/mod.rs#L95) the default migrations directory, so migrations won't be run if they are configured and stored somewhere else.

Eijebong:
This is the follow up PR of #868

Fixes #857

### DIFF
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -12,13 +12,8 @@ pub fn build_cli() -> App<'static, 'static> {
         .about("A group of commands for generating, running, and reverting \
                 migrations.")
         .setting(AppSettings::VersionlessSubcommands)
-        .arg(Arg::with_name("MIGRATION_DIRECTORY")
-            .long("migration-dir")
-            .help("The location of your migration directory. By default this \
-                   will look for a directory called `migrations` in the \
-                   current directory and its parents.")
-            .takes_value(true)
-            .global(true)
+        .arg(
+            migration_dir_arg()
         ).subcommand(SubCommand::with_name("run")
             .about("Runs all pending migrations")
         ).subcommand(SubCommand::with_name("revert")
@@ -46,12 +41,16 @@ pub fn build_cli() -> App<'static, 'static> {
         ).setting(AppSettings::SubcommandRequiredElseHelp);
 
     let setup_subcommand = SubCommand::with_name("setup")
-        .about("Creates the migrations directory, creates the database \
+        .arg(
+            migration_dir_arg()
+        ).about("Creates the migrations directory, creates the database \
                 specified in your DATABASE_URL, and runs existing migrations.");
 
     let database_subcommand = SubCommand::with_name("database")
         .alias("db")
-        .about("A group of commands for setting up and resetting your database.")
+        .arg(
+            migration_dir_arg()
+        ).about("A group of commands for setting up and resetting your database.")
         .setting(AppSettings::VersionlessSubcommands)
         .subcommand(SubCommand::with_name("setup")
             .about("Creates the database specified in your DATABASE_URL, \
@@ -102,4 +101,14 @@ pub fn build_cli() -> App<'static, 'static> {
         .subcommand(generate_bash_completion_subcommand)
         .subcommand(infer_schema_subcommand)
         .setting(AppSettings::SubcommandRequiredElseHelp)
+}
+
+fn migration_dir_arg<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name("MIGRATION_DIRECTORY")
+        .long("migration-dir")
+        .help("The location of your migration directory. By default this \
+               will look for a directory called `migrations` in the \
+               current directory and its parents.")
+        .takes_value(true)
+        .global(true)
 }
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -172,19 +172,37 @@ fn migrations_dir(matches: &ArgMatches) -> PathBuf {
 }
 
 fn run_setup_command(matches: &ArgMatches) {
-    migrations::find_migrations_directory()
-        .unwrap_or_else(|_| {
-            create_migrations_directory()
-                .unwrap_or_else(handle_error)
+    let migrations_dir = create_migrations_dir(matches).unwrap_or_else(handle_error);
+
+    database::setup_database(matches, &migrations_dir).unwrap_or_else(handle_error);
+}
+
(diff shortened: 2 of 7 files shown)


diesel_cli/tests/database_setup.rs
```diff
@@ -48,8 +48,8 @@ fn database_setup_runs_migrations_if_no_schema_table() {
     let db = database(&p.database_url());
 
     p.create_migration("12345_create_users_table",
-                       "CREATE TABLE users ( id INTEGER )",
```
Eijebong: Derp !

diesel_cli/src/cli.rs
```diff
@@ -46,12 +46,26 @@ pub fn build_cli() -> App<'static, 'static> {
...
     let database_subcommand = SubCommand::with_name("database")
         .alias("db")
-        .about("A group of commands for setting up and resetting your database.")
+        .arg(Arg::with_name("MIGRATION_DIRECTORY")
+            .long("migration-dir")
+            .help("the location of your migration directory. by default this \
+                   will look for a directory called `migrations` in the \
+                   current directory and its parents.")
+            .takes_value(true)
+            .global(true)
```
killercup: Can you

```rust
fn migration_dir_arg() -> Arg {
    Arg::with_name("MIGRATION_DIRECTORY")
        .long("migration-dir")
        .help("the location of your migration directory. by default this \
            will look for a directory called `migrations` in the \
            current directory and its parents.")
        .takes_value(true)
        .global(true)
}
```

instead of duplicating the lines?

diesel_cli/src/main.rs
```diff
@@ -198,14 +216,11 @@ fn generate_bash_completion_command(_: &ArgMatches) {
...
-fn create_migrations_directory() -> DatabaseResult<PathBuf> {
-    let project_root = try!(find_project_root());
-    println!("Creating migrations/ directory at: {}", project_root
-                                                        .join("migrations")
-                                                        .display());
-    try!(fs::create_dir(project_root.join("migrations")));
-    try!(fs::File::create(project_root.join("migrations/.gitkeep")));
-    Ok(project_root)
+fn create_migrations_directory(path: &Path) -> DatabaseResult<PathBuf> {
+    println!("Creating migrations directory at: {}", path.display());
```
killercup: Casual memo to self to get a CLI logging story together (cc #150)

diesel_cli/tests/database_reset.rs
```diff
@@ -75,3 +75,59 @@ fn reset_handles_postgres_urls_with_username_and_password() {
...
+    let result = p.command("database")
+        .arg("reset")
+        .env("MIGRATION_DIRECTORY", "bar")
+        .run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(!db.table_exists("posts"));
+    assert!(db.table_exists("users"));
+    assert!(db.table_exists("__diesel_schema_migrations"));
+}
```
killercup: The only diff is the way `result` is constructed, right? There might be a nicer way to do that, but probably not without adding a testing framework like the rspec crate… (I tried to write [a simple macro](https://play.rust-lang.org/?gist=803dc8c8662f595c4faf647405f41db2&version=stable&backtrace=0) but couldn't get it to work after a few minutes)

Eijebong: Yep, this is the only difference. I duplicated every test (one for arg and one for env).

sgrif: Could refactor to a higher order function that has the signature `fn(Project) -> WhatEverRunReturns` pretty easily I think

diesel_cli/tests/setup.rs
```diff
@@ -98,3 +98,37 @@ fn setup_doesnt_notify_when_not_creating_a_database() {
...
+    // make sure the project builder doesn't create it for us
+    assert!(!p.has_file("migrations"));
+    assert!(!p.has_file("foo"));
+
+    let result = p.command("setup").arg("--migration-dir=foo").run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(!p.has_file("migrations"));
+    assert!(p.has_file("foo"));
+
```
sgrif: ✂️

diesel_cli/tests/setup.rs
```diff
@@ -98,3 +98,37 @@ fn setup_doesnt_notify_when_not_creating_a_database() {
...
+    assert!(!p.has_file("bar"));
+
+    let result = p.command("setup")
+        .env("MIGRATION_DIRECTORY", "bar")
+        .run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(!p.has_file("migrations"));
+    assert!(p.has_file("bar"));
+
```
sgrif: ✂️

diesel_cli/src/main.rs
```diff
@@ -172,19 +172,37 @@ fn migrations_dir(matches: &ArgMatches) -> PathBuf {
...
+    let dir = matches.value_of("MIGRATION_DIRECTORY")
+        .map(PathBuf::from)
+        .or_else(|| {
+            env::var("MIGRATION_DIRECTORY").map(PathBuf::from).ok()
+        }).unwrap_or_else(|| {
+            find_project_root().unwrap_or_else(handle_error).join("migrations")
         });
 
-    database::setup_database(matches).unwrap_or_else(handle_error);
+    if !dir.is_dir() {
```
sgrif: This seems like a weird condition to me. Why do we want to handle the case where the expected path exists but is a file/symlink rather than a directory? It seems like just checking whether it exists would be more clear.

PR #903, opened 2017-05-13
Turn `to_sql` into a standard AST pass

sgrif:
We're not quite to the point where we have a single uniform function, as
we still need to eliminate some of the branches that come from the
existence of `Changeset` and `InsertValues`.

I've been wanting to introduce a `context` parameter that gets passed
around to enable things like expression index support in PG upsert, and
I suspect that this will fix the problem.

I suspect this will alter our performance characteristics in a few cases
(particularly when the query is boxed). I don't expect any major
regressions from this, but I do want to see if it's worth sticking
`#[inline]` in a few places (or even `#[inline(always)]` in some
particularly hot paths like `SelectStatement` and tuples)

### DIFF
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -225,10 +225,8 @@ impl<DB> StatementCacheKey<DB> where
     }
 
     fn construct_sql<T: QueryFragment<DB>>(source: &T) -> QueryResult<String> {
-        use result::Error::QueryBuilderError;
-
         let mut query_builder = DB::QueryBuilder::default();
-        try!(source.to_sql(&mut query_builder).map_err(QueryBuilderError));
+        try!(source.to_sql(&mut query_builder));
         Ok(query_builder.finish())
     }
 }
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -67,21 +67,15 @@ impl<T, U, DB> QueryFragment<DB> for In<T, U> where
     T: QueryFragment<DB>,
     U: QueryFragment<DB> + MaybeEmpty,
 {
-    fn to_sql(&self, out: &mut DB::QueryBuilder) -> BuildQueryResult {
+    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
         if self.values.is_empty() {
             out.push_sql("1=0");
         } else {
-            try!(self.left.to_sql(out));
+            self.left.walk_ast(out.reborrow())?;
             out.push_sql(" IN (");
-            try!(self.values.to_sql(out));
+            self.values.walk_ast(out.reborrow())?;
             out.push_sql(")");
         }
-        Ok(())
-    }
-
-    fn walk_ast(&self, mut pass: AstPass<DB>) -> QueryResult<()> {
-        self.left.walk_ast(pass.reborrow())?;
-        self.values.walk_ast(pass.reborrow())?;
         Ok(())
     }
 }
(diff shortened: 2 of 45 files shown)


diesel/src/connection/statement_cache.rs
```diff
@@ -225,10 +225,8 @@ impl<DB> StatementCacheKey<DB> where
     }
 
     fn construct_sql<T: QueryFragment<DB>>(source: &T) -> QueryResult<String> {
-        use result::Error::QueryBuilderError;
-
         let mut query_builder = DB::QueryBuilder::default();
-        try!(source.to_sql(&mut query_builder).map_err(QueryBuilderError));
+        try!(source.to_sql(&mut query_builder));
```
killercup: `?`

diesel/src/query_builder/ast_pass.rs
```diff
@@ -1,21 +1,30 @@
...
 {
     internals: AstPassInternals<'a, DB>,
 }
 
 impl<'a, DB> AstPass<'a, DB> where
     DB: Backend,
+    DB::QueryBuilder: 'a,
     DB::BindCollector: 'a,
 {
+    #[cfg_attr(feature="clippy", allow(wrong_self_convention))]
```
killercup: Finally, after all this diff noise, this is where it gets interesting!

diesel/src/query_builder/ast_pass.rs
```diff
@@ -46,15 +56,41 @@ impl<'a, DB> AstPass<'a, DB> where
...
         DB: HasSqlType<T>,
         U: ToSql<T, DB>,
     {
-        if let AstPassInternals::CollectBinds(ref mut out) = self.internals {
-            out.push_bound_value(bind)?;
+        use self::AstPassInternals::*;
+        match self.internals {
+            ToSql(ref mut out) => out.push_bind_param(),
+            CollectBinds(ref mut out) => out.push_bound_value(bind)?,
+            _ => {}, // noop
```
killercup: Sorry, that comment doesn't really add anything here

diesel/src/query_builder/ast_pass.rs
```diff
@@ -46,15 +56,41 @@ impl<'a, DB> AstPass<'a, DB> where
         }
     }
 
+    pub fn push_sql(&mut self, sql: &str) {
+        if let AstPassInternals::ToSql(ref mut builder) = self.internals {
+            builder.push_sql(sql);
+        }
+    }
```
killercup: I like these methods that are no-ops for most variants, by the way. Assuming rustc can track which variant we are looking at, this should also be an easy target for dead code elimination.

diesel/src/expression/array_comparison.rs
```diff
@@ -193,19 +181,16 @@ impl<T, DB> QueryFragment<DB> for Many<T> where
...
+        out.unsafe_to_cache_prepared();
+        let mut first = true;
         for value in &self.0 {
-            value.walk_ast(pass.reborrow())?;
+            if first {
+                first = false;
+            } else {
+                out.push_sql(", ");
+            }
+            value.walk_ast(out.reborrow())?;
```
killercup: Nice win for readability! I'd probably add a unit test for that, but it's simple enough that it's not really necessary (and it's tested in a whole bunch of other tests anyway)

diesel/src/query_builder/ast_pass.rs
sgrif: Yes. I think we will need to stick `#[inline(always)]` on the impls for tuples and `SelectStatement`, as I believe that both of them are currently considered too expensive to inline and are extreme hot paths. I was planning on leaving that for a separate commit though.

diesel/src/query_builder/ast_pass.rs
sgrif: I tend to leave these to note that I am intentionally doing nothing in the else case. If we think they add little value, we should probably remove them across the entire code base

Issue #184, opened 2016-02-03
Add support for date/time to SQLite


sgrif:
Currently SQLite has no support for any of the date/time types. Let's add support back for these, as well as `SystemTime`. We should probably have an intermediate struct which others can reason about, similar to how we handled `PgDate` and friends.

aperezdc:
There's an attempt to implement this in PR #344 — which is a bit old and seems to need some love.

euclio:
Any way I can help get this moving? What still needs to be done?

kballard:
Same here. I'd love to see this get done, and I'm happy to help if someone wants to provide some direction.

killercup:
@euclio @kballard sorry, I wasn't watching this issue. We would definitely appreciate your help! I'm not sure I can really follow it, but I think the main issue with #344 is that we need to make sure that SQLite gives us date/times in a format we can definitely parse.

Maybe you can drop by [our Gitter room](https://gitter.im/diesel-rs/diesel) and we see how we can tackle this?

Eijebong:
Oh, I missed this one :o 

Can this be closed now or is SQLite still missing something regarding date/time ? (see #887)

sgrif:
Yeah, I think so. I had envisioned broader support than #887 added, but I don't see any way to do that since we *have* to do string parsing with these types.

gentoo90:
Are there any plans to add schema inference for `DATE`, `TIME`, `DATETIME` and `TIMESTAMP`?
Currently (`master`, 560f650a89ffd283888362) they all fail:
```
diesel/diesel_cli $ sqlite3 test_datetime.db 'CREATE TABLE test (id INTEGER PRIMARY KEY, dt DATETIME)'
diesel/diesel_cli $ echo 'DATABASE_URL=test_datetime.db' > .env
diesel/diesel_cli $ RUST_BACKTRACE=1 cargo run print-schema
    Finished release [optimized] target(s) in 0.0 secs
     Running `/home/igor/projects/rust/diesel/target/release/diesel print-schema`
thread 'main' panicked at 'Could not load table `test`: StringError("Error determining type of test.dt: Unsupported type: datetime")', /checkout/src/libcore/result.rs:859
```

Eijebong:
Yeah, it makes sense. Fixing it

Eijebong:
This is now fixed. @gentoo90, thanks for the report :)

gentoo90:
Thanks for the fix!

PR #904, opened 2017-05-13
Correctly infer datetime types for SQLite

Eijebong:
cc #184

### DIFF
--- a/diesel_infer_schema/src/sqlite.rs
+++ b/diesel_infer_schema/src/sqlite.rs
@@ -85,25 +85,36 @@ pub fn get_primary_keys(conn: &SqliteConnection, table: &TableData) -> QueryResu
         .collect())
 }
 
+fn diesel_type(t: &str) -> Vec<String> {
+    vec!["diesel".into(), "types".into(), t.into()]
+}
+
 pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box<Error>> {
     let type_name = attr.type_name.to_lowercase();
     let path = if is_bool(&type_name) {
-        vec!["diesel".into(), "types".into(), "Bool".into()]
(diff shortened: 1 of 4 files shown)


diesel_infer_schema/src/sqlite.rs
```diff
@@ -103,7 +103,14 @@ pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box
         vec!["diesel".into(), "types".into(), "Float".into()]
     } else if is_double(&type_name) {
         vec!["diesel".into(), "types".into(), "Double".into()]
-    } else {
+    } else if type_name == "datetime" || type_name == "timestamp" {
+        vec!["diesel".into(), "types".into(), "Timestamp".into()]
+    } else if type_name == "date" {
+        vec!["diesel".into(), "types".into(), "Date".into()]
+    } else if type_name == "time" {
+        vec!["diesel".into(), "types".into(), "Time".into()]
```
killercup: I think it's time to add a

```rust
fn diesel_type(t: &str) -> Vec<String> {
    vec!["diesel".into(), "types".into(), t.into()]
}
```

diesel_infer_schema/src/sqlite.rs
```diff
@@ -85,25 +85,36 @@ pub fn get_primary_keys(conn: &SqliteConnection, table: &TableData) -> QueryResu
         .collect())
 }
 
+fn diesel_type(t: &str) -> Vec<String> {
+    vec!["diesel".into(), "types".into(), t.into()]
```
sgrif: I think we can actually just do `t.into()` now.

Eijebong: Nope, `vec![t.into()]` didn't work

PR #906, opened 2017-05-14
Allow bind parameters to be supplied to raw SQL queries

sgrif:
This is a first pass at a much requested feature. While this does not
provide the richer raw SQL story I eventually want to provide, this
should serve as a sufficient stopgap. Bind parameters are included in
the raw SQL string, and then bound after the fact by calling `.bind`
once for each parameter (supplying the SQL type of the parameter at the
call site).

It should be noted that I have explicitly *not* implemented `Expression`
for the new AST node that this added, as it seems like it'll be too easy
to mess up the parameter index when it's used as a fragment of a query
constructed using the query builder. This can be expanded in the future
(maybe only allowed on SQLite and MySQL which don't index bind params),
but I wanted to be conservative to start.

### DIFF
--- a/diesel/src/expression/sql_literal.rs
+++ b/diesel/src/expression/sql_literal.rs
@@ -4,6 +4,7 @@ use backend::Backend;
 use expression::*;
 use query_builder::*;
 use result::QueryResult;
+use super::unchecked_bind::UncheckedBind;
 use types::HasSqlType;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


diesel/src/expression/sql_literal.rs
```diff
@@ -20,12 +21,50 @@ pub struct SqlLiteral<ST> {
...
+    #[doc(hidden)]
     pub fn new(sql: String) -> Self {
         SqlLiteral {
             sql: sql,
             _marker: PhantomData,
         }
     }
+
+    /// Bind a value for use with this SQL query. This function should be used
+    /// with care, as Diesel cannot validate that the value is of the right type.
```
Eijebong: Nor can we that you're binding the right number of values

sgrif: Ironically I originally had that, and then started to go down a path where `.bind` appended the actual bind parameter SQL, and then decided that it was a bad API and scrapped it and forgot to add the docs back

diesel/src/expression/sql_literal.rs
```diff
@@ -20,12 +21,52 @@ pub struct SqlLiteral<ST> {
...
         SqlLiteral {
             sql: sql,
             _marker: PhantomData,
         }
     }
+
+    /// Bind a value for use with this SQL query. This function should be used
+    /// with care, as Diesel cannot validate that the value is of the right type
+    /// nor can it validate that you have passed the correct number of
+    /// parameters.
```
killercup: I'd put the second sentence in a section with a `# Safety` heading

diesel/src/expression/sql_literal.rs
```diff
@@ -20,12 +21,52 @@ pub struct SqlLiteral<ST> {
...
+    /// #[cfg(not(feature="postgres"))]
+    /// let query = sql::<Integer>("SELECT id FROM users WHERE name = ?");
+    /// let seans_id = query.clone().bind::<Text, _>("Sean")
+    ///     .get_result(&connection);
+    /// assert_eq!(Ok(1), seans_id);
+    /// let tess_id = query.bind::<Text, _>("Tess")
+    ///     .get_result(&connection);
+    /// assert_eq!(Ok(2), tess_id);
+    /// # }
+    /// ```
```
killercup: So, after reading this I was immediately wondering if I could use bind multiple times. I thing this should be part of the docs.

And I know this will add a bunch of copypasta code but I'd split this up into _three_ examples: {non-postgres, postgres} ⨯ {single bind, multiple binds}. Maybe a `# PostgreSQL syntax` section with a sentence explaining that this is backend-specific and the two examples with `$n` instead of `?`.

diesel/src/expression/unchecked_bind.rs
```diff
@@ -0,0 +1,57 @@
...
+    value: Value,
+    _marker: PhantomData<ST>,
+}
+
+impl<Query, Value, ST> UncheckedBind<Query, Value, ST> {
+    pub fn new(query: Query, value: Value) -> Self {
+        UncheckedBind {
+            query,
+            value,
+            _marker: PhantomData,
```
killercup: Oh, is this syntax in stable yet? If so, only very recently. This probably bumped us off to "only works with 1.17+".

sgrif: I think it was stable in 1.16

Eijebong: https://github.com/rust-lang/rust/commit/8e97693b2cbcaeb745643ee1b37058e8dde2efa1#diff-5cafcf46bfb947f8e58fa5c16d39018cR7

1.17

PR #908, opened 2017-05-14
Remove a forgotten FIXME

Eijebong:


sgrif:
Turned out to not be needed

PR #913, opened 2017-05-15
Add an insert_default_values function

### LINKED ISSUE #872: INSERT DEFAULT VALUES?
I have this table

```sql
CREATE TABLE users
(
	id serial NOT NULL,
	created_at timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
	PRIMARY KEY (id)
) WITHOUT OIDS;
```

and want to create an user using default values.

I tried to define this structure:

```rust
#[derive(Insertable)]
#[table_name="users"]
pub struct NewUser {}
```

but it raises this error:

```
error: no rules expected the token `]`
  --> model/src/lib.rs:23:19
   |
23 | #[derive(Insertable)]
   |                   ^
```

It is reasonable because in SQL, inserting empty values is not allowed

```
db=# insert into users () values (); 
ERROR:  syntax error at or near ")"  
LINE 1: insert into users () values (); 
```

However, with, `DEFAULT VALUE` you can insert values with default values.

```
actcast=# INSERT INTO users DEFAULT VALUES;
INSERT 0 1
actcast=# SELECT * FROM USERS;
 id |         created_at
----+----------------------------
  1 | 2017-04-26 01:12:56.90735
(1 row)
```

How can I use this feature through diesel? Or if not supported, could you support this (feature request)?

Eijebong:
This will produce the following SQL `INSERT INTO table DEFAULT VALUES`.

Fixes #872


diesel/src/query_builder/insert_statement.rs
```diff
@@ -351,3 +351,66 @@ impl<'a, T, Table> UndecoratedInsertRecord<Table> for &'a Vec<T> where
     &'a [T]: UndecoratedInsertRecord<Table>,
 {
 }
+
+/// The structure returned by [`insert_default_values`](/diesel/fn.insert_default_values.html). The
+/// only thing that can be done with it is call `into`.
```
killercup: Awesome! This'll be helpful for people following function signatures in the API docs.

diesel/src/query_builder/insert_statement.rs
```diff
@@ -351,3 +351,66 @@ impl<'a, T, Table> UndecoratedInsertRecord<Table> for &'a Vec<T> where
...
+    returning: Ret,
+}
+
+impl IncompleteDefaultInsertStatement {
+    #[doc(hidden)]
+    pub fn new() -> Self {
+        IncompleteDefaultInsertStatement {}
+    }
+
+    /// Specify which table the data passed to `insert` should be added to.
```
killercup: Not in this PR, but maybe in the future we should add examples to these incomplete methods to show them used in context. Or at least link to where they are used in context.

diesel/src/query_builder/functions.rs
```diff
@@ -124,3 +124,10 @@ pub fn select<T>(expression: T) -> Select<SelectStatement<()>, T> where
 {
     SelectStatement::simple(()).select(expression)
 }
+
+/// Creates an insert statement. Will insert default values in the table. This
+/// function is not exported by default. As with other commands, the resulting
+/// query can return the inserted rows if you choose.
```
Eijebong: The problem with moving the test here is that I don't have access to schema_dsl which means 3 SQL queries to create the table with default values.

killercup: True. We should think of something to make it easier to write these doc tests (and maybe also write it down somewhere)

diesel_compile_tests/tests/compile-fail/empty_insertable_struct.rs
```diff
@@ -0,0 +1,20 @@
...
+use diesel::*;
+
+table! {
+    users {
+        id -> Integer,
+    }
+}
+
+#[derive(Insertable)]
+//~^ ERROR no rules expected the token `]`
```
killercup: Is this the best error we get? No "cannot derive Insertable for struct with no fields"? doesn't matter, should be a new issue in that case

Eijebong: Yeah, it's not a good error message, but at least it doesn't compile ^^

Eijebong: Can we even customize the error message there ?

sgrif: Yes, we should be panicking inside of codegen and never getting to the Rust macro

diesel/src/query_builder/insert_statement.rs
```diff
@@ -351,3 +351,66 @@ impl<'a, T, Table> UndecoratedInsertRecord<Table> for &'a Vec<T> where
     &'a [T]: UndecoratedInsertRecord<Table>,
 {
 }
+
+/// The structure returned by [`insert_default_values`](/diesel/fn.insert_default_values.html). The
+/// only thing that can be done with it is call `into`.
+#[derive(Debug, Copy, Clone, Default)]
+pub struct IncompleteDefaultInsertStatement {}
```
sgrif: I feel like it should be possible to return `IncompleteInsertStatement<DefaultValues, Insert>` where `DefaultValues` implements `Insertable` for all tables rather than introducing a whole new query type for this

diesel_tests/tests/insert.rs
```diff
@@ -214,3 +214,41 @@ fn insert_empty_slice_with_returning() {
     assert_eq!(Ok(None), insert_one.optional());
     assert_eq!(Ok(vec![]), insert_all);
 }
+
+#[test]
+#[cfg(not(feature = "mysql"))]
```
sgrif: Wait seriously? MySQL doesn't support this?

diesel_compile_tests/tests/compile-fail/empty_insertable_struct.rs
killercup: Created #920

diesel_tests/tests/insert.rs
Eijebong: We're dropping users and creating it again which would commit the transaction IIRC

PR #914, opened 2017-05-15
Add an example/doctest for BoxableExpression

Eijebong:
The example is a bit stupid and useless but it shows what needs to be shown I think.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -165,6 +165,47 @@ use query_builder::{QueryFragment, QueryId};
 /// Helper trait used when boxing expressions. This exists to work around the
 /// fact that Rust will not let us use non-core types as bounds on a trait
 /// object (you could not return `Box<Expression+NonAggregate>`)
+///
+/// # Examples
+///
+/// ```rust
+/// # #[macro_use] extern crate diesel_codegen;
+/// # #[macro_use] extern crate diesel;
+/// # use diesel::types;
+/// # include!("src/doctest_setup.rs");
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+///
+/// # #[derive(PartialEq, Eq, Debug)]
+/// #[derive(Queryable)]
+/// struct User {
+///     id: i32,
+///     name: String,
+/// }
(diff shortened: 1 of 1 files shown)


diesel/src/expression/mod.rs
```diff
@@ -165,6 +165,47 @@ use query_builder::{QueryFragment, QueryId};
 /// Helper trait used when boxing expressions. This exists to work around the
 /// fact that Rust will not let us use non-core types as bounds on a trait
 /// object (you could not return `Box<Expression+NonAggregate>`)
+///
+/// # Examples:
```
killercup: Would look real pretty without the colon but an empty line below it

diesel/src/expression/mod.rs
```diff
@@ -165,6 +165,47 @@ use query_builder::{QueryFragment, QueryId};
...
+///
+/// fn main() {
+///     let conn = establish_connection();
+///     let where_clause: Box<BoxableExpression<users::table, _, SqlType=types::Bool>>;
+///     let search_by_id = true;
+///
+///     if search_by_id {
+///         where_clause = Box::new(users::id.eq(1))
+///     }
+///     else {
```
killercup: We usually put the `else` just after the `}`

diesel/src/expression/mod.rs
Eijebong: Haha, I really love those colons

killercup: No. Those are wrong. There are no colons after headlines when the headline actually printed as a headline. That's just as bad as people who put a space before an exclamation mark ! It's not even correct to do that in French because there you use a _thin space_ or even a _hair space_  ! Typography is art!!11one `</rant>` 😆

Eijebong: Ok ok, sorry 😢

sgrif:
Don't want to reference unstable syntax in the docs, but hypothetically does `let where_clause = box if  {...` look less weird to you?

killercup:
> does let where_clause = box if {... look less weird to you?

A bit less weird, yeah. But I'm weird myself because `if x { 1 } else { 2 }.boxed()` looks fine to me 😄

killercup:
Okay, let's merge this and refine later if someone gets around to it :)

PR #916, opened 2017-05-15
Add an example for the transactions function

### LINKED ISSUE #750: Obsolete reference to TransactionError::UserReturnedError in docs
See: https://github.com/diesel-rs/diesel/blob/6b8d35c48db7e74384af02248730262929988cfa/diesel/src/connection/mod.rs#L33-L36

Should probably also add an example when updating this.

Eijebong:
Fixes #750

I don't know if I should break this one in two parts

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -38,6 +38,61 @@ pub trait Connection: SimpleConnection + Sized + Send {
     ///
     /// The error returned from the function must implement
     /// `From<diesel::result::Error>`.
+    ///
+    /// # Examples:
+    ///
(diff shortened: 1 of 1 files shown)


Eijebong:
🎶  Hello deadlock my old friend... 🎶

Eijebong:
As I said, I could split it in two examples,one with a commit and one with a rollback

killercup:
We talked a bit about this yesterday and my idea was to to make this a bit easier to read by using an empty table as a starting point and adding one row and then one per transaction.

But if that is too much work, I'm fine with merging this as-is. We can always refine this later.

PR #917, opened 2017-05-16
Add a link to the bind method on the sql() function page

Eijebong:
This improves discoverability of bound values.

[ci skip]

### DIFF
--- a/diesel/src/expression/sql_literal.rs
+++ b/diesel/src/expression/sql_literal.rs
@@ -139,6 +139,12 @@ impl<ST> NonAggregate for SqlLiteral<ST> {
 /// DSL. You will need to provide the SQL type of the expression, in addition to
 /// the SQL.
 ///
+/// # Bound parameters
+///
+/// If you need to pass arguments to your query, you should use [`.bind()`].
+///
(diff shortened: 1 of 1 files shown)


killercup:
More links are more good.

diesel/src/expression/sql_literal.rs
```diff
@@ -139,6 +139,12 @@ impl<ST> NonAggregate for SqlLiteral<ST> {
 /// DSL. You will need to provide the SQL type of the expression, in addition to
 /// the SQL.
 ///
+/// # Bound parameters
+///
+/// If you need to pass arguments to your query, you should use [`.bind()`].
+///
+/// [`.bind()`]: struct.SqlLiteral.html#method.bind
```
killercup: Does this actually work? This is rendered on the `/diesel/expression/dsl/fn.sql.html` page, right? And `bind` is on `/diesel/expression/sql_literal/struct.SqlLiteral.html#method.bind`

Eijebong: No it doesn't... I think I didn't wait for `cargo doc` to finish before testing.

killercup: Someone should write an RFC for automatic linking in rustdoc! :trollface:

PR #918, opened 2017-05-20
Fix PgInterval docs

FliegendeWurst:
Please see the [offical docs](https://www.postgresql.org/docs/8.2/static/datatype-datetime.html):
> Internally interval values are stored as months, days, and seconds.


Eijebong:
Yep, thank you very much !

PR #924, opened 2017-05-26
Fix sqlite appveyor

mcgoo:
Recent changes to cargo mean that `cargo test` won't look for dlls in paths outside of the target dir even if a build script tries to add them - the path/dylib path must be manually updated outside of cargo.


mcgoo:
Added a commit to generate an import library to fix the msvc tests also.

killercup:
I have experience in dealing with appveyor or sqlite on windows, but this looks good… and green! So, 👍 from me. Anyone else want to have a look? @sgrif?

mcgoo:
It should be there as long as the AppVeyor build machine doesn't change radically. It's the version from Visual Studio 2013, but any of the 10-odd tool chains on the image would be fine. I somewhat arbitrarily chose the version from VS2103 since mysql also requires VS2013 at this point (although I have an incoming PR that allows 2015 and 2017 to work with mysql also.)

The other alternative I was able to think of was to put the VS tools in the path, but that is a global change that will indirectly affect rustc and this way will not.

PR #926, opened 2017-05-30
Emit a message on stderr when infering a varchar[]

Eijebong:
Since postgres does not coerce varchar[] to text[] and diesel does not
make any difference between varchar and text, you cannot use array
specific operators with a varchar[].

### DIFF
--- a/diesel_infer_schema/src/pg.rs
+++ b/diesel_infer_schema/src/pg.rs
@@ -1,4 +1,5 @@
 use std::error::Error;
+use std::io::{stderr, Write};
 
 use data_structures::*;
 
@@ -9,6 +10,14 @@ pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box
     } else {
         &attr.type_name
     };
+
+    let tpe_is_varchar = tpe.to_lowercase() == "varchar";
+
+    // Postgres doesn't coerce varchar[] to text[] so print out a message to inform
+    // the user.
+    if tpe_is_varchar && is_array {
+        writeln!(&mut stderr(), "The column `{}` is of type `varchar[]`. This will cause problems when using Diesel. You should consider changing the column type to `text[]`.", attr.column_name)?;
+    }
 
     Ok(ColumnType {
         path: vec!["diesel".into(), "types".into(), capitalize(tpe)],


diesel_infer_schema/src/pg.rs
```diff
@@ -10,6 +11,14 @@ pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box
         &attr.type_name
     };
 
+    let tpe_is_varchar = tpe.to_lowercase() == "varchar";
+
+    // Postgres doesn't coerce varchar[] to text[] so print out a message to inform
+    // the user.
+    if tpe_is_varchar && is_array {
+        writeln!(&mut stderr(), "The column {} is of type varchar[]. Array operations are not supported on the varchar[] type. You should change your schema to be text[] as it's more efficient anyway.", attr.column_name)?;
```
killercup: Can you add some back ticks around the column and type names?

> more efficient anyway

citation needed? 😉

Eijebong: Hu, @sgrif said that :p Maybe I should remove that part.

diesel_infer_schema/src/pg.rs
```diff
@@ -10,6 +11,14 @@ pub fn determine_column_type(attr: &ColumnInformation) -> Result<ColumnType, Box
         &attr.type_name
     };
 
+    let tpe_is_varchar = tpe.to_lowercase() == "varchar";
+
+    // Postgres doesn't coerce varchar[] to text[] so print out a message to inform
+    // the user.
+    if tpe_is_varchar && is_array {
+        writeln!(&mut stderr(), "The column `{}` is of type `varchar[]`. Array operations are not supported on the `varchar[]` type. You should change your column type to `text[]`.", attr.column_name)?;
```
sgrif: The second sentence is misleading. I'd change it to: `This will cause problems when using Diesel.`

killercup:
Do you want to add a test for this? Otherwise this looks good to get merged.

Eijebong:
I don't think it's possible to add a test as we cannot capture stderr in a test and check it without doing some crazy things with libc or something like that...

killercup:
Oh, I thought that'd end up as part of the compile error in compile test (when provoking a compile error). But, you could of course call this as a subcommand and… nope, not worth it ;)

PR #928, opened 2017-06-01
Set up the groundwork for multi-table joins

sgrif:
This solves the fundamental problem of needing
`SelectableExpression<Join<Left, Right>> for column` to work when we
don't know where the table concretely appears, just that it appears
somewhere. With this commit, the logic is now that a column can be
selected from any join where its table appears within the join exactly
once. This disallows `users.inner_join(posts.inner_join(users))`, which
we don't want to allow as it would require the second instance of
`users` to be aliased.

This does require that there is an `impl ContainsTable<left> for right`
for every two tables that would appear. Right now I'm just brute-forcing
these when the association between the two tables is defined. For
multi-table joins to work when two of the tables in the join do not have
an association, this will require a manual impl. We'll probably provide
something like `enable_multi_table_joins!(users, comments)` or similar.
This *may* go away eventually with specialization, if projections on
default associated types are eventually allowed in the fully monomorphic
case.

It should be noted that this commit does not enable multi-table joins on
itself. `Join` still does not implement `JoinDsl`, and tables do not
implement `JoinTo<Join<Mid, Right>>`. The code required to enable that
is pretty minimal, but I don't want to flip the switch just yet, as
we're actually generating invalid SQL at the moment and that needs to be
resolved.

However, with this change, all of the limitations that prevent
multi-table joins at the type level are eliminated.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -26,47 +26,31 @@ macro_rules! __diesel_column {
         impl SelectableExpression<$($table)::*> for $column_name {
         }
 
-        impl AppearsOnTable<$($table)::*> for $column_name {
-        }
-
-        impl<Right, Kind> SelectableExpression<
-            Join<$($table)::*, Right, Kind>,
-        > for $column_name where
-            $column_name: AppearsOnTable<Join<$($table)::*, Right, Kind>>,
+        impl<QS> AppearsOnTable<QS> for $column_name where
+            QS: ContainsTable<$($table)::*, Count=Once>,
         {
         }
 
-        impl<Left> SelectableExpression<
-            Join<Left, $($table)::*, Inner>,
+        impl<Left, Right> SelectableExpression<
+            Join<Left, Right, LeftOuter>,
         > for $column_name where
-            Left: $crate::JoinTo<$($table)::*>
+            $column_name: AppearsOnTable<Join<Left, Right, LeftOuter>>,
+            Left: ContainsTable<$($table)::*, Count=Once>,
+            Right: ContainsTable<$($table)::*, Count=Never>,
         {
         }
 
-        impl<Right, Kind> AppearsOnTable<
-            Join<$($table)::*, Right, Kind>,
+        impl<Left, Right> SelectableExpression<
+            Join<Left, Right, Inner>,
         > for $column_name where
-            Right: QuerySource,
-            $($table)::*: $crate::JoinTo<Right>
-        {
-        }
-
-        impl<Left, Kind> AppearsOnTable<
-            Join<Left, $($table)::*, Kind>,
-        > for $column_name where
-            Left: $crate::JoinTo<$($table)::*>
-        {
-        }
-
-        // FIXME: Remove this when overlapping marker traits are stable
-        impl<Join, On> AppearsOnTable<JoinOn<Join, On>> for $column_name where
-            $column_name: AppearsOnTable<Join>,
+            $column_name: AppearsOnTable<Join<Left, Right, Inner>>,
+            Join<Left, Right, Inner>: ContainsTable<$($table)::*, Count=Once>,
         {
         }
 
         // FIXME: Remove this when overlapping marker traits are stable
         impl<Join, On> SelectableExpression<JoinOn<Join, On>> for $column_name where
-            $column_name: SelectableExpression<Join>,
+            $column_name: SelectableExpression<Join> + AppearsOnTable<JoinOn<Join, On>>,
         {
         }
 
@@ -339,6 +323,7 @@ macro_rules! table_body {
             use $crate::associations::HasTable;
             use $crate::query_builder::*;
             use $crate::query_builder::nodes::Identifier;
+            use $crate::query_source::{ContainsTable, Once};
             $(use $($import)::+;)+
             pub use self::columns::*;
 
@@ -418,6 +403,10 @@ macro_rules! table_body {
                 }
             }
 
+            impl ContainsTable<table> for table {
+                type Count = Once;
+            }
+
             impl_query_id!(table);
 
             /// Contains all of the columns of this table
@@ -426,7 +415,8 @@ macro_rules! table_body {
                 use $crate::{Expression, SelectableExpression, AppearsOnTable, QuerySource};
                 use $crate::backend::Backend;
                 use $crate::query_builder::{QueryFragment, AstPass};
-                use $crate::query_source::joins::{Join, JoinOn, Inner};
+                use $crate::query_source::joins::{Join, JoinOn, Inner, LeftOuter};
+                use $crate::query_source::{ContainsTable, Once, Never};
                 use $crate::result::QueryResult;
                 $(use $($import)::+;)+
 
@@ -534,6 +524,9 @@ macro_rules! joinable_inner {
         primary_key_ty = $primary_key_ty:ty,
         primary_key_expr = $primary_key_expr:expr,
     ) => {
+        impl $crate::query_source::ContainsTable<$right_table_ty> for $left_table_ty {
+            type Count = $crate::query_source::Never;
+        }
         impl $crate::JoinTo<$right_table_ty> for $left_table_ty {
             type JoinOnClause = $crate::expression::helper_types::Eq<
                 $crate::expression::nullable::Nullable<$foreign_key>,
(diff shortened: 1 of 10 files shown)


diesel/src/macros/mod.rs
```diff
@@ -27,24 +27,24 @@ macro_rules! __diesel_column {
         }
 
         impl<QS> AppearsOnTable<QS> for $column_name where
-            QS: ContainsTable<$($table)::*, Count=Once>,
+            QS: AppearsInFromClause<$($table)::*, Count=Once>,
```
Eijebong: Yeah, this is much more readable

diesel_compile_tests/tests/compile-fail/right_side_of_left_join_requires_nullable.rs
```diff
@@ -27,13 +27,17 @@ fn main() {
     let join = users::table.left_outer_join(posts::table);
 
     // Invalid, only Nullable<title> is selectable
-    let _ = join.select(posts::title); //~ ERROR E0277
+    let _ = join.select(posts::title);
```
Eijebong:
```
error[E0271]: type mismatch resolving `<users::table as diesel::query_source::AppearsInFromClause<posts::table>>::Count == diesel::query_source::Succ<diesel::query_source::Never>`
  --> src/main.rs:30:18
   |
30 |     let _ = join.select(posts::title);
   |                  ^^^^^^ expected struct `diesel::query_source::Never`, found struct `diesel::query_source::Succ`
   |
   = note: expected type `diesel::query_source::Never`
              found type `diesel::query_source::Succ<diesel::query_source::Never>`
```

This is really hard to understand what's going on here. Is there any way this could be better ?

diesel_compile_tests/tests/compile-fail/right_side_of_left_join_requires_nullable.rs
```diff
@@ -27,13 +27,17 @@ fn main() {
...
+    //~^ ERROR E0271
+    //~| ERROR E0271
     // Valid
     let _ = join.select(posts::title.nullable());
     // Valid -- NULL to a function will return null
     let _ = join.select(lower(posts::title).nullable());
     // Invalid, only Nullable<title> is selectable
-    let _ = join.select(lower(posts::title)); //~ ERROR E0277
+    let _ = join.select(lower(posts::title));
```
Eijebong: Same error here

diesel/src/query_source/mod.rs
```diff
@@ -48,3 +48,13 @@ pub trait Table: QuerySource + AsQuery + Sized {
     fn primary_key(&self) -> Self::PrimaryKey;
     fn all_columns() -> Self::AllColumns;
 }
+
+pub trait AppearsInFromClause<QS> {
```
killercup: ℹ️ Tracks how many times a table occurs in QuerySource (from/join clause)

Maybe add ↑ as doc comment?

diesel/src/query_source/joins.rs
```diff
@@ -180,3 +174,33 @@ impl<DB: Backend> QueryFragment<DB> for LeftOuter {
         Ok(())
     }
 }
+
+use super::{Succ, Never, AppearsInFromClause};
+
+impl<T, Left, Right, Kind> AppearsInFromClause<T> for Join<Left, Right, Kind> where
+    Left: AppearsInFromClause<T>,
+    Right: AppearsInFromClause<T>,
+    Left::Count: Plus<Right::Count>,
```
killercup: ℹ️ Constraint D: Table `T` occurs on left side of join one more time than on right side

Actually, why is that?

diesel/src/query_source/joins.rs
```diff
@@ -180,3 +174,33 @@ impl<DB: Backend> QueryFragment<DB> for LeftOuter {
...
+{
+    type Count = <Left::Count as Plus<Right::Count>>::Output;
+}
+
+impl<T, Join, On> AppearsInFromClause<T> for JoinOn<Join, On> where
+    Join: AppearsInFromClause<T>,
```
killercup: ℹ️ Expose inner count in JoinOn wrapper

diesel/src/macros/mod.rs
```diff
@@ -26,47 +26,31 @@ macro_rules! __diesel_column {
...
-            $column_name: AppearsOnTable<Join<$($table)::*, Right, Kind>>,
-        {
-        }
-
-        impl<Left> SelectableExpression<
-            Join<Left, $($table)::*, Inner>,
-        > for $column_name where
-            Left: $crate::JoinTo<$($table)::*>
+        impl<QS> AppearsOnTable<QS> for $column_name where
+            QS: AppearsInFromClause<$($table)::*, Count=Once>,
```
killercup: ℹ️ Constraint A: Can only select column if column's table occurs exactly once in from/join

diesel/src/macros/mod.rs
```diff
@@ -26,47 +26,31 @@ macro_rules! __diesel_column {
...
-        impl<Right, Kind> AppearsOnTable<
-            Join<$($table)::*, Right, Kind>,
+        impl<Left, Right> SelectableExpression<
+            Join<Left, Right, LeftOuter>,
         > for $column_name where
-            Right: QuerySource,
-            $($table)::*: $crate::JoinTo<Right>
+            $column_name: AppearsOnTable<Join<Left, Right, LeftOuter>>,
+            Left: AppearsInFromClause<$($table)::*, Count=Once>,
+            Right: AppearsInFromClause<$($table)::*, Count=Never>,
```
killercup: ℹ️ Constraint B: Can only select column when column's table appears on left side of Left Outer Join and not on right side

diesel/src/macros/mod.rs
```diff
@@ -26,47 +26,31 @@ macro_rules! __diesel_column {
...
         > for $column_name where
-            Left: $crate::JoinTo<$($table)::*>
-        {
-        }
-
-        // FIXME: Remove this when overlapping marker traits are stable
-        impl<Join, On> AppearsOnTable<JoinOn<Join, On>> for $column_name where
-            $column_name: AppearsOnTable<Join>,
+            $column_name: AppearsOnTable<Join<Left, Right, Inner>>,
+            Join<Left, Right, Inner>: AppearsInFromClause<$($table)::*, Count=Once>,
```
killercup: ℹ️ Constraint C: Can only select column when column's table appears once on either left or right side of Left Inner Join

diesel/src/macros/mod.rs
```diff
@@ -418,6 +403,10 @@ macro_rules! table_body {
                 }
             }
 
+            impl AppearsInFromClause<table> for table {
```
killercup: Why use the same name for the type param as for the struct?

diesel/src/macros/mod.rs
```diff
@@ -534,6 +524,9 @@ macro_rules! joinable_inner {
         primary_key_ty = $primary_key_ty:ty,
         primary_key_expr = $primary_key_expr:expr,
     ) => {
+        impl $crate::query_source::AppearsInFromClause<$right_table_ty> for $left_table_ty {
+            type Count = $crate::query_source::Never;
```
killercup: ℹ️ Right side of left inner join does not contain table on left side

diesel_compile_tests/tests/compile-fail/right_side_of_left_join_requires_nullable.rs
killercup: Yep, it's one of those errors we should cover in the docs. Or, [you know](https://deterministic.space/hook-into-rustc-errors.html) customize.

sgrif: Once all of this is merged and stable functionality-wise, I think I will replace `Succ` with `Once` and `MoreThanOnce` just for the minor improvement in the error messages.

diesel/src/query_source/joins.rs
sgrif: I'm not sure I understand the question here? This is basically just saying that both counts are either `Never` or `Succ`.

diesel/src/macros/mod.rs
sgrif: Because they're the same type. This is where we say a table contains itself exactly once. If we can get the changes we need to specialization, I'd have these two blanket impls instead:

```rust
impl<T, U> AppearsInFromClause<T> for U {
    type Count = Never;
}

// The table constraint is to make this impl disjoint from the impl we provide for `Join`
impl<T: Table> AppearsInFromClause<T> for T {
    type Count = Once;
}
```

diesel/src/query_source/joins.rs
killercup: Ah, that makes sense. (I added the question today, but wrote the comment on Saturday)

PR #929, opened 2017-06-02
Set default busy timeout to 5 seconds

king6cong:
similar to:
https://github.com/jgallagher/rusqlite/commit/05b03ae2cec9f9f630095d5c0e89682da334f4a4

this should reduce SQLITE_BUSY situation

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -25,6 +25,7 @@ time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.6.0", optional = true, features = ["use_std"] }
 ipnetwork = { version = "0.12.2", optional = true }
+log = "0.3"
 
 [dev-dependencies]
 cfg-if = "0.1.0"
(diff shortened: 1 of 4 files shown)


killercup:
Thanks. Can you provide a bit more information what this "SQLITE_BUSY situation" is and how this solves it? :)

king6cong:
yeah, this deal with multiple concurrent sqlite writers.

### there are some discussions like here: 
https://github.com/mattn/go-sqlite3/issues/274

### and also these documentations: 

* http://www.sqlite.org/cvstrac/wiki?p=MultiThreading
* https://www.sqlite.org/c3ref/busy_timeout.html

### tldr:

* SQLITE_BUSY will occur inevitably when there are multiple thread writers and we need to handle this
* add busy timeout will reduce the SQLITE_BUSY situation
* at the moment, diesel will panic in `step` when SQLITE_BUSY occurred, I think log it maybe better for users at the moment. Better error handling should be saved for another commit :)

sgrif:
I'd rather not introduce the log crate or warnings here. We need to properly return an error here.

king6cong:
Updated

king6cong:
@sgrif @killercup These changes turn out working well in our product, any ideas?

sgrif:
> We need to properly return an error here.

We cannot accept a PR that simply has `TODO: Better error handling`

sgrif:
Going to close this, as it's been sitting around for a while. Feel free to re-submit if you're still interested in working on this. However for this to be accepted we'd need to present this error case to the user. Silently treating it as an empty result set is not an acceptable solution.

Issue #298, opened 2016-04-23
*ANY* breaking changes that we want to make in the near future


sgrif:
Right now the plan is to have 0.7 be the last release before 1.0. That means that we should start seriously thinking about anything we might want to break in the near future. This does _not_ mean that we should feel rushed to get remaining major features out. Only that we should address anything we expect to cause breaking changes.

A few things that we should look at off the top of my head
- `group_by` support, and how this affects `NonAggregate`. Honestly, I think we're so far from this that it'll have to wait until 2.0 if it needs breaking changes (it might not need breaking changes though). This likely requires some form of negative reasoning in the language. Regardless, I need to take a swing at it before 1.0
- Logging. Does this eliminate the need for `debug_sql!` and the `Debug` backend?
- Reforming joins to work with multiple tables. Does the improved association support in 0.7 eliminate the majority of the need for this?

iqualfragile:
maybe support for composite primary keys? I guess that would mean some changes, not sure if the api would change though.

sgrif:
Yes you are right we should attempt to explore that in 0.7 or 0.8

sgrif:
Everything in this issue has been resolved. `group_by` is going to wait for 2.0. Logging has been removed from the milestone (we can add it backwards compatibly I think, though it'll require a default impl of the `set_logger` method on `Connection` which panics). Joins have been sufficiently reformed.

sgrif:
Also...

> Right now the plan is to have 0.7 be the last release before 1.0

Yeah, that worked out well... >_>

PR #930, opened 2017-06-03
Allow joins which contain more than 2 tables

### LINKED ISSUE #89: Associations: what should they look like?
Let's talk about associations. First what's in the public API today, what's in the private API today, and what they actually need to do. We can use those things to drive what the API should look like.
# Public API today

Right now `Table` has `inner_join` and `left_outer_join` which can take any other table which implements `JoinTo`. This is limited to exactly _one_ table today, due to some limitations that I think will be resolved by specialization. There also needs to be some boilerplate implementations of `SelectableExpression` for all the various types of query sources, which will go away when specialization (or https://github.com/rust-lang/rust/issues/29864) lands and replaced with:

``` rust
impl<Left, Right, ST, T> SelectableExpression<InnerJoinSource<Left, Right>, ST>
    for T where
    T: SelectableExpression<Left, ST>,
    Left: JoinTo<Right>,
{}

impl<Left, Right, ST, T> SelectableExpression<InnerJoinSource<Left, Right>, ST>
    for T where
    T: SelectableExpression<Right, ST>,
    Left: JoinTo<Right>,
{}
```

This will also basically allow us to implement the join methods on the various join sources as well, though we might need to do some tuple hackery.
# Private API today

The annotations `#[has_many]` and `#[belongs_to]` are part of the internal API today. Both implement `JoinTo` automatically. `belongs_to` additionally implements `BelongingToDsl<Parent>` for the child. I wrote up some thoughts on where that is today at https://github.com/sgrif/diesel/issues/86#issuecomment-171031933.
# What do Associations actually need to do?

Ultimately we're either eager loading the children for a collection of parents, or we're loading the children for a single parent. I believe that `BelongingTo` sufficiently handles the latter, but it'll effectively be handled by the former.
# One To Many

I'm pretty reasonably confident that associations should be non invasive (user doesn't know it has many posts). That means that the type of a user and all of its posts is `(User, Vec<Post>)`. This can get tricky when dealing with multiple levels of nesting here. For example, the type of a user, all the posts they've written, and all the comments written by that user is `(User, Vec<Post>, Vec<Comment>)`. By contrast, the type of a user, all of the posts they've written, and all of the comments left on each of those posts would be `(User, Vec<(Post, Vec<Comment>)>)`.

I'd imagine the way to specify that you want to load the comments for the posts and not for the users is by writing `users.left_outer_join(posts.left_outer_join(comments))` as opposed to `users.left_outer_join(posts).left_outer_join(comments)`.

The case of a user and all of its posts would be written today as:

``` rust
users.left_outer_joins(posts::table).load(&connection)
    .group_by(|(user, post)| user)
    .map(|(user, posts)| posts.into_iter().filter_map(|p| p).collect())
    .collect()
```

The case involving comments is effectively impossible today.

We do not need to have specific code to handle loading the children for a single record, as `Post::belonging_to(&user)` is sufficient.

It should also be noted that I don't want to restrict what you're able to work with. We should be able to get a user, and the title of all of their posts by doing `users.left_outer_joins(posts::table).select((users.all_columns(), posts::title))`.
# One To One

The type of a one to one relationship is either `(A, B)` or `(A, Option<B>)`. Technically only a `belongs_to` can be mandatory, but you can still end up with the first signature with an inner join. Regardless of if we're going child to parent or parent to child, the way this is written today is simply:

``` rust
users.inner_join(profile)
```

And loading a single record is handled by `Post::belonging_to(&user)`. I do not believe we need any additional code to handle this case, but we could potentially rename the join methods to have parity with whatever we come up with for one to many.
# Composition

Associations should be composable. If we want to get all of the comments that have been written for any posts written by a user, we should be able to re-use our existing logic, without having to actually load the posts.

### LINKED ISSUE #398: Many-To-Many Joins
Currently it seems to not be possible to use many-to-many joins in diesel. I was unable to find some out-of-the-box "many-to-many from A to B via cross-table C" -- and for manual implementation (one-to-many from  A to C, many-to-one from C to B) support for id-less tables (C obiously has a compound PK) is missing.

Thus, I wanted to open this issue as a reminder (I'm quite astonished that there is none yet :))

### LINKED ISSUE #759: Unify `LeftOuterJoinSource` and `InnerJoinSource`
The main reason that these differ is because we need to change the SQL type of columns on the right side of the join. This required that these structs be separate types back in the pre-0.1 days when I added them, but we have sufficient information now to separate them. I'd like to have just a single `JoinSource<Left, Right, JoinType>`.

sgrif:
This builds on #928 so I've pointed this at that branch to make it easier to review. This pull request is split into 4 commits, and I recommend reviewing each commit separately. Each message has more context about the individual changes. Beyond this PR, I still need to add some doctests for multi-table joins, and figure out the right way to document `enable_multi_table_joins!`.

Fixes #398 
Fixes #89 
Partially addresses #298 
Fixes #759

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -61,6 +61,7 @@ mod query_dsl;
 pub mod query_source;
 pub mod result;
 pub mod row;
+mod util;
 
 pub mod helper_types {
     //! Provide helper types for concisely writing the return type of functions.
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -54,6 +54,12 @@ macro_rules! __diesel_column {
         {
         }
 
+        // FIXME: Remove this when overlapping marker traits are stable
+        impl<From> SelectableExpression<SelectStatement<From>> for $column_name where
+            $column_name: SelectableExpression<From> + AppearsOnTable<SelectStatement<From>>,
+        {
+        }
+
         impl $crate::expression::NonAggregate for $column_name {}
 
         impl $crate::query_source::Column for $column_name {
@@ -414,7 +420,7 @@ macro_rules! table_body {
                 use super::table;
                 use $crate::{Expression, SelectableExpression, AppearsOnTable, QuerySource};
                 use $crate::backend::Backend;
-                use $crate::query_builder::{QueryFragment, AstPass};
+                use $crate::query_builder::{QueryFragment, AstPass, SelectStatement};
                 use $crate::query_source::joins::{Join, JoinOn, Inner, LeftOuter};
                 use $crate::query_source::{AppearsInFromClause, Once, Never};
                 use $crate::result::QueryResult;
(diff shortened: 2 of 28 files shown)


sgrif:
Reopened against master in #933

PR #931, opened 2017-06-03
Add error message for derive(Insertable) on empty struct

### LINKED ISSUE #920: Nice codegen errors
Make sure our codegen (`#[derive(…)]`) generates useful error messages. They should be tested by a compile-fail test.

Especially edge cases like empty structs should be caught where necessary. For example,

```rust
#[derive(Insertable)]
#[table_name="users"]
pub struct NewUser {}
```

currently fails with

> no rules expected the token `]`

It should probably say something like this instead (maybe with a link to docs or some other feature the user might've been meaning to use):

> Failed to derive `Insertable` on `NewUser` struct because it has no fields.

We already alert the user to some things, e.g. how we currently don't support generic structs, with code like [this](https://github.com/diesel-rs/diesel/blob/62dbbecc7fa689a95ded2618b9eb4cecc5051bcc/diesel_codegen/src/insertable.rs#L14-L16).

Feel free to add more examples in the comments.

DorianListens:
This change adds a new error message when attempting to derive
Insertable on a struct with no fields, and a compile-fail test to
ensure it is displayed appropriately.

See #920

### DIFF
--- a/diesel_codegen/src/insertable.rs
+++ b/diesel_codegen/src/insertable.rs
@@ -21,6 +21,11 @@ pub fn derive_insertable(item: syn::MacroInput) -> quote::Tokens {
     let lifetimes = model.generics.lifetimes;
     let fields = model.attrs.as_slice();
 
+    if fields.is_empty() {
+        panic!("Failed to derive `Insertable` for `{}`: `Insertable` \
+           cannot be used on structs with empty fields", struct_name);
+    }
+
     quote!(impl_Insertable! {
         (
             struct_name = #struct_name,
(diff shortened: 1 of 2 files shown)


DorianListens:
I imagine there are probably other instances of codegen producing less-than-helpful errors, but I don't really have any familiarity with the codegen system, so I don't know where they are. I'd be happy to merge this as is, since it makes things better, but I'm also happy to wait and add more improvements to this PR directly.

diesel_codegen/src/insertable.rs
```diff
@@ -21,6 +21,10 @@ pub fn derive_insertable(item: syn::MacroInput) -> quote::Tokens {
     let lifetimes = model.generics.lifetimes;
     let fields = model.attrs.as_slice();
 
+    if fields.is_empty() {
+        panic!("Failed to derive `Insertable` for `{}` because it has no fields.", struct_name);
```
killercup: What do you think of rephrasing that to

> Failed to derive `Insertable` for `{}`: `Insertable` cannot be used on structs with empty fields

DorianListens: Works for me!

PR #940, opened 2017-06-07
[WIP] Prototype of infering enums from the database schema

weiznich:
* This will only work with postgres in the current version

This is a first draft version to get some feedback on this. 
Things to fix before merging:
- [ ] Formating
- [ ] Documentation
- [ ] Tests
- [ ] Fix compiling other backends

Using an implementation similar to the shown, will restrict the produced binary to exactly the same database that was used to compile the code because the type oid's of the enums are hardcoded in `HasSqlType::metadata`. As far as I understand those oid's could/will change if you run your migrations on a other database instance, so the provided mapping will be invalid.

As far as I see there are two possible solutions:
- Document and accept the behavior 
- Add some way to query the database on startup/connect for the actual oid's and store them in some kind of static map. Then use this map to generate the metadata

### DIFF
--- a/diesel/src/macros/macros_from_codegen.rs
+++ b/diesel/src/macros/macros_from_codegen.rs
@@ -122,3 +122,24 @@ macro_rules! embed_migrations {
         }
     }
 }
+
+#[macro_export]
+macro_rules! infer_enum {
+    ($database_url: expr) => {
+        mod __diesel_infer_enum {
+            #[derive(InferEnum)]
+            #[infer_enum_options(database_url=$database_url)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_enum::*;
+    };
+
+    ($database_url: expr, $schema_name: expr) => {
+        mod __diesel_infer_enum {
+            #[derive(InferEnum)]
+            #[infer_enum_options(database_url=$database_url, schema_name=$schema_name)]
+            struct _Dummy;
+        }
+        pub use self::__diesel_infer_enum::*;
+    };
+}
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -90,7 +90,16 @@ pub fn build_cli() -> App<'static, 'static> {
              .help("Use table list as blacklist")
              .conflicts_with("whitelist"));
 
-    App::new("diesel")
+    let infer_enums = SubCommand::with_name("print-enums")
+        .setting(AppSettings::VersionlessSubcommands)
+        .about("Print enum definitions for database enum.")
+        .arg(Arg::with_name("schema")
+             .long("schema")
+             .short("s")
+             .takes_value(true)
+             .help("The name of the schema."));
+
+    let mut diesel = App::new("diesel")
         .version(env!("CARGO_PKG_VERSION"))
         .setting(AppSettings::VersionlessSubcommands)
         .after_help("You can also run `diesel SUBCOMMAND -h` to get more information about that subcommand.")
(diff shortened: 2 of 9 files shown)


weiznich:
[Example code](https://gist.github.com/weiznich/1b4ff9b804efac8fb389be7e3283ed25)

sgrif:
As you've mentioned here, this solution can't work, as the OID needs to be found at runtime not compile time. We cannot assume the OID is stable in any way for types that are not core to PG. Querying at startup is one possibility, but I'd prefer something that is lazily done when constructing the query that requires that type. They can't be handled in a static map, since you can connect to more than one database.

I think that problem needs to be solved first, as a standalone pull request that handles only that problem.

weiznich:
I've prototyped an solution for the runtime cacheing [here](https://github.com/weiznich/diesel/commit/b16584a72b72b1837111785a00372c033b3b9bac). This allows to use the enum feature across different database instances.

The caching is implemented on connection layer, so it should work even if different instances are used in parallel.
The implementation is missing some error handling and the handling of enum array type combinations. So see it as something to start iterating over.

sgrif:
The idea of having a static vs dynamic enum is interesting, but I don't think it pans out to the `ToSql` case. Ultimately I think the way to go about this is as follows:

1. Add a `MetadataLookup` type to `TypeMetadata` (set to `()` for all cases)
2. Change the signature of `ToSql` to take some struct that wraps `W` instead of `W` directly (this type implements `Write` where `W: Write`. This type would also be generic over `DB` and have a field of type `DB`
3. Change the signature of `HasSqlType::metadata` to take a reference to `DB::TypeMetadata`
4. Introduce an actual metadata lookup type for PG, which wraps a reference to `PgConnection` (this will almost 100% require doing a `mem::transmute` to have a `&'static PgConnection`, which is unfortunate, but until we get ATCs I don't see anyway around that without adding lifetimes to `Pg` or to `TypeMetadata` which are places I really don't want to add a lifetime parameter)

sgrif:
Also for the record, I'm less concerned about the caching aspect of it, and more concerned about the API. An initial implementation could make the request every time and that's fine. It's easy to add caching later.

weiznich:
I've used a [slightly](https://github.com/weiznich/diesel/commit/e046377e684adea788872ce69fe71d0e299f68a8) different approach to get the type lookup working inside of `ToSql`. This reuses most of the already written code and does not change that many parts of the existing interface. Furthermore it does not require any hacks including `mem::transmute` or `&'static` references.

sgrif:
I'd prefer to avoid exposing the entire connection object to `ToSql`, especially since until now it's been entirely concerned about the backend and not the actual underlying connection. Part of that layout is so that we can provide some amount of encapsulation.

weiznich:
> I'd prefer to avoid exposing the entire connection object to ToSql, especially since until now it's been entirely concerned about the backend and not the actual underlying connection.

As far as I can understand we must expose the connection in some kind to `ToSql`, because otherwise we could not reason about dynamic sql types (We may need to query the database. To query the database we will need the connection.)

Furthermore, we do not really expose the connection for all backend. We expose a type, that implements `MetadataLookup` in theory this could be anything. In practice in case of the Pg backend it is because of the restriction that we may need to query the database the actual connection object. For other back ends this must not be true.

If there is any better solution to this, I'm quite happy to change it. But as explained above, I think we need to expose the connection in some kind, because we need to execute queries in some cases.

weiznich:
@sgrif  I've updated the branch to do not expose `PgConnection` to `ToSql`

PR #941, opened 2017-06-09
Refactor `LoadDsl` to make it easier to introduce the raw SQL API

sgrif:
One of the things that's going to make our upcoming raw SQL escape hatch
different from our current `sql` function is that it's not going to
require specifying the SQL type fo the query, and it's going to load
parameters by name instead of by index. This does mean that we'll need
to provide a parallel trait to `Queryable`, but I'd prefer to avoid
introducing a parallel version of `LoadDsl`.

So in order to make the type we will introduce for this compatible with
`LoadDsl`, we needed to remove all mentions of `Queryable` from the
trait itself, and move it to bounds on the impl instead. This means we
need a new trait that is generic over the type we're deserializing, and
changed `LoadDsl` to entirely base its bounds on that.

I've opted for the name `LoadQuery` instead of something like
`InternalLoadDsl`, as this trait is one that may end up in bounds for
user code, and will certainly show up in error messages.

I had to keep a bunch of redundant bounds on `LoadDsl` in order to make
the impls of `SaveChangesDsl` remain disjoint. This is because
associated types are basically ignored for purposes of disjointness.
While we know that `Update<T, T>: !LoadQuery<SqliteConnection, U>`,
since there is that second unconstrained type parameter, we cannot use
`LoadQuery` for disjointness either since someone could in theory write
a blanket impl of `impl<T, Conn> LoadQuery<Conn, MyType> for T`
(possibly slightly more specific than that).

Finally, as some bits for clarity, I've moved `Connection#query_one`
over to `LoadDsl`, since it had no real reason to exist on the
connection. I've renamed `query_all` to `query_by_index`, since that's
how it will differ from other methods starting with `query_`. Finally,
I've moved `LoadDsl#first` off to its own trait, since the bounds are
verbose to write if you wanted to call `.first` otherwise.

### DIFF
--- a/diesel/src/query_dsl/save_changes_dsl.rs
+++ b/diesel/src/query_dsl/save_changes_dsl.rs
@@ -1,72 +1,66 @@
 use associations::HasTable;
-use backend::SupportsReturningClause;
-use connection::Connection;
 use helper_types::*;
 use query_builder::{AsChangeset, IntoUpdateTarget};
 use query_dsl::*;
-use query_source::Queryable;
 use result::QueryResult;
-use types::HasSqlType;
 #[cfg(any(feature = "sqlite", feature = "mysql"))]
 use associations::Identifiable;
 
-pub trait SaveChangesDsl<Conn, ST> where
-    Conn: Connection,
-    Conn::Backend: HasSqlType<ST>,
-{
-    fn save_changes<T>(self, connection: &Conn) -> QueryResult<T> where
-        T: Queryable<ST, Conn::Backend>;
+pub trait InternalSaveChangesDsl<Conn, T>: Sized {
+    fn internal_save_changes(self, connection: &Conn) -> QueryResult<T>;
 }
 
-impl<T, ST, Conn> SaveChangesDsl<Conn, ST> for T where
-    Conn: Connection,
-    Conn::Backend: HasSqlType<ST> + SupportsReturningClause,
+impl<T, U, Conn> InternalSaveChangesDsl<Conn, U> for T where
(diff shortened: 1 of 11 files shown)


diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -1,72 +1,66 @@
...
+    fn internal_save_changes(self, connection: &Conn) -> QueryResult<T>;
 }
 
-impl<T, ST, Conn> SaveChangesDsl<Conn, ST> for T where
-    Conn: Connection,
-    Conn::Backend: HasSqlType<ST> + SupportsReturningClause,
+impl<T, U, Conn> InternalSaveChangesDsl<Conn, U> for T where
     T: Copy + AsChangeset<Target=<T as HasTable>::Table> + IntoUpdateTarget,
-    Update<T, T>: LoadDsl<Conn, SqlType=ST>,
+    Update<T, T>: LoadDsl<Conn> + LoadQuery<Conn, U>,
```
killercup: Doesn't LoadQuery already imply LoadDsl? Or is this one of the points where rustc plays dumb?

sgrif:
> Nice. Let's hope that's enough new internal traits for a while, though ;)

Darn that ruins my plans for a followup PR adding `InternalLoadDslInternals`

diesel/src/query_dsl/save_changes_dsl.rs
sgrif: `LoadQuery` implies `LoadDsl`, but we need to explicitly mention it here in order to make this impl disjoint from the sqlite/mysql ones.

PR #945, opened 2017-06-11
Add a general purpose `NOT` function

### LINKED ISSUE #944: Add a general purpose not function
I had a very short discussion with @sgrif about this in gitter and he asked me to open an issue for it.

Right now, there does not seem to be a way to make a query for finding all rows where an array does not contain a different array. Something to do the opposite of this:
```rust
files::table.filter(files::tags.contains(tags))
```

I assume adding a `filter_not` function that generates a `WHERE NOT` query would be the most general way of implementing this.

sgrif:
I was rather surprised that we didn't already have this. I didn't stick
this in `expression/predicates.rs`, as we didn't already have a macro
for prefix operators, and I want to refactor that to add it later. I may
move this to that macro in the future, and just have the return type be
`Not<Grouped<T::Expression>>` to reduce the amount of required code
here.

We need parenthesis in the resulting SQL to ensure that it always
applies to its arguments. `not(true.and(false))` should return `true`.
However, `NOT true AND false` is equivalent to `(NOT true) AND false`.

Fixes #944.

### DIFF
--- a/diesel/src/expression/not.rs
+++ b/diesel/src/expression/not.rs
@@ -1,0 +1,72 @@
+use expression::*;
+use query_builder::*;
+use result::QueryResult;
+use types::Bool;
+
+/// Creates a SQL `NOT` expression
+///
+/// # Example
+///
+/// ```rust
+/// # #[macro_use] extern crate diesel;
+/// # include!("src/doctest_setup.rs");
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+/// #
+/// # fn main() {
+/// #     use self::users::dsl::*;
+/// #     let connection = establish_connection();
+/// use diesel::expression::not;
+///
+/// let users_with_name = users.select(id).filter(name.eq("Sean"));
+/// let users_not_with_name = users.select(id).filter(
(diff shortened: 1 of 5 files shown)


Eijebong:
Would `users.filter(not(name.ilike("%coucou")))` do the same thing as `users.filter(name.not_ilike("%coucou"))` ?

diesel/src/expression/not.rs
```diff
@@ -0,0 +1,72 @@
...
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+/// #
+/// # fn main() {
+/// #     use self::users::dsl::*;
+/// #     let connection = establish_connection();
+/// use diesel::expression::dsl::not;
```
Eijebong: Why not `use diesel::expression::not` ? That's what rust told me to use when it was missing

sgrif: Good point. Will fix.

sgrif:
I'd prefer to keep `ne`, `not_like`, `not_between`, etc in cases where there is an explicit operator, since that's what people will be looking for. I even want to keep `ne_any` even though it's literally `not(eq_any`, since I think most people think of `NOT IN` as a single operator and not as `NOT(IN(...))`

PR #946, opened 2017-06-11
Refactor `infix_predicate!`, add a prefix version

sgrif:
I've refactored `infix_predicate!` and `postfix_predicate!` to share
nearly all of their code, and added a prefix version. I've moved `not`
to use the generated operator, and renamed all of these to names I'm
comfortable with for 1.0. I want a `diesel_` prefix since these aren't
likely to be called from user code much, only plugins. I like `operator`
better here, since it's not a predicate if the return type is something
other than bool, and it's not an expression if the return type is ().

As a result of this change, `diesel_postfix_operator!` (previously
`postfix_predicate!`) gained all the functionality it was missing that
the infix form has (specifically, the ability to specify a backend).

Most of the macro was actually extremely easy to share. The only
difference between `infix_predicate_body` and `postfix_predicate_body!`
for most of the code was the number of type parameters, and the names of
the fields. This is easily solved with macro repeaters.

The two main places where the code had to differ I've moved to seprate
submacros. The first was the behavior of the body of `walk_ast`, which
needed to change based on whether it was prefix, postfix, or infix.
Secondly, if we're generating the `QueryFragment` impl for a specific
backend rather than all backends, we need to generate a second impl for
the `Debug` backend as well.

Both of these were pretty easy to solve. The resulting code is perhaps a
bit more gnarly if you're not used to macros, but I think it's better
composed than the `global_infix_predicate_to_sql!` stuff we had before.


diesel/src/expression/operators.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+    };
+
+    // At least one backend_ty_param was given. We generated a generic impl
+    // which will cover `Debug`.
+    (backend_ty_params = ($($backend_ty_params:ident,)+), $($rest:tt)*) => {};
+}
+
+/// Useful for libraries adding support for new SQL types. Apps should never
+/// need to call this
```
killercup: Period.

diesel/src/expression/operators.rs
```diff
@@ -0,0 +1,392 @@
...
+/// If the operator is specific to a single backend, you can specify this by
+/// adding `backend: Pg` or similar as the last argument.
+///
+/// It should be noted that the generated impls will not constrain the SQL
+/// types of the arguments. You should ensure that they are of the right
+/// type in your function which constructs the operator.
+///
+/// Typically you would not expose the type that this generates directly. You'd
+/// expose a function (or trait) used to construct the expression, and a helper
+/// type which represents the return type of that function. See the
```
killercup: > See the [source of]

diesel/src/expression/operators.rs
```diff
@@ -0,0 +1,392 @@
...
+            ty_params = (T, U,),
+            field_names = (left, right,),
+            backend_ty_params = (),
+            backend_ty = $backend,
+        );
+    };
+}
+
+/// Useful for libraries adding support for new SQL types. Apps should never
+/// need to call this
```
killercup: Period.

diesel/src/expression/operators.rs
```diff
@@ -0,0 +1,392 @@
...
+            ty_params = (Expr,),
+            field_names = (expr,),
+            backend_ty_params = (),
+            backend_ty = $backend,
+        );
+    };
+}
+
+/// Useful for libraries adding support for new SQL types. Apps should never
+/// need to call this
```
killercup: Period? Period.

sgrif:
> we need to remember to update diesel_full_text_search though.

Already have -- waiting to merge this to push so I can point it at git instead of a path dependency

diesel/src/expression/operators.rs
sgrif: Apps should never need to call this. Period.

I like it. It's forceful.

killercup: Gives a whole other meaning to 'should'

PR #948, opened 2017-06-11
Remove `#[has_many]`, it has become useless

sgrif:
This attribute was originally required back when we were generating
brute force impls of `SelectableExpression` for every column on both
sides of a join. That went away a few versions back, and this attribute
became mostly useless (in 0.13 you can do `child.joins(parent)` with
just `BelongsTo`, but `parent.joins(child)` requires the `has_many`).

One accidental side effect of how I implemented multi-table joins was
that the `has_many` became mandatory again. This was very unfortunate
(and means that 0.14 will generate really confusing error messages
during the upgrade) so I've just gone ahead and removed it entirely.

I expected this to be a lot simpler than it actually was. The problem is
that `#[belongs_to]` does not directly have access to the parent table.
The only way it can access it is by doing `<User as HasTable>::Table`.
Even though that is a fully monomorphic projection on a local type,
where the result is also a local type, Rust does not currently consider
this to be a local type (I think this might be a bug or oversight).

So to work around this, we instead implement the child to parent join
twice, wrapping the parent table in a marker struct for the second impl.
We then have blanket impls for every table that look for this marker,
and generate the inverse impl that we actually want.

It's a kludge, but ultimately an invisible one. The blanket impls need
to be in the generated code for tables and not in Diesel core, since if
you replace the parameter to `PleaseGenerateInverseJoinImpls` with a
type parameter, Rust will start checking infinitely recursive forms of
`PleaseGenerateInverseJoinImpls<PleaseGenerateInverseJoinImpls<...>>`
whenever a projection on that parameter is evaluated.

The problem with the impls being in the generated code is that
downstream crates cannot rely on the fact that `SelectStatement` and
friends are not tables. This was resolved by simply not having a `Table`
constraint, which lets us remove a bunch of impls in Diesel core
instead.


diesel/src/query_source/joins.rs
```diff
@@ -292,6 +261,22 @@ impl<T, Join, On> AppearsInFromClause<T> for JoinOn<Join, On> where
     type Count = Join::Count;
 }
 
+#[allow(missing_debug_implementations, missing_copy_implementations)]
+#[doc(hidden)]
+/// A hack to allow bidirectional joins to be generated from `#[belongs_to]`
+///
+/// This type needs to exist because it is illegal in Rust today to write
```
Eijebong: Will this ever be possible ? If so, maybe add a FIXME here ?

diesel/src/associations/mod.rs
```diff
@@ -47,24 +46,19 @@
 //! # }
 //! ```
 //!
-//! Note that in addition to the `#[has_many]` and `#[belongs_to]` annotations, we also need to
+//! Note that in addition to the `#[has_many]` annotation, we also need to
```
Eijebong: Hu ? You remove `has_many`, not `belongs_to`

diesel/src/query_source/joins.rs
```diff
@@ -292,6 +261,22 @@ impl<T, Join, On> AppearsInFromClause<T> for JoinOn<Join, On> where
...
+/// `impl JoinTo<posts> for <User as HasTable>::Table`, even though the type
+/// is fully monomorphic and projects to a local type. If this restriction
+/// were ever lifted in the future, this type could be removed.
+///
+/// Instead, after generating `impl JoinTo<<User as HasTable>::Table> for
+/// posts`, we *also* generate `impl JoinTo<PleaseGenerateInverseJoinImpls<<User
+/// as HasTable>::table> for posts`, and rely on the fact that `users::table`
+/// will have a blanket impl on itself for anything that joins to
+/// `PleaseGenerateInverseJoinImpls`.
+pub struct PleaseGenerateInverseJoinImpls<T>(T);
```
Eijebong: That struct name though 😆

diesel/src/associations/mod.rs
```diff
@@ -47,24 +46,19 @@
...
 //! [Identifiable]: trait.Identifiable.html
 //!
 //! If the name of your foreign key doesn't follow the convention `tablename_id`, you can specify a
-//! custom one to `#[has_many]` and `#[belongs_to]` by adding a `foreign_key` argument to the
-//! attribute like so `#[has_many, foreign_key="mykey"]`.
-//!
-//! `#[has_many]` actually has no behavior on its own. It only enables joining between the two
-//! tables. If you are only writing `children.inner_join(parents)` or
-//! `Child::belonging_to(&parents)`, you only need to define the `#[belongs_to]` side.
+//! custom one `#[belongs_to]` by adding a `foreign_key` argument to the
```
Eijebong: custom on to `#[belongs_to]`

diesel/src/query_source/joins.rs
sgrif: I'm happy to put a `ThankYou` at the end if you think the current name is too rude

diesel/src/query_source/joins.rs
sgrif: I don't know. I need to discuss w/ some folks. If nothing else, I don't think this doc comment is likely to get out of sync with the code.

diesel/src/associations/mod.rs
```diff
@@ -47,24 +46,19 @@
 //! # }
 //! ```
 //!
-//! Note that in addition to the `#[has_many]` and `#[belongs_to]` annotations, we also need to
+//! Note that in addition to the `#[belongs_to]` annotation, we also need to
 //! `#[derive(Associations)]`
 //!
-//! `#[has_many]` should be passed the name of the table that the children will be loaded from,
-//! while `#[belongs_to]` is given the name of the struct that represents the parent. Both types
+//! `#[belongs_to]` is given the name of the struct that represents the parent. Both types
```
killercup: "Both types" means "the annotated struct and the parent", right? Maybe write that out?

CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 [not-0.14.0]: http://docs.diesel.rs/diesel/expression/dsl/fn.not.html
 
+### Removed
+
+* `#[has_many]` has been removed. Its functionality is now provided by
+  `#[belongs_to]` on the child struct. If there is no child struct to
+  put `#[belongs_to]` on, you can invoke `joinable!` directly instead.
```
killercup: Maybe link to the joinable! docs (once that macro os no longer hidden in the docs)?

diesel/src/query_source/joins.rs
withoutboats: This should be possible in a post-chalk world. Would not be possible if `User` was a type parameter, but after chalk we should be able to normalize this to `users` or w/e.

sgrif: Yeah, that was why I haven't opened an issue. I figured the answer would either be "this is intentional" or "this is blocked on chalk". :D

PR #957, opened 2017-06-19
Add some docs to infer schema generated code

killercup:
Now that #849 is merged (:tada:), let's add some documentation to the generated schema!


diesel_infer_schema/src/codegen.rs
```diff
@@ -90,7 +99,18 @@ fn column_def_tokens(
     if column_type.is_nullable {
         tpe = quote!(Nullable<#tpe>);
     }
-    Ok(quote!(#column_name -> #tpe))
+
+    let docs = format!("The `{}` column of the `{}` table.\
+        \n\n\
+        Its SQL type is `{}`.
```
killercup: A follow-up PR could link some easily identifiable types to their docs

killercup:
Ah, well, restricting the macro to doc comments disallows `#[doc]` attributes. I'll work on that.

weiznich:
The macro is able to parse this, because we disallowed every attribute but  `#[doc=…]`.
There are some tests are failing. (diesel_cli::print_schema).

killercup:
> The macro is able to parse this

Yes, my bad. I wasn't able to parse my own `Cargo.toml` content ;)

> There are some tests are failing. (diesel_cli::print_schema).

Yep, I'll try to make the codegen produce nice doc comments (instead of the attributes).

@dtolnay, is there a way to write a `quote!` call that results in something like `/// foo bar`? (Incl. substitutions)

To get nice output when showing the generated code (we do that in the `diesel print-schema` CLI tool), I would like to write something like

```rust
quote!(table! {
    /// Representation of the `#table_path` table
    ///
    /// (Generated by Diesel's `infer_schema!` macro.)
    #table_path (#(#primary_keys),*) {
        #(#tokens),*,
    }
})
```

but the `#table_path` isn't substituted. (It doesn't have to be exactly this code, just something that results in the doc comment.)

Otherwise, we'll have to expand our pretty printing (no big deal, but would be nice to not have to do this ;)).

killercup:
Over lunch, I managed to get it this far:

```rust
table! {
    /// Representation of the `balances` table.
 ///
 /// (Generated by Diesel's `infer_schema!` macro.)
 balances (id) {
        /// The `id` column of the `balances` table.
 ///
 /// Its SQL type is `Integer`.
 ///
 /// (Generated by Diesel's `infer_schema!` macro.)
 id -> Integer,
        /// The `created_at` column of the `balances` table.
 ///
 /// Its SQL type is `Text`.
 ///
 /// (Generated by Diesel's `infer_schema!` macro.)
 created_at -> Text,
 …
    }
}
```

So, 'only' some minor indentation adjustments to the pretty printer (and test fixes) left to do!

@weiznich, I'll probably not get around to this before tomorrow evening. As you wrote most of the pretty printer, feel free to chime in on this PR :)

Also unpinging @dtolnay, I managed to work out how to do it (there might still be a better way).

killercup:
Rebased and fixed recursion limit fun (see last commit).

r?

Eijebong:
So if I understand correctly, `diesel print-schema` will also print those comments ?
Can we have something like a flag to disable that behavior ?

killercup:
Good idea, @Eijebong. Shouldn't be too hard, just a matter of passing one more flag down to the expand fns. Do you think we should do that here or in a follow-up PR?

Eijebong:
No opinion on that

killercup:
Okay, then let's do is in another PR, this one's already large enough.

r? @sgrif

sgrif:
Should this wait until after #1080 ?

killercup:
I don't mind rebasing this, it's probably easier than updating #1080

Sean Griffin <notifications@github.com> schrieb am Mo. 7. Aug. 2017 um
18:33:

sgrif:
Sorry that was Ruby

killercup:
Okay, ~~rebased~~ rewritten on top of #1080. This now also fixes/refactors some related parts of the code. Let's see what CI says! (My local mysql setup has apparently gone up in flames during a recent update.)

diesel_cli/src/print_schema.rs
```diff
@@ -100,21 +107,10 @@ impl<'a> Display for ColumnDefinitions<'a> {
...
-                if column.ty.is_array {
-                    write!(out, ">")?;
-                }
-                if column.ty.is_nullable {
-                    write!(out, ">")?;
+                for d in column.docs.lines() {
+                    writeln!(out, "///{}{}", if d.is_empty() { "" } else { " " }, d)?;
                 }
-                writeln!(out, ",")?;
+                writeln!(out, "{} -> {},", column.name, column.ty)?;
```
Eijebong: Nice :)

diesel_infer_schema/src/inference.rs
```diff
@@ -146,20 +146,42 @@ pub fn load_foreign_key_constraints(database_url: &str, schema_name: Option<&str
     }
 }
 
+macro_rules! doc_comment {
```
Eijebong: I think this is fine but @sgrif already emitted concerns about using macros for so few calls (2 in this case)

diesel_cli/src/pretty_printing.rs
```diff
@@ -1,183 +0,0 @@
-use std::fmt::{Write, Error as FmtError};
```
Eijebong: I guess this file wasn't removed in the PR that changed how pretty printing was done ?

killercup: Yep, as mentioned in the commit message, #1080 made this obsolete :)

diesel_infer_schema/src/inference.rs
killercup: I would've made it a regular fn but the first arg would've always been `format!(…)`, so I made it a format-like macro

killercup:
Thanks for the review. Will merge this and then begin a PR to make the comment optional in the CLI output

PR #959, opened 2017-06-21
Allow trailing commas in sql_function! arguments list

Eijebong:


killercup:
cc https://github.com/rust-lang-nursery/rustfmt/issues/1739

PlasmaPower:
@killercup The latest commit tests trailing commas by modifying an existing test.

killercup:
@PlasmaPower thanks, I totally missed that!

Ideally, I want to have a larger discussion about how to design our macros in general, but that is not a pressing matter right now. `sql_function!` specifically does not seem like one that is usually written over multiple lines, though. @Eijebong, was this just triggered by the rustfmt bug? I'm fine with adding this, but I don't really want to stabilize/guarantee that this macro parses ≥1 trailing comma.

Eijebong:
Yeah, this was triggered by the rustfmt bug (which is now fixed and released). Maybe we can close this since I don't think there is a way to guarantee that we only parse <= 1 trailing comma without duplicating every declaration.

killercup:
I've opened #977 to get a general discussion going. I'm going to close this for now, but depending on what we decide, we can reopen and merge this in the future.

PR #960, opened 2017-06-21
Bump chrono to 0.4

Eijebong:
This allow the usage of serde 1.0 with diesel again since chrono 0.3.1 has been
yanked.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["database"]
 
 [dependencies]
 byteorder = "1.0"
-chrono = { version = "0.3", optional = true }
+chrono = { version = "0.4", optional = true }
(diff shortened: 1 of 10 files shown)


killercup:
Nice, saw the chrono 0.4 announcement earlier and thought we should bump the dep in diesel asap :)

Should we enable chrono's serde feature when our serde feature is enabled? Can we?

quodlibetor:
@killercup you mean conditionally depending on chrono's serde feature if Diesel's serde is enabled? I think that would be reasonable, reducing the "why isn't serde working" stumbling block.

[It seems like it's possible](https://github.com/rust-lang/cargo/issues/633#issuecomment-59089252) ([stackoverflow example](https://stackoverflow.com/a/41440621/25616)).

PR #961, opened 2017-06-21
Activate Chrono's serde feature in extras

quodlibetor:
AFAICT #rust-lang/cargo/633 doesn't allow _only_ activating a feature.
Activating a downstream feature means that you unconditionally pull it into the
crate.

That means we can't do:

```
[features]
serde_json = ["chrono/serde"]
```

because it would always pull chrono in even if it's unused.

This was spurred by #960

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -35,7 +35,7 @@ tempdir = "^0.3.4"
 
 [features]
 default = ["with-deprecated"]
-extras = ["chrono", "serde_json", "uuid", "deprecated-time", "network-address"]
+extras = ["chrono/serde", "serde_json", "uuid", "deprecated-time", "network-address"]
 unstable = []
 lint = ["clippy"]
 large-tables = []


killercup:
Thanks! Let's talk a bit more about the use case of this. I think I was a bit quick to suggest this.

Currently, we can convert to/from `serde_json::Value` and `chrono::DateTime`. Can the two ever interact in a way that the user can omit chrono's serde feature in their own `Cargo.toml`? I don't think so. This would mean automatically activating this would be superfluous in some cases, and may actually be confusing in others (features are additive and diesel silently activating this may be confusing in a "why does this even work?" way).

Very interested in your opinions!

quodlibetor:
I'm not actually sure how the mapping from `serde_json::Value` -> `chrono::DateTime` works right now, is it mostly the `To/FromSql` traits? Looking through diesel's chrono code I don't see any specific benefit to enabling the serde feature. It would mostly just enable the `#[derive...]` calls for upstream, but if users are using chrono then they probably should just enable the feature themselves.

So, yeah, I don't see a a tremendous amount of benefit to auto-enabling the serde feature in chrono if _not_ having it is fine. This PR was largely a "oh, huh, that _does_ work" situation.

killercup:
> This PR was largely a "oh, huh, that does work" situation.

And that's totally fine 😄 

> It would mostly just enable the #[derive...] calls for upstream, but if users are using chrono then they probably should just enable the feature themselves.

Yeah I too think that's the only thing happening with the PR. And user's should do that explicitly.

I'll close this PR for now, but feel free to ping me/reopen when you can think of another reason to have this :)

quodlibetor:
I agree with that decision 👍

PR #962, opened 2017-06-22
Updated Chrono to 0.4.0

beefsack:
0.3.1 was yanked this morning so this is the only available version of Chrono supporting Serde 1.0.


beefsack:
Dupe of #960.

PR #981, opened 2017-06-30
Add a rust-toolchain file

Eijebong:
Since rustup now supports overrides via a rust-toolchain file, I think
it makes sense to have it since clippy breaks very often and it prevents
us from running the test suite.


killercup:
Seems like a good idea in general, even if it's just to show people what nightly we run on CI for clippy.

We need to make sure to update both this and travis.yml at the same time in the future, though.

sgrif:
Reverting since this broke Appveyor.

PR #985, opened 2017-07-02
on appveyor, install rust version specified in rust-toolchain file

mcgoo:
@Eijebong, not sure if this is what you intended, but this will run all appveyor builds using the Rust version specified in the rust-toolchain file, falling back to stable if that file is not present.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,7 +17,8 @@ cache:
 
 install:
   - curl -fsS --retry 3 --retry-connrefused -o rustup-init.exe https://win.rustup.rs/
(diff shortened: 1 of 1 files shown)


Eijebong:
Nah the rust-toolchain file is only here for developers since it tells rustup to use that particular rust version (allowing devs to run tests without disabling/updating clippy).
I think appveyor should still run stable only.

@killercup btw, #981 broke appveyor since it's using rustup (travis isn't).

mcgoo:
Hmm, maybe just make the appveyor script delete it before it does anything then?

PR #990, opened 2017-07-04
Automatically generate `numeric_expr!` for known numeric types

### LINKED ISSUE #966: Document `column + 1` needs `numeric_expr!(column)`
See https://gitter.im/diesel-rs/diesel?at=594d2b24743aa6f70890dad8

Full example:

```rust

use diesel::prelude::*;

table! {
    files(id) {
        id -> Integer,
        views -> Int8,
    }
}

fn main() {
    // Help people find the next two lines!
    numeric_expr!(files::id);
    numeric_expr!(files::views);

    files::table.select(files::id + 1);

    diesel::update(files::table.filter(files::id.eq(0)))
            .set(files::views.eq(files::views + 1))
            .returning(files::views);
}
```

sgrif:
This was a much larger pain in the ass then I expected. It turns out
that capturing a token as a `ty` doesn't just prevent breaking it back
apart later, it prevents any matches against it period. So we have to
carry the ty through as a series of `tt` until the very end.

The problem is that this is impossible to do unambiguously, and we
can't even reliably get just the "common" cases since ty params use `<>`
which aren't a single token tree.

So this implementation basically does a best effort case to match the
"common" cases, and then falls back to just capturing in a `ty`
fragment afterwards. For the cases that don't match `$($x:tt)::*
$(<$($y:tt)::*>)*`, they would never be numeric anyway (we will need to
increase the nesting to 3 for the ty params once unsigned comes back
though, since `Nullable<Unsigned<Integer>>` will be a thing)

Ideally I'd like to eventually revert this, and instead be able to write
the impl with a `where Self::SqlType: types::ops::Add`, but that is
currently invalid in Rust.

Close #966.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## Unreleased
+
+### Changed
+
+* Diesel will now automatically invoke `numeric_expr!` for your columns in the
+  common cases. You will likely need to delete any manual invocations of this
+  macro.
+
 ## [0.14.0] - 2017-07-04
 
 ### Added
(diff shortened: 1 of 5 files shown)


sgrif:
Added it to the changelog entry while rebasing.

PR #991, opened 2017-07-04
Don't magically set up timestamp helpers, create a migration file

### LINKED ISSUE #522: Migrations not supported on CockroachDB

sgrif:
This file is where our initial "plumbing" can go. At the moment we only
provide it for PG, and its only behavior is safe to add to existing
projects.

However, since we're placing this file in other people's projects, and
in a structure where once it's run we can't modify it, we should talk
about how this file will evolve over time if needed. If we ever do need
to add more stuff here, I'd expect this file to just be changed for new
projects. For existing projects, I'm thinking we'd provide a `diesel
update` command which checks a metadata table with the last version
`diesel update` was run against. It will then generate new migrations
(versioned normally, we don't want to affect existing tables) to bring
everything up to speed.

Fixes #522.
Close #848.


sgrif:
Based on the value of the metadata table I mentioned previously

Eijebong:
Sure, but since we're putting everything in one file, does it mean we'll have to parse this file when running  `diesel update` ?

sgrif:
`diesel update` wouldn't modify this file. It would generate a new migration that has all of the changes between "current metadata version entry" and "version of diesel_cli being run"

Eijebong:
How do you retrieve the version of the file at the "current metadata version entry" ?

sgrif:
I'm not sure I understand what you're asking. We would set up a file that is "here's how to migrate from version 1 to version 2". If we read the table and see that the version is less than 2 (or the table doesn't exist), we will generate a migration with that file. We'd change the file that is generated for new projects to already have those changes in the initial migration.

Eijebong:
Ohh I think I misunderstood your comment. I thought you wanted every setup SQL to be in one file in diesel...

sgrif:
Only for new projects. We cannot modify this file for existing projects.

Eijebong:
Yep, it all makes sense now.

sgrif:
@killercup I'm going to hold off on merging this until you've had a chance to review since you had some strong opinions about this change.

diesel_cli/src/setup_sql/postgres/initial_setup/up.sql
```diff
@@ -0,0 +1,36 @@
+-- This file was automatically created by Diesel to setup helper functions
+-- and other internal bookkeeping. This file is safe to edit, any future
+-- changes will be added to existing projects as new migrations.
```
killercup: Isn't it "set up" and "book keeping" here?

diesel_cli/tests/setup.rs
```diff
@@ -28,6 +31,33 @@ fn setup_creates_migrations_directory() {
...
+
+    let result = p.command("setup").run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(p.has_file(Path::new("migrations").join("00000000000000_diesel_initial_setup")));
+}
+
+#[test]
+#[cfg(feature="postgres")]
+fn setup_creates_default_migration_file_if_project_is_otherwise_setup() {
```
killercup: This right here is all I ever wanted! 😍

killercup: This right here is all I ever wanted! 😍

diesel_cli/src/setup_sql/postgres/initial_setup/up.sql
sgrif: [no](https://www.merriam-webster.com/dictionary/setup) and [no](https://www.merriam-webster.com/dictionary/bookkeeping) English is about as consistent as PHP

sgrif:
> The workflow cockroach will be to run diesel setup, and empty out the file, right?

Pretty much. I'm hesitant to actually call out a specific workflow for that case, since we don't "officially" support it. But yes, that is what I intend for those folks.

carols10cents:
>  workflow cockroach 

wait wut

🐜🐛🐞🕷

killercup:
> 🐜🐛🐞🕷

such resilient
very flow
much working

PR #994, opened 2017-07-05
[WIP] Feature/infer enums

### LINKED ISSUE #343: Support enumerated types

weiznich:
follow up for #940 

* This will only work with postgres in the current version

This is a first draft version to get some feedback on this. 
Things to fix before merging:
- [ ] Formating
- [ ] Documentation
- [ ] Tests
- [x] Fix compiling other backends

Using an implementation similar to the shown, will restrict the produced binary to exactly the same database that was used to compile the code because the type oid's of the enums are hardcoded in `HasSqlType::metadata`. As far as I understand those oid's could/will change if you run your migrations on a other database instance, so the provided mapping will be invalid.

As far as I see there are two possible solutions:
- Document and accept the behavior 
- Add some way to query the database on startup/connect for the actual oid's and store them in some kind of static map. Then use this map to generate the metadata 

Fixes #343

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -25,6 +25,7 @@ time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.6.0", optional = true, features = ["use_std"] }
 ipnetwork = { version = "0.12.2", optional = true }
+chashmap = "2.2"
 
 [dev-dependencies]
 cfg-if = "0.1.0"
--- a/diesel/src/backend.rs
+++ b/diesel/src/backend.rs
@@ -4,9 +4,11 @@ use query_builder::QueryBuilder;
 use query_builder::bind_collector::{BindCollector, RawBytesBindCollector};
 use query_builder::debug::DebugQueryBuilder;
 use types::{self, HasSqlType};
+use result::QueryResult;
 
 pub trait Backend where
     Self: Sized,
+    Self: TypeMetadata,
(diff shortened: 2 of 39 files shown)


sgrif:
Can you remove all the enum stuff from this PR? I'd like to focus *only* on the runtime lookup of OIDs for now

weiznich:
Opened #995 for the runtime lookup part.

PR #995, opened 2017-07-05
Feature/runtime oid lockup

weiznich:
Implement a mechanism to lookup type oids at runtime.


sgrif:
So this strays pretty far from what I'd laid out in https://github.com/diesel-rs/diesel/pull/940#issuecomment-307481087. There's a lot of other issues, the biggest of which being that it doesn't compile, and has no tests. I decided that it'd be less work to just finish up the branch I had for this rather than work on cleaning this PR up. Closing in favor of #996

PR #996, opened 2017-07-05
Allow lookup of OIDs at runtime

sgrif:
This lays all the groundwork for a custom type (or type implementing an extension) to perform an OID lookup at runtime. I'm not interested in supporting any extensions or user defined types in Diesel proper, but I do want to see third party crates provide that capability. This change is required for those crates to be able to exist.

Ultimately the root issue that this PR needs to solve is that `TypeMetadata::metadata` needs to have access to the database connection for PG. Of course I don't want to expose the *actual* connection, since that function should not have the capability to execute arbitrary queries. This does mean that we have to make the wrapper struct a little funky, as putting a lifetime on it would require either adding a lifetime to `Pg` or `TypeMetadata`, neither of which are things I'd like to do. We can refactor the funky `new` signature away once ATCs land.

To keep the scope of the PR small, I have explicitly omitted error handling and caching from this implementation. They will be done in a follow-up PR.

I've tried to implement this in a way that has as little unnecessary churn as possible (e.g. wrapping the parameter of `ToSql` instead of passing a second parameter so we only have to change definitions but not call sites).

### DIFF
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -1,0 +1,32 @@
+use std::mem;
+
+use prelude::*;
+use super::{PgConnection, PgTypeMetadata};
+
+#[allow(missing_debug_implementations)]
+pub struct PgMetadataLookup {
+    conn: PgConnection,
+}
+
+impl PgMetadataLookup {
+    pub(crate) fn new(conn: &PgConnection) -> &Self {
+        unsafe { mem::transmute(conn) }
+    }
+
+    pub fn lookup_type(&self, type_name: &str) -> PgTypeMetadata {
+        use self::pg_type::dsl::*;
+
+        pg_type.select((oid, typarray))
+            .filter(typname.eq(type_name))
+            .first(&self.conn)
+            .unwrap_or_default()
+    }
+}
+
+table! {
+    pg_type (oid) {
+        oid -> Oid,
+        typname -> Text,
+        typarray -> Oid,
+    }
+}
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -275,10 +275,10 @@ pub type VarChar = Text;
 pub use pg::types::sql_types::*;
 
 pub trait HasSqlType<ST>: TypeMetadata {
-    fn metadata() -> Self::TypeMetadata;
+    fn metadata(lookup: &Self::MetadataLookup) -> Self::TypeMetadata;
 
-    fn row_metadata(out: &mut Vec<Self::TypeMetadata>) {
-        out.push(Self::metadata())
+    fn row_metadata(out: &mut Vec<Self::TypeMetadata>, lookup: &Self::MetadataLookup) {
+        out.push(Self::metadata(lookup))
     }
 }
 
@@ -345,26 +345,27 @@ pub struct ToSqlOutput<'a, T, DB> where
     DB::MetadataLookup: 'a,
 {
     out: T,
-    marker: ::std::marker::PhantomData<&'a DB::MetadataLookup>,
+    metadata_lookup: &'a DB::MetadataLookup,
 }
 
 impl<'a, T, DB: TypeMetadata> ToSqlOutput<'a, T, DB> {
-    pub fn new(out: T) -> Self {
-        ToSqlOutput {
-            out,
-            marker: ::std::marker::PhantomData,
-        }
+    pub fn new(out: T, metadata_lookup: &'a DB::MetadataLookup) -> Self {
+        ToSqlOutput { out, metadata_lookup }
     }
 
     pub fn with_buffer<U>(&self, new_out: U) -> ToSqlOutput<'a, U, DB> {
         ToSqlOutput {
             out: new_out,
-            marker: self.marker,
+            metadata_lookup: self.metadata_lookup,
         }
     }
 
     pub fn into_inner(self) -> T {
         self.out
+    }
+
+    pub fn metadata_lookup(&self) -> &'a DB::MetadataLookup {
+        self.metadata_lookup
     }
 }
 
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -20,6 +20,7 @@ pub mod impls;
 mod fold;
 
 use std::fmt;
+use std::ops::{Deref, DerefMut};
 
 #[doc(hidden)]
 pub mod structs {
@@ -43,7 +44,7 @@ pub use self::fold::Foldable;
 use backend::{Backend, TypeMetadata};
 use row::Row;
 use std::error::Error;
-use std::io::Write;
+use std::io::{self, Write};
 
 /// The boolean SQL type. On SQLite this is emulated with an integer.
 ///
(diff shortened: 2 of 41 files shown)


diesel/src/types/mod.rs
```diff
@@ -274,10 +275,10 @@ pub type VarChar = Text;
 pub use pg::types::sql_types::*;
 
 pub trait HasSqlType<ST>: TypeMetadata {
-    fn metadata() -> Self::TypeMetadata;
+    fn metadata(lookup: &Self::MetadataLookup) -> Self::TypeMetadata;
```
weiznich: The `PgTypeMetadata::Dynamic`/`PgTypeMetadata::Static` approach does not require to change this for all backends. So this may be preferred? Or is this required for any future feature on other backends?

diesel/src/types/mod.rs
```diff
@@ -337,18 +338,100 @@ pub enum IsNull {
     No,
 }
 
+#[derive(Debug, Clone, Copy)]
+#[doc(hidden)]
+pub struct ToSqlOutput<'a, T, DB> where
```
weiznich: In my opinion this struct wraps two totally unrelated things. The only common point here is that both parts are passed to `ToSql::to_sql`. I don't see any advantage over passing those fields directly as arguments to `to_sql`. 
Creating this wrapping struct needs ~100 lines boilerplate code for nothing (as far as I understand).

weiznich:
It would have been somewhat easier if for me to implement this, with a clearer perspective how to get this merged and clear points what exactly needs to be fixed (And why in a certain way;)).

Anyhow, I'm happy to see some movement on this topic.

diesel/src/pg/metadata_lookup.rs
```diff
@@ -0,0 +1,32 @@
+use std::mem;
+
+use prelude::*;
+use super::{PgConnection, PgTypeMetadata};
+
+#[derive(Debug)]
+pub struct PgMetadataLookup {
```
weiznich: This could not implement `Debug` because `PgConnection` does not implement `Debug`

diesel/src/types/mod.rs
```diff
@@ -337,18 +338,100 @@ pub enum IsNull {
...
+    /// Returns a `ToSqlOutput` suitable for testing `ToSql` implementations.
+    /// Unsafe to use for testing types which perform dynamic metadata lookup.
+    pub fn test() -> Self {
+        use std::mem;
+        Self::new(Vec::new(), unsafe { mem::uninitialized() })
+    }
+}
+
+impl<'a, T: Write, DB: TypeMetadata> Write for ToSqlOutput<'a, T, DB> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
```
sgrif: There are places that it is passed to a function which expects something which implements `Write`. I could manually add `&mut *out` to those places, but the goal of this version is to avoid churning lines which don't need to change.

diesel/src/types/mod.rs
sgrif: > I don't see any advantage over passing those fields directly as arguments to to_sql.

Well there's two main points (I explained some of this in the PR description, not sure if you read it). The big one is that less existing code will need to change this way. If we add a second parameter, we'll need to modify every definition *and* every use of it. The second reason I'm preferring this is that I don't want to have to modify all this code *again* when we realize we needed a third or a fourth parameter.

> Creating this wrapping struct needs ~100 lines boilerplate code

Lines of code is not a valuable metric to measure things against.

diesel/src/types/mod.rs
sgrif: You're right, either one could work. I felt that this version was more flexible, and required less churn.

weiznich: As long as one ignores the error handling this version seems to be easier. 
As soon as you add error handling you will need to touch `HasSqlType` again, because the lookup will return an `Result`. Therefore `HasSqlType::metadata` will also return an `Result`. So each Implementation and each call of `HasSqlType::metadata` will be touched again.

Using the `Dynamic/Static` approach the lookup is done inside of `ToSql` (already returns `Result`) or inside of `BindCollector::push_bound_value` (already returns `Result`). So only an `try!` around those lookup calls will be needed.

sgrif: I'll take another look

diesel/src/types/mod.rs
weiznich: Makes sense.

According to the [rust-api-guidelines](https://github.com/brson/rust-api-guidelines#c-deref) only smart pointers should implement `Deref`/`DerefMut`. Furthermore `ToSql` only requires `Write`, so maybe we should drop the `Deref` implementations?

diesel/src/types/mod.rs
sgrif: I implemented this and compared the two versions. I decided that I did not like the enum form. Reasoning laid out in the commit message. https://github.com/diesel-rs/diesel/commit/2a84869c38b0d69bddf86021029e5c003c7e74f0

diesel/src/types/mod.rs
```diff
@@ -337,18 +338,109 @@ pub enum IsNull {
...
+
+impl<'a, T, DB: TypeMetadata> ToSqlOutput<'a, T, DB> {
+    pub fn new(out: T, metadata_lookup: &'a DB::MetadataLookup) -> Self {
+        ToSqlOutput { out, metadata_lookup }
+    }
+
+    pub fn with_buffer<U>(&self, new_out: U) -> ToSqlOutput<'a, U, DB> {
+        ToSqlOutput {
+            out: new_out,
+            metadata_lookup: self.metadata_lookup,
```
killercup: Future-proof this even more against diff churn with `... self }` instead of listing the other fields (like `metadata_lookup` here)

diesel/src/types/mod.rs
sgrif: Yes, deref coercions are used all over hte place

diesel/src/types/mod.rs
```diff
@@ -420,6 +420,15 @@ impl<'a, T, U, DB> PartialEq<U> for ToSqlOutput<'a, T, DB> where
     }
 }
 
+impl<'a, T, DB> fmt::Debug for ToSqlOutput<'a, T, DB> where
+    T: fmt::Debug,
+    DB: TypeMetadata,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.out.fmt(f)
```
sgrif: We use `assert_eq!` in tests.

diesel/src/types/mod.rs
sgrif: Can't use `..self` here, as the type is changing.

diesel/src/pg/metadata_lookup.rs
```diff
@@ -0,0 +1,32 @@
...
+use super::{PgConnection, PgTypeMetadata};
+
+#[allow(missing_debug_implementations)]
+pub struct PgMetadataLookup {
+    conn: PgConnection,
+}
+
+impl PgMetadataLookup {
+    pub(crate) fn new(conn: &PgConnection) -> &Self {
+        unsafe { mem::transmute(conn) }
```
sgrif: I generally prefer to avoid code comments which can get out of sync with the code they're referencing. I explained the reasoning for this in the commit message, which is where I prefer to document these things. This function doesn't have any invariants which we could violate without rewriting the function.

diesel/src/types/mod.rs
killercup: Ok, fin— _Well, actually_ I get only 3 errors when removing the impls which are all from the `Array` which assumes `out` to be slice-like thing, and all other method calls on `out` are part of the `Write` impl (or rustc just gives up when after these errors, I can never tell).

No, seriously, it's fine.

diesel/src/pg/metadata_lookup.rs
killercup: Generally a good practice, which I try to follow as well. `unsafe` makes me question these practices, though :) I guess it's okay in this case, as its intend is pretty clear and the implementation obvious ("oh hey compiler, this PgConnection reference is now a PgMetadataLookup reference"). I was more surprised to see a `new` method that returns a reference, though!

sgrif: Do you think it would be valuable to add a comment which points to `Path::new` from the standard library as a reference point?

PR #1002, opened 2017-07-06
Don't implement expression methods on tuples

### LINKED ISSUE #998: Syntax error for delete operation
### Versions

- **Rust:** 1.18
- **Diesel:** 0.14
- **Database:** postgress
- **Operating System:** macos

### Feature Flags

- **diesel:** postgres
- **diesel_codegen:** postgres

## Problem Description
It seems that generated SQL for delete operation is not valid in some cases.

For query which described below in **Steps to reproduce** section `print_sql!` macros prints next:
```sql
DELETE FROM subs
WHERE subs.user_id, subs.show_id IN
      (SELECT
         subs.user_id,
         subs.show_id
       FROM
         (users
           INNER JOIN (subs
             INNER JOIN shows ON subs.show_id = shows.id) ON subs.user_id = users.id)
       WHERE users.id = ? AND shows.title = ?)
```
but posgtres expects parens in WHERE clause:
```sql
DELETE FROM subs
WHERE (subs.user_id, subs.show_id) IN
      (SELECT
         subs.user_id,
         subs.show_id
       FROM
         (users
           INNER JOIN (subs
             INNER JOIN shows ON subs.show_id = shows.id) ON subs.user_id = users.id)
       WHERE users.id = ? AND shows.title = ?)
```

I'm trying to delete object from table with nested select.

`Success!` in console and successful execution of delete operation.

```
Err(DatabaseError(__Unknown, "syntax error at or near \",\""))
```

### Steps to reproduce
**up.sql**
```sql
-- Your SQL goes here
CREATE TABLE users (
  id INTEGER PRIMARY KEY
);

CREATE TABLE shows (
  id BIGSERIAL PRIMARY KEY,
  title TEXT NOT NULL
);

CREATE TABLE subs (
  show_id BIGINT NOT NULL REFERENCES shows ON DELETE CASCADE,
  user_id INTEGER NOT NULL REFERENCES users ON DELETE CASCADE,
  PRIMARY KEY (show_id, user_id)
);

INSERT INTO users (id) VALUES
  (1);

INSERT INTO shows (id, title) VALUES
  (1, 'show-title');

INSERT INTO subs (show_id, user_id) VALUES
  (1, 1);
```

**models.rs**
```rust
use schema::{users, shows, subs};

#[derive(Identifiable, Queryable, Insertable, Associations, Debug)]
#[table_name="users"]
pub struct User {
    pub id: i32
}

#[derive(Identifiable, Queryable, Insertable, Associations, Debug)]
#[table_name="shows"]
pub struct Show {
    pub id: i64,
    pub title: String
}

#[derive(Queryable, Insertable, Associations, Debug)]
#[belongs_to(Show)]
#[belongs_to(User)]
#[table_name="subs"]
pub struct Subscription {
    pub show_id: i64,
    pub user_id: i32,
}

enable_multi_table_joins!(users, shows);
```
**schema.rs**
```rust
```

**lib.rs**
```rust
#[macro_use]
extern crate diesel;
#[macro_use]
extern crate diesel_codegen;
extern crate dotenv;

pub mod models;
pub mod schema;

use diesel::prelude::*;
use diesel::pg::PgConnection;
use dotenv::dotenv;
use std::env;

pub fn establish_connection() -> PgConnection {
    dotenv().ok();

    let database_url = env::var("DATABASE_URL")
        .expect("DATABASE_URL must be set");
    PgConnection::establish(&database_url)
        .expect(&format!("Error connecting to {}", database_url))
}
```

**main.rs**
```rust
extern crate diesel_delete_example;
extern crate diesel;

use diesel::prelude::*;
use diesel_delete_example::*;
use diesel_delete_example::schema::*;

fn main() {

    use diesel_delete_example::schema::subs::*;
    use diesel_delete_example::schema::shows::title;
    use diesel_delete_example::schema::users::id;

    let test_user_id = 1;
    let test_title = "show-title";

    let query = diesel::delete(subs::table.filter((user_id, show_id).eq_any(
        users::table.inner_join((subs::table).inner_join(shows::table))
            .select((user_id, show_id))
            .filter(id.eq(test_user_id).and(title.eq(test_title))))
    ));

    match query.execute(&connection) {
        Ok(_) => println!("Success!"),
        Err(e) => println!("{:?}", e)
    }
}
```

## Checklist

sgrif:
With the exception of `.nullable`, all of these methods would generate
invalid SQL if used. Some people mistakenly think that we map rust
tuples to a PG tuple. We don't actually support PG tuples, and if we did
we'd need to have it be a separate type to differentiate `SELECT id,
name` from `SELECT (id, name)`.

This is not a full solution to the problem. Ideally we would actually
properly represent in the type system that a tuple is a list, which is
only valid in a few contexts. Instead, this solution simply prevents
their construction. I don't think this mistake is common enough to
warrant more time being spent on it.

My hope is that if/when variadic generics happen, they are not built
around tuples, which would allow us to better separate this case from
normal values. Until that time, this should suffice.

I'm not 100% sold on the name `SingleValue`, as it seems odd that
`Array<T>` implements it. However, the only other name I could think of
was `NotTuple`, which will feel odd if we ever do support PG tuples.
Something like `NotColumnList` could work, but that name feels odd to
me. I think I want to stick with a name that says what the type is, not
what it isn't.

Fixes #998.

### DIFF
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -153,6 +153,9 @@ use std::io::Write;
 #[cfg(not(feature="postgres"))]
 impl NotNull for Numeric {}
 
+#[cfg(not(feature="postgres"))]
+impl SingleValue for Numeric {}
+
 /// The text SQL type.
 ///
 /// On all backends strings must be valid UTF-8.
@@ -294,6 +297,12 @@ impl<T: NotNull> IntoNullable for T {
 
 impl<T: NotNull> IntoNullable for Nullable<T> {
     type Nullable = Nullable<T>;
+}
+
+pub trait SingleValue {
+}
+
+impl<T: NotNull + SingleValue> SingleValue for Nullable<T> {
 }
 
 /// How to deserialize a single field of a given type. The input will always be
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -47,6 +47,7 @@ impl<'a, Parent, Child> BelongingToDsl<&'a Parent> for Child where
     Child: HasTable + BelongsTo<Parent>,
     Id<&'a Parent>: AsExpression<<Child::ForeignKeyColumn as Expression>::SqlType>,
     <Child as HasTable>::Table: FilterDsl<Eq<Child::ForeignKeyColumn, Id<&'a Parent>>>,
+    Child::ForeignKeyColumn: ExpressionMethods,
 {
     type Output = FindBy<
         Child::Table,
@@ -64,6 +65,7 @@ impl<'a, Parent, Child> BelongingToDsl<&'a [Parent]> for Child where
     Child: HasTable + BelongsTo<Parent>,
     Vec<Id<&'a Parent>>: AsInExpression<<Child::ForeignKeyColumn as Expression>::SqlType>,
     <Child as HasTable>::Table: FilterDsl<EqAny<Child::ForeignKeyColumn, Vec<Id<&'a Parent>>>>,
+    Child::ForeignKeyColumn: ExpressionMethods,
 {
     type Output = Filter<
         Child::Table,
--- a/diesel/src/expression_methods/global_expression_methods.rs
+++ b/diesel/src/expression_methods/global_expression_methods.rs
@@ -1,6 +1,7 @@
 use expression::{Expression, AsExpression, nullable};
 use expression::array_comparison::{In, NotIn, AsInExpression};
 use expression::operators::*;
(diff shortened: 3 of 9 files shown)


diesel/src/types/mod.rs
```diff
@@ -296,6 +299,12 @@ impl<T: NotNull> IntoNullable for Nullable<T> {
     type Nullable = Nullable<T>;
 }
 
+pub trait SingleValue {
```
Eijebong: Maybe add a comment to explain why this is here ?

PR #1003, opened 2017-07-06
Alias Macaddr to MacAddr

Eijebong:
This fixes the macaddr feature when used with type inference as the
macaddr type is capitalized to Macaddr and not MacAddr.

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -347,6 +347,11 @@ pub mod sql_types {
     #[derive(Debug, Clone, Copy, Default)] pub struct MacAddr;
 
     #[cfg(feature = "network-address")]
+    #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


diesel/src/pg/types/mod.rs
```diff
@@ -347,6 +347,10 @@ pub mod sql_types {
     #[derive(Debug, Clone, Copy, Default)] pub struct MacAddr;
 
     #[cfg(feature = "network-address")]
+    /// Alias for `MacAddr` to be able to use it with `infer_schema`.
+    pub type Macaddr = ::types::MacAddr;
```
sgrif: Don't need `::types`

sgrif: This needs to be `#[doc(hidden)]`

Eijebong: Yeah, I tried to compile it without the extras features and found it weird that I'd have to add `::types` and forgot it

PR #1008, opened 2017-07-08
Document `AstPass`

sgrif:
Given that implementing `QueryFragment` is something that we want
plugins to be able to do, this type is part of our public API. I do not
want constructing this type to be part of our public API (it might be
required today by someone who is implementing their own connection
adapter, but we should try to make that unneccessary), so those methods
are hidden. There are also a few methods that care about the current AST
pass which I'd like to refactor out when I have the time. Those methods
are also hidden.

### DIFF
--- a/diesel/src/query_builder/ast_pass.rs
+++ b/diesel/src/query_builder/ast_pass.rs
@@ -3,8 +3,21 @@ use query_builder::{BindCollector, QueryBuilder};
 use result::QueryResult;
 use types::{ToSql, HasSqlType};
 
-#[doc(hidden)]
 #[allow(missing_debug_implementations)]
+/// The primary type used when walking a Diesel AST during query execution
+///
+/// Executing a query is generally done in multiple passes. This list includes,
+/// but is not limited to:
+///
+/// - Generating the SQL
+/// - Collecting and serializing the values sent separately from the SQL
+/// - Determining if a query is safe to store in the prepared statement cache
+///
+/// When adding a new type that is used in a Diesel AST, you don't need to care
+/// about which specific passes are being performed, nor is there any way for
+/// you to find out what the current pass is. You should simply call the
+/// relevant methods and trust that they will be a no-op if they're not relevant
+/// to the current pass.
 pub struct AstPass<'a, DB> where
     DB: Backend,
     DB::QueryBuilder: 'a,
@@ -17,6 +30,7 @@ pub struct AstPass<'a, DB> where
 impl<'a, DB> AstPass<'a, DB> where
     DB: Backend,
 {
+    #[doc(hidden)]
     #[cfg_attr(feature="clippy", allow(wrong_self_convention))]
     pub fn to_sql(query_builder: &'a mut DB::QueryBuilder) -> Self {
         AstPass {
@@ -24,6 +38,7 @@ impl<'a, DB> AstPass<'a, DB> where
         }
     }
 
+    #[doc(hidden)]
     pub fn collect_binds(
         collector: &'a mut DB::BindCollector,
         metadata_lookup: &'a DB::MetadataLookup,
@@ -33,15 +48,22 @@ impl<'a, DB> AstPass<'a, DB> where
         }
     }
 
+    #[doc(hidden)]
     pub fn is_safe_to_cache_prepared(result: &'a mut bool) -> Self {
         AstPass {
             internals: AstPassInternals::IsSafeToCachePrepared(result),
         }
     }
 
-    /// Effectively copies `self`, with a narrower lifetime. This method
-    /// matches the semantics of the implicit reborrow that occurs when passing
-    /// a reference by value in Rust.
+    /// Call this method whenever you pass an instance of `AstPass` by value.
+    ///
+    /// Effectively copies `self`, with a narrower lifetime. When passing a
+    /// reference or a mutable reference, this is normally done by rust
+    /// implicitly. This is why you can pass `&mut Foo` to multiple functions,
+    /// even though mutable references are not `Copy`. However, this is only
+    /// done implicitly for references. For structs with lifetimes it must be
+    /// done explicitly. This method matches the semantics of what Rust would do
+    /// implicitly if you were passing a mutable reference
     pub fn reborrow(&mut self) -> AstPass<DB> {
         use self::AstPassInternals::*;
         let internals = match self.internals {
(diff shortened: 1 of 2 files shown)


sgrif:
Relevant for anyone outside the core team who may be interested in reviewing this: I'm planning on writing a guide around extending the query builder using [this code](https://github.com/sgrif/crates.io/blob/6d7ab2ea21dac6b1faa902386a6a37278bbc5547/src/pagination.rs) as the final goal. I'm feeling pretty good about the level of abstraction in that code (maybe not specifically for pagination, but it felt like a good self-contained example. And there will always be use cases that need to drop down to this level). However, `AstPass` needs to be documented for that guide (and specifically `reborrow` feels off to me if it's part of our public API. Unfortunately there's no good place that reborrowing for the language is documented that I can link to)

diesel/src/query_builder/ast_pass.rs
```diff
@@ -3,8 +3,21 @@ use query_builder::{BindCollector, QueryBuilder};
 use result::QueryResult;
 use types::{ToSql, HasSqlType};
 
-#[doc(hidden)]
 #[allow(missing_debug_implementations)]
+/// The primary type used when walking a Diesel AST during query execution
```
Eijebong: period

diesel/src/query_builder/ast_pass.rs
```diff
@@ -3,8 +3,21 @@ use query_builder::{BindCollector, QueryBuilder};
...
+///
+/// Executing a query is generally done in multiple passes. This list includes,
+/// but is not limited to:
+///
+/// - Generating the SQL
+/// - Collecting and serializing the values sent separately from the SQL
```
Eijebong: This sentence is weird IMO.

What about "Collecting and serializing SQL bound values" ?

diesel/src/query_builder/ast_pass.rs
```diff
@@ -3,8 +3,21 @@ use query_builder::{BindCollector, QueryBuilder};
 use result::QueryResult;
 use types::{ToSql, HasSqlType};
 
-#[doc(hidden)]
 #[allow(missing_debug_implementations)]
+/// The primary type used when walking a Diesel AST during query execution
+///
+/// Executing a query is generally done in multiple passes. This list includes,
+/// but is not limited to:
```
Eijebong: To be consistent with L85-94, add a period at the end of each bullet point

diesel/src/query_builder/ast_pass.rs
```diff
@@ -50,25 +72,73 @@ impl<'a, DB> AstPass<'a, DB> where
...
+    /// prepared statement cache.
+    ///
+    /// Diesel caches prepared statements as much as possible. However, it is
+    /// important to ensure that this doesn't result in unbounded memory usage
+    /// on the database server. To ensure this is the case, any logical query
+    /// which could generate a potentially unbounded number of prepared
+    /// statements *must* call this method. Examples of AST nodes which do this
+    /// are:
+    ///
+    /// - `SqlLiteral`. We have no way of knowing if the sql string was
```
Eijebong: SQL

diesel/src/query_builder/ast_pass.rs
```diff
@@ -50,25 +72,73 @@ impl<'a, DB> AstPass<'a, DB> where
...
+    ///     }
+    /// }
+    /// ```
     pub fn push_sql(&mut self, sql: &str) {
         if let AstPassInternals::ToSql(ref mut builder) = self.internals {
             builder.push_sql(sql);
         }
     }
 
+    /// Push the given SQL identifier on the end of the query being constructed.
```
Eijebong: This would confuse me if I didn't know what it was doing. What about "Add the given SQL identifier to the query [...]"

diesel/src/query_builder/ast_pass.rs
```diff
@@ -50,25 +72,73 @@ impl<'a, DB> AstPass<'a, DB> where
...
+    /// - `UpdateStatement`. The number of potential queries is actually
+    ///   technically bounded, but the upper bound is the number of columns on
+    ///   the table factorial which is too large to be safe.
     pub fn unsafe_to_cache_prepared(&mut self) {
         if let AstPassInternals::IsSafeToCachePrepared(ref mut result) = self.internals {
             **result = false
         }
     }
 
+    /// Push the given SQL string on the end of the query being constructed.
```
Eijebong: Same thing as `push_identifier`

diesel/src/query_builder/ast_pass.rs
```diff
@@ -50,25 +72,73 @@ impl<'a, DB> AstPass<'a, DB> where
...
+    /// The identifier will be quoted using the rules specific to the backend
+    /// the query is being constructed for.
     pub fn push_identifier(&mut self, identifier: &str) -> QueryResult<()> {
         if let AstPassInternals::ToSql(ref mut builder) = self.internals {
             builder.push_identifier(identifier)?;
         }
         Ok(())
     }
 
+    /// Push a value onto the given query to be sent separate from the SQL
```
Eijebong: "Push a bind parameter onto the given query. The value will automatically be bound at the execution" ?

diesel/src/query_builder/ast_pass.rs
sgrif: > This would confuse me if I didn't know what it was doing.

That's semi-intentional. This is the method least likely to be used by plugins. I should perhaps call that out more explicitly

Eijebong: I meant that the "at the end" in the sentence was confusing.

diesel/src/query_builder/ast_pass.rs
sgrif: Given that "bind parameter" isn't a standard term in SQL, I don't want to document a method called `push_bind_param` assuming knowledge of what `bind_param` means. Do you have any suggestions on how we can improve this without relying on knowing what the method name means already?

diesel/src/query_builder/ast_pass.rs
sgrif: I guess I mostly added that to give a definition to the `push` portion of the method name. I agree that the only reasonable place for it to be added is at the end. I like your suggestion

diesel/src/query_builder/ast_pass.rs
Eijebong: I thought it was a standard term. I don't have any other idea :/

sgrif: It's certainly the most common, but it's just a term that many ORMs use. It's not part of the standard.

PR #1010, opened 2017-07-09
Up the limit for composite primary key to 5

Eijebong:
Someone on IRC had a composite primary key with 5 columns. This allows
them to use diesel with that table.


Eijebong:
I'll see if I can add a test yeah

Issue #830, opened 2017-03-29
No support for some MySQL data types


Razican:
Diesel currently supports only the main MySQL data types, defined [here](http://docs.diesel.rs/diesel/mysql/enum.MysqlType.html). In that documentation page the [MySQL documentation](https://dev.mysql.com/doc/refman/5.7/en/c-api-prepared-statement-type-codes.html) is linked, were it can be seen that there are more types (`MEDIUMINT`, `DECIMAL`, `VARCHAR`, `VARBINARY`, `TINYBLOB`, `TINYTEXT`, `MEDIUMBLOB`, `MEDIUMTEXT`, `LONGBLOB`, `LONGTEXT` and  `BIT`) that are not supported by the current implementation.

Trying to use these types gives these errors:
```text
error[E0412]: cannot find type `Mediumint` in this scope
 --> src/db/schema.rs:3:1
  |
3 | infer_schema!("dotenv:DATABASE_URL");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
  |
  = note: this error originates in a macro outside of the current crate

error[E0412]: cannot find type `Decimal` in this scope
 --> src/db/schema.rs:3:1
  |
3 | infer_schema!("dotenv:DATABASE_URL");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
  |
  = note: this error originates in a macro outside of the current crate
```

Reported in Rust user's forum [here](https://users.rust-lang.org/t/diesel-cannot-find-derive-macro-inferschema/10095/4?u=razican).

Thomspoon:
Since the diesel_infer_schema seems to only work off of the default SQL types, this might require some direction on how a maintainer can help out.

Eijebong:
The Decimal type is currently being implemented for postgres in #837. I'll probably implement it for MySQL once it's done for pg.

mediumint cannot be deserialized as is since there is not i24 type. Would considering it as an integer be ok ? @sgrif

Thomspoon:
I figured primitives were the only acceptable types. If mediumint could just be a i32, that would solve a lot of problems at the expense of one byte per use.

sgrif:
[...]

PR #1012, opened 2017-07-09
add mysql aliases for Text and Binary types

Fiedzia:
Handle mysql text and binary types by adding suitable aliases
We may want to have something more ergonomic for bit type

### DIFF
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -174,6 +174,8 @@ impl SingleValue for Numeric {}
 /// On all backends strings must be valid UTF-8.
 /// On PostgreSQL strings must not include nul bytes.
 ///
+/// On Mysql, it is also aliased by Tinytext, Mediumtext, Longtext, Char and Varchar.
+///
 /// ### [`ToSql`](/diesel/types/trait.ToSql.html) impls
 ///
 /// - [`String`][String]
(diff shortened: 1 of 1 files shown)


diesel/src/types/mod.rs
```diff
@@ -174,6 +174,8 @@ impl SingleValue for Numeric {}
 /// On all backends strings must be valid UTF-8.
 /// On PostgreSQL strings must not include nul bytes.
 ///
+/// On Mysql, it is also aliased by Tinytext, Mediumtext, Longtext, Char and Varchar.
```
killercup: Just cosmetics, but can you change it to "MySQL" and put the type names in back ticks (like <code>\`this\`</code>).

diesel/src/types/mod.rs
```diff
@@ -188,9 +190,15 @@ impl SingleValue for Numeric {}
...
 pub type VarChar = Text;
 #[doc(hidden)] pub type Varchar = VarChar;
+#[doc(hidden)] pub type Char       = Text;
+#[doc(hidden)] pub type Tinytext   = Text;
+#[doc(hidden)] pub type Mediumtext = Text;
+#[doc(hidden)] pub type Longtext   = Text;
 
 /// The binary SQL type.
 ///
+/// On Mysql, it is also aliased by Tinyblob, Mediumblob, Longblob, Bit and Varbinary.
```
killercup: Same as above

Fiedzia:
> I'm not entirely sure what the semantic differences of these types are

None I believe. If I read MySQL docs correctly, Mysql will handle them differently on the server side (at storage level), but there is no difference in the transport layer. At storage level those types simply differ in allowed max size.

> What use cases does this add for you?

Schema generated by diesel print-schema compiles. Also I'd like to have ability to generate create table sql from diesel schema definition (I am not sure if/how that works yet, but being specific here would help in achieving that).

diesel/src/types/mod.rs
```diff
@@ -188,9 +190,15 @@ impl SingleValue for Numeric {}
 #[derive(Debug, Clone, Copy, Default)] pub struct Text;
 pub type VarChar = Text;
 #[doc(hidden)] pub type Varchar = VarChar;
+#[doc(hidden)] pub type Char       = Text;
```
sgrif: Please avoid visual alignment.

Fiedzia:
This is part of solution for #830

killercup:
@sgrif are you good with merging this? I'd probably want to add a test for inferring all these types, but I think that's not that easy right now.

sgrif:
LGTM

killercup:
Thanks again!

PR #1021, opened 2017-07-13
Add support for the postgres range type

Eijebong:


diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+use backend::Debug;
+use pg::{Pg, PgTypeMetadata};
+use types::*;
+use std::collections::Bound;
+use std::error::Error;
+use query_source::Queryable;
+
+bitflags! {
+    struct RangeBound: u8 {
+        const RANGE_EMPTY = 0x01;
```
sgrif: Since we are relying on specific bits being set here, can we use binary literals instead?

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
+
+use byteorder::{ReadBytesExt, WriteBytesExt, NetworkEndian};
+use backend::Debug;
+use pg::{Pg, PgTypeMetadata};
+use types::*;
+use std::collections::Bound;
```
sgrif: I usually group "external uses" separately from "internal uses"

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct RangeMetadata {
+    oid: u32,
+    array_oid: u32
+}
+
+pub trait CanAppearInRange {
+    fn metadata() -> RangeMetadata;
```
sgrif: Hm... This feels off to me. I think I'd prefer this to be something like:

```rust
pub trait CanAppearInRange where
    Pg: HasSqlType<Self::RangeType>,
{
    type RangeType;
}
```

That way we can just call `metadata` on `RangeType`

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+
+
+impl<T, ST> FromSql<Range<ST>, Pg> for (Bound<T>, Bound<T>) where
+    ST: CanAppearInRange,
+    T: FromSql<ST, Pg>,
+    Pg: HasSqlType<ST>,
+{
+    fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
+        let mut bytes = not_none!(bytes);
+        let flags: RangeBound = RangeBound::from_bits(bytes.read_u8()?).unwrap();
```
sgrif: Let's avoid using `unwrap`

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+        let mut lbound = Bound::Unbounded;
```
sgrif: Can we avoid abbreviations for this?

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+            let v = T::from_sql(Some(elem_bytes))?;
+            if flags.contains(RANGE_LB_INC){
+                lbound = Bound::Included(v);
+            }
+            else {
+                lbound = Bound::Excluded(v);
+            }
+        }
+        if !flags.contains(RANGE_UB_INF) {
+            let _ = bytes.read_i32::<NetworkEndian>()?; // Just get the size out of the way, we don't need it
```
sgrif: How about we add a `debug_assert!` here that the size matches

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        if !flags.contains(RANGE_LB_INF) {
+            let elem_size = bytes.read_i32::<NetworkEndian>()?;
+            let (elem_bytes, new_bytes) = bytes.split_at(elem_size as usize);
+            bytes = new_bytes;
+            let v = T::from_sql(Some(elem_bytes))?;
+            if flags.contains(RANGE_LB_INC){
+                lbound = Bound::Included(v);
+            }
+            else {
```
sgrif: Formatting is a bit weird here.

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
...
+    T: FromSql<ST, Pg> + Queryable<ST, Pg>,
+    Pg: HasSqlType<ST>,
+{
+    type Row = Self;
+    fn build(row: Self) -> Self {
+        row
+    }
+}
+
+impl<T, ST> FromSqlRow<Range<ST>, Pg> for (Bound<T>, Bound<T>) where
```
sgrif: We need the nullable impl too

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,125 @@
+
```
sgrif: Can you run this file through `rustfmt` when you're done? (Latest version default settings)

I'm planning on doing it to the whole codebase soon, one less place that needs to churn.

sgrif:
Unless I'm missing something, you're missing `AsExpression` impls.

diesel/src/pg/types/ranges.rs
Eijebong: I just copy/pasted it from the postgres source code but yeah I can change that to binary literals if you want to

sgrif: I don't feel super strongly either way, I just find the binary version easier to read.

Eijebong: I don't think they're easier to read when they're not visually aligned

sgrif: That's a good point.

diesel/src/pg/types/ranges.rs
Eijebong: How would I know the expected size ?

diesel/src/pg/types/ranges.rs
killercup: Make sure to `cargo install rustfmt-nightly` which is the version with all the latest changes. (Or better yet, `cargo install --git https://github.com/rust-lang-nursery/rustfmt` directly)

Eijebong: Done :)

diesel_tests/tests/schema_inference.rs
```diff
@@ -163,3 +163,48 @@ mod sqlite {
...
+        date: (Bound<NaiveDate>, Bound<NaiveDate>),
+    }
+
+    #[test]
+    fn ranges_are_correctly_inferred() {
+        let conn = connection();
+        let numeric = PgNumeric::Positive{
+            weight: 1,
+            scale: 1,
+            digits: vec![1]
```
killercup: trailing comma

diesel/src/pg/types/ranges.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+impl HasSqlType<Range<Int8>> for Pg {
+    fn metadata(_: &PgMetadataLookup) -> PgTypeMetadata {
+        PgTypeMetadata {
+            oid: 3926,
+            array_oid: 3927,
+        }
+    }
+}
```
killercup: Do we have a macro for these impls?

Eijebong: `primitive_impls` but it doesn't like generic stuff and I think it also implements other things.

PR #1025, opened 2017-07-14
Don't deny warning when developing, only on CI

killercup:
This changes the lint settings to only emit warnings by default. This means that during development, you'll get to see warnings, but rustc will continue on with compiling diesel.

When explicitly checking the code for code style (on CI in the clippy build step, or when using `bin/check`), we supply `--deny warnings` as a flag to `rustc` to error out on warnings.

This does not change anything when using diesel as a dependency, as cargo automatically sets all lints to allow in that case.


killercup:
I finally came around to it, @llogiq! :)

PR #1027, opened 2017-07-14
Explicitly set the time zone and client encoding at connection startup

### LINKED ISSUE #1024: SET TIMEZONE 'UTC' on connection by default
Filed from: https://github.com/rust-lang/crates.io/pull/846

khan timezones meme

sgrif:
Diesel generally assumes that the database time zone is UTC, and that
the encoding is UTF8. These are the most common configurations, but this
explicitly sets our assumptions when the connection is established.

SQLite does not have a way to configure either of these, it is always
UTF8 and UTC.

Fixes #1024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   remove the `Nullable` portion (Unless you are using it with fields that are
   actually nullable)
 
+* Connections will now explicitly set the session time zone to UTC when the
+  connection is established
+
 ## [0.14.1] - 2017-07-10
 
 ### Changed
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -115,7 +115,11 @@ impl MysqlConnection {
     }
 
     fn set_config_options(&self) -> QueryResult<()> {
-        try!(self.execute("SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT'))"));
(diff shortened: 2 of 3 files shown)


diesel/src/mysql/connection/mod.rs
```diff
@@ -115,7 +115,11 @@ impl MysqlConnection {
     }
 
     fn set_config_options(&self) -> QueryResult<()> {
-        try!(self.execute("SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT'))"));
+        self.execute("SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT'))")?;
+        self.execute("SET time_zone = '+00:00';")?;
+        self.execute("SET character_set_client = 'utf8mb4'")?;
```
Eijebong: Won't this break everything with indexed `VARCHAR(255)`

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   remove the `Nullable` portion (Unless you are using it with fields that are
   actually nullable)
 
+* Connections will now explicitly set the session time zone to UTC when the
```
Eijebong: and encoding

sgrif: This is not a relevant user-facing detail

diesel/src/mysql/connection/mod.rs
sgrif: No

sgrif: This doesn't affect data storage, it affects transmission.

PR #1033, opened 2017-07-21
Add support for Citext Postgres Type

### LINKED ISSUE #1000: Support citext postgres type
Likely this should be fine:

```
type Citext = Text
```

Also need to check if the right coersions exist and that codegen will place nice with citext[]

Notes from gitter^ :)

alexcameron89:
Fixes #1000


killercup:
Thanks! Just a quick heads up: If you think this fixes #1000, add "Fixes #1000" to the commit message. This way, Github will automatically close the issue when this gets merged.

alexcameron89:
Thanks @killercup! I always forget to use that feature , l will remember to in the future😄

PR #1034, opened 2017-07-23
[DO NOT MERGE] Potential API change to `table!`

sgrif:
The goal that I'm trying to accomplish with this change is to allow you to reference `users` without the `::table`, but not have to `use users::dsl::*`. I've found that when I'm writing functions which are working with a large number of tables (e.g. I can't import `dsl`), having `::table` all over the place feels really verbose. This change is possible because Rust has two namespaces: "values" (constants, functions, etc) and "types" (structs, enums, type aliases, modules, etc).

However, I'm not 100% sure I want to make this change on its own. A common point of confusion among new users is when the `dsl` module should be imported, and what exactly that does. (I've also seen people do `users::table` when `users::dsl::table` is in scope, which gives a confusing error because `users::table` will be referencing `http://docs.diesel.rs/diesel/associations/trait.HasTable.html#tymethod.table`). This will just make that situation more confusing. Knowing when you can do `users::id` vs `users.id` vs `id` requires knowledge of exactly what we generate, and thorough knowledge of name resolution in Rust. It's something that's extremely hard for us to document well in the places people will be looking for it. Additionally, this many `use` lines just feels even more verbose. I'd want a solution that requires a single `use` line, or a way to eliminate them entirely.

So I think if we make this change, we should try to make this the canonical way to do things, or at least have `users.id` *always* work. This is actually possible, since we can re-export just the constant but not the type at the same level as the module. The main downside to doing that is all code which looks like this would break:

```rust
use schema::*;

let users = ...
```

This would error since you cannot have a let binding with the same name as a constant. This only affects code that was not using the `dsl` module, since you also cannot have a let binding with the same name as a unit struct.

I generally think that `let users = users::table.stuff` is a bit funky looking anyway, so I'm not sure if I'm actually concerned about this or not. If nothing else it'll cause a lot of breakage in our own test suite since I *always* `use schema::*`, but it's possible that the migration will be easier. If nothing else, I want to see how much work it is to migrate our test suite and crates.io, and also how much this improves the codebase of crates.io

If we do make this change, I think we can remove the `dsl` module entirely, and code which wants that can do `use users::columns::*` instead which is much more explicit about exactly what happens.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -577,7 +577,14 @@ macro_rules! table_body {
             ///
             /// This is the type which provides the base methods of the query
             /// builder, such as `.select` and `.filter`.
-            pub struct table;
+            pub struct table {
+                $(pub $column_name: $column_name,)+
+            }
+
+            #[allow(non_upper_case_globals)]
+            pub const table: table = table {
+                $($column_name: $column_name,)+
+            };
 
             impl table {
                 #[allow(dead_code)]
--- a/diesel_tests/tests/joins.rs
+++ b/diesel_tests/tests/joins.rs
@@ -455,6 +455,11 @@ fn selecting_crazy_nested_joins() {
 }
 
 fn connection_with_fixture_data_for_multitable_joins() -> (TestConnection, TestData) {
+    use schema::comments::dsl::comments;
+    use schema::followings::dsl::followings;
+    use schema::likes::dsl::likes;
+    use schema::posts::dsl::posts;
+
     let connection = connection_with_sean_and_tess_in_users_table();
 
     let sean = find_user_by_name("Sean", &connection);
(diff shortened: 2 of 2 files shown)


sgrif:
Immediately found code which validates my concerns in crates.io. Approximately this:

```rust
use schema::versions;

#[derive(Identifiable)] // <--- This requires `schema::versions` be in scope for the module
pub struct Version {
    // ...
}

impl Version {
    //       v---- This variable name will always error if `schema::versions` is in scope for the module
    fn max(versions: Vec<Self>) -> Self {
        // This function does not do anything Diesel related
    }
}
```

If we made this change, I'd probably end up fixing our test suite by either limiting the imports we do from `schema` or always importing them at the function level instead of the module level. The problem is, we have code that enforces that you do the import at the module level. With this change, you could no longer use a variable with the same name as the table anywhere in the file, even though it's often the most convenient variable name for a collection in functions which have nothing to do with Diesel.

sgrif:
Yeah so I think this is pretty much dead unless we're ok with this change as is. I explored if we could have all our derives which take the table name take a path instead, and it's basically impossible unless we change Rust to allow a `path` fragment to be followed by `::`

PR #1035, opened 2017-07-23
Hide the token patterns we look for in `table!`

sgrif:
The generated documentation for the `table!` macro is really bad. Macros
document all of the patterns that they match on. For complex macros such
as this one, this is not useful information to the user.

This was maybe useful back in 0.1 when the docs looked like
https://docs.rs/diesel/0.1.0/diesel/macro.table!.html, but now it's just
a bunch of opaque macro soup that the user doesn't care about. The
actual documentation is buried more than two full page scrolls down.

This makes the documented token pattern as small as possible, so that
the actual documentation can be front and center.

Before:
screen shot 2017-07-23 at 6 47 35 am

After:
screen shot 2017-07-23 at 6 47 42 am

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -243,6 +243,14 @@ macro_rules! __diesel_column {
 /// ```
 #[macro_export]
 macro_rules! table {
+    ($($tokens:tt)*) => {
+        __diesel_table_impl!($($tokens)*);
+    }
+}
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __diesel_table_impl {
     // Put imports into the import field
     (
         @parse


sgrif:
Immediately found code which validates my concerns in crates.io. Approximately this:

```rust
use schema::versions;

#[derive(Identifiable)] // <--- This requires `schema::versions` be in scope for the module
pub struct Version {
    // ...
}

impl Version {
    //       v---- This variable name will always error if `schema::versions` is in scope for the module
    fn max(versions: Vec<Self>) -> Self {
        // This function does not do anything Diesel related
    }
}
```

If we made this change, I'd probably end up fixing our test suite by either limiting the imports we do from `schema` or always importing them at the function level instead of the module level. The problem is, we have code that enforces that you do the import at the module level. With this change, you could no longer use a variable with the same name as the table anywhere in the file, even though it's often the most convenient variable name for a collection in functions which have nothing to do with Diesel.

sgrif:
Yeah so I think this is pretty much dead unless we're ok with this change as is. I explored if we could have all our derives which take the table name take a path instead, and it's basically impossible unless we change Rust to allow a `path` fragment to be followed by `::`

PR #1039, opened 2017-07-24
Enable joining from boxed queries

natboehm:
Looks like this commit didn't make it in the 0.15.0 release, we think this is what we need for the `group_by` statement. This is the only thing that wasn't included from the branch we were testing with and isn't currently in master. 

When we try crates.io with Diesel 0.15.0, we're getting the error 
```
ERROR:conduit_log_requests: 127.0.0.1:64825 [2017-07-24T10:37:29-04:00] 
Get /api/v1/crates - 4ms 500: syntax error at or near "GROUP" syntax error at or near "GROUP"
```


sgrif:
Can you amend the commit message to note that this is for group by not join?

natboehm:
Yup, done

sgrif:
Releasing now. Thanks.

Issue #1037, opened 2017-07-24
Query dsl generated by select leak private types to public interface


weiznich:
### Versions

- **Rust:** 1.19
- **Diesel:** 0.15
- **Database:** Postgres
- **Operating System** Ubuntu

### Feature Flags

- **diesel:** default-features = false, features = ["postgres", "chrono", "serde_json"]
- **diesel_codegen:** default

### What are you trying to accomplish?

Return a common select query part from a function (I know of boxed queries)

### What is the expected output?

The code compiles (even if the the returning type is somewhat complex)

### What is the actual output?

It is impossible to import all needed types, because some of them are inside of a private module of diesel. Therefore it is not possible to write down those type.

(In theory there should be a lint([private_in_public]( https://github.com/rust-lang/rust/issues/34537)) for this in rustc, but it seems like the lint failed to catch this types)

### Steps to reproduce

```rust
table! {
    groups {
        id -> Integer,
        group_name -> Text,
    }
}

fn query_groups() -> ::diesel::query_builder::SelectStatement<self::groups::table,
                                                              ::diesel::query_builder::select_clause::DefaultSelectClause,
                                                              ::diesel::query_builder::distinct_clause::NoDistinctClause,
                                                              ::diesel::query_builder::where_clause::WhereClause<
                                                                      ::diesel::expression::operators::Eq<self::groups::columns::id,
                                                                                                          ::diesel::expression::bound::Bound<::diesel::types::Integer, i32>>>>{
    use self::groups::dsl::*;
    groups.filter(id.eq(1))
}
```

```
error[E0603]: module `select_clause` is private
  --> src/lib.rs:77:63
   |
77 |                                                               ::diesel::query_builder::select_clause::DefaultSelectClause,
   |                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0603]: module `distinct_clause` is private
  --> src/lib.rs:78:63
   |
78 |                                                               ::diesel::query_builder::distinct_clause::NoDistinctClause,
   |                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### Cause

[...]

PR #1040, opened 2017-07-24
Add Helper types for join dsl

weiznich:
Add some helper types for the join dsl. See #1037


sgrif:
Can you add a changelog entry?

weiznich:
I've added a changelog entry.

PR #1041, opened 2017-07-24
Prototype of infering enums from the database schema

weiznich:
* It only works with postgres

This is a resubmitted update version of #940.
Some small things should be fixed(see the todo list) before the pull request is merged. Feel free to comment on the other parts in the meanwhile.

Todo:

- [ ] Pretty Printing
- [ ] Documentation

### DIFF
--- a/diesel_codegen/src/schema_inference.rs
+++ b/diesel_codegen/src/schema_inference.rs
@@ -6,6 +6,28 @@ use diesel_infer_schema::*;
 
 use util::{get_options_from_input, get_option, get_optional_option};
 
+#[cfg(feature = "postgres")]
+pub fn derive_infer_enum(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `infer_schema`!");
+    }
+    let options = get_options_from_input("infer_enum_options", &input.attrs, bug)
+        .unwrap_or_else(|| bug());
+    let database_url = extract_database_url(get_option(&options, "database_url", bug)).unwrap();
+    let schema_name = get_optional_option(&options, "schema_name");
+    let schema_name = schema_name.as_ref().map(|s| &**s);
+
+    let enums = load_enums(&database_url, schema_name)
+        .expect(&format!("Could not load enums from database {}", database_url))
+        .into_iter()
+        .map(|e| expand_enum(e, ExpandEnumMode::Codegen));
+    quote!{
+        #(#enums)*
+    }
+
+}
+
 pub fn derive_infer_schema(input: syn::MacroInput) -> quote::Tokens {
     fn bug() -> ! {
         panic!("This is a bug. Please open a Github issue \
--- a/diesel_infer_schema/src/codegen.rs
+++ b/diesel_infer_schema/src/codegen.rs
@@ -4,9 +4,251 @@ use quote;
 use syn;
 
 use table_data::TableData;
-use data_structures::ColumnInformation;
+use data_structures::{ColumnInformation, EnumInformation};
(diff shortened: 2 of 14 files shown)


killercup:
Thanks! That looks like a lot of features!

It's also a huge amount of code, though… I'll try my best to find some time to play around with it. One single commit adding 472 lines is not quick to read through 😅 I personally love to read big additions like these in multiple commits that build on each other and whose commit messages tell me the indent of each change.

IIRC, @sgrif mentioned in an earlier discussion about enum inference that he'd like to try to have this in (an) external crate(s). Maybe he can chime to supply his original arguments before I quote him wrong :). @weiznich, do you think this is do-able as an add-on?

diesel_codegen/src/schema_inference.rs
```diff
@@ -6,6 +6,28 @@ use diesel_infer_schema::*;
 
 use util::{get_options_from_input, get_option, get_optional_option};
 
+#[cfg(feature = "postgres")]
+pub fn derive_infer_enum(input: syn::MacroInput) -> quote::Tokens {
+    fn bug() -> ! {
+        panic!("This is a bug. Please open a Github issue \
+               with your invocation of `infer_schema`!");
```
sgrif: This isn't `infer_schema!`

diesel_infer_schema/src/codegen.rs
```diff
@@ -4,10 +4,252 @@ use quote;
...
+                        match s {
+                            #(#field_mapping,)*
+                            _ => unreachable!(),
+                        }
+                    }
+                }
+            }
+        };
+        quote!{
+            #from_1
```
sgrif: Given that the latest stable rust ups the default recursion limit to 1024, this is probably not necessary.

diesel_infer_schema/src/codegen.rs
```diff
@@ -4,10 +4,252 @@ use quote;
...
+            impl From<String> for #type_name {
+                fn from(s: String) -> Self {
+                    match s.as_str() {
+                        #(#field_mapping,)*
+                        _ => unreachable!(),
```
sgrif: `From` mappings should never fail. This should be `TryFrom` or just left to the `FromSql` implementation.

sgrif:
Yeah so as @killercup mentioned, I don't think this is a feature that's appropriate for Diesel itself. I'm happy to help you get it set up as a third party crate though. I thought I had mentioned this before, but I can't find it on the issue tracker so it must have been in gitter or I just never mentioned it. Sorry if I haven't made this clear before now.

In general, I think this code should be broken up differently. I think a simpler starting point would be `#[derive(PgEnum)]`, which you place on enums defined in your code. `infer_enums!` could be built on top of that pretty easily (though I'm not really sure that this needs a macro to infer it. The number of enum types in an application are going to be pretty small compared to the number of tables).

weiznich:
> Yeah so as @killercup mentioned, I don't think this is a feature that's appropriate for Diesel itself. I'm happy to help you get it set up as a third party crate though. I thought I had mentioned this before, but I can't find it on the issue tracker so it must have been in gitter or I just never mentioned it. Sorry if I haven't made this clear before now.

I remember only some discussion about mapping enums to primitive types like integers, but it is ok. Would it be possible to host this crate as part of the diesel-rs organization?

> In general, I think this code should be broken up differently. I think a simpler starting point would be #[derive(PgEnum)], which you place on enums defined in your code. infer_enums! could be built on top of that pretty easily (though I'm not really sure that this needs a macro to infer it. The number of enum types in an application are going to be pretty small compared to the number of tables).

Good point:+1: This will hide the implementation details fro the actual users. It should be quite easily possible to rewrite the current code to that approach.

PR #1042, opened 2017-07-25
Remove migration output from setup, including test

### LINKED ISSUE #1023: Don't print migration output when running the initial setup migration
I mostly want the initial setup migration to be invisible. We shouldn't have this in the output:

```
Running migration 00000000000000
```

This just requires special casing that version in the function which prints that line.

juliusdelta:
fix #1023 

I'm not 100% sure why but it's showing I added `#[test]` and `[cfg(feature="postgres")]` to the spec below it instead of to itself but I didn't 🤔 ... 

Well I ran the spec and got it to pass. Let me know if something isn't up to par and I'll get it resolved.

### DIFF
--- a/diesel_cli/tests/setup.rs
+++ b/diesel_cli/tests/setup.rs
@@ -32,6 +32,16 @@ fn setup_creates_migrations_directory() {
 
 #[test]
 #[cfg(feature="postgres")]
+fn setup_initial_migration_returns_nothing_to_console() {
+    let p = project("setup_intial_migration_returns_nothing_to_console").build();
+
+    let result = p.command("setup").run();
+
+    assert!(!result.stdout().contains("Running migration"));
+}
+
+#[test]
+#[cfg(feature="postgres")]
 fn setup_creates_default_migration_file() {
     let p = project("setup_creates_default_migration_file").build();
 
--- a/diesel/src/migrations/mod.rs
+++ b/diesel/src/migrations/mod.rs
@@ -269,7 +269,9 @@ fn run_migration<Conn>(conn: &Conn, migration: &Migration, output: &mut Write)
         Conn: MigrationConnection,
 {
     conn.transaction(|| {
-        try!(writeln!(output, "Running migration {}", migration.version()));
+        if migration.version() != "00000000000000" {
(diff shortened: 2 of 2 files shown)


diesel_cli/tests/setup.rs
```diff
@@ -32,6 +32,16 @@ fn setup_creates_migrations_directory() {
 
 #[test]
 #[cfg(feature="postgres")]
+fn setup_initial_migration_returns_nothing_to_console() {
+    let p = project("setup_intial-migration_returns_nothing_to_console").build();
+
+    let result = p.command("setup").run();
+
+    assert_eq!(result.stdout().contains("00000000000000"), false);
```
Eijebong: You can just use `assert!(!result.stdout().contains("00000000000000"))` here

Eijebong:
> I'm not 100% sure why but it's showing I added #[test] and [cfg(feature="postgres")] to the spec below it instead of to itself but I didn't 🤔 ...

It's just how git does things, sometimes it's a bit messed up, but it looks ok to me

diesel_cli/tests/setup.rs
juliusdelta: 👍 I'll fix that and squash the commits

juliusdelta:
Done @Eijebong 💪

diesel_cli/tests/setup.rs
```diff
@@ -32,6 +32,16 @@ fn setup_creates_migrations_directory() {
 
 #[test]
 #[cfg(feature="postgres")]
+fn setup_initial_migration_returns_nothing_to_console() {
+    let p = project("setup_intial_migration_returns_nothing_to_console").build();
+
+    let result = p.command("setup").run();
+
+    assert!(!result.stdout().contains("00000000000000"));
```
sgrif: Can we change the assertion to make sure it doesn't contain `Running migration` instead?

juliusdelta: Sure thing!

juliusdelta: Done

juliusdelta:
Ty guys. That was my first non-documentation PR to an open source project. Thanks for being so helpful. @Eijebong @sgrif

PR #1043, opened 2017-07-25
Fix typo and add rust version info to contributing guide

juliusdelta:
Just some minor documentation fixes and upgrades. I didn't see anywhere where it explicitly described Diesel's support for Rust Nightly and intermittent support for Beta, as I understand it, so I opted to make that clearer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,8 +54,11 @@ Thank you! We'll try to get back to you as soon as possible.
 
 1. Install Rust using [rustup], which allows you to easily switch between Rust
    versions.
+
+   Diesel currently supports Rust Stable, Nightly, Rust Beta.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -54,8 +54,11 @@ Thank you! We'll try to get back to you as soon as possible.
 
 1. Install Rust using [rustup], which allows you to easily switch between Rust
    versions.
+
+   Diesel currently supports Rust Stable & Rust Nightly, but does not guarantee support for Rust Beta.
```
sgrif: This is incorrect. We support stable, beta, and nightly. There's an issue with our doctests failing the beta build at this moment, but that doesn't affect our support.

juliusdelta: Fixed

killercup:
Thanks!

PR #1047, opened 2017-07-29
Include bind parameters in `debug_sql!`

sgrif:
I've implemented this as a separate AST pass, rather than the
`CollectBinds` pass of the `Debug` backend. This is because we will
ultimately want to use this same code for logging, which won't have the
`QueryFragment<Debug>` constraint. Ultimately I think I want to get rid
of the `Debug` backend entirely once logging is implemented.

I had originally implemented this by adding two additional lifetime
parameters onto `AstPass`. However, I really disliked how opaque and
random these two extra parameters felt, and it made the type harder to
document (and use). Shortening an invariant lifetime like this is
actually one of the documented use cases that `mem::transmute` is useful
for. I don't like introducing unsafe code in general, but I felt that
the tradeoff was worth it in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added helper types for inner join and left outer join
 
-* `diesel::debug_sql` has been added as a replacement for `debug_sql!`. This
+* `diesel::debug_query` has been added as a replacement for `debug_sql!`. This
   function differs from the macro by allowing you to specify the backend, and
-  will generate the actual query which will be run.
+  will generate the actual query which will be run. The returned value will
+  implement `Display` and `Debug` to show the query in different ways
 
 * `diesel::pg::PgConnection`, `diesel::mysql::MysqlConnection`, and
   `diesel::sqlite::SqliteConnection` are now exported from `diesel::prelude`.
@@ -47,7 +48,7 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Removed
 
-* `debug_sql!` has been deprecated in favor of `diesel::debug_sql`.
+* `debug_sql!` has been deprecated in favor of `diesel::debug_query`.
 
 * `print_sql!` has been deprecated without replacement.
 
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -144,7 +144,7 @@ pub mod prelude {
 
 pub use prelude::*;
 #[doc(inline)]
-pub use query_builder::debug_sql;
+pub use query_builder::debug_query;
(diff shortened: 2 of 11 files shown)


diesel/src/query_builder/debug_binds.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+}
+
+impl<'a, T, DB> Display for DebugBinds<'a, T, DB> where
+    DB: Backend,
+    T: QueryFragment<DB>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("binds: ")?;
```
Eijebong: Can we find something to add between the query and the parameters ? It's a bit confusing IMO

sgrif: How do you feel about treating the output as an unnamed struct? So the output would be:

```
{ query: "SELECT * FROM users WHERE id = ?", binds: [1] }
```

And then we could theoretically work properly with pretty-printing as well

killercup: My first thought was something like

```sql
SELECT * FROM users WHERE id = ? -- binds: [1]
```

sgrif: I was hoping to keep them semantically separate since if we go with `slog` we'll actually want to keep them separate in that structure. If we did want to do the json-ish thing, perhaps instead of a `debug_sql` function/macro, we should have a `.debug()` instance method which just returns a struct that implements `std::fmt::Debug`

sgrif: Actually here's a thought -- what if `DebugQuery` implemented `Display` as you've described, and `Debug` with the anonymous struct form?

Eijebong: :+1: for me

sgrif: I'm going to merge #1051 first, and redo this on top of that, as a lot of the surrounding code has changed here.

sgrif:
I've updated this to include the API we've discussed. `debug_sql` is now `debug_query` which returns a type that implements `Display` and `Debug`. The display implementation will be the SQL query with the binds as a comment. The `Debug` impl pretends it's a struct called `Query`, which contains the SQL and the binds as fields.

killercup:
Oh, @sgrif, that CI failure looks noteworthy:

> assertion failed: `(left == right)` (left: `"SELECT \"users\".\"id\", \"users\".\"n\" FROM \"users\" WHERE \"users\".\"n\" = $1 -- binds: [1]"`, right: `"SELECT \"users\".\"id\", \"users\".\"n\" FROM \"users\" WHERE \\\n        \"users\".\"n\" = ? -- binds: [1]"`)

1. Escaping
2. `$1` vs. `?`

(This is Postgres 9.5 on Windows.)

PR #1048, opened 2017-07-29
update the diesel_cli readme

rbalicki2:
I'm not sure If there are other prerequisites to installing the `diesel_cli` that I'm unaware of. Happy to add more if necessary.

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -8,8 +8,8 @@ Installation
 ---------------
 
 The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
-and [`mysql`][mysql], some of which already be included on your system. See which
-dependencies are already present with:
+and [`mysql`][mysql], some of which may already be included on your system. See
+which dependencies are already present with:
 
 ```sh
 which psql
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -8,16 +8,8 @@ Installation
 ---------------
 
 The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
-and [`mysql`][mysql], some of which may already be included on your system. See
-which dependencies are already present with:
-
-```sh
-which psql
-which sqlite3
-which mysql
-```
-
-Once these dependencies are installed, you can run `cargo install diesel_cli`.
+and [`mysql`][mysql]. Once these dependencies are installed, you can run
+`cargo install diesel_cli`.
 
 To install the cli without these dependencies, omit the unneeded dependencies from
 the following command:
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -7,11 +7,30 @@ bi-directional changes to your database that get applied sequentially.
 Installation
 ---------------
 
-The diesel cli requires `mysql` to be installed on your system. If you have
-homebrew installed, you can run `which mysql || brew install mysql` to install
-`mysql` if necessary.
+The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
+and [`mysql`][mysql], some of which already be included on your system. See which
+dependencies are already present with:
 
-Finally, you can install the diesel CLI by running `cargo install diesel_cli`.
+```sh
+which psql
+which sqlite3
+which mysql
+```
+
+Once these dependencies are installed, you can run `cargo install diesel_cli`.
+
+To install the cli without these dependencies, omit the unneeded dependencies from
+the following command:
+
+```
+cargo install diesel_cli --no-default-features --features postgres,sqlite,mysql
+```
+
+The argument passed to `--features` must include no spaces, and be comma-separated.
+
+[postgres]: https://www.postgresql.org/download/
+[sqlite]: http://www.sqlitetutorial.net/download-install-sqlite/
+[mysql]: https://dev.mysql.com/doc/refman/5.7/en/installing.html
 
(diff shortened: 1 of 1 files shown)


diesel_cli/README.md
```diff
@@ -4,6 +4,15 @@ Diesel CLI
 Diesel CLI is a tool that aids in managing your database schema. Migrations are
 bi-directional changes to your database that get applied sequentially.
 
+Installation
+---------------
+
+The diesel cli requires `mysql` to be installed on your system. If you have
+homebrew installed, you can run `which mysql || brew install mysql` to install
+`mysql` if necessary.
```
killercup: I'm very skeptical of suggesting to use brew as not everyone is using macOS, and even there, there are other ways to install it (even though I use brew myself and would recommend it).

diesel_cli/README.md
```diff
@@ -87,6 +96,9 @@ Runs the `down.sql` for the most recent migration.
 #### `migration redo`
 Runs the `down.sql` and then the `up.sql` for the most recent migration.
 
+## `diesel print-schema`
+Prints table definitions for database schema.
```
killercup: Ah, thanks! We apparently totally forgot to mention this here!

diesel_cli/README.md
rbalicki2: ah, the reason I suggested brew is that the `mysql` website doesn't really help at all :-D the "install for mac os x" page, for example, references a DMG but provides no links to download it...

rbalicki2: Aaaaand it does.

rbalicki2:
@killercup I've only ever installed the entire postgres package. If there's a way to install `libpq` that doesn't involve installing all of postgres, I could mention that. I figure for most users, they want postgres installed if they are going to be using the postgres features, but that may not be true if this is being used programmatically (e.g. as part of a build process)

diesel_cli/README.md
```diff
@@ -4,6 +4,34 @@ Diesel CLI
...
+Once these dependencies are installed, you can run `cargo install diesel_cli`.
+
+To install the cli without these dependencies, omit the unneeded dependencies from
+the following command:
+
+```
+cargo install diesel_cli --no-default-features --features postgres,sqlite,mysql
+```
+
+The argument passed to `--features` must include no spaces, and be comma-separated.
```
killercup: This works? Interesting! I've only ever seen the space-separated `"postgres sqlite mysql"` syntax, which is also what [cargo install --help](https://github.com/rust-lang/cargo/blob/305bc25d5e105e84ffe261655b46cf74570f6e5b/src/bin/install.rs#L55) mentions. Where did you see this?

rbalicki2:
```
546 ~/Documents/code/diesel > cargo install diesel_cli --force --features sqlite postgres --no-default-features
error: Invalid arguments.

Usage:
    cargo install [options] [<crate>]
    cargo install [options] --list
547 ~/Documents/code/diesel > cargo install diesel_cli --force --no-default-features --features sqlite postgres
error: Invalid arguments.

Usage:
    cargo install [options] [<crate>]
    cargo install [options] --list
```

Can you double check that the space separate version works for you? Maybe I have something weird on my machine

rbalicki2: (omitting either of the features allows the install to proceed)

killercup: Please note the quotes so it's seen as one string argument: `cargo install diesel_cli --force --features "sqlite postgres" --no-default-features`

rbalicki2: 👍  awesome. will fix

diesel_cli/README.md
```diff
@@ -4,6 +4,34 @@ Diesel CLI
...
+Once these dependencies are installed, you can run `cargo install diesel_cli`.
+
+To install the cli without these dependencies, omit the unneeded dependencies from
+the following command:
+
+```
+cargo install diesel_cli --no-default-features --features "postgres sqlite mysql"
+```
+
+The argument passed to `--features` must include no spaces, and be comma-separated.
```
sgrif: I think you meant to remove this line?

diesel_cli/README.md
```diff
@@ -4,6 +4,34 @@ Diesel CLI
...
+---------------
+
+The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
+and [`mysql`][mysql], some of which may already be included on your system. See
+which dependencies are already present with:
+
+```sh
+which psql
+which sqlite3
+which mysql
```
sgrif: I don't think this will work for linux users, where `mysql` the binary is included in a separate package from `libmysql` the object we need to link against.

Eijebong: `ldconfig -p | grep mysql`

diesel_cli/README.md
Eijebong: Yeah, it's weird, in the example it's separated by spaces and has no comma. Maybe you meant it the other way around ?

diesel_cli/README.md
sgrif: screen shot 2017-07-29 at 4 22 42 pm

I don't think there's any command we can give here which will work in a cross-platform way. Do you think we should we just focus on the Diesel CLI specific stuff and assume that folks can figure out whether they have something installed on their own? (Worst case, attempting to install diesel_cli will tell them)

rbalicki2: Sounds good, definitely

PR #1049, opened 2017-07-29
Finish the update guide, actually run its tests

sgrif:
This finishes all the TODOs I had left previously. I also noticed that
we weren't actually running the test suite for this example, so I've
updated it to be run. The suite was failing due to `numeric_expr!` being
inferred, which prompted me to remove the point in the guide mentioning
it as well.

The explanation of `get_result` and `get_results` assumes that someone
is already familiar with `Queryable` and/or `load`, which I don't like.
However I think we will have a separate guide that explains those in
depth, and we can cross-link to that here when we do so.


sgrif:
I'll start using that style for future guides. I don't think it's worth the time to convert this one though.

PR #1050, opened 2017-07-29
Add a schema in depth guide

sgrif:
This guide is meant to go through everything that gets generated by
`table!`. I've also explained `infer_schema!` and friends here, as this
felt like a good place to explain what exactly they do. I'd like to make
the piece on the DSL module more prominent, since that's a common source
of confusion.

This guide does not include an associated example, since this is about
what these macros do more than how to use them.


guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
+Schema in Depth
+===============
+
+In this guide we're going to look at what exactly `infer_schema!`, `infer_table_from_schema!`, and `table!` do. For `table!` we will show a simplified version of the actual code that gets generated, and explain how each piece is relevant to you. If you've ever been confused about what exactly is getting generated, or what `use schema::posts::dsl::*` means, this is the right place to be.
```
killercup: I'd add an intro sentence like

> If you've followed the Getting Started guide, you've already used some of Diesel's tools to work with database schemas.

to set the expectation that people should already know when to use these macros and where to look for more basic info.

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+
+`infer_schema!` is a macro provided by `diesel_codegen` when you have enabled the feature for one or more database backends. The macro will establish a database connection at compile time, query for a list of all the tables, and generate `infer_table_from_schema!` for each one. `infer_schema` will skip any table names which start with `__`.
```
killercup: > is a macro that `diesel_codegen` provides for the backends you specified in your `Cargo.toml`.

killercup: Put "skip if starts with __" on its own line, maybe a note block

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+
+`infer_table_from_schema!` has a similar role. It will establish a database connection at compile time, and query the database for the list of columns for the given table. It will then use this information to generate an invocation of `table!` for you.
+
```
killercup: Maybe add a summary of the relationship between the macros, like

> So, `infer_schema!` generates `infer_table_from_schema!` calls, and `infer_table_from_schema!` in turn generates `table!` calls.

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+You can get the same `table!` invocations that `infer_schema!` would have generated by running `diesel print-schema`. If you do not want to require a database with the proper schema be present to compile your app, `diesel print-schema` is the recommended way to create your schema file. Typically these invocations would be placed in `src/schema.rs`.
```
killercup: Would move this whole paragraph below the next one and maybe even into a note block

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+
+`table!` is where the bulk of the code gets generated. If you wanted to, you could see the actual exact code that gets generated by running `cargo rustc -- -Z unstable-options --pretty=expanded`. However, the output will be quite noisy, and there's a lot of code that won't actually be relevant to you. Instead we're going to go step by step through a *simplified* version of this output, which only has the code which you would use directly.
```
killercup: maybe also link to https://github.com/dtolnay/cargo-expand (which `rustfmt`s the `--pretty=expanded` output and even supports colorized output)

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+        id -> Integer,
+        name -> Text,
+        hair_color -> Nullable<Text>,
+    }
+}
+```
+
+If you just want to see the full simplified code and look through it yourself, you will find it at the end of this guide.
+
+The output of `table!` is always a Rust module with the same name. If you are using `infer_schema!` there will actually be one additional level of nesting like this:
```
killercup: I can't help but ask "why" :) maybe add a little sentence on only exposing certain items?

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+pub use self::infer_users::users;
+```
+
+However, if you're using `table!` directly, you will just have `mod users`. The first and most important part of this module will be the definition of the table itself:
+
+```rust
+pub struct table;
+```
+
+This is the struct that represents the user's table for the purpose of constructing SQL queries. It's usually referenced in code as `users::table` (or sometimes just `users`, more on that in a bit). Next, we'll see there's a module called `columns`, with one struct per column of the table.
```
killercup: "This is the [unit] struct"

I'd actually like to link to official docs or even just a blog post about using unit structs but I couldn't find one.

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+
+```rust
+pub struct id;
+pub struct name;
+pub struct hair_color;
+```
+
+Each of these structs uniquely represents each column of the table for the purpose of constructing SQL queries. Each of these structs will implement a trait called [`Expression`][], which indicates the SQL type of the column.
```
killercup: Replace second "Each of [these] will" with "They will all"

killercup: Explain where types like `Integer` and `Text` come from? (or did you mention `diesel::types` anywhere?)

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+pub struct star;
+
+impl Expression for star {
+    type SqlType = ();
+}
+```
+
+The `star` struct represents `users.*` in the query builder. This struct is only intended to be used for generating count queries. It should never be used directly. Diesel loads your data from a query by index, not by name. In order to ensure that we're actually getting the data for the column we think we are, Diesel never uses `*` when we actually want to get the data back out of it. We will instead generate an explicit select clause such as `SELECT users.id, users.name, users.hair_color`.
+
+Everything in the columns module will be re-exported from the parent module. This is why we can reference columns as `users::id`, and not `users::columns::id`.
```
killercup: "and not [just]"

guide_drafts/schema-in-depth.md
```diff
@@ -0,0 +1,218 @@
...
+Queries can often get quite verbose when everything has to be prefixed with `users::`. For this reason, Diesel also provides a convenience module called `dsl`.
+
+```rust
+pub mod dsl {
+    pub use super::columns::{id, name, hair_color};
+    pub use super::table as users;
+}
+```
+
+This module re-exports everything in `columns` module (except for `star`), and also re-exports the table but renamed to the actual name of the table. This means that instead of writing `users::table.filter(users::name.eq("Sean")).filter(users::hair_color.eq("black"))`, we can instead write `users.filter(name.eq("Sean")).filter(hair_color.eq("black"))`. The `dsl` module should only ever be imported for single functions. You should never have `use schema::users::dsl::*;` at the top of a module. Code like `#[derive(Insertable)]` will assume that `users` points to the module, not the table struct.
```
killercup: Move last to sentences to own paragraph, maybe even note block

guide_drafts/schema-in-depth.md
sgrif: I think the note block calls more attention to it than necessary.

sgrif: I'm not sure I understand what your first correction means? The point I was trying to make was that the macro is only present if you've enabled at least one backend (otherwise it's just the derives)

killercup: I just wanted to give an alternative sentence structure that is a bit easier to read IMO. Took a second to comprehend what you meant by "is a macro when you have enabled …"

guide_drafts/schema-in-depth.md
sgrif: I intentionally left out the `use` statements. You're right that we should specify which exact type these are though.

sgrif:
I'll address the feedback on this as I move it to the website.

PR #1051, opened 2017-07-29
Remove the `Debug` backend, deprecate those macros

sgrif:
This removes the Debug backend, and deprecates the macros which used it
in favor of a new `debug_sql` function, which takes the backend as an
argument. This requires that the `QueryBuilder` implement `Default`
(true for all backends currently supported).

`backend::Debug` has been a common source of bugs. We frequently forget
to make sure that new backend specific types also implement the query
builder for the debug backend. Additionally, it is impossible for us to
make boxed queries work with that structure.

I opted to deprecate those macros in favor of a function, since there's
no established syntax for a type parameter to a macro. The backend can
be inferred for boxed queries, so I would want to make it optional. That
just leads to awful error messages when someone tries to call it without
a backend in a place where it can't be inferred.

Really there was no reason for these to be macros in the first place.
I've attempted to retain their behavior as much as I could, by choosing
an arbitrary backend based on what's enabled.

This change also commits us to doing at least one more release before
1.0. I want to remove all the deprecated things in the 1.0 release.


diesel_tests/tests/debug/mod.rs
```diff
@@ -1,15 +1,25 @@
+use schema::TestBackend;
 use diesel::*;
 
 #[test]
 fn test_debug_count_output() {
     use schema::users::dsl::*;
-    let sql = debug_sql!(users.count());
-    assert_eq!(sql, "SELECT COUNT(*) FROM `users`");
+    let sql = debug_sql::<TestBackend, _>(&users.count());
```
Eijebong: This is the exact same test as https://github.com/diesel-rs/diesel/pull/1051/files#diff-b33826ad230edbe4ba2ad2925b396345R162
Not sure if it's worth keeping

sgrif: Probably not. I suspect this one just predates me writing doctests

sgrif: Perhaps instead of a `debug_sql` function, we should have a `.debug()` instance method which just returns a struct that implements `std::fmt::Debug`

PR #1054, opened 2017-07-30
Export the connection types in `prelude`

sgrif:
It is very rare for you to be importing `prelude::*`, and not want these
types. They are named uniquely enough to not be concerning

### DIFF
--- a/diesel_compile_tests/tests/compile-fail/pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs
+++ b/diesel_compile_tests/tests/compile-fail/pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs
@@ -2,7 +2,6 @@
 #[macro_use] extern crate diesel_codegen;
 
 use diesel::*;
-use diesel::sqlite::SqliteConnection;
 use diesel::types::*;
 use diesel::expression::dsl::*;
 use diesel::pg::upsert::*;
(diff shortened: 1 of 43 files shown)


diesel_compile_tests/tests/compile-fail/pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs
```diff
@@ -1,40 +1,5 @@
...
-    users.select(id).filter(name.is_not_distinct_from("Sean"))
-        .load::<i32>(&connection);
-    //~^ ERROR E0277
-    users.select(id).filter(now.eq(now.at_time_zone("UTC")))
-        .load::<i32>(&connection);
-    //~^ ERROR E0277
-    insert(&NewUser("Sean").on_conflict_do_nothing()).into(users)
-        .execute(&connection);
-    //~^ ERROR type mismatch resolving `<diesel::sqlite::SqliteConnection as diesel::Connection>::Backend == diesel::pg::Pg`
+::SqliteConnection as diesel::Connection>::Backend == diesel::pg::Pg`
```
killercup: Woah, please teach us this extraordinary vim-fu so we too can reduce test cases like this to one-liners :trollface:

Eijebong: Nice one :D

sgrif: Lol I must have hit `dn` on accident instead of `dd` where the last thing I searched for was `sqlite::SqliteConnection`

PR #1055, opened 2017-08-01
Port associations derive to proc macro

weiznich:
Most likely this will break some doc tests to compile because they continue to use the old macros (This needs to be changed before merging)


killercup:
We discussed the possibility of doing this [on Gitter](https://gitter.im/diesel-rs/diesel?at=59804aaa45fc670746e4d170).

sgrif:
My concerns with this PR are the same as the last time it has been brought up. Has anything changed to address these?

- It's harder to test
  - We can no longer test the behavior of the derived code inline, it has to go into the integration testing crate, which makes things harder to isolate and makes feedback cycles longer
- We need to improve our test coverage
  - The last two PRs which did this both introduced regressions. I'm not comfortable accepting a complete rewrite of this code unless we improve our test coverage first
- We lose a convenient place to document things
  - For some of the derives we can document options for the derive on that trait, but that is not the case here. It was nice to have this as a place we could link to in rustdoc that was isolated.
- Circular dependencies have complicated our build process
  - The fact that Diesel now needs diesel_codegen has significantly complicated the release process, and has made running `bin/test` recompile the same crate multiple times, slowing feedback cycles.
- It's not in service of anything specific
  - I'm not against having refactorings in general, but we're right in the middle of a push to 1.0. This doesn't feel like the right time to accept rewrites of this scope.

weiznich:
As stated in the Gitter discussion, this code was originally written to experiment with joining a table multiple times. So it is no big deal to close this pull request.

>  It's harder to test
        We can no longer test the behavior of the derived code inline, it has to go into the integration testing crate, which makes things harder to isolate and makes feedback cycles longer

There is already no test inline for this functionality(As stated below those tests are missing.) Furthermore, I think most likely those tests will also call `#[derive(Associations)]` or implement those traits manually.

>    We need to improve our test coverage
        The last two PRs which did this both introduced regressions. I'm not comfortable accepting a complete rewrite of this code unless we improve our test coverage first

Agreed. For this custom derive the following test cases are needed?

- [ ] Test simple join `child.join(parent)`
- [ ] Test reverse join `parent.join(child)`
- [ ] Test self join `table.join(table)`
- [ ] Test join with custom foreign key name 

>    We lose a convenient place to document things
        For some of the derives we can document options for the derive on that trait, but that is not the case here. It was nice to have this as a place we could link to in rustdoc that was isolated.

Serde [documents](https://serde.rs/derive.html) the usage of the codegen options in a custom guide outside of rustdoc.

>    Circular dependencies have complicated our build process
        The fact that Diesel now needs diesel_codegen has significantly complicated the release process, and has made running bin/test recompile the same crate multiple times, slowing feedback cycles.

The circular dependency is already introduced, so this should not affect this decision. 

>    It's not in service of anything specific
        I'm not against having refactorings in general, but we're right in the middle of a push to 1.0. This doesn't feel like the right time to accept rewrites of this scope.

If diesel 1.0 is released with the current macro based approach this means those macros are part of the stable api. Therefore it will not be possible to remove them without releasing version 2.0 of diesel.

sgrif:
> As stated in the Gitter discussion, this code was originally written to experiment with joining a table multiple times.

I'm not clear on how this makes that any easier? Joining to a table multiple times is something that will require more than just changes to internal code. We would need to introduce a public API for controlling table aliasing, which is currently out of scope for 1.0. Otherwise any time you reference a column from a table it would be ambiguous which instance of the table you're trying to reference.

> The circular dependency is already introduced, so this should not affect this decision.

It is something I would like to eventually remove.

> If diesel 1.0 is released with the current macro based approach this means those macros are part of the stable api. Therefore it will not be possible to remove them without releasing version 2.0 of diesel.

That is a good point. I think we need to discuss whether we want that to happen or not, and what we should do if not.

weiznich:
> I'm not clear on how this makes that any easier? Joining to a table multiple times is something that will require more than just changes to internal code. We would need to introduce a public API for controlling table aliasing, which is currently out of scope for 1.0. Otherwise any time you reference a column from a table it would be ambiguous which instance of the table you're trying to reference.

I've played a bit with implementing this, but got no really useful solution. (At least I've reached a point where it was possible to have multiple `#[belongs_to(XY)]` annotations and where one could join two tables on different foreign keys, but only one (The aliasing was missing). On @killercup pointed out a much simpler solution for my initial problem, so I've stopped there.)
While playing I needed to decide which traits needs to be implemented if there are multiple `#[belongs_to(XY)]` annotations for the same parent table to prevent conflicting implementations for `JoinTo` and `BelongsTo`. Filter those implementations was much easier on proc macro side, than in the original macro rules code, so I've ported things over. 
This pull request contains only the cleaned part of porting the dervie, without any of the other experiments.

sgrif:
I understand. In general I don't think that's going to be a good solution. If you have two `belongs_to` with different foreign keys on the same table, then we don't know which one you meant when you say `parent.inner_join(child)`. I think part of how we solve that is by separating join code from `#[belongs_to]`, which would solve some of your use cases. It would also let you have self-referential associations as long as you aren't trying to write a join.

I've actually been warming up to the idea of inferring `joinable!` from foreign keys. If we were to do that we'd need to warn when two constraints would conflict. One other benefit of doing that is `infer_schema!` and `diesel_print_schema!` could then also generate `enable_multi_table_joins!` for every combination of tables (the last part might not actually be desirable)

weiznich:
> If you have two belongs_to with different foreign keys on the same table, then we don't know which one you meant when you say `parent.inner_join(child)`

My idea for solving this problem was to let the argument of the join methods be something that implement some trait (call it `JoinOnClause` for example). In case of there is only one association between child and parent `JoinOnClause` would be implemented for `parent` and `child`. Otherwise there exist some "default" implementation for `(parent, foreign_key)` and `(child, foreign_key)` which would clearly state which key should be used. (It is quite simple to find out which traits should be implemented in proc macro derives, therefore those code)

> I've actually been warming up to the idea of inferring joinable! from foreign keys. If we were to do that we'd need to warn when two constraints would conflict. 

I've managed to go down this route some steps. I've created a trait `ForeignKey` that got implemented for every foreign key column and contained informations about the child and the parent table. The problem with this is that as soon as you start to add generic implementation of for example `JoinTo` based on this, it is impossible (I've found no way) to also add a generic implementation for the reversed implementation.

killercup:
Re: @sgrif's concerns

- We need to write more tests for this derive
  - → please open new PRs for that -- tests for this can land right now
- Ideally, we should also have a test suite explicitly for derive macros. Not sure how much precedence there is aside from compile-test.
- We need to document the codegen crate. This is already true and becomes more important with each line of code we add to it.
  - Can rustdoc be used with proc-macro crates? I haven't tested this in a while.
  - We can already document this in Readmes and guides, and link to them from the diesel main rustdoc page.
  - → Let's write more docs! (As always ;))
- (This doesn't add build/release issues AFAIK, those already exist.)
- I think this _does_ service Diesel developers, because one can argue it makes the code base easier to contribute to. I'm not sure which of the tow, proc-macros or macro_rules, are harder to write, but currently you need to concern yourself with both. This removes the need to deal with this macro_rules macro in this specific area.

weiznich:
> Can rustdoc be used with proc-macro crates? I haven't tested this in a while.

Yes this seems to be possible.
bildschirmfoto vom 2017-08-02 11-56-42

> I think this does service Diesel developers, because one can argue it makes the code base easier to contribute to. I'm not sure which of the tow, proc-macros or macro_rules, are harder to write, but currently you need to concern yourself with both. This removes the need to deal with this macro_rules macro in this specific area.

For me it is much easier to read proc macro code than macro_rules code parsing struct definitions, because the proc macro code uses the same syntax as normal rust code and could be structured in more or less small functions.

weiznich:
> Circular dependencies have complicated our build process
    The fact that Diesel now needs diesel_codegen has significantly complicated the release process, and has made running bin/test recompile the same crate multiple times, slowing feedback cycles.

I think it should be possible to remove this circular dependency even if we transform all derives to proc macro implementations. The circular dependency exists because `infer_schema!` depends on diesel_infer_schema, that depends on diesel. So if we reorder some things it should work without the circular dependency.
I propose the following change: 

1. Remove everything related to `infer_schema!` from diesel and diesel_codegen (Should be the `infer_schema!` macros and the `derive_infer_schema` custom derive)

-> diesel_codegen does not depend on diesel itself anymore

2. Rename diesel_infer_schema to diesel_infer_schema_intern
3. Create a new crate diesel_infer_schema_codegen for the custom derive.

Optionally apply what @dtolnay has proposed [here](https://github.com/quodlibetor/diesel-newtype/pull/1#issuecomment-315560923):
4. Reexport the diesel_codegen macros from diesel. This eliminates the need to track two distinct crates to use diesel, and potentially unbreaks `embed_migrations!` (It is not possible to build diesel without the needed custom derive anymore)

5. Create a new crate diesel_infer_schema, that contains the `infer_schema!` wrapper macro and reexports the custom derive from diesel_infer_schema_codegen

weiznich:
Rebased to master and removed the `JoinTo` implementation.

weiznich:
I've rebased this again to pull in the new test cases. Everything seems to pass.
@sgrif: Any decision on this yet?

sgrif:
@weiznich If it's ok with you, I'd like to defer this change until after 1.0. We're not planning on doing another release until 1.0, and I'd *really* like to avoid rewriting a major part of the framework in that release.

killercup:
@sgrif wasn't there the concern of this being a breaking change? Can we
avoid this somehow by hiding certain items?


Sean Griffin <notifications@github.com> schrieb am So. 27. Aug. 2017 um
12:19:

weiznich:
@sgrif It's ok for me, I fully understand your concerns.

As mentioned before: changing this after the 1.0 release will be a breaking change, because:
a) It removes public macros from diesel (this could not be solved by hidding some items, because the macros must be public :disappointed: )
b) If we go with the change proposed [here](https://github.com/diesel-rs/diesel/pull/1055#issuecomment-320454637) it will alter the crate structure

So this means such a change cannot happen without releasing diesel 2.0.

killercup:
@weiznich I meant `#[doc(hidden)]`, so it's not perceived as part of the public API. This is, while certainly not ideal, good enough for now; and we already do that in some other places.

killercup:
Also FYI, I have no problem releasing Diesel 2.0 the day general proc-macros become stable. I don't care about the _numbers_ in version numbers as long as the breakage for our users is manageable (say, less than 15min of work for a codebase that casually uses diesel).

weiznich:
> Also FYI, I have no problem releasing Diesel 2.0 the day general proc-macros become stable.

Porting the custom derives only need macros 1.1.

> I don't care about the numbers in version numbers as long as the breakage for our users is manageable (say, less than 15min of work for a codebase that casually uses diesel).

One could use this argument also to justify to release version 0.17 before going to 1.0 (just saying)

killercup:
Sorry to derail the discussion, that was not my intention!

> Porting the custom derives only need macros 1.1.

Sure, but replacing/refactoring _all_ macros (incl. `table!`) would be much more… whole step. One that could potentially allow some other features, like better error reporting for examples.

> One could use this argument also to justify to release version 0.17 before going to 1.0 (just saying)

That argument, sure. But there is also another one for 1.0, which is: It's about focussing on the last remaining features and getting this project ready for prime time. Which is a huge step, and one I'm both excited and cautious about. :)

sgrif:
@weiznich I think we're good to start working on these (one at a time please don't flood us with review requests 😄 ). However, I'm going to close *this* PR since the entire thing basically needs to be rewritten and the discussion on this PR isn't relevant to the actual code being merged.

PR #1057, opened 2017-08-01
Clarify the wording on an infer_schema error message

klieth:
This error message could be interpreted to be saying that when the
dotenv feature is enabled, the string must begin with `dotenv:`. I
believe the new wording clarifies that the issue is that the dotenv
feature has not been enabled in the first place.

### DIFF
--- a/diesel_codegen/src/database_url.rs
+++ b/diesel_codegen/src/database_url.rs
@@ -29,8 +29,8 @@ fn load_dotenv_file() -> Result<(), String> {
 
 #[cfg(not(feature = "dotenv"))]
 fn load_dotenv_file() -> Result<(), String> {
-    Err(String::from("The dotenv feature is required to use strings starting \
-        with `dotenv:`"))
+    Err(String::from("diesel_codegen must be compiled with the dotenv feature \
+        to use strings starting with `dotenv:`"))
 }
 
 #[test]
--- a/diesel_codegen/src/database_url.rs
+++ b/diesel_codegen/src/database_url.rs
@@ -29,8 +29,8 @@ fn load_dotenv_file() -> Result<(), String> {
 
 #[cfg(not(feature = "dotenv"))]
 fn load_dotenv_file() -> Result<(), String> {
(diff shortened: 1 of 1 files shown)


diesel_codegen/src/database_url.rs
```diff
@@ -29,8 +29,8 @@ fn load_dotenv_file() -> Result<(), String> {
 
 #[cfg(not(feature = "dotenv"))]
 fn load_dotenv_file() -> Result<(), String> {
-    Err(String::from("The dotenv feature is required to use strings starting \
-        with `dotenv:`"))
+    Err(String::from("diesel_codegen must be compiled with the dotenv feature \
```
killercup: Can you put "dotenv" in quotes?

diesel_codegen/src/database_url.rs
```diff
@@ -29,8 +29,8 @@ fn load_dotenv_file() -> Result<(), String> {
...
+        to use strings starting with `dotenv:`"))
```
killercup: This is fine, but if we're already editing this, I'd probably replace the "use strings" with something like "correctly substitute strings starting with `dotenv:`"

klieth: using "substitute" seems to me to be exposing the internals of how the library works a little too much, although I admittedly haven't looked at much of the code and don't know what the standard is. How about "to correctly interpret strings starting with" or "to allow passing strings starting with"? I'm also happy to go with your suggestion if you think it's best.

killercup: Good intuition to not expose too many internals!

But… Hm. Dotenv reads environment variables (from the actual environment and from a `.env` file). The `dotenv:FOO` syntax means "replace this with the environment variable `FOO`". So I'm not sure what it could be doing if it's not substituting a string 😅

"Interpreting" is good, too, but a bit vaguer. I like that it anthropomorphizes diesel_codegen a bit, though! :)

klieth: I went with the "interpreting" one; if it still reads poorly when you first read it I'll change it to your suggestion.

PR #1058, opened 2017-08-02
percent_decode for mysql url

### LINKED ISSUE #871: Unable to pass URL components ('#', '/', '?') in password for DATABASE_URL
Hello,

Currently, the connection code uses the url crate to parse the connection string. This is fine, however, percent_encoding/decoding is not currently used in the project which leads the user to have to change their passwords if they use the characters '#', '/', or '?'.

For example: `mysql://root:pass#word@localhost:3006/testing` would result in a panic/error.

The solution to this is to allow the connection string to be: `mysql://root:pass%23word@localhost:3006/testing` and then use the [percent_encoding](http://docs.diesel.rs/url/percent_encoding/index.html) functions to decode the password where it is matched.

This seems to be a limitation of the rust-url crate, where I've reached out to the developers, but to no avail. They are just following the url standard.

nottxy:
The url crate `percent_encode` the input string when `set_password` and `set_username`, so `percent_decode` is needed when `password` and `username` were got.

And this will fix #871.

https://github.com/servo/rust-url/blob/bb1100baf0873b2502b637d7245088f5d5c00a39/src/lib.rs#L1299

https://github.com/servo/rust-url/blob/bb1100baf0873b2502b637d7245088f5d5c00a39/src/lib.rs#L1351

### DIFF
--- a/diesel/src/mysql/connection/url.rs
+++ b/diesel/src/mysql/connection/url.rs
@@ -2,6 +2,7 @@ extern crate url;
 
 use std::ffi::{CStr, CString};
 use self::url::Url;
+use self::url::percent_encoding::percent_decode;
 
 use result::{ConnectionError, ConnectionResult};
 
@@ -32,9 +33,19 @@ impl ConnectionOptions {
             Some(host) => Some(try!(CString::new(host.as_bytes()))),
             None => None,
         };
-        let user = try!(CString::new(url.username().as_bytes()));
+        let username = match percent_decode(url.username().as_ref()).decode_utf8() {
+            Ok(username) => username,
+            Err(_) => return Err(connection_url_error()),
+        };
+        let user = try!(CString::new(username.as_ref().as_bytes()));
         let password = match url.password() {
-            Some(password) => Some(try!(CString::new(password.as_bytes()))),
+            Some(password) => {
+                let password = match percent_decode(password.as_ref()).decode_utf8() {
+                    Ok(password) => password,
+                    Err(_) => return Err(connection_url_error()),
+                };
+                Some(try!(CString::new(password.as_ref().as_bytes())))
+            }
             None => None,
         };
         let database = match url.path_segments().and_then(|mut iter| iter.nth(0)) {
@@ -115,4 +126,28 @@ fn first_path_segment_is_treated_as_database() {
             .unwrap()
             .database()
     );
+}
+
+#[test]
+fn userinfo_should_be_percent_decode() {
+    use self::url::percent_encoding::{USERINFO_ENCODE_SET, utf8_percent_encode};
+
+    let username = "x#gfuL?4Zuj{n73m}eeJt0";
+    let encoded_username = utf8_percent_encode(username, USERINFO_ENCODE_SET);
+
+    let password = "x/gfuL?4Zuj{n73m}eeJt1";
+    let encoded_password = utf8_percent_encode(password, USERINFO_ENCODE_SET);
+
+    let db_url = format!(
+        "mysql://{}:{}@localhost/bar",
+        encoded_username,
+        encoded_password
+    );
+    let db_url = Url::parse(&db_url).unwrap();
+
+    let conn_opts = ConnectionOptions::parse(db_url.as_str()).unwrap();
+    let username = CString::new(username.as_bytes()).unwrap();
+    let password = CString::new(password.as_bytes()).unwrap();
+    assert_eq!(username, conn_opts.user);
+    assert_eq!(password, conn_opts.password.unwrap());
 }
--- a/diesel/src/mysql/connection/url.rs
+++ b/diesel/src/mysql/connection/url.rs
@@ -33,19 +33,9 @@ impl ConnectionOptions {
             Some(host) => Some(try!(CString::new(host.as_bytes()))),
             None => None,
         };
-        let username = match percent_decode(url.username().as_ref()).decode_utf8() {
-            Ok(username) => username,
-            Err(_) => return Err(connection_url_error()),
-        };
-        let user = try!(CString::new(username.as_ref().as_bytes()));
+        let user = try!(decode_into_cstring(url.username()));
         let password = match url.password() {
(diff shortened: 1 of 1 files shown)


golddranks:
Note that the default percent encoding of the `url` crate doesn't encode every special characters in URLs. You need to define your own encoding for them. (For reference: https://github.com/servo/rust-url/issues/353 )

In my experience, Postgres doesn't like having the character `$` in the URL; the URL doesn't work if one's password contains `$`.

Here's an encoding set I defined in my code to circumvent the problem. Feel free to use:

        define_encode_set! {
            // This encode set encodes all the reserved URL delimiters defined
            // in [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.2).
            // Contains the `SIMPLE_ENCODE_SET` and additionally characters :,
            //  /, ?, #, [, ], @, !, $, &, ` , (, ), *, +, ,, ;, =
            pub RESERVED_ENCODE_SET = [SIMPLE_ENCODE_SET] | {
                ':', '/', '?', '#', '[', ']', '@', '!', '$',
                '&','\'', '(', ')', '*', '+', ',', ';', '='
            }
        }

Eijebong:
@golddranks Thanks for the heads up :) MySQL seems to like both "$" and "%24" though and this PR is only targeting MySQL

@nottxy Sorry it took so long for someone to have a look at it :/
Also can you add a test that checks that percent decoding is working ? (There are examples at the bottom of the file).

nottxy:
@Eijebong The test has been added

qmx:
confirmed that it fixes #871

sgrif:
Note that the new PG driver will do URL parsing on our end when it ships, so we will eventually be able to have the behavior of all of these conform to the same thing. The behavior I'm trying to target is roughly "the behavior of libpq, but following the URI standard whenever those two differ"

diesel/src/mysql/connection/url.rs
```diff
@@ -32,9 +33,13 @@ impl ConnectionOptions {
             Some(host) => Some(try!(CString::new(host.as_bytes()))),
             None => None,
         };
-        let user = try!(CString::new(url.username().as_bytes()));
+        let username = percent_decode(url.username().as_ref()).decode_utf8_lossy();
```
sgrif: Is lossy the correct behavior here? If there's an invalid UTF-8 character, I think that should be considered an error condition.

diesel/src/mysql/connection/url.rs
```diff
@@ -110,3 +115,18 @@ fn first_path_segment_is_treated_as_database() {
...
     );
 }
+
+#[test]
+fn userinfo_should_be_percent_decode() {
+    let mut db_url = Url::parse("mysql://localhost/bar").unwrap();
+    let username = "x#gfuL?4Zuj{n73m}eeJt0";
+    let password = "x/gfuL?4Zuj{n73m}eeJt1";
+    db_url.set_username(username).unwrap();
+    db_url.set_password(Some(password)).unwrap();
```
sgrif: I'm finding this test rather difficult to read. Rather than relying on the (implicit to this test) behavior of `set_username` and `set_password` doing percent encoding, can we just have a URL string which includes percent encoded/decoded characters?

diesel/src/mysql/connection/url.rs
nottxy: Thanks, make sense, updated

diesel/src/mysql/connection/url.rs
nottxy: Sure, updated

nottxy:
Guys, Is there any update about this PR?

diesel/src/mysql/connection/url.rs
```diff
@@ -32,9 +33,19 @@ impl ConnectionOptions {
             Some(host) => Some(try!(CString::new(host.as_bytes()))),
             None => None,
         };
-        let user = try!(CString::new(url.username().as_bytes()));
+        let username = match percent_decode(url.username().as_ref()).decode_utf8() {
```
sgrif: Can you use `as_bytes` here to be more explicit what we want?

diesel/src/mysql/connection/url.rs
```diff
@@ -32,9 +33,19 @@ impl ConnectionOptions {
...
+            Ok(username) => username,
+            Err(_) => return Err(connection_url_error()),
+        };
+        let user = try!(CString::new(username.as_ref().as_bytes()));
```
sgrif: Do we need this `as_ref`?

diesel/src/mysql/connection/url.rs
```diff
@@ -32,9 +33,19 @@ impl ConnectionOptions {
...
         let password = match url.password() {
-            Some(password) => Some(try!(CString::new(password.as_bytes()))),
+            Some(password) => {
+                let password = match percent_decode(password.as_ref()).decode_utf8() {
```
sgrif: Can we use `as_bytes` here to be more explicit about what you want?

diesel/src/mysql/connection/url.rs
```diff
@@ -32,9 +33,19 @@ impl ConnectionOptions {
...
+                    Ok(password) => password,
+                    Err(_) => return Err(connection_url_error()),
+                };
+                Some(try!(CString::new(password.as_ref().as_bytes())))
```
sgrif: Do we need this `as_ref`?

diesel/src/mysql/connection/url.rs
sgrif: Can we pull this out into a function?

```rust
fn decode_into_cstring(s: &str) -> ConnectionResult<CString> {
    let decoded = percent_decode(s.as_bytes)
        .decode_utf8()
        .map_err(|_| connection_url_error())?;
    CString::new(decoded.as_bytes()).map_err(Into::into)
}
```

nottxy: updated

nottxy:
@sgrif updated

PR #1059, opened 2017-08-03
Codegen README clarification

alexander-alvarez:


killercup:
Thanks! Didn't know we still had that in there 😅

I don't think the Readme needs to contain the history, though. Can you maybe reduce it down to something like this?

> It depends on features introduced in Rust 1.15. Make sure to always use the latest stable release for optimal performance and feature support.

alexander-alvarez:
I'm new to Rust and Diesel so if you feel the above is better, I have no problem doing it.

killercup:
Thanks. It's not really a matter of Rust or Diesel, it's more my personal opinion that the Readme only needs to describe what is currently true :)

In general, Rust developers tend to update their compiler version quite often, as there are stable releases every 6 weeks and you only need `rustup update` to get the latest version. This makes describing the behavior of older compiler versions less important.

killercup:
Actually, since this is such a small change, I'll just go ahead and do it so you don't have to mess around with adding a commit to that branch (from the "patch-1" name I guessed you used Github's interface). Hope you don't mind.

Thanks again :)

PR #1061, opened 2017-08-03
Assume the type is in scope for {queryable,expression}_impls

Eijebong:
This allow the usage of those two macros with custom types that are not
in diesel::types.

### DIFF
--- a/diesel/src/types/impls/primitives.rs
+++ b/diesel/src/types/impls/primitives.rs
@@ -2,7 +2,8 @@ use std::error::Error;
 use std::io::Write;
 
 use backend::Backend;
-use types::{self, HasSqlType, FromSql, ToSql, ToSqlOutput, IsNull, NotNull};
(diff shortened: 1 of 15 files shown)


diesel/src/types/impls/primitives.rs
```diff
@@ -2,7 +2,8 @@ use std::error::Error;
 use std::io::Write;
 
 use backend::Backend;
-use types::{self, HasSqlType, FromSql, ToSql, ToSqlOutput, IsNull, NotNull};
+use types::{self, HasSqlType, FromSql, ToSql, ToSqlOutput, IsNull, NotNull, Binary, BigInt, Bool,
+            Date, Double, Float, Integer, SmallInt, Text, Time, Timestamp};
```
sgrif: Can we avoid visual indent?

Eijebong: Sure

PR #1062, opened 2017-08-03
Add some "compile" tests for deriving Associations

weiznich:
Any future cases that should be added?

### DIFF
--- a/diesel_codegen/tests/associations.rs
+++ b/diesel_codegen/tests/associations.rs
@@ -40,6 +40,8 @@ fn simple_belongs_to() {
         title: String,
     }
 
+    joinable!(posts -> users(user_id));
+
     let _can_join_tables = posts::table.inner_join(users::table)
         .select((users::id, users::name, posts::id))
         .filter(posts::id.eq(1)
@@ -95,6 +97,8 @@ fn custom_foreign_key() {
         belongs_to_user: i32,
         title: String,
     }
+
+    joinable!(posts -> users(belongs_to_user));
 
 
     let _can_join_tables = posts::table.inner_join(users::table)
--- a/diesel_codegen/tests/associations.rs
+++ b/diesel_codegen/tests/associations.rs
@@ -1,5 +1,12 @@
 #![allow(dead_code, unused_must_use)]
 
+#[cfg(feature = "sqlite")]
+type Backend = ::diesel::sqlite::Sqlite;
+#[cfg(feature = "mysql")]
+type Backend = ::diesel::mysql::Mysql;
+#[cfg(feature = "postgres")]
+type Backend = ::diesel::pg::Pg;
+
 #[test]
 fn simple_belongs_to() {
     use diesel::*;
@@ -25,7 +32,7 @@ fn simple_belongs_to() {
         name: String
     }
 
-    #[derive(Associations)]
+    #[derive(Associations, Identifiable)]
     #[belongs_to(User)]
     pub struct Post {
         id: i32,
@@ -43,7 +50,17 @@ fn simple_belongs_to() {
     users::table.inner_join(posts::table)
         .select((posts::id, posts::user_id, posts::title))
         .filter(users::id.eq(1)
-            .and(users::name.eq("Sean")));
+                .and(users::name.eq("Sean")));
+
+    let t = User { id: 42, name: "Sean".into() };
+
+    let belong_to = Post::belonging_to(&t);
+    let filter = posts::table.filter(posts::user_id.eq(42));
+
+    assert_eq!(
+        debug_sql::<Backend, _>(&belong_to),
+        debug_sql::<Backend, _>(&filter)
+    )
 }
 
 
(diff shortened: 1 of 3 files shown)


sgrif:
I'd like us to test that these are actually generating the code we expect. We should execute the queries and expect to see some data back.

killercup:
Thanks!

I fear this will fail CI with `must_use` errors, though. Let's add insert some data and assert the results.

weiznich:
> I'd like us to test that these are actually generating the code we expect. We should execute the queries and expect to see some data back.

There are already tests doing this in diesel_test. So I'm not sure if we should add those tests on two location? 

> I fear this will fail CI with `must_use errors`, though. Let's add insert some data and assert the results.

There is an `#![allow(unused_must_use)]` at the top of the file to prevent this errors.

killercup:
> There is an #![allow(unused_must_use)] at the top of the file to prevent this errors.

Lol, I apparently can't read. …And I've just been drinking tea today before anyone asks! 😅

sgrif:
The only test I can see which executes a query involving a `foreign_key` option is the test for self referencing associations.

sgrif:
We need to provide *some* tests for this behavior, even if it's not testing the actual results of executing the query. I'd like to at least test that `foreign_key` is doing the right thing, and probably a test like:

```rust
assert_eq!(
    debug_sql(Foo::belonging_to(&bar)),
    debug_sql(foos.filter(bar_id.eq(bar.id))),
);
```

weiznich:
>  I'd like to at least test that foreign_key is doing the right thing, and probably a test like:

I've adjusted the diesel_codegen tests to check the generated sql

> The only test I can see which executes a query involving a foreign_key option is the test for self referencing associations.

I've added one test to diesel_tests for this scenario.

diesel_codegen/tests/associations.rs
```diff
@@ -0,0 +1,149 @@
...
+        user_id: i32,
+        title: String,
+    }
+
+
+    posts::table.inner_join(users::table)
+        .select((users::id, users::name))
+        .filter(posts::id.eq(1)
+            .and(posts::user_id.eq(2))
+                .and(posts::title.eq("Bar")));
```
killercup: Can you name these "compile-success assertions" by making them let statements? E.g., `let _can_select_columns_from_join = posts::table.inn…` (or a better name)

diesel_codegen/tests/associations.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+
+    #[derive(Associations, Identifiable)]
+    #[belongs_to(Tree, foreign_key = "parent_id")]
+    pub struct Tree {
+        id: i32,
+        parent_id: Option<i32>,
+    }
+    let t = Tree{ id: 42, parent_id: None };
```
killercup: pls add space before `{` (sorry my brain zooms in on inconsistencies like that)

diesel_tests/tests/associations.rs
```diff
@@ -187,3 +187,70 @@ fn conn_with_test_data() -> (TestConnection, User, User, User) {
...
+        id: i32,
+        name: String
+    }
+    #[derive(Debug, Queryable, Associations, Identifiable, Clone, PartialEq)]
+    #[belongs_to(User, foreign_key="belongs_to_user")]
+    pub struct Post {
+        id: i32,
+        belongs_to_user: i32,
+        title: String,
+    }
```
killercup: Please add empty lines before each of the above items where missing

diesel_tests/tests/associations.rs
```diff
@@ -187,3 +187,70 @@ fn conn_with_test_data() -> (TestConnection, User, User, User) {
...
+
+    table! {
+        posts {
+            id -> Integer,
+            belongs_to_user -> Integer,
+            title -> Text,
+        }
+    }
+
+    #[derive(Debug, Queryable, Identifiable, PartialEq, Clone)]
```
killercup: Can you put diesel's derives last? Or… sort them alphabetically? Or anything that I can identify as 'order'? Or reveal the system you ordered them by? :D Sorry for commenting on all the nits

diesel_tests/tests/associations.rs
```diff
@@ -187,3 +187,70 @@ fn conn_with_test_data() -> (TestConnection, User, User, User) {
...
+        belongs_to_user: i32,
+        title: String,
+    }
+    let sean = User { id: 1, name: "Sean".into() };
+    let tess = User { id: 2, name: "Tess".into() };
+    let post1 = Post { id: 1, belongs_to_user: 1, title: "Hello".into() };
+    let post2 = Post { id: 2, belongs_to_user: 2, title: "World".into() };
+    let post3 = Post { id: 3, belongs_to_user: 1, title: "Hello 2".into() };
+
+
```
killercup: ✂️

diesel_codegen/tests/associations.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    let t = User { id: 42, name: "Sean".into() };
+
+    let belong_to = Post::belonging_to(&t);
+    let filter = posts::table.filter(posts::user_id.eq(42));
+
+    assert_eq!(
+        debug_sql::<Backend, _>(&belong_to),
+        debug_sql::<Backend, _>(&filter)
+    )
```
killercup: The macro may return `()`, but in other tests we always append a `;`

weiznich:
Any idea why the tests are failing? It builds fine for me locally using the same rust version (1.19.0)

PR #1064, opened 2017-08-03
Document the `optional` method

sgrif:
I think this was the first time I've written a doctest for Diesel which
required 0 setup. \o/

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -82,7 +82,29 @@ pub enum ConnectionError {
 pub type QueryResult<T> = Result<T, Error>;
 pub type ConnectionResult<T> = Result<T, ConnectionError>;
 
+/// See the [method documentation](#tymethod.optional).
 pub trait OptionalExtension<T> {
+    /// Converts a `QueryResult<T>` into a `QueryResult<Option<T>>`.
+    ///
+    /// By default, Diesel treats 0 rows being returned from a query that is expected to return 1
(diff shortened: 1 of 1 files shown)


diesel/src/result.rs
```diff
@@ -82,7 +82,26 @@ pub enum ConnectionError {
 pub type QueryResult<T> = Result<T, Error>;
 pub type ConnectionResult<T> = Result<T, ConnectionError>;
 
+/// See the [method documentation](#tymethod.optional).
```
killercup: This link depends on the way rustdoc adds ids to certain elements on the page. I don't think this is public API and will work in future version of rustdoc. But let's keep it for now and hopefully just upgrade to glorious intra-rustdoc links later!

diesel/src/result.rs
```diff
@@ -82,7 +82,26 @@ pub enum ConnectionError {
...
 pub trait OptionalExtension<T> {
+    /// Converts a `QueryResult<T>` into a `QueryResult<Option<T>>`.
+    ///
+    /// By default, Diesel treats 0 rows being returned from a query that is
+    /// expected to return 1 row as an error. This method will handle that
```
killercup: "…as constructed with `get_result`"

sgrif: I'm not sure I understand where you want me to put that

killercup: Sorry, more context:

> By default, Diesel treats 0 rows being returned from a query that expected to return 1 row (as e.g. constructed with `get_result`) as an error.

PR #1065, opened 2017-08-03
Document the `ops` module

sgrif:
This is the same doc comment for each of the traits, other than the
operation name changing. It really bothers me how many other words need
to change in English when you change "add" to "divide"

### DIFF
--- a/diesel/src/types/ops.rs
+++ b/diesel/src/types/ops.rs
@@ -2,23 +2,60 @@
 //!
 //! Crates which add new types which allow numeric operators should implement these traits to
 //! specify what the output is for a given right hand side
+
+/// Represents SQL types which can be added.
+///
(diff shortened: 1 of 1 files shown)


diesel/src/types/ops.rs
```diff
@@ -2,23 +2,60 @@
...
     type Rhs;
+    /// The SQL type of the result of multiplying `Self` by `Rhs`
     type Output;
 }
 
+/// Represents SQL types which can be divided.
+///
+/// Similar to `std::ops::Div`, but this only includes information about the SQL types that will
+/// result from the operation. Unlike `std::ops::Div`, the right side is an associated type rather
+/// than a type parameter. This means that a given SQL type can only divided with one other SQL
```
killercup: "[be] divided [by]"

PR #1070, opened 2017-08-04
Infer `joinable!` from foreign keys in the DB

sgrif:
I had hoped this would be less of a pain than it turned out to be. This
was another case where MySQL doesn't properly follow the
`information_schema` standard and we had to give it special treatment.
We also had several tests which were dropping a table which now has a
foreign key (no backend agnostic way to do that), as well as one test
which was violating referential integrity (it was less of a pain to
disable/defer the checks than to insert the required data).

There are several things this PR does not yet handle, but we should do
before we release:

- Dump this in `diesel print-schema` as well
- Ensure that the child and parent table are both in the tables we are
  generating `table!` invocations for
- Skip FKs which don't point at the primary key
- Skip FKs with more than one column
- Skip FKs when there are more than one of them linking two tables

### DIFF
--- a/diesel_infer_schema/src/inference.rs
+++ b/diesel_infer_schema/src/inference.rs
@@ -4,7 +4,7 @@ use diesel::prelude::*;
 use diesel::result::Error::NotFound;
 
 use table_data::TableData;
-use data_structures::{ColumnInformation, ColumnType};
+use data_structures::{ColumnInformation, ColumnType, ForeignKeyConstraint};
 
 pub enum InferConnection {
     #[cfg(feature = "sqlite")]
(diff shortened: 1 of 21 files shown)


diesel_infer_schema/src/sqlite.rs
```diff
@@ -45,6 +59,28 @@ pub fn load_table_names(connection: &SqliteConnection, schema_name: Option<&str>
...
+                .map(|row| {
+                    let parent_table = TableData::from_name(row.parent_table);
+                    ForeignKeyConstraint {
+                        child_table: child_table.clone(),
+                        parent_table,
+                        foreign_key: row.foreign_key,
+                    }
+                }).collect())
+        }).collect::<QueryResult<Vec<Vec<_>>>>()?;
+    Ok(rows.into_iter().flat_map(|x| x).collect())
```
Eijebong: Oh, I didn't know `flat_map`, neat

diesel_infer_schema/src/sqlite.rs
```diff
@@ -24,6 +24,20 @@ table!{
...
+        _table -> VarChar,
+        from -> VarChar,
+        to -> VarChar,
+        on_update -> VarChar,
+        on_delete -> VarChar,
+        _match -> VarChar,
+    }
+}
+
+
```
Eijebong: ✂️ ?

diesel_infer_schema/src/inference.rs
```diff
@@ -130,3 +130,18 @@ pub fn get_primary_keys(
...
+pub fn load_foreign_key_constraints(database_url: &str, schema_name: Option<&str>)
+    -> Result<Vec<ForeignKeyConstraint>, Box<Error>>
+{
+    let connection = try!(establish_connection(database_url));
+
+    match connection {
+        #[cfg(feature = "sqlite")]
+        InferConnection::Sqlite(c) => ::sqlite::load_foreign_key_constraints(&c, schema_name),
+        #[cfg(feature = "postgres")]
+        InferConnection::Pg(c) => ::information_schema::load_foreign_key_constraints(&c, schema_name).map_err(Into::into),
```
Eijebong: Why isn't this one in `::postgres` ?

sgrif: It's not doing anything PG specific.

Eijebong: Well, it only works for PG, I'd say it's PG specific :p

sgrif: It will also work for SQL Server, Oracle, Spanner, and any other database which correctly implements the `information_schema` ANSI standard.

Eijebong: Assuming they do. But yeah ok, it makes sense

PR #1074, opened 2017-08-05
Deriving AsChangeset on a structure containing only the pk is now unsupported 

Eijebong:
It will now panic with a meaningful message instead of throwing a syntax
error.


sgrif:
Meaningful advice "If you want to change the primary key of a row, you should do so with `.set(table::id.eq(new_id))`. `AsChangeset` never changes the primary key of a row."

Eijebong:
Yeah, I can add that 👍

sgrif:
Can we add a test as wellL?

Eijebong:
Wait, that commit message is so bad

PR #1075, opened 2017-08-05
Flush pending results in `MysqlConnection#execute` and `batch_execute`

### LINKED ISSUE #728: Additional issues with `r2d2-diesel` and MySQL
Thanks for [PR 727](https://github.com/diesel-rs/diesel/pull/727) and the new release (`0.11.1`)! Your speed and willingness to help is amazing!

Unfortunately, I'm encountering further issues when using r2d2 with MySQL. Whenever I try to run a query, like:

    users::table.limit(5).load::<User>(&*conn).expect("Users load failed")

I get 

    DatabaseError(__Unknown, "Commands out of sync; you can\'t run this command now")

When I get a connection directly using `MysqlConnection::establish`, everything works fine.

I dug around a little bit, and it is definitely related to the query in the `is_valid` health check. I put together an alternate impl for `::r2d2::ManageConnection` where `is_valid` always returned `Ok(())`, and it worked fine. (Of course, that is not a viable solution for the managed connection.)

sgrif:
Whenever any call to `mysql_query` (the fake one) or `mysql_real_query`
(make sure you don't use the fake one) contains anything which could
return a result, the caller is responsible for what is effectively
flushing the network buffer. MySQL is not reading any packets from the
response until after we call things like `mysql_next_result`. If we
don't ensure that we've cleared those packets out, we'll get an error
indicating that commands are being run out of sync. This means that any
migration containing a `SELECT` will fail, and our `SELECT 1` ping from
r2d2-diesel won't work.

Fixes #728

### DIFF
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -79,7 +79,9 @@ impl RawConnection {
                 query.len() as libc::c_ulong,
             );
         }
-        self.did_an_error_occur()
+        self.did_an_error_occur()?;
+        self.flush_pending_results()?;
+        Ok(())
     }
 
     pub fn enable_multi_statements<T, F>(&self, f: F) -> QueryResult<T> where
@@ -94,12 +96,6 @@ impl RawConnection {
         self.did_an_error_occur()?;
 
         let result = f();
-
-        unsafe {
-            while ffi::mysql_next_result(self.0) != -1 {
-                self.did_an_error_occur()?;
-            }
-        }
 
         unsafe {
             ffi::mysql_set_server_option(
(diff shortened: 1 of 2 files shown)


diesel/src/mysql/connection/raw.rs
```diff
@@ -142,6 +138,35 @@ impl RawConnection {
             ))
         }
     }
+
+    fn flush_pending_results(&self) -> QueryResult<()> {
+        self.consume_current_result()?;
+        while self.next_result()? {
+            self.consume_current_result()?;
+        }
+        self.consume_current_result()?;
```
killercup: We just talked about this, so I know that for some reason this needs to be used in this way, but otherwise I'd have guessed that at least one of those calls to `self.consume_current_result` was a copypasta error… Can you add a short comment citing some mysql doc/source and say that this is on purpose?

diesel/src/mysql/connection/raw.rs
```diff
@@ -142,6 +138,35 @@ impl RawConnection {
...
+        Ok(())
+    }
+
+    fn consume_current_result(&self) -> QueryResult<()> {
+        unsafe {
+            let res = ffi::mysql_store_result(self.0);
+            if !res.is_null() {
+                ffi::mysql_free_result(res);
+            }
+            self.did_an_error_occur()
```
killercup: This is not actually unsafe to call, is it? (I'd keep the unsafe section as short as possible)

diesel/src/mysql/connection/raw.rs
```diff
@@ -142,6 +138,35 @@ impl RawConnection {
...
+        }
+    }
+
+    /// Calls `mysql_next_result` and returns whether there are more results
+    /// after this one.
+    fn next_result(&self) -> QueryResult<bool> {
+        unsafe {
+            let more_results = ffi::mysql_next_result(self.0) == 0;
+            self.did_an_error_occur()?;
+            Ok(more_results)
```
killercup: Same here. Would only put the ffi call in an unsafe block

PR #1077, opened 2017-08-07
Don't infer `joinable!` for joins which are unsupported

sgrif:
We will no longer infer `joinable!` for any of the following types of
foreign keys:

- Foreign keys between tables excluded from `infer_schema!`
- Self-referential FKs
- Foreign keys which reference a column other than the primary key
- Composite foreign keys
- Foreign keys where there are more than one foreign key linking two
  tables (we could theoretically infer joinable for one of them, but I
  think figuring out which one would be confusing to users)

There are no explicit tests needed for this, as the mere presence of
these tables makes `infer_schema!` blow up without this change.


sgrif:
> Should we add something to the changelog to explain what kind of relations will not get inferred.

I think it should be documented, but the changelog is not the place to do it. Maybe in the docs for `joinable!` itself? (with links in the docs for `infer_schema!` and `diesel print-schema`)

PR #1079, opened 2017-08-07
Move `default_schema` handling into table loading

sgrif:
The concept of `default_schema` is only a thing for `information_schema`
implementors, so we should handle it there.


killercup:
Not sure why but this breaks some print_schema tests

sgrif:
It's due to how we're doing whitelisting. I'm going to roll this into another PR which rewrites `print-schema`

PR #1080, opened 2017-08-07
Remove codegen from `diesel_infer_schema`

sgrif:
This separates the code used by `infer_schema!` from the code used by `diesel print-schema`. When I changed `diesel print-schema` to include `joinable!` as well, the formatting was messed up and I didn't want to try and finangle our crappy string formatter to do the right thing.

I've never felt good about having a module called `codegen` in `diesel_infer_schema`, as that was a crate that was purely meant to include the functions to return information about the schema. It shouldn't care about how it's presented. The code itself was messier than it should have been as a result.

Additionally, the way we were doing formatting for `diesel print-schema` was about as backwards as it could possibly be. Rather than formatting the closest thing we have to an AST for this, we were instead converting that to tokens, then converting *that* to a String, and trying to format the string instead.

There's no reason for diesel CLI to deal with Rust tokens here, since we're only dealing with macro invocations which would not otherwise be valid Rust syntax. I've rewritten that code to be based on the `std::fmt` module instead. This also fixed a long-standing bug with the indentation of modules.


diesel_cli/src/print_schema.rs
```diff
@@ -0,0 +1,163 @@
...
+
+struct TableDefinition<'a>(&'a TableData);
+
+impl<'a> Display for TableDefinition<'a> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "table! {{")?;
+        {
+            let mut out = PadAdapter::new(f);
+            write!(out, "\n{} (", self.0.name)?;
+            for (i, pk) in self.0.primary_key.iter().enumerate() {
```
sgrif: Only on slices, and that would require allocating an intermediate string

diesel_cli/src/print_schema.rs
```diff
@@ -0,0 +1,163 @@
...
+struct TableDefinitions(Vec<TableData>);
+
+impl Display for TableDefinitions {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let mut is_first = true;
+        for table in &self.0 {
+            if is_first {
+                is_first = false;
+            } else {
+                writeln!(f, "")?;
```
sgrif: IIRC clippy warns but I will give it a shot

PR #1081, opened 2017-08-07
Refactor encoding `BigDecimal` on PG

sgrif:
I had originally worked on this as part of 0.15.2, but have been
sitting on it until a new version was released so we could use
`to_bigint_and_exponent`.

The new code takes a different approach to the encoding. I've
specifically broken the code into a `From` impl to make it easier for us
to test the output.

There was an impedence mismatch in the old code. Ultimately the way
`BigDecimal` is represented is as a `BigInteger` with a scale. The way
we're transmitting it is as a `BigInteger` with a weight and scale.
Splitting it into separate integral and fractional parts, and applying
different logic was making the code difficult to understand, and
introducing unneccessary complications.

There are two differences to how `BigUint` is represented vs how PG
wants it. The first is that PG is base 10000, while `BigUint` is base
2^16. The second is that in `BigUint`, `scale` represents both the
number of significant digits, *and* the position of the decimal point.
In PG, `scale` only represents the number of significant digits, and
there is a separate `weight` to represent the position of the decimal
point.

This means that `BigUint` will store trailing zeroes that we need
to strip, but PG will need to ensure that the decimal falls on a digit
boundary in base 10k. The majority of the logic here is to handle that
portion. I've tried to make it as clear as possible what's going on
through proper naming. Unfortunately the fairly simple code of iterating
over the digits in base 10k gets drowned out by the cleanup.

The code could potentially be simplified if we could easily iterate over
the digits in big-endian order, but I could find an easy way to do that
mathematically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   `grandchild.join(child.join(parent))`. Previously only
   `parent.join(child.join(grandchild))` would compile.
 
+* When encoding a `BigDecimal` on PG, `1.0` is no longer encoded as if it were
+  `1`.
+
 [bigdecimal-0.16.0]: https://crates.io/crates/bigdecimal
 [range-0.16.0]: http://docs.diesel.rs/diesel/pg/types/sql_types/struct.Range.html
 
--- a/diesel/src/pg/types/numeric.rs
+++ b/diesel/src/pg/types/numeric.rs
@@ -5,91 +5,62 @@ mod bigdecimal {
     extern crate num_integer;
     extern crate bigdecimal;
 
+    use self::bigdecimal::BigDecimal;
+    use self::num_bigint::{Sign, BigInt, BigUint};
+    use self::num_integer::Integer;
+    use self::num_traits::{Signed, Zero, ToPrimitive};
     use std::error::Error;
     use std::io::prelude::*;
 
     use pg::Pg;
-
-    use self::num_traits::{Signed, Zero, ToPrimitive};
-    use self::num_bigint::{Sign, BigInt, BigUint, ToBigInt};
-    use self::num_integer::Integer;
-    use self::bigdecimal::BigDecimal;
-
     use pg::data_types::PgNumeric;
     use types::{self, FromSql, ToSql, ToSqlOutput, IsNull};
 
-    type Digits = Vec<i16>;
+    /// Iterator over the digits of a big uint in base 10k.
+    /// The digits will be returned in little endian order.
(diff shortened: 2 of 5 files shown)


diesel/src/pg/types/numeric.rs
```diff
@@ -5,91 +5,62 @@ mod bigdecimal {
...
+    /// Iterator over the digits of a big uint in base 10k.
+    /// The digits will be returned in little endian order.
+    struct ToBase10000(Option<BigUint>);
+
+    impl Iterator for ToBase10000 {
+        type Item = i16;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            self.0.take().map(|v| {
+                let (div, rem) = v.div_rem(&BigUint::from(10000u16));
```
sgrif: /cc @cuviper It would be great if there were a way to do this without allocating.

cuviper: Hmm, yeah, we now have scalar ops, but not for a combined `div_rem`. I could see adding something like `div_rem_scalar`, maybe generic or just for `u32`.

sgrif: If `Div<u16>` returned `u16` instead of `BigUint` I'd happily use that instead. It seems like all the scalar operations are just converting to the `num` type under the hood?

cuviper: You're right that the output of `Rem` could have matched the divisor type. I guess we were just thinking to use consistent output for all ops. That could be reconsidered in a breaking change.

You do need all digits for computing `Rem` though, unless the divisor shares all prime factors with the base. (In this case just powers of 2.)

CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   `grandchild.join(child.join(parent))`. Previously only
   `parent.join(child.join(grandchild))` would compile.
 
+* When encoding a `BigDecimal` on PG, `1.0` is no encoded as if it were `1`.
```
Eijebong: now

diesel/src/pg/types/numeric.rs
```diff
@@ -5,91 +5,62 @@ mod bigdecimal {
...
+            let digits_after_decimal = scale as u16 / 4 + 1;
+            let weight = digits.len() as i16 - digits_after_decimal as i16 - 1;
+            let index_of_decimal = (weight + 1) as usize;
+
+            let unneccessary_zeroes = digits[index_of_decimal..]
+                .iter()
+                .rev()
+                .take_while(|i| i.is_zero())
+                .count();
+            let relevent_digits = digits.len() - unneccessary_zeroes;
```
Eijebong: relevant

PR #1082, opened 2017-08-08
Add support for the Datetime type for MySQL

Eijebong:
This is mostly the same as for the Timestamp type so I reused both FromSql and
ToSql implementations.

### DIFF
--- a/diesel/src/mysql/types/date_and_time.rs
+++ b/diesel/src/mysql/types/date_and_time.rs
@@ -8,7 +8,7 @@ use std::os::raw as libc;
 use std::{ptr, mem, slice};
 
 use mysql::Mysql;
-use types::{ToSql, ToSqlOutput, FromSql, IsNull, Timestamp, Time, Date};
+use types::{ToSql, ToSqlOutput, FromSql, IsNull, Timestamp, Time, Date, Datetime};
 
 macro_rules! mysql_time_impls {
     ($ty:ty) => {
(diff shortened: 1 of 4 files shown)


diesel/src/mysql/types/date_and_time.rs
```diff
@@ -41,10 +41,24 @@ macro_rules! mysql_time_impls {
     }
 }
 
+mysql_time_impls!(Datetime);
+primitive_impls!(Datetime -> NaiveDateTime);
 mysql_time_impls!(Timestamp);
 mysql_time_impls!(Time);
 mysql_time_impls!(Date);
 
+impl ToSql<Datetime, Mysql> for NaiveDateTime {
```
sgrif: Could this be `impl<T: ToSql<Timestamp, Mysql>> ToSql<Datetime, Mysql> for T`?

Eijebong:
```
error[E0119]: conflicting implementations of trait `types::FromSql<mysql::types::Datetime, mysql::backend::Mysql>` for type `std::borrow::Cow<'_, _>`:
  --> src/mysql/types/date_and_time.rs:56:1
   |
56 | / impl<T: ToSql<Timestamp, Mysql>> FromSql<Datetime, Mysql> for T {
57 | |     fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
58 | |         <T as FromSql<Timestamp, Mysql>>::from_sql(bytes)
59 | |     }
60 | | }
   | |_^ conflicting implementation for `std::borrow::Cow<'_, _>`
   | 
  ::: src/types/impls/primitives.rs
   |
91 | / impl<'a, T: ?Sized, ST, DB> FromSql<ST, DB> for Cow<'a, T> where
92 | |     T: 'a + ToOwned,
93 | |     DB: Backend + HasSqlType<ST>,
94 | |     T::Owned: FromSql<ST, DB>,
...  |
98 | |     }
99 | | }
   | |_- first implementation here
```

sgrif: :+1:

alanpoon:
The trait `diesel::Expression` is not implemented for `diesel::types::Datetime`, when I try to derive Insertable a Row struct that has field of type diesel::mysql::types::Datetime.

Issue #967, opened 2017-06-26
Allow to rename table columns in table! macro and do it automatically for invalid names


Fiedzia:
I have a table with a column named "mod", which is not valid identifier in Rust.
Even for valid identifiers, I want to be able to use different name than what database provides,
is that possible?

Same thing should be applied to table names.

killercup:
Yes, this is something we want to allow. There is a previous discussion (probably outdated a bit by now) in #424.

@Fiedzia do you have a concrete suggestion for the syntax to rename columns in `table!`?

Fiedzia:
The naive solution would be to add some for of "real name" as foo -> Integer,
but ideally table! enhancement should take into account other things one may want to set there.

I consider Django ORM to be a golden standard in this regard because it incorporates
a lot of domain knowledge in this aspect (among other reasons), and for example here are some of the things it allows to set in field definition:

* nullability
* can the value be blank
* choices (this one is also important for me)
* column name
* if an index should be created for this field (multifield indexes are property of a table)
* tablespace
* default value
* if it is primary key (multifield indexes are property of a table)
* human-readable field name
* validators
* (some other django-related bits I'll ignore here)

as well as properties specific for particular field type:

* precision (for decimal fields)
* max_length for text fields

https://docs.djangoproject.com/en/1.11/ref/models/fields/

Perhaps not all of it makes sense for Diesel, but I'd prefer to have generic and extensible solution for setting field and table attributes.

As  for syntax, I'd suggest something similar to Django:

table! {
    users {
        id -> Integer(primary_key: true),
        name -> VarChar(db_column: "username", max_length:120),
        properties:
            indexes: ...
            db_table: "user"
    }
}

(I find ":" more readable than "=")

sgrif:
Let's keep the feature focused. The use case we want to support here is renames, we don't need to complicate things by adding a bunch of other information that Diesel does not care about (and likely never will).

I'm more in favor of using attributes for this, but want to make sure the naming makes it very clear what is the Rust side and what is the SQL side.

```rust
table! {
    users {
        id -> Integer,
        #[database_name="username"]
        name -> Text
    }
}
```

I'd also like to see a concrete proposal for the conventional renames for Rust keywords.

Fiedzia:
> #[database_name="username"]

I don't mind the format for setting attributes, but I'd prefer to all this attribute column_name or database_column_name - that makes clear what it does even for someone not familiar with diesel. 

[...]

PR #1084, opened 2017-08-09
Add a way to rename a column in the table! macro

Eijebong:
The renaming is done with a `real_name` attribute.

2 passes were added to the table macro.
The first one parses the `real_name` attribute if there is no
`current_column_real_name`.
The second one adds a default `current_column_real_name` if we find a column
definition without any.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -1,7 +1,7 @@
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __diesel_column {
-    ($($table:ident)::*, $column_name:ident -> ($($Type:tt)*), $($doc:expr),*) => {
+    ($($table:ident)::*, $column_name:ident -> ($($Type:tt)*),  $sql_name:expr, $($doc:expr),*) => {
         $(
             #[doc=$doc]
         )*
@@ -20,7 +20,7 @@ macro_rules! __diesel_column {
             fn walk_ast(&self, mut out: $crate::query_builder::AstPass<DB>) -> $crate::result::QueryResult<()> {
                 $($table)::*.from_clause().walk_ast(out.reborrow())?;
                 out.push_sql(".");
-                out.push_identifier(stringify!($column_name))
+                out.push_identifier($sql_name)
             }
         }
 
@@ -69,7 +69,7 @@ macro_rules! __diesel_column {
             type Table = $($table)::*;
 
             fn name() -> &'static str {
-                stringify!($column_name)
+                $sql_name
             }
         }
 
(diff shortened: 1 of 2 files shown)


killercup:
Sweet! Will review shortly

cc #967 #342

diesel/src/macros/mod.rs
```diff
@@ -198,6 +198,23 @@ macro_rules! __diesel_column {
 /// # fn main() {}
 /// ```
 ///
+/// If you have a column with the same name as a rust reserved keyword, you can use
```
killercup: Nit: Is the name of the programming language we use written as proper name with a capital letter at the start or is lowercase okay? This looks wrong.

diesel/src/macros/mod.rs
```diff
@@ -198,6 +198,23 @@ macro_rules! __diesel_column {
...
+/// the `real_name` attribute like this:
+///
+/// ```
+/// #[macro_use] extern crate diesel;
+///
+/// table! {
+///     posts {
+///         id -> Integer,
+///         #[real_name = "type"]
+///         /// This column is named `mytype` but references the table `type` column.
```
killercup: Hm. Should the comment be above or below the attr? Intuitively. I'd say above.

diesel/src/macros/mod.rs
```diff
@@ -411,10 +429,11 @@ macro_rules! table_body {
         table_name = $name:ident,
         primary_key_ty = $primary_key_ty:ty,
         primary_key_expr = $primary_key_expr:expr,
-        columns = [$($column_name:ident -> $Type:tt; doc = [$($doc:expr)*],)*],
+        columns = [$($column_name:ident -> $Typey:tt; doc = [$($doc:expr)*]; real_name = $real_name:expr,)*],
```
killercup: "Type Y"? What's the "y" for? (Personally, I'd of course change it to Typey McTypeface!)

Eijebong: Haha woups, I forgot that, it was to see which invocation was failing. (I had to recompile rust to remove the "something happened in this macro" but then I couldn't distinguish which one it was, so I had typey, typee, typei, well you see the idea

diesel/src/macros/mod.rs
Eijebong: Should be `Rust`, you're right

sgrif:
I don't like `real_name` as the attribute name, but we can change that later LGTM (I do need to refactor the `$doc` stuff though. :\)

Eijebong:
Oh yeah, we never actually settled on anything.

kardeiz:
I haven't looked too closely but does this cover renaming table names as well? Sorry if I've missed something or if this is handled elsewhere&mdash;I haven't followed this too closely since my PR #424 (which did cover renaming tables).

Eijebong:
Nope, it only covers column renaming.

Issue #1056, opened 2017-08-01
Infer `operator_allowed!` for Timestamp types


sgrif:
We already do this for numeric types, but date/time types specifically only allow add and sub. We should infer the required calls to `operator_allowed!` for those, the same way that we infer `numeric_expr!` in `table!` today.

juliusdelta:
Should this issue be closed?

Eijebong:
Yep, thanks

PR #1094, opened 2017-08-13
Added operator_allowed! for Timestamp types

csharad:
For issue #1056. This is my first PR ever. Hope it works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ target
 Cargo.lock
 !diesel_cli/Cargo.lock
 .env
-.idea/
-*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ target
 Cargo.lock
 !diesel_cli/Cargo.lock
 .env
+.idea/
+*.iml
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@ target
 Cargo.lock
 !diesel_cli/Cargo.lock
 .env
+.idea/
+*.iml
```
killercup: Can you instead add those to you global gitignore file?

Eijebong:
I think that error is because we're manually allowing `Add` for this column in the test suite. The line needs to be removed.

csharad:
Only `Timestamp` and `Date` are `Add` and `Sub`. Should I remove others from `__diesel_generate_ops_impls_if_date_time` or add impls for them?

killercup:
This LGTM, but

r? @Eijebong

Eijebong:
Why aren't other types `Add` and `Sub` ? I guess it doesn't really matter for this PR since we can easily add them later to the list.

killercup:
Thanks! We should probably add a test for this as well, but can do so in a follow-up PR

PR #1096, opened 2017-08-13
Fix the Decimal serializing for postgres

Eijebong:
Values with four zeroes after the `.` were failing to serialize because
weight is negative and thus unnecessary_zeroes was negative ending in a
panic when we were indexing digits with it.

I'm not sure this is valid though, would appreciate if someone that actually knows what's going on with the serialization could confirm it to be ok :)


diesel/src/pg/types/numeric.rs
```diff
@@ -50,14 +50,20 @@ mod bigdecimal {
             digits.reverse();
             let digits_after_decimal = scale as u16 / 4 + 1;
             let weight = digits.len() as i16 - digits_after_decimal as i16 - 1;
-            let index_of_decimal = (weight + 1) as usize;
 
-            let unneccessary_zeroes = digits[index_of_decimal..]
+            let unnecessary_zeroes = if weight >= 0 {
+                let index_of_decimal = (weight + 1) as usize;
+                digits[index_of_decimal..]
```
killercup: How about we use `.get` here and ~~fail gracefully or~~ panic with a custom message like we do above?

PR #1102, opened 2017-08-14
Fix a warning when compiling diesel_cli

Eijebong:
```
warning: path `/home/eijebong/code/diesel/diesel_cli/src/main.rs` was erroneously implicitly accepted for binary `diesel`,
please set bin.path in Cargo.toml
```


killercup:
Explicit is better than ~~implicit~~ warnings 😆

PR #1104, opened 2017-08-15
link to jsonb type page for convenience

lancecarlson:


diesel/src/pg/types/mod.rs
```diff
@@ -124,7 +124,7 @@ pub mod sql_types {
     /// The JSON SQL type.  This type can only be used with `feature =
     /// "serde_json"`
     ///
-    /// Normally you should prefer `Jsonb` instead, for the reasons
+    /// Normally you should prefer [`Jsonb`](/diesel/pg/types/sql_types/struct.Jsonb.html) instead, for the reasons
```
killercup: Can you make this a relative link so it works on <https://docs.rs/diesel/0.15.2/diesel/pg/types/sql_types/struct.Json.html>

(I know this linking in Rustdoc is hack-y, I wrote an RFC for it 😄)

lancecarlson: Absolute URL?

killercup:
Thanks!

PR #1120, opened 2017-08-19
Make doc comment output optional for CLI

killercup:


killercup:
Follow-up to #957

sgrif:
Could you rebase this onto #1122 instead of master? <3

PR #1121, opened 2017-08-20
Include inferred `joinable!` in `diesel print-schema`

sgrif:
This was a pretty straightforward addition. I've not included an
explicit test that the filtering is applied, or that the unsafe
constraints are removed here. It may be worth more exhaustively testing
this.


diesel_cli/src/print_schema.rs
```diff
@@ -67,6 +75,14 @@ impl Display for TableDefinitions {
             }
             writeln!(f, "{}", TableDefinition(table))?;
         }
+
+        if !self.1.is_empty() {
+            write!(f, "\n")?;
+        }
+
+        for foreign_key in &self.1 {
+            write!(f, "{}", Joinable(foreign_key))?;
```
killercup: I'd use `writeln!` here and `write!` in the `<Joinable<'a> as Display>::fmt` method

diesel_cli/tests/print_schema.rs
```diff
@@ -390,3 +390,80 @@ r"pub mod custom_schema {
...
 ");
 }
+
+#[test]
+#[cfg(feature = "postgres")]
+fn print_schema_with_foreign_keys() {
+    let p = project("print_schema_with_foreign_keys").build();
+    let db = database(&p.database_url());
+
+    // Make sure the project is setup
```
killercup: Can't help but notice the redundant comment that is also spelt wrong (should be "is set up") which was successfully cargo culted into every single test case 😆

diesel_cli/src/print_schema.rs
```diff
@@ -118,6 +134,20 @@ impl<'a> Display for ColumnDefinitions<'a> {
     }
 }
 
+struct Joinable<'a>(&'a ForeignKeyConstraint);
+
+impl<'a> Display for Joinable<'a> {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        writeln!(
+            f,
+            "joinable!({} -> {} ({}));",
```
killercup: I see you're using rustfmt style now. I'd put the first two args on the first line, but whatever

sgrif: I'm going to run it on the whole codebase one of these days. Trying to reduce the churn in the short term

diesel_cli/tests/print_schema.rs
sgrif: Lol I'll get rid of it

PR #1122, opened 2017-08-20
Refactor print-schema tests to be easier to read and write

sgrif:
These tests were not ever testing the output on MySQL, were a pain to
write, and also to read. This moves the setup and expected output to
backend specific separate files, where they can be easily read and
written

I've also changed the tests to display a diff between the expected and
actual output on failures, as the failure messages were impossible to
read.


diesel_cli/Cargo.toml
```diff
@@ -23,6 +23,7 @@ diesel_infer_schema = "0.15.0"
 clippy = { optional = true, version = "=0.0.138" }
 
 [dev-dependencies]
+difference = "1.0"
```
killercup: Ha, I've added the exact same thing locally to test something but didn't refactor anything so I didn't push it!

Issue #1130, opened 2017-08-25
Crash with special timestamp value in mysql (`0000-00-00 00:00:00`) through chrono


konstin:
### Versions

- **Rust:** rustc 1.21.0-nightly (8c303ed87 2017-08-20)
- **Diesel:** 0.15.2
- **Database:** mysql 5.7.1
- **Operating System:** Ubuntu

### Feature Flags

- **diesel:** `["mysql", "chrono", "large-tables"]`
- **diesel_codegen:** `["mysql"]`

## Problem Description

When trying to load a value such as `0000-00-00 00:00:00` into a NaiveDateTime, there is panic inside chrono, which is called by diesel. 

### Steps to reproduce

 * Create a table in mysql with a timestamp field and add an entry with `0000-00-00 00:00:00`
 * Try to load that row into a struct with a `NaiveDateTime` field.
 * The process will panic: `thread '<unnamed>' panicked at 'invalid or out-of-range date', /checkout/src/libcore/option.rs:819:4`

### Full backtrace 

This crash happened trying to read a table with multiple columns, therefore the bigger generic. I'm still sure that the problem is the `0000-00-00 00:00:00` as the crash doesn't occur after removing them.

```
thread '<unnamed>' panicked at 'invalid or out-of-range date', /checkout/src/libcore/option.rs:819:4
stack backtrace:
   0:     0x561ed0753873 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::h80d78ba3b40687b5
                               at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1:     0x561ed074fa14 - std::sys_common::backtrace::_print::h47b9b32fe06dd6eb
                               at /checkout/src/libstd/sys_common/backtrace.rs:71
   2:     0x561ed0756263 - std::panicking::default_hook::{{closure}}::h006dcf643a2d1ee4
                               at /checkout/src/libstd/sys_common/backtrace.rs:60
                               at /checkout/src/libstd/panicking.rs:381
   3:     0x561ed0755fc2 - std::panicking::default_hook::h1e56c296d63316e2
                               at /checkout/src/libstd/panicking.rs:397
   4:     0x561ed0756767 - std::panicking::rust_panic_with_hook::h218401524ff20a29
                               at /checkout/src/libstd/panicking.rs:611
   5:     0x561ed07565c4 - std::panicking::begin_panic::h1668556d5aa9a913
[...]

PR #1137, opened 2017-09-01
Fix MySQL NaiveDate deserialization

alexeyzab:
Old behavior could cause panic in `chrono`.

See #1130.

If this approach is correct I can replace `from_ymd` with `from_ymd_opt` in other places too. Same with `and_hms_micro`.

### DIFF
--- a/diesel/src/mysql/types/date_and_time.rs
+++ b/diesel/src/mysql/types/date_and_time.rs
@@ -78,16 +78,17 @@ impl ToSql<Timestamp, Mysql> for NaiveDateTime {
 impl FromSql<Timestamp, Mysql> for NaiveDateTime {
     fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
         let mysql_time = <ffi::MYSQL_TIME as FromSql<Timestamp, Mysql>>::from_sql(bytes)?;
-        Ok(NaiveDate::from_ymd(
+
+        NaiveDate::from_ymd_opt(
             mysql_time.year as i32,
             mysql_time.month as u32,
             mysql_time.day as u32,
-        ).and_hms_micro(
+        ).and_then(|v| v.and_hms_micro_opt(
             mysql_time.hour as u32,
             mysql_time.minute as u32,
             mysql_time.second as u32,
             mysql_time.second_part as u32,
-        ))
+        )).ok_or_else(|| format!("Cannot parse this date: {:?}", mysql_time).into())
     }
 }
 
@@ -249,5 +250,8 @@ mod tests {
         let january_first_2018 = NaiveDate::from_ymd(2018, 1, 1);
         let query = select(sql::<Date>("CAST('2018-1-1' AS DATE)"));
         assert_eq!(Ok(january_first_2018), query.get_result::<NaiveDate>(&connection));
+
+        let query = select(sql::<Date>("CAST('0000-0-0' AS DATE)"));
+        assert!(query.get_result::<NaiveDate>(&connection).is_err());
     }
 }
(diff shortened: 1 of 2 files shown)


diesel/src/mysql/types/date_and_time.rs
```diff
@@ -78,16 +78,19 @@ impl ToSql<Timestamp, Mysql> for NaiveDateTime {
 impl FromSql<Timestamp, Mysql> for NaiveDateTime {
     fn from_sql(bytes: Option<&[u8]>) -> Result<Self, Box<Error+Send+Sync>> {
         let mysql_time = <ffi::MYSQL_TIME as FromSql<Timestamp, Mysql>>::from_sql(bytes)?;
-        Ok(NaiveDate::from_ymd(
+
+        // Clippy complains if you use `.into()` after `.ok_or()`
```
weiznich: Clippy complains here because this calls a function even in the success case. Creating a `String` from a `&'static str` will allocate memory.

A better solution would be to replace the `ok_or` call with [`ok_or_else`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else) and construct the error message string there.

alexeyzab: Thanks for pointing this out!

diesel/src/mysql/types/date_and_time.rs
```diff
@@ -78,16 +78,17 @@ impl ToSql<Timestamp, Mysql> for NaiveDateTime {
...
             mysql_time.month as u32,
             mysql_time.day as u32,
-        ).and_hms_micro(
+        ).and_then(|v| v.and_hms_micro_opt(
             mysql_time.hour as u32,
             mysql_time.minute as u32,
             mysql_time.second as u32,
             mysql_time.second_part as u32,
-        ))
+        )).ok_or_else(|| "Invalid date.".into())
```
killercup: As this is a cold path (I hope!), can you make the error message a bit more elaborate to make this easier to debug? Something like `.ok_or_else(|| format!("Cannot parse this date: {:?}", mysql_time).into())`

alexeyzab: Sure thing! I copied the error message from over [here](https://github.com/diesel-rs/diesel/blob/3ae353c3aff8a7ea64ed6cb39b3a045ac86cd60e/diesel/src/sqlite/types/date_and_time/chrono.rs#L25-L28). I am guessing it would be a good idea to add something like this everywhere `from_ymd` is used.

alexeyzab:
> That would be awesome. Do you want to do that here, or in another PR?

I can do that here, just wanted to make sure I was on the right track. :)

killercup:
> I can do that here, just wanted to make sure I was on the right track. :)

Awesome, you most definitely are!

alexeyzab:
After looking at the code again, it seems that all the other calls to `from_ymd` involve specific dates that won't fail, like `pg_epoch` or all the unit tests. I've fixed the sqlite error message I linked to earlier, but not sure what else I could do here.

sgrif:
Tests pass locally for me as well. I'm going to try merging this and see what happens on master.

sgrif:
The reason this fails on CI but not locally is that this PR only changed `NaiveDateTime`, not `NaiveDate`. On my machine (and presumably yours), `CAST('0000-0-0' AS DATE)` returns `NULL`, so it's erroring before it gets to the deserialization code. On Travis it's using a different MySQL version which is parsing that string.

alexeyzab:
@sgrif Ohh, I see, thank you for pointing it out! Is there anything I can help with here?

sgrif:
No, I fixed it in 347cd24

alexeyzab:
Yep, I saw it, wanted to make sure.

PR #1139, opened 2017-09-01
Documentation Syntax results in paniced macro.

alatiera:
When followed the (outdated?) syntax provided by the documentation the macro would panic.
Changed it instead to what the compiler example suggests.
```
error: proc-macro derive panicked
  --> src/models.rs:36:21
   |
36 | #[derive(Queryable, Identifiable)]
   |                     ^^^^^^^^^^^^
   |
   = help: message: `table_name` must be in the form `#[table_name="something"]`
```

### DIFF
--- a/diesel/src/macros/associations/belongs_to.rs
+++ b/diesel/src/macros/associations/belongs_to.rs
@@ -1,7 +1,7 @@
 /// Defines a one-to-one association for the child table. This macro should be
 /// called with the name of the parent struct, followed by any options, followed
 /// by the entire struct body. The struct *must* be annotated with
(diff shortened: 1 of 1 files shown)


konstin:
Thanks!

Eijebong:
As I said on gitter, I don't think this is the right fix.

`BelongsTo!` does take `#[table_name(...)]`, what you're usually looking for when you're doing associations is `#[belongs_to(...)]`.

sgrif:
The `table_name(foo)` was really only a thing with `impl_Identifiable!` which is essentially deprecated. AFAIK there is nothing that takes `table_name(foo)` when using `diesel_codegen`, so I think this is the right fix.

Issue #1143, opened 2017-09-04
Consider linting test code on CI


killercup:
We are currently telling contributors to use the [`bin/test`] script to test their changes. This script runs a ALL the `cargo test`s—with clippy enabled. Which means it catches more than our CI setup does, as the Travis only runs [the equivalent of cargo clippy](https://github.com/diesel-rs/diesel/blob/ca7b989d4457aee33eef6c5e373238b77bf74b9a/.travis.yml#L43-L50).

I think we can tell Travis to compile in a way so code in `#[cfg(test)]` gets included, but this may conflict with compiling all backends at once.

[`bin/test`]: https://github.com/diesel-rs/diesel/blob/ca7b989d4457aee33eef6c5e373238b77bf74b9a/bin/test

sgrif:
I'm actually wondering if we should just remove clippy from `bin/test` as it increases build times quite a bit. I don't think linting the tests extensively adds nearly as much value

killercup:
True. On the other hand, it's frustrating to see tests pass locally only to have CI fail on clippy. Do we mention `bin/check` enough in the Contributing Guide?

sgrif:
If we removed it from `bin/test` I would say that we should not lint tests on CI either. Only lint the main code, which is what CI does currently.

killercup:
Yep. But if people *only* use bin/test, they wont ever see clippy's wonderful suggestions for their changes

sgrif:
Given that it's test code, I don't think that's a huge loss

killercup:
I'm also talking about regular code. With this change, the only way to get clippy to check anything is `bin/check`

sgrif:
I guess it's a question of whether people are running `bin/test` or `bin/check` more.

killercup:
Well, let's just hope people find `bin/check` then 😄

PR #1144, opened 2017-09-04
Add underscores to long literals in test code

willmurphyscode:
Updating to rust nightly 2017-08-31 made clippy start
failing builds on long literals. (e.g., 72400 should be 72_400).
This adds underscores to long literals in test code.

See https://github.com/diesel-rs/diesel/issues/1143 on why this wasn't caught in CI.

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -297,11 +297,11 @@ pub mod sql_types {
     /// #     )").unwrap();
     /// let new_item = NewItem {
     ///     name: "Shiny Thing".into(),
-    ///     price: Cents(123456),
+    ///     price: Cents(123_456),
     /// };
     /// let inserted_item = insert(&new_item).into(items)
     ///     .get_result::<Item>(&connection).unwrap();
-    /// assert_eq!(Cents(123456), inserted_item.price);
+    /// assert_eq!(Cents(123_456), inserted_item.price);
     /// # }
     /// ```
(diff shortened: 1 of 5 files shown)


diesel_tests/tests/types.rs
```diff
@@ -383,10 +383,10 @@ fn timestamp_from_sql() {
     use diesel::data_types::PgTimestamp;
 
     let query = "'2015-11-13 13:26:48.041057-07'::timestamp";
-    let expected_value = PgTimestamp(500736408041057);
+    let expected_value = PgTimestamp(500_736_408_041_057);
```
willmurphyscode: This line, and others in the same file, were not caught by `bin/test`. I changed them anyway, because it seems like they should have been caught by `bin/test`, and I assume future development might make `bin/test` catch them.

diesel/src/pg/types/mod.rs
```diff
@@ -297,11 +297,11 @@ pub mod sql_types {
     /// #     )").unwrap();
     /// let new_item = NewItem {
     ///     name: "Shiny Thing".into(),
-    ///     price: Cents(123456),
+    ///     price: Cents(123_456),
```
willmurphyscode: Also, this line and the other line in this file were not caught by `bin/test`, but I changed them because I assume they should have been linted the same way as the rest of the code, and in the future they might be.

killercup:
Awesome, thanks @willmurphyscode!

Issue #1135, opened 2017-08-31
Where do we want to use associated consts?


killercup:
Rust 1.20 stabilized [associated consts](https://github.com/rust-lang/rfcs/blob/master/text/0195-associated-items.md)!

…and, in contrast to every other ~~hipster~~ unstable Rust feature under the sun, we apparently have neither code comments nor commit messages in diesel's code base telling us what places could benefit from this new feature. But not to worry, we can easily[^1] find some!

Wait—should we? It makes the API a bit more idiomatic at the cost of requiring at least 1.12 to build diesel. Additionally, we _could_ declare the current methods as `#[deprecated]` and not break any builds. The constants should be in `SCREAMING_SNAKE_CASE` anyway, so the names won't conflict. I'd say: Go for it, and actually replace these static methods. Right now, before 1.0, we can afford these breaking changes, after that we need to live with the status quo.

Here are the places[^1] that I found that could benefit from associated consts:

- `diesel::query_builder::QueryId`: `const HAS_STATIC_QUERY_ID: bool` instead of `fn has_static_query_id() ->bool`
- `diesel::query_source::Column`: `const NAME: &'static str` instead of `fn name() -> &'static str`
- `diesel::types::FromSqlRow`: `const FIELDS_NEEDED: usize` (with default) instead of `fn fields_needed() -> usize { ... }`

I'm not entirely sure if we want/can/should turn methods like `foo() -> Self::Foo` into associated consts. This would add a few more cases (`<Foo as diesel::associations::HasTable>::table()`, `<Foo as diesel::query_source::Table>::all_columns()`, and `<Foo as diesel::associations::BelongsTo>::foreign_key_column()`).

- - -

[^1]: `rg -i '    (pub | ) fn (.*?)\(\) ->'`

sgrif:
:+1: for using associated constants for those 3 cases

willmurphyscode:
Is anyone presently working on this? If not, I'd like to volunteer to give this a try.

Eijebong:
I don't think anyone is working on it so feel free to work on it.
If you have any question don't hesitate to ask here or on our gitter room (link it the README)

PR #1145, opened 2017-09-04
Issue 1135: Associated constants

willmurphyscode:
This PR will make will use [associated constances](https://doc.rust-lang.org/1.6.0/book/associated-constants.html) instead of functions in 3 places:

- `diesel::query_builder::QueryId: const HAS_STATIC_QUERY_ID:` bool instead of `fn has_static_query_id() ->bool`
- `diesel::query_source::Column: const NAME: &'static str` instead of `fn name() -> &'static str`
- `diesel::types::FromSqlRow: const FIELDS_NEEDED: usize` (with default) instead of `fn fields_needed() -> usize { ... }`

These are the three places specifically mentioned in https://github.com/diesel-rs/diesel/issues/1135.

I have one stylistic question: Currently the `const WHATEVER` line replaces the first line of `fn whatever()`. If the line with `const WHATEVER` followed a line like `type SomeType`, I took out the blank line above. Otherwise, I left the blank lines as is. Is that how we want the style to be?


diesel/src/expression/unchecked_bind.rs
```diff
@@ -32,10 +32,7 @@ impl<Query, Value, ST> QueryId for UncheckedBind<Query, Value, ST> where
     ST: QueryId,
 {
     type QueryId = UncheckedBind<Query::QueryId, (), ST::QueryId>;
-
-    fn has_static_query_id() -> bool {
-        Query::has_static_query_id() && ST::has_static_query_id()
-    }
+    const HAS_STATIC_QUERY_ID: bool = Query::HAS_STATIC_QUERY_ID && ST::HAS_STATIC_QUERY_ID;
```
killercup: Wow, that works? Nice!

diesel/src/macros/query_id.rs
```diff
@@ -17,31 +14,22 @@ macro_rules! impl_query_id {
             $($ty_param: $crate::query_builder::QueryId),*
         {
             type QueryId = $name<$($ty_param::QueryId),*>;
-
-            fn has_static_query_id() -> bool {
-                $($ty_param::has_static_query_id() &&)* true
-            }
+            const HAS_STATIC_QUERY_ID: bool = $($ty_param::HAS_STATIC_QUERY_ID &&)* true;
```
killercup: Now that is just awesome

diesel/src/query_builder/query_id.rs
```diff
@@ -3,10 +3,10 @@ use super::QueryFragment;
 
 pub trait QueryId {
     type QueryId: Any;
-    fn has_static_query_id() -> bool;
+    const HAS_STATIC_QUERY_ID: bool = true;
```
Eijebong: This one doesn't respect what you said about letting a blank line between each type of declaration

willmurphyscode: Thanks! I must've missed that one. I'll push up a change in a minute.

willmurphyscode:
It looks like the `diesel_tests` project failed to build on Windows for Postgres. I'm a bit surprised, since I would expect this change either to succeed or to fail everything everywhere. 

I'm also surprised that this build: https://ci.appveyor.com/project/sgrif/diesel/build/1.0.1006 passed while this build: https://ci.appveyor.com/project/sgrif/diesel/build/1.0.1007 failed when the commit between them was this: https://github.com/diesel-rs/diesel/pull/1145/commits/c0bbc9f55263c95d514236c796e9c1965ab5f2f8 which has only a whitespace change. 

I'll attempt to repro this myself, but honestly my Windows computer is not set up for development at all, so I could use a little bit of assistance. (Also, it might be worth investigating a pipeline issue, since apparently a whitespace-only commit broke a build.)

willmurphyscode:
It looks like https://github.com/diesel-rs/diesel/pull/1146 is failing CI builds with the same error, but also failing them for Postgres on Linux, not just Windows.

PR #1146, opened 2017-09-04
Provide a predictable sort for `joinable!` calls in print-schema.

dikaiosune:
**TLDR: I can haz stable `joinable!` output from `diesel print-schema`?**

## Why?

In switching https://github.com/dikaiosune/rust-dashboard over to use `diesel print-schema` instead of calling `infer_schema!` directly in the code, I've hit a small snag.

My goal is to run `diff src/schema_which_is_committed_to_git.rs schema_which_i_generate_in_ci_from_the_migrations_i_ran.rs` and fail the CI job if the committed migrations result in a schema different from what's committed to source control. Unfortunately, today it seems that identical versions of `diesel_cli` can produce [identical `joinable!` calls but in different orders](https://travis-ci.org/dikaiosune/rust-dashboard/jobs/270927391#L622-L650). Ideally, the output of the foreign key constraints would be predictably sorted. So, uh, that's what I did.

## Risks

This seems like a pretty small change code-wise, although it would involve a publicly visible trait impl which would break backcompat to later remove.

### DIFF
--- a/diesel_infer_schema/src/inference.rs
+++ b/diesel_infer_schema/src/inference.rs
@@ -208,7 +208,7 @@ pub fn load_foreign_key_constraints(
 ) -> Result<Vec<ForeignKeyConstraint>, Box<Error>> {
     let connection = try!(establish_connection(database_url));
 
-    match connection {
+    let constraints = match connection {
         #[cfg(feature = "sqlite")]
         InferConnection::Sqlite(c) => ::sqlite::load_foreign_key_constraints(&c, schema_name),
         #[cfg(feature = "postgres")]
@@ -219,7 +219,9 @@ pub fn load_foreign_key_constraints(
         InferConnection::Mysql(c) => {
             ::mysql::load_foreign_key_constraints(&c, schema_name).map_err(Into::into)
         }
-    }
+    };
+
+    constraints.map(|mut ct| { ct.sort(); ct })
 }
 
 macro_rules! doc_comment {
--- a/diesel_cli/tests/print_schema/print_schema_with_foreign_keys/postgres/expected.rs
+++ b/diesel_cli/tests/print_schema/print_schema_with_foreign_keys/postgres/expected.rs
@@ -52,5 +52,5 @@ table! {
     }
 }
 
-joinable!(posts -> users (user_id));
 joinable!(comments -> posts (post_id));
+joinable!(posts -> users (user_id));
--- a/diesel_infer_schema/src/data_structures.rs
+++ b/diesel_infer_schema/src/data_structures.rs
@@ -92,7 +92,7 @@ where
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ForeignKeyConstraint {
     pub child_table: TableName,
     pub parent_table: TableName,


diesel_infer_schema/src/inference.rs
```diff
@@ -143,14 +143,20 @@ pub fn load_foreign_key_constraints(database_url: &str, schema_name: Option<&str
...
+    let mut constraints = match connection {
         #[cfg(feature = "sqlite")]
         InferConnection::Sqlite(c) => ::sqlite::load_foreign_key_constraints(&c, schema_name),
         #[cfg(feature = "postgres")]
         InferConnection::Pg(c) => ::information_schema::load_foreign_key_constraints(&c, schema_name).map_err(Into::into),
         #[cfg(feature = "mysql")]
         InferConnection::Mysql(c) => ::mysql::load_foreign_key_constraints(&c, schema_name).map_err(Into::into),
+    };
+
+    if let Ok(ref mut c) = constraints.as_mut() {
```
killercup: That's pretty… noisy code with many ref/mut/as_mut thingies. Do you think `let constraints = constraints.map(|mut cs| { cs.sort(); cs });` is better?

dikaiosune: Good point! Just pushed a cleanup for that. No need for an assignment, that can be the terminating expression for the function.

(related: i simply cannot wait for pattern ergonomics improvements!)

dikaiosune:
Re: `ORDER BY` vs. sorting in Rust, I agree it's nice to not be backend dependent. Also fewer places to make errors as a new contributor :).

My assumption re: runtime perf is that this code is called on very small (surely <1000) vectors, rather infrequently. I guess the inference results aren't cached, so if one were using `infer_schema!` in their code--as opposed to `print-schema`--this might have an observable performance impact for very large databases. Seems unlikely?

dikaiosune:
Related question: I modified the existing test for printing foreign keys to pass with this ordering, but I also see that there is a specific ordering test for table descriptions. Given that this is a small change, and not that complex, I'm assuming you're OK with not adding a new test?

dikaiosune:
Looks like the Windows CI failure is due to rustc type inference:

```
The column `tags` is of type `varchar[]`. This will cause problems when using Diesel. You should consider changing the column type to `text[]`.
error[E0282]: type annotations needed
  --> tests\select.rs:99:22
   |
99 |     let actual_data: Vec<_> = source.load(&connection).unwrap();
   |         -----------  ^^^^^^ cannot infer type for `_`
   |         |
   |         consider giving `actual_data` a type

```

Doesn't seem related to any of my changes here, but I could be wrong. A minimal version of the code which is failing seems to work for me: http://play.integer32.com/?gist=363dae41488a865fbcc18bbd3f971c8d&version=stable

willmurphyscode:
@dikaiosune It looks like https://github.com/diesel-rs/diesel/pull/1145 is failing with the same error, which makes it seem like indeed some change outside these two PRs is causing the failure. 

I can repro this Linux failure locally, but I haven't been able to pin down the cause yet.

killercup:
@dikaiosune failing test issue fixed in and more details in 1f3d7b1d521d616ebd969a5c19c375807579d864.

Can you rebase on master?

dikaiosune:
@killercup rebased, looks like tests are passing.

PR #1147, opened 2017-09-05
Add missing clone implementation for Many

weiznich:


sgrif:
I was hoping to leave this for a new contributor, but thanks for fixing this. :)

PR #1148, opened 2017-09-05
Fixes #1138 - Utilize compile_error! macro.

### LINKED ISSUE #1138: Use compile_error!
Rust 1.20 stabilized the [`compile_error!`](https://doc.rust-lang.org/stable/std/macro.compile_error.html) macro, which means we can finally output useful errors from macros and unusable feature-flag combinations!

Here are the places with code comments just waiting to use this macro:

https://github.com/diesel-rs/diesel/blob/b26472f4290c658822334b148494550b943ef991/diesel/src/macros/mod.rs#L406-L408

https://github.com/diesel-rs/diesel/blob/453d8ffbfe80bada40b79ddc2834d2b3c190b7be/diesel/src/test_helpers.rs#L50

https://github.com/diesel-rs/diesel/blob/453d8ffbfe80bada40b79ddc2834d2b3c190b7be/diesel_codegen/tests/test_helpers.rs#L50

https://github.com/diesel-rs/diesel/blob/4f87b471128061887040c703c6aa1f1ae62218d6/diesel/src/doctest_setup.rs#L144

https://github.com/diesel-rs/diesel/blob/148d0e448c95120baaa721578fe03e519b0bbf40/diesel/src/macros/insertable.rs#L501

notryanb:
compile_error! was introduced in Rust 1.20. Use to give concise compile-time error messages.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -401,9 +401,8 @@ macro_rules! __diesel_table_impl {
         table_doc = [$($doc:expr,)*];
         $($rest:tt)*
     ) => {
-        // Remove this workaround as soon as rust issue 40872 is implemented
-        // https://github.com/rust-lang/rust/issues/40872
-        const ERROR: i32 = env!("invalid table! syntax");
+        compile_error!("Invalid `table!` syntax. Please see the `table!` macro docs for more info. \
+        `http://docs.diesel.rs/diesel/macro.table.html`");
     };
 
     // Put a parse annotation and empty fields for imports and documentation on top
--- a/diesel_infer_schema/src/inference.rs
+++ b/diesel_infer_schema/src/inference.rs
@@ -114,6 +114,13 @@ pub(crate) fn establish_connection(database_url: &str) -> Result<InferConnection
                 database_url,
             ).into(),
         ),
+        #[cfg(not(any(feature = "mysql", feature = "sqlite", feature = "postgres")))]
+        _ => compile_error!(
+                "At least one backend must be specified for use with this crate.\n \
+                In Cargo.toml, please specify `features = [\"postgresql\"]`, \
+                `features = [\"mysql\"]`, or `features = [\"sqlite\"]`\n\n \
+                ex. `diesel_codegen { features = [\"postgres\"] }`\n "
+        )
     }
 }
 
--- a/diesel/src/doctest_setup.rs
+++ b/diesel/src/doctest_setup.rs
@@ -140,8 +140,12 @@ cfg_if! {
             connection
         }
     } else {
-        // FIXME: https://github.com/rust-lang/rfcs/pull/1695
-        // compile_error!("At least one backend must be enabled to run tests");
+        compile_error!(
+            "At least one backend must be used to test this crate.\n \
+            Pass argument `--features \"<backend>\"` with one or more of the following backends, \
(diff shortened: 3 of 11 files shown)


diesel_infer_schema/src/inference.rs
```diff
@@ -56,15 +56,17 @@ pub(crate) fn establish_connection(database_url: &str) -> Result<InferConnection
                 It must start with postgres:// or postgresql://",
                 database_url,
             ).into())
-        }
+        },
```
sgrif: These commas shouldn't be required

diesel_infer_schema/src/inference.rs
```diff
@@ -56,15 +56,17 @@ pub(crate) fn establish_connection(database_url: &str) -> Result<InferConnection
...
         _ => {
             Err(format!(
                 "{} is not a valid MySQL database URL. \
                 It must start with mysql://",
                 database_url,
             ).into())
-        }
+        },
+        #[cfg(not(any(feature = "mysql", feature = "sqlite", feature = "postgres")))]
+        _ => compile_error!("No database backend specified."),
```
sgrif: Since this message is user-facing, can we provide some information about how to fix this? It should mention that `features = ["postgres"]`, etc are required.

diesel/src/macros/mod.rs
```diff
@@ -401,9 +401,7 @@ macro_rules! __diesel_table_impl {
         table_doc = [$($doc:expr,)*];
         $($rest:tt)*
     ) => {
-        // Remove this workaround as soon as rust issue 40872 is implemented
-        // https://github.com/rust-lang/rust/issues/40872
-        const ERROR: i32 = env!("invalid table! syntax");
+        compile_error!("invalid table! syntax");
```
killercup: Can we link to some docs here? (If we have them)

diesel_infer_schema/src/inference.rs
```diff
@@ -114,6 +114,10 @@ pub(crate) fn establish_connection(database_url: &str) -> Result<InferConnection
                 database_url,
             ).into(),
         ),
+        #[cfg(not(any(feature = "mysql", feature = "sqlite", feature = "postgres")))]
+        _ => compile_error!("At least one backend must be used with this crate.\n \
+                            Please specify `features = [\"postgresql\"]`, \
```
killercup: Specify where? Add "`Cargo.toml` dependency settings" somewhere?

diesel_infer_schema/src/inference.rs
```diff
@@ -114,6 +114,10 @@ pub(crate) fn establish_connection(database_url: &str) -> Result<InferConnection
...
+                            `features = [\"mysql\"]`, or `features = [\"sqlite\"]`")
```
killercup: Why not copypasta this good message to all other places this PR changes? :)

diesel/src/macros/mod.rs
notryanb: Should that link appear in the error message or a comment just above the code?

killercup: in the error message, so users can see how to use `table!` correctly

notryanb:
Updated a few things @sgrif @killercup @Eijebong 
- invalid `table!` macro error is now more actionable & tests were updated
- `diesel_infer_schema/src/inference.rs` error now references the correct crate
- **diesel_cli** now has a `compile_error!` for being installed without any backend specified.

PR #1151, opened 2017-09-05
Document all items in `expression::helper_types`

sgrif:
I've also de-macro'd the code, as the macros weren't adding much value.

### DIFF
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -5,37 +5,48 @@ use super::{AsExpression, Expression};
 use super::grouped::Grouped;
 use types;
 
+/// The SQL type of an expression
 pub type SqlTypeOf<Expr> = <Expr as Expression>::SqlType;
+
+/// The type of `Item` when converted to an expression with the same type as `TargetExpr`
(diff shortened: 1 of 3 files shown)


diesel/src/expression/helper_types.rs
```diff
@@ -5,37 +5,48 @@ use super::{AsExpression, Expression};
...
+pub type GtEq<Lhs, Rhs> = super::operators::GtEq<Lhs, AsExpr<Rhs, Lhs>>;
+
+/// The return type of `lhs.lt(rhs)`
+pub type Lt<Lhs, Rhs> = super::operators::Lt<Lhs, AsExpr<Rhs, Lhs>>;
+
+/// The return type of `lhs.le(rhs)`
+pub type LtEq<Lhs, Rhs> = super::operators::LtEq<Lhs, AsExpr<Rhs, Lhs>>;
+
+/// The return type of `lhs.and(rhs)`
+pub type And<Lhs, Rhs> = super::operators::And<Lhs, AsExprOf<Rhs, Bool>>;
```
killercup: Oh, and this needs to be `types::Bool` (same for the following VarChars)

PR #1152, opened 2017-09-05
Document the items on `AsQuery` [ci skip]

sgrif:


diesel/src/query_builder/mod.rs
```diff
@@ -131,9 +131,14 @@ impl<DB: Backend> QueryFragment<DB> for () {
 /// internally to automatically add the right select clause when none is
 /// specified, or to automatically add `RETURNING *` in certain contexts
 pub trait AsQuery {
+    /// The SQL type of `Self::Query`
     type SqlType;
+
+    /// What kind of query does this type represent?
```
killercup: This is a very abstract thing… maybe add an example for a concrete instance of this to illustrate what it can be?

PR #1159, opened 2017-09-11
Run rustfmt on CI

sgrif:


.travis.yml
```diff
@@ -48,6 +48,12 @@ matrix:
     - (cd diesel_codegen && cargo rustc --no-default-features --features "lint dotenv sqlite postgres mysql" -- -Zno-trans)
     - (cd diesel_infer_schema && cargo rustc --no-default-features --features "lint dotenv sqlite postgres mysql" -- -Zno-trans)
     - (cd diesel_compile_tests && travis-cargo test)
+  include:
+  - rust: nightly
+    env: RUSTFMT=YESPLEASE
```
Eijebong: Why do we need this ?

sgrif: To make it clear what this run is.

Eijebong: Oh in the travis UI, makes sense 👍

sgrif: Well we also need some env var that is different to make it a different matrix item than the normal nightly test runs.

PR #1160, opened 2017-09-14
feat: Support batch insert for sqlite

xinghun92:
Support batch insert in sqlite with default values when using infer_schema macro.
If using table macro directly without define default value, the default value is NULL

Change-Id: I5c7bed2c8018efbb452350752b9c43755158da03

### DIFF
--- a/diesel/src/backend.rs
+++ b/diesel/src/backend.rs
@@ -31,5 +31,4 @@ pub trait TypeMetadata {
 }
 
 pub trait SupportsReturningClause {}
-pub trait SupportsDefaultKeyword {}
 pub trait UsesAnsiSavepointSyntax {}
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -1,6 +1,6 @@
 use std::iter;
 
-use backend::{Backend, SupportsDefaultKeyword};
+use backend::Backend;
 use result::QueryResult;
 use query_builder::AstPass;
 use query_source::Table;
@@ -39,7 +39,7 @@ type ValuesFn<Item, T, DB> = fn(Item)
 impl<Iter, T, DB> Insertable<T, DB> for Iter
 where
     T: Table,
-    DB: Backend + SupportsDefaultKeyword,
+    DB: Backend,
     Iter: IntoIterator,
     Iter::Item: Insertable<T, DB>,
     Iter::IntoIter: Clone,
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -243,6 +243,7 @@ mod tests {
         users {
             id -> Integer,
             name -> VarChar,
+            #[default_value="'Green'"]
             hair_color -> Nullable<VarChar>,
         }
     }
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -1,7 +1,7 @@
 #[macro_export]
 #[doc(hidden)]
 macro_rules! __diesel_column {
-    ($($table:ident)::*, $column_name:ident -> ($($Type:tt)*),  $sql_name:expr, $($doc:expr),*) => {
+    ($($table:ident)::*, $column_name:ident -> ($($Type:tt)*),  $sql_name:expr, $default_value:expr, $($doc:expr),*) => {
         $(
             #[doc=$doc]
         )*
@@ -69,6 +69,7 @@ macro_rules! __diesel_column {
             type Table = $($table)::*;
 
             const NAME: &'static str = $sql_name;
+            const DEFAULT: &'static str = $default_value;
         }
 
         impl<T> $crate::EqAll<T> for $column_name where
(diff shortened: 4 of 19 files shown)


xinghun92:
I have no idea how to use users table in batch_insert_with_defaults and insert_with_defaults, so I create a new table named users_with_default.

sgrif:
Thank you for the PR. However, this takes several steps that I'm not comfortable with right now. I would strongly prefer that Diesel does *not* know or care about the default values of columns for the time being (this is something that we could potentially revisit after 1.0, but it has major compatibility concerns that I don't want to commit to for 1.0 stability).

I'd also want to see a more compelling justification for doing batch insert on SQLite in a single query. We already support batch insert by doing one query per row. While this would be a performance concern on other backends, it is not a concern for SQLite, as there is no round trip time.

xinghun92:
Although there is no round trip time for SQLite, the performance is quite different for batch insert by one query or multiple queries.

this is my test code
```
lib.rs
#![recursion_limit="128"]
#![allow(unused_features)]
#![feature(test)]
#[macro_use]
extern crate diesel;
extern crate dotenv;
#[macro_use]
extern crate diesel_codegen;

use std::env;
use diesel::Connection;
use dotenv::dotenv;

#[derive(Queryable, Clone, Debug)]
pub struct User {
    pub id: i64,
    pub name: String,
    pub type_: i32,
    pub avatar_key: String,
    pub update_time: i64,
    pub name_pinyin: String,
    pub creator_id: Option<i64>,
    pub is_resigned: Option<bool>
}

#[derive(Insertable, Debug)]
#[table_name = "users"]
pub struct NewUser<'a> {
    pub id: i64,
    pub name: &'a str,
    pub type_: i32,
    pub avatar_key: &'a str,
    pub update_time: i64,
    pub name_pinyin: &'a str,
    pub creator_id: Option<i64>,
    pub is_resigned: Option<bool>
}

#[allow(dead_code)]
fn establish_connection() ->  SqliteConnection {
    dotenv().ok();

    let database_url = env::var("DATABASE_URL")
        .expect("DATABASE_URL must be set");
     SqliteConnection::establish(&database_url).unwrap()
}

#[allow(dead_code)]
fn create_new_users<'a>(num: i64) -> Vec<NewUser<'a>> {
    let mut users = Vec::new();
    for i in 0..num {
        let user = NewUser {
            id: i,
            name: "name",
            type_: 1,
            avatar_key: "avatar_key",
            update_time: 1,
            name_pinyin: "name",
            creator_id: Some(0),
            is_resigned: Some(true)
        };
        users.push(user);
    }
    users
}

#[cfg(test)]
mod tests {
    extern crate test;
    use ::diesel::ExecuteDsl;
    use self::test::Bencher;

    #[bench]
    fn bench_batch_insert_users(b: &mut Bencher) {
        let test_num = 1;
        let conn = ::establish_connection();
        let new_users = ::create_new_users(test_num);
        b.iter(|| {
            let num = ::diesel::insert_or_replace(&new_users)
                .into(::users::table)
                .execute(&conn).unwrap();
            assert_eq!(test_num, num as i64);
        });
    }
}

migrations/20170620103016_users/up.sql
CREATE TABLE users (
  id BIGINT NOT NULL PRIMARY KEY,
  name TEXT NOT NULL,
  type_ INTEGER NOT NULL,
  avatar_key TEXT NOT NUll DEFAULT '',
  update_time BIGINT NOT NULL DEFAULT 0,
  name_pinyin TEXT NOT NULL,
  creator_id BIGINT,
  is_resigned BOOLEAN
);

migrations/20170620103016_users/down.sql
drop table users;

.env
DATABASE_URL=test.db
```

and the time cost goes here:

```
batch insert by one query:
1 row ----> 0.45ms
10 row ----> 0.76ms
100 row ---> 2.55ms

batch insert by multiple queries:
1 row ----> 0.42ms
10 row ----> 10.20ms
100 row ---> 47.38ms
```
Batch insert by one query is faster than multiple queries. And I think support batch insert by single query for SQLite may be better.

sgrif:
You're comparing apples to oranges. One is atomic, the other is not. If you wrap your benchmark in a transaction, the performance is virtually identical.

xinghun92:
Why not implement the batch insert of sqlite in a transaction like this? Or add a new batch_insert with transaction. I think there are some needs for faster insert, and add a transaction in ever insert statement out of diesel seems bad.
```
#[cfg(feature="sqlite")]
impl<'a, T, U, Op, Ret> ExecuteDsl<::sqlite::SqliteConnection>
    for BatchInsertStatement<T, &'a [U], Op, Ret> where
        InsertStatement<T, &'a U, Op, Ret>: ExecuteDsl<::sqlite::SqliteConnection>,
        T: Copy,
        Op: Copy,
        Ret: Copy,
{
    fn execute(self, conn: &::sqlite::SqliteConnection) -> QueryResult<usize> {
        if self.records.is_empty() {
            Ok(0)
        } else {
            conn.transaction::<usize, ::result::Error, _> (|| {
                let mut result = 0;
                for record in self.records {
                    result += InsertStatement::new(self.target, record, self.operator, self.returning)
                        .execute(conn)?;
                }
                Ok(result)
            })
        }
    }
}

sgrif:
https://github.com/diesel-rs/diesel/issues/1177

PR #1162, opened 2017-09-14
Refactor `InsertValues` for tuples

sgrif:
Note: This depends on #1161. The second commit is the only one that is new here.

This changes the implementation of `InsertValues` on tuples to not care
about its interior type, and moves the SQLite special handling to
`ColumnInsertValue` directly.

The addition of the `Table` parameter on `InsertValues` is required for
the tuple implementation to enforce that all values are targeting the
same table.

One side effect of this structure is that it will (eventually) allow
arbitrary nesting of `Insertable` types. This is important for use cases
like rocket, where you will likely have a `user_id` field which comes
from the session, but we want you to derive `Insertable` and `FromForm`
on the same struct. This means we should be able to write code like the
following:

    #[derive(Insertable, FromForm)]
    #[table_name = "posts"]
    struct NewPost {
        title: String,
        body: String,
    }

    #[post("/posts", data = "<form>")]
    fn create_post(data: NewPost, user: AuthenticatedUser, conn: DbConn)
        -> QueryResult<Redirect>
    {
        use schema::posts::dsl::*;

        let created = insert(&(user_id.eq(user.id), data))
            .into(posts)
            .returning(id)
            .get_result::<i32>(&*conn)?;
        let url = format!("/posts/{}", created);
        Ok(Redirect::to(&url))
    }

This should also let us simplify the `default_values` code dramatically.

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -1,9 +1,10 @@
-use std::marker::PhantomData;
-
 use backend::{Backend, SupportsDefaultKeyword};
+use expression::Expression;
 use result::QueryResult;
-use query_builder::AstPass;
-use query_source::Table;
+use query_builder::{AstPass, QueryBuilder, QueryFragment};
+use query_source::{Column, Table};
+#[cfg(feature = "sqlite")]
+use sqlite::Sqlite;
 
 /// Represents that a structure can be used to insert a new row into the
 /// database. This is automatically implemented for `&[T]` and `&Vec<T>` for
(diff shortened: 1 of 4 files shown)


diesel/src/insertable.rs
```diff
@@ -33,18 +45,71 @@ pub enum ColumnInsertValue<Col, Expr> {
...
+        }
+        Ok(())
+    }
+
+    fn is_noop(&self) -> bool {
+        if let ColumnInsertValue::Expression(..) = *self {
+            false
+        } else {
+            true
+        }
```
killercup: I'd probably flip this to `if let ColumnInsertValue ::Default(..) = *self { true } else { false }` (and maybe even add a method to the enum if we ever need that for another thing).

sgrif: :+1: will follow up with that once I'm done moving this code around

PR #1163, opened 2017-09-14
Replace `insert_default_values` with `insert(&default_values())`

sgrif:
This API is more consistent with the rest of Diesel, and requires
significantly less code. I think I will likely end up removing the need
to pass a reference to `insert`. If so I'll get rid of the `&` impls on
this and just have it be `insert(default_values)`


diesel_tests/Cargo.toml
```diff
@@ -12,7 +12,7 @@ dotenv = ">=0.8, <0.11"
 [dependencies]
 assert_matches = "1.0.1"
 chrono = { version = "0.4" }
-diesel = { path = "../diesel", default-features = false, features = ["quickcheck", "chrono", "uuid", "serde_json", "network-address", "numeric"] }
+diesel = { path = "../diesel", default-features = false, features = ["quickcheck", "chrono", "uuid", "serde_json", "network-address", "numeric", "with-deprecated"] }
```
Eijebong: Is this the first deprecated stuff that is actually tested ? Weird

sgrif: Everything else we've deprecated is tested through doctests, not integration tests.

PR #1164, opened 2017-09-15
Allow ad-hoc inserts of single column rows

sgrif:
This change allows for simple inserts where only one column is being
assigned. This will be followed by a PR to allow multi-column inserts
using tuples, but the number of tests required for that is significantly
higher, so I wanted to break this out into a separate piece.


sgrif:
Tests were green, I had just forgotten to run `rustfmt` on it

PR #1165, opened 2017-09-15
Allow ad-hoc inserts using tuples

### LINKED ISSUE #789: Provide a lightweight API for one-off inserts
We allow tuples to be passed directly to `.set` for updates, which is nice when I just want to do a one-off update of a single column. It'd be nice to have a similarly lightweight API for inserts.

sgrif:
This provides a more lightweight API for inserts that mirrors what is
possible with `.set` for updates. The implementation provided ensures
that all elements within the tuple are valid for the same table, and
that there's no funkiness like sticking an upsert value in there.
(Secret way to break Diesel -- It will totally allow you to stick a
`Vec` in there, and will generate invalid SQL as a result. We should fix
that one eventually, but it's pretty low priority)

One thing I would also like to allow is to allow mixing ad-hoc values
and `Insertable` structs. (See 10374a5026bcab2202f7f67f057493c4636d547b
for rationale). I tried a naive approach where I moved the parens out
from the `InsertValues` impl on tuples, and into `InsertStatement`, but
that breaks PG upsert, so I'll need to spend more time on it. I've left
the test cases for that in place, but I will defer actually making them
pass for a later PR in order to keep the PR size and scope reasonable.

I expected to need more tests/compile-tests, but what's included here
felt like it pretty exhaustively tests everything.

Fixes #789

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -48,7 +48,13 @@ pub trait InsertValues<T: Table, DB: Backend> {
 #[derive(Debug, Copy, Clone)]
 pub enum ColumnInsertValue<Col, Expr> {
     Expression(Col, Expr),
-    Default(Col),
+    Default,
+}
+
+impl<Col, Expr> Default for ColumnInsertValue<Col, Expr> {
+    fn default() -> Self {
+        ColumnInsertValue::Default
+    }
 }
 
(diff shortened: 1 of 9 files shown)


sgrif:
Going to mark the failing test as `#[should_panic]`, as its solution is the same as the other 2 mentioned in the PR description (the issue is that we're currently generating `VALUES (('foo'), ('bar'))` which is valid with most expressions, but invalid with `DEFAULT`.

Issue #1171, opened 2017-09-16
Diesel cli generates invalid joinable!() when using postgres schema


antab:
Using the following SQL:
```sql
CREATE SCHEMA test;
CREATE TABLE test.a(
        id serial NOT NULL,
        CONSTRAINT a_pkey PRIMARY KEY (id)

);
CREATE TABLE test.b(
        id serial NOT NULL,
        parent serial NOT NULL,
        CONSTRAINT b_pkey PRIMARY KEY (id)

);
ALTER TABLE test.b ADD CONSTRAINT ab_fkey FOREIGN KEY (parent)
REFERENCES test.a (id) MATCH FULL
ON DELETE NO ACTION ON UPDATE NO ACTION;
```
Then `diesel print-schema --schema test --database-url postgresql://...` will then output:

```rust
pub mod test {
    table! {
        test.a (id) {
            id -> Int4,
        }
    }

    table! {
        test.b (id) {
            id -> Int4,
            parent -> Int4,
        }
    }

    joinable!(test.b -> test.a (parent));
}
```
This won't compile and removing `test.` from the joinable!() fixes the issue.

PR #1172, opened 2017-09-16
Diesel cli: Only print the table name for joinable!().

antab:
This fixes diesel-rs/diesel#1171


Eijebong:
Can you also add a test for this ? :)

antab:
Take 3, I added a test and removed the stray file from the last one.

PR #1175, opened 2017-09-18
Generate compile time error for table! double import

### LINKED ISSUE #624: infer_schema! errors when column is the same name with the table it contains
Table which has a column name same to the table such as

```sql
CREATE TABLE message(
id serial PRIMARY KEY NOT NULL,
message character varying
);
```

will error with:

```rust
error[E0252]: a type named `message` has already been imported in this module
 --> src/schema.rs:3:1
  |
3 | infer_schema!("dotenv:DATABASE_URL");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | previous import of `message` here
  | already imported
```
A workaround is to simply not use a column with the same name as the table, such as
```sql
CREATE TABLE message(
id serial PRIMARY KEY NOT NULL,
body character varying
);
```
This is a minor issue, but the rust error message is a bit misleading.

notryanb:
Generates a better error message when a column
shares the same name as its table.

Fixes #624

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -654,7 +654,23 @@ macro_rules! table_body {
             /// table struct renamed to the module name. This is meant to be
             /// glob imported for functions which only deal with one table.
             pub mod dsl {
-                pub use super::columns::{$($column_name),+};
+                $(static_cond! {
+                    if $table_name == $column_name {
+                        compile_error!(concat!(
+                            "Column `",
+                            stringify!($column_name),
+                            "` cannot be named the same as its table.\n \
+                            You may use `#[sql_name = \"",
+                            stringify!($column_name),
+                            "\"]` to reference the table's `",
+                            stringify!($column_name),
+                            "` column. \n \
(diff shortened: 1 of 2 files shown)


diesel/src/macros/mod.rs
```diff
@@ -632,6 +632,15 @@ macro_rules! table_body {
         current_column_doc = [],
         current_column_sql_name = [],
     ) => {
+        $(static_cond! {
+            if $table_name == $column_name {
+                compile_error!(concat!(
+                    "Column ",
+                    stringify!($column_name),
+                    " cannot be named the same as its table"
```
Eijebong: Maybe you should point to
 ```
#[sql_name = "users"]
users_
```

notryanb: Ah, good point! I totally forgot about this.

sgrif:
Should we skip generating the DSL module here too? (Otherwise that error message will still appear)

notryanb:
@sgrif I think that makes sense. It's redundant and confusing if they're both reporting the same error with different messages.

Issue #1166, opened 2017-09-15
Our PG upsert API is clunky and we should fix it


sgrif:
I've really disliked using our upsert API in practice. You basically always need to pull `records.on_conflict(...)` into a local variable since it's going to span multiple lines, and it's really hard to name that variable. https://github.com/rust-lang/crates.io/blob/9c8947c4d171706f490dfd8dfeb932a714d252be/src/download.rs#L43-L50 https://github.com/rust-lang/crates.io/blob/9c8947c4d171706f490dfd8dfeb932a714d252be/src/categories.rs#L125-L140

I'd like to change our API to something more fluent, which more naturally allows for line breaks where they're likely to occur. I'd like to make the following changes:

```rust
// OLD
insert(records.on_conflict_do_nothing()).into(table)

// NEW
insert(records).into(table).on_conflict_do_nothing()

// OLD
insert(records.on_conflict(target, do_nothing())).into(table)

// NEW
insert(records).into(table).on_conflict(target).do_nothing()

// OLD
insert(records.on_conflict(target, do_update().set(...))).into(table)

// NEW
insert(records).into(table).on_conflict(target).do_update().set(...)

// ALTERNATE
insert(records).into(table).on_conflict(target).do_update_set(...)

// ALTERNATE 2
insert(records).into(table).on_conflict(target).do_update(...)
```

alexcameron89:
I'll give this one a try 👍

PR #1182, opened 2017-09-21
Deprecate `insert` in favor of `insert_into`

### LINKED ISSUE #1167: Consider changing `insert(records).into(table)` to `insert_into(table).values(records)`
There are two main things that got me thinking about this change. The first is [this code](https://github.com/diesel-rs/diesel/pull/1165/files#diff-0f3c021dc67b888eef8a6619689208cbR19). It's a papercut, but I don't like the fact that if you get the argument to `insert` wrong, the message is going to be that you passed the wrong number of arguments to `into`. Maybe this isn't a big deal (I don't think it's ever come up in gitter), but it's a paper cut I'd like to avoid, and I think it will be easier to run into once #789 is done.

The second is that I've been thinking about an API to represent `INSERT INTO table (column_list) SELECT ...`. The best idea I've seen based on our current API is `insert(select_statement).into(column_list)`, but that feels a little implicit to me. If we changed the insert API, I could see something like `insert_into(table).columns(column_list).values(select_statement)` or `insert_into(table).from_select(column_list, select_statement)` being much more reasonable.

sgrif:
This change deprecates inserts in the form `insert(values).into(table)`
in favor of `insert_into(table).values(values)`.

There are several reasons for this change:

- The new API more closely mirrors the SQL it creates.
- The old API conflicted with `Into::into`, meaning that if the argument
  to `insert` was wrong, the error message would be that `into` expected
  0 arguments and got 1.
- Things like `default_values` fit more nicely with this API (especially
  after the upsert changes in #1166 land)
- The API for inserting a select statement with an explicit column list
  will work better with this API

I've replaced all instances of `insert` with the new API, except for the
doctests of the deprecated functions.

(As a reminder in case you're wondering what the 0.99.0 is, the plan is
that the next version will be 1.0. 0.99.0 will be identical to 1.0,
except 1.0 will have no deprecated code)

Fixes #1167.


sgrif:
IDK why @diesel-rs/reviewers isn't showing up in the list I can request code review from... Maybe I have to ping it to convince GH that it exists.

Also FYI, the stats on this are huge, but the PR is actually quite small. The *vast* majority of the changes are churn from changing all examples/tests to use the new API

sgrif:
A-ha! I figured it out

.travis.yml
```diff
@@ -15,10 +15,11 @@ before_script:
 script:
 - |
   if [[ "$TRAVIS_RUST_VERSION" == nightly* ]]; then
-    (cd diesel && travis-cargo test -- --no-default-features --features "unstable chrono $BACKEND")
+    (cd diesel && travis-cargo test -- --no-default-features --features "unstable extras $BACKEND")
```
Eijebong: Oh...

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -20,7 +20,7 @@ pub trait ExpressionMethods: Expression + Sized {
     /// # }
     /// #
     /// # fn main() {
-    /// #     use self::users::dsl::*;
```
Eijebong: 👍 This commit didn't have enough git churn in it :D

CHANGELOG.md
```diff
@@ -11,17 +11,32 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
...
-* Added `insert(&default_values())` as a replacement for
+* Added `insert_into(table).default_values()` as a replacement for
   `insert_default_values()`
 
+* Added `insert_into(table).values(values)` as a replacement for
+  `insert(values).into(table)`.
+
+* Added support for MySQL's `REPLACE INTO` as `replace_into(table)`.
+
+* Added `replace_into(table).values(values)` as a replacement for
```
Eijebong: Why isn't you commit talking about that ? Should this have been in another PR ?

sgrif: Given that `insert_or_replace` is exactly analogous to `insert`, it made sense to do them at the same time. While working on it, I noticed that having it be `INSERT OR REPLACE` was unnecessary, and it could just be `REPLACE` which is compatible with MySQL.

PR #1184, opened 2017-09-21
Allow `#[derive(Queryable)]` for structs with `Queryable` fields

### LINKED ISSUE #1006: [feature request] Support  nested Queryable, Identifiable (and like those)
Current diesel doesn't support nested Queryables nor Identifiables but I'd like to write code like below:

```rust

#[derive(Eq, PartialEq, Hash)]
#[derive(Queryable)]
struct UserId(pub i32);

#[derive(Queryable, Identifiable)]
struct User {
    // nested Queryable
    pub id: UserId,
    pub name: String,
}

#[derive(Queryable, Identifiable)]
#[table_name = "users"]
struct UserDetail {
    // nested Queryable, Identifiable.
    //  May need a certain attribute like `#[have_primary_key(id)]`
    pub user: User,
    pub email_address: String,
    pub screen_name: String,
}

fn find_user(cn: &SomeConnection, user_id: UserId) -> Result<User, Error> {
    //...
    // If `UserId` is `AsExpression` and can be use like `find(user_id)`, it's also awesome but not scope of this issue
    users.find(user_id.0)
        .select(((id), name))
        // or, allowing `select((id, name))` will be useful
        .get_result(cn)
 
}
fn find_user_detail(cn: &SomeConnection, user_id: UserId) -> Result<UserDetail, Error> {
    //...
    users.find(user_id.0)
        .select((((id), name), email_address, screen_name))
        // the same above. allowing `select((id, name, email_address, screen_name))` will be usefull
        .get_result(cn)
}
```

In summary,

* A struct is Queryable when its direct fields *and* fields of its child structs are isomorphic to the selecting sql fields
* A struct is Identifiable when it has a primary key *or* one of its child structs have a primary key

As far as `Queryable` is concerned, nested deriving itself is possible but impossible to `get_result` from any isomorphic sql types. (This may not the matter of `Queryabel` but the matter of `FromSql` and `ToSql`.)

Structs like `struct UserId(i32);` is very useful to avoid confusions when handling many table ids. Is it possibel to implement these features?

sgrif:
Previously we only allowed structs to have fields which implementd
`ToSqlRow`. This meant that `Queryable` could not be composed. This
lifts the restriction to allow any `Queryable` type. This does mean that
things which implement `FromSqlRow` will also need to implement
`Queryable` to be used here, but they should have already been doing
that.

Fixes #1006.

### DIFF
--- a/diesel/src/types/impls/primitives.rs
+++ b/diesel/src/types/impls/primitives.rs
@@ -128,6 +128,19 @@ where
     }
 }
 
+impl<'a, T: ?Sized, ST, DB> ::Queryable<ST, DB> for Cow<'a, T>
+where
+    T: 'a + ToOwned,
+    DB: Backend + HasSqlType<ST>,
+    Self: ::types::FromSqlRow<ST, DB>,
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> Self {
+        row
+    }
+}
+
 use expression::bound::Bound;
 use expression::{AsExpression, Expression};
 impl<'a, T: ?Sized, ST> ::expression::AsExpression<ST> for Cow<'a, T>
--- a/diesel_codegen/src/lib.rs
+++ b/diesel_codegen/src/lib.rs
@@ -1,3 +1,5 @@
+#![recursion_limit = "1024"]
+
 // Built-in Lints
 #![deny(warnings, missing_copy_implementations)]
 // Clippy lints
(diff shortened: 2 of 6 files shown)


sgrif:
I actually think this might potentially improve the error messages, but I'll have to play with it more. Is there anything in particular you had in mind for a compile-fail test that isn't already covered by the existing tests?

killercup:
Just checking that we get some sensible error messages that mention the
right types as well as Queryable. I know we don't usually match against
concrete error messages, but I think for this we could use partial matches
without expecting much breakage from rustc updates.

Sean Griffin <notifications@github.com> schrieb am So. 24. Sep. 2017 um
03:39:

PR #1185, opened 2017-09-21
Call `Insertable::values` much earlier in the chain

sgrif:
This change has two main benefits:

1. If an error message that is not related to the backend being used is
  going to occur, it will occur on the call to `values`, not the call to
  `execute`. This generally means better error messages.
2. It removes the `Copy` constraint from `InsertStatement`, which will
  let us deal with owned values more frequently (and follow the same
  rules as `SelectStatement`, which can be executed multiple times only if
  it is `Copy`


notryanb:
👍

PR #1187, opened 2017-09-22
Deprecate `on_conflict_do_nothing`, add it to `InsertStatement` instead

sgrif:
This deprecates the method `on_conflict_do_nothing` which was present on
`Insertable` structs, and instead moves it onto `InsertStatement`
instead. Code which was previously written as
`.values(x.on_conflict_do_nothing())` can be written as
`.values(x).on_conflict_do_nothing()`. This allows a more fluent form
where line breaks can be more naturally introduced (which will be a much
bigger deal for `on_conflict(..., do_update().set(...))`

/cc @alexcameron89 who originally paired with me on this

### DIFF
--- a/diesel/src/pg/upsert/on_conflict_extension.rs
+++ b/diesel/src/pg/upsert/on_conflict_extension.rs
@@ -1,5 +1,8 @@
-pub use super::on_conflict_clause::*;
-pub use super::on_conflict_target::*;
+use expression::operators::Eq;
+use query_builder::insert_statement::{InsertStatement, UndecoratedInsertRecord};
+use super::on_conflict_actions::*;
+use super::on_conflict_clause::*;
+use super::on_conflict_target::*;
 
 /// Adds extension methods related to PG upsert
 pub trait OnConflictExtension {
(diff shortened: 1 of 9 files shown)


diesel/src/pg/upsert/on_conflict_extension.rs
```diff
@@ -199,4 +204,76 @@ pub trait OnConflictExtension {
...
+    /// let inserted_row_count = diesel::insert_into(users)
+    ///     .values(&vec![user, user])
+    ///     .on_conflict_do_nothing()
+    ///     .execute(&conn);
+    /// assert_eq!(Ok(1), inserted_row_count);
+    /// # }
+    /// ```
+    pub fn on_conflict_do_nothing(
+        self,
+    ) -> InsertStatement<T, OnConflictValues<U, NoConflictTarget, DoNothing>, Op, Ret> {
```
notryanb: @sgrif Do you think this line warrants line breaks (just over 80 chars)? I'm referencing #120 long function signatures.

sgrif: This is what `rustfmt` decided to do.

PR #1190, opened 2017-09-23
Change `between` to take two arguments instead of a range

sgrif:
I want to be able to write `.between(now, now - 90.days())`. Having a
range requires that the lower and upper bound be the same type in Rust,
rather than requiring that they be the same type in SQL.

### DIFF
--- a/diesel/src/expression_methods/global_expression_methods.rs
+++ b/diesel/src/expression_methods/global_expression_methods.rs
@@ -290,7 +290,8 @@ pub trait ExpressionMethods: Expression + Sized {
         LtEq::new(self, other.as_expression())
     }
 
-    /// Creates a SQL `BETWEEN` expression using the given range.
+    /// Creates a SQL `BETWEEN` expression using the given lower and upper
+    /// bounds.
     ///
     /// # Example
     ///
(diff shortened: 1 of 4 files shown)


diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -249,25 +249,25 @@ pub trait ExpressionMethods: Expression + Sized {
     }
 
     /// Creates a SQL `BETWEEN` expression using the given range.
```
notryanb: No longer takes range? Should this say "using given bound"?

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -249,25 +249,25 @@ pub trait ExpressionMethods: Expression + Sized {
...
-        )
+    fn between<T, U>(self, lower: T, upper: U) -> Between<Self, And<T::Expression, U::Expression>>
+    where
+        T: AsExpression<Self::SqlType>,
+        U: AsExpression<Self::SqlType>,
+    {
+        Between::new(self, And::new(lower.as_expression(), upper.as_expression()))
     }
 
     /// Creates a SQL `NOT BETWEEN` expression using the given range.
```
notryanb: Same thing here.

notryanb:
LGTM 👍 . Also - are there any performance benefits to using two args instead of a range?

sgrif:
No. A struct with 2 fields will have the same in memory representation as passing the args separately (more or less)

Issue #1093, opened 2017-08-13
Explicitly document --features and their dependencies in Getting Started Guide


katrinabrock:
I'm happy to submit a PR for this soon.

## Problem Description
"A Note on Installing diesel_cli" is helpful, but could be way more helpful to make install painless.

 Two main things
1. Possible arguments to `--features` (backends) should be explicitly listed. Maybe there is a cargo way to figure out what they are? Even if there is, it would be nice to explicitly list them in the getting started guide to make it rust-beginner friendly.

2. For each backend, it would be nice to list the exact package name needed for various OS package managers (`brew`, `apt-get`, `dnf`, do we need to cover Windows?)

## Checklist

- [X ] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

killercup:
Yes please! I'm not sure I'd put all that in the Getting Started guide, though. Maybe add a new document about _all_ the feature toggles Diesel has? The 3 backends are the relevant ones for the CLI, and a good start; but the same document could also talk about the features you need to enable in `diesel` (and `diesel-codegen`) to get DateTime, JSON, etc. support.

katrinabrock:
For a guide to installing dependencies, do we need to cover windows or just Mac and common Linux flavors?

killercup:
Eventually, we should cover every platform our users use. :) But we can absolutely start out with just Mac and Debian, for example!

PR #1191, opened 2017-09-23
Add db client installation guide

katrinabrock:
#1093 

This basically expands the getting started guide's direction to install database client libraries "using the usual way to do this depending on your operating system".

I tested all of these except OS X commands which I got from @sgrif . `apt-get` I tested on Debian. `yum` I tested on CentOS. Of course, it would be nice to add Windows.

I'm a little concerned that `sudo apt-get install postgresql-server-dev-all` is installing way more than necessary and there might be a skinnier package that would work. Not sure how big a deal that is.

### DIFF
--- a/guide_drafts/backend_installation.md
+++ b/guide_drafts/backend_installation.md
@@ -1,24 +1,34 @@
 # Installing Backend Client Libraries
 
-Diesel supports SQLite, PostgreSQL, and MySQL as backend databases. By default, `diesel_cli` requires the client library of all three backends to be installed. If one is missing, then `cargo install diesel_cli` you will throw an error like:
+Diesel supports SQLite, PostgreSQL, and MySQL as backend databases.
+By default, `diesel_cli`
+requires the client library of all three backends to be installed.
+If one is missing, then `cargo install diesel_cli` you will throw an error like:
 
 ```
 note: ld: library not found for -lmysqlclient
 clang: error: linker command failed with exit code 1 (use -v to see invocation)
 ```
-To install `diesel_cli` without all backends, specify `--no-default-features`. Use `--features` option to specify `postgres`, `sqlite`, and/or `mysql`. For example, to install with sqlite only, run:
+To install `diesel_cli` without all backends,
+specify `--no-default-features`.
+Use `--features` option to specify `postgres`, `sqlite`, and/or `mysql`.
+For example, to install with sqlite only, run:
 
 ```
 cargo install diesel_cli --no-default-features --features sqlite
 ```
 
-For projects that depend on diesel, you can specify which backend is required in the `Cargo.toml`. For example:
+For projects that depend on diesel,
+you can specify which backend is required in the `Cargo.toml`.
+For example:
 ```
 [dependencies]
 diesel = { version = "X.X.X", features = ["sqlite"] }
 ```
 
-Below are command to run to install appropriate database clients with various package managers.
+Below are command to run
+to install appropriate database clients
+with various package managers.
 
 ## Debian/Ubuntu
 
--- a/guide_drafts/backend_installation.md
+++ b/guide_drafts/backend_installation.md
@@ -1,6 +1,6 @@
 # Installing Backend Client Libraries
 
-Diesel supports SQLite, PostgreSQL, and MySQL as backend databases.
+Diesel supports SQLite, PostgreSQL, and MySQL as database backends.
 By default, `diesel_cli`
 requires the client library of all three backends to be installed.
 If one is missing, then `cargo install diesel_cli` you will throw an error like:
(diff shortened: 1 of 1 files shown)


sgrif:
This looks great! This will be perfect to link to from the getting started guide as well. I'll leave this open for the rest of the team to leave feedback on if they have any, otherwise I'll deploy this to the website tomorrow.

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
+# Installing Backend Client Libraries
+
+Diesel supports SQLite, PostgreSQL, and MySQL as backend databases. By default, `diesel_cli` requires the client library of all three backends to be installed. If one is missing, then `cargo install diesel_cli` you will throw an error like:
```
killercup: > as backend databases

Just an FYI: I think in most other documents we call it 'database backend'

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+note: ld: library not found for -lmysqlclient
+clang: error: linker command failed with exit code 1 (use -v to see invocation)
+```
+To install `diesel_cli` without all backends, specify `--no-default-features`. Use `--features` option to specify `postgres`, `sqlite`, and/or `mysql`. For example, to install with sqlite only, run:
```
killercup: "Use [the] `--features` option"

Or maybe "`--features` flag"? These parameters are often referred to as "CLI flags"

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
+# Installing Backend Client Libraries
+
+Diesel supports SQLite, PostgreSQL, and MySQL as backend databases. By default, `diesel_cli` requires the client library of all three backends to be installed. If one is missing, then `cargo install diesel_cli` you will throw an error like:
+
+```
+note: ld: library not found for -lmysqlclient
+clang: error: linker command failed with exit code 1 (use -v to see invocation)
+```
```
killercup: I always add empty lines between code blocks and paragraphs as it improves readability a bit (easier to see segments of text) but Markdown doesn't require it

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+To install `diesel_cli` without all backends, specify `--no-default-features`. Use `--features` option to specify `postgres`, `sqlite`, and/or `mysql`. For example, to install with sqlite only, run:
+
+```
+cargo install diesel_cli --no-default-features --features sqlite
+```
+
+For projects that depend on diesel, you can specify which backend is required in the `Cargo.toml`. For example:
```
killercup: You can also specify multiple backends, so maybe use plural here?

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+```
+[dependencies]
+diesel = { version = "X.X.X", features = ["sqlite"] }
+```
+
+Below are command to run to install appropriate database clients with various package managers.
```
killercup: "command[s]"

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Debian/Ubuntu
+
+### SQLite
+`sudo apt-get install libsqlite3-dev`
```
killercup: Again, I tend to add empty lines between blocks (here headline and paragraph)

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+
+### PostgreSQL
+`sudo apt-get install postgresql-server-dev-all`
+
+### MySQL
+
+1. Install add the MySQL APT repository.
```
killercup: "Install add"? What to you mean? "Install the following to add …"?

guide_drafts/backend_installation.md
```diff
@@ -0,0 +1,74 @@
...
+`sudo apt-get install libsqlite3-dev`
+
+### PostgreSQL
+`sudo apt-get install postgresql-server-dev-all`
+
+### MySQL
+
+1. Install add the MySQL APT repository. 
+
+```
```
killercup: If you mean for this block (and following line of text) to be part of the list item, I think you need to indent it by at least 3 spaces

notryanb:
LGTM 👍

guide_drafts/backend_installation.md
katrinabrock: I checked the cargo man page and it's listed under `Options:`. I always think of a "CLI flag" as an option that doesn't take an argument. How about:
>Use cargo's `--features` option

katrinabrock:
@killercup thanks for your suggestions. I have made those edits.

guide_drafts/backend_installation.md
killercup: > I always think of a "CLI flag" as an option that doesn't take an argument

That's a good point! I hadn't thought about it before and that totally makes sense

Issue #1091, opened 2017-08-12
Print name of migration when running/reverting


mwillsey:
Right now, the output of running or reverting migrations only lists the timestamp:

```
$ diesel migration revert
Rolling back migration 20170812171255
$ diesel migration run
Running migration 20170812171255
```

These aren't very informative, so it would be nice if it printed the name as well. Even just the full migration folder name would be preferable.

Diesel CLI version: 0.15.0

juliusdelta:
@killercup I Can take this since I'm pretty familiar with that code if that's cool!

sgrif:
@juliusdelta Go ahead

juliusdelta:
So the first way I can think to do this would be:

1. I could just print & parse the `migration.file_path()` which returns something like: `"/path/diesel_cli_test/migrations/20170924041631_create_posts"`. Then parse it to just print `20170924041631 -> Create Posts` (or something similar) and use that to replace https://github.com/diesel-rs/diesel/blob/master/diesel/src/migrations/mod.rs#L313 and corresponding `revert` output.
  a. Might use a regex possibly to pull data after the last `/` then split it up by `_`s.

It'd have to be it's own method for sure. I'd probably call it `migration_output_information()` or something.

Thoughts?

juliusdelta:
Actually `std::Path` might help https://doc.rust-lang.org/std/path/enum.Component.html

henryboisdequin:
Can this issue be closed as this is fixed?

PR #1195, opened 2017-09-24
Change migration output information to show file_path()

juliusdelta:
connect #1091 

So this is my first iteration of the solution for #1091. 

When running/reverting migrations it will look like:
```bash
~/path/diesel_cli_test(master*) » diesel migration run                                                                                                                jd@Jonathans-MacBook-Pro
Running migration 20170924041631_create_posts
------------------------------------------------------------
~/path/diesel_cli_test(master*) » diesel migration revert                                                                                                             jd@Jonathans-MacBook-Pro
Rolling back migration 20170924041631_create_posts
```

This may need to be refactored. The first thing I tried was to use 
```rust
fn format_migration_output(migration: &Migration) -> String {
    let output = migration.file_path().unwrap()
        .components()
        .map(|x| x.as_os_str().to_str())
        .collect::<Vec<_>>()
        .last()
        .unwrap()
        .unwrap();
    (format!("{}", output))
}
```

Anyone have any thoughts on refactoring this? Also is this a good solution? Or do we want to parse the output string further to be prettier?

#### Todo
- [x]  Refactor? (Pending Feedback)
- [x] Add/update test coverage
- [ ] Add to Changelog

Edit: Update comment as of [this commit](https://github.com/diesel-rs/diesel/pull/1195/commits/680228a1be9c3f4881d798e97945130b62c0672f)

### DIFF
--- a/diesel_migrations/migrations_internals/src/migration.rs
+++ b/diesel_migrations/migrations_internals/src/migration.rs
@@ -2,6 +2,7 @@ use diesel::connection::SimpleConnection;
 use super::{MigrationError, RunMigrationsError};
 
 use std::path::{Path, PathBuf};
+use std::fmt;
 
 pub trait Migration {
     fn version(&self) -> &str;
@@ -9,6 +10,33 @@ pub trait Migration {
     fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
     fn file_path(&self) -> Option<&Path> {
         None
+    }
+}
+
+#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy)]
+pub struct MigrationName<'a> {
+    pub migration: &'a Migration,
+}
+
+pub fn name(migration: &Migration) -> MigrationName {
+    MigrationName {
+        migration: migration,
+    }
+}
+
+impl<'a> fmt::Display for MigrationName<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let file_name = self.migration
+            .file_path()
+            .and_then(|file_path| file_path.file_name())
+            .and_then(|file| file.to_str());
+        if let Some(name) = file_name {
+            f.write_str(name)?;
+        } else {
+            f.write_str(self.migration.version())?;
+        }
+        Ok(())
     }
 }
 
--- a/diesel_cli/tests/migration_redo.rs
+++ b/diesel_cli/tests/migration_redo.rs
@@ -15,8 +15,8 @@ fn migration_redo_runs_the_last_migration_down_and_up() {
     let result = p.command("migration").arg("redo").run();
 
     let expected_stdout = "\
-Rolling back migration 12345
-Running migration 12345
+Rolling back migration 12345_create_users_table
+Running migration 12345_create_users_table
 ";
 
     assert!(result.is_success(), "Result was unsuccessful {:?}", result);
@@ -47,8 +47,8 @@ fn migration_redo_respects_migration_dir_var() {
         .run();
 
     let expected_stdout = "\
-Rolling back migration 12345
-Running migration 12345
+Rolling back migration 12345_create_users_table
+Running migration 12345_create_users_table
 ";
 
     assert!(result.is_success(), "Result was unsuccessful {:?}", result);
@@ -79,8 +79,8 @@ fn migration_redo_respects_migration_dir_env() {
         .run();
 
     let expected_stdout = "\
-Rolling back migration 12345
-Running migration 12345
+Rolling back migration 12345_create_users_table
+Running migration 12345_create_users_table
 ";
 
     assert!(result.is_success(), "Result was unsuccessful {:?}", result);
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -323,11 +323,7 @@ where
 {
     conn.transaction(|| {
         if migration.version() != "00000000000000" {
-            try!(writeln!(
-                output,
-                "Running migration {}",
-                migration.version()
-            ));
+            try!(writeln!(output, "Running migration {}", name(&migration)));
         }
         try!(migration.run(conn));
         try!(conn.insert_new_migration(migration.version()));
@@ -344,7 +340,7 @@ fn revert_migration<Conn: Connection>(
         try!(writeln!(
             output,
             "Rolling back migration {}",
-            migration.version()
+            name(&migration)
         ));
         try!(migration.revert(conn));
         let target = __diesel_schema_migrations.filter(version.eq(migration.version()));


diesel/src/migrations/mod.rs
```diff
@@ -297,6 +297,17 @@ where
     Ok(())
 }
 
+fn format_migration_output(migration: &Migration) -> String {
```
weiznich: I think we should rather implement `Display` for Migration instead of implementing the formating as custom function.

killercup: Indeed, that's a good suggestion. That would also help us get rid of all the `unwrap` (but we'll have to convert them to `fmt::Error`s which dont carry any information)

juliusdelta:
I've updated the commits after some help from @sgrif this _should_ be good to go. If I need to add to changelog.md just let me know and I'll rebase and get that done.

sgrif:
CI says no

juliusdelta:
I managed to change CI's mind.

diesel/src/migrations/migration.rs
```diff
@@ -97,6 +98,31 @@ pub fn version_from_path(path: &Path) -> Result<String, MigrationError> {
         })
 }
 
+#[allow(missing_debug_implementations)]
```
sgrif: Surprised this didn't trigger a missing copy implementations as well. Should this type implement `Copy`?

diesel/src/migrations/migration.rs
```diff
@@ -97,6 +98,31 @@ pub fn version_from_path(path: &Path) -> Result<String, MigrationError> {
...
+    MigrationName {
+        migration: migration,
+    }
+}
+
+impl<'a> fmt::Display for MigrationName<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if let Some(file_path) = self.migration.file_path() {
+            if let Some(file) = file_path.file_name() {
+                if let Some(name) = file.to_str() {
```
sgrif: How about this:

```rust
let file_name = self.migration.file_path()
    .and_then(|file| file_path.file_name())
    .and_then(|file| file.to_str());
if let Some(name) = file_name {
    f.write_str(name)?;
}
Ok(())
```

diesel/src/migrations/mod.rs
```diff
@@ -270,7 +270,6 @@ fn migrations_in_directory(path: &Path) -> Result<Vec<Box<Migration>>, Migration
         .map(|e| migration_from(e.path()))
         .collect()
 }
-
```
sgrif: I don't think you meant to delete this

diesel/src/migrations/migration.rs
sgrif: Also this needs to fall back to printing the version if no file path is present. Embedded migrations return `None` for the file path

juliusdelta:
Sorry just got back from a vacation of sorts. Checking into these changes now.

diesel/src/migrations/migration.rs
juliusdelta: Tbh I'm no1 100% sure if it _should_ as I'm unfamiliar with the case in which it's necessary. If you think it should I can add it as well.

juliusdelta:
I rebased on master to handle conflicts.

diesel/src/migrations/migration.rs
sgrif: @juliusdelta What do you think about changing this pyramid to use `and_then` like my comment above?

diesel/src/migrations/migration.rs
sgrif: This type should implement `Copy`

sgrif:
The rebase needs more work. The `migrations` module was moved out of Diesel and into another crate. All these files should have been deleted, I'm not sure why they're still around on master. Hopefully CI fails (I'm really scared if it doesn't)

diesel/src/migrations/migration.rs
juliusdelta: @sgrif I had but changed it back for some reason, and I can't quite remember why. Anyway, since an embedded migration would return `None` for `file_path()` I can just add an `else` like:

```rust
impl <'a> fmt::Display for MigrationName<'a> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let file_name = self.migration.file_path()
            .and_then(|file| file_path.file_name())
            .and_then(|file| file.to_str());
        if let Some(name) = file_name {
            f.write_str(name)?;
        } else {
            f.write_str(self.migration.version())?;
        }
        Ok(())
    }
}
```
to handle that case. Right?

juliusdelta:
Ok will take a look

diesel/src/migrations/migration.rs
sgrif: Yes

sgrif: Or if you wanted to get really fancy:

```rust
impl <'a> fmt::Display for MigrationName<'a> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let file_name = self.migration.file_path()
            .and_then(|file| file_path.file_name())
            .and_then(|file| file.to_str())
            .map(|name| f.write_str(name))
            .unwrap_or_else(|| f.write_str(self.migraiotn.version))
    }
}

Comment (killercup) [reply diesel/src/migrations/migration.rs]: Fancy, you say? Then by all means, try and go with this!

```rust
let file_name = self.migration.file_path()
    .and_then(Path::file_name)
    .and_then(OsStr::to_str) // ...
```

juliusdelta:
I only got moderately fancy. I called bankruptcy and archived the old branch and made a new one and copied my changes over to the new files. I didn't have the brain power to figure out the "proper git solution" lol. Hopefully this works as intended. Shall I add to the changelog @sgrif

sgrif:
I don't think this needs a changelog entry (it's a minor change in behavior, but not one that our users need to care about during migration)

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -12,6 +13,31 @@ pub trait Migration {
     }
 }
 
+#[allow(missing_debug_implementations)]
+pub struct MigrationName<'a> {
```
sgrif: This type should still implement `Copy`. ;) (I'm still really surprised that the `missing_copy_implementations` lint isn't triggering here)

juliusdelta:
Ok Implemented `Copy` on the `MigrationName` struct.

sgrif:
CI says "fuck you run `cargo fmt`"

juliusdelta:
Done

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -12,6 +13,33 @@ pub trait Migration {
     }
 }
 
+#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy)]
```
sgrif: Oh I wonder if the lint only triggers if the type has a clone impl. That would make sense.

Issue #290, opened 2016-04-20
Find a good place to document this thing


sgrif:
### TL;DR: UFCS can improve our error messages, and we should document this, but I'm not sure where the right place is.

Random trick for debugging when Diesel gives an unhelpful error message. Yesterday when I was implementing `QueryId` for our entire AST, I got the following error for a node I had missed:

```
tests/expressions/date_and_time.rs:35:10: 35:14 error: no method named `load` found for type `diesel::query_builder::SelectStatement<diesel::types::Integer, expressions::date_and_time::has_timestamps::columns::id, expressions::date_and_time::has_timestamps::table, diesel::query_builder::distinct_clause::NoDistinctClause, diesel::query_builder::where_clause::WhereClause<diesel::expression::predicates::Lt<expressions::date_and_time::has_timestamps::columns::created_at, diesel::expression::now>>>` in the current scope
tests/expressions/date_and_time.rs:35         .load::<i32>(&connection);
                                               ^~~~
[...]

PR #1198, opened 2017-09-25
Feedback on updating API landing page

katrinabrock:
I'd like to make the [API landing page](http://docs.diesel.rs/diesel/index.html) a little more useful/welcoming. This is my first incremental step. Looking for feedback not just on this, but on what else should go here. Perhaps a debugging section including the info in #290? API Overview? (What would that entail?) I've talked to @sgrif a bit about this an d am interested in getting feedback from others.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -1,6 +1,6 @@
 //! # Diesel
 //!
-//! Diesel is an ORM and query builder designed to reduce the boilerplate for database interactions. 
+//! Diesel is an ORM and query builder designed to reduce the boilerplate for database interactions.
 //!
 //! * [Homepage](http://diesel.rs/)
 //! * [Getting Started Guide](http://diesel.rs/guides/getting-started/)
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -1,6 +1,17 @@
-//! Diesel is an ORM and query builder designed to reduce the boilerplate for database
-//! interactions. [A getting started guide](http://diesel.rs/guides/getting-started/) can be
-//! found on our website.
+//! # Diesel
+//!
+//! Diesel is an ORM and query builder designed to reduce the boilerplate for database interactions. 
+//!
+//! * [Homepage](http://diesel.rs/)
+//! * [Getting Started Guide](http://diesel.rs/guides/getting-started/)
+//! * [Source Code](https://github.com/diesel-rs/diesel/)
+//!
+//! ## Crates
+//!
+//! * **Diesel**: core libary, documented here
+//! * [**Diesel Codegen**](https://github.com/diesel-rs/diesel/tree/master/diesel_codegen): library of procedural macros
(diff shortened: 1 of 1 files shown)


diesel/src/lib.rs
```diff
@@ -1,6 +1,17 @@
...
-//! found on our website.
+//!# Diesel
+//!
+//! Diesel is an ORM and query builder designed to reduce the boilerplate for database interactions. 
+//!
+//! * [Homepage](http://diesel.rs/)
+//! * [Getting Started Guide](http://diesel.rs/guides/getting-started/)
+//! * [Source Code](https://github.com/diesel-rs/diesel/)
+//!
+//!## Crates
```
katrinabrock: Not sure this is the best heading, but this list needs a heading. Suggestions?

juliusdelta: [Rocket uses 'Libraries'](https://api.rocket.rs/rocket/) which I think is what I would suggest here. Super stoked for this to be done btw.

juliusdelta:
Maybe a configuration section would be good with different config options and supported DB's? Rocket has a brief one on their [docs](https://api.rocket.rs/rocket/#configuration).

Like:
```
diesel = { version = "0.15.0", features = ["postgres"] }
diesel_codegen = { version = "0.15.0", features = ["postgres"] }
```

diesel/src/lib.rs
katrinabrock: "Libraries" sounds nicer, but is a CLI tool really a library?

juliusdelta: Good point. Hm. I'd almost suggest `Crates & Libraries` but there's only 3 things.

diesel/src/lib.rs
```diff
@@ -1,6 +1,17 @@
-//! Diesel is an ORM and query builder designed to reduce the boilerplate for database
-//! interactions. [A getting started guide](http://diesel.rs/guides/getting-started/) can be
-//! found on our website.
+//!# Diesel
```
killercup: Can you add a space after the `//!` and before the `#`? (Same below)

diesel/src/lib.rs
killercup: So I spend a surprisingly long amount of time reading the wikipedia pages for gasoline and diesel to see what the things that make up the fuel are called… and all I found was "component"

notryanb: `Tools & Libraries`? 🤔 I think what you have written does a good job (Crates as the heading). Technically, each one of those "components" is a crate and there is a short description detailing its purpose. I don't see any ambiguous language that would confuse someone or inhibit them from getting to the correct documentation from your links. 

... although, they are "containers" that "hold diesel" code. We can call them `Tanks`.. 😆

katrinabrock:
fixed spacing :-)

killercup:
[Very close!](https://travis-ci.org/diesel-rs/diesel/jobs/282150517) I'll make rustfmt happy and merge this :)

katrinabrock:
Merp. I need to start running that locally, huh? Thanks @killercup.

killercup:
No worries :) (I set you editor to strip trailing whitespace in .rs files)

PR #1201, opened 2017-09-25
Fixing passing --database-url as global cli arg

weiznich:
--database-url is passed as global cli argument, this means it may be
  not part of the current ArgMatches. To solve this we need to look recursivly in all current subcommands for the database-url

### DIFF
--- a/diesel_cli/tests/migration_run.rs
+++ b/diesel_cli/tests/migration_run.rs
@@ -191,7 +191,7 @@ fn migration_run_runs_pending_migrations_custom_database_url_1() {
 
     assert!(!db.table_exists("users"));
 
-    let result = p.command("migration")
+    let result = p.command_without_database_url("migration")
         .arg("run")
         .arg("--database-url")
         .arg(db_url)
@@ -223,7 +223,7 @@ fn migration_run_runs_pending_migrations_custom_database_url_2() {
 
     assert!(!db.table_exists("users"));
 
-    let result = p.command("migration")
+    let result = p.command_without_database_url("migration")
         .arg("--database-url")
         .arg(db_url)
         .arg("run")
--- a/diesel_cli/tests/migration_run.rs
+++ b/diesel_cli/tests/migration_run.rs
@@ -173,3 +173,67 @@ fn any_pending_migrations_after_running_and_creating() {
 
     assert!(result.stdout().contains("true\n"));
 }
+
+#[test]
+fn migration_run_runs_pending_migrations_custom_database_url_1() {
+    let p = project("migration_run").folder("migrations").build();
+    let db_url = p.database_url();
+    let db = database(&db_url);
+
+    // Make sure the project is setup
+    p.command("setup").run();
+
+    p.create_migration(
+        "12345_create_users_table",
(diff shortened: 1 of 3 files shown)


sgrif:
Does setting the [`global` config](https://docs.rs/clap/2.26.2/clap/struct.Arg.html#method.global) accomplish the same thing? If not I think we should fix this in clap.

weiznich:
> Does setting the global config accomplish the same thing? If not I think we should fix this in clap.

The documentation of global states explicitly the current behavior, so this seems to be desired or at least not fixable without changing larger parts of clap.

> NOTE: Global arguments, when matched, only exist in the command's matches that they were matched to. For example, if you defined a --flag global argument in the top most parent command, but the user supplied the arguments top cmd1 cmd2 --flag only cmd2's ArgMatches would return true if tested for ArgMatches::is_present("flag").

weiznich:
There is also [this issue](https://github.com/kbknapp/clap-rs/issues/978) on the clap repo, but setting this flag does not solve the problem…

sgrif:
Can you add a few test cases?

weiznich:
Something like the last commit?

diesel_cli/tests/migration_run.rs
```diff
@@ -173,3 +173,67 @@ fn any_pending_migrations_after_running_and_creating() {
 
     assert!(result.stdout().contains("true\n"));
 }
+
+#[test]
+fn migration_run_runs_pending_migrations_custom_database_url_1() {
+    let p = project("migration_run").folder("migrations").build();
```
sgrif: Every test case needs to have a unique project name.

sgrif:
The last commit is the sort of thing I had in mind, yes. However, those tests pass without your code changes

sgrif:
Based on some cursory testing, it looks like we *should* only need the `.setting(AppSettings::PropagateGlobalValuesDown)`. However, that setting causes `diesel --database-url=... database drop` to work, but not `diesel database --database-url=... drop`. The addition of the `database_url_from_cli` function doesn't seem to affect it.

weiznich:
>Based on some cursory testing, it looks like we should only need the .setting(AppSettings::PropagateGlobalValuesDown). However, that setting causes diesel --database-url=... database drop to work, but not diesel database --database-url=... drop. The addition of the database_url_from_cli function doesn't seem to affect it.

Executing `target/debug/diesel migration run --migration-dir some/migrations/dir  --database-url /tmp/test.db` fails for me without the `database_url_from_cli` function.

Also this function may fix the issue with `diesel --database-url=...`

weiznich:
>  However, those tests pass without your code changes.

I've modified those tests to run without the `DATABASE_URL` environment variable. Now `migration_run_runs_pending_migrations_custom_database_url_1` fails without the using `database_url_from_cli`

sgrif:
Can you add a note that we can remove that function once https://github.com/kbknapp/clap-rs/issues/978 is fixed?

sgrif:
You need to run rustfmt

weiznich:
Fixed by #1304

PR #1202, opened 2017-09-25
Allow subselects to reference columns from the outer table

sgrif:
This changes the semantics of `exists`, and the various array_comparison
functions/methods to allow the subselect passed to them to reference
columns from the outer table. In other words, this allows subselects
which are not valid queries on their own, but are valid expressions in
this context.

Note that this only allows the where clause to reference the outer
table, as it's much less useful to do so from elsewhere in the query.

There are a few side effects of this change that will need to be
addressed later on. The first is that `exists` can now take any
expression, not just select statements. We should limit this to
`BoxedSelectStatement` and `SelectStatement`. Second is that you can no
longer pass a subselect to the same table. I'm not entirely sure how we
go about fixing this just yet, but that's a much more niche use case
anyway.

There is still additional work to be done to make this useable though,
as it will still be impossible to construct many queries while most of
our DSL methods still have `AsQuery` bounds. They will need to be
loosened in a similar fashion to #1200.

### DIFF
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -139,7 +139,6 @@ pub trait MaybeEmpty {
 
 impl<ST, S, F, W, O, L, Of, G, FU> AsInExpression<ST> for SelectStatement<S, F, W, O, L, Of, G, FU>
 where
-    Self: Query<SqlType = ST>,
     Subselect<Self, ST>: Expression<SqlType = ST>,
 {
     type InExpression = Subselect<Self, ST>;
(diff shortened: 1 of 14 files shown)


leshow:
> Second is that you can no
longer pass a subselect to the same table. I'm not entirely sure how we
go about fixing this just yet, but that's a much more niche use case
anyway.

For clarity, does this mean something like `SELECT foo FROM tablea IN (SELECT bar FROM table_a)` is no longer possible, because they select from the same table?

sgrif:
Well that's not a valid query either way, but yes you are getting the general idea.

PR #1205, opened 2017-09-26
Remove special cased ordering in `migration list`

sgrif:
Diesel is not meant to have an opinion on versioning. The use of
timestamps is a convention that gives us alphabetically increasing
versions, but we should not give it special treatment. Additionally,
this behavior did not mirror the order in which the migrations will be
run.


juliusdelta:
@killercup I'm confident it wont.

PR #1207, opened 2017-09-26
Adds rust types to reserved names.

### LINKED ISSUE #1204: `table!` fails when column name is named `bool`
related #1110 

https://gitter.im/diesel-rs/diesel?at=59ca4bbbf7299e8f5373ae61

Emilgardis:
Fixes #1204


Emilgardis:
I considered sorting the (un)signed types properly, but I thought it wouldn't really matter.
Also, should we care about columns like `"i32"`?

sgrif:
I'd rather avoid speculatively doing this for all types, as there's no reason you couldn't just have a column called `char`. The only reason that `bool` is causing an issue is that it appears in the signature of a trait. I believe we can fix that by just changing `impl_query_id!` to use `::std::bool` instead of `bool`, and not touch codegen at all

sgrif:
I've opened https://github.com/rust-lang/rust/issues/44865 to see if we can get a better way to do this. In the short term I think we should add this to the crate root:

```rust
// Referenced by `impl_query_id!` to ensure we get the right type if users have
// a column called `bool`. This should be removed when the language lets us
// reference `bool` with an absolute path
#[doc(hidden)]
pub type RealBool = bool;
```

and change `impl_query_id!` to use `$crate::RealBool` instead of `bool`. We should also add tests for this. Simply adding a migration with a column that has the same name as each primitive should suffice. Your list is also missing `str`

sgrif:
I'm going to close this, as it's gotten out of date. I would still like to see the underlying issue fixed, but through the method I mentioned above rather than making these names reserved.

PR #1210, opened 2017-09-26
Add a method generic version of `AsExpression`

sgrif:
This effectively removes any reason for users to reference
`AsExpression` other than in where clauses. In cases where the SQL type
could not be inferred, `as_expression` was incredibly verbose to use.

The naming `as_` and `into_` are the standard rust conventions for
owned/borrowed conversions. I'm not super thrilled about `sql` as the
suffix, but it's the only thing I could think of that was concise,
specific enough to Diesel, and makes sense for people who aren't
implementing Diesel and don't necessarily know/care what an "expression"
is.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -69,6 +69,7 @@ pub mod dsl {
 pub use self::sql_literal::SqlLiteral;
 
 use backend::Backend;
+use dsl::AsExprOf;
 
 /// Represents a typed fragment of SQL.
 ///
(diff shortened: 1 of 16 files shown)


diesel/src/expression/mod.rs
```diff
@@ -89,15 +90,16 @@ impl<'a, T: Expression + ?Sized> Expression for &'a T {
...
-/// These types couldn't just implement [`Expression`](trait.Expression.html)
-/// directly, as many things can be used as an expression of multiple types.
-/// (`String` for example, can be used as either
-/// [`VarChar`](../types/type.VarChar.html) or
-/// [`Text`](../types/struct.Text.html)).
+/// Converts a type to its representation for use in Diesel's query builder.
 ///
-/// This trait allows us to use primitives on the right hand side of various
-/// expressions. For example `name.eq("Sean")`
+/// Implementations of this trait will generally do one of 3 things:
```
Eijebong: I like this, this is way clearer than the doc that was there before :+1:

PR #1212, opened 2017-09-26
Don't allocate when encoding `char` in `EscapeExpressionMethods`

sgrif:
Not that the 4 byte heap allocation is a huge bottleneck, but we can do
this without allocating so why not


sgrif:
On second thought, I don't want to add this much code to optimize something which is not a bottleneck on a function that is likely being used by nobody.

PR #1213, opened 2017-09-27
feat: return directly if records is empty when do batch insert in sqlite

xinghun92:
There is no need to start a transaction if no data to be inserted.

### DIFF
--- a/diesel/src/query_builder/insert_statement.rs
+++ b/diesel/src/query_builder/insert_statement.rs
@@ -195,6 +195,9 @@ where
 {
     fn execute(self, conn: &SqliteConnection) -> QueryResult<usize> {
         use connection::Connection;
+        if self.records.is_empty() {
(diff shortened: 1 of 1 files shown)


sgrif:
Why? What's the downside?

xinghun92:
I think there is no need to create a new transaction and do nothing. Does create a transaction in sqlite is zero cost?

sgrif:
The cost of an empty transaction is basically the cost of a ping. If you benchmark this change in isolation, you will see that there is a 2 microsecond difference (cost of doing nothing vs the cost of doing not nothing). However, if you were to execute `SELECT 1` in there (something we do every time a connection is removed or checked into the connection pool), the performance is virtually identical. If there were a more meaningful difference here, I'd be fine with adding this, but I don't want to speculatively add code that does exactly what occurs otherwise unless there's a real gain.

PR #1227, opened 2017-09-30
Rejigger our peano number implementation, document it

sgrif:
`Succ<Never>` is a really poor type to have appear in error messages.
Since we never care about the precise value unless that value is 1, I've
replaced it with two concrete types to represent one and more than one.
The error message still isn't *great*, but it's at least passable.

### DIFF
--- a/diesel/src/query_source/peano_numbers.rs
+++ b/diesel/src/query_source/peano_numbers.rs
@@ -1,0 +1,44 @@
+//! A simple implementation of peano numbers.
+//!
+//! This is used to enforce that columns can only be selected from a given from
+//! clause if their table appears exactly one time.
+
+/// A table never appears in the from clause.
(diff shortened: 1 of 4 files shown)


diesel/src/query_source/peano_numbers.rs
```diff
@@ -0,0 +1,44 @@
+//! A simple implementation of peano numbers.
+//!
+//! This is used to enforce that columns can only be selected from a given from
+//! clause if their table appears exactly one time.
+
+/// A table never appears in the from clause.
+#[allow(missing_debug_implementations, missing_copy_implementations)]
```
weiznich: Are there any particular reasons why this should not implement `Debug`, `Clone` and `Copy`?

sgrif: Because instances of this type never exist, and more code = slower build times

PR #1228, opened 2017-09-30
Documentation pass over the top level `query_builder` module

sgrif:
[ci skip]

### DIFF
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -50,6 +50,7 @@ use result::QueryResult;
 
 #[doc(hidden)]
 pub type Binds = Vec<Option<Vec<u8>>>;
+/// A specialized Result type used with the query builder.
 pub type BuildQueryResult = Result<(), Box<Error + Send + Sync>>;
 
 /// Apps should not need to concern themselves with this trait.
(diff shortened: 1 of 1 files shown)


diesel/src/query_builder/mod.rs
```diff
@@ -73,23 +74,40 @@ pub trait QueryBuilder<DB: Backend> {
...
-/// A complete SQL query with a return type. This can be a select statement, or
-/// a command such as `update` or `insert` with a `RETURNING` clause. Unlike
-/// [`Expression`](../expression/trait.Expression.html), types implementing this
+/// A complete SQL query with a return type.
+///
+/// This can be a select statement, or a command such as `update` or `insert`
+/// with a `RETURNING` clause. Unlike [`Expression`], types implementing this
 /// trait are guaranteed to be executable on their own.
+///
+/// Just because a type doesn't implement this trait doesn't mean that it
```
juliusdelta: Now that I re-read these, I'd probably change this to the inverse to 2 sentences.  "A type that doesn't implement this trait means that it doesn't have a return type. It doesn't mean that it isn't a complete SQL query. For example,..."

killercup: Yeah, three "doesn't"s is too much. Maybe:

> *Note:* There are executable SQL queries that do not implement this type. For example, an `INSERT` statement without a `RETURNING` clause can be executed, but will not implement `Query`. This is because it does not have a return type.

PR #1231, opened 2017-10-01
Restructure codegen crates

weiznich:
This pull request splits diesel_codegen into 2 crates.

One crate containing all model custom derives (`Queryable`,
`Identifiable`, `Insertable`, `AsChangeset`, `Associations`). This
crate (`diesel_codegen`) only consist of code not depending on diesel
itself. This allows us to reexport all model derives through diesel
itself. (A user only needs to write `#[macro_use] extern crate
diesel;` instead of `#[macro_use] extern crate diesel; #[macro_use]
extern crate diesel_codegen;`)

All codegen code that depends on diesel is moved to a new crate called
`diesel_proc_macro_internal`. This crate contains the custom derives
for `infer_schema!`, `infer_table_from_schema!` and
`embed_migrations!`. The corresponding macros are removed from diesel
itself and moved to a new crate called `diesel_proc_macro`, which
reexports the underlying custom derives. This ensures those macros
works without importing another crate.

The used mechanism is very similar to that what [serde](https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L283) does.

I propose this change for the following reasons:

- Makes it easier and less error prone to use the provided custom derives
- Allows us to include the documentation for the custom derives in diesels api documentation (currently suppressed by `#[doc(hidden)]`
- Makes it less error prone to use our proc macro 2.0 workaround. (There is now no way to use this macros without having the underlying custom derives imported.)

This pull request should show the general impact of such a change. Feel free to close it if this kind of change is not wanted.

Things to do:

- [x] Write changelog
- [x] See if all tests passes


.appveyor.yml
```diff
@@ -41,12 +41,14 @@ before_test:
   - mysql -e "create database diesel_test; create database diesel_unit_test;" -uroot
 
 test_script:
-  - cd diesel
+  - cd diesel_codegen
+  - cargo test
+  - cd ../diesel
   - cargo test --features "%backend% extras"
+  - cd ../diesel_proc_macro
```
sgrif: This needs to test `diesel_proc_macro_internal` as well doesn't it?

diesel_codegen/tests/associations.rs
```diff
@@ -1,11 +1,6 @@
 #![allow(dead_code, unused_must_use)]
 
-#[cfg(feature = "sqlite")]
 type Backend = ::diesel::sqlite::Sqlite;
-#[cfg(feature = "mysql")]
```
sgrif: I'd still prefer to test this against all backends. Even if we're not generating anything that directly hits the database, we should still be sure that we're generating code that is valid for all backends

diesel_proc_macro/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "diesel_proc_macro"
+version = "0.16.0"
+authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+#diesel_proc_macro_internal = "0.16.0"
```
sgrif: You can specify a version and a path

sgrif:
So I like this, but I think it's a bit odd now that `diesel_codegen` doesn't require `extern crate` but `diesel_proc_macro` does. I think I'd also like to split up our bang macros by behavior, not by the fact that they're a macro. What I think I'd like to do is this:

Move the `infer_schema!` and `infer_table_from_schema!` macro definitions into `diesel_infer_schema`. Move the derives for that into `infer_schema_internals` (note: `diesel` is explicitly left out of the name so it doesn't show up in searches). Then we also create a new `diesel_embed_migrations` crate (we can probably move this into its own repo TBH), and have it depend on a new `embed_migrations_internals` crate.

There's also a tradeoff here that the `infer_schema!` macro (and others) will no longer show up in the documentation for the main crate. We'll have to play with that to see what we think is best, but I don't think that should block this PR.

@diesel-rs/core Anyone else have thoughts on this?

sgrif:
> I believe that some people might read this change as us taking a step back from promoting infer_schema! as the way to develop Diesel apps.

I'm completely fine with people taking that interpretation. I think we're at the point where we can say that `diesel print-schema` is the canonical way to develop, and `infer_schema!` is more useful for prototyping and/or tests

weiznich:
@diesel-rs/core Any idea what is causing the appveyor failures?

diesel/Cargo.toml
```diff
@@ -33,13 +34,12 @@ bitflags = { version = "0.9", optional = true }
...
 [dev-dependencies]
 cfg-if = "0.1.0"
-diesel_codegen = "0.16.0"
 dotenv = ">=0.8, <0.11"
 quickcheck = "0.3.1"
 tempdir = "^0.3.4"
 
 [features]
-default = ["with-deprecated"]
+default = ["with-deprecated", "sqlite"]
```
killercup: Wait, why are we including sqlite by default now?

diesel/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["database"]
 
 [dependencies]
 byteorder = "1.0"
+diesel_derives = {version = "0.16.0", path = "../diesel_derives/" }
```
killercup: Remember to remove the `path` attr when this is released!

Also: Can you add a space after the opening brace to make it consistent with the other lines?

diesel/src/associations/mod.rs
```diff
@@ -356,7 +348,7 @@ impl<'a, T: HasTable> HasTable for &'a T {
 ///
 /// ### Deriving
 ///
-/// This trait can be automatically derived using `diesel_codegen` by adding
+/// This trait can be automatically derived using `diesel_derives` by adding
```
killercup: If this is automatically included by diesel now, we should remove the

> using `diesel_derives`

diesel/src/insertable.rs
```diff
@@ -13,7 +13,7 @@ use sqlite::Sqlite;
 ///
 /// ### Deriving
 ///
-/// This trait can be automatically derived using `diesel_codegen` by adding
+/// This trait can be automatically derived using `diesel_derives` by adding
```
killercup: If this is automatically included by diesel now, we should remove the

> using `diesel_derives`

same below in all other instances of this text

diesel/Cargo.toml
weiznich: Not really (It is only a artifact from testing…)

diesel/src/associations/mod.rs
weiznich: There is much more documentation wrong/incorrect/broken after this. I think it is more meaningful to first finish the actual implementation and fix all those documentation things afterwards.

weiznich:
I've hopefully fixed all now broken parts of the documentation and added a changelog entry. From my side this is ready to be reviewed and merged.

sgrif:
I hadn't even thought of bundling the `migrations` module with the `embed_migrations!` macro, but I really like that idea now that I've seen it.

diesel_migrations/migrations_internals/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "migrations_internals"
```
sgrif: I don't understand the need for this crate. Everything in here was previously part of the public API, and I don't see the reasoning for making it private. Why does this need to be separate from `diesel_migrations`?

sgrif:
@weiznich Given that this PR is getting quite large, can you give an updated summary on everything that's changed?

diesel_migrations/migrations_internals/Cargo.toml
weiznich: We need a extra crate here because some parts of this are used [inside of migrations_macros](https://github.com/diesel-rs/diesel/pull/1231/files#diff-d004f1c44c15da1ffeae40bf38f49fafR12). Everything that was public before inside the migrations module remains public, because it is [reexported](https://github.com/diesel-rs/diesel/pull/1231/files#diff-27119e6fe3360c1e2e3dd6a25aa5e811R19) through diesel_migrations

weiznich:
> @weiznich Given that this PR is getting quite large, can you give an updated summary on everything that's changed?

Basically the following things are changed:

* Remove `diesel_codegen`
* Remove `diesel::migrations`
* Remove(Or better rename) the old `diesel_infer_schema` crate

* Add `diesel_derive`, containing all custom derives (`Queryable`, `QueryableByName`, `Identifiable`, `Insertable`, `AsChangeset` and `Associated`)
* Add `diesel_infer_schema` containing all `infer_schema!` related code
* Add `diesel_migrations` containing all migration related code

weiznich:
@diesel-rs/core What is missing to get this merged?

.appveyor.yml
```diff
@@ -42,20 +42,33 @@ before_test:
...
+  - cd ../../diesel_infer_schema/infer_schema_macros
+  - cargo test --no-default-features --features "%backend% dotenv"
+  - cd ../../diesel_infer_schema
+  - cargo test --no-default-features --features "%backend% dotenv_macro"
+  - cd ../diesel_migrations/migrations_internals
+  - cargo test
+  - cd ../../diesel_migrations/migrations_macros
+  - cargo test
+  - cd ../../diesel_migrations/
+  - cargo test --no-default-features --features "%backend%"
```
killercup: Super unimportant and probably better in a separate PR: Instead of two lines per crate, you could also try to call this as `cargo test --no-default-features --features "%backend% extras" --manifest-path=diesel/Cargo.toml` (same is true for travis and this could make the files look more alike)

CHANGELOG.md
```diff
@@ -57,8 +57,19 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
...
+ derive procedural macro implementations (`Queryable`, `QueryableByName`,
+ `Identifiable`, `Insertable`, `AsChangeset` and `Associated`) are now
+ provided by `diesel_derives`. Furthermore those annotations are reexported
+ by `diesel` itself, so there is no need to depend on a second crate for
+ custom derives anymore. The `infer_schema!` and `infer_table_from_schema!`
+ macro is moved from diesel into the `diesel_infer_schema` crate. All
+ functionality related to migrations (The `diesel::migrations` module and the
+ `embed_migrations!` macro) is moved to the `diesel_migrations` crate. If you
+ are using any of those, simply depend on the new crates and your code will
+ continue to build.
```
killercup: This is quite detailed and well written. I fear people will not read it, though… Can you add a small example of the change needed here?

> This means, if you were using `diesel_codegen` before, you can now replace this dependency
> with `diesel_infer_schema` (or remove it completely if you're not using `infer_schema!` in your code):
> 
> ```diff
> [dependencies]
> diesel = { version = "0.10.1", features = ["sqlite"] }
> -diesel_codegen = { version = "0.16", features = ["sqlite"] }
> +diesel_infer_schema = { version = "0.16", features = ["sqlite"] }
> ``` 
>
> and change the imports to
>
> ```rust
> #[macro_use] extern crate diesel;
> #[macro_use] extern crate diesel_infer_schema;
> ```
>
> (again, the second line is only necessary of you're using `infer_schema!`).
> In general, a lot of project will now only need to add diesel itself as a dependency.

diesel_derives/README.md
```diff
@@ -0,0 +1,15 @@
+# Diesel Codegen
```
killercup: Diesel Derives?

examples/postgres/all_about_updates/src/lib.rs
```diff
@@ -44,7 +42,7 @@ fn examine_sql_from_publish_all_posts() {
 
     assert_eq!(
         "UPDATE \"posts\" SET \"draft\" = $1 -- binds: [false]",
-        debug_query(&diesel::update(posts).set(draft.eq(false))).to_string()
+        debug_query::<Pg, _>(&diesel::update(posts).set(draft.eq(false))).to_string()
```
killercup: Hm, why is this annotation necessary here _now_?

CHANGELOG.md
weiznich: This is not that simple :wink: 
If someone is using `infer_schema!` he needs to import `diesel_infer_schema` now.
If someone is using `embed_migrations!` he needs to import `diesel_migrations` now.

PR #1233, opened 2017-10-03
Allow tables to be renamed

### LINKED ISSUE #1217: Provide a way to rename tables in table! macro
Database table names, like column names, can clash with Rust keywords and identifiers, and the library should provide a way to rename them in defining a schema.

#1084 provides a way to rename columns but not tables. The same pseudo-attribute format could be used for renaming tables.

See also #424 #967 etc.

Eijebong:
The renaming is done via a `sql_name` attribute like columns renaming.

Fixes #1217


sgrif:
Agreed that we should add a custom type to the test

PR #1234, opened 2017-10-04
Add rustfmt info to contributing guide

katrinabrock:
Pretty self explanatory.

CONTRIBUTING.md could probably use some reorganization once some of the `draft_guides` are deployed to the website. Maybe even formalize it into a contributor guide that itself is deployed to the website?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -130,15 +130,26 @@ To run rustfmt tests locally:
    ```
 
 3. Run rustfmt using cargo from the root of your diesel repo.
+   
+   To see changes that need to be made, run
 
    ```
    cargo fmt --all -- --write-mode=diff
    ```
 
    If all code is properly formatted (e.g. if you have not made any changes),
-   this should run without error.
+   this should run without error or output.
    If your code needs to be reformatted,
    you will see a diff between your code and properly formatted code.
+   If you see code here that you didn't make any changes to
+   then you are probably running the wrong version of rustfmt.
+   Once you are ready to apply the formatting changes, run 
+
+   ```
+   cargo fmt --all
+   ```
+
+   You won't see any output, but all your files will be corrected.
 
 You can also use rustfmt to make corrections or highlight issues in your editor.
 Check out [their README](https://github.com/rust-lang-nursery/rustfmt) for details.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,7 @@ Thank you! We'll try to respond as quickly as possible.
 
 ## Contribute code to Diesel
 
+
 ### Setting up Diesel locally
 
 1. Install Rust using [rustup], which allows you to easily switch between Rust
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -108,6 +112,37 @@ Thank you! We'll try to respond as quickly as possible.
...
+3. Run rustfmt using cargo from the root of your diesel repo.
+
+   ```
+   cargo fmt --all -- --write-mode=diff
+   ```
+
+   If all code is properly formatted (e.g. if you have not made any changes),
+   this should run without error.
+   If your code needs to be reformatted,
+   you will see a diff between your code and properly formatted code.
```
killercup: I'd tell people to just run `cargo fmt --all`, which will rewrite their code to be in Good Style. No need to read diffs :)

katrinabrock: My concern with that is it might be a little too magical / hidden. Maybe there's a `-v` option?

killercup: Good point, that may throw people off when they are new to rustfmt and don't trust it yet. On the other hand… git diff before/after?

I only mention this because I have never seen anyone use rustfmt's diff thing outside of CI, but it is a valid feature :)

killercup:
> Maybe even formalize it into a contributor guide that itself is deployed to the website?

I'd like to have a "Diesel internals" guide that points contributors to where some common things live and how Diesel itself is organized. This will be hard to keep up-to-date, though. I'm not sure a meta-guide on how our issue tracker is organized or how to submit PRs will be good content for a full-blown guide, as it's more specific to the Github than the project itself.

CONTRIBUTING.md
notryanb: @killercup I've used `write-mode=diff` recently because the local rustfmt version I have is newer than CI and caused my PR to fail after formatting. I suppose it would show up in a `git diff` anyway, but I did find it helpful.

katrinabrock: Is there a way to just automate this, ie have travis just run rustfmt and then add a commit fixing everything?

sgrif: Not that I'm aware of but if you know of one please do open a PR

CONTRIBUTING.md
```diff
@@ -50,6 +50,7 @@ Thank you! We'll try to respond as quickly as possible.
 
 ## Contribute code to Diesel
 
+
```
sgrif: ✂️

CONTRIBUTING.md
```diff
@@ -60,7 +61,10 @@ Thank you! We'll try to respond as quickly as possible.
    continuous integration. Find it by looking for a line like
    `rust: nightly-2017-06-06` in the `.travis.yml` file. You can install and
    set a custom nightly version for a project using
-   `rustup override add nightly-2017-06-06`.
+
+   ```
+   rustup override add nightly-2017-06-06
```
sgrif: Should we go ahead and change this to the currently used date?

PR #1235, opened 2017-10-04
address the FIXME in assocations module docs

notryanb:
Code examples in this portion of the docs were untested.
Add doctests for each of the five ignored examples
and use relative urls for links.

### DIFF
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -1,4 +1,3 @@
-// FIXME: Replace these examples with executable tests
 //! Traits related to relationships between multiple tables.
 //!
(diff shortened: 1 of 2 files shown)


diesel/src/associations/mod.rs
```diff
@@ -63,23 +67,99 @@
...
+//! #   use users::dsl::*;
+//! #   use posts::dsl::*;
+//! #   let connection = establish_connection();
+//! #
+//! let data: Vec<(User, Post)> = users.inner_join(posts)
+//!     .load(&connection)
+//!     .expect("Couldn't load query");
 //!
+//! assert_eq!(data,
+//! vec![
```
sgrif: Can we move this to a local variable?

PR #1243, opened 2017-10-08
Always implement `AsExpression` for double references

### LINKED ISSUE #1242: Error deriving Insertable for a SQLite table with a nullable timestamp column
### Versions

- **Rust:** 1.20.0 (stable)
- **Diesel:** 0.16.0
- **Database:** SQLite 3.7.17
- **Operating System** CentOS 7.4

### Feature Flags

- **diesel:** sqlite chrono
- **diesel_codegen:** sqlite

I have a table defined as follows:

```sql
create table books (
    id integer not null primary key,
    title varchar null,
    save_date timestamp null 
)
```
I'm trying to derive an `Insertable` implementation for it following roughly the *Getting Started* guide, using `infer_schema!` and the following struct:

```rust
	#[derive(Insertable)]
	#[table_name="books"]
	pub struct NewBook<'a> {
	    pub title : Option<&'a str>,
	    pub save_date : Option<&'a chrono::NaiveDateTime>, 
	}
```

A successful compilation, as is the case with diesel 0.14

A series of error messages complaining for unsatisfied trait bounds, such as these:

    error[E0277]: the trait bound `chrono::NaiveDateTime: diesel::Expression` is not satisfied
      --> src/main.rs:30:11
       |
    30 |    #[derive(Insertable)]
       |             ^^^^^^^^^^ the trait `diesel::Expression` is not implemented for `chrono::NaiveDateTime`
       |
       = note: required because of the requirements on the impl of `diesel::Expression` for `&'a chrono::NaiveDateTime`
    
    error[E0277]: the trait bound `&'insert domain::NewBook<'a>: diesel::Insertable<schema::__diesel_infer_schema::infer_books::books::table, DB>` is not satisfied
      --> src/main.rs:30:11
       |
    30 |    #[derive(Insertable)]
       |             ^^^^^^^^^^ the trait `diesel::Insertable<schema::__diesel_infer_schema::infer_books::books::table, DB>` is not implemented for `&'insert domain::NewBook<'a>`
       |
                 <&'insert domain::NewBook<'a> as diesel::Insertable<schema::__diesel_infer_schema::infer_books::books::table, DB>>
       = note: required by `diesel::Insertable`

The full list of errors can be found in [this gist](https://gist.github.com/nmusatti/89d541ec7582ed32a36ab2185fd7b483).

### Steps to reproduce

[This project](https://github.com/nmusatti/diesel_timestamp_error) exemplifies the problem; uncomment the sections that are commented out and build. If you change the `diesel` and `diesel_codegen` version to 0.14 and `dotenv`'s to 0.9 you'll see that it compiles and runs without problems.

## Checklist

sgrif:
The only time that a double reference `AsExpression` impl comes up is
from the derived impl of `Insertable` and `AsChangeset`. We can't vary
the generated code based on whether the type of a field is `Copy`, so we
have to assume that it isn't and always take a reference.

Previously we only did this for `str` and slices, as `String` and `Vec`
are the only non-copy types that we support (I think `BigDecimal` may
fall into that category now, but it's unlikely that avoiding a single
clone of that type is ever relevant).

Still, there's no reason that we shouldn't allow references of any type
to appear in `Insertable` structs. To avoid code bloat, I've cleaned
things up a bit and implemented more stuff on `str` and `[T]` directly,
letting the blanket reference impl apply to those types when it's
relevant.

Fixes #1242.

### DIFF
--- a/diesel/src/macros/insertable.rs
+++ b/diesel/src/macros/insertable.rs
@@ -467,6 +467,40 @@ mod tests {
         assert_eq!(Ok(expected), saved);
     }
 
+    #[test]
+    fn named_struct_with_unusual_reference_type() {
+        struct NewUser<'a> {
+            name: &'a String,
+            hair_color: Option<&'a String>,
+        }
+
+        impl_Insertable! {
+            (users)
+            struct NewUser<'a> {
+                name: &'a String,
+                hair_color: Option<&'a String>,
+            }
+        }
+
+        let conn = connection();
+        let sean = "Sean".to_string();
+        let black = "Black".to_string();
+        let new_user = NewUser {
+            name: &sean,
+            hair_color: Some(&black),
+        };
+        ::insert_into(users::table)
+            .values(&new_user)
+            .execute(&conn)
+            .unwrap();
+
+        let saved = users::table
+            .select((users::name, users::hair_color))
+            .load(&conn);
+        let expected = vec![(sean.clone(), Some(black.clone()))];
+        assert_eq!(Ok(expected), saved);
+    }
+
     cfg_if! {
         if #[cfg(feature = "sqlite")] {
             fn connection() -> ::test_helpers::TestConnection {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * MySQL URLs will now properly percent decode the username and password.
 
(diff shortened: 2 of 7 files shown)


diesel/src/macros/insertable.rs
```diff
@@ -344,6 +344,14 @@ mod tests {
         }
     }
 
+    test_struct_definition! {
```
weiznich: There is no impl for &String: From<&str>, so this needs some more explicit conversion.

PR #1247, opened 2017-10-08
Fix build warnings

bippityboppity:
A few warnings that kept coming up when building / testing locally. Thought I'd fix them as they were easy enough to resolve.


sgrif:
These may not be used by SQLite, but they are used by the other backends as you can see from the build failures.

PR #1253, opened 2017-10-10
Fix our SQLite chrono implementation

sgrif:
The previous implementation would only handle the specific output format
of the `datetime` function. However, it is incorrect for us to assume
that we would get that specific function, as times may not have been
created by that function (which is why the range of input it accepts is
wider than the output it can give).

While values that were stored by using Diesel with chrono will be in the
expected format, we need to work with databases that have data from
other sources. The value in the column will just be whatever string was
given.

This implementation is updated to take every value that is defined as
valid input to the `strftime` function in SQLite, with the exception of
`'now'`. I've updated the tests to exercise all possible values. I've
also changed/removed tests which were testing for values that SQLite
does not handle. It documents its range as being from midnight of Jan 1
0000 to Dec 31 9999 23:59:59.999999. I explicitly opted not to test the
behavior of out of range values, as SQLite defines it as undefined.

### DIFF
--- a/diesel/src/sqlite/types/date_and_time/chrono.rs
+++ b/diesel/src/sqlite/types/date_and_time/chrono.rs
@@ -2,45 +2,18 @@ extern crate chrono;
 
 use std::error::Error;
 use std::io::Write;
-use self::chrono::{Datelike, NaiveDate, NaiveDateTime, NaiveTime};
+use self::chrono::{NaiveDate, NaiveDateTime, NaiveTime};
 
 use sqlite::Sqlite;
 use sqlite::connection::SqliteValue;
 use types::{Date, FromSql, IsNull, Text, Time, Timestamp, ToSql, ToSqlOutput};
 
-const SQLITE_TIME_FMT: &'static str = "%0H:%0M:%0S%.6f";
-
-fn parse_sqlite_date(date: &str) -> Result<NaiveDate, Box<Error + Send + Sync>> {
-    let negative_year = date.starts_with('-');
-
-    let ymd: Vec<u32> = date.split('-').flat_map(str::parse).collect();
-    if ymd.len() != 3 {
-        return Err(
-            "Cannot parse date, too many parts. The date should be in the form: YYYY-MM-DD".into(),
-        );
-    }
-
-    if negative_year {
-        return Ok(NaiveDate::from_ymd(-(ymd[0] as i32), ymd[1], ymd[2]));
-    }
-
-    NaiveDate::from_ymd_opt(ymd[0] as i32, ymd[1], ymd[2])
-        .ok_or_else(|| format!("Cannot parse this date: {:?}", ymd).into())
-}
-
-fn dump_sqlite_date<T: Datelike>(date: &T) -> String {
-    format!(
-        "{:0>4}-{:0>2}-{:0>2}",
-        date.year(),
-        date.month(),
-        date.day()
-    )
-}
+const SQLITE_DATE_FORMAT: &'static str = "%F";
 
 impl FromSql<Date, Sqlite> for NaiveDate {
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -174,11 +219,40 @@ mod tests {
...
+            "00:00:00.000000",
+            "00:00Z",
+            "00:00:00Z",
+            "00:00:00.000Z",
+            "00:00:00.000000Z",
+            "00:00+00:00",
+            "00:00:00+00:00",
+            "00:00:00.000+00:00",
+            "00:00:00.000000+00:00",
+            "01:00+01:00",
```
weiznich: This tests [fails](https://travis-ci.org/diesel-rs/diesel/jobs/286125994#L634) on CI

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -127,42 +128,86 @@ mod tests {
     fn unix_epoch_encodes_correctly() {
```
weiznich: Do we really need this test twice? The same value is also tested by the next test.

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -174,11 +219,40 @@ mod tests {
...
+            let query = select(time(format));
+            assert_eq!(
+                Ok(midnight),
+                query.get_result::<NaiveTime>(&connection),
+                "format {} failed",
+                format
+            );
+        }
 
         let noon = NaiveTime::from_hms(12, 0, 0);
```
weiznich: Why are this two values not in the array above, but rather tested separated?

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -67,24 +40,50 @@ impl ToSql<Time, Sqlite> for NaiveTime {
...
+
+        for format in sqlite_datetime_formats {
+            if let Ok(dt) = Self::parse_from_str(text, format) {
+                return Ok(dt);
+            }
         }
 
-        let time = NaiveTime::parse_from_str(dt[1], SQLITE_TIME_FMT)?;
-        let datetime = parse_sqlite_date(dt[0])?.and_time(time);
+        if let Ok(julian_days) = text.parse::<f64>() {
```
quodlibetor: Technically I believe these are all Gregorian not Julian. Chrono certainly doesn't actually deal with the Julian calendar, and sqlite [documents itself](https://sqlite.org/datatype3.html#date_and_time_datatype) as being gregorian.

I really doubt many people would be confused by the distinction unless they have to deal with Julian calendars, which is probably just historians?

diesel/src/sqlite/types/date_and_time/chrono.rs
sgrif: This is testing serialization. The next test is testing deserialization.

diesel/src/sqlite/types/date_and_time/chrono.rs
sgrif: Because noon is not midnight.

diesel/src/sqlite/types/date_and_time/chrono.rs
sgrif: Chrono doesn't deal with julian days, but a float representing julian days is one of the documented valid date formats at https://sqlite.org/lang_datefunc.html. It's quite trivial to convert julian to gregorian, which is what this is doing.

quodlibetor: Oh no. You're obviously right, I should have read the whole function before I got confused by the variables.

PR #1259, opened 2017-10-13
Make raw connection handles available

jethrogb:
Sometimes, you need to access specialized DB functionality that is not available through Diesel directly. This PR makes the raw connection handles available for that purpose.

This is a pretty common practice:
* https://doc.rust-lang.org/std/os/unix/io/trait.AsRawFd.html
* https://docs.rs/native-tls/0.1.4/native_tls/backend/openssl/trait.TlsStreamExt.html
* https://docs.rs/curl/0.4.8/curl/easy/struct.Easy2.html#method.raw
* etc.

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -153,3 +153,12 @@ pub trait Connection: SimpleConnection + Sized + Send {
     #[doc(hidden)]
     fn transaction_manager(&self) -> &Self::TransactionManager;
 }
+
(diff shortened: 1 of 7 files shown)


jethrogb:
I don't think the build failures are related to this PR.

sgrif:
I'm not interested in exposing this. The fact that we happen to use libpq is an implementation detail, and one that will change between minor versions. SQLite is the only backend likely to continue using the C library for the foreseeable future

PR #1260, opened 2017-10-16
Ensure PG results always have an error message

### LINKED ISSUE #1255: Panic while formatting an error when Pg connection is reset
## Problem Description

When a Postgres database connection is closed unexpectedly, trying to debug-format the error returned by Diesel panics. Here's a simple test program (using a table named `config`):

```rust
fn main() {
    loop {
        println!("{:?}", config::table.count().first::<i64>(&connection));
        std::thread::sleep(std::time::Duration::from_secs(1));
    }
}
```

Have the DB server running before starting the program. After you see an `Ok(...)` line, forcefully terminate the DB server, for example using `kill -9`. I'm getting the following panic:

```
thread 'main' panicked at 'internal error: entered unreachable code: Per PGs documentation, all errors should have a message', .cargo/registry/src/github.com-1ecc6299db9ec823/diesel-0.16.0/src/pg/connection/result.rs:97:20
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:380
             at /checkout/src/libstd/panicking.rs:396
             at /checkout/src/libstd/panicking.rs:611
   5: std::panicking::begin_panic_new
             at /checkout/src/libstd/panicking.rs:553
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:521
   7: <diesel::pg::connection::result::PgErrorInformation as diesel::result::DatabaseErrorInformation>::message
             at .cargo/registry/src/github.com-1ecc6299db9ec823/diesel-0.16.0/src/pg/connection/result.rs:97
   8: <diesel::result::DatabaseErrorInformation + core::marker::Sync + core::marker::Send + 'static as core::fmt::Debug>::fmt
             at .cargo/registry/src/github.com-1ecc6299db9ec823/diesel-0.16.0/src/result.rs:52
   9: <alloc::boxed::Box<T> as core::fmt::Debug>::fmt
             at /checkout/src/liballoc/boxed.rs:534
  10: <&'a T as core::fmt::Debug>::fmt
             at /checkout/src/libcore/fmt/mod.rs:1482
  11: core::fmt::write
             at /checkout/src/libcore/fmt/mod.rs:951
  12: core::fmt::builders::DebugTuple::field
             at /checkout/src/libcore/fmt/builders.rs:0
  13: <diesel::result::Error as core::fmt::Debug>::fmt
             at .cargo/registry/src/github.com-1ecc6299db9ec823/diesel-0.16.0/src/result.rs:6
  14: <&'a T as core::fmt::Debug>::fmt
             at /checkout/src/libcore/fmt/mod.rs:1482
  15: core::fmt::write
             at /checkout/src/libcore/fmt/mod.rs:951
  16: core::fmt::builders::DebugTuple::field
             at /checkout/src/libcore/fmt/builders.rs:0
  17: <core::result::Result<T, E> as core::fmt::Debug>::fmt
             at /checkout/src/libcore/result.rs:250
  18: core::fmt::write
             at /checkout/src/libcore/fmt/mod.rs:951
  19: <std::io::stdio::Stdout as std::io::Write>::write_fmt
             at /checkout/src/libstd/io/mod.rs:1117
             at /checkout/src/libstd/io/stdio.rs:461
  20: std::io::stdio::_print
             at /checkout/src/libstd/io/stdio.rs:679
             at /checkout/src/libstd/io/stdio.rs:701
  21: dtest::main
             at src/main.rs:27
  22: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  23: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:458
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:59
  24: main
  25: __libc_start_main
  26: _start
```

### Versions

- **Rust:** 1.20
- **Diesel:** 0.16.0
- **Database:** CockroachDB 1.0
- **Operating System** Ubuntu 16.04
- **libpq**: 9.5.8-0ubuntu0.16.04.1

### Feature Flags

- **diesel:** default
- **diesel_codegen:** default

sgrif:
In certain cases (for example, if the database connection unexpectedly
terminated), we will have a `PGResult` pointer which is not null, but
does not have an error response from the server (and therefore no result
error fields). In this case, PG will copy the connection's error message
onto the result. However, this needs to be retrieved with a separate
function, as it is not part of an error response from the server.

I'd like to explore if there are any other cases that can result in
this, and potentially look at providing a specific `DatabaseErrorKind`
variant for this case.

Fixes #1255.

### DIFF
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -132,6 +132,12 @@ impl RawResult {
     pub fn as_ptr(&self) -> *mut PGresult {
         self.0
     }
+
+    pub fn error_message(&self) -> &str {
+        let ptr = unsafe { PQresultErrorMessage(self.0) };
+        let cstr = unsafe { CStr::from_ptr(ptr) };
+        cstr.to_str().unwrap_or_default()
+    }
 }
 
 impl Drop for RawResult {
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -97,10 +97,8 @@ struct PgErrorInformation(RawResult);
 
 impl DatabaseErrorInformation for PgErrorInformation {
     fn message(&self) -> &str {
-        match get_result_field(self.0.as_ptr(), ResultField::MessagePrimary) {
-            Some(e) => e,
-            None => unreachable!("Per PGs documentation, all errors should have a message"),
-        }
+        get_result_field(self.0.as_ptr(), ResultField::MessagePrimary)
+            .unwrap_or_else(|| self.0.error_message())
     }
 
     fn details(&self) -> Option<&str> {


diesel/src/pg/connection/raw.rs
```diff
@@ -132,6 +132,12 @@ impl RawResult {
     pub fn as_ptr(&self) -> *mut PGresult {
         self.0
     }
+
+    pub fn error_message(&self) -> &str {
+        let ptr = unsafe { PQresultErrorMessage(self.0) };
```
weiznich: This pointer is never null in any case? Otherwise it may be better to add a additional check for this. I think it may be better to add the additional check for null here because in theory this could be null…

sgrif: The pointer is never null in any case here

PR #1264, opened 2017-10-20
ARRAY[] function

### LINKED ISSUE #1263: Support for ARRAY[]
We currently lack support for the `ARRAY[]` constructor in Postgres, and it'd be nice to have so we can use array arguments to functions.

(Context: me trying to work out what I want and @sgrif finally working out what I meant happened in chat here: https://gitter.im/diesel-rs/diesel?at=59e8bb40b20c642429219b6f)

kivikakk:
Fixes #1263.

TODO:

- [x] Proper testing.
- [x] Put things where they belong. It should be somewhere pg-specific.
- [x] Improve the interface. We shouldn't have to use different names for different argument counts.


kivikakk:
https://github.com/kivikakk/diesel/blob/a0fce398cfc2338564e6c7fa9225d7e92ec38489/diesel_tests/tests/expressions/mod.rs#L364

This is an eyesore, but I'm not sure how to convince the compiler to infer `r` as equal(ish?!) to the `$arg_struct_name` idents …

kivikakk:
I'd be curious on anyone's comments before I proceed with trying to clean this up!

sgrif:
I don't like this approach. I'd much rather we provide a single function called `Array` which takes a tuple.

kivikakk:
I absolutely agree! Hence asking for comments. :) I'll work on that shortly.

sgrif:
So I'd roughly take this approach:

```rust
// Not super stoked on this name
pub trait IntoSingleTypeExpressionList<ST> {
    type Expression;

    fn into_single_type_expression_list(self) -> Self::Expression;
}

// This would go in types/impls/tuples.rs
impl<$($T,)+ ST> IntoSingleTypeExpressionList<ST> for ($($T,)+)
where
    $($T: AsExpression<ST>,)+
{
    type Output = ($(AsExprOf<$T, ST>,)+);

    fn into_single_type_expression_list(self) -> Self::Expression {
        ($(self.$idx.into_sql::<ST>(),)+)
    }
}

pub struct Array<T, ST> {
    elements: T,
    _marker: PhantomData<ST>,
}

// Order here is important, it's likely people will sometimes need to specify the sql type, but infer the actual argument being passed.
pub fn array<ST, T>(elements: T) -> Array<T::Expression, ST>
where
    T: IntoSingleTypeExpressionList<ST>,
{
    Array {
        elements: elements.into_single_type_expression_list(),
        _marker: PhantomData,
    }
}

impl<T, ST> Expression for Array<T, ST>
where
    T: Expression,
{
    type SqlType = types::Array<ST>;
}
```

It's worth noting that a single element array literal would look a little funky, as it'd be passed like: `array((foo,))`.

kivikakk:
This is now looking better. Thanks for your impeccable guidance there; it slotted right in with few changes.

Does the current positioning of the traits and structs for supporting this look OK?

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,64 @@
+use std::marker::PhantomData;
+use expression::{AppearsOnTable, Expression, IntoSingleTypeExpressionList, SelectableExpression};
+
+#[derive(Debug)]
+pub struct Array<T, ST> {
+    elements: T,
+    _marker: PhantomData<ST>,
+}
+
+pub fn array<ST, T>(elements: T) -> Array<T::Expression, ST>
```
weiznich: It would be great to add some small documentation with an example to this function.

kivikakk: Good point, thank you! Added in bb35442.

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,103 @@
...
+/// # include!("../../doctest_setup.rs");
+/// #
+/// # table! {
+/// #     users {
+/// #         id -> Integer,
+/// #         name -> VarChar,
+/// #     }
+/// # }
+/// #
+/// # #[cfg(feature = "postgres")]
```
weiznich: Because this lives in `diesel/src/pg` I think we  do not need the `#[cfg(feature = "postgres")]` annotation.

(Same for the empty main function for the not-pg case)

diesel/src/expression/mod.rs
```diff
@@ -300,3 +300,9 @@ impl<QS, ST, DB> QueryId for BoxableExpression<QS, DB, SqlType = ST> {
 
     const HAS_STATIC_QUERY_ID: bool = false;
 }
+
+pub trait IntoSingleTypeExpressionList<ST> {
```
weiznich: I think this trait should also live in `diesel/src/pg/expression/array.rs` because it is Postgres specific.

diesel/src/pg/expression/array.rs
kivikakk: :+1:

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,100 @@
...
+        elements: elements.into_single_type_expression_list(),
+        _marker: PhantomData,
+    }
+}
+
+impl<T, ST> Expression for Array<T, ST>
+where
+    T: Expression,
+{
+    type SqlType = ::pg::types::sql_types::Array<ST>;
```
sgrif: This can just be `types::Array`

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,100 @@
...
+use backend::Backend;
+
+impl<T, ST, DB> QueryFragment<DB> for Array<T, ST>
+where
+    DB: Backend,
+    for<'a> (&'a T): QueryFragment<DB>,
+{
+    fn walk_ast(&self, mut out: AstPass<DB>) -> ::result::QueryResult<()> {
+        out.push_sql("ARRAY[");
+        QueryFragment::walk_ast(&(&self.elements,), out.reborrow())?;
```
sgrif: Why the tuple?

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<T, ST, QS> AppearsOnTable<QS> for Array<T, ST>
+where
+    T: AppearsOnTable<QS>,
+    Array<T, ST>: Expression,
+{
+}
+
+impl<T, ST> ::expression::NonAggregate for Array<T, ST>
```
sgrif: Can we import this?

diesel/src/pg/expression/mod.rs
```diff
@@ -1,5 +1,7 @@
 #[doc(hidden)]
 pub mod array_comparison;
+#[doc(hidden)]
+pub mod array;
```
sgrif: Does this need to be public?

diesel/src/expression/mod.rs
kivikakk: 👍

diesel/src/pg/expression/array.rs
kivikakk: Not in this context? `types` isn't imported here.

diesel/src/pg/expression/array.rs
kivikakk: Yep.

diesel/src/pg/expression/array.rs
kivikakk: Good question! Dropping it.

diesel/src/pg/expression/mod.rs
kivikakk: Now that `IntoSingleTypeExpressionList` is in that module, yes.

sgrif:
This needs some compile-fail tests as well, probably checking that: only usable with pg, expressions have to be the right type, expressions have to have the same type

kivikakk:
Good call! Done.

diesel/src/pg/expression/array.rs
sgrif: Or `::types::Array`. Is there a reason for us to not import it?

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl<T, ST> Expression for Array<T, ST>
+where
+    T: Expression,
+{
+    type SqlType = ::pg::types::sql_types::Array<ST>;
+}
+
+use query_builder::{AstPass, QueryFragment};
+use backend::Backend;
```
sgrif: Can we keep all the `use` statements at the top?

kivikakk: Yep, done.

diesel/src/pg/expression/array.rs
kivikakk: Yeah, good point. Fixed.

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+use types;
+
+pub trait IntoSingleTypeExpressionList<ST> {
+    type Expression;
+
+    fn into_single_type_expression_list(self) -> Self::Expression;
+}
+
+#[derive(Debug)]
+pub struct Array<T, ST> {
```
sgrif: What do you think about naming this `ArrayLiteral` so it's clear what the distinction from `types::Array` is?

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+use query_builder::{AstPass, QueryFragment};
+use types;
+
+pub trait IntoSingleTypeExpressionList<ST> {
+    type Expression;
+
+    fn into_single_type_expression_list(self) -> Self::Expression;
+}
+
+#[derive(Debug)]
```
sgrif: This should implement `Clone` and `Copy`.

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+pub struct Array<T, ST> {
+    elements: T,
+    _marker: PhantomData<ST>,
+}
+
+/// Creates an `ARRAY[...]` expression.  The argument should be a tuple of
```
sgrif: The first sentence should be its own paragraph.

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+/// #     use diesel::*;
+/// #     use diesel::dsl::*;
+/// #     use users::dsl::*;
+/// #     let connection = establish_connection();
+/// let ints = select(array::<types::Int4, _>((1, 2)))
+///     .get_result::<Vec<i32>>(&connection);
+/// // An array is returned as a Vec.
+/// assert_eq!(Ok(vec![1, 2]), ints);
+///
+/// // The type of `id` is known so we don't have to specify the SQL type here.
```
sgrif: This doc comment mentions type inference more than anything else. Do you think this is likely to be used in contexts where the type can't be inferred enough to warrant this much attention?

diesel/src/pg/expression/array.rs
```diff
@@ -0,0 +1,106 @@
...
+{
+    fn walk_ast(&self, mut out: AstPass<DB>) -> ::result::QueryResult<()> {
+        out.push_sql("ARRAY[");
+        QueryFragment::walk_ast(&&self.elements, out.reborrow())?;
+        out.push_sql("]");
+        Ok(())
+    }
+}
+
+impl_query_id!(Array<T, ST>);
```
sgrif: Note: If #1391 is merged before this, this needs to get updated.

diesel/src/pg/expression/mod.rs
sgrif: Can we have a `#[doc(hidden)]` reexport of this from somewhere public and make the module private?

diesel/src/types/impls/tuples.rs
```diff
@@ -250,6 +250,18 @@ macro_rules! tuple_impls {
                     ($($T,)+ next)
                 }
             }
+
+            #[cfg(feature = "postgres")]
+            impl<$($T,)+ ST> $crate::pg::expression::array::IntoSingleTypeExpressionList<ST> for ($($T,)+) where
```
sgrif: There's no need to use fully qualified paths vs just `use`ing the trait here. This macro doesn't get called from user code.

diesel_compile_tests/tests/compile-fail/array_expressions_must_be_correct_type.rs
```diff
@@ -0,0 +1,20 @@
...
+extern crate diesel;
+
+use diesel::*;
+use diesel::dsl::*;
+
+fn main() {
+    let connection = PgConnection::establish("").unwrap();
+    select(array((1, 3))).get_result::<Vec<i32>>(&connection);
+    select(array((1f64, 3f64))).get_result::<Vec<i32>>(&connection);
+    //~^ ERROR E0277
```
sgrif: Can we be specific on at least one of these? (Also why are there so many?)

diesel/src/pg/expression/array.rs
kivikakk: :+1:

diesel/src/pg/expression/array.rs
kivikakk: I noticed behaviour where it wasn't, but I think that was in a previous iteration. I'll drop the excess.

diesel/src/pg/expression/array.rs
kivikakk: Understood.

diesel/src/types/impls/tuples.rs
kivikakk: :+1:

diesel/src/pg/expression/array.rs
kivikakk: :+1:

diesel/src/pg/expression/array.rs
kivikakk: :+1:

diesel_compile_tests/tests/compile-fail/array_expressions_must_be_correct_type.rs
kivikakk: They're like this:

```
unexpected errors (from JSON output): [
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:5: 10:11: the trait bound `f64: diesel::SelectableExpression<()>` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:33: 10:43: the trait bound `f64: diesel::expression::NonAggregate` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:33: 10:43: the trait bound `f64: diesel::SelectableExpression<()>` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:33: 10:43: the trait bound `f64: diesel::query_builder::QueryId` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:12: 10:17: the trait bound `f64: diesel::Expression` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:33: 10:43: the trait bound `f64: diesel::query_builder::QueryFragment<_>` is not satisfied [E0277]"
    },
    Error {
        line_num: 10,
        kind: Some(
            Error
        ),
        msg: "10:12: 10:31: the trait bound `f64: diesel::Expression` is not satisfied [E0277]"
    }
]
```

I'll add some specifics.

sgrif: Ughhhhh actually add none of those. Seriously Rust. Just fucking say "`f64: AsExpression<Array<_>>` is not satisfied" these errors are ridiculous.

diesel/src/pg/expression/mod.rs
kivikakk: I'm not exactly sure what shape this should take, having tried a few ways. I've tried something in 99783c00, but it feels janky and wrong. Some documentation added too.

diesel_compile_tests/tests/compile-fail/array_expressions_must_be_correct_type.rs
kivikakk: Got it.

sgrif:
Thank you for all your hard work, this is an excellent PR! <3

sgrif:
I manually rebased, and made additional changes. This was merged in 8fb0429

PR #1266, opened 2017-10-21
Add an example to the LimitDsl::limit function

Eijebong:


diesel/src/query_dsl/limit_dsl.rs
```diff
@@ -16,6 +16,36 @@ where
 {
     type Output = <T::Query as LimitDsl>::Output;
 
+    /// Example:
```
killercup: Can you add a summary line and then use `# Example` as headline? This way it shows up in rustdoc as "Example:" in some places. Maybe something like

> Limit of rows to return

or just

> Set the `LIMIT` on the query

diesel/src/query_dsl/limit_dsl.rs
```diff
@@ -16,6 +16,38 @@ where
...
+    ///
+    /// # #[derive(Queryable, PartialEq, Debug)]
+    /// # struct User {
+    /// #     id: i32,
+    /// #     name: String,
+    /// # }
+    ///
+    /// # fn main() {
+    /// #     let connection = establish_connection();
+    /// let user1 = NewUser { name: "Sean".into() };
```
sgrif: This could probably use an ad-hoc insert more and be more clear, couldn't it?

```rust
insert_into(users::table)
    .values(vec![
        users::name.eq("Sean"),
        users::name.eq("Tess"),
        users::name.eq("Pascal"),
    ])
```

diesel/src/query_dsl/limit_dsl.rs
```diff
@@ -16,6 +16,38 @@ where
...
+    /// #         name -> VarChar,
+    /// #     }
+    /// # }
+    ///
+    /// # #[derive(Queryable, PartialEq, Debug)]
+    /// # struct User {
+    /// #     id: i32,
+    /// #     name: String,
+    /// # }
+    ///
```
sgrif: We don't need to render all this whitespace

PR #1268, opened 2017-10-22
Make query_dsl more discoverable

katrinabrock:
Previously in docs, query_dsl attributes were doced on the prelude page
this makes the mod public so it has its own doc page.

### DIFF
--- a/diesel/src/query_dsl/belonging_to_dsl.rs
+++ b/diesel/src/query_dsl/belonging_to_dsl.rs
@@ -1,3 +1,4 @@
+#![deny(missing_docs)]
 /// Constructs a query that finds record(s) based on directional association with other record(s).
 ///
 /// # Example
--- a/diesel/src/query_dsl/belonging_to_dsl.rs
+++ b/diesel/src/query_dsl/belonging_to_dsl.rs
@@ -1,5 +1,44 @@
+/// Constructs a query that finds record(s) based on directional association with other record(s).
+///
+/// # Example
+///
+/// ```rust
+/// # #[macro_use] extern crate diesel;
+/// # #[macro_use] extern crate diesel_codegen;
+/// # include!("../doctest_setup.rs");
+/// # use schema::{posts, users};
+/// #
+/// # #[derive(Identifiable, Queryable)]
+/// # pub struct User {
+/// #     id: i32,
+/// #     name: String,
+/// # }
+/// #
+/// # #[derive(Debug, PartialEq)]
+/// # #[derive(Identifiable, Queryable, Associations)]
+/// # #[belongs_to(User)]
+/// # pub struct Post {
+/// #     id: i32,
+/// #     user_id: i32,
+/// #     title: String,
+/// # }
+/// #
+/// # fn main() {
+/// # let connection = establish_connection();
+/// # use users::dsl::*;
+/// # let user = users.find(2).get_result::<User>(&connection).unwrap();
+/// let posts = Post::belonging_to(&user)
+/// #    .load::<Post>(&connection);
+/// #
+/// # assert_eq!(posts,
+/// #     Ok(vec![Post { id: 3, user_id: 2, title: "My first post too".to_owned() }])
+/// # );
+/// # }
+/// ```
 pub trait BelongingToDsl<T> {
+    /// The query returned by `belonging_to`
     type Output;
 
+    /// Get the record(s) belonging to record(s) `other`
     fn belonging_to(other: T) -> Self::Output;
 }
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -1,3 +1,12 @@
+//! Traits that construct SELECT statements
+//!
+//! Traits in this module have methods that generally map to the keyword for the corresponding clause in SQL,
+//! unless it conflicts with a Rust keyword (such as `WHERE`/`where`).
+//!
+//! See also [`expression_methods`][expression_methods] and [`dsl`][dsl].
+//!
+//! [expression_methods]: ../expression_methods/index.html
+//! [dsl]: ../dsl/index.html
 mod belonging_to_dsl;
 #[doc(hidden)]
 pub mod boxed_dsl;
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -68,7 +68,7 @@ pub mod pg;
 pub mod sqlite;
 
 pub mod migrations;
-mod query_dsl;
+pub mod query_dsl;
 pub mod query_source;
 pub mod result;
 pub mod row;
--- a/diesel/src/query_dsl/distinct_dsl.rs
+++ b/diesel/src/query_dsl/distinct_dsl.rs
@@ -1,3 +1,4 @@
+#![deny(missing_docs)]
 use query_source::Table;
 
 /// Adds the `DISTINCT` keyword to a query.
@@ -31,7 +32,10 @@ use query_source::Table;
 /// # }
 /// ```
 pub trait DistinctDsl {
+    /// Query with DISTINCT added
     type Output;
+
+    /// Should return query with DISTINCT added
     fn distinct(self) -> Self::Output;
 }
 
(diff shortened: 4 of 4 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -1,18 +1,13 @@
 mod belonging_to_dsl;
-#[doc(hidden)]
```
sgrif: These should all still be `#[doc(hidden)]`. I don't want to expose the internal structure of this module, just have it be public for documentation purposes.

katrinabrock:
4 character PR

sgrif:
Hah I figured it'd add some module level docs too but that can always be done separately

katrinabrock:
yeah, I'd like to do that but still don't know enough to know what to write.

sgrif:
Something vaguely along the lines of "methods which are used to construct select statements. Methods in this module will generally map to the keyword for the corresponding clause in SQL, unless it conflicts with a Rust keyword (such as `where`). Methods in this module are generally focused at the query level. For expression level methods, see `expression_methods` and `dsl`"

I don't like the terms "query level" and "expression level", but I'm struggling to come up with a better term to differentiate them. Also saying that it's for select statements is slightly misleading since update and delete statements also implement `FilterDsl`, but I think it's fine to just refer to these as methods around select statements, since `UpdateStatement` and `DeleteStatement` are concrete types that I'm fine with people relying on, but `SelectStatement` is internal

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,46 @@
...
+/// # Example
+///
+/// ```rust
+/// // You need to have the table definitions from `infer_schema!` or `table!` in scope to
+/// // derive Identifiable.
+/// # #[macro_use] extern crate diesel;
+/// # #[macro_use] extern crate diesel_codegen;
+/// # include!("../doctest_setup.rs");
+/// # use schema::{posts, users};
+///
```
sgrif: These whitespace lines will all be rendered if you don't have ` #` won't they?

diesel/src/query_dsl/mod.rs
```diff
@@ -1,3 +1,4 @@
+#[deny(missing_docs)]
```
sgrif: I think you meant `#!` here?

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,46 @@
...
+/// let posts = Post::belonging_to(&user)
+///     .load::<Post>(&connection);
+///
+/// assert_eq!(posts,
+///     Ok(vec![Post { id: 3, user_id: 2, title: "My first post too".to_owned() }])
+/// );
+/// # }
+/// ```
 pub trait BelongingToDsl<T> {
+    ///
```
sgrif: Did you mean to leave this blank?

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,46 @@
...
     type Output;
 
+    /// Get the record(s) belonging to record(s) `other`
```
sgrif: Do you think that this implies we're loading the data instead of returning a query that needs to be executed?

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,46 @@
+/// Loads record(s) by a directional association with (an)other record(s).
+///
+/// # Example
+///
+/// ```rust
+/// // You need to have the table definitions from `infer_schema!` or `table!` in scope to
+/// // derive Identifiable.
```
sgrif: Do we need to restate this here? Is it stated in the docs for `Identifiable`? Do you think this line is confusing without seeing the `use` lines?

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,46 @@
+/// Loads record(s) by a directional association with (an)other record(s).
```
sgrif: This trait doesn't actually load anything. Do you think this statement is misleading?

diesel/src/query_dsl/mod.rs
katrinabrock: that would explain it...

diesel/src/query_dsl/belonging_to_dsl.rs
katrinabrock: It's a placeholder, this push was mainly to debug my rendering issue.

diesel/src/query_dsl/belonging_to_dsl.rs
katrinabrock: This example is copy pasta from the associations mod. This didn't jump out at me due to rendering issue.

katrinabrock:
@sgrif *now* ready for review. Still needs examples, of course.

I don't think `query_dsl` is for query level and `expression_dsl` is for expression level really adds much, but it would be nice to have something about why methods might be in one place or another.

diesel/src/query_dsl/belonging_to_dsl.rs
```diff
@@ -1,5 +1,44 @@
+/// Constructs a query that finds record(s) based on directional association with other record(s).
```
killercup: > based on directional association with other record(s)

wow so many long words and generic singular/plural, it's almost a german sentence :O (I like it)

diesel/src/query_dsl/mod.rs
```diff
@@ -1,3 +1,12 @@
+//! Traits with methods to construct select statements
+//!
+//! Methods in this module generally map to the keyword for the corresponding clause in SQL,
+//! unless it conflicts with a Rust keyword (such as `WHERE`/`where`).
+//!
+//! See also [`expression_methods`][expression_methods] and [`dsl`][dsl].
+//!
+//! [expression_methods]: ../expression_methods/index.html
+//! [dsl]: ../dsl/index.html
```
killercup: I totally missed this being a `//!` upper-scope comment as there was no empty line between this and the next line with the `mod` on it

diesel/src/query_dsl/mod.rs
```diff
@@ -1,3 +1,12 @@
+//! Traits with methods to construct select statements
+//!
+//! Methods in this module generally map to the keyword for the corresponding clause in SQL,
```
killercup: > Methods in this module

I know what you mean, but modules don't have methods directly. Maybe

> The traits in this module have methods that generally map to the keyword for the corresponding clause in SQL

diesel/src/query_dsl/mod.rs
```diff
@@ -1,3 +1,12 @@
+//! Traits with methods to construct select statements
```
killercup: Hm. Can we find something other than "Traits with methods to" to say this is a thing that contains things to do select things? Maybe

> Module with traits to construct SELECT statements

diesel/src/query_dsl/mod.rs
katrinabrock: Existing mods don't have a blank line here. https://github.com/diesel-rs/diesel/blob/master/diesel/src/expression_methods/mod.rs  ¯\_(ツ)_/¯

killercup: fair enough 👍

katrinabrock:
So...I'm thinking the main thing is PR still needs is an example. Any thought as to what should be in it? What should a mod-level doc have as an example that's distinct from the examples in trait-level docs?

katrinabrock:
I'm working my way through the `query_dsl` mod. I know "one PR per file", but some of these are just adding a couple lines.

killercup:
Sorry for not following up on this earlier! Do you want to land this right now and add more stuff in a follow-up PR?

katrinabrock:
merging as is would be fine with me

killercup:
Let's get this landed, then!

killercup:
Ah, damn, on master this has `#![deny(missing_docs)]` so it [fails CI](https://travis-ci.org/diesel-rs/diesel/jobs/307970280#L975-L994) currently. I'll fix it.

katrinabrock:
oops! should have caught that.

PR #1276, opened 2017-10-27
I wrote some docs on a plane

sgrif:


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -2,6 +2,9 @@ use expression::{AsExpression, Expression};
 use super::operators::*;
 use types::{Array, Text};
 
+/// PostgreSQL expression methods which are not type specific.
+///
+/// This trait is implemented for all types that implement `Expression`.
```
killercup: Hmm, a general note: I think docs like

> This trait is implemented for […]

are pretty much worthless for people who know how to use rustdoc. They only appear on detail pages for the trait, and these pages include a list of implementors (see [rendered docs](https://docs.rs/diesel/0.16.0/diesel/expression_methods/trait.PgExpressionMethods.html#implementors)).

_But_ they may be useful in other places, e.g., tooltips in IDEs; or, when they convey more the indent of which types are expected to implemented this (like in `SortExpressionMethods`).

diesel/src/query_builder/insert_statement.rs
```diff
@@ -1,3 +1,5 @@
+/// Types related to the construction of an `INSERT` statement.
```
killercup: You meant `//!`, right?

diesel/src/query_source/mod.rs
```diff
@@ -82,6 +82,17 @@ pub trait Table: QuerySource + AsQuery + Sized {
...
+/// Determines how many times `Self` appears in `QS`
+///
+/// This trait is primarily used to determine whether or not a column is
+/// selectable from a given from clause. A column can be selected if its table
+/// appears in the from clause *exactly once*.
+///
+/// We do not allow the same table to appear in a query multiple times in any
+/// context where referencing that table would be ambiguous (depending on the
+/// context and backend being used, this may or may not be something that would
+/// otherwise result in a runtime error).
```
killercup: This is great! I'm pretty sure people will get error messages with `AppearsInFromClause` in them, and if the search for that, these are the docs they'll find. I can't think of any right now, but if we had a few "usual error cases" we could describe here (with fixes), that could be quite helpful. (Like the [error index](https://doc.rust-lang.org/1.21.0/error-index.html).)

PR #1279, opened 2017-10-27
Stop recompiling, I have shit to do

sgrif:
I wrote this wrong... It prints `rerun-if-changed=postgresql`, not the
actual directory. Interestingly, this causes Cargo to just
unconditionally recompile the crate, regardless of whether anything has
changed. That is undesirable. Instead, let's only recompile if there is
a new migration. That sounds like a good plan.

### DIFF
--- a/diesel_tests/build.rs
+++ b/diesel_tests/build.rs
@@ -54,7 +54,7 @@ fn main() {
     let migrations_dir = migrations::find_migrations_directory()
         .unwrap()
         .join(MIGRATION_SUBDIR);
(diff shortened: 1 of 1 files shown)


killercup:
> Interestingly, this causes Cargo to just
> unconditionally recompile the crate, regardless of whether anything has
> changed.

It's because cargo looks for a file at `./postgresql` and when it's not there it assumes it's been deleted: https://github.com/rust-lang/cargo/issues/4213

PR #1285, opened 2017-10-30
Add SQLite COLLATE operator support

gentoo90:
> When SQLite compares two strings, it uses a collating sequence or collating function (two words for the same thing) to determine which string is greater or if the two strings are equal. SQLite has three built-in collating functions: `BINARY`, `NOCASE`, and `RTRIM`.
> 
> * `BINARY` - Compares string data using `memcmp()`, regardless of text encoding.
> * `NOCASE` - The same as binary, except the 26 upper case characters of ASCII are folded to their lower case equivalents before the comparison is performed. Note that only ASCII characters are case folded. SQLite does not attempt to do full UTF case folding due to the size of the tables required.
> * `RTRIM` - The same as binary, except that trailing space characters are ignored.
>
https://sqlite.org/datatype3.html#collating_sequences

### DIFF
--- a/diesel/src/sqlite/expression/collate.rs
+++ b/diesel/src/sqlite/expression/collate.rs
@@ -1,0 +1,101 @@
+use backend::Backend;
+use expression::*;
+use query_builder::*;
+use result::QueryResult;
+
+pub trait Collation {
+    fn get_collation<'a>(&'a self) -> &'a str;
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct Collate<Lhs: Expression, Coll: Collation> {
+    left: Lhs,
+    collation: Coll,
+}
+
+impl<Lhs: Expression, Coll: Collation> Collate<Lhs, Coll> {
+    pub fn new(left: Lhs, collation: Coll) -> Self {
+        Collate { left, collation }
+    }
+}
+
+impl<Lhs, Coll> QueryId for Collate<Lhs, Coll>
+where
+    Lhs: 'static + Expression + QueryId,
+    Coll: 'static + Collation,
+{
+    type QueryId = Collate<Lhs, Coll>;
+    const HAS_STATIC_QUERY_ID: bool = Lhs::HAS_STATIC_QUERY_ID;
+}
+
+impl<Lhs, Coll, QS> SelectableExpression<QS> for Collate<Lhs, Coll>
+where
+    Collate<Lhs, Coll>: AppearsOnTable<QS>,
+    Lhs: Expression + SelectableExpression<QS>,
+    Coll: Collation,
+{
+}
+
+impl<Lhs, Coll, QS> AppearsOnTable<QS> for Collate<Lhs, Coll>
+where
+    Collate<Lhs, Coll>: Expression,
+    Lhs: Expression + AppearsOnTable<QS>,
+    Coll: Collation,
+{
+}
+
+impl<Lhs, Coll> Expression for Collate<Lhs, Coll>
+where
+    Lhs: Expression,
+    Coll: Collation,
+{
+    type SqlType = Lhs::SqlType;
+}
+
+impl<Lhs, Coll> NonAggregate for Collate<Lhs, Coll>
+where
+    Lhs: Expression + NonAggregate,
+    Coll: Collation,
+{
+}
+
+impl<Lhs, Coll, DB> QueryFragment<DB> for Collate<Lhs, Coll>
+where
+    Lhs: Expression + QueryFragment<DB>,
+    Coll: Collation,
+    DB: Backend,
+{
+    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
+        self.left.walk_ast(out.reborrow())?;
+        out.push_sql(" COLLATE ");
+        out.push_sql(self.collation.get_collation());
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct CollationBinary;
+
+impl Collation for CollationBinary {
+    fn get_collation<'a>(&'a self) -> &'a str {
+        "BINARY"
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct CollationNoCase;
+
+impl Collation for CollationNoCase {
+    fn get_collation<'a>(&'a self) -> &'a str {
+        "NOCASE"
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct CollationRTrim;
+
+impl Collation for CollationRTrim {
+    fn get_collation<'a>(&'a self) -> &'a str {
+        "RTRIM"
+    }
+}
--- a/diesel/src/sqlite/expression/expression_methods.rs
+++ b/diesel/src/sqlite/expression/expression_methods.rs
@@ -2,17 +2,20 @@ use expression::Expression;
 use super::operators::*;
 
 pub trait SqliteExpressionMethods: Sized {
-    /// Creates SQLite `COLLATE BINARY` expression
+    /// Creates SQLite `COLLATE BINARY` expression.
+    /// https://sqlite.org/datatype3.html#collating_sequences
     fn collate_binary(self) -> CollateBinary<Self> {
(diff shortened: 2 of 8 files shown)


diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -0,0 +1,21 @@
+use expression::Expression;
+use super::operators::*;
+
+pub trait SqliteExpressionMethods: Sized {
+    /// Creates SQLite `COLLATE BINARY` expression
```
killercup: Can you link to the sqlite docs in each of these comments so people who have no idea what that is have a link to click?

diesel_tests/tests/filter.rs
```diff
@@ -460,3 +460,16 @@ fn filter_subselect_with_pg_any() {
...
+#[cfg(feature = "sqlite")]
+fn filter_by_string_equality_case_insensitive() {
+    use schema::users::dsl::*;
+
+    let connection = connection_with_sean_and_tess_in_users_table();
+
+    let sean = User::new(1, "Sean");
+    let tess = User::new(2, "Tess");
+    assert_eq!(Ok(sean), users.filter(name.eq("sean").collate_nocase()).first(&connection));
+    assert_eq!(Ok(tess), users.filter(name.eq("Tess ").collate_rtrim()).first(&connection));
```
killercup: This is a pretty much perfect PR but rustfmt still complains that these lines should be broken up :) See CI log for details

gentoo90:
> Thanks! That's a This should only work on strings, though, right? Or does it work on numbers and binaries, too?

Well, applying it to `INTEGER` doesn't cause any syntax errors.

diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -0,0 +1,24 @@
+use expression::Expression;
+use super::operators::*;
+
+pub trait SqliteExpressionMethods: Sized {
```
weiznich: To prevent having the "same" method thrice we could do something like this:
```rust
pub trait SqlExpressionMethods: Sized {
    fn collate<T>(self, collate: T) -> Collate<Self, T> where T: CollateKind {
        Collate::new(self)
    }
}

pub trait CollateKind{}
pub mod collate {
    pub struct Binary;
    pub struct NoCase; 
    pub struct RTrim;

    impl CollateKind for Binary {}
    impl CollateKind for NoCase {}
    impl CollateKinde for RTrim {}
}
```

(This is only a suggestion to discuss this with @diesel-rs/contributors )

diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    /// Creates SQLite `COLLATE RTRIM` expression.
+    /// https://sqlite.org/datatype3.html#collating_sequences
+    fn collate_rtrim(self) -> CollateRTrim<Self> {
+        CollateRTrim::new(self)
+    }
+}
+
+impl<T: Expression> SqliteExpressionMethods for T {}
```
sgrif: It looks like this is only valid after a specific set of expressions. Can we limit this to that set, and get a compile test that it fails for something that it shouldn't apply to (e.g. `LIKE`)?

gentoo90: > The `COLLATE` operator is a unary postfix operator that assigns a collating sequence to an expression. The `COLLATE` operator has a higher precedence (binds more tightly) than any binary operator and any unary prefix operator except `~`. (`COLLATE` and `~` are associative so their binding order does not matter.) The collating sequence set by the `COLLATE` operator overrides the collating sequence determined by the `COLLATE` clause in a table column definition.

https://sqlite.org/lang_expr.html#collateop

For example:
```sql
SELECT * FROM [projects]
    WHERE [name] COLLATE NOCASE = 'test' COLLATE RTRIM
    ORDER BY NAME COLLATE NOCASE COLLATE BINARY;
```
is syntactically correct.

diesel/src/sqlite/expression/expression_methods.rs
gentoo90: Probably a good idea.
>An application can register additional collating functions using the `sqlite3_create_collation()` interface.

https://sqlite.org/datatype3.html#collating_sequences

diesel/src/sqlite/expression/collate.rs
```diff
@@ -0,0 +1,101 @@
+use backend::Backend;
+use expression::*;
+use query_builder::*;
+use result::QueryResult;
+
+pub trait Collation {
+    fn get_collation<'a>(&'a self) -> &'a str;
```
weiznich: I think this should be a associated constant instead of a member function.
```rust
const COLLATION: &'static str;
```

diesel/src/sqlite/expression/collate.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct Collate<Lhs: Expression, Coll: Collation> {
+    left: Lhs,
+    collation: Coll,
```
weiznich: If we use a associated constant, we do not need to store collation explicitly. So maybe use `std::marker::PhantomData` here?

diesel/src/sqlite/expression/collate.rs
gentoo90: I thought about this too, but then I found out that other DB engines also have `COLLATE`, and it's used a little differently.
https://www.postgresql.org/docs/9.6/static/collation.html
https://dev.mysql.com/doc/refman/5.7/en/charset-collate.html

If this trait will be reused for them, then it's probably better to allow non-static strings here.

sgrif: This trait would need to be generic over the backend for it to be used by more than just SQLite. If we think we will want to re-use this for other backends, we need to do that in this PR. We're a 1.0 library, so we can't change this later.

sgrif:
So I think this goes the wrong direction for this feature. We should support this for all backends since they all use nearly identical syntax. I also don't think we should take a type here. All backends support the registration of additional collations, and the built-in collations available depend on the platform the server is running on. We could try to do schema inference on this, but I don't think it's a big enough win for the amount of work that would require. This is a pretty low risk method. So I think the implementation aught to look a bit more like this:


```
pub struct Collate<'a, Expr> {
    expr: Expr,
    collation: Cow<'a, str>,
}

impl QueryFragment<Mysql> for Collate<'a, Expr> {
    fn walk_ast(&self, mut out: AstPass<Mysql>) -> QueryResult<()> {
        self.expr.walk_ast(out.reborrow())?;
        out.push_sql(" COLLATE ");
        out.push_sql(&collation);
        Ok(())
    }
}

impl QueryFragment<Sqlite> for Collate<'a, Expr> {
    fn walk_ast(&self, mut out: AstPass<Sqlite>) -> QueryResult<()> {
        self.expr.walk_ast(out.reborrow())?;
        out.push_sql(" COLLATE ");
        out.push_sql(&collation);
        Ok(())
    }
}

impl QueryFragment<Pg> for Collate<'a, Expr> {
    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
        self.expr.walk_ast(out.reborrow())?;
        out.push_sql(" COLLATE \"");
        out.push_sql(&collation);
        out.push_sql("\"");
        Ok(())
    }
}

impl<'a, Expr::SqlType> Expression for Collate<'a, Expr> {
    type SqlType = Expr::SqlType;
}

impl<'a, Expr> QueryId for Collate<'a, Expr> {
    type QueryId = ();

}

pub trait TextExpressionMethods {
    fn collate<'a, T>(self, collation: T) -> Grouped<Collate<'a, Self>>
    where
        T: Into<Cow<'a, T>>,
    {
        Grouped::new(Collate::new(self, collation))
    }
}
```

sgrif:
@gentoo90 Are you still interested in working on this?

gentoo90:
@sgrif Write now I don't have time for this. Also I lack knowledge of MySQL and PostgreSQL to do it properly.

sgrif:
Thanks for taking the time for the first attempt

PR #1290, opened 2017-11-02
Fix the large tables (and above) features

Eijebong:
In 6cfafef7f, in the build function, a generic parameter named `R` was
introduced for the build function. It's colliding with the name of our
generic tuple stuff so I just renamed it to `RowT`


NotBad4U:
Maybe we should add the feature `large-table` to [diesel_tests/Cargo.toml](https://github.com/diesel-rs/diesel/blob/master/diesel_tests/Cargo.toml) and add a test with a `struct` that use this feature to avoid to encounter this bug again ?

PR #1292, opened 2017-11-03
open SQLite connection with SQLITE_OPEN_URI

euclio:
This allows the user to set options such as in-memory caching on the database.

See the documentation here for notes on backwards-compatibility: https://www.sqlite.org/uri.html.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   They are no longer dependent on the current working directory (for all
   directories in the same project)
 
+* The SQLite backend is now configured to interpret URIs.
+  See [the SQLite URI documentation] for additional details.
+
+[the SQLite URI documentation]: https://www.sqlite.org/uri.html
+
 ### Deprecated
 
 * `diesel_(prefix|postfix|infix)_operator!` have been deprecated. These macros
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -23,9 +23,16 @@ pub struct RawConnection {
 impl RawConnection {
     pub fn establish(database_url: &str) -> ConnectionResult<Self> {
         let mut conn_pointer = ptr::null_mut();
-        let database_url = CString::new(database_url.trim_start_matches("sqlite://"))?;
-        let connection_status =
-            unsafe { ffi::sqlite3_open(database_url.as_ptr(), &mut conn_pointer) };
+
+        let database_url = if database_url.starts_with("sqlite://") {
+            CString::new(database_url.replacen("sqlite://", "file:", 1))?
+        } else {
+            CString::new(database_url)?
+        };
+        let flags = ffi::SQLITE_OPEN_READWRITE | ffi::SQLITE_OPEN_CREATE | ffi::SQLITE_OPEN_URI;
+        let connection_status = unsafe {
+            ffi::sqlite3_open_v2(database_url.as_ptr(), &mut conn_pointer, flags, ptr::null())
+        };
 
         match connection_status {
             ffi::SQLITE_OK => {
--- a/diesel_tests/tests/connection.rs
+++ b/diesel_tests/tests/connection.rs
@@ -82,3 +82,22 @@ fn strips_sqlite_url_prefix() {
     path.push("diesel_test_sqlite.db");
     assert!(SqliteConnection::establish(&format!("sqlite://{}", path.display())).is_ok());
 }
+
+#[test]
+#[cfg(feature = "sqlite")]
+fn file_uri_created_in_memory() {
+    use std::path::Path;
+
+    assert!(SqliteConnection::establish("file::memory:").is_ok());
+    assert!(!Path::new("file::memory:").exists());
+    assert!(!Path::new(":memory:").exists());
+}
+
+#[test]
+#[cfg(feature = "sqlite")]
+fn sqlite_uri_prefix_interpreted_as_file() {
+    let mut path = std::env::temp_dir();
+    path.push("diesel_test_sqlite_readonly.db");
+    assert!(SqliteConnection::establish(&format!("sqlite://{}?mode=rwc", path.display())).is_ok());
+    assert!(path.exists());
+}


sgrif:
This looks like it has more to do with how you compiled SQLite than Diesel. Am I missing something?

euclio:
SQLite has support for opening databases with URIs such as `file::memory:?cache=shared`, which allows you to turn various knobs while opening the database (caching, read-only, etc.). However, this support is disabled by default. An application may request URI support by supplying the correct flags to `sqlite_open_v2`. Observe that if you use that URI with diesel a file named `file::memory:?cache=shared` will be created instead of an in-memory database (at least, it will for a brewed sqlite on Mac OS).

euclio:
Perhaps an example will make the issue clearer?

This program creates the database in memory, which is expected when using a URI.

```rust
extern crate rusqlite;

use rusqlite::Connection;

fn main() {
    Connection::open("file::memory:?cache=shared").unwrap();
}
```

On the other hand, this program creates a the database in a file called `file::memory:?cache=shared`.

```rust
extern crate diesel;

use diesel::prelude::*;

fn main() {
    SqliteConnection::establish("file::memory:?cache=shared").unwrap();
}
```

sgrif:
Sorry for not replying to this issue sooner. I didn't mean to ignore it, but we will likely not be able to address it until after 1.0.0

diesel/src/sqlite/connection/raw.rs
```diff
@@ -17,8 +19,11 @@ impl RawConnection {
     pub fn establish(database_url: &str) -> ConnectionResult<Self> {
         let mut conn_pointer = ptr::null_mut();
         let database_url = try!(CString::new(database_url));
-        let connection_status =
-            unsafe { ffi::sqlite3_open(database_url.as_ptr(), &mut conn_pointer) };
+        let flags =
+            SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_NOMUTEX | SQLITE_OPEN_URI;
```
sgrif: Why are we specifying `OPEN_NOMUTEX` here? It is not a flag set by `sqlite3_open`, and it seems unrelated to your change.

sgrif:
So if I'm reading this correctly, the only time this has a change in behavior is when the path starts with `file:`? (which means this is technically a breaking change for anyone who has a file starting with `file:` but I think we can safely ignore that)

diesel/src/sqlite/connection/raw.rs
euclio: My understanding is that this flag is required to use the shared memory cache mode when opening a file with URI `file::memory:?cache=shared`.

euclio:
That's correct.

diesel/src/sqlite/connection/raw.rs
sgrif: I'd prefer we hold off on making that change for the time being. We certainly can't just do it by default, but we can possibly expose an API for this in the future.

sgrif:
@euclio Are you still interested in finishing this?

euclio:
Yes, will address your comments within the week.

euclio:
Addressed comments.

vemoo:
What's the status of this PR?
I'm interested because I want to be able to create a readonly connection, and I think this is the only way to achieve it.

sgrif:
The status appears to be that it has a failing build and needs a rebase. The test suite would need to pass before this can be further reviewed

sgrif:
I'd also like to see a more meaningful test -- The test provided passes without this change

euclio:
Cool, will address feedback within the week. I haven't had much impetus to work on this since my own project that uses Diesel has stalled, but if there's more interest than just me I'm happy to work on it :)

vemoo:
Awesome, thank you!

euclio:
@sgrif This is ready for another review. Failure on nightly is unrelated.

sgrif:
I agree that the clippy failure seems unrelated, but it only occurs on this PR. I can't reproduce on master or locally

sgrif:
Never mind I see now it is affecting other PRs

joelgallant:
@euclio did you want to rebase and finish this? I could also try to finish it off

euclio:
Thanks for the ping, @joelgallant. Rebased.

euclio:
Ping @sgrif

kevincox:
Any updates on this? The docs say that this is supported so the lack of support for URIs is very confusing. https://docs.rs/diesel/1.4.4/diesel/sqlite/struct.SqliteConnection.html

weiznich:
@kevincox I would happily accept a rebased version of this PR, as long as the CI passes Otherwise a fix for the documentation would also be welcome.

euclio:
@weiznich I went ahead and rebased this PR.

CHANGELOG.md
```diff
@@ -39,6 +39,13 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   functionality of `NonAggregate`. See [the upgrade
   notes](#2-0-0-upgrade-non-aggregate) for details.
 
+### Fixed
```
weiznich, line 42: There is already a `Fixed` section in line 94. Could you move this point there?
Additionally I'm not that happy with the formulation "This allows
  applications to use in-memory databases" as it is already possible to do this using `":memory:"` as connection string passed to `SqliteConnection::establish`. Maybe just drop that point and only say that we support full sqlite URIs now?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -24,8 +24,10 @@ impl RawConnection {
     pub fn establish(database_url: &str) -> ConnectionResult<Self> {
         let mut conn_pointer = ptr::null_mut();
         let database_url = CString::new(database_url.trim_start_matches("sqlite://"))?;
```
weiznich, line 26: Just a random idea here: Should we treat `sqlite://URI` the same as `file://URI`? If so we should change the code in such a way that we replace `sqlite://` at the beginning of a connection URI with `file://` instead of just removing that part.

euclio: Makes sense to me. Done.

CHANGELOG.md
euclio: Done.

euclio:
Looks like the CI failures are spurious.

euclio:
Happy this finally got merged! Thanks :)

PR #1293, opened 2017-11-04
Change links to https

Ppjet6:
More TLS love!


Ppjet6:
So many READMEs, more commits coming!

Ppjet6:
I haven't converted the link to [www.apache.org](https://www.apache.org) in LICENSE-APACHE, not sure what to do here. If it's fine I'll push one more.
For the rest I checked and they don't support yet.

notryanb:
LGTM 👍

@Ppjet6 I think a redirect might be a good idea if we want to force TSL for some reason? I don't really know enough to make a sound decision, but I'm assuming this helps with SEO?
side note - is there a way to `301` http -> https for docs pages?

killercup:
Sorry to leave this hanging so long!

PR #1298, opened 2017-11-09
Use the scale from postgres to construct our BigDecimal in FromSql

Eijebong:
This `with_scale` method didn't exist when we added support for the
bigdecimal crate and it was causing 0.01 to be interpreted as 0.01000.
Now that it exists, use it.


Eijebong:
Woups rustfmt :D  I knew I forgot about something

What do you mean by explicit values ? Something like "0.01000" ?

killercup:
What? 0.01? No, 4.200000 obviously!

PR #1304, opened 2017-11-14
Fix global cli argument handling

### LINKED ISSUE #1301: --migration-dir option does not work
### Versions

- **Rust:** 1.21
- **Diesel:** 0.16
- **Database:** SQLite
- **Operating System** MacOS 10.13.1

## Problem Description

The CLI does not respect the --migration-dir option.

Perform diesel migration commands on a custom migration directory. Specifically the following command: `diesel migration --database-url my.db --migration-dir migrations/sqlite list`

The `migrations/sqlite` folder contains a folder called `2017-11-11-060348_initial`, which contains only `up.sql` and `down.sql`. `my.db` has had no migrations applied.

The command should show the list of available migrations.

`Invalid migration directory, the directory's name should be <timestamp>_<name_of_migration>, and it should only contain up.sql and down.sql.`

weiznich:
Fixes #1301 

This is a improved/newer version of #1201

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/main.rs"
 
 [dependencies]
 chrono = "0.4"
-clap = "2.20.3"
+clap = ">=2.27.0"
 diesel = { version = "0.16.0", default-features = false }
 dotenv = ">=0.8, <0.11"
 diesel_infer_schema = "0.16.0"
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -54,32 +54,32 @@ fn main() {
 
 fn run_migration_command(matches: &ArgMatches) {
     match matches.subcommand() {
-        ("run", Some(args)) => {
+        ("run", Some(_)) => {
             let database_url = database::database_url(matches);
-            let dir = migrations_dir(args);
+            let dir = migrations_dir(matches);
             call_with_conn!(
                 database_url,
                 migrations::run_pending_migrations_in_directory(&dir, &mut stdout())
             ).unwrap_or_else(handle_error);
         }
-        ("revert", Some(args)) => {
+        ("revert", Some(_)) => {
             let database_url = database::database_url(matches);
-            let dir = migrations_dir(args);
+            let dir = migrations_dir(matches);
             call_with_conn!(
                 database_url,
                 migrations::revert_latest_migration_in_directory(&dir)
             ).unwrap_or_else(handle_error);
         }
-        ("redo", Some(args)) => {
+        ("redo", Some(_)) => {
             let database_url = database::database_url(matches);
-            let dir = migrations_dir(args);
+            let dir = migrations_dir(matches);
             call_with_conn!(database_url, redo_latest_migration(&dir));
         }
-        ("list", Some(args)) => {
+        ("list", Some(_)) => {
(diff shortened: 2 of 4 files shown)


diesel_cli/src/main.rs
```diff
@@ -153,10 +153,20 @@ fn migration_version<'a>(matches: &'a ArgMatches) -> Box<Display + 'a> {
         .unwrap_or_else(|| Box::new(Utc::now().format(TIMESTAMP_FORMAT)))
 }
 
-fn migrations_dir(matches: &ArgMatches) -> PathBuf {
+fn migrations_dir_from_cli(matches: &ArgMatches) -> Option<PathBuf> {
     matches
         .value_of("MIGRATION_DIRECTORY")
         .map(PathBuf::from)
+        .or_else(|| {
```
notryanb: @weiznich - I'm still becoming familiar with rust idioms and clap! 😅. From what I understand, `Pathbuf::from` indicates that you expect the `MIGRATION_DIRECTORY` value will always convert without fail. Why do we need to chain `.or_else` and then match on the subcommand (name?, not sure which value of the tuple `1` represents)?

weiznich: > From what I understand, `Pathbuf::from` indicates that you expect the `MIGRATION_DIRECTORY` value will always convert without fail. 

As far as I understood the stdlib docs/code this conversion is guaranteed to succeed. If the user passes a nonsense path a `PathBuf` pointing to a non existing (the nonsense path) location is generated and the following operations executed by diesel will end with a error message indication a invalid/non existing path.

>Why do we need to chain `.or_else` and then match on the subcommand (name?, not sure which value of the tuple 1 represents)?

The `or_else` is needed to workaround some architectural decisions/limitations in clap. A clap app consists of several subcommands (migration, setup, print-schema , … for diesel_cli). Each subcommand could also consists of several subcommands (list, run, … for diesel migration). Arguments are defined in a subcommand and are valid for the defined subcommand and all child subcommands.  `ArgMatches` contains all arguments passed to a certain subcommand. Arguments are stored in that `ArgMatches` to which they are passed, not in which they are defined. So if someone is calling `diesel migration --migration_dir /some/dir list` the migration_dir argument is stored in the `ArgMatches` of migration. It is also possible to call `diesel migration list --migration_dir /some/dir` where migration_dir is stored in the `ArgMatches` of list. (There is a global option that allows a argument globally, but that does not solve this problem here). Therefore we need to traverse the results for each subcommand. This is done by recursively calling this method on each subcommand. (See the [clap docs](https://docs.rs/clap/2.27.1/clap/struct.ArgMatches.html#method.subcommand) for why the we access `.1` on the tuple)

notryanb: Awesome!!! 😄  Thanks so much! I'll finish review ASAP this week (unless someone else beats me to it).

notryanb:
@killercup @Eijebong - I'd like to merge this. Good to go?

killercup:
LGTM! Go ahead, @notryanb!

diesel_cli/Cargo.toml
```diff
@@ -16,7 +16,7 @@ path = "src/main.rs"
 
 [dependencies]
 chrono = "0.4"
-clap = "2.20.3"
+clap = ">=2.27.0"
```
sgrif: Diesel CLI 0.99 is going to break when a new version of clap is released...

PR #1305, opened 2017-11-16
Revise associations and joins documentation

notryanb:
`Associations` docs currently portray a role in the `JoinDsl`. Updates documentation where they reference each other to make a clear distinction in use-cases and responsibilities.

### DIFF
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -65,17 +65,17 @@
 //! custom one to `#[belongs_to]` by adding a `foreign_key` argument to the
 //! attribute like so `#[belongs_to(Foo, foreign_key="mykey")]`.
 //!
-//! `Associations` are separate from joins in Diesel. Please see docs on [`JoinDsl!`]
+//! `Associations` are separate from joins in Diesel. Please see docs on [`JoinDsl`]
 //! to learn more about joining tables.
 //!
-//! [`JoinDsl!`]: ../prelude/trait.JoinDsl.html
+//! [`JoinDsl`]: ../query_dsl/trait.JoinDsl.html
 //!
 //! You can load the children for a single parent or multiple parents (`Vec<ParentType>`) using the
 //! [`belonging_to()`][belonging-to] method.
 //! This amounts to the `SQL` statements `SELECT * FROM posts WHERE posts.user_id = $1`
 //! or `SELECT * FROM posts WHERE posts.user_id IN ($1, $2, etc...)`
 //!
-//! [belonging-to]: ../prelude/trait.BelongingToDsl.html#tymethod.belonging_to
+//! [belonging-to]: ../query_dsl/trait.BelongingToDsl.html#tymethod.belonging_to
 //!
 //! ```rust
 //! # #[macro_use] extern crate diesel;
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -1,14 +1,34 @@
 //! Traits related to relationships between multiple tables.
 //!
-//! Associations in Diesel are bidirectional, but primarily focus on the child-to-parent
-//! relationship. You can declare an association between two records with
-//! `#[belongs_to]`.
+//! **Note: This feature is under active development, and we are seeking feedback on the APIs that
+//! have been released. Please feel free to [open issues][open-issue], or join [our chat][gitter]
+//! to provide feedback.**
+//!
+//! [open-issue]: https://github.com/diesel-rs/diesel/issues/new
+//! [gitter]: https://gitter.im/diesel-rs/diesel
+//!
+//! Associations in Diesel are unidirectional and focus on the child-to-parent
+//! relationship.
+//!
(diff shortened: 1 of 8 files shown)


diesel/src/associations/mod.rs
```diff
@@ -1,18 +1,24 @@
...
+//! In order to declare this relationship, you must annotate your child struct
+//! with `#[derive(Associations)]` and `#[belongs_to(ParentStructName)]`.
+//! `#[belongs_to]` is given the name of the struct that represents the parent. Both parent and
+//! child types must implement the [`Identifiable`][identifiable] trait.
+//! The struct or table referenced in your association has to be in scope,
+//! so you'll need `use schema::posts` or similar to bring the
+//! table into scope.
 //!
 //! ```rust
 //! // You need to have the table definitions from `infer_schema!` or `table!` in scope to
```
weiznich: Just add those table definitions here?

diesel/src/associations/mod.rs
```diff
@@ -149,6 +99,7 @@
 //! #   use users::dsl::*;
 //! #   let connection = establish_connection();
 //! #
+//! // Loading the posts for a single parent user.
 //! let user = users.find(1).first::<User>(&connection).expect("Error loading user");
```
weiznich: This example seems to be very similar to that one above, so do we really need both? (At least we could change this one to load all posts for multiple users?)

diesel/src/query_dsl/join_dsl.rs
```diff
@@ -88,7 +88,7 @@ where
 /// [`allow_tables_to_appear_in_same_query!`]: ../macro.allow_tables_to_appear_in_same_query.html
 pub trait JoinDsl: Sized {
     /// Join two tables using a SQL `INNER JOIN`. The `ON` clause is defined
-    /// via the [associations API](../associations/index.html).
+    /// via the [JoinDsl](./trait.JoinOnDsl.html#method.on).
```
weiznich: We should mention here that a default/implicit on clause is used if `JoinOnDsl::on` is not called.

diesel/src/query_dsl/join_dsl.rs
```diff
@@ -97,7 +97,7 @@ pub trait JoinDsl: Sized {
     }
 
     /// Join two tables using a SQL `LEFT OUTER JOIN`. The `ON` clause is defined
-    /// via the [associations API](../associations/index.html).
+    /// via the [JoinDsl](./trait.JoinOnDsl.html#method.on).
```
weiznich: Same as for `inner_join`

diesel/src/query_source/joins.rs
```diff
@@ -154,8 +154,8 @@ where
 // }
 
 /// Indicates that two tables can be used together in a JOIN clause.
-/// Implementations of this trait will be generated for you automatically by
-/// the [association annotations](../associations/index.html) from codegen.
+/// Implementations of this trait will be generated for you when using
+/// the [joinable!](../macro.joinable.html) macro.
```
weiznich: We should mention here that calls to that macro are generated by `infer_schema!` and `print-schema` if the database relations exists in SQL (`REFERENCES`)

Also we should mention here the [`allow_tables_to_appear_in_same_query!`](http://docs.diesel.rs/diesel/macro.allow_tables_to_appear_in_same_query.html) macro that enables multi table joins.

diesel/src/associations/mod.rs
notryanb: @weiznich Add the definitions for the doc-test example or re-word the preceding paragraph? The tables are imported via `doctest_setup.rs`.

weiznich: I think adding adding the actual definitions here will make the example easier to understand because the all needed code is included in the actual example. (This is not needed everywhere, but because associations are quite complex I think the docs may be improved by adding them here.)

diesel/src/query_dsl/join_dsl.rs
```diff
@@ -44,6 +44,84 @@ where
...
+/// aliases.
+///
+/// You may also need to call [`allow_tables_to_appear_in_same_query!`][] (particularly if
+/// you see an unexpected error about `AppearsInFromClause`). See the
+/// documentation for [`allow_tables_to_appear_in_same_query!`][] for details.
+///
+/// Diesel expects multi-table joins to be semantically grouped based on the
+/// relationships. For example, `users.inner_join(posts.inner_join(comments))`
+/// is not the same as `users.inner_join(posts).inner_join(comments)`. The first
+/// would deserialize into `(User, (Post, Comment))` and generate the following
```
pickfire, line 69: Is there an example for this? I saw `(User, Post, Comment)` but I don't know why it isn't working, it would be good to have example so easier to follow. The example below uses explicit select.

weiznich: @pickfire Please don't ask questions on old PR's. That's definitively not the right place for such things. Use the gitter channel or our forum instead.

PR #1312, opened 2017-11-27
Someone broke master

weiznich:
Add some documentation for DistinctOnDsl

(To prevent this in future maybe we should use something like [bors](https://bors.tech/)?)


killercup:
Let's merge this as soon as CI is green :)

PR #1315, opened 2017-11-28
Add a missing ToSql impl for postgres ranges

Eijebong:
For some reasons the range implementation was missing the
`ToSql<Nullable<ST>> for T` impl.

Thanks @valeriansaliou for the bug report :)

### DIFF
--- a/diesel/src/pg/types/ranges.rs
+++ b/diesel/src/pg/types/ranges.rs
@@ -181,6 +181,20 @@ where
     }
 }
 
+
+impl<ST, T> ToSql<Nullable<Range<ST>>, Pg> for (Bound<T>, Bound<T>)
+where
+    Pg: HasSqlType<Range<ST>>,
+    (Bound<T>, Bound<T>): ToSql<Range<ST>, Pg>
+{
+    fn to_sql<W: Write>(
+        &self,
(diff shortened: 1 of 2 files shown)


diesel/src/pg/types/ranges.rs
```diff
@@ -181,6 +181,20 @@ where
     }
 }
 
+
+impl<ST, T> ToSql<Nullable<Range<ST>>, Pg> for (Bound<T>, Bound<T>)
+where
+    Pg: HasSqlType<ST> + HasSqlType<Range<ST>>,
+    T: ToSql<ST, Pg>,
```
sgrif: Should this be `(Bound<T>, Bound<T>): ToSql<Range<ST>, Pg>`?

diesel/src/pg/types/ranges.rs
```diff
@@ -181,6 +181,20 @@ where
     }
 }
 
+
+impl<ST, T> ToSql<Nullable<Range<ST>>, Pg> for (Bound<T>, Bound<T>)
+where
+    Pg: HasSqlType<ST> + HasSqlType<Range<ST>>,
```
sgrif: Can we drop the `HasSqlType<ST>` if the other comment is addressed?

PR #1323, opened 2017-11-30
Fix docs for dsl::count and dsl::count_star

dbrgn:
It looks like the docs weren't fully updated yet :)


dbrgn:
Oh, don't merge this yet, I found another problem in these docs...

dbrgn:
Done: Updated the target of the `count` link :)

weiznich:
Tests are failing because master is broken again :disappointed: (Maybe I've some time to fix this today)

PR #1327, opened 2017-11-30
Make `#[derive(QueryableByName)]` work without a table

sgrif:
crates.io required 5 manual `QueryableByName` implementations. Of those
5, 4 of them were only required because they were for use with a struct
that didn't map to fields on a single table. This change allows use with
those structs by requiring that the SQL type be specified on each field
instead.

Since the default behavior is based on the table name, and forgetting to
annotate the fields is an easy mistake to make, I've changed
`QueryableByName` to no longer use the inferred table name. Instead the
annotation will be required to determine the SQL type based on the
column.

### DIFF
--- a/diesel_derives/src/queryable_by_name.rs
+++ b/diesel_derives/src/queryable_by_name.rs
@@ -1,6 +1,7 @@
 use quote::Tokens;
 use syn;
 
+use attr::Attr;
 use model::Model;
 use util::wrap_item_in_const;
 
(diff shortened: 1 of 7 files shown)


diesel_derives/src/queryable_by_name.rs
```diff
@@ -43,15 +43,30 @@ pub fn derive(item: syn::DeriveInput) -> Tokens {
...
+fn sql_type(attr: &Attr, model: &Model) -> Tokens {
+    let table_name = model.table_name();
+    let column_name = attr.column_name();
+
+    match attr.sql_type() {
+        Some(st) => quote!(#st),
+        None => if model.has_table_name_annotation() {
+            quote!(diesel::dsl::SqlTypeOf<#table_name::#column_name>)
+        } else {
+            panic!(
```
weiznich: Instead of panicing in error cases we should emit a `compile_error!` in place of the actual `impl QueryableByName`. This should result in better readable error messages.

sgrif: The error message is the same either way. Panicking is how proc macros communicate errors to the user.

sgrif: screen shot 2017-11-30 at 10 55 13 am

sgrif: I guess having the message at the top is marginally better

notryanb:
@sgrif Is this slated for the `1.0` release?

sgrif:
Yes, I'm going to include some minor changes to this derive in 1.0

PR #1329, opened 2017-11-30
 Rewrite ci stuff to enable bors

weiznich:
* For pull request now only the following builds are performed:
  1. Travis stable postgres
  2. Travis stable mysql
  3. Travis stable sqlite
  4. Travis clippy
  5. Travis rustfmt

  Those 5 builds seems to run in parallel and finish in <
  20min (Postgres build is the slowest)

* For bors try and bors r+ all builds are performed.
* Added a test builds for our minimal supported rust version (1.20 as
  time of writing) to detect breaking changes by raising this version

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,8 +5,8 @@ image: Visual Studio 2013
 
 branches:
   only:
-    - master
-    - appveyor_nonsense
+    - staging
+    - trying
 
 services:
   - postgresql95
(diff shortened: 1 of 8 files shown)


sgrif:
I'm not sure I see the need to add bors to this project at the moment. CI is not nearly enough of a bottleneck to cause issues, and we don't have nearly enough commit volume to need to worry about CI passing on a branch but failing on merge

killercup:
I personally like the idea behind bors quite a lot, and use it for most of my projects, but I also see its downsides. For me it's mostly about being sure that master always works, and me not having to do anything. If we keep doing regular releases and basically no-one is pointing at our master branch, it's fine for master to sometimes break. Another data point: We often have PRs that are open for a few weeks. If we make sure to rebase them before merging, we can greatly reduce the risk of breaking master.

Bors also adds another feature: We can delegate the ability to "r+" PRs to people who don't have the ability to merge Diesel PRs on Github.

Eijebong:
Seeing how many "Fix build on master" commits we're getting recently, I'm :+1: on the idea

sgrif:
We could just as easily fix that by making master require green CI before merging.

weiznich:
> We could just as easily fix that by making master require green CI before merging.

There was already a [case](https://github.com/diesel-rs/diesel/commit/5d820ceb6e26b6a491b0067bcabf1cc636cc2105) where merging a pull request with green CI broke master. (Note the broken build for the merge commit, but the green build for the last commit in the pull request.)
Such situations are not prevented by requiring to have the CI green.

sgrif:
I'm still not fundamentally opposed to this, but I think we should go ahead and close this for the time being. Our CI is in a weird place right now where it's slow enough that I hesitate to add anything that will make it slower (see the backlog right now), but it's not *so slow* that waiting on CI before merging something is a major problem.

I get that I've been a bit reckless at times with merging things I shouldn't have, but that doesn't result in any serious problems with the current state of the project (there's zero chance that a bad master for an hour results in nightly not getting published for example). Pretty much every case where something has passed CI but failed post-merge has been due to an update of clippy/rustfmt, and I don't think that warrants this kind of tooling.

PR #1332, opened 2017-11-30
Document the `pg` module

sgrif:


diesel/src/pg/backend.rs
```diff
@@ -7,12 +9,16 @@ use super::PgMetadataLookup;
 use super::query_builder::PgQueryBuilder;
 use types::Oid;
 
+/// The PostgreSQL backend
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct Pg;
 
+/// The OIDs for a SQL type
```
killercup: Two weeks from now:

> Random Diesel user: Dafuq is an OID? Oh, there's a link to [the PG docs](https://www.postgresql.org/docs/current/static/datatype-oid.html), that's nice!
>
> Pascal: Of course!
> 
> Pascal (whispering): Good thing I urged Sean to add _yet another_ link to the docs.

diesel/src/pg/types/date_and_time/mod.rs
```diff
@@ -44,12 +44,21 @@ pub struct PgTime(pub i64);
...
     pub months: i32,
 }
 
 impl PgInterval {
+    /// Constructs a new `PgInterval`
+    ///
+    /// No conversion occurs on the arguments. It is valid to provide a number
+    /// of microseconds greater than the longest possible day, or a number of
+    /// days greater than the longest possible month, as it is impossible to say
+    /// how many months are in "40 days" without knowing a precise date.
```
killercup: Maybe also add an answer to: What happens when this kind of invalid data is used in a query?

diesel/src/types/impls/option.rs
```diff
@@ -60,6 +61,25 @@ where
     }
 }
 
+impl<T, DB> QueryableByName<DB> for Option<T>
```
killercup: Was this addition meant as part of the PG docs PR?

PR #1334, opened 2017-11-30
Minor changes to the backend installation guide

sgrif:
This has some grammar fixes, adds details about how to specify multiple backends, and removes the bit about passing features to Diesel, as it implies to me that your `Cargo.toml` affects `cargo install diesel_cli`, which is not true.

### DIFF
--- a/guide_drafts/backend_installation.md
+++ b/guide_drafts/backend_installation.md
@@ -3,14 +3,14 @@
 Diesel supports SQLite, PostgreSQL, and MySQL as database backends.
 By default, `diesel_cli`
 requires the client library of all three backends to be installed.
-If one is missing, then `cargo install diesel_cli` you will throw an error like:
+If one is missing, then `cargo install diesel_cli` you will receive an error like:
 
 ```
 note: ld: library not found for -lmysqlclient
 clang: error: linker command failed with exit code 1 (use -v to see invocation)
 ```
 
-To install `diesel_cli` without all backends,
+To install Diesel CLI with only the backends you need,
 specify `--no-default-features`.
 Use cargo's `--features` option to specify `postgres`, `sqlite`, and/or `mysql`.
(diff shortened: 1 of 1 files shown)


guide_drafts/backend_installation.md
```diff
@@ -3,14 +3,14 @@
 Diesel supports SQLite, PostgreSQL, and MySQL as database backends.
 By default, `diesel_cli`
 requires the client library of all three backends to be installed.
-If one is missing, then `cargo install diesel_cli` you will throw an error like:
+If one is missing, then `cargo install diesel_cli` you will receive an error like:
```
killercup: "…then `foo` you will receive" doesn't read like English to me. Wasn't the original meant like

> then `cargo install diesel_cli` will throw (you) an error like

guide_drafts/backend_installation.md
```diff
@@ -19,13 +19,12 @@ For example, to install with sqlite only, run:
 cargo install diesel_cli --no-default-features --features sqlite
 ```
 
-For projects that depend on diesel,
-you can specify which backends are required in the `Cargo.toml`.
+To install Diesel CLI with support for more than one backend,
+use spaces to separate the features.
```
killercup: My spidey senses are tingling. There should be a link here. To the cargo-install docs. But… where are they? [Here](https://doc.rust-lang.org/nightly/cargo/)? [H̸ere](https://github.com/rust-lang/cargo/blob/19cfb6742f5ff7a53220056dccdfd5626b647428/src/etc/man/cargo-install.1)?? [Here̴](https://github.com/rust-lang/cargo/blob/75f8541d32b276ef0764535fcfb63befbc285126/src/bin/install.rs#L40-L103)??? Cargo team, _where_ a̢re the `cargo ins̸ţall` docs?̵?̶!́?

PR #1338, opened 2017-12-02
Document the MySQL module

sgrif:
Though it would be keeping with MySQL tradition to leave this module
undocumented, I figure it's better to document it.


sgrif:
TBH I would actually prefer to just `#[allow(missing_docs)]` in a lot of these places. Unfortunately, there's no way for me to say "allow this module to have no module level doc-comment" without also allowing all items in the module to omit documentation.

PR #1340, opened 2017-12-02
Update out of date docs in the `types` module

sgrif:
I've gone through the module and read all existing doc comments, and
updated things that have either become out of date, or which needed
additional wording.

A lot of these docs predated MySQL support or even SQLite support, and
said things which made no sense in the context of those backends. Others
were just really badly worded (for example `SqlOrd` implied that it
affected what could be passed to `order_by`)

### DIFF
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -43,7 +43,8 @@ use row::Row;
 use std::error::Error;
 use std::io::{self, Write};
 
-/// The boolean SQL type. On SQLite this is emulated with an integer.
+/// The boolean SQL type. On backends without a native boolean type this is
+/// emulated with the smallest supported integer.
 ///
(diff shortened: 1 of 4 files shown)


diesel/src/types/mod.rs
```diff
@@ -368,16 +360,39 @@ pub trait SingleValue {}
...
+/// When possible, implementations of this trait should prefer to use an
+/// existing implementation, rather than reading from `bytes`. (For example, if
+/// you are implementing this for an enum which is represented as an integer in
+/// the database, prefer `i32::from_sql(bytes)` over reading from `bytes`
+/// directly)
+///
+/// ### Backend specific details
+///
+/// - For PostgreSQL, the bytes will be sent using the binary protocol, not text.
+/// - For SQLite, the acutal type of `DB::RawValue` is private API. All
```
Eijebong: `actual`

PR #1343, opened 2017-12-03
Unify most of the traits in `query_dsl` into a single trait

sgrif:
This moves most traits that represent a single method into a new
`QueryBuilder` trait. While I said I didn't want to do any more big
changes until post-1.0, this change is primarily motivated by making
our documentation easier to browse. This change will need a 0.99.2
release.

In addition to making our docs easier to read, this may improve our
error messages slightly in a few places that required using UFCS to get
a decent error message. `SelectStatement`, `BoxedSelectStatement`, and
`Table` now unconditionally implement this trait, so the method will
always be present and error messages will be about why the where clause
failed.

That said, those cases will be pretty small. Most places where you end
up doing the UFCS trick are related to `.execute` and `.load`. I think
it would make sense to do a similar change there (maybe we call it
`RunQueryDsl`?)

There were two traits which did not make sense to keep around --
`CountDsl` and `JoinDsl`. There's perhaps an argument for keeping
`CountDsl` separate, as you could have relied on it from generic code.
However, any code that was using `JoinDsl` already would have needed the
where clauses required to use it from `QueryDsl`.

I've opted to add an inherent method for `filter` back onto update and
delete statements, rather than implementing `QueryDsl` for them.
`filter` is the only method that would ever work on those types, and
having it be an inherent method makes the docs for those methods more
visible.

I've opted to leave the `impl FilterDsl` in place though, because having
some generic code which adds a where clause to both select statements
and update statements is totally a reasonable thing to do 🙃.


sgrif:
The compile test failures are mostly due to https://github.com/rust-lang/rust/issues/34260. We would have to go through major contortions to avoid the overflow here, and I really think this is a bug in rustc. For now I'm going to comment out the bad lines and push to get this fixed in Rust. (All places that the overflow happens should be failing with `E0277` -- There are also failures due to us expecting `E0599` which are legit and part of the logic behind this change)

CHANGELOG.md
```diff
@@ -12,15 +12,26 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   table. If the `#[table_name]` annotation is left off, you must annotate each
   field with `#[sql_type = "Integer"]`
 
+
```
Eijebong: Hu ? ✂️

diesel/src/query_dsl/mod.rs
```diff
@@ -29,18 +35,559 @@ mod order_dsl;
...
+    /// ```
+    #[cfg(feature = "postgres")]
+    fn distinct_on<Expr>(self, expr: Expr) -> DistinctOn<Self, Expr>
+    where
+        Self: methods::DistinctOnDsl<Expr>,
+    {
+        methods::DistinctOnDsl::distinct_on(self, expr)
+    }
+
+    // FIXME: Needs usage example and doc rewrite
```
Eijebong: Maybe we should open an issue for that ?

diesel/src/query_dsl/mod.rs
```diff
@@ -29,18 +35,559 @@ mod order_dsl;
...
+        Self: methods::SelectDsl<CountStar>,
+    {
+        use dsl::count_star;
+
+        QueryDsl::select(self, count_star())
+    }
+
+    /// Join tow tables using a SQL `INNER JOIN`.
+    ///
+    ///
```
Eijebong: That's a lot of new lines :o

diesel/src/query_dsl/mod.rs
sgrif: I'm going to do a pass on this module soon.

CHANGELOG.md
sgrif: Rebasing is hard

PR #1346, opened 2017-12-03
Remove some keywords from CLI and Infer Schema

sgrif:
If people are browsing the "database" keyword on crates.io, these are
not the crates that they're looking for. I'm going to encourage some
other libs that have this keyword unneccessarily to do the same.


notryanb:
Has this caused problems in the past (anyone complained about those crates showing up) or do you think it just generates "noise" for `crates.io` search results? Either way, it makes sense for these specific crates.

sgrif:
Just noise

PR #1349, opened 2017-12-03
Upload docs to `/edge` instead of the root

sgrif:
This moves the documentation off of the root of `docs.diesel.rs`, and
into a subdirectory called `edge`. After this PR is merged,
`docs.diesel.rs` will redirect to the docs for the latest release, and
the docs for the master branch will live at `docs.diesel.rs/edge`.

Before merging this, I will set up the gh-pages branch to contain a
CNAME file, an `index.html` which redirects to 0.99.1, and the docs for
0.99.0 and 0.99.1 in appropriate directories. At the moment I intend to
have updating docs for releases be a manual process, and I don't think
we need to include releases prior to 0.99.0.

This will break *all* existing links to `docs.diesel.rs/stuff`, which is
probably fine. The website should be updated to always link to the docs
for the version it was written against.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,11 +82,7 @@ env:
 after_success:
 - |
   if [[ "$TRAVIS_RUST_VERSION" == stable ]]; then
-    (cd diesel && travis-cargo doc -- --features "postgres sqlite mysql extras")
-    mkdir diesel/target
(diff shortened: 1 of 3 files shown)


sgrif:
I still want to do this at some point, but I'll revisit it at a later date.

PR #1350, opened 2017-12-04
Merge `ExecuteDsl` and friends into a single trait.

sgrif:
This applies the same strategy as #1343 to `ExecuteDsl`, `LoadDsl`, and
`FirstDsl`. Ultimately I didn't think that `LoadDsl` and `FirstDsl` were
pulling their weight enough to continue to exist.

This will dramatically improve our error messages in some cases, since
we should never see `no method named "first" for <INSERT PAGE LONG TYPE
HERE OMG THIS IS PROOF DIESEL IS THE MOST COMPLEX THING EVER RUN FOR
YOUR LIFE>` again. Users should see the actual issue that caused
the query to be invalid. Many of those errors still need to be improved,
but it's a step in the right direction. This ends up killing the UFCS
trick.

There was one unfortunate side effect, which is that I had to convert an
impl which previously applied to all backends which support the
returning keyword to one that is PG specific. This will go away with
either specialization (maybe) or disjointness on associated types.

I've opted to change `ExecuteDsl` to not take a `self` parameter here,
as it seemed like most code using it with a where clause was also
imporitng `RunQueryDsl`.

This also forces an explicit impl of `RunQueryDsl` on types which we
want to be able to execute as a full query. Previously `Query` did this
for cases where we were using `get_result`, but it's actually possible
to do `id.eq(1).execute(&conn)` before this commit (which would of
course error).

### DIFF
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -94,7 +94,7 @@ use std::fs::DirEntry;
 use std::io::{stdout, Write};
 
 use diesel::expression_methods::*;
-use diesel::{Connection, ExecuteDsl, QueryDsl, QueryResult};
+use diesel::{Connection, QueryDsl, QueryResult, RunQueryDsl};
 use self::schema::__diesel_schema_migrations::dsl::*;
 
 use std::env;
@@ -216,7 +216,7 @@ pub fn revert_migration_with_version<Conn: Connection>(
 ) -> Result<(), RunMigrationsError> {
     migration_with_version(migrations_dir, ver)
         .map_err(|e| e.into())
-        .and_then(|m| revert_migration(conn, m, output))
+        .and_then(|m| revert_migration(conn, &m, output))
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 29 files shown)


Eijebong:
`imporitng`

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -216,7 +216,7 @@ pub fn revert_migration_with_version<Conn: Connection>(
 ) -> Result<(), RunMigrationsError> {
     migration_with_version(migrations_dir, ver)
         .map_err(|e| e.into())
-        .and_then(|m| revert_migration(conn, m, output))
+        .and_then(|m| revert_migration(conn, &m, output))
```
Eijebong: Why is this here ?

sgrif: Because Rust is super inconsistent about whether it runs lints or not, and suddenly this clippy lint (which should have always been failing) is now failing as a result of these unrelated changes.

Eijebong: Lol, okay'

PR #1353, opened 2017-12-05
Document all items in the `types` module

sgrif:
I'm actually surprised this was it. I expected there to be many more
undocumented items in here than this.

### DIFF
--- a/diesel/src/types/fold.rs
+++ b/diesel/src/types/fold.rs
@@ -2,7 +2,9 @@ use types::{self, NotNull};
 
 /// Represents SQL types which can be used with `SUM` and `AVG`
 pub trait Foldable {
+    /// The SQL type of `sum(this_type)`
     type Sum;
+    /// The SQL type of `avg(this_type)`
     type Avg;
 }
 
--- a/diesel/src/types/mod.rs
+++ b/diesel/src/types/mod.rs
@@ -171,6 +171,10 @@ pub type Float8 = Double;
 /// [`bigdecimal::BigDecimal`]: /bigdecimal/struct.BigDecimal.html
 #[derive(Debug, Clone, Copy, Default)]
 pub struct Numeric;
+
+/// The `DECIMAL` SQL type
+///
+/// This type is always an alias for `NUMERIC` on all backends.
 pub type Decimal = Numeric;
 
 #[cfg(not(feature = "postgres"))]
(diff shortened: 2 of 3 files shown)


diesel/src/types/fold.rs
```diff
@@ -2,7 +2,9 @@ use types::{self, NotNull};
 
 /// Represents SQL types which can be used with `SUM` and `AVG`
 pub trait Foldable {
+    /// The SQL type of `sum(this_type)`
```
weiznich: Maybe transform the function call into a link to the actual function?

diesel/src/types/mod.rs
```diff
@@ -334,17 +348,38 @@ pub use pg::types::sql_types::*;
 #[cfg(feature = "mysql")]
 pub use mysql::types::*;
 
+/// Indicates that a SQL type exists for a backend.
 pub trait HasSqlType<ST>: TypeMetadata {
+    /// Fetch the metadata for the given type
+    ///
+    /// This method may use `lookup` to do dynamic runtime lookup. Implementors
+    /// of this method should not do dynamic lookup unless absolutely necessary
```
weiznich: Maybe this should include a example case where dynamic lookup is appropriated? (Some sentence mentioning custom sql type definitions?)

diesel/src/types/mod.rs
```diff
@@ -380,6 +421,7 @@ impl<T: NotNull + SingleValue> SingleValue for Nullable<T> {}
 ///
 /// [`MysqlType`]: ../mysql/enum.MysqlType.html
```
weiznich: We should mention somewhere here that all types that implement `FromSql` also must also implement `FromSqlRow` (and link to `FromSqlRow`)

diesel/src/types/mod.rs
```diff
@@ -396,6 +438,7 @@ pub trait FromSqlRow<A, DB: Backend + HasSqlType<A>>: Sized {
     /// the number of times you would call `row.take()` in `build_from_row`
```
weiznich: I think transform one of the places `FromSql` into a link would improve the trait documentation. (Because if someone implements this he will go forth and back between the `FromSql` and `FromSqlRow` documentation)

diesel/src/types/mod.rs
sgrif: I think that should be left to the documentation for the backend specific lookup types. I prefer not to put PG specifics here.

PR #1360, opened 2017-12-09
Clean up the docs for `.distinct()`

sgrif:
This example was noisy and I found it hard to read with all the
`.clone()`s. Even though `Vec<&str>: PartialEq<Vec<String>>`, that
does not apply to `Result`, so I had to restructure the test to allow
`?` (since we want to avoid `unwrap` in our docs when possible)

### DIFF
--- a/diesel/src/query_dsl/distinct_dsl.rs
+++ b/diesel/src/query_dsl/distinct_dsl.rs
@@ -11,10 +11,10 @@ use expression::SelectableExpression;
 ///
 /// [`QueryDsl`]: ../trait.QueryDsl.html
 pub trait DistinctDsl {
-    /// Query with DISTINCT added
+    /// The type returned by `.distinct`
     type Output;
 
-    /// Adds `DISTINCT` to to the query.
+    /// See the trait documentation.
     fn distinct(self) -> Self::Output;
 }
 
@@ -25,7 +25,6 @@ where
 {
     type Output = <T::Query as DistinctDsl>::Output;
 
-    /// Returns query with DISTINCT added
     fn distinct(self) -> Self::Output {
         self.as_query().distinct()
     }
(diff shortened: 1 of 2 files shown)


diesel/src/query_dsl/distinct_dsl.rs
```diff
@@ -40,10 +39,10 @@ where
 /// [`QueryDsl`]: ../trait.QueryDsl.html
 #[cfg(feature = "postgres")]
 pub trait DistinctOnDsl<Selection> {
-    /// Query with `DISTINCT ON()` added
+    /// The type returned by `.distinct_on`
     type Output;
 
-    /// Should return query with `DISTINCT ON` added
+    /// See the trait documentation
```
killercup: This is pretty non-documentary. If this is what you really want people to do, at least include a link

diesel/src/query_dsl/distinct_dsl.rs
```diff
@@ -11,10 +11,10 @@ use expression::SelectableExpression;
 ///
 /// [`QueryDsl`]: ../trait.QueryDsl.html
 pub trait DistinctDsl {
-    /// Query with DISTINCT added
+    /// The type returned by `.distinct`
     type Output;
 
-    /// Adds `DISTINCT` to to the query.
+    /// See the trait documentation.
```
killercup: This is pretty non-documentary. If this is what you really want people to do, at least include a link

diesel/src/query_dsl/distinct_dsl.rs
sgrif: I would be linking to the page they are currently on.

diesel/src/query_dsl/distinct_dsl.rs
sgrif: I'd be linking to the page the user is currently on. There are plenty of examples in libstd where a module that basically only has one trait puts all its documentation in the module docs, and then has the trait documented as "see the module docs". I figured the same applies reasonably well to traits with only one method (in general the goal here being to have the docs render as high up the page as possible).

TL;DR: Given that the trait only has one method, and the trait docs are written as though they were the docs for this method, I think this is fine. The only real alternative is moving the trait docs to this method, and linking to them in the trait docs, but the trait docs render higher up on the page.

PR #1362, opened 2017-12-09
Rewrite the `select` docs

sgrif:
While most of the bounds related here are really more relevant to those
traits than `select`, but this is where users will encounter it, so I
think it's worth mentioning here. I'm unsure if we should mention the
aggregate/non-aggregate stuff here or not, but I've left it off for now.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -158,12 +158,120 @@ pub trait QueryDsl: Sized {
     }
 
     // FIXME: Needs usage example and doc rewrite
-    /// Sets the select clause of a query.
+    /// Adds a `SELECT` clause to the query.
     ///
-    /// If there was already a select clause, it will be overridden. The
-    /// expression passed to `select` must actually be valid for the query (only
-    /// contains columns from the target table, doesn't mix aggregate +
-    /// non-aggregate expressions, etc).
+    /// If there was already a select clause present, it will be overridden.
+    /// For example, `foo.select(bar).select(baz)` will produce the same
+    /// query as `foo.select(baz)`.
+    ///
+    /// By default, the select clause will be roughly equivalent to `SELECT *`
+    /// (however, Diesel will list all columns to ensure that they are in the
+    /// order we expect).
+    ///
+    /// `select` has slightly stricter bounds on its arguments than other
+    /// methods. In particular, when used with a left outer join, `.nullable`
+    /// must be called on columns that come from the right side of a join. It
+    /// can be called on the column itself, or on an expression containing that
+    /// column. `title.nullable()`, `lower(title).nullable()`, and `(id,
+    /// title).nullable()` would all be valid.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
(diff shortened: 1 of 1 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -163,12 +163,120 @@ pub trait QueryDsl: Sized {
...
+    /// let all_users = users.load::<(i32, String)>(&connection)?;
+    /// let all_names = users.select(name).load::<String>(&connection)?;
+    ///
+    /// assert_eq!(vec![(1, String::from("Sean")), (2, String::from("Tess"))], all_users);
+    /// assert_eq!(vec!["Sean", "Tess"], all_names);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    ///
+    /// ### When used with a left join
```
killercup: 2nd level headline?

diesel/src/query_dsl/mod.rs
```diff
@@ -163,12 +163,120 @@ pub trait QueryDsl: Sized {
...
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::users::dsl::*;
+    /// #     let connection = establish_connection();
+    /// // By default, all columns will be selected
+    /// let all_users = users.load::<(i32, String)>(&connection)?;
+    /// let all_names = users.select(name).load::<String>(&connection)?;
+    ///
+    /// assert_eq!(vec![(1, String::from("Sean")), (2, String::from("Tess"))], all_users);
+    /// assert_eq!(vec!["Sean", "Tess"], all_names);
```
killercup: for symmetry with the next example, i'd write it it like load/assert/load/assert

diesel/src/query_dsl/mod.rs
```diff
@@ -163,12 +163,120 @@ pub trait QueryDsl: Sized {
...
-    /// expression passed to `select` must actually be valid for the query (only
-    /// contains columns from the target table, doesn't mix aggregate +
-    /// non-aggregate expressions, etc).
+    /// If there was already a select clause present, it will be overridden.
+    /// For example, `foo.select(bar).select(baz)` will produce the same
+    /// query as `foo.select(baz)`.
+    ///
+    /// By default, the select clause will be roughly equivalent to `SELECT *`
+    /// (however, Diesel will list all columns to ensure that they are in the
+    /// order we expect.)
```
killercup: swap ) and .

diesel/src/query_dsl/mod.rs
```diff
@@ -163,12 +163,120 @@ pub trait QueryDsl: Sized {
...
+    ///
+    /// `select` has slightly stricter bounds on its arguments than other
+    /// methods. In particular, when used with a left outer join, `.nullable`
+    /// must be called on columns that come from the right side of a join. It
+    /// can be called on the column itself, or on an expression containing that
+    /// column. `title.nullable()`, `lower(title).nullable()`, and `(id,
+    /// title).nullable()` would all be valid.
+    ///
+    /// # Example
```
killercup: plural

PR #1371, opened 2017-12-12
Document all items on `RunQueryDsl`

sgrif:


diesel/src/query_dsl/mod.rs
```diff
@@ -752,6 +870,44 @@ pub trait RunQueryDsl<Conn>: Sized {
...
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test();
+    /// # }
+    /// #
+    /// # #[cfg(feature = "postgres")]
```
weiznich: There should be a visible sign that this code only works with postgres.

diesel/src/query_dsl/mod.rs
```diff
@@ -752,6 +870,44 @@ pub trait RunQueryDsl<Conn>: Sized {
     /// `Err(NotFound)` will be returned if the query affected 0 rows. You can
     /// call `.optional()` on the result of this if the command was optional to
     /// get back a `Result<Option<U>>`
+    ///
+    /// When this method is called on an insert, update, or delete statement,
+    /// it will implicitly add a `RETURNING *` to the query,
+    /// unless a returning clause was already specified.
+    ///
+    /// # Example
```
weiznich: I think a second example showing how to use `get_result` with a select query would be great. Maybe add this as first example and then add a caption like postgres specific use case, before the existing example.

diesel/src/query_dsl/mod.rs
```diff
@@ -760,6 +916,9 @@ pub trait RunQueryDsl<Conn>: Sized {
     }
 
     /// Runs the command, returning an `Vec` with the affected rows.
+    ///
+    /// This method is an alias for `load`, but with a name that makes more
```
weiznich: Add a actual link to `load`?

diesel/src/query_dsl/mod.rs
sgrif: This method isn't really meant to be used with select statements normally. `.first` is intended for that case. I don't think wanting to skip `LIMIT 1` is common enough to warrant an example here.

diesel/src/query_dsl/mod.rs
sgrif: I don't think we need to state "sqlite and mysql don't support returning clauses" every time we say "RETURNING" in our docs. Especially since trying to run this code on SQLite will give you "`Sqlite: SupportsReturningClause` is not satisfied"

diesel/src/query_dsl/mod.rs
weiznich: What about other queries known to return only one item? For example filtering by the primary key?

sgrif: You'd still want `LIMIT 1` there.

PR #1374, opened 2017-12-12
Rework the docs on `SaveChangesDsl`.

sgrif:
The previous example was just throwing away the results, which rather
defeats the purpose of `save_changes` entirely. I've updated the docs to
show a more meaningful use case.

### DIFF
--- a/diesel/src/query_dsl/save_changes_dsl.rs
+++ b/diesel/src/query_dsl/save_changes_dsl.rs
@@ -69,8 +69,10 @@ where
 pub trait SaveChangesDsl<Conn> {
     /// Sugar for types which implement both `AsChangeset` and `Identifiable`
     ///
+    /// On backends which support the `RETURNING` keyword,
     /// `foo.save_changes(&conn)` is equivalent to
-    /// `update(foo::table().find(foo.id())).set(&foo).get_result(&conn)`
+    /// `update(&foo).set(&foo).get_result(&conn)`.
+    /// On other backends, two queries will be executed.
     ///
     /// # Example
     ///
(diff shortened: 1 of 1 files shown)


diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -69,8 +69,10 @@ where
 pub trait SaveChangesDsl<Conn> {
     /// Sugar for types which implement both `AsChangeset` and `Identifiable`
     ///
+    /// On backends which support the `RETURNING` keyword,
     /// `foo.save_changes(&conn)` is equivalent to
-    /// `update(foo::table().find(foo.id())).set(&foo).get_result(&conn)`
+    /// `update(&foo).set(&foo).get_result(&conn)`.
+    /// On other backends, two queries will be executed.
```
killercup: Oh, these two queries are not executed in a transaction. Maybe they should be.

diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -79,30 +81,38 @@ pub trait SaveChangesDsl<Conn> {
...
-    ///     .first::<Animal>(&connection);
-    ///
-    /// assert_eq!(Ok(Animal { id: 2, species: "solifuge".to_string(), legs: 10, name: None }), changed_animal);
+    /// let form = AnimalForm { id: 2, name: "Super scary" };
+    /// let changed_animal = form.save_changes(&connection)?;
+    /// let expected_animal = Animal {
+    ///     id: 2,
+    ///     species: String::from("spider"),
+    ///     legs: 8,
+    ///     name: Some(String::from("Super scary")),
```
killercup: Really? "Super scary"? That's what you'd name your pet spider? I'd go for something like "Charlotte".

diesel/src/query_dsl/save_changes_dsl.rs
sgrif: It only executes one query that modifies data (the second query is a select). A transaction is only really needed if you're executing two or more queries that modify things, which you'd like to have be atomic. If that were the case, you would need an outer transaction regardless of backend.

diesel/src/query_dsl/save_changes_dsl.rs
sgrif: Originally I was going with "might eat your fucking face if you take your eyes off it"

diesel/src/query_dsl/save_changes_dsl.rs
killercup: I was more concerned with the data being changed by another user/connection while you're waiting for the second query :)

sgrif: This is a post 1.0 thing either way. I'll open an issue.

diesel/src/query_dsl/save_changes_dsl.rs
killercup: Damn, what kind of spiders do you have in New Mexico…

sgrif: Never trust something with more than 2 eyes

sgrif: Feel free to change the name, I'm just being lazy

PR #1375, opened 2017-12-12
Document all items in `query_dsl`.

sgrif:


diesel/src/query_dsl/join_dsl.rs
```diff
@@ -44,34 +44,35 @@ where
...
+/// # #[macro_use] extern crate diesel;
+/// # include!("../doctest_setup.rs");
+/// # use schema::{users, posts};
+/// #
+/// # fn main() {
+/// #     let connection = establish_connection();
+/// let data = users::table
+///     .left_join(posts::table.on(
+///         users::id.eq(posts::user_id).and(
+///             posts::title.eq("My first post"))
```
killercup: Btw, I'd format this as

```rust
users::id.eq(posts::user_id)
.and(posts::title.eq("My first post"))
```

diesel/src/query_dsl/save_changes_dsl.rs
```diff
@@ -66,45 +66,46 @@ where
...
+/// #    legs: i32,
+/// #    name: Option<String>,
+/// # }
+/// #
+/// # fn main() {
+/// #     use animals::dsl::*;
+/// #     let connection = establish_connection();
+/// let mut spider = animals.filter(species.eq("spider"))
+///     .first::<Animal>(&connection)
+///     .expect("Too scary to load");
```
killercup: maybe change this test to use a wrapper fn and `?` too?

diesel/src/query_dsl/join_dsl.rs
sgrif: This is just moved from the other location. I wonder if we can set up a way to run our docs code through rustfmt.

diesel/src/query_dsl/save_changes_dsl.rs
sgrif: This is just moved from its original location, and I also have a PR up that rewrites this example

PR #1379, opened 2017-12-12
(derive) Use absolute path to Result in QueryableByName derive

### LINKED ISSUE #1372: `#[derive(QueryableByName)]` fails to compile if there is a type called `Result` in scope
I really wish people wouldn't shadow types from Rust's prelude, but we should probably explicitly reference `std::result::Result`

theduke:
This prevents name collisions when a custom Result type is in scope.

Fixes #1372


theduke:
Doing a PR for this feels kind of emberassingly pointless. ;)

sgrif:
We tend to leave "embarrassingly simple" things which are not critical open so that first time contributors have something easy to start with. I appreciate the PR though.

theduke:
Actually, then let's close this and wait for someone else to find it. ;)

PR #1382, opened 2017-12-13
Document `Queryable`

sgrif:
This is one of the traits that felt most likely to be manually
implemented (if it was never manually implemented, it would have no
reason to exist alongside `FromSqlRow`). Because of that, I figured we
should have at least one example showing how you'd implement it (focused
on how I'd write it in my app, not the code we generate which is
substantially more generic).

### DIFF
--- a/diesel/src/query_source/mod.rs
+++ b/diesel/src/query_source/mod.rs
@@ -16,15 +16,104 @@ use types::{FromSqlRow, HasSqlType};
 pub use self::joins::JoinTo;
 pub use self::peano_numbers::*;
 
-/// Trait indicating that a record can be queried from the database. This trait
-/// can be derived automatically using `#[derive(Queryable)]`. This trait can
-/// only be derived for structs, not enums.
+/// Trait indicating that a record can be queried from the database.
+///
+/// Types which implement `Queryable` represent the result of a SQL query. This
+/// does not necessarily mean they represent a single database table.
+///
+/// This trait can be derived automatically using `#[derive(Queryable)]`. This
+/// trait can only be derived for structs, not enums.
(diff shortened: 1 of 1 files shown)


diesel/src/query_source/mod.rs
```diff
@@ -16,15 +16,104 @@ use types::{FromSqlRow, HasSqlType};
...
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// let first_user = users.first(&connection)?;
+/// let expected = User { id: 1, name: "Sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+///
+/// If we want to do additional work during deserializaiton, we can implement
```
notryanb: <3

diesel/src/query_source/mod.rs
```diff
@@ -16,15 +16,104 @@ use types::{FromSqlRow, HasSqlType};
...
 
-/// Trait indicating that a record can be queried from the database. This trait
-/// can be derived automatically using `#[derive(Queryable)]`. This trait can
-/// only be derived for structs, not enums.
+/// Trait indicating that a record can be queried from the database.
+///
+/// Types which implement `Queryable` represent the result of a SQL query. This
+/// does not necessarily mean they represent a single database table.
+///
+/// This trait can be derived automatically using `#[derive(Queryable)]`. This
```
notryanb: Should we mention `#[macro_use] extern crate diesel;` anywhere?

sgrif: I think we can safely assume that is always present at this level

PR #1383, opened 2017-12-13
Document everything in the `query_source` module

sgrif:
As part of this commit I also reviewed all existing docs to make sure
they were still accurate (I skipped over really recent ones like
`QueryableByName`).

Generally most things were documented and up to date. Some of the join
related traits had cruft from when `#[belongs_to]` generated
`joinable!`.

Even though `QuerySource` is mostly internal, I felt it was important
enough to warrant documenting. Plugins may need to deal with this trait.


diesel/src/query_source/joins.rs
```diff
@@ -153,9 +153,19 @@ where
...
-/// Indicates that two tables can be used together in a JOIN clause.
-/// Implementations of this trait will be generated for you automatically by
-/// the [association annotations](../associations/index.html) from codegen.
+/// Indicates that two tables can be joined without an explicit `ON` clause.
+///
+/// Implementations of this trait are generated by invoking [`joinable!`].
+/// Implementing this trait means that you can call
+/// `left_table.inner_join(right_table)`, without supplying the `ON` clause
+/// explicitly. To join two tables which do not implement this trait, you will
+/// need to call [`.on`].
```
weiznich: I think this sentence should state on which part of the query `on` should be called.

diesel/src/query_source/mod.rs
```diff
@@ -23,8 +23,12 @@ pub trait Queryable<ST, DB>
 where
     DB: Backend + HasSqlType<ST>,
 {
+    /// The Rust type you'd like to map from.
```
weiznich: Those changes are also part of the document `Queryable` pull request.

sgrif: I probably copied them over before I just decided to `#[allow(missing_docs)]`

diesel/src/query_source/joins.rs
sgrif: Do you think that is sufficiently covered by the documentation for `.on`?

PR #1388, opened 2017-12-13
Document all items in `query_builder`

sgrif:
I've reviewed everything rendered in this module, and updated it if
needed, documented if it was undocumented, or hid it if it was not
intentionally public.

I've added `#[allow(missing_docs)]` on modules documented by other PRs,
in a way that will cause a merge conflict and force me remove whatever
annotation is there once both are merged.

`QueryId` is probably the one I'm most concerned about the docs for.
It's the one that is hardest to understand I think. Really the answer
for 99% of people is "do `impl_query_id!(YourType<A, B, C>);` but
knowing whether to put `noop: ` there or not is important. (As an aside,
I will 100% be replacing that macro with a derive in 1.1)


diesel/src/macros/query_id.rs
```diff
@@ -1,5 +1,36 @@
...
+///
+/// ```rust,ignore
+/// impl<Left, Right> QueryId for And<Left, Right> {
+///     type QueryId = ();
+///
+///     const HAS_STATIC_QUERY_ID: bool = false;
+/// }
+/// ```
 #[macro_export]
-#[doc(hidden)]
```
weiznich: If we want to replace this macro with a custom derive in 1.1 we should hide the documentation about it, to "remove" the item from the public api.

diesel/src/query_builder/mod.rs
```diff
@@ -10,7 +10,11 @@ mod query_id;
 mod clause_macro;
 
 mod ast_pass;
+<<<<<<< d4704b75743cb3050c02dd634e631a52ffbf9f02
```
weiznich: Resolve this merge conflict :wink:

diesel/src/query_builder/mod.rs
```diff
@@ -29,7 +33,7 @@ mod select_clause;
 mod select_statement;
 mod sql_query;
 pub mod where_clause;
-#[deny(missing_docs)]
+mod where_clause;
```
weiznich: Why do we need `mod where_clause` twice?

diesel/src/query_builder/query_id.rs
```diff
@@ -1,11 +1,54 @@
 use std::any::{Any, TypeId};
 use super::QueryFragment;
 
+/// Uniquely identifies queries by their type for the purpose of prepared
+/// statement caching.
+///
+/// All types which implement `QueryFragment` should also implement this trait
+/// (It is not an actual supertrait of `QueryFragment` for boxing purposes).
+///
+/// This trait is generally implemented by invoking the [`impl_query_id!`]
```
weiznich: As written above: If we want to replace this macro, we should not mention it in the documentation.

diesel/src/query_builder/select_statement/mod.rs
```diff
@@ -9,6 +9,8 @@
 //! Of: Offset Clause
 //! G: Group By Clause
 //! FU: For Update Clause
+#![allow(missing_docs)] // The missing_docs lint triggers even though this is hidden
+
```
weiznich: This empty line should be removed.

diesel/src/macros/query_id.rs
sgrif: We can deprecate the macro in 1.1.

PR #1390, opened 2017-12-14
Update documentation for all items in `expression_methods`

sgrif:
In addition to normal doc cleanup, I've hidden the submodules of
`expression_methods`, as well as `pg::expression_methods`. I've also
made sure that we consistently prefix the PG expression methods with
`Pg`.

### DIFF
--- a/diesel/src/expression_methods/bool_expression_methods.rs
+++ b/diesel/src/expression_methods/bool_expression_methods.rs
@@ -3,8 +3,40 @@ use expression::operators::{And, Or};
 use expression::{AsExpression, Expression};
 use types::Bool;
 
+/// Methods present on boolean expressions
 pub trait BoolExpressionMethods: Expression<SqlType = Bool> + Sized {
     /// Creates a SQL `AND` expression
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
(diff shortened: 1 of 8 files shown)


diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -3,17 +3,82 @@ use expression::operators::{And, Or};
...
+    ///         (species.eq("ferret"), legs.eq(4), name.eq("Freddy")),
+    ///         (species.eq("ferret"), legs.eq(4), name.eq("Jack")),
+    ///     ])
+    ///     .execute(&connection)?;
+    ///
+    /// let data = animals.select((species, name))
+    ///     .filter(species.eq("ferret").or(name.eq("Jack")))
+    ///     .load(&connection)?;
+    /// let expected = vec![
+    ///     (String::from("dog"), Some(String::from("Jack"))),
```
weiznich: I think this example would be clearer if all entries are part of the insert statement above.

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -3,6 +3,7 @@ use expression::array_comparison::{AsInExpression, In, NotIn};
 use expression::operators::*;
 use types::SingleValue;
 
+/// Methods present on all expressions, except tuples
 pub trait ExpressionMethods: Expression + Sized {
```
weiznich: We should remove the `unwrap` statements and raw sql strings in the example code of:
* `eq_any` 
* `ne_any`

sgrif: I didn't think it was worth the churn since those lines aren't rendered.

PR #1391, opened 2017-12-14
Revamp the docs in `expression::helper_types`

sgrif:
I've converted everything to links. The function links are `../dsl/fn`
instead of just `fn`, because these types are rendered both in the `dsl`
module and the `helper_types` module.

I've added a bunch of missing helper types, restructured the module so
that everything is ordered the same as they're defined on the traits
involved, to make it easier for me to ensure we have a helper type for
every method in `expression_methods`.

I've also added a bunch of helper types for the various bare functions,
and converted `not` to use our convention for function helper types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * Added `.or_filter`, which behaves identically to `.filter`, but using `OR`
   instead of `AND`.
 
+* `helper_types` now contains a type for every method defined in
+  `expression_methods`, and every function in `dsl`.
+
 ### Deprecated
 
 * *IMPORTANT NOTE* Do to [several][rust-deprecation-bug-1]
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## Unreleased
 
+### Added
+
+* `helper_types` now contains a type for every method defined in
+  `expression_methods`, and ever function in `dsl`.
```
killercup: s/ever/every

diesel/src/expression/functions/helper_types.rs
```diff
@@ -0,0 +1,24 @@
+#![allow(non_camel_case_types)]
```
killercup: What is your argument for making these types lowercase? (Maybe I missed it)

sgrif: It's what's generated by `sql_function!`

sgrif:
Shit I forgot to merge this before 1.0

Issue #1105, opened 2017-08-15
Improving API Documentation Ergonomics


lancecarlson:
I'm been going through various searches of the docs and observing what pages appear first in the list based on things I searched for in the past when I ran into issues. I think pages that appear first in the results should link to other pages with examples, or should somehow be moved in the order they appear in search (no idea if this is possible). One example of this is nullable. If you do a search for nullable, Nullable appears before nullable. To me, the nullable page is much more useful because it includes an example. Should the Nullable page ALSO include an example, or should it just link to the nullable page OR is there a way to hijack the search ranking?

Anyway, I wanted to reserve this ticket for areas where I see more of these things crop up and general questions I know I will be having as I move forward with trying to improve the docs in general.

lancecarlson:
Another suggestion I was thinking about was to have a table for each database that shows a list of the rust types that correspond with the schema types and the exact database type. Currently it is a bit cumbersome to have to click on each type to look these up. I'm sure it will also reduce the amount of questions that come through gitter about what types need to be used.

PR #1395, opened 2017-12-14
Support retrieving error codes for database errors

jethrogb:
Most databases have error codes for errors they return [1] [2] [3].
This patch extends the DatabaseErrorInformation trait to make those
codes available. The only implementation is for Postgres right now,
because the other backends don't define their own error types.

[1]: https://www.postgresql.org/docs/9.5/static/errcodes-appendix.html
[2]: https://sqlite.org/rescode.html
[3]: https://dev.mysql.com/doc/en/error-messages-server.html

### DIFF
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -129,6 +129,10 @@ impl DatabaseErrorInformation for PgErrorInformation {
     fn constraint_name(&self) -> Option<&str> {
         get_result_field(self.0.as_ptr(), ResultField::ConstraintName)
     }
+
+    fn code(&self) -> Option<i64> {
+        get_result_field(self.0.as_ptr(), ResultField::SqlState).and_then(|code| code.parse().ok())
+    }
 }
 
 /// Represents valid options to
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -120,6 +120,15 @@ pub trait DatabaseErrorInformation {
     /// Currently this method will return `None` for all backends other than
     /// PostgreSQL.
     fn constraint_name(&self) -> Option<&str>;
+
+    /// The numeric error code that was returned by the database if the backend
+    /// supports retrieving that information.
+    ///
+    /// Currently this method will return `None` for all backends other than
+    /// PostgreSQL.
+    fn code(&self) -> Option<i64> {
+        None
+    }
 }
 
 impl fmt::Debug for DatabaseErrorInformation + Send + Sync {
@@ -146,6 +155,9 @@ impl DatabaseErrorInformation for String {
         None
     }
     fn constraint_name(&self) -> Option<&str> {
+        None
+    }
+    fn code(&self) -> Option<i64> {
         None
     }
 }


jethrogb:
My immediate use case is [CockroachDB client-side transaction retries](https://www.cockroachlabs.com/docs/stable/transactions.html#client-side-transaction-retries) for which you need to check error code 40001.

sgrif:
Thank you for the pull request. I'm hesitant about exposing this, because most useful cases here should be given a corresponding variant in `DatabaseErrorKind`. If we do decide to go this route, what do you think about returning the SQLSTATE code, rather than a backend specific one?

Either way, this won't be merged until after 1.0 has been released.

jethrogb:
> what do you think about returning the SQLSTATE code, rather than a backend specific one?

I'm not sure what this distinction is.

> Either way, this won't be merged until after 1.0 has been released.

I'd recommend merging it now while you can still change the `DatabaseErrorInformation` trait without affecting compatibility.

sgrif:
> I'm not sure what this distinction is.

One is an ANSI standard that is consistent across backends. I actually didn't even think about the fact that PG has no error codes besides SQL state codes. I was thrown off by the fact that this returns `i64`. This function can't return `i64`, SQLSTATE codes are 5 character strings. e.g. `"2BP01"`

I also don't like the name `code` since other backends have "error codes" which are not SQLSTATE identifiers.

> I'd recommend merging it now while you can still change the DatabaseErrorInformation trait without affecting compatibility.

This pull request is fully backwards compatible.

jethrogb:
> One is an ANSI standard that is consistent across backends. I actually didn't even think about the fact that PG has no error codes besides SQL state codes. 
> 
> I also don't like the name code since other backends have "error codes" which are not SQLSTATE identifiers.

I see. That seems fine with me, although I can't find any evidence that SQLite implements this particular part of the standard.

> I was thrown off by the fact that this returns i64. This function can't return i64, SQLSTATE codes are 5 character strings. e.g. "2BP01"

I didn't realize this. I can easily change it if/when you want to merge this.

> This pull request is fully backwards compatible.

It adds a trait method. Implementors of `DatabaseErrorInformation` will not compile with this change without a change in the implementation.

sgrif:
> I can't find any evidence that SQLite implements this particular part of the standard

It's not part of the SQL standard, it is a separate ANSI standard (similar to information_schema). SQLite does not support it. MySQL does, and AFAIK SQL Server and Oracle do as well.

> It adds a trait method. Implementors of DatabaseErrorInformation will not compile with this change without a change in the implementation.

It adds a trait method with a default implementation. Implementors don't need to care. It can cause minor breakage in downstream code if some code is calling `x.code()` where `x` implements a different trait which also has a method called `code`. However, this is considered a minor breaking change (aka semver backwards compatible) by [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)

sgrif:
If you're still interested in working on this, feel free to open a new PR, but I'm going to go ahead and close this as it's not currently in a state that's close to being merged, and there has been no activity for several months. For this PR to be considered, it'd need to be based on the SQLSTATE standard rather than a backend specific code, and return an appropriate type. I'd also like to see some use cases that justify a generic error code as opposed to better handling of specific errors.

jethrogb:
Again, SQLSTATE seems pretty useless:

> I can't find any evidence that SQLite implements this particular part of the standard. Also it looks like about 80% of the error codes MySQL returns correspond to SQLSTATE HY000, so I'm not sure how useful that is.

PR #1396, opened 2017-12-14
Update the documentation for `associations`

sgrif:
Wow, this really hasn't been touched since 0.7...

I'm still not super happy with these docs. Specifically I don't think
I've made it clear enough exactly what `grouped_by` does. I wish there
were a way for me to demonstrate it that is a bit less visually noisy...

I think having the smaller first example will help at least. If nothing
else, these docs are now all up to date, and at least are factually
correct.

### DIFF
--- a/diesel_tests/tests/schema.rs
+++ b/diesel_tests/tests/schema.rs
@@ -60,32 +60,15 @@ impl Comment {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Queryable, Insertable, Associations)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Queryable, Insertable, Associations, Identifiable)]
 #[belongs_to(User)]
 #[belongs_to(Post)]
 #[table_name = "followings"]
+#[primary_key(user_id, post_id)]
 pub struct Following {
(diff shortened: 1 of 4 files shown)


diesel_tests/tests/schema.rs
```diff
@@ -60,34 +60,17 @@ impl Comment {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Queryable, Insertable, Associations)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Queryable, Insertable, Associations, Identifiable)]
```
notryanb: Should we follow a particular style for `derive`s (or is this rustfmt's fault...)? There are a few `derive`s in this PR that split lines like this into the following format, which group traits (by origin?). I'm fine with the inconsistency - just checking.
```
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[derive(Queryable, Insertable, Associations, Identifiable)]
```

notryanb:
@sgrif - I will close out #1305 once we can merge this - as this PR does a much better job.

sgrif:
Shit I didn't even see that there was an open PR for this. Sorry :(

PR #1398, opened 2017-12-15
Fix link, syntax highlighting

sanmai-NL:


diesel_cli/README.md
```diff
@@ -108,7 +108,7 @@ Runs the `down.sql` and then the `up.sql` for the most recent migration.
 ## `diesel print-schema`
 Prints table definitions for database schema.
 
-[pending-migrations]: https://docs.diesel.rs/diesel/migrations/fn.run_pending_migrations.html
+[pending-migrations]: https://docs.rs/migrations_internals/*/migrations_internals/fn.run_pending_migrations.html
```
killercup: Oh, wow, we published that as "migrations-internals" and not "diesel-migrations-internals"? @sgrif, wanna change that?

sgrif: I think we need to `#[doc(inline)]` some stuff, that function should be exposed only in `diesel_migrations`

sgrif: It looks like we do have `#[doc(inline)]`, but that doesn't work for functions from other crates... Looks like we'll need to do some work there

sanmai-NL:
It should be possible to close this one quickly 🙂

diesel_cli/README.md
weiznich: The correct link is now: https://docs.rs/diesel_migrations/*/diesel_migrations/fn.run_pending_migrations.html
It would be great to fix that before merging

sgrif:
Oh whoops I didn't see that the link was still wrong

PR #1400, opened 2017-12-15
Document all items in `Backend`

sgrif:
This one is a little tricky, since users may end up having to care about
`Backend`, and what it represents (mostly so they know when it's needed
as a type constraint), but most of the items in this module are only
relevant if you are implementing support for a new backend.

### DIFF
--- a/diesel/src/backend.rs
+++ b/diesel/src/backend.rs
@@ -1,9 +1,23 @@
+//! Types which represent various database backends
+
 use byteorder::ByteOrder;
 
 use query_builder::QueryBuilder;
 use query_builder::bind_collector::BindCollector;
 use types::{self, HasSqlType};
 
+/// A database backend
+///
+/// This trait represents the concept of a backend (e.g. "MySQL" vs "SQLite").
+/// It is separate from a [`Connection`](../connection/trait.Connection.html)
+/// to that backend.
+/// One backend may have multiple concrete connection implementations.
+///
+/// Implementations of this trait should not assume details about how the
+/// connection is implemented.
+/// For example, the `Pg` backend does not assume that `libpq` is being used.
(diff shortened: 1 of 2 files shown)


diesel/src/backend.rs
```diff
@@ -19,17 +33,43 @@ where
     Self: HasSqlType<types::Time>,
     Self: HasSqlType<types::Timestamp>,
 {
+    /// The concrete `QueryBuilder` implementation for this backend.
```
weiznich: Link to `QueryBuilder`

diesel/src/backend.rs
```diff
@@ -19,17 +33,43 @@ where
...
     type QueryBuilder: QueryBuilder<Self>;
+    /// The concrete `BindCollector` implementation for this backend.
```
weiznich: Link to `BindCollector`

diesel/src/backend.rs
```diff
@@ -19,17 +33,43 @@ where
...
+    ///
+    /// Most backends should use [`RawBytesBindCollector`].
+    ///
+    /// [`RawBytesBindCollector`]: ../query_builder/bind_collector/struct.RawBytesBindCollector.html
     type BindCollector: BindCollector<Self>;
+    /// The raw representation of a database value given to `FromSql`.
```
weiznich: Link to `FromSql`

diesel/src/backend.rs
sgrif: This renders underneath a link to `QueryBuilder`.

diesel/src/backend.rs
sgrif: This renders underneath a link to `BindCollector`

PR #1401, opened 2017-12-15
Document all items in `connection`.

sgrif:


diesel/src/connection/mod.rs
```diff
@@ -12,83 +14,77 @@ pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager};
...
     #[doc(hidden)]
     type TransactionManager: TransactionManager<Self>;
 
-    /// Establishes a new connection to the database at the given URL. The URL
-    /// should be a valid connection string for a given backend. See the
-    /// documentation for the specific backend for specifics.
+    /// Establishes a new connection to the database
+    ///
+    /// The argument to this method varies by backend.
+    /// See the documentation for that backend's connection class
```
weiznich: Maybe we should override exactly this documentation in the implementation of each Connection

sgrif: It doesn't render in a prominent enough location. I've been toying with the idea of deprecating `Connection::establish` and just having inherent methods for each class (would let us vary the signature too which would be nice)

PR #1403, opened 2017-12-15
Fix a bunch of incorrect documentation links

sgrif:


sgrif:
It got removed since it showed up in the results of the regex I was using to find these.

PR #1405, opened 2017-12-16
First draft of an INSERT guide

sgrif:
This guide mirrors the form of "All About Updates", but covers the
various ways to use `insert_into`.

I opted not to cover upsert in depth, as really PG upsert should have
its own guide if we feel that is useful to have. The API docs are really
comprehensive for upsert, so I think just linking to the docs is fine
for now.

Right now I only included code examples for PG, I'll port that to the
other backends once the guide is ready to be published.


sgrif:
I just realized I never mentioned a tuple like `(user_id.eq(session.get("user_id")), &post_form)`

guide_drafts/all-about-insert.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+This will generate the same SQL as if we had used a tuple.
+
+```
+INSERT INTO "users" ("name", "hair_color")
+VALUES ($1, $2) -- binds: ["Sean", "Black"]
+```
+
+If one of the fields on our tuple is `None`,
```
killercup: s/tuple/struct

or actually completely skip the "on our tuple"

guide_drafts/all-about-insert.md
```diff
@@ -0,0 +1,445 @@
...
+While there are a lot of examples in this guide,
+ultimately the only difference between various kinds of insert statements
+is the argument passed to `.values`.
+
+All examples in this guide are run as part of Diesel's test suite.
+You can find the full code examples for each backend at these links:
+
+- [`PostgreSQL`]
+- [`MySQL`]
+- [`SQLite`]
```
killercup: (This comment is just so we can see this **TODO** in the review.)

PR #1407, opened 2017-12-16
Rewrite the root docs

sgrif:
This is inspired by the module docs for `std`. The goal of these docs is
to help you find the most important modules, and generally learn where
to find things.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -1,16 +1,98 @@
 //! # Diesel
 //!
 //! Diesel is an ORM and query builder designed to reduce the boilerplate for database interactions.
+//! If this is your first time reading this documentation,
+//! we recommend you start with the [getting started guide].
+//! We also have [many other long form guides].
 //!
-//! * [Homepage](https://diesel.rs/)
-//! * [Getting Started Guide](https://diesel.rs/guides/getting-started/)
-//! * [Source Code](https://github.com/diesel-rs/diesel/)
+//! [getting started guide]: https://diesel.rs/guides/getting-started/
+//! [many other long form guides]: https://diesel.rs/guides
 //!
-//! ## Crates
+//! # Where to find things
 //!
-//! * **Diesel**: core libary, documented here
-//! * [**Diesel CLI**](https://github.com/diesel-rs/diesel/tree/master/diesel_cli): command line tool that aids in managing your database schema
+//! ## Declaring your schema
 //!
+//! So Diesel is able to validate your queries at compile time,
+//! it requires you to specify your schema in your code,
+//! which you can do with [the `table!` macro][`table!`].
+//! `diesel print-schema` or `infer_schema!` can be used
+//! to automatically generate these macro calls
+//! (by connecting to your database and querying its schema).
+//!
+//! [`table!`]: macro.table.html
+//!
+//! ## Getting started
+//!
+//! Queries usually start from either a table, or a function like [`update`].
+//! Those functions can be found [here](#functions).
+//!
+//! Diesel provides a [`prelude` module](prelude),
+//! which exports most of the typically used traits and types.
+//! We are conservative about what goes in this module,
+//! and avoid anything which has a generic name.
+//! Files which use Diesel are expected to have `use diesel::prelude::*;`.
+//!
+//! [`update`]: fn.update.html
+//!
+//! ## Constructing a query
+//!
+//! The tools the query builder gives you can be put into these three categories:
+//!
+//! - "Query builder methods" are things that map to portions of a whole query
+//!   (such as `ORDER` and `WHERE`). These methods usually have the same name
+//!   as the SQL they map to, except for `WHERE` which is called `filter` in Diesel
+//!   (To not conflict with the Rust keyword).
+//!   These methods live in [the `query_dsl` module](query_dsl).
+//! - "Expression methods" are things you would call on columns
+//!   or other individual values.
+//!   These methods live in [the `expression_methods` module](expression_methods)
+//!   You can often find these by thinking "what would this be called"
+//!   if it were a method
+//!   and typing that into the search bar
+//!   (e.g. `LIKE` is called `like` in Diesel).
+//!   Most operators are named based on the Rust function which maps to that
+//!   operator in [`std::ops`]
+//!   (For example `==` is called `.eq`, and `!=` is called `.ne`).
+//! - "Bare functions" are normal SQL functions
+//!   such as `sum`.
+//!   They live in [the `dsl` module](dsl).
+//!   Diesel only supports a very small number of these functions.
+//!   You can declare additional functions you want to use
+//!   with [the `sql_function!` macro][`sql_function!`].
+//!
+//! [`std::ops`]: //doc.rust-lang.org/stable/std/ops/index.html
+//! [`sql_function!`]: macro.sql_function.html
+//!
+//! ## Serializing and Deserializing
+//!
+//! Types which represent the result of a SQL query implement
+//! a trait called [`Queryable`].
+//!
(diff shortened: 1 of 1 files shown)


diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//! # Where to find things
 //!
-//! * **Diesel**: core libary, documented here
-//! * [**Diesel CLI**](https://github.com/diesel-rs/diesel/tree/master/diesel_cli): command line tool that aids in managing your database schema
+//! ## Declaring your schema
 //!
+//! Diesel ensures that your queries are valid at compile time.
+//! This means it needs to know what your schema is.
+//! This is done using [the `table!` macro][`table!`].
+//! This macro is usually generated by `diesel print-schema` or [`infer_schema!`]
```
notryanb: Lots of "this". Maybe reword this block to something similar to...
> Diesel ensures that your queries are valid at compile time by knowing about your schema. This is done using  [the `table!` macro][`table!`], which is usually generated by `diesel print-schema` or [`infer_schema!`]

?

killercup: Yeah. I want to keep the "it needs to know about" aspect, though.

> So Diesel is able to validate your queries at compile-time, it requires you to specify your database's schema in your code, which you can do with the `table!` macro. `diesel print-schema` or `infer_schema!` can be used to automatically generate these macro calls (by connecting to your database and parsing its schema).

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//!
+//! [`table!`]: macro.table.html
+//! [`infer_schema!`]: FIXME: Host this crate on docs.diesel.rs because docs.rs won't render docs for this crate
```
killercup: (This comment is just to make this **FIXME** show up in the comments sections so it gets fixed before merging)

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//!
+//! ## Getting started
+//!
+//! Queries usually start from either a table, or a function like [`update`].
+//! Those functions can be found [here](#functions).
+//!
+//! Diesel provides a [`prelude` module](prelude),
+//! which exports most of the most important traits and types.
```
killercup: Maybe `s/most of the most important/most of the typically used`?

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//! We are conservative about what goes in this module,
+//! and avoid anything which has a generic name.
+//! Files which use Diesel are expected to have `use diesel::prelude::*;`.
```
killercup: > are expected to

What expects this? Our derives? Macros? Maybe add this as a reason

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//! [`update`]: fn.update.html
+//!
+//! ## Constructing a query
+//!
+//! Diesel breaks things into three categories:
+//!
+//! - "Query builder methods" are things that map to portions of a whole query
+//!   (such as `ORDER` and `WHERE`). These methods usually have the same name
+//!   as the SQL they map to, except for `WHERE` which is called `filter` in Diesel.
+//!   These methods live in [the `query_dsl` module](query_dsl).
```
killercup: No semantics linefeeds in the paragraph?

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//! which exports most of the most important traits and types.
+//! We are conservative about what goes in this module,
+//! and avoid anything which has a generic name.
+//! Files which use Diesel are expected to have `use diesel::prelude::*;`.
+//!
+//! [`update`]: fn.update.html
+//!
+//! ## Constructing a query
+//!
+//! Diesel breaks things into three categories:
```
killercup: These "things" are… what exactly? "Rust items"? "Operations"? "Functions"? "Methods"?

Maybe something like

> The tools the query builder gives you can be put into these three categories:

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//!
+//! - "Query builder methods" are things that map to portions of a whole query
+//!   (such as `ORDER` and `WHERE`). These methods usually have the same name
+//!   as the SQL they map to, except for `WHERE` which is called `filter` in Diesel.
```
killercup: > (to not conflict with the Rust keyword)

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//!   These methods live in [the `query_dsl` module](query_dsl).
+//! - "Expression methods" are things you would call on columns
+//!   or other individual values.
+//!   These methods live in [the `expression_methods` module](expression_methods)
+//!   You can often find these by thinking "what would this be called"
+//!   if it were a method
+//!   and typing that into the search bar
+//!   (e.g. `LIKE` is called `like` in Diesel).
+//!   Most operators are named based on the Rust function which maps to that
+//!   operator in `std::ops`
```
killercup: Not sure how many people are familiar with the ops module, most probably only use the operators themselves. So, maybe link to <https://doc.rust-lang.org/1.22.1/std/ops/index.html>?

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//!
+//! Diesel breaks things into three categories:
+//!
+//! - "Query builder methods" are things that map to portions of a whole query
+//!   (such as `ORDER` and `WHERE`). These methods usually have the same name
+//!   as the SQL they map to, except for `WHERE` which is called `filter` in Diesel.
+//!   These methods live in [the `query_dsl` module](query_dsl).
+//! - "Expression methods" are things you would call on columns
+//!   or other individual values.
+//!   These methods live in [the `expression_methods` module](expression_methods)
```
killercup: This (and query_dsl) are weird relative links. I assume it's supposed to lead to `./expression_methods/index.html`?

diesel/src/lib.rs
```diff
@@ -1,16 +1,95 @@
...
+//! [`sql_function!`]: macro.sql_function.html
+//!
+//! ## Serializing and Deserializing
+//!
+//! Types which represent the result of a SQL query implement
+//! a trait called [`Queryable`].
+//!
+//! Diesel maps "Rust types" (e.g. `i32`) to and from "SQL types"
+//! (e.g. [`diesel::types::Integer`]).
+//! You can find all the types supported by Diesel in [the `types` module](types).
```
killercup: Maybe also mention backend specific types?

sgrif:
> I just hope they find it ;)

This is the most prominent place I can put it, so they better. >_>

diesel/src/lib.rs
sgrif: It's just convention. All our docs are written assuming that line. All our macros and derives should work without it.

diesel/src/lib.rs
sgrif: Yes. Static file servers typically render `index.html` if a directory is requested.

diesel/src/lib.rs
sgrif: The docs for the `types` module shows links to backend specific types.

diesel/src/lib.rs
killercup: Yeah, it just looked like a markdown ref link at first :)

sgrif:
I was going to wait for the build but ok :P

killercup:
Is there anything in here that is tested by CI? Did you add a link checker without me noticing? :D

sgrif:
No, I just generally wait all the time now after that time that I broke CI like 3 times in a day with PRs that "definitely aren't doing anything that could break CI"

killercup:
Pfff and you expect to learn from _your_ failures? I hardly learn from my own!

PR #1408, opened 2017-12-16
Update the tests for `avg`

### LINKED ISSUE #841: Better handling of Numeric type
Some operations like `avg()` on an integer field return `Numeric`, which isn't currently mapped to any Rust type. I didn't find out how to actually retrieve the resulting value. Is it even possible?

See #840.

sgrif:
Right now there is no support for the `Numeric` type at all on SQLite
(since SQLite has no type that maps to it). This is a problem that we
need to fix later (probably by just treating it as an alias for
`Double` there).

The CFG attrs are really funky here. This is because tests for the main
crate run with multiple backends enabled, but the doctests need to run
against a single backend, and they have the equivalent to this:

    #[cfg(feature = "postgres")]
    type DB = Pg;

    #[cfg(all(not(feature = "postgres"), feature = "sqlite"))]
    type DB = Sqlite;

    #[cfg(all(not(any(feature = "postgres", feature = "sqlite")), feature = "mysql"))]
    type DB = Mysql;

In other words, the doctests run against the first of ["pg", "sqlite",
"mysql"] that is enabled, in that order... So this test needs to run if
pg is enabled (overrides SQLite) or SQLite is not enabled (so we must be
running on a backend that will pass).

Fixes #841 since this is really the last thing there that isn't fixed. I
will open a separate issue for the SQLite problem.

### DIFF
--- a/diesel/src/expression/functions/aggregate_folding.rs
+++ b/diesel/src/expression/functions/aggregate_folding.rs
@@ -72,23 +72,48 @@ fold_function!(
     avg,
     Avg,
     "AVG",
-    "Represents a SQL `AVG` function. This function can only take types which are
+    r#"Represents a SQL `AVG` function. This function can only take types which are
 Foldable.
 
 # Examples
 
 ```rust
 # #[macro_use] extern crate diesel;
-# include!(\"../../doctest_setup.rs\");
+# include!("../../doctest_setup.rs");
 # use diesel::dsl::*;
+# #[cfg(feature = "bigdecimal")]
+# extern crate bigdecimal;
 #
 # fn main() {
-#     use schema::animals::dsl::*;
-#     let connection = establish_connection();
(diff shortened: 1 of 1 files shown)


diesel/src/expression/functions/aggregate_folding.rs
```diff
@@ -72,23 +72,48 @@ fold_function!(
     avg,
     Avg,
     "AVG",
-    "Represents a SQL `AVG` function. This function can only take types which are
+    r#"Represents a SQL `AVG` function. This function can only take types which are
```
notryanb: What does the `r#` do here or is it a typo?

killercup: This is a [raw string literal](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals). You can use it when you have lots of quotes in a string and don't want to escape them. In addition to the usual quotes you add the same number of `#`s before and after the quotes.

PR #1410, opened 2017-12-17
Fix inlining docs for diesel_migrations

weiznich:


diesel_migrations/src/lib.rs
```diff
@@ -17,7 +78,52 @@ extern crate migrations_macros;
...
+#[doc(inline)]
+pub use migrations_internals::migration_paths_in_directory;
+#[doc(inline)]
+pub use migrations_internals::migration_from;
+#[doc(inline)]
+pub use migrations_internals::find_migrations_directory;
+#[doc(inline)]
+pub use migrations_internals::search_for_migrations_directory;
+#[doc(inline)]
+pub use migrations_internals::version_from_path;
```
notryanb: docs are hidden for [this function](https://github.com/diesel-rs/diesel/blob/master/diesel_migrations/migrations_internals/src/migration.rs#L113) Does it still need to be inlined?

PR #1414, opened 2017-12-19
Explicit joinable! documentation.

alatiera:
I don't really feel qualified to write docs and I am not even sure if what I wrote is accurate.

I had trouble figuring out `joinable!` at first and Georg Semmler suggested making a PR about it.

My original problem was that I wasn't sure in what the `table(argument)` was supposed to be in the example. Turns out in my cause it specified the `ON` clause.

I don't know much about sql in general and I've got some stuff wrong probably so feel free to let me know what to change.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -992,13 +992,21 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// ```
 ///
-/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
-/// the tables and the ON clause in the following way:
+/// In the example above, the line `joinable!(posts -> users (user_id));`
+///
+/// specifies the relation of the tables and the ON clause in the following way:
+///
 /// `parent_table -> child_table (foreign_key)`
-/// Where `parent_table` is the Table with the Primary key,
-/// `child_table` is the Table with the Foreighn key,
-/// So for the Table decaration in [Associations docs](http://docs.diesel.rs/diesel/associations/index.html),
-/// The parent table would be `User`, the child table would be `Post` and the Foreighn key would be `Post.user_id`
+///
+/// * `parent_table` is the Table with the Primary key.
+///
+/// * `child_table` is the Table with the Foreighn key.
+///
+/// So given the Table decaration from [Associations docs](http://docs.diesel.rs/diesel/associations/index.html)
+///
+/// * The parent table would be `User`
+/// * The child table would be `Post`
+/// * and the Foreighn key would be `Post.user_id`
 ///
 /// For joins that do not explicitly use on clauses via [JoinOnDsl](http://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
 /// the following on clause is generated implicitly:
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -990,16 +990,20 @@ macro_rules! __diesel_table_query_source_impl {
 /// assert_eq!(implicit_on_clause_sql, explicit_on_clause_sql);
 /// # }
 ///
-/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
-/// the tables and the ON clause in the following way:
-///
-/// `table_with_foreighn_key -> table_that_foreign_key_points_to (foreign_key)`
-/// and the Sql generated is:
-/// # I am not sure to the exact sql that would be generated, Please correct me.
-/// ```sql
-/// posts JOIN users ON posts.user_id = users.id
 /// ```
 ///
+/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
+/// the tables and the ON clause in the following way:
+/// `parent_table -> child_table (foreign_key)`
+/// Where `parent_table` is the Table with the Primary key,
+/// `child_table` is the Table with the Foreighn key,
+/// So for the Table decaration in [Associations docs](http://docs.diesel.rs/diesel/associations/index.html),
+/// The parent table would be `User`, the child table would be `Post` and the Foreighn key would be `Post.user_id`
+///
+/// For joins that do not explicitly use on clauses via [JoinOnDsl](http://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
+/// the following on clause is generated implicitly:
+/// ```sql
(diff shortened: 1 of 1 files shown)


diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// assert_eq!(implicit_on_clause_sql, explicit_on_clause_sql);
 /// # }
+///
```
weiznich: The example code block should end here. (Put ``` in this line)

Also as said in the gitter conversion. I think the example above would benefit from adding the actual table! calls. (I would add them in line 978 before the call to `joinable!`)

diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
...
+/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
+/// the tables and the ON clause in the following way:
+///
+/// `table_with_foreighn_key -> table_that_foreign_key_points_to (foreign_key)`
```
weiznich: table_with_foreig**h**n_key -> table_with_foreign_key

diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// assert_eq!(implicit_on_clause_sql, explicit_on_clause_sql);
 /// # }
+///
+/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
```
weiznich: in line -> the line

diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
...
+/// the tables and the ON clause in the following way:
+///
+/// `table_with_foreighn_key -> table_that_foreign_key_points_to (foreign_key)`
+/// and the Sql generated is:
```
weiznich: I would use a different formulation for this part of the documentation:

For joins that do not explicitly use on clauses via [`JoinOnDsl`](link to JoinOnDsl) the following on clause is generated implicitly:
```sql
post JOIN users ON posts.user_id = users.id
```

diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
...
+/// # I am not sure to the exact sql that would be generated, Please correct me.
+/// ```sql
+/// posts JOIN users ON posts.user_id = users.id
+/// ```
+///
 /// ```
```
weiznich: As noted above: This needs to be moved to line 992

sgrif:
> I don't really feel qualified to write docs and I am not even sure if what I wrote is accurate.

Keep in mind that we're always happy to answer questions. Typically this process is what results in better docs, since you know what you were looking for and/or confused by better than we can guess. :)

diesel/src/macros/mod.rs
sgrif: What do you think about calling this `child_table` and `parent_table`, and then specifying that the child table is the one with the foreign key, and the parent table is the one with the primary key that the child table points to? (This is how the association docs lay it out)

diesel/src/macros/mod.rs
sgrif: SQL should be capitalized.

diesel/src/macros/mod.rs
```diff
@@ -989,6 +989,17 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// assert_eq!(implicit_on_clause_sql, explicit_on_clause_sql);
 /// # }
+///
+/// In the example above, in line `joinable!(posts -> users (user_id));` specifies the relation of
+/// the tables and the ON clause in the following way:
+///
+/// `table_with_foreighn_key -> table_that_foreign_key_points_to (foreign_key)`
+/// and the Sql generated is:
+/// # I am not sure to the exact sql that would be generated, Please correct me.
```
sgrif: You can always inspect the SQL of a query using [`debug_query`](http://docs.diesel.rs/diesel/fn.debug_query.html)

diesel/src/macros/mod.rs
sgrif: Do you think it'd make sense to show how this would be accomplished without `joinable!` instead of demonstrating in terms of generated SQL? e.g. mentioning that `users.inner_join(posts)` becomes equivalent to `users.inner_join(posts.on(users::id.eq(posts::user_id)))`?

diesel/src/macros/mod.rs
alatiera: Yes that sound better, and less confusing. I took a look at the [associations docs](http://docs.diesel.rs/diesel/associations/index.html), do you think we could link or use those definitions of User and Post tables?

diesel/src/macros/mod.rs
alatiera: That's way better worded and informative, I will go ahead and replace it, thanks!

alatiera: I think the SQL terms would help people map diesel syntax to something they already know, where the other way would already require a familiarity with diesel. Though both could probably be mentioned.

sgrif:
@alatiera Are you still interested in finishing this?

alatiera:
@sgrif Yes sorry. I guess I missed a review and the callback was not triggered?

Sorry, still getting used to github's Pull request mechanism.

sgrif:
No problem, that's why I pinged you. It looks like the outstanding issue is that your last commit didn't address https://github.com/diesel-rs/diesel/pulls#discussion_r157878831 :)

diesel/src/macros/mod.rs
alatiera: Sorry for the late response, I missed that.

As of 4695673e434127b20deb2e4a05d179052c6e4321 it was fixed  but seems like git  didn't picked it up as changed. Sorry for that.

Here's how it renders on the latest commit (2f1575e4d6affea2844cf0b756e578127d202285):

screenshot from 2018-02-04 00-28-48

alatiera:
@sgrif It seems that https://github.com/diesel-rs/diesel/pull/1414#discussion_r157878831 still isn't marked as resolved and I am not sure how do it. Can I request a manual review?

sgrif:
@weiznich Feel free to take over this

PR #1415, opened 2017-12-19
Use fully qualified name for ::std::result::Result

fluxxu:
The following code fails to compile
`E0244: wrong number of type arguments: expected 1, found 2`

```rust
#[macro_use]
extern crate diesel;

use diesel::types::*;

type Result<T> = ::std::result::Result<T, ()>;

#[derive(Debug, QueryableByName)]
pub struct Row {
  #[sql_type = "Text"]
  pub column: String,
}
```

Related issue:
https://github.com/diesel-rs/diesel/issues/1372


sgrif:
Can you add a change log entry for this?

fluxxu:
Sorry I have to create a new one: https://github.com/diesel-rs/diesel/pull/1426 because I can't figure out how to edit this pull request.
It says `from unknown repository`.

PR #1417, opened 2017-12-19
Set order for distinct integration tests

cpick:
Explicitly set the order for the integration tests in
'diesel_tests/tests/distinct.rs'.

Previously, the tests could fail with an error like the following:

    $ bin/test integration postgres -- distinct::simple_distinct
    ...
    ---- distinct::simple_distinct stdout ----
        thread 'distinct::simple_distinct' panicked at 'assertion failed: `(left == right)`
      left: `["Sean", "Tess"]`,
     right: `["Tess", "Sean"]`', tests/distinct.rs:17:4
    note: Run with `RUST_BACKTRACE=1` for a backtrace.
    ...


sgrif:
CI failure is legit.

killercup:
Thanks!

PR #1418, opened 2017-12-19
Add `insert_or_ignore_into()` "INSERT [OR] IGNORE"

### LINKED ISSUE #1303: Support for `insert or ignore` for Sqlite.
Atm you kinda have to isolate the query and throw away the result in order to do that.

```rust
let _ = diesel::insert(foo)
    .into(atable)
    .execute(con);
```
It would be nice if you could do something like this:
```rust
diesel::insert_into(atable)
    .values(foo)
    .on_conflict()
    .do_nothing() 
    .execute(con)?;
```
Or even on_conflict().update() run [replace_into](http://docs.diesel.rs/diesel/fn.replace_into.html) instead?

I think this sort of functionality exists for Postgres with .on_conflict().do_nothing() but It would be nice if something like that existed for Sqlite too.

cpick:
Add `insert_or_ignore_into()` which provides Sqlite's "INSERT OR IGNORE" or
Mysql's "INSERT IGNORE".

When using `insert_or_ignore_into()`, if a constraint violation fails, the
database will ignore the offending row and continue processing any subsequent
rows.

This was implemented as a separate function (as opposed to something like
`insert.or(ignore)`) based on the logic layed out in #297 (0ee8363).

Fixes #1303.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -267,7 +267,8 @@ pub use prelude::*;
 #[doc(inline)]
 pub use query_builder::debug_query;
 #[doc(inline)]
-pub use query_builder::functions::{delete, insert_into, replace_into, select, sql_query, update};
+pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
+                                   select, sql_query, update};
 pub use result::Error::NotFound;
 #[doc(inline)]
 pub use types::structs::data_types;
--- a/diesel/src/query_builder/functions.rs
+++ b/diesel/src/query_builder/functions.rs
@@ -2,7 +2,7 @@ use dsl::Select;
 use expression::Expression;
 use query_dsl::methods::SelectDsl;
 use super::delete_statement::DeleteStatement;
-use super::insert_statement::{Insert, Replace};
+use super::insert_statement::{Insert, InsertOrIgnore, Replace};
 use super::{IncompleteInsertStatement, IncompleteUpdateStatement, IntoUpdateTarget,
             SelectStatement, SqlQuery};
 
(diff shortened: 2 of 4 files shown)


diesel/src/query_builder/functions.rs
```diff
@@ -276,6 +276,46 @@ pub fn insert_into<T>(target: T) -> IncompleteInsertStatement<T, Insert> {
     IncompleteInsertStatement::new(target, Insert)
 }
 
+/// Creates an `INSERT [OR] IGNORE` statement.
+///
+/// If a constraint violation fails, the database will ignore the offending
+/// row and continue processing any subsequent rows. This function is only
+/// available with MySQL and SQLite.
```
sgrif: Should we link to `on_conflict_do_nothing` here?

diesel_compile_tests/tests/compile-fail/sqlite_insert_or_ignore_cannot_be_used_on_pg.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Insertable)]
+#[table_name="users"]
+struct User {
+    id: i32,
+}
+
+fn main() {
+    let connection = PgConnection::establish("").unwrap();
+    insert_or_ignore_into(users::table)
+        .values(&User { id: 1 })
```
sgrif: Nit: It'd be more ideomatic to use `users::id.eq(1)` here (not that it matters too much for a compile-fail test

sgrif:
I manually cancelled the CI build because I want master to run so we rebuild our docs. I will re-run once the docs have been rebuilt.

sgrif:
The CI failure is legit

diesel/src/lib.rs
```diff
@@ -196,7 +196,8 @@ pub use prelude::*;
 #[doc(inline)]
 pub use query_builder::debug_query;
 #[doc(inline)]
-pub use query_builder::functions::{delete, insert_into, replace_into, select, sql_query, update};
+pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
+                                   select, sql_query, update};
```
killercup: Oh wow is that was rustfmt suggests? (It usually doesn't use visual indentation anywhere)

diesel/src/query_builder/functions.rs
```diff
@@ -276,6 +276,50 @@ pub fn insert_into<T>(target: T) -> IncompleteInsertStatement<T, Insert> {
...
+///
+/// With PostgreSQL, similar functionality is provided by [`on_conflict_do_nothing`].
+///
+/// [`on_conflict_do_nothing`]: ../diesel/query_builder/insert_statement/struct.InsertStatement.html#method.on_conflict_do_nothing
```
sgrif: `../diesel` is redundant.

diesel/src/query_builder/functions.rs
```diff
@@ -276,6 +276,50 @@ pub fn insert_into<T>(target: T) -> IncompleteInsertStatement<T, Insert> {
...
+/// # #[macro_use] extern crate diesel;
+/// # include!("../doctest_setup.rs");
+/// #
+/// # #[cfg(not(feature = "postgres"))]
+/// # fn main() {
+/// #     use schema::users::dsl::*;
+/// #     use diesel::insert_or_ignore_into;
+/// #
+/// #     let conn = establish_connection();
+/// #     conn.execute("DELETE FROM users").unwrap();
```
sgrif: Even though this line is hidden, can we avoid using `conn.execute`? This is trivial in the public API: `diesel::delete(users).execute(&conn);`

diesel/src/query_builder/functions.rs
```diff
@@ -276,6 +276,50 @@ pub fn insert_into<T>(target: T) -> IncompleteInsertStatement<T, Insert> {
...
+/// insert_or_ignore_into(users)
+///     .values((id.eq(1), name.eq("Jim")))
+///     .execute(&conn)
+///     .unwrap();
```
sgrif: We try to avoid using `unwrap` in documentation. Recent doctests follow the convention below. Can you update this test to do the same?

```rust
# #[macro_use] extern crate diesel;
# include!("doctest_setup.rs");
#
# fn main() {
#     run_test().unwrap();
# }
#
# fn run_test() -> QueryResult<()> {
// actual test goes here and uses `?`
#     Ok(())
# }
```

cpick:
Does it look like any more changes are needed?  I think I got everything so far.

sgrif:
Only thing missing is a CHANGELOG entry but I can add that.

PR #1428, opened 2017-12-20
Add a more "advanced" example

sgrif:
This app is eventually going to be used as part of the advanced querying
guide. Unlike the getting started guide, that guide will not be a step
by step "building this app" style guide, but will instead refer to some
of the code in this example for its code examples, as I want to have
them presented with broader context if the reader wants it.

Since this example is meant to be separate from the guide, I figured it
made sense to submit it on its own.

Originally I chose a CLI blog for the getting started guide because I
wanted the example purely focused on Diesel, with little to no noise
from CLI stuff or a web framework. Clearly that is not the case here. In
fact, a web app probably would be a better example. However, I'm so
amused by the idea of a "CLI blog" that I decided to double down on it
and make a multi-tenant blog wiht commenting functionality and
pagination. Of course this is a pointless exercise, since having auth
logic makes no sense when the client requires a direct database
connection...

Anyway, this example is meant to provide some examples for the following
more complex use cases:

- Various types of joins and associations. Demonstrating data with
  multiple associations that go to the same table multiple times, and is
  difficult to describe unambiguously
- When to use joins vs `belonging_to`
- Nesting queryables (see the structs used in auth)
- Dealing with structs that don't contain all fields from a table
- Demonstrating that `Queryable` represents the results of a query, not
  a single table
  - I want to restructure how `all_posts` works. I really want to have
    a struct that is used for "comment with username" and "comment
    with post title" but that doesn't work well today. I'm going to
    revisit for 1.1
- Real world use of `returning`
- Real world constraint error handling (I may change this to be
  `.on_conflict(username).do_nothing()` and handle it there so this maps
  more easily to SQLite and MySQL. If we're being pedantic I should be
  checking the constraint name in that match arm)
- Showing how to deal with "sometimes update this field" without using
  `AsChangeset`.
- Showing that various query builder methods compose together
- Generally just having more examples of non-trivial queries.

### DIFF
--- a/examples/postgres/advanced-blog-cli/migrations/00000000000000_diesel_initial_setup/down.sql
+++ b/examples/postgres/advanced-blog-cli/migrations/00000000000000_diesel_initial_setup/down.sql
@@ -1,0 +1,2 @@
+DROP FUNCTION IF EXISTS diesel_manage_updated_at(_tbl regclass);
+DROP FUNCTION IF EXISTS diesel_set_updated_at();
--- a/examples/postgres/advanced-blog-cli/src/auth.rs
+++ b/examples/postgres/advanced-blog-cli/src/auth.rs
@@ -1,0 +1,164 @@
+use bcrypt::*;
+use diesel::prelude::*;
+use diesel::{self, insert_into};
+use dotenv;
+
+use schema::users;
+
+#[derive(Debug)]
+pub enum AuthenticationError {
+    IncorrectPassword,
+    NoUsernameSet,
+    NoPasswordSet,
+    EnvironmentError(dotenv::Error),
+    BcryptError(BcryptError),
+    DatabaseError(diesel::result::Error),
+}
+
+impl From<BcryptError> for AuthenticationError {
+    fn from(e: BcryptError) -> Self {
+        AuthenticationError::BcryptError(e)
+    }
+}
+
+pub use self::AuthenticationError::{IncorrectPassword, NoPasswordSet, NoUsernameSet};
+
+#[derive(Queryable, Identifiable, Debug, PartialEq)]
+pub struct User {
+    pub id: i32,
+    pub username: String,
+}
+
+#[derive(Queryable)]
+pub struct UserWithPassword {
+    user: User,
+    password: String,
+}
+
+pub fn current_user_from_env(conn: &PgConnection) -> Result<Option<User>, AuthenticationError> {
+    let username = get_username()?;
+    let password = get_password()?;
+    find_user(conn, &username, &password)
+}
+
+pub fn register_user_from_env(conn: &PgConnection) -> Result<User, AuthenticationError> {
+    let username = get_username()?;
+    let password = get_password()?;
+    register_user(conn, &username, &password)
+}
+
+fn find_user(
+    conn: &PgConnection,
+    username: &str,
+    password: &str,
+) -> Result<Option<User>, AuthenticationError> {
+    let user_and_password = users::table
+        .filter(users::username.eq(username))
+        .select(((users::id, users::username), users::hashed_password))
+        .first::<UserWithPassword>(conn)
+        .optional()
+        .map_err(AuthenticationError::DatabaseError)?;
+
+    if let Some(user_and_password) = user_and_password {
+        if verify(password, &user_and_password.password)? {
+            Ok(Some(user_and_password.user))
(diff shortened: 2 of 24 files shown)


examples/postgres/advanced-blog-cli/src/auth.rs
```diff
@@ -0,0 +1,164 @@
...
+
+fn get_username() -> Result<String, AuthenticationError> {
+    if_not_present(dotenv::var("BLOG_USERNAME"), NoUsernameSet)
+}
+
+fn get_password() -> Result<String, AuthenticationError> {
+    if_not_present(dotenv::var("BLOG_PASSWORD"), NoPasswordSet)
+}
+
+fn if_not_present<T>(
```
killercup: That's a weird name… I'd probably write an `fn get_dotenv_or_err_with(&str, AuthenticationError)` that also includes the `dotenv::var`. Making this a very generic function doesn't really help here, as it internally depends on `dotenv::Error` anyway.

examples/postgres/advanced-blog-cli/src/auth.rs
```diff
@@ -0,0 +1,164 @@
+use bcrypt::*;
+use diesel::prelude::*;
+use diesel::{self, insert_into};
+use dotenv;
+
+use schema::users;
+
+#[derive(Debug)]
+pub enum AuthenticationError {
```
killercup: I'd move this (and the From impl) just above the helper functions (that start with `find_user`).

examples/postgres/advanced-blog-cli/src/pagination.rs
```diff
@@ -0,0 +1,69 @@
+use diesel::prelude::*;
+use diesel::query_dsl::methods::LoadQuery;
+use diesel::query_builder::*;
+use diesel::pg::Pg;
+use diesel::types::BigInt;
+
+pub trait Paginate: Sized {
```
killercup: Not that it is particularly useful here, but in the 'leading by example' sense we should probably add some doc comments to the public items in this module. At least, we should add a module level doc comment that links to the #1437 guide (once it is online)

examples/postgres/advanced-blog-cli/src/post.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub id: i32,
+    pub user_id: i32,
+    pub title: String,
+    pub body: String,
+    pub created_at: NaiveDateTime,
+    pub updated_at: NaiveDateTime,
+    pub status: Status,
+}
+
+pub enum Status {
```
killercup: I'd call more attention to this enum here and its Queryable impl as it's not just boilerplate but actually one of the 'advanced' features :) So maybe add a doc comment that describes how this is used to map `NULL` to `Draft`, and how it's a good idea to do this instead of just using  `Option<NaiveDateTime>`.

examples/postgres/advanced-blog-cli/src/schema.rs
```diff
@@ -0,0 +1,38 @@
+table! {
```
killercup: Maybe add

```rust
// generated by `diesel print-schema`
```

examples/postgres/advanced-blog-cli/migrations/00000000000000_diesel_initial_setup/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP FUNCTION IF EXISTS diesel_manage_updated_at(_tbl regclass);
```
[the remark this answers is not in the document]
sgrif: Open an issue? ;)

examples/postgres/advanced-blog-cli/src/pagination.rs
sgrif: I agree. I'm going to merge without doing this so I can publish a guide referencing code here, but I will make sure to come back and do this after we ship 1.0

examples/postgres/advanced-blog-cli/src/post.rs
sgrif: Advanced querying guide will

examples/postgres/advanced-blog-cli/src/schema.rs
sgrif: But then I can't do `diesel print-schema > src/schema.rs`

PR #1432, opened 2017-12-22
feat: add run_with_callback for embedded_migration

xinghun92:
My Application meets a problem that some data is saved in the form of BLOB in SQLite, and I want to transfer them to a new struct after some migrations run successfully. This transformation can not be done only by sql query,  so a callback is needed for each successfully ran migration.
Change-Id: I7b24bf53afbec227c6b9ed927fc25f1452687ee4

### DIFF
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -133,7 +133,7 @@ where
     Conn: MigrationConnection,
 {
     let all_migrations = try!(migrations_in_directory(migrations_dir));
-    run_migrations(conn, all_migrations, output)
+    run_migrations(conn, all_migrations, output, None)
 }
 
 /// Compares migrations found in `migrations_dir` to those that have been applied.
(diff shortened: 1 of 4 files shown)


sgrif:
Thank you for the pull request, but I don't think we're interested in this feature at this point in time. For changes which cannot be represented in SQL, you should consider a one-off task rather than a Diesel migration.

xinghun92:
But the one-off task is triggered by a specific migration. If I change the schema from A to B and need to do some data conversion, I need check whether B is in pending migrations, and do my conversion after all migrations finish in current diesel version. And this check is also in embed_migrations::run(), so I add a callback for each migration in which user can do some additional task. Is there a more elegant way to solve this problem? thanks.

xinghun92:
Is there a more elegant way to do this work with functions supported by current diesel version? Thanks.

king6cong:
I also think a more powerful migrations method is helpful, considering sometimes sql alone is not enough.

PR #1433, opened 2017-12-22
Simplify advanced blog CLI using `structopt`

killercup:
Structopt allows generating clap-based CLI arg parsers by using derives
on structs/enums. Crucially, it also handles converting arguments to
their expected types.

My goal here was to reduce the CLI parsing boilerplate (this code lives
in Diesel's example directory and should thus focus on Diesel) while
also switching to a style that will most likely be used by the next
version of clap itself.

There still is more boilerplate[^1] than I'd like to have, but at least
it's now a well-typed enum, and we can skip all the type conversion
steps. Also note that the line count is slightly skewed by my additional
help texts.

[^1]: For example: If there was a "convert enum variant name to
      kebab-case[^2] option I'd love to use it instead of all these
      `name = "foo_bar"` attributes.
[^2]: Yes we should totally use kebab-case and not snake_case here.

### DIFF
--- a/examples/postgres/advanced-blog-cli/Cargo.toml
+++ b/examples/postgres/advanced-blog-cli/Cargo.toml
@@ -6,10 +6,11 @@ authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
 [dependencies]
 bcrypt = "0.1.0"
 chrono = "0.4.0"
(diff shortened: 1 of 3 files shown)


killercup:
**Heads up:** This targets #1428's branch

PR #1436, opened 2017-12-23
Guide draft: Composing Applications with Diesel

sgrif:
This guide goes into how to write functions which return both queries
and expressions, how to use the helper types, and how to use their boxed
forms.

This was originally going to be part of the advanced querying guide, but
that guide got too large so I split out an "extending diesel" guide from
it, but this felt separate from everything else in that guide, so I
decided it made sense as its own stand-alone guide

Unlike our other guides, there's no code in the examples directory for
this one. The links are going to point at crates.io. The code is
slightly modified from the real thing, but I'd still prefer to point the
links at the real app than our example copy of it.

### DIFF
--- a/guide_drafts/composing-diesel-applications.md
+++ b/guide_drafts/composing-diesel-applications.md
@@ -1,0 +1,246 @@
+Composing Applications with Diesel
+----------------------------------
+
+One of the main benefits of using a query builder over raw SQL
+is that you can pull bits of your query out into functions and reuse them.
+In this guide,
+we'll look at common patterns for extracting your code into re-usable pieces.
+We'll also look at best practices for how to structure your code.
+
+All of our code examples are based on code from crates.io,
+a real world application which uses Diesel extensively.
+All of our examples will be focused on functions which *return*
+queries or pieces of queries.
+None of these examples will include a function which takes a database
+connection.
+We will go into the benefits of this structure at the end of the guide.
+
+crates.io has a `canon_crate_name` SQL function
+which is always used when comparing crate names.
+Rather than continuously writing
+`canon_crate_name(crates::name).eq("some name")`,
+we can instead pull this out into a function.
+
+```
+use diesel::dsl::Eq;
+use diesel::types::Text;
+
+sql_function!(canon_crate_name, CanonCrateName, (x: Text) -> Text);
+
+type WithName<'a> = Eq<canon_crate_name<crates::name>, canon_crate_name<&'a str>>;
+
+fn with_name(name: &str) -> WithName {
+    canon_crate_name(crates::name).eq(canon_crate_name(name))
+}
+```
+
+Now when we want to find a crate by name, we can write
+`crates::table.filter(with_name("foo"))` instead.
+If we want to accept types other than a string,
+we can make the method generic.
+
+```
+use diesel::dsl::Eq;
+use diesel::types::Text;
+
+sql_function!(canon_crate_name, CanonCrateName, (x: Text) -> Text);
+
+type WithName<T> = Eq<canon_crate_name<crates::name>, canon_crate_name<T>>;
+
+fn with_name<T>(name: T) -> WithName<T>
+where
+    T: AsExpression<Text>,
+{
+    canon_crate_name(crates::name).eq(canon_crate_name(name))
+}
+```
+
+It's up to you whether you make your functions generic,
+or only take a single type.
+We recommend only making these functions generic if it's actually needed,
+since it requires additional bounds in your `where` clause.
+The bounds you need might not be clear,
+unless you are familiar with Diesel's lower levels.
+
+In these examples,
+we are using helper types from `diesel::dsl`
+to write the return type explicitly.
+Nearly every method in Diesel has a helper type like this.
+The first type parameter is the method receiver
+(the thing before the `.`).
+The remaining type parameters are the arguments to the method.
+If we want to avoid writing this return type,
+or dynamically return a different expression,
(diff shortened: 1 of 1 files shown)


sgrif:
I think I may expand on this with a second section in the future. I want to hammer home somewhere that "Queryable is one-to-one with the queries you run (which doesn't always mean a table)", and "Insertable and AsChangeset are one-to-one with web forms, API endpoints, or whatever your user input source is". I'm not sure if this guide is the right place to do it, but a second section that talks about when to use each, how to compose structs (especially for insert), etc probably makes sense.

Either way, I think this first section is fine to review on its own.

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+use diesel::types::Text;
+
+sql_function!(canon_crate_name, CanonCrateName, (x: Text) -> Text);
+
+fn with_name<'a, T>(name: T) -> Box<BoxableExpression<crates::table, Pg, SqlType = Bool> + 'a>
+where
+    T: AsExpression<Text>,
+    T::Expression: BoxableExpression<crates::table, Pg>,
+{
+    canon_crate_name(crates::name).eq(canon_crate_name(name))
```
weiznich: I think this is missing a `Box::new()` or something.

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+connection.
+We will go into the benefits of this structure at the end of the guide.
+
+crates.io has a `canon_crate_name` SQL function
+which is always used when comparing crate names.
+Rather than continuously writing
+`canon_crate_name(crates::name).eq("some name")`,
+we can instead pull this out into a function.
+
+```
```
weiznich: As noted on gitter: This should have a language hint.

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+    canon_crate_name(crates::name).eq(canon_crate_name(name))
+}
+```
+
+Now when we want to find a crate by name, we can write
+`crates::table.filter(with_name("foo"))` instead.
+If we want to accept types other than a string,
+we can make the method generic.
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+to write the return type explicitly.
+Nearly every method in Diesel has a helper type like this.
+The first type parameter is the method receiver
+(the thing before the `.`).
+The remaining type parameters are the arguments to the method.
+If we want to avoid writing this return type,
+or dynamically return a different expression,
+we can box the value instead.
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+another table.
+
+In addition to extracting expressions,
+you can also pull out entire queries into functions.
+Going back to crates.io,
+the `Crate` struct doesn't use every column from the `crates` table.
+Because we almost always select a subset of these columns,
+we have an `all` function which selects the columns we need.
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+        crates::table.select(ALL_COLUMNS)
+    }
+}
+```
+
+We also frequently found ourselves writing
+`Crate::all().filter(with_name(crate_name))`.
+We can pull that into a function as well.
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+Note that in all of our examples,
+we are writing functions which *return* queries or expressions.
+None of these functions execute the query.
+In general you should always prefer functions which return queries,
+and avoid functions which take a connection as an argument.
+This allows you to re-use and compose your queries.
+
+For example, if we had written our `by_name` function like this:
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+}
+```
+
+Then we would never be able to use this query in another context,
+or modify it further.
+By writing the function as one that returns a query,
+rather than executing it,
+we can do things like use it as a subselect.
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
```diff
@@ -0,0 +1,246 @@
...
+let version_id = versions
+    .select(id)
+    .filter(crate_id.eq_any(Crate::by_name(crate_name).select(crates::id)))
+    .filter(num.eq(version))
+    .first(&*conn)?;
+```
+
+Or use it to do things like get all of its downloads:
+
+```
```
weiznich: language hint

guide_drafts/composing-diesel-applications.md
sgrif: I dont bother adding them for drafts. I have to do a ton of work on code blocks when I publish

guide_drafts/composing-diesel-applications.md
sgrif: Good catch

sgrif: Good catch

PR #1437, opened 2017-12-24
Guide draft: Extending diesel

sgrif:
This guide covers adding new capabilities to Diesel's query builder.
It's main goal is to demystify things which might otherwise be
considered "internal"

### DIFF
--- a/guide_drafts/extending-diesel.md
+++ b/guide_drafts/extending-diesel.md
@@ -1,0 +1,385 @@
+Extending Diesel's Query Builder
+================================
+
+Diesel provides a lot of capabilities out of the box.
+However, it doesn't necessarily provide everything your app may want to use.
+One of Diesel's greatest strengths,
+is that it can be extended to add new functionality.
+
+In this guide we'll look at several ways to hook into Diesel's query builder,
+both to add new capabilities,
+and to introduce new abstractions.
+
+This guide is only going to cover extending the query builder.
+How to add support for new SQL types will be covered in a future guide.
+
+`sql_function!`
+---------------
+
+The easiest and most common way to extend Diesel's query builder
+is by declaring a new SQL function.
+This can be used for functions defined by your database,
+or for built-in functions that Diesel doesn't support out of the box.
+
+Functions in SQL often have multiple signatures,
+making them difficult or impossible to represent in Rust.
+Because of this, Diesel only provides support for a small number
+of built-in SQL functions.
+Consider `COALESCE`.
+This function can take any number of arguments,
+and its return type changes based on whether any arguments are `NOT NULL`.
+While we can't easily represent that in Rust,
+we can use `sql_function!` to declare it with the exact signature we're using.
+
+```
+use diesel::types::{Nullable, Text};
+sql_function!(coalesce, Coalesce, (x: Nullable<Text>, y: Text) -> Text);
+
+users.select(coalesce(hair_color, "blue"))
+```
+
+As this example shows,
+`sql_function!` converts it's argument like other parts of the query builder.
+This means that the generated function can take both Diesel expressions,
+and Rust values to be sent with the query.
+
+The macro takes three arguments:
+
+- A function name
+- A type name
+- A type signature
+
+The type signature uses the same syntax as a normal Rust function.
+However, the types given are SQL types,
+not concrete Rust types.
+This is what allows us to pass both columns and Rust strings.
+If we defined this function manually, it would look like this:
+
+```
+fn coalesce<X, Y>(x: X, y: Y) -> Coalesce<X::Expression, Y::Expression>
+where
+    X: AsExpression<Nullable<Text>>,
+    Y: AsExpression<Text>,
+```
+
+The type name given as the second argument is almost never used.
+Instead, a helper type is generated with the same name as the function.
+This helper type handle's Diesel's argument conversion.
+This lets us write `coalesce<hair_color, &str>`
+instead of `Coalsece<hair_color, Bound<Text, &str>>`.
+
+Custom SQL
+----------
+
+Often times it's useful to encapsulate a common SQL pattern.
+For example, if you're doing pagination on your queries,
+PostgreSQL is capable of loading the total count in a single query.
+The query you would want to execute would look like this:
+
+```
+SELECT *, COUNT(*) OVER () FROM (subselect t) LIMIT $1 OFFSET $1
+```
+
+However, as of version 1.0,
+Diesel doesn't support window functions, or selecting from a subselect.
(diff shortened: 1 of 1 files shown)


guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+    Y: AsExpression<Text>,
+```
+
+The type name given as the second argument is almost never used.
+Instead, a helper type is generated with the same name as the function.
+This helper type handle's Diesel's argument conversion.
+This lets us write `coalesce<hair_color, &str>`
+instead of `Coalsece<hair_color, Bound<Text, &str>>`.
+
+Custom SQL
```
sgrif: I would love an idea for a better header to use here.

sgrif: I originally had this as the third section (swapped with "custom operators"), since this is definitely "lower level" than custom operators are.

However, this seemed like code an app is more likely to write. Operators are more likely to be handled by libraries (and that example might even belong in the "supporting custom SQL types" guide).

Anyway, this order is roughly "most likely to be done by apps to least likely". If we swap this with "custom operators", the order becomes "highest level to lowest level"

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+
+```
+use diesel::types::{Nullable, Text};
+sql_function!(coalesce, Coalesce, (x: Nullable<Text>, y: Text) -> Text);
+
+users.select(coalesce(hair_color, "blue"))
+```
+
+As this example shows,
+`sql_function!` converts it's argument like other parts of the query builder.
```
alexcameron89: it's -> its

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+```
+fn coalesce<X, Y>(x: X, y: Y) -> Coalesce<X::Expression, Y::Expression>
+where
+    X: AsExpression<Nullable<Text>>,
+    Y: AsExpression<Text>,
+```
+
+The type name given as the second argument is almost never used.
+Instead, a helper type is generated with the same name as the function.
+This helper type handle's Diesel's argument conversion.
```
alexcameron89: handle's -> handles

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+impl<T> RunQueryDsl<PgConnection> for Paginated<T> {}
+```
+
+Now that we've implemented all of these things,
+let's look at how we would go about constructing this.
+We'll want to add a `paginate` method to all Diesel queries,
+which specifies which page we're on,
+as well as a `per_page` method which specifies the number of elements per page.
+
+Once again, we can extend existing types with a trait.
```
killercup: once again? in this guide didn't we didn't write any new traits to extend existing types IIRC

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+    fn paginate(self, page: i64) -> Paginated<Self::Query> {
+        Paginated {
+            query: self.as_query(),
+            page,
+            per_page: DEFAULT_PER_PAGE,
+         }
+    }
+}
+
+impl<T: AsQuery> Paginate for T {}
```
killercup: the idea of `impl MyTrait for T` might be new to people reading this. not sure if there is a good documentation on this. <https://doc.rust-lang.org/1.22.1/book/second-edition/ch10-02-traits.html#using-trait-bounds-to-conditionally-implement-methods> is a bit short

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+}
+```
+
+This is one of the rare cases where we want to define a function that takes a
+connection.
+One benefit of defining the function this way
+is that if we wanted to support backends other than PostgreSQL,
+we could have this function execute two queries.
+
+You can find the full code for this example in [the "advanced blog" example]
```
killercup: period.

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+The code is very similar, but without the trait.
+Here's now `not` is defined in Diesel.
+
+```
+pub fn not<T: AsExpression<Bool>>(expr: T) -> Not<Grouped<T::Expression>> {
+    super::operators::Not::new(Grouped(expr.as_expression()))
+}
+```
+
+In this case we're using `Grouped` to add parenthesis around our argument.
```
killercup: Maybe mention that this is `expression::grouped::Grouped`, which is currently only used internally (technically: hidden in the docs). Maybe we want to expose this in the future?

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+This ensures that the operator precedence in SQL matches what's expected.
+For example, if we would expect `not(true.and(false))` to return `true`.
```
killercup: cut the 'if' here

guide_drafts/extending-diesel.md
killercup: How about "Using custom SQL and how to extend the Query DSL"

I like that this takes such a prominent position in the guide; it's very useful information and more people should make use of this to build cool extension crates!

sgrif:
Recommendations applied. I'm going to publish rather than merging to drafts.

guide_drafts/extending-diesel.md
```diff
@@ -0,0 +1,385 @@
...
+see the documentation for [`AstPass`].
+One important question to ask whenever you implement `QueryBuilder`
+is whether you are generating a query that is safe to cache.
+The way to answer this question is by asking
+"does this struct generate an unlimited number of potential SQL queries"?
+Typically that is only the case if the body of `walk_ast` contains a for loop.
+If your query is not safe to cache, you *must* call
+`out.unsafe_to_cache_prepared`.
+
+Whenever you implement `QueryFragment`, you also need to implement `QueryId`.
```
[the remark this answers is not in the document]
sgrif: I like how you're always like "why do relative instead of https only" and then you paste a link to http :P

guide_drafts/extending-diesel.md
sgrif: We do in section 3 which used to come before this >_<

guide_drafts/extending-diesel.md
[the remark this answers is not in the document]
killercup: pff is it my fault it doesn't redirect to https? :P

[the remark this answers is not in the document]
sgrif: Given you added https support in the first place, yes. :P

PR #1444, opened 2018-01-01
Support `BINARY` type name for SQLite

### LINKED ISSUE #1355: `BINARY` should be a supported type name on SQLite
It'd be given the `NUMERIC` affinity by SQLite, but I don't think that's relevant. The relevant code for this lives [here](https://github.com/diesel-rs/diesel/blob/52f306ee53dc9646f8247795b769676125739ead/diesel_infer_schema/infer_schema_internals/src/sqlite.rs#L168-L194)

MarcManiez:
I believe this fixes #1355.

My understanding is that this issue is a matter of taking columns labeled `BINARY` and fitting them under [Diesel's `Binary` type](https://docs.diesel.rs/diesel/types/struct.Binary.html). I've just been getting into the source code for diesel and I'm slowly gaining more context for how `determine_column_type` fits into everything, so feel to correct and provide more context if I have the wrong idea!

Let me know if this is something we should add to the changelog. I'm happy to do that with some guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## Unreleased
+
+### Added
+
+* The `BINARY` column type name is now supported for SQLite.
+
 ## [1.0.0-rc1] - 2017-12-23
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


MarcManiez:
@weiznich I think I might be confused: don't we already have something like it [in the same file, a few lines down](https://github.com/diesel-rs/diesel/blob/master/diesel/src/types/mod.rs#L252)?

sgrif:
Can you a CHANGELOG entry?

MarcManiez:
I didn't see any other recent additions to the changelog, so I added an "Unreleased" section. Please do let me know if you had something more specific in mind!

weiznich:
> @weiznich I think I might be confused: don't we already have something like it in the same file, a few lines down?

My mistake, I forgot that the sqlite type mapping does work in a other way than the postgres mapping.

PR #1448, opened 2018-01-03
Deprecate `impl_query_id!` in favor of a derive

sgrif:
This is part of an ongoing effort to remove all macros with the word
`impl` in the name, and replace them with a derive.

I was actually surprised that using our own derives in Diesel itself
didn't just work. Unfortunately, since we can't access `$crate` in
`diesel_derives`, we did a hack where we wrap the whole thing in a dummy
const and do `extern crate diesel;` inside the block. You cannot do
`extern crate diesel` for code inside of Diesel. Luckily, it seems that
you can create new modules inside of const blocks as well, so I've
exported a macro that does `pub use $crate::*;` in Diesel and stuck that
in a module called `diesel`. This seems to work both inside and outside
of Diesel.

I've opted to exclude the `noop: ` form of the macro from the derive.
The number of types which need this is quite small, and are likely to
only be internal to Diesel. The actual impl you have to write in that
case is trivial, and I don't think it's worth adding machinery to the
derive to handle it.


diesel/src/query_builder/query_id.rs
```diff
@@ -7,15 +7,43 @@ use super::QueryFragment;
...
 /// See the documentation of [the `QueryId` type] and [`HAS_STATIC_QUERY_ID`]
 /// for more details.
 ///
-/// [`impl_query_id!`]: ../macro.impl_query_id.html
 /// [the `QueryId` type]: #associatedtype.QueryId
 /// [`HAS_STATIC_QUERY_ID`]: #associatedconstant.HAS_STATIC_QUERY_ID
+///
+/// ### Deriving
+///
+/// This trait can be automatically derived by `diesel_derives`.
```
weiznich: We should not mention `diesel_derives` here, because this derive is provided by diesel itself for all users outside of diesel.

diesel/src/query_builder/query_id.rs
```diff
@@ -7,15 +7,43 @@ use super::QueryFragment;
...
+/// For example, given this struct:
+///
+/// ```rust,ignore
```
weiznich: Is there any reason why this doc test should be ignored?

diesel/src/query_builder/query_id.rs
sgrif: Oh right hence why we can use it in Diesel >_<

Issue #1339, opened 2017-12-02
`pg::types::Array` should implement `SqlOrd`


sgrif:
Turns out that `MAX(int[])` totally works. This is a trivial thing to fix, but I don't want us to start adding a ton of new features until after 1.0 is out. This needs to wait for 1.1.

virome:
With 1.0 released, I'd like to take a stab at this one.

killercup:
Go for it, @virome!

sgrif:
This was fixed by #1449

PR #1449, opened 2018-01-04
Impl sqlord for array

virome:
Implements #1339.

### DIFF
--- a/diesel/src/types/ord.rs
+++ b/diesel/src/types/ord.rs
@@ -14,3 +14,5 @@ impl SqlOrd for types::Interval {}
 impl SqlOrd for types::Time {}
 impl SqlOrd for types::Timestamp {}
 impl<T: SqlOrd + NotNull> SqlOrd for types::Nullable<T> {}
+#[cfg(feature = "postgres")]
+impl<T: SqlOrd> SqlOrd for types::Array<T> {}
--- a/diesel_tests/tests/expressions/mod.rs
+++ b/diesel_tests/tests/expressions/mod.rs
@@ -1,7 +1,7 @@
 mod date_and_time;
 mod ops;
 
-use schema::{connection, NewUser, TestBackend};
+use schema::{connection, DropTable, NewUser, TestBackend};
 use schema::users::dsl::*;
 use diesel::*;
 use diesel::backend::Backend;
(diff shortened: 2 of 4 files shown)


diesel_tests/tests/expressions/mod.rs
```diff
@@ -67,6 +87,51 @@ fn test_count_max() {
...
+#[cfg(feature = "postgres")]
+fn test_min_max_of_array() {
+    use self::number_arrays::columns::*;
+    use self::number_arrays::table as number_arrays;
+
+    let connection = connection();
+
+    connection
+        .execute("CREATE TABLE number_arrays (na int[])")
+        .unwrap();
```
virome: So I think (?) this is wrong and I should be using a file in in https://github.com/diesel-rs/diesel/tree/master/migrations/postgresql

But what is the format for naming those folders?

virome:
Hmm, looks like some module failure on stable. Fixing now...

diesel/src/types/ord.rs
```diff
@@ -14,3 +16,5 @@ impl SqlOrd for types::Interval {}
 impl SqlOrd for types::Time {}
 impl SqlOrd for types::Timestamp {}
 impl<T: SqlOrd + NotNull> SqlOrd for types::Nullable<T> {}
+#[cfg(feature = "postgres")]
+impl<T: SqlOrd> SqlOrd for pg::types::sql_types::Array<T> {}
```
sgrif: You don't need to go through the `pg` module for this. Just `types::Array<T>` is enough.

diesel_tests/tests/expressions/mod.rs
```diff
@@ -67,6 +87,46 @@ fn test_count_max() {
...
+
+#[test]
+#[cfg(feature = "postgres")]
+fn test_min_max_of_array() {
+    use self::number_arrays::columns::*;
+    use self::number_arrays::table as number_arrays;
+
+    let connection = connection();
+    connection
+        .execute("INSERT INTO number_arrays (na) VALUES ('{1, 20, 300}'), ('{0, 2}'), ('{500}')")
```
sgrif: We prefer to stick to public API for new tests, and avoid using `execute`.

diesel_tests/tests/expressions/mod.rs
```diff
@@ -67,6 +87,46 @@ fn test_count_max() {
...
+    number_arrays (na) {
+        na -> Array<Integer>,
+    }
+}
+
+#[test]
+#[cfg(feature = "postgres")]
+fn test_min_max_of_array() {
+    use self::number_arrays::columns::*;
+    use self::number_arrays::table as number_arrays;
```
sgrif: `use self::number_arrays::dsl::*;`

migrations/postgresql/2018-01-04-033400_array_test_tables/up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE number_arrays (
```
sgrif: Can we just create this table in the test directly since it's only used once?

diesel_tests/tests/expressions/mod.rs
```diff
@@ -48,7 +48,27 @@ table! {
 }
 
 #[test]
-fn test_count_max() {
+fn test_min() {
+    use self::numbers::columns::*;
```
sgrif: `use self::numbers::dsl::*;`

diesel_tests/tests/expressions/mod.rs
```diff
@@ -48,7 +48,27 @@ table! {
...
+    use self::numbers::table as numbers;
+
+    let connection = connection();
+    connection
+        .execute("INSERT INTO numbers (n) VALUES (2), (1), (5)")
```
sgrif: Can we use the public API for this? `diesel::insert_into(numbers).values(&vec![n.eq(1), n.eq(2), n.eq(3)])`

diesel_tests/tests/expressions/mod.rs
```diff
@@ -48,7 +48,27 @@ table! {
...
+        .unwrap();
+    let source = numbers.select(min(n));
+
+    assert_eq!(Ok(Some(1)), source.first(&connection));
+    connection
+        .execute("DELETE FROM numbers WHERE n = 1")
```
sgrif: Can we use the public API for this? `diesel::delete(numbers).filter(n.eq(1))`

diesel_tests/tests/expressions/mod.rs
```diff
@@ -48,7 +48,27 @@ table! {
...
+        .unwrap();
+    assert_eq!(Ok(Some(2)), source.first(&connection));
+    connection.execute("DELETE FROM numbers").unwrap();
```
sgrif: Can we use the public API for this? `diesel::delete(numbers)`

diesel_tests/tests/expressions/mod.rs
```diff
@@ -67,6 +87,56 @@ fn test_count_max() {
...
+        .execute(&connection)
+        .unwrap();
+    assert_eq!(Ok(Some(vec![1, 5, 5])), max_query.first(&connection));
+    assert_eq!(Ok(Some(vec![1, 5, 5])), min_query.first(&connection));
+
+    delete(number_arrays).execute(&connection).unwrap();
+    assert_eq!(Ok(None::<Vec<i32>>), max_query.first(&connection));
+    assert_eq!(Ok(None::<Vec<i32>>), min_query.first(&connection));
+
+    connection.execute("DROP TABLE number_arrays").unwrap();
```
sgrif: If the test fails, or someone interrupts the test mid-run, this will break. We need to either change the `CREATE TABLE` line to be `CREATE TABLE IF NOT EXISTS`, or (preferably) do this `DROP TABLE` in a `Drop` impl

diesel_tests/tests/expressions/mod.rs
```diff
@@ -48,7 +48,27 @@ table! {
...
+    let source = numbers.select(min(n));
+
+    assert_eq!(Ok(Some(1)), source.first(&connection));
+    connection
+        .execute("DELETE FROM numbers WHERE n = 1")
+        .unwrap();
+    assert_eq!(Ok(Some(2)), source.first(&connection));
+    connection.execute("DELETE FROM numbers").unwrap();
+    assert_eq!(Ok(None::<i32>), source.first(&connection));
+}
```
sgrif: Sorry I just realized that all this code was moved from another part in the file. Can we leave this code where it was before? This just adds unnecessary git churn.

diesel_tests/tests/expressions/mod.rs
virome: This is the test that I just moved up from below, which is why I didn't change it when you said to use the api for new tests.
But I'll go ahead and migrate to diesel public api for both the test_min and test_max functions since they're related.

diesel_tests/tests/expressions/mod.rs
virome: Sounds good

PR #1452, opened 2018-01-06
Remove `queryable_impls!` in favor of `#[derive(FromSqlRow)]`

sgrif:
This generates a slightly different impl than we had before.
Rather than explicitly stating the SQL type,
the generated impl is now generic for any SQL type
as long as there is a `FromSql` impl.
This is basically the blanket impl that I want to write,
but can't for various reasons.

This means that certain `queryable_impl!` invocations were redundant.
Some impls will now be present regardless of enabled features.

This derive will also generate a `Queryable` impl,
which you always want when implementing `FromSqlRow`.

Since the majority of impls in Diesel are for types located outside of Diesel, I've added an undocumented `foreign_derive` flag. The error handling around it is light, since it's intended to only be used by Diesel

### DIFF
--- a/diesel/src/pg/types/network_address.rs
+++ b/diesel/src/pg/types/network_address.rs
@@ -25,6 +25,19 @@ primitive_impls!(MacAddr -> ([u8; 6], pg: (829, 1040)));
 primitive_impls!(Inet -> (IpNetwork, pg: (869, 1041)));
 primitive_impls!(Cidr -> (IpNetwork, pg: (650, 651)));
 
+#[allow(dead_code)]
+mod foreign_derives {
+    use super::IpNetwork;
+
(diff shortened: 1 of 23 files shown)


diesel/src/pg/types/network_address.rs
```diff
@@ -25,6 +25,19 @@ primitive_impls!(MacAddr -> ([u8; 6], pg: (829, 1040)));
 primitive_impls!(Inet -> (IpNetwork, pg: (869, 1041)));
 primitive_impls!(Cidr -> (IpNetwork, pg: (650, 651)));
 
+#[allow(dead_code)]
```
Eijebong: This is kinda meh but I guess we don't have any other choice

PR #1453, opened 2018-01-06
Remove `expression_impls!` in favor of a derive

sgrif:
The nullable `ToSql` impl feels funky here,
but I've left it there to keep parity with the old macro.
Similar to `FromSqlRow`, there is an undocumented `foreign_derive` flag.
Any external crates attempting to use that flag would actually fail to
compile,
since `impl AsExpression<Nullable<Type>, DB> for NotLocal` is a
coherence violation.

This derive does not work for types which need to be generic on the SQL
type, so `Option`, arrays, and ranges can't use it.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -109,6 +109,26 @@ impl<'a, T: Expression + ?Sized> Expression for &'a T {
 ///   [`Timestamp`]: ../types/struct.Timestamp.html
 ///   [`Timestamptz`]: ../../pg/types/sql_types/struct.Timestamptz.html
 ///   [`ToSql`]: ../types/trait.ToSql.html
+///
+/// ## Deriving
+///
+/// This trait can be automatically derived for any type which implements `ToSql`.
+/// The type must be annotated with `#[sql_type = "SomeType"]`.
+/// If that annotation appears multiple times,
+/// implementations will be generated for each one of them.
+///
+/// This will generate the following impls:
+///
+/// - `impl AsExpression<SqlType> for YourType`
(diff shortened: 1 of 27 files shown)


diesel/src/expression/mod.rs
```diff
@@ -109,6 +109,26 @@ impl<'a, T: Expression + ?Sized> Expression for &'a T {
...
+/// This will generate the following impls:
+///
+/// - `impl AsExpression<SqlType> for YourType`
+/// - `impl AsExpression<Nullable<SqlType>> for YourType`
+/// - `impl AsExpression<SqlType> for &'a YourType`
+/// - `impl AsExpression<Nullable<SqlType>> for &'a YourType`
+/// - `impl AsExpression<SqlType> for &'a &'b YourType`
+/// - `impl AsExpression<Nullable<SqlType>> for &'a &'b YourType`
+///
+/// If your type is unsized,
```
Eijebong: That's a weird way to put that. Did rustfmt do that ?

diesel_derives/src/util.rs
```diff
@@ -62,6 +66,11 @@ fn str_value_of_attr<'a>(attr: &'a Attribute, name: &str) -> &'a str {
     str_value_of_meta_item(&attr.value, name)
 }
 
+pub fn ty_value_of_attr<'a>(attr: &'a Attribute, name: &str) -> Ty {
+    parse::ty(str_value_of_attr(attr, name))
+        .expect(&format!("#[{}] did not contian a valid Rust type", name))
```
Eijebong: contain

PR #1460, opened 2018-01-07
Implement `or_filter`

### LINKED ISSUE #1389: Add `or_filter` (this name is terrible we need a better one)
This method would basically behave the same as `filter`, but if an existing where clause exists, it would combine with `OR` rather than `AND`. As an example use case, crates.io has a very large, complex, dynamically constructed query. However, regardless of whatever other conditions are added, they virtually always want to include a crate "if its name is an exact match for the search term". It would be very useful for that case to be able to put at the very end of the query construction, something like `.or_filter(exact_match)`.

I actually considered a method like this back in 0.1, but considered it to generally be a smell, since we already have `.or` on `ExpressionMethods`. However, I'm starting to think that it's reasonable to provide this, since `filter` is the *only* method in our query builder which concatenates rather than overrides, and it seems reasonable to allow concatenating with `OR` rather than `AND`.

sgrif:
I'm still not sold on this method name, but I haven't thought of a
better one. I'm open to any suggestions on better names.

This method behaves identically to `filter`, but it appends to existing
queries using `.or` instead of `.and`. This is useful for cases where
you have a large query being constructed dynamically, but you always
want some other condition to be included. A motivating use case in
crates.io is that we want to be able to do
`.or_filter(has_name(q_string))` in crates::index,
so that crates with an exact match are
always included regardless of what else we've done with the query (the
function itself is over a hundred lines of query construction).

Fixes #1389

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -477,6 +477,55 @@ pub trait QueryDsl: Sized {
         methods::FilterDsl::filter(self, predicate)
     }
 
+    /// Adds to the `WHERE` clause of a query using `OR`
+    ///
+    /// If there is already a `WHERE` clause, the result will be `(old OR new)`.
+    /// Calling `foo.filter(bar).or_filter(baz)`
+    /// is identical to `foo.filter(bar.or(baz))`.
+    /// However, the second form is much harder to do dynamically.
+    ///
+    /// # Example:
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use schema::animals::dsl::*;
+    /// #     let connection = establish_connection();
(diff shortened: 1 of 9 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -477,6 +477,55 @@ pub trait QueryDsl: Sized {
         methods::FilterDsl::filter(self, predicate)
     }
 
+    /// Adds to the `WHERE` clause of a query using `OR`
+    ///
+    /// If there is already a `WHERE` clause, the result will be `(old AND new)`.
```
jbcden: I might be misunderstanding this but should this be `(old OR new)`?

sgrif: Good catch

sgrif:
I'm going to leave this open for a day or two in case anybody has a better idea for the method name.

sgrif:
@jbcden FYI, we usually avoid the "Request changes" button unless the changes are substantial enough that you think a second review will be needed after the changes are addressed.

jbcden:
>  FYI, we usually avoid the "Request changes" button unless the changes are substantial enough that you think a second review will be needed after the changes are addressed.

@sgrif good to know!

jbcden:
@sgrif I'm not sure if this would be any better but what do you think of `or_filter_by`?

Issue #1356, opened 2017-12-06
Support `BEGIN EXCLUSIVE` or `BEGIN IMMEDIATE` for Sqlite.


Lymia:
Sqlite uses relatively weak locking by default when just `BEGIN` is used to start a transaction. Normally, you would use `BEGIN EXCLUSIVE` or `BEGIN IMMEDIATE` rather than `BEGIN` to do such a transaction with stronger locking, however, Diesel has no support for using these.

sgrif:
Ideally we'd handle this and #408 at the same time.

PR #1462, opened 2018-01-07
Support specifying details about a PG transaction

### LINKED ISSUE #408: Transaction isolation
Does Diesel allow control of the isolation level of a transaction? It doesn't appear that it does right now, in which case consider this a feature request!

Specifically, for my use case, I want to use the "serializable" transaction isolation level in Postgres for an operation where a transaction should fail if a value it read in a query has changed by the time the transaction is committed. In other words, it queries a table, and if the data returned matches the expected condition, adds a new row to another table, but that insert should fail if the row previously queried from the first table has changed. (If there's a better way to do that, I'm all ears!) In other other words, defending against a time of check to time of use error.

sgrif:
This adds support for all options which can be given to `SET
TRANSACTION`. We only expose an API for this when opening the
transaction though, as `SET TRANSACTION` errors unless it's the first
query run after opening the transaction.

Most of the tests are really just there to make sure the code compiles
and generates valid SQL. Seeing *a* usage example is still worth having,
but I don't think we can effectively demonstrate/test the differences
between isolation levels in a reasonable manner.

This lays some groundwork that will also be used by #1356, but I haven't
implemented SQLite support yet. The APIs for this will be backend
specific, not trait based.

Fixes #408.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## Unreleased
 
+### Added
+
+* Added support for specifying `ISOLATION LEVEL`, `DEFERRABLE`, and `READ ONLY`
+  on PG transactions. See [`PgConnection::build_transaction`] for details.
+
+[`PgConnection::build_transaction`]: http://docs.diesel.rs/diesel/pg/struct.PgConnection.html#method.build_transaction
+
 ### Deprecated
 
 * `ne_any` has been renamed to `ne_all`.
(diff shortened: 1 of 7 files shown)


sgrif:
This is ready for review regardless of the CI failures. I haven't had a chance to look into why this is failing on CI since `READ ONLY` and `READ WRITE` are not new. There might be order dependence in versions of PG older than what I tested on locally

sgrif:
I'm hoping that a rebase magically fixes the CI failures because I have no clue why `READ ONLY` and `READ WRITE` are somehow not working

jbcden:
It looks like adding
```rust
            let connection = PgConnection::establish(&connection_url).unwrap();
            connection.execute("CREATE TABLE IF NOT EXISTS users (
                id SERIAL PRIMARY KEY,
                name VARCHAR NOT NULL
            )").unwrap();
```
to the `connection_no_transaction` helper function seems to fix this. Not sure if that's ideal or not though.

sgrif:
Lol yup, that'd do it. Explains why it wasn't failing locally >_>

sgrif:
Hm, that's actually a problem since I don't think we can modify schema in a `READ ONLY` transaction

PR #1463, opened 2018-01-07
Remove `primitive_impls!` in favor of a derive.

sgrif:
This adds `#[derive(SqlType)]`, which is responsible for implementing
`NotNull`, `SingleValue`, and potentially also implementing `HasSqlType`
on one or more backends. The body of `HasSqlType` is specified through
additional attributes. The only one of note is the `postgres` attribute,
which can do either static OIDs or dynamic type lookup.

`Array` and `Range` are not able to make full use of this derive, since
`Range` has a bunch of static OIDs and `Array` has a unique
implementation of `HasSqlType`.

At this point, the `custom_types` test is about as small as it's ever
going to get. The only way we could possibly eliminate any more code
there is by making it so you don't have to write as much of the
signature.


diesel/src/types/mod.rs
```diff
@@ -292,12 +316,10 @@ pub struct Date;
 ///
 /// [`PgInterval`]: ../pg/data_types/struct.PgInterval.html
 /// [`IntervalDsl`]: ../pg/expression/extensions/trait.IntervalDsl.html
-#[derive(Debug, Clone, Copy, Default, QueryId)]
+#[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
+#[postgres(oid = "1186", array_oid = "1187")]
 pub struct Interval;
 
-#[cfg(not(feature = "postgres"))]
```
weiznich: The new derive adds this impl also for Postgres. Is this intended?

diesel_derives/src/sql_type.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    fn error() -> ! {
+        panic!(
+            "#[postgres] must be in the form \
+             #[postgres(oid = \"1\", array_oid = \"2\")] \
+             or #[postgres(type_name = \"my_type\")]"
+        );
+    }
+
+    if cfg!(not(feature = "postgres")) {
```
weiznich: I case that "postgres" is not enabled the `#[postgres]` attribute is not checked. I think we should either check the attribute unconditionally or report an error if the attribute exists but "postgres" is not enabled.

diesel_derives/src/sql_type.rs
```diff
@@ -0,0 +1,133 @@
...
+
+fn pg_tokens(
+    attrs: &[syn::Attribute],
+    generics: &syn::Generics,
+    struct_ty: &syn::Ty,
+) -> Option<Tokens> {
+    use syn::{Lit, MetaItem, NestedMetaItem};
+
+    fn error() -> ! {
+        panic!(
```
weiznich: Didn't we decide [here](https://github.com/diesel-rs/diesel/pull/1327#pullrequestreview-80271119) that `compile_error!` messages are more readable?

diesel_derives/src/sql_type.rs
sgrif: That would mean Diesel would error every time you compile without all three backends. You should be able to declare how a backend supports a type, and then only have the code generated if that backend is enabled for Diesel.

diesel/src/types/mod.rs
sgrif: The impl was always present it just came from a different place if PG was enabled.

diesel_derives/src/sql_type.rs
sgrif: Very marginally, but it would be significantly harder to do that here than it was in that case

diesel_derives/src/sql_type.rs
weiznich: Then we should at least check if the attribute would be valid, without generating the actual impl.

PR #1464, opened 2018-01-08
Add type aliases for the return types of `ToSql` and `FromSql`

sgrif:
These types were annoying to write out, and often involved importing
several types you wouldn't otherwise need to import.

These aliases are `serialize::Result` and `deserialize::Result`. I
intend to eventually move the relevant traits into these modules, but
this felt like a good standalone change.

### DIFF
--- a/diesel/src/mysql/types/numeric.rs
+++ b/diesel/src/mysql/types/numeric.rs
@@ -2,20 +2,15 @@
 pub mod bigdecimal {
     extern crate bigdecimal;
 
-    use std::error::Error;
+    use self::bigdecimal::BigDecimal;
     use std::io::prelude::*;
 
     use mysql::Mysql;
-
-    use self::bigdecimal::BigDecimal;
-
     use types::{self, FromSql, IsNull, ToSql, ToSqlOutput};
+    use {deserialize, serialize};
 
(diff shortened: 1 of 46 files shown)


diesel/src/mysql/types/numeric.rs
```diff
@@ -2,28 +2,23 @@
...
 
     impl ToSql<types::Numeric, Mysql> for BigDecimal {
-        fn to_sql<W: Write>(
-            &self,
-            out: &mut ToSqlOutput<W, Mysql>,
-        ) -> Result<IsNull, Box<Error + Send + Sync>> {
+        fn to_sql<W: Write>(&self, out: &mut ToSqlOutput<W, Mysql>) -> serialize::Result {
             write!(out, "{}", *self)
                 .map(|_| IsNull::No)
                 .map_err(|e| e.into())
```
weiznich: This should be written as `Into::into` 
I think there is a clippy lint for this, but it seems to have this overseen.

sgrif: I don't think existing code that does this is worth touching.

sgrif:
This was merged but I'm bad at rebasing.

PR #1465, opened 2018-01-08
Refactor the `ToSql` implementation of `bool` on PG

sgrif:


jbcden:
@sgrif Looks like a nice refactoring to me!

weiznich:
@diesel-rs/contributors 
As far as I remember the representation of rusts `bool` type is considered a internal implementation detail and may change at any time. So this pull request introduces at least implementation defined behavior. 
See [the most up to date comment](https://github.com/rust-lang/rfcs/pull/954#issuecomment-83811006) I've found about this.

Eijebong:
Maybe this should be added as a comment in the code ? (Would've been nice in the commit message but it's too late for that :p)

sgrif:
Reverting (going to be a manual commit since there was rebasing involved)

sgrif:
Actually I lied. This isn't implementation defined behavior. We're not relying on the internal representation of bool, we're relying on the behavior of casting a bool to `u8`. This is not an implementation detail, it is public API which cannot be changed without a major version bump. (If we were using `mem::transmute` that would be a different story)

PR #1466, opened 2018-01-08
Merge r2d2-diesel into Diesel itself

### LINKED ISSUE #1348: Explore merging `r2d2-diesel` into Diesel itself
This was brought up in the impl period working group channel. The more I think about it, the more I like this idea. I started thinking about this again recently when I was looking at [the "database" keyword on crates.io](https://crates.io/keywords/database). As of the time of this issue, r2d2 shows up before any library that people there are likely actually looking for (diesel, rusqlite, or postgres). It makes sense for that to be the case. Anybody using Diesel (except for sqlite users) or rust-postgres are almost certainly going to rely on r2d2. It'd make sense for them to remove the `database` keyword, but at the same time, neither of those libraries mentions r2d2 in their docs, so I'm not sure where people would find it.

The point of all this is that I'm warming up to the idea of merging `r2d2-diesel` into Diesel itself. I think it still needs some exploration though. I'd like to see how much of the code from https://gitter.im/rust-impl-period/WG-libs-diesel?at=5a1d24628b3a9e2c0c27d69f actually goes away if we moved it into Diesel (just having aliases in Diesel is not good enough IMO, we need to abstract over some of those types to the extent that users don't have to care about them).

If nothing else, having `impl<T: Connection> Connection for PooledConnection<T>` will help a ton, and remove the need for explicit derefs which can be annoying and confusing for newcomers.

I've placed this issue on the 1.1 milestone, but again, this issue is for "exploring" this option, not for sure making it happen.

sgrif:
The main benefit of doing this is that we can implement `Connection` for
`PooledConnection`, removing the requirement to do `&*conn` in your
outermost functions. Additionally, we can just re-export r2d2 from
within Diesel itself, so we can hopefully get fewer bug reports saying
"Hey why does it say PgConnection doesn't implement Connection". Instead
people can just `use diesel::r2d2::Pool` instead of explicitly depending
on r2d2.

The code and the tests are both stripped straight from r2d2-diesel.
(Yeah, the tests are a bit... opaque. They come from the initial commit
of that repo, with the commit message "The tests have been pretty much
ripped from r2d2-postgres, as I really don't know what else to do to
test this". So blame r2d2-postgres I guess)

The only changes between this code and what is in r2d2-diesel 1.0 is the
`Connection` impl. This impl has to be a little bit funky. I decided to
just restrict to `AnsiTransactionManager` since that's what all three
backends use at the moment. If we wanted to support anything that
implemented `diesel::Connection`, we would need to add the constraint
`C::TransactionManager: TransactionManager<Self>`. However,
`TransactionManager` has the bound that it's parameter be `Connection`,
so we just recurse infinitely. Maybe on day the language will be smarter
about this. Until then, we just have to do this weird hack.

Fixes #1348.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -31,6 +31,7 @@ num-traits = { version = "0.1.35", optional = true }
 num-integer = { version = "0.1.32", optional = true }
 bigdecimal = { version = "0.0.10", optional = true }
 bitflags = { version = "1.0", optional = true }
+r2d2 = { version = ">= 0.7, < 0.9", optional = true }
 
 [dev-dependencies]
 cfg-if = "0.1.0"
@@ -40,7 +41,7 @@ tempdir = "^0.3.4"
 
 [features]
 default = ["with-deprecated"]
-extras = ["chrono", "serde_json", "uuid", "deprecated-time", "network-address", "numeric"]
+extras = ["chrono", "serde_json", "uuid", "deprecated-time", "network-address", "numeric", "r2d2"]
 unstable = []
 lint = ["clippy"]
 large-tables = []
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -1,0 +1,216 @@
+//! Connection pooling via r2d2
+
+extern crate r2d2;
+
+pub use self::r2d2::*;
+
+use std::convert::Into;
+use std::fmt;
+use std::marker::PhantomData;
+
+use backend::UsesAnsiSavepointSyntax;
+use deserialize::QueryableByName;
+use prelude::*;
+use connection::{AnsiTransactionManager, SimpleConnection};
+use query_builder::{AsQuery, QueryFragment, QueryId};
+use sql_types::HasSqlType;
+
+/// An r2d2 connection manager for use with Diesel.
+///
+/// See the [r2d2 documentation] for usage examples.
+///
+/// [r2d2 documentation]: ../../r2d2
+#[derive(Debug, Clone)]
+pub struct ConnectionManager<T> {
+    database_url: String,
+    _marker: PhantomData<T>,
+}
+
+unsafe impl<T: Send + 'static> Sync for ConnectionManager<T> {}
+
+impl<T> ConnectionManager<T> {
+    /// Returns a new connection manager,
+    /// which establishes connections to the given database URL.
+    pub fn new<S: Into<String>>(database_url: S) -> Self {
+        ConnectionManager {
+            database_url: database_url.into(),
+            _marker: PhantomData,
+        }
+    }
+}
+
+/// The error used when managing connections with `r2d2`.
+#[derive(Debug)]
+pub enum Error {
+    /// An error occurred establishing the connection
+    ConnectionError(ConnectionError),
+
+    /// An error occurred pinging the database
+    QueryError(::result::Error),
+}
+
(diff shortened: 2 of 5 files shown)


diesel/src/r2d2.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl<C> Connection for PooledConnection<ConnectionManager<C>>
+where
+    C: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
+    C::Backend: UsesAnsiSavepointSyntax,
+{
+    type Backend = C::Backend;
+    type TransactionManager = C::TransactionManager;
+
+    fn establish(_: &str) -> ConnectionResult<Self> {
```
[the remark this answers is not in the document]
sgrif: Not backwards compatibly, no

diesel/src/r2d2.rs
```diff
@@ -0,0 +1,216 @@
...
+{
+    fn batch_execute(&self, query: &str) -> QueryResult<()> {
+        (&**self).batch_execute(query)
+    }
+}
+
+impl<C> Connection for PooledConnection<ConnectionManager<C>>
+where
+    C: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
+    C::Backend: UsesAnsiSavepointSyntax,
```
weiznich: Why this constrain is required?
This code should work for any kind of backend as far as I see

sgrif: This is explained in both the commit message and PR description.

sgrif: >  If we wanted to support anything that
> implemented `diesel::Connection`, we would need to add the constraint
> `C::TransactionManager: TransactionManager<Self>`. However,
> `TransactionManager` has the bound that it's parameter be `Connection`,
> so we just recurse infinitely. Maybe on day the language will be smarter
> about this. Until then, we just have to do this weird hack.

diesel/Cargo.toml
```diff
@@ -31,6 +31,7 @@ num-traits = { version = "0.1.35", optional = true }
 num-integer = { version = "0.1.32", optional = true }
 bigdecimal = { version = "0.0.10", optional = true }
 bitflags = { version = "1.0", optional = true }
+r2d2 = { version = ">= 0.7, < 0.9", optional = true }
```
Kerollmops: There is a problem here, the r2d2 version you chose here [doesn't make `r2d2::Error` public](https://docs.rs/r2d2/0.7.4/r2d2/) and make the diesel r2d2 reexport to not compile. In the diesel 0.8 version, [the `Error` type is now public](https://docs.rs/r2d2/0.8.2/r2d2/).

```rust
error[E0603]: trait `Error` is private
  --> /Users/crenault/.cargo/registry/src/github.com-1ecc6299db9ec823/diesel-1.1.1/src/r2d2.rs:11:22
   |
11 | pub type PoolError = self::r2d2::Error;
   |                      ^^^^^^^^^^^^^^^^^
```

Kerollmops: Whooops, it seems to be fixed in the master...

weiznich: Yes, [this PR](https://github.com/diesel-rs/diesel/pull/1513) fixed that problem.

weiznich: It is possible to fix this locally by executing `cargo update -p r2d2`

PR #1468, opened 2018-01-09
Port `#[derive(Associations)]` to entirely use a custom derive

weiznich:
* Deprecated the `BelongsTo` macro
* Add a small check if all options passed to the custom derive are valid

Follow up to #1055


diesel/src/macros/associations/belongs_to.rs
```diff
@@ -1,5 +1,7 @@
 #[macro_export]
 #[doc(hidden)]
+#[cfg(feature = "with-deprecated")]
+#[deprecated(since = "1.1.0", note = "Use `#[derive(Associations)]` instead")]
```
sgrif: This isn't public API. We don't need to deprecate it.

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
     let foreign_key_name = options
         .foreign_key_name
         .unwrap_or_else(|| to_foreign_key(parent_struct.as_ref()));
+    let foreign_key_name = &foreign_key_name;
     let child_table_name = model.table_name();
-    let fields = model.attrs.as_slice();
+    let child_table_name = &child_table_name;
+    let foreign_key_attr = fields
+        .iter()
+        .find(|attr| attr.column_name().as_ref() == foreign_key_name.as_ref())
```
sgrif: Do we need any of these `as_ref`s?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
+        .unwrap();
```
sgrif: Can we use `expect` here? This is going to give a very poor error message if the model doesn't have a field with the right column name.

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
 
 fn expand_belongs_to(model: &Model, options: AssociationOptions) -> quote::Tokens {
     let parent_struct = options.name;
     let struct_name = &model.name;
+    let fields = model.attrs.as_slice();
 
     let foreign_key_name = options
         .foreign_key_name
         .unwrap_or_else(|| to_foreign_key(parent_struct.as_ref()));
+    let foreign_key_name = &foreign_key_name;
```
sgrif: Why is this needed?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
     let child_table_name = model.table_name();
-    let fields = model.attrs.as_slice();
+    let child_table_name = &child_table_name;
+    let foreign_key_attr = fields
+        .iter()
+        .find(|attr| attr.column_name().as_ref() == foreign_key_name.as_ref())
+        .unwrap();
+    let foreign_key_ty = &foreign_key_attr.ty;
```
sgrif: `foreign_key_attr` is never used again. Can we just do `.map(|a| a.ty)`?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
+
+    // we need to special case foreign keys on with an Option type
+    // to allow self referencing joins
+    let (foreign_key, foreign_key_ty) = match *foreign_key_ty {
+        syn::Ty::Path(None, ref p) if p.segments[0].ident == "Option" => {
```
sgrif: Don't we already have a util function for getting the inner part of an option?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
+            let segment = &p.segments[0];
+            let t = match segment.parameters {
+                syn::PathParameters::AngleBracketed(ref p) => p.types[0].clone(),
+                syn::PathParameters::Parenthesized(_) => unreachable!(),
+            };
+            (quote!(self.#foreign_key_name.as_ref()), t)
```
sgrif: Why `as_ref` and not `&`?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
+        }
+        ref t => (quote!(Some(&self.#foreign_key_name)), t.clone()),
```
sgrif: What is `t`? The type? Can we at least call it `ty`?

diesel_derives/src/associations.rs
```diff
@@ -2,55 +2,84 @@ use syn;
...
 }
 
-fn build_association_options(
-    attr: &syn::Attribute,
-    association_kind: &str,
-) -> Result<AssociationOptions, String> {
-    let usage_error = Err(format!(
-        "`#[{}]` must be in the form `#[{}(table_name, option=value)]`",
-        association_kind, association_kind
+fn build_association_options(attr: &syn::Attribute) -> Result<AssociationOptions, quote::Tokens> {
```
sgrif: The changes to this function feel unrelated to the rest of this PR. Can we leave them out?

diesel/src/macros/associations/belongs_to.rs
weiznich: Technically this is public API, because a downstream crate could use this. Removing this macro will cause those (theoretical existing) crates to break.

diesel_derives/src/associations.rs
weiznich: We can move them to a separate pull request. These changes are unrealated to the actual porting, but include
* some cleanup (removing a unused function parameter)
* improved error handling, compile error and checking the syntax of the `#[belongs_to()]` attribute more deeply

So just open a follow up pullrequest containing this changes, as soon as we merged this one, or only put those changes into a separate commit?

diesel/src/macros/associations/belongs_to.rs
sgrif: Our definition of public API is whether or not it appears on documentation. `#[doc(hidden)]` items can be freely changed without deprecation.

diesel_compile_tests/tests/compile-fail/associations_errors.rs
```diff
@@ -0,0 +1,32 @@
+#[macro_use] extern crate diesel;
+
+#[derive(Associations)]
+//~^ ERROR Could not derive `Associations` without any `#[belongs_to(table_name)]` attribute
+struct Post1 {
+    a: i32,
+}
+
+#[derive(Associations)]
+//~^ ERROR The foreign key column post_id is not found
```
sgrif: Can we improve this error message to say "No field found that corresponds to the column `post_id`"

sgrif:
LGTM after the old macro is removed.

sgrif:
Also the test for the old macro need to be ported over I think

weiznich:
> Also the tests for the old macro need to be ported over I think

There are no tests for the old macro :worried:

sgrif:
> There are no tests for the old macro 😟

Awesome...

PR #1471, opened 2018-01-10
💣 `types`

sgrif:
I know the diff numbers look really scary but I promise the vast
majority of the line count is from renaming files (a few of which are
our longest files at nearly 2k lines).

Since there's a *lot* of churn in this commit, I'll break from the norm
and actually list everything that's done.

`diesel::types` has been deprecated. All items in that module have been
moved to either `diesel::sql_types`, `diesel::deserialize`, or
`diesel::serialize`. Additionally, `Queryable` and `QueryableByName`
have been moved to `diesel::deserialize`, and `TypeMetadata` has been
moved to `sql_types`. `ToSqlOutput` has been renamed to `deserialize::Output`.
Any `sql_type` structs which were previously behind feature flags have
had the flags removed (your schema containing a json column should compile
even without `serde-json`)

This makes it much easier to find the traits or types involved in each
portion of Diesel's query serialization/deserialization cycle, and makes
it more apparent what the struct `Integer` is.

Most of this churn is just fixing `use` statements, but there were a few
files that basically ended up with `use diesel::sql_types::{self,
Numeric}` where `sql_types` is only ever used as `sql_types::Numeric`...
For those files I've also updated the usage a bit.


weiznich:
Travis says this needs a rustfmt run

diesel/src/lib.rs
```diff
@@ -69,7 +69,7 @@
 //! a trait called [`Queryable`].
 //!
 //! Diesel maps "Rust types" (e.g. `i32`) to and from "SQL types"
-//! (e.g. [`diesel::types::Integer`]).
+//! (e.g. [`diesel::sql_types::Integer`]).
 //! You can find all the types supported by Diesel in [the `types` module](types).
```
weiznich: This line references the `types` module which is now "empty". I think this should better point to `sql_types` now

diesel/src/lib.rs
```diff
@@ -83,7 +83,7 @@
 //! and find the Rust type you want to use.
 //!
 //! [`Queryable`]: query_source/trait.Queryable.html
```
weiznich: Point to `deserialize::Queryable`?

diesel/src/lib.rs
```diff
@@ -83,7 +83,7 @@
...
-//! [`diesel::types::Integer`]: types/struct.Integer.html
+//! [`diesel::sql_types::Integer`]: sql_types/struct.Integer.html
 //! [`ToSql`]: types/trait.ToSql.html
 //! [`FromSql`]: types/trait.FromSql.html
```
weiznich: Also point to the new locations?

diesel/src/sql_types/ord.rs
```diff
@@ -0,0 +1,19 @@
...
+impl SqlOrd for sql_types::Float {}
+impl SqlOrd for sql_types::Double {}
+impl SqlOrd for sql_types::Text {}
+impl SqlOrd for sql_types::Date {}
+impl SqlOrd for sql_types::Interval {}
+impl SqlOrd for sql_types::Time {}
+impl SqlOrd for sql_types::Timestamp {}
+impl<T: SqlOrd + NotNull> SqlOrd for sql_types::Nullable<T> {}
+
+#[cfg(feature = "postgres")]
```
weiznich: All other impl in `sql_types` are unconditional to the enabled backends, so I think this should also be implemented in every case.

diesel/src/type_impls/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod date_and_time;
+pub mod floats;
```
weiznich: floats and option should be private, not public

diesel/src/types/mod.rs
```diff
@@ -1,720 +1,109 @@
...
+#![cfg(feature = "with-deprecated")]
+#![deprecated(since = "1.1.0", note = "Use `sql_types`, `serialize`, or `deserialize` instead")]
 
-pub mod ops;
-mod ord;
-#[macro_use]
-#[doc(hidden)]
-pub mod impls;
-mod fold;
+#[deprecated(since = "1.1.0", note = "Use `sql_types` instead")]
```
weiznich: The message is not 100% correct, the user should use the types inside the `sql_types` module instead the ones inside the `types` module. If I intentionally try to misinterpret the message I read something like: Instead of `types::Integer` use simply `sql_types`"

diesel/src/types/mod.rs
```diff
@@ -1,720 +1,109 @@
...
-        pub use pg::data_types::*;
-    }
-}
+#[deprecated(since = "1.1.0", note = "Use `serialize` instead")]
+pub use serialize::{IsNull, ToSql};
 
-pub use self::ord::SqlOrd;
-pub use self::fold::Foldable;
+#[deprecated(since = "1.1.0", note = "Use `sql_types::Bool` instead")]
+pub type Bool = ::sql_types::Bool;
```
weiznich: We reexport the whole `sql_types` module above so this should not be needed again?
(Same for all other types here)

diesel/src/types/mod.rs
```diff
@@ -1,720 +1,109 @@
...
-#[derive(Debug, Clone, Copy, Default)]
-pub struct Nullable<ST: NotNull>(ST);
-
-#[cfg(feature = "postgres")]
-pub use pg::types::sql_types::*;
+#[deprecated(since = "1.1.0", note = "Use `sql_types::Timestamp` instead")]
+pub type Timestamp = ::sql_types::Timestamp;
 
+#[deprecated(since = "1.1.0", note = "Use `sql_types::Datetime` instead")]
 #[cfg(feature = "mysql")]
```
weiznich: The types in `sql_types` don't have this cfg's anymore, so maybe skip them here too?

diesel_derives/src/as_expression.rs
```diff
@@ -6,7 +6,7 @@ use util::*;
 pub fn derive(item: syn::DeriveInput) -> Tokens {
     let item_name = item.ident.as_ref().to_uppercase();
     let is_sized = !flag_present(&item.attrs, "not_sized");
-    let sql_types = item.attrs
```
weiznich: Why rename this variable? The old name was better :wink:

Eijebong: `sed -i s/types/sql_types/` probably :D

sgrif: Lol whoops

diesel/src/sql_types/ord.rs
sgrif: The `Array` type doesn't exist unless this feature is enabled.

weiznich: Then maybe enable it also unconditionally?
(It seems a bit strange that this type is only there if "postgres" is one of the enabled backend)

diesel/src/type_impls/mod.rs
sgrif: This code was just moved unchanged.

weiznich: Git does not see this as a code-move (it is shown in the diff). Therefore we could make those two also modules private (they do not contain any public items, as far as I have seen)

diesel/src/types/mod.rs
sgrif: This is to work around bugs in Rust.

diesel/src/types/mod.rs
sgrif: The `Datetime` type is a MySQL specific type, and is only present when the `mysql` feature is enabled.

diesel/src/type_impls/mod.rs
sgrif: Yeah, the replacement of `types` with `sql_types` probably touched enough lines for it to not be seen as a rename in git.

PR #1474, opened 2018-01-12
Add a TryFrom<PgNumeric> for BigDecimal impl

Eijebong:
Since TryFrom isn't stable yet, put it behind the diesel unstable
feature. Once stabilized we can remove the code duplication by using
try_from directly within from_sql.

### DIFF
--- a/diesel/src/pg/types/numeric.rs
+++ b/diesel/src/pg/types/numeric.rs
@@ -17,6 +17,11 @@ mod bigdecimal {
     use serialize::{self, Output, ToSql};
     use sql_types::Numeric;
 
+    #[cfg(feature = "unstable")]
+    use std::convert::{TryFrom, TryInto};
+    #[cfg(feature = "unstable")]
+    use std::error::Error;
+
     /// Iterator over the digits of a big uint in base 10k.
     /// The digits will be returned in little endian order.
     struct ToBase10000(Option<BigUint>);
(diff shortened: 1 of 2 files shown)


sgrif:
Rather than using `feature = "unstable"` for this (which IIRC is actually unused right now), can we change this to be `#[cfg(diesel_unstable)]`? Rayon does this and I think it's a really good idea. It requires that the *outermost compilation unit* (aka the actual application) build with `RUSTFLAGS = --cfg diesel_unstable`, which means you can't accidentally transitively depend on an unstable Diesel feature. This mirrors how relying on nightly works (if one of your deps needs nightly, you have to opt into nightly), and makes much more sense than cargo features (which are additive and you can rely on without knowing it)

diesel/src/pg/types/numeric.rs
```diff
@@ -246,5 +293,27 @@ mod bigdecimal {
             };
             assert_eq!(expected, decimal.into());
         }
+
+        #[test]
+        #[cfg(feature = "unstable")]
```
sgrif: Can you confirm this test is actually run by travis?

diesel/src/pg/types/numeric.rs
```diff
@@ -102,6 +148,7 @@ mod bigdecimal {
 
     impl FromSql<types::Numeric, Pg> for BigDecimal {
         fn from_sql(numeric: Option<&[u8]>) -> deserialize::Result<Self> {
+            // FIXME: Use the TryFrom impl when try_from is stable
             let (sign, weight, scale, digits) = match PgNumeric::from_sql(numeric)? {
```
sgrif: This is maybe a nit, but I feel like in the interim we should pull this into a function so we don't *actually* have 40 lines of code copy-pasted like htis

sgrif:
To be clear, this is where I saw the idea to use a cfg flag instead of a Cargo feature, and I think it makes a lot of sense. https://github.com/rayon-rs/rayon#semver-policy-the-rayon-core-crate-and-unstable-features

sgrif:
I was tired when I reviewed this. I retract my thing about not using the unstable feature. This is about unstable Rust not unstable features of the library.

PR #1481, opened 2018-01-15
bugfix: Eliminate shadowing of r2d2 public rexports by diesel types.

YetAnotherMinion:
bug: Alias `r2d2::Error` as `diesel::r2d2::PoolError` to prevent shadowing of public rexports from r2d2 crate by diesel types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## [1.1.1] - 2018-01-15
+
+### Added
+
+* Added `diesel::r2d2::PoolError` as an alias for `r2d2::Error`. Previously this
+  type was inaccessible due to `diesel::r2d2::Error`.
+
 ## [1.1.0] - 2018-01-15
 
 ### Added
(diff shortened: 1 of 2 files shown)


diesel/src/r2d2.rs
```diff
@@ -3,6 +3,8 @@
 extern crate r2d2;
 
 pub use self::r2d2::*;
+/// Rexport r2d2 crate's `Error` to prevent shadowing by this module's `Error` type.
```
sgrif: This reads more like a code comment than documentation. Can we say something like "A re-export of `r2d2::Error`, which is only used by methods on `r2d2::Pool`"?

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## [1.1.1] - 2018-01-15
+
+### Fixed
+
+* `r2d2::Error` is now aliased as `diesel::r2d2::PoolError`. Previously the
```
sgrif: I would change this to `Added` and reword it as

> Added `diesel::r2d2::PoolError` as an alias for `r2d2::Error`. Previously this type was inaccessible due to `diesel::r2d2::Error`.

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All user visible changes to this project will be documented in this file.
...
+  `r2d2::Error` type public reexport in `diesel::r2d2` was shadowed by
+  the module's own Error type.
+
+
```
sgrif: ✂️

Issue #1106, opened 2017-08-15
Support for insert into tbl1 from select * from tbl2 queries


lancecarlson:
Is this something diesel would/should support? (Correct me if Diesel already supports this)

lancecarlson:
Thinking this through, I would imagine the syntax might look like this:

```rust
diesel::insert(diesel::select_table).into(diesel::insert_into_table);
```

I'm not sure that the dsl adds anything over the SQL unless it can generate the columns from the struct because (at least in postgres) you have to cast the select output into a table with an explicit column manifest/schema.

killercup:
If the tables have the exact same columns, you could try

```rust
diesel::insert(table1::table.select(table1::all_columns)).into(table2::table)
```

sgrif:
Update: https://github.com/diesel-rs/diesel/commit/73778af0 was very much in service of this eventual feature. Any version of this that lands will require explicitly listing the columns. e.g. We will not support `INSERT INTO table SELECT ...` but we will support `INSERT INTO table (col1, col2) SELECT ...`.

I would love to see ideas on concrete API proposals for this. Right now the best I've got is:

```rust
insert_into(table1)
    .from_select(
        (col1, col2, col3),
        table2.select(...).etc,
    )
```

I think we can do better though. In particular, I don't like the amount of rightward drift there. It reminds me very much of our old `on_conflict` function from 0.16, which we deprecated in 0.99 for the same ergonomic reasons

sgrif:
I think I'm going to go with this API:

```
insert_into(table1)
    .from_select(table2.select(...))
    .into_columns((col1, col2, col3))
```

The order is backwards from what you'd write in SQL, but it lets us default to `table1::all_columns`, and any method that allowed you to specify the columns doesn't make sense unless you know that you're inserting from a select statement.

sgrif:
So I prototyped this out, and found that the select statement generally was so long we had the same rightward drift problem. You can work around it by sticking stuff in locals, but then the variable is awkward to name and really "you can work around this crappy API" doesn't make it acceptable. @weiznich had a really good idea in the gitter room though. `select_statement.insert_into(table).columns(col_list)`, which is even more backwards from the SQL, but definitely lets you insert line breaks where it's natural. TBH this is also pretty close to how I read these queries anyway, since the select statement is by far the most important part.

[...]

PR #1486, opened 2018-01-17
Remove `Changeset`

sgrif:
This is a long overdue refactoring. I'm going to do a similar
refactoring for `InsertValues` as part of #1106, but this shares a lot
of groundwork for that and is much smaller in scope. `Changeset` has
always been entirely redundant with `QueryFragment`, but we had to keep
it as a separate trait for two places where the implementation changed:

- `Eq` doesn't qualify the column name
- tuples skip commas when `is_noop` returned true (for `Changeset` this
  was only the case for `None` or a tuple where all elements were `None`)

The first one is easy to solve, we just return a type other than `Eq` in
`AsChangeset` (there was very little reason to use `Eq` in the first
place. We weren't able to re-use any of its code.)

The second one is a bit trickier, since we need to replicate the
`is_noop` method in `QueryFragment`. I'm fine with doing this, since
`InsertValues` has the same method (and tuples implement `walk_ast`
identically for `InsertValues` and `Changeset`), so it makes sense to
share them.

Originally I added an explicit `noop` method to `AstPass` which we would
call. However, it felt weird that the impl for `()` which literally just
does `Ok(())` didn't call `noop`. At that point I realized I could just
generalize this to "did a method that generates SQL get called?" This
works fine for updates. I'm not 100% sure if it'll work for inserts, but
it's worth a shot.

I should note that the semantics of the new `is_noop` are slightly
different than the one on `InsertValues`, which explicitly states that
an empty array will return `false`. This will make it return `true` when
we switch inserts to use this. Since we already have to explicitly
handle empty arrays long before we start checking `is_noop`, I think
that's fine.

### DIFF
--- a/diesel/src/query_builder/update_statement/changeset.rs
+++ b/diesel/src/query_builder/update_statement/changeset.rs
@@ -1,6 +1,8 @@
 use backend::Backend;
-use query_builder::AstPass;
-use query_source::QuerySource;
+use expression::AppearsOnTable;
+use expression::operators::Eq;
+use query_builder::*;
+use query_source::{Column, QuerySource};
 use result::QueryResult;
 
 /// Types which can be passed to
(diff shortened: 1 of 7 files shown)


diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -49,15 +40,37 @@ impl<T: AsChangeset> AsChangeset for Option<T> {
...
+    type Target = Left::Table;
+    type Changeset = Assign<Left, Right>;
 
-    fn walk_ast(&self, out: AstPass<DB>) -> QueryResult<()> {
-        match *self {
-            Some(ref c) => c.walk_ast(out),
-            None => Ok(()),
+    fn as_changeset(self) -> Self::Changeset {
+        Assign {
+            _column: self.left,
```
weiznich: Why does this use a underscore name?

sgrif: It's never actually used. Only the type is. I didn't feel like using `PhantomData` for it since we actually have the value at the point of construction.

PR #1488, opened 2018-01-17
Remove most of `InsertValues`

sgrif:
This removes every method on `InsertValues` except for `column_names` on
`InsertValues`. These methods were redundant with the versions on
`QueryFragment`, and we can just directly use those instead.

I'd like to actually get rid of this trait completely, but that is
impossible without a breaking change to the API of `Insertable`. As I
suspected, our implementation of `is_noop` doesn't qutie work here. The
issue is that `walk_ast` always outputs at least `()`, which is not
"nothing". I've basically added a hacky workaround to `Grouped` in the
short term (and added a test demonstrating the problem, we didn't
previously have one).

I think this hack and problem both generally go away by introducing a
`ValuesClause` struct which is responsible for both the `VALUES`
keyword, and the parenthesis. (And I guess probably the column list,
too). All of this is required if we don't want to completely
re-implement `InsertStatement` (and more importantly PG upsert,
`insert_or_ignore`, etc) for to insert from a select statement.

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -1,3 +1,5 @@
+use std::marker::PhantomData;
+
 use backend::{Backend, SupportsDefaultKeyword};
 use expression::{AppearsOnTable, Expression, Grouped};
 use result::QueryResult;
@@ -37,12 +39,12 @@ where
     }
 }
 
-impl<T, DB> CanInsertInSingleQuery<DB> for [T]
+impl<'a, T, Tab, DB> CanInsertInSingleQuery<DB> for BatchInsert<'a, T, Tab>
 where
     DB: Backend + SupportsDefaultKeyword,
 {
     fn rows_to_insert(&self) -> usize {
-        self.len()
+        self.records.len()
     }
 }
 
@@ -65,20 +67,8 @@ where
     }
 }
 
-pub trait InsertValues<T: Table, DB: Backend> {
+pub trait InsertValues<T: Table, DB: Backend>: QueryFragment<DB> {
     fn column_names(&self, out: AstPass<DB>) -> QueryResult<()>;
-    fn walk_ast(&self, out: AstPass<DB>) -> QueryResult<()>;
-
-    /// Whether or not `column_names` and `walk_ast` will perform any action
-    ///
-    /// This method will return `true` for values which semantically represent
-    /// `DEFAULT` on backends which don't support it, or `DEFAULT VALUES` on
-    /// any backend.
-    ///
-    /// Note: This method only has semantic meaning for types which represent a
-    /// single row. Types which represent multiple rows will always return
-    /// `false` for this, even if they will insert 0 rows.
-    fn is_noop(&self) -> bool;
 }
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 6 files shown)


sgrif:
Forgot to mention in the commit message. The reason for the `BatchInsert` struct is that I'm not comfortable implementing `QueryFragment` for slice.

diesel/src/insertable.rs
```diff
@@ -193,31 +188,50 @@ where
...
 }
 
-impl<'a, T, Tab, DB> InsertValues<Tab, DB> for &'a [T]
+#[derive(Debug, Clone, Copy)]
+pub struct BatchInsert<'a, T: 'a, Tab> {
+    records: &'a [T],
+    _marker: PhantomData<Tab>,
+}
+
+impl<'a, T, Tab> IntoIterator for BatchInsert<'a, T, Tab> {
```
weiznich: Why is this impl needed?
You use `records.iter()` below

diesel/src/query_builder/insert_statement.rs
```diff
@@ -176,7 +176,7 @@ where
 }
 
 #[cfg(feature = "sqlite")]
-impl<'a, T, U, Op> ExecuteDsl<SqliteConnection> for InsertStatement<T, &'a [U], Op>
```
weiznich: This is a breaking change, because this will break code that uses this as a trait bound

diesel/src/query_builder/insert_statement.rs
```diff
@@ -335,7 +335,7 @@ where
 {
 }
 
-impl<T, Table> UndecoratedInsertRecord<Table> for [T]
```
weiznich: This is also a breaking change

diesel/src/insertable.rs
sgrif: Need to otherwise access `records` in another file and this felt cleaner than `.records().records`

diesel/src/query_builder/insert_statement.rs
sgrif: This is not a breaking change. There is no way to construct this type in public API other than by constructing it by passing a slice to `values`. All of that continues to work.

diesel/src/query_builder/insert_statement.rs
sgrif: I guess technically? We can keep it around, it literally does nothing though

diesel/src/query_builder/insert_statement.rs
sgrif: I guess I can leave this impl in place -- it's straight up impossible to construct this type though

weiznich: It is not impossible, because I've a codebase where something like this exists:
```rust
trait MyOwnInsert<&'insert, T, Res, E> {
    type Ret;
    fn insert(&'insert self, conn: &SqliteConnection) -> QueryResult<Self::Ret>;
}
```
Now if I implement this trait for `C` where `C: AsRef<[E]> I need (amongst other) the following trait bound:
```rust
 InsertStatement<T, <&'insert [E] as Insertable<T>>::Values>: ExecuteDsl<SqliteConnection>,
```
(Note that this only uses code from our public API)

sgrif: The code you just pasted would continue to work, and is the intended way to do that.

PR #1490, opened 2018-01-17
Handle query strings in database url

### LINKED ISSUE #1480: Query string in database URL is treated as part of the name by diesel database setup
When setting up the database with `diesel database setup` any query string provided as part of the DSN is considered part of the database name. In the following example, I expect a database called "example" to be created:

### Steps to reproduce

```
$ export DATABASE_URL=postgres://postgres@localhost/example?sslmode=disable
$ diesel database setup
Creating database: example?sslmode=disable
$ psql
postgres=# SELECT datname FROM pg_database WHERE datname LIKE 'example%';
        datname         
------------------------
 example?sslmode=disable
```

rokob:
Fixes #1480 

If the DATABASE_URL is set to has a query string, for example:

`postgres://postgres@localhost/example?sslmode=disable`

Then database is parsed to be `example?sslmode=disable` rather than `example`.

I added a test which fails without the additional code in this change. The test ensures that `change_database_of_url` returns the proper database name along with the full DSN including the query string as the second return value.

I used the url crate to handle the parsing, but followed the style of the function that current exists and just used unwrap instead of more friendly error handling.

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -22,6 +22,7 @@ dotenv = ">=0.8, <0.11"
 infer_schema_internals = { version = "1.1.0" }
 clippy = { optional = true, version = "=0.0.174" }
 migrations_internals = { version = "1.1.0" }
+url = { version = "1.4.0", optional = true }
 
 [dev-dependencies]
 difference = "1.0"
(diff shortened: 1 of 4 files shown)


sgrif:
Thanks, but this exact approach has been opened before and rejected. I'd prefer to use an actual URL parser rather than rolling our own.

rokob:
Alrighty, here it is using the url crate. The current function using `pop().unwrap()` led me to believe that the error handling here is to panic on bad input, so I just liberally used unwrap.

diesel_cli/Cargo.toml
```diff
@@ -22,12 +22,12 @@ dotenv = ">=0.8, <0.11"
 infer_schema_internals = { version = "1.1.0" }
 clippy = { optional = true, version = "=0.0.174" }
 migrations_internals = { version = "1.1.0" }
+url = { version = "1.4.0" }
```
sgrif: Should this be optional? It's only used with the mysql and postgres features.

sgrif:

diesel_cli/src/main.rs
```diff
@@ -17,6 +17,8 @@ extern crate diesel;
 extern crate dotenv;
 extern crate infer_schema_internals;
 extern crate migrations_internals;
+#[cfg(any(feature = "mysql", feature = "postgres"))]
```
sgrif: Last nit. This should be `#[cfg(feature = "url")]`.

Issue #1103, opened 2017-08-15
SQLite custom functions


Boscop:
Are there plans to support creating [custom functions](https://sqlite.org/c3ref/create_function.html) when using SQLite?
It would allow using Rust functions in sqlite queries, and can be made accessible in diesel through the `sql_function` macro.

E.g. in C you can define custom sqlite functions like this:
```c
int sqlite3_extension_init(sqlite3*db,const char**err,const struct sqlite3_api_routines*api) {
  sqlite3_api=api;
  sqlite3_create_function(db,"PCRE_COMPILE",2,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_compile,0,0);
  sqlite3_create_function(db,"PCRE_CONFIG",1,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_config,0,0);
  sqlite3_create_function(db,"PCRE_EXEC",-1,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_exec,0,0);
  sqlite3_create_function(db,"PCRE_FIRSTCHAR",1,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_firstchar,0,0);
  sqlite3_create_function(db,"PCRE_QUOTE",1,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_quote,0,0);
  sqlite3_create_function(db,"PCRE_REPLACE",-1,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_pcre_replace,0,0);
  sqlite3_create_function(db,"REGEXP",2,SQLITE_UTF8|SQLITE_DETERMINISTIC,0,fn_regexp,0,0);
  sqlite3_create_module(db,"PCRE_VTABLE",&the_module,0);
  pcre_callout=my_callout;
  return SQLITE_OK;
}
```
[(Source)](http://zzo38computer.org/sql/sqlite.txt)

killercup:
I haven't heard of any plans. But this seems interesting! Are there any sqlite bindings/crates that support this? What is the expected API for this?

Boscop:
The sqlite crate that diesel uses exposes these functions here:
https://github.com/jgallagher/rusqlite/blob/master/libsqlite3-sys/sqlite3/bindgen_bundled_version.rs#L1281-L1360

killercup:
[...]

PR #1492, opened 2018-01-18
Sqlite custom function

maghoff:
This is a work in progress, for issue #1103.

The goal here is to support exposing application defined functions to SQL with SQLite, and this is currently implemented. See [the tests](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/mod.rs#L253-L436) for examples.

There are problems I want to address still, but I'd like some input/discussion for what's a good way to do that.

1. The [`create_scalar_function`](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/mod.rs#L139-L148) API is a bit clunky. It is especially round-about how the arguments to the application-defined function must be queried from the context-object. It would be nice to have an API that would let you stuff in a function, and the argument count and necessary conversions would be discovered automatically. Can we do this?
2. In [FromSqliteValue](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/mod.rs#L139-L148) and [into_sqlite_result](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/into_sqlite_result.rs#L8-L16) I'm passing raw pointers. This is my first foray into `unsafe` code, so I am a bit uncertain. I suppose I should really create types that semantically convey guarantees such as "this is a valid sqlite3_context pointer" and "this is a valid sqlite3_value pointer that is 'protected' in the sqlite3-sense". I am not sure how I would go about doing this without introducing some inadvertent cost, for example when [converting a raw array of `sqlite3_value` pointers](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/app_defined_fn.rs#L60). (Maybe do it one at a time in [fn get(..)](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/app_defined_fn.rs#L37)?)
3. `FromSqliteValue` for `String` and `Vec<u8>` is [not zero cost](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/from_sqlite_value.rs#L46-L51). I think it is impossible to create a safe zero-cost abstraction here. Maybe the ideal API would offer two options: unsafe zero-cost and safe non-zero-cost.
4. `IntoSqliteResult` [should support blob](https://github.com/maghoff/diesel/blob/d200a6344fda92445a3776c0538ad1bda0f9431f/diesel/src/sqlite/connection/into_sqlite_result.rs#L6). I'd like this to be zero-cost, like the CString-impl, but I don't know how to go about it. (I can only pass a (non-fat) pointer into the FFI, and I have to properly drop the value based on this later, but `slice::as_ptr` gives me a fat pointer)
5. I'm not entirely sure that `IntoSqliteResult` is ideal because of the recursive `impl`s of `Option` and `Result`. Not only is it possible to convert an `Option<i32>` into a nullable sqlite result, it is also possible to convert `Option<Option<Option<i32>>>` into one. This doesn't actually make any sense. Is it worth it to add additional types so that while `Option<i32>` works, `Option<Option<i32>>` would fail to compile?

I'm hoping for some input/discussion to help me focus on what's most important going forward. Thanks :)

(CC @killercup )

### DIFF
--- a/diesel/src/sqlite/connection/app_defined_fn.rs
+++ b/diesel/src/sqlite/connection/app_defined_fn.rs
@@ -1,0 +1,69 @@
+//! Plumbing for the implementation of the application defined function API
+
+use std::os::raw as libc;
+
+use super::ffi;
+use super::into_sqlite_result::IntoSqliteResult;
+use super::from_sqlite_value::FromSqliteValue;
+
+/// Context is a wrapper for the SQLite function evaluation context.
+#[derive(Debug)]
+pub struct Context<'a> {
+    ctx: *mut ffi::sqlite3_context,
+    args: &'a [*mut ffi::sqlite3_value],
+}
+
+// Context is translated from rusqlite
+impl<'a> Context<'a> {
+    /// Returns the number of arguments to the function.
+    pub fn len(&self) -> usize {
+        self.args.len()
+    }
+
+    /// Returns `true` when there is no argument.
+    pub fn is_empty(&self) -> bool {
+        self.args.is_empty()
+    }
+
+    /// Returns the `idx`th argument as a `T`.
+    ///
+    /// # Failure
+    ///
+    /// Will panic if `idx` is greater than or equal to `self.len()`.
+    pub fn get<T>(&self, idx: usize) -> T
+    where
+        T: FromSqliteValue
+    {
+        let arg = self.args[idx];
+
+        T::from_sqlite_value(arg)
+    }
+}
+
+pub unsafe extern "C" fn free_boxed_value<T>(p: *mut ::std::os::raw::c_void) {
+    let _: Box<T> = Box::from_raw(::std::mem::transmute(p));
+}
+
+pub unsafe extern "C" fn call_boxed_closure<F, T>(
+    ctx: *mut ffi::sqlite3_context,
+    argc: libc::c_int,
+    argv: *mut *mut ffi::sqlite3_value
+)
+where
+    F: FnMut(&Context) -> T,
+    T: IntoSqliteResult
+{
+    use std::{slice, mem};
+
+    let ctx = Context {
+        ctx: ctx,
+        args: slice::from_raw_parts(argv, argc as usize),
+    };
+
+    let boxed_f: *mut F = mem::transmute(ffi::sqlite3_user_data(ctx.ctx));
+    assert!(!boxed_f.is_null(), "Internal error - null function pointer");
+
+    let t = (*boxed_f)(&ctx);
+
+    t.into_sqlite_result(ctx.ctx);
+}
--- a/diesel/src/sqlite/connection/app_defined_fn.rs
+++ b/diesel/src/sqlite/connection/app_defined_fn.rs
@@ -32,7 +32,7 @@ impl<'a> Context<'a> {
     /// Will panic if `idx` is greater than or equal to `self.len()`.
     pub fn get<T>(&self, idx: usize) -> T
     where
-        T: FromSqliteValue
+        T: FromSqliteValue,
     {
         let arg = self.args[idx];
 
@@ -47,13 +47,12 @@ pub unsafe extern "C" fn free_boxed_value<T>(p: *mut ::std::os::raw::c_void) {
 pub unsafe extern "C" fn call_boxed_closure<F, T>(
     ctx: *mut ffi::sqlite3_context,
     argc: libc::c_int,
-    argv: *mut *mut ffi::sqlite3_value
-)
-where
+    argv: *mut *mut ffi::sqlite3_value,
+) where
     F: FnMut(&Context) -> T,
-    T: IntoSqliteResult
+    T: IntoSqliteResult,
 {
-    use std::{slice, mem};
+    use std::{mem, slice};
 
     let ctx = Context {
         ctx: ctx,
--- a/diesel/src/sqlite/connection/from_sqlite_value.rs
+++ b/diesel/src/sqlite/connection/from_sqlite_value.rs
@@ -1,14 +1,12 @@
-extern crate libsqlite3_sys as ffi;
+use super::ffi;
 
-use std::ffi::CStr;
-use std::os::raw as libc;
-
-// The sqlite3_value_*type*-functions are only safe for so-called "protected" values.
-
-// Per https://www.sqlite.org/c3ref/value.html, values sent to application-defined functions are protected,
-// making it safe to use those values with FromSqliteValue.
-
+/// The sqlite3_value_*type*-functions are only safe for so-called "protected" values.
+///
+/// Per https://www.sqlite.org/c3ref/value.html, values sent to application-defined functions are protected,
+/// making it safe to use those values with FromSqliteValue.
 pub trait FromSqliteValue {
+    /// * `value` must be a valid pointer
+    /// * `value` must be "protected" as defined in the sqlite documentation
     fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self;
 }
 
@@ -50,6 +48,8 @@ impl FromSqliteValue for String {
             // >can be invalidated by a subsequent call to sqlite3_value_bytes(), sqlite3_value_bytes16(),
             // >sqlite3_value_text(), or sqlite3_value_text16().
 
+            // This copy makes this abstraction non-zero cost
+
             let buf = slice::from_raw_parts(ptr, len as usize);
 
             str::from_utf8(buf)
@@ -72,6 +72,8 @@ impl FromSqliteValue for Vec<u8> {
             // >Please pay particular attention to the fact that the pointer returned from ... sqlite3_value_blob(), ...
             // >can be invalidated by a subsequent call to sqlite3_value_bytes(), sqlite3_value_bytes16(),
             // >sqlite3_value_text(), or sqlite3_value_text16().
+
+            // This copy makes this abstraction non-zero cost
 
             let buf = slice::from_raw_parts(ptr as *const u8, len as usize);
 
--- a/diesel/src/sqlite/connection/from_sqlite_value.rs
+++ b/diesel/src/sqlite/connection/from_sqlite_value.rs
@@ -1,0 +1,81 @@
+extern crate libsqlite3_sys as ffi;
+
+use std::ffi::CStr;
+use std::os::raw as libc;
+
+// The sqlite3_value_*type*-functions are only safe for so-called "protected" values.
+
+// Per https://www.sqlite.org/c3ref/value.html, values sent to application-defined functions are protected,
+// making it safe to use those values with FromSqliteValue.
+
+pub trait FromSqliteValue {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self;
+}
+
+impl FromSqliteValue for i32 {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        unsafe {
+            ffi::sqlite3_value_int(value)
+        }
+    }
+}
+
+impl FromSqliteValue for i64 {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        unsafe {
+            ffi::sqlite3_value_int64(value)
+        }
+    }
+}
+
+impl FromSqliteValue for f64 {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        unsafe {
+            ffi::sqlite3_value_double(value)
+        }
+    }
+}
+
+impl FromSqliteValue for String {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        use std::{slice, str};
+
+        unsafe {
+            let len = ffi::sqlite3_value_bytes(value); // Must not be called after sqlite3_value_text
+            let ptr = ffi::sqlite3_value_text(value);
+            assert!(!ptr.is_null());
+
+            // The buffer must be copied immediately (https://www.sqlite.org/c3ref/value_blob.html):
+            // >Please pay particular attention to the fact that the pointer returned from ... sqlite3_value_text(), ...
+            // >can be invalidated by a subsequent call to sqlite3_value_bytes(), sqlite3_value_bytes16(),
+            // >sqlite3_value_text(), or sqlite3_value_text16().
+
+            let buf = slice::from_raw_parts(ptr, len as usize);
+
+            str::from_utf8(buf)
+                .expect("SQLite guarantees UTF-8 encoding")
+                .to_string()
+        }
+    }
+}
+
+impl FromSqliteValue for Vec<u8> {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        use std::slice;
+
+        unsafe {
+            let len = ffi::sqlite3_value_bytes(value); // Must not be called after sqlite3_value_blob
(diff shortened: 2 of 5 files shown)


diesel/src/sqlite/connection/from_sqlite_value.rs
```diff
@@ -0,0 +1,83 @@
+use super::ffi;
+
+/// The sqlite3_value_*type*-functions are only safe for so-called "protected" values.
+///
+/// Per https://www.sqlite.org/c3ref/value.html, values sent to application-defined functions are protected,
+/// making it safe to use those values with FromSqliteValue.
+pub trait FromSqliteValue {
```
sgrif: Why do we need this trait? It's entirely redundant with `FromSql`

diesel/src/sqlite/connection/into_sqlite_result.rs
```diff
@@ -0,0 +1,141 @@
+use std::ffi::{CString, CStr};
+use std::os::raw as libc;
+
+use super::ffi;
+
+// TODO: Support BLOB values
+
+pub trait IntoSqliteResult {
```
sgrif: Do we need a separate trait for this? Can we use `ToSql` instead?

diesel/src/sqlite/connection/from_sqlite_value.rs
maghoff: The `sqlite3_create_function_v2` gives us `sqlite3_value` pointers to work with. As far as I can tell, the only way to get the actual values out of these pointers is with the `sqlite3_value_*` functions. `FromSql` seems to use (via `SqliteValue`) the functions named `sqlite3_column_*`.

It might very well be that I have misunderstood something, but it seems to me that we need this in addition to `FromSql`.

diesel/src/sqlite/connection/into_sqlite_result.rs
maghoff: I'm not sure. I feel that I don't know `ToSql` well enough to give a really decisive response. Again, the point here is to be able to use the correct sqlite API calls, in this case `sqlite3_result_*`. These functions are not called anywhere else in the code base.

Now I'm thinking that you might be mainly concerned about the additional `trait`, not the implementation. I started out trying to make it work with `FromSql` and `ToSql` and couldn't make it. I think they are the slightly wrong abstraction. However, it might be that somebody more familiar with the Diesel code base could be able to make it work.

diesel/src/sqlite/connection/from_sqlite_value.rs
sgrif: Can we change `SqliteValue` to operate on `sqlite3_value_*` functions, and construct it by calling `sqlite3_column_value`?

diesel/src/sqlite/connection/into_sqlite_result.rs
sgrif: `ToSql` always just writes some bytes to a buffer. In the case of `Sqlite`, we then look at the `SqliteType` variant for each value, and figure out the right bind function to call based on that. We should be able to do pretty much the exact same thing here, but calling `sqlite3_result_*` instead.

sgrif: > However, it might be that somebody more familiar with the Diesel code base could be able to make it work.

Yeah, this is a pretty hairy feature to take on for your first PR

diesel/src/sqlite/connection/from_sqlite_value.rs
maghoff: Barring any hidden dragons, I think it would be possible. It would, however, require an extra copy, via [`sqlite3_value_dup`](https://www.sqlite.org/c3ref/value_dup.html). The documentation of [`sqlite3_column_value`](https://www.sqlite.org/c3ref/column_blob.html) seems to recommend against this design:

>**Warning:** The object returned by `sqlite3_column_value()` is an unprotected `sqlite3_value object`. In a multithreaded environment, an unprotected `sqlite3_value` object may only be used safely with `sqlite3_bind_value()` and `sqlite3_result_value()`. If the unprotected `sqlite3_value` object returned by `sqlite3_column_value()` is used in any other way, including calls to routines like `sqlite3_value_int()`, `sqlite3_value_text()`, or `sqlite3_value_bytes()`, the behavior is not threadsafe. Hence, the `sqlite3_column_value()` interface is normally only useful within the implementation of application-defined SQL functions or virtual tables, not within top-level application code.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -129,6 +135,59 @@ impl SqliteConnection {
...
+        F: FnMut(&Context) -> T,
+        T: IntoSqliteResult
+    {
+        // create_scalar_function is translated from rusqlite
+
+        use self::app_defined_fn::{call_boxed_closure, free_boxed_value};
+
+        // https://www.sqlite.org/c3ref/create_function.html
+        // >The length of the name is limited to 255 bytes in a UTF-8 representation, excluding the zero-terminator.
+        assert!(fn_name.as_bytes().len() <= 255);
```
killercup: This is a programmer error, right, as it's only dependent on the function name size? Can you add a custom error message?

diesel/src/sqlite/connection/app_defined_fn.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.args.is_empty()
+    }
+
+    /// Returns the `idx`th argument as a `T`.
+    ///
+    /// # Failure
+    ///
+    /// Will panic if `idx` is greater than or equal to `self.len()`.
+    pub fn get<T>(&self, idx: usize) -> T
```
killercup: It's atypical Rust to have a `get` method panic; it should return an Option, or you should `impl Index for Context`

diesel/src/sqlite/connection/app_defined_fn.rs
```diff
@@ -0,0 +1,69 @@
...
+        T: FromSqliteValue
+    {
+        let arg = self.args[idx];
+
+        T::from_sqlite_value(arg)
+    }
+}
+
+pub unsafe extern "C" fn free_boxed_value<T>(p: *mut ::std::os::raw::c_void) {
+    let _: Box<T> = Box::from_raw(::std::mem::transmute(p));
```
killercup: Can Box::from_raw panic, e.g. when p == NULL? Do we need to deal with this?

diesel/src/sqlite/connection/app_defined_fn.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+pub unsafe extern "C" fn call_boxed_closure<F, T>(
+    ctx: *mut ffi::sqlite3_context,
+    argc: libc::c_int,
+    argv: *mut *mut ffi::sqlite3_value
+)
+where
+    F: FnMut(&Context) -> T,
+    T: IntoSqliteResult
```
killercup: trailing comma

diesel/src/sqlite/connection/app_defined_fn.rs
```diff
@@ -0,0 +1,69 @@
...
+{
+    use std::{slice, mem};
+
+    let ctx = Context {
+        ctx: ctx,
+        args: slice::from_raw_parts(argv, argc as usize),
+    };
+
+    let boxed_f: *mut F = mem::transmute(ffi::sqlite3_user_data(ctx.ctx));
+    assert!(!boxed_f.is_null(), "Internal error - null function pointer");
```
killercup: Can you add a bit more context to the assert message, like "Got a null pointer when expecting a function pointer when calling SQLite custom function"?

Let's step a bit back though: This assert can panic, and thus unwind the stack. But it is in an extern fn that gets called by SQLite, right? Isn't panicking in a C context undefined behavior? Shouldn't we try to catch_unwind this whole function body if possible and set t to an error state when that happens?

diesel/src/sqlite/connection/from_sqlite_value.rs
```diff
@@ -0,0 +1,85 @@
...
+            let ptr = ffi::sqlite3_value_blob(value);
+            let len = ffi::sqlite3_value_bytes(value);
+            assert!(!ptr.is_null());
+
+            // The buffer must be copied immediately (https://www.sqlite.org/c3ref/value_blob.html):
+            // >Please pay particular attention to the fact that the pointer returned from ... sqlite3_value_blob(), ...
+            // >can be invalidated by a subsequent call to sqlite3_value_bytes(), sqlite3_value_bytes16(),
+            // >sqlite3_value_text(), or sqlite3_value_text16().
+
+            // This copy makes this abstraction non-zero cost
```
killercup: Thanks for documenting this <3

diesel/src/sqlite/connection/app_defined_fn.rs
maghoff: In that case, I'm partial to `impl Index`, since `Context` is essentially a wrapper around an array.

diesel/src/sqlite/connection/app_defined_fn.rs
maghoff: Fixed with rustfmt, which took some other style issues as well

diesel/src/sqlite/connection/app_defined_fn.rs
maghoff: Yeah, the message could be improved. This assert can only cause a panic in case of logic errors in diesel or sqlite, but it still makes sense to have a descriptive error message.

I like your step back. I think catch_unwind-ing could be the correct way to go here.

diesel/src/sqlite/connection/app_defined_fn.rs
sgrif: I'd prefer we have this function take `*mut T` and cast the function pointer rather than the other way around. The call to `transmute` is too likely to accidentally cause problems in the future.

diesel/src/sqlite/connection/from_sqlite_value.rs
sgrif: > Hence, the sqlite3_column_value() interface is normally only useful within the implementation of application-defined SQL functions

Isn't that exactly what we're doing? `unprotected` just means that there's no mutex around the value, but that literally just means that it needs to be `!Sync` (which it will be by default if it involves a pointer)

diesel/src/sqlite/connection/from_sqlite_value.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+}
+
+impl FromSqliteValue for Vec<u8> {
+    fn from_sqlite_value(value: *mut ffi::sqlite3_value) -> Self {
+        use std::slice;
+
+        unsafe {
+            // Call sqlite3_value_blob(), then _bytes(), as per https://www.sqlite.org/c3ref/column_blob.html
```
sgrif: Either this comment is wrong, the comment below is wrong, or SQLite's docs are wrong. Both the comment below and SQLite's docs claim that calling `sqlite3_value_bytes` can invalidate the pointer returned by `sqlite3_value_blob`. This is of course identical to the implementation in Diesel today, which means that this definitely isn't happening (or nobody is using binary on SQLite with Diesel... Or we're relying on UB).. Either way, something is lying. I'd like to either fix the comment here, or fix the implementation.

sgrif:
@killercup I want to land it incrementally, but I think we need to answer some higher level questions about the API. I also have some strong concerns with this implementation. I'm not convinced that there's any reason to duplicate the functionality of `FromSqlRow` and `ToSql` here.

I don't like the API that's exposed. We should be able to figure out the number of arguments based on the type of the function given. I don't think we should be exposing anything like `Context`. We should deal with conversions on our end, and just give the user what they ask for.

There are also major memory safety issues in this implementation. At absolute minimum we need to be enforcing that both the function given and it's return type are `'static`. I need to look more closely to see if there are any other issues in that regard.

diesel/src/sqlite/connection/from_sqlite_value.rs
maghoff: >Isn't that exactly what we're doing?
Of course it is. The question I am trying to address is whether or not it is a good idea to change the preexisting `FromSql` code for handling the column values from normal database reads.

>`unprotected` just means that there's no mutex around the value
Ok. The way I read the documentation it seemed to me that the only safe way to access the value of an `unprotected` `sqlite3_value` would be via `sqlite3_value_dup`. I see now that the documentation is not as strict, it seems we can handle the thread safety ourselves. Good.

I'll look into refactoring to using `sqlite3_value`s for columns as well, so we can get rid of the duplication.

diesel/src/sqlite/connection/from_sqlite_value.rs
maghoff: Yeah, the SQLite documentation is confusing at best, contradictory at worst.

https://www.sqlite.org/c3ref/column_blob.html says:

>The safest policy is to invoke these routines in one of the following ways:
>
> * sqlite3_column_text() followed by sqlite3_column_bytes()
> * sqlite3_column_blob() followed by sqlite3_column_bytes()
> * sqlite3_column_text16() followed by sqlite3_column_bytes16()
>
>In other words, you should call sqlite3_column_text(), sqlite3_column_blob(), or sqlite3_column_text16() first to force the result into the desired format, then invoke sqlite3_column_bytes() or sqlite3_column_bytes16() to find the size of the result. Do not mix calls to sqlite3_column_text() or sqlite3_column_blob() with calls to sqlite3_column_bytes16(), and do not mix calls to sqlite3_column_text16() with calls to sqlite3_column_bytes().

https://www.sqlite.org/c3ref/value_blob.html says:
>Please pay particular attention to the fact that the pointer returned from sqlite3_value_blob(), sqlite3_value_text(), or sqlite3_value_text16() can be invalidated by a subsequent call to sqlite3_value_bytes(), sqlite3_value_bytes16(), sqlite3_value_text(), or sqlite3_value_text16().

It seems that the documentation is imprecise and that the current implementation is recommended. It looks like the problems would only appear if we cause SQLite to transform a data representation, but they stop short of actually promising that in the documentation.

Perhaps we could add a comment explaining that while the documentation looks contradictory, we trust this implementation because it is clearly recommended by the documentation in https://www.sqlite.org/c3ref/column_blob.html ?

maghoff:
FWIW I agree with all of @sgrif's points about what needs improvement. This is a work in progress that I posted as a PR for discussion, as recommended by @killercup 🙂

sgrif:
@maghoff What's your plan here? Do you want to continue working on this? If so I think we should get you in touch with someone on the team to help implement the rest of this (and probably start by agreeing on an API to target before anything else)

maghoff:
@sgrif I was planning on refactoring to unify FromSql and FromSqliteValue. I think I have enough to go on to succeed with that. And then we could get me in touch with someone on the team?

sgrif:
This was implemented in #1691

maghoff:
Thanks for this feature, and sorry for disappearing.

I've been able to make use of this in [sausagewiki](https://github.com/maghoff/sausagewiki) and Diesel 1.3 solves my problem nicely! 🙂

PR #1493, opened 2018-01-18
Move the responsibility of the `VALUES` keyword off of `InsertStatement`

sgrif:
This change is in order to allow a select statement to be used instead.
There's a little bit of duplication between this and batch insert, but
I'm relatively comfortable with it overall. This should be the last
structural change required to support insert from select

We can safely make the change to this impl, since it has bounds on
traits that are not public API. The only types which could have been
used here are types defined by Diesel.

### DIFF
--- a/diesel/src/query_builder/insert_statement.rs
+++ b/diesel/src/query_builder/insert_statement.rs
@@ -1,4 +1,5 @@
 use std::any::*;
+use std::marker::PhantomData;
 
 use backend::Backend;
 use expression::{Expression, NonAggregate, SelectableExpression};
@@ -145,7 +146,7 @@ where
     DB: Backend,
     T: Table,
     T::FromClause: QueryFragment<DB>,
-    U: InsertValues<T, DB> + CanInsertInSingleQuery<DB>,
+    U: QueryFragment<DB> + CanInsertInSingleQuery<DB>,
     Op: QueryFragment<DB>,
     Ret: QueryFragment<DB>,
 {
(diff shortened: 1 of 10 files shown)


diesel/src/query_builder/insert_statement.rs
```diff
@@ -398,38 +399,77 @@ impl<'a, Tab> Insertable<Tab> for &'a DefaultValues {
...
+}
+
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy)]
+pub struct ValuesClause<T, Tab> {
+    pub(crate) values: T,
+    _marker: PhantomData<Tab>,
+}
+
+impl<T: Default, Tab> Default for ValuesClause<T, Tab> {
```
weiznich: Couldn't we just `#[derive(Default)]`?

diesel_migrations/migrations_internals/src/connection.rs
```diff
@@ -26,7 +25,7 @@ where
     T: Connection,
     String: FromSql<VarChar, T::Backend>,
     // FIXME: HRTB is preventing projecting on any associated types here
-    for<'a> InsertStatement<__diesel_schema_migrations, Grouped<ColumnInsertValue<version, &'a Bound<VarChar, &'a str>>>>: ExecuteDsl<T>,
+    for<'a> InsertStatement<__diesel_schema_migrations, ValuesClause<ColumnInsertValue<version, &'a Bound<VarChar, &'a str>>, __diesel_schema_migrations>>: ExecuteDsl<T>,
```
weiznich: Shouldn't this not better use `<InsertType as Insertable<__diesel_schema_migrations>>::Values`?

sgrif: See the comment on the line above

diesel/src/query_builder/insert_statement.rs
sgrif: That would require `Tab: Default`

PR #1494, opened 2018-01-18
Partially fix our test suite on beta/nightly

sgrif:
Right now on beta the behavior of `file!` (or more specifically, the
cwd when compiling the crate) has changed when running `cargo test`.
It's unclear whether that change will be reverted or not, but it does
point out how brittle these tests are. We want an absolute path, not a
relative one, so using `CARGO_MANIFEST_DIR` instead seems like the way
to go.


sgrif:
cc @alexcrichton I can't actually test this on beta/nightly due to the other regressions, but does this seem more reasonable to you? There's no chance of this breaking in the future, right?

alexcrichton:
@sgrif yeah `CARGO_MANIFEST_DIR` should always be an absolute path

Issue #1498, opened 2018-01-19
Do we need ToSql Cow to require T::Owned = ToSql?


SoniEx2:
## Problem Description

It seems redundant (and annoying) that T::Owned needs ToSql, especially considering that there isn't a blanket ToSql for Borrow types. Do we need this requirement?

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

sgrif:
Good point. No we don't.

sgrif:
Fixed by #1499

PR #1499, opened 2018-01-19
Remove ToSql impl for Cow

jbcden:
This addresses issue https://github.com/diesel-rs/diesel/issues/1498.


sgrif:
We don't want to remove the impl. We just want to remove the requirement that `T::Owned: ToSql`

jbcden:
@sgrif Sorry about that I'm still pretty new to Rust. I just added 7dd2853. If this still isn't quite right I can drop into Gitter and perhaps we can talk it over a bit.

sgrif:
Yeah, that sounds like a good idea to me. :)

PR #1504, opened 2018-01-22
Port #[derive(Insertable)] to proc macro derive

weiznich:


diesel_derives/src/attr.rs
```diff
@@ -5,8 +5,8 @@ use util::*;
 
 #[derive(Debug)]
 pub struct Attr {
-    column_name: Option<syn::Ident>,
-    field_name: Option<syn::Ident>,
+    pub column_name: Option<syn::Ident>,
```
sgrif: Why do we need to make this public? Things should be using the `column_name()` method.

diesel_derives/src/attr.rs
```diff
@@ -5,8 +5,8 @@ use util::*;
...
+    pub field_name: Option<syn::Ident>,
```
sgrif: Why do we need to make this public? Things should be using the `field_name()` method.

diesel_derives/src/attr.rs
weiznich: Because otherwise the shorthand pattern for the destruction of the struct won't work. 
(In detail [this](https://github.com/diesel-rs/diesel/pull/1504/files#diff-cecdfd4f581eb6c37ace4cbea4fc8575R80) line would not be possible without making that thing public)
Maybe this should be moved to `Attr` as member funtcion?

sgrif: I don't see any shorthand destructuring of the struct. That line can and should be using the method form.

> Maybe this should be moved to Attr as member funtcion?

It already is there as a method. That is my point.

diesel_derives/src/attr.rs
sgrif: Same answer...

diesel_derives/src/model.rs
```diff
@@ -55,6 +62,29 @@ impl Model {
         let model_name = self.name.as_ref().to_uppercase();
         format!("_IMPL_{}_FOR_{}", trait_name, model_name).into()
     }
+
+    pub fn destruct_pattern(&self) -> ::quote::Tokens {
```
sgrif: We don't need this method. `let #struct_name { #(#field_name: #column_name,)* }` works in all cases. Alternatively, just skip destructuring entirely (which was mostly done to make the `macro_rules!` form easier), and use `self.#field_name`

diesel_derives/src/model.rs
```diff
@@ -10,18 +10,25 @@ pub struct Model {
     pub generics: syn::Generics,
     pub primary_key_names: Vec<syn::Ident>,
     table_name_from_annotation: Option<syn::Ident>,
+    is_tuple_struct: bool,
 }
 
 impl Model {
     pub fn from_item(item: &syn::DeriveInput, derived_from: &str) -> Result<Self, String> {
-        let attrs = match item.body {
+        let (attrs, is_tuple_struct) = match item.body {
```
sgrif: We should never need a conditional on whether something is a tuple struct.

PR #1506, opened 2018-01-23
Display the correct migration order in `diesel migration list`

### LINKED ISSUE #1500: diesel lists migrations in wrong chronological order
Diesel lists migrations in wrong chronological order after the date format changed.
It lists `2017-11-23-064836` before `20170505070309`:
http://pastebin.centos.org/504376/64198071/

sgrif:
The old code was incorrectly assuming that the migration version was
equal to the file name, which caused issues on projects that have
migrations using the old timestamp format. This continues to display the
actual file name, but sorts on migration version instead.

Fixes #1500

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -77,33 +77,22 @@ fn run_migration_command(matches: &ArgMatches) {
             call_with_conn!(database_url, redo_latest_migration(&dir));
         }
         ("list", Some(_)) => {
-            use std::ffi::OsStr;
-
             let database_url = database::database_url(matches);
             let dir = migrations_dir(matches);
-            let migrations =
+            let mut migrations =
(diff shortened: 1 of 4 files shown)


diesel_cli/src/main.rs
```diff
@@ -77,33 +77,22 @@ fn run_migration_command(matches: &ArgMatches) {
...
-
-            sorted.sort();
+            migrations.sort_by_key(|&(ref m, _)| m.version().to_string());
 
             println!("Migrations:");
-            for (name, applied) in sorted {
+            for (migration, applied) in migrations {
+                let name = migration
+                    .file_path()
+                    .unwrap()
```
killercup: keep the nicer errors?

sgrif: Migrations that came from a `_in_directory` function always have a file path.

PR #1512, opened 2018-01-24
Rewrite `AsChangeset` using syn 0.12

sgrif:
In order to update to syn 0.12, we're going to have to rewrite a lot of
`diesel_derives`. A lot of code is incompatible with no obvious
migration path. I also want to take this as an opportunity to do a lot
of long needed refactoring to this library.

Despite continuous assertions that outputting `compile_error!` somehow
makes our error messages significantly better (it really doesn't, it
just avoids "proc-macro panicked") -- I am panicking in error cases
here. One of the benefits of syn 0.12 is that it retains span
information, which we can use on nightly to output errors with real
diagnostics. I haven't made that jump here quite yet, but I plan to in a
followup PR.

Since this is the first derive moved over to the new crate, it includes
much more code than PRs for the rest will.


diesel_derives2/src/model.rs
```diff
@@ -0,0 +1,82 @@
...
+        let name = self.name.as_ref().to_uppercase();
+        format!("_IMPL_{}_FOR_{}", trait_name, name).into()
+    }
+
+    pub fn fields(&self) -> &[Field] {
+        &self.fields
+    }
+}
+
+pub fn camel_to_snake(name: &str) -> String {
```
sgrif: This function and `infer_table_name` were both copied over from the old codebase. I think I want to deprecate the inferred table name behavior (I want to hold off on that until everything is moved over though)

sgrif:
After playing around with [`Diagnostic`](https://doc.rust-lang.org/nightly/proc_macro/struct.Diagnostic.html), I need to structure this to return errors, not panic.

sgrif:
I guess I'll restructure to not-panic separately..

diesel_derives2/src/as_changeset.rs
```diff
@@ -0,0 +1,91 @@
...
+                .expect_bool_value()
+        })
+        .unwrap_or(false);
+    let model = Model::from_item(&item);
+    let struct_name = &model.name;
+    let table_name = model.table_name();
+
+    let (_, ty_generics, where_clause) = item.generics.split_for_impl();
+    let mut impl_generics = item.generics.clone();
+    impl_generics.params.push(parse_quote!('update));
```
dtolnay: You are going to want another split_for_impl after this line.

```rust
let (impl_generics, _, _) = impl_generics.split_for_impl();
```

Otherwise it does not work for types that have generic type parameter defaults.

```rust
#[derive(AsChangeset)]
#[table_name = "users"]
#[primary_key(name)]
struct Changes2<T = String> {
    name: T,
    hair_color: String,
}
```

```
error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions.
   --> tests/update.rs:390:14
    |
390 |     #[derive(AsChangeset)]
    |              ^^^^^^^^^^^
    |
    = note: #[deny(invalid_type_param_default)] on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #36887 <https://github.com/rust-lang/rust/issues/36887>
```

diesel_derives2/src/field.rs
```diff
@@ -0,0 +1,56 @@
...
+    column_name_from_attribute: Option<syn::Ident>,
+    identifier: Identifier,
+}
+
+impl Field {
+    pub fn from_struct_field(field: &syn::Field, index: usize) -> Self {
+        let column_name_from_attribute = MetaItem::with_name(&field.attrs, "column_name")
+            .ok()
+            .map(|m| m.expect_ident_value());
+        let identifier = match field.ident {
```
weiznich: Wouldn't 
```rust
field.ident.map(Identifier::Named).unwrap_or_else(|| Identifier::Unnamed(index.to_string().into())
```
be a bit more idiomatic?

diesel_derives2/src/field.rs
```diff
@@ -0,0 +1,56 @@
...
+        };
+
+        Self {
+            ty: field.ty.clone(),
+            column_name_from_attribute,
+            identifier,
+        }
+    }
+
+    pub fn column_name(&self) -> &syn::Ident {
```
weiznich: `Ident` seems to be `Copy` now. Maybe return no reference than?

diesel_derives2/src/as_changeset.rs
sgrif: This code doesn't currently work with generic structs, but good catch.

diesel_derives2/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "1.1.0"
+authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+documentation = "https://diesel.rs/guides/"
+homepage = "https://diesel.rs"
+repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
+
+[dependencies]
+syn = { version = "0.12.0", features = ["full"] }
```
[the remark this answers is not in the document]
sgrif: Our test suite fails without the full feature.

PR #1519, opened 2018-01-31
Move various `AsChangeset` tests into derives2, more exhaustively test

sgrif:
The tests for `AsChangeset` mostly lived in the integration suite (there
is still at least one test that is redundant there, but I do want the
integration suite to have at least one "AsChangeset works" test).

I have not moved the tests that are specific to `save_changes` over,
since those have nothing to do with codegen. I specifically structured
the tests so that the *only* derive used is `AsChangeset`, since these
are meant to be unit tests for that derive.

I've added a few tests that weren't covered before (like a tuple
struct), and in the process learned that tuple structs were broken. I
haven't tested anything that we don't explicitly support yet or that we
know is broken, like generic structs or structs with where clauses.

### DIFF
--- a/diesel_derives2/tests/helpers.rs
+++ b/diesel_derives2/tests/helpers.rs
@@ -1,0 +1,76 @@
+use diesel::prelude::*;
+use std::sync::{Once, ONCE_INIT};
+
+cfg_if! {
+    if #[cfg(feature = "sqlite")] {
+        embed_migrations!("../migrations/sqlite");
+
+        pub type TestConnection = SqliteConnection;
+
+        fn connection_no_transaction() -> TestConnection {
+            SqliteConnection::establish(":memory:").unwrap()
+        }
+    } else if #[cfg(feature = "postgres")] {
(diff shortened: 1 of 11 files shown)


diesel_derives2/tests/helpers.rs
```diff
@@ -0,0 +1,72 @@
...
+            ex. cargo test --features \"mysql postgres sqlite\"\n"
+        );
+     }
+}
+
+static RUN_MIGRATIONS: Once = ONCE_INIT;
+
+pub fn connection() -> TestConnection {
+    let connection = connection_no_transaction();
+    RUN_MIGRATIONS.call_once(|| {
```
weiznich: This causes failures in all but the first test for sqlite, because each test is executed in a new database.
I would vote for simply calling `embedded_migrations::run()` unconditionally.

sgrif: I stuck an `if cfg!` in there. Running it unconditionally requires additional synchronization that would have been more painful.

PR #1520, opened 2018-02-01
Enable 32 column tables by default, move features around

sgrif:
This deprecates the names `large-tables` and `huge-tables` in favor of
`32-column-tables` and `64-column-tables`. `32-column-tables` is enabled
by default. On my machine this changes the compile time of Diesel from 8
seconds to 18 seconds, which I think is an acceptable loss. The feature
can still be turned off for cases where compile times are important.

The way we do the deprecation warning is really funky but I don't know a
better way to do it


sgrif:
Release mode timings:

- 16 columns
  - 9.87 secs
- 32 columns
  - 20.56 secs
- 64 columns
  - 82.36 secs
- 128 columns
  - ¯\\\_(ツ)\_/¯ probably forever

sgrif:
Note: Diesel itself has so little code that release mode doesn't generally affect it. Only release mode for libs using Diesel.

kushwahashiv:
> 128 columns
> * ¯\_(ツ)_/¯ probably forever

that's true. I approve that by my experience and if you are building a docker image with 128 you will never be able to deploy in production for sure.

PR #1521, opened 2018-02-01
Get diesel_derives2 working with hygiene turned on

sgrif:
When `proc_macro2` has the nightly feature enabled, a lot of the rules
around hygiene change. Most of the changes needed appear to be due to
bugs in Rust (e.g. to access a tuple struct field, the span of the dot has to
be resolved specially, but not for a named struct field), but the span
changes for table names in particular is definitely required.

The change to hygiene rules randomly broke our const workaround (the
ability to `use` items defined in an expression like that was
potentially unintended to begin with). However, there's no reason it
needs to be a const, we can just have it be a module with no additional
problems.

I've added a few UI tests for the failure cases that are actually
affected by this change (there is more work to be done on this, of
course). Note that we can't actually run those tests yet, as they
require a nightly compiler, and we won't build on nightly until
tomorrow. However, I've left the tests in place so that the changes in
output can be seen.

The error for `table_name` isn't quite there, but due to a bug in Rust
it's the best we can do. Right now it looks like this:

```
1 | #[table_name = "users"]
  | ^ Use of undeclared type or module `users`
```

instead of like this:

```
1 | #[table_name = "users"]
  |                ^^^^^^^ Use of undeclared type or module `users`
```

### DIFF
--- a/diesel_derives2/src/field.rs
+++ b/diesel_derives2/src/field.rs
@@ -1,5 +1,6 @@
-use syn;
+use proc_macro2::Span;
 use quote;
+use syn;
 
 use meta::*;
 
@@ -16,7 +17,11 @@ impl Field {
             .map(|m| m.expect_ident_value());
         let name = match field.ident {
             Some(x) => FieldName::Named(x),
-            None => FieldName::Unnamed(index.into()),
+            None => FieldName::Unnamed(syn::Index {
+                index: index as u32,
+                // https://github.com/rust-lang/rust/issues/47312
+                span: Span::call_site(),
+            }),
         };
 
         Self {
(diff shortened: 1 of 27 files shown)


sgrif:
Note: Build is going to be red until tomorrow. Nothing I can do about that, but this should be fine to review

dtolnay:
What happens tomorrow?

sgrif:
https://github.com/rust-lang/rust/pull/47738 will be in tomorrow's nightly.

diesel_derives2/src/field.rs
```diff
@@ -42,6 +47,13 @@ pub enum FieldName {
 
 impl quote::ToTokens for FieldName {
     fn to_tokens(&self, tokens: &mut quote::Tokens) {
+        use proc_macro2::{Spacing, TokenNode, TokenTree};
+
+        // https://github.com/rust-lang/rust/issues/47312
+        tokens.append(TokenTree {
```
dtolnay: [`quote_spanned!`](https://docs.rs/quote/0.4/quote/macro.quote_spanned.html) can make this more concise but it is up to you whether you prefer the explicit way or the concise way.

```rust
quote_spanned!(call_site=> .)
```

PR #1522, opened 2018-02-02
Improve the errors generated by `#[derive(AsChangeset)]` when possible

sgrif:
This removes all uses of `panic!` or `println!` from derives2, and
instead using the nightly-only `Diagnostic` API when the nightly feature
is enabled. This does require a change in our structure, as anywhere
that we were relying on `panic!` halting, we either need to return some
"default" value, or use `Result` to halt instead.

Whether we actually want to halt or not for each case is actually not a
simple question. Ideally we only halt if continuing with a dummy would
result in nonsense errors. I've added UI tests for every single place
that previously panicked, and I'm reasonably happy with the error output
of every single one.

### DIFF
--- a/diesel_compile_tests/tests/ui/as_changeset_bad_options.stderr
+++ b/diesel_compile_tests/tests/ui/as_changeset_bad_options.stderr
@@ -1,0 +1,38 @@
+error: `treat_none_as_null` must be in the form `treat_none_as_null = "true"`
+ --> <macro expansion>:1:1
+  |
+1 | #[changeset_options(treat_none_as_null("true"))]
+  | ^
+
+error: `treat_none_as_null` must be in the form `treat_none_as_null = "true"`
+ --> <macro expansion>:1:1
+  |
+1 | #[changeset_options(treat_none_as_null)]
+  | ^
+
+error: `treat_none_as_null` must be in the form `treat_none_as_null = "true"`
+ --> <macro expansion>:1:1
+  |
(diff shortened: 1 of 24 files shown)


diesel_derives2/src/field.rs
```diff
@@ -35,7 +36,15 @@ impl Field {
...
                 FieldName::Named(x) => x,
-                _ => panic!("All fields of tuple structs must be annotated with `#[column_name]`"),
+                _ => {
+                    self.ty
+                        .span()
+                        .error(
+                            "All fields of tuple structs must be annotated with `#[column_name]`",
+                        )
+                        .emit();
+                    "unknown_column".into()
```
weiznich: Maybe this should explicitly be replaced by a call to `unreachable!()`

diesel_compile_tests/tests/ui/as_changeset_bad_options.stderr
```diff
@@ -0,0 +1,38 @@
+error: `treat_none_as_null` must be in the form `treat_none_as_null = "true"`
+ --> <macro expansion>:1:1
+  |
+1 | #[changeset_options(treat_none_as_null("true"))]
```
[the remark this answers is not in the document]
sgrif: No, see the comment and code related to this which point at a rustc bug.

diesel_derives2/src/field.rs
sgrif: It's not unreachable.

PR #1523, opened 2018-02-02
Rewrite `#[derive(AsExpression)]` in derives2

sgrif:
This rewrite was pretty straightforward. Getting the span correct on the
type required a bit of work. I'm not sure if there's a better way to do
this or not (perhaps syn just needs a `parse_str_span` function).

This is also the first derive that's used by Diesel itself to get moved.
This required some funkyness in the module wrapping. If
`__diesel_use_everything!()` has `Span::def_site()`, for some reason
that macro doesn't resolve when used inside of Diesel (but it works fine
when used from a third crate). However, if the `;` after it *doesn't*
have `Span::def_site()`, none of the imports are visible to the macro...
This was fun to figure out.

This derive is pretty lightly tested, since we use it for *everything*
in Diesel itself, and that leads to a pretty exhaustive test that way.
This macro also has far fewer error cases than "model" derives.


diesel_derives2/src/as_expression.rs
```diff
@@ -0,0 +1,122 @@
...
+    let dummy_mod = format!(
+        "_impl_as_expression_for_{}",
+        item.ident.as_ref().to_lowercase()
+    );
+    let flags =
+        MetaItem::with_name(&item.attrs, "diesel").unwrap_or_else(|| MetaItem::empty("diesel"));
+    let is_sized = !flags.has_flag("not_sized");
+
+    let sql_types = MetaItem::all_with_name(&item.attrs, "sql_type");
+    let any_sql_types = !sql_types.is_empty();
```
Eijebong: What's the point of `AsExpression` without any sql_type ? Should we maybe make this an error instead ?

sgrif: The point would be this:

```
#[derive(Debug, AsExpression)]
#[cfg_attr(something, sql_type = "SomeType")]
struct Foo;
```

If we error here, we force them to write

```
#[derive(Debug)]
#[cfg_attr(something, derive(AsExpression))]
#[cfg_attr(something, sql_type = "SomeType")]
struct Foo;
```

An example of this in Diesel is:

```
    #[derive(FromSqlRow, AsExpression)]
    #[diesel(foreign_derive)]
    #[cfg_attr(feature = "mysql", sql_type = "::sql_types::Tinyint")]
    struct I8Proxy(i8);
```

sgrif: Also it takes more code to error here than to generate no code. ;)

diesel_compile_tests/tests/ui/as_expression_bad_sql_type.rs
```diff
@@ -0,0 +1,11 @@
+#[macro_use]
+extern crate diesel;
+
+#[derive(AsExpression)]
+#[sql_type(Foo)]
+#[sql_type]
+#[sql_type = "@%&&*"]
+#[sql_type = "1omg"]
+struct Lol;
```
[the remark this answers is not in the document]
sgrif: Lol

PR #1525, opened 2018-02-02
Rewrite `#[derive(FromSqlRow)]` in derives2

sgrif:
This is by far our simplest derive, and didn't even need an explicit
compiletest. There are no explicit unit tests for this derive, as it's
very exhaustively tested by its use in Diesel itself.

For derives that are used both inside of Diesel and out, hygiene can be
really finicky. In this case the requirements of
`__diesel_use_everything!()` were really weird. The macro itself needs
to have `call_site`, but the semicolon after it needs to have
`def_site`.


diesel_derives2/src/from_sql_row.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use meta::*;
+use util::*;
+
+pub fn derive(mut item: syn::DeriveInput) -> Result<Tokens, Diagnostic> {
+    let flags =
+        MetaItem::with_name(&item.attrs, "diesel").unwrap_or_else(|| MetaItem::empty("diesel"));
+    let struct_ty = ty_for_foreign_derive(&item, &flags)?;
+
+    item.generics.params.push(parse_quote!(__ST));
```
Eijebong: `parse_quote!` really is awesome ❤️

diesel_derives2/src/from_sql_row.rs
```diff
@@ -0,0 +1,57 @@
...
+    item.generics.params.push(parse_quote!(__DB));
+    {
```
Eijebong: Why is this in a block ?

diesel_derives2/src/util.rs
```diff
@@ -1,16 +1,20 @@
...
+use proc_macro2::Span;
 use quote::Tokens;
+use syn::*;
 
 pub use diagnostic_shim::*;
+use meta::*;
 
 pub fn wrap_in_dummy_mod(const_name: Ident, item: Tokens) -> Tokens {
+    let call_site = Span::call_site();
+    let use_everything = quote_spanned!(call_site=> __diesel_use_everything!());
```
Eijebong: `call_site => __diesel_use_everything!()` ? It looks weird without the space

diesel_derives2/src/from_sql_row.rs
sgrif: `where_clause` is an exclusive reference to a member of `item.generics`. The line after this block ends creates a shared reference to `item.generics`. I believe this block can go away when NLL lands.

diesel_derives2/src/util.rs
sgrif: I agree. Blame @dtolnay. https://docs.rs/quote/0.4.2/quote/macro.quote_spanned.html

> There should be no space before the => token.

sgrif: (But maybe we should just ignore that because I agree it looks really wrong without a space)

sgrif:
For reference, I think the span requirements for this might be a bug in rustc. If so, https://github.com/rust-lang/rust/issues/47988 is the issue.

diesel_derives2/src/util.rs
dtolnay: Yeah the design constraints here are a bit unusual. If you have something like `quote_spanned!(a ??? b)` the objective was to *make* it look visibly off-balance and draw the eye a particular way, due to `a` being evaluated in the context of the procedural macro and `b` being evaluated in the generated code. I am happy with this design and I believe it has the intended effect.

sgrif: Thanks for the insight. <3

PR #1527, opened 2018-02-02
Rewrite `#[derive(Insertable)]` in derives2

sgrif:
I didn't realize just how much we'd simplified this derive at this
point. This used to be by far our most involved derive.

I've explicitly opted not to include a UI test for when there's a type
mismatch, as I cannot get errors that I'm satisfied with yet, and I
suspect that it will have to wait for some changes in the language.


diesel_derives2/src/insertable.rs
```diff
@@ -0,0 +1,82 @@
...
+use field::*;
+use model::*;
+use util::*;
+
+pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
+    let model = Model::from_item(&item)?;
+
+    if model.fields().is_empty() {
+        return Err(Span::call_site()
+            .error("Cannot derive Insertable for unit structs")
```
retep007: I think the previous error was slightly better, however what about explicitly saying `Cannot derive Insertable for empty Structs`

diesel_derives2/src/insertable.rs
```diff
@@ -0,0 +1,82 @@
...
+            .help(format!(
+                "Use `insert_into({}::table).default_values()` if you want `DEFAULT VALUES`",
```
retep007: How does this help me when my structure fields are missing? What is the use case?

sgrif: I'm assuming that if you derive this on a struct with no fields, you're likely trying to do `INSERT DEFAULT VALUES`. Otherwise I'm not sure what they're trying to accomplish.

PR #1528, opened 2018-02-02
Rewrite `#[derive(QueryId)]` in derives2

sgrif:
This was a really weird one to get working with hygiene turned on. I'm
not sure if this is a bug in rustc, or something we need to explicitly
deal with. Until we know for sure, we need to do this hack to make this
work. At the moment proc_macro2 doesn't expose `Span::source` even with
`procmacro2_semver_exempt`.


diesel_derives2/src/util.rs
```diff
@@ -45,3 +47,18 @@ fn option_ty_arg(ty: &Type) -> Option<&Type> {
...
+}
+
+#[cfg(feature = "nightly")]
+/// There's an issue with the resolution of `__diesel_use_everything` if the
+/// derive itself was generated from within a macro. This is a shitty workaround
+/// until we figure out the expected behavior.
+fn root_span(span: Span) -> Span {
+    use std::mem;
+
+    unsafe { mem::transmute(span.unstable().source()) }
```
sgrif: @dtolnay any idea on whether this is a bug or something I actually need to handle? The problem comes from code that looks like this:

```rust
macro_rules! gen_struct {
    () => {
        #[derive(QueryId)]
        struct Foo;
}

gen_struct!();
```

with hygiene turned on, the body of the derive can't see any of the uses that come from `__diesel_use_everything!();`

dtolnay: I don't know. :( Serde hasn't hit this because we don't generate macros from macros. File a Rust issue with a minimal repro if you convince yourself it is a bug.

sgrif:
For reference, I think the weird span behavior with `__diesel_use_everything!()` is due to a bug in rustc. https://github.com/rust-lang/rust/issues/47988 is the issue.

PR #1532, opened 2018-02-03
Rewrite `#[derive(SqlType)]` in derives2

sgrif:
The thing that makes this derive tricky is the parsing of options on the
PG side. There's a bunch of different combinations of options, and we
also need an error for if none of the options are present. This makes
the parsing functions tricky to structure, because we want them to
return `None` if the option isn't present, but they must return `Some`
if it was, even if the value isn't something we can work with.

The returning `Some` on bad values is important so we don't display the
"you need to pass one of these three options" error. We also want to
avoid aborting at any point mid-processing, so we display all
errors/warnings. This is most clear in the UI test that does `#[oid =
"NaN", ary_oid = "1"]`, where we need to show an error for `NaN`,
another error for `array_oid` being missing, and a warning for
`ary_oid`, but we need to *not* display the "you need to pass one of
these options" error that we'd give if `oid` wasn't present.

One further change in behavior is that we now parse/validate the
attributes even if the relevant feature is not enabled. This should help
prevent accidental bugs from someone forgetting to test with the right
feature turned on downstream.


sgrif:
For reference, I think the weird span behavior with `__diesel_use_everything!()` here is due to a bug in rustc. https://github.com/rust-lang/rust/issues/47988 is the issue.

PR #1537, opened 2018-02-04
Reimplement SqlLiteral::Bind

### LINKED ISSUE #1347: Bring back `SqlLiteral::bind`
That method was deprecated when `sql_query` was added, and subsequently removed in 1.0. The interface it provided was impossible to use safely on PG or Oracle. While most places that were using it were probably able to migrate to `sql_query`, it's still a useful method to have. However, for it to be safe to embed elsewhere in the query builder, it needs to *also* be responsible for adding the bind SQL. For example, `sql("id > ? AND name != ?").bind::<Integer, _>(1).bind::<Text, _>("Sean")` should become: `sql("id > ").bind::<Integer, _>(1).sql(" AND name != ").bind::<Text, _>("Sean")`

notryanb:
Reimplement `SqlLiteral::bind()`.

Fixes #1347

### DIFF
--- a/diesel/src/expression/sql_literal.rs
+++ b/diesel/src/expression/sql_literal.rs
@@ -7,56 +7,187 @@ use query_dsl::RunQueryDsl;
 use result::QueryResult;
 
 #[derive(Debug, Clone)]
+#[must_use = "Queries are only executed when calling `load`, `get_result`, or similar."]
 /// Returned by the [`sql()`] function.
 ///
 /// [`sql()`]: ../dsl/fn.sql.html
-pub struct SqlLiteral<ST> {
+pub struct SqlLiteral<ST, T = ()> {
     sql: String,
+    inner: T,
     _marker: PhantomData<ST>,
 }
 
-impl<ST> SqlLiteral<ST> {
+impl<ST, T> SqlLiteral<ST, T> {
     #[doc(hidden)]
-    pub fn new(sql: String) -> Self {
+    pub fn new(sql: String, inner: T) -> Self {
         SqlLiteral {
             sql: sql,
+            inner: inner,
             _marker: PhantomData,
         }
     }
+
+    /// Bind a value for use with this SQL query.
+    ///
+    /// # Safety
+    ///
+    /// This function should be used with care, as Diesel cannot validate that
+    /// the value is of the right type nor can it validate that you have passed
+    /// the correct number of parameters.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #    users {
+    /// #        id -> Integer,
+    /// #        name -> VarChar,
+    /// #    }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// #     use diesel::dsl::sql;
+    /// #     use diesel::sql_types::{Integer, Text};
+    /// #     let connection = establish_connection();
+    /// let seans_id = users
+    ///     .select(id)
+    ///     .filter(sql("name = ").bind::<Text, _>("Sean"))
+    ///     .get_result(&connection);
+    /// assert_eq!(Ok(1), seans_id);
+    ///
+    /// let tess_id = sql::<Integer>("SELECT id FROM users WHERE name = ")
+    ///     .bind::<Text, _>("Tess")
+    ///     .get_result(&connection);
+    /// assert_eq!(Ok(2), tess_id);
+    /// # }
+    /// ```
+    ///
+    /// ### Multiple Bind Params
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
+    /// # include!("../doctest_setup.rs");
+    ///
+    /// # table! {
+    /// #    users {
+    /// #        id -> Integer,
+    /// #        name -> VarChar,
+    /// #    }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// #     use diesel::dsl::sql;
+    /// #     use diesel::sql_types::{Integer, Text};
+    /// #     let connection = establish_connection();
+    /// #     diesel::insert_into(users).values(name.eq("Ryan"))
+    /// #           .execute(&connection).unwrap();
+    /// let query = users
+    ///     .select(name)
+    ///     .filter(
+    ///         sql("id > ")
+    ///         .bind::<Integer,_>(1)
+    ///         .sql(" AND name <> ")
+    ///         .bind::<Text, _>("Ryan")
+    ///     )
+    ///     .get_results(&connection);
+    /// let expected = vec!["Tess".to_string()];
+    /// assert_eq!(Ok(expected), query);
+    /// # }
+    /// ```
+    pub fn bind<BindST, U>(self, bind_value: U) -> UncheckedBind<Self, U::Expression>
+    where
+        U: AsExpression<BindST>,
+    {
+        UncheckedBind::new(self, bind_value.as_expression())
+    }
+
+    /// Use literal SQL in the query builder
+    ///
+    /// This function is intended for use when you need a small bit of raw SQL in
+    /// your query. If you want to write the entire query using raw SQL, use
+    /// [`sql_query`](../fn.sql_query.html) instead.
+    ///
+    /// # Safety
+    ///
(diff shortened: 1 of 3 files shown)


diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +174,70 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+
+impl<Query, Value, ST, DB> QueryFragment<DB> for UncheckedBind<Query, Value, ST>
+where
+    DB: Backend + HasSqlType<ST>,
+    Query: QueryFragment<DB>,
+    Value: ToSql<ST, DB>,
+{
+    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
+        self.query.walk_ast(out.reborrow())?;
+        out.push_bind_param(&self.value)?;
```
sgrif: Rather than duplicating bind param handling here, what do you think about structuring this differently so that it just holds the bind parameter?

```
struct UncheckedBind<Query, Value> {
    query: Query,
    value: Value,
}

impl<ST, T> SqlLiteral<ST, T> {
    pub fn bind<BindST, U>(self, value: U) -> UncheckedBind<Self, U::Expression>
    where
        U: AsExpression<BindST>,
```

sgrif:
Can you rebase and run rustfmt so it's easier to see what's new in this PR?

diesel/src/expression/sql_literal.rs
```diff
@@ -5,48 +5,128 @@ use expression::*;
...
+    /// #        name -> VarChar,
+    /// #    }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     use self::users::dsl::*;
+    /// #     use diesel::dsl::sql;
+    /// #     use diesel::sql_types::{Integer, Text};
+    /// #     let connection = establish_connection();
+    /// let seans_id = sql::<Integer>("SELECT id FROM users WHERE name = ")
```
weiznich: Maybe we should use some other example here, because `sql` is meant to implement small missing query fragments not whole queries. So maybe just show something where a custom filter is set?

notryanb: This is a good point. I'll update this test case.

diesel/src/expression/sql_literal.rs
```diff
@@ -10,43 +10,130 @@ use result::QueryResult;
...
+    /// # }
+    /// ```
+    pub fn bind<BindST, U>(self, bind_value: U) -> UncheckedBind<Self, U::Expression>
+    where
+        U: AsExpression<BindST>,
+    {
+        UncheckedBind::new(self, bind_value.as_expression())
+    }
+
+    pub fn sql(self, sql: &str) -> SqlLiteral<ST, Self> {
```
sgrif: I'm surprised that this isn't failing due to `#[deny(missing_docs)]`. It needs docs.

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+
+impl<Query, Value> UncheckedBind<Query, Value>
+where
+    Query: Expression,
+{
+    pub fn new(query: Query, value: Value) -> Self {
+        UncheckedBind { query, value }
+    }
+
+    pub fn sql(self, sql: &str) -> SqlLiteral<Query::SqlType, Self> {
```
sgrif: Needs docs.

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+pub struct UncheckedBind<Query, Value> {
+    query: Query,
+    value: Value,
+}
+
+impl<Query, Value> UncheckedBind<Query, Value>
+where
+    Query: Expression,
+{
+    pub fn new(query: Query, value: Value) -> Self {
```
sgrif: This should be `pub(crate)`

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
 /// # }
 /// ```
 pub fn sql<ST>(sql: &str) -> SqlLiteral<ST> {
-    SqlLiteral::new(sql.into())
+    SqlLiteral::new(sql.into(), ())
+}
+
+#[derive(Debug, Clone, Copy)]
+#[must_use = "Queries are only executed when calling `load`, `get_result`, or similar."]
+pub struct UncheckedBind<Query, Value> {
```
sgrif: Needs docs.

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+}
+
+impl<Query, Value> Expression for UncheckedBind<Query, Value>
+where
+    Query: Expression,
+{
+    type SqlType = Query::SqlType;
+}
+
+impl<Query, Value> QueryId for UncheckedBind<Query, Value>
```
sgrif: This impl is wrong. We should just derive it.

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+impl<Q, Value> Query for UncheckedBind<Q, Value>
+where
+    Q: Query,
+{
+    type SqlType = Q::SqlType;
+}
+
+impl<Query, Value> NonAggregate for UncheckedBind<Query, Value>
+where
+    UncheckedBind<Query, Value>: Expression,
```
sgrif: This should have `Query: NonAggregate` and `Value: NonAggregate` as bounds.

sgrif: Can we prefer `Self` to explicitly stating the full type?

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+{
+}
+
+impl<QS, Query, Value> SelectableExpression<QS> for UncheckedBind<Query, Value>
+where
+    Query: Expression,
```
sgrif: This should be `Self: AppearsOnTable<QS>`.

diesel/src/expression/sql_literal.rs
```diff
@@ -94,5 +181,81 @@ impl<ST> NonAggregate for SqlLiteral<ST> {}
...
+{
+}
+
+impl<QS, Query, Value> AppearsOnTable<QS> for UncheckedBind<Query, Value>
+where
+    Query: Expression,
```
sgrif: This should be `Self: Expression`.

diesel/src/expression/sql_literal.rs
notryanb: Any examples of this elsewhere?

weiznich: Just put a `#[derive(QueryId)]` on the definition of `UncheckedBind`

notryanb: Ack, I totally forgot about this!

diesel/src/expression/sql_literal.rs
notryanb: @sgrif Do we need this method? If we don't plan on supporting`.filter(sql("id > 1").sql(" AND name <> 'Ryan'"))...` I think I can omit this.

diesel/src/expression/sql_literal.rs
notryanb: @sgrif Do we need this method? If we don't plan on supporting.filter(sql("id > 1").sql(" AND name <> 'Ryan'"))... I think I can omit this.

sgrif: We don't *strictly speaking* need it, but I like the symmetry of having both methods on both types. It makes it more clear to me what you can call on `UncheckedBind` by seeing it here. Otherwise I have to click through to the return type to see the methods on the other type. If we just have the same methods on both, those methods (and docs) apply in both places.

notryanb:
@sgrif @weiznich Just checking in. The CI failure looks to be unrelated to the implementation.

Eijebong:
@notryanb Yeah, the failure is the same as in #1595 
I think we need to rustup because cargo_metadata isn't compatible with the nightly we're using anymore

Eijebong:
#1597

sgrif:
Merging with red CI #yolo

PR #1538, opened 2018-02-04
rm `IncompleteUpdateStatement`

sgrif:
This type is redundant and can be replaced with a session type in the
third parameter of `UpdateStatement`. Otherwise I'll end up needing
`IncompleteBoxedUpdateStatement` too. I've left a deprecated type alias
in place for backwards compatibility.


weiznich:
What happens here if someone calls the following code:
```rust
update(some::table).execute(conn)?;
```
It does just not work because `SetNotCalled` does not implement `QueryFragment`, right?
If so we may want to add a compile tests for that?

sgrif:
compile test added

PR #1545, opened 2018-02-06
Fix `SelectStatement` being incorrectly allowed in an expression context

sgrif:
This unfortunately requires a breaking change, but it is in order to fix
a bug. Any code that was relying on this impl would have generated an
invalid query, or be subject to other bugs. Additionally, all types
affected by this are internal API. Finally, the SQL types are no longer
changed by enabling the "postgres" feature, and we can start relying on
`SelectStatement` always doing what we want (there are some hacks
working around that which I have not removed in this commit).

In order to continue to be able to at least somewhat abstract over how
to get the SQL type of a select clause, I've introduced `SelectQuery`
into the public API. This trait is basically `Expression`, but only
implemented for select statements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * `#[derive(Identifiable)]` now correctly associates `#[primary_key]` with the
   column name, not field name.
 
+* Select statements can no longer incorrectly appear in an expression context.
+
+* `exists` can no longer incorrectly receive values other than select
+  statements.
+
 ### Jokes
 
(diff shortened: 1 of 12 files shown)


weiznich:
This commit introduced a bug into what queries are allowed as subqueries for `eq_any`
After this change the following code compiles, but generates invalid sql:
```rust
table!{
    users{
       id -> Integer,
       name -> Text,
    }
}

table! {
    posts {
        id -> Integer,
        user_id -> Integer,
    }
}

fn test() {
    let subquery = users::table.filter(users::id.eq(1));
    let query = posts::table.filter(posts::user_id.eq_any(subquery));
}
```
This does only happen/work for normal select statements, not for boxed select statements.

PR #1547, opened 2018-02-07
Allow to pass boxed subselect queries as eq_any on nullable fields

weiznich:
This matches the behaviour of normal subselect queries


sgrif:
> This matches the behaviour of normal subselect queries

Where? I don't see that impl

weiznich:
> > This matches the behaviour of normal subselect queries
>
> Where? I don't see that impl

It just works. It may also be related to the bug written below #1545 .

(Adding the same impl for `SelectStatement` generates a conflicting implementation error with the existing impl for `SelectStatement`)

PR #1548, opened 2018-02-08
Clarify use of `execute` versus `load` in documentation

### LINKED ISSUE #1136: error message could explain `execute` vs `load`
### Versions

- **Rust:** nightly 1.21 2017-08-29
- **Diesel:** 0.14
- **Database:** postgresql 9.5
- **Operating System** ubuntu 16.04

### Feature Flags

- **diesel:** postgres, uuid, serde_json, chrono
- **diesel_codegen:** postgres

## Problem Description

It took me a long time to figure out that `execute` returns at most one value (I'm assuming) while `load` returns (or can return) a collection of values. 

I got the following error message: 

```
  --> src/main.rs:95:9
   |
95 | /         sql::<(Integer, pg_types::Uuid,)>(query)
96 | |             .execute(&db)
97 | |             .expect("query failed");
   | |_______________________________________^ expected struct `std::vec::Vec`, found usize
   |
   = note: expected type `std::vec::Vec<(i32, uuid::Uuid)>`
              found type `usize`
```
(You can't see it in the error message, but I was trying to assign the query result to a `Vec<(i32, uuid::Uuid)>`).

It would be helpful if this error message could offer some hint about the difference between `execute` vs `load`. Just thought I would share a point of confusion from someone trying to learn the library. Thanks!

## Checklist

- [ yes ] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

brandur:
Adds cross-references to the documentation of the `execute` and `load`
functions that help clarify when it's appropriate to use each one, and
suggests the use of the other when it's not.

Fixes #1136.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -875,8 +875,12 @@ impl<T: Table> QueryDsl for T {}
 pub trait RunQueryDsl<Conn>: Sized {
     /// Executes the given command, returning the number of rows affected.
     ///
-    /// Used in conjunction with [`insert_into`](../fn.insert_into.html),
-    /// [`update`](../fn.update.html) and [`delete`](../fn.delete.html)
+    /// `execute` is usually used in conjunction with [`insert_into`](../fn.insert_into.html),
+    /// [`update`](../fn.update.html) and [`delete`](../fn.delete.html) where the number of
+    /// affected rows is often enough information.
+    ///
+    /// When asking the database to return data from a query, [`load`](fn.load.html) should
+    /// probably be used instead.
     ///
     /// # Example
     ///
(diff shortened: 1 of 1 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -875,8 +875,10 @@ impl<T: Table> QueryDsl for T {}
 pub trait RunQueryDsl<Conn>: Sized {
     /// Executes the given command, returning the number of rows affected.
     ///
-    /// Used in conjunction with [`insert_into`](../fn.insert_into.html),
-    /// [`update`](../fn.update.html) and [`delete`](../fn.delete.html)
+    /// `execute` is usually used in conjunction with [`insert_into`](../fn.insert_into.html),
+    /// [`update`](../fn.update.html) and [`delete`](../fn.delete.html) where the number of
+    /// affected rows is often enough information. When asking the database to return information
```
retep007: Could you move `When asking...` on a new line?

Also what you say on `When asking the database to return data from query`.

- Information is also a number of the affected rows
- Operation feels to me a bit too much agnostic

brandur: Hi @retep007, thanks for the review!

Can you take another look? I integrated your feedback here, and also tightened up the phrasing to be a little more prescriptive: "`load` should probably be used instead" instead of "`load` might be useful instead".

Issue #1553, opened 2018-02-12
Add support for `skip_locked` 


Diggsey:
This is a must-have feature for implementing queues on top of postgres. External work queues are not suitable when you need full transactional semantics and have strong ordering and delivery guarantees to uphold.

sgrif:
Fixed by #1554

PR #1554, opened 2018-02-12
Implement support for "FOR UPDATE" modifiers

Diggsey:
Quick attempt at #1553


diesel/src/query_dsl/locking_dsl.rs
```diff
@@ -30,3 +30,63 @@ where
         self.as_query().for_update()
     }
 }
+
+/// The `skip_locked` method
+///
+/// This trait should not be relied on directly by most apps. Its behavior is
+/// provided by [`QueryDsl`]. However, you may need a where clause on this trait
+/// to call `for_update` from generic code.
```
weiznich: There is a copy-past error in here. This trait is required for `skip_locked` not for `for_update`.

Diggsey:
I added some tests which verify that the "no wait" and "skip locked" behaviours work as intended and added the compile fail tests.

I also added a new database error kind for use with `.no_wait()`, which makes it possible to detect whether a query failed for this reason.

diesel/src/query_builder/for_update_clause.rs
```diff
@@ -12,4 +12,15 @@ impl<DB: Backend> QueryFragment<DB> for NoForUpdateClause {
 }
 
 #[derive(Debug, Clone, Copy, QueryId)]
-pub struct ForUpdateClause;
+pub struct ForUpdateClause<Modifier> {
```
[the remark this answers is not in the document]
sgrif: It doesn't need to for backwards compatibility, but I think we should default it anyway. Less churn.

diesel/src/pg/query_builder/query_fragment_impls.rs
```diff
@@ -1,11 +1,26 @@
...
 
-impl QueryFragment<Pg> for ForUpdateClause {
+impl QueryFragment<Pg> for ForUpdateClause<NoModifier> {
     fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
         out.push_sql(" FOR UPDATE");
         Ok(())
     }
 }
+
+impl QueryFragment<Pg> for ForUpdateClause<SkipLockedModifier> {
```
sgrif: I'd rather we implement `QueryFragment` for the modifiers rather than duplicate this impl 3 times.

diesel/src/query_builder/for_update_clause.rs
```diff
@@ -12,4 +12,15 @@ impl<DB: Backend> QueryFragment<DB> for NoForUpdateClause {
...
+    pub(crate) modifier: Modifier,
+}
+
+#[derive(Debug, Clone, Copy, QueryId)]
+pub struct NoModifier;
+
+#[derive(Debug, Clone, Copy, QueryId)]
+pub struct SkipLockedModifier;
```
sgrif: Do we need `Modifier` at the end of every type? I think just `SkipLocked` is fine.

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -302,7 +311,53 @@ impl<F, S, W, O, L, Of> ForUpdateDsl for SelectStatement<F, S, NoDistinctClause,
             self.limit,
             self.offset,
             self.group_by,
-            ForUpdateClause,
+            ForUpdateClause {
```
sgrif: Can we move this to `ForUpdateClause::new`?

diesel_compile_tests/tests/compile-fail/select_for_update_skip_locked_cannot_be_used_on_mysql.rs
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+
+fn main() {
+    let conn = MysqlConnection::establish("").unwrap();
+    users::table
+        .for_update()
+        .skip_locked()
+        .load(&conn)
+        //~^ ERROR: E0271
```
sgrif: MySQL 8 does support these modifiers, so we should allow them on that backend.

diesel/src/result.rs
```diff
@@ -94,6 +94,9 @@ pub enum DatabaseErrorKind {
     /// An example of a case where this would occur is if you attempted to send
     /// a query with more than 65000 bind parameters using PostgreSQL.
     UnableToSendCommand,
+    /// A lock could not be acquired, perhaps because the query was executed with
+    /// `FOR UPDATE NOWAIT` and the lock was already held.
+    LockNotAvailable,
```
sgrif: We've been very conservative about adding variants to this in the past. Can you provide some example use cases that need this variant?

diesel/src/pg/query_builder/query_fragment_impls.rs
Diggsey: Will do

diesel/src/query_builder/for_update_clause.rs
Diggsey: OK

diesel/src/query_builder/select_statement/dsl_impls.rs
Diggsey: Yep

diesel_compile_tests/tests/compile-fail/select_for_update_skip_locked_cannot_be_used_on_mysql.rs
Diggsey: How do you handle that right now? Should I just enable it outright for the MySQL back-end, or is there a way to specify the minimum supported MySQL version at compile time?

sgrif: I'd just enable it outright for the backend, and put a note in the docs about the supported versions.

sgrif: We should mention that `SKIP LOCKED` is only on PG 9.5+ as well

diesel/src/result.rs
Diggsey: When using `FOR UPDATE NOWAIT` it's important to know whether the query failed due to the lock having been taken, or for some other reason. See the tests I added for a concrete example. I don't know how you'd use `NOWAIT` without that ability.

sgrif: It's worth noting that this is the only way to create a select statement that can fail at runtime in Diesel (other than network errors)

Diggsey: That's not true, there is at least one other way a select could fail: timing out. In fact I use this in the test by setting a 1 second timeout to determine if `NOWAIT` is working correctly. (I deliberately set up a deadlock, and then if it times out instead of returning `LockNotAvailable` then it didn't work correctly).

I imagine there are also various encoding-related things that could go wrong in a select statement.

Diggsey: Also if the schema as defined in rust doesn't match the database schema, that could also cause issues, or if permissions are not correct. There are many ways select could fail.

sgrif: I'm inclined to accept the variant, but I'd prefer that it be kept as a separate PR so we can keep the changes atomic, and to give that error a place to be discussed in isolation.

Diggsey: What should I do about the test? I can't test `NOWAIT` without that error.

sgrif: Just test `is_err` for now

Diggsey: OK, but that will always pass 😦

sgrif: Oh I see what you mean, it'd time out if it didn't fail immediately. Perhaps check that the error message contained an expected string?

diesel/src/pg/connection/result.rs
```diff
@@ -39,8 +39,7 @@ impl PgResult {
                         _ => DatabaseErrorKind::__Unknown,
                     };
                 let error_information = Box::new(PgErrorInformation(internal_result));
-                Err(Error::DatabaseError(error_kind, error_information))
-            }
+                Err(Error::DatabaseError(error_kind, error_information))            }
```
sgrif: I think you need to run rustfmt. ;)

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -289,8 +289,17 @@ where
...
+    type Output = SelectStatement<
+        F,
+        S,
+        NoDistinctClause,
+        W,
+        O,
+        L,
+        Of,
+        NoGroupByClause,
+        ForUpdateClause<NoModifier>,
```
sgrif: I don't think this line needed to change.

diesel/src/query_dsl/locking_dsl.rs
```diff
@@ -30,3 +30,63 @@ where
...
+    /// convenient access to this type.
+    ///
+    /// [`dsl::SkipLocked`]: ../../dsl/type.SkipLocked.html
+    type Output;
+
+    /// See the trait level documentation
+    fn skip_locked(self) -> Self::Output;
+}
+
+impl<T> SkipLockedDsl for T
```
sgrif: We can drop this impl. The only types which implement this already have a `FOR UPDATE` clause, which tables do not.

diesel/src/query_dsl/locking_dsl.rs
```diff
@@ -30,3 +30,63 @@ where
...
+    /// convenient access to this type.
+    ///
+    /// [`dsl::NoWait`]: ../../dsl/type.NoWait.html
+    type Output;
+
+    /// See the trait level documentation
+    fn no_wait(self) -> Self::Output;
+}
+
+impl<T> NoWaitDsl for T
```
sgrif: We can drop this impl, too

sgrif:
This is looking great, but the `Table` impls made me realize something -- You haven't implemented this for boxed select statements, and there is no way to safely do so (we can't know if a boxed select statement already has a `FOR UPDATE` clause or not). I think rather than `.for_update().skip_locked()`, we should go with the API `.for_update_skip_locked()`, which overrides the previous `FOR UPDATE` clause if there was one. This way we can reasonably implement it for boxed select statements.

Diggsey:
> we should go with the API .for_update_skip_locked(), which overrides the previous FOR UPDATE clause if there was one. 

That's not going to extend well with other modifiers - `FOR UPDATE` is a complex clause and `NOWAIT`/`SKIP LOCKED` are only one of the ways that its behaviour can be modified.

sgrif:
Ughhhhhhhh.... 
screen shot 2018-02-26 at 5 25 08 pm

sgrif:
I need to spend some time thinking about this.

sgrif:
SQL vendors please stop adding features which are extremely complex and ever so slightly different in an incompatible way with nearly identical feature from other SQL vendor kthx

Diggsey:
@sgrif tests are passing, good to go?

sgrif:
It's the weekend. I'll take a look on Monday. :)

Diggsey:
Bump 😛

sgrif:
I never said *which* Monday :trollface: #perl6 (Sorry I've been a little swamped lately I have pinned this to my browser, and I super promise to review this week. I was about to do it right now but literally as I was writing this my wife said it's time to go)

sgrif:
Let's go with the API you've proposed. My main concern with it is that it allows:

```
foo::table
    .for_update()
    .filter(stuff)
    .skip_locked()
```

which is nonsense, but I think that's OK. This still has unresolved feedback to be addressed though

sgrif:
Note: We *can* work around this by having the return type of `.for_update` be something like `LockClauseBuilder`, and making everything else be inherent methods on that type -- But unfortunately the way things are structured today we wouldn't be able to get away with just implementing `AsQuery`. We'd also need to implement every DSL trait since the blanket trait on tables won't apply there. It'd also require making the concrete type of `<SelectStatement as ForUpdateDsl>::Output` part of the public API which I don't want to do.

Diggsey:
@sgrif we recently started actually using this at work (on our fork of diesel) and discovered we also needed `for_no_key_update`. I'm starting to wonder if the cases I considered rare enough to not worry about are actually quite common.

What about an API like this:

```rust
foo::table
    .locking(foo::table.for_update())
```

This better matches how the query is parsed, and allows us to handle the more complicated cases:
```rust
<some query involving foo and bar>
    .locking(foo::for_no_key_update().skip_locked())
    .locking(bar::for_share())
```

Mentioning the same table multiple times is allowed, so we don't have to prevent that. What do you think?

We would probably need to add global methods, so you could do:
```rust
<some query involving foo and bar>
    .locking(for_no_key_update().skip_locked())
```

And it would affect both mentioned table, as this is the only form allowed in mysql, although personally I like being explicit about which tables are locked (and we actually had a production issue caused by a joined table being unexpectedly locked...)

Diggsey:
We would of course need to keep around `for_update()` for backwards compatibility but it can easily be implemented on top of the above model.

sgrif:
I'm not sure I see the benefit of having a separate `locking` method vs just `.for_no_key_update` and `.for_share` on the table itself

Diggsey:
@sgrif I'm not quite sure how that would look? As it stands, the API does not allow you to specify which tables are locked in a multi-table query, or to specify different locking behaviours for different tables.

sgrif:
It would look like the API you've proposed.

```
.for_update().skip_locked()
.for_no_key_update().skip_locked()
```

WRT specifying the table, I suspect we'd add something along the lines of `.for_update_of(table)`. Granted, this is a large amount of methods -- Your API is somewhat along the lines of something I considered when I was thinking about this (but I was thinking something more along the lines of `.locking().for_update().skip_locked.for_table(table)`, which has all sorts of problems). Ultimately the only thing gained by going with a separate builder style API in any form is that we would need one method to specify the table instead of 4, but I'm not sure that's a big enough win?

Diggsey:
It has the advantage of not cluttering the top-level DSL. I also think having a bit of structure is preferable to just a flat namespace - do you disagree? The closer the DSL is to the way the query is parsed, the simpler and less magical it feels to use.

It's obvious from looking at the nested form *what* the `skip_locked()` modifier actually applies to, even if you are not already familiar with it: nested better matches the reality of how the query is executed, which is why it's a bit of a pain to pull those methods up onto the root QueryDsl.

sgrif:
> It has the advantage of not cluttering the top-level DSL.

This really depends on what you consider the "top level" DSL. Your proposal clutters a different top level, which is much harder to document. Ultimately this is a documentation problem more than anything else. We could certainly choose *not* to document the methods from `LockingDsl` on `QueryDsl` if we choose to keep them separate.

> I also think having a bit of structure is preferable to just a flat namespace - do you disagree?

I do not disagree, I just haven't seen an API that I think is preferable.

> It's obvious from looking at the nested form what the skip_locked() modifier actually applies to, even if you are not already familiar with it: nested better matches the reality of how the query is executed, which is why it's a bit of a pain to pull those methods up onto the root QueryDsl.

Sure, this is the concern I raised in my original comment today. I think it's not an actual concern in practice however, since people are just going to write `.skip_locked` immediately after the line it applies to.

Diggsey:
Let's say we implement these as flat methods, we run into the problem you brought up earlier about implementing a `LockClauseBuilder`.

Implementing all the DSL traits for each builder doesn't really scale very well... The nested version of the API solves this because the "closing" of the locking clause is explicit. 

Documentation wise, the "locking(...)" method is the perfect place to document what all the different modifiers are that can be used.

sgrif:
@diesel-rs/core Need some more opinions here

Diggsey:
@sgrif let's forget about the "per table" locking modes for now? I've updated the PR to include the other basic locking modes (`FOR SHARE`/`FOR KEY SHARE`) which I ended up needing.

sgrif:
Brief look over looks fine. I don't like the amount of churn here. I'll try to make some time to give it a more thorough review in the morning.

Diggsey:
Hm, bit confused about the compile-fail tests - it seems to be complaining that `E0277` is unexpected, but that's exactly the error that's annotated in the test...

sgrif:
@Diggsey it is time

Diggsey:
@sgrif it is

PR #1557, opened 2018-02-15
Allow to construct a column for a generic parameter

weiznich:


sgrif:
Adding a non-defaulted method to a trait in the public API is a breaking change.

PR #1558, opened 2018-02-15
Allow to select a column as nullable

weiznich:
This makes the following code compile (and work):
```rust
users::table.select(users::name.nullable());
```

This is for example useful to write subselect statements on nullable columns.
(If this is accepted #1547 is not needed anymore.)


sgrif:
I'd like to hold off on this for now. If the motivating use case is `nullable_column.eq_any(non_nullable_subselect)`, I think there's a better solution (possibly #1547, or possibly just adding a `.nullable` method on `SelectStatement`, need to look at it more closely). I think this is probably the least ergonomic solution though.

PR #1559, opened 2018-02-15
Implement `Default` for `Column`

weiznich:
This allows to construct columns from generic parameters

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -6,7 +6,7 @@ macro_rules! __diesel_column {
             #[doc=$doc]
         )*
         #[allow(non_camel_case_types, dead_code)]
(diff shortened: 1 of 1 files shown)


diesel/src/query_source/mod.rs
```diff
@@ -45,7 +45,7 @@ pub trait QuerySource {
 
 /// A column on a database table. Types which implement this trait should have
 /// been generated by the [`table!` macro](../macro.table.html).
-pub trait Column: Expression {
+pub trait Column: Expression + Default {
```
sgrif: Adding this bound is a breaking change.

weiznich: I've removed this bound.

sgrif:
This needs a changelog entry

PR #1561, opened 2018-02-17
Add Support for Unsigned Types in Mysql

joshleeb:
Ref. Handling unsigned integer types from MySQL #1181

This functionality was originally added in #782 and then reverted in #912.

PR #1566, opened 2018-02-20
add support for the Point type on PG.

YetAnotherMinion:
I am using the built in geometry types in Postgres to implement a collaborative drawing app. I was able to create a custom type for my own use, but I would be happy to do the work to upstream geometric types into diesel.

I looked at the existing implementations of postgres types and they were different enough to leave me with some questions:

- Should I use a wrapper type such as `PgPoint`, just like `PgMoney` for `Money` does, or should I just use a raw tuple of `(f64, f64)` like `Range` does?
- I see all of the wrapper types have the `#[sql_type = ""]` attribute, but when I add it to the PgPoint I get a compiler error about unknown attribute. I don't understand why. Even when I copied and pasted the `PgMoney` implementation and changed only the name of `PgMoney` to `PgPoint`, and `sql_type` to `Point`, I still got the same compiler error.
- The doc test for `Point` does not compile, however the doc tests for Inet,Cidr, Jsonb, and MacAddr also do not compile on my system, all with the same error that diesel::Expression is not implemented. Can you point me towards the right place in the documentation so I can understand what is going on.

Probably unrelated, when I attempt to use this branch directly from local path with a [single file example](https://gist.github.com/YetAnotherMinion/00a2a17880b485677ba0530df3214bfe), I get a compiler error about impl_Identifiable! not found when deriving Identifiable. The same error is present in the diesel master I used c12bea51f75fa51b482a51386dffe54c070a5647. Is this a known bug from code changes in between releases? I did not see anything the changelog about identifiable going away?


diesel/src/pg/types/geometric.rs
```diff
@@ -0,0 +1,62 @@
...
+use expression::AsExpression;
+use expression::bound::Bound;
+use pg::Pg;
+use serialize::{self, Output, ToSql, IsNull};
+use sql_types::{Point};
+
+/// Point is represented in Postgres as a tuple of 64 bit floating point values (x, y).  This
+/// struct is a dumb wrapper type, meant only to indicate the tuple's meaning.
+#[derive(Debug, Clone, PartialEq, Copy, FromSqlRow)]
+pub struct PgPoint(pub f64, pub f64);
```
YetAnotherMinion: Should this be just a raw tuple i.e. `(f64, f64)`?

diesel/src/pg/types/geometric.rs
```diff
@@ -0,0 +1,62 @@
...
+    type Expression = Bound<Point, &'a PgPoint>;
+
+    fn as_expression(self) -> Self::Expression {
+        Bound::new(self)
+    }
+}
+
+impl FromSql<Point, Pg> for PgPoint {
+    fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
+        let mut bytes = not_none!(bytes);
```
YetAnotherMinion: Do I need a debug_assert! counting the number of bytes I expect (16bytes for point)? Some types have it, some don't.

YetAnotherMinion:
Looking at #1561 which also adds a type to diesel, do I need to implement `SelectableExpression`, `AppearsOnTable`, `Expression`?

diesel/src/pg/types/geometric.rs
sgrif: Nope.

diesel/src/pg/types/geometric.rs
sgrif: No, we include them on integers because it's common to accidentally end up with the wrong type. That's less of a concer here.

diesel/src/pg/types/geometric.rs
```diff
@@ -0,0 +1,64 @@
...
+        let x = bytes
+            .read_f64::<NetworkEndian>()
+            .map_err(|e| Box::new(e) as Box<Error + Send + Sync>)?;
```
sgrif: This `map_err` uneccessary.

diesel/src/pg/types/geometric.rs
```diff
@@ -0,0 +1,64 @@
...
+        let y = bytes
+            .read_f64::<NetworkEndian>()
+            .map_err(|e| Box::new(e) as Box<Error + Send + Sync>)?;
```
sgrif: This `map_err` is unnecessary.

diesel/src/pg/types/mod.rs
```diff
@@ -446,4 +447,55 @@ pub mod sql_types {
...
+    ///
+    /// ### [`ToSql`] impls
+    ///
+    /// - [`PgPoint`]
+    ///
+    /// ### [`FromSql`] impls
+    ///
+    /// - [`PgPoint`]
+    ///
+    /// [`ToSql`]: ../../../serialize/trait.ToSql.html
```
sgrif: Can you use the new intra-module link form that rustdoc provides?

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * Queries that treat a subselect as a single value (e.g. `foo = (subselect)`)
   are now supported by calling [`.single_value()`].
 
+* Added support for the Point type on PG.
```
sgrif: Link to the docs?

YetAnotherMinion:
- Added link to Postgresql docs for Point datatype in changelog.
- Used intra module documentation links
- Removed unnecessary box errors
- Rustfmtted

The doc test example still fails. Should I choose a different example (like using the point as a member on a struct instead of querying on the value of the point) or does query fragment need to be implemented so a point's value can be used in a where clause?

sgrif:
Your doctest example is fine, but you're missing `#[derive(AsExpression)]`. I'd also like to see some tests (not doctests) showing that this can be used in Queryable and Insertable contexts.

YetAnotherMinion:
I looked into adding the Postgres geometric operators to diesel in `src/pg/expression/operators.rs` and `src/pg/expression/expression_methods.rs`. The design constraint I ran into there is that some geometric operators have restrictions on operand types for both left and right hand sides.  For example, the **is horizontally aligned** operator `?-` is implemented by PG for `point -> point -> bool`, but not for `polygon -> polygon -> bool`.

I want to verify my understanding of the design of expression methods with you:
- The operator structs are in the public API where anyone can shove any type into them?
- The SQL generated from raw use of operator structs may not be supported by the backend?
- Traits are used to provide builder pattern dsl methods for valid combinations only?
- If you use the `.method` style dsl then the generated SQL will be valid for the backend, i.e. no operator errors?

I am leary to blindly create a trait like `PgGeometricExpressionMethods` that happens to work for all `point -> point -> result` operators, but then puts the next person who wants to add another geometric type such as circle into a bind because the trait bounds on my naive `PgGeometricExpressionMethods` were too loose.

A naive forward compatible design would be to just have a separate trait for each geometric operator that is generic over right and left hand side types, with impls for each possible type signature. I would  use a script to brute force find valid type signatures for each operator.

All other approaches I could think of were more complicated.

sgrif:
> I looked into adding the Postgres geometric operators to diesel

I really don't know that those belong in Diesel. I think they probably should be in a separate crate. TBH the only reason I didn't close this PR immediately is because these types are in ever PG installation, not an extension. Even with that, I'm still not sure that these types belong in Diesel itself vs a more full featured crate.

> The operator structs are in the public API where anyone can shove any type into them?

It depends on the use case, but generally yes, we enforce type safety at the point of construction.

> The SQL generated from raw use of operator structs may not be supported by the backend?

No. Anything that generates backend specific SQL should reflect that in it's `QueryFragment` impl.

> Traits are used to provide builder pattern dsl methods for valid combinations only?

I don't know what this means.

> If you use the .method style dsl then the generated SQL will be valid for the backend, i.e. no operator errors?

We're talking about backend specific features. It'll fail to compile on the wrong backend.

> I am leary to blindly create a trait like PgGeometricExpressionMethods that happens to work for all point -> point -> result operators, but then puts the next person who wants to add another geometric type such as circle into a bind because the trait bounds on my naive PgGeometricExpressionMethods were too loose.

I agree. You shouldn't do that.

> A naive forward compatible design would be to just have a separate trait for each geometric operator that is generic over right and left hand side types, with impls for each possible type signature. I would use a script to brute force find valid type signatures for each operator.

I wouldn't make it generic at all. I'd introduce one trait per type that you care about, as we've done for virtually every other type specific operator.

YetAnotherMinion:
Having another crate for geometric operators is fine with us (what we are doing already). Do you think that crate should be owned by `diesel-rs` just like `diesel_full_text_search` or operated by a random third party?

The only reason I thought of upstreaming into diesel is like you said:
>  because these types are in ever PG installation, not an extension.

I brought up adding the operators because I noticed when writing tests that the `Eq` operator is not valid for point type in Postgres, but diesel lets it compile anyway (fails at runtime with `DatabaseError`). The runtime error when using `schema::dsl::column_name.eq(PgPoint(3,4))` is what I was referring to with:
> If you use the .method style dsl then the generated SQL will be valid for the backend, i.e. no operator errors?

Please let me know if there are any more changes, more tests, etc.. that you would like me to make.

sgrif:
I'll need to think about it more and discuss with the rest of the core
team. :)

vjousse:
I'm also very interested in using the geometric types with diesel. @YetAnotherMinion would it be possible for you to publish your crate while we're waiting for a decision from the core team?

sgrif:
@diesel-rs/core We never had much of a discussion about this PR, and I think we should -- and that we should do so in public.

I have serious reservations around supporting geometric types in Diesel. Our policy has been that extensions live in third party crates, and while this isn't an extension to PG it certainly has the weight of one. My main reservation here is that any type we deserialize to should be richer than what we provide, e.g. `chrono::NaiveDateTime` vs `PgTimestamp`.

Even if we were to provide a "primitive" type with no functionality like we have for others, we get into the question of whether we should support all the various operators for these types in Diesel or not. Again, I seriously hesitate to do that in Diesel. It really feels closer to `diesel_full_text_search` to me in spirit than something that should live in Diesel itself. I'm also concerned with our ability to maintain those operators, since I don't think anybody on the team is using PG's geometry types directly.

YetAnotherMinion:
@vjousse https://github.com/ThinkAlexandria/diesel_geometry

vjousse:
@YetAnotherMinion thanks a lot!

killercup:
> Our policy has been that extensions live in third party crates

Yes.

> we get into the question of whether we should support all the various operators for these types in Diesel

This leads to ask another question: In which context are you actually using this? Does it make sense on its own? Without operators, and helper methods around it? PgMoney is pretty simple and Pg itself doesn't give you much more than numeric ops on it. SO: Is there a small subset of Point that diesel can include without opening the door to having to add _all_ the features?

Alternatively: Do you usually use these types in combination with other crates? If so, should we try and get [`geo::Point`](https://docs.rs/geo/0.7.4/geo/struct.Point.html) and diesel to work together?

- - -

My current feeling is that geometric types are something that (a) can live outside of diesel and (b) is easier to iterate on and add features to if it's a 3rd-party-crate.

weiznich:
Given that it is easily possible to provide implementation for those types outside of diesel I also feel that this live in a 3rd-party-crate. If we see a big demand in supporting this type we could include the support at any future point of time easily from the third party crate. 
The only downside I see in this approach is that things like `infer-schema`/`print-schema` won't work out of the box for this custom type. (But this is a separate issue, for which we could provide a general solution)

Eijebong:
:+1: It should be pretty easy to provide support for those types in a 3rd-party crate.

I see another advantage to that, this way we don't depend on `geo` which means that breaking changes in that crate don't mean breaking change in diesel (which would force us to go 2.0 for example).

For the `infer_schema`/`print-schema`, this is another argument for a config file IMO.

sgrif:
Sounds like there's consensus around this living as a third party crate. I've been thinking a lot about the config file lately, I'll open an issue.

PR #1567, opened 2018-02-20
Derive insertable also on the type, not only on the reference

### LINKED ISSUE #1501: Derive Insertable for struct / Arc / Rc rather then for plain reference
Currently there's only implementation for `&'a struct` when `Insertable` is auto-derived https://github.com/diesel-rs/diesel/blob/master/diesel/src/macros/insertable.rs#L103

This is kind of painful when working with multithreading, because multithreading doesn't understand lifetimes. E.g. I have this method:

```rust
    pub fn create(&self, payload: NewUser) -> Box<Future<Item=User, Error=Error> {
        let conn = self.get_connection();

        Box::new(
            self.cpu_pool.spawn_fn(move || {
                let query = diesel::insert_into(users).values(&payload);
                query.get_result(&*conn).map_err(|e| Error::from(e))
            })
        )
    }
```

I want to generalize all the async stuff into one method to reduce boilerplate. This function works fine for `'static` queries:

```rust
    fn execute_query<T: Send + 'static, U: LoadQuery<PgConnection, T> + Send + 'static>(&self, query: U) -> RepoFuture<T> {
        let conn = match self.r2d2_pool.get() {
            Ok(connection) => connection,
            Err(_) => return Box::new(future::err(Error::Connection("Cannot connect to users db".to_string())))
        };

        Box::new(
            self.cpu_pool.spawn_fn(move || {
                query.get_result::<T>(&*conn).map_err(|e| Error::from(e))
            })
        )
    }
```

But this thing `let query = diesel::insert_into(users).values(&payload);` isn't static, but has `'a` lifetime that is equal to payload. To avoid this, I tried to implement smth like this:

```rust
    fn execute_query_fn<U, Q, F, P>(&self, payload: P, query_fn: F) -> RepoFuture<U>
        where
            U: Send + 'static,
            Q: LoadQuery<PgConnection, U> ,
            F: for<'a> Fn(&'a P) -> Q<'a> + Send + 'static,
            P: Send + 'static
    {
        let conn = match self.r2d2_pool.get() {
            Ok(connection) => connection,
            Err(_) => return Box::new(future::err(Error::Connection("Cannot connect to users db".to_string())))
        };

        Box::new(
            self.cpu_pool.spawn_fn(move || {
                let pl = payload;
                let query = query_fn(&pl);
                query.get_result::<U>(&*conn).map_err(|e| Error::from(e))
            })
        )
    }
```

But this fails due to rust compile limitations `F: for<'a> Fn(&'a P) -> Q<'a> + Send + 'static`, i.e. it doesn't allow to specify lifetime for generic parameter. The corresponding Rust RFC is in progress https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md. 

All this pain comes from the fact that I have to use reference to payload here `let query = diesel::insert_into(users).values(&payload);`. Because only reference auto-derived for Insertable. If it was smth like `struct` or `Arc` everything would work ok.

Does it make sense to implement these?

weiznich:
(Partially) Fixes #1501

### DIFF
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -21,15 +21,31 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     let table_name = model.table_name();
     let struct_name = item.ident;
 
-    let (impl_generics, ty_generics, where_clause) = item.generics.split_for_impl();
-    let mut ref_generics = item.generics.clone();
-    ref_generics.params.push(parse_quote!('insert));
-    let (ref_generics, ..) = ref_generics.split_for_impl();
+    let (_, ty_generics, where_clause) = item.generics.split_for_impl();
+    let mut impl_generics = item.generics.clone();
+    impl_generics.params.push(parse_quote!('insert));
+    let (impl_generics, ..) = impl_generics.split_for_impl();
 
-    let (field_ty, field_assign): (Vec<_>, Vec<_>) = model
+    let (direct_field_ty, direct_field_assign): (Vec<_>, Vec<_>) = model
         .fields()
         .iter()
-        .map(|f| (field_ty(f, table_name), field_expr(f, table_name)))
+        .map(|f| {
+            (
+                (field_ty(f, table_name, None)),
+                (field_expr(f, table_name, None)),
+            )
+        })
+        .unzip();
+
+    let (ref_field_ty, ref_field_assign): (Vec<_>, Vec<_>) = model
+        .fields()
+        .iter()
+        .map(|f| {
+            (
+                (field_ty(f, table_name, Some(quote!(&'insert)))),
+                (field_expr(f, table_name, Some(quote!(&)))),
+            )
+        })
         .unzip();
 
     Ok(wrap_in_dummy_mod(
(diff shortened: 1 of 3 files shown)


diesel_derives/src/insertable.rs
```diff
@@ -59,7 +85,23 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     ))
 }
 
-fn field_ty(field: &Field, table_name: syn::Ident) -> syn::Type {
+fn direct_field_ty(field: &Field, table_name: syn::Ident) -> syn::Type {
+    if field.has_flag("embed") {
+        let field_ty = &field.ty;
+        parse_quote!(#field_ty)
```
sgrif: This can just be `field.ty.clone()`, can't it?

weiznich: Why? This line is required because of the `parse_quote!` below. We can change it to `field.ty.clone()` but I do not see the point to do this.

weiznich:
> This has more duplication than I'd like, but this seems fine.

An alternative to prevent some duplication would be to unify `field_ty` and `field_expr`. I decided against this because both functions consists more or less only of parts (all `quote_parse!` calls) that needs to be handled separately.

diesel_derives/src/insertable.rs
```diff
@@ -59,33 +85,41 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     ))
 }
 
-fn field_ty(field: &Field, table_name: syn::Ident) -> syn::Type {
+fn field_ty(field: &Field, table_name: syn::Ident, life_time: quote::Tokens) -> syn::Type {
```
sgrif: `lifetime`.

PR #1568, opened 2018-02-21
Ensure that IPv6 MySQL hosts are not wrapped in brackets

kulshrax:
Previously, attempting to connect to a MySQL server via an IPv6 address (e.g., `mysql://user@[::1]/foo`) would result in an error (e.g., `BadConnection("Unknown MySQL server host \'[::1]\' (0)")`) because `ConnectionOptions::host()` would return the host wrapped in brackets due to [the underlying behavior of `Url::host_str()`](https://docs.rs/url/1.7.0/url/enum.Host.html#variant.Ipv6). Unfortunately, `mysql_real_connect()` in `libmysqlclient` appears to only allow IPv6 addresses to be passed without brackets, so this situation needs to be special-cased in order to make this work.

### DIFF
--- a/diesel/src/mysql/connection/url.rs
+++ b/diesel/src/mysql/connection/url.rs
@@ -1,7 +1,7 @@
 extern crate url;
 
 use std::ffi::{CStr, CString};
-use self::url::Url;
+use self::url::{Host, Url};
 use self::url::percent_encoding::percent_decode;
 
 use result::{ConnectionError, ConnectionResult};
@@ -29,8 +29,9 @@ impl ConnectionOptions {
             return Err(connection_url_error());
         }
 
-        let host = match url.host_str() {
-            Some(host) => Some(try!(CString::new(host.as_bytes()))),
+        let host = match url.host() {
+            Some(Host::Ipv6(host)) => Some(try!(CString::new(host.to_string()))),
+            Some(host) => Some(try!(CString::new(host.to_string()))),
             None => None,
         };
         let user = try!(decode_into_cstring(url.username()));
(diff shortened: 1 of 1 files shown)


diesel/src/mysql/connection/url.rs
```diff
@@ -29,8 +29,10 @@ impl ConnectionOptions {
             return Err(connection_url_error());
         }
 
-        let host = match url.host_str() {
-            Some(host) => Some(try!(CString::new(host.as_bytes()))),
+        let host = match url.host() {
+            // Avoid brackets around IPv6 addresses.
```
sgrif: Can we delete this comment? The commit message should have all the context needed.

diesel/src/mysql/connection/url.rs
```diff
@@ -29,8 +29,10 @@ impl ConnectionOptions {
...
+            Some(Host::Ipv6(host)) => Some(try!(CString::new(host.to_string()))),
+            Some(host) => Some(try!(CString::new(host.to_string()))),
```
sgrif: Can this be written as `Some(Host::Ipv6(host)) | Some(host) => Some(CString::new(host.to_string())?),`

kulshrax: Unfortunately it can't because `host` has different types in the two match arms (`std::net::Ipv6Addr` vs `mysql::connection::url::url::Host`). 

I can change it to use the `?` operator instead of the `try!` macro -- I only chose to use the latter to maintain consistency with the surrounding code.

sgrif: No, it's fine to leave it using `try!` for now.

PR #1569, opened 2018-02-21
Include diesel_compile_tests in the root workspace

Eijebong:
I don't know why it wasn't there in the first place but just removing
the `[workspace]` line from its Cargo.toml and adding it to the
workspace members seems to be working.


Eijebong:
Ha tests are failing, that's why this wasn't done

PR #1570, opened 2018-02-21
Derive `AsChangeset` also for the plain struct, similar to `Insertable`

weiznich:


diesel_derives/src/as_changeset.rs
```diff
@@ -20,22 +20,28 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     let struct_name = &model.name;
     let table_name = model.table_name();
 
-    let (_, ty_generics, where_clause) = item.generics.split_for_impl();
-    let mut impl_generics = item.generics.clone();
-    impl_generics.params.push(parse_quote!('update));
-    let (impl_generics, _, _) = impl_generics.split_for_impl();
+    let (impl_generics, ty_generics, where_clause) = item.generics.split_for_impl();
+    let mut ref_generics = item.generics.clone();
+    ref_generics.params.push(parse_quote!('update));
```
sgrif: Do we actually need to split these? Rust doesn't mind if there's an unused lifetime in the generics.

diesel_derives/src/as_changeset.rs
```diff
@@ -71,27 +88,33 @@ fn field_changeset_ty(
     field: &Field,
     table_name: syn::Ident,
     treat_none_as_null: bool,
+    lifetime: quote::Tokens,
```
sgrif: This feels to implicit to me. Can we take `Option<syn::Lifetime>` instead?

CHANGELOG.md
```diff
@@ -95,9 +98,6 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * `exists` can no longer incorrectly receive values other than select
   statements.
 
-* `MysqlConnection::establish` can now properly handle IPv6 addresses wrapped in
```
sgrif: Did you mean to remove this?

PR #1571, opened 2018-02-21
Derive `Identifiable` similar to `Insertable` and `AsChangeset`

weiznich:


diesel_derives/src/identifiable.rs
```diff
@@ -14,12 +14,14 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
...
     let (ref_generics, ..) = ref_generics.split_for_impl();
 
-    let (field_ty, field_access): (Vec<_>, Vec<_>) = model
+    let (ref_field_ty, ref_field_access): (Vec<_>, Vec<_>) = model
         .primary_key_names
         .iter()
         .filter_map(|&pk| model.find_column(pk).emit_error())
         .map(|f| (&f.ty, f.name.access()))
         .unzip();
+    let direct_field_ty = ref_field_ty.clone();
```
sgrif: Can we not use the same variable twice in `quote!`?

Eijebong: @sgrif I personally didn't find a way to do it when I needed it

weiznich: So it seems it is possible to use the same variable in `quote!` twice.
Internally `quote!` seems to use `into_iter` so passing a `Vec<_>` to quote will cause an error about a moved variable, but if we use a reference instead it works :wink:

diesel_derives/src/identifiable.rs
```diff
@@ -14,12 +14,13 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     ref_generics.params.push(parse_quote!('ident));
     let (ref_generics, ..) = ref_generics.split_for_impl();
 
-    let (field_ty, field_access): (Vec<_>, Vec<_>) = model
+    let t: (Vec<_>, Vec<_>) = model
```
sgrif: Why the temporary variable here? `let (ref field_ty, ref field_access)` should work fine.

diesel_derives/src/identifiable.rs
```diff
@@ -14,7 +14,7 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     ref_generics.params.push(parse_quote!('ident));
     let (ref_generics, ..) = ref_generics.split_for_impl();
 
-    let (field_ty, field_access): (Vec<_>, Vec<_>) = model
+    let &(ref field_ty, ref field_access): &(Vec<_>, Vec<_>) = &model
```
sgrif: None of the `&`s are needed.

weiznich:
@diesel-rs/core Adding this implementation is not considered as breaking change, right? (Asking because I needed to change the tests to compile)

sgrif:
Looking at the tests that had to change, I'm not comfortable with this. Strictly speaking it could be called a minor breaking change, but this seems likely to actually affect apps.

weiznich:
> Looking at the tests that had to change, I'm not comfortable with this. Strictly speaking it could be called a minor breaking change, but this seems likely to actually affect apps.

I'm fine with not having this. I've found a other solution for my affected code, so feel free to close this.

PR #1572, opened 2018-02-22
add statement position to error info

wg:
PostgreSQL can return the position of an error which is useful when
executing arbitrary SQL statements. Include this position in the
DatabaseErrorInformation implementation for PG.

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -120,6 +120,13 @@ pub trait DatabaseErrorInformation {
     /// Currently this method will return `None` for all backends other than
     /// PostgreSQL.
     fn constraint_name(&self) -> Option<&str>;
+
+    /// An optional integer indicating an error cursor position as an index into
+    /// the original statement string.
+    ///
+    /// Currently this method will return `None` for all backends other than
+    /// PostgreSQL.
+    fn statement_position(&self) -> Option<&str>;
 }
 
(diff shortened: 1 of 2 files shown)


diesel/src/result.rs
```diff
@@ -120,6 +120,13 @@ pub trait DatabaseErrorInformation {
     /// Currently this method will return `None` for all backends other than
     /// PostgreSQL.
     fn constraint_name(&self) -> Option<&str>;
+
+    /// An optional integer indicating an error cursor position as an index into
```
sgrif: If this is an optional integer, why is the return type `Option<&str>`?

staninprague:
This is really a string value containing a decimal integer, per: https://www.postgresql.org/docs/13/libpq-exec.html:

> PG_DIAG_STATEMENT_POSITION
> A string containing a decimal integer indicating an error cursor position as an index into the original statement string. The first character has index 1, and positions are measured in characters not bytes.

Question is if @sgrif would see the problem with it returned as per postgresql definition, as a string, containing a decimal integer.

weiznich:
> This is really a string value containing a decimal integer, per: https://www.postgresql.org/docs/13/libpq-exec.html:

That sounds like an implementation detail then. If it essentially returns a integer we should return an integer from this function.

henryboisdequin:
Closed as #2652 took over this PR

PR #1573, opened 2018-02-22
Adding `barrel` feature

spacekookie:
This is the initial version of integrating `barrel` into diesel. A few things are still not optimal and I'd like to get some feedback regarding that.

1. There are some barrel specific functions and structs in the `migrations_internal` module which are only compiled when using barrel. I would love to keep those parts in the barrel crate but can't because of cyclic dependencies. Is there a different way to solve this? If so, how?

2. Currently (for testing) the `barrel` feature is always enabled on diesel. How would I add `features = [ "barrel"]` to the `migration_internals` crate, depending on the external flag? Is there a more elegant way of doing this?

3. Should the code added to `diesel_cli` be conditionally added with a `barrel` feature flag too?

Additionally the barrel dependency currently points to `git` which should be changed once `barrel 0.2` is released

### DIFF
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -6,9 +6,7 @@ use std::convert::From;
 use std::{fmt, io};
 use std::path::PathBuf;
 use std::error::Error;
-
 use result;
-
 /// Errors that occur while preparing to run migrations
 #[derive(Debug)]
 pub enum MigrationError {
@@ -76,7 +74,7 @@ impl From<io::Error> for MigrationError {
     }
 }
 
-/// Errors that occur while running migrations
+///Errors that occur while running migrations
 #[derive(Debug, PartialEq)]
 #[cfg_attr(feature = "clippy", allow(enum_variant_names))]
 pub enum RunMigrationsError {
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -1,16 +1,27 @@
-use diesel::result;
+//! Error types that represent migration errors.
+//! 
+//! These are split into multiple segments, depending on
+//! where in the migration process an error occurs.
 
 use std::convert::From;
 use std::{fmt, io};
 use std::path::PathBuf;
 use std::error::Error;
 
+use result;
+
+/// Errors that occur while preparing to run migrations
 #[derive(Debug)]
 pub enum MigrationError {
+    /// The migration directory wasn't found
     MigrationDirectoryNotFound,
+    /// Provided migration was in an unknown format
     UnknownMigrationFormat(PathBuf),
+    /// General system IO error
     IoError(io::Error),
+    /// Provided migration had an incompatible version number
     UnknownMigrationVersion(String),
+    /// No migrations had to be/ could be run
     NoMigrationRun,
 }
 
@@ -63,11 +74,15 @@ impl From<io::Error> for MigrationError {
     }
 }
 
+/// Errors that occur while running migrations
 #[derive(Debug, PartialEq)]
 #[cfg_attr(feature = "clippy", allow(enum_variant_names))]
 pub enum RunMigrationsError {
+    /// A general migration error occured
     MigrationError(MigrationError),
+    /// The provided migration included an invalid query
     QueryError(result::Error),
+    /// The provided migration was empty
     EmptyMigration,
 }
 
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -1,5 +1,4 @@
 //! Error types that represent migration errors.
-//! 
 //! These are split into multiple segments, depending on
 //! where in the migration process an error occurs.
 
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -22,6 +22,9 @@ pub enum MigrationError {
     UnknownMigrationVersion(String),
     /// No migrations had to be/ could be run
     NoMigrationRun,
+    /// 
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl Error for MigrationError {
@@ -83,6 +86,9 @@ pub enum RunMigrationsError {
     QueryError(result::Error),
     /// The provided migration was empty
     EmptyMigration,
+    /// 
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl Error for RunMigrationsError {
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -12,19 +12,19 @@ use result;
 /// Errors that occur while preparing to run migrations
 #[derive(Debug)]
 pub enum MigrationError {
-/// The migration directory wasn't found
-MigrationDirectoryNotFound,
-/// Provided migration was in an unknown format
-UnknownMigrationFormat(PathBuf),
-/// General system IO error
-IoError(io::Error),
-/// Provided migration had an incompatible version number
-UnknownMigrationVersion(String),
-/// No migrations had to be/ could be run
-NoMigrationRun,
-/// 
-#[doc(hidden)]
-__NonExhaustive,
+    /// The migration directory wasn't found
+    MigrationDirectoryNotFound,
+    /// Provided migration was in an unknown format
+    UnknownMigrationFormat(PathBuf),
+    /// General system IO error
+    IoError(io::Error),
+    /// Provided migration had an incompatible version number
+    UnknownMigrationVersion(String),
+    /// No migrations had to be/ could be run
+    NoMigrationRun,
+    ///
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl Error for MigrationError {
@@ -80,15 +80,15 @@ impl From<io::Error> for MigrationError {
 #[derive(Debug, PartialEq)]
 #[cfg_attr(feature = "clippy", allow(enum_variant_names))]
 pub enum RunMigrationsError {
-/// A general migration error occured
-MigrationError(MigrationError),
-/// The provided migration included an invalid query
-QueryError(result::Error),
-/// The provided migration was empty
-EmptyMigration,
-/// 
-#[doc(hidden)]
-__NonExhaustive,
+    /// A general migration error occured
+    MigrationError(MigrationError),
+    /// The provided migration included an invalid query
+    QueryError(result::Error),
+    /// The provided migration was empty
+    EmptyMigration,
+    ///
+    #[doc(hidden)]
+    __NonExhaustive,
 }
 
 impl Error for RunMigrationsError {
--- a/diesel/src/migration/errors.rs
+++ b/diesel/src/migration/errors.rs
@@ -44,6 +44,7 @@ impl Error for MigrationError {
             MigrationError::NoMigrationRun => {
                 "No migrations have been run. Did you forget `diesel migration run`?"
             }
+            MigrationError::__NonExhaustive => unreachable!()
         }
     }
 }
@@ -97,6 +98,7 @@ impl Error for RunMigrationsError {
             RunMigrationsError::MigrationError(ref error) => error.description(),
             RunMigrationsError::QueryError(ref error) => error.description(),
             RunMigrationsError::EmptyMigration => "Attempted to run an empty migration.",
+            RunMigrationsError::__NonExhaustive => unreachable!()
         }
     }
 }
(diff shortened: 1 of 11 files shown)


diesel_cli/src/cli.rs
```diff
@@ -51,6 +51,13 @@ pub fn build_cli() -> App<'static, 'static> {
                              for most use cases.",
                         )
                         .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("type")
+                        .long("type")
+                        .short("t")
+                        .takes_value(true)
+                        .help("Specify the migration type."),
```
sgrif: Can we add `.possible_values(&["sql", "barrel"])` and `.default_value("sql")`?

diesel_cli/src/cli.rs
```diff
@@ -51,6 +51,13 @@ pub fn build_cli() -> App<'static, 'static> {
                              for most use cases.",
                         )
                         .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("type")
```
sgrif: Arg names should be uppercase by convention.

diesel_cli/src/main.rs
```diff
@@ -111,27 +111,50 @@ fn run_migration_command(matches: &ArgMatches) {
...
-            let down_path = migration_dir.join("down.sql");
-            println!(
-                "Creating {}",
-                migration_dir_relative.join("down.sql").display()
-            );
-            let mut down = fs::File::create(down_path).unwrap();
-            down.write_all(b"-- This file should undo anything in `up.sql`")
-                .unwrap();
+            match migration_type {
+                Some("barrel") => {
```
sgrif: This needs to be `#[cfg]`d.

diesel_cli/src/main.rs
```diff
@@ -111,27 +111,50 @@ fn run_migration_command(matches: &ArgMatches) {
...
+                    // TODO: Is there a nier way to print this?
+                    let mut barrel_migr = fs::File::create(migr_path).unwrap();
+                    barrel_migr.write(b"/// Handle up migrations \n").unwrap();
+                    barrel_migr
+                        .write(b"fn up(migr: &mut Migration) {} \n\n")
+                        .unwrap();
+                    barrel_migr.write(b"/// Handle down migrations \n").unwrap();
+                    barrel_migr
+                        .write(b"fn down(migr: &mut Migration) {} \n")
+                        .unwrap();
```
sgrif: All of this code should live in Barrel, not Diesel.

diesel_cli/src/main.rs
```diff
@@ -111,27 +111,50 @@ fn run_migration_command(matches: &ArgMatches) {
...
+                }
+                _ => {
+                    let up_path = migration_dir.join("up.sql");
```
sgrif: Can we pull this out into a function?

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -49,6 +58,39 @@ pub fn migration_from(path: PathBuf) -> Result<Box<Migration>, MigrationError> {
     }
 }
 
+#[cfg(feature = "barrel")]
+struct BarrelMigration(PathBuf, String, String, String);
```
sgrif: Can we just move the migration trait from `diesel_migrations` into `diesel` so this code can live in Barrel?

sgrif:
CI is failing.

spacekookie:
Yea, there is a problem with the code I don't know how to solve. I messaged you on gitter about it

sgrif:
Oh whoops! Sorry!

diesel/src/migration/mod.rs
```diff
@@ -19,3 +20,67 @@ pub trait Migration {
         None
     }
 }
+
+/// A migration name
+#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy)]
+pub struct MigrationName<'a> {
```
sgrif: Why did this need to move?

diesel/src/migration/mod.rs
```diff
@@ -19,3 +20,67 @@ pub trait Migration {
         None
     }
 }
+
+/// A migration name
```
sgrif: Can we document what this actually does rather than just repeating the name of the type?

diesel/src/migration/mod.rs
```diff
@@ -19,3 +20,67 @@ pub trait Migration {
...
+#[allow(missing_debug_implementations)]
+#[derive(Clone, Copy)]
+pub struct MigrationName<'a> {
+    /// Wraps around a migration
```
sgrif: Can we provide more information than just repeating the type?

diesel/src/migration/mod.rs
```diff
@@ -19,3 +20,67 @@ pub trait Migration {
...
+    pub migration: &'a Migration,
+}
+
+/// Get the name of a migration
```
sgrif: Can we provide more information than just repeating the function/argument names? Maybe at least a usage example?

spacekookie:
That's a fair point, I moved the `MigrationName` stuff back to `migrations_internals`. I don't really know what else to document about it. Considering that nothing else in that module is documented I think this should be alright.

sgrif:
There's still CI failures to be addressed

diesel_cli/Cargo.toml
```diff
@@ -23,6 +23,7 @@ infer_schema_internals = { version = "1.1.0" }
 clippy = { optional = true, version = "=0.0.185" }
 migrations_internals = { version = "1.1.0" }
 url = { version = "1.4.0", optional = true }
+barrel = { git = "https://github.com/spacekookie/barrel", optional = true, features = ["diesel"] }
```
sgrif: Are you planning on releasing before this is merged? I'm not comfortable putting a git dependency here.

diesel_cli/Cargo.toml
```diff
@@ -36,6 +37,7 @@ lint = ["clippy"]
 postgres = ["diesel/postgres", "infer_schema_internals/postgres", "url"]
 sqlite = ["diesel/sqlite", "infer_schema_internals/sqlite"]
 mysql = ["diesel/mysql", "infer_schema_internals/mysql", "url"]
+barrels = ["migrations_internals/barrel", "barrel"]
```
sgrif: Why `barrels`?

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -77,22 +77,25 @@
...
-mod migration_error;
 #[doc(hidden)]
 pub mod schema;
 
 #[doc(inline)]
 pub use self::connection::MigrationConnection;
 #[doc(inline)]
 pub use self::migration::*;
-pub use self::migration_error::*;
+// pub use self::migration_error::*;
```
sgrif: ?

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -41,6 +32,12 @@ impl<'a> fmt::Display for MigrationName<'a> {
 }
 
 pub fn migration_from(path: PathBuf) -> Result<Box<Migration>, MigrationError> {
+    #[cfg(feature = "barrel")]
+    match ::barrel::diesel::migration_from(&path) {
```
sgrif: What do you think about changing this to `if let Some(migration) = ::barrel::diesel::migration_from(path)`?

diesel_migrations/src/lib.rs
```diff
@@ -78,15 +78,11 @@ extern crate migrations_macros;
...
-pub use migrations_internals::Migration;
-#[doc(inline)]
-pub use migrations_internals::MigrationName;
-#[doc(inline)]
 pub use migrations_internals::MigrationConnection;
-#[doc(inline)]
-pub use migrations_internals::MigrationError;
-#[doc(inline)]
-pub use migrations_internals::RunMigrationsError;
+// #[doc(inline)]
```
sgrif: Why the commented lines?

sgrif:
@spacekookie Looks like there's some minor cleanup left but this looks good overall. I think there's work to be done on the docs for the migration trait but I can do that after merging. Additionally, right now this constitutes a major breaking change. Anybody who is depending on the existence of `diesel_migrations::Migration` will break. We need to re-export anything moved into the core diesel crate from `diesel_migrations` publicly.

spacekookie:
@sgrif Sounds good :+1:
I just published `barrel v0.2.0` on crates.io and changed the dependencies. I'm gonna do the final round of cleanups tomorrow.

There is still an issue that's not caught by the CI which fails to compile diesel when using barrel. I'm wondering if it's some transient fault in diesel or in my own code. Would be nice to get that fixed before shipping the feature.

diesel_migrations/migrations_internals/src/migration.rs
spacekookie: That would be more elegant although rustc tells me `attributes are not yet allowed on 'if' expressions`. The stabilisation is being tracked here: [rust-lang/rust#15701](https://github.com/rust-lang/rust/issues/15701). 

I'd suggest we change the syntax when the feature becomes stable, as it is more elegant.

spacekookie:
Eh, nevermind. Not sure what happened to the compile errors but they've gone away. I've cleaned up the code and adjusted some last things when calling barrel functions. The new feature seems to work and I hope I didn't break the CI again :smile: :crossed_fingers:

diesel_cli/src/cli.rs
```diff
@@ -51,6 +51,15 @@ pub fn build_cli() -> App<'static, 'static> {
                              for most use cases.",
                         )
                         .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("TYPE")
+                        .long("type")
+                        .short("t")
+                        .possible_values(&["sql", "barrel"])
```
killercup: I'd either go with "raw" + "barrel", or "sql" + "rust".

diesel/src/migration/errors.rs
```diff
@@ -1,16 +1,26 @@
...
 use std::convert::From;
 use std::{fmt, io};
 use std::path::PathBuf;
 use std::error::Error;
 
+use result;
+
+/// Errors that occur while preparing to run migrations
 #[derive(Debug)]
 pub enum MigrationError {
```
killercup: We should probably make this non-exhaustive now that it's public, right?

diesel/src/migration/mod.rs
```diff
@@ -0,0 +1,55 @@
+//! Representation of migrations
+
+pub mod errors;
+pub use self::errors::*;
```
killercup: This looks super dangerous adds a hard to see amount of items to our public API!

I does in fact only expose `MigrationError` and `RunMigrationsError`. So let's `pub use` those specifically if needed.

diesel/src/migration/mod.rs
```diff
@@ -0,0 +1,55 @@
+//! Representation of migrations
+
+pub mod errors;
```
killercup: Does this need to be public or is `pub(crate)` enough?

diesel/src/migration/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+pub use self::errors::*;
+
+use connection::SimpleConnection;
+use std::path::Path;
+
+/// Represents a migration that interacts with diesel
```
killercup: Just noticed that's it's super hard to see what's up with SimpleConnection – it's not public and the only way aside from reading the code where you can find it is in the docs for Connection (which it's a supertrait of). Not something for this PR to fix but seemed worth commenting on.

diesel/src/migration/errors.rs
```diff
@@ -63,11 +73,15 @@ impl From<io::Error> for MigrationError {
     }
 }
 
+/// Errors that occur while running migrations
 #[derive(Debug, PartialEq)]
 #[cfg_attr(feature = "clippy", allow(enum_variant_names))]
 pub enum RunMigrationsError {
```
killercup: We should probably make this non-exhaustive now that it's public, right?

diesel/src/migration/mod.rs
spacekookie: This needs to be public, it's re-exported from diesel_migrations

spacekookie:
@sgrif @killercup Is there anything else that should be changed? Otherwise it would be great if we could get this merged :)

sgrif:
@spacekookie Sorry for the delay in replies (both on this and your other PR on the website which I still need to reply to). I've been fairly swamped at work and have been mostly away from this project. My understanding is that this is blocked on the PR bumping clippy. I'm not entirely sure why that is, since your CI failures look like what was fixed on master a while ago. Perhaps try rebasing?

sgrif:
I'm at a conference, but one of the other committers can take care of
merging. I'd like to see this squashed down.

PR #1575, opened 2018-02-26
Replace most raw pointer usage with `NonNull`

sgrif:
This newly stabilized feature indicates that a pointer cannot be null.
This allows certain compiler optimizations, including but not limited to
the "null pointer optimization", which means that `Option<NonNull<T>>`
has the same size as `NonNull<T>`.

Using `NonNull` over raw pointers as much as possible in our code
potentially enables additional compiler optimizations, and better
expresses intent in several places.

Since this type was only stabilized in Rust 1.25, I have stuck the use
of the real thing behind the unstable flag, and provided a polyfill
by default. This keeps with our policy of "only bump the minimum Rust
version if you have a good reason". I don't think this is a good enough
reason to bump on its own. The polyfill lets us structure our code the
same regardless, and it should be trivial to switch to always using the
real thing once we no longer support Rust 1.24 and earlier.

### DIFF
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -8,24 +8,24 @@ use std::sync::{Once, ONCE_INIT};
 use result::{ConnectionError, ConnectionResult, QueryResult};
 use super::url::ConnectionOptions;
 use super::stmt::Statement;
+use util::NonNull;
 
-pub struct RawConnection(*mut ffi::MYSQL);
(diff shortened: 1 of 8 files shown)


sgrif:
> Have we ever written down/settled on a min Rust version policy?

Last time we discussed it, I believe the consensus was "it's fine to bump as long as there's a good reason", and that "makes something easier to implement" is not a good reason if the thing can be implemented without bumping Rust. So basically any time it's required for a new feature or to fix a bug.

PR #1577, opened 2018-02-27
Allow use of uuid 0.6.0

Idolf:
There is a new version of uuid. This is similar to #878, however version 0.6.0 of the uuid crates has renamed the `use_std` feature to `std` and turned it on by default.

As far as I can tell, it is not possible to support all versions in one Cargo.toml file, without either adding the `use_std` flag back into uuid as compatibility or getting them to use the semver-trick.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -24,7 +24,7 @@ quickcheck = { version = "0.4", optional = true }
 serde_json = { version = ">=0.8.0, <2.0", optional = true }
 time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
-uuid = { version = ">=0.2.0, <0.6.0", optional = true, features = ["use_std"] }
+uuid = { version = ">=0.2.0, <0.7.0", optional = true, features = ["use_std"] }
 ipnetwork = { version = "0.12.2", optional = true }
 num-bigint = { version = "0.1.41", optional = true }
 num-traits = { version = "0.1.35", optional = true }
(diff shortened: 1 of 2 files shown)


sgrif:
I'd prefer not to drop support for older versions of UUID. We should just be able to remove the `features =` section, as the `use_std`/`std` feature is enabled by default.

sgrif:
Ugh. Well I guess let's open a PR that adds a backwards compatibility shim? Otherwise it's impossible for us to support both.

sgrif:
This is blocked on https://github.com/uuid-rs/uuid/pull/157

sgrif:
UUID 0.6.1 was released which does include the `use_std` feature, so this should be good to go. Can you re-add that feature usage? (I don't think we can specifically exclude 0.6.0 from the range, but I also don't think we need to)

Idolf:
Done.

Idolf:
Are there any plans to also do a new release for this?

weiznich:
I think this will be part of the diesel 1.2 release.

Idolf:
Do you know an approximate timeframe for that? It is preventing us from upgrading our internal repos to uuid 0.6. I'd prefer not to have an internal fork of diesel.

weiznich:
I think I heard somewhere something like soon (as in the next weeks), but take with some caution, because I'm not the one that decides to make a release.

Eijebong:
@sgrif said this week, so probably next week ;)

PR #1578, opened 2018-02-27
[DO NOT MERGE] use match_default_bindings

sgrif:
This PR is not intended to be merged, it is only to show what our codebase looks like using the `match_default_bindings` feature, as an example to be shown in the stabilization discussion for that feature.

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -247,13 +247,13 @@ impl From<NulError> for Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            Error::InvalidCString(ref nul_err) => nul_err.fmt(f),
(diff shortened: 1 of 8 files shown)


diesel/src/result.rs
```diff
@@ -279,43 +279,43 @@ impl StdError for Error {
...
 
     fn cause(&self) -> Option<&StdError> {
-        match *self {
-            Error::InvalidCString(ref e) => Some(e),
-            Error::QueryBuilderError(ref e) => Some(&**e),
-            Error::DeserializationError(ref e) => Some(&**e),
-            Error::SerializationError(ref e) => Some(&**e),
+        match self {
+            Error::InvalidCString(e) => Some(e),
+            Error::QueryBuilderError(e) => Some(&**e),
```
sgrif: I was surprised that I had to keep `&**` here. I would have expected to be able to remove at least one `*`.

PR #1581, opened 2018-03-01
Add a Insertable implementation for Vec<T>

weiznich:
Similar to the Insertable implementation on structs itself this makes it
easier/possible to write generic code

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -111,6 +111,15 @@ where
     }
 }
 
+impl<T, Tab, DB> CanInsertInSingleQuery<DB> for OwnedBatchInsert<T, Tab>
+where
+    DB: Backend + SupportsDefaultKeyword,
+{
+    fn rows_to_insert(&self) -> Option<usize> {
+        Some(self.values.len())
+    }
+}
+
 impl<T, U, DB> CanInsertInSingleQuery<DB> for ColumnInsertValue<T, U>
 where
     DB: Backend,
@@ -218,6 +227,20 @@ where
     }
 }
 
+impl<T, Tab> Insertable<Tab> for Vec<T>
+where
+    T: Insertable<Tab>,
+{
+    type Values = OwnedBatchInsert<T::Values, Tab>;
+
+    fn values(self) -> Self::Values {
+        OwnedBatchInsert {
+            values: self.into_iter().map(Insertable::values).collect(),
+            _marker: PhantomData,
+        }
+    }
+}
+
 impl<T, Tab> Insertable<Tab> for Option<T>
 where
     T: Insertable<Tab>,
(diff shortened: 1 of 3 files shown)


weiznich:
> This needs explicit tests (as does the derive TBH), but other than that this looks fine.

To write this test we need to merge #1567 first.

diesel/src/insertable.rs
```diff
@@ -267,3 +290,29 @@ where
         Ok(())
     }
 }
+
+#[derive(Debug)]
+pub struct OwnedBatchInsert<V, Tab> {
```
sgrif: I don't like this name, but I don't have a better suggestion.

diesel/src/insertable.rs
```diff
@@ -267,3 +290,29 @@ where
...
+    pub(crate) values: Vec<V>,
+    _marker: PhantomData<Tab>,
```
sgrif: Since this is owned, we don't need to defer grabbing the values. We can Just have this wrap `Vec<T>` where `T = <U as Insertable<Tab>>::Values`

diesel/src/query_builder/insert_statement/mod.rs
```diff
@@ -240,6 +240,28 @@ where
     }
 }
 
+#[cfg(feature = "sqlite")]
+impl<T, U, Op> ExecuteDsl<SqliteConnection>
```
sgrif: This impl bothers me more than the rest of the duplication. I'd like to mull on it for a bit.

diesel/src/insertable.rs
weiznich: It is already done [that way](https://github.com/diesel-rs/diesel/pull/1581/files#diff-6b84e53c349d8eac5892281f4b21cc4dR238) if I understand the comment correctly.

sgrif: Why do we need this field, then?

compressed:
I believe this was a breaking change. I had code that was doing something like this:

```rust
let mut v = Vec::new();
// populate v
v.insert_into(table).execute(conn);
v.clear();
// populate v again
..
```

Now with this impl, this code fails to compile because the `insert_into` is going to move `v`, whereas before `v` would have just been borrowed by `insert_into`. The fix is easy enough though `(&v).insert_into()`. Maybe makes sense to call this out more clearly in the CHANGELOG?

weiznich:
> I believe this was a breaking change

According to [RFC 1105](https://github.com/nox/rust-rfcs/blob/master/text/1105-api-evolution.md) this is not considered as a breaking change that requires a major version bump. Specifically see [this section] (https://github.com/nox/rust-rfcs/blob/master/text/1105-api-evolution.md#minor-change-implementing-any-non-fundamental-trait).

> Maybe makes sense to call this out more clearly in the CHANGELOG?

Feel free to open a PR adding a notice about this to the change log.

PR #1582, opened 2018-03-02
Allow to set fields to `Default` when constructing struct from query

weiznich:
This commit adds a `#[diesel(default)]` field attribute to
`#[derive(Queryable)]`, that allows a user to set a field to a default
value instead of loading it from the database. This is usefull if the
same struct is used to serve some API that requires additional
information to be added.


sgrif:
:-1: from me on this idea. It's too unclear what this does, and I don't think there's an easy way to make it obvious. I don't think it's obvious whether this ignores a database column, populates it on null, or changes the expected SQL type.

weiznich:
Do you see a problem with this feature in general, or with the naming?
In the second case we could do something similar to serde and use `#[diesel(skip)]`.

weiznich:
@sgrif To point to a practical use of this, see [this example](https://github.com/weiznich/wundergraph/blob/master/wundergraph_example/src/main.rs#L143) from my `wundergraph` crate.
If there is any other way to do something similar that does not involve manually implementing `Queryable` I'm fine with using it.

sgrif:
With the feature in general. I think this affects more things in non-obvious ways than serde-skip does.

weiznich:
So is there any other way to solve this without manually implementing `Queryable` on each struct needing this?

PR #1594, opened 2018-03-15
fixed example

Leopard2A5:
`posts::table` should just be `posts` in insert statements.

`posts::table` fails with `function or associated item not found in 'schema::posts::table'`

### DIFF
--- a/examples/postgres/getting_started_step_2/src/lib.rs
+++ b/examples/postgres/getting_started_step_2/src/lib.rs
@@ -26,7 +26,7 @@ pub fn create_post(conn: &PgConnection, title: &str, body: &str) -> Post {
         body: body,
     };
 
-    diesel::insert_into(posts::table)
+    diesel::insert_into(posts)
         .values(&new_post)
         .get_result(conn)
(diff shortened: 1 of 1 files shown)


weiznich:
As [travis](https://travis-ci.org/diesel-rs/diesel/jobs/353733260#L1382) suggests, this does not "fix" the example but break it.

Probably you did this change because it does not work using this variant in your own code? 
The issue here is that there are actually two ways to access the table and column definition.
1. `use schema::my_table::dsl::*;` and then access the table with `my_table` and the columns with the name of the column
2. `use schema::my_table;` and then access the table with `my_table::table` and the columns with `my_table::my_column`
The first variant leads to shorter code, while the second variant is clearer to which table a certain column belongs to (useful for writing complicated queries affecting multiple tables)

So my guess is that you have a `use schema::posts::dsl::*;` somewhere in your code so that the code from the example doesn't work. The example code uses the [second variant](https://github.com/diesel-rs/diesel/pull/1594/files#diff-857b2cf480346591aee31831ef3ccbabL22)

PR #1595, opened 2018-03-16
Bump num-traits to 0.2

Eijebong:


weiznich:
This is a breaking change because it removes compatibility with the old num-trait version.

weiznich:
> Unless I missed it, this isn't used in our public API

Yes, your right :+1: 

I'm not sure what causes the ci error. Otherwise this look fine.

sgrif:
CI is red due to #1602. Assuming this is fine if that weren't a problem.

PR #1596, opened 2018-03-20
Allow function overloading in Postgres via variant_sql_function macro

moore3071:
## Summary

This pull request adds another macro for extending diesel, `variant_sql_function`. This macro is only available to Postgres, and provides a way to overload SQL functions (MySQL doesn't provide function overloading). The format is: `variant_sql_function(rust_function_name, struct_name, sql_name, type signature)`.

`sql_function_body` has been modified to take a sql function name so `variant_sql_function` can use it too.

`sql_function` patterns were reordered to flow more logically (the third pattern doesn't call the first).

## Does this break backwards combatibility at all?

Nope!

## What needs improving?

Just like `sql_function`, it would be nice to need less arguments. Unfortunately, `concat_idents` isn't that useful. Luckily, macros 2.0 seems to be moving and would allow less arguments.

### DIFF
--- a/diesel/src/expression/functions/mod.rs
+++ b/diesel/src/expression/functions/mod.rs
@@ -131,7 +131,7 @@ macro_rules! sql_function {
 }
 
 #[macro_export]
-#[cfg(feature = "postgres")]
+#[cfg(any(feature = "postgres", feature = "sqlite"))]
 /// Declare a variant of an already defined sql function for use in your code. Useful if you have
 /// your own SQL functions that you'd like to use that are overloaded. You can optionally provide a
 /// doc string as well. `$struct_name` and `$fn_name` should be unique names. You will not need to
--- a/diesel/src/expression/functions/mod.rs
+++ b/diesel/src/expression/functions/mod.rs
@@ -131,7 +131,7 @@ macro_rules! sql_function {
 }
 
 #[macro_export]
-#[cfg(feature= "postgres" )]
+#[cfg(feature = "postgres")]
 /// Declare a variant of an already defined sql function for use in your code. Useful if you have
 /// your own SQL functions that you'd like to use that are overloaded. You can optionally provide a
 /// doc string as well. `$struct_name` and `$fn_name` should be unique names. You will not need to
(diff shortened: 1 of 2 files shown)


diesel_tests/tests/macros.rs
```diff
@@ -5,9 +5,10 @@
 #![cfg(feature = "postgres")]
 use schema::*;
 use diesel::*;
-use diesel::sql_types::{BigInt, VarChar};
+use diesel::sql_types::{BigInt, Int4, VarChar};
```
weiznich: Better use `Integer` instead of `Int4`.

diesel/src/expression/functions/mod.rs
```diff
@@ -114,18 +119,47 @@ macro_rules! sql_function {
...
+#[macro_export]
+#[cfg(feature = "postgres")]
+/// Declare a variant of an already defined sql function for use in your code. Useful if you have
+/// your own SQL functions that you'd like to use that are overloaded. You can optionally provide a
+/// doc string as well. `$struct_name` and `$fn_name` should be unique names. You will not need to
+/// reference `$struct_name` in your code.
+///
+/// This will generate a rust function with the same name to construct the expression, and a helper
+/// type which represents the return type of that function. The function will automatically convert
+/// its arguments to expressions.
```
weiznich: There should be an example here.

diesel_tests/tests/macros.rs
```diff
@@ -38,6 +46,10 @@ fn test_sql_function() {
             .load(&connection)
             .unwrap()
     );
+    assert_eq!(
+        vec![sean],
+        users.filter(my_lower_id(name, 1)).load(&connection).unwrap()
```
weiznich: This does generate invalid sql, therefore the compilation fails.

moore3071:
@weiznich, all problems from your review should be fixed now. It's just the Clippy part that's failing the build now (but it looks like the Diesel team already brought that up with the Clippy team).

diesel/src/expression/functions/mod.rs
```diff
@@ -114,18 +119,66 @@ macro_rules! sql_function {
...
         $docs:expr,
         $helper_ty_docs:expr
     ) => {
-        sql_function_body!($fn_name, $struct_name, $args -> $return_type, $docs, $helper_ty_docs);
+        sql_function_body!($fn_name, $struct_name, $fn_name, $args -> $return_type, $docs, $helper_ty_docs);
+    };
+}
+
+#[macro_export]
+#[cfg(feature = "postgres")]
```
sgrif: Is this PG specific? `COALESCE` can have many signatures and is ANSI SQL.

moore3071: MySQL doesn't support function overloading. SQLite is an odd case. It doesn't have `CREATE FUNCTION` and functions are created through the C API. I Thought that it didn't support overloading, but after rereading the [functions page](https://www.sqlite.org/c3ref/create_function.html) it appears that they can be. I'll add a commit so that this works for SQLite.

moore3071:
I don't know much about `macros 1.1` (yet), but I agree that that's the better approach for this issue. It's more concise, logical, and forward thinking. I'd love to help out with this; my main concern would be the support for `macros 1.1`. Are they supported in stable currently?

If so, would you mind, @sgrif, if I closed this PR and opened an issue for discussing this new sql_function macro?

sgrif:
Yes, Macros 1.1 has been stable for well over a year. I think it was Rust 1.17? I know it was by far the biggest release of Rust. It's the reason Serde and Diesel work on stable.

sgrif:
> If so, would you mind, @sgrif, if I closed this PR and opened an issue for discussing this new sql_function macro?

Up to you. It's your PR. ;)

PR #1598, opened 2018-03-21
Add a trait to abstract away the access to foreign key fields in belongs_to

weiznich:
This allows to use wrapper constructs as foreign key fields. Without
this change a foreign key is required to have "the same" type as the
primary key of the parent struct. After this change it is possible to
do some kind of transformation to get a compatible type out of the
wrapper struct

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -8,6 +8,24 @@ use super::{HasTable, Identifiable};
 use std::borrow::Borrow;
 use std::hash::Hash;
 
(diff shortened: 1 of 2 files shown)


sgrif:
I don't understand how this trait differs from `BelongsTo`.

weiznich:
The difference to `BelongsTo` is that this must not be implemented for every relation, but only once for a wrapper type. See the linked commit from wundergraph for a specific example. (This feature is not for normal users of diesel, but rather for people wanting to abstract over certain features in diesel.)

PR #1600, opened 2018-03-26
Bump clippy

killercup:


killercup:
Currently, CI fails because cargo_metadata doesn't compile on the nightly use for clippy.

Eijebong:
@killercup This is going to fail because of https://github.com/rust-lang-nursery/rust-clippy/issues/2560

Eijebong:
You have to bump compiletest too

PR #1610, opened 2018-04-02
Add support for serializing `std::time::Duration` and

weiznich:
`chrono::Duration` to `Interval`

We add only a `ToSql` implementation because it is possible to map
every instance to a `Interval`. `FromSql` is not added because it is
not possible to represent every instance of as on of this two
`Duration` types.
Normally we only support types that allows us to implement `ToSql` and
`FromSql`, but in this case we make a exception to make it possible to
use the `Duration` types in a context like `now() - Duration::minutes(42)`

### DIFF
--- a/diesel/src/pg/types/date_and_time/std_time.rs
+++ b/diesel/src/pg/types/date_and_time/std_time.rs
@@ -59,6 +59,19 @@ fn duration_to_usecs(duration: Duration) -> u64 {
     seconds + u64::from(subseconds)
 }
 
+impl ToSql<sql_types::Interval, Pg> for Duration {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        use dsl::IntervalDsl;
+        let seconds = self.as_secs() as i64;
+        let pg_interval = seconds
+            .checked_mul(1000 * 1000)
+            .and_then(|v| v.checked_add(i64::from(self.subsec_nanos() / NANO_PER_USEC)))
+            .map(|v| v.microseconds())
+            .unwrap_or_else(|| seconds.seconds());
+        <_ as ToSql<sql_types::Interval, Pg>>::to_sql(&pg_interval, out)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     extern crate dotenv;
@@ -69,7 +82,7 @@ mod tests {
     use dsl::{now, sql};
     use prelude::*;
     use select;
-    use sql_types::Timestamp;
+    use sql_types::{Timestamp, Interval};
 
     fn connection() -> PgConnection {
         dotenv().ok();
@@ -104,6 +117,19 @@ mod tests {
 
         let time = SystemTime::now() - Duration::from_secs(60);
         let query = select(now.at_time_zone("utc").gt(time));
+        assert!(query.get_result::<bool>(&connection).unwrap());
+    }
+
+    #[test]
+    fn interval_encode_correctly() {
+        let connection = connection();
+
+        let seconds = Duration::from_secs(42);
+        let query = select(sql::<Interval>("'42 seconds'::interval").eq(seconds));
+        assert!(query.get_result::<bool>(&connection).unwrap());
+
+        let millis = Duration::from_millis(42);
+        let query = select(sql::<Interval>("'42 milliseconds'::interval").eq(millis));
         assert!(query.get_result::<bool>(&connection).unwrap());
     }
 }
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -286,6 +286,10 @@ pub struct Date;
 /// ### [`ToSql`](../serialize/trait.ToSql.html) impls
 ///
 /// - [`PgInterval`] which can be constructed using [`IntervalDsl`]
+/// - [`std::time::Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html)
+///   [^microseconds]
+/// - [`chrono::Duration`](/chrono/struct.Duration.html)
+///   [^microseconds] with `feature = "chrono"`
 ///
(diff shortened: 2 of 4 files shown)


diesel/src/pg/types/date_and_time/mod.rs
```diff
@@ -80,6 +80,75 @@ impl PgInterval {
     pub fn from_months(months: i32) -> Self {
         Self::new(0, 0, months)
     }
+
+    fn from_duration(mut seconds: i64, microseconds: i64) -> Self {
```
sgrif: This is very much not the right approach. We cannot assume that `25.hours()` means `1.day() + 1.hour()`. `chrono::Duration` is lossy. We should not try to infer what they meant when they constructed it. We should convert the types with their direct translation:

```
PgInterval {
    microseconds: duration.num_microseconds.expect("Cannot convert a duration larger than 292471 years"),
    ..Default::default(),
}
```

diesel/src/type_impls/date_and_time.rs
```diff
@@ -7,6 +7,11 @@ use std::time::SystemTime;
 #[sql_type = "::sql_types::Timestamp"]
 struct SystemTimeProxy(SystemTime);
 
+#[derive(FromSqlRow, AsExpression)]
+#[diesel(foreign_derive)]
+#[cfg_attr(feature = "postgres", sql_type = "::sql_types::Interval")]
+struct SystemDuration(::std::time::Duration);
```
sgrif: Can we follow the convention used for every other type in the name?

diesel/src/type_impls/date_and_time.rs
```diff
@@ -34,4 +39,9 @@ mod chrono {
     #[diesel(foreign_derive)]
     #[cfg_attr(feature = "postgres", sql_type = "::sql_types::Timestamptz")]
     struct DateTimeProxy<Tz: TimeZone>(DateTime<Tz>);
+
+    #[derive(FromSqlRow, AsExpression)]
+    #[diesel(foreign_derive)]
+    #[cfg_attr(feature = "postgres", sql_type = "::sql_types::Interval")]
+    struct IntervalProxy(Duration);
```
sgrif: DurationProxy

sgrif:
You might want to give a better motivating example since the one you gave will be 1 hour off from what you expect in most time zones for roughly half the year. :)

diesel/src/pg/types/date_and_time/std_time.rs
```diff
@@ -57,6 +57,19 @@ fn duration_to_usecs(duration: Duration) -> u64 {
...
 
+impl ToSql<sql_types::Interval, Pg> for Duration {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        use dsl::IntervalDsl;
+        let seconds = self.as_secs() as i64;
+        let pg_interval = seconds
+            .checked_mul(1000 * 1000)
+            .and_then(|v| v.checked_add(i64::from(self.subsec_nanos() / NANO_PER_USEC)))
+            .map(|v| v.microseconds())
+            .unwrap_or_else(|| seconds.seconds());
```
sgrif: On the one hand I don't like that we're silently losing precision rather than returning an error on overflow. On the other hand this only occurs for durations greater than thousands of years so I don't really care.

diesel/src/sql_types/mod.rs
```diff
@@ -285,13 +285,21 @@ pub struct Date;
...
+///   [^microseconds] with `feature = "chrono"`
 ///
 /// ### [`FromSql`](../deserialize/trait.FromSql.html) impls
 ///
 /// - [`PgInterval`] which can be constructed using [`IntervalDsl`]
 ///
 /// [`PgInterval`]: ../pg/data_types/struct.PgInterval.html
 /// [`IntervalDsl`]: ../pg/expression/extensions/trait.IntervalDsl.html
+/// [^microseconds]:  Note: The conversion is done on microseconds basis,
+///   that means any timeshift (like leepseconds, daylaight saving,
```
sgrif: "leap seconds", "daylight savings time"

diesel/src/sql_types/mod.rs
```diff
@@ -285,13 +285,21 @@ pub struct Date;
...
+///   …) is ignored. Also this conversion will overflow for a
```
sgrif: The implementation doesn't overflow so maybe let's not mention this

diesel/src/sql_types/mod.rs
```diff
@@ -285,13 +285,21 @@ pub struct Date;
 /// ### [`ToSql`](../serialize/trait.ToSql.html) impls
 ///
 /// - [`PgInterval`] which can be constructed using [`IntervalDsl`]
+/// - [`std::time::Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html)
+///   [^microseconds]
+/// - [`chrono::Duration`](/chrono/struct.Duration.html)
```
sgrif: This link will break on docs.rs.

weiznich: This means that the docs for `Time` and `Timestamp` are also broken, because they do the same. 
What is the preferred way to handle this? Just link explicitly to docs.rs?

joelgallant: Can it not just be `[chrono::Duration]` with backticks? ([eg](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md#implied-shortcut-reference-links))

McDic:
Are we ever going to add `chrono::Duration` support? Currently the built-in support is only available for `chrono::DateTime`, which sounds a bit strange.

weiznich:
@McDic This kind of question is highly discouraged on old stale PR's like this one. I'm closing this PR now as it's heavily outdated and has merge conflicts. I do not plan to continue working on this feature anytime soon.

You are welcome to submit a new PR that implements this feature and adds the necessary tests + documentation.

PR #1615, opened 2018-04-05
Add a flag to help users better debug bad `Queryable` impls

sgrif:
Since the impl we generate is so generic, when a bad `Queryable` impl is
given, the error doesn't occur until they actually try to load a query.
This ends up being really non-local, and gives little to no feedback
about which field is actually the problem.

There is a new option called `#[check_types]` which completely changes
how the impl is generated, in such a way that when the `unstable`
feature is enabled the error will point at the specific field that is
the problem. The actual message is the same, but where it points makes
debugging much easier.

This flag also generates code which validates that your fields are in
the right order to match your columns.

Unfortunately, we cannot have this generate an actual usable impl.
Trait bounds not matching in types or where clauses will always result
in an error pointing at the derive itself, which is no more helpful than
where we started. For that reason, we can't actually have any usable
`Row` type. So this flag is for debugging purposes only.

This does not help with cases where the number of fields is incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   rounding issues. This is primarily to support things like `avg(int_col)`,
   which we define as returning `Numeric`
 
+* `#[check_types]` has been added to help debug missing `Queryable` impls. See
+  [the docs][check_types-1.2.0] for details.
+
+[check_types-1.2.0]: http://docs.diesel.rs/diesel/deserialize/trait.Queryable.html#debugging
+
 ### Changed
 
 * The bounds on `impl ToSql for Cow<'a, T>` have been loosened to no longer
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -98,6 +98,93 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
 /// assert_eq!(expected, first_user);
 /// #     Ok(())
 /// # }
+/// ```
+///
+/// # Debugging
+///
+/// When your struct has the wrong types for your query, the error messages
+/// can often be hard to interpret. For example, take this struct where
+/// `hair_color` and `updated_at` should both be `Option`s. Additionally,
+/// `created_at` and `updated_at` are in the wrong order.
+///
+/// ```ignore
+/// table! {
+///     users {
+///         id -> Integer,
+///         name -> Text,
+///         hair_color -> Nullable<Text>,
+///         created_at -> Timestamp,
+///         updated_at -> Nullable<Timestamp>,
+///     }
+/// }
+///
+/// #[derive(Queryable)]
+/// struct User {
+///     id: i32,
+///     name: String,
+///     hair_color: String,
+///     updated_at: SystemTime,
+///     created_at: SystemTime,
+/// }
+/// ```
+///
+/// We get this error:
+///
+/// ```text
+///    |
+/// 28 |     users::table.load::<User>(&conn).unwrap();
+///    |                  ^^^^ the trait `diesel::deserialize::FromSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamp>, diesel::pg::Pg>` is not implemented for `std::time::SystemTime`
+///
+///    |
+/// 28 |     users::table.load::<User>(&conn).unwrap();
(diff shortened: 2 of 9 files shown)


diesel/src/deserialize.rs
```diff
@@ -98,6 +98,93 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
+/// `created_at` and `updated_at` are in the wrong order.
+///
+/// ```ignore
+/// table! {
+///     users {
+///         id -> Integer,
+///         name -> Text,
+///         hair_color -> Nullable<Text>,
+///         created_at -> Timestamp,
+///         updated_at -> Timestamp,
```
Eijebong: Shouldn't this be `Nullable<Timestamp>`

sgrif: Yes good catch

diesel/src/deserialize.rs
```diff
@@ -98,6 +98,93 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
+///    |     ^^^^^^^^^^ the trait `diesel::deserialize::FromSql<diesel::sql_types::Nullable<diesel::sql_types::Text>, diesel::pg::Pg>` is not implemented for `*const str`
+///
+///    |
+/// 24 |     updated_at: SystemTime,
+///    |     ^^^^^^^^^^ the trait `diesel::deserialize::FromSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamp>, diesel::pg::Pg>` is not implemented for `std::time::SystemTime`
+///
+///   |
+/// 24| /     updated_at: SystemTime,
+///   |
+///   = note: expected type `users::columns::created_at`
```
Eijebong: what's going on here ?

Eijebong: Ohhh I didn't see the order was wrong :D

diesel_derives/src/queryable.rs
```diff
@@ -49,3 +63,100 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
...
+            pub enum DummyRow {}
+
+            impl<ST> FromSqlRow<ST, #backend> for DummyRow {
+                fn build_from_row<R: self::diesel::row::Row<#backend>>(_: &mut R)
+                    -> deserialize::Result<Self>
+                {
+                    use self::std::result::Result::Err;
+                    use self::std::convert::Into;
+
+                    Err("`#[check_types]` is only for debugging purposes".into())
```
weiznich: I think that should rather panic than return a `Err`, because this is clearly a programmer mistake. 
(We could even think about using `compile_error!` if this typechecks correctly.)

diesel/src/deserialize.rs
```diff
@@ -98,6 +98,93 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
+///              found type `users::columns::updated_at`
+///
+///   |
+/// 25| /     created_at: SystemTime,
+///   |
+///   = note: expected type `users::columns::updated_at`
+///              found type `users::columns::created_at`
+/// ```
+///
+/// Keep in mind that `#[check_types]` is only for debugging purposes. The impl
```
weiznich: We should mention somewhere here that this feature needs the unstable feature and a nightly compiler.

diesel_derives/src/queryable.rs
Eijebong: Nah, because you could have lots of code, add a column to you `Queryable` struct and still want the project to compile instead of throwing 20 000 compile errors telling you that `#[check_types]` is for debugging only.

CHANGELOG.md
```diff
@@ -82,6 +82,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   rounding issues. This is primarily to support things like `avg(int_col)`,
   which we define as returning `Numeric`
 
+* `#[check_types]` has been added to help debug missing `Queryable` impls. See
```
Eijebong: I don't think this is right

diesel/src/deserialize.rs
weiznich: @sgrif I do not see this anywhere mentioned.

CHANGELOG.md
sgrif: What's wrong with it?

Eijebong: They're not missing, just wrong impls

sgrif: I worded it this way because the message the user gets is literally "`Queryable<...>` is not implemented"

Eijebong: Makes sense I guess

weiznich:
That's now basically implemented by https://github.com/diesel-rs/diesel/pull/3359

PR #1616, opened 2018-04-05
Tighten bounds on select statements used as a subselect

### LINKED ISSUE #1587: Invalid SQL generated for SelectStatements and subqueries.
### Versions

- **Diesel:** master (ac3074338) / 1.1
- **Database:** postgres

### Feature Flags

- **diesel:** postgres

## Problem Description

Without a explicit `select` clause diesel will generate invalid sql when using normal select statement as subquery in a `IN` clause.
This does not happen with `BoxedSelectStatements`.

```rust

table!{
    users{
       id -> Integer,
       name -> Text,
    }
}

table! {
    posts {
        id -> Integer,
        user_id -> Integer,
    }
}

fn test(conn: &PgConnection) {
    let subquery = users::table.filter(users::id.eq(1));
    let query = posts::table.filter(posts::user_id.eq_any(subquery));
    query.execute(conn).unwrap();
}
```

A compiler error.

The program compiles fine, but errors with a invalid query at runtime.
`DatabaseError(__Unknown, "subquery has too many columns")`

No

sgrif:
We were accidentally allowing them to coerce to any SQL type.
Fixes #1587.

### DIFF
--- a/diesel_compile_tests/tests/compile-fail/subselect_requires_correct_type.rs
+++ b/diesel_compile_tests/tests/compile-fail/subselect_requires_correct_type.rs
@@ -1,0 +1,24 @@
+#[macro_use] extern crate diesel;
+
+use diesel::*;
+
+table!{
+    users{
+       id -> Integer,
+       name -> Text,
+    }
+}
+
+table! {
+    posts {
+        id -> Integer,
+        user_id -> Integer,
+    }
+}
+
+fn main() {
+    let conn = PgConnection::establish("").unwrap();
+    let subquery = users::table.filter(users::id.eq(1));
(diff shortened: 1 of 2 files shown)


diesel_compile_tests/tests/compile-fail/subselect_requires_correct_type.rs
```diff
@@ -0,0 +1,24 @@
...
+
+table! {
+    posts {
+        id -> Integer,
+        user_id -> Integer,
+    }
+}
+
+fn main() {
+    let conn = PgConnection::establish("").unwrap();
```
Eijebong: Why is this needed ?

PR #1618, opened 2018-04-05
Rebase of #1561

### LINKED ISSUE #1181: Handling unsigned integer types from MySQL
I've got a lot of MySQL tables with unsigned integer types (e.g. `id INT(10) UNSIGNED`).

I'm struggling to run queries against these with Diesel when their values exceed the positive `i32` max value.

E.g. using:

```
extern crate diesel;

use diesel::prelude::*;
use diesel::mysql::MysqlConnection;
use diesel::expression::sql;
use diesel::types::Integer;

#[derive(PartialEq, Debug, Queryable)]
struct Thing {
    id: i32
}

fn main() {
    let url = "mysql://user:pass@localhost/test"
    let conn = MysqlConnection::establish(&url).unwrap();

    let query = sql::<Integer>("SELECT 4000000000");
    query.load::<Thing>(&conn);
}
```
results in a runtime error:
```
Err(DeserializationError(StringError("Numeric overflow/underflow occurred")))
```
which is expected, as it overflows `i32`, but is an acceptable `u32` value.

Changing my struct to:
```
struct Thing {
    id: u32
}
```
instead results in a compiler error:
```
the trait `diesel::types::FromSqlRow<diesel::types::Integer, _>` is not implemented for `(u32,)
```

Is this just unsupported by Diesel, or is there a different approach I should be taking for dealing with MySQL's unsigned integers?

sgrif:
Close #1561 
Close #1181

### DIFF
--- a/diesel/src/mysql/types/mod.rs
+++ b/diesel/src/mysql/types/mod.rs
@@ -5,36 +5,228 @@ mod date_and_time;
 mod numeric;
 
 use std::io::Write;
-use byteorder::WriteBytesExt;
+use byteorder::{ReadBytesExt, WriteBytesExt};
 
-use deserialize::{self, FromSql};
-use mysql::Mysql;
+use row::Row;
+use deserialize::{self, FromSql, FromSqlRow, Queryable};
+use backend::Backend;
+use mysql::{Mysql, MysqlType};
 use serialize::{self, IsNull, Output, ToSql};
-use sql_types;
+use expression::{AppearsOnTable, Expression, SelectableExpression};
+use query_builder::{AstPass, QueryFragment, QueryId};
+use result::QueryResult;
+use sql_types::*;
 
-impl ToSql<sql_types::Tinyint, Mysql> for i8 {
+impl ToSql<Tinyint, Mysql> for i8 {
     fn to_sql<W: Write>(&self, out: &mut Output<W, Mysql>) -> serialize::Result {
         out.write_i8(*self).map(|_| IsNull::No).map_err(Into::into)
     }
 }
 
-impl FromSql<sql_types::Tinyint, Mysql> for i8 {
+impl FromSql<Tinyint, Mysql> for i8 {
     fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
         let bytes = not_none!(bytes);
         Ok(bytes[0] as i8)
     }
 }
 
-impl ToSql<sql_types::Bool, Mysql> for bool {
+/// Represents the MySQL unsigned type.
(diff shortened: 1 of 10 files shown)


sgrif:
@diesel-rs/reviewers This needs a re-review. This needed some pretty substantial work.

diesel/src/mysql/types/mod.rs
```diff
@@ -8,36 +8,88 @@ use std::io::Write;
...
 }
 
-impl ToSql<sql_types::Bool, Mysql> for bool {
+/// Represents the MySQL unsigned type.
+#[derive(Debug, Clone, Copy, Default, SqlType, QueryId)]
+pub struct Unsigned<ST>(ST);
+
+impl ToSql<Unsigned<SmallInt>, Mysql> for u16 {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Mysql>) -> serialize::Result {
+        ToSql::<SmallInt, Mysql>::to_sql(&(*self as i16), out)
```
weiznich: Just to be sure. It's okay here to convert this to a signed integer? 
This does not use some kind of undefined behavior in case of self > 128? 
I assume that the cast does not change the binary representation and only that matters when it is written to the database? (Because of that, maybe it would be clearer to do the writing directly here?)

(This applies to all `ToSql`/`FromSql` impls below.)

diesel/src/mysql/types/mod.rs
```diff
@@ -8,36 +8,88 @@ use std::io::Write;
...
 }
 
-impl FromSql<sql_types::Bool, Mysql> for bool {
+impl FromSql<Bool, Mysql> for bool {
     fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
         Ok(not_none!(bytes).iter().any(|x| *x != 0))
     }
 }
 
+impl<ST> HasSqlType<Unsigned<ST>> for Mysql
```
weiznich: I think this impl is to broad. This makes `Mysql: HasSqlType<Unsigend<Text>>` a valid constrain. 

(So maybe better allow only a manual written subset of `Unsigned` types?)

diesel/src/mysql/types/mod.rs
sgrif: No, this does not invoke any UB, nor does it change the representation in memory.

diesel/src/mysql/types/mod.rs
sgrif: As I mentioned in the original PR, this impl doesn't enable anything by itself, and I don't think we need to make it more specific.

PR #1619, opened 2018-04-05
Implement boxed update statements

### LINKED ISSUE #1128: Feature request: Boxed where clauses for delete statements
A query that has been created dynamically with `into_boxed()` can't be passed to `diesel::delete`. Is there an alternative for dynamically adding filters to deletes or is there no workaround for this feature that looks like it's missing?

sgrif:
Just like delete statements, the only thing we actually want to box here
is the where clause. I considered boxing the values clause, but I
couldn't come up with any use case to do so that isn't already handled
by `Option`.

I felt silly re-implementing every method on the statement just to box
the where clause *again* so I factored out that handling, and got rid of
the explicit `BoxedDeleteStatement`.

Fixes #1128.


diesel/src/query_builder/where_clause.rs
```diff
@@ -126,3 +126,67 @@ where
     Expr: AppearsOnTable<QS>,
 {
 }
+
+#[allow(missing_debug_implementations)] // We can't...
```
weiznich: We could do something like that:
```
impl<'a, DB> Debug for BoxedWhereClause<'a, DB> where DB: Backend, DB::QueryBuilder: Default, {
    fn fmt(&self, fmt: Formatter) -> io::Result<…> {
        fmt.debug_tuple("BoxedWhereClause").field(if let Where(q) = self {debug_query(&q).to_string()} else {"None"}).finish()
    }
}
```
But I'm not sure what I should think about that idea…

PR #1623, opened 2018-04-06
Add support for PG tuples

sgrif:
This commit looks like it changes a lot of code. I promise that it
doesn't actually change a lot of code. I generalized the way that our
tuple impls work so that we can add new shit for tuples without making
`impls/tuples` even longer than it already is. That unfortunately churns
7k+ lines of code, but really -- I promise, this commit isn't that big.

This has come up in various contexts enough times for me to take another
look at it. Turns out it wasn't actually that bad to support other than
the fact that we cannot -- nor will we ever be -- able to support a
`ToSql` impl for PG tuple types. I've managed to emulate it with a
decent `AsExpression` impl.

This shouldn't affect most people, as you very rarely care about whether
`ToSql` is *actually* implemented. The only time you would care are if
you're using `sql_query`, which I wrote a long bit in the docs about
*exactly* what we generate, so folks can do the same. The other time you
might care is if you are implementing support for a named composite type
local to your application, in which case you definitely want that impl
to exist. However, I have not done that in this PR, as I need some time
to think over how to do that in a way that doesn't give an impl that
accidentally gets used and generates runtime errors.


diesel/src/test_helpers.rs
```diff
@@ -1,3 +1,5 @@
+extern crate dotenv;
```
weiznich: This change seems to be unrelated to the other changes.

sgrif: Needed to pull out the PG specific functions to use them for testing.

weiznich:
@sgrif Github means this needs a rebase.

Issue #1629, opened 2018-04-10
Introduce a config file for Diesel CLI


sgrif:
The main goal of this file is to allow us to deprecate `infer_schema!`, and provide ways to customize the output of `diesel print-schema`. If there are other things that it'd make sense for us to configure, leave a comment on this issue. This file should not handle things like database URLs, as it's meant to be checked into version control.

# Getting rid of `infer_schema!`

`infer_schema!` predates both `diesel print-schema` and Diesel CLI. The goal was to not force people to write their `table!` declarations by hand, when they represent something we can lookup elsewhere. However, that macro complicates builds since it requires a database connection at compile time. It is very opaque what's happening (especially if something goes wrong), and it's impossible to change the output.

However, we didn't deprecate the macro when `diesel print-schema` was added. I still think it is a useful macro to have at the very beginning of your project. Early on, the schema changes much more frequently, and it's a pain to have to remember to run this second command when you run migrations.

Having a config file will fix this. We can have an option in this file for where the schema file should be located. If this option is present, we will automatically run `print-schema` after any migration command, outputting to that file. Directly running `diesel print-schema` should continue to print to STDOUT, and should not be affected by this option.

# Customizing the output

`diesel print-schema` still has its own problems though. One of the biggest reasons we moved towards this is that it allows you to change the output (just by editing the file). However, this is extremely annoying since you have to redo those changes every time.

I do not think we can ever catch every change that people want to make to this file, but we can catch some of the most common ones. I'd like this file to support:

- Ignoring certain tables
- Adding additional type imports

This list is pretty short. There are other common customizations that are omitted (skipping certain joins, adding additional joins, renaming columns, changing the type of a column). The two customizations I've listed have obvious and clear ways to configure them. Renaming columns for example, I would want to see a concrete proposal for the configuration format.

Finally, as a catch-all, we should also support providing a `.patch` file which gets applied.

# Minor details

The config file should be found by the first of the following which is present:

- A global `--config-file` flag on the CLI
- The `DIESEL_CONFIG_FILE` environment variable
- A file called `diesel.toml` located in the same directory as `Cargo.toml`

For any configuration options which can be configured both by the config file and by command line options, the config file should win.

sgrif:
Unresolved question: if generating schema fails, should we abort the transaction used to run the migrations? I'm leaning towards no, since we don't have hooks to do so, and we don't run migrations in a transaction on MySQL

Eijebong:
[...]

PR #1630, opened 2018-04-10
Add a config file for Diesel CLI, autorun `print-schema`

sgrif:
As discussed in #1629 and elsewhere, we'd like to replace
`infer_schema!` with a config file that re-writes your schema file
automatically when it changes.

This file will eventually be expanded to support most of the CLI args to
`print-schema`, and gain additional options to further customize the
output. However, this is the minimal first step.

I would like to change the default file generated by `diesel setup` to
include `file = "src/schema.rs"`, but before we do that I think we
should change the behavior on tables with no PK to print a warning that
it's being skipped instead of panicking.

### DIFF
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -122,6 +122,17 @@ pub fn build_cli() -> App<'static, 'static> {
                 .help("Render documentation comments for tables and columns"),
         );
 
+    let config_arg = Arg::with_name("CONFIG_FILE")
+        .long("config-file")
+        .help(
+            "The location of the configuration file to use. Falls back to the \
+             `DIESEL_CONFIG_FILE` environment variable if unspecified. Defaults \
+             to `diesel.toml` in your project root. See \
+             diesel.rs/guides/configuring-diesel-cli for documentation on this file.",
+        )
+        .global(true)
+        .takes_value(true);
+
     App::new("diesel")
         .version(env!("CARGO_PKG_VERSION"))
         .setting(AppSettings::VersionlessSubcommands)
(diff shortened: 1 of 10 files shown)


diesel_cli/src/cli.rs
```diff
@@ -122,13 +122,25 @@ pub fn build_cli() -> App<'static, 'static> {
                 .help("Render documentation comments for tables and columns"),
         );
 
+    let config_arg = Arg::with_name("CONFIG_FILE")
+        .long("config-file")
+        .help(
+            "The location of the configuration file to use. Falls back to the \
+             `DIESEL_CONFIG_FILE` environment variable if unspecified. Defaults \
+             to `diesel.toml` in your project root. See \
+             diesel.rs/guides/configuring-diesel-cli for documentation on this file.",
```
weiznich: Is there a https:// missing?

diesel_cli/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+impl Config {
+    pub fn file_path(matches: &ArgMatches) -> PathBuf {
+        matches
+            .value_of("CONFIG_FILE")
+            .map(PathBuf::from)
+            .or_else(|| env::var_os("DIESEL_CONFIG_FILE").map(PathBuf::from))
+            .unwrap_or_else(|| {
+                find_project_root()
+                    .unwrap_or_else(handle_error)
+                    .join("diesel.toml")
```
notryanb: Does this create `diesel.toml` in the project root if it doesn't already exist?

sgrif: No. https://github.com/diesel-rs/diesel/pull/1630/files/ae22270600aeddeb8b95e6b88f48060bd514ab9a#diff-15619a8ba441d58f7250a6aa4a4b5273R200 does.

PR #1631, opened 2018-04-10
Deprecate the `xXX-column-tables` features

sgrif:
The leading x had to get added at the last second due to a bug in
crates.io. That bug has been fixed, so we can remove it. Unfortunately,
we are a stable library comitted to semver, so these features have to
stick around until 2.0... Either way, we can remove it from our docs and
deprecate it.

(Also sorry everyone that my fix for the bug broke the ability to publish new
crates for a day or so, and thanks @ashleygwilliams for helping me deal
with the fallout from that)


sgrif:
@diesel-rs/core Unless anyone objects, I'm planning on cherry-picking this against v1.2.0 and releasing as 1.2.1 once it's approved

PR #1635, opened 2018-04-12
Allow a `.patch` file to be provided to `print-schema`

sgrif:
Ultimately we can't have a config flag for every possible change someone
might want to make to the output. As a catch-all, we can allow a
`.patch` file to be provided which gets applied at the very end. To do
this, we need to shell out to GNU `patch`.

Unfortunately, `patch` does not provide any way to work on stdin rather
than a file, so we have to persist everything to a tempfile
(specifically a named one). This sucks, but I'm not sure there's any
other way for us to go about doing this.

One thing we need to set up a story for is what happens when a merge
conflict occurs, but we can deal with that later.


diesel_cli/src/print_schema.rs
```diff
@@ -65,6 +77,22 @@ pub fn output_schema<W: io::Write>(
     } else {
         write!(out, "{}", definitions)?;
     }
+
+    if let Some(ref patch_file) = config.patch_file {
+        let output = Command::new("patch")
+            .arg(out_path)
+            .arg(patch_file)
+            .output()?;
```
killercup: This looks like a place where we could special-case the `unknown command: patch` case and tell people they need to install it

sgrif: I don't know how we could do that other than checking with `which`. But then we have the same problem

sgrif:
Random note (I should probably open an issue for this), I think we need an explicit merge conflict flow, and probably also a CLI subcommand which regenerates the patch file

sgrif:
Another unresolved question which I will perhaps figure out by tinkering with this on crates.io -- Do we want to have the patch file not existing be an error? I think maybe yes if we expect to generate this from a Diesel subcommand.

At least for crates.io, where we have the modified schema already committed, and want to generate the patch file (vs the flow for new apps, where they'll have the unmodified schema committed, and then edit it), it was annoying to generate the patch file... So probably?

PR #1640, opened 2018-04-13
Feature: expose raw sqlite connection

dbkaplun:
Related: #150, #1103, #1492

This is a simple change that allows diesel applications to hook directly into sqlite's API. This PR exposes diesel's version for libsqlite3_sys bindings, and an `SqliteConnection::get_raw_conn` method. In particular, this allows query tracing/profiling, as well as custom functions via SQLite's API (see also #1103).

Thanks for considering!


sgrif:
Thank you for the pull request, but we've rejected similar features in the past, as we're not interested in exposing the concrete implementations of our connections as part of our public API.

PR #1644, opened 2018-04-16
Add a `nullable` methods for `SelectStatement` and `BoxedSelectStatement`

weiznich:
This allows to use a not nullable query in places that only accept
nullable expressions.

A notable use case is passing a subquery that selects a not nullable
primary key column as `IN`/`eq_any` filter to a nullable column.

cc #1547 and #1558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added support for PG tuples. See [`sql_types::Record`][record-1-3-0] for details.
 
+* Add a `nullable` method on `SelectStatement` and `BoxedSelectStatement` that allows to
+  use those queries in places that except only nullable types
+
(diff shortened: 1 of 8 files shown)


sgrif:
Is there a reason this can't just call `.nullable` on the select clause?

sgrif:
Or return `Nullable<Subselect<Self>>`?

weiznich:
> Is there a reason this can't just call `.nullable` on the select clause? 

Adding a `nullable` method will require to return a new type. In case of `SelectClause<T>` this seems to be easy (`SelectClause<Nullable<T>`), but for `DefaultSelectClause` this will require some sort of wrapper type. If we add a wrapper type I think it is easier to just use this type everywhere.

> Or return `Nullable<Subselect<Self>>`?

`Nullable<_>` works only with `Expression` not with `SelectableExperession`  as far as I see. Also using `Nullable<_>` on the `Subselect` seems to be pretty similar to #1547 .

sgrif:
>  but for DefaultSelectClause this will require some sort of wrapper type.

I don't think so. We should just reify the type when `.nullable` is called.

Issue #1622, opened 2018-04-06
`sql` is in `dsl` but `select` isn't and I don't know why


sgrif:
It's `diesel::select` but `diesel::dsl::sql`. For consistency, we should re-export everything from `query_builder::functions` through `dsl` instead of just at the crate root.

tgturner:
I would love to take a look at this one!

PR #1649, opened 2018-04-19
Re-export query_builder::functions through dsl

tgturner:
This re-exports all `query_builder::functions` through `dsl`.

I am new to rust and the the diesel project, but hope this helps some! Please let me know if there are incorrect patterns here of if there is anything I need to change.

Addresses #1622

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -189,6 +189,10 @@ pub mod dsl {
 
     #[doc(inline)]
     pub use expression::dsl::*;
+
+    #[doc(inline)]
+    pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
+                                       select, sql_query, update};
 }
 
 pub mod helper_types {


diesel_tests/tests/internal_details.rs
```diff
@@ -1,6 +1,6 @@
 use schema::*;
 use diesel::sql_types::*;
-use diesel::dsl::sql;
+use diesel::dsl::{sql, select};
```
weiznich: Why do we import select here and then use a qualified call to `insert_into` below.
(In general I would vote for having more imports and better remove the qualified calls in our test suite. This means instead of doing `select(something)` → `dsl::select(something)` better do one `use diesel::dsl::select` on the top of the file. This applies for all changes in diesel_test and diesel_compile_test)

tgturner:
@sgrif @weiznich 

Thanks for the feedback! I thought it seemed like on heck of a change but figured it was best to just get something out there.

So from what I am hearing, the only change that needs to be made is to update lib.rs to look like this?

``` rust
pub mod dsl {
    //! Includes various helper types and bare functions which are named too
    //! generically to be included in prelude, but are often used when using Diesel.

    #[doc(inline)]
    pub use helper_types::*;

    #[doc(inline)]
    pub use expression::dsl::*;

    #[doc(inline)]
    pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
                                       select, sql_query, update};
}
```

Does it make sense to add new tests for these methods somewhere?

weiznich:
> So from what I am hearing, the only change that needs to be made is to update lib.rs to look like this?

:+1: 

> Does it make sense to add new tests for these methods somewhere?

Maybe change one or two tests to the newly added methods. Just in case somebody tries to remove them for whatever reason.

tgturner:
Will do! Thanks for the guidance!

tgturner:
@weiznich I made the requested changes and squashed the commits (no one needs to see that giant commit I made). Let me know if there are any other changes I can make! 

I also updated some tests to use the newly exported methods.

sgrif:
None of our tests should have needed to change for this

Issue #1646, opened 2018-04-17
Rollback transactions on panic


Diggsey:
Related to https://github.com/sfackler/r2d2/issues/31

Transactions should be rolled back if the inner function panics. The rollback code should take care not to panic in this case, to avoid double panics. Failure to do this can result in connections being returned to the pool with open transactions, which means database locks are held indefinitely, and future uses of the connection will cause further panics.

sgrif:
I disagree that this should be handled in Diesel. Transactions that do not explicitly commit will be rolled back when the connection is dropped. This is a problem *specific* to connection pooling, and should be handled there.

Diggsey:
I strongly disagree with this decision: the transaction function creates a scope for the transaction, and it is fully expected that the transaction not be open after the end of that scope, regardless of how the scope is left. That's not a problem that's specific to connection pooling.

PR #1652, opened 2018-04-20
Roll back transactions on panic and report connection status

Diggsey:
Rolling back transactions on panic is the correct behaviour regardless of whatever else is going on in the system, and is not specific to connection pooling: panics may be unrelated to diesel, in which case diesel should continue to function normally.

See also: scoped thread local storage, which has a similar API, and restores the previous value even in the event that the callback panics.

This is not within the remit of diesel *as an ORM*, but it is within the remit of diesel as a whole because it provides back-end specific implementations of the `Connection` trait.

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -11,7 +11,7 @@ use query_builder::{AsQuery, QueryFragment, QueryId};
 use result::*;
 use sql_types::HasSqlType;
 
-pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager};
+pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager, ScopedTransaction};
 #[doc(hidden)]
 pub use self::statement_cache::{MaybeCached, StatementCache, StatementCacheKey};
 
@@ -95,13 +95,15 @@ pub trait Connection: SimpleConnection + Sized + Send {
     {
         let transaction_manager = self.transaction_manager();
         try!(transaction_manager.begin_transaction(self));
+
+        let tx = ScopedTransaction::new(transaction_manager, self);
         match f() {
             Ok(value) => {
-                try!(transaction_manager.commit_transaction(self));
+                try!(tx.commit());
                 Ok(value)
             }
             Err(e) => {
-                try!(transaction_manager.rollback_transaction(self));
+                try!(tx.rollback());
                 Err(e)
             }
         }
@@ -186,4 +188,11 @@ pub trait Connection: SimpleConnection + Sized + Send {
 
     #[doc(hidden)]
     fn transaction_manager(&self) -> &Self::TransactionManager;
+
+    /// Returns `false` if the connection to the database has been lost.
+    /// Implementations are not required to recognise a "lost connection"
+    /// state, in which case this function will always return `true`.
+    fn is_connected(&self) -> bool {
+        true
+    }
 }
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -129,3 +129,39 @@ where
         self.transaction_depth.get() as u32
     }
 }
+
+/// Manages the state of a transaction, ensuring it is rolled back in the event of a panic
+#[allow(missing_debug_implementations)]
+pub struct ScopedTransaction<'a, Conn: Connection + 'a, T: TransactionManager<Conn> +'a> {
+    tx_manager: &'a T,
+    conn: &'a Conn,
+    closed: bool
+}
+
+impl <'a, Conn: Connection + 'a, T: TransactionManager<Conn> + 'a> ScopedTransaction<'a, Conn, T> {
+    /// Create a new scoped transaction. The transaction should have already been opened.
+    pub fn new(tx_manager: &'a T, conn: &'a Conn) -> Self {
+        ScopedTransaction { tx_manager, conn, closed: false }
+    }
+    /// Commit the transaction
+    pub fn commit(mut self) -> QueryResult<()> {
+        self.closed = true;
+        self.tx_manager.commit_transaction(self.conn)
+    }
+    /// Rollback the transaction
+    pub fn rollback(mut self) -> QueryResult<()> {
+        self.closed = true;
+        self.tx_manager.rollback_transaction(self.conn)
+    }
+}
+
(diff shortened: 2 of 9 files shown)


sgrif:
Panics are not exceptions. You should not be catching them in general. There is no reason that normal usage would result in a database connection continuing to be open when a panic occurs, unless you are using connection pooling. I have laid this out in #1646.

Your argument is like saying that `Vec` should have special behavior on panics because it can get left in an inconsistent state when using `Mutex`. Exception safety is the responsibility of the types that cause values to outlive an exception, not the types that have internal invariants.

This PR is making the exact same argument made there. Unless there is something new to bring to the discussion, continuing to go in circles is not productive. Additionally, continuing to re-open an issue after it's closed is not acceptable behavior.

Diggsey:
> Your argument is like saying that Vec should have special behavior on panics because it can get left in an inconsistent state when using Mutex. Exception safety is the responsibility of the types that cause values to outlive an exception, not the types that have internal invariants.

No, a `Vec` does not manage internally mutable state. The appropriate comparison would be to a `Mutex`, which *does* unlock the mutex (and poisons it) in the event of a panic.

If you look in the standard libary, or on crates.io, you will find that types with internally mutable state, *do* have special behaviour regarding panics. This includes `RefCell`, `Mutex`, `Rwlock`.

> Additionally, continuing to re-open an issue after it's closed is not acceptable behavior.

This is a PR, not an issue.

Diggsey:
> Panics are not exceptions. You should not be catching them in general.

It's considered bad practice to use panics for fine-grained exception handling, however catching panics is still completely legitimate, hence the stabilisation of `catch_unwind`.

> There is no reason that normal usage would result in a database connection continuing to be open when a panic occurs, unless you are using connection pooling. I have laid this out in #1646.

You've asserted it without any basis: you can't know there's no other reason, there could be any number of reasons why that might be desirable: connection pooling is just the example we've encountered.

For example, here's another issue: even if the calling code does drop the connection after a panic, it is not required to do so promptly. Destructors may block: many web frameworks wait for the server to finish in a destructor, scoped thread locals "join" on their child threads before returning, even in the event of a panic. In all these cases, the transaction should not be held open during that time.

> Exception safety is the responsibility of the types that cause values to outlive an exception, not the types that have internal invariants.

This is not about exception safety, it's about clean up. If a connection was smart enough to "poison" (disconnect) itself whenever there was a panic from within the connection implementation, then that would be an example of exception safety, and that would be how you could implement `UnwindSafe` for a connection.

However, even for types that are not `UnwindSafe`, that doesn't mean they can just leak resources when there's a panic. `Vec` still frees its memory when dropped, `RefCell`s are still unlocked, *even though they are not `UnwindSafe`*, etc. etc.

sgrif:
Destructors being run is very different from general exception safety. The entire reason that `Mutex` poisons at all is that types in Rust are allowed to have invariants which get potentially violated in Rust. Of course, any type is required to ensure that destructors are always able to safely run (e.g. `Vec` has to make sure it doesn't try to `Drop` an element which is actually uninitialized memory, or free the wrong amount).

Every single one of our built-in connection types is `!UnwindSafe`. They should be assumed to be invalid if a panic has occurred, period. Transactions aside, we internally use other types which are not `!UnwindSafe`. Making these types `UnwindSafe` is not something I'm willing to commit to. I'm going to quote the docs for `UnwindSafe` to make the rest of my point:

> Types such as &mut T and &RefCell<T> are examples which are not unwind safe. The general idea is that any mutable state which can be shared across catch_unwind is not unwind safe by default. This is because it is very easy to witness a broken invariant outside of catch_unwind as the data is simply accessed as usual.

> Types like &Mutex<T>, however, are unwind safe because they implement poisoning by default. They still allow witnessing a broken invariant, but they already provide their own "speed bumps" to do so.

Connection pooling is akin to `Mutex` here. It is the thing responsible for making a type potentially appear after a panic occurs. Similar to `Mutex`, it is that types responsibility to implement poisoning to ensure that no broken invariants can be witnessed if that occurs.

Diggsey:
> Destructors being run is very different from general exception safety.

That's my point: this issue is about destructors running, not exception safety: an open transaction is a resource, no different than an open file handle, and that resource should be dropped when it goes out of scope. It's a completely separate issue from making the connection itself unwind-safe.

This API from the standard library has an *identical* contract to the `transaction` method:
https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.replace_with

As you can see, it unlocks the `RefCell` even in the event of a panic. Please give one reason why this should have different semantics from those implemented by diesel?

sgrif:
When a connection is dropped, the transaction is rolled back. This is the behavior of every database server out there.

sgrif:
> This API from the standard library has an identical contract to the transaction method:
https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.replace_with

> As you can see, it unlocks the RefCell even in the event of a panic.

I'm not sure what you mean by "contract". You're referring to undocumented behavior, which could incidentally change in the future. `RefCell` is `!UnwindSafe`. It should not be used across panic boundaries, period. Whether broken invariants happen to get observed in one case or another is completely irrelevant, and dangerous to assume.

For the same reason, Diesel's connections are `!UnwindSafe`, period. We use types which are not unwind safe. Changing how transactions behave does not make us `UnwindSafe`. You should not be using a diesel connection across panic boundaries.

Diggsey:
> When a connection is dropped, the transaction is rolled back. This is the behavior of every database server out there.

And the behaviour of every other database library or ORM providing a *scoped transaction* is to rollback the transaction in the event of a panic/unchecked exception/etc. Diesel is the odd one out here. You can look at java's `@Transactional` decorator, sqlalchemy's or django ORM's context managers for transaction, ruby's ActiveRecord, I've yet to find one that didn't do this.

> I'm not sure what you mean by "contract". You're referring to undocumented behavior, which could incidentally change in the future.

The fact that it doesn't "leak" a ref-guard is a pretty important part of `RefCell`'s API. I doubt if you speak to any of the rust core team that they will agree with you on this.

> Changing how transactions behave does not make us UnwindSafe.

I never said it would. In fact I specifically pointed out previously what it would take for that to be the case, and am not currently advocating for that.

> You should not be using a diesel connection across panic boundaries.

I'm not, and yet this is still a problem. A transaction being held open any longer than the scope of the `transaction` method call is a bug. Even if I don't use it again, and do not use a connection pool, the fact that the transaction is still open while destructors are running for types used *outside* that transaction can result in interference with other connections to the database.

PR #1653, opened 2018-04-21
Rename a few things related to barrel support

sgrif:
"type" is an overloaded term. "format" makes more sense with regards to
migration generation. I'm also not a fan of short form args in general,
and `-f` implies force instead of format, so I just removed the short
form version. Finally, I renamed the "rust" part of both the argument
and feature to "barrel", both to make it clear what library is being
used, and to leave room for other rust migration libraries to appear in
the future.

Unfortunately, we couldn't just name the feature "barrel", since Cargo
doesn't allow dependencies and features to share a name. We really
should submit an RFC to fix this. In the mean time, the feature is now
`barrel-migrations`.


killercup:
cc @spacekookie

PR #1655, opened 2018-04-22
Implement `FromSql`/`ToSql` for `Vec<u8>` for Postgres `json`/`jsonb`

brandur:
Implement `FromSql`/`ToSql` for `Vec<u8>` for Postgres `json`/`jsonb`

This patch implements `FromSql` and `ToSql` for `Vec<u8>` for use with
the Postgres `json` and `jsonb` data types. Currently, these traits are
only implemented for `serde_json::Value`.

Here's a use case: while being able to get a `serde_json::Value` back is
tremendously powerful, I want to be able to store one of a number of
different classes of JSON blob into a single field in a Postgres table.
When I retrieve one of them, I'll switch off the value in a separate
column to figure out how I want to handle them. The database column
should still be a `jsonb` (as opposed to a byte array or something else)
because that lets me introspect and query on JSON from Postgres.

Philosophically speaking, I think this is very much a case that the
addition of `json`/`jsonb` to Postgres was designed to handle. If I knew
the structure of my data in advance and expected it to be stable, I'd
have have built it into my schema rather than add a field with flexible
keys/values. Extracting one of these values as free form data allows me
to deserialize it into one of many custom `serde`-compatible type.

I don't know Diesel's internals very well, so any advice on this
implementation would be greatly appreciated. I few questions in
particular came to mind:

1. Is this the right thing to do? (i.e., Am I missing some other obvious
   alternative?)
2. Should I also have an implementation for `[u8]`? I tried doing that,
   but had some trouble with the tests.
3. Is the naming/order of my tests okay? I opted to add the prefix
   `byte_slice` to each like `byte_slice_json_to_sql`, but I could also
   see this being a suffix instead (`json_to_sql_byte_slice`).

Thank you!

### DIFF
--- a/diesel/src/pg/types/json.rs
+++ b/diesel/src/pg/types/json.rs
@@ -28,9 +28,24 @@ impl FromSql<sql_types::Json, Pg> for serde_json::Value {
     }
 }
 
+impl FromSql<sql_types::Json, Pg> for Vec<u8> {
+    fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
+        let bytes = not_none!(bytes);
+        Ok(bytes.to_vec())
+    }
+}
+
 impl ToSql<sql_types::Json, Pg> for serde_json::Value {
     fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
         serde_json::to_writer(out, self)
+            .map(|_| IsNull::No)
+            .map_err(Into::into)
+    }
+}
+
+impl ToSql<sql_types::Json, Pg> for Vec<u8> {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
+        out.write_all(self.as_slice())
             .map(|_| IsNull::No)
             .map_err(Into::into)
     }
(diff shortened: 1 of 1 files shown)


weiznich:
In general your policy for the implementation of  `FromSql` / `ToSql` for certain types in diesel is something like the three following questions:

1. Is this a commonly used type. (Yes because there is already support for `Jsonb` in diesel.
2. Could we provide a `FromSql` and a `ToSql` implementation for that type. (In this case Yes)
3. Could the conversion go somehow wrong because the sql type or the rust type could hold values that are not representable in the other type. (In this case the conversion from `Vec<u8>` to `Jsonb` could go wrong, because obviously not every arbitrary byte sequence is a json string)

Only if we could answer those three questions with:

* It is a commonly used type
* We could provide both conversions 
* Nothing could go wrong

we add support for those types. 

Because the third question is not answered positively I do not think we want to add support for this conversion (Maybe the members of @diesel-rs/core have a different opinion here?)

> Is this the right thing to do? (i.e., Am I missing some other obvious
alternative?)

As far as I understood your use case you want to load values into some struct that implements `Deserialize` or you want to store values from some struct that implements `Serialize` in a jsonb column, right? 
I see three alternatives here:

1. Why not use [`serde_json::from_value`](https://docs.serde.rs/serde_json/fn.from_value.html) / [`serde_json::to_value`](https://docs.serde.rs/serde_json/fn.to_value.html) for construction and destruction of the value loaded from/stored in the database?
2. If variant 1 is to noise because you have many such operations in your database, whats about adding support for a custom type (for diesel) to your application. Something like `JsonValue<T>` and then implement for `FromSql` and `ToSql` for this type for all `T` that implements `Serialize` / `Deserialize`. After that you could just use  `JsonValue<MyJsonType>` at every location that currently supports `serde_json::Value`.
3. If you have any performance concerns, you could implement `FromSql` / `ToSql` for each of your json types and use them directly.

sgrif:
:-1: from me. Semantically, `Vec<u8>` is not a type which represents json

brandur:
Thanks for the detailed response @weiznich. I think as you suggested, `serde_json::from_value` / `serde_json::to_value` will indeed be sufficient for my use case here.

PR #1663, opened 2018-04-26
Support generic and aggregate `sql_function!`s

sgrif:
This extends the new syntax of `sql_function!` to allow generics. To
test this, I wanted to use it internally by replacing some of our
"special" function macros with normal `sql_function!`.

To do this for our aggregate functions, I also needed to add a way to
skip the `NonAggregate` impl. Generic functions were easier than
expected, though capturing the type parameters and their bounds was
annoying as all hell since `<>` isn't a token tree...

Right now we don't support multiple bounds on a type either, which
prevents me from moving `max` and `min` over to this. I'm not sure how
I'm going to handle that, since you can't capture a sequence separated
by `+`... Ideally I'd just be able to do `$($param:ident $(:
$bound:ty)*),*` since `Foo + Bar` parses as a valid type. However, Rust
doesn't allow a type token in the bounds position, only `path` and
`lifetime`. I think we'll probably just have to support where clauses,
and say "if you need more than one bound, put it in a where clause".

Dealing with `#[aggregate]` was a bit more annoying, but it's something
we will need to eventually do in order to support `#[sql_name]` anyway.
We just do our typical token munching strategy, going through one meta
item at a time and letting `macro_rules!` branching do the work for us
as we recursively call ourselves.

The main thing that makes this annoying is that we can never capture
anything as a `meta` token, since once you do that it no longer can be
matched against, ever. Luckily for attributes in particular this is not
a huge annoyance, as the attribute itself is surrounded by `[]` making
it a single token tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,22 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 [record-1-3-0]: http://docs.diesel.rs/diesel/pg/types/sql_types/struct.Record.html
 
+* `sql_function!` now supports generic functions. See [the documentation for
+  `sql_function!`][sql-function-1-3-0] for more details.
+
+* `sql_function!` now supports aggregate functions like `sum` and `max`, by
+  annotating them with `#[aggregate]`. This skips the implementation of
+  `NonAggregate` for your function. See [the documentation for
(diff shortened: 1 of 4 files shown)


sgrif:
> Multiple generic parameters
> Where clause

Neither of these are supported right now. They're both a PITA to parse. I have a followup branch which handles them.

PR #1670, opened 2018-05-01
Use `sql_function!` for `count`

sgrif:
This change technically allows more than we did before, but it doesn't
feel like a major problem to me.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -17,6 +17,9 @@
 #[macro_use]
 #[doc(hidden)]
 pub mod ops;
+#[doc(hidden)]
+#[macro_use]
+pub mod functions;
 
 #[doc(hidden)]
 pub mod array_comparison;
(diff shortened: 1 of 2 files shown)


diesel/src/expression/mod.rs
```diff
@@ -47,9 +47,12 @@ pub mod sql_literal;
 pub mod subselect;
 
 #[doc(hidden)]
+#[allow(non_camel_case_types)]
 pub mod dsl {
+    use dsl::SqlTypeOf;
+
     #[doc(inline)]
-    pub use super::count::{count, count_star};
+    pub use super::count::*;
```
weiznich: pub using items with `#[doc(inline)]` does not actually show the documentation in that place (Rustdoc will only render `pub use super::count::*;` here not the actual imported types.

sgrif: I don't think that's correct. http://docs.diesel.rs/diesel/dsl/index.html https://github.com/diesel-rs/diesel/blob/master/diesel/src/lib.rs#L187-L191

weiznich: Hmm, then this only works inside of an crate. (There was [this PR](https://github.com/diesel-rs/diesel/pull/1410) for diesel_migrations)

sgrif: Correct, you cannot inline documentation across crates.

PR #1671, opened 2018-05-01
Ensure CLI still works with no config file

sgrif:
We can handle an empty config file just fine, but I forgot to handle
cases where there's *no* config file.

### DIFF
--- a/diesel_cli/src/config.rs
+++ b/diesel_cli/src/config.rs
@@ -9,7 +9,7 @@ use toml;
 use super::{find_project_root, handle_error};
 use print_schema;
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Default)]
 #[serde(deny_unknown_fields)]
 pub struct Config {
     #[serde(default)]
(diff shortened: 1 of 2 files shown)


diesel_cli/src/config.rs
```diff
@@ -31,9 +31,14 @@ impl Config {
 
     pub fn read(matches: &ArgMatches) -> Result<Self, Box<Error>> {
         let path = Self::file_path(matches);
-        let mut bytes = Vec::new();
-        fs::File::open(path)?.read_to_end(&mut bytes)?;
-        toml::from_slice(&bytes).map_err(Into::into)
+
+        if path.exists() {
+            let mut bytes = Vec::new();
+            fs::File::open(path)?.read_to_end(&mut bytes)?;
```
killercup: If you too are randomly wondering if `Read::read_to_end` is cleverly buffered: [It totally is](https://doc.rust-lang.org/1.25.0/src/std/io/mod.rs.html#357)

diesel_cli/tests/migration_run.rs
```diff
@@ -346,3 +346,29 @@ file = "src/my_schema.rs"
     );
     assert!(p.has_file("src/my_schema.rs"));
 }
+
+#[test]
+fn migrations_can_be_run_with_no_config_file() {
+    let p = project("migration_run").folder("migrations").build();
+    let db = database(&p.database_url());
+
+    p.command("database setup").run();
```
killercup: CI says

> Failed to open connection to postgres://postgres@localhost/diesel_migration_run: BadConnection("FATAL:  database "diesel_migration_run" does not exist

which leads me to ask: Isn't it just "setup" here? Also, `run` doesn't seem to assert the command exited with zero, so it can't catch wrong calls.

sgrif: `diesel setup` will create a config file

killercup: ah that's true, damn

sgrif: I just forgot to change the project name in this test. It's a race condition, not an actual bug.

weiznich:
This is missing a rustfmt run otherwise it looks fine.

PR #1302, opened 2017-11-13
Enable sqlite/bundle for diesel cli

simonhdickson:
Trying to install the CLI on windows can be problematic, when creating a project you can just import the libsqlite3-sys feature but you can't do that when installing the cli tools. With this change you can install on Windows (or other platforms without an easy way of installing sqlite3).

Installation can be achieved using this: `cargo install diesel_cli --no-default-features --features sqlite-bundled`

PR #1672, opened 2018-05-01
Enable sqlite/bundle for diesel cli redux

nickbabcock:
Rebased #1302 onto master

> Installation can be achieved using this: cargo install diesel_cli --no-default-features --features sqlite-bundled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * Added support for specifying `ISOLATION LEVEL`, `DEFERRABLE`, and `READ ONLY`
   on PG transactions. See [`PgConnection::build_transaction`] for details.
 
+* Added `sqlite-bundled` feature to `diesel_cli` to make installing on
+  some platforms easier. 
+
 [`PgConnection::build_transaction`]: http://docs.diesel.rs/diesel/pg/struct.PgConnection.html#method.build_transaction
 
 * Added support for `BEGIN IMMEDIATE` and `BEGIN EXCLUSIVE` on SQLite.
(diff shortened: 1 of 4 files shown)


killercup:
LGTM at first glance, but is blocked by #1674 and CI needs some work to run correctly. Also, speaking of CI, it would make sense to also test this on AppVeyor but its builds look broken, too :(

sgrif:
I don't want to add this to the travis matrix. We don't need to rerun the whole suite for this. See https://github.com/diesel-rs/diesel/pull/1302#issuecomment-358111591. A manual matrix addition which runs *only* the CLI tests w/ bundled enabled once, on stable is enough.

nickbabcock:
Rebased onto master, so tests pass.

I've also updated the travis file so that running the bundle isn't a one off that is ran on nightly and allowed to fail. It also contains the SQLITE_BUNDLED env as "sqlite-bundled" so that it can be used in features and allows for more tests on the bundled functionality. The downside of this is that it does add two full fledged builds in .travis so now mysql, psql, sqlite, and sqlite bundled are ran on stable and beta.

I can understand how this would not be desirable, so I can switch it back to the one off (but run it on stable instead of nightly)

CHANGELOG.md
```diff
@@ -86,6 +86,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * Added support for specifying `ISOLATION LEVEL`, `DEFERRABLE`, and `READ ONLY`
   on PG transactions. See [`PgConnection::build_transaction`] for details.
 
+* Added `sqlite-bundled` feature to `diesel_cli` to make installing on
```
sgrif: This needs to be placed in the `Unreleased` section.

nickbabcock:
Ah, I've been ninja'd! 😄

PR #1673, opened 2018-05-02
Add Cluster Database Connections [WIP]

hodgesds:
This PR is a work in progress and before going too far I'd like to get some feedback. The feature that I'm looking to implement is the ability to use a clustered database connection(s). The idea is that it would be configurable so that reads work on replicas and writes get directed to the master. If this is a feature that sounds reasonable to implement I'd be more than willing to continue on this, but I'd like the blessing of the maintainers first. Thanks!

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -11,9 +11,9 @@ use query_builder::{AsQuery, QueryFragment, QueryId};
 use result::*;
 use sql_types::HasSqlType;
 
-pub use self::transaction_manager::{AnsiTransactionManager, TransactionManager};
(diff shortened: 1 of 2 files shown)


weiznich:
In my opinion this is something that should be handled at connection pooling level.
So implementing this would require to rewrite the current pooling architecture. There are other issues in the current pooling system so this is something on our backlog. I'm not sure when the rewrite takes place. (Maybe we should first collect some ideas what the new system should be able to handle.)

hodgesds:
I'm going to open an issue for now to track this as that seems to be the right place for continuing what needs to be discussed.

PR #1674, opened 2018-05-02
Bump all of the nightly things

sgrif:
parking_lot decided to bump their minimum required Rust version in a
patch release, which means that we have to update to a newer nightly.
While this pretty much always means updating our clippy version,
the version of `proc_macro2` we were using stopped working. So we had to
bump syn as well...

It appears that https://github.com/rust-lang/rust/issues/47941 has
gotten worse, so a lot of our error messages have regressed on newer
nightlies.

I also bumped rustfmt because why not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-  - rust: nightly-2018-02-02
+  - rust: nightly-2018-04-19
     env: CLIPPY_AND_COMPILE_TESTS=YESPLEASE
     script:
     - (cd diesel && cargo rustc --no-default-features --features "lint sqlite postgres mysql extras" -- -Zno-trans)
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
     - (cd diesel_migrations && cargo rustc --no-default-features --features "lint dotenv sqlite postgres mysql" -- -Zno-trans)
     - (cd diesel_infer_schema && cargo rustc --no-default-features --features "lint dotenv sqlite postgres mysql" -- -Zno-trans)
     - (cd diesel_compile_tests && travis-cargo test)
-  - rust: nightly-2018-02-02
+  - rust: nightly-2018-03-15
     env: RUSTFMT=YESPLEASE
     script:
     - cargo install rustfmt-nightly --vers 0.3.7 --force
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -117,7 +117,7 @@
 #![cfg_attr(feature = "clippy", plugin(clippy(conf_file = "../../clippy.toml")))]
 #![cfg_attr(feature = "clippy",
             allow(option_map_unwrap_or_else, option_map_unwrap_or, match_same_arms,
-                  type_complexity))]
+                  type_complexity, redundant_field_names))]
 #![cfg_attr(feature = "clippy",
             warn(option_unwrap_used, result_unwrap_used, print_stdout,
                  wrong_pub_self_convention, mut_mut, non_ascii_literal, similar_names,
--- a/diesel/src/query_builder/ast_pass.rs
+++ b/diesel/src/query_builder/ast_pass.rs
@@ -89,6 +89,8 @@ where
     /// done implicitly for references. For structs with lifetimes it must be
     /// done explicitly. This method matches the semantics of what Rust would do
     /// implicitly if you were passing a mutable reference
+    // Clippy is wrong, this cannot be expressed with pointer casting
+    #[cfg_attr(feature = "clippy", allow(transmute_ptr_to_ptr))]
     pub fn reborrow(&mut self) -> AstPass<DB> {
         use self::AstPassInternals::*;
         let internals = match self.internals {
--- a/diesel/src/query_builder/debug_query.rs
+++ b/diesel/src/query_builder/debug_query.rs
@@ -81,6 +81,8 @@ where
     DB: Backend,
     T: QueryFragment<DB>,
 {
+    // Clippy is wrong, this cannot be expressed with pointer casting
+    #[cfg_attr(feature = "clippy", allow(transmute_ptr_to_ptr))]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut list = f.debug_list();
         {
--- a/diesel_compile_tests/tests/ui/insertable_missing_table_or_column.stderr
+++ b/diesel_compile_tests/tests/ui/insertable_missing_table_or_column.stderr
@@ -11,28 +11,30 @@ error[E0433]: failed to resolve. Use of undeclared type or module `posts`
   | ^ Use of undeclared type or module `posts`
 
 error[E0412]: cannot find type `name` in module `users`
-  --> $DIR/insertable_missing_table_or_column.rs:24:5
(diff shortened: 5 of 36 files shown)


diesel/src/pg/types/network_address.rs
```diff
@@ -185,12 +195,38 @@ fn some_v4address_from_sql() {
...
+        ($ty:ty, $net_type:expr) => {
             let mut bytes = Output::test();
-            let test_address = IpNetwork::V6(Ipv6Network::new(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1), 64).unwrap());
+            let test_address = IpNetwork::V6(
+                Ipv6Network::new(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1), 64).unwrap(),
+            );
             ToSql::<$ty, Pg>::to_sql(&test_address, &mut bytes).unwrap();
-            assert_eq!(bytes, vec![PGSQL_AF_INET6, 64, $net_type, 16, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]);
-        }
+            assert_eq!(
```
Eijebong: Not sure how I feel about this one...

diesel/src/query_builder/debug_query.rs
```diff
@@ -81,6 +81,8 @@ where
     DB: Backend,
     T: QueryFragment<DB>,
 {
+    // Clippy is wrong, this cannot be expressed with pointer casting
```
Eijebong: Is this worth reporting a bug on clippy ?

diesel_compile_tests/tests/ui/insertable_missing_table_or_column.stderr
```diff
@@ -11,28 +11,30 @@ error[E0433]: failed to resolve. Use of undeclared type or module `posts`
   | ^ Use of undeclared type or module `posts`
 
 error[E0412]: cannot find type `name` in module `users`
-  --> $DIR/insertable_missing_table_or_column.rs:24:5
+  --> $DIR/insertable_missing_table_or_column.rs:21:10
    |
-24 |     name: String
-   |     ^^^^ not found in `users`
+21 | #[derive(Insertable)]
```
Eijebong: :(

diesel/src/query_builder/debug_query.rs
sgrif: I did

diesel_derives/src/meta.rs
```diff
@@ -220,7 +222,7 @@ impl MetaItem {
 }
 
 #[cfg_attr(rustfmt, rustfmt_skip)] // https://github.com/rust-lang-nursery/rustfmt/issues/2392
```
weiznich: This is fixed now :wink:

diesel_derives/src/util.rs
```diff
@@ -71,7 +71,7 @@ pub fn ty_for_foreign_derive(item: &DeriveInput, flags: &MetaItem) -> Result<Typ
 }
 
 pub fn fix_span(maybe_bad_span: Span, fallback: Span) -> Span {
-    let bad_span_debug = "Span(Span { lo: BytePos(0), hi: BytePos(0), ctxt: #0 })";
+    let bad_span_debug = "#0 bytes(0..0)";
```
weiznich: I think doing this is quite risky, because this may break at a minor version bump of `proc_macro2` 
Maybe there is some way to just construct a bad span itself? (This would at least make this code more robust against possible future changes of the debug output)

.travis.yml
```diff
@@ -41,7 +41,7 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-  - rust: nightly-2018-02-02
+  - rust: nightly-2018-04-30
```
killercup: This is so last month already ;)

diesel/src/expression/operators.rs
```diff
@@ -258,7 +246,7 @@ macro_rules! diesel_postfix_operator {
         diesel_postfix_operator!($name, $operator, $crate::sql_types::Bool);
     };
 
-    ($name:ident, $operator:expr, backend: $backend:ty) => {
```
killercup: This is a weird change. Something funky we should tell rustfmt people about?

diesel/src/lib.rs
```diff
@@ -93,36 +93,61 @@
...
+    )
+)]
+#![cfg_attr(
+    feature = "clippy",
+    warn(
+        option_unwrap_used, result_unwrap_used, print_stdout, wrong_pub_self_convention, mut_mut,
+        non_ascii_literal, similar_names, unicode_not_nfc, enum_glob_use, if_not_else,
+        items_after_statements, used_underscore_binding
+    )
+)]
```
killercup: It feels wrong for me to like this syntax

diesel/src/query_builder/ast_pass.rs
```diff
@@ -89,6 +89,8 @@ where
     /// done implicitly for references. For structs with lifetimes it must be
     /// done explicitly. This method matches the semantics of what Rust would do
     /// implicitly if you were passing a mutable reference
+    // Clippy is wrong, this cannot be expressed with pointer casting
+    #[cfg_attr(feature = "clippy", allow(transmute_ptr_to_ptr))]
```
killercup: can you put this on the unsafe block with the transmute?

diesel_infer_schema/infer_schema_internals/src/data_structures.rs
```diff
@@ -68,7 +68,7 @@ impl ColumnInformation {
         ColumnInformation {
             column_name: column_name.into(),
             type_name: type_name.into(),
-            nullable: nullable,
+            nullable,
```
killercup: this syntax landed quite a while ago, right? would be weird to bump min required rustc version for something like this

examples/postgres/advanced-blog-cli/src/cli.rs
```diff
@@ -1,7 +1,7 @@
 #[derive(StructOpt)]
 #[structopt(
     name = "blog",
-    after_help = "You can also run `blog SUBCOMMAND -h` to get more information about that subcommand.",
+    after_help = "You can also run `blog SUBCOMMAND -h` to get more information about that subcommand."
```
killercup: huh, no trailing comma here seems like a bug in rustfmt

diesel/src/macros/mod.rs
```diff
@@ -1050,6 +1053,7 @@ macro_rules! joinable {
 
 #[macro_export]
 #[doc(hidden)]
+#[cfg_attr(rustfmt, rustfmt_skip)] // It turns this into soup
```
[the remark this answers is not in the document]
sgrif: I'm planning to, just haven't gotten around to it

diesel_derives/src/util.rs
sgrif: This has less to do with `proc_macro2` and more to do with Rust. I agree that it's brittle and will break again in the future. That said, the only reason we do this at all is to work around a bug in Rust itself, and if this code breaks it just means that a handful of error cases point at the macro invocation instead.

I couldn't find any way to directly construct a bad span, no.

diesel_infer_schema/infer_schema_internals/src/data_structures.rs
sgrif: Yes, it landed in the same version as macros 1.1. We already use it elsewhere.

PR #1675, opened 2018-05-04
Update supported composite primary key columns count

ivanovaleksey:


diesel/src/macros/mod.rs
```diff
@@ -119,7 +119,7 @@ macro_rules! __diesel_column {
 /// ```
 ///
 /// You may also specify a primary key if it's called something other than `id`.
-/// Tables with no primary key, or composite primary containing more than 3
+/// Tables with no primary key, or composite primary key containing more than 5
```
weiznich: If we change things here, we should also fix the sentence grammatically.
I think "Tables with no primary key, or **a** composite primary key containing more than 5 columns are not supported."

ivanovaleksey: I am not sure if we should make changes here at all but it seemed to me that a word "key" was missing.
I am not so good in English grammar, so I can't decide here :)

ivanovaleksey:
@weiznich I have updated grammar :)

PR #1676, opened 2018-05-04
Adding a way to provide a backend specific create table statement for

weiznich:
the migration table

Different databases uses different sql dialects. While the current
table statement for the migration table works on all supported
backends it is a show stopper while implementing a new backend for
supporting a oracle database.

Technically this commit is a breaking change. This may affect only
custom backend implementation. We are not aware of any such
implementation. (Because of that I've requested a review from the other @diesel-rs/core members)

cc @pgab

### DIFF
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -10,6 +10,7 @@ use super::backend::Mysql;
 use super::bind_collector::MysqlBindCollector;
 use connection::*;
 use deserialize::{Queryable, QueryableByName};
+use migration::MigrationConnection;
 use query_builder::*;
 use result::*;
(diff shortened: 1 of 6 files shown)


diesel/src/mysql/connection/mod.rs
```diff
@@ -10,6 +10,7 @@ use super::backend::Mysql;
 use super::bind_collector::MysqlBindCollector;
 use connection::*;
 use deserialize::{Queryable, QueryableByName};
+use migration::MigrationConnection;
```
JohnTitor: Conflict occurred here

weiznich:
Something like this is still required for our oracle backend, but we need to disuse the api in the @diesel-rs/contributors team.

PR #1678, opened 2018-05-04
Simplify the internal trait structure of locking select statements

sgrif:
The public API for these features remains the same, but we don't need
this many traits to support it. Similar to joins, we can represent this
with a single trait that has many methods accessing it.

Since `ForUpdateDsl` is part of the public API, we cannot directly
remove it.

### DIFF
--- a/diesel/src/query_builder/select_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_statement/dsl_impls.rs
@@ -288,124 +288,22 @@ where
     }
 }
 
-impl<F, S, W, O, L, Of> ForUpdateDsl for SelectStatement<F, S, NoDistinctClause, W, O, L, Of> {
-    type Output = SelectStatement<
-        F,
-        S,
-        NoDistinctClause,
-        W,
-        O,
-        L,
-        Of,
-        NoGroupByClause,
-        LockingClause<ForUpdate, NoModifier>,
-    >;
-
-    fn for_update(self) -> Self::Output {
-        SelectStatement::new(
-            self.select,
-            self.from,
-            self.distinct,
-            self.where_clause,
-            self.order,
-            self.limit,
-            self.offset,
-            self.group_by,
-            LockingClause::new(ForUpdate, NoModifier),
-        )
-    }
-}
-
-impl<F, S, W, O, L, Of> ForNoKeyUpdateDsl for SelectStatement<F, S, NoDistinctClause, W, O, L, Of> {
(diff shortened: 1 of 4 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -797,13 +800,38 @@ pub trait QueryDsl: Sized {
     /// // Executes `SELECT * FROM users FOR UPDATE`
     /// users.for_update().load(&connection)
     /// ```
+    #[cfg(feature = "with-deprecated")]
+    #[allow(deprecated)]
     fn for_update(self) -> ForUpdate<Self>
```
Diggsey: Can't `ForUpdateDsl` extend, and be implemented for all, `LockingDsl<ForUpdate>` avoiding the need for two definitions of this method?

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -415,17 +313,17 @@ impl<F, S, D, W, O, L, Of, G, LM> SkipLockedDsl
...
             self.group_by,
-            LockingClause::new(self.locking.lock_mode, SkipLocked),
+            LockingClause::new(lock, NoModifier),
         )
     }
 }
 
-impl<F, S, D, W, O, L, Of, G, LM> NoWaitDsl
-    for SelectStatement<F, S, D, W, O, L, Of, G, LockingClause<LM, NoModifier>>
+impl<F, S, D, W, O, L, Of, G, LC, LM, Modifier> ModifyLockDsl<Modifier>
```
Diggsey: Was it intentional to allow setting the modifier multiple times here?

eg. `for_update().skip_locked().no_wait()`

diesel/src/query_dsl/mod.rs
Diggsey: Hm, I suppose adding a super trait would be a breaking change. If only trait aliases were stable...

diesel/src/query_builder/select_statement/dsl_impls.rs
sgrif: Yes. We generally allow overriding clauses later. This should be no different. I'm planning on a followup PR to allow overriding the rest of the lock clause as well

diesel/src/query_dsl/mod.rs
sgrif: This would break any existing code that implements `ForUpdateDsl`.

PR #1685, opened 2018-05-09
Allow migrations to have attached metadata, avoid running in transaction

sgrif:
On PostgreSQL, certain commands cannot be run inside of a transaction,
such as `CREATE INDEX CONCURRENTLY` and `ALTER TYPE`. This causes
problems when attempting to use Diesel migrations, since we always run
these migrations inside a transaction.

To allow opting out of this, we introduce the concept of migration
metadata. I've opted to keep this incredibly general, rather than adding
a specific `should_run_in_migration` function, as there's been interest
expressed in having this capability to enable alternate migration
runners.

Since we need to retain object safety, we can't actually have any
defined structure for the metadata. The best we can do is either
something that is conceptually `Map<String, String>`, or an enum like
`serde_json::Value` or `toml::Value`. I really don't want to restrict
what can or cannot be in here, so I've opted for the more general
option.

/cc @Diggsey you had shown interest in having this feature.
/cc @spacekookie You should consider if you want to support this, and how it would be represented in barrel. I can try to figure out a good place to publicly re-export the toml metadata struct if you'd like to just use that.

### DIFF
--- a/diesel/src/migration/mod.rs
+++ b/diesel/src/migration/mod.rs
@@ -5,17 +5,26 @@ pub use self::errors::{MigrationError, RunMigrationsError};
 
 use connection::SimpleConnection;
 use std::path::Path;
+use std::borrow::Cow;
 
 /// Represents a migration that interacts with diesel
 pub trait Migration {
     /// Get the migration version
     fn version(&self) -> &str;
+
     /// Apply this migration
     fn run(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
+
     /// Revert this migration
     fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError>;
+
     /// Get the migration file path
     fn file_path(&self) -> Option<&Path> {
+        None
+    }
+
+    /// Get the metadata associated with this migration, if any
+    fn metadata(&self) -> Option<&Metadata> {
         None
     }
 }
@@ -32,8 +41,13 @@ impl Migration for Box<Migration> {
     fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError> {
         (&**self).revert(conn)
     }
+
     fn file_path(&self) -> Option<&Path> {
         (&**self).file_path()
+    }
+
+    fn metadata(&self) -> Option<&Metadata> {
+        (&**self).metadata()
     }
 }
 
@@ -49,7 +63,27 @@ impl<'a> Migration for &'a Migration {
     fn revert(&self, conn: &SimpleConnection) -> Result<(), RunMigrationsError> {
         (&**self).revert(conn)
     }
+
     fn file_path(&self) -> Option<&Path> {
         (&**self).file_path()
     }
+
+    fn metadata(&self) -> Option<&Metadata> {
+        (&**self).metadata()
+    }
+}
+
+/// Represents metadata associated with a migration.
+///
+/// The format of a migration's metadata is dependent on the migration format
+/// being used.
+///
+/// For Diesel's built in SQL file migrations, metadata is stored in a file
+/// called `metadata.toml`. Diesel looks for a single key, `run_in_transaction`.
+/// By default, all migrations are run in a transaction on SQLite and
+/// PostgreSQL. This behavior can be disabled for a single migration by setting
+/// this to `false`.
+pub trait Metadata {
+    /// Get the metadata at the given key, if present
+    fn get(&self, key: &str) -> Option<Cow<str>>;
 }
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -76,6 +76,7 @@
 //! ```
 #[macro_use]
 extern crate diesel;
+extern crate toml;
 
 #[cfg(feature = "barrel")]
 extern crate barrel;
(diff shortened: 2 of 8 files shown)


diesel/src/migration/mod.rs
```diff
@@ -49,7 +63,27 @@ impl<'a> Migration for &'a Migration {
...
+/// being used.
+///
+/// For Diesel's built in SQL file migrations, metadata is stored in a file
+/// called `metadata.toml`. Diesel looks for a single key, `run_in_transaction`.
+/// By default, all migrations are run in a transaction on SQLite and
+/// PostgreSQL. This behavior can be disabled for a single migration by setting
+/// this to `false`.
+pub trait Metadata {
+    /// Get the metadata at the given key, if present
+    fn get(&self, key: &str) -> Option<Cow<str>>;
```
Diggsey: I'm concerned that this may be too limited an interface: I wished to store a list of previous revisions as part of the metadata for example, which means they've have to be double-encoded in the toml file.

Perhaps this "get" method could use [erased_serde](https://github.com/dtolnay/erased-serde) to allow it to support structured data? That way I can specify the type when I access the metadata?

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -320,14 +321,33 @@ fn run_migration<Conn>(
...
         }
         try!(migration.run(conn));
         try!(conn.insert_new_migration(migration.version()));
         Ok(())
-    })
+    };
+
+    let run_in_transaction = migration
+        .metadata()
+        .and_then(|m| m.get("run_in_transaction"));
```
Diggsey: The metadata key that diesel understands should probably be exported as constants somewhere.

Diggsey: I meant publicly exported.

diesel/src/migration/mod.rs
Diggsey: Alternatively, to keep things simpler we could just be more opinionated about what format metadata comes in, and have the migration supply a raw TOML/RON/JSON/whatever string.

sgrif: > I'm concerned that this may be too limited an interface: I wished to store a list of previous revisions as part of the metadata for example, which means they've have to be double-encoded in the toml file.

I had the same concern and thought about it for a good while. You absolutely do not need to store it double encoded. You're right that this function will return a TOML string for tables/arrays though, and you'll end up calling `toml::from_str` on the result. There's actually a benefit to doing that, since you can then turn the returned string into a type other than `toml::Value` etc.

> Perhaps this "get" method could use erased_serde to allow it to support structured data? That way I can specify the type when I access the metadata?

I'd prefer not to couple to any single serialization framework here. Especially since that assumes that metadata is serialized in a file, which may not be true for Rust migrations, or migrations which decide to use magic comments, etc.

diesel_migrations/migrations_internals/src/lib.rs
sgrif: Why?

diesel/src/migration/mod.rs
Diggsey: Oh, I see, it's serialising back to TOML before returning if it's not a string 🙁 

What if barrel decides not to use TOML to store its metadata? Then I won't know how to deserialize an array without knowing where a migration came from?

> I'd prefer not to couple to any single serialization framework here.

Serde is the de facto standard for serialization, and diesel already has it as an optional dependency, so this doesn't seem overly risky.

> Especially since that assumes that metadata is serialized in a file, which may not be true for Rust migrations, or migrations which decide to use magic comments, etc.

I don't follow this logic? The only assumption would be that the metadata could be converted to a format which is compatible with serde's object model. For embedding migrations, the internal format would not be exposed, we'd just pick any format which supports the whole of that object model (eg. RON).

Overall I'd rather diesel picked a standardised object model (either serde/RON or JSON) and expected migrations to convert their metadata to that format. Given one of those object models, there's a choice of exposing an API based on serde, or exposing the raw RON or JSON, depending on which you'd rather be coupled to (the latter would allow you to entirely avoid coupling to serde).

diesel_migrations/migrations_internals/src/lib.rs
Diggsey: So that if someone else wants to look at the same metadata they don't have to duplicate the strings, they can import the constant from diesel, and for someone reading the code it's obvious how everything fits together. The constant would also serve as a convenient place to put the documentation.

Also, for someone working on diesel, it's more obvious that the constant is part of the public API (ie. it can't just be changed without breaking stuff) if it is actually part of the public API.

diesel/src/migration/mod.rs
sgrif: > What if barrel decides not to use TOML to store its metadata? Then I won't know how to deserialize an array without knowing where a migration came from?

I was sort of assuming that if you're caring about specific metadata keys, you're also assuming a migration format, but you're right that isn't ideal. I'll need to think about this more. I was about to say we could have `get_map` and `get_array` as separate functions, but there's no type we can really return from `get_array` unless we have an enum. Perhaps we should just have a *very* naive enum representation of `Value(String), Array(Vec<ThisEnum>), Map(HashMap<String, ThisEnum>)`. (This now is making some assumptions about how maps are represented but I think that's probably reasonable.) Either way I do need to give this more thought.

> Serde is the de facto standard for serialization, and diesel already has it as an optional dependency, so this doesn't seem overly risky.

I'm not implying that there's another serialization framework to use instead, just that some types are difficult to represent with Serde, and I'd rather not push it here.

> For embedding migrations, the internal format would not be exposed, we'd just pick any format which supports the whole of that object model (eg. RON).

This is less about embedded migrations, and more about cases where the metadata is entirely computed.

leoank:
Is this pull request going to be merged any time soon? Or there are any other workarounds for running commands such as `CREATE INDEX CONCURRENTLY` and `ALTER TYPE`.

sgrif:
This has pretty much stalled so I'm going to close this. If someone wants to rebase this onto master and push it over the finish line, I think folks are still pretty interested in this general feature (there may need to be more discussion over whether this is the right way to handle it or not)

msrd0:
@sgrif you are right, folks are still interested. I have created #2061, I hope this is not going to stall again.

PR #1687, opened 2018-05-10
Use include/exclude instead of whitelist/blacklist.

dreid:
Retain whitelist/blacklist options (hidden from help) and add a deprecation warning to stderr.

While the origins of the term "black list" may have nothing to do with race language evolves and so we find ourselves in a place where the idiom is interpreted that way.

This is hopefully a relatively uncontroversial change but it is perhaps important to note that is not an isolated one.

- https://github.com/whatwg/html/pull/281
- https://github.com/kickstarter/rack-attack/pull/181

And cargo uses `include`/`exclude` for a similar purpose: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields-optional


sgrif:
I'm good with changing the terminology, but I'm not sure `include` is the right term to use, since it's actually saying "exclude everything not in this list". Perhaps `only` instead? I think I'd like to mention `table` in the option as well.

dreid:
`only-tables` And `except-tables`?

diesel_cli/tests/print_schema/print_schema_only_tables/diesel.toml
```diff
@@ -1,4 +1,4 @@
 [print_schema]
 file = "src/schema.rs"
 with_docs = true
-filter = { blacklist = ["users1"] }
+filter = { only-tables = ["users1"] }
```
sgrif: This is backwards isn't it? Shouldn't this be `except-tables`?

diesel_cli/tests/print_schema/print_schema_except_tables/diesel.toml
```diff
@@ -1,4 +1,4 @@
 [print_schema]
 file = "src/schema.rs"
 with_docs = true
-filter = { whitelist = ["users1"] }
+filter = { except-tables = ["users1"] }
```
sgrif: This is backwards, isn't it? Shouldn't this be `only-tables`?

sgrif: I think git is just confused.

dreid: I might have confused the git rename detection?

diesel_cli/src/cli.rs
```diff
@@ -108,21 +108,39 @@ pub fn build_cli() -> App<'static, 'static> {
                 .index(1)
                 .takes_value(true)
                 .multiple(true)
-                .help("Table names to filter (default whitelist if not empty)"),
+                .help("Table names to filter (default include if not empty)"),
```
sgrif: Should this say `only-tables`? TBH I wonder if we should just remove any "default" behavior (I think we have an issue open saying that the default is broken anyway)

diesel_cli/src/main.rs
```diff
@@ -352,10 +352,14 @@ fn run_infer_schema(matches: &ArgMatches) -> Result<(), Box<Error>> {
         })
         .collect();
 
-    if matches.is_present("whitelist") {
-        config.filter = Filtering::Whitelist(filter)
-    } else if matches.is_present("blacklist") {
-        config.filter = Filtering::Blacklist(filter)
+    if matches.is_present("whitelist") || matches.is_present("blacklist") {
+        eprintln!("Deprecation warning: whitelist/blacklist will be removed in a future version, use only-tables/except-tables instead.");
```
sgrif: Can we give a specific error message for each one? I'd prefer something like this:

```
if matches.is_present("whitelist") {
    eprintln!("The `whitelist` option has been deprecated and renamed to `only-tables`. ");
}

if matches.is_present("blacklist") {
    eprintln!("The `blacklist` option has been deprecated and renamed to `only-tables`. ");
}
```

sgrif:
Can you add a `CHANGELOG` entry for this as well?

diesel_cli/src/cli.rs
dreid: If we remove the default behavior would it make more sense to have `--only-tables` and `--except-tables` take a comma separated list of arguments instead of relying on the positional arguments to the command? (This was confusing when I first encountered it, from glancing at the help I really expected `--blacklist=foo,bar,baz` instead of `--blacklist foo bar baz`)

sgrif:
@dreid FYI I'm hoping to release 1.3 tonight, and I'd like to include this PR. If you're able to address the few remaining comments today that would be awesome.

diesel_cli/src/main.rs
```diff
@@ -353,9 +353,17 @@ fn run_infer_schema(matches: &ArgMatches) -> Result<(), Box<Error>> {
...
 
     if matches.is_present("whitelist") {
-        config.filter = Filtering::Whitelist(filter)
-    } else if matches.is_present("blacklist") {
-        config.filter = Filtering::Blacklist(filter)
+        eprintln!("The `whitelist` option has been deprecated and renamed to `only-tables`.");
+    }
+
+    if matches.is_present("blacklist") {
+        eprintln!("The `whitelist` option has been deprecated and renamed to `except-tables`.");
```
PeterW-LWL: `whitelist` -> `blacklist`

PR #1690, opened 2018-05-12
Don't wrap `sqlite::RawConnection` in an Rc

sgrif:
The reason we originally did the Rc wrapping was to ensure that the
prepared statements were finalized before the connection was closed. If
we don't finalize all prepared statements before closing the connection,
attempting to close the connection will fail, and we will panic.

At the time this code was originally written, the drop order of struct
fields was considered an implementation detail of Rust. However, that
was stabilized in [RFC #1857] last year, so we no longer need to hold an
Rc for that reason. We do now use the DB connection to provide better
error messages, but we can get that from the statement object quite
easily.

[RFC #1857]: https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -9,7 +9,6 @@ mod sqlite_value;
 pub use self::sqlite_value::SqliteValue;
 
 use std::os::raw as libc;
-use std::rc::Rc;
 
 use connection::*;
 use deserialize::{Queryable, QueryableByName};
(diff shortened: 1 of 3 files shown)


Diggsey:
Might be worth adding a comment to the struct where field order is important so that it doesn't get re-ordered by accident?

sgrif:
A comment is a poor way to ensure something like that remains true. The tests will unconditionally fail if the fields are re-ordered.

PR #1693, opened 2018-05-15
Abstract away `Option` special handling while deriving `Associations`

weiznich:
This commit introduces a new trait that abstracts the way foreign keys
are handled. When deriving `Associations` we need a way to extract the
value that is used to compare the foreign key with the referenced
primary key. For not nullable foreign keys this is simply the key
itself, for nullable keys only the not null value, otherwise the key
is skipped. Currently this is handled by some special casing in
`diesel_derives`, after this commit the newly introduced trait will be
used.
There are two reasons for this change:

* This simplifies `diesel_derives`. (In detail it removes the error
  prone detection of the `std::Option` type. This could theoretical
  break if someone uses a other type named `Option` in foreign key
  position, because currently there is no way to detect the origin of some type in
  custom derives.)
* It adds a way for down stream crates to introduce `Option` like
  wrapper types on there own. Using this change it should be possible
  to create a foreign key type wrapper that allows to store the
  referenced values inline later.

This is a more general successor of #1598. 

I really want to have such a thing because otherwise it is not possible for downstream crates to add a new wrapper type that could be used in position of a foreign key and works with the provides derive for `Associations`. For downstream end user crates this may not be that tragic because as Sean pointed out in #1598 this could be worked around by manually implementing `BelongsTo`. For downstream crates that provides additional functionality for other users as library this may be a no go. Currently there are only two bad solutions for this problem: 
1. Provide a own custom derive that implements `BelongsTo` -> Bad because the user needs to remember which custom derive is used in which position. Also this may break anytime as soon as diesel changes the own custom derive.
2. Require the users of that library to manually implement `BelongsTo` for every struct they use the custom type. -> Much repetitive code, therefore the `derive` exists.

The general idea of this PR is to add a third option: Allow that library to `hook` into the custom derive by implementing a trait indicating how to get a foreign key value out of a specific type. As Sean noted in #1598 this is quite similar to use case of `BelongsTo` but works on a other level. `BelongsTo` is meant to define how to get a foreign key out of a "model" type, meaning a type that represents some database entry. `ForeignKey` on the other hand is meant to describe how to get a foreign key value out of some wrapper type, describing a single field of a model type. The easiest type this could be applied to is  `Option<T>`. 

The use case for a custom type that I've in my mind is the following: I want something that could store referenced parent items inside the child type. The mean motivation for this is that this greatly simplifies serialization to json/…. It is quite easy to write some type 
```
enum HasOne<F, I>{
    Id(F),
    Item(I)
}
```
that does exactly this. Implementing support for nearly all operations in diesel is quite easily possible. The only missing bit is that this type won't work with the current `Associations` derive because it would need a similar special casing like `Option<T>`. Because we couldn't add special cases for all possible existing types people will write it seems like a better option to remove the special casing in the custom derive and provide a abstraction point to do this in the type implementation.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -2,6 +2,7 @@ use dsl::{Eq, EqAny, Filter, FindBy};
 use expression::AsExpression;
 use expression::array_comparison::AsInExpression;
 use query_dsl::methods::FilterDsl;
+use sql_types::{NotNull, Nullable};
 use prelude::*;
 use super::{HasTable, Identifiable};
 
@@ -26,6 +27,23 @@ pub trait BelongsTo<Parent> {
     fn foreign_key(&self) -> Option<&Self::ForeignKey>;
     /// Returns the foreign key column of this struct's table
     fn foreign_key_column() -> Self::ForeignKeyColumn;
+}
+
+/// A trait to abstract which types could be used in a foreign key context.
+///
+/// The use case of this trait is to abstract away wrapper types used for example
+/// to represent if a foreign key could be null.
+///
+/// You will not need to interact with this trait, unless you are writing an
+/// implementation of custom type wrapper that could be used in foreign key position
+pub trait ForeignKey<ST> {
+    /// The (rust) type of the primary key that is referenced by the foreign key
+    type KeyType;
+
+    /// A helper function to extract the foreign key value from the wrapper struct.
+    ///
+    /// Returns `None` if the value is null otherwise `Some(value)`.
+    fn key(&self) -> Option<&Self::KeyType>;
 }
 
 /// The `grouped_by` function groups records by their parent.
@@ -172,5 +190,18 @@ where
 
     fn belonging_to(parents: &'a Vec<Parent>) -> Self::Output {
         Self::belonging_to(&**parents)
+    }
+}
+
+impl<T, ST> ForeignKey<Nullable<ST>> for Option<T>
+where
+    T: Hash + ::std::cmp::Eq,
+    Option<T>: AsExpression<Nullable<ST>>,
+    ST: NotNull,
+{
+    type KeyType = T;
+
+    fn key(&self) -> Option<&Self::KeyType> {
+        self.as_ref()
     }
 }
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -329,7 +329,7 @@ use std::hash::Hash;
 
 use query_source::Table;
 
-pub use self::belongs_to::{BelongsTo, GroupedBy};
+pub use self::belongs_to::{BelongsTo, GroupedBy, ForeignKey};
 
 /// This trait indicates that a struct is associated with a single database table.
 ///
--- a/diesel_derives/src/as_expression.rs
+++ b/diesel_derives/src/as_expression.rs
@@ -98,6 +98,14 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
                         Bound::new(self)
                     }
                 }
+
+                impl#impl_generics diesel::associations::ForeignKey<#sql_type> for #struct_ty {
+                    type KeyType = #struct_ty;
+
+                    fn key(&self) -> Option<&Self::KeyType> {
+                        Some(self)
+                    }
+                }
             )
         } else {
             tokens
--- a/diesel_derives/src/associations.rs
+++ b/diesel_derives/src/associations.rs
@@ -40,13 +40,11 @@ fn derive_belongs_to(
     let foreign_key_field = model.find_column(foreign_key)?;
     let struct_name = model.name;
     let foreign_key_access = foreign_key_field.name.access();
-    let foreign_key_ty = inner_of_option_ty(&foreign_key_field.ty);
+    let foreign_key_ty = &foreign_key_field.ty;
     let table_name = model.table_name();
-
-    let foreign_key_expr = if is_option_ty(&foreign_key_field.ty) {
-        quote!(self#foreign_key_access.as_ref())
-    } else {
-        quote!(std::option::Option::Some(&self#foreign_key_access))
+    let foreign_key_trait = quote!{
+        <#foreign_key_ty
+            as diesel::associations::ForeignKey<diesel::dsl::SqlTypeOf<#table_name::#foreign_key>>>
     };
 
     Ok(quote! {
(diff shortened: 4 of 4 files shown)


diesel/src/associations/belongs_to.rs
```diff
@@ -28,6 +29,23 @@ pub trait BelongsTo<Parent> {
...
+/// You will not need to interact with this trait, unless you are writing an
+/// implementation of custom type wrapper that could be used in foreign key position
+pub trait ForeignKey<ST> {
+    /// The (rust) type of the primary key that is referenced by the foreign key
+    type KeyType;
+
+    /// A helper function to extract the foreign key value from the wrapper struct.
+    ///
+    /// Returns `None` if the value is null otherwise `Some(value)`.
+    fn key(&self) -> Option<&Self::KeyType>;
```
sgrif: This signature won't be able to represent composite keys. Should this take `self` by value, and be implemented on a reference instead?

diesel/src/associations/belongs_to.rs
```diff
@@ -28,6 +29,23 @@ pub trait BelongsTo<Parent> {
...
 }
 
+/// A trait to abstract which types could be used in a foreign key context.
+///
+/// The use case of this trait is to abstract away wrapper types used for example
+/// to represent if a foreign key could be null.
+///
+/// You will not need to interact with this trait, unless you are writing an
+/// implementation of custom type wrapper that could be used in foreign key position
+pub trait ForeignKey<ST> {
```
sgrif: Should we just have `Hash + Eq` as bounds here?

diesel/src/associations/belongs_to.rs
```diff
@@ -174,3 +192,16 @@ where
         Self::belonging_to(&**parents)
     }
 }
+
+impl<T, ST> ForeignKey<Nullable<ST>> for Option<T>
+where
+    T: Hash + ::std::cmp::Eq,
+    Option<T>: AsExpression<Nullable<ST>>,
```
sgrif: Can we use `Self` here?

diesel/src/associations/belongs_to.rs
```diff
@@ -174,3 +192,16 @@ where
...
+    ST: NotNull,
+{
+    type KeyType = T;
```
sgrif: Should this implementation be more generic? e.g.

```
impl<T> ForeignKey<ST::Nullable> for Option<T>
where
    T: ForeignKey<ST>,
    ST: IntoNullable,
{
    type KeyType = T::KeyType;

    fn key(&self) -> Option<&Self::KeyType> {
        self.as_ref().map(T::key)
    }
}
```

diesel/src/associations/belongs_to.rs
sgrif: I don't understand why we need `ST` here.

sgrif:
Would this change serve your use case equally well?

```patch
diff --git a/diesel_derives/src/associations.rs b/diesel_derives/src/associations.rs
index e2882fc..a6cae7f 100644
--- a/diesel_derives/src/associations.rs
+++ b/diesel_derives/src/associations.rs
@@ -46,7 +46,7 @@ fn derive_belongs_to(
     let foreign_key_expr = if is_option_ty(&foreign_key_field.ty) {
         quote!(self#foreign_key_access.as_ref())
     } else {
-        quote!(std::option::Option::Some(&self#foreign_key_access))
+        quote!(&self#foreign_key_access.into())
     };
 
     Ok(quote! {
```

weiznich:
> I'm not sure I entirely understand the use case here. There certainly seems to be a more generic concept here than just foreign keys. It seems like this is trying to recreate Into.

My use case is basically to write a generic type `HasOne` that contains either a referenced id or the referenced item itself.

```rust
pub enum HasOne<R, T> {
    Id(R),
    Item(T),
}
```

I want to use this in my graphql resolver crate, that is build on top of diesel. (Graphql leads to quite nested structures.) For serialization to json/… it is easier to store the referenced values inline.

Basically I want to be able to write something like
```rust
#[derive(Queryable, Serialize)]
struct Post {
    title: String,
    content: Option<String>,
    user: HasOne<i32, User>,
}
```

So the next step is that I need some way to transform graphql "queries" into sql queries to load data from the database. My strategie here is to load data by different levels. Given a query that requests all posts with informations about the authors, this means to load all posts first (and populate the `HasOne::Id` variant). After that load all corresponding users (using the id form `HasOne::Id` to populate the `HasOne::Item` variant). To prevent N+1 queries I want to use the existing `BelongsTo` infrastructure in diesel. This means I need to find some way to implement `BelongsTo`. Manually implementing it would result in writing a lot of this implementations, so I look for some way to make the existing `#[derive(Associations)]` work with custom types. In the end it boils down to have some way describing how to handle a certain type in foreign key positions. Currently this is done entirely in `diesel_derive` by special casing `Option`.

> Would this change serve your use case equally well?

Unfortunately not, because I'm not allowed to write an generic impl for `Option<T>`

```rust
impl<'a, R, T> From<&'a HasOne<R, T>> for Option<&'a R>
where
    &'a T: Identifiable<Id = &'a R>,
{
    fn from(m: &'a HasOne<R, T>) -> Self {
        match *m {
            HasOne::Id(ref id) => Some(id),
            HasOne::Item(ref item) => Some(item.id()),
        }
    }
}
```

sgrif:
We can use `Into` instead.

```
impl<'a, R, T> Into<Option<&'a R>> for &'a HasOne<R, T>
where
    &'a T: Identifiable<Id = &'a R>,
{
    fn into(self) -> Option<&'a R> {
        match *self {
            HasOne::Id(ref id) => Some(id),
            HasOne::Item(ref item) => Some(item.id()),
        }
    }
}
```

weiznich:
That seems to work. (I didn't even try that because I assumed this would conflict with the default implementation for `From`)

weiznich:
Wait, stop: Turns out this kind of solves only one part of the issue because we have no way to know the correct type for `ForeignKey` in `BelongsTo`. This type will change depending on the foreign key wrapper...

sgrif:
I think this is what we need then:

```patch
diff --git a/diesel_derives/src/associations.rs b/diesel_derives/src/associations.rs
index e2882fc..e785601 100644
--- a/diesel_derives/src/associations.rs
+++ b/diesel_derives/src/associations.rs
@@ -43,12 +43,6 @@ fn derive_belongs_to(
     let foreign_key_ty = inner_of_option_ty(&foreign_key_field.ty);
     let table_name = model.table_name();
 
-    let foreign_key_expr = if is_option_ty(&foreign_key_field.ty) {
-        quote!(self#foreign_key_access.as_ref())
-    } else {
-        quote!(std::option::Option::Some(&self#foreign_key_access))
-    };
-
     Ok(quote! {
         impl #impl_generics diesel::associations::BelongsTo<#parent_struct>
             for #struct_name #ty_generics
@@ -58,7 +52,7 @@ fn derive_belongs_to(
             type ForeignKeyColumn = #table_name::#foreign_key;
 
             fn foreign_key(&self) -> std::option::Option<&Self::ForeignKey> {
-                #foreign_key_expr
+                (&self#foreign_key_access).into()
             }
 
             fn foreign_key_column() -> Self::ForeignKeyColumn {
```

sgrif:
I'm a bit too tired to look at this more in depth. I suspect the trait you have in mind is correct in that case, but it should probably have a name other than `ForeignKey`. I'd like to look at it more in depth though.

PR #1706, opened 2018-05-19
Automatically generate a schema file by default

sgrif:
Since we're deprecating `infer_schema!`, we want to generate this file
by default. I think we should eventually change our behavior on tables
with no primary key to warn and skip instead of error, but I want to do
a bigger restructuring once we can get rid of `infer_schema!`.

### DIFF
--- a/diesel_cli/src/default_files/diesel.toml
+++ b/diesel_cli/src/default_files/diesel.toml
@@ -1,2 +1,5 @@
 # For documentation on how to configure this file,
 # see diesel.rs/guides/configuring-diesel-cli
+
+[print_schema]
+file = "src/schema.rs"
--- a/diesel_cli/tests/migration_list.rs
+++ b/diesel_cli/tests/migration_list.rs
@@ -16,7 +16,7 @@ fn migration_list_lists_pending_applied_migrations() {
 
     p.create_migration(
         "12345_create_users_table",
-        "CREATE TABLE users ( id INTEGER )",
+        "CREATE TABLE users (id INTEGER PRIMARY KEY)",
         "DROP TABLE users",
     );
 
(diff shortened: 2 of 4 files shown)


Diggsey:
Is there a previous discussion about this? I saw the message in this commit: https://github.com/diesel-rs/diesel/commit/ae22270600aeddeb8b95e6b88f48060bd514ab9a but the issue number seems incorrect as that links to my PR?

This seems to make some sense for the "single developer" workflow - however, I'm not sure about how well it will work with a team, or when running migrations on production via the CLI.

Presumably, `schema.rs` will be checked into version control, so when you pull the latest version and run the migrations, you don't necessarily want it to modify your schema. Likewise, on production it may not even be possible to modify the schema file if it's running of a read-only file-system.

sgrif:
I'm not sure I understand your concern. Why would pulling the latest version and running migrations modify this file?

Diggsey:
AIUI, whenever you run migrations via the CLI tool now, it will regenerate and overwrite your `schema.rs` file, correct?

Are you saying that it should rewrite it exactly the same because the database state should be the same? It's mostly just a bit of a red flag for me whenever there's something in version control which is transparently updated like this.

I think I would be less concerned if there was a `--validate-only` option for use in CI/CD pipelines, which runs the migrations, infers the schema, and then instead of overwriting the schema file, simply compares the two and returns an error if they don't match?

sgrif:
@Diggsey I think that's a good idea, but ultimately unrelated to this PR. :smile: This is just setting the convention that schema stuff should live in `src/schema.rs` and setting that by default for new projects. As usual, it's easy to opt out of this if it's problematic.

Diggsey:
Sure, that's why I wanted to know if there was a previous discussion of the broader feature 😛

sgrif:
I don't think there was one on Github.

PR #1707, opened 2018-05-19
Deprecate `infer_schema!`

sgrif:
`diesel print-schema` is now the preferred way to generate schema.rs.


limira:
Personally, I never install the `diesel-cli` because `infer_schema` is enough for me. 

`diesel.rs/guides/configuring-diesel-cli` is not available right now so the only thing I can say is that I wish/hope it is as easy as using `infer_schema`.

PR #1712, opened 2018-05-21
Check if there is a newer version of diesel in the dependency graph.

### LINKED ISSUE #1704: Yank old versions of diesel_codegen
We seeing quite periodic questions/bugreports about something broken in diesel that boils down to someone tries to use the old `diesel_codegen`. 
I suggest to yank at least the 0.16 version to indicate that this crate shouldn't be used anymore.

weiznich:
This means most likely that the build will break because of a changed codegen.

Fixes #1704 

@sgrif The 0.16.x branch is missing. Could you crate one?

### DIFF
--- a/diesel_codegen/Cargo.toml
+++ b/diesel_codegen/Cargo.toml
@@ -17,6 +17,9 @@ dotenv = { version = ">=0.8, <0.11", optional = true, default-features = false }
 diesel = { version = "0.16.0", default-features = false }
 diesel_infer_schema = { version = "0.16.0", default-features = false, optional = true }
 clippy = { optional = true, version = "=0.0.138" }
+serde_json = "1"
+serde = "1"
+serde_derive = "1"
 
 [dev-dependencies]
 tempdir = "0.3.4"
--- a/diesel_codegen/Cargo.toml
+++ b/diesel_codegen/Cargo.toml
@@ -17,9 +17,8 @@ dotenv = { version = ">=0.8, <0.11", optional = true, default-features = false }
 diesel = { version = "0.16.0", default-features = false }
 diesel_infer_schema = { version = "0.16.0", default-features = false, optional = true }
 clippy = { optional = true, version = "=0.0.138" }
-serde_json = "1"
-serde = "1"
-serde_derive = "1"
+serde_json = "1.0"
+serde = { version = "1.0", features = ["derive"] }
 
 [dev-dependencies]
 tempdir = "0.3.4"
--- a/diesel_codegen/src/check_version.rs
+++ b/diesel_codegen/src/check_version.rs
@@ -1,0 +1,42 @@
+
+use std::process::Command;
+
+#[derive(Deserialize)]
+struct Metadata {
+    packages: Vec<Package>,
+}
+
+#[derive(Deserialize, Debug)]
+struct Package {
+    name: String,
+    version: String,
+}
+#[allow(warnings)]
+pub fn check_version() {
+    let contents = String::from_utf8(
+        Command::new("cargo")
+            .arg("metadata")
+            .output()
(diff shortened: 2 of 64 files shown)


sgrif:
@weiznich You should have all the rights you need to create it

weiznich:
> You should have all the rights you need to create it.

Right. (I totally forgot about that 😟 )

diesel_codegen/src/check_version.rs
```diff
@@ -0,0 +1,42 @@
+
```
Eijebong: :scissors:

weiznich:
> Any reason why you aren't you running this in a build.rs ?

Because I've found no way to get the working directory of a down stream crate from `build.rs`. `CARGO_MANIFEST_DIR` is set there to the directory where `diesel_codegen` lives.

sgrif:
Are we unable to run `cargo metadata` in a build script?

diesel_codegen/src/check_version.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Deserialize)]
+struct Metadata {
+    packages: Vec<Package>,
+}
+
+#[derive(Deserialize, Debug)]
+struct Package {
+    name: String,
+    version: String,
+}
```
sgrif: Newline please :smile:

sgrif:
The `.unwrap`s worry me a bit. I don't want to risk breaking old builds with this. Can we make sure we just return and do nothing if something along the way errors?

diesel_codegen/src/check_version.rs
```diff
@@ -0,0 +1,42 @@
...
+            .unwrap()
+            .stdout,
+    ).unwrap();
+
+    let data: Metadata = ::serde_json::from_str(&contents).unwrap();
+
+    let diesel = data.packages
+        .into_iter()
+        .filter(|p| p.name == "diesel")
+        .collect::<Vec<_>>();
```
sgrif: We don't need this `.collect`

diesel_codegen/Cargo.toml
```diff
@@ -17,6 +17,9 @@ dotenv = { version = ">=0.8, <0.11", optional = true, default-features = false }
 diesel = { version = "0.16.0", default-features = false }
 diesel_infer_schema = { version = "0.16.0", default-features = false, optional = true }
 clippy = { optional = true, version = "=0.0.138" }
+serde_json = "1"
+serde = "1"
+serde_derive = "1"
```
sgrif: Can we do `serde = { version = "1.0", features = ["derive"] }` instead?

sgrif:
Can you edit `.travis.yml` to run against this branch in your PR too please?

weiznich:
> Are we unable to run cargo metadata in a build script?

It is possible to run `cargo metadata ` in a build-script, but it is run in the diesel_codegen directory and therefore will result in the dependency list of diesel_codegen. Theoretically we could change the working directory, but I've found no way to get the directory of the downstream crate to run `cargo metadata` there. (Totally possible that I miss something)

weiznich:
> Can you edit .travis.yml to run against this branch in your PR too please?

I've added the branch to the `.travis.yml` but it seems like the test suite is broken on current stable + some dependency is broken on the old nightly.

sgrif:
Ah right, we'll need to cherry pick the fix for that.

weiznich:
I got finally a green CI on this. I've disabled the clippy CI entry because some dependency (pq-sys) does not support that old rustc version.
If there are no objectives I will try to release diesel_codegen as 0.16.1 tomorrow. Do this need a changelog entry? Or a distinct tag?

sgrif:
Yes to both

PR #1714, opened 2018-05-22
Our minimum supported Rust version is 1.24.

sgrif:
This was already true, as we require
https://github.com/rust-lang/rust/pull/46094 to compile. This commit
ensures that we test for this on CI so we know if this changes in the
future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,11 @@ matrix:
     - SQLITE_DATABASE_URL=/tmp/test.db
     script:
     - (cd diesel_cli && travis-cargo test -- --no-default-features --features "sqlite-bundled")
+  - rust: 1.24.1
+    env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -61,6 +61,11 @@ matrix:
     - SQLITE_DATABASE_URL=/tmp/test.db
     script:
     - (cd diesel_cli && travis-cargo test -- --no-default-features --features "sqlite-bundled")
+  - rust: 1.24.1
+    env:
+    - ENSURE_SUPPORTED_RUST_VERSION=1.24.1
```
Eijebong: Does this do anything ?

sgrif: Puts a nice label in the matrix so it's clear what this does

Eijebong: Oh :+1:

PR #1720, opened 2018-05-22
rm `diesel_infer_schema`

sgrif:
RIP `infer_schema!`
0.3.0 - 1.3.0
You served us well

`infer_schema_internals` is still used by `diesel_cli`, but there's no
reason for it to continue to be its own crate, so I've just inlined it
into Diesel CLI. We can almost certainly refactor it to establish
significantly fewer database connections now, but I've limited this to
the smallest possible diff that compiles.


weiznich:
I vote against removing the `infer_schema_internals` crate because I plan eventually to provide a `diesel print-schema` like utility for wundergraph at some later point. It would be nice to reuse the existing code to load a schema from the database to write such an tool. So maybe it is possible to keep it as a separate crate?

sgrif:
I think that a crate which exposes this as public API is better rewritten from scratch. The current API is not something I'm comfortable publishing for external use. If nothing else, it will definitely have to live in a crate with a name other than "_internals"

xianghx:
I use it. Somthing like this:
xdiesel print-struct
  ```
    #[derive(Identifiable, Queryable)]
    pub struct User {
        pub id: i32,
        pub name: String,
        pub favorite_color: Option<String>,
        pub created_at: NaiveDateTime,
    }

    #[derive(Insertable)]
    #[table_name="users"]
    pub struct NewUser<'a> {
        pub name: &'a str,
        pub favorite_color: Option<&'a str>,
        pub created_at: NaiveDateTime,
    }

    #[derive(AsChangeset)]
    #[table_name="users"]
    pub struct UpdateUser<'a> {
        pub name: &'a str,
        pub favorite_color: Option<&'a str>,
        pub created_at: NaiveDateTime,
    }

    NewUser {
        name,
        favorite_color,
        created_at,
    }

    UpdateUser {
        name,
        favorite_color,
        created_at,
    }

    #[derive(Identifiable, Queryable)]
    pub struct User {
        pub id: i32,
        pub name: String,
        pub favorite_color: Option<String>,
        pub created_at: NaiveDateTime,
    }

    #[derive(Insertable)]
    #[table_name="users"]
    pub struct NewUser<'a> {
        pub name: &'a str,
        pub favorite_color: Option<&'a str>,
        pub created_at: NaiveDateTime,
    }

    #[derive(AsChangeset)]
    #[table_name="users"]
    pub struct UpdateUser<'a> {
        pub name: &'a str,
        pub favorite_color: Option<&'a str>,
        pub created_at: NaiveDateTime,
    }

    NewUser {
        name,
        favorite_color,
        created_at,
    }

    UpdateUser {
        name,
        favorite_color,
        created_at,
    }

```
or  xdiesel print-html
```
<label for="id">id</label>
<label for="title">title</label><input type="text" name="title" id="title"><br>
<label for="body">body</label><input type="text" name="body" id="body"><br>
<label for="published">published</label><input type="checkbox" name="published" id="published"><br>

```

weiznich:
I'm with @sgrif on this. A crate that exposes this functionality as public api should be written from scratch. I'm fine with temporally moving this inside of diesel-cli. As soon as someone starts working on that new crate those code could be used as starting point. It will not be more work to copy the code from diesel-cli instead of from infer_schema_internals.


@xianghx Existing code using the old crate will just continue to work, so there is no problem. If we release a rewritten `infer_schema` crate at some point in time you could just update to the new crate.

xianghx:
I'm ok. Just give me a new crate.

sgrif:
@weiznich It's continuing to use 1.3.0 from crates.io

jhpratt:
@sgrif `infer_schema!` is still referenced in [the docs](https://docs.diesel.rs/diesel/index.html?search=#declaring-your-schema).

PR #1735, opened 2018-05-27
Add a helper trait for writing `ToSql` impls for PG tuples

sgrif:
I had meant to do this when adding support for `Record`, but didn't get
around to it. However, @ebkalderon needed this today, so now is as
good a time as any.

We avoid implementing `ToSql` directly, because we don't want
`Bound<Record<(Integer, Text)>, (i32, String)>` to become valid.

### DIFF
--- a/diesel/src/pg/types/record.rs
+++ b/diesel/src/pg/types/record.rs
@@ -1,11 +1,14 @@
-use byteorder::{NetworkEndian, ReadBytesExt};
+use byteorder::*;
+use std::io::Write;
+
 use deserialize::{self, FromSql, FromSqlRow, Queryable};
 use expression::{AppearsOnTable, AsExpression, Expression, NonAggregate, SelectableExpression};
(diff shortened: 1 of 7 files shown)


diesel/src/pg/types/record.rs
```diff
@@ -190,4 +221,34 @@ mod tests {
             .get_result(&conn);
         assert_eq!(Ok(true), res);
     }
+
+    #[test]
+    fn serializing_named_composite_types() {
```
weiznich: I think this test case should live in `diesel_tests/tests/custom_types.rs` because we link that file everywhere.

ebkalderon:
Looks like the failing rustfmt checks in CI have been fixed. Is this ready to merge?

xpe:
Ah, I think this is the write up: http://docs.diesel.rs/diesel/serialize/trait.WriteTuple.html

PR #1742, opened 2018-05-30
update deps

rizakrko:
Updates:
`quote`: 0.5 => 0.6
`syn`: 0.13 => 0.14
`proc-macro2`: 0.3 => 0.4

### DIFF
--- a/diesel_derives/Cargo.toml
+++ b/diesel_derives/Cargo.toml
@@ -9,10 +9,10 @@ homepage = "https://diesel.rs"
 repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
 
 [dependencies]
-syn = { version = "0.13.0", features = ["full", "fold"] }
-quote = "0.5.0"
+syn = { version = "0.14", features = ["full", "fold"] }
+quote = "0.6"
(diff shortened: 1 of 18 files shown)


sgrif:
Is there a specific reason we need any of these changes?

rizakrko:
There is no particular reason, just a routine update

sgrif:
I'd prefer not to arbitrarily churn code unless there's a specific reason, especially WRT things as unstable as syn and proc_macro2. If someone on the core team is reading and strongly disagrees, feel free to reopen.

sgrif:
@Eijebong disagrees

Eijebong:
I don't think they're that unstable and I'd prefer to review `0.13 -> 0.14` and then `0.14 -> whatever` than jumping two versions.

rizakrko:
Failure is related to changes in nightly compiler, so [#51042](https://github.com/rust-lang/rust/pull/51042) has to be merged, and then `clippy` and related nightly needs to be updated.

xfix:
If somebody needs a reason, this reduces amount of compilation if using diesel with another crate which has newest version of those dependencies.

PR #1749, opened 2018-06-01
Fix the handling of unsigned values on MySQL

### LINKED ISSUE #1745: MySQL u32 overflow at i32::max_value()
### Versions

- **Rust:** rustc 1.26.1 (827013a31 2018-05-25)
- **Diesel:** 1.3.0
- **Database:** MySQL 8
- **Operating System** CentOS

### Feature Flags

- **diesel:** diesel = { version = "1.0.0", features = ["mysql", "chrono"] }

## Problem Description
Trying to insert a u32 above i32::max_value causes an overflow and database error on MySQL

### Steps to reproduce (reduced testcase)

```rust
#[macro_use]
extern crate diesel;
extern crate dotenv;

use diesel::prelude::*;
use diesel::insert_into;
use dotenv::dotenv;
use std::env;

table! {
    DATA (id) {
        id -> Integer,
        record_id -> Unsigned<Integer>,
        session_id -> Unsigned<Integer>,
    }
}

#[derive(Debug, Insertable)]
#[table_name = "DATA"]
pub struct Data {
    pub record_id: u32,
    pub session_id: u32,
}

fn main() {
    use DATA::dsl::*;

    dotenv().ok();
    let database_url = env::var("DATABASE_URL").expect("DATABASE_URL must be set");
    let connection = MysqlConnection::establish(&database_url)
        .expect(&format!("Error connecting to {}", database_url));

    let i32_max = i32::max_value() as u32;
    let data_ok = Data {
        record_id: 1,
        session_id: i32_max
    };

    insert_into(DATA).values(&data_ok).execute(&connection).unwrap();

    let data_error = Data {
        record_id: 1,
        session_id: i32_max + 1
    };

    insert_into(DATA).values(&data_error).execute(&connection).unwrap();
}
```

### Output
```
    Finished dev [unoptimized + debuginfo] target(s) in 10.45 secs
     Running `target/debug/overflow`

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(__Unknown, "Out of range value for column \'session_id\' at row 1")', libcore/result.rs:945:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:207
             at libstd/panicking.rs:223
             at libstd/panicking.rs:402
   5: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:349
   6: rust_begin_unwind
             at libstd/panicking.rs:325
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:72
   8: core::result::unwrap_failed
             at /checkout/src/libcore/macros.rs:26
   9: <core::result::Result<T, E>>::unwrap
             at /checkout/src/libcore/result.rs:782
  10: overflow::main
             at src/main.rs:46
  11: std::rt::lang_start::{{closure}}
             at /checkout/src/libstd/rt.rs:74
  12: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:306
  13: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
  14: std::rt::lang_start_internal
             at libstd/panicking.rs:285
             at libstd/panic.rs:361
             at libstd/rt.rs:58
  15: std::rt::lang_start
             at /checkout/src/libstd/rt.rs:74
  16: main
  17: __libc_start_main
  18: <unknown>


```
Actual SQL executed via the general log
```
2018-06-01T04:52:07.152401Z        74 Query     SET sql_mode=(SELECT CONCAT(@@sql_mode, ',PIPES_AS_CONCAT'))
2018-06-01T04:52:07.152678Z        74 Query     SET time_zone = '+00:00'
2018-06-01T04:52:07.152931Z        74 Query     SET character_set_client = 'utf8mb4'
2018-06-01T04:52:07.153124Z        74 Query     SET character_set_connection = 'utf8mb4'
2018-06-01T04:52:07.153405Z        74 Query     SET character_set_results = 'utf8mb4'
2018-06-01T04:52:07.154014Z        74 Prepare   INSERT INTO `DATA` (`record_id`, `session_id`) VALUES (?, ?)
2018-06-01T04:52:07.154222Z        74 Execute   INSERT INTO `DATA` (`record_id`, `session_id`) VALUES (1, 2147483647)
2018-06-01T04:52:07.323184Z        74 Close stmt
2018-06-01T04:52:07.323492Z        74 Prepare   INSERT INTO `DATA` (`record_id`, `session_id`) VALUES (?, ?)
2018-06-01T04:52:07.323738Z        74 Execute   INSERT INTO `DATA` (`record_id`, `session_id`) VALUES (1, -2147483648)
2018-06-01T04:52:07.324176Z        74 Close stmt
2018-06-01T04:52:07.324698Z        74 Quit
```
## Checklist

sgrif:
We were sending the right bytes, but we weren't properly setting the
`is_unsigned` flag on the bind parameters, resulting in it always being
sent as signed. This results in a client error about overflow when
deserializing large values, and either a server error or incorrect
results when serializing them.

The fact that the tests for this all used `-1` should have been a huge
red flag, but it slipped through.

Unfortunately, there's no way for us to reasonably fix this without a
breaking change or crazy hacks (we could do something like write an
additional byte at the end if it's unsigned, but that's probably not
worth it to avoid a breaking change). For that reason, we will need to
make the next version of Diesel 2.0. We shouldn't go crazy with this
change, but this would also allow us to make #1457 happen as well.

The other alternative would be to introduce a `Mysql2` backend, and
deprecate the current one.

Fixes #1745

### DIFF
--- a/diesel/src/mysql/backend.rs
+++ b/diesel/src/mysql/backend.rs
@@ -22,12 +22,20 @@ pub struct Mysql;
 pub enum MysqlType {
     /// Sets `buffer_type` to `MYSQL_TYPE_TINY`
     Tiny,
+    /// Sets `buffer_type` to `MYSQL_TYPE_TINY` and `is_unsigned` to `true`
+    UnsignedTiny,
     /// Sets `buffer_type` to `MYSQL_TYPE_SHORT`
     Short,
+    /// Sets `buffer_type` to `MYSQL_TYPE_SHORT` and `is_unsigned` to `true`
+    UnsignedShort,
     /// Sets `buffer_type` to `MYSQL_TYPE_LONG`
     Long,
+    /// Sets `buffer_type` to `MYSQL_TYPE_LONG` and `is_unsigned` to `true`
+    UnsignedLong,
     /// Sets `buffer_type` to `MYSQL_TYPE_LONGLONG`
     LongLong,
+    /// Sets `buffer_type` to `MYSQL_TYPE_LONGLONG` and `is_unsigned` to `true`
+    UnsignedLongLong,
     /// Sets `buffer_type` to `MYSQL_TYPE_FLOAT`
     Float,
     /// Sets `buffer_type` to `MYSQL_TYPE_DOUBLE`
@@ -44,6 +52,19 @@ pub enum MysqlType {
     String,
     /// Sets `buffer_type` to `MYSQL_TYPE_BLOB`
     Blob,
+}
+
+impl MysqlType {
+    /// Is this an unsigned type?
+    pub fn is_unsigned(&self) -> bool {
+        match *self {
+            MysqlType::UnsignedTiny
+            | MysqlType::UnsignedShort
+            | MysqlType::UnsignedLong
+            | MysqlType::UnsignedLongLong => true,
+            _ => false,
+        }
+    }
 }
 
 impl Backend for Mysql {
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -35,9 +35,13 @@ impl Binds {
     }
 
     pub fn from_result_metadata(fields: &[ffi::MYSQL_FIELD]) -> Self {
+        const UNSIGNED_FLAG: libc::c_uint = 32;
         let data = fields
             .iter()
-            .map(|field| field.type_)
+            .map(|field| {
+                let unsigned = field.flags & UNSIGNED_FLAG > 0;
+                (field.type_, unsigned as _)
+            })
             .map(BindData::for_output)
             .collect();
 
@@ -89,6 +93,7 @@ struct BindData {
     length: libc::c_ulong,
     is_null: ffi::my_bool,
     is_truncated: Option<ffi::my_bool>,
+    is_unsigned: ffi::my_bool,
 }
 
 impl BindData {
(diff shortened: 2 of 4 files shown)


Diggsey:
@sgrif given that this previously caused an error, my view is that it should be introduced as a point release. If you consider the number of applications which are inadvertently broken today and would be fixed by this change vs programs which *would* be broken by introducing this change, the former is likely to be quite large, and the latter is likely to be zero.

That is, unless it is possible today to round-trip a value through the database without an error, and get a different result from what you'd get with this change applied?

sgrif:
@Diggsey The problem is that it's not just a change in behavior, it could cause code which was previously compiling to no longer compile. While I do think that it's incredibly unlikely that anyone is exhaustively matching this enum (there's no reason to do so unless you're implementing a connection), it's documented and part of our public API. Specifically this enum is structured to allow exhaustively matching, so I think it would be against the spirit of our commitment to semver if we can avoid it. (We can avoid it, see #1749)

Diggsey:
Makes sense 👍

PR #1751, opened 2018-06-01
Fix our handling of unsigned types on MySQL

### LINKED ISSUE #1745: MySQL u32 overflow at i32::max_value()

sgrif:
We were sending the right bytes, but we weren't properly setting the
`is_unsigned` flag on the bind parameters, resulting in it always being
sent as signed. This results in a client error about overflow when
deserializing large values, and either a server error or incorrect
results when serializing them.

The fact that the tests for this all used `-1` should have been a huge
red flag, but it slipped through.

While changing the type of `Mysql::BindCollector` could technically be
considered a breaking change under the strictest definition, I do not
consider this to be one. We have stated in the past that folks should
not rely on the exact types used in `impl Backend`. And unlike changing
`RawValue`, there's literally no reason for anybody to have depended on
the exact value of this type, unless they're implementing their own
Connection.

If we really want to avoid any potential source incompatibility, we can
actually do this without the custom bind collector, if we have the
`ToSql` implementations for `Unsigned` add an extra byte at the end, and
check to see if the length of the buffer is 1 byte larger than the
expected length. However, this is super hacky and much more likely to
break applications than this approach.

Fixes #1745.
Close #1749.

### DIFF
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -401,17 +401,44 @@ pub trait HasSqlType<ST>: TypeMetadata {
     /// of this method should not do dynamic lookup unless absolutely necessary
     fn metadata(lookup: &Self::MetadataLookup) -> Self::TypeMetadata;
 
-    /// Fetch the metadata for a tuple representing an entire row
-    ///
-    /// The default implementation of this method simply calls `Self::metadata`.
-    /// You generally should not need to override this method.
-    ///
-    /// However, if you are writing an implementation of `HasSqlType` that
-    /// simply delegates to an inner type (for example, `Nullable` does this),
-    /// then you should ensure that you delegate this method as well.
+    #[doc(hidden)]
+    #[cfg(feature = "with-deprecated")]
+    #[deprecated(
+        since = "1.4.0",
+        note = "This method is no longer used, and has been deprecated without replacement"
+    )]
     fn row_metadata(out: &mut Vec<Self::TypeMetadata>, lookup: &Self::MetadataLookup) {
         out.push(Self::metadata(lookup))
     }
+
+    #[doc(hidden)]
+    #[cfg(feature = "mysql")]
+    fn is_signed() -> IsSigned {
+        IsSigned::Signed
+    }
+
+    #[doc(hidden)]
+    #[cfg(feature = "mysql")]
+    fn mysql_metadata(lookup: &Self::MetadataLookup) -> (Self::TypeMetadata, IsSigned) {
+        (Self::metadata(lookup), Self::is_signed())
+    }
+
+    #[doc(hidden)]
+    #[cfg(feature = "mysql")]
+    fn mysql_row_metadata(
+        out: &mut Vec<(Self::TypeMetadata, IsSigned)>,
+        lookup: &Self::MetadataLookup,
+    ) {
+        out.push(Self::mysql_metadata(lookup))
+    }
+}
+
+#[doc(hidden)]
+#[cfg(feature = "mysql")]
+#[derive(Debug, Clone, Copy)]
+pub enum IsSigned {
+    Signed,
+    Unsigned,
 }
 
 /// Information about how a backend stores metadata about given SQL types
--- a/diesel/src/mysql/backend.rs
+++ b/diesel/src/mysql/backend.rs
@@ -2,9 +2,9 @@
 
 use byteorder::NativeEndian;
 
+use super::bind_collector::MysqlBindCollector;
 use super::query_builder::MysqlQueryBuilder;
 use backend::*;
-use query_builder::bind_collector::RawBytesBindCollector;
 use sql_types::TypeMetadata;
 
 /// The MySQL backend
@@ -48,7 +48,7 @@ pub enum MysqlType {
 
 impl Backend for Mysql {
     type QueryBuilder = MysqlQueryBuilder;
-    type BindCollector = RawBytesBindCollector<Mysql>;
+    type BindCollector = MysqlBindCollector;
     type RawValue = [u8];
     type ByteOrder = NativeEndian;
 }
--- a/diesel/src/mysql/bind_collector.rs
+++ b/diesel/src/mysql/bind_collector.rs
@@ -1,0 +1,38 @@
+use super::{Mysql, MysqlType};
+use query_builder::BindCollector;
+use result::Error::SerializationError;
+use result::QueryResult;
+use serialize::{IsNull, Output, ToSql};
+use sql_types::{HasSqlType, IsSigned};
+
+#[derive(Default)]
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub struct MysqlBindCollector {
+    pub(crate) binds: Vec<(MysqlType, IsSigned, Option<Vec<u8>>)>,
+}
+
+impl MysqlBindCollector {
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl BindCollector<Mysql> for MysqlBindCollector {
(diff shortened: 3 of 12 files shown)


sgrif:
@diesel-rs/reviewers Between this and #1744, we have enough for a 1.3.2 release. Need some feedback here.

diesel/src/sql_types/mod.rs
```diff
@@ -401,17 +401,44 @@ pub trait HasSqlType<ST>: TypeMetadata {
...
     }
+
+    #[doc(hidden)]
+    #[cfg(feature = "mysql")]
+    fn is_signed() -> IsSigned {
+        IsSigned::Signed
+    }
+
+    #[doc(hidden)]
+    #[cfg(feature = "mysql")]
```
weiznich: I do not really like to have those two mysql specific functions here. (On the other hand I see no better solution + `row_metadata` was also only used by mysql`)

Eijebong:
While I'm not happy with the patch as it's technically a breaking change, the path to update seems pretty straightforward and it fixes a bug that shouldn't be here in the first place and that users might hit pretty easily...
My main problem is that accepting that as a non breaking change sets a precedent that I'm really reluctant to set as I'm pretty attached to semver but we also aren't ready for 2.0...

killercup:
Ugh, yeah, that's… not as great as it possibly can be. I too would rather not pollute a generic trait with backend specific methods. But as a compromise between adding this (touching code that probably only we will care about) and a major breaking change that users will notice? I'm actually okay with it. Can we maybe do a "mini crater run" with known production users?

Do we want to _really_ prevent making a breaking change? How far would we go with this? We could try and specialize `RawBytesBindCollector`. Well, we can't use actual specialization right now… Let's imagine this: We add another type parameter with concrete default type, like `RawBytesBindCollector<DB, Mode=ProperBytesBindCollector>` and change all our impls to be based on that. For MySQL, we impl `Backend` with `type BindCollector = RawBytesBindCollector<Mysql, Mode=AcutallyImproperMysqlSpecificBindCollector>`. Is this now a breaking change? Or an opt-in bugfix for everyone who actually had that trait bound in their code? Or am I just talking about weird stuff that would never work anyway (likely)?

sgrif:
> Can we maybe do a "mini crater run" with known production users?

Not that I'm aware of. The only other possible change here that I can come up with is to change the implementation of `ToSql` for unsigned types to add an extra byte at the end. I think this is the *most likely* of the options to break real code (there's very little reason for someone to exhaustively match `MysqlType`, and even less reason to depend on the exact type of this associated type. However, we encourage people to rely on the behavior of `ToSql` impls).

I'm not sure I entirely understand your second point, since structs can't have associated types.

Now that everybody has commented, it sounds like we're all in agreement that the 4 choices we have are:

- This
- #1749 
- This change, but instead of changing the type of `<Mysql as Backend>::BindCollector`, we instead change the `ToSql` impl of `Unsigned<T>` to write an extra byte at the end (aka crazy hacks)
- Release 2.0

And that we are all feel uneasy about this option, but agree that it is the best of those 4 choices. Unless anyone specifically objects, I'll merge this tomorrow and release it as 1.3.2

PR #1743, opened 2018-05-30
Update rustfmt

sgrif:
Old versions of rustfmt-nightly are no longer compiling since they don't
disable default features of serde, and the default features of serde
require Rust 1.26. Unfortunately, the rustfmt that ships with 1.26 is
super buggy, but all previous stable versions error on lines they can't
reformat that are >=100 characters, so we'll have to work around them.

PR #1762, opened 2018-06-17
Update rustfmt install instruction

galuhsahid:
Update rustfmt install instruction following the change in #1743.


sgrif:
Should we mention the specific version we run against?

galuhsahid:
Does the specific version change often? 

If so, I think we can write something like:

> The line `rust: x.x.x` above `rustup component add rustfmt-preview` in `travis.yml` is the specific version we run rustfmt against

 to avoid updating this file often.

Otherwise we can just mention the specific version. What do you think?

Eijebong:
The problem with specifying it there is that it's going to be forgotten IMO

weiznich:
What about just adding a `rust-toolchain` file with the corresponding version?

weiznich:
> What about nightly then ?

It is possible to override the used toolchain using `cargo +version ...`, correct? 
Otherwise we could just use rustfmt from the nightly that we use for clippy.

(Also given that clippy is deprecating the plugin in favour of the command line tool our need for nightly will be gone...)

sgrif:
I'm fine with pointing people to .travis.yml, my point was just that we should indicate that we're running a specific version that may not be the most recent stable (I don't plan on immediately updating when 1.27 comes out for example)

sgrif:
CI failure looks legit

weiznich:
I think all builders now use the toolchain specified in rust-toolchain.toml. That makes clippy failing. (I'm not sure if the normal channel builder picking the right toolchain.)

When I remember correctly it is possible to override that using `cargo +toolchain command ` in place of `cargo command` where toolchain ist something like stable, beta, a stable version or a nightly version. So I think at least the CI command to execute the clippy check must be adjusted.

AndrewSpeed:
👋 Would anyone object to me working on this issue?

sgrif:
Nope

weiznich:
Done by #1924

PR #1787, opened 2018-07-17
Future proof us against upcoming changes to derives

### LINKED ISSUE #1785: names from parent modules are not accessible without an explicit import
### Versions

- **Rust:1.29.0-nightly (254f8796b 2018-07-13)**
- **Diesel:1.3.2**
- **Database:sqlite**
- **Operating System:win10**

### Feature Flags

- **diesel:**

## Problem Description

I got a lot of warns after rustup and cargo update, like follow:

```
warning: cannot find type `ThingDefine` in this scope
  --> src\data\thing\mod.rs:71:41
   |
71 | #[derive(Serialize, Deserialize, Debug, Queryable, Clone, PartialOrd, PartialEq)]
   |                                         ^^^^^^^^^ names from parent modules are not accessible without an explicit import
   |
   = note: #[warn(proc_macro_derive_resolution_fallback)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #50504 <https://github.com/rust-lang/rust/issues/50504>

warning: cannot find type `NaiveDateTime` in this scope
  --> src\data\thing\mod.rs:71:41
   |
71 | #[derive(Serialize, Deserialize, Debug, Queryable, Clone, PartialOrd, PartialEq)]
   |                                         ^^^^^^^^^ names from parent modules are not accessible without an explicit import
   |
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #50504 <https://github.com/rust-lang/rust/issues/50504>

```
I try to add #[allow(proc_macro_derive_resolution_fallback)] on the structure used for query to disable the errors, but the warning exists yet.

I want to remove these warnings

sgrif:
Macro hygiene used to ignore module scope, which meant that we could
reference a type from outside our wrapper module even if we didn't
import it. That is changing in an upcoming version of Rust.

Unfortunately, we can't just do either of the recommended solutions
(either adding `use super::*` or changing to use a function instead of a
module). `use super::*` doesn't work for types defined inside of a
function, and changing ot use a function instead of a module broke our
workaround to pretend we had `$crate` when we really didn't.

With this change, an item named `diesel` must be present at the crate
root. In the case of Diesel itself, this is just a module, but for all
users they will now have to put `extern crate diesel` at the crate root.
It's unlikely folks were renaming it, but we will no longer work if they
do (the import was already located at the crate root, since you cant do
`#[macro_use] extern crate` anywhere else).

Fixes #1785.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 [write-tuple-1-4-0]: docs.diesel.rs/diesel/serialize/trait.WriteTuple.html
 
+### Changed
+
+* Diesel's derives now require that `extern crate diesel;` be at your crate root
+  (e.g. `src/lib.rs` or `src/main.rs`)
+
 ## [1.3.2] - 2018-06-13
 
 ### Fixed
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -354,3 +354,7 @@ pub use query_builder::debug_query;
 pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
                                    select, sql_query, update};
 pub use result::Error::NotFound;
+
+pub(crate) mod diesel {
+    pub use super::*;
+}
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -1042,15 +1042,6 @@ macro_rules! allow_tables_to_appear_in_same_query {
     () => {};
 }
 
-#[macro_export]
-#[doc(hidden)]
-/// Used by `diesel_derives`, which can't access `$crate`
-macro_rules! __diesel_use_everything {
-    () => {
-        pub use $crate::*;
-    };
-}
-
 /// Gets the value out of an option, or returns an error.
 ///
 /// This is used by `FromSql` implementations.
--- a/diesel_derives/src/as_changeset.rs
+++ b/diesel_derives/src/as_changeset.rs
@@ -61,8 +61,8 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
     Ok(wrap_in_dummy_mod(
         model.dummy_mod_name("as_changeset"),
         quote!(
-            use self::diesel::query_builder::AsChangeset;
-            use self::diesel::prelude::*;
+            use diesel::query_builder::AsChangeset;
+            use diesel::prelude::*;
 
             impl #impl_generics AsChangeset for &'update #struct_name #ty_generics
             #where_clause
--- a/diesel_derives/src/as_expression.rs
+++ b/diesel_derives/src/as_expression.rs
@@ -108,10 +108,10 @@ pub fn derive(item: syn::DeriveInput) -> Result<quote::Tokens, Diagnostic> {
         Ok(wrap_in_dummy_mod(
             dummy_mod.into(),
             quote! {
-                use self::diesel::expression::AsExpression;
-                use self::diesel::expression::bound::Bound;
-                use self::diesel::sql_types::Nullable;
-                use self::diesel::serialize::{self, ToSql, Output};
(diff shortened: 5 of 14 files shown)


djc:
So these warnings are now also triggered on stable Rust with 1.29.0, with the result that my little Diesel (1.3.3) project on stable Rust now generates pages and pages of warnings on every compile. Could this be backported to an 1.3.x release, or will be 1.4.0 be released soon?

sgrif:
https://github.com/diesel-rs/diesel/issues/1785#issuecomment-422577018

sergeysova:
Does it already release?

weiznich:
@sergeysova Unfortunately not because rustdoc does not build our documentation we cannot do a release currently.

weiznich:
@qrilka https://github.com/rust-lang/rust/issues/54524

r-arias:
@weiznich that issue should be resolved in stable 1.30.1 (`cargo doc` works for me on this version); is there anything else blocking a release?

weiznich:
@r-arias Yes, our ci is [failing on windows only](https://dev.azure.com/diesel-rs/887f3fe2-6950-4a87-8d32-b03825055869/_build/results?buildId=9). Unfortunately I have no access to a windows machine, so no way for me to track down that issue. Hopefully someone other from @diesel-rs/core (or someone other) with access to a windows machine could help here.

(That said: That error looks quite strange, because it's only happening on windows and the code that causes this according to rustc has no conditional thing based on the OS 😟. I also opened a issue on the rustc repo for this: https://github.com/rust-lang/rust/issues/55832 )

weiznich:
@r-arias Yes only sqlite causes problems.

r-arias:
Could it be that sqlite is wrongly configured in the win environment? Just a guess, I am not familiar with how the test stage is set up...

weiznich:
@r-arias That's not a configuration issue of the environment, because it does not fail at the linking stage, but while compiling the rust code. In detail while trying to resolve some traits. Those traits do not depend on any os specific impls and work on all other platforms. (That's probably a bug in rustc but without windows I'm not able to dig further into this.)

r-arias:
Hmm, I thought maybe the schema is wrongly generated:

```
2018-11-08T16:12:12.5030063Z error[E0433]: failed to resolve. Could not find `users` in `schema`
2018-11-08T16:12:12.5053093Z  --> diesel_tests\tests\select.rs:8:17
2018-11-08T16:12:12.5055022Z   |
2018-11-08T16:12:12.5055296Z 8 |     use schema::users::dsl::*;
2018-11-08T16:12:12.5059000Z   |                 ^^^^^ Could not find `users` in `schema`
```
([here](https://dev.azure.com/diesel-rs/diesel/_build/results?buildId=9&view=logs))

weiznich:
That would be possible, but how? The same database url is used for the build.rs file to run the migrations, so I do not see how running the migrations would be ok, but running infer_schema on the same url would fail afterwards.

r-arias:
Can't say, I'm just not familiar enough with this :/

r-arias:
Looks like it's building on azure, now?

https://dev.azure.com/diesel-rs/diesel/_build/results?buildId=10

I can't look at appveyor, where it is still failing.

icefoxen:
Sorry for reviving this, but I thought with Rust 2018 `#[macro_use]` and `extern crate ...;` were supposed to not be needed anymore.  I can't come up with the right combination of `use` decl's to get all the right bits of the macro in the same scope though.  Is there a good way to do this yet, or should we just stick with `#[macro_use]`?

Thanks a lot.

weiznich:
@icefoxen For supporting the new macro import feature we need to change some internal macros. (The issue is basically that we call internally some macros from the exported ones. Because of the macro hygiene rules a user is not able to provide reference to those internal macros.) Till that is done continue to use `#[macro_use]`.

jhpratt:
@weiznich Is there a tracking issue for proc macros? I'm unable to find any.

PR #1789, opened 2018-07-18
Update to nightly-2018-07-17

sgrif:
I would have preferred to do this in smaller pieces, but unfortunately
we had to do this all in one go. Clippy is no longer publishing to
crates.io, so we have to switch to using `cargo clippy`. There are also
a few bugs in recent versions that we have to work around.

The version of proc-macro that we were using no longer compiles on the
most recent nightlies, so we had to bump that as well (and therefore
also bump our version of `syn`). The biggest change here is that
`syn::Ident` is now `proc_macro2::Ident`, which does not implement
`From<&str>`, `AsRef<str>`, or `Copy`.

Finally, our UI tests regressed significantly, as Rust is no longer
giving us any useable span information for any piece of attributes in
derive macros. There's nothing we can do to work around this, we'll just
have to wait for Rust to fix it on their end.


diesel/src/connection/statement_cache.rs
```diff
@@ -108,7 +108,7 @@ pub struct StatementCache<DB: Backend, Statement> {
     pub cache: RefCell<HashMap<StatementCacheKey<DB>, Statement>>,
 }
 
-#[cfg_attr(feature = "clippy", allow(len_without_is_empty, new_without_default_derive))]
+#[cfg_attr(feature = "cargo-clippy", allow(len_without_is_empty, new_without_default_derive))]
```
Eijebong: I guess this is set by clippy when you run it ?

diesel_cli/src/infer_schema_internals/foreign_keys.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "cargo-clippy", allow(expect_fun_call))] // My calls are so fun
```
Eijebong: :laughing:

diesel_derives/src/queryable_by_name.rs
```diff
@@ -78,8 +78,8 @@ fn sql_type(field: &Field, model: &Model) -> syn::Type {
             parse_quote!(diesel::dsl::SqlTypeOf<#table_name::#column_name>)
         } else {
             let field_name = match field.name {
-                FieldName::Named(ref x) => x.as_ref(),
```
Eijebong: This could also be `to_string` instead of creating another `Ident`. That would make it clearer that it's not going to be used in a `quote!` call.

diesel_derives/src/util.rs
```diff
@@ -1,14 +1,13 @@
...
 use syn::*;
 
 pub use diagnostic_shim::*;
 use meta::*;
 
-pub fn wrap_in_dummy_mod(const_name: Ident, item: Tokens) -> Tokens {
+pub fn wrap_in_dummy_mod(const_name: Ident, item: TokenStream) -> TokenStream {
     quote! {
         #[allow(non_snake_case, unused_extern_crates, unused_imports)]
-        mod #const_name {
```
Eijebong: Hu ? This isn't a dummy mod then

diesel_derives/src/util.rs
```diff
@@ -60,14 +59,20 @@ pub fn ty_for_foreign_derive(item: &DeriveInput, flags: &MetaItem) -> Result<Typ
...
         Ok(parse_quote!(#ident #ty_generics))
     }
 }
 
-pub fn fix_span(maybe_bad_span: Span, fallback: Span) -> Span {
+pub fn fix_span(maybe_bad_span: Span, mut fallback: Span) -> Span {
     let bad_span_debug = "#0 bytes(0..0)";
+
+    if format!("{:?}", fallback) == bad_span_debug {
+        // On recent rust nightlies, even our fallback span is bad.
```
Eijebong: :/

diesel_derives/src/util.rs
sgrif: Sure -- Just didn't really feel like churning every call. Probably should have named it `dummy_scope` or something since we've now changed it from const to mod to fn

nabijaczleweli:
Does this also fix the docs not building on latest nightly?

sgrif:
I have no idea. We build our docs on stable.

nabijaczleweli:
Well, FTR, [this](https://travis-ci.com/Galactim/Sudoku/jobs/135225196#L3096) is what the error looks like (can you `cfg()` on is-doc?, seems like simplest solution).

nabijaczleweli:
Whoops! I based this project off an another, older one, and there seemed to not be any kind of (migration, deprecation?) notice on `diesel_codegen`, so I matched `diesel` version to `diesel_codegen`'s. Will explore that.

However, the error still occurs on latest `diesel`:

```
note: lint level defined here
   --> D:\Users\nabijaczleweli\.cargo\registry\src\github.com-1ecc6299db9ec823\diesel-1.3.2\src\lib.rs:131:9
    |
131 | #![deny(warnings, missing_debug_implementations, missing_copy_implementations, missing_docs)]
    |         ^^^^^^^^
    = note: #[deny(intra_doc_link_resolution_failure)] implied by #[deny(warnings)]
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`
```

diesel/src/connection/statement_cache.rs
killercup: yes

nabijaczleweli:
Okay, so I've the *simple* solution for the aforementioned problem;, that is: change [`srd/lib.rs#L131`](https://github.com/diesel-rs/diesel/pull/1789/files#diff-7ab3c70a207718b2fce878a8f925a415R131) to:
```rust
#![cfg_attr(feature = "dox", deny(warnings, missing_debug_implementations, missing_copy_implementations, missing_docs))]
```

This will only disallow warnings on non-doc builds, meaning that docs, while still warninging like hell (I'll probably PR that), do actually build.

PR #1795, opened 2018-07-23
Grab connection before finalizing statement for SQLite

xinghun92:
When dropping statement, inner_statement is finalized first,
and then is used to get the raw connection, which may cause a crash.
https://www.sqlite.org/c3ref/finalize.html


sgrif:
It seems like we just need to grab the connection before calling finalize

xinghun92:
@sgrif Are there other problems? Please review this commit once more.

weiznich:
The ci needs to pass before any further review is meaningful.

weiznich:
Already fixed by 1ba00c9eda7a3dbd6ec8b229c91c2f2965a34c54

PR #1798, opened 2018-07-26
R2D2 ConnectionManager: allow to pass queries executed for new conne…

Fiedzia:
…ctions

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -46,10 +46,10 @@ impl<T> ConnectionManager<T> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


xinghun92:
You can do this via CustomizeConnection 
 in [r2d2](https://github.com/sfackler/r2d2/blob/f56a081384578045acd515e45bcafc72d9cf723b/src/lib.rs#L122)

sgrif:
The above comment is correct. https://docs.rs/r2d2/0.8.2/r2d2/struct.Builder.html#method.connection_customizer

PR #1799, opened 2018-07-26
Remove some UB from our statement cache

sgrif:
From https://doc.rust-lang.org/beta/reference/behavior-considered-undefined.html

- Invalid values in primitive types, even in private fields and locals:
  - Dangling or null references and boxes.

This means that our "create an `&mut T` pointing to arbitrary memory"
strategy is UB, even if we never dereference it. Instead of trying to
lift the error out of `RefMut::map`, this new version instead calls the
function outside of `RefMut::map`, allowing us to just do normal error
handling. We then turn the result into a raw pointer to eliminate its
lifetime, so we can stick it back into a `RefMut` later.

This does create a pointer which alises a live `&mut` binding, but as
far as I can tell, this is only UB if we access both -- which we don't.

### DIFF
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -239,25 +239,17 @@ where
 ///
 /// If we were in Haskell (and if `RefMut` were a functor), this would just be
 /// `sequenceA`.
-fn refmut_map_result<T, U, F>(refmut: RefMut<T>, mapper: F) -> QueryResult<RefMut<U>>
+fn refmut_map_result<T, U, F>(mut refmut: RefMut<T>, mapper: F) -> QueryResult<RefMut<U>>
 where
     F: FnOnce(&mut T) -> QueryResult<&mut U>,
 {
-    use std::mem;
-
-    let mut error = None;
-    let result = RefMut::map(refmut, |mutref| match mapper(mutref) {
-        Ok(x) => x,
-        Err(e) => {
-            error = Some(e);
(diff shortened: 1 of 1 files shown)


sgrif:
Thanks to everyone in the Rust IRC for helping me find a solution to this that wasn't UB.

diesel/src/connection/statement_cache.rs
```diff
@@ -239,25 +239,17 @@ where
...
-        Some(e) => Err(e),
-        None => Ok(result),
-    }
+    // We can't just use `RefMut::map` here, since to lift the error out of that
+    // closure we'd need to return *something*.
+    //
+    // Instead we will very briefly convert to a raw pointer to eliminate
+    // lifetimes from the equation. Ultimately the cast is safe since the input
+    // and output lifetimes are identical. However, without the raw pointer
+    // we would have two live mutable references at the same time.
```
weiznich: Doesn't that comment mean in cleartext that this continues to be undefined behaviour because we have at one point in time 2 mutable pointers for a given memory location and the compiler assumes that there is on aliasing?

Or is the argument here that we have one mutable pointer/reference to the outer type (`RefMut`) and one to the new inner type `U`  and that that are two different memory locations?

sgrif: What's important here is that one is just a normal pointer, not a reference. Rust makes no guarantees about raw pointers. If aliasing here is UB then virtually every cast from reference to raw pointer is UB. What's important is that we never use the aliasing pointers in a way that actually violates aliasing rules. While both pointers exist at the same time, we never dereference more than one of them.

PR #1800, opened 2018-07-28
Regenerate schema.rs after running `diesel database reset`

dsander:
Especially in the early stages of a project it can be useful to modify
the database structure by changing old(er) migrations and recreate
everything with `diesel database reset`.

Updating the schema file saves one additional step and ensures it is
always up to date.

### DIFF
--- a/diesel_cli/tests/database_reset.rs
+++ b/diesel_cli/tests/database_reset.rs
@@ -170,3 +170,23 @@ fn reset_sanitize_database_name() {
         result.stdout()
     );
 }
+
+#[test]
+fn reset_updates_schema_if_config_present() {
+    let p = project("reset_updates_schema_if_config_present")
+        .folder("migrations")
+        .file(
+            "diesel.toml",
+            r#"
+[print_schema]
+file = "src/my_schema.rs"
+"#,
+        )
+        .build();
+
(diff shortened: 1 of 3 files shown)


diesel_cli/tests/database_reset.rs
```diff
@@ -170,3 +170,23 @@ fn reset_sanitize_database_name() {
...
 }
+
+#[test]
+fn reset_updates_schema_if_config_present() {
+    let p = project("reset_updates_schema_if_config_present")
+        .folder("migrations")
+        .file(
+            "diesel.toml",
+            r#"
+[print_schema]
```
sgrif: I don't think TOML cares about whitespace does it? Can we have this indented normally?

dsander: You are correct, I changed the indentation in `migration_run_updates_schema_if_config_present` as well.

djc:
Ping, please take another look as this makes for very surprising UX.

dsander:
@djc Would to expect `diesel database reset` not to update `schema.rs` if it changed?

sgrif:
Please rebase onto master

dsander:
@sgrif Done

PR #1801, opened 2018-07-29
Update code examples for inserting posts to use tuples

galuhsahid:
The previous insert examples in the Getting Started guide uses the `Insertable` trait to insert new posts. Now that we can insert new data using tuples, we should be using tuples since we already have the value for the title and body of our post.


sgrif:
Can you squash and rebase?

PR #1806, opened 2018-08-01
bump bigdecimal, num-bigint and num-integer

leodasvacas:


marioidival:
god

diesel/Cargo.toml
```diff
@@ -25,10 +25,10 @@ time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.7.0", optional = true, features = ["use_std"] }
 ipnetwork = { version = "0.12.2", optional = true }
-num-bigint = { version = "0.1.41", optional = true }
+num-bigint = { version = "0.2", optional = true }
```
sgrif: This is a breaking change. Are we unable to support both versions?

diesel/Cargo.toml
```diff
@@ -25,10 +25,10 @@ time = { version = "0.1", optional = true }
...
 num-traits = { version = "0.2", optional = true }
-num-integer = { version = "0.1.32", optional = true }
-bigdecimal = { version = ">= 0.0.10, < 0.0.13", optional = true }
+num-integer = { version = "0.1.39", optional = true }
```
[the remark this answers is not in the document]
sgrif: Assuming that we're able to support `0.1.32`, yes

diesel/Cargo.toml
sgrif: `> 0.1, <= 0.3`

leodasvacas:
@sgrif thanks for the review! Does it look correct now?

weiznich:
Someone needs to fix the ci. (We will not merge a PR that breaks the ci :wink:)

> Is there any reason that these dependencies have an upper bound on version selection?

We need a upper bound there because otherwise those dependencies could break diesel if they break the functions that are used inside of diesel. Using only `> some version` would allow cargo to pull in a new possible incompatible major version.

kestred:
@weiznich: please see https://github.com/diesel-rs/diesel/pull/1837 (tests finished successfully)

sorccu:
This issue can be closed since #1837 is the exact same thing and it's been merged.

weiznich:
As @sorccu pointed out this was already done by #1837.

PR #1803, opened 2018-07-31
Implement Unsigned<Tinyint> for Mysql

joshleeb:
Add the implementation for `Unsigned<Tinyint>` for Mysql.

Closes #1771

PR #1807, opened 2018-08-01
Rename Tinyint to TinyInt

joshleeb:
Rename `Tinyint` to `TinyInt` for consistent naming. 

Ref. #1803

### DIFF
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -44,7 +44,7 @@ pub struct Bool;
 ///
 /// This is only available on MySQL.
 /// Keep in mind that `infer_schema!` will see `TINYINT(1)` as `Bool`,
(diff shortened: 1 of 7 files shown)


diesel/src/sql_types/mod.rs
```diff
@@ -57,7 +57,7 @@ pub struct Bool;
 /// [i8]: https://doc.rust-lang.org/nightly/std/primitive.i8.html
 #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
 #[mysql_type = "Tiny"]
-pub struct Tinyint;
```
sgrif: We still need to keep this name around. We can't just make a major breaking change like this

joshleeb:
@sgrif updated the language for the changelog entry and added `#[doc(hidden)]` to the type alias.

PR #1812, opened 2018-08-06
AsChangeset for mut ref

nottxy:
AsChangeset for mut ref, so that it could work:

```rust
#[derive(AsChangeset)]
#[table_name = "datas"]
struct UpdateData<'a> {}

impl<'a> UpdateData<'a> {
    fn update(&mut self, conn: &PgConnection) -> Result<()>{
        self.field = some_mut_action(self.field);
        diesel::update(datas.find(self.id)).set(self).execute(conn)
    }
}
```


sgrif:
You can assign `&*self` instead. I don't see any need to implement this for mutable references in general.

PR #1816, opened 2018-08-08
Fix build

weiznich:
Fix the CI

(I don't know why it broke without any warning. Those errors are caused by [this commit](https://github.com/diesel-rs/diesel/commit/ed09d955770b4c37d217b1c69c3d50dfd881703f))

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -9,6 +9,7 @@ documentation = "https://diesel.rs/guides/getting-started"
 homepage = "https://diesel.rs"
 repository = "https://github.com/diesel-rs/diesel"
 keywords = ["diesel", "migrations", "cli"]
+autotests = false
 
 [[bin]]
 name = "diesel"
--- a/diesel_derives/Cargo.toml
+++ b/diesel_derives/Cargo.toml
@@ -7,6 +7,7 @@ description = "You should not use this crate directly, it is internal to Diesel.
 documentation = "https://diesel.rs/guides/"
 homepage = "https://diesel.rs"
 repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
+autotests = false
 
 [dependencies]
 syn = { version = "0.14.0", features = ["full", "fold"] }
--- a/diesel_migrations/migrations_internals/src/migration.rs
+++ b/diesel_migrations/migrations_internals/src/migration.rs
@@ -49,7 +49,7 @@ impl<'a> fmt::Display for MigrationFileName<'a> {
             None => return Err(fmt::Error),
             Some(v) => v.join(self.sql_file),
         };
-        f.write_str(fpath.to_str().unwrap())?;
+        f.write_str(fpath.to_str().unwrap_or("Invalid utf8 in filename"))?;
         Ok(())
     }
 }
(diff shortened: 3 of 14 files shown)


diesel_cli/Cargo.toml
```diff
@@ -9,6 +9,7 @@ documentation = "https://diesel.rs/guides/getting-started"
 homepage = "https://diesel.rs"
 repository = "https://github.com/diesel-rs/diesel"
 keywords = ["diesel", "migrations", "cli"]
+autotests = false
```
sgrif: Does this skip running unit tests?

diesel_derives/Cargo.toml
```diff
@@ -7,6 +7,7 @@ description = "You should not use this crate directly, it is internal to Diesel.
 documentation = "https://diesel.rs/guides/"
 homepage = "https://diesel.rs"
 repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
+autotests = false
```
sgrif: Does this skip running unit tests?

diesel_cli/Cargo.toml
weiznich: It does not change anything on the number of executed unit tests.

Help statement from cargo:

> warning: An explicit [[test]] section is specified in Cargo.toml which currently
disables Cargo from automatically inferring other test targets.
>This inference behavior will change in the Rust 2018 edition and the following
files will be included as a test target:
>
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/helpers.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/as_changeset.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/associations.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/queryable.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/identifiable.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/schema.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/queryable_by_name.rs
> * /home/weiznich/Dokumente/rust/diesel/diesel_derives/tests/insertable.rs
> This is likely to break cargo build or cargo test as these files may not be
ready to be compiled as a test target today. You can future-proof yourself
and disable this warning by adding `autotests = false` to your [package]
section. You may also move the files to a location where Cargo would not
automatically infer them to be a target, such as in subfolders.
> For more information on this warning you can consult
https://github.com/rust-lang/cargo/issues/5330

(The same applies for diesel_derive)

sgrif: Those are the integration tests. Are the tests in `src` still run (if there are any)?

sgrif: Also I don't think we need to handle warnings for the 2018 edition just yet -- I'm pretty sure we actually want to fix this the other way, and remove the `[[test]]` section (not 100% sure yet, waiting until 2018 is in beta to figure it out)

weiznich: See the linked Cargo issue. Setting `autotest = false` is basically just the current behaviour. This only disables the detection of integration tests in `tests/`. For this this means that we have only on entry point into our integration tests as is it currently defined in our `Cargo.toml` 
I do not see any downside of doing this now, this will just silence the warnings by explicitly opting in to the current default behaviour.

sgrif: Ok sounds good

PR #1818, opened 2018-08-09
Added single-element tuple example for select

WillSquire:


WillSquire:
Yeah sure, I'm easy :) . The most important part for me is a quick example just in case they're not 'au fait' with tuples. Would having something under the `select` documentation also be useful? (This was the first place I looked)

PR #1820, opened 2018-08-10
Add a `nullable` methods for `SelectStatement` and `BoxedSelectStatem…

weiznich:
…ent`

This allows to use a not nullable query in places that only accept
nullable expressions.

A notable use case is passing a subquery that selects a not nullable
primary key column as IN/eq_any filter to a nullable column.

Successor of #1644

### DIFF
--- a/diesel/src/query_builder/select_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_statement/dsl_impls.rs
@@ -19,7 +19,6 @@ use query_dsl::boxed_dsl::BoxedDsl;
 use query_dsl::methods::*;
 use query_dsl::*;
 use query_source::joins::{Join, JoinOn, JoinTo};
-use query_source::QuerySource;
 use sql_types::{BigInt, Bool};
 
 impl<F, S, D, W, O, L, Of, G, LC, Rhs, Kind, On> InternalJoinDsl<Rhs, Kind, On>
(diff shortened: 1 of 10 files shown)


sgrif:
Hm.. This seems more complicated than I had anticipated it being. I will poke at it more on Monday

weiznich:
@sgrif ping

diesel_tests/tests/filter.rs
```diff
@@ -439,6 +439,101 @@ fn filter_subselect_with_boxed_query() {
...
+            integer("home_world"),
+        ),
+    ).execute(&connection)
+        .unwrap();
+
+    connection
+        .execute("INSERT INTO home_worlds (name) VALUES ('Tatooine');")
+        .unwrap();
+    connection
+        .execute(
```
sgrif: Let's avoid using private API in new tests. `sql_query` if we must, but this can just use the normal insert API

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -470,3 +469,51 @@ where
         InsertFromSelect::new(self)
     }
 }
+
+impl<'a, F, S, D, W, O, L, Of, G> NullableSelectDsl
+    for SelectStatement<F, SelectClause<S>, D, W, O, L, Of, G>
+where
+    SelectClause<S>: SelectClauseExpression<F>,
+    <SelectClause<S> as SelectClauseExpression<F>>::SelectClauseSqlType: NotNull,
```
sgrif: I'm not clear why this is needed. We allow `.nullable` on expressions which are already nullable, and `Nullable<T>` works fine regardless of the SQL type for `T`. I'd rather be consistent and not have this bound.

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -366,7 +367,6 @@ where
         )
     }
 }
-
```
sgrif: Let's not delete whitespace :)

diesel_tests/tests/filter.rs
```diff
@@ -439,6 +439,111 @@ fn filter_subselect_with_boxed_query() {
     assert_eq!(expected, data);
 }
 
+#[test]
+fn filter_subselect_with_nullable_column() {
```
sgrif: I don't think we need this test if we have the doctest.

diesel/src/query_dsl/nullable_select_dsl.rs
```diff
@@ -0,0 +1,14 @@
+/// The `nullable` method
+///
+/// This trait should not be relied on directly by most apps. Its behavior is
+/// provided by [`QueryDsl`]. However you may need a where clause on this trait
+/// to call `nullable` from generic code.
+///
+/// [`QueryDsl`]: ../trait.QueryDsl.html#method.nullable
+pub trait NullableSelectDsl {
```
sgrif: WDYT about `SelectNullableDsl` instead?

PR #1830, opened 2018-08-29
Fix name of SQLite in-memory database name

djc:
The `:memory` name just opens a database on disk.


weiznich:
Rebasing the commit to the current master branch should fix the ci.

sgrif:
Merging with failing CI since this only affects docs (which we aren't building ATM)

PR #1833, opened 2018-09-04
[WIP] Migration to opaque PgValue struct for Diesel 2.0

Thomspoon:
Migrating the Pg connection to use an opaque struct rather than a slice for diesel dynamic schema handling. 

Currently broken.


diesel/src/macros/mod.rs
```diff
@@ -1055,6 +1055,19 @@ macro_rules! not_none {
     };
 }
 
+/// Gets the value out of an option, or returns an error.
+///
+/// This is used by `FromSql` implementations for Postgres PgValue.
+#[macro_export]
+macro_rules! not_none_pg {
```
sgrif: Hm... I was more envisioning `fn bytes(&self) -> &[u8]` on `PgValue`, and then we'd just do `not_none!(value.bytes())`

diesel/src/pg/types/array.rs
```diff
@@ -48,7 +48,10 @@ where
                 } else {
                     let (elem_bytes, new_bytes) = bytes.split_at(elem_size as usize);
                     bytes = new_bytes;
-                    T::from_sql(Some(elem_bytes))
+                    T::from_sql(Some(&PgValue {
+                        data: elem_bytes.to_vec(),
+                        oid: 1, //TODO FIXFIXFIX
```
sgrif: Oh this is interesting. We could probably just stick a reference to the metadata lookup on `PgValue` but I'm not sure I want to do that.

diesel/src/pg/types/date_and_time/mod.rs
```diff
@@ -143,18 +143,21 @@ impl ToSql<sql_types::Interval, Pg> for PgInterval {
...
-            ))),
-            days: try!(FromSql::<sql_types::Integer, Pg>::from_sql(Some(
-                &bytes[8..12]
-            ))),
-            months: try!(FromSql::<sql_types::Integer, Pg>::from_sql(Some(
-                &bytes[12..16]
-            ))),
+            microseconds: try!(FromSql::<sql_types::BigInt, Pg>::from_sql(Some(&PgValue {
+                data: bytes[..8].to_vec(),
+                oid: 1, // TODO: FIX
```
sgrif: In these cases we don't care since we know it's not looking at the oid. We should probably encode this somehow though if we don't make arbitrary lookup possible here.

diesel/src/pg/types/array.rs
sgrif: Actually no, we don't need to do any of that. `_oid` above is the right value.

diesel/src/pg/backend.rs
```diff
@@ -13,6 +13,11 @@ use sql_types::{Oid, TypeMetadata};
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct Pg;
 
+pub struct PgValue {
+    pub data: Vec<u8>,
```
sgrif: I'd like all these fields to be private. The struct should not be destructurable

diesel/src/pg/types/date_and_time/mod.rs
Thomspoon: I was worried about having an Enum for all the possible OIDs, but I'm not sure how feasible that is.

diesel/src/pg/types/array.rs
Thomspoon: Fixed

diesel/src/pg/backend.rs
Thomspoon: Fixed, added new method.

diesel/src/pg/types/date_and_time/mod.rs
sgrif: OIDs aren't fixed, which is why we have the dynamic lookup in the first place.

diesel/src/pg/backend.rs
```diff
@@ -13,6 +13,29 @@ use sql_types::{Oid, TypeMetadata};
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct Pg;
 
+#[derive(Debug, Clone)]
+pub struct PgValue {
+    data: Vec<u8>,
```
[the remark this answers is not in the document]
sgrif: Once https://github.com/rust-lang/rust/issues/48055 lands, we can just have `data: [u8]` as the final field of the struct. Until then, we'll probably need to use `*const [u8]`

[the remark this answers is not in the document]
weiznich: So as requested by @Thomspoon I've played a bit around with how to write such a struct. As Sean has written above, currently the probably easiest way is to use  `NonNull<[u8]>` as field here.

This would result in the following code:
```rust 
struct PgValue {
    raw_bytes: NonNull<[u8]>,
    oid: u32
}

impl PgValue {
    unsafe fn new(value_ptr: *mut u8, byte_count: usize) -> Self {
        let slice =  from_raw_parts_mut(value_ptr, byte_count);
        Self {
            raw_bytes: NonNull::new(slice).expect("cannot be null"),
        }
    }

    fn as_bytes(&self) -> &[u8] {
        unsafe { self.raw_bytes.as_ref() }
    }

    fn oid(&self) -> u32 {
        self.oid
    }
}
```

The new function should be only called in `PgResult::get`. For constructing `PgValue` from a other `PgValue` from inside a `from_sql` a additional function should be added.

diesel/src/deserialize.rs
```diff
@@ -166,7 +166,7 @@ where
 /// [`MysqlType`]: ../mysql/enum.MysqlType.html
 pub trait FromSql<A, DB: Backend>: Sized {
     /// See the trait documentation.
-    fn from_sql(bytes: Option<&DB::RawValue>) -> Result<Self>;
+    fn from_sql<'a>(bytes: Option<<DB::RawValue as FamilyLt<'a>>::Out>) -> Result<Self>;
```
weiznich: I think we should provide some type def for that type :wink:

```rust
type RawValue<'a, DB> = <DB::RawValue as FamilyLt<'a>>::Out;
```

Then use it as `RawValue<'a, DB>`.

diesel/src/backend.rs
```diff
@@ -6,6 +6,18 @@ use query_builder::bind_collector::BindCollector;
 use query_builder::QueryBuilder;
 use sql_types::{self, HasSqlType};
 
+use std::marker::PhantomData;
+
+pub trait FamilyLt<'a> {
```
weiznich: The trait and the type below should be `#[doc(hidden)]` to remove them from our public api.

diesel/src/mysql/types/numeric.rs
```diff
@@ -20,7 +20,7 @@ pub mod bigdecimal {
     }
 
     impl FromSql<Numeric, Mysql> for BigDecimal {
-        fn from_sql(bytes: Option<&<Mysql as Backend>::RawValue>) -> deserialize::Result<Self> {
+        fn from_sql<'a>(bytes: Option<<<Mysql as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich: Just use the actual type here:
```suggestion
        fn from_sql<'a>(bytes: Option<&'a [u8]>) -> deserialize::Result<Self> {
```

diesel/src/pg/backend.rs
```diff
@@ -9,10 +9,41 @@ use prelude::Queryable;
...
 
+use std::slice;
+use std::ptr::NonNull;
+
 /// The PostgreSQL backend
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct Pg;
 
+#[derive(Debug)]
+pub struct PgValue {
```
weiznich: Is there any reason why we don't use a slice here?
Beside of that: Oid's should be `u32` (http://docs.diesel.rs/diesel/pg/types/sql_types/struct.Oid.html)
So that `PgValue would look like this:

```rust
pub struct PgValue<'a> {
    bytes: &'a [u8],
    oid: u32,
}

pub struct PgValueProxy;

impl<'a> FamilyLt<'a> for PgValueProxy {
    type Out = PgValue<'a>;
}
```

diesel/src/pg/backend.rs
```diff
@@ -39,7 +70,7 @@ impl Queryable<(Oid, Oid), Pg> for PgTypeMetadata {
 impl Backend for Pg {
     type QueryBuilder = PgQueryBuilder;
     type BindCollector = RawBytesBindCollector<Pg>;
-    type RawValue = [u8];
+    type RawValue = RefFamily<PgValue>;
```
weiznich: Why this needs to be `RefFamily<PgValue>`? `PgValue` itself implements the necessary trait. (In your version, if the change above is applied, this should be `PgValueProxy`)

diesel/src/type_impls/floats.rs
```diff
@@ -1,25 +1,25 @@
...
-            bytes.len() <= 4,
-            "Received more than 4 bytes while decoding \
-             an f32. Was a double accidentally marked as float?"
-        );
-        bytes
-            .read_f32::<DB::ByteOrder>()
-            .map_err(|e| Box::new(e) as Box<Error + Send + Sync>)
-    }
-}
+// impl<DB: Backend<RawValue = [u8]>> FromSql<sql_types::Float, DB> for f32 {
```
weiznich: Those should be moved to pg/mysql.

diesel/src/pg/backend.rs
Thomspoon: Mostly because above we were discussing using NonNull :)

diesel/src/pg/backend.rs
```diff
@@ -9,10 +9,57 @@ use prelude::Queryable;
...
 
+/// The raw value representation of the postgres backend
+#[derive(Debug)]
+pub struct PgValue<'a> {
+    raw_bytes: &'a [u8],
+    oid: u32,
+}
+
+impl<'a> PgValue<'a> {
+    pub(crate) fn new(value_ptr: *mut u8, byte_count: usize) -> PgValue<'a> {
```
weiznich: Any reasons why we need a `*mut u8` and not a `*const u8` here?
```suggestion
    pub(crate) fn new(value_ptr: *const u8, byte_count: usize) -> PgValue<'a> {
```

diesel/src/pg/connection/row.rs
```diff
@@ -20,10 +20,10 @@ impl<'a> PgRow<'a> {
 }
 
 impl<'a> Row<Pg> for PgRow<'a> {
-    fn take(&mut self) -> Option<&[u8]> {
+    fn take(&mut self) -> Option<PgValue> {
         let current_idx = self.col_idx;
         self.col_idx += 1;
-        self.db_result.get(self.row_idx, current_idx)
+        self.db_result.get(self.row_idx, current_idx).map(|result| result)
```
weiznich: This map call seems to be redundant.

```suggestion
        self.db_result.get(self.row_idx, current_idx)
```

diesel/src/pg/connection/row.rs
```diff
@@ -43,7 +43,7 @@ impl<'a> PgNamedRow<'a> {
 }
 
 impl<'a> NamedRow<Pg> for PgNamedRow<'a> {
-    fn get_raw_value(&self, index: usize) -> Option<&[u8]> {
+    fn get_raw_value<'b>(&'b self, index: usize) -> Option<PgValue> {
```
weiznich: Either that lifetime is not required or the return value should use a explicit lifetime to make it easier to reason about which lifetime is used there.

diesel/src/pg/types/array.rs
```diff
@@ -48,7 +49,7 @@ where
                 } else {
                     let (elem_bytes, new_bytes) = bytes.split_at(elem_size as usize);
                     bytes = new_bytes;
-                    T::from_sql(Some(elem_bytes))
+                    T::from_sql(Some(PgValue::with_oid(elem_bytes.as_ptr() as *mut u8, elem_bytes.len(), oid as u32)))
```
weiznich: Maybe we should make the `PgValue::with_oid` constructor just taking a `&[u8]` slice and do the constructing of that slice only once in `pg/connection/cursor.rs`?

diesel/src/pg/types/date_and_time/mod.rs
```diff
@@ -143,17 +143,18 @@ impl ToSql<sql_types::Interval, Pg> for PgInterval {
...
 impl FromSql<sql_types::Interval, Pg> for PgInterval {
-    fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
-        let bytes = not_none!(bytes);
+    fn from_sql(value: Option<PgValue>) -> deserialize::Result<Self> {
+        let value = not_none!(value);
+        let bytes = value.bytes();
         Ok(PgInterval {
             microseconds: try!(FromSql::<sql_types::BigInt, Pg>::from_sql(Some(
-                &bytes[..8]
+                PgValue::with_oid(bytes[..8].as_ptr() as *mut u8, 8, 0) // TODO Find OID
```
weiznich: We need to find some way to get the oid from the `Pg: HasSqlType<BigInt>` implementation. That basically means we need to find some way to construct a fake `PgMetadataLookup` struct here. (The implementation will not use it internally because the oid is static for `BigInt`)

diesel/src/pg/types/integers.rs
```diff
@@ -1,14 +1,77 @@
...
+}
+
+impl FromSql<sql_types::Integer, Pg> for i32 {
+    fn from_sql(value: Option<PgValue>) -> deserialize::Result<Self> {
+        let value = not_none!(value);
+        let mut bytes = value.bytes();
+        debug_assert!(
+            bytes.len() <= 4,
+            "Received more than 4 bytes decoding i32. \
+             Was a BigInteger expression accidentally identified as Integer?"
```
weiznich: Time to fix those ancient error message
```suggestion
             Was a BigInt expression accidentally identified as Integer?"
```

diesel/src/pg/types/integers.rs
```diff
@@ -1,14 +1,77 @@
...
+        let mut bytes = value.bytes();        
+        debug_assert!(
+            bytes.len() <= 8,
+            "Received more than 8 bytes decoding i64. \
+             Was an expression of a different type misidentified as BigInteger?"
+        );
+        debug_assert!(
+            bytes.len() >= 8,
+            "Received fewer than 8 bytes decoding i64. \
+             Was an Integer expression misidentified as BigInteger?"
```
weiznich:
```suggestion
             Was an Integer expression misidentified as BigInt?"
```

diesel/src/pg/types/primitives.rs
```diff
@@ -1,14 +1,28 @@
...
 use serialize::{self, IsNull, Output, ToSql};
 use sql_types;
 
+/// The returned pointer is *only* valid for the lifetime to the argument of
+/// `from_sql`. This impl is intended for uses where you want to write a new
+/// impl in terms of `String`, but don't want to allocate. We have to return a
+/// raw pointer instead of a reference with a lifetime due to the structure of
+/// `FromSql`
+impl FromSql<sql_types::Text, Pg> for *const str {
+    fn from_sql<'a>(bytes: Option<PgValue>) -> deserialize::Result<Self> {
```
weiznich: Remove the unused lifetime
```suggestion
    fn from_sql(bytes: Option<PgValue>) -> deserialize::Result<Self> {
```

diesel/src/pg/types/ranges.rs
```diff
@@ -88,7 +89,9 @@ where
             let elem_size = bytes.read_i32::<NetworkEndian>()?;
             let (elem_bytes, new_bytes) = bytes.split_at(elem_size as usize);
             bytes = new_bytes;
-            let value = T::from_sql(Some(elem_bytes))?;
+            let value = T::from_sql(Some(
+                PgValue::with_oid(elem_bytes.as_ptr() as *mut u8, 1, 0) // TODO Find OID
```
weiznich: Basically the same as for `PgInterval` implementation above: We need to find a way to use `HasSqlType::metadata` here.

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -12,7 +12,7 @@ use sqlite::Sqlite;
 const SQLITE_DATE_FORMAT: &str = "%F";
 
 impl FromSql<Date, Sqlite> for NaiveDate {
-    fn from_sql(value: Option<&<Sqlite as Backend>::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(value: Option<<<Sqlite as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich: Just use the actual RawValue type
```suggestion
    fn from_sql(value: Option<&SqliteValue>) -> deserialize::Result<Self> {
```

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -27,7 +27,7 @@ impl ToSql<Date, Sqlite> for NaiveDate {
 }
 
 impl FromSql<Time, Sqlite> for NaiveTime {
-    fn from_sql(value: Option<&<Sqlite as Backend>::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(value: Option<<<Sqlite as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(value: Option<&SqliteValue>) -> deserialize::Result<Self> {
```

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -54,7 +54,7 @@ impl ToSql<Time, Sqlite> for NaiveTime {
 }
 
 impl FromSql<Timestamp, Sqlite> for NaiveDateTime {
-    fn from_sql(value: Option<&<Sqlite as Backend>::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(value: Option<<<Sqlite as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(value: Option<&SqliteValue>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/floats.rs
```diff
@@ -1,26 +1,12 @@
-use byteorder::{ReadBytesExt, WriteBytesExt};
+use byteorder::{WriteBytesExt}; // ReadBytesExt
 use std::error::Error;
 use std::io::prelude::*;
 
 use backend::Backend;
-use deserialize::{self, FromSql};
+//use deserialize::{self, FromSql};
```
weiznich:
```suggestion
```

diesel/src/type_impls/floats.rs
```diff
@@ -1,26 +1,12 @@
-use byteorder::{ReadBytesExt, WriteBytesExt};
+use byteorder::{WriteBytesExt}; // ReadBytesExt
```
weiznich:
```suggestion
use byteorder::WriteBytesExt;

File: diesel/src/type_impls/integers.rs [hunk 26]
```diff
@@ -1,32 +1,11 @@
-use byteorder::{ReadBytesExt, WriteBytesExt};
+use byteorder::{WriteBytesExt}; // ReadBytesExt
```
weiznich:
```suggestion
use byteorder::WriteBytesExt; 
```

diesel/src/type_impls/integers.rs
```diff
@@ -1,32 +1,11 @@
...
 use std::error::Error;
 use std::io::prelude::*;
 
-use backend::Backend;
-use deserialize::{self, FromSql};
+use backend::{Backend};
```
weiznich:
```suggestion
use backend::Backend;
```

diesel/src/type_impls/option.rs
```diff
@@ -52,7 +52,7 @@ where
     DB: Backend,
     ST: NotNull,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<<<DB as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -116,27 +116,14 @@ where
     DB: Backend,
     *const str: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<<<DB as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -160,25 +147,14 @@ where
     DB: Backend,
     *const [u8]: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<<<DB as Backend>::RawValue as FamilyLt<'a>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -216,7 +192,7 @@ where
     DB: Backend,
     T::Owned: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'b>(bytes: Option<<<DB as Backend>::RawValue as FamilyLt<'b>>::Out>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql<'b>(bytes: Option<RawValue<'b, DB>>) -> deserialize::Result<Self> {
```

diesel/src/pg/connection/result.rs
```diff
@@ -69,7 +71,7 @@ impl PgResult {
         PgRow::new(self, idx)
     }
 
-    pub fn get(&self, row_idx: usize, col_idx: usize) -> Option<&[u8]> {
+    pub fn get(&self, row_idx: usize, col_idx: usize) -> Option<PgValue> {
```
weiznich: We need to load the oid for each type here somehow.
Postgresql provided `PQftype(PqResult*, usize*)` for this.
I'm not sure if that does a additional client->server round trip or just reads the value from same internal libpq field. In the second case we can just call `PQftype` here, otherwise this should be done while constructing `PQResult` above.
cc @sgrif because I'm not sure how libpq works internally

diesel/src/type_impls/primitives.rs
Thomspoon: This doesn't seem to be a valid change because the expected type is:
fn(std::option::Option<<::RawValue as backend::FamilyLt<'a>>::Out>)

diesel/src/type_impls/primitives.rs
Thomspoon: This doesn't seem to be a valid change because the expected type is:
fn(std::option::Option<<::RawValue as backend::FamilyLt<'a>>::Out>)

diesel/src/type_impls/primitives.rs
Thomspoon: This doesn't seem to be a valid change because the expected type is:
fn(std::option::Option<<::RawValue as backend::FamilyLt<'a>>::Out>)

diesel/src/pg/connection/result.rs
Thomspoon: I'll test the ftype call and see if I get valid OIDs out of it!

diesel/src/pg/backend.rs
```diff
@@ -48,7 +47,7 @@ impl<'a> PgValue<'a> {
 
     /// Get the type oid of the value represented by this raw value
     pub fn oid(&self) -> u32 {
```
weiznich: I think just returning `Option<NonZeroU32>` here is the better solution, so we can handle that at the calling side, because it may be ok to have no oid.

diesel/src/mysql/types/mod.rs
```diff
@@ -25,6 +34,65 @@ impl FromSql<TinyInt, Mysql> for i8 {
...
+    }
+}
+
+impl FromSql<BigInt, Mysql> for i64 {
+    fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
+        let mut bytes = not_none!(bytes);
+        debug_assert!(
+            bytes.len() <= 8,
+            "Received more than 8 bytes decoding i64. \
+             Was an expression of a different type misidentified as BigInteger?"
```
weiznich:
```suggestion
             Was an expression of a different type misidentified as BigInt?"
```

diesel/src/mysql/types/mod.rs
```diff
@@ -25,6 +34,65 @@ impl FromSql<TinyInt, Mysql> for i8 {
...
+        );
+        debug_assert!(
+            bytes.len() >= 8,
+            "Received fewer than 8 bytes decoding i64. \
+             Was an Integer expression misidentified as BigInteger?"
```
weiznich:
```suggestion
             Was an Integer expression misidentified as BigInt?"
```

diesel/src/pg/backend.rs
weiznich:
```suggestion
    pub fn oid(&self) -> Option<NonZeroU32> {
```

diesel/src/pg/types/date_and_time/mod.rs
```diff
@@ -140,12 +140,19 @@ impl ToSql<sql_types::Interval, Pg> for PgInterval {
...
-        let bytes = not_none!(bytes);
+    fn from_sql(value: Option<PgValue>) -> deserialize::Result<Self> {
+        let value = not_none!(value);
+        let bytes = value.bytes();
         Ok(PgInterval {
-            microseconds: FromSql::<sql_types::BigInt, Pg>::from_sql(Some(&bytes[..8]))?,
-            days: FromSql::<sql_types::Integer, Pg>::from_sql(Some(&bytes[8..12]))?,
-            months: FromSql::<sql_types::Integer, Pg>::from_sql(Some(&bytes[12..16]))?,
+            microseconds: FromSql::<sql_types::BigInt, Pg>::from_sql(Some(
+                PgValue::with_oid(&bytes[..8], None) // TODO Find OID
```
weiznich: We should set an oid here. Basically we need to find a way to call `HasSqlType::metadata` for types with a static oid.

diesel/src/pg/types/integers.rs
```diff
@@ -1,14 +1,77 @@
...
+}
+
+impl FromSql<sql_types::BigInt, Pg> for i64 {
+    fn from_sql(value: Option<PgValue>) -> deserialize::Result<Self> {
+        let value = not_none!(value);
+        let mut bytes = value.bytes();        
+        debug_assert!(
+            bytes.len() <= 8,
+            "Received more than 8 bytes decoding i64. \
+             Was an expression of a different type misidentified as BigInteger?"
```
weiznich:
```suggestion
             Was an expression of a different type misidentified as BigInt?"
```

diesel/src/row.rs
```diff
@@ -11,7 +11,7 @@ use deserialize::{self, FromSql};
 /// [`FromSqlRow`]: ../deserialize/trait.FromSqlRow.html
 pub trait Row<DB: Backend> {
     /// Returns the value of the next column in the row.
-    fn take(&mut self) -> Option<&DB::RawValue>;
+    fn take<'a>(&'a mut self) -> Option<RawValue<'a, DB>>;
```
weiznich: That lifetime should not be required
```suggestion
    fn take(&mut self) -> Option<RawValue<DB>>;
```

diesel/src/row.rs
```diff
@@ -62,5 +62,5 @@ pub trait NamedRow<DB: Backend> {
     #[doc(hidden)]
     fn index_of(&self, column_name: &str) -> Option<usize>;
     #[doc(hidden)]
-    fn get_raw_value(&self, index: usize) -> Option<&DB::RawValue>;
+    fn get_raw_value<'a>(&'a self, index: usize) -> Option<RawValue<'a, DB>>;
```
weiznich:
```suggestion
    fn get_raw_value(&self, index: usize) -> Option<RawValue<DB>>;
```

diesel/src/sqlite/connection/sqlite_value.rs
```diff
@@ -115,7 +115,7 @@ impl<'a> NamedRow<Sqlite> for SqliteNamedRow<'a> {
         self.column_indices.get(column_name).cloned()
     }
 
-    fn get_raw_value(&self, idx: usize) -> Option<&SqliteValue> {
+    fn get_raw_value<'b>(&'b self, idx: usize) -> Option<&'b SqliteValue> {
```
weiznich:
```suggestion
    fn get_raw_value(&self, idx: usize) -> Option<&SqliteValue> {
```

diesel/src/type_impls/option.rs
```diff
@@ -52,7 +52,7 @@ where
     DB: Backend,
     ST: NotNull,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(bytes: Option<RawValue<DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -107,27 +107,14 @@ where
     DB: Backend,
     *const str: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(bytes: Option<RawValue<DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -151,25 +138,14 @@ where
     DB: Backend,
     *const [u8]: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'a>(bytes: Option<RawValue<'a, DB>>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(bytes: Option<RawValue<DB>>) -> deserialize::Result<Self> {
```

diesel/src/type_impls/primitives.rs
```diff
@@ -207,7 +183,7 @@ where
     DB: Backend,
     T::Owned: FromSql<ST, DB>,
 {
-    fn from_sql(bytes: Option<&DB::RawValue>) -> deserialize::Result<Self> {
+    fn from_sql<'b>(bytes: Option<RawValue<'b, DB>>) -> deserialize::Result<Self> {
```
weiznich:
```suggestion
    fn from_sql(bytes: Option<RawValue<DB>>) -> deserialize::Result<Self> {
```

diesel/src/pg/backend.rs
```diff
@@ -17,7 +17,8 @@ use std::num::NonZeroU32;
 pub struct Pg;
 
 /// The raw value representation of the postgres backend
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug, SqlType)]
+#[postgres(oid = "0", array_oid = "0")]
```
weiznich: That should not be added.

Thomspoon: removed...

diesel/src/pg/types/date_and_time/mod.rs
Thomspoon: @weiznich It looks like the impl for PgMetadataLoop and PgTypeMetadata both query pg_type, which wouldn't be performant. What does the PGftype call in pg/connection/result.rs do that requires us to redefine the OID later?

diesel/src/mysql/connection/stmt/iterator.rs
```diff
@@ -120,7 +120,7 @@ impl<'a> NamedRow<Mysql> for NamedMysqlRow<'a> {
         self.column_indices.get(column_name).cloned()
     }
 
-    fn get_raw_value(&self, idx: usize) -> Option<&[u8]> {
+    fn get_raw_value<'b>(&'b self, idx: usize) -> Option<&'b [u8]> {
```
sgrif: This lifetime should be elided

sgrif:
I'd like some more time to review this in depth, but at first glance I'm not sure how I feel about the approach taken to get around the lack of GATs. I really don't like that we've removed the ability to say `<DB as Backend>::RawValue`. We've generally tried to encourage folks to use that vs explicit types, as it makes their code less likely to break if the type changes in the future. This ends up breaking anybody using that form.

That said, this definitely may be the best/only approach, but I'd like some time to look at it more closely.

sgrif:
So it's clear to me that keeping `&DB::RawValue` around is possible, but not something we want to do in the long run. 2.0 seems like a reasonable time to change to `backend::RawValue<DB>` instead.(note: I do think including the module should be encouraged by example, similar to `serialize::Input`)

That said, this whole structure of `RefFamily<'a>` is incredibly opaque to me, and I think it's overly abstracted from what we're trying to do. The helper type should be a forwards compatibility shim for the eventual switch to GATs at most, not required to even understand what this is doing. I've opened #1976 as an alternative to that piece of this PR, which can be independently merged into master (and will hopefully make this a little smaller for review purposes if we go that direction)

weiznich:
Closed by #2134

PR #1835, opened 2018-09-04
Allow Pg ilike to be used on Nullable<Text>

carlosdp:


carlosdp:
It looks like the fmt is running on 1.26 but I'm on the latest stable 1.28, and maybe they format differently?

weiznich:
You need to use rustfmt from rust 1.26.1.

carlosdp:
Gotchya, should be good to go now @weiznich

PR #1837, opened 2018-09-09
Bump bigdecimal and num-* crate versions

kestred:
This is a reimplementation of https://github.com/diesel-rs/diesel/pull/1806
with an additional change to fix the failing test case on nightly.


diesel/src/pg/types/numeric.rs
```diff
@@ -94,7 +94,10 @@ mod bigdecimal {
 
             // Ensure that the decimal will always lie on a digit boundary
             for _ in 0..(4 - scale % 4) {
-                integer = integer * 10;
+                // NOTE: Assignment to `mul` is intended to avoid `#[deny(assign_op_pattern)]` on nightly
+                //       despite lack of `std::ops::MulAssign` impl for `BigInt`.
```
sgrif: Can we just add an `allow` for that instead?

sgrif: Either way, this seems unrelated to your change, and should probably be done in a PR that bumps clippy, not here

kestred:
@sgrif: now using allow rather than changing the syntax.

I'd intuitively guess that newer versions of clippy would not have a fix for this ? (unless they've removed the lint entirely). The `allow` change is related to code affected by the version bump, so I'd strongly prefer merging this PR as is over blocking on another clippy-related PR.

weiznich:
Could we move the clippy attribute to the affected function, so that we do not apply it to the whole module?

PR #1844, opened 2018-09-11
Add zsh completions

kpcyrd:
There are currently only bash completions available, this adds a subcommand to generate zsh completions as well.

### DIFF
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -1,4 +1,4 @@
-use clap::{App, AppSettings, Arg, SubCommand};
+use clap::{App, AppSettings, Arg, SubCommand, Shell};
 
 pub fn build_cli() -> App<'static, 'static> {
     let database_arg = Arg::with_name("DATABASE_URL")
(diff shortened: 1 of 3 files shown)


kpcyrd:
@weiznich I've changed this to a subcommand so `diesel completions $x` accepts bash, fish and zsh.

diesel_cli/src/cli.rs
```diff
@@ -90,8 +90,18 @@ pub fn build_cli() -> App<'static, 'static> {
         )
         .setting(AppSettings::SubcommandRequiredElseHelp);
 
-    let generate_bash_completion_subcommand = SubCommand::with_name("bash-completion")
-        .about("Generate bash completion script for the diesel command.");
+    let generate_completions_subcommand = SubCommand::with_name("completions")
```
weiznich: What about making the shell variants just arguments?
Something like this:

```rust
SubCommand::with_name("completions")
    .about("Generate shell completion scripts for the diesel command.")
    .arg(Arg::with_name("SHELL")
        .index(1)
        .required(true)
        .possible_values(&Shell::variants())
    .setting(AppSettings::SubcommandRequiredElseHelp);

```

This would simplify the implementation and automatically support all shells that are added in a future version of clap.

kpcyrd: I somehow missed that function, I'm now using your snippet and also found the `value_t!` macro, this removes every reference to bash or zsh and generates completions for everything supported in clap.

sgrif: Removing the `bash-completion` command is a breaking change, and shouldn't be done in a 1.x version. We need to keep that command and print a deprecation warning.

kpcyrd: @sgrif I've re-introduced the command and also added a deprecation warning to stderr.

weiznich:
You need to use rustfmt from rust 1.26.1.

kpcyrd:
@weiznich I've pushed a commit to correct the rustfmt issue.

Issue #1845, opened 2018-09-11
Add a function to embedded_migrations! to check if migration is needed


kpcyrd:
It seems right now two functions are exported in the module that is generated by `embedded_migrations!`:

- `run`
- `run_with_output`

I would like to have a 3rd function, along the lines of `fn migrations_needed() -> bool`, that can be used to check if migrations are available/needed.

The idea is that this function can be used to determine if `embedded_migrations::run` needs to be executed and also to inform the user that migrations are going to be executed:

```rust
if embedded_migrations::migrations_needed() {
    println!("executing migrations);
    embedded_migrations::run(&db)?;
}
```

sgrif:
Thank you for the feature request. We're currently in the process of a cleanup of the issue tracker, and aren't currently accepting feature requests (there will be an official policy written soon, but the TL;DR is that open issues should reflect something that is a bug or on our immediate roadmap).

We're happy to discuss feature requests, but the place to do so is discourse.diesel.rs, not the issue tracker.

PR #1846, opened 2018-09-12
Add any_pending_migrations to `embed_migrations!`

kpcyrd:
Implements #1845

It turned out the logic already exists, but hasn't been exposed for the `embed_migrations!` macro yet.

```rust
if embedded_migrations::any_pending_migrations()? {
    println!("executing migrations);
    embedded_migrations::run(&db)?;
}
```

### DIFF
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -190,7 +190,10 @@ where
 ///
 /// See the [module level documentation](index.html) for information on how migrations should be
 /// structured, and where Diesel will look for them by default.
-pub fn any_pending_migrations_from<Conn, List>(conn: &Conn, migrations: List) -> Result<bool, RunMigrationsError>
+pub fn any_pending_migrations_from<Conn, List>(
+    conn: &Conn,
+    migrations: List,
+) -> Result<bool, RunMigrationsError>
 where
     Conn: MigrationConnection,
     List: IntoIterator,
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -169,8 +169,8 @@ where
     Ok(migrations)
 }
 
-// Returns true if there are outstanding migrations in the migrations directory, otherwise
-// returns false. Returns an `Err` if there are problems with migration setup.
+/// Returns true if there are outstanding migrations in the migrations directory, otherwise
+/// returns false. Returns an `Err` if there are problems with migration setup.
 ///
 /// See the [module level documentation](index.html) for information on how migrations should be
 /// structured, and where Diesel will look for them by default.
@@ -181,9 +181,24 @@ where
     let migrations_dir = find_migrations_directory()?;
     let all_migrations = migrations_in_directory(&migrations_dir)?;
     setup_database(conn)?;
+
+    any_pending_migrations_from(conn, all_migrations)
+}
+
+/// Returns true if there are outstanding migrations in the list provided to the function,
+/// otherwise returns false. Returns an `Err` if there are problems with migration setup.
+///
+/// See the [module level documentation](index.html) for information on how migrations should be
+/// structured, and where Diesel will look for them by default.
+pub fn any_pending_migrations_from<Conn, List>(conn: &Conn, migrations: List) -> Result<bool, RunMigrationsError>
+where
+    Conn: MigrationConnection,
+    List: IntoIterator,
+    List::Item: Migration,
+{
     let already_run = conn.previously_run_migration_versions()?;
 
-    let pending = all_migrations
+    let pending = migrations
         .into_iter()
         .any(|m| !already_run.contains(&m.version().to_string()));
 
--- a/diesel_migrations/migrations_macros/src/embed_migrations.rs
+++ b/diesel_migrations/migrations_macros/src/embed_migrations.rs
@@ -50,6 +50,10 @@ pub fn derive_embed_migrations(input: &syn::DeriveInput) -> proc_macro2::TokenSt
                 unreachable!()
             }
         }
+
+        pub fn any_pending_migrations<C: MigrationConnection>(conn: &C) -> Result<bool, RunMigrationsError> {
+            any_pending_migrations_from(conn, ALL_MIGRATIONS.iter().map(|v| *v))
+        }
     );
 
     let run_fns = quote!(
--- a/diesel_migrations/src/lib.rs
+++ b/diesel_migrations/src/lib.rs
@@ -83,6 +83,8 @@ extern crate migrations_macros;
 #[doc(inline)]
 pub use migrations_internals::any_pending_migrations;
 #[doc(inline)]
+pub use migrations_internals::any_pending_migrations_from;
+#[doc(inline)]
 pub use migrations_internals::find_migrations_directory;
 #[doc(inline)]
 pub use migrations_internals::mark_migrations_in_directory;


dmcguire81:
@kpcyrd doesn't look like your `rustfmt` patch succeeded:

```
$ cargo fmt --all -- --write-mode=diff
Diff in /home/travis/build/diesel-rs/diesel/diesel_migrations/migrations_internals/src/lib.rs at line 183:
 /// structured, and where Diesel will look for them by default.
 pub fn any_pending_migrations_from<Conn, List>(
     conn: &Conn,
-    migrations: List
+    migrations: List,
 ) -> Result<bool, RunMigrationsError>
 where
     Conn: MigrationConnection,
The command "cargo fmt --all -- --write-mode=diff" exited with 4.
```

I'm assuming that's why there are no takers on reviewing it.

dmcguire81:
@weiznich can you confirm that a single missing comma is holding this up?

kpcyrd:
@dmcguire81 I somehow missed that, thanks for pointing it out. rustfmt should pass with the patch I just pushed.

Boscop:
Any update on this? :)

weiznich:
I'm not sure I understand why this is necessary. `embedded_migrations::run(&db)?;` internally checks already if any migration needs to run.

kpcyrd:
My example, was probably confusing, I'm more interested in the `println!` part:
```rust
if embedded_migrations::any_pending_migrations()? {
    println!("executing migrations);
}
embedded_migrations::run(&db)?;
```
Having access to these things allows me to write a better user interface. @Boscop may have a similar usecase.

peterschwarz:
This would be great to see in the library, as it enables several use-cases. 

- Confirmation of running migrations
- Running migrations in a separate CLI, and verifying they have been run by a daemon process.

The latter case supports the use-case where migrations can be run via a sysadmin/dba with permissions to create/destroy tables, but the application runs with a user that only has read/write permissions for the tables .

veeg:
This functionality is indeed useful as a standalone method for `embedded_migrations`, mostly due to the fact that I would like to deny further processing in my application if the database has yet to be (manually) migrated.

peterschwarz:
Looks like there are some conflicts that need to be resolved - any possibility of more action on this @kpcyrd ?

disconsented:
> I'm not sure I understand why this is necessary. `embedded_migrations::run(&db)?;` internally checks already if any migration needs to run.

I've just been dealing with a power loss during a migration scenario using Sqlite which resulted in database corruption, to avoid this in the future we run migrations on a working copy and then replace the old database on success.

Without the ability to check whether we need to run migrations, we need to create the working copy each time which increases startup times and decreases NAND durability unnecessarily.

weiznich:
I'm probably willing to accept such a change if there is at least:
* A complete documentation why this can be useful
* A changelog entry
* A test case 
* Passing CI
* No merge conflict

disconsented:
Can you please expand upon exactly what you'd like for "A complete documentation why this can be useful"?

kpcyrd:
@weiznich would you accept a patch that would instead expose functions necessary to build a custom progress bar? Something like:

```
let n = embedded_migrations::number_of_missing_transactions()?;
if n > 0 {
    for i in 0..n {
        println!("Applying migration ({}/{})", i+1, n);
        embedded_migrations::run_next(&db)?;
    }
} else {
    println!("Database already up-to-date!");
}
```

weiznich:
@disconsented 
> Can you please expand upon exactly what you'd like for "A complete documentation why this can be useful"?

The current PR description does only briefly mention potential use cases. Also I feel that all use cases listed in the comments, are quite incomplete. I can understand why this feature would be interesting but I would like to have a concrete written down example why this must be implemented that way and not some other way.

@kpcyrd Having tough a bit more about this, I came to the conclusion that the probably best solution for this and similar problems would be to just make the `ALL_MIGRATIONS` constant part of our public API. Given that all this functionality could be implemented in `diesel_migrations` itself as function not specific to `embed_migration!`, which also would make all that much more discoverable in my opinion. I think I would be open for a PR doing that instead.

genbattle:
I'm working with @disconsented on this. While I respect the need to maintain a tidy API, why can this sort of refactoring not be done later as an additional PR, instead of requiring all of the work up-front and further delaying a change that would satisfy affected users in the meantime?

weiznich:
@genbattle There are multiple reasons for that:

* Removing an API which is added "just" as quick fix is really hard. That does require releasing a new major diesel version, which is something I rather would not like to do often.
* Documenting the currently proposed extensions in a discoverable manner is not really possible in my opinion, as all of those methods would not be part of any diesel crate, but rather generated by some opaque proc-macro. That makes it hard to discover them easily and at least for me that feels like to much magic. If possible we should try to hide as less information behind a proc-macro as possible.
* I feel that the solution proposed I've proposed above is easier to implement and understand than adding a bunch of semi-hidden functions to the output generated by `embed_migration!`

Additionally I do my work on diesel in my free time and share that for free with you. So maybe requesting "requiring all of the work up-front and further delaying a change that would satisfy affected users in the meantime?" is not the right way to convince me that that's the right way to implement such an API, because for me that sounds like: You need that for your potential commercial use case, you want to have a quick fix for that and rather then spending the time yourself to implement a sustainable solution yourself you want to outsource that for free to the diesel maintainers. That's not something I'm willing to accept here.

disconsented:
I think there has been a bit of a miscommunication here, I have **already** spent a couple of hours working on updating the original work (with credit to kpcyrd), so we're not trying to "outsource that for free to the diesel maintainers". Which is why I reached out to you on gitter with a question about how much leeway I'd have with tests because I wanted to write something a bit more comprehensive. 

Your reply makes it seem like you didn't read the question either.

I won’t comment on the first however I disagree on the second, proc macros in general make discovery harder but that cost was already willingly paid about 7 months ago adding in an example in the docs isn't any worse.

As for the third, I want to reiterate that the code in this PR still works rewriting code is harder than not rewriting code.

Not sure how this could be easier than using the existing 27 lines, we're happy to spend the time doing this, but we're hesitant to spend a couple of weeks going through the entire process.

weiznich:
@disconsented I think I should clarify a few points about the general process here:

1. At least till last week I would consider this PR have going stale. It was not something there was active work on + there was no indication that someone from the contributor team would be willing to merge it as it was or even that the API was something we would like to have.
2. You comment 6 days ago put that PR back to my notification. You've given a good example why something like this can be useful, so that's a totally fine way to do this.
3. I had another look at the change and figured out that this would be OK given the points I wrote up above.
4. You've asked a question about some detail of my points which should be clarified. That's also totally fine.
5. Then the original author of this PR stepped in and asked for yet another slightly different API, at this point it is clear for me that there is more than one uncovered use case here and we should have a discussion about the exact API. In my understanding it's my job as maintainer to sit down an think about such cases then and propose a solution that is maintainable, opens up the minimal required API and is generally discoverable. In this case that seems quite easy for me, so I've proposed a different API quite soon. At that point I consider that proposal open for discussion and wait for arguments made that this is the acceptable approach or not.
6. At that point you've asked questions about how to test the old proposed solution in our gitter channel, where I've answered that I would rather like to go with my new proposal for the reasons listed in the corresponding comment here. It's fine to ask that question, but after pointing out that there is a new proposal I would at least expect that you put in some argument why you would like to go with the old variant. (Note: I do not really consider: "But I've already started with implementing that" a valid argument here, because at this point that was not clear for me or anyone else in this thread.)
7. Someone unrelated starts to comment in this thread requesting in quite hard words that the original solution is implemented as "quick way" to "satisfy users". That's probably not the right way to give a technical argument.


May main points here are:

* The discussion on this feature restarted less then a week before. 
* The original author expressed that they are interested in a other slightly different API. That means the proposed API is not set in stone yet.
* You basically expect that I'm willing to decide on an API in that time, while I'm only working on and of in my free time on this project. That means at least for me it's totally OK to not comment for a few days on something, because I'm occupied with other things like weekend, my paid work and so on. If that's not a acceptable level of support for you, you need find a way to pay someone for providing a support in a more timely manner.

Additionally I do not see any concrete plans to put out a release as soon as some potential PR is merged here. Therefore I do not see any reason to rush the API design, to have this done in less then a week or so.

PR #1850, opened 2018-09-12
Add examples for typical trivial `ToSql`/`FromSql` impls

### LINKED ISSUE #1783: Add an example of FromSql/ToSql for custom types
I've spent a few hours trying to understand how to use custom types instead of one from `sql_types`, but not results.

Let's say I have a `SmallInt` column that I want to represent as `enum`. How can I do this?

sgrif:
We point people to a test case way too frequently, and should have some
example impls in our code (especially for the simpler case of "I'm using
an integer/string/whatever column and not creating a custom SQL type")

Fixes #1783.

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -164,6 +164,38 @@ where
 /// - For third party backends, consult that backend's documentation.
 ///
 /// [`MysqlType`]: ../mysql/enum.MysqlType.html
+///
+/// ### Examples
+///
+/// Most implementations of this trait will be defined in terms of an existing
+/// implementation.
+///
+/// ```rust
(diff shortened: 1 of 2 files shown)


sgrif:
> These changes are looking fine on themself, but adding those two impl will not allow a user to really use the enum as custom type, because it is missing the derive of `FromSqlRow` and `AsExpression`.

Yes, I tried to keep this focused on specifically the impls for the traits being documented, rather than discussion different levels of abstractions here. It seems like folks generally understand that you need `#[derive(AsExpression, FromSqlRow)]`, but get hung up on how to implement these two traits.

> Maybe we should add such an example also to the guide as part of the Extending Diesel section?

Probably a separate guide, but yes.

PR #1851, opened 2018-09-12
Refactor our handling of nested options

sgrif:
Today our code for nested options assumes that all nested options are
required, and that option not being there is an error. This is probably
because when I wrote that function, it was for
`#[changeset_options(treat_none_as_null = "true")]`, which is the only
case where a nested option is required...

All other uses of this function immediately called `.ok` on it, meaning
if the error came from the meta item being malformed rather than the
nested item not being there, we either gave a suboptimal error message,
or ignored the error entirely.

For example, if you wrote `#[postgres = "Foo"]` when you meant
`#[postgres(sql_type = "Foo")]`, you'd get an error saying "missing
required options" rather than one telling you that `#[postgres]` was
malformed.

This forces callers to handle the error from malformed syntax, and
allows future features to distinguish from "the place we looked is
malformed" from "it's not there".


diesel_compile_tests/tests/ui/sql_type_bad_options.stderr
```diff
@@ -36,17 +34,17 @@ warning: Option ary_oid has no effect
...
-error: Missing required option array_oid
+error: Missing required option `array_oid`
   --> $DIR/sql_type_bad_options.rs:21:3
    |
 21 | #[postgres(oid = "NaN", ary_oid = "1")]
    |   ^^^^^^^^
 
-error: Expected a number
-  --> $DIR/sql_type_bad_options.rs:21:18
+error: `postgres` must be in the form `postgres(...)`
```
weiznich: Could we provide something better here. Like this: `postgres` must be in the form `postgres(type_name = "…")` or `postgres(oid = "…", array_oid = "…")`?

Eijebong: Isn't that wrong ?

sgrif: Not without removing the entire abstraction and special casing this one place :\

sgrif: No, the diff is just a bit confusing because the `expected a number` error appeared twice before

PR #1852, opened 2018-09-12
Add first class support for newtypes in `Queryable` and `QueryableByN…

sgrif:
…ame`

A consistent problem that's existed for us has been "what do we do about
library X when Diesel doesn't want to care about X, and X doesn't want
to care about Diesel?" Another permutation of this problem has been
folks wanting type conversions that we're never going to support (such
as `impl FromSql<Binary, Sqlite> for Uuid`).

Ultimately the answer for this has pretty much always come down to
newtypes. Whether you're doing it in your app, or having them provided
by a library, that's the only real solution to the orphan rule here.

This commit introduces a new API to all our relevant deserialization
derives: `#[deserialize_as]`. The API is pretty straightforward. You
give us the name of a type. That type has to satisfy whatever
constraints would exist if you used that type directly on your struct.
It also has to implement `Into` for whatever type is on your struct.

This is similar to, but slightly different than
`#[serde(deserialize_with)]`, in that it takes a type name rather than a
function. We cannot mirror serde's API here, as we use this in a
situation where we cannot infer the type you need based on the argument
to the function.

I intend to follow this PR with an equivalent `#[serialize_as]`
attribute for `Insertable` and `AsChangeset`.

Unresolved Questions
--------------------

I'm really not liking using `#[diesel(...)]` to avoid namespace issues.
The main problem with doing that is that we cannot communicate to the
compiler what is allowed inside of it. We also cannot use our normal
"warn if there's unrecognized options" strategy in this case, since
there can be multiple derives that use `#[diesel(...)]` on the same
attribute.

The result of this is that we silently ignore any abnormal options in a
`#[diesel]` attribute. While writing this, I wrote `deserialize_with`
instead of `deserialize_as` twice and couldn't figure out what went
wront. There's a few options here, but I'm not sure which is best.

One would be for us to maintain a global list of all options to
`#[diesel]` that all derives recognize, and warn if something doesn't
match that list. The main downsides to that implemenation is that it's
very brittle, and we won't warn if you use something only recognized by
`#[derive(QueryableByName)]` when that derive isn't present.

The other option is to deprecated `#[diesel(...)]`, and replace all
uses with `#[diesel_...]`. This is the option I'm currently leaning
towards (though I do not think it needs to be done in this PR), since
the only downside is that if you are passing more than one thing to
`#[diesel]`, it will need to be on multiple lines. I do not think we
currently have any cases where that is even reasonable.

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -15,17 +15,25 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
 /// Types which implement `Queryable` represent the result of a SQL query. This
 /// does not necessarily mean they represent a single database table.
 ///
-/// This trait can be derived automatically using `#[derive(Queryable)]`. This
-/// trait can only be derived for structs, not enums.
-///
 /// Diesel represents the return type of a query as a tuple. The purpose of this
 /// trait is to convert from a tuple of Rust values that have been deserialized
 /// into your struct.
+///
+/// # Deriving
+///
+/// This trait can be derived automatically using `#[derive(Queryable)]`. This
+/// trait can only be derived for structs, not enums.
 ///
 /// When this trait is derived, it will assume that the order of fields on your
 /// struct match the order of the fields in the query. This means that field
 /// order is significant if you are using `#[derive(Queryable)]`. Field name has
 /// no effect.
+///
+/// To provide custom deserialization behavior for a field, you can use
+/// `#[diesel(deserialize_as = "Type")]`. If this attribute is present, Diesel
+/// will deserialize into that type, rather than the type on your struct and
+/// call `.into` to convert it. This can be used to add custom behavior for a
+/// single field, or use types that are otherwise unsupported by Diesel.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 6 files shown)


diesel/src/deserialize.rs
```diff
@@ -55,8 +63,59 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
+/// impl Into<String> for LowercaseString {
+///     fn into(self) -> String {
+///         self.0
+///     }
+/// }
+///
+/// impl<DB, ST> Queryable<ST, DB> for LowercaseString
+/// where
+///     DB: Backend,
+///     String: Queryable<ST, DB>,
```
killercup: (Oh? Is writing that constraint (still) necessary?)

diesel/src/deserialize.rs
```diff
@@ -55,8 +63,59 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
+///     fn build(row: Self::Row) -> Self {
+///         LowercaseString(String::build(row).to_lowercase())
+///     }
+/// }
+///
+/// #[derive(Queryable, PartialEq, Debug)]
+/// struct User {
+///     id: i32,
+///     #[diesel(deserialize_as = "LowercaseString")]
+///     name: String,
```
killercup: Huh, from the description I somehow expected this to be the other way around: Deserialize it "like a string", and then call `LowercaseString::from(string)` so you end up with a `LowercaseString` here as type of `name`. I can see now that this would move the whole logic into the `Into` impl and you can't work with the raw row data, though, so you might be converting the data twice (which in this example isn't really necessary).

sgrif: I'm not sure I understand what you're saying. Do you mean that you expected this to be helpful for when you *want* to use the newtype on your struct, requiring only a `From` impl instead of the relevant trait?

The goal of this feature is to let you write impls on newtypes for coherence purposes, but not have to use the newtype on your struct (since they tend to be painful to use). Keep in mind, a lot of this cases this is meant to enable are cases where the inner type isn't supported by Diesel (e.g. all the folks who want `impl FromSql<Binary, Sqlite> for Uuid`)

diesel/src/deserialize.rs
```diff
@@ -15,18 +15,26 @@ pub type Result<T> = result::Result<T, Box<Error + Send + Sync>>;
...
 /// into your struct.
 ///
+/// # Deriving
+///
+/// This trait can be derived automatically using `#[derive(Queryable)]`. This
+/// trait can only be derived for structs, not enums.
+///
 /// When this trait is derived, it will assume that the order of fields on your
 /// struct match the order of the fields in the query. This means that field
 /// order is significant if you are using `#[derive(Queryable)]`. Field name has
```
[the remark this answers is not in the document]
sgrif: I don't think the folks asking questions about it have generally ever been to this page in the first place

PR #1857, opened 2018-09-18
Add some derives for common trivial traits

sgrif:
`impl_selectable_expression!` was one of the few pre 1.1 "fake internal
derive" macros that I hadn't bothered to replace yet. I've also added a
derive for `NonAggregate`, since most types which can derive
`SelectableExpression` and `AppearsOnTable` can also derive
`NonAggregate`.

Spotting places we could remove `NonAggregate` impls was a bit trickier,
since it's not always obvious at a glance whether every field had a
bound or not. The tests always caught it when I messed up though, and
when adding manual impls back I spotted at least one place where the
impls were wrong (`count_star.in(vec![1, 2, 3])` was incorrectly treated
as `NonAggregate`)

I've probably missed a handful of manual impls for `AppearsOnTable` or
`SelectableExpression` that could be derived, but I got all of the
`NonAggregate` impls and uses of the legacy macro.


diesel/src/expression/array_comparison.rs
```diff
@@ -53,12 +53,14 @@ where
 
 impl<T, U> NonAggregate for In<T, U>
```
weiznich: Why don't we derive this two impls too?

diesel_derives/src/non_aggregate.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use util::*;
+
+pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagnostic> {
+    let mut generics = item.generics.clone();
+
+    {
+        let where_clause = generics.make_where_clause();
+        for type_param in item.generics.type_params() {
+            where_clause.predicates.push(parse_quote!(#type_param: NonAggregate));
```
weiznich: Several of the types that now derive `NonAggregate` had `Self: Expression` on their impls before. Not sure if that matters?

sgrif: It's just from back when `Expression` was a supertrait of `NonAggregate`. It's safe to remove.

diesel/src/expression/array_comparison.rs
sgrif: Because we don't want `U: NonAggregate` as a bound.

PR #1865, opened 2018-09-23
Bring back `__diesel_use_everything`

weiznich:
This allows to rename diesel while importing the crate

This rolls back the breaking change that it was required to import diesel as `#[macro_use] extern crate diesel;` in your crate root.

Site note:

It seems like it is [possible to use items from a mod defined inside of a constant by using the full path without self](https://play.rust-lang.org/?gist=8cd4e24b186df6ba50f926202c4e3eac&version=stable&mode=debug&edition=2015), but it seems not to be possible to [create use statements for those things](https://play.rust-lang.org/?gist=89eb3a97a41d3ce2d2980ec817b68dca&version=stable&mode=debug&edition=2015). Not sure if that should be reported as bug against rustc?

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -354,7 +354,3 @@ pub use query_builder::debug_query;
 pub use query_builder::functions::{delete, insert_into, insert_or_ignore_into, replace_into,
                                    select, sql_query, update};
 pub use result::Error::NotFound;
-
-pub(crate) mod diesel {
-    pub use super::*;
-}
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -2,6 +2,14 @@
 
 #[macro_export]
 #[doc(hidden)]
+macro_rules! __diesel_use_everything {
+    () => {
+        pub use $crate::*;
+    }
+}
+
+#[macro_export]
+#[doc(hidden)]
 macro_rules! __diesel_column {
     (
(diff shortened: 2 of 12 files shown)


weiznich:
> It seems like it is possible to use items from a mod defined inside of a constant by using the full path without self, but it seems not to be possible to create use statements for those things. Not sure if that should be reported as bug against rustc?

Opened https://github.com/rust-lang/rust/issues/54525 about that.

sgrif:
I don't think this is worth it. Rust 2018 effectively removes this as an option, since `extern crate` statements are gone.

weiznich:
> Rust 2018 effectively removes this as an option, since extern crate statements are gone. 

a) Not everybody will update to Rust 2018, it is also possible to continue using the 2015 edition.
b) It is also possible to [rename the crate on in rust 2018 by changing the name in your `Cargo.toml` ](https://github.com/rust-lang/cargo/blob/0dd53841d30816ec0d79d879de71d7adc360bdc5/src/doc/src/reference/unstable.md#rename-dependency)
c) This broke some of our internal code. I can also fix this there, but again, this was a breaking change and if it's possible to avoid such a thing we should do that, because this will probably affect more users.


> I don't think we should remove our ability to write use in derives for something that requires writing unidiomatic code to do.

It's quite unfortunate that use statements are broken there, but I consider this a bug in rustc that should be fixed there.

PR #1871, opened 2018-10-03
Support `diesel_manage_updated_at` on SQLite

sgrif:
There's a little bit of funkiness required to make this work. The first
piece is that we need to modify the function registration to provide
access to the connection for our internal functions. This isn't
*strictly* necessary, we could `transmute` the `&self` to be `&'static
self`, since the function will never be called after `self` is dropped.
That felt like it would add some unneccessary unsafety though.

The second bit of funkiness is that we have to have *some* return type,
so we can't just return `()`. I think I want to fix this in the future
by providing a `Null` SQL type, which provides `ToSql` and `FromSql`
impls only for `()`. This requires adding a variant to `SqliteType` and
`MysqlType` though, so it will have to be done in 2.0.

The function itself is subtly different from the PostgreSQL version,
since it runs even if no values were actually changed (we can't do
something like `NEW IS DISTINCT FROM OLD` here). This also cannot be run
on tables without ROWIDs. I think we should probably declare this as a
SQL function in code somewhere for documentation purposes, but I want to
wait until our docs are building again to follow up with that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   example, if one of the fields has the type `Cow<'a, str>`). To define an
   association to such a type, write `#[belongs_to(parent = "User<'_>")]`
 
+* `diesel_manage_updated_at('table_name')` is now available on SQLite. This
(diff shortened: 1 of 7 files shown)


weiznich:
@sgrif Needs rebase.

weiznich:
Closed by #1955

PR #1875, opened 2018-10-05
diesel/macros: Fix a typo in the dsl docs

behnam:
Columns names fall under `columns` module of the table instance module.


behnam:
Actually, I later realized that it's optional. So, feel free to Close if you think the current way is preferred.

weiznich:
I would vote for continue to use the short form because it is (in my opinion) easier to read.

Issue #1399, opened 2017-12-15
Generate a file called `.keep` instead of `.gitkeep`.


sanmai-NL:
```sh 
diesel migration generate
```
creates a `.gitkeep` as well. Do not do this if the crate isn’t under Git VCS.

sgrif:
I guess we can call the file `.keep`.

ocamlmycaml:
I’d like to take a stab at this issue. This would be my first rust project contribution here so i’m hoping this will be ok for that as well

joelgallant:
If this was changed, could it also remove the old `.gitkeep` file?

ocamlmycaml:
> If this was changed, could it also remove the old `.gitkeep` file?

hmm is that a good idea? it'll be dead code after you use `diesel_cli` once (in the newer release). I guess it's ok if I can add a TODO or issue to clean it up in a much later release

hcheng826:
Sorry might be a dump question, what does the `.keep` file do? Is it supposed to be uploaded to the repo? For me it's an empty file. Thanks!

Arjentix:
@hcheng826 , I come here with the same question. After googling a little bit more I've found the asnwer.

So `git` does not track empty directories. And by placing some file in directory we can make `git` track it. So `.keep` is just a stub-file for cases when `migrations` directory is empty.

So you can either delete this file if you have some files in `migrations`, or commit if you don't, or commit just for fun.

[Source](https://www.reddit.com/r/linuxquestions/comments/dae860/what_is_the_keep_file_and_why_do_i_see_it_in_many/).

hcheng826:
@Arjentix Ah I get it. Appreciate the sharing!

PR #1878, opened 2018-10-06
Added first pass of solution

ocamlmycaml:
Issue diesel-rs/diesel#1399

I changed the file name of `.gitkeep` in the `migrations/` directory to `.keep`. I also added some logic in the diesel_cli function to get the migrations directory path (`migrations_dir`) to remove the `.gitkeep` if it already exists. This way the next time someone runs a diesel_cli migration, it'll get deleted.

I'm also not sure how to deal with cleaning up that code after a few releases so any suggestions for that would also be appreciated.

This is also a first attempt at contributing to a rust project in general so I'll welcome any and all criticism :)

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -188,9 +188,27 @@ fn migrations_dir_from_cli(matches: &ArgMatches) -> Option<PathBuf> {
 }
 
 fn migrations_dir(matches: &ArgMatches) -> PathBuf {
-    migrations_dir_from_cli(matches)
+    let dir_path = migrations_dir_from_cli(matches)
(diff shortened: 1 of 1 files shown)


joelgallant:
Question - does hg even require a .keep file like git does? I believe svn doesn't (it tracks directories), not sure about other vcs's.

weiznich:
CI says this needs a rustfmt (1.26.1) and clippy run.

weiznich:
> Not sure if I missed something on the contrib guide, but do I need to do something else to get approval and merge?

You don't missed something, it's just that nobody currently has time to review PR's 🙈 
(More reviewers would be great …)

Issue #1854, opened 2018-09-18
SQLite: on_conflict().do_udpate() support


theduke:
SQLite supports full UPSERT functionality equivalent to Postgres now.

Like: `INSERT INTO phonebook(name,phonenumber) VALUES('Alice','704-555-1212')
  ON CONFLICT(name) DO UPDATE SET phonenumber=excluded.phonenumber;`

See: https://www.sqlite.org/lang_UPSERT.html

Would be really great to support this, the same way it works in PG:

```rust
value.insert_into(table)
  .on_conflict(id)
  .do_update()
  .set(...)
```

sgrif:
Thank you for the feature request. We're currently in the process of a cleanup of the issue tracker, and aren't currently accepting feature requests (there will be an official policy written soon, but the TL;DR is that open issues should reflect something that is a bug or on our immediate roadmap).

We're happy to discuss feature requests, but the place to do so is discourse.diesel.rs, not the issue tracker.

theduke:
Understood.
The README should reflect this new policy, though.

Would you still accept PRs for new features though?

sgrif:
> there will be an official policy written soon

Thanks for your patience.

theduke:
Guess I missed that ;).

sgrif:
I'm actually going to re-open this, since I think implementing it will be fairly easy and would be a nice addition for 1.4

sgrif:
Go for it

h-michael:
@sgrif
Thanks to your reply.
BTW, both of Postgress and Sqlite have same query syntax for upsert.
How do you think how to pg and sqlite share query builder mdule?
If you have any ideas, please tell me your thought.
postgres https://www.postgresql.org/docs/10/static/sql-insert.html
sqlite https://www.sqlite.org/lang_UPSERT.html

weiznich:
There is already an implementation of ` on_conflict().do_update()` for postgresql [here](https://github.com/diesel-rs/diesel/tree/master/diesel/src/pg/upsert).

What needs to be done:
* Move that module to `diesel::upsert` and add reeports at the current location (otherwise this is a breaking change). Also make this module only available if the `sqlite` or `postgres` feature is given.
* See what parts are available on both systems, all other parts should be moved back to the postgres specific module.
[...]

PR #1884, opened 2018-10-14
impliment upsert for SQLite

h-michael:
discuss with #1854.

This pr include breaking change that `pg::upsert` move to `diesel::upsert`.

We can use upsert with SQLite from SQLite version 3.24.0.
https://www.sqlite.org/releaselog/3_24_0.html

[Postgres's doc abount `on conflict`](https://www.postgresql.org/docs/10/static/sql-insert.html)
[Sqlite's doc about `on conflict`](https://www.sqlite.org/lang_UPSERT.html)


sgrif:
@joelgallant Honestly, we haven't done extensive testing on that. Given that all the databases we support are generally backwards compatible in their wire protocol, I believe we suspect virtually every version of PG and MySQL in use today. We do of course support features that are only available in newer versions of those libraries, and if you're using those features (which you would have to explicitly do), you will need to be on a version that supports it. AFAIK the only backend which we have a meaningful version requirement on is SQLite, because we use a function that's only present in newer versions. I honestly can't remember what function it is or what version it was added in, but I know we didn't work with the version of sqlite that shipped with macOS a few years ago.

If you want to do some more extensive testing to figure out what our version support is, I would certainly love to document that. It's valuable to do, but just not something we've prioritized.

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -163,14 +164,15 @@ where
     /// See the documentation for [`on_constraint`] and [`do_update`] for
     /// more examples.
     ///
-    /// [`on_constraint`]: ../pg/upsert/fn.on_constraint.html
-    /// [`do_update`]: ../pg/upsert/struct.IncompleteOnConflict.html#method.do_update
-    pub fn on_conflict<Target>(
+    /// [`on_constraint`]: ../upsert/fn.on_constraint.html
+    /// [`do_update`]: ../upsert/struct.IncompleteOnConflict.html#method.do_update
+    pub fn on_conflict<DB, Target>(
```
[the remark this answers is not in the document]
h-michael: @weiznich 
Sorry, I don't know how to implement you said.
Please tell me in detail, if you could.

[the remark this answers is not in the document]
weiznich: So the `OnConflictTarget` trait [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/upsert/on_conflict_target.rs#L57) says that it has `Queryable<Pg>` as supertrait. Changing that to be generic over `DB` without providing a default type for the generic type parameter is a breaking change. On way around this is to just remove the super trait (`Queryable`) there and then add it as additional constrain everywhere where the compiler tells you that it is missing. (As far as I see this should only be on this function)

h-michael:
@weiznich @sgrif 
Do you know what is version of sqlite on travis ?
I think that that sqlite version is not support upsert.

diesel/src/sqlite/backend.rs
```diff
@@ -50,4 +50,7 @@ impl TypeMetadata for Sqlite {
     type MetadataLookup = ();
 }
 
+impl SupportsDefaultKeyword for Sqlite {}
+impl SupportsReturningClause for Sqlite {}
```
weiznich: Sqlite does not support the default key word. Also returning clauses are not supported. So both traits should not be implemented because otherwise invalid sql will be generated.

diesel/src/upsert/on_conflict_target.rs
```diff
@@ -1,9 +1,10 @@
+use backend::{Backend, SupportsOnConflictClause};
 use expression::SqlLiteral;
-use pg::Pg;
 use query_builder::*;
 use query_source::Column;
 use result::QueryResult;
 
+#[cfg(feature = "postgres")]
```
weiznich: Move postgres specific implementations back to the `pg/upsert` module and then re-export them here.

diesel/src/upsert/on_conflict_target.rs
```diff
@@ -54,14 +56,14 @@ pub struct OnConstraint<'a> {
...
-pub trait OnConflictTarget<Table>: QueryFragment<Pg> {}
+pub trait OnConflictTarget<Table> {}
 
 #[doc(hidden)]
 #[derive(Debug, Clone, Copy)]
 pub struct NoConflictTarget;
 
-impl QueryFragment<Pg> for NoConflictTarget {
-    fn walk_ast(&self, _: AstPass<Pg>) -> QueryResult<()> {
+impl<DB: Backend + SupportsOnConflictClause> QueryFragment<DB> for NoConflictTarget {
```
weiznich: Move the trait bound to an explicit where clause.

diesel_compile_tests/tests/compile-fail/columns_cannot_be_rhs_of_insert.rs
```diff
@@ -1,24 +0,0 @@
-#[macro_use]
```
weiznich: Do not remove tests that are unrelated to the PR

diesel_compile_tests/tests/compile-fail/delete_statement_does_not_support_returning_methods_on_sqlite.rs
```diff
@@ -1,25 +0,0 @@
-#[macro_use]
```
weiznich: Returning clauses are not supported on sqlite so do not remove that test.

diesel_compile_tests/tests/compile-fail/insert_statement_does_not_support_returning_methods_on_sqlite.rs
```diff
@@ -1,53 +0,0 @@
-#[macro_use]
```
weiznich: Returning clauses are not supported on sqlite so do not remove that test.

diesel_compile_tests/tests/compile-fail/pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs
```diff
@@ -1,38 +0,0 @@
...
-    users.select(id).filter(name.eq(any(Vec::<String>::new())))
-        .load::<i32>(&connection);
-    //~^ ERROR type mismatch resolving `<diesel::SqliteConnection as diesel::Connection>::Backend == diesel::pg::Pg`
-    users.select(id).filter(name.is_not_distinct_from("Sean"))
-        .load::<i32>(&connection);
-    //~^ ERROR type mismatch resolving `<diesel::SqliteConnection as diesel::Connection>::Backend == diesel::pg::Pg`
-    users.select(id).filter(now.eq(now.at_time_zone("UTC")))
-        .load::<i32>(&connection);
-    //~^ ERROR type mismatch resolving `<diesel::SqliteConnection as diesel::Connection>::Backend == diesel::pg::Pg`
-    insert_into(users).values(&NewUser("Sean"))
```
weiznich: Only remove this statement all other parts of the test should not be changed.

diesel_compile_tests/tests/compile-fail/update_statement_does_not_support_returning_methods_on_sqlite.rs
```diff
@@ -1,27 +0,0 @@
-#[macro_use]
```
weiznich: Returning clauses are not supported on sqlite so do not remove that test.

h-michael:
@weiznich 
> Do not remove unrelated tests and remove that impls for SupportsDefaultKeyword and SupportsReturningClause for sqlite because both are not supported on sqlite and may lead to invalid sql.

I'm sorry, I was variously misunderstanding.
I've fixed it.

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -53,14 +53,15 @@ where
     /// #     let conn = establish_connection();
     /// #     #[cfg(feature = "postgres")]
     /// #     conn.execute("TRUNCATE TABLE users").unwrap();
-    /// #     #[cfg(feature = "sqlite")]
-    /// #     conn.execute("DELETE FROM users").unwrap();
+    /// #[cfg(feature = "postgres")]
```
h-michael: I think that these how to use cfg attr is bad, but I don't know how to write better especially features documentation test.

h-michael:
@weiznich 
Failed tests on travis are success on my desktop.
I think that sqlite3 running on the travis is not support upsert version.

h-michael:
@weiznich 
Failed tests on travis are success on my desktop.
I think that sqlite3 running on the travis is not support upsert version.

diesel/src/pg/mod.rs
```diff
@@ -6,14 +6,16 @@
...
+#[deprecated(since = "1.4.0", note = "Use `diesel::upsert` instead")]
+pub use super::upsert;
 
 mod backend;
 mod connection;
 mod metadata_lookup;
 mod query_builder;
 pub(crate) mod serialize;
 mod transaction;
+pub mod on_constraint;
```
weiznich: just go with `pub(crate) mod on_constraint;` and reexport that module from `::upsert::on_constraint`

diesel/src/upsert/on_conflict_actions.rs
```diff
@@ -13,8 +13,8 @@ pub fn excluded<T>(excluded: T) -> Excluded<T> {
 #[derive(Debug, Clone, Copy)]
 pub struct DoNothing;
 
-impl QueryFragment<Pg> for DoNothing {
-    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+impl<DB: Backend + SupportsOnConflictClause> QueryFragment<DB> for DoNothing {
```
weiznich: Use an explicit where clause for more than one (simple) constrain.

diesel/src/upsert/on_conflict_extension.rs
weiznich: Maybe move those examples to a postgres specific section  (== below a headline/something indicating that this is only for postgres) of this doc comment?
For the cfg attr usage: This is also done in other places, so this is ok: https://github.com/diesel-rs/diesel/blob/2ce0e4ea0fda474459139042247512f0c8b254cf/diesel/src/query_builder/sql_query.rs#L62

diesel/src/upsert/on_conflict_target.rs
weiznich: We should do `pub pg::on_constraint::*;` to have the old module layout in place. Otherwise this would be a breaking change.

diesel/src/upsert/on_conflict_target.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        impl<T, $($col),+> OnConflictTarget<T::Table> for ConflictTarget<(T, $($col),+)> where
+            T: Column,
+            $($col: Column<Table=T::Table>,)+
+        {
+        }
+    }
+}
+
+on_conflict_tuples!(U);
```
weiznich: Unrelated (so not needed for this PR): We should replace this with a [`__diesel_for_each_tuple`](https://github.com/diesel-rs/diesel/blob/master/diesel/src/macros/tuples.rs#L3) call. 
(This is more or less as side note for the core team)

diesel_compile_tests/tests/compile-fail/pg_specific_expressions_cant_be_used_in_a_sqlite_query.rs
weiznich: After looking at the code above I think this should be changed to contain a `on_constrain` clause, so verify that this extension is only supported on postgresql (Sorry for changing my mind so often…)

diesel/src/pg/mod.rs
```diff
@@ -6,7 +6,6 @@
 
 pub mod expression;
 pub mod types;
-pub mod upsert;
```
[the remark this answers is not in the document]
weiznich: This should also have a `#[doc(hidden)]` attribute to prevent it from showing up in the docs. Also there is the enabled-by-default feature `with-deprecated` for enabling deprecated things. So as the name suggests this reexport should only be included if that feature is enabled. That means just add `#[cfg(feature = "with-deprecated")]` to it.

h-michael:
@weiznich 
Thank you so match for your polite review!
I'll fix that today or tomorrow.
After fix it, should I rebase and clean commit logs ?

h-michael:
@sgrif Would you review this PR?

h-michael:
[memo]
in travisci
```
$ sqlite3 --version
3.8.2 2013-12-06 14:53:30 27392118af4c38c5203a04b8013e1afdb1cebd0d
```

h-michael:
@weiznich @sgrif 
Python depends on libsqlite3-0, so we should build python from source in order to using latest sqlite3. :(

pascalw:
Wouldn't it be easier to run tests with `libsqlite3-sys`? I'm not familiar with Diesels Travis setup, so perhaps there's a good reason not to, but just throwing it out there?

.travis.yml
```diff
@@ -8,10 +8,47 @@ rust:
...
+                   --enable-dynamic-extensions \
+                   --libdir=/usr/lib/x86_64-linux-gnu \
+                   --libexecdir=/usr/lib/x86_64-linux-gnu/sqlite3) &&
+      (cd sqlite-autoconf-3250200; sudo make; sudo make install)
+    fi
 before_script:
   - pip install 'travis-cargo<0.2' --user
   - export PATH=$HOME/.local/bin:$PATH
   - mysql -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'travis'@'%';" -uroot
+  - sudo chown travis:travis /home/travis/.gitconfig
```
h-michael: It changes it's owner ship to root:root. 
I think this is a bug by TravisCI or git.

h-michael: It changes it's owner ship to root:root.
I think this is a bug by TravisCI or git.

h-michael:
@weiznich @sgrif
I've pass the CI.
Would you review this PR?

diesel/src/pg/mod.rs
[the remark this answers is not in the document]
h-michael: Fixed.

h-michael:
@weiznich I've fixed. 
Can I ignore azure pipelines failure?

weiznich:
@h-michael Yes azure pipeline could be ignored. It is just enabled but missing the corresponding pipeline file. (Just started to get #1889 working…)

diesel/src/upsert/mod.rs
```diff
@@ -1,4 +1,5 @@
 //! Types and functions related to PG's and Sqlite's `ON CONFLICT` clause
+//! If you use upsert, you need use PostgreSQL since version 9.0 or Sqlite3 since version 3.24.0.
```
weiznich: There should be an empty line between the "title" line and all other content.

h-michael: Fixed

h-michael:
@Eijebong @killercup @sgrif 
Would you review this PR?

diesel/src/upsert/on_conflict_extension.rs
weiznich: Reorder the doc tests to indicate what's only working on postgres, and then hide/remove the cfg attribute.

h-michael:
@weiznich 
Fix and update CHANGELOG.md.

h-michael:
@weiznich
Travis has passed.
But azzure has not passed yet.
How can I debug azzure pipelines?

azure-pipelines.yml
```diff
@@ -64,7 +65,39 @@ jobs:
...
+                      -DSQLITE_ENABLE_PREUPDATE_HOOK \
+                      -DSQLITE_ENABLE_SESSION \
+                      -DSQLITE_ENABLE_STMTVTAB \
+                      -DSQLITE_MAX_VARIABLE_NUMBER=250000" \
+          ./configure --prefix=/usr \
+                      --enable-threadsafe \
+                      --enable-dynamic-extensions \
+                      --libdir=/usr/lib/x86_64-linux-gnu \
+                      --libexecdir=/usr/lib/x86_64-linux-gnu/sqlite3 &&
+          cd sqlite-autoconf-3240000 &&
```
weiznich: This line is not required
```suggestion
```

azure-pipelines.yml
```diff
@@ -239,7 +272,38 @@ jobs:
         rust_version: 1.31.0
     - bash: |
         sudo apt-get update &&
-        sudo apt-get -y install libsqlite3-dev libpq-dev libmysqlclient-dev
+        sudo apt-get install wget &&
```
weiznich: This misses libpq and mysqlclient
```suggestion
        sudo apt-get install wget libpq-dev libmysqlclient-dev &&
```

azure-pipelines.yml
```diff
@@ -64,7 +65,39 @@ jobs:
     setup:
       - bash: |
           sudo apt-get update &&
-          sudo apt-get -y install sqlite3 libsqlite3-dev
+          sudo apt-get install wget &&
```
weiznich: Use curl retry here?
```suggestion
          sudo apt-get install curl &&
```

azure-pipelines.yml
```diff
@@ -64,7 +65,39 @@ jobs:
...
+          wget --quiet -c https://www.sqlite.org/2018/sqlite-autoconf-3240000.tar.gz &&
```
weiznich:
```suggestion
          curl -fsS --retry 3 --retry-connrefused -o https://www.sqlite.org/2018/sqlite-autoconf-3240000.tar.gz &&
```

azure-pipelines.yml
```diff
@@ -239,7 +272,38 @@ jobs:
         rust_version: 1.31.0
     - bash: |
         sudo apt-get update &&
-        sudo apt-get -y install libsqlite3-dev libpq-dev libmysqlclient-dev
+        sudo apt-get install wget &&
+        wget --quiet -c https://www.sqlite.org/2018/sqlite-autoconf-3240000.tar.gz &&
```
weiznich:
```suggestion
       curl -fsS --retry 3 --retry-connrefused -o https://www.sqlite.org/2018/sqlite-autoconf-3240000.tar.gz &&
```

azure-pipelines.yml
```diff
@@ -64,7 +65,39 @@ jobs:
     setup:
       - bash: |
           sudo apt-get update &&
```
weiznich: Any chance to put the linux sqlite setup code into a own template, so we could remove the duplication here?

weiznich:
@h-michael 
For MacOS it seems like the [wrong sqlite instalation](https://dev.azure.com/diesel-rs/diesel/_build/results?buildId=68&view=logs&jobId=c6526e78-27dc-5fff-edb7-6e55c7ad2f01&taskId=49df001c-c83b-52c1-9a17-1b44fb3a0916&lineStart=14&lineEnd=26&colStart=1&colEnd=41) is used

For the linux builds see the inline comments.

h-michael:
```
2019-01-08T11:19:01.0765437Z curl: option --retry-connrefused: is unknown
2019-01-08T11:19:01.0777480Z curl: try 'curl --help' or 'curl --manual' for more information
2019-01-08T11:19:01.0902731Z ##[error]Bash exited with code '2'.
2019-01-08T11:19:01.0986070Z ##[section]Finishing: Install sqlite
```
oh...

h-michael:
@weiznich cc: @sgrif 
Now, CI has passed except nightly rust issue.
Please please review this PR.

azure-pipelines.yml
weiznich: Take a look at the install_rust template. Something similar should be done here.

h-michael:
@weiznich Now all passed :tada:

weiznich:
I've just noticed that the current implementation does not work for insert statements based on select statements. (See the tests in `diesel_tests/tests/insert_from_select.rs`. In such cases invalid sql is generated for sqlite. That needs to be fixed before we could merge this :(

This means that we need to have a specialized impl of queryable for insertstatements based on select statements for sqlite.

h-michael:
@weiznich 
> I've just noticed that the current implementation does not work for insert statements based on select statements. (See the tests in diesel_tests/tests/insert_from_select.rs. In such cases invalid sql is generated for sqlite.

What case you said? Now all test cases passed.
Dese you said is I should adding more test case?
If that's, could you show me tha case?

h-michael:
@weiznich 
I see. 
> 1. Somehow add a default where clause like suggested by the sqlite docs for the sqlite insert statement with select statements and conflict clause

I think query builder should not automatically insert query that the user unintended.
I think it should be fail build.
But I don't know diesel team policy.
How do you think about that?

h-michael:
> Inserting a additional WHERE true does not alter the query, but only works around a bug in sqlites query parser. (They even state that in their docs…)

This is just my personal opinion.
That is either parser bug or spec, we can't guarantee that it remains the same in the future version.
If that behavior changed, we should add version specific codes.
That is not good.

weiznich:
> That is either parser bug or spec, we can't guarantee that it remains the same in the future version.
If that behavior changed, we should add version specific codes.

Sure we cannot guarantee that this won't change in a future sqlite version, but adding a `WHERE true` clause if no where clause is existing can never lead to a invalid query. Even if a future sqlite version fixes that, those query will remain valid and behave the same.

h-michael:
@weiznich 
Ok, I try to implement adding `WHERE true` when `SELECT` clause that does not contain `WHERE` clause.

weiznich:
Changed Milestone to 2.0 because this needs a solution for the insert by select case. (2.0 is just a placeholder for the next release after 1.4)

h-michael:
@weiznich 
> Changed Milestone to 2.0 because this needs a solution for the insert by select case. (2.0 is just a placeholder for the next release after 1.4)

Then, shall we merge this as 1.4 after adding fail test.

weiznich:
If you get this ready before this evening (UTC) I'm willing to think about adding it to 1.4, otherwise I won't block the release on this. (The release is already long overdue...)

h-michael:
@weiznich 
Ok, now I've rebased.
And I'm going to add fail test soon.

h-michael:
@weiznich 
I added some fail test cases.
And CI will passed soon.

h-michael:
@weiznich CI has passed.

h-michael:
@weiznich Oh, sorry.
I'm trying implement now.
Does diesel have predicate which express `TRUE` or `1=1` ?

weiznich:
There is `true.into_sql::<Bool>` that should work here.

h-michael:
@weiznich 
Sorry for late reply.
I've see all insert statement codebase. But I don't know how to implement add `WHERE true` in such case.
Would you help me?

Skeen:
I would love to see this merged at some point.

Right now I have a hard time finding a purely diesel way to achieve a database agnostic `REPLACE` / `ON DUPLICATE KEY UPDATE`. Postgres does not support `REPLACE` and thus `replace_into` will not be supported without an emulation of `REPLACE` on postgres, however `ON DUPLICATE KEY UPDATE` seems to be supported by PostgreSQL, SQLite and MySQL as well.

weiznich:
@Skeen This will be merged as soon as the issues pointed out above are fixed. So if someone really need this, feel free to submit a updated version of this as new PR.

h-michael:
Ok, I will restart this PR.

theduke:
@h-michael will you have time to work on this?

h-michael:
@theduke I've restarted now :)

h-michael:
@JohnTitor 
This PR have to solve these comments https://github.com/diesel-rs/diesel/pull/1884#issuecomment-454443616, https://github.com/diesel-rs/diesel/pull/1884#issuecomment-454794947, https://github.com/diesel-rs/diesel/pull/1884#issuecomment-454797372.

But if there is an `ON CONFLICT DO NOTHING` statement and there is no WHERE clause for the SELECT clause, we have no idea how to write` WHERE 1 = 1` after the SELECT clause.

Do you have any idea?

JohnTitor:
> Make those code not compiling for sqlite. (This will need a compile-fail test) and we try to fix that after merging this PR.

I agree with @weiznich 's suggestion, it's one of the solutions to merge this.

weiznich:
@h-michael I've found some time to figure out how to handle the missing where clause cases on sqlite. https://github.com/diesel-rs/diesel/pull/1884/commits/fad61d1f0458d0901856a03a59d7668f67f1cd9d implements a solution for that by disallowing plain select statements without where clause and just patching a where clause in for boxed select statements. I'm not sure if that's a good idea or if we better just patch the query for all variants. (Asking @diesel-rs/reviewers here)

I would like to have the code reordered a bit before finally merging it. All SQL construction stuff (so everything related to `QueryFragment` and so on) should belong to `diesel::query_builder::upsert`, everything related to the query construction dsl traits can be left in `diesel::upsert`.

(Yes I know, CI is failing for beta + nightly, that's because of rustc regressions…)

h-michael:
@weiznich 
I'm sorry to have left the PR for a long time.
Thanks for implementing my shortfall.
Is there anything I can help?

weiznich:
No worries. A final test + review would be a great help, otherwise I think that this PR is mostly finished.

h-michael:
@weiznich Thank you for completing this PR.

kpcyrd:
Any estimates when this is going to be released? :)

weiznich:
@kpcyrd I've already written in quite a lot of places that the next feature release is estimated to be released when it's done. I will not give any clearer timeline.
Additionally considering that we run diesel in our free time, please consider what you are asking for.

diesel/src/pg/query_builder/on_constraint.rs
```diff
@@ -0,0 +1,62 @@
...
+///
+/// let pk_conflict_result = diesel::insert_into(users)
+///     .values(&same_id_different_name)
+///     .on_conflict(on_constraint("users_name"))
+///     .do_nothing()
+///     .execute(&conn);
+/// assert!(pk_conflict_result.is_err());
+/// # }
+/// ```
+pub fn on_constraint(constraint_name: &str) -> OnConstraint {
```
pickfire, line 44: Why do we still need this when we can put the original column?

PR #1885, opened 2018-10-14
Add a note to diesel::r2d2 specifying it requires the r2d2 feature

vmalloc:


diesel/src/r2d2.rs
```diff
@@ -1,4 +1,6 @@
-//! Connection pooling via r2d2
+//! Connection pooling via r2d2.
+//! Note that In order to use this module, you need to have Diesel
```
weiznich: "Note that *in* order to use this module you need to enable the `r2d2` feature on diesel." 
sounds better for me.

vmalloc: Oops, my bad. Fixed

diesel/src/r2d2.rs
```diff
@@ -1,4 +1,6 @@
-//! Connection pooling via r2d2
+//! Connection pooling via r2d2.
+//! Note that in order to use this module, you need to have Diesel
```
sgrif: There should be an empty line here.

diesel/src/r2d2.rs
```diff
@@ -1,4 +1,6 @@
...
+//! installed with the `r2d2` feature
```
sgrif: I don't like the use of the word "installed" for this. What do you think of "Note: This module requires enabling the `r2d2` feature"

vmalloc: You're right. That's a lot better.

Issue #239, opened 2016-03-18
Cargo.lock


jgillich:
I'd like to package diesel_cli for Nix, however this requires a Cargo.lock file. Would you mind adding it?

PR #1886, opened 2018-10-14
Add Cargo.lock to diesel_cli

lschuermann:
I'm going to give another shot at packaging `diesel_cli` for Nix (#239). Somehow the Cargo.lock file vanished. Would you mind it being added again?

### DIFF
--- a/diesel_cli/Cargo.lock
+++ b/diesel_cli/Cargo.lock
@@ -1,0 +1,700 @@
+[[package]]
+name = "aho-corasick"
+version = "0.6.8"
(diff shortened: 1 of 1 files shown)


weiznich:
The lock file was removed in 79af07cc0. If I understand that commit correctly readding it is useless because diesel_cli is part of a workspace, so the workspace lock file is used.

sgrif:
Yeah, I'd prefer we not give ourselves a false sense of security here. Crates.io/cargo will not use this lockfile, and we shouldn't pretend it will.

PR #1890, opened 2018-10-20
Update rustfmt install instructions

AndrewSpeed:
This PR takes work from https://github.com/diesel-rs/diesel/pull/1762 to update the instructions to contributors, and specifies the version of rust used to run `cargo fmt` and `cargo clippy`.


weiznich:
> It looks like CI is failing due to unstable features being tested for the stable and beta versions, but I'm not sure what the cause of this is

I think this is caused by the rust-toolchain file. 

@diesel-rs/contributors Maybe we should not add a rust-toolchain file for a compiler version that is not supported any more by diesel? So maybe just bumping rustfmt to 1.30?

PR #1897, opened 2018-10-27
Reserved names replacement in tables! is also applied in joinable!

### LINKED ISSUE #1609: Changed column names in `table!` don't make it to `joinable!`
## Setup
Latest version

## Problem Description
Using `print-schema` on a pre-existing database, columns named `type` are correctly changed to `type_` with a `#[sql_name = "type"]` attribute, but the underscore is not added in the `joinable!` macro later, leading to errors.

## Checklist

meven:
Fixes #1609 

I expect some feedback.
I have a hard time testing.
I want to add more tests.

https://github.com/diesel-rs/diesel/commit/ce056c6f06046a17ec0cc97b4409f6b292fedbd9
should be a good example

cc @Eijebong


diesel_cli/src/print_schema.rs
```diff
@@ -245,10 +245,15 @@ struct Joinable<'a>(&'a ForeignKeyConstraint);
 
 impl<'a> Display for Joinable<'a> {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        let fk = if self.0.foreign_key_rust_name.is_some() {
+            self.0.foreign_key_rust_name.unwrap()
```
JohnTitor: This line has an error because `fmt` doesn't have ownership of `self.0` maybe.

```
error[E0507]: cannot move out of borrowed content
   --> diesel_cli/src/print_schema.rs:249:13
    |
249 |             self.0.foreign_key_rust_name.unwrap()
    |             ^^^^^^ cannot move out of borrowed content
```

diesel_cli/src/print_schema.rs
```diff
@@ -245,10 +245,15 @@ struct Joinable<'a>(&'a ForeignKeyConstraint);
...
+        } else {
+            self.0.foreign_key
```
JohnTitor: Same here

meven:
The code has been fixed and rebased.
The previous test that was not valid has been replaced a proper test for the fix.

The replacement of reserved keywords in the foreign constraint primary key field is not tested currently.

JohnTitor:
This is a off-topic but you should use rustfmt.

meven:
Ran rustfmt, remove the unused primary_key_rust_name field, squashed commits

diesel_cli/src/infer_schema_internals/inference.rs
```diff
@@ -109,6 +109,12 @@ pub fn load_foreign_key_constraints(
 
     constraints.map(|mut ct| {
         ct.sort();
+        ct.iter_mut().for_each(|foreign_key_constraint| {
+            if RESERVED_NAMES.contains(&foreign_key_constraint.foreign_key.as_str()) {
```
meven: This check could be done in `impl<'a> Display for Joinable<'a> ` alternatively, but this way the field is quite explicit.

meven:
I feel the code is in good shape and ready to land, unless the alternate implementation would be favored.

meven:
@Eijebong anytime for last review ?

mrcosta:
Ironically we're facing this issue right now haha. Nice to see the fix coming. Nice job you two.

meven:
@Eijebong Can't wait for last review.

meven:
I really hope, this would get fixed, it has been ready for a while.

diesel_cli/src/infer_schema_internals/data_structures.rs
```diff
@@ -102,6 +102,7 @@ pub struct ForeignKeyConstraint {
     pub child_table: TableName,
     pub parent_table: TableName,
     pub foreign_key: String,
+    pub foreign_key_rust_name: Option<String>,
```
weiznich: Is there any specific reason why we need an extra field for this? As far as I can see we will only use the rust name, so we can just set/modify the existing field to contain the right name?
Alternatively this field shouldn't be an option field, but contain the original name in the current `None` case

meven: There were no particular reason to have a foreign_key_rust_name, it was just that I thought foreign_key could be used elsewhere and https://github.com/diesel-rs/diesel/commit/ce056c6f06046a17ec0cc97b4409f6b292fedbd9, which I drew inspiration upon, added a rust_name field for similar needs. I have tested your sugestion of editing the foreign_key field with the rust_name and it works fine.
I suggested another alternative would be to do the name to rust_name conversion in print_schema.rs `impl<'a> Display for Joinable<'a> ` avoiding adding a field or changing foreign_key field.

I guess we mainly want to avoid adding a field here.
Which solution would we prefer though ?

weiznich: I think we can add the field, but we should not use `Option<String>` as type. Just set the normal name in the current none case.

meven: I have updated the PR following your suggestion.

PR #1898, opened 2018-10-27
Allow libsqlite3-sys 0.10

SergioBenitez:


weiznich:
Thanks

SergioBenitez:
Any chance you'd be willing to push a point release with this change? Due to the linking restrictions for `-sys` libraries, this is preventing updating other libraries.

weiznich:
Unfortunately not because we cannot publish new documentation because of a rustdoc issue. See https://github.com/rust-lang/rust/issues/54524 for details.

PR #1902, opened 2018-10-31
Sort database migrations before embedding

### LINKED ISSUE #1901: Reproducible builds: Non deterministic order of embedded migrations
## Setup

I'm trying to make a program pass reprotest for [reproducible-builds][rb]. I've noticed that embedded_migrations introduce randomness into the binary that may prevent successful rebuilding:

```sh
reprotest -vv --vary=-time,-domain_host,-build_path --store-dir artifacts/ --source-pattern 'Cargo.* src/ migrations/' '
    CARGO_HOME="$PWD/.cargo" RUSTUP_HOME='"$HOME/.rustup"' \
        RUSTFLAGS="--remap-path-prefix=$HOME=/remap-home --remap-path-prefix=$PWD=/remap-pwd" \
        cargo build --release --verbose --locked' \
    target/release/fsord
```

This builds a small example project that has a few migrations embedded:
```rust
extern crate diesel;
#[macro_use] extern crate diesel_migrations;

use diesel::sqlite::*;
use diesel::Connection;

embed_migrations!();

fn main() {
    let path = "foo.db";
    let db = SqliteConnection::establish(&path).unwrap();
    embedded_migrations::run(&db).unwrap();
    println!("Hello, world!");
}
```

This currently fails since the `embed_migrations!` macro depends on the dirent order, which may be undeterministic:
https://github.com/diesel-rs/diesel/blob/master/diesel_migrations/migrations_internals/src/lib.rs#L264-L279
```rust
#[doc(hidden)]
pub fn migration_paths_in_directory(path: &Path) -> Result<Vec<DirEntry>, MigrationError> {
    try!(path.read_dir())
        .filter_map(|entry| {
            let entry = match entry {
                Ok(e) => e,
                Err(e) => return Some(Err(e.into())),
            };
            if entry.file_name().to_string_lossy().starts_with('.') {
                None
            } else {
                Some(Ok(entry))
            }
        })
        .collect()
}
```

migrations

[rb]: https://reproducible-builds.org/

### Versions

- **diesel:** 1.3.3
- **diesel_migrations:** 1.3.0

### Feature Flags

- **diesel:** sqlite

## Suggested solution

The list of migrations should be sorted, either directly in `migration_paths_in_directory`, or in the `embed_migrations!` macro.

I assume this list needs to be sorted at runtime anyway to ensure migrations are applied in the correct order.

## Checklist

kpcyrd:
Resolves #1901

### DIFF
--- a/diesel_migrations/migrations_macros/src/embed_migrations.rs
+++ b/diesel_migrations/migrations_macros/src/embed_migrations.rs
@@ -78,7 +78,11 @@ pub fn derive_embed_migrations(input: &syn::DeriveInput) -> quote::Tokens {
 }
 
 fn migration_literals_from_path(path: &Path) -> Result<Vec<quote::Tokens>, Box<Error>> {
-    try!(migration_paths_in_directory(path))
+    let mut migrations = migration_paths_in_directory(path)?;
+
+    migrations.sort_by_key(|a| a.path());
+
+    migrations
         .into_iter()
         .map(|e| migration_literal_from_path(&e.path()))
         .collect()


kpcyrd:
@weiznich thanks, I've updated my pr

diesel_migrations/migrations_macros/src/embed_migrations.rs
```diff
@@ -78,7 +78,13 @@ pub fn derive_embed_migrations(input: &syn::DeriveInput) -> quote::Tokens {
 }
 
 fn migration_literals_from_path(path: &Path) -> Result<Vec<quote::Tokens>, Box<Error>> {
-    try!(migration_paths_in_directory(path))
+    let mut migrations = migration_paths_in_directory(path)?;
+
+    // Sort the migrations to make sure embedded migrations are deterministic
+    // and the binary is reproducible.
+    migrations.sort_by(|a, b| a.path().cmp(&b.path()));
```
sgrif: What do you think about `migrations.sort_by_key(|a| a.path())` instead?

diesel_migrations/migrations_macros/src/embed_migrations.rs
```diff
@@ -78,7 +78,13 @@ pub fn derive_embed_migrations(input: &syn::DeriveInput) -> quote::Tokens {
 }
 
 fn migration_literals_from_path(path: &Path) -> Result<Vec<quote::Tokens>, Box<Error>> {
-    try!(migration_paths_in_directory(path))
+    let mut migrations = migration_paths_in_directory(path)?;
+
+    // Sort the migrations to make sure embedded migrations are deterministic
```
sgrif: These sort of comments tend to get out of date with the code, and we prefer not to have them. Context like this is better kept in the commit message instead.

diesel_migrations/migrations_macros/src/embed_migrations.rs
kpcyrd: Good catch, updated my PR.

diesel_migrations/migrations_macros/src/embed_migrations.rs
kpcyrd: I've moved the comment to the commit message.

PR #1903, opened 2018-11-04
sqlite extension loading tests

alamminsalo:
Tests loading with/without calling `enable_load_extension()` call and calling extension function with result assert.

Added `libspatialite5` as test dependency.


.travis.yml
```diff
@@ -89,6 +90,7 @@ branches:
   only:
     - master
     - ಠ_ಠ
+    - sg-enable-load-extension
```
weiznich: You should not change this branches.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -106,8 +106,8 @@ impl RawConnection {
         }
     }
 
-    pub fn enable_load_extension(&self) {
-        unsafe { ffi::enable_load_extension(self.internal_connection.as_ptr(), true) }
+    pub fn enable_load_extension(&self) -> i32 {
```
weiznich: Why do we return a `i32` here?

alamminsalo: changed to return QueryResult instead

.travis.yml
alamminsalo: removed this line

alamminsalo:
What is the status of this pull? The requested changes were added

weiznich:
This PR is opened against a old stale branch.

PR #1889, opened 2018-10-19
Play with azure-pipeline

weiznich:
This builds everything on all platforms that are supported by azure pipelines.

Someone with admin access to the diesel organisation needs to add a organisation on azure pipelines and configure the setup. (Maybe @sgrif :wink:)

A build can be seen here: https://dev.azure.com/weiznich/diesel-test/_build/results?buildId=187&view=logs. Some of the builds are currently failing but I'm not sure why 😟 

In detail the following to errors occure:

- [x]  All windows builder: There seems to be some issue with the tests for diesel_cli,
- [x] On MacOS the postgres builder is failing at diesel_tests with the following error:
```
---- select::select_for_update_modifiers stdout ----
 thread 'select::select_for_update_modifiers' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(UniqueViolation, "duplicate key value violates unique constraint \"pg_type_typname_nsp_index\"")', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Beside of that I would suggest to move all the ci things to azure because it seems that all those builds are approximately as fast the current travis builds.

Closes #1838 
Closes #1880

PR #1910, opened 2018-11-07
Fix rustdoc error (Missing link)

weiznich:
As @oli-obk pointed out in https://github.com/rust-lang/rust/issues/54524 this link was broken


weiznich:
Picked to #1889

PR #1915, opened 2018-11-13
Workaround a uuid bump breaking change

### LINKED ISSUE #1900: uuid 0.7 not being used because of features in manifest
### Versions

- **Rust:** `rust version 1.30.0 (da5f414c2 2018-10-24)` 
- **Diesel:** 1.3.3
- **Database:** Any
- **Operating System:** Any

### Feature Flags

- **diesel:** `uuid`

## Problem Description

As far as I can tell, uuid 0.7 will never be used because it doesn't match the requirements. While 0.7 is indeed >= 0.2 and < 0.8.0, it does not have a `use_std` feature, which locks uuid to 0.6. The feature was [renamed to `std` in 0.7 and is a default feature](https://github.com/uuid-rs/uuid/blob/master/Cargo.toml#L73).

I forked diesel and [changed Cargo.toml to use `std`](https://github.com/jkcclemens/diesel/blob/master/diesel/Cargo.toml#L26), then used a patch section in one of my projects. Updating to my fork removed uuid 0.6 and added uuid 0.7. Reverting back to the official git adds uuid 0.6 back.

## Checklist

weiznich:
(This needs a cargo nightly feature so we cannot do this now, but as
soon as this hit's stable)

Tracking issue for cargo: https://github.com/rust-lang/cargo/issues/5653
Fixes #1900

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -24,6 +24,7 @@ serde_json = { version = ">=0.8.0, <2.0", optional = true }
 time = { version = "0.1", optional = true }
 url = { version = "1.4.0", optional = true }
 uuid = { version = ">=0.2.0, <0.7.0", optional = true, features = ["use_std"] }
+uuidv07 = { version = "0.7.0", optional = true, package = "uuid"}
 ipnetwork = { version = ">=0.12.2, <0.14.0", optional = true }
 num-bigint = { version = ">=0.1.41, <0.3", optional = true }
 num-traits = { version = "0.2", optional = true }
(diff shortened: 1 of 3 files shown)


weiznich:
[Rust 1.31 stabilizes](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#cargo) the needed feature, so merge this as soon as we bump our  minimal rustc version to 1.31.

jonas-schievink:
#1936 is now merged :tada:

PR #1916, opened 2018-11-13
Revert "Update Uuid to 0.7"

weiznich:
This reverts commit 8973d8427b2c82d168ef641627c3ecdefaf601f3.

Short term solution for #1900


weiznich:
> Makes me sad but 👍 at least we'll show it's not compatible

We can add support for uuid 0.7 as soon as cargo supports renaming crates(#1915) or we release diesel 2.0

PR #1923, opened 2018-11-26
Fix a compilation error in diesel_derives with the latest version of syn

### LINKED ISSUE #1921: syn 0.15.22 breaks diesel_derives
### Versions

- **Rust:** 1.30.1
- **Diesel:** 1.33
- **Database:** PostgreSQL
- **Operating System** macOS 10.14.1

### Feature Flags

- **diesel:** default

## Problem Description

The latest version of the `syn` crate exports a new `Result` type alias, which collides with `std::result::Result` and makes `diesel_derives` fail to compile.

Use `diesel_derives`.

`diesel_derives` compiles successfully.

```
error[E0107]: wrong number of type arguments: expected 1, found 2
  --> diesel_derives/src/util.rs:48:84
   |
48 | pub fn ty_for_foreign_derive(item: &DeriveInput, flags: &MetaItem) -> Result<Type, Diagnostic> {
```

No

### Steps to reproduce

Compile diesel_derives from master.

## Checklist

trha:
This PR replaces all glob imports in `diesel_derives/src/utils.rs` with concrete imports to prevent future occurrences of this error.
Closes #1921.

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -1,8 +1,8 @@
-use proc_macro2::*;
-use syn::*;
+pub use diagnostic_shim::{Diagnostic, DiagnosticShim, EmitErrorExt};
 
-pub use diagnostic_shim::*;
-use meta::*;
+use proc_macro2::{TokenStream, Span};
+use syn::{Ident, Type, DeriveInput, Data, GenericArgument};
+use meta::MetaItem;
 
 pub fn wrap_in_dummy_mod(const_name: Ident, item: TokenStream) -> TokenStream {
     quote! {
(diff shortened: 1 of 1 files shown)


Eijebong:
This won't compile with syn < 0.15.22 right ? If so please change the version of sun in the Cargo.toml to 0.15.22

Also is this in some unstable API ? If it.s not maybe the right call would be to ask for that version to be yanked

weiznich:
@Eijebong I do not see why it should not work with syn 0.15.21. As far as I understood the issue that is breaking diesel_derives is that syn 0.15.22 exports some type called `Result<T>` from it's main module. We are doing a wild card import of all items from that module, but assume that `Result` refers to the `Result<T, E>` type from the std lib.Therefore compilation will break with the new version. 
The fix just explicitly imports the items that we actually use (so no "wrong" `Result<T>` type is pulled in scope) Those types should exist on previous syn versions, because we used them before.

Eijebong:
Ha then disregard :p

weiznich:
So this seems fine modulo that clippy compiler error. That seems like we need to update clippy … (Will try to look into this this evening)

PR #1931, opened 2018-12-06
Remove `syn` glob import in `diesel_derives`.

icefoxen:
Resolves issue #1921.


weiznich:
Thanks for opening this PR, but there is already one PR open to fix that problem.
Duplicate of #1923

PR #1947, opened 2019-01-06
Mark as edition 2018 and fix crate-root imports (also lets users drop macro_use)

mehcode:
Working on an async connection interface as a side-project. This is a required first step for async functions.

Didn't see a nice way to run tests locally so this build probably won't be green. Do we have something setup with docker to run all the tests? If not, should we?

---

Neat bonus here, that was easy while I was in the middle of it, is all Diesel macros and derives work via standard `use` imports now. `#[macro_use] extern crate diesel;` can be dropped from usage. They still work fine in Rust 2015 (and 2018) using `#[macro_use]` though (as proven by tests as I only changed 1 to use the new import style).

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -20,17 +20,17 @@ else
   (cd diesel && cargo test --no-default-features --features "extras sqlite postgres mysql" $*)
 
   (cd diesel_cli && cargo test --features "sqlite" --no-default-features $*)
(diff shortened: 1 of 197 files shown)


mehcode:
I'll give it a go later. Thanks for the link.

mehcode:
Decided to tear this PR apart into several small PRs to slowly move us forward here. See #1956 for the first one.

PR #1949, opened 2019-01-09
Update libsqlite-sys again, now that the old feature flag continue to…

weiznich:
… exists


lnicola:
Would it be possible to upgrade to a newer libsqlite3-sys, or do you consider that a breaking change?

weiznich:
@lnicola As long as that does not break support with older version, no.
Otherwise: The next diesel release will be 2.0 anyway, so it does not matter. Maybe open a PR to discuss the details there?

PR #1953, opened 2019-01-10
Ignore build failures for rust nightly builds

weiznich:
Just trying


weiznich:
So that's not working :(

PR #1952, opened 2019-01-10
Fix homebrew error on azure pipelines

h-michael:
Fix problem of this ci failing.
https://dev.azure.com/diesel-rs/diesel/_build/results?buildId=81

PR #1954, opened 2019-01-12
Mention ilike in changelog.

kaj:
Now that PR #1835 is merged, I think it should be mentioned in the changelog that `Nullable<Text>` now supports ilike on PostgreSQL.


kaj:
The broken azure build seems to be fixed in #1952 .

weiznich:
Thanks :+1:

Issue #1601, opened 2018-03-26
SqlOrd missing from Timestamptz


dsvensson:
As subject says,

https://github.com/diesel-rs/diesel/blob/master/diesel/src/sql_types/ord.rs

gasabr:
I would like to try to implement this issue

Eijebong:
@gasabr Sure, go for it :+1:

PR #1959, opened 2019-01-19
impl SqlOrd for Timestamptz

stephenmuss:
Added SqlOrd for Timestamptz as per #1601 .


weiznich:
Thanks :+1:

Close #1601

PR #1968, opened 2019-01-25
Fix a hanging test

weiznich:
The "underlying issue here is that this stalls as soon as we try to
get the second connection.

Seperate them as a quick fix into 2 test cases.


sgrif:
Sorry XD

PR #1974, opened 2019-01-30
Update lib.rs to reflect 2018 same crate path syntax

stevensonmt:
Does not compile without this change. Alternatively stipulate upper limit of rust version for which the example is valid.

### DIFF
--- a/examples/postgres/getting_started_step_2/src/lib.rs
+++ b/examples/postgres/getting_started_step_2/src/lib.rs
@@ -20,7 +20,7 @@ pub fn establish_connection() -> PgConnection {
 }
 
 pub fn create_post(conn: &PgConnection, title: &str, body: &str) -> Post {
-    use schema::posts;
+    use crate::schema::posts;
 
     let new_post = NewPost { title, body };
 


weiznich:
I do not think that it is a good idea to migrate only parts of the code/examples to the 2018 edition. That should be done in one step. There is already some effort done by @mehcode on this.

Also I rechecked that the examples are building fine using the last stable(1.32.0) and last nightly( c1c3c4e95 2019-01-29) compiler. Therefore I see no reason to change this now. 
(The Cargo.toml files are part of the examples. They state that `edition = "2018"` is not set, therefore this code is fine.)

stevensonmt:
Does not compile with rustc 1.31.1.

mehcode:
You're welcome to help out here. Waiting on me for things is a bad idea (my free time is always odd). 

I would turn on the `future_incompatible` warning for one crate at a time and fix anything that pops up.

weiznich:
> Does not compile with rustc 1.31.1.

I cannot reproduce that with rustc 1.31.1. (It builds fine for me...)

Can you provide detailed steps to reproduce this?

stevensonmt:
> > Does not compile with rustc 1.31.1.
> 
> I cannot reproduce that with rustc 1.31.1. (It builds fine for me...)
> 
> Can you provide detailed steps to reproduce this?

Weird. Here's the code I'm using: https://github.com/stevensonmt/diesel_pg_demo

The error returned is 
>error[E0658]: imports can only refer to extern crate names passed with `--extern` on stable channel (see issue #53130)                                                                        
  --> src/lib.rs:22:9                                                                                                                                                                         
   |                                                                                                                                                                                          
10 | pub mod schema;                                                                                                                                                                          
   | --------------- not an extern crate passed with `--extern`                                                                                                                               
...                                                                                                                                                                                           
22 |     use schema::posts;                                                                                                                                                                   
   |         ^^^^^^                                                                                                                                                                           
   |                                                                                                                                                                                          
note: this import refers to the module defined here                                                                                                                                           
  --> src/lib.rs:10:1                                                                                                                                                                         
   |                                                                                                                                                                                          
10 | pub mod schema;                                                                                                                                                                          
   | ^^^^^^^^^^^^^^^        

```
$ rustc -V
rustc 1.31.1 (b6c32da9b 2018-12-18)
```

Manjaro Linux box.

weiznich:
> 
Weird. Here's the code I'm using: https://github.com/stevensonmt/diesel_pg_demo

As I guessed before: That's not the example code :wink:. The problem there is that you've not copied the whole project, but only the lib.rs file. The example also includes a Cargo.toml file, **without `edition ="2018"**. Your project sets that flag. (It's a bit unfortunately that `cargo new` by default now generates projects with edition 2018 enabled by default...)

Anyway: I'm open to change those examples. But such a change should update all examples at once, not only a single one. Feel free to open a new PR for updating all projects in the example folder.

stevensonmt:
The problem is not that I did not clone the repo, but that I followed the tutorial. The tutorial does not mention anything about that flag. So the tutorial should be updated to reflect one of:

1. An upper bound of rust version
2. Instructions to remove that flag from `Cargo.toml` if the rust version is a 2018 edition
3. Instructions to add the same-crate syntax where appropriate for rust versions that are 2018 editions.

I am happy to make edits and PRs for all the examples, but I'd like to know which of these three strategies is preferred.

weiznich:
> I am happy to make edits and PRs for all the examples, but I'd like to know which of these three strategies is preferred.

As written above: I'm OK with a similar change like this one, as long as that updates all examples. (And given the proposed alternatives I would prefer this variant.)

PR #1975, opened 2019-01-30
Adds `BoxedSqlQuery`

ZeGentzy:
Signed-off-by: Hal Gentz <zegentzy@protonmail.com>

### DIFF
--- a/diesel/src/query_builder/sql_query.rs
+++ b/diesel/src/query_builder/sql_query.rs
@@ -4,7 +4,7 @@ use std::marker::PhantomData;
 use backend::Backend;
 use connection::Connection;
 use deserialize::QueryableByName;
-use query_builder::{AstPass, QueryFragment, QueryId, self, debug_query};
+use query_builder::{self, debug_query, AstPass, QueryFragment, QueryId};
 use query_dsl::{LoadQuery, RunQueryDsl};
 use result::QueryResult;
 use serialize::ToSql;
@@ -180,7 +180,9 @@ where
 impl<Conn, Query, Value, ST> RunQueryDsl<Conn> for UncheckedBind<Query, Value, ST> {}
 
 #[must_use = "Queries are only executed when calling `load`, `get_result`, or similar."]
-/// See `SqlQuery`
+/// See [`SqlQuery::into_boxed`].
+///
+/// [`SqlQuery::into_boxed`]: ../struct.SqlQuery.html#method.into_boxed
 pub struct BoxedSqlQuery<'f, DB: Backend, Inner> {
     inner: Inner,
     sql: String,
@@ -205,20 +207,20 @@ impl<'f, DB: Backend, Inner> BoxedSqlQuery<'f, DB, Inner> {
         }
     }
 
-    /// See `SqlQuery`'s `bind`
+    /// See [`SqlQuery::bind`].
+    ///
+    /// [`SqlQuery::bind`]: ../struct.SqlQuery.html#method.bind
     pub fn bind<BindSt, Value>(mut self, b: Value) -> Self
     where
-        BindSt: QueryId,
         DB: HasSqlType<BindSt>,
         Value: ToSql<BindSt, DB> + 'f,
     {
-        self.binds.push(Box::new(move |mut out| {
-            out.push_bind_param_value_only(&b)
-        }));
+        self.binds
+            .push(Box::new(move |mut out| out.push_bind_param_value_only(&b)));
         self
     }
 
-    /// See `SqlQuery`'s `sql`
+    /// Appends a piece of SQL code at the end.
     pub fn sql(mut self, sql: &str) -> Self {
         self.sql += sql;
         self
--- a/diesel/src/query_builder/sql_query.rs
+++ b/diesel/src/query_builder/sql_query.rs
@@ -88,8 +88,9 @@ impl<Inner> SqlQuery<Inner> {
     }
 
     /// Appends a piece of SQL code at the end.
-    pub fn sql<T: Into<String>>(self, sql: T) -> SqlQuery<Self> {
-        SqlQuery::new(self, sql.into())
+    pub fn sql<T: AsRef<str>>(mut self, sql: T) -> Self {
+        self.query += sql.as_ref();
+        self
     }
 }
 
@@ -227,8 +228,8 @@ impl<'f, DB: Backend, Query> BoxedSqlQuery<'f, DB, Query> {
     /// See [`SqlQuery::sql`].
     ///
     /// [`SqlQuery::sql`]: ../struct.SqlQuery.html#method.sql
-    pub fn sql<T: Into<String>>(mut self, sql: T) -> Self {
-        self.sql += &sql.into();
+    pub fn sql<T: AsRef<str>>(mut self, sql: T) -> Self {
+        self.sql += sql.as_ref();
         self
     }
 }
--- a/diesel/src/query_builder/sql_query.rs
+++ b/diesel/src/query_builder/sql_query.rs
@@ -1,10 +1,9 @@
-use std::fmt::{self, Debug};
 use std::marker::PhantomData;
 
 use backend::Backend;
 use connection::Connection;
 use deserialize::QueryableByName;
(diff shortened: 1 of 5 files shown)


diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +178,85 @@ where
 }
 
 impl<Conn, Query, Value, ST> RunQueryDsl<Conn> for UncheckedBind<Query, Value, ST> {}
+
+#[must_use = "Queries are only executed when calling `load`, `get_result`, or similar."]
+/// See `SqlQuery`
```
weiznich: This should link to SqlQuery or better to the `into_boxed` calls

diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +178,85 @@ where
...
+pub struct BoxedSqlQuery<'f, DB: Backend, Inner> {
+    inner: Inner,
+    sql: String,
+    binds: Vec<Box<dyn Fn(AstPass<DB>) -> QueryResult<()> + 'f>>,
+}
+
+impl<DB: Backend, Inner> Debug for BoxedSqlQuery<'_, DB, Inner>
```
weiznich: cc @diesel-rs/core This seems to be a clever idea. Maybe we could do similar things for other query dsl types?

diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +178,85 @@ where
...
+impl<'f, DB: Backend, Inner> BoxedSqlQuery<'f, DB, Inner> {
+    pub(crate) fn new(inner: Inner) -> Self {
+        BoxedSqlQuery {
+            inner,
+            sql: "".to_string(),
+            binds: vec![],
+        }
+    }
+
+    /// See `SqlQuery`'s `bind`
```
weiznich: Link to the actual documentation

diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +178,85 @@ where
...
+        DB: HasSqlType<BindSt>,
+        Value: ToSql<BindSt, DB> + 'f,
+    {
+        self.binds.push(Box::new(move |mut out| {
+            out.push_bind_param_value_only(&b)
+        }));
+        self
+    }
+
+    /// See `SqlQuery`'s `sql`
```
weiznich: `SqlQuery` has no `sql` function, because the whole query is provided at once. Therefore that function is not necessary and should be removed. (Same for the sql field, that also seems to be unused then.)

diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +178,85 @@ where
...
+            inner,
+            sql: "".to_string(),
+            binds: vec![],
+        }
+    }
+
+    /// See `SqlQuery`'s `bind`
+    pub fn bind<BindSt, Value>(mut self, b: Value) -> Self
+    where
+        BindSt: QueryId,
```
weiznich: I do not see why this bound is required. (Possibly I miss something 🙈 )

diesel/src/query_builder/sql_query.rs
ZeGentzy: I did not realize `SqlQuery` has no `sql` function, however, normally you can add extra sql via [sql](file:///home/gentz/Documents/0ad/source/tools/rdb/diesel/target/doc/diesel/expression/struct.SqlLiteral.html#method.sql).

As far as I can tell, an `sql` function in `BoxedSqlQuery` is necessary for doing things like this in a loop:
```
                        replays_query = replays_query
                            .sql("`")
                            .sql(&id_str)
                            .sql("`.`")
                            .sql(id_mem)
                            .sql("`")
                            .sql(op)
                            .sql(match equal {
                                OrEqual::True => "=",
                                OrEqual::False => "",
                            })
                            .sql("? ")
                            .bind::<sql_types::Binary, _>(val)
```

I guess we could loop twice, once for compiling the sql query, and another for adding the binds, however, that strikes me as unergonomic.

weiznich: That example won't work. `SqlQuery` assumes that you manually write the whole sql query, including the bind parameters.

The `sql` function is something entirely different. (Though this function needs also be extended to support a variable number of joins, but that's another PR)

weiznich:
CI says you need to run rustfmt.

diesel/src/query_builder/sql_query.rs
ZeGentzy: This example does work, however, with `BoxedSqlQuery`.

ZeGentzy: Should be noted that users can also do things like `diesel::dsl::sql::<sql_types::Integer>("Hello ").sql("World!");`

ZeGentzy:
@weiznich Have I addressed all your concerns?

diesel/src/query_builder/sql_query.rs
weiznich: I would like to hear the opinion of at least one other core team member on this. 
In my opinion the API should be symmetrical to `SqlQuery`, that means we need to add such an function there or remove it here.

diesel/src/query_builder/sql_query.rs
sgrif: It's an interesting idea, but I think it needs work to be added (delimiters, indicating what it actually is, etc). There's a discussion to be had here, but either way that should be a separate PR, not part of this one.

diesel/src/query_builder/sql_query.rs
ZeGentzy: Made it symmetrical.

diesel/src/query_builder/sql_query.rs
ZeGentzy: Removed

ZeGentzy:
Done changes.

ZeGentzy:
Say, what's this waiting on?

weiznich:
> Say, what's this waiting on?

[:point_up:](https://github.com/diesel-rs/diesel/pull/1975#discussion_r252708806)

ZeGentzy:
Rebased against master. Pleeeeeeeeease can we get this merged?

diesel/src/query_builder/sql_query.rs
```diff
@@ -76,36 +77,53 @@ impl SqlQuery {
...
+    ///
+    /// This allows doing things you otherwise couldn't do, e.g. `bind`ing in a
+    /// loop.
+    pub fn into_boxed<'f, DB: Backend>(self) -> BoxedSqlQuery<'f, DB, Self> {
+        BoxedSqlQuery::new(self)
+    }
+
+    /// Appends a piece of SQL code at the end.
+    pub fn sql<T: Into<String>>(self, sql: T) -> SqlQuery<Self> {
+        SqlQuery::new(self, sql.into())
```
weiznich: Why do we create a new wrapper type here and not just append to sql string of the existing one, similar to `BoxedSqlQuery`?

diesel/src/query_builder/sql_query.rs
```diff
@@ -164,3 +190,81 @@ where
...
+        self.binds
+            .push(Box::new(move |mut out| out.push_bind_param_value_only(&b)));
+        self
+    }
+
+    /// See [`SqlQuery::sql`].
+    ///
+    /// [`SqlQuery::sql`]: ../struct.SqlQuery.html#method.sql
+    pub fn sql<T: Into<String>>(mut self, sql: T) -> Self {
+        self.sql += &sql.into();
```
weiznich: It should be possible to avoid calling `Into<String>` here, so that we could save one allocation.
Maybe something like:

```rust
pub fn sql<T: AsRef<str>>(mut self, sql: T) -> Self {
        self.sql += sql.as_ref()
        self
)
```

(The same should be done for the sql function of `SqlQuery`)

diesel/src/query_builder/sql_query.rs
ZeGentzy: No reason, fixed.

weiznich:
👍 Thanks

PR #1976, opened 2019-01-31
Restructure `Backend::RawValue` to allow non-references

sgrif:
As part of Diesel 2.0 we want to change all our backends which currently
have `type RawValue = [u8]` to instead return some opaque struct which
can possibly carry additional information. There are a lot of motivating
use cases for that change, such as including the OID on PG to allow for
working with dynamic schemas, and reducing our reliance on implicit
casts in `libmysqlclient` as attempted by #1457.

Regardless of what these opaque structs do or don't do, we will need to
move away from the signature of `&DB::RawValue`. If we keep this
signature, we would basically be forcing all backends to do what the
SQLite backend did prior to 718a32ea.

In an ideal world, [GATs] would be stable and we would just write `type
RawValue<'a>;`. However, they aren't stable. Implementation on them
appears to have barely started, so it's unlikely that they will be
stable by the time 2.0 comes out. (If it does, or appears close to
happening, we should switch to using that before releasing 2.0).

This introduces a new trait, `HasRawValue`, which acts as the "GAT" form
of the `RawValue` type. The choice to make this a separate trait vs
adding a lifetime to `Backend` was very explicit. Most code which
involves the `Backend` trait doesn't care about the type of `RawValue`,
and adding a lifetime would require virtually all bounds on `Backend`
(including the code which actually does care about `RawValue`) to change
`DB: Backend` to `DB: for<'a> Backend<'a>`.

This does have the unfortunate side effect of breaking code that isn't
currently relying on the type of `Backend` in any concrete fashion.
There's no way around this without keeping the signature of `FromSql` as
`&Something` which is a burden we really don't want to have. The
`backend::RawValue` helper type has been introduced as a forwards
compatibility shim, so code using this new helper type will continue to
work if we switch to GATs in 3.0. The only code which couldn't use this
is the code which was generic over `RawValue = [u8]`. I expect we can
abstract that away in a follow up PR (more on that below).

This was spun off from #1833 as an alternative to a major piece of that
PR which I was unhappy with. One important note is that the motivating
use case that #1833 is trying to address naturally means we can no
longer have the generic "I have some bytes please don't make me rewrite
the impls for integers/floats/byte arrays" impls. This PR is meant to be
standalone against master, so it does not attempt to address everything
done in #1833, and therefore includes some ugly bounds to keep these
impls around. I would actually like to see us keep them if possible, and
I think we can do so with an explicit trait implementation. But that
will be experimented in another PR.

[GATs]: https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md

### DIFF
--- a/diesel/src/backend.rs
+++ b/diesel/src/backend.rs
@@ -32,6 +32,7 @@ where
     Self: HasSqlType<sql_types::Date>,
     Self: HasSqlType<sql_types::Time>,
     Self: HasSqlType<sql_types::Timestamp>,
+    Self: for<'a> HasRawValue<'a>,
 {
     /// The concrete `QueryBuilder` implementation for this backend.
     type QueryBuilder: QueryBuilder<Self>;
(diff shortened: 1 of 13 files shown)


weiznich:
cc @Thomspoon

diesel/src/backend.rs
```diff
@@ -41,17 +42,27 @@ where
...
     ///
     /// This type is only used if `RawValue` is `[u8]`.
     type ByteOrder: ByteOrder;
 }
 
+/// The raw representation of a database value given to `FromSql`.
+///
+/// This trait is separate from `Backend` to imitate `type RawValue<'a>`. It
+/// should only be referenced directly by implementors. Users of this type
+/// should instead use the [`RawValue`] helper type instead.
```
weiznich: This (and the next) `RawValue` link is not rendered correctly.

Thomspoon:
This seems cleaner than the FamilyLT/RefFamily solution that we came up with in #1833

sgrif:
Fixed the doc links (apparently intra-rustdoc links don't work on type aliases) if someone wants to merge when CI is green.

PR #1978, opened 2019-02-04
Make `migrations_internals::setup_database` public

jonas-schievink:
This is the only additional method you need to call for automatically setting up an empty database from inside the application.

### DIFF
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -256,7 +256,7 @@ fn migration_with_version(
     }
 }
 
-#[doc(hidden)]
+/// Creates all diesel-internal tables and data (if needed).
 pub fn setup_database<Conn: Connection>(conn: &Conn) -> QueryResult<usize> {
     create_schema_migrations_table_if_needed(conn)
 }


weiznich:
`migrations_internals` is (as the name suggests) not considered part of the public API, therefore that's not the right place to expose that function. 
The public API resides in `diesel_migration`, therefore the "stabilisation" should happen there.

cc @diesel-rs/contributors Do we want to expose such a functionality?

sgrif:
I'd be interested in hearing more about the use case since any function that interacts with migrations should be calling this already

jonas-schievink:
My usecase was for running a test suite on a different database, without having the user deal with migrations. The test runner would run all needed migrations automatically and execute each test in a test transaction that is never committed.

sgrif:
You don't need this function in order to do that. You can just call `run_pending_migrations` and it will do this for you

jonas-schievink:
Ah yeah, I just checked the code again and I'm calling `any_pending_migrations` before `run_pending_migrations` to log a message when migrations need to be run, and *that* fails if the migrations table doesn't exist. `setup_database` will fix that.

I could live without this though, so feel free to close if you think this doesn't make sense. Or maybe the better fix would be to call `setup_migrations` from `any_pending_migrations`.

sgrif:
Yeah, let's just fix `any_pending_migrations` to either call this function or short circuit if the database isn't set up

PR #1981, opened 2019-02-09
Allow setting a specific test bin

kpcyrd:
This is related to packaging, I noticed that every test paniced with no specific error and it turned out this is because I ran `cargo test --release` instead of `cargo test`.

I'm currently running `cargo test` for that reason instead, which means diesel is built twice, once for packaging and once for testing. This patch would allow me to specify the release binary instead. It also adds some context in case this fails for any reason.

### DIFF
--- a/diesel_cli/tests/support/command.rs
+++ b/diesel_cli/tests/support/command.rs
@@ -37,7 +37,10 @@ impl TestCommand {
     }
 
     pub fn run(self) -> CommandResult {
-        let output = self.build_command().output().unwrap();
+        let output = self
+            .build_command()
+            .output()
+            .expect("failed to execute process");
         CommandResult { output: output }
     }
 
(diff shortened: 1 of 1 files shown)


sgrif:
I'm not sure I understand the use case of running our test suite in release mode

kpcyrd:
I'm basically trying to do something like this:

```bash
build() {
  cd "diesel-${pkgver}/${pkgname}"
  cargo build --release
}

check() {
  cd "diesel-${pkgver}/${pkgname}"
  DIESEL_TEST_BIN=../target/release/diesel_cli cargo test --release --no-default-features --features sqlite
}

package() {
  cd "diesel-${pkgver}"
  install -Dm755 "target/release/diesel" "${pkgdir}/usr/bin/diesel"

  install -d "${pkgdir}/usr/share/bash-completion/completions" \
             "${pkgdir}/usr/share/zsh/site-functions" \
             "${pkgdir}/usr/share/fish/vendor_completions.d"
  "${pkgdir}/usr/bin/diesel" completions bash > "${pkgdir}/usr/share/bash-completion/completions/diesel"
  "${pkgdir}/usr/bin/diesel" completions zsh > "${pkgdir}/usr/share/zsh/site-functions/_diesel"
  "${pkgdir}/usr/bin/diesel" completions fish > "${pkgdir}/usr/share/fish/vendor_completions.d/diesel.fish"

  install -Dm644 LICENSE-MIT -t "${pkgdir}/usr/share/licenses/${pkgname}"
}
```

When the package is built, `build()`, `check()` and `package()` are executed in that order. Running `cargo build` in release mode and `cargo test` in debug mode means the whole dependency tree is built twice. Since I already built everything in release mode, building the tests in release mode as well would significantly speedup the build.

sgrif:
I guess I'm just not following the need to run Diesel's test suite outside of developing Diesel

thomasetter:
> I'm not sure I understand the use case of running our test suite in release mode

There is sometimes slightly different behavior in release and debug mode which can lead to bugs.
It makes sense to test all compile modes (also in the CI) to decrease the probability of introducing hard-to-debug issues.

weiznich:
This seems to be reasonable for me + this allows to workaround cases like having a custom target dir :+1:

PR #1982, opened 2019-02-09
Remove all deprecated code

sgrif:
This removes all items marked `#[deprecated]` and all code with
`#[cfg(feature = "with-deprecated")]`. Unfortunately the compiler
wouldn't warn in all cases of deprecated items, particularly deprecated
re-exports. This hit us extremely hard in the case of the deprecated
re-export of `Queryable` in `prelude`, and `#[derive(Queryable)]` would
actually generate invalid code if the `with-deprecated` feature was
missing.

While I did change a lot of our code to work with the re-export of
`Queryable` removed, I wasn't actually able to remove it since
`diesel_infer_schema` breaks. We can remove it once we're no longer
using `infer_schema`, but we should also consider whether that
deprecation is worthwhile for the problems it caused.


diesel/src/query_source/mod.rs
```diff
@@ -13,13 +13,6 @@ use query_builder::*;
 pub use self::joins::JoinTo;
 pub use self::peano_numbers::*;
 
-#[cfg(feature = "with-deprecated")]
-#[deprecated(since = "1.1.0", note = "Use `deserialize::Queryable` instead")]
-pub use deserialize::Queryable;
```
weiznich: Removing this breaks the CI for diesel with the postgres feature enabled.

diesel/src/sql_types/mod.rs
```diff
@@ -403,16 +403,6 @@ pub trait HasSqlType<ST>: TypeMetadata {
     /// of this method should not do dynamic lookup unless absolutely necessary
     fn metadata(lookup: &Self::MetadataLookup) -> Self::TypeMetadata;
 
-    #[doc(hidden)]
-    #[cfg(feature = "with-deprecated")]
-    #[deprecated(
-        since = "1.4.0",
-        note = "This method is no longer used, and has been deprecated without replacement"
```
weiznich: Something like this is currently required for the oracle backend 😟

diesel/src/query_source/mod.rs
sgrif: Whoops this is the one I meant to leave in

diesel/src/sql_types/mod.rs
sgrif: We should discuss this more and see if there is a better option for this

weiznich: I've opend #1985 for this.

weiznich:
@sgrif CI says this needs a rustfmt run.

PR #1989, opened 2019-02-17
Introduce a derive for `NonAggregate`

sgrif:
This introduces a derive for the `NonAggregate` trait, which we're
planning on removing as laid out in
https://discourse.diesel.rs/t/proposed-change-replace-nonaggregate-with-something-less-broken-that-can-support-group-by/18.
The purpose of this derive is mostly to reduce the churn caused by the
changes laid out there in the most common case. Because we don't yet
have a PR opened for that change, and it's not clear that is actually
the direction we will end up going, I have written this as a standalone
change that can be shipped in terms of documentation and CHANGELOG
entries.

As part of this, I reviewed every impl of `NonAggregate`. A *lot* of
these impls were missing the bounds required to make them sound. That's
been fixed for most of them by using the new derive. A few others that
can't use the derive have been fixed manually. In a lot of cases there
was also a useless `Self: Expression` bound which is no longer required
and has been removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ## Unreleased
 
+### Added
+
+* `NonAggregate` can now be derived for simple cases.
+
 ### Removed
 
 * All previously deprecated items have been removed.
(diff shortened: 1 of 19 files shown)


sgrif:
I had meant to name the derive `ValidGrouping` since the plan was to reduce churn... But I'm also starting to hate that name, so I think we should go with this for the derive, if we do go through with the plan laid out in Discourse we can have a poorly named derive to start, and then rename the derive when we decide on a final name.

sgrif:
Not sure if it's worth an issue since ultimately one of us needs to look into it, I'm not sure we will just accept a PR from someone saying they investigated and it's fine.

I don't feel strongly about it though if you want to open an issue.

PR #1992, opened 2019-02-21
Use more compact encoding for NUMERIC with leading group(s) of 0

mguillemot-tel:
Diesel is not using the most compact possible encoding for NUMERIC values that end with at least one group of "0000" in their decimal representation.

For example, the value `'100000000'::numeric` can be encoded two ways using the Postgres binary protocol:

1/ `weight==2 scale==0 sign==0 digits==[1, 0, 0]`
2/ `weight==2 scale==0 sign==0 digits==[1]`

The Postgres server accepts both encodings, but the second one should be preferred since it is more compact on the wire (it is also the encoding used by the official `libpq` client library).

We discovered this because some other Postgres-compatible servers (CockroachDB in our case) handle the first form (used by Diesel) incorrectly, leading to data corruption issues. To be fair, this is a problem on their side, and not a sufficient reason to change Diesel's behaviour, but as this PR demonstrates, producing the compact format actually makes the Diesel code simpler AND more efficient.

This PR has been tested on both PostgreSQL 10 and CockroackDB 2.1.5, and works in both cases.

### DIFF
--- a/diesel/src/pg/types/numeric.rs
+++ b/diesel/src/pg/types/numeric.rs
@@ -104,10 +104,7 @@ mod bigdecimal {
             let weight = digits.len() as i16 - digits_after_decimal as i16 - 1;
 
             let unnecessary_zeroes = if weight >= 0 {
-                let index_of_decimal = (weight + 1) as usize;
                 digits
-                    .get(index_of_decimal..)
-                    .expect("enough digits exist")
                     .iter()
                     .rev()
                     .take_while(|i| i.is_zero())
@@ -185,7 +182,7 @@ mod bigdecimal {
             let expected = PgNumeric::Positive {
                 weight: 1,
                 scale: 0,
-                digits: vec![1, 0],
+                digits: vec![1],
             };
             assert_eq!(expected, decimal.into());
 
@@ -201,7 +198,7 @@ mod bigdecimal {
             let expected = PgNumeric::Positive {
                 weight: 2,
                 scale: 0,
-                digits: vec![1, 0, 0],
+                digits: vec![1],
             };
             assert_eq!(expected, decimal.into());
         }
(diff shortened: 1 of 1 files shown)


JohnTitor:
Sorry for being late. When I re-ran travis, this may require to rebase.

mguillemot-tel:
Rebased on latest master.

mguillemot-tel:
CI tests seem to be failing for an unrelated reason. Should I try to fix those here even though it's unrelated?

JohnTitor:
@mguillemot-tel Nightly pipelines are sometimes failed regardless of the PR. I want to check stable pipelines and Travis mainly.

weiznich:
The CI passes on other PR's so I don't see any reason why it shouldn't pass here.

JohnTitor:
Hmm, the log seems unrelated
```
error: variable does not need to be mutable
   --> diesel_cli/src/main.rs:430:17
    |
430 |             let mut file = fs::File::create(path)?;
    |                 ----^^^^
    |                 |
    |                 help: remove this `mut`
    |
note: lint level defined here
   --> diesel_cli/src/main.rs:2:9
    |
2   | #![deny(warnings, missing_copy_implementations)]
    |         ^^^^^^^^
    = note: #[deny(unused_mut)] implied by #[deny(warnings)]

error: aborting due to previous error
```

We should fix this but I think to separate pull-requests.

mguillemot-tel:
Sorry for the bump, is there a plan to merge this PR since it seems to be ok for everyone?

weiznich:
Is there any official documentation available for this?

mwilliammyers:
Are there still plans to merge this? If so, what can I do to help push it over the edge? Is adding those explicit test cases a blocker?

weiznich:
@mwilliammyers This seems fine for me, I've probably just missed it back then. As there was no merge conflict I've merged it now.

PR #1999, opened 2019-02-23
Update dotenv to 0.13 (or allow it when it was a range already)

Eijebong:


weiznich:
CI says that there is coming a 0.10.1 dotenv version from somewhere that fails to build.

Eijebong:
Yeah, it's coming from infer_schema_macros 1.4.0 which is used in diesel_infer_schema, no idea why it makes compilation fail though...

mehcode:
With the state that dotenv is in I'd rather see it get removed or rewritten and replaced. Hope the original creator is doing okay.

JohnTitor:
ping @Eijebong, any updates?

examples/postgres/advanced-blog-cli/src/main.rs
```diff
@@ -4,6 +4,7 @@ extern crate bcrypt;
 extern crate chrono;
 #[macro_use]
 extern crate diesel;
+extern crate failure;
```
JohnTitor: At least, this order is bad(failed in `cargo fmt`).

examples/postgres/advanced-blog-cli/src/main.rs
```diff
@@ -194,6 +195,7 @@ fn register_user(conn: &PgConnection) -> Result<(), Box<Error>> {
 }
 
 fn convert_auth_error(err: auth::AuthenticationError) -> Box<Error> {
+    use failure::Fail;
```
JohnTitor: Same here

examples/postgres/advanced-blog-cli/src/main.rs
```diff
@@ -204,7 +206,7 @@ fn convert_auth_error(err: auth::AuthenticationError) -> Box<Error> {
         NoPasswordSet => {
             "No password given. You need to set the BLOG_PASSWORD environment variable.".into()
         }
-        EnvironmentError(e) => e.into(),
+        EnvironmentError(e) => e.compat().into(),
```
AlterionX: I may have missed something, but why does this need the failure crate if we're only using it here? Is there a way to do this without the crate?

Just trying to avoid adding another dependency since it doesn't seem to be used in many places.

weiznich:
Already fixed by #2528

PR #2007, opened 2019-03-11
Bumping minimum supported barrel version

spacekookie:
This PR bumps the minimum version of barrel that is supported.
There have recently been a few fixes in the integration and also
starting with `0.5.0` the API will become more stable, leading up
to the `1.0.0` release soon™


spacekookie:
That's kinda by design. I would argue that users always being on the latest `barrel` version is preferable to having to re-do PRs updating the maximum version

sgrif:
Sorry this took so long to review.

Just so folks are clear, @spacekookie owns any code related to barrel integration and gets the final say -- It's not maintained by the Diesel team, it's just in our code base because it sorta has to be to allow this sort of plugin functionality.

PR #2010, opened 2019-03-13
ungate try_from because it has been stabilized

KeenS:
See also https://github.com/rust-lang/rust/pull/58302.
This fixes failures of other CIs


azure-pipelines.yml
```diff
@@ -178,7 +178,8 @@ jobs:
     - template: _build/install-rust.yml
       parameters:
         platform: Linux
-        rust_version: nightly-2018-11-27
+        # least nightly release that stabilize try_from
+        rust_version: nightly-2019-02-26
```
KeenS: Updated nightly version to pass test. Is this acceptable? To say more, what's the policy of supported nightly version?

azure-pipelines.yml
```diff
@@ -178,7 +178,8 @@ jobs:
     - template: _build/install-rust.yml
       parameters:
         platform: Linux
-        rust_version: nightly-2018-11-27
+        # least nightly release that stabilize try_from
```
sgrif: I don't think we need this comment

azure-pipelines.yml
sgrif: Yeah, that's fine. This is only the compile tests. We don't officially support nightly, we just bump this periodically as needed to be able to run the compile tests (which require nightly)

diesel/src/lib.rs
```diff
@@ -93,7 +93,7 @@
 //! You can come ask for help at
 //! [gitter.im/diesel-rs/diesel](https://gitter.im/diesel-rs/diesel)
 
-#![cfg_attr(feature = "unstable", feature(specialization, try_from))]
+#![cfg_attr(feature = "unstable", feature(specialization))]
```
sgrif: We should remove the gating on the uses of this feature as well (grep `try_from`)

KeenS: try_from has been stabilized (ungated) at nightly but it is still nightly only API.

azure-pipelines.yml
KeenS: I see.

diesel/src/lib.rs
sgrif: Ah I see

sgrif: Can you update the FIXME comment to specify which Rust version we can use it?

diesel_compile_tests/rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2018-11-27
+nightly-2019-02-26
```
weiznich: That changes the output of our compile tests, which means you need to go though that errors and adjust them. (There is a script for that)

diesel/src/pg/types/numeric.rs
```diff
@@ -156,7 +153,7 @@ mod bigdecimal {
 
     impl FromSql<Numeric, Pg> for BigDecimal {
         fn from_sql(numeric: Option<&[u8]>) -> deserialize::Result<Self> {
-            // FIXME: Use the TryFrom impl when try_from is stable
+            // FIXME: Use the TryFrom impl after Rust 1.34 get released.
```
sgrif: Rust 1.34 is released

diesel/src/pg/types/numeric.rs
```diff
@@ -112,7 +108,8 @@ mod bigdecimal {
                     .expect("enough digits exist")
                     .iter()
                     .rev()
-                    .take_while(|i| i.is_zero())
+                    .cloned()
```
sgrif: Why do we need to clone this now?

weiznich: Good question. Got an error on my first try so added cloned there. Now that error is gone if I remove it again 🤷‍♂️

diesel/src/pg/types/numeric.rs
```diff
@@ -153,7 +152,6 @@ mod bigdecimal {
 
     impl FromSql<Numeric, Pg> for BigDecimal {
         fn from_sql(numeric: Option<&[u8]>) -> deserialize::Result<Self> {
-            // FIXME: Use the TryFrom impl after Rust 1.34 get released.
```
sgrif: Shouldn't this actually do what this comment says? We can get rid of that function and just put the body in the `TryFrom` impl

weiznich: Right, didn't read the code, sorry 😥

diesel/src/pg/types/numeric.rs
weiznich: Ok, that was not that easy.
So clippy suggests to replace the closure with `Zero::is_zero`. The iterator yields `&&i16`, `Zero` is implemented for `i16`, `is_zero` takes the value by reference. So this closure works due to autoderef, whereas passing the function directly (as clippy suggests) does not work because there is no impl for `&i16`.
So there are two options here:
* Ignore that clippy warning
* Clone all values (not that expensive on a iterator over `i16`)

sgrif: I don't really have a preference as long as we open an issue with Clippy

sgrif: (Personally I hate this lint and wouldn't mind just turning it off)

weiznich: That's interesting: It's [not reproducible](https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=6d82809d4a8fd6552a2498bb000718ea) by simply creating an iterator over `&&i16` and it's [not reproducible by creating](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6de2bc8d499ef1b3b510f05296d4d278). Also tried locally with the a newer nightly. That does not emit this warning, so this seems to be already fixed.
Anyway, I've added a comment to [this issue](https://github.com/rust-lang/rust-clippy/issues/3942)

PR #2014, opened 2019-03-17
Convert type names from snake_case to CamelCase

### LINKED ISSUE #2012: Converting snake_case to CamelCase
### Versions

- **Rust:** rustc 1.33.0-nightly (b43986184 2019-01-11)
- **Diesel:** 1.4.1
- **Database:** PostgreSQL 10.4
- **Operating System** macOS 10.14

### Feature Flags

- **diesel:** postgres

## Problem Description
snake_case types in Diesel get converted to an unusual `Capitalized_snake_case`

I'm trying to add a postgres enum to my schema. I'm defining the type name as `action_type`, but when it gets added to the schema it's converted to `Action_type`. IF I try using CamelCase in the SQL it gets converted to Actiontype (no surprise since case is generally ignored in SQL)

`ActionType`

`Action_type`

Nope

### Steps to reproduce
Here's the migration that generates the types and corresponding tables:
```
CREATE TABLE roles (
id SERIAL PRIMARY KEY,
name VARCHAR NOT NULL
);

CREATE TYPE ActionType AS ENUM ('create', 'read', 'update', 'delete');
CREATE TYPE ActionModifier AS ENUM ('all', 'self');

CREATE TABLE permissions (
id SERIAL PRIMARY KEY,
role_id INTEGER REFERENCES roles(id) NOT NULL,
resource_name VARCHAR,
action ActionType,
modifier ActionModifier
);

CREATE TABLE user_roles (
id SERIAL PRIMARY KEY,
user_id INTEGER REFERENCES users(id) NOT NULL,
role_id INTEGER REFERENCES roles(id) NOT NULL
);
```

Or just check [this](https://github.com/NicholasLYang/diesel-camel-case) out (you'll need to add an .env file with the DATABASE_URL).
## Checklist

NicholasLYang:
Fixes #2012
I was writing some postgres enums and I couldn't help but notice that my `action_type` enum was getting converted to `Action_type`, which imo, is a little ugly. I just wrote up a quick snake_case to CamelCase function and have it convert any types.

One potential issue is any built in types that are snake_case. So far I've only found `txid_snapshot` in Postgres. Any idea if this will cause a problem? I can special case that situation if so.

### DIFF
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -3,6 +3,7 @@ use prelude::*;
 
 /// Determines the OID of types at runtime
 #[allow(missing_debug_implementations)]
+#[repr(transparent)]
 pub struct PgMetadataLookup {
     conn: PgConnection,
 }


kiljacken:
I can recommend using [heck](https://crates.io/crates/heck) instead of a homebrew solution.

NicholasLYang:
Sounds like a good idea!

diesel_cli/src/infer_schema_internals/.#pg.rs
```diff
@@ -0,0 +1 @@
+nicholas@172-18-33-126.dynapool.nyu.edu.356
```
kiljacken: Including this looks like a mistake, you might want to remove it :)

NicholasLYang: Oops. Will do

weiznich:
In my opinion we should not use a "complicated" algorithm there to change those generated names, because no matter how advanced such an algorithm is there will be cases where this does not work. Beside of that it is quite simple to just change the name in the generated file by using the patch mechanism. 

That said: I would like to hear the opinion from some other @diesel-rs/core members here before doing a final decision.

sgrif:
> In my opinion we should not use a "complicated" algorithm there to change those generated names, because no matter how advanced such an algorithm is there will be cases where this does not work. 

I agree with this sentiment in general (e.g. pluralization), but converting snake case to camel case seems fine to me.

> Beside of that it is quite simple to just change the name in the generated file by using the patch mechanism.

This is true, but patch files are also a pain to work with -- especially for this kind of case. Before we make this argument, I think we need to improve that experience. Better support for generating them (e.g. you just modify the schema file however you want and run `diesel generate-schema-patch` or something), making merge conflicts less likely to occur (if any of the context lines have changed it'll fail, not sure if falling back to a 3 way merge will help), and giving a better workflow if a merge conflict we can't resolve does arise

derekdreery:
For evidence of prior art: wasm-bindgen (a pretty official project) [uses heck](https://github.com/rustwasm/wasm-bindgen/blob/5ae6de5e9e2656c73f5fb309be1cafcc8ca061a1/crates/webidl/src/util.rs#L52).

derekdreery:
FWIW, the way I would like this to work for enums is given the schema

```sql
CREATE TYPE my_type AS ENUM ('foo', 'bar');
CREATE TABLE custom_types (
    id SERIAL PRIMARY KEY,
    custom_enum my_type NOT NULL
);
```

generate

```rust
table! {
    use diesel::sql_types::*;
    use super::MyTypeRaw;
    custom_types {
        id -> Integer,
        custom_enum -> MyTypeRaw,
    }
}

#[derive(SqlType)]
#[postgres(type_name = "my_type")]
struct MyTypeRaw;

#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, FromSqlRow, AsExpression)]
#[sql_type = "MyTypeRaw"]
pub enum MyType {
    Foo,
    Bar,
}

impl ToSql<MyTypeRaw, Pg> for MyType {
    fn to_sql<W: Write>(&self, out: &mut Output<W, Pg>) -> serialize::Result {
        match *self {
            MyType::Foo => out.write_all(b"foo")?,
            MyType::Bar => out.write_all(b"bar")?,
        }
        Ok(IsNull::No)
    }
}

impl FromSql<MyTypeRaw, Pg> for MyType {
    fn from_sql(bytes: Option<&[u8]>) -> deserialize::Result<Self> {
        match not_none!(bytes) {
            b"foo" => Ok(MyType::Foo),
            b"bar" => Ok(MyType::Bar),
            _ => Err("unrecognized enum variant".into()),
        }
    }
}
```

where the `Raw` suffix is bikesheaddable.

derekdreery:
I'm just going to ping here to say I very much think this should be merged. The CI errors are from the stabilization of TryFrom/To, and so master will also fail CI.

weiznich:
This needs a testcase and an entry to the changelog.

NicholasLYang:
@weiznich Could you walk me through the process of adding a test? I looked at diesel_test and `schema_inference.rs` seems like the best place to put the tests, but I'm a little confused on how the test infrastructure works. Is there an example schema that we're testing against? I assume I should add an enum there and then test if it gets renamed properly.

weiznich:
The right place to add a test case for this is in my opinion [here](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/tests/print_schema.rs). Add a new testcase there and add the corresponding sql migration and the expected generated schema.rs [here](https://github.com/diesel-rs/diesel/tree/master/diesel_cli/tests/print_schema). Hopefully that setup is quite self explaining.

dimensi:
Any news?

derekdreery:
I'm going to **ping** this because I believe it is ready for merge. The test failures are spurious. I miss this functionality because it makes me do patching to my schema.rs file, and I don't really want to do that!

weiznich:
@derekdreery  @dimensi It's missing the requested test cases.

derekdreery:
@weiznich gotcha

@NicholasLYang Do you mind if I finish this off?

NicholasLYang:
@derekdreery, no, go for it! I've just been busy with school. I actually added a test case but got stuck on getting the tests to pass on my machine (without my new test, that is).

derekdreery:
@NicholasLYang You could always push your test and see what the CI says?

NicholasLYang:
Oops, this got closed for some reason. Reopened here #2060

PR #2018, opened 2019-03-19
Wrap all math operations in parens

### LINKED ISSUE #2017: Mathematical operations need to wrap in parenthesis
### Versions

- **Rust:** any
- **Diesel:** any
- **Database:** any
- **Operating System** any

### Feature Flags

- **diesel:**

## Problem Description

Parenthesis inserted in Rust to override order of operations are lost in the generated SQL. This can also occur when a value is assigned to a local variable. This will at best result in an incorrect value being generated, and at worst will result in a runtime error because of incorrect types (e.g. `interval * (integer - integer)` is valid but `interval * integer - integer` is not)

`14`

`10`

### Steps to reproduce

```
fn test(conn: &Connection) {
    let x = select(2.into_sql::<Integer>() * (3.into_sql::<Integer>() + 4))
        .get_result(conn);
    assert_eq!(Ok(14), x);
}
```

## Checklist

sgrif:
Previously if you used local variables or parens in Rust to override the
default order of operations, we would completely lose that information
in the generated SQL. This would at best result in an unexpected value,
but at worse would cause a query like `now + 2.minutes() * (some_int +
1)` to result in a runtime error complaining that there's no operator
for `interval + integer`

I intend to backport this to 1.4

Fixes #2017

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   have been. All types in Diesel are now correctly only considered
   non-aggregate if their parts are.
 
+* Parenthesis are now inserted around all mathematical operations. This means
+  that `(2.into_sql() + 3) * 4` will correctly evaluate to 20 as expected.
+  Previously we would generate SQL that evaluated to 14. This could even result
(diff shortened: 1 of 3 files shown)


Eijebong:
Lgtm with ci fixed

PR #2020, opened 2019-03-25
R2D2 transaction check

mikethetike:
Add a check to see if there is a transaction still open before giving the connection back to the pool.

We found that if the thread panics, the connection kept its locks, and potentially the next thread to pick up the connection could commit the data even if it was incorrect. There has been some fixes to R2D2 to check whether the thread has panicked, but this check should also catch any cases where a developer may have not called commit.

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -94,6 +94,7 @@ where
 
     fn is_valid(&self, conn: &mut T) -> Result<(), Error> {
         if conn.transaction_manager().get_transaction_depth() > 0 {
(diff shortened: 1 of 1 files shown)


sgrif:
Thanks, but this is something we're trying to fix on the R2D2 side. Just refusing to put a connection in the pool if there's an open transaction is problematic, since it's common to create a pool with a size of 1 in tests (where you want the open transaction returned). You can follow https://github.com/sfackler/r2d2/issues/31 and https://github.com/sfackler/r2d2/pull/70

PR #2023, opened 2019-03-28
Store sign information in `Mysql::TypeMetadata`

sgrif:
This is the concrete implementation of the proposal described at
https://discourse.diesel.rs/t/proposed-change-add-unsigned-types-to-mysqltype/66.
I have opted for a different implementation than originally imagined --
Rather than adding variants to `MysqlType`, I have kept the flag
separate. MySQL is unlikely to add variants to the `enum_field_types`
enum for backwards compatibility, but I don't want to rule out them
paying attention to the `is_signed` flag for existing variants in the
future.

This makes the impact greater in some ways, and lesser in others. Folks
who were exhaustively matching on `MysqlType` will not have their code
break as they would have in the original code. However, anyone
implementing or calling `<Mysql as HasSqlType<_>>::metadata` will have
to work with this new struct if they were manually constructing or
maniuplating the result.

Ultimately the expected impact is the same as before -- nobody. Unless
someone is writing their own Mysql connection adapter, there's no reason
folks would have been working with this type in the first place.

The proposal left whether we should make the enum non-exhaustive as an
unresolved question. I've opted to leave it exhaustive, since exhaustive
matching is a worthwhile thing to do, and I don't expect libmysqlclient
to add new variants on their end. Under similar logic, I've opted to
allow destructuring/public construction of the new struct. The only
field that we aren't currently mapping is `is_null`, which doens't
currently make sense in our type system. I think before 2.0 ships, we'll
want to add an `is_null` field on this struct, which could be used by a
theoretical future `Null` type. With this field, we will have covered
everything in libmysqlclient's API, which I don't see them adding to,
and thus I think we are safe leaving this struct public

Since the unsigned flag is now passed around as a named field of a
struct, I don't think the enum used before carries any value over a
boolean


diesel/src/mysql/bind_collector.rs
```diff
@@ -1,15 +1,15 @@
...
 use result::QueryResult;
 use serialize::{IsNull, Output, ToSql};
-use sql_types::{HasSqlType, IsSigned};
+use sql_types::HasSqlType;
 
 #[derive(Default)]
 #[doc(hidden)]
 #[allow(missing_debug_implementations)]
 pub struct MysqlBindCollector {
-    pub(crate) binds: Vec<(MysqlType, IsSigned, Option<Vec<u8>>)>,
```
weiznich: Any reason why we need to keep the custom bind collector? Wouldn't it be easier to just use `RawBytesBindCollector` here?

sgrif:
Removed the mysql bind collector

PR #2028, opened 2019-04-03
More general implementations of Connection trait for r2d2 PooledConne…

### LINKED ISSUE #2027: Implement `Connection` for all `r2d2::PooledConnection<T> where T: r2d2::ManageConnection`
Currently `Connection` is implemented for `r2d2::PooledConnection<ConnectionManager<C>> where C: Connection`.

However, sometimes you might want to use a custom connection manager. In this case there is no way to implement the `Connection` trait for `r2d2::PooledConnection<MyConnectionManager>` because of the coherence rules.

Instead, diesel could provide a broader implementation for `r2d2::PooledConnection<T> where T: r2d2::ManageConnection, T::Connection: Connection`.

Diggsey:
…ction

Fixes #2027

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -97,22 +97,24 @@ where
     }
 }
 
-impl<T> SimpleConnection for PooledConnection<ConnectionManager<T>>
+impl<M> SimpleConnection for PooledConnection<M>
 where
-    T: Connection + Send + 'static,
+    M: ManageConnection,
+    M::Connection: Connection + Send + 'static,
 {
     fn batch_execute(&self, query: &str) -> QueryResult<()> {
         (&**self).batch_execute(query)
     }
 }
 
-impl<C> Connection for PooledConnection<ConnectionManager<C>>
+impl<M> Connection for PooledConnection<M>
 where
-    C: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
-    C::Backend: UsesAnsiSavepointSyntax,
+    M: ManageConnection,
+    M::Connection: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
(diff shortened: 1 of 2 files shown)


kiljacken:
The change could probably be made less noisy by using `M` (for manager), or even just staying with `C` as the type parameter instead of `T`, making the actual changes more visible.

Diggsey:
Updated

sgrif:
Seems reasonable. Thanks!

sgrif:
Can you add a changelog entry for this?

Diggsey:
Done.

diesel/src/r2d2.rs
```diff
@@ -97,22 +97,24 @@ where
...
     }
 }
 
-impl<C> Connection for PooledConnection<ConnectionManager<C>>
+impl<M> Connection for PooledConnection<M>
 where
-    C: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
-    C::Backend: UsesAnsiSavepointSyntax,
+    M: ManageConnection,
+    M::Connection: Connection<TransactionManager = AnsiTransactionManager> + Send + 'static,
```
weiznich: Is there any reason why we need `TransactionManager = AnsiTransactionManager` here? That will prevent connection implementation that use another TransactionManager from using this impl. (Yes I know that the original version had the same restrictions.…)

sgrif: Yes. If we remove it we end up with:

```
C: Connection,
C::TransactionManager: TransactionManager<Self>,
```

but the definition of `TransactionManger<C>` requires `C: Connection`, so we've got an infinite loop. https://github.com/rust-lang/rfcs/blob/master/text/1647-allow-self-in-where-clauses.md specifies behavior in this case, but that portion of the RFC has not been implemented

Diggsey:
I'm assuming these test failures are spurious and I don't need to do anything more here?

PR #2035, opened 2019-04-15
Bump bigdecimal-rs to 0.1.0

vorot93:


weiznich:
Is it possible to continue to support the old versions?

vorot93:
Perhaps it is.

Issue #2040, opened 2019-04-18
Replace `http` with `https` in some files


JohnTitor:
## Summary

We use domain `diesel.rs` and can visit not just `http://diesel.rs` but also `https://diesel.rs`. This is not important change, but we might have to unify these url in some files (e.g. `Cargo.toml`), perhaps `https`?

(And we can replace url in repository's website line.)

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)

TomPridham:
looks like this can be closed?

PR #2041, opened 2019-04-18
Replace `http` with `https`

JohnTitor:
I replaced `diesel.rs` and `docs.diesel.rs`'s `http` with `https` in some files.

cc #2040

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -930,13 +930,13 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// * `child_table` is the Table with the Foreign key.
 ///
-/// So given the Table decaration from [Associations docs](http://docs.diesel.rs/diesel/associations/index.html)
+/// So given the Table decaration from [Associations docs](https://docs.diesel.rs/diesel/associations/index.html)
 ///
 /// * The parent table would be `User`
 /// * The child table would be `Post`
 /// * and the Foreign key would be `Post.user_id`
 ///
-/// For joins that do not explicitly use on clauses via [`JoinOnDsl`](http://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
+/// For joins that do not explicitly use on clauses via [`JoinOnDsl`](https://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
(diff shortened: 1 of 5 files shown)


diesel/src/macros/mod.rs
```diff
@@ -928,13 +928,13 @@ macro_rules! __diesel_table_query_source_impl {
 ///
 /// * `child_table` is the Table with the Foreign key.
 ///
-/// So given the Table decaration from [Associations docs](http://docs.diesel.rs/diesel/associations/index.html)
+/// So given the Table decaration from [Associations docs](https://docs.diesel.rs/diesel/associations/index.html)
```
weiznich: Those link should rather be relative instead of absolute

```suggestion
/// So given the Table decaration from [Associations docs](associations/index.html)
```

diesel/src/macros/mod.rs
```diff
@@ -928,13 +928,13 @@ macro_rules! __diesel_table_query_source_impl {
...
 ///
 /// * The parent table would be `User`
 /// * The child table would be `Post`
 /// * and the Foreign key would be `Post.user_id`
 ///
-/// For joins that do not explicitly use on clauses via [`JoinOnDsl`](http://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
+/// For joins that do not explicitly use on clauses via [`JoinOnDsl`](https://docs.diesel.rs/diesel/prelude/trait.JoinOnDsl.html)
```
weiznich:
```suggestion
/// For joins that do not explicitly use on clauses via [`JoinOnDsl`](prelude/trait.JoinOnDsl.html)
```

diesel/src/macros/mod.rs
JohnTitor: Thank you, I changed these.

diesel/src/macros/mod.rs
AlterionX: For this, does simply ``[`JoinOnDsl`]`` or ``[`JoinOnDsl`](crate::prelude::JoinOnDsl)`` work? I remember using crate relative links and they worked preety nicely.

JohnTitor: I think it's okay as it is.

PR #2056, opened 2019-05-07
Add `#[repr(transparent)]` to `PgMetadataLookup`

weiznich:
Otherwise the constructor exploits implementation defined behaviour by
assuming a specific memory layout which may change in the future

### DIFF
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -3,6 +3,7 @@ use prelude::*;
 
 /// Determines the OID of types at runtime
 #[allow(missing_debug_implementations)]
+#[repr(transparent)]
 pub struct PgMetadataLookup {
     conn: PgConnection,
 }


diesel/src/pg/metadata_lookup.rs
```diff
@@ -3,6 +3,7 @@ use prelude::*;
 
 /// Determines the OID of types at runtime
 #[allow(missing_debug_implementations)]
+#[repr(transparent)]
 pub struct PgMetadataLookup {
     conn: PgConnection,
```
Eijebong: Doesn't that need to be repr something too then ? (the `PgConnection`)

weiznich: Not as far as I understand `#[repr(transparent)]`. 

In my understanding `#[repr(transparent)]` requires that the type as exactly one not zero sized type. Then it is valid to apply `#[repr(transparent)]` and it's guaranteed that this type has exactly the same layout as the inner wrapped type. That means this type is only a compile time construct that could be transmuted/pointer casted/… away, which is exactly what we are doing a few lines below. Beside of that `#[repr(transparent)]` does not require the inner type to have any specific layout. 
[RFC](https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md)

Anyway, while rereading this code I'm asking myself why we even needed that new type wrapper in the first place 🤔

Eijebong: I'm more concerned about passing a non `#[repr(c)]` struct as a pointer.I'm not sure I get all the repr stuff though

weiznich: We don't pass it as pointer anywhere.
That pointer cast mess below is just a `transmute` variant that was suggested by clippy. Or have we copied that casting from [stdlib](https://doc.rust-lang.org/src/std/path.rs.html#1798-1800)?

Anyway I think passing a non `#[repr(C)]` struct as pointer is fine as long as you don't assume anything about the inner layout of such structs. `#[repr(transparent)]` just allows us to assume that both types are the same.

Eijebong: Oh right, I just saw the `as *const PgConnection as *const PgMetadataLookup` when I grepped for it. I guess this is fine then

PR #2059, opened 2019-05-09
Add SimpleConnection::as_pg_connection()

Boscop:
This PR adds a new method `as_pg_connection` to the `SimpleConnection` trait:
```rust
    fn as_pg_connection(&self) -> Option<&PgConnection>;
```
The purpose of this is to allow Rust structs that impl `Migration` to do something useful.
Why is this method necessary? Currently, `Migration::run` only gets a `SimpleConnection` which only provides the method [`batch_execute`](https://docs.rs/diesel/1.4.0/diesel/connection/trait.SimpleConnection.html?search=#tymethod.batch_execute) that takes a raw sql query string and returns `QueryResult<()>`, thus it cannot be used to fetch values from any table, convert them in Rust code and then update the table with the new values (which is our primary use case for wanting to have migrations written in Rust).
With this method, it allows us to downcast the `SimpleConnection` to a `PgConnection` so that we can do useful (e.g. data-converting) migrations with it (e.g. re-hashing passwords or other things that wouldn't be easily possible in PL/pgSQL).

E.g. with this method, we can then write:
```rust
trait RustMigration {
    const VERSION: &'static str;

    fn run(&self, conn: &PgConnection) -> Result<(), RunMigrationsError>;

    ...
}

struct RustMigrationWrapper<T: RustMigration>(T);

impl<T: RustMigration> Migration for RustMigrationWrapper<T> {
    fn version(&self) -> &str {
        T::VERSION
    }

    fn run(&self, conn: &dyn SimpleConnection) -> Result<(), RunMigrationsError> {
        if let Some(conn) = conn.as_pg_connection() {
            <T as RustMigration>::run(&self.0, conn)
        } else {
            // Err
        }
    }

    ...
}
```

This is the first step towards being able to integrate Rust-based migrations with embedded migrations.
The next step is to extend the `embed_migrations` macro to accept as argument a list of structs that impl `Migration`, and to include those (chronologically) in the embedded migration array (`ALL_MIGRATIONS`).

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -10,6 +10,8 @@ use deserialize::{Queryable, QueryableByName};
 use query_builder::{AsQuery, QueryFragment, QueryId};
 use result::*;
 use sql_types::HasSqlType;
+#[cfg(feature = "postgres")]
+use crate::pg::PgConnection;
 
 #[doc(hidden)]
 pub use self::statement_cache::{MaybeCached, StatementCache, StatementCacheKey};
@@ -24,6 +26,10 @@ pub trait SimpleConnection {
     /// This function is used to execute migrations,
     /// which may contain more than one SQL statement.
     fn batch_execute(&self, query: &str) -> QueryResult<()>;
+
+    /// Return the underlying `PgConnection` if it is one.
+    #[cfg(feature = "postgres")]
+    fn as_pg_connection(&self) -> Option<&PgConnection>;
 }
 
 /// A connection to a database
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -5,13 +5,13 @@ mod transaction_manager;
 
 use std::fmt::Debug;
 
+#[cfg(feature = "postgres")]
+use crate::pg::PgConnection;
 use backend::Backend;
 use deserialize::{Queryable, QueryableByName};
 use query_builder::{AsQuery, QueryFragment, QueryId};
 use result::*;
 use sql_types::HasSqlType;
-#[cfg(feature = "postgres")]
-use crate::pg::PgConnection;
 
 #[doc(hidden)]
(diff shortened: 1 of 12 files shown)


diesel/src/connection/mod.rs
```diff
@@ -10,6 +10,8 @@ use deserialize::{Queryable, QueryableByName};
 use query_builder::{AsQuery, QueryFragment, QueryId};
 use result::*;
 use sql_types::HasSqlType;
+#[cfg(feature = "postgres")]
+use crate::pg::PgConnection;
```
JohnTitor: According to rustfmt, we should correct ordering.

weiznich:
Thanks for your PR :+1:
I totally understand the motivation for this feature, but I think this is the wrong approach. 
IMHO a API fitting well into diesel should full fill the following points:

* It must work with all three backend, not only with one
* It should try to minimize the backend related code (i.e sprinkling `#[cfg(feature = "backend")]` around in central trait definitions is a bad idea in my opinion)
* It should support out of tree third party backends
* If it's a breaking change it requires a impact analysis for existing code

Unfortunately I haven't currently the time and bandwidth to think about how exactly such an API would look like, but I'm willing to discuss a proposed API that would full fill those points.

> This is the first step towards being able to integrate Rust-based migrations with embedded migrations.
The next step is to extend the embed_migrations macro to accept as argument a list of structs that impl Migration, and to include those (chronologically) in the embedded migration array (ALL_MIGRATIONS).

IMHO this does not belong into the existing `embed_migrations!` macro, at least not in the first step. This should rather be done as new macro, maybe in a third party crate to get some experience on the API.

sgrif:
Strong agree with the points @weiznich made

weiznich:
This PR contains adds a wild mix of different functionality. For each feature there should be an own PR. Also none of the concerns mentioned above are addressed so I'm closing this now.
Feel free to reopen this or a new PR as as there is any change.

PR #2061, opened 2019-05-14
Allow migrations to have attached metadata, avoid running in transaction

msrd0:
This is #1685 from @sgrif rebased onto master.

On PostgreSQL, certain commands cannot be run inside of a transaction,
such as `CREATE INDEX CONCURRENTLY` and `ALTER TYPE`. This causes
problems when attempting to use Diesel migrations, since we always run
these migrations inside a transaction.

To allow opting out of this, we introduce the concept of migration
metadata. I've opted to keep this incredibly general, rather than adding
a specific `should_run_in_migration` function, as there's been interest
expressed in having this capability to enable alternate migration
runners.

Since we need to retain object safety, we can't actually have any
defined structure for the metadata. The best we can do is either
something that is conceptually `Map<String, String>`, or an enum like
`serde_json::Value` or `toml::Value`. I really don't want to restrict
what can or cannot be in here, so I've opted for the more general
option.


diesel/src/migration/mod.rs
```diff
@@ -49,7 +63,27 @@ impl<'a> Migration for &'a Migration {
...
+/// being used.
+///
+/// For Diesel's built in SQL file migrations, metadata is stored in a file
+/// called `metadata.toml`. Diesel looks for a single key, `run_in_transaction`.
+/// By default, all migrations are run in a transaction on SQLite and
+/// PostgreSQL. This behavior can be disabled for a single migration by setting
+/// this to `false`.
+pub trait Metadata {
+    /// Get the metadata at the given key, if present
+    fn get(&self, key: &str) -> Option<Cow<str>>;
```
sgrif: I think I'd like a more strongly typed API here. Something like:

```rust
fn get<T>(&self, key: &str) -> Option<Result<T, MigrationError>>
where
    T: serde::DeserializeOwned;
```

The TOML impl can then be:

```rust
fn get<T>(&self, key: &str) -> Option<Result<T, MigrationError>>
where
    T: DeserializeOwned,
{
    self.0.get(key)
        .map(|v| v.try_into::<T>().map_err(|e| MigrationError::InvalidMetadata(Box::new(e))))
}
```

Then we can drop the manual boolean parsing code, and more complex consumers can tightly couple themselves to deeper data structures if they want.

msrd0:
@sgrif I changed it as you requested so that the `get` method now returns a generic type.

Since returning generic types in trait methods is not so trivial, I had to add `Self : Sized` restrictions. This required the `Migration` trait to become generic as well, since it now had to return a generic type of `Metadata + Sized` in its `metadata()` function.

To keep the `embed_migrations!` macro working I introduced `DummyMetadata` so that a `Migration` type that doesn't want to specify metadata at all can then use `impl Migration<DummyMetadata>`.

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -348,17 +351,8 @@ where
...
-        Some("false") | None => false,
-        Some(_) => {
-            return Err(MigrationError::InvalidMetadata(
-                "run_in_transaction must be a boolean".into(),
-            )
-            .into())
-        }
-    };
+        .and_then(|m| m.get("run_in_transaction"))
+        .unwrap_or(Ok(false))?;
```
msrd0: @sgrif You set the default for `run_in_transaction` to false [here](https://github.com/diesel-rs/diesel/pull/1685/files#diff-ed7f6c154205d63bf776e00598c41eb9R338), is that the intended behaviour? Or would this possibly break migrations that rely on the previous behaviour of always running inside a transaction?

sgrif:
Adding a generic type to `Migration` kinda defeats the whole purpose of using trait objects. :) I'd forgotten that we need `Metadata` to be object safe. WDYT about having `get` return `&Any` instead?

msrd0:
Well that's true, I wasn't entirely happy about that solution as well.

I've never used `Any` in rust before, so I played around with it a bit. However, using `&str` seems to be better because I could then manually deserialize it to whatever type I want, although I might just not know how to get it work ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4920940e6896a2e12f30db636b03928f))

sgrif:
You still to represent as some concrete type internally, like we are today and then covert when `get` is called.

e.g.

```rust
fn get(&self, key: &str) -> Option<&dyn Any> {
    use toml::Value::*;

    match self.0.get(key)? {
        String(s) => s,
        Integer(i) => i,
        // etc
    }
}
```

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -348,19 +348,13 @@ where
...
-        Some(_) => {
-            return Err(MigrationError::InvalidMetadata(
-                "run_in_transaction must be a boolean".into(),
-            )
-            .into())
-        }
-    };
+        .and_then(|m| {
+            m.get("run_in_transaction")
+                .and_then(|x: &dyn Any| x.downcast_ref::<bool>())
```
sgrif: This should return an error if the value wasn't a boolean, not silently ignore it.

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -177,11 +177,17 @@ fn run_sql_from_file(conn: &SimpleConnection, path: &Path) -> Result<(), RunMigr
...
-    fn get(&self, key: &str) -> Option<Cow<str>> {
-        self.0.get(key).map(|v| {
-            v.as_str()
-                .map(Into::into)
-                .unwrap_or_else(|| v.to_string().into())
+    fn get(&self, key: &str) -> Option<&dyn Any> {
+        use toml::Value::*;
+
+        self.0.get(key).map(|v| match v {
+            String(s) => s as &dyn Any,
```
sgrif: All of these can be `as _`

msrd0:
Ok, will add a test tomorrow.

Still wondering if changing the default behaviour from always in transaction to run outside a transaction unless explicitly requested is intended.

sgrif:
No that is unintentional.

msrd0:
Ok so I've added the test and changed the default behaviour to run inside a transaction unless specified otherwise.

msrd0:
@sgrif any updates?

msrd0:
@weiznich Done. Spacing could be a bit more unified, e.g. in `diesel/src/migration/mod.rs`, where there is sometimes a newline between methods and sometimes not. Definitely out of scope for this PR though.

weiznich:
CI says there is one more `dyn` missing:

```
   --> diesel_migrations/migrations_internals/src/migration.rs:159:35
    |
159 |     fn metadata(&self) -> Option<&Metadata> {
    |                                   ^^^^^^^^ help: use `dyn`: `dyn Metadata`
    |
```

diesel/src/migration/errors.rs
```diff
@@ -44,9 +45,24 @@ impl Error for MigrationError {
...
             }
+            MigrationError::InvalidMetadata(_) => "Migration metadata was invalid",
             MigrationError::__NonExhaustive => unreachable!(),
         }
     }
+
+    fn cause(&self) -> Option<&dyn Error> {
+        match *self {
+            MigrationError::IoError(ref err) => Some(err),
+
```
weiznich:
```suggestion
```

diesel/src/migration/errors.rs
```diff
@@ -44,9 +45,24 @@ impl Error for MigrationError {
...
+            MigrationError::InvalidMetadata(ref err) => Some(&**err),
+
```
weiznich:
```suggestion
```

msrd0:
@weiznich true. What do you suggest? Should we simply include the content of the `metadata.toml` file and parse it upon request of metadata, or can/should we parse the toml compile-time and include the `toml::Value` struct, or even map it to some other type (which makes the implementation more complicated, but removes the dependency on the toml crate for parsing/representing metadata)?

weiznich:
As long as the toml dependency is reeported as implementation detail through `diesel_migration` I'm fine with any of the proposed solutions. Just try the simplest.

msrd0:
@weiznich I've decided to map the input toml file into a precompiled `toml::Value` using reexports to make the implementation as easy as possible while keeping error handling minimal. However, I found no better method to serialize `Datetime` values other than using `Display`/`FromStr` because the internal fields of `Datetime` aren't exposed to the public in any way. Also, I had to increase the recursion limit to be able to add the `metadata()` method to the `Migration` impl.

diesel_migrations/migrations_macros/src/embed_migrations.rs
```diff
@@ -99,8 +108,70 @@ fn migration_literal_from_path(path: &Path) -> Result<proc_macro2::TokenStream,
...
         up_sql: include_str!(#sql_file_path),
+        metadata: #metadata
     }))
 }
+
+fn migration_metadata_from_path(path: &Path) -> Result<proc_macro2::TokenStream, Box<dyn Error>> {
+    let mut buf = Vec::new();
+    let mut file = File::open(path)?;
+    file.read_to_end(&mut buf)?;
+    let value = TomlMetadata::from_slice(&buf)?.0;
```
weiznich: I think it is clearer and much shorter to just do the following:
```rust
quote! {
    TomlMetadata::from_slice(#buf)?
}
```
especially because we fallback to string paring  for datetime.

msrd0:
I'm afraid I never had the time to finish this

Issue #1459, opened 2018-01-07
`left_join` incorrectly compiling if additional joins are done afterwards


sgrif:
Reported via gitter. This query compiles successfully:

```rust
build::table
        .left_join(revision::table)
        .inner_join(job::table)
        .select((
                job::name,
                build::build_no,
                revision::hash,
        ))
        .load::<Build>(conn)
```

It should fail to compile unless `.nullable` has been called on `revision::hash`. Reversing the order of the join calls causes this code to fail to compile as it should

sgrif:
I actually don't know if we can fix this until disjointness based on associated types lands. We need to change this impl:

```
        impl<Left, Right> SelectableExpression<
            Join<Left, Right, Inner>,
        > for $column_name where
            $column_name: AppearsOnTable<Join<Left, Right, Inner>>,
            Join<Left, Right, Inner>: AppearsInFromClause<$($table)::*, Count=Once>,
        {
        }
```

to be these two impls:

```

        impl<Left, Right> SelectableExpression<
            Join<Left, Right, Inner>,
        > for $column_name where
            $column_name: SelectableExpression<Left>,
            Right: AppearsInFromClause<$($table)::*, Count=Never>,
        {
        }

        impl<Left, Right> SelectableExpression<
            Join<Left, Right, Inner>,
        > for $column_name where
            $column_name: SelectableExpression<Right>,
            Left: AppearsInFromClause<$($table)::*, Count=Never>,
        {
        }
```

I need to think about this more to come up with a way to fix this today.

Ten0:
I have the same issue.
The following incorrectly compiles : 

```sql
CREATE TABLE "integrations" (
	"id" Integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
	"label" Character Varying ( 100 ) NOT NULL UNIQUE,
	"field_z" Text [] NOT NULL
);

CREATE TABLE "user" ( 
	"id" Integer GENERATED ALWAYS AS IDENTITY,
	"name" Character Varying( 128 ) NOT NULL UNIQUE,
	"integration"  Integer NOT NULL REFERENCES integrations(id),
	PRIMARY KEY ( "id" )
);

CREATE TABLE "params1" (
	"retailer_id" Integer NOT NULL REFERENCES users(id) PRIMARY KEY,
	"field_x" Boolean NOT NULL
);

CREATE TABLE "params2" (
[...]

PR #2063, opened 2019-05-14
`NULL` does not deserialize to `false` on Postgres

Ten0:
This could happen due to #1459.
We're better off triggering an error rather than having `false`s propagating in the rest of the application logic.


weiznich:
See @sgrif comments on #1459, 26720d7 and #104 why this is a bad idea.

sgrif:
Yeah, this isn't something we can do unless you have a plan to avoid regressing issues from 4 years ago.

PR #2069, opened 2019-05-16
Interpret paths in diesel.toml as relative to the project root

### LINKED ISSUE #2057: Running diesel migrations generates schema.rs at path that depends on cwd
We would expect to have it generated at a path that depends on the project root (src/schema.rs) but if I run diesel migrations while in a subdirectory, it will create an src directory and a schema.rs file whereever I am.

Is this intended ?

sgrif:
I went ahead and also changed the search for project root to look for
`diesel.toml` first, which should improve our behavior in workspaces.

Fixes #2057.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,14 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   any columns from `posts`. That will now fail to compile, and any selections
   from `posts` will need to be made explicitly nullable.
 
+* Diesel CLI will now look for `diesel.toml` to determine the project root
+  before looking for `Cargo.toml`.
+
+* Any relative paths in `diesel.toml` will now be treated as relative to the
+  project root (the directory containing either `diesel.toml` or `Cargo.toml`).
+  They are no longer dependent on the current working directory (for all
+  directories in the same project)
+
 ### Deprecated
 
 * `diesel_(prefix|postfix|infix)_operator!` have been deprecated. These macros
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,14 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   have been. All types in Diesel are now correctly only considered
   non-aggregate if their parts are.
 
+* Diesel CLI will now look for `diesel.toml` to determine the project root
+  before looking for `Cargo.toml`.
+
+* Any relative paths in `diesel.toml` will now be treated as relative to the
+  project root (the directory containing either `diesel.toml` or `Cargo.toml`).
```
AlterionX: I think that this clarification makes it a little more confusing. It might be more effective to state in the previous paragraph what the project root is, then just reuse the term, especially since there can be multiple `Cargo.toml` files (which this PR is trying to fix anyways).

diesel_cli/src/database_error.rs
```diff
@@ -38,8 +38,8 @@ impl From<result::ConnectionError> for DatabaseError {
 impl Error for DatabaseError {
     fn description(&self) -> &str {
         match *self {
-            CargoTomlNotFound => {
-                "Unable to find Cargo.toml in this directory or any parent directories."
+            ProjectRootNotFound => {
+                "Unable to find diesel.toml or Cargo.toml in this directory or any parent directories."
```
AlterionX: It may be useful to indicate what "this" directory is. I don't know much about efficiency if you include a `format!` here, though. That would entail passing in either the url into the error, or something similar, which might cause enum size differences? But I'm not sure.

sezna:
I was looking on the pipeline to see what failed and Azure says the build wasn't found -- does anybody know why it failed the pipeline?

PR #2076, opened 2019-05-25
Rename extension traits according to rfc 445

toshokan:
as of rfc 445, the the Rust convention for extension traits is `Ext` suffix. This PR makes `OptionalExtension` conform to this convention.


weiznich:
I'm not sure if that's enough reason to do such a breaking change, given that this will likely have a bigger impact.

toshokan:
It could always be done next time there is a larger breaking change, but since the contribution guide recommends the Rust style guide I think it's reasonable to conform where possible. 

Since this is in `prelude` I can't imagine many users are depending on the name of the trait explicitly anyway.

weiznich:
That's all fine, but this remains to be a breaking change. So we cannot simply change out of the blue. This needs a more extended motivation + a migration strategy and so on.
Given all that is a lot of work I'm not sure that is justified for this purely cosmetic change.

sgrif:
Yeah I agree, this seems like needless breakage without a strong motivation

weiznich:
Closed because it seems to be the consent of the core that this is needless breakage.

PR #2077, opened 2019-05-28
Allow "out" columns

### LINKED ISSUE #2074: Schema fails to compile when column is named `out`
### Versions

- **Rust:** 1.35.0
- **Diesel:** 1.4.2

## Problem Description

When using `print-schema` on a database with columns named "out", the resulting schema will fail to compile with many errors like the following:

```
error[E0530]: function parameters cannot shadow unit structs
   --> src\schema.rs:287:1
    |
287 |    table! {
    |   _^
    |  |_|
    | ||
288 | ||     execution_instance (id) {
289 | ||         id -> Binary,
290 | ||         time_executed -> Nullable<Datetime>,
...   ||
306 | ||     }
307 | || }
    | || ^
    | ||_|
    | |__cannot be named the same as a unit struct
    |    the unit struct `out` is defined here
    |
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

## Checklist

Diggsey:
Fixes #2074

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -23,10 +23,11 @@ macro_rules! __diesel_column {
             DB: $crate::backend::Backend,
             <$table as QuerySource>::FromClause: QueryFragment<DB>,
         {
-            fn walk_ast(&self, mut out: $crate::query_builder::AstPass<DB>) -> $crate::result::QueryResult<()> {
-                $table.from_clause().walk_ast(out.reborrow())?;
-                out.push_sql(".");
-                out.push_identifier($sql_name)
+            #[allow(non_snake_case)]
+            fn walk_ast(&self, mut __out: $crate::query_builder::AstPass<DB>) -> $crate::result::QueryResult<()> {
+                $table.from_clause().walk_ast(__out.reborrow())?;
+                __out.push_sql(".");
+                __out.push_identifier($sql_name)
             }
         }
 
@@ -808,9 +809,10 @@ macro_rules! __diesel_table_impl {
                 impl<DB: Backend> QueryFragment<DB> for star where
                     <table as QuerySource>::FromClause: QueryFragment<DB>,
                 {
-                    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
-                        table.from_clause().walk_ast(out.reborrow())?;
-                        out.push_sql(".*");
+                    #[allow(non_snake_case)]
+                    fn walk_ast(&self, mut __out: AstPass<DB>) -> QueryResult<()> {
+                        table.from_clause().walk_ast(__out.reborrow())?;
+                        __out.push_sql(".*");
                         Ok(())
                     }
                 }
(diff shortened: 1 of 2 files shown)


diesel/src/macros/mod.rs
```diff
@@ -23,10 +23,11 @@ macro_rules! __diesel_column {
             DB: $crate::backend::Backend,
             <$table as QuerySource>::FromClause: QueryFragment<DB>,
         {
-            fn walk_ast(&self, mut out: $crate::query_builder::AstPass<DB>) -> $crate::result::QueryResult<()> {
-                $table.from_clause().walk_ast(out.reborrow())?;
-                out.push_sql(".");
-                out.push_identifier($sql_name)
+            #[allow(non_snake_case)]
+            fn walk_ast(&self, mut diesel__out: $crate::query_builder::AstPass<DB>) -> $crate::result::QueryResult<()> {
```
sgrif: In procedural macros we tend to just prefix with `__`. Do you think it's worth being consistent with that convention here?

diesel_cli/src/infer_schema_internals/inference.rs
```diff
@@ -14,6 +14,12 @@ static RESERVED_NAMES: &[&str] = &[
     "type", "typeof", "unsafe", "unsized", "use", "virtual", "where", "while", "yield",
 ];
 
+fn is_reserved(name: &str) -> bool {
+    RESERVED_NAMES.contains(&name) || (
+        name.starts_with("diesel__") && !name.ends_with("_")
```
sgrif: I'm a little confused by the `ends_with` check. It seems unrelated?

diesel/src/macros/mod.rs
Diggsey: To me, a leading underscore is an indication that the variable is not necessarily used, as it will disable "unused variable" warnings in the compiler.

Also, it seems plausible that people could have columns that start with an underscore. It seems less plausible that they will have columns starting `diesel__`.

diesel_cli/src/infer_schema_internals/inference.rs
Diggsey: It's not strictly necessary, it's just that we append an underscore as a way to make something "not reserved", so it seemed simpler and more consistent to just treat everything ending in an underscore as not a reserved word.

diesel/src/macros/mod.rs
Diggsey: (To clarify: reserving this namespace is a breaking change because it will change people's existing schema, and reserving `diesel__` seems less likely to cause problems than reserving `__`)

sgrif: > Also, it seems plausible that people could have columns that start with an underscore.

Right, which is why we use double leading underscore

Diggsey: I do not feel particularly strongly either way so have updated the PR.

Diggsey:
These build failures seem unrelated to my changes.

PR #2081, opened 2019-05-31
Trait objects without dyn are deprecated in the latest nightly, so replace all of them

weiznich:


weiznich:
Not sure why the azure build is failing. I will try to retrigger it later.

PR #2083, opened 2019-06-03
Mysql UUID support

Diggsey:
Adds support for querying UUIDs to/from Mysql Binary columns.

### DIFF
--- a/diesel/src/mysql/types/mod.rs
+++ b/diesel/src/mysql/types/mod.rs
@@ -3,6 +3,8 @@
 #[cfg(feature = "chrono")]
 mod date_and_time;
 mod numeric;
+#[cfg(feature = "uuidv07")]
+mod uuid_v0_7;
 
 use byteorder::WriteBytesExt;
 use std::io::Write;
--- a/diesel/src/mysql/types/uuid_v0_7.rs
+++ b/diesel/src/mysql/types/uuid_v0_7.rs
@@ -1,0 +1,59 @@
+extern crate uuidv07 as uuid;
+
+use std::io::prelude::*;
+
+use deserialize::{self, FromSql};
+use mysql::Mysql;
+use serialize::{self, IsNull, Output, ToSql};
+use sql_types::Binary;
+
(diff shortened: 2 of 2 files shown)


weiznich:
In general your policy for the implementation of `FromSql` /` ToSql` for certain types in diesel is something like the three following questions:

*   Is this a commonly used type. (Yes because there is already support for `Uuid` in diesel.
*   Could we provide a `FromSql` and a `ToSql` implementation for that type. (In this case Yes)
*   Could the conversion go somehow wrong because the sql type or the rust type could hold values that are not representable in the other type. (In this case the conversion from `Vec<u8>` to `Uuid` could go wrong, because not every arbitrary byte sequence is a uuid. Therefore the corresponding constructor returns a `Result`)

Only if we could answer those three questions with:

*   It is a commonly used type
*   We could provide both conversions
*   Nothing could go wrong

we add support for those types.

Because the third question is not answered positively I do not think we want to add support for this conversion (Maybe the members of @diesel-rs/core have a different opinion here?)

Diggsey:
I see. It's unfortunate because in practice, my conversion cannot fail (my columns are `BINARY(16)`).

weiznich:
Unfortunately we cannot enforce this on diesel's side.

theduke:
I know this has been the policy for a while.

Personally I think this badly hurts the usability of Diesel.

I've run into this with `u32/u64` in Postgres, `Uuid` in Mysql and Sqlite, ...

It requires either creating additional structs with the real type, and writing (Try)From/Into impls for easy conversion or coming up with a wrapper type, like shown in https://github.com/diesel-rs/diesel/issues/2098.

Both solutions are incredibly annoying in practice and introduce a lot of boilerplate.

I understand the argument - types should be coherent with the underlying database representation.

But I think the value is limited. Yes, you may get errors due to invalid type mappings. 
But the worst thing that would happen is a conversion error when querying data. And it can be circumvented with a custom wrapper type anyway.

In my opinion, this decision should be left to the developer and  the  strict coherence policy should be loosened in favour of usability.

PR #2088, opened 2019-06-05
Cache custom types OIDs on Postgres

### LINKED ISSUE #2071: Postgres: We should avoid sending one query per custom type bind
(From gitter)

@Ten0: 
I'm using a custom type in postgres in an app where I need performance and minimal database charge.
However, the serialization system seems to be sending queries everytime I serialize one of them:
`SELECT "pg_type"."oid", "pg_type"."typarray" FROM "pg_type" WHERE "pg_type"."typname" = $1 LIMIT $2`
This results in a very large amount of queries that should be avoided.
Do you know why that is, and how I can avoid that ?

@weiznich : 
Postgres requires passing the oid of a bind/return value while executing query. For builtin types those oid's are static so they are hard coded in diesel itself. For custom types they my change, therefore we need to query the oid on use. Probably we could do some caching there so that this query is only executed once, but it seems like nobody hit that as a performance problem yet. I'm willing to provide some mentoring on that if you want.

@Ten0:
I'm very interested :) How would you go about implementing this ?

@weiznich: 
The building blocks are already here. PgMetadataLookup needs to be extended to contain/access some kind of connection local lookup table for cashing the results of the queries executed in fn lookup_type(). I would think a HashMap<String, Oid> will be sufficient there.
Additionally we need to figure out a way to store that lookup table inside of a PgConnection and then pass everything together to the relevant functions. (That's the hard part.) The probably right approach is to look at the change sean has done to Backend::RawValue diesel-rs/diesel@ddf5b46 and do something similar with the Metadata lookup infrastructure. (Which allows us having lifetimes there, which should simplify everything)

DL;DR rest is implementations discussions

@Ten0:
 That is exactly what I would have done (though not as complex as what is done with RawValue). One of my issues issue was that prepare_query takes a non-mutable reference to the connection and yet has to edit its cache, and I was wondering what was the best way to get around this and what were the constraints on PgConnection (Sync?...), but I've realized that's the same thing for the query cache, which uses a RefCell, so I would use that. (is that ok?)
My second issue is that the DB::MetadataLookup has to hold a reference to the PgConnection, which means it can't be directly stored in PgConnection itself. So I would still instantiate it in prepare_query and just pass it a reference to the RefCell<HashMap<...>> stored in PgConnection. Storing it directly would not be very flexible and somewhat exposes the internals of PgMetadataLookup, which is unclean. So we would need to create an intermediate structure (PgMetadataCache) holding the RefCell<HashMap<...>> and stored in PgConnection, which then passes a reference to the RefCell to a newly instantiated PgMetadataLookup, but then we get to a point where we have RefCells with sub-structures that hold references on RefCells and it starts to feel somewhat unclean as well.
Do you think it's acceptable to change the public API a bit by making the &PgMetadataLookups mut in addition to the previously described implementation, to get rid of the &RefCell ? This would aply to AstPass that is already accessible mutably whereever it is used (correct me if I'm wrong), and it would seem cleaner to me than having an &RefCell roaming around. Plus it exposes the idea that this may be used to store cache, which would indeed be one of its purposes. Otherwise, would you be satisfied with the previously described implementation ?

@weiznich: 
> I was wondering what was the best way to get around this and what were the constraints on PgConnection (Sync?...), but I've realized that's the same thing for the query cache, which uses a RefCell, so I would use that. (is that ok?)

PgConnection is Send but not Sync and that should not be changed, so using a RefCell seems fine.

> Storing it directly would not be very flexible and somewhat exposes the internals of PgMetadataLookup, which is unclean. So we would need to create an intermediate structure (PgMetadataCache) holding the RefCell<HashMap<...>> and stored in PgConnection, which then passes a reference to the RefCell to a newly instantiated PgMetadataLookup, but then we get to a point where we have RefCells with sub-structures that hold references on RefCells and it starts to feel somewhat unclean as well.

Using a wrapper struct for RefCell<HashMap<…>> seems like a good idea, but I'm not sure I understand where the second RefCell is coming from. I do not think changing the API from &PgMetadataLookup to &mut PgMetadataLookup is a great idea because that would require a lot of code to be changed.

> Otherwise, would you be satisfied with the previously described implementation ?

From the rough outline this sounds good, but it's easier to discuss such things on the actual changes on a pull request :wink:

--- 
So here's the tracking issue.

Additionally:  
> I'm not sure I understand where the second RefCell is coming from

Right, we could just store a reference to the `PgMetadataCache`, but it still means moving around a `!Sync` value that could be `Sync` and some having some `store_metadata(&self)` which feels weird to me since it obviously mutates data. I read the code and it felt like only ~10 lines would need to be changed but I gues we could try the implementation in a PR and see how it feels.

Opening a PR soon.

Ten0:
Resolves #2071

This avoids sending one query per custom type bind, which caused a lot of back and forth to the database when inserting a large numbers of lines that used a custom type.

Note: The implementation was also discussed on the issue (#2071).

### DIFF
--- a/diesel/src/pg/mod.rs
+++ b/diesel/src/pg/mod.rs
@@ -17,7 +17,6 @@ mod transaction;
 
 pub use self::backend::{Pg, PgTypeMetadata};
 pub use self::connection::PgConnection;
-pub use self::metadata_lookup::PgMetadataCache;
 pub use self::metadata_lookup::PgMetadataLookup;
 pub use self::query_builder::DistinctOnClause;
 pub use self::query_builder::PgQueryBuilder;
--- a/diesel/src/pg/mod.rs
+++ b/diesel/src/pg/mod.rs
@@ -17,6 +17,7 @@ mod transaction;
 
 pub use self::backend::{Pg, PgTypeMetadata};
 pub use self::connection::PgConnection;
+pub use self::metadata_lookup::PgMetadataCache;
 pub use self::metadata_lookup::PgMetadataLookup;
 pub use self::query_builder::DistinctOnClause;
 pub use self::query_builder::PgQueryBuilder;
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -14,7 +14,7 @@ use self::result::PgResult;
 use self::stmt::Statement;
 use connection::*;
 use deserialize::{Queryable, QueryableByName};
-use pg::{Pg, PgMetadataCache, PgMetadataLookup, TransactionBuilder};
+use pg::{metadata_lookup::PgMetadataCache, Pg, PgMetadataLookup, TransactionBuilder};
 use query_builder::bind_collector::RawBytesBindCollector;
 use query_builder::*;
 use result::ConnectionError::CouldntSetupConfiguration;
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -14,7 +14,7 @@ use self::result::PgResult;
 use self::stmt::Statement;
 use connection::*;
 use deserialize::{Queryable, QueryableByName};
-use pg::{Pg, PgMetadataLookup, TransactionBuilder};
+use pg::{Pg, PgMetadataCache, PgMetadataLookup, TransactionBuilder};
 use query_builder::bind_collector::RawBytesBindCollector;
 use query_builder::*;
 use result::ConnectionError::CouldntSetupConfiguration;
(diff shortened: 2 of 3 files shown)


diesel/src/pg/mod.rs
```diff
@@ -17,6 +17,7 @@ mod transaction;
 
 pub use self::backend::{Pg, PgTypeMetadata};
 pub use self::connection::PgConnection;
+pub use self::metadata_lookup::PgMetadataCache;
```
weiznich: That should not be public, it's an implementation detail.
```suggestion
```

Ten0: Should be good now :)

Ten0:
Test fails are due to `rand` being down. x)

https://github.com/rust-random/rand/issues/818

JohnTitor:
I just made azure and travis re-run.

PR #2108, opened 2019-07-07
Replace ONCE_INIT with Once::new()

JohnTitor:
`ONCE_INIT` will be deprecated in the future release (see https://github.com/rust-lang/rust/pull/61757).
For the future, we can now use `Once::new()`.

### DIFF
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -3,7 +3,7 @@ extern crate mysqlclient_sys as ffi;
 use std::ffi::CStr;
 use std::os::raw as libc;
 use std::ptr::{self, NonNull};
-use std::sync::{Once, ONCE_INIT};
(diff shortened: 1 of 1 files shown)


JohnTitor:
If we don't allow deprecated items in #2117, I think we should merge this to make CI green. What about? @sgrif

JohnTitor:
Uh, I cannot re-run pipelines but it should be green.

sgrif:
Rerun isn't doing anything for me either. Can you close/re-open?

JohnTitor:
Hmm, the `inner__join` test is failed in macOS_mysql. This also appears in #2126.

JohnTitor:
It seems the reason is MySQL version. `8.0.16` passes but `8.0.17` fails.

weiznich:
Could you rebase this PR on top of #2131 to fix the CI?

PR #2117, opened 2019-07-12
Allow some deprecated items

killercup:
These were deprecated recently but we don't want to use them yet to not
break compatibility with older rustc versions unnecessary. Explicitly
allowing them should fix the build for nightly with `#[deny(warnings)]`.

### DIFF
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -3,7 +3,7 @@ extern crate mysqlclient_sys as ffi;
 use std::ffi::CStr;
 use std::os::raw as libc;
 use std::ptr::{self, NonNull};
-use std::sync::{Once, ONCE_INIT};
+use std::sync::Once;
 
 use super::stmt::Statement;
 use super::url::ConnectionOptions;
@@ -195,7 +195,7 @@ impl Drop for RawConnection {
 /// > any other client library call.
 ///
 /// <https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html>
-static MYSQL_THREAD_UNSAFE_INIT: Once = ONCE_INIT;
+static MYSQL_THREAD_UNSAFE_INIT: Once = Once::new();
 
 fn perform_thread_unsafe_library_initialization() {
     MYSQL_THREAD_UNSAFE_INIT.call_once(|| {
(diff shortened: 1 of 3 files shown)


killercup:
Might conflict with #2108, sorry

diesel/src/mysql/connection/raw.rs
```diff
@@ -195,6 +196,7 @@ impl Drop for RawConnection {
 /// > any other client library call.
 ///
 /// <https://dev.mysql.com/doc/refman/5.7/en/mysql-init.html>
+#[allow(deprecated)] // min 1.38.0
```
JohnTitor: In my opinion, `Once::new()` has been stabilized in 1.2.0 so using it might not break compatibility...?

killercup: When did it become a `const fn`, though? Otherwise we can't use it in a static AFAIK

JohnTitor: Ah, I might wrong but I think we can in static from 1.24.0, not 1.2.0 [godbolt](https://rust.godbolt.org/z/0hHkDO).

JohnTitor:
`mem::initialized` will be deprecated in 1.39.0, not 1.38.0 by [this PR](https://github.com/rust-lang/rust/pull/62599) but it's reasonable to allow that before it is too late.

diesel/src/mysql/connection/raw.rs
killercup: Ah, very cool! Didn't know it's been available that long. Thanks for checking!

killercup:
Rebased on top of your branch, @JohnTitor, so it now includes the `Once::new()`. Let's see what CI says.

JohnTitor:
Hmm, an error unrelated to this PR has occurred in `macOS mysql beta`.

killercup:
Yeah, the tests even passed but

> Hosted Agent lost communication with the server

when running rustdoc.

I'll read this as "CI is green" :D

sgrif:
Given that many folks on the unsafe code guidelines feel that `mem::uninitialized` can *never* be used safely without UB, should we just switch to using `MaybeUninit` instead of allowing deprecated here? Or is it not stable on the current stable yet?

JohnTitor:
The problem with using `MaybeUninit` is that it has recently been stabilized. As mentioned above, we need to bump up the minimal version. If bumping up it isn't a big problem, it's good to replace `uninitialized()` with `MaybeUninit`.

sgrif:
Our policy is that we only bump if we have a good reason (e.g. not just for a convenience function). I think that meets this bar. We usually want to have some amount of lag so folks using their distro's Rust have a chance of being able to get the right version, but I think right now we're at least a few months from 2.0 shipping so it's fine.

JohnTitor:
Yeah, it seems we have a good reason to bump up the version.

weiznich:
I would vote for closing this PR.
#2131 fixes the CI by allowing all those warnings on nightly builds.
#2108 Fixes the warning regarding to `ONCE_INIT`
The deprecation of `mem::uninitialized` should be handled in a different PR

killercup:
👍

PR #2121, opened 2019-07-13
Add sqlite-bundled and sqlite-bundled-windows features

kpcyrd:
This allows passing through 2 features to libsqlite3-sys:

- `sqlite-bundled`: enables the bundled libsqlite source in libsqlite3-sys
- `sqlite-bundled-windows`: does the same, but is specifically restricted to windows and uses dynamic linking everywhere else

Note this is specific to `libsqlite3-sys 0.15.0`.


Eijebong:
This is the same as #829. You can achieve this by adding `libsqlite3` to your `Cargo.toml` with the features you'd like.

PR #2126, opened 2019-07-23
Added helpful error message when patch_file is used without patch

runfalk:
Currently `diesel print-schema` will print *not found* if `patch` is not installed and the `patch_file` argument is used. This bit me as I thought it could not find my patch file (which wasn't the case).

Hopefully this patch helps others in my position.

### DIFF
--- a/diesel_cli/src/print_schema.rs
+++ b/diesel_cli/src/print_schema.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Deserializer};
 use std::error::Error;
 use std::fmt::{self, Display, Formatter, Write};
 use std::fs::File;
-use std::io::{self, Write as IoWrite};
+use std::io::{self, Error as IoError, ErrorKind, Write as IoWrite};
 use std::path::Path;
 use std::process::Command;
 use tempfile::NamedTempFile;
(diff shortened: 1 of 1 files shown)


sgrif:
The change seems fine, but I don't think you should need to reformat all the code underneath it for such a small change.

runfalk:
There we go. I've made the change less invasive.

diesel_cli/src/print_schema.rs
```diff
@@ -51,6 +51,13 @@ pub fn run_print_schema<W: IoWrite>(
     Ok(())
 }
 
+fn simplify_patch_error(err: IoError) -> Box<dyn Error> {
+    match err.kind() {
+        ErrorKind::NotFound => "Unable to find 'patch' command, is it installed?".into(),
```
JohnTitor: IMO: I'm not sure but using backticks instead of single quotations would be good?

runfalk: I don't really have a strong preference for either. The only intention here is to make sure the user doesn't read it as the word *patch* instead of the command `patch`. I've updated the commit to use backticks.

weiznich:
Could you rebase this PR on top of #2131 to fix the CI?

runfalk:
It seems like there is still some issues with the CI. I've looked at both the failing Travis and Azure pipelines logs and it seems to be timeouts unrelated to my changes. I've retriggered the build.

runfalk:
The re-run sems to have fixed Azure, but Travis is still not playing nice. If I look at the failed logs they don't event get to setting up Rust.

weiznich:
Seems like travis has some issues in the last time regarding to downloading stuff. Hopefully the will fix that soon.
(Anyway no need to wait for this.)

PR #2130, opened 2019-07-31
Bump pq-sys to 0.4.1

xfix:
This avoids errors when building with `-Z minimal-versions`


weiznich:
Could you elaborate on which errors exactly occur while building and how those are fixed with downgrading libpq-sy with `-Z minimal-versions`?

xfix:
```
error[E0642]: patterns aren't allowed in methods without bodies
   --> /home/xfix/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-serialize-0.3.19/src/serialize.rs:147:45
    |
147 |                                             &f_name: &str,
    |                                             ^^^^^^^
```

This is because `pq-sys` 0.3 uses `bindgen 0.22.0` which itself depends on `rustc-serialize 0.3.19`.

PR #2131, opened 2019-07-31
Allow nightly builds to fail

weiznich:
If a nightly build fails a warning instead of an error is shown in the
build console


weiznich:
Hopefully this makes #2117 obsolete.

weiznich:
Would be great if azure pipelines would report that there were warnings. They are shown in the [build status over there](https://dev.azure.com/diesel-rs/diesel/_build/results?buildId=435&view=results), but I'm not sure why they are not reported to github

JohnTitor:
Ugh, `ONCE_INIT` is deprecated in `nightly-2019-08-01` so compiletest is failed again...

JohnTitor:
This PR seems good to me so we can merge this now then we can go to https://github.com/diesel-rs/diesel/pull/2108, or this can also include its changes.

JohnTitor:
Can we go? Or should we wait for another review?

weiznich:
I would like to hear at least another opinion about the use of `--cap-lints` here.

Eijebong:
So IIUC, the cap-lints part is to make depreciation warning stay as warnings even with a `deny(deprecated)` ?

weiznich:
> So IIUC, the cap-lints part is to make depreciation warning stay as warnings even with a deny(deprecated) ?

Yes, but only for the nightly compiler. It is basically meant as a poor mans `allow_failures: nightly`. Ideally azure would indicate that some job generates warnings, but I've not found any way to do this yet.

Eijebong:
That would not work on the event of something erroring for real then ? That's kind of a bummer because a dependency might be broken on nightly for some reasons and we don't want that to make the build red in that case either

weiznich:
Right, if we want to prevent that we would need to set the [this return](https://github.com/diesel-rs/diesel/pull/2131/files#diff-f125f8b9ff87d6369d6c4296253e8813R17) here to true. I'm not sure if we really want to do that, because ~all of our breakage from the last time was basically rustc introducing some new lint/deprecation/… in nightly.

Eijebong:
Yeah but I've seen stuff erroring due to errors in dependencies in the past on other projects. I guess this is not common enough to worry about especially since it's fixed quickly most of the time.

I wish rustc would add a cfg for nightly compilers so we could do `#[cfg(not(nightly), deny(warnings))]` or something...

I guess this is fine then :+1:

PR #2134, opened 2019-08-03
Separate raw value representations of postgresql and mysql backend

weiznich:
This commit separates the raw value representation of the mysql and
postgresql backend from raw `&[u8]` to specific types that could hold
additional information. To avoid code duplication for commonly used
types an additional trait (`BinaryRawValue`) is introduced, which is
basically equivalent to `AsRef<[u8]>` but has the existing
`HasRawValue` trait as super trait. This is required to workaround
some higher ranked life time issues in rustc.
Additionally some type information are added to the postgresql value
type. Those information may allow to do dynamic type conversations in
future diesel versions.

Basically a improved reimplementation of #1833 on top of #1976. As @sgrif stated there this PR keeps the common impl's in place by adding a `BinaryRawValue` trait.

That said: I'm not super sure if the newly added public types/functions have meaningful names or if we should choose something different.

cc @pgab

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,12 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * The minimal officially supported rustc version is now 1.36.0
 
+* The `RawValue` types for the `Mysql` and `Postgresql` backend where changed
+  from `[u8]` to distinct opaque types. If you used the concrete `RawValue` type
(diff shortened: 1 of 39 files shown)


weiznich:
All remaining build failures are fixed by #2131

weiznich:
@sgrif Stated that he has some comments on this.

weiznich:
@sgrif I've done the changes we have discussed some weeks ago.

weiznich:
@JohnTitor Would it possible to get a final review on this (assuming that the CI now succeeds)?
Seans comments are now addressed.

PR #2135, opened 2019-08-04
Add OpenSSL req + note on libpq in PATH

khionu:
Hopefully this prevents some confusion when trying to use the CLI

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -7,9 +7,11 @@ bi-directional changes to your database that get applied sequentially.
 Installation
 ---------------
 
-The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
+The diesel cli, by default, requires [`openssl`][openssl] [`libpq`][postgres], [`sqlite`][sqlite]
 and [`mysql`][mysql]. Once these dependencies are installed, you can run
 `cargo install diesel_cli`.
+
+> Note: Make sure that both the `bin` and `lib` directories for postgres are added to your PATH
 
 To install the cli without these dependencies, omit the unneeded dependencies from
 the following command:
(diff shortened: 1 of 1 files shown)


diesel_cli/README.md
```diff
@@ -7,17 +7,20 @@ bi-directional changes to your database that get applied sequentially.
 Installation
 ---------------
 
-The diesel cli, by default, requires [`libpq`][postgres], [`sqlite`][sqlite]
+The diesel cli, by default, requires [`openssl`][openssl] [`libpq`][postgres], [`sqlite`][sqlite]
```
JohnTitor: You should add a comma after `openssl`. And for readability, it may be good that the lines aren't long like this.

```
The diesel cli, by default, requires [`openssl`][openssl], [`libpq`][postgres],
[`sqlite`][sqlite] and [`mysql`][mysql]. Once these dependencies are installed,
you can run `cargo install diesel_cli`.
```

diesel_cli/README.md
```diff
@@ -7,17 +7,20 @@ bi-directional changes to your database that get applied sequentially.
...
+> Note: Make sure that both the `bin` and `lib` directories for postgres are added to your PATH
+
 To install the cli without these dependencies, omit the unneeded dependencies from
 the following command:
 
 ```sh
 cargo install diesel_cli --no-default-features --features "postgres sqlite mysql"
 ```
 
+[openssl]: https://indy.fulgan.com/SSL/
```
JohnTitor: Why isn't it `https://www.openssl.org/source/`?

khionu: I grabbed it from the Wiki. Also, noticing now that it's Windows binaries only. Will change to that URL

diesel_cli/README.md
khionu: Adding comma, but not sure how I would break it up? Also, when rendered, it's significantly shorter.

JohnTitor: In the editor, long lines aren't good to read. At a quick glance, many other lines are separated to no more than 80 characters. You can shorten the length of a line by separating newlines instead of a space where appropriate.

khionu: Fixed, I forgot Markdown doesn't respect a single newline, which I can now appreciate!

PR #2089, opened 2019-06-10
Adjust clippy config to reflect 'cyclomatic' -> 'cognitive' rename

rye:
In only one case is `allow` called with this lint, and there is also a reference in the `clippy.toml` configuration file. This (teensy) PR just updates the lint name to reflect the changes that have been made in the naming of this lint.

PR #2137, opened 2019-08-06
Fix/remove mem uninitialized

### LINKED ISSUE #2136: Remove cyclomatic-complexity-threshold from clippy.toml
### Versions

- **Rust:** 1.36
- **Diesel:**  1.4.2 patched to github master
- **Database:** sqlite3
- **Operating System** Linux 4.15

### Feature Flags

- **diesel:** sqlite

## Problem Description
When running `cargo clippy` on my project that depends on diesel the following error is reported, and clippy build aborts:
```
    Checking diesel v1.4.2 (https://github.com/diesel-rs/diesel#0b6c59e3)
error: error reading Clippy's configuration file `/home/rui/.cargo/git/checkouts/diesel-6e3331fb3b9331ec/0b6c59e/clippy.toml`: found deprecated field `cyclomatic-complexity-threshold`. Please use `cognitive-complexity-threshold` instead.
```

Run clippy

`Finished dev [unoptimized + debuginfo] target(s) in 20.88s`

See above

No

### Steps to reproduce
Run `cargo clippy`

## Checklist

I replaced `cyclomatic-complexity-threshold` with `cognitive-complexity-threshold` in clippy.toml, and `cargo clippy` now builds.

weiznich:
Follow up to #2117 

Closes #2089
Closes #2136


diesel/src/serialize.rs
```diff
@@ -76,7 +76,7 @@ impl<DB: TypeMetadata> Output<'static, Vec<u8>, DB> {
     /// Unsafe to use for testing types which perform dynamic metadata lookup.
     pub fn test() -> Self {
         use std::mem;
-        #[allow(clippy::invalid_ref)]
+        #[allow(clippy::invalid_ref, deprecated)]
```
JohnTitor: Why do we allow this?

weiznich: The alternative for the line below would be `Self::new(Vec::new(), unsafe { mem::MaybeUninit::uninit().assume_init() })` which would basically be also UB.
Therefore I think it is not worth changing it. 
(cc @diesel-rs/core just in case)

JohnTitor: Ah okay, I got it. Could we add a comment about it?

weiznich: A comment for what exactly? There is already a comment describing what this function tries to achieve.

JohnTitor: I mean it's reasonable to leave a comment on why we use the deprecated method, what about?

PR #1824, opened 2018-08-22
Align getting started example imports with repo

tyre:
While the prelude imports `PgConnection`, it's a confusing mis-match for beginners.

Fixes #1758

PR #2143, opened 2019-08-14
[rebase] Align getting started example imports with repo

Mcat12:
Rebased #1824 onto master so it can pass the CI and get merged.

Original author: @tyre


JohnTitor:
But CI is red due to `mem::uninitialized` again so we should remove it for maintenance even if it is meaningless? @weiznich

JohnTitor:
CI should be green now, could you rebase again?

PR #2144, opened 2019-08-14
Allow to customize r2d2 `is_valid` checks per backends

weiznich:
The current query is not portable for all possible backends. Oracle needs
`SELECT 1 FROM DUAL` whereas all other backends are using `SELECT
1`. This change indroduces a trait implemented by the connection
implementation that provides the correct query string as associated
constant.

Technically that's a breaking change, for third party backends at least. The only affected crate that I'm aware of that could be affected by this is `diesel-oci`, but I'm proposing the change from there. So I'm not sure if this needs a changelog entry?

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -75,9 +75,36 @@ impl ::std::error::Error for Error {
     }
 }
 
+/// A trait indicating a connection could be used inside a r2d2 pool
+pub trait R2D2Connection: Connection {
+    /// Check if a connection is still valid
+    fn ping(&self) -> QueryResult<()>;
+}
+
+#[cfg(feature = "postgres")]
+impl R2D2Connection for ::pg::PgConnection {
+    fn ping(&self) -> QueryResult<()> {
+        self.execute("SELECT 1").map(|_| ())
+    }
+}
+
+#[cfg(feature = "mysql")]
+impl R2D2Connection for ::mysql::MysqlConnection {
+    fn ping(&self) -> QueryResult<()> {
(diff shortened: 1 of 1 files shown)


sgrif:
I am extremely mad that this isn't portable

diesel/src/r2d2.rs
```diff
@@ -75,9 +75,31 @@ impl ::std::error::Error for Error {
     }
 }
 
+/// A trait indicating a connection could be used inside a r2d2 pool
+pub trait R2D2Connection: Connection {
+    /// Query used to check if a connection is still alive
+    ///
+    /// Should result in a database error if that's not the case
+    const CHECK_QUERY_STRING: &'static str;
```
sgrif: If `SELECT 1` isn't broad enough, should we just drop all assumptions and have this be `fn ping(&self) -> QueryResult<()>`?

weiznich: That would probably be the even better solution because this would theoretically allow to use sql independent connection checks provided be the libraries below.

diesel/src/r2d2.rs
```diff
@@ -75,9 +75,35 @@ impl ::std::error::Error for Error {
...
+        self.execute("SELECT 1").map(|_| ())
+    }
+}
+
+#[cfg(feature = "mysql")]
+impl R2D2Connection for ::mysql::MysqlConnection {
+    fn ping(&self) -> QueryResult<()> {
+        self.execute("SELECT 1").map(|_| ())
+    }
+}
```
sgrif: Newline needed

PR #2146, opened 2019-08-16
Remove `mem::uninitialized` from testcase to fix ci

weiznich:
(Yes we know that's UB, but that was ok there …)

### DIFF
--- a/diesel/src/serialize.rs
+++ b/diesel/src/serialize.rs
@@ -38,7 +38,7 @@ where
     DB::MetadataLookup: 'a,
 {
     out: T,
-    metadata_lookup: &'a DB::MetadataLookup,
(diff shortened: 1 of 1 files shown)


diesel/src/serialize.rs
```diff
@@ -38,15 +38,15 @@ where
     DB::MetadataLookup: 'a,
 {
     out: T,
-    metadata_lookup: &'a DB::MetadataLookup,
+    metadata_lookup: Option<&'a DB::MetadataLookup>,
```
Eijebong: This is a breaking change, do we want that ?

Eijebong: No it's not, I'm going back to bed...

PR #2147, opened 2019-08-16
Add a special warning when multiple parents are used in one belongs_to

### LINKED ISSUE #2091: #[belongs_to()] silently ignores additional parameters
### Versions

- **Rust:** rustc 1.37.0-nightly (400b409ef 2019-06-09)
- **Diesel:** 1.4.2

## Problem Description

When a model has multiple associations, the developer might intuitively do the following:
- belongs_to(Parent1, Parent2)

...whereas they are supposed to do:

- belongs_to(Parent1)
- belongs_to(Parent2)

When passing multiple parameters to belongs_to the compiler will silently ignore the second association and the code will compile. The problem is not exposed until the developer tries to call belonging_to(Parent2).

Create a model with two parent associations so I can call belonging_to() for either of those parents.

The compiler should warn about the incorrect syntax for belongs_to() and advise on the proper syntax. From a usability perspective, allowing multiple associations to be passed to belongs_to() would be an intuitive solution.

Nothing. The code will compile silently without error (unless belonging_to(Parent2) is called.

## Checklist

### Thanks

A massive thanks to the maintainers and contributors to Diesel. This is a really great ORM.

Mcat12:
There was already a warning being emitted (unrecognized option), but this new warning is more user-friendly.

Closes #2091

### DIFF
--- a/diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
+++ b/diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
@@ -3,10 +3,10 @@ warning: belongs_to takes a single parent. Change
 to
     belongs_to(Bar)
     belongs_to(Baz)
-  --> $DIR/belongs_to_second_parent.rs:23:19
+  --> $DIR/belongs_to_second_parent.rs:23:3
    |
 23 | #[belongs_to(Bar, Baz)]
-   |                   ^^^
+   |   ^^^^^^^^^^^^^^^^^^^^
 
 error[E0433]: failed to resolve: use of undeclared type or module `foos`
   --> $DIR/belongs_to_second_parent.rs:24:8
--- a/diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
+++ b/diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
@@ -1,0 +1,19 @@
+warning: belongs_to takes a single parent. Change
+    belongs_to(Parent1, Parent2)
+to
+    belongs_to(Parent1)
+    belongs_to(Parent2)
+  --> $DIR/belongs_to_second_parent.rs:23:19
+   |
+23 | #[belongs_to(Bar, Baz)]
+   |                   ^^^
+
+error[E0433]: failed to resolve: use of undeclared type or module `foos`
+  --> $DIR/belongs_to_second_parent.rs:24:8
+   |
+24 | struct Foo {
+   |        ^^^ use of undeclared type or module `foos`
+
+error: aborting due to previous error
+
+For more information about this error, try `rustc --explain E0433`.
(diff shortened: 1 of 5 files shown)


sgrif:
This needs a compile-test

Mcat12:
@sgrif is this still needed?

diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
```diff
@@ -0,0 +1,19 @@
...
+    belongs_to(Parent1, Parent2)
+to
+    belongs_to(Parent1)
+    belongs_to(Parent2)
+  --> $DIR/belongs_to_second_parent.rs:23:19
+   |
+23 | #[belongs_to(Bar, Baz)]
+   |                   ^^^
+
+error[E0433]: failed to resolve: use of undeclared type or module `foos`
```
weiznich: This error message does not belong here. Seems like the table should be named `foos` not `foo`.

diesel_derives/src/associations.rs
```diff
@@ -147,13 +147,29 @@ impl AssociationOptions {
...
-                    path_to_string(&ignored.name())
-                ))
-                .emit();
+            match ignored.path() {
+                Ok(_) => {
+                    ignored
+                        .span()
+                        .warning(
+                            "belongs_to takes a single parent. Change\n\
+                            \tbelongs_to(Parent1, Parent2)\n\
```
weiznich: Any possibility that this error message contains the actual parameters passed to the attribute, so that for `#[belongs_to(Bar, Baz)]` the following warning is emitted:

```
warning: belongs_to takes a single parent. Change
    belongs_to(Bar, Baz)
to
    belongs_to(Bar)
    belongs_to(Baz)
````

Mcat12:
@weiznich Is this special warning necessary? I asked about this on the issue but did not receive a response:
> Based on the issue, it seems that the unrecognized option warning wasn't sufficient. If this is no longer the case, we can close the issue and PR.

https://github.com/diesel-rs/diesel/issues/2091#issuecomment-522589979

Mcat12:
@weiznich Checking one last time before I close this.

weiznich:
I think better error/warning messages are not bad, so I'm in favour of merging this as soon as the points mentioned are fixed.
- [ ] UI compile-test
- [ ] My review comments

diesel_compile_tests/tests/ui/belongs_to_second_parent.stderr
Mcat12: This was taken from the existing `belongs_to_invalid_option_syntax` test. I've fixed it there as well.

Mcat12:
@weiznich I think I've addressed all of the comments. Thanks for reviewing!

PR #2152, opened 2019-08-25
Reduce cache on Travis

JohnTitor:
Fixes that Travis jobs are terminated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,10 @@ rust:
   - stable
   - beta
   - nightly
-cache: cargo
+cache:
+  directories:
+    - $HOME/.cargo
+    - $HOME/.rustup
 addons:
   postgresql: '9.5'
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
   directories:
     - $HOME/.cargo
     - $HOME/.rustup
+before_cache:
+  - rm -rf /home/travis/.cargo/registry
 addons:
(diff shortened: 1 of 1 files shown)


JohnTitor:
It seems fine, what about? @weiznich

JohnTitor:
Attempt to reduce more following https://levans.fr/rust_travis_cache.html

killercup:
This might also be interesting: https://github.com/rust-analyzer/rust-analyzer/blob/04c2961d0c5493962e948dc8101445cc76f1d460/.travis.yml#L2-L5

JohnTitor:
@killercup Good point! It seems reasonable. If we don't drop Travis right away, we can go for reducing cache for now.

.travis.yml
```diff
@@ -4,7 +4,15 @@ rust:
...
   - beta
   - nightly
-cache: cargo
+cache:
+  directories:
+    - $HOME/.cargo
+    - $HOME/.rustup
+before_cache:
+  - rm -rf /home/travis/.cargo/registry
+  - find ./target/debug -maxdepth 1 -type f -delete
```
weiznich: We don't cache `target` therefore we don't need to delete there something right?

JohnTitor: Ah, indeed, so we don't need a2d36107e770b07fbf164174727535a00f113871?

weiznich: I think we don't need it.

JohnTitor: Okay, dropped the commit.

PR #2111, opened 2019-07-07
Fix incorrect error message

derekdreery:
Fixes #1977

PR #2159, opened 2019-09-02
Fix formatting error with embedded migrations

### LINKED ISSUE #1977: formatter error

geomaniac:
In `Display` impl for`MigrationName`, use the migration version if `Migration::file_path` returns `None`.

This is based on the work from @sorccu in #1993.

Fixes #1977.

### DIFF
--- a/diesel_migrations/migrations_internals/src/migration.rs
+++ b/diesel_migrations/migrations_internals/src/migration.rs
@@ -45,12 +45,12 @@ pub fn file_name<'a>(migration: &'a dyn Migration, sql_file: &'a str) -> Migrati
 
 impl<'a> fmt::Display for MigrationFileName<'a> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let fpath = match self.migration.file_path() {
-            None => return Err(fmt::Error),
-            Some(v) => v.join(self.sql_file),
-        };
-        f.write_str(fpath.to_str().unwrap_or("Invalid utf8 in filename"))?;
-        Ok(())
+        if let Some(path) = self.migration.file_path() {
+            let fpath = path.join(self.sql_file);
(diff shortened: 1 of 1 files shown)


JohnTitor:
Could you rebase it to prevent travis from being terminated?

geomaniac:
> Could you rebase it to prevent travis from being terminated?

Ah your PR just got merged. Sure, give me a sec.

geomaniac:
Not quite sure what to do about the 1 failing test, seems to be something SSL related:

> Caused by:
>
>A  [35] SSL connect error (error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure)
>
>The command "rustc +$TRAVIS_RUST_VERSION --version &&[..]" exited with 101.

Btw, I just noticed there is another open PR #2111, that goes in a somewhat similar direction. I am not too attached to my changes, so if you rather want to go with #2111, I can close this PR.

lutter:
I am glad to see this fixed. When will this appear in a release?

weiznich:
@lutter We do not give any ETA on release dates other than "when it is done".

PR #1895, opened 2018-10-25
Strip scheme from sqlite database URL

AndrewSpeed:
This PR strips the `sqlite://` scheme from sqlite database URLs, as mentioned in https://github.com/diesel-rs/diesel/issues/1124#issuecomment-333334217

I wasn't able to locate anywhere that tested the functionality I was editing, please let me know if I just missed it and I'll be happy to update this PR.

Issue #1124, opened 2017-08-20
Raise more useful error if mismatch between feature enabled and db url


katrinabrock:
@sgrif 
## Problem Description

If proper features are not enabled, error messages to not point to this being the issue.

### What are you trying to accomplish?
Started project with one backend, switched to a different backend and forgot to change `Cargo.toml`.

### What is the expected output?
If `DATABASE_URL` is for a database where feature is not enabled, there should at least be a warning about this. 

### What is the actual output?
`
help: message: Could not load table names from database `postgresql://sammy:sam
my@localhost/diesel_demo`: StringError("Failed to establish a database connection a
t postgresql://sammy:sammy@localhost/diesel_demo. Error: BadConnection(\"Unable to 
open the database file\")")
`
### Steps to reproduce
Cargo.toml 
`
[dependencies]
diesel = { version = "0.15.0", features = ["sqlite"] }
diesel_codegen = { version = "0.15.0", features = ["sqlite"] }
dotenv = "0.9.0"
`
.env
`
DATABASE_URL=postgres://username:password@localhost/diesel_demo > .env
`

## Checklist

- [ ] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

sgrif:
To clarify, specifically I think we should be checking for the URL starting with `mysql://`, `postgres://` or `postgresql://` when those features are not enabled. I think we can safely assume that the user is not trying to connect to a SQLite database in that case.

sgrif:
Related: We should strip `sqlite://` from the URL if it's present. I've had a couple people try that expecting it to work.

bbqsrc:
This just broke my mind for a while, as I changed computers and couldn't understand the error that diesel-cli was giving me. Trying to connect to a postgres db with only the sqlite feature enabled provides the following:

```
$ diesel migration run
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ConnectionError(BadConnection("Unable to open the database file"))', libcore/result.rs:945:5
```
So yeah, this is still less than ideal. :)

christiansakai:
I literally just encountered this problem and went on for 2 days.... Glad I found this issue.

AndrewSpeed:
I'll take a look at implementing this

apiraino:
I literally just encountered this problem and went on for ~2 hours ... Glad I found this issue.

porglezomp:
[...]

PR #2176, opened 2019-09-30
Strip scheme from sqlite database urls

porglezomp:
This PR rebases #1895 and adds a test case as requested there.

> This PR strips the `sqlite://` scheme from sqlite database URLs, as mentioned in [#1124 (comment)](https://github.com/diesel-rs/diesel/issues/1124#issuecomment-333334217)
> 
> I wasn't able to locate anywhere that tested the functionality I was editing, please let me know if I just missed it and I'll be happy to update this PR.


JohnTitor:
Could you run rustfmt to pass CI?

Issue #2163, opened 2019-09-11
Feature : configure the migration dir in diesel.toml


theredfish:
Hello,

Could we use the `diesel.toml` file to configure the migrations directory path? It could be easier than using the `--migration-dir` argument. It seems to be the right place for this kind of configuration. 

If you agree I can provide a pull request for this feature.

sgrif:
Sounds reasonable to me

theredfish:
Hello,
I'm trying to run tests, but I'm getting some errors about duplicate code. It seems that macros for features are ignored. For example : 

```rust
#[cfg(feature = "sqlite")]
const BACKEND: &str = "sqlite";
#[cfg(feature = "postgres")]
const BACKEND: &str = "postgres";
#[cfg(feature = "mysql")]
const BACKEND: &str = "mysql";
```

```bash
    |
119 | const BACKEND: &str = "sqlite";
    | ------------------------------- previous definition of the value `BACKEND` here
120 | #[cfg(feature = "postgres")]
121 | const BACKEND: &str = "postgres";
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `BACKEND` redefined here
    |
    = note: `BACKEND` must be defined only once in the value namespace of this module
```

This is what the compiler returns for the command `cargo test` in `diesel_cli`. I didn't found any specific information in the `CONTRIBUTING.md` about the way to run the tests. Could you help me with this @sgrif ?

sgrif:
The tests for the CLI need to be run for a specific backend. Add `--no-default-features --features some_backend` or look at `bin/test`

theredfish:
@sgrif  I tried to put the logic inside the diesel_cli crate, but I'm not sure if it's the right place. In `diesel_cli/main.rs` the function `migrations_dir` calls `migrations::find_migrations_directory`. 

Should I manage to read the `diesel.toml` file in the `diesel_migrations` crate? It's where I found the function `find_migrations_directory` (so in my case I need to add the feature to load the path from the `diesel.toml` file). 

Thanks

ralpha:
It seems like this is (still) not merges with [diesel_cli](https://crates.io/crates/diesel_cli) 
As I get:
```
unknown field `migrations_directory`, expected `print_schema`
```
And current version is: 1.4.0 From Jan 21, 2019. This issue was created after that date.

Just saying that this would be a nice feature to have.

weiznich:
[...]

PR #2179, opened 2019-10-01
Configure the migration directory in diesel.toml

theredfish:
This pull request is related to #2163 .

By default the diesel.toml is configured with the migrations directory :
```
[print_schema]
file = "src/schema.rs"

[migrations_directory]
file = "migrations"
```

This configuration allow users to change the default migrations directory. The migration directory is read with the following priority :

1. From the CLI argument
2. From the environment variable
3. From the configuration file `diesel.toml`
4. From the default directory : migrations

If the following pull request is accepted i can open a new one to update the [documentation](https://github.com/sgrif/diesel.rs-website) concerning the way to configure the migration directory from `diesel.toml`.


diesel_cli/src/main.rs
```diff
@@ -194,12 +195,27 @@ fn migrations_dir_from_cli(matches: &ArgMatches) -> Option<PathBuf> {
...
     migrations_dir_from_cli(matches)
         .or_else(|| env::var("MIGRATION_DIRECTORY").map(PathBuf::from).ok())
+        .or_else(|| migrations_dir_from_config(matches).ok())
         .unwrap_or_else(|| migrations::find_migrations_directory().unwrap_or_else(handle_error))
 }
 
+/// Looks for a migration directory configured in `diesel.toml`.
+/// Returns a `MigrationError::MigrationDirectoryNotFound` if
+/// no path to the migration directory is found.
+fn migrations_dir_from_config(matches: &ArgMatches) -> Result<PathBuf, MigrationError> {
```
AlterionX: Possibly bump this function above `migrations_dir`, same as `migrations_dir_from_cli`.

diesel_cli/src/config.rs
```diff
@@ -64,3 +66,15 @@ impl PrintSchema {
...
+#[derive(Default, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct MigrationsDirectory {
+    pub file: PathBuf,
+}
+
+impl MigrationsDirectory {
+    pub fn file(&self) -> &PathBuf {
+        &self.file
+    }
```
AlterionX: The field is already exposed as `pub`, so I'm not sure if this is necessary.

diesel_cli/src/main.rs
```diff
@@ -210,9 +226,15 @@ fn create_migrations_dir(matches: &ArgMatches) -> DatabaseResult<PathBuf> {
         .map(PathBuf::from)
         .or_else(|| env::var("MIGRATION_DIRECTORY").map(PathBuf::from).ok())
         .unwrap_or_else(|| {
-            find_project_root()
-                .unwrap_or_else(handle_error)
-                .join("migrations")
+            let migrations_dir_from_config = migrations_dir_from_config(matches);
```
AlterionX: Possibly lift out the common component between `migrations_dir` and this function instead of adding to this function? But not necessary.

diesel_cli/src/config.rs
```diff
@@ -64,3 +66,15 @@ impl PrintSchema {
...
+}
```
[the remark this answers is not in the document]
theredfish: Thanks for the review. I wonder what's better between a direct access to the field or through a function?

diesel_cli/src/main.rs
theredfish: What do you mean by the common component?

AlterionX:
```rust
fn migrations_dir(matches: &ArgMatches) -> PathBuf {
    migrations_dir_from_cli(matches)
        .or_else(|| env::var("MIGRATION_DIRECTORY").map(PathBuf::from).ok())
        .or_else(|| migrations_dir_from_config(matches).ok())
        .unwrap_or_else(|| migrations::find_migrations_directory().unwrap_or_else(handle_error))
}

// -- snip

fn create_migrations_dir(matches: &ArgMatches) -> DatabaseResult<PathBuf> {
    // This statment seems to be (almost) a c+p of the `migrations_dir` method.
    // That is it looks like it could be `let dir = matches.migrations_dir()` instead.
   // If this isn't the case, then can we refactor a little so that the `migration_dir` can
   // be directly called instead?
    let dir = matches
        .value_of("MIGRATION_DIRECTORY")
        .map(PathBuf::from)
        .or_else(|| env::var("MIGRATION_DIRECTORY").map(PathBuf::from).ok())
        .unwrap_or_else(|| {
            let migrations_dir_from_config = migrations_dir_from_config(matches);
            let project_root = find_project_root().unwrap_or_else(handle_error);

            // Retreives and transforms the path of the migrations
            // directory from the configuration file or returns the
            // default one.
            migrations_dir_from_config
                .ok()
                .unwrap_or_else(|| project_root.join("migrations"))
        });

    // -- snip
}
```

AlterionX: I wrote it in a comment in the code block, but:

The statement in `create_migrations_dir` looks really similar to the `migrations_dir` function, so can we call `migrations_dir` instead of typing it out again?

If this isn't the case, then can we refactor a little so that this can happen?

diesel_cli/src/config.rs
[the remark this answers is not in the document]
AlterionX: Depends on what you want to achieve, namely:

A field access allows a super set of the accessor function (i.e. you're not really trying to maintain some state). For example, if you pass an `&mut` to the struct, you can replace the file field with an alternative. If you pass an `&` to the struct, it will be almost the same as an accessor method.

A function allows you to either hide some of the complexity or return a conglomeration/derived field as well. This doesn't seem necessary here. An analogous case is present in the struct abobe. If we look there, I believe file is treated as a field and there is no accessor function for the field. This is because it is solely meant to hold data and it's not pertinent to hide the details behind the field.

Now that I think about it again, can we call the field in the struct `dir` instead of `file` as well? Since we're no longer looking for a single file.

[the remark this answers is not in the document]
theredfish: Thank you for the explanantions. I made the changes,

diesel_cli/src/main.rs
theredfish: Ok thank you for the details. I will see what I can do soon.

theredfish:
Hello @AlterionX !
I'm at the Rustfest and it's impl days! So i'm taking the time to finish my PRs. I would like to know if I should rebase from master once I push my changes? Thank you!

weiznich:
@theredfish I'm also here at the impl days at room 218. Maybe just come around :wink:

theredfish:
Thank you @weiznich and @AlterionX  :)

Issue #2148, opened 2019-08-16
Cargo documentation generation fails


ivegotasthma:
## Setup
NixOS 19.03

```
[nix-shell:~/work]$ rustup --version
rustup 1.16.0 ( )

[nix-shell:~/work]$ rustc --version
rustc 1.38.0-nightly (c43d03a19 2019-08-14)

[nix-shell:~/work]$ cargo --version
cargo 1.38.0-nightly (e853aa976 2019-08-09)
```

### Versions

- **Diesel:** 1.4.2

### Feature Flags

- **diesel:** sqlite, chrono

## Problem Description
Documentation generation fails. Just running `cargo doc`

### What are you trying to accomplish?
Local documentation

### What is the expected output?
Bunch of red text

### What is the actual output?
```
[nix-shell:~/work/work]$ cargo doc -p diesel
warning: output filename collision.
The lib target `lock_api` in package `lock_api v0.3.1` has the same output filename as the lib target `lock_api` in package `lock_api v0.2.0`.
Colliding filename is: /home/asthma/work/work/target/doc/lock_api/index.html
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `parking_lot` in package `parking_lot v0.9.0` has the same output filename as the lib target `parking_lot` in package `parking_lot v0.8.0`.
Colliding filename is: /home/asthma/work/work/target/doc/parking_lot/index.html
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `parking_lot_core` in package `parking_lot_core v0.6.2` has the same output filename as the lib target `parking_lot_core` in package `parking_lot_core v0.5.0`.
Colliding filename is: /home/asthma/work/work/target/doc/parking_lot_core/index.html
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `rand_core` in package `rand_core v0.4.2` has the same output filename as the lib target `rand_core` in package `rand_core v0.3.1`.
Colliding filename is: /home/asthma/work/work/target/doc/rand_core/index.html
The targets should have unique names.
[...]

PR #2175, opened 2019-09-29
Fix a rustdoc issue with nightly

derekdreery:
This cherry-picks the fix for this issue from master

PR #2164, opened 2019-09-17
Update syn related dependencies

JohnTitor:
Follow up of #2145

PR #2102, opened 2019-06-28
bump libsqlite3-sys up to 0.15

arve0:
issue #2100

PR #2181, opened 2019-10-04
Backport a set of fixes to 1.4.x

helaan:
As requested in #2148, backport some commits to fix the documentation error. As requested in that commit, the SQLite bump is also included, as are a set of commits to make Diesel compile without warning on 1.38.0


weiznich:
I see now that libsqlite-sys 0.16 is released. Maybe we should bump that crate again to 0.16 before the release. Can you do add that to this PR?

helaan:
I've bumped libsqlite3-sys in #2183, will backport that when that one is merged.

weiznich:
I've just realized that this PR is not #2175 so I've just drop the list from there here :see_no_evil: 

#2019
#2056
#2108
#2102
#2164

I think #2019, #2056 and #2164 are missing. It would be great to also get them.

helaan:
I've incorporated #2019 and #2056, however #2164 requires additional work, which I'm not sure are worth it at this point, as Diesel doesn't expose types from syn and users of syn-1.0 libraries can just pull in another version of syn.

weiznich:
I've added #2164 to minimize duplicated syn versions (and therefore build time) in complexer projects.

PR #2188, opened 2019-10-08
Json support for MySQL

jiaminglu:
Add Json support for MySQL.

Because there is a name conflict with the pg's Json type, I moved it into the common `sql_types` module, while kept the `Jsonb` type in pg.

### DIFF
--- a/diesel/src/mysql/types/json.rs
+++ b/diesel/src/mysql/types/json.rs
@@ -1,0 +1,20 @@
+use deserialize::{self, FromSql};
+use mysql::{Mysql, MysqlValue};
+use serialize::{self, IsNull, Output, ToSql};
+use sql_types::*;
+use std::io::Write;
+
+impl FromSql<Json, Mysql> for serde_json::Value {
+    fn from_sql(value: Option<MysqlValue<'_>>) -> deserialize::Result<Self> {
+        let value = not_none!(value);
+        serde_json::from_slice(value.as_bytes()).map_err(Into::into)
+    }
+}
+
+impl ToSql<Json, Mysql> for serde_json::Value {
+    fn to_sql<W: Write>(&self, out: &mut Output<W, Mysql>) -> serialize::Result {
+        serde_json::to_writer(out, self)
+            .map(|_| IsNull::No)
+            .map_err(Into::into)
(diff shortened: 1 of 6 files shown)


weiznich:
In general your policy for the implementation of `FromSql` / `ToSql` for certain types in diesel is something like the three following questions:

 1.   Is this a commonly used type. (Yes because there is already support for `Jsonb `in diesel.
 2.   Could we provide a `FromSql` and a `ToSql` implementation for that type. (In this case Yes)
 3.   Could the conversion go somehow wrong because the sql type or the rust type could hold values that are not representable in the other type. (In this case the conversion from `String` to `Json` could go wrong, because obviously not every arbitrary string is a json object)

Because of the third point is not fulfilled we won't add support for this mapping. 
(Beside of that the same code will work fine in a third party crate, so no real need to have it in diesel at all costs)

jiaminglu:
@weiznich MySQL supports JSON types. In current versions, if the DB contains any JSON field, there will be `Json` types in the `schema.rs`, but the type is missing fo MySQL, so it won't compile.

jiaminglu:
@weiznich You can verify this by write a migration with the following table:
```sql
create table test (
    id integer not null primary key auto_increment,
    value json not null
);
```

The error message:
```rust
error[E0412]: cannot find type `Json` in this scope
  --> src/schema.rs:25:18
   |
25 |         value -> Json,
   |                  ^^^^ not found in this scope
help: there is an enum variant `config::FileFormat::Json`; try using the variant's enum
   |
25 |         value -> config::FileFormat,
   |                  ^^^^^^^^^^^^^^^^^^
```

weiznich:
Mysql supports internally [this types](https://dev.mysql.com/doc/refman/8.0/en/c-api-prepared-statement-type-codes.html). Everything else cannot be really enforced.

jiaminglu:
@weiznich So how to resolve the above error? I don't want to modify the `schema.rs` code every time I run a migration.

And according to [this page](https://dev.mysql.com/doc/refman/5.7/en/json.html), there won't be invalid json strings.
> Automatic validation of JSON documents stored in JSON columns. Invalid documents produce an error.

PR #2194, opened 2019-10-10
Make some fields of insertable public

weiznich:
This allows custom backend implementation like diesel-oci to write
their own `QueryFragment` implementations for those structs

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -268,7 +268,7 @@ where
 
 #[derive(Debug, Clone, Copy)]
 pub struct BatchInsert<'a, T: 'a, Tab> {
-    pub(crate) records: &'a [T],
+    pub records: &'a [T],
     _marker: PhantomData<Tab>,
 }
 
@@ -295,7 +295,7 @@ where
 
 #[derive(Debug)]
 pub struct OwnedBatchInsert<V, Tab> {
-    pub(crate) values: Vec<V>,
+    pub values: Vec<V>,
     _marker: PhantomData<Tab>,
 }
 
--- a/diesel/src/query_builder/insert_statement/mod.rs
+++ b/diesel/src/query_builder/insert_statement/mod.rs
@@ -448,7 +448,7 @@ where
 #[doc(hidden)]
 #[derive(Debug, Clone, Copy)]
 pub struct ValuesClause<T, Tab> {
-    pub(crate) values: T,
+    pub values: T,
     _marker: PhantomData<Tab>,
 }
 


Mingun:
I think, then you now need to document them

weiznich:
There is a `#![forbid(missing_docs)]` in the crate root, so ci will fail if documentation is required :wink:

Mingun:
Yes, and as this now is public API for backends my question is: why this is not failed?

weiznich:
It hasn't failed because the corresponding module is hidden from the documentation and therefore not considered part of the public API that wont be changed without major version bumps. 
It is currently not clear how those rules should apply for custom backends because they need to relay on internals that are not documented and may change between minor version bumps.

Mingun:
I understand the reasons for the lack of warnings, but I think the reason and consequence have changed places:
```
NOT public => IS hidden
IS public => NOT hidden
```
And now the situation looks like this:
```
IS hidden => NOT public
NOT hidden => IS public
```
I just point to this strange logic.

May be you think, that API isn't well-developed. But even then, it would be good to leave a reminder of how to use it, for you, for colleagues, for newcomers, if only to describe what is uncomfortable in the current API. If we can say that public and stable API is very difficult to use without documentation, what to say about unstable API? It needs documentation a thousand times more.

> It is currently not clear how those rules should apply for custom backends because they need to relay on internals that are not documented and may change between minor version bumps.

May be introduce the concept of unstable API -- you do not guarantee that it will not change within the major version, but at the same time document it. Thus, it can be officially tried?

weiznich:
> I understand the reasons for the lack of warnings, but I think the reason and consequence have changed places:

No this hasn't changed. There are already a lot of undocumented public types. 

> May be you think, that API isn't well-developed. But even then, it would be good to leave a reminder of how to use it, for you, for colleagues, for newcomers, if only to describe what is uncomfortable in the current API. If we can say that public and stable API is very difficult to use without documentation, what to say about unstable API? It needs documentation a thousand times more.

Undocumented public things are there for several reasons:
* Some macro needs them. We consider this as implementation detail, so nobody should see/use those types.
* Marker types returned by some functions that are basically opaque. Their only functionality should used for such types are wild card trait impls that are documented in other places.

Yes all of those could be documented, but then our documentation would be even more cluttered with unrelated types. It's even now quite common that newcomers don't find things in our documentation because there is already to much in it. Adding more unrelated stuff there won't make this better. 

> May be introduce the concept of unstable API -- you do not guarantee that it will not change within the major version, but at the same time document it? Thus, it can be officially tried.

Again: None of those types need to bother anyone using diesel in an application or even in a library. The only case where those types matter are implementation of custom backends. There won't be much of such backends and to write such a backend you need a really deep understanding of diesel internals. Documentation on this few fields won't help there in any way. You need to understand how those fields interact with the other code and in my opinion the only way to learn this is by reading the actual code.

sgrif:
@Mingun Diesel has always considered "public API" to be "what is rendered in the documentation". This distinction has been required since 0.1, as a large part of our API is based around macros, which may need to call code from Diesel that we don't want you to call directly. Diesel is also split into multiple crates, and there are many functions which need to be called from another Diesel crate, but we do not want our users depending on directly. Having "public, stable API == documented" is a fairly common convention.

This commit does not add anything to the public API. It does allow third party crates to rely on private API if they need to, similar to the undocumented, unstable `group_by` function we added at the request of users some time ago that we know we will break eventually, but lets them do their job in the meantime.

I don't know what the path to stability or if there even is one for these structs. That said, they're pretty stable, and it's reasonable for diesel-oci to depend on them until we have a better answer (though it should probably have a very strict version range since it relies on internals, similar to what we do for reliance on internals for our own crates)

PR #2195, opened 2019-10-10
Bring back `HasSqlType::row_metadata` as this method is not only

### LINKED ISSUE #1985: Find a replacement for the deprecated `HasSqlType::row_metadata` function for use in third party backends
### Versions

- **Diesel:** 2.0

## Problem Description

Diesel 2.0 will remove `HasSqlType::row_metadata()` because it was deprecated and should not be used any more.  Unfortunately implementing a function like this is required to implement a third party backend that adds support for oracle databases. In detail such a backend implementation requires (One need to tell oci what types are coming back…) that there is some way to get a list of response sql types that could be [iterated at runtime](https://github.com/GiGainfosystems/diesel-oci/blob/re_rebalance_coherence/src/oracle/connection/stmt.rs#L177). (Basically just that what `HasSqlType::row_metadata()` has done before.)

Without any deeper investigation I see 2 solutions for this problem:
(Just a unordered mind dump…)

* Add something like `HasSqlType::row_metadata()` back to diesel. Probably just as new trait implemented for all supported tuple sizes?
* Specify that such a backend would be responsible on their own to implement such a trait proposed in the first variant. That would imply that they need to implement trait for tuples up to the same size that diesel implements it's own traits. A way to simplify that would to make the internal `__diesel_for_each_tuple!` macro part of the public api. That would easily allow custom backends (and other crates) to integrate with diesel implement some trait for all tuples up to a certain length approach.

After writing this up I have the feeling that the second solution may be slightly more fitting because that would allow other things than this specific use case. The downside of such an approach would be that we commit to implementing our traits for tuples with a fixed number of elements. That will hold true as long as we do not move to another internal representation (hlists for example) or rust gains the ability to use variadic generics. (Both events would justify a major release therefore exposing that macro should be fine) 

cc @pgab

weiznich:
useful for the mysql backend

Fixes #1985

### DIFF
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -72,7 +72,7 @@ impl Connection for MysqlConnection {
 
         let mut stmt = self.prepare_query(&source.as_query())?;
         let mut metadata = Vec::new();
-        Mysql::mysql_row_metadata(&mut metadata, &());
+        Mysql::row_metadata(&mut metadata, &());
         let results = unsafe { stmt.results(metadata)? };
         results.map(|mut row| {
             U::Row::build_from_row(&mut row)
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -404,8 +404,7 @@ pub trait HasSqlType<ST>: TypeMetadata {
     fn metadata(lookup: &Self::MetadataLookup) -> Self::TypeMetadata;
 
     #[doc(hidden)]
-    #[cfg(feature = "mysql")]
-    fn mysql_row_metadata(out: &mut Vec<Self::TypeMetadata>, lookup: &Self::MetadataLookup) {
+    fn row_metadata(out: &mut Vec<Self::TypeMetadata>, lookup: &Self::MetadataLookup) {
         out.push(Self::metadata(lookup))
     }
 }
--- a/diesel/src/type_impls/option.rs
+++ b/diesel/src/type_impls/option.rs
@@ -19,9 +19,8 @@ where
         <DB as HasSqlType<T>>::metadata(lookup)
     }
 
-    #[cfg(feature = "mysql")]
-    fn mysql_row_metadata(out: &mut Vec<DB::TypeMetadata>, lookup: &DB::MetadataLookup) {
-        <DB as HasSqlType<T>>::mysql_row_metadata(out, lookup)
+    fn row_metadata(out: &mut Vec<DB::TypeMetadata>, lookup: &DB::MetadataLookup) {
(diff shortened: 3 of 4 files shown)


sgrif:
I have the same reservations about this that I had in the issue this is referencing. This function is only relevant to certain backends, as an implementation detail, and I think it's far too easy to shoot yourself in the foot by not implementing it properly. It was originally introduced in https://github.com/diesel-rs/diesel/commit/c8bf33f4ab1669eac1f484239f0d9ad5190566c2, for a reason that is *very* specifically coupled to the use of libmysqlclient. I still believe the solution proposed in #1457 is the right way forward. The only reason it exists at all is because we were passing `&[u8]` as the `RawValue` for `Mysql`. We're no longer doing that, which means we can now remove this function entirely, as I stated I want to do in https://github.com/diesel-rs/diesel/pull/1751#issuecomment-395570018. I would hope other backends can find their own backend specific solutions. Making `__diesel_for_each_tuple!` public API still feels generally right to me. Unfortunately the link at https://github.com/diesel-rs/diesel/issues/1985#issuecomment-463406757 is now dead (protip: always press `y` before copying a link to a line of code), so I can't see the historical counterargument. At the end of the day, it makes me extremely uncomfortable to have code in Diesel which does not need to exist for this crate, but is required for third party crates to exist. There will be other such needs that we didn't anticipate, and I think we should address the reason that the backend specific solutions can't exist in other crates.

weiznich:
@sgrif  Unfortunately it is not possible to fix the issue in diesel-oci with using some equivalent of `__diesel_for_each_tuple!` in combination with a own trait.  The basic issue there is that we [need to create a bind](https://github.com/GiGainfosystems/diesel-oci/blob/595062f6303bec08bef898326b4bd1d98a95ec25/src/oracle/connection/stmt.rs#L508-L520) for each field returned by a query. This bind needs to know the field type. The whole code needs to be called inside of `Connection::query_by_index` which leaves only the methods defined by the trait bounds of this method, so in the end it needs to be some method of a trait defined in diesel to get this information. 
That said I'm not sold to exactly this solution, it's only the best I've coming up with yet.

sgrif:
Specialization would be one way to solve that. A blanket impl over `T where Oracle: HasSqlType<T>` with the default impl and then specialized impls where needed would mean that the bounds on that function imply your local trait.

Given that there's at least one workaround in future versions of the compiler, I don't want to add this to the public API. But if having it private unblocks diesel-oci for now, it seems fine

weiznich:
> Specialization would be one way to solve that. A blanket impl over T where Oracle: HasSqlType<T> with the default impl and then specialized impls where needed would mean that the bounds on that function imply your local trait.

That works, thanks for the hint. As diesel-oci depends on specialization anyway (at least for now) I'm fine with this workaround.

PR #2196, opened 2019-10-10
Do not hard code the setup of the __diesel_schema_migrations table

weiznich:
Different backends (oracle…) could require a different sql comment
here. Similar to the r2d2 method we just provide a function where
everything could be easily handled.

Closes #1676

### DIFF
--- a/diesel/src/migration/mod.rs
+++ b/diesel/src/migration/mod.rs
@@ -57,11 +57,10 @@ impl<'a> Migration for &'a dyn Migration {
 
 /// Create table statement for the `__diesel_schema_migrations` used
 /// used by the postgresql, sqlite and mysql backend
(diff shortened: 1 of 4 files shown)


diesel/src/migration/mod.rs
```diff
@@ -53,3 +54,54 @@ impl<'a> Migration for &'a dyn Migration {
         (&**self).file_path()
     }
 }
+
+/// Create table statement for the `__diesel_schema_migrations` used
+/// used by the postgresql, sqlite and mysql backend
+pub const CREATE_MIGRATIONS_TABLE: &str = "CREATE TABLE IF NOT EXISTS __diesel_schema_migrations (\
```
sgrif: What do you think about moving this to a separate file and using `include_str!` here?

PR #2197, opened 2019-10-10
Cleanup the Backend trait

weiznich:
The `ByteOrder` associated type is documented as only be used if the
`RawValue` is `[u8]` but that's was never true.
We've now a trait (`BinaryRawValue`) that indicates if a backend uses
raw bytes as raw values representation and therefore has a use case
for the `ByteOrder` associated type, so move that type
there. Additionally fix those cases where the type was used
"accidentally" even if that results in some code duplication.


diesel/src/sqlite/types/floats.rs
```diff
@@ -0,0 +1,22 @@
+use byteorder::{NativeEndian, WriteBytesExt};
+use serialize::{self, IsNull, Output, ToSql};
+use sql_types::{Double, Float};
+use sqlite::Sqlite;
+use std::error::Error;
+use std::io::Write;
+
+impl ToSql<Double, Sqlite> for f64 {
```
sgrif: Is there a reason this can't be:

```rust
impl<DB, ByteOrder> ToSql<Double, DB> for f64
where
    DB: Backend + for<'a> BinaryRawValue<'a, ByteOrder = ByteOrder>,
{
    fn to_sql<W: Write>(&self, out: &mut Output<W, DB>) -> serialize::Result {
        out.write_f64::<ByteOrder>(*self)
            .map(|_| IsNull::No)
            .map_err(|e| Box::new(e) as Box<dyn Error + Send + Sync>)
    }
}
```

It seems like that impl would eliminate most of the fallout from this.

weiznich: The problem here is that `Sqlite` not implements `BinaryRawValue`, but uses `SqliteValue` (which is an Enum) as raw value representation.

weiznich:
As discussed the fallout of this is to big compared to potential gains, so just don't do this.

PR #2200, opened 2019-10-14
Update all crates to the 1.4.3 release to fix build

weiznich:
Fixes the ci


JohnTitor:
We should also update here.

https://github.com/diesel-rs/diesel/blob/00f170137e1f38ad920c17f35eaa8c07fc089345/diesel_compile_tests/Cargo.toml#L12-L14

PR #2213, opened 2019-10-31
Make generated schema the same on all platforms

### LINKED ISSUE #2202: Generated schema depends on platform
### Versions

- **Rust:** Latest stable
- **Diesel:** CLI from cargo
- **Database:** Postgres
- **Operating System:** Linux/MacOS (schema differs depending which one we're on)

Reliably generate a `schema.rs` on whichever platform based on a set of migrations.
We are then patching the schema using `diff -u`/`patch`. The order being different breaks the patch, as well as deploy-time checks that ensure the git-tracked file is consistent with the database schema.

On both platforms, `schema.rs` lines are ordered the same way.

The tables are not ordered in the same way.
On Linux:
- d_c_e
- d_cs
- ds
- d_s_c

On Mac:
- d_c_e
- d_cs
- d_s_c
- ds

(Can be noted that the Mac ordering is more consistent. Looks like the Linux sort gets rid of the underscores before sorting.)

### Any clue on how to solve?

I suspect there are no `sort`s on the rust side when generating the schema, and this inconsistency may come more from the postgres side than from the rust side.
(Everytime schema generation was run on Mac so far, the database was also there. I may soon check for what happens when generating on MacOS a schema for a database that lives on a Linux postgres.)

It is possible that the fix is as simple as applying a sort on the rust side.
To limit breakage, the sort applied could be the same as is applied on the Linux version.

Ten0:
Resolves #2202

It appears that the following query does not behave the same way on all platforms:
```sql
SELECT table_name, table_schema from information_schema.tables 
WHERE table_schema = 'your schema' AND table_type LIKE 'BASE TABLE' ORDER BY table_name;
```

The order changes depending on the platform and the version of postgres (more details [in the PostgreSQL bug mailing list](https://www.postgresql.org/message-id/11505.1571936496%40sss.pgh.pa.us)).

This query is used when generating the `schema.rs` in the Diesel CLI.
The fact the order changes depending on the platform creates issues when applying, say, an unified-diff patch to the schema afterwards.

This patch sorts the table names on the rust side so that the order of the tables will not depend on the collation of the `information_schema` table.

PostgreSQL >= 12 has *by default* on all platforms a collation of the `information_schema` table that results in the same generated `schema.rs`s as this patch enforces on all versions/platforms/configurations of Postgres, so this is a *one-shot* controlled somewhat-breaking change that is otherwise already happening *gradually* on all platforms.

### DIFF
--- a/diesel_cli/src/infer_schema_internals/information_schema.rs
+++ b/diesel_cli/src/infer_schema_internals/information_schema.rs
@@ -176,22 +176,25 @@ where
     use self::information_schema::tables::dsl::*;
 
     let default_schema = Conn::Backend::default_schema(connection)?;
-    let schema_name = match schema_name {
+    let db_schema_name = match schema_name {
         Some(name) => name,
         None => &default_schema,
     };
 
     let mut table_names = tables
-        .select((table_name, table_schema))
-        .filter(table_schema.eq(schema_name))
+        .select(table_name)
+        .filter(table_schema.eq(db_schema_name))
         .filter(table_name.not_like("\\_\\_%"))
         .filter(table_type.like("BASE TABLE"))
-        .order(table_name)
-        .load::<TableName>(connection)?;
-    for table in &mut table_names {
-        table.strip_schema_if_matches(&default_schema);
-    }
-    Ok(table_names)
+        .load::<String>(connection)?;
+    table_names.sort_unstable();
+    Ok(table_names
+        .into_iter()
+        .map(|name| TableName {
+            name,
+            schema: schema_name.map(|schema| schema.to_owned()),
+        })
+        .collect())
 }
 
 #[allow(clippy::similar_names)]
--- a/diesel_cli/src/infer_schema_internals/information_schema.rs
+++ b/diesel_cli/src/infer_schema_internals/information_schema.rs
@@ -133,7 +133,7 @@ where
     columns
         .select((column_name, type_column, is_nullable))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name.as_ref()))
+        .filter(table_schema.eq(schema_name))
         .order(ordinal_position)
         .load(conn)
 }
@@ -160,7 +160,7 @@ where
         .select(column_name)
         .filter(constraint_name.eq_any(pk_query))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name.as_ref()))
+        .filter(table_schema.eq(schema_name))
         .order(ordinal_position)
         .load(conn)
 }
@@ -177,10 +177,7 @@ where
     use self::information_schema::tables::dsl::*;
 
     let default_schema = Conn::Backend::default_schema(connection)?;
-    let db_schema_name = match schema_name {
-        Some(name) => name,
-        None => &default_schema,
-    };
+    let db_schema_name = schema_name.unwrap_or(&default_schema);
 
     let mut table_names = tables
         .select(table_name)
@@ -216,10 +213,7 @@ where
     use self::information_schema::table_constraints as tc;
 
     let default_schema = Conn::Backend::default_schema(connection)?;
-    let schema_name = match schema_name {
-        Some(name) => name,
-        None => &default_schema,
-    };
+    let schema_name = schema_name.unwrap_or(&default_schema);
 
     let constraint_names = tc::table
         .filter(tc::constraint_type.eq("FOREIGN KEY"))
--- a/diesel_cli/src/infer_schema_internals/information_schema.rs
+++ b/diesel_cli/src/infer_schema_internals/information_schema.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::error::Error;
 
 use diesel::backend::Backend;
@@ -124,15 +125,15 @@ where
     use self::information_schema::columns::dsl::*;
 
     let schema_name = match table.schema {
-        Some(ref name) => name.clone(),
-        None => Conn::Backend::default_schema(conn)?,
+        Some(ref name) => Cow::Borrowed(name),
+        None => Cow::Owned(Conn::Backend::default_schema(conn)?),
     };
 
     let type_column = Conn::Backend::type_column();
     columns
         .select((column_name, type_column, is_nullable))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name))
+        .filter(table_schema.eq(schema_name.as_ref()))
         .order(ordinal_position)
         .load(conn)
 }
@@ -151,15 +152,15 @@ where
         .filter(constraint_type.eq("PRIMARY KEY"));
 
     let schema_name = match table.schema {
-        Some(ref name) => name.clone(),
-        None => Conn::Backend::default_schema(conn)?,
+        Some(ref name) => Cow::Borrowed(name),
+        None => Cow::Owned(Conn::Backend::default_schema(conn)?),
     };
 
     key_column_usage
         .select(column_name)
         .filter(constraint_name.eq_any(pk_query))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name))
+        .filter(table_schema.eq(schema_name.as_ref()))
         .order(ordinal_position)
         .load(conn)
 }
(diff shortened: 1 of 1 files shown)


diesel_cli/src/infer_schema_internals/information_schema.rs
```diff
@@ -175,23 +176,25 @@ where
...
         .filter(table_name.not_like("\\_\\_%"))
         .filter(table_type.like("BASE TABLE"))
-        .order(table_name)
-        .load::<TableName>(connection)?;
-    for table in &mut table_names {
-        table.strip_schema_if_matches(&default_schema);
-    }
-    Ok(table_names)
+        .load::<String>(connection)?;
+    table_names.sort_unstable();
```
weiznich: If we care about order here we should use `sort` instead of `sort_unstable`.
Additionally why not just use `.sort_by_key()` and then return the table name there instead of doing that dance with the schema name below?

Ten0: > If we care about order here we should use sort instead of sort_unstable.

There are only two differences between a [stable sort](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort) and an [unstable sort](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_unstable):
- The stable sort guarantees the non-reordering of equal elements
- The unstable sort is usually fastest

But this query is not supposed to return any duplicates, so there will never be equal elements.
Besides, since here the comparison function is the default, that is that of two strings being equal, even if there were duplicates, it wouldn't change anything to what happens afterwards. 

Using stable sort instead of unstable sort may have an effect when using `sort_by` or `sort_by_key` in very specific scenarios where you need to preserve the order of initial elements in case two elements are equal, but in the general case unstable sort is strictly better.

We are not in such a case, which is why I favored the unstable sort.

> Why not just use .sort_by_key() and then return the table name there instead of doing that dance with the schema name below?

Because it took me more time than I would have liked to understand exactly what was happening because what the code was doing is:
- Get the name of the default schema
- Query the table name AND the schema name with a constraint on the schema name being something we already know
- Post-patch the schema name so that if we initially queried None which led to the schema name being that of the default schema we get None back in the table name (and doing so by checking every `TableName` even though they could only all match or all not match)

Though what we want to do is simply:
- Query the table names for a specific schema
- Create objects that pack both the table name and the schema name that we asked for.

The previous approach raised questions like:
- Why is the check on whether we should erase the schema made on all the tables one by one ? Do they have different values ?
- Why do we query the schema name with the table name ? Is there some weird case somewhere where it could be different from what we called the function with ?

Overall I feel like this new approcach is much more straightforward hence easier to understand and feels much less as a dance to me than the previous one. (And as a bonus may dodge a query on MySQL backends).

Ten0: I have just realized while going to bed that there's actually a difference between my implementation and the previous one: if the argument is `Some(default_schema_name)` it used to be turned into `None`, while my implementation didn't do that.

I've updated the code to preserve that previous behaviour, still in a more explicit way than used to be.

However the systematic use of `strip_schema_if_matches` on the database output also guaranteed that behaviour so there was more to it than I originally thought, so I cannot advocate for my implementation as much as I did. However if we revert this to the previous implementation, we should add some documentation about why it's there because it was non-obvious enough that I went and rewrote it.

Ten0:
The error on nightly seems unrelated.

Ten0:
Hello,
I'd be interested to have more thoughts/a merge on this. It is currently causing issues in my team where only 20% of the developers can automatically refresh the schema, and with the upcoming updates in Postgres more and more people are going to start having this issue.

weiznich:
I've currently no time to deal with that now, sorry:pensive:. If not somebody other from the contributors team find some time to do a final review this needs to wait at least till the week after the next one.
(In that context: Please note that we are currently doing most if the work on diesel in our free time. If you need in time support for a commercial project consider to found someone (for example Sean) that actively works on diesel.)

Ten0:
> If not somebody other from the contributors team find some time to do a final review this needs to wait at least till the week after the next one

Thank you for the update! :)

> Please note that we are currently doing most if the work on diesel in our free time. If you need in time support for a commercial project consider to found someone (for example Sean) that actively works on diesel

I am aware of this, and I thank you very much for all of this work! <3
Diesel is amazing and it enables us to write very reliable software with DB storage in Rust! Unfortunately though, we are still a very small structure, which has the upside of not having technical debt, hence our ability to use such an amazing programming language and software, but also the downside that we don't have enough money to start funding open-source projects maintainers yet, only for contributing to them on an as-needed basis. Hopefully in a few years... ;)

diesel_cli/src/infer_schema_internals/information_schema.rs
```diff
@@ -124,15 +125,15 @@ where
...
+        Some(ref name) => Cow::Borrowed(name),
+        None => Cow::Owned(Conn::Backend::default_schema(conn)?),
     };
 
     let type_column = Conn::Backend::type_column();
     columns
         .select((column_name, type_column, is_nullable))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name))
+        .filter(table_schema.eq(schema_name.as_ref()))
```
weiznich: `Cow` already implements `AsExpression` so no need to call `.as_ref()` here explicitly.
```suggestion
        .filter(table_schema.eq(schema_name))
```

diesel_cli/src/infer_schema_internals/information_schema.rs
```diff
@@ -151,15 +152,15 @@ where
...
+        Some(ref name) => Cow::Borrowed(name),
+        None => Cow::Owned(Conn::Backend::default_schema(conn)?),
     };
 
     key_column_usage
         .select(column_name)
         .filter(constraint_name.eq_any(pk_query))
         .filter(table_name.eq(&table.name))
-        .filter(table_schema.eq(schema_name))
+        .filter(table_schema.eq(schema_name.as_ref()))
```
weiznich:
```suggestion
        .filter(table_schema.eq(schema_name))
```

diesel_cli/src/infer_schema_internals/information_schema.rs
```diff
@@ -176,22 +177,27 @@ where
     use self::information_schema::tables::dsl::*;
 
     let default_schema = Conn::Backend::default_schema(connection)?;
-    let schema_name = match schema_name {
+    let db_schema_name = match schema_name {
```
weiznich: Just 
```suggestion
    let db_schema_name = schema_name.unwrap_or(&default_schema);
```
should be fine

diesel_cli/src/infer_schema_internals/information_schema.rs
weiznich: >      There are only two differences between a stable sort and an unstable sort:
>>    The stable sort guarantees the non-reordering of equal elements
>>   The unstable sort is usually the fastest
> 
> But this query is not supposed to return any duplicates, so there will never be equal elements.
> Besides, since here the comparison function is the default, that is that of two strings being equal, even if there were     duplicates, it wouldn't change anything to what happens afterwards.

> Using stable sort instead of unstable sort may have an effect when using sort_by or sort_by_key in very specific scenarios where you need to preserve the order of initial elements in case two elements are equal, but in the general case unstable sort is strictly better.

> We are not in such a specific case, which is why I favored the unstable sort.

Right, forgot that `table` names should be unique here 🤦‍

> Overall I feel like this new approcach is much more straightforward hence easier to understand and feels much less as a dance to me than the previous one. (And as a bonus may dodge a query on MySQL backends).

Normally we try to leave old code in place because `git blame` then points at the correct commits that have introduced that code and contain that contain additional information about the implementation. In this case the change seems fine because it actually improves the implementation quite significantly.

weiznich:
> I am aware of this, and I thank you very much for all of this work! <3
Diesel is amazing and it enables us to write very reliable software with DB storage in Rust! Unfortunately though, we are still a very small structure, which has the upside of not having technical debt, hence our ability to use such an amazing programming language and software, but also the downside that we don't have enough money to start funding open-source projects maintainers yet, only for contributing to them on an as-needed basis. Hopefully in a few years... ;)

Another welcome contributing would be someone helping out with code reviews and similar things from time to time:wink:.

diesel_cli/src/infer_schema_internals/information_schema.rs
Ten0: Following your principle of not touching too much I hadn't touched this part but I agree :)

I've also updated in the same way a few lines under.

PR #2218, opened 2019-11-11
remove confusing panic messages from examples

r-arias:
The `publish_post` examples contained a confusing `panic!()` that
suggested the update statement would return an error, when a post to be
updated could not be found. In reality, however, that case results in an
`Ok(0)` value.

To avoid confusion, this commit replaces the panicking
`unwrap_or_else()` calls with a simpler `unwrap()`.

### DIFF
--- a/examples/postgres/getting_started_step_3/src/bin/publish_post.rs
+++ b/examples/postgres/getting_started_step_3/src/bin/publish_post.rs
@@ -19,6 +19,6 @@ fn main() {
     let post = diesel::update(posts.find(id))
         .set(published.eq(true))
         .get_result::<Post>(&connection)
-        .unwrap_or_else(|_| panic!("Unable to find post {}", id));
+        .unwrap();
     println!("Published post {}", post.title);
 }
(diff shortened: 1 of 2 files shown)


r-arias:
woah, are all these failures my fault?

weiznich:
The ci is  broken on beta because of https://github.com/rust-lang/rust/issues/66295

JohnTitor:
Could you rebase this to fix the CI?

r-arias:
Done, I think :)

examples/mysql/getting_started_step_3/src/bin/publish_post.rs
```diff
@@ -16,10 +16,7 @@ fn main() {
         .expect("Invalid ID");
     let connection = establish_connection();
 
-    let post: Post = posts
-        .find(id)
-        .first(&connection)
-        .unwrap_or_else(|_| panic!("Unable to find post {}", id));
+    let post: Post = posts.find(id).first(&connection).unwrap();
```
weiznich: The error message was correct in that case, or do I miss something?

r-arias: uh, I might have been too quick there. let me look at it again.

r-arias: Yeah, sorry! Good catch! :flushed:

PR #2221, opened 2019-11-17
Update major versions of dependencies

paolobarbolini:
This updates all major versions of dependencies.

I wasn't able to update `structopt` in [examples/postgres/advanced-blog-cli/Cargo.toml](https://github.com/paolobarbolini/diesel/blob/e19fa752d0c2e8c4d3534421eafb9b96299a2228/examples/postgres/advanced-blog-cli/Cargo.toml) because the newer versions apparently don't allow setting `default_value` on `Option`.

I replaced `tempdir` with `tempfile` which was already being used in other parts of the code.

I wasn't able to run tests for all database backends so I'll rely on CI to find out if there are any problems left to fix.


paolobarbolini:
I had to remove the `dotenv` update from `diesel_tests` because `diesel_infer_schema` still depends on older versions of it and breaks the test build when running with newer versions.

https://github.com/diesel-rs/diesel/blob/1cfdd44460534556cc1e5aa69fdb42ca52665542/diesel_infer_schema/Cargo.toml#L18

h-michael:
If you upgrade many library versions at once, the reviewer will be hard, so why not split it into several PRs?

PR #2229, opened 2019-11-27
Allow `unused_parens` temporarily

JohnTitor:
It's annoying on CI so we should allow this lint temporarily, I think.

### DIFF
--- a/diesel_tests/tests/joins.rs
+++ b/diesel_tests/tests/joins.rs
@@ -1,5 +1,5 @@
-use schema::*;
 use diesel::*;
+use schema::*;
 
 #[test]
 fn belongs_to() {
--- a/diesel_tests/tests/joins.rs
+++ b/diesel_tests/tests/joins.rs
@@ -1,4 +1,4 @@
-use super::schema::*;
+use schema::*;
 use diesel::*;
 
 #[test]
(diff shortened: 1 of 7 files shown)


weiznich:
I would rather work on fixing that in rustc. The critical line in rustc is [that](https://github.com/rust-lang/rust/blob/a44774c3a9739b2eea8923e09d67b14312c78ef3/src/librustc_lint/unused.rs#L358-L359) one. If someone knows how to check there if we are in a macro expansion it should be simple to be fixed.

JohnTitor:
This is just a stopgap. Actually, the lint errors hide test failure so it's worth to avoid the lint failure, for now, I think. And it will take some time until the fix lands on beta unless backport.

diesel_tests/tests/joins.rs
```diff
@@ -678,13 +678,13 @@ fn connection_with_fixture_data_for_multitable_joins() -> (TestConnection, TestD
         .order(posts::id)
         .load::<Post>(&connection)
         .unwrap();
-    let new_comments = vec![
+    let new_comments: &[NewComment<'static>] = &[
         NewComment(posts[0].id, "First Comment"),
```
weiznich: Any idea why this fails on nightly? Should we maybe report a regression against rustc?

JohnTitor: I have no idea. We haven't changed behavior for it and the failure only appears on nightly so I guess recent changes on nightly cause the failure (I'm not sure it's a regression or expected behavior though).

weiznich: I've retriggered the ci. Seems to be fixed now. You need to run rustfmt.

PR #2230, opened 2019-11-27
Use migration directory during pending CLI command

Koisell:
The diesel_cli migration pending wasn't using the
`--migration-dir <MIGRATION_DIRECTORY>` option. We now use it at
parameter of any_pending_migrations in the CLI.

### DIFF
--- a/diesel_migrations/migrations_internals/src/lib.rs
+++ b/diesel_migrations/migrations_internals/src/lib.rs
@@ -179,6 +179,13 @@ where
     Conn: MigrationConnection,
 {
     let migrations_dir = find_migrations_directory()?;
+    any_pending_migrations_in_directory(conn, &migrations_dir)
+}
+
+pub fn any_pending_migrations_in_directory<Conn>(conn: &Conn, migrations_dir: &Path) -> Result<bool, RunMigrationsError>
+where
+    Conn: MigrationConnection,
+{
     let all_migrations = migrations_in_directory(&migrations_dir)?;
     setup_database(conn)?;
     let already_run = conn.previously_run_migration_versions()?;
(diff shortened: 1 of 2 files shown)


diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -174,11 +174,10 @@ where
 ///
 /// See the [module level documentation](index.html) for information on how migrations should be
 /// structured, and where Diesel will look for them by default.
-pub fn any_pending_migrations<Conn>(conn: &Conn) -> Result<bool, RunMigrationsError>
+pub fn any_pending_migrations<Conn>(conn: &Conn, migrations_dir: &Path) -> Result<bool, RunMigrationsError>
```
weiznich: That's a breaking change. I would rather like to not doing that. Instead add a new method named `any_pending_migrations_in_directory` containing the relevant part of this method. Then the existing method should just internally call the new method with the infered migration directory.

Koisell:
I understand perfectly. I will submit an another version.

Le jeu. 28 nov. 2019 à 21:00, Georg Semmler <notifications@github.com> a
écrit :

Ten0:
Can't compile diesel cli master through cargo install now due to the old "migrations" crate still being used. Would it be possible to publish?

weiznich:
Publishing a new version of `diesel_cli` is currently not planed because master could contain potential breaking changes. If you really need to install `diesel_cli` from the master branch you need to clone the repo and call `cargo install --path diesel_cli` from there, but be aware that this may result in a diesel_cli version that is not compatible with the 1.4 release.

Ten0:
Nice, didn't know that would work. I will use that as a workaround. What's the ETA for 2.0 and will this be published then?

weiznich:
ETA for 2.0 is: When we've implemented everything that is planed to require a breaking change. I do not give any estimate on this because it's an open question how many time contributors find to work on this. This change will be published then if we don't backport it to the 1.4 branch for another point release.

PR #2234, opened 2019-12-05
Correctly enforce nullability with nested joins

### LINKED ISSUE #2001: left_join makes it possible to try to stuff nullable values into a non-nullable Queryable
## Setup

Pardon the vagueness, my database-fu is too weak to know how to explain this more concisely.  This is also slightly simplified so that I can try to explain it at all.

We have two tables, TableA and TableB.  We have two `Queryable` structs, `QueryA` and `QueryB`.  TableA and TableB have no nullable fields, and `QueryA` and `QueryB` contain no `Options`, and life is good.

Now we add this:

```rust
#[derive(Queryable, ...)]
struct QueryC {
    a: QueryA,
    b: QueryB,
}
```

We can write a diesel query that does an inner join on A and B and stuffs the result into `QueryC`, and it works:  `table_a::dsl::table_a.inner_join(table_b::dsl::table_b).select(...).get_results<QueryC>()`.

However, when we do a `left_join()` instead, it *compiles* but panics at runtime with `'Could not get crate list?: DeserializationError(UnexpectedNullError)'`.  The `left_join()` produces, basically, `(QueryA, Option<QueryB>)` but that's not what `QueryC` expects, and trying to make it contain that fails to compile.

...except now that I test more, it *does* not-compile, as is correct, but when you do what I'm actually trying to do (which is wrap the whole bloody thing in *yet another* inner join) it fails.  Heck, I'm just going to give you the raw code:

```rust
table! {
    crates (id) {
        id -> Int8,
        name -> Varchar,
    }
}

table! {
    crate_versions (id) {
        id -> Int8,
        crate_id -> Int8,
        version -> Varchar,
        checksum -> Varchar,
        yanked -> Bool,
    }
}

table! {
    analyze_cratefile (id) {
        id -> Int8,
        crate_versions_id -> Int8,
        size -> Int8,
        valid -> Bool,
        checksum_matches -> Bool,
    }
}

    // This is the bit that compiles and shouldn't, and panics at runtime
    let crate_versions = c::crates
        .inner_join(cv::crate_versions.left_join(ac::analyze_cratefile))
        .select((
            (cv::id, c::name, cv::version, cv::checksum, cv::yanked),
            (ac::size, ac::valid, ac::checksum_matches),
        ))
        .order_by(c::name)
        .get_results::<CrateWithAnalysis>(conn)
        .expect("Could not get crate list?");
```

I hope you can make heads or tails of this, 'cause I can't.  Please let me know if you need any more information and I'll try to provide, once I've had more than six hours of sleep.

### Versions

- **Rust:** 1.32
- **Diesel:** 1.4
- **Database:** PostgreSQL 11
- **Operating System** Linux

### Feature Flags

- **diesel:** postgres, extras

## Checklist

sgrif:
When multiple joins are nested together, the resulting type is
essentially a binary tree of each individal join. What this means is
that Diesel only has to consider one join at a time in trait
implementations.

Prior to this commit, when considering nullability, we mistakenly only
considered the outermost join. "Considering nullability" in this case
means `impl SelectableExpression<TheJoin> for column`. If that trait is
implemented, we allow selecting the column directly. If it appears on
the right side of a left join, we do not want that trait to be
implemented, but we do want it implemented for `Nullable<column>`.

The problem code lived in two generated impls in the `table!` macro. For
each individual column, we generate two relevant `SelectableExpression`
impls: one for inner joins, and one for left joins. Both of these impls
were wrong.

The impl for left joins was written as:

    Left: AppearsInFromClause<$table, Count=Once>,
    Right: AppearsInFromClause<$table, Count=Never>,

which in English means "where my table appears once in the left side of
the join, and doesn't appear in the right side of the join". This
incorrectly assumes that `$table` is the left-most table, or all joins
contained in `Left` are inner joins. If `Left == Join<other_table,
$table, LeftOuter>`, this constraint incorrectly holds. This first case
was much easier to solve. The first line is changed to:

    Self: SelectableExpression<Left>

The inner join case was much harder to solve. The constraint was written
as:

    Join<Left, Right, Inner>: AppearsInFromClause<$table, Count=Once>,

In English, this means "my table appears in this join clause exactly
once". This fails to consider that the position the table appears in may
be the right side of a left join. What we really want to write here is
something like this:

    Self: SelectableExpression<Left> | SelectableExpression<Right>,

Where the pipe means "either of these constraints are true". Of course
no such impl exists. If we write two separate impls, they are
overlapping (and overlapping marker traits are not stable). So we need
something different.

Since the problem case applies only to columns, we know that the only
reason that `SelectableExpression<Left>` and
`SelectableExpression<Right>` both hold is if the table appears more
than once in the from clause, which we already reject. Because of this,
we know that when we want this impl to apply, `(<Left as
AppearsInFromClause<$table>>::Count, <Right as
AppearsInFromClause<$table>>::Count>)` will either be `(Once, Never)` or
`(Never, Once)`.

Now that we know we have two disjoint types, we can use that to
implement a new trait, which gives us either `Left` or `Right` depending
on which of the two counts was `Once`, and use that as the parameter for
`SelectableExpression`. In English, the new bounds mean "I am selectable
from whichever of these two type parameters contains my table"

Since this new trait is only meant as a hack for this single impl, and
is not public API, I wanted to avoid having it appear in compiler
errors. For this reason, I've added additional impls for all the invalid
count combinations. This will lead to an error such as:

> table3::id: SelectableExpression<this_table_does_not_appear_in_the_from_clause>

instead of

> (Never, Never): Select<table1, table2>

I did not add explicit compile tests for these errors, as I want to
allow for the deletion of those impls if they become a problem in the
future.

This lead to the final challenge which is that our `impl
SelectableExpression<Join<...>> for Nullable<T>` actually relied on this
bug. We considered `Nullable<T>: SelectableExpression<AnyJoin>` if `T:
SelectableExpression<InnerJoin>`. This of course would break if the
actual table is on the right side of a nested left join. To work around
this, we've introduced another new trait which recursively converts all
joins in a tree to inner joins, which we then use for our check. This
also simplified the various impls for `Nullable<T>`, and should
correctly ensure that `Nullable<T>` is always usable.

Fixes #2001.


diesel/src/query_source/mod.rs
```diff
@@ -76,3 +76,83 @@ pub trait AppearsInFromClause<QS> {
...
+///     Right: AppearsInFromClause<T::Table>,
+///     (Left::Count, Right::Count): Select<Left, Right>,
+///     T: SelectableExpression<
+///         <(Left::Count, Right::Count) as Select<Left, Right>>::Selection,
+///     >,
+/// ```
+///
+/// In order to aquire the counts in the first place, we must already know
+/// the table we're searching for.
+pub trait Select<Left, Right> {
```
weiznich: I would vote for another name, because `Select` collides with select from SQL statements.
`MatchByOnce`?

diesel_compile_tests/tests/compile-fail/right_side_of_left_join_requires_nullable.rs
```diff
@@ -18,26 +18,122 @@ table! {
...
     // Valid -- NULL to a function will return null
     let _ = join.select(lower(posts::title).nullable());
     // Invalid, only Nullable<title> is selectable
     let _ = join.select(lower(posts::title));
     //~^ ERROR E0271
+    //~| ERROR E0277
+    // Invalid, Nullable<title> is selectable, but lower expects not-null
+    let _ = join.select(lower(posts::title.nullable()));
+    //~^ ERROR E0271
     //~| ERROR E0271
```
weiznich: We should probably add one test for `pets` here? Just to be sure…
(Also for the other tests below?)

sgrif: We've got a lot of tests for the right side of the outermost join in this file and elsewhere. We can always add more, but they'd be redundant

weiznich: Oh right, that's the outer type wrapper in the final type. I've missed that. So just ignore this comment.

diesel/src/query_source/mod.rs
sgrif: Renamed to `Pick`

Ten0:
I've tested this on the huge code we're about to set in production and it highlighted a few errored joins.

I've also noted that it refused some code that was somewhat valid (in the sense that it worked):
```rust
table_a::table.left_join(table_b::table)
	.select(table_a::field_a.is_not_null().and(
		table_a::field_a
		.eq(table_b::field_b))
```

I had to change it to:
```rust
table_a::table.left_join(table_b::table)
	.select(table_a::field_a.is_not_null().and(
		table_a::field_a
		.eq(table_b::field_b.nullable()))
```
even though `field_b` is already a `Nullable<Integer>`.

PR #2241, opened 2019-12-13
Remove unused metadata field from `PgValue`

sgrif:
This was originally included in #2134 with the following comment:

> Additionally some type information are added to the postgresql value
> type. Those information may allow to do dynamic type conversations in
> future diesel versions.

While this may be useful in the future, we should wait until we have a
concrete use case for doing so. At the moment this means we have to
thread a `PgConnection` through various types that otherwise wouldn't
need to carry it, and are stabilizing an API that we aren't even sure
if/how we want to eventually use. Given that every use of `PgConnection`
complicates providing an async implementation, we should be more
conservative about speculative APIs like this.

### DIFF
--- a/diesel/src/pg/connection/cursor.rs
+++ b/diesel/src/pg/connection/cursor.rs
@@ -9,15 +9,15 @@ use std::marker::PhantomData;
 
 /// The type returned by various [`Connection`](struct.Connection.html) methods.
 /// Acts as an iterator over `T`.
-pub struct Cursor<'a, ST, T> {
+pub struct Cursor<ST, T> {
     current_row: usize,
-    db_result: PgResult<'a>,
+    db_result: PgResult,
     _marker: PhantomData<(ST, T)>,
 }
 
-impl<'a, ST, T> Cursor<'a, ST, T> {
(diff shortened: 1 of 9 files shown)


weiznich:
I totally understand the motivation to remove this field, but getting the metadata lookup is required[ to implement `FromSql` in a dynamic way. That`s required for dynamic select clauses, see for example [this impl](https://github.com/diesel-rs/diesel-dynamic-schema/pull/10/files#diff-ff3ba1a32fc76af46480206b6549dfd5R18-R40). For the standard types it may be possible to just hard code the oid values there. For custom types this is not possible, so the users needs some way to lookup the correct oid in the database. This requires doing a database call in the implementation of `FromSql`, so we need a connection there in some way. In my opinion the most straight forward approach here is to just "reuse" the `HasSqlType<ST>` impls in such cases. Yes that's currently not fully implemented in diesel master, but there are open PR's pointing out this API. (That's more than for the async implementation).

weiznich:
As discussed it makes sense to remove this to unblock async. We may want to look into a different API to unblock this if someone needs to support custom types as part of the dynamic loading API.

PR #2242, opened 2019-12-14
Remove a workaround for an old rust version

weiznich:
Our minimal supported rust version is now >= 1.30, so remove this workaround


JohnTitor:
Hm, compilation is failed?

weiznich:
Yes, I've already noticed this. Seems to be related to this change. I will try to look into this.
(That was a drive by change I've noticed while reading the corresponding code. Seems like I've should have done more tests :see_no_evil:

sgrif:
You need to remove the `inner_of_option_ty` here: https://github.com/diesel-rs/diesel/pull/2242/files#diff-781e467abf942d7a01a3334a0c5a18feL46

JohnTitor:
Uhm, the failure on Pipelines is unrelated?

PR #2243, opened 2019-12-16
Update dotenv repo url

dns2utf8:
I also added a command that does not require root access and updated the completions automatically with the binary.

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -120,7 +120,7 @@ Runs the `down.sql` and then the `up.sql` for the most recent migration.
 Prints table definitions for database schema.
 
 [pending-migrations]: https://docs.rs/diesel_migrations/*/diesel_migrations/fn.run_pending_migrations.html
-[rust-dotenv]: https://github.com/slapresta/rust-dotenv#examples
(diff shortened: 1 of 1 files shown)


diesel_cli/README.md
```diff
@@ -131,7 +131,7 @@ Diesel can generate a bash completion script for itself:
 #### linux
 
 ```sh
-$ diesel bash-completion > /etc/bash_completion.d/diesel
+$ echo 'source <(diesel completions bash)' >> ~/.bashrc
```
weiznich: I've tried this command locally. It does not work for me, but this is probably related to my shell. We should probably just reword this section to just saying something like: 
Diesel can generate shell completion scripts via `diesel completions <shell>`. See the help message of this command for a list of supported shells. Checkout the documentation of your shell to see how you could use this completions.

dns2utf8: Good point. Another hint, this works with bash only and you have to start a new shell or run `source ~/.bashrc` in the current one.

TaKO8Ki:
The problem @dns2utf8 tried to solve in this pull request is already solved in the following PRs So, I closed this PR.
- https://github.com/diesel-rs/diesel/pull/2530
- https://github.com/diesel-rs/diesel/pull/2511

PR #2246, opened 2019-12-17
Use unnamed free constants to wrap derive impls

sgrif:
As of Rust 1.37, we no longer need to come up with unique names to use
for everything, we can just use an unnamed constant. We originally
switched away from wrapping in a constant to work around a bug in Rust
that appears to have since been fixed.

I suspect there are some old workarounds floating in the codebase that
we can remove with this change, but none immediately stood out to me, so
they can be removed the next time they get touched (if at all)


.travis.yml
```diff
@@ -65,7 +65,7 @@ matrix:
     - SQLITE_DATABASE_URL=/tmp/test.db
     script:
     - (cd diesel_cli && cargo +$TRAVIS_RUST_VERSION test --no-default-features --features "sqlite-bundled")
-  - rust: 1.36.0
+  - rust: 1.37.0
     name: "Minimal supported rust version == 1.36.0"
```
JohnTitor:
```suggestion
     name: "Minimal supported rust version == 1.37.0"
```

PR #2248, opened 2019-12-17
Update our macros to work in the 2018 style

sgrif:
Note: This includes #2247. That PR should be merged first. Look at the last commit only for a smaller diff.

This updates our crate to work without `#[macro_use] extern crate
diesel` at the crate root. I've audited all uses of `#[macro_export]`.
If it was obvious at a glance that inner macros are local, I've added
`local_inner_macros`. Otherwise I changed it to use `$crate::`. Most of
the non-public macros can probably have the `__diesel_` prefix removed,
but I've left it on to keep things reasonable for 2015 edition users and
to minimize the diff.

We actually have a surprisingly small number of non-derive macros that
are public API. It's pretty much limited to our schema macros and
operator generation. The schema macros are exported from the prelude,
the operator generation macros are not as they had their prefix removed
in 2.0 with the expectation they are always invoked with `diesel::`.

I've removed all instances of `extern crate diesel` from the doctests.
Note that this does not change anything outside the main crate. In
particular, `diesel_cli` will still generate a `schema.rs` that assumes
`#[macro_use] extern crate diesel;`. It will need to import the prelude
at the top of that file. This will be addressed once that crate is moved
to the 2018 edition.

Close #1956


diesel/src/lib.rs
```diff
@@ -292,6 +291,9 @@ pub mod helper_types {
 
 pub mod prelude {
     //! Re-exports important traits and types. Meant to be glob imported when using Diesel.
+    #[doc(inline)]
+    pub use diesel_derives::*;
```
weiznich: I think the better strategy here is to put these reexports in those places where the actual traits are defined that are implemented by the corresponding custom derive. If that trait is reexported via prelude, the proc macro will also be reexported.

diesel_derives/src/lib.rs
```diff
@@ -59,56 +90,278 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
...
+///
+/// ## Required type attributes
+///
+/// * `#[table_name = "some_table_name"]`, specifies the table this type
+/// is insertable into. Requires that `some_table_name` is in scope.
+///
+/// ## Optional field attributes
+///
+/// * `#[column_name = "some_table_name"]`, overrides the column the current
+/// field maps to to `some_table_name`. By default the field name is used
```
pickfire:
```suggestion
/// field maps to `some_table_name`. By default the field name is used
```

diesel_derives/src/lib.rs
```diff
@@ -59,56 +90,291 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
...
 #[proc_macro_derive(Insertable, attributes(table_name, column_name, diesel))]
 pub fn derive_insertable(input: TokenStream) -> TokenStream {
     expand_proc_macro(input, insertable::derive)
 }
 
+/// Implements `NonAggregate`
+///
+/// This derive can be used for structs with no type parameters,
+/// which are not aggregate, as well as for struct which are
+/// `NonAggregate` if all type parameters arse `NonAggregate`
```
pickfire:
```suggestion
/// `NonAggregate` if all type parameters ars `NonAggregate`.
```

diesel_derives/src/lib.rs
```diff
@@ -59,56 +90,291 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
...
+///   single database column, but is a type that implements
+///   `QueryableByName` on it's own
 #[proc_macro_derive(QueryableByName, attributes(table_name, column_name, sql_type, diesel))]
 pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
     expand_proc_macro(input, queryable_by_name::derive)
 }
 
+/// Implement necessary traits for adding a new sql type
+///
+/// This trait implements all necessary traits  to define a
```
pickfire:
```suggestion
/// This trait implements all necessary traits to define a
```

weiznich:
I would like to merge this PR soon. 
Things to address in a review:
* Are the new documentation for the derives understandable?
* Are all public attributes for the custom derives documented?
* Are the reexport locations for the proc macros reasonable?
* Are we willing to break existing crates that use `#[macro_use] extern crate diesel;` to import the custom derives? Otherwise we need to readd the wildcard reexport in the crate root. (Unfortunately there is no good way to [deprecate a reexport](https://github.com/rust-lang/rust/issues/47236) 😞 ).

diesel/src/associations/mod.rs
```diff
@@ -389,25 +386,8 @@ impl<'a, T: HasTable> HasTable for &'a T {
 /// `update(YourStruct::table().find(&your_struct.primary_key())`).
 ///
 /// This trait is usually implemented on a reference to a struct,
-/// not the struct itself.
-///
-/// ### Deriving
+/// not the struct itself. It can be [derived](derive.Identifiable.html).
```
martell: `not on the struct itself` ?

diesel_tests/Cargo.toml
```diff
@@ -17,6 +17,7 @@ chrono = { version = "0.4" }
 diesel = { path = "../diesel", default-features = false, features = ["quickcheck", "chrono", "uuidv07", "serde_json", "network-address", "numeric", "with-deprecated"] }
 diesel_migrations = { version = "1.4.0" }
 diesel_infer_schema = { version = "1.4.0" }
+diesel_derives = { version = "1.4" }
```
martell: not exactly `1.4.0` ?

martell:
> Are the new documentation for the derives understandable?

The only thing nit I could find I added in a comment.

> Are we willing to break existing crates that use #[macro_use] extern crate diesel; to import the custom derives? Otherwise we need to readd the wildcard reexport in the crate root. (Unfortunately there is no good way to deprecate a reexport 😞 ).

I am not a @diesel-rs/contributors but I guess it makes sense for it to be okay do that given that we are moving to 2.0 - this is the best time for breaking changes.

I can't comment on the other 2 questions but I just wanted to bump this with the 2 potential nits I found.

Hoping you can merge this soon :)

diesel/src/associations/mod.rs
pickfire: What should I change that to? But I believe that is for another issue, maybe not in this pull request.

diesel_tests/Cargo.toml
pickfire: No, I did not change these too, it should be in another issue.

pickfire:
> 
>>   Are the new documentation for the derives understandable?
>
> The only thing nit I could find I added in a comment.

@martell But I believe those should be in another issue right? Or if you want, can just send in a suggestions. I think it would be better to just keep as it is an have another issue to solve it, or maybe in another commit?

diesel/src/associations/mod.rs
martell: Ahh you are correct, it had this exact language before, I will create a PR after this is merged.

For context I think
`This trait is usually implemented on a reference to a struct, not the struct itself.`
should be
`This trait is usually implemented on a reference to a struct, not on the struct itself.`

diesel_tests/Cargo.toml
martell: I'm a little lost on the context of the reply.

I was just trying to say that 

`diesel_derives = { version = "1.4" }`
should be
`diesel_derives = { version = "1.4.0" }`

like the others around it
```
diesel_migrations = { version = "1.4.0" }
diesel_infer_schema = { version = "1.4.0" }
```

I should really be using the change suggestion feature on github.

pickfire: Sorry, @martell I just noticed that my context was also wrong, I thought I was in another repository and another pull request, it also have a similar name.

diesel/src/associations/mod.rs
pickfire: Sorry, @martell I just noticed that my context was also wrong, I thought I was in another repository and another pull request, it also have a similar name.

Do we need the `usually`? Or should we have an example for this?

martell:
🎉

PR #2253, opened 2019-12-20
Fix travis build

sgrif:
The build on master is currently broken. This is because the lockfile
created by Cargo in 1.41 is incompatible with earlier versions of Cargo.
Because the `test_all` files in the example directory aren't adding
`+$TRAVIS_RUST_VERSION`, our examples get run with the version specified
in `rust-toolchain` instead of the intended version for that piece of
the matrix.

Since everything in the main workspace uses a single lockfile, this
means that beta started breaking. We run the tests for the main crate
and derives first, which (correctly) runs on 1.41 and generates a
lockfile. When we then try to run the examples, they (incorrectly) use
the version in `rust-toolchain`, which cannot read the lockfile
generated by 1.41.

To fix this I just nuke the file before the build starts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,28 +21,29 @@ before_script:
   - pip install 'travis-cargo<0.2' --user
   - export PATH=$HOME/.local/bin:$PATH
   - mysql -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'travis'@'%';" -uroot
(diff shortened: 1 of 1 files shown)


sgrif:
image

Going to merge once the build is green

sgrif:
Travis is green (which is what this fixes). I do see windows PG failing on pipelines, I'll open a separate PR for that if it's not intermittent.

PR #2254, opened 2019-12-20
Implement support for aliasing

weiznich:
**This is currently WIP and should only be seen as demo and a request for comments**
This PR takes another try on implement aliasing support. 

# Example
As first point just a short example how the API would look like:

```rust
    let user_alias = alias!(users as user_alias);

    posts::table
        .inner_join(user_alias)
        .inner_join(users::table)
        .select((user_alias.field(users::name), users::name))
        .load::<(String, String)>(&connection)
        .unwrap();
```

# Feature

The current implementation allows:
* to define aliases
* to access columns of a defined alias (maybe we should also implement the API proposed by sean in the other PR here?)
* to use multiple aliases of the same/different tables to write subqueries
* to join aliases of the same/different table using explicit and implicit joins

# Implementation

One of our problems with providing support for aliasing was that we were not sure how a good type safe API would look like. Sean already opened #1773 which would provide a type safe API, that requires more steps to setup an alias then we would like to have. Another idea floating around was to have an alias type in diesel itself, that would just wrap tables (and maybe later also other things), which would allow us to provide a much concrete API that wouldn't require that much steps to setup. We thought this was blocked on specialization because we would need to implement `AppearsInFromClause` for every combination and alias. In detail we need to implement that trait in such a way that the associated type `Count` is `Never` for all combination besides table and alias type are the same on both sides. It seemed impossible to solve that without having some kind of specialization. This PR does also not really solve this problem, it just takes another way to implement all combinations of `AppearsInFromClause`:
* It reuses the `allow_tables_in_same_query!` macro by extending the generated code there to also contain cases for aliased tables. This allows to mix aliases of different tables with other tables as long as you can use the tables together. This requires the recently stabilized `re-rebalance-coherence` feature, that is currently in the rust beta channel.
* Unfortunately it is not possible to implement `AppearsInFromClause` there in such a way that would allow to use two aliases of different tables together, because of rusts coherence rules (all types are not fundamental and not from the current crate). This is solved by the `AliasNotEqualHelper` trait which just moves this job to something (a table) for which we could implement traits there and a generic implementation of `AppearsInFormClause` for all alias alias combination that internally uses the concrete implementations of `AliasNotEqualHelper`
* We need to provide a implementation of `AppearsInFromClause` for the alias with itself. This again uses the `AliasNotEqualHelper`, but now inside the alias macro. This prevents that a alias is used more than once in a given query.
* As last open point we need to find some way to allow multiple aliases for the same table to be used in the same query. This requires another implementation of `AppearsInFromClause` for cases where two aliases share the same table, but are different. This impl would conflict again with already implemented ones, so I choose a hack to workaround this one. By declaring multiple aliases together in a single `alias!` macro call we are able to brute force this impls by reusing `AliasNotEqualHelper` again.

Anything else in this PR is basically just implementing already existing stuff also for aliases to write some simple test cases. The tests just check if the query compiles and is successfully executed.

# Open questions

* Which features did I miss?
* Did I miss an important usecase?
* Are there any unnecessary restrictions in this API?
* Is this API sound?
* To we even want to have a API in this shape?

# Things needed to be fixed for a final PR

- [x] Correctly implement query dsl support for aliases (it requires currently that you call `as_query()` explicitly in most cases)
- [x] revisit code organisation (suggestions welcome)
- [ ] discuss naming
- [x] documentation
- [x] more and better tests


diesel/src/query_source/alias.rs
```diff
@@ -0,0 +1,374 @@
...
+                type Count = $crate::query_source::Never;
+            }
+
+            impl $crate::query_source::AliasNotEqualHelper<$right_table::table, $left_alias, $right_alias> for $left_table::table {
+                type Count = $crate::query_source::Never;
+            }
+        }}
+
+        __internal_alias_helper!($left_table as $left_alias, $($table as $alias,)*);
+        __internal_alias_helper!($right_alias as $right_alias, $($table as $alias,)*);
```
weiznich:
```suggestion
        __internal_alias_helper!($right_table as $right_alias, $($table as $alias,)*);
```

dessalines:
Assuming these conflicts are solved, is this working? There hasn't been movement on this in a year but if it works I'd love to use it.

weiznich:
As there seems to be some interest in getting this feature somehow into diesel, I will try to writeup what I expect from a complete implementation. To start of with something: This PR is merely a demonstration of the underlying concept extracted from some of my own personal projects and somehow integrated into diesel. 
To finish this implementation at least the following things are required:
- [ ] Rebasing this PR to the current diesel master branch
- [ ] Correctly implement all `QueryDsl` traits that make sense for aliases (That's already started with at least some of the traits)
- [ ] Review the test cases + maybe add more
- [ ] Review the compile tests + maybe add more
- [ ] Review the code structure + maybe move things into the right place
- [ ] Write quite a lot of documentation how to use this feature
- [ ] Have a look about other cases where aliases can occur, like expression aliases and aliases for select statements. Before merging this we should at least have and idea if it's possible to support such cases with the same/a similar implementation or if that's impossible. This likely will require quite a lot of experimentation.
- [ ] Have a look about how to write types (for example as return types) for queries containing an alias

So if anyone is interested feel free to take this PR as basis to work on. If you need some pointers feel free to ping me in our gitter channel.

cc @dessalines

Ten0:
I have time to work on this.

I believe this is the correct design :)

I've updated this to the current master (or almost) at https://github.com/Ten0/diesel/tree/feature/aliasing (tests pass again).

If you can allow edits:
allow-maintainers-to-make-edits-sidebar-checkbox
I can probably push that work here :)

weiznich:
@Ten0 Thanks for working on this :+1: Edits are already allowed 🙈

Ten0:
> Edits are already allowed :see_no_evil:

Ah, then it looks like my permissions don't allow me to do this. (Maybe I have merge but not push or something)

Should I open a separate draft then?

weiznich:
@Ten0 I've added you as Collaborator to my diesel fork, that should allow you to push to this branch.

diesel/src/query_source/alias.rs
Ten0: 09d166c09f068374e548973c7aa77a33b2e8bf14

Ten0:
So I've tried some changes and I'm pretty happy with them:
- Instead of having `Alias<T, F>` where `T` is the table and `F` represents the name (which requires two generic parameters in most places and allows the incorrect representation `Alias<T2, F1>`), turn that into a single generic parameter that implements an AliasSource trait that provides both the table and the alias name.
- Reduce the amount of code generated in macros:
  * `allow_tables_to_appear_in_same_query!` now implements a single new trait: `TableNotEqual<T2> for T1` for all pairs of distinct tables. This impl is then used within the `diesel` crate to generate the required implementations of `AppearsInFromClause` for both tables and aliases when tables are distinct. Implementations where tables are *not* distinct are still implemented by the `table!` macro (otherwise would conflict with previous impl because a downstream crate could `impl TableNotEqual<T> for T`), but that is O(n) generated code instead of O(n²) generated code.
  * Alias joins are implemented mostly within Diesel (rest generically by the `table!` macro). This is done by extending `FieldAliasMapper` so that it can convert a lot more expressions than just directly tables, leaving those that are from other tables untouched. This allows us to convert the join on clauses to their aliased counterparts in generic code.
    (Although pretty minor, this will also enable the user to convert several fields at a time in composite structures, e.g.:
    ```rust
    let posts = users::table.inner_join(posts::table.on(...).inner_join(users_alias))
    .select(users_alias.fields((posts::title, posts::body, users::username, users::last_seen)))
    .load::<Post>(db)?;
    ```
    )

Ticki84:
@Ten0 Thanks for your work on this feature!

It seems that making more than 3 aliases to the same table in the same alias! macro causes an implementation conflict:

```
let (u_1, u_2, u_3, u_4) =
            alias!(users as u_1, users as u_2, users as u_3, users as u_4);

conflicting implementations of trait `diesel::query_source::AliasAliasAppearsInFromClause<schema::users::table, data_access::u_4, data_access::u_3>` for type `schema::users::table`
    --> src/data_access.rs:1070:13
     |
1070 |             alias!(users as u_1, users as u_2, users as u_3, users as u_4);
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |             |
     |             first implementation here
     |             conflicting implementation for `schema::users::table`
     |
     = note: this error originates in the macro `__static_cond` (in Nightly builds, run with -Z macro-backtrace for more info)

```

Ten0:
> @Ten0 Thanks for your work on this feature!
> 
> It seems that making more than 3 aliases to the same table in the same alias! macro causes an implementation conflict:

Hahah that's so funny! I think know where the issue comes from. (https://github.com/diesel-rs/diesel/pull/2254/commits/ba470e4c5b22240f2616fbd7154b72e27406ef94).
It's crazy that you found it so fast! Thanks!

Ten0:
@Ticki84 this should be fixed by 76e4530e4a

diesel/src/query_source/alias.rs
```diff
@@ -0,0 +1,637 @@
...
+/// In order to be able to implement within diesel a lot of traits on what will represent the alias,
+/// we cannot use directly that new type within the query builder. Instead, we will use `Alias<S>`,
+/// where `S: AliasSource`.
+///
+/// This trait should never be implemented by an end-user directly.
+pub trait AliasSource: Copy + Default {
+    /// The name of this alias in the query
+    const NAME: &'static str;
+    /// The table it maps to
+    type Table: Table;
```
weiznich, line 32: I'm not sure if we really need to require table here. Yes for the currently implemented use case that always will be a table. But in the future we may want to add other kinds of alias, where this could be either a select statement or a column (not sure if we really need to add the second one someday)

weiznich (commented):
Thanks for the great update :+1:
I had no time for a in-depth review yet, but here are a few comments from going quickly over the code.

The other thing we need to talk about is the timeline. Do we want to get that into the 2.0 release or is it ok to skip this later. At least in my opinion the 2.0 release is nearly finished. I would like to include merge #2935 and #2996. After that I would like to release a first rc version.

diesel/src/query_source/alias.rs
```diff
@@ -0,0 +1,637 @@
...
+}
+
+impl<S, C> QueryId for AliasedField<S, C>
+where
+    S: AliasSource + 'static,
+    S::Table: 'static,
+    C: Column<Table = S::Table> + 'static,
+{
+    type QueryId = Self;
+    const HAS_STATIC_QUERY_ID: bool = true;
```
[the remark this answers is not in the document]
Ten0: If a different name is used for the alias, (so different `S`) wouldn't `QueryId=Self` differ by `S`? (as `Self = AliasedField<S,C>`)

Ten0:
> Do we want to get that into the 2.0 release or is it ok to skip this later.

I think it's fine to get `Alias` in a bit later if we don't manage to close [the open topics](https://github.com/diesel-rs/diesel/pull/2254#issuecomment-745502853) before the rest is ready, however don't we want to get the `allow_tables_to_appear_in_same_query!`/`TableNotEqual` change in there so we don't risk doing anything breaking related to this after?

diesel/src/query_source/alias.rs
```diff
@@ -0,0 +1,637 @@
...
+pub struct AliasedField<S, F> {
+    _alias_source: S,
+    _field: F,
+}
+
+impl<S> QueryId for Alias<S> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
```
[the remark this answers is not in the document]
weiznich: You are right. I've somehow missed that `AliasSource` is a trait and we have a different type for each alias. 🙈 
That implies this should be the correct impl:
```suggestion
impl<S> QueryId for Alias<S> {
    type QueryId = Self;

    const HAS_STATIC_QUERY_ID: bool = true;
}
```

weiznich:
> , however don't we want to get the allow_tables_to_appear_in_same_query!/TableNotEqual change in there so we don't risk doing anything breaking related to this after?

Adding this later shouldn't be a breaking change as there is no old code/behaviour that relies on this. If we add that in an hypothetical 2.1 release that would be fine, because it adds just now functionality + new code

diesel/src/query_source/alias.rs
Ten0: bd0269628a8afae5e7cacf13d4c30247412d3282

Ten0:
> Have a look about other cases where aliases can occur, like expression aliases and aliases for select statements

I've experimented with this a bit and found it would actually be pretty hard to support within the query builder. (Parentheses are not neutral, `SelectableExpression !: Expression`, we can't reference fields of an aliased subselect because their name is not typed)...

Given this I think it's reasonable to ignore this for now, and either:
 - Have a separate structure to handle this later
 - Have that in Diesel 3

weiznich:
@Ten0 Thanks for the summary ❤️ . I will try to have a look hopefully later this week. If someone else from @diesel-rs/reviewers  want to review this sooner feel free to do so.

diesel/src/query_source/aliasing/alias.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl<S> QueryId for Alias<S>
+where
+    Self: 'static,
+    S: AliasSource,
+    S::Table: Table,
+{
+    type QueryId = Self;
+    const HAS_STATIC_QUERY_ID: bool = true;
```
weiznich, line 56: That should be depended on `AliasSource::Table`. If that has a static query id, that's here the case as well. If not than the query id of the whole alias is not stable.

diesel/src/query_source/aliasing/alias.rs
```diff
@@ -0,0 +1,171 @@
...
+where
+    T1: TableNotEqual<T2> + Table,
+    T2: Table,
+    S1: AliasSource<Table = T1>,
+    S2: AliasSource<Table = T2>,
+{
+    type Count = Never;
+}
+
+impl<S> AppearsInFromClause<Alias<S>> for ()
```
weiznich, line 166: This impl shoud not be needed anymore, as we don't allow `()` as query source anymore.

diesel/src/query_source/aliasing/aliased_field.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl<S, C> QueryId for AliasedField<S, C>
+where
+    S: AliasSource + 'static,
+    S::Table: 'static,
+    C: Column<Table = S::Table> + 'static,
+{
+    type QueryId = Self;
+    const HAS_STATIC_QUERY_ID: bool = true;
```
weiznich, line 31: Again, this has only a static query id if the inner column has a static query id.

diesel/src/query_source/aliasing/aliased_field.rs
```diff
@@ -0,0 +1,108 @@
...
+where
+    S: AliasSource,
+    C: Column<Table = S::Table>,
+    Self: ExpressionMethods,
+    <Self as Expression>::SqlType: sql_types::SqlType,
+    T: AsExpression<<Self as Expression>::SqlType>,
+    dsl::Eq<Self, T>: Expression<SqlType = sql_types::Bool>,
+{
+    type Output = dsl::Eq<Self, T>;
+    fn eq_all(self, rhs: T) -> Self::Output {
```
weiznich, line 105:
```suggestion
    type Output = dsl::Eq<Self, T>;
    
    fn eq_all(self, rhs: T) -> Self::Output {
```

diesel/src/query_source/aliasing/dsl_impls.rs
```diff
@@ -0,0 +1,269 @@
...
+    S::Table: Table + HasTable<Table = S::Table>,
+    <S::Table as Table>::PrimaryKey: FieldAliasMapper<S>,
+    <<S::Table as Table>::PrimaryKey as FieldAliasMapper<S>>::Out: EqAll<PK>,
+    Self: FilterDsl<
+        <<<S::Table as Table>::PrimaryKey as FieldAliasMapper<S>>::Out as EqAll<PK>>::Output,
+    >,
+{
+    type Output = dsl::Filter<
+        Self,
+        <<<S::Table as Table>::PrimaryKey as FieldAliasMapper<S>>::Out as EqAll<PK>>::Output,
```
weiznich, line 53: Looks like we should add a type alias to `diesel::dsl` for `<T as FieldAliasMapper<S>>::Out` and use that one here.

diesel/src/query_source/aliasing/field_alias_mapper.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn map(column: C, alias: &Alias<S>) -> Self::Out;
+}
+impl<S, C> FieldAliasMapper<S> for C
+where
+    S: AliasSource,
+    C: Column,
+    S::Table: FieldAliasMapperAssociatedTypesDisjointnessTrick<C::Table, S, C>,
+{
+    type Out = <S::Table as FieldAliasMapperAssociatedTypesDisjointnessTrick<C::Table, S, C>>::Out;
+    fn map(self, alias: &Alias<S>) -> Self::Out {
```
weiznich, line 42:
```suggestion
    type Out = <S::Table as FieldAliasMapperAssociatedTypesDisjointnessTrick<C::Table, S, C>>::Out;
    
    fn map(self, alias: &Alias<S>) -> Self::Out {
```

diesel/src/query_source/aliasing/field_alias_mapper.rs
```diff
@@ -0,0 +1,121 @@
...
+diesel_derives::__diesel_for_each_tuple!(field_alias_mapper);
+
+// The following `FieldAliasMapper` impls are useful for the generic join implementations.
+// More may be added.
+impl<SPrev, SNew, F> FieldAliasMapper<SNew> for AliasedField<SPrev, F>
+where
+    SNew: AliasSource,
+{
+    type Out = Self;
+    fn map(self, _alias: &Alias<SNew>) -> Self::Out {
```
weiznich, line 97:
```suggestion
    type Out = Self;
    
    fn map(self, _alias: &Alias<SNew>) -> Self::Out {
```

diesel/src/query_source/aliasing/field_alias_mapper.rs
```diff
@@ -0,0 +1,121 @@
...
+        self
+    }
+}
+
+impl<S, F> FieldAliasMapper<S> for expression::nullable::Nullable<F>
+where
+    F: FieldAliasMapper<S>,
+{
+    type Out = expression::nullable::Nullable<<F as FieldAliasMapper<S>>::Out>;
+    fn map(self, alias: &Alias<S>) -> Self::Out {
```
weiznich, line 108:
```suggestion
    type Out = expression::nullable::Nullable<<F as FieldAliasMapper<S>>::Out>;
    
    fn map(self, alias: &Alias<S>) -> Self::Out {
```

diesel/src/query_source/aliasing/field_alias_mapper.rs
```diff
@@ -0,0 +1,121 @@
...
+        expression::nullable::Nullable::new(self.0.map(alias))
+    }
+}
+
+impl<S, F> FieldAliasMapper<S> for expression::grouped::Grouped<F>
+where
+    F: FieldAliasMapper<S>,
+{
+    type Out = expression::grouped::Grouped<<F as FieldAliasMapper<S>>::Out>;
+    fn map(self, alias: &Alias<S>) -> Self::Out {
```
weiznich, line 118:
```suggestion
    type Out = expression::grouped::Grouped<<F as FieldAliasMapper<S>>::Out>;
    
    fn map(self, alias: &Alias<S>) -> Self::Out {
```

diesel/src/query_source/aliasing/macros.rs
```diff
@@ -0,0 +1,128 @@
...
+/// ---------------------
+/// It may sometimes be useful to declare an alias at the module level, in such a way that the type
+/// of a query using it can be expressed (to not declare it anonymously).
+///
+/// This can be achieved in the following way
+/// ```rust
+/// # include!("../../doctest_setup.rs");
+/// use diesel::{query_source::Alias, dsl};
+///
+/// diesel::alias!(schema::users as users_alias: UsersAlias);
```
weiznich, line 37: I would prefer something like:

```rust
const USER_ALIAS: Alias<user_alias> = diesel::alias!(schema::users as user_alias);
``` 

here if possible. Making the declared identifier visible, instead of hiding it behind a macro. That makes it easier to search for the corresponding definition as it looks like any other rust code.

weiznich (commented):
I did not finish the whole PR yet, but I rather like to leave the comments I've already made. 
Thanks for the update :+1:

diesel/src/query_source/aliasing/alias.rs
Ten0: Fine for me to add that (I'll do it), but I thought that was already implied by the fact it was a `Table` (hence supposedly generated by the `table!` macro) -> why do we prefer writing it like this?

diesel/src/query_source/aliasing/aliased_field.rs
Ten0: Same as https://github.com/diesel-rs/diesel/pull/2254#discussion_r788717334

diesel/src/query_source/aliasing/alias.rs
weiznich: `diesel_dynamic_schema::Table` implements `Table` as well, but that does not have a static query id for obvious reasons.

https://github.com/diesel-rs/diesel/blob/master/diesel_dynamic_schema/src/table.rs#L75

diesel/src/query_source/aliasing/macros.rs
Ten0: I don't think that would be possible because that would make this macro an expression macro so the `user_alias` type would be scoped to inside the construction of that expression at best.

My goal was to make it feel like `table!` on use (I had to use a const for that instead of a unit struct because of the alias wrapper).

weiznich: If we could not make it an expression maco, I would like to suggest the same solution as for `sql_function!` here. Just use `diesel::alias!(const MY_ALIAS_IDENT: UserAlias = schema::users as UserAlias);` (or something slightly different as long as it matches the normal rust const definition) to define an alias at  module level.

diesel/src/query_source/aliasing/dsl_impls.rs
Ten0: So after having tried it in all places where `FieldAliasMapper` was used I feel like when used on generic types these typically make the link between the trait bounds `T: FieldAliasMapper<S>` and the other bounds or type constructions (`dsl::MappedFields<S, T>`) slightly less obvious hence harder to read, so I'd favor having the user use this alias with concrete types (so I've used it as signature of `Alias::fields`), but I haven't updated to using it for our internal trait bounds.

WDYT?

weiznich: Thats fine for me :+1:

diesel/src/query_source/aliasing/macros.rs
Ten0: So the writing would then probably be:
```rust
//                               const name             AliasSource                   alias name within the SQL query
diesel::alias!(const USERS_ALIAS: Alias<UsersAlias> = schema::users as users_alias);
```
I'm not sure the fact it's a `const` instead of e.g. a unit struct that can be instantiated implicitly is typically relevant to the end user - or what they would focus on.
I think in our codebase we'd favor the `table!`-like writing to the `fn`-like writing when declaring these at the module level.
@Elrendio wdyt? (vs ```rustdiesel::alias!(schema::users as users_alias: UsersAlias);`)

Ten0: Although maybe it would be relevant considering if you're declaring at the module level that might be because you need to express the types so making the magic not magic would help with that.

Ten0: For now I'm supporting both

Elrendio: Hi :wave:

After discussing it IRL with @Ten0 I'm in favor of `diesel::alias!(schema::users as users_alias: UsersAlias);` and even better `diesel::alias!(schema::users as users_alias);` when you don't need to reference the type.

# Arguments:
1. I find `diesel::alias!(schema::users as users_alias);` simplier than `diesel::alias!(const USERS_ALIAS: Alias<UsersAlias> = schema::users as users_alias);` as there's less boiler plate. Seems to me it will be simplier for developers to adopt the feature
2. I don't think developpers need to know the internals of the feature before using it. Most people @Stockly for example will just want to be able to reference an alias in a query. 
4. Having the same writing for when using an alias in a function and in a module feels simplier to me. Notably, it will enable to just move the line of code out of the function.
5. We can support both writings for people that want to deep dive a bit more 

Have a nice day,
Elrendio

Ten0: TODO if we keep the `const` writing we probably want to use `static_cond!` to prevent from writing:
```rust
diesel::alias! {
    const USERS_ALIAS: Alias<UsersAlias> = schema::users as users_alias;
    const USERS_ALIAS_2: Alias<UsersAlias2> = schema::users as users_alias;
}
```

weiznich:
I was quite busy with other things the last two weeks. Hopefully I'm able to have a deeper look at this tomorrow.

weiznich:
> I've experimented with this a bit (Ten0@107ae92) and found it would actually be pretty hard to support within the query builder. (Parentheses are not neutral, SelectableExpression !: Expression

I've played a bit with this and got that here working: https://github.com/weiznich/diesel/commit/2e23b87bb98a3c7ef0df3c77308b9e5d73305a7a This obviously needs some more work to figure out how to reference those aliases from somewhere else (group by, order, …) + to polish the interface.
That does not mean we need to support this in the initial version, just that we should name stuff so that this is possible (+ maybe add that `QueryFragment` impl for the generated `table` structs as this could be a potentially breaking change).

> we can't reference fields of an aliased subselect because their name is not typed, 

Give expression aliases we could enforce that users could only reference aliased expressions from the select clause of the inner alias query in the outer query. So something like this:

```rust
let max_user_id = alias!(max(users::id) as inner_user_id);
let user_query_alias = alias!(users::table.filter(…).select(user_id) as filtered_users);

posts::table.inner_join(user_query_alias).filter(max_user_id.gt(5))
```

Again that does not need to be implemented now, we just should look if there are no major blockers for this. (I know even that is not trivial…)

> table does not impl QueryFragment)

As written above, that's the simplest thing to fix 🙈  (we should probably just do that, it think that's purely for historic reasons that table does not implement `QueryFragement` directly...)

weiznich (commented):
Looks good modulo those two small points :tada:

weiznich:
One other thing I forgot to mention above: This misses a changelog entry :wink:

bitemyapp:
Pretty excited about this PR btw, table aliasing is one of the few things that forces me to write raw SQL queries. Thank you everyone for your hard work!

diesel/src/query_source/aliasing/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+/// In order to be able to implement within diesel a lot of traits on what will represent the alias,
+/// we cannot use directly that new type within the query builder. Instead, we will use `Alias<S>`,
+/// where `S: AliasSource`.
+///
+/// This trait should never be implemented by an end-user directly.
+pub trait AliasSource {
+    /// The name of this alias in the query
+    const NAME: &'static str;
+    /// The table it maps to
+    type Table;
```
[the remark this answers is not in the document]
Ten0: fefdca5315f0a9de906152c9d7b600605bb6c6fb

Ten0:
> > I've experimented with this a bit ([Ten0/diesel@107ae92](https://github.com/Ten0/diesel/commit/107ae92)) and found it would actually be pretty hard to support within the query builder. (Parentheses are not neutral, SelectableExpression !: Expression
> 
> I've played a bit with this and got that here working: [weiznich@2e23b87](https://github.com/weiznich/diesel/commit/2e23b87bb98a3c7ef0df3c77308b9e5d73305a7a) This obviously needs some more work to figure out how to reference those aliases from somewhere else (group by, order, …) + to polish the interface. That does not mean we need to support this in the initial version, just that we should name stuff so that this is possible (+ maybe add that `QueryFragment` impl for the generated `table` structs as this could be a potentially breaking change).

So the issue I had with this approach when getting to this point is the "Parentheses are not neutral, `SelectableExpression !: Expression`" issue, that is:
```sql
SELECT (("users"."name" || '!')) AS "name_alias" FROM "users";
```
is valid, but none of these are:
```sql
SELECT (("users"."name" || '!')) AS "name_alias" AS "name_alias_2" FROM "users";
SELECT ((("users"."name" || '!')) AS "name_alias") || '!' FROM "users";
SELECT "users"."name" FROM "users" WHERE "users"."name" AS "name_alias" = 'Sean';
-- or even
SELECT (("users"."name") AS "name_alias") FROM "users";
```
and with these trait bounds, all of these could be written.

(as branch: https://github.com/Ten0/diesel/commit/36292a98bf153b0f8b3565d82e8ccb30e3905b42)

So it looks like `impl<S> Expression for Alias<S>` is not really correct, and we would need to have a different trait for pseudo-expressions that can only be used at the top-level of a select statement. (I mistakenly confused this with `SelectableExpression` in my previous message but it looks like this trait currently doesn't exist).

In addition, methods such as `.field()`/`.fields()` don't really make sense... It would feel more natural to me to have a different type for expression aliases anyway (esp. since it's located in `query_source`).

However I think there's a reasonably high chance that we'd use this for named subselects eventually, and I also think `Table` would not be appropriate there, so I'm updating it to `Target`. (I think `Source` may be confusing since there's also `AliasSource` which is a different concept.)

weiznich (commented):
Looks ready now 👍 
So lets move on merging this and see how it goes. 
(That implies that I need to update my documentation/private item reform PR again. Hopefully we can land that one next)

weiznich:
CI is falling due to to the already merged clap fixes. So let's move on and merge this one as well :tada:

PR #2263, opened 2020-01-10
Update the test suite to the 2018 edition

sgrif:
Because `cargo fix --edition` also fixes any warnings, which would
remove a bunch of dead code if run for sqlite or mysql, I went ahead and
fixed the warnings for those backends figuring it'd take less time than
fixing the 2018 edition errors for those backends manually (spoiler: it
wasnt)

The warnings for those backends were 100% related to dead code, mostly
unused imports. For the most part, fixing these was just moving imports
where they were actually used, or cfging them. In one case I fixed the
unused imports by adding some missing tests for float
seralization/deserializion on SQLite and MySQL. This was a bit funky
since neither have a way to specify the type of a float literal, nor do
they have a way to specify NaN or Infinity (though SQLite will coerce an
out of range number to infinity). For the `schema_dsl` module, I stuck
the slightly odd `#[cfg_attr(not(feature = "postgres"),
allow(dead_code))]`. This is because going through that module and
annotating what schema creation methods are used by what backends just
isn't worth the time. We do want to warn if something is *actually*
unused though, so I'm assuming that if it's used at all it's used by at
least PG.

The benchmarks haven't built for ~2 years... They build again now.

### DIFF
--- a/diesel_tests/tests/schema.rs
+++ b/diesel_tests/tests/schema.rs
@@ -1,6 +1,4 @@
 use diesel::*;
-use dotenv::dotenv;
-use std::env;
 
 #[cfg(all(feature = "postgres", feature = "backend_specific_database_url"))]
 infer_schema!("dotenv:PG_DATABASE_URL");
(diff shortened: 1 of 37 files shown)


diesel_tests/tests/schema.rs
```diff
@@ -241,7 +237,7 @@ pub fn drop_table_cascade(connection: &TestConnection, table: &str) {
         .unwrap();
 }
 
-#[cfg(not(feature = "sqlite"))]
+#[cfg(feature = "postgres")]
```
weiznich: Is it right that this test isn't run for mysql any more?

sgrif: This isn't a test. It's a regular function that MySQL never calls

sgrif: This isn't a test. It's a regular function that MySQL never calls

PR #2267, opened 2020-01-14
Fix the CI

weiznich:
Caused by the release of url 2.1.1
See https://github.com/servo/rust-url/issues/579 for details


weiznich:
Something seems to be currently broken with azure pipelines:
```
E: Failed to fetch https://packages.microsoft.com/ubuntu/16.04/prod/dists/xenial/main/binary-amd64/Packages  Writing more data than expected (622228 > 606754)
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

That's obviously not related to this change, but a (hopefully) temporary issue with their apt mirror. I will try to retrigger the builds tomorrow, hopefully it's fixed then…

PR #2268, opened 2020-01-15
Remove support for old uuid versions

weiznich:
uuid 0.7.0 was released back in 2018 introducing breaking changes to
API's diesel is using internally. Because we exposed in our public api
we couldn't remove it back then.
The next release will be diesel 2.0, so a breaking change is now
possible. Give the time span since the release of the initial
compatible version I'm in favour of removing support for even older
version, as support for them comes with a lot of code duplication. (As
it turns out this support was already broken in our master branch
because we don't test it anymore...)


JohnTitor:
Hm, this still has a conflict and CI doesn't run :/

JohnTitor:
nightly failures are deprecation warnings, would you include the fix or follow-up later?

JohnTitor:
Ah, wait, it's just a warning, overflow occurs, and it's reported on rust-lang/rust.

weiznich:
Yes nightly failures are because of https://github.com/rust-lang/rust/issues/68264. So if someone approves the change I see no reason not to merge it yet.
(We should probably fix the warnings too, but that should be done in a separate PR.)

PR #2269, opened 2020-01-15
Experimental ci with github actions

weiznich:
If that actually works I'm proposing moving all of our ci/automation
to this service, because of:

* Integration with github seems to be vastly better
    + No need to click n times to go to logs
    + Rerun a failed job is easier
    + Maybe? Inline error messages in PR's
* Error messages are automatically extracted
* I think they have 20 free concurrent jobs, azure has only 10 (so
faster CI)

(Most of the config is copied from wundergraph)


JohnTitor:
`fail-fast: false` might be useful here.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions-rs/cargo@v1
+      with:
+        command: doc
+        args: --manifest-path diesel/Cargo.toml --features "postgres sqlite mysql extras"
+
+    - name: Publish documentation
+      if: success()
+      uses: JamesIves/github-pages-deploy-action@releases/v3
+      with:
+        ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
weiznich: We need to create this token somewhere.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,39 @@
...
+        BRANCH: gh-pages # The branch the action should deploy to.
+        FOLDER: target/doc # The folder the action should deploy.
```
weiznich: This means we need to change the documentation link on our web page

weiznich:
This seems to work now modulo the expected nightly failures, so it's ready to get comments.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,373 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v1
+
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-${{ matrix.backend }}-cargo-registry-${{ hashFiles('**/Cargo.toml') }}
```
JohnTitor: Just a question: the reason that we use `Cargo.toml` instead of `Cargo.lock` is to make the cache last longer?

weiznich: I've used the cargo.toml's because the cargo.lock file is not checked into the repository

JohnTitor: How about generating it on the CI using cargo command such as `cargo generate-lockfile`? I haven't tried it yet, so the CI time may increase, but I feel that the current cash has a little too long lifetime.

weiznich: I've had another thought about that: The problem is to generate the lock file you need to already have a cloned copy of cargos registry. This step caches the repository, so it would not be meaningful  to make that depended on downloading the index first, then build the lock file and then download the cached index again.
Yes this cache probably lives longer than in has to, but using a incomplete probably outdated cache should be faster than just rebuilding the index on each build. (We should probably measure that, so it's just blind guessing…)

JohnTitor: Alright, cache is just a side effect. Let's reconsider when we find a better way.

weiznich:
Just got a mail that says we need to change our azure pipeline config till March 23. (Have to update some runner image versions till then at least.)
I would say we should try to get this PR done till then, so we don't have to handle that other change.

JohnTitor:
If deploying blocks merge, we can skip it for now and make it follow-up work since Travis does that work, unrelated to pipelines.
@weiznich How about?

weiznich:
I would really avoid adding another CI without removing any of the existing ones. One solution would be to leave travis temporary in place and remove azure_pipelines. 🤔 

@diesel-rs/contributors  Any opinions here?

.github/workflows/doc.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - master
+      - '0.[0-9]+.x'
+      - '1.[0-9]+.x'
+
+
+name: Publish Docs
+
+jobs:
+  publish_docs:
```
martell: Would it be possible to guard the publish docs workflow with something like
`if ${GITHUB_REPOSITORY} == "diesel-rs/diesel"`
so that workflows continue to work with forks?

martell: I managed to do this in a fork, will have to add a PR here to test?

.github/workflows/doc.yml
martell: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

seems like it is under the repo settings if you are an admin?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,404 @@
...
+              --libexecdir=/usr/lib/x86_64-linux-gnu/sqlite3
+          sudo make
+          sudo make install
+          echo '::set-env name=SQLITE_DATABASE_URL::/tmp/test.db'
+
+      - name: Install mysql (Linux)
+        if: runner.os == 'Linux' && matrix.backend == 'mysql'
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install mysql-server libmysqlclient-dev
```
JohnTitor, line 103: I think this is related to the failures: https://github.blog/changelog/2020-02-21-github-actions-breaking-change-ubuntu-virtual-environments-will-no-longer-start-the-mysql-service-automatically/
So we should add the script from that blog to here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,404 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        rust: ["stable", "beta", "nightly"]
+        backend: ["postgres", "sqlite", "mysql"]
+        os: [ubuntu-18.04, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v1
```
JohnTitor, line 25: v2 has some performance-related improvements so I'd recommend using here.

```suggestion
        uses: actions/checkout@v2
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,404 @@
...
+
+      - name: Cache cargo registry
+        uses: actions/cache@v1
```
JohnTitor, line 28: actions/cache v2 is also released now.

```suggestion
        uses: actions/cache@v2
```

weiznich:
Related things that should be fixed before merging:

- [ ] [Badges](https://github.com/diesel-rs/diesel/blob/2b4340d882fbf94f5a83532b7f11dd95f7aa241f/diesel/Cargo.toml#L62-L64) section in `Cargo.toml` (Likely also for other crates)
- [ ] [Badges](https://github.com/diesel-rs/diesel/blob/2b4340d882fbf94f5a83532b7f11dd95f7aa241f/README.md) section in Readme.md
- [ ] Add a subpage behind the documentation link on the web page to list correctly the documentation for different versions
- [ ] Backport this pr at least to the 1.4.x branch

JohnTitor:
> Badges section in Cargo.toml (Likely also for other crates)

crates.io doesn't display CI-related badge now so we could just remove them.

weiznich:
@diesel-rs/reviewers Should be ready for a final review

.github/workflows/ci.yml
```diff
@@ -0,0 +1,408 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-${{ matrix.backend }}-cargo-registry-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Cache cargo index
+        uses: actions/cache@v2
+        with:
+          path: ~/.cargo/git
+          key: ${{ runner.os }}-${{ matrix.backend }}-cargo-index-${{ hashFiles('**/Cargo.toml') }}
```
JohnTitor, line 37: We could shrink like this if you'd like:
```
with:
  path: |
    ~/.cargo/registry
    ~/.cargo/git
  key: ${{ runner.os }}-${{ matrix.backend }}-cargo-${{ hashFiles('**/Cargo.toml') }}
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,53 @@
...
+        override: true
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install libmysqlclient-dev libsqlite3-dev libpq-dev
+    - name: Build documentation
+      uses: actions-rs/cargo@v1
+      with:
+        command: doc
+        args: --manifest-path diesel/Cargo.toml --features "postgres sqlite mysql extras"
```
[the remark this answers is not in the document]
weiznich: As we host the docs ourself I think we want to include dependencies here, because otherwise for example the docs for the derive macros will not show up correctly.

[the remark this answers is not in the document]
JohnTitor: Then I think `--workspace` is a better way, how about?

PR #2271, opened 2020-01-17
Remove deprecated `std::error::Error` methods

weiznich:


JohnTitor:
It seems we should also remove `description()` from tests like this:

https://github.com/diesel-rs/diesel/blob/bd349ce9d43de202176e754cedbb2555a76d0f4c/diesel/src/pg/types/uuid.rs#L58

diesel/src/pg/types/json.rs
```diff
@@ -76,14 +76,14 @@ fn some_json_from_sql() {
 fn bad_json_from_sql() {
     let uuid: Result<serde_json::Value, _> =
         FromSql::<sql_types::Json, Pg>::from_sql(Some(PgValue::for_test(b"boom")));
-    assert_eq!(uuid.unwrap_err().description(), "JSON error");
+    assert_eq!(uuid.unwrap_err().to_string(), "JSON error");
```
JohnTitor: This shows `serde_json`'s detail error so assertion failed. We should tweak the right value also.

diesel/src/pg/types/json.rs
```diff
@@ -108,15 +108,15 @@ fn some_jsonb_from_sql() {
 fn bad_jsonb_from_sql() {
     let uuid: Result<serde_json::Value, _> =
         FromSql::<sql_types::Jsonb, Pg>::from_sql(Some(PgValue::for_test(b"\x01boom")));
-    assert_eq!(uuid.unwrap_err().description(), "JSON error");
+    assert_eq!(uuid.unwrap_err().to_string(), "JSON error");
```
JohnTitor: Same here.

weiznich:
I would like to merge this soon as an updated serde_json version now returns another error message description (mentioning that `.decscription()`  is deprecated). This breaks the CI for us.

PR #2272, opened 2020-01-17
Add a `contains` method for postgres range types

### LINKED ISSUE #2210: PG ranges and `contains`
## Problem

It's currently not possible to use PG ranges in a meaningful way, because helper methods such as `@>`, or `.contains` in diesel are not defined.

For example, I would like to be able to do something like the following:

```
table! {
    extra (id) {
        id -> Int4,
        client -> Int4,
        range -> Range<Date>,
    }

    client (id) {
        id -> Int4,
        name -> Text,
    }
}

client
        .left_join(
            extra.on(other_table::dsl::client
                .eq(client_table::dsl::id)
                .and(other_table::dsl::range.contains(diesel::dsl::date(diesel::dsl::now))))
        )
```

the compiler complains that `.contains` is not implemented, and only suggests `diesel::PgArrayExpressionMethods`, which implies that there is no trait which implements these methods for PG ranges.

Currently the columns are implemented as separate `start_date` and `end_date`, but there is a requirement for no overlapping ranges, which is trivial with a PG EXCLUDE + DATERANGE and not so trivial with separate columns.

I think this is something I could make a start at -- or at least get `.contains` working, but it would be nice to have some guidance / feedback that this is indeed a missing feature and is something that should be worked on.

## Checklist

weiznich:
Fixes #2210

(I've just needed this method, so I can simply put it into upstream
because it's probably useful for others too)

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -2,7 +2,7 @@
 
 use super::operators::*;
 use crate::expression::{AsExpression, Expression};
-use crate::sql_types::{Array, Nullable, Text};
+use crate::sql_types::{Array, Nullable, Range, Text};
 
 /// PostgreSQL specific methods which are present on all expressions.
 pub trait PgExpressionMethods: Expression + Sized {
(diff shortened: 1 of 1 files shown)


a1ien:
How to allow nullable Range? 
I add 
```
impl<ST> RangeHelper for Nullable<Range<ST>> {
    type Inner = ST;
}
``` 
But how `impl<T, ST> PgRangeExpressionMethods for T` for nullable range?

weiznich:
@a1ien It's already implemented for `Nullable<Range<ST>>`, but not released yet.

PR #2275, opened 2020-01-22
Fix UnexpectedNullError when deserializing (NULL,NotNull) into Option

### LINKED ISSUE #2274: UnexpectedNullError when deserializing (Null,Boolean,Boolean) into Option<(a,b,c)>
### Versions

- **Rust:** 1.40.0
- **Diesel:** master (bd349ce9d43de202176e754cedbb2555a76d0f4c)
- **Database:** Postgresql
- **Operating System** Linux

### Feature Flags

- **diesel:** "postgres", "r2d2", "serde_json", "chrono"

## Problem Description
Hitting an unexpected UnexpectedNullError at runtime

```rust
    xxx::table
        .find(xxx_id)
        .inner_join(yyy::table)
        .left_join(zzz::table)
        .select((
            xxx::integer,
            (
                zzz::integer,
                zzz::nullable_integer.is_not_null(),
                yyy::nullable_integer
                    .is_not_null()
                    .or(zzz::nullable_enum.is_not_null()),
            )
                .nullable(),
        ))
        .get_result::<(Option<i32>, Option<(i32, bool, bool)>)>(db)?;
```
where postgres' tuple result of that query is `NULL, NULL, FALSE, FALSE`

`Ok((None, None))`

`Err(DeserializationError(UnexpectedNullError))`

Stack trace of where this error ends up being triggered is:
```
  14: diesel::type_impls::integers::<impl diesel::deserialize::FromSql<diesel::sql_types::Integer,DB> for i32>::from_sql
             at /home/ten/Projects/diesel/diesel/src/type_impls/integers.rs:47
  15: diesel::type_impls::primitives::foreign_impls::_::<impl diesel::deserialize::FromSqlRow<__ST,__DB> for i32>::build_from_row
             at /home/ten/Projects/diesel/diesel/src/type_impls/primitives.rs:28 <-- This probably comes from that I'm actually using an intermediate type but it's unrelated to our issue: same thing happens when using directly an i32
  16: diesel::type_impls::tuples::<impl diesel::deserialize::FromSqlRow<(SA, SB, SC, SD),__DB> for (A, B, C, D)>::build_from_row
             at ./src/lib.rs:1
  17: diesel::type_impls::option::<impl diesel::deserialize::FromSqlRow<diesel::sql_types::Nullable<ST>,DB> for core::option::Option<T>>::build_from_row
             at /home/ten/Projects/diesel/diesel/src/type_impls/option.rs:98
  18: diesel::type_impls::tuples::<impl diesel::deserialize::FromSqlRow<(SA, SB),__DB> for (A, B)>::build_from_row
             at ./src/lib.rs:1
  19: <diesel::pg::connection::cursor::Cursor<ST,T> as core::iter::traits::iterator::Iterator>::next
             at /home/ten/Projects/diesel/diesel/src/pg/connection/cursor.rs:41
```
(replaced the `return Err(UnexpectedNullError)` of the `not_none!` macro by a panic to get the stack trace so I hope this error would not have been caught and handled, but it's pretty consistent with why i'd expect this error to trigger so I think this is the actual issue.

## Checklist

## Solving

Digged into that stacktrace a bit, and found this:
https://github.com/diesel-rs/diesel/blob/bd349ce9d43de202176e754cedbb2555a76d0f4c/diesel/src/type_impls/option.rs#L84-L101

where in particular, `next_is_null` is documented as making sure that "the next `count` columns are **all** `NULL`", hence the  current behaviour.

Note: Just checking whether **any** of these columns is null is not satisfactory because we still want to deserialize `1, NULL` into `Option<i32, Option<i32>>` as `Some(1, None)` and not just `None`.

### Whether it should be solved

Clearly in my case whether these booleans are defined doesn't have any meaning if I couldn't join on the `zzz` table, so I'm very happy to not have access to them if just any of these fields is `NULL`, which is why I'm expecting this behaviour.

I can't think of a scenario where I'd prefer a runtime error: cases where I could unintentionally "forget" a field should be covered by the typing after the query: "Why can't I get the value of this field? Oh it's because it was packed in this `Option`...)

Ten0:
Resolves #2274

I'm not sure if this was originally a bug or not, but in any case as described on the issue, often the NotNull field (say, `zzz::some_nullable_thing.is_not_null()`) doesn't have any meaning if we couldn't left_join on the `zzz` table and select the NULL field (say, `zzz::id`).

This is why I'm very happy to not have access to the second field if the first field is NULL.

However I can't think of a scenario where I'd prefer a runtime error: cases where I could unintentionally "forget" a field should be covered by the typing after the query: "Why can't I get the value of this field? Oh it's because it was packed in this `Option`...)

This is why I beleive that this behaviour that ends up giving `Ok(None)` instead of `Err(DeserializationError(UnexpectedNullError))` is better.

~~Accorging to the tests this implementation seems to work properly, though I'm not sure why recovering from the error skips the proper amount of fields... Can anyone confirm that (and why) this indeed behaves as expected ? :)~~
EDIT: This indeed did not skip the proper amount of fields, and is now fixed.

Also open to any comments :)

### DIFF
--- a/diesel/src/row.rs
+++ b/diesel/src/row.rs
@@ -19,6 +19,12 @@ pub trait Row<DB: Backend> {
     /// would all return `None`.
     fn next_is_null(&self, count: usize) -> bool;
 
+    /// Returns the index of the current column (next to be returned by take, zero based)
+    ///
+    /// May be used to skip the right number of fields when recovering for an `UnexpectedNullError` when
+    /// deserializing an `Option`
+    fn column_index(&self) -> usize;
+
     /// Skips the next `count` columns. This method must be called if you are
     /// choosing not to call `take` as a result of `next_is_null` returning
     /// `true`.
--- a/diesel/src/mysql/connection/stmt/iterator.rs
+++ b/diesel/src/mysql/connection/stmt/iterator.rs
@@ -59,6 +59,10 @@ impl<'a> Row<Mysql> for MysqlRow<'a> {
     fn next_is_null(&self, count: usize) -> bool {
         (0..count).all(|i| self.binds.field_data(self.col_idx + i).is_none())
     }
+
+    fn column_index(&self) -> usize {
+        self.col_idx
+    }
 }
 
 pub struct NamedStatementIterator<'a> {
--- a/diesel/src/pg/connection/row.rs
+++ b/diesel/src/pg/connection/row.rs
@@ -31,6 +31,10 @@ impl<'a> Row<Pg> for PgRow<'a> {
     fn next_is_null(&self, count: usize) -> bool {
         (0..count).all(|i| self.db_result.is_null(self.row_idx, self.col_idx + i))
     }
+
+    fn column_index(&self) -> usize {
+        self.col_idx
+    }
 }
 
 pub struct PgNamedRow<'a> {
--- a/diesel/src/sqlite/connection/functions.rs
+++ b/diesel/src/sqlite/connection/functions.rs
@@ -30,7 +30,7 @@ where
     }
 
     conn.register_sql_function(fn_name, fields_needed, deterministic, move |conn, args| {
-        let mut row = FunctionRow { args };
+        let mut row = FunctionRow { args, col_idx: 0 };
         let args_row = Args::Row::build_from_row(&mut row).map_err(Error::DeserializationError)?;
         let args = Args::build(args_row);
 
@@ -55,11 +55,13 @@ where
 
 struct FunctionRow<'a> {
     args: &'a [*mut ffi::sqlite3_value],
+    col_idx: usize,
 }
 
 impl<'a> Row<Sqlite> for FunctionRow<'a> {
     fn take(&mut self) -> Option<&SqliteValue> {
         self.args.split_first().and_then(|(&first, rest)| {
+            self.col_idx += 1;
             self.args = rest;
             unsafe { SqliteValue::new(first) }
         })
@@ -69,5 +71,9 @@ impl<'a> Row<Sqlite> for FunctionRow<'a> {
         self.args[..count]
             .iter()
             .all(|&p| unsafe { SqliteValue::new(p) }.is_none())
+    }
+
+    fn column_index(&self) -> usize {
+        self.col_idx
     }
 }
--- a/diesel/src/sqlite/connection/sqlite_value.rs
+++ b/diesel/src/sqlite/connection/sqlite_value.rs
@@ -105,6 +105,10 @@ impl Row<Sqlite> for SqliteRow {
             tpe == ffi::SQLITE_NULL
         })
     }
+
+    fn column_index(&self) -> usize {
+        self.next_col_index as usize
+    }
 }
 
 pub struct SqliteNamedRow<'a> {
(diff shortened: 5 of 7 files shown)


weiznich:
Could you add a testcase that would trigger the bad behaviour to the test suite? Otherwise this looks fine :+1:

Ten0:
Hello,

I've added the test case (it indeed did not work where I expected it to not work) and fixed it :)

Comments (or merge ^^) welcome. :)

diesel/src/row.rs
```diff
@@ -19,6 +19,12 @@ pub trait Row<DB: Backend> {
     /// would all return `None`.
     fn next_is_null(&self, count: usize) -> bool;
 
+    /// Returns the index of the current column (next to be returned by take, zero based)
+    ///
+    /// May be used to skip the right number of fields when recovering for an `UnexpectedNullError` when
+    /// deserializing an `Option`
+    fn column_index(&self) -> usize;
```
weiznich: As this is a public API I would prefer not to add this method here, because it only has an internal meaning. That said I'm not sure what would be the right approach here.

Ten0: If `Row` has to be public for others to be able to connect new backends I feel like this is close to being the only alternative. Another way I see would be to make sure that when the error is thrown all the fields have been consumed, but that is useless for every other error case and that possibly means a lot of places where we would have to take care of this and possibly make a mistake (enums, structs with Queryable derive,...) .

Given the other functions that this trait requires, this one doesn't seem too absurd to me, especially since the field is already required by most impls.

Ten0:
Hello,

I keep hitting this as a runtime error regularly, and code that makes requests work is far from being as clear as the one this PR enables.

I understand this changes the API a bit so it requires a thorough examination.
Is there anything I can do to help this advance or should we just wait until more reviewers happen to pass by ?

weiznich:
I have currently rather limited bandwidth, so I was not able to think about a better API yet, so this needs to wait till I find some time or any of the other reviewers comments. In general we should try to unify the change to the `Row` trait with that one done in #2181

diesel/src/row.rs
Ten0: Since it is consistent with `advance`, the meaning is not only internal.

weiznich: I've not forget about this issue, it should definitively be fixed with the diesel 2.0 release, so I've added it the the corresponding mile stone. That said, unfortunately I was pretty occupied with other things in the last time, so I had not really the bandwidth to think much about that. It will be done, but takes probably some time.

Ten0:
Hello,

Just an up, as I keep hitting this error/having to write and maintain intermediate structures when fetching from the database due to this, and would really benefit from this moving forward.

weiznich:
There has nothing changed from my side. I do currently not have the bandwidth to work on so much things in parallel due to my real life engagements, a fix for this is scheduled to be included in a 2.0 release  (I see it as blocker for 2.0) and I want to see some experimentation about how to unify that with the changes required in #2182 (Especially about how it interacts with `FIELDS_NEEDED` for dynamic clauses where this is not a static member and how that could be integrated into each other)

Ten0:
This is eventually being solved in #2182 and there's probably a problem in the implementation of this PR that triggers when nullable booleans come into play (due to #104).
=> closing

PR #2276, opened 2020-01-23
Generalize creating a `BoxedWhereClause`

weiznich:
This commit generalizes the creating of a boxed where clause from an
empty where clause in such a way that this works for any lifetime
instead only for static lifetimes. This is required to write
`dsl::IntoBoxed<some::table>` for general queries that may contain a
restriction on their lifetime. (For example because of they contain a
`.eq_any(some_borrowed_array)` expression)

### DIFF
--- a/diesel/src/query_builder/where_clause.rs
+++ b/diesel/src/query_builder/where_clause.rs
@@ -59,8 +59,8 @@ where
     }
 }
 
-impl<DB> Into<BoxedWhereClause<'static, DB>> for NoWhereClause {
-    fn into(self) -> BoxedWhereClause<'static, DB> {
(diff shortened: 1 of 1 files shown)


weiznich:
@Razican Thanks for the review. Would you be interested to join the diesel reviewer team. We are actively looking for people there. You don't need to know any details about diesel internals. You would be pinged on every newly opened PR. When there is anything unclear for some submitted code it's totally fine to ask how something works there. We will then try to explain it as good as possible. I would say it's one of the best ways to actual get into diesels code base.

Razican:
@weiznich Thanks for the idea! I'm already in the reviewers team if I'm not mistaken (I receive the notifications) but I try to only reply to those where I more or less understand what's happening, in order to not mess things up :)

PR #2278, opened 2020-01-24
generate standard ops impls for Unsigned<T> types

WebeWizard:
needed for mysql updates using unsigned types.
`      .set(my_unsigned_col.eq(my_unsigned_col + 1))
`


diesel/src/macros/ops.rs
```diff
@@ -78,6 +80,7 @@ macro_rules! date_time_expr {
 #[doc(hidden)]
 macro_rules! __diesel_generate_ops_impls_if_date_time {
     ($column_name:ident, Nullable<$($inner:tt)::*>) => { __diesel_generate_ops_impls_if_date_time!($column_name, $($inner)::*); };
+    ($column_name:ident, Unsigned<$($inner:tt)::*>) => { __diesel_generate_ops_impls_if_date_time!($column_name, $($inner)::*); };
```
weiznich: There are no unsigned date/time types in diesel, therefore we should not add this here.
```suggestion
```

WebeWizard:
should be good to go for both mysql and sqlite. thanks

migrations/mysql/20170209180355_add_one_off_tables_from_integration_tests/up.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE numbers (n INTEGER PRIMARY KEY);
 CREATE TABLE precision_numbers (n DOUBLE PRECISION NOT NULL PRIMARY KEY);
 CREATE TABLE nullable_doubles (id INT PRIMARY KEY AUTO_INCREMENT, n DOUBLE PRECISION);
-CREATE TABLE users_with_name_pk (name VARCHAR(255) PRIMARY KEY);
+CREATE TABLE users_with_name_pk (name VARCHAR(50) PRIMARY KEY);
```
weiznich: Could you explain why this change is required? It seems unrelated to the other changes.

diesel_tests/tests/expressions/ops.rs
```diff
@@ -199,3 +199,75 @@ fn precedence_with_parens_is_maintained() {
...
+        .unwrap();
+
+    let expected_data = vec![1, 2];
+    let data = unsigned_table.select(value / 1).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data = vec![1, 1];
+    let data = unsigned_table.select(value / value).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+}
```
weiznich: Can we add another test checking that operations with two unsigned columns just work fine?

diesel_tests/tests/expressions/ops.rs
```diff
@@ -199,3 +199,75 @@ fn precedence_with_parens_is_maintained() {
     let x = select((2.into_sql::<Integer>() + 3) * 4).get_result::<i32>(&connection());
     assert_eq!(Ok(20), x);
 }
+
+#[test]
+#[cfg(not(feature = "postgres"))]
```
weiznich: That should only be tested for mysql. Sqlite has no support for unsigned integers. (Or better, it accepts that you write `unsigned integer`, but will use just a normal integer instead).
```suggestion
#[cfg(feature = "mysql")]
```

diesel_tests/tests/expressions/ops.rs
```diff
@@ -199,3 +199,75 @@ fn precedence_with_parens_is_maintained() {
...
+    let data = unsigned_table.select(value + 1).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data = vec![2, 4];
+    let data = unsigned_table.select(value + value).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+}
+
+#[test]
+#[cfg(not(feature = "postgres"))]
```
weiznich:
```suggestion
#[cfg(feature = "mysql")]
```

diesel_tests/tests/expressions/ops.rs
```diff
@@ -199,3 +199,75 @@ fn precedence_with_parens_is_maintained() {
...
+    let data = unsigned_table.select(value - 1).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data = vec![0, 0];
+    let data = unsigned_table.select(value - value).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+}
+
+#[test]
+#[cfg(not(feature = "postgres"))]
```
weiznich:
```suggestion
#[cfg(feature = "mysql")]
```

diesel_tests/tests/expressions/ops.rs
```diff
@@ -199,3 +199,75 @@ fn precedence_with_parens_is_maintained() {
...
+    let data = unsigned_table.select(value * 1).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+
+    let expected_data = vec![1, 4];
+    let data = unsigned_table.select(value * value).load(&connection);
+    assert_eq!(Ok(expected_data), data);
+}
+
+#[test]
+#[cfg(not(feature = "postgres"))]
```
weiznich:
```suggestion
#[cfg(feature = "mysql")]
```

migrations/mysql/20170209180355_add_one_off_tables_from_integration_tests/up.sql
WebeWizard: It has something to do with the version of MariaDB I'm using to test with (and may be mysql version 5.6).  With length 255 the migration would fail with this error https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes.  From looking at the diesel tests, we only need a few characters, and not 255.  I thought this would help other users testing on non-dockerized mysql.

diesel_tests/tests/expressions/ops.rs
WebeWizard: Although sqlite does not rely on the Unsigned<T> type, I suggest that we keep these tests turned on for sqlite as the tests do verify that sqlite unsigned columns continue to work as you described above (currently not covered by any tests)

weiznich: Just to clarify that: Sqlite does handle `UNSIGNED INTEGER` just the same way as `INTEGER` (or every other type that contains `INT`). Also diesel does infer these types as `Integer` not as `Unsigned<Integer>`. That means for sqlite this test would be just a duplicate of the existing tests above. That would be confusing, because the tests would be named `unsigned_*` but test something differently.

diesel_tests/tests/expressions/ops.rs
WebeWizard: Sure thing.  I turned "id" column into an unsigned integer, and then add a test for "value" / "id"

diesel_tests/tests/expressions/ops.rs
WebeWizard: Done.  I removed the sqlite unsigned migration and set feature = mysql on all of the tests.

Issue #2289, opened 2020-02-04
Update libsqlite3-sys to 0.18


tyrylu:
Hello.
Would it be possible to bump libsqlite3-sys to 0.18? In one of my projects i would like to use the bundled sqlite3, but the one bundled with 0.17 crashes with some extensions in a specific circumstance.

weiznich:
We would be happy to receive a PR bumping the version constrains.

JADSN:
# Update libsqlite3-sys to 0.20.1

Using the depencies:

- r2d2 = '0.8.9'
- r2d2_sqlite = '0.17.0'

```
package `libsqlite3-sys v0.9.1`
    ... which is depended on by `diesel v1.4.5`
links to native library `sqlite3`

package `libsqlite3-sys v0.20.1`
    ... which is depended on by `rusqlite v0.24.1`
    ... which is depended on by `r2d2_sqlite v0.17.0`
also links to native library `sqlite3`
```

### I tried:

- `cargo update -p libsqlite3-sys:0.9.1 --precise=0.20.1`

### Maybe a solution:
[Disel Cargo.toml](https://github.com/diesel-rs/diesel/blob/master/diesel/Cargo.toml)

- Bump version from "0.20.0" to "0.20.1"

**Thanks**

weiznich:
@JADSN as written above: We are happy to recieve a PR for that. Otherwise that's currently not on top of my priority list. I will likely bump supported versions of our dependencies shortly before the next release (without ETA) if requiered.

PR #2292, opened 2020-02-08
feat: bump to libsqlite3-sys 0.17, no other changes

FauxFaux:
New `libsqlite3-sys` is out, released 2019-12-13 and has a patch release.

Everything seems to work on linux.

Most of the changes seem to be around the `bundled` feature, which we don't expose (yet?): https://github.com/jgallagher/rusqlite/commits/master/libsqlite3-sys


weiznich:
Thanks :+1:

Fixes #2289

PR #2293, opened 2020-02-08
Remove references to infer_schema macro from the docs

### LINKED ISSUE #2281: Docs: infer_schema is deprecated
[This line](https://github.com/diesel-rs/diesel/blob/master/diesel/src/lib.rs#L18) should be updated to reflect the deprecation of `infer_schema` [as of 1.3.0](https://github.com/diesel-rs/diesel/blob/v1.3.0/CHANGELOG.md#deprecated).

blasrodri:
Fixes #2281

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -891,8 +891,8 @@ macro_rules! __diesel_table_query_source_impl {
 /// The generated `ON` clause will always join to the primary key of the parent
 /// table. This macro removes the need to call [`.on`] explicitly, you will
 /// still need to invoke [`allow_tables_to_appear_in_same_query!`] for these two tables to
-/// be able to use the resulting query, unless you are using `infer_schema!` or
-/// `diesel print-schema` which will generate it for you.
+/// be able to use the resulting query, unless you are using `diesel print-schema`
+/// which will generate it for you.
 ///
 /// If you are using `infer_schema!` or `diesel print-schema`, an invocation of
 /// this macro will be generated for every foreign key in your database unless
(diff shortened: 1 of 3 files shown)


diesel/src/macros/mod.rs
```diff
@@ -891,8 +891,8 @@ macro_rules! __diesel_table_query_source_impl {
 /// The generated `ON` clause will always join to the primary key of the parent
 /// table. This macro removes the need to call [`.on`] explicitly, you will
 /// still need to invoke [`allow_tables_to_appear_in_same_query!`] for these two tables to
-/// be able to use the resulting query, unless you are using `infer_schema!` or
-/// `diesel print-schema` which will generate it for you.
+/// be able to use the resulting query, unless you are using `diesel print-schema`
```
jdlambert: A comma seems appropriate here, before the "which":

> ...unless you are using `diesel print-schema`, which will generate it for you.

diesel/src/macros/mod.rs
```diff
@@ -891,8 +891,8 @@ macro_rules! __diesel_table_query_source_impl {
...
+/// which will generate it for you.
```
jdlambert: It's not immediately apparent to me what this "it" refers to.

blasrodri: I agree. And it's not even clear to me either.

weiznich: It refers to `allow_tables_to_appear_in_same_query!.

blasrodri: @weiznich Should I rephrase it, or leave it as it is then?

weiznich: If you have an idea how to rephrase it, then yes. Otherwise it is not required for merging this pr.

PR #2295, opened 2020-02-10
Add note clarifying that placeholder syntax is engine-dependent

tmcw:
I got a little lost in this example because it wasn't clear that the example was using SQLite as a backend, or that the query syntax (`?` placeholders) would be different for PostgreSQL. In hindsight this was obvious, but I figure it'd potentially save a future-me if the documentation noted up-front that this query syntax varied depending on the database type.

### DIFF
--- a/diesel/src/query_builder/sql_query.rs
+++ b/diesel/src/query_builder/sql_query.rs
@@ -28,7 +28,10 @@ impl<Inner> SqlQuery<Inner> {
         SqlQuery { inner, query }
     }
 
-    /// Bind a value for use with this SQL query.
+    /// Bind a value for use with this SQL query. The given query should have
+    /// placeholders that vary based on the database type,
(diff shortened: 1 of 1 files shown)


diesel/src/query_builder/sql_query.rs
```diff
@@ -28,7 +28,10 @@ impl<Inner> SqlQuery<Inner> {
         SqlQuery { inner, query }
     }
 
-    /// Bind a value for use with this SQL query.
+    /// Bind a value for use with this SQL query. The given query should have
+    /// placeholders that vary based on the database type,
+    /// like [SQLite Parameters](https://sqlite.org/lang_expr.html#varparam) or
+    /// the [PostgreSQL PREPARE syntax](https://www.postgresql.org/docs/current/sql-prepare.html).
```
weiznich: Could you add a reference for the mysql bind syntax here as well?

tmcw: Yep, done!

PR #2304, opened 2020-02-17
Additional `DatabaseErrorKind` variants

PhilipTrauner:
Introduces additional [`DatabaseErrorKind`](https://github.com/PhilipTrauner/diesel/blob/5d7cfb69ca8251db7447898341029c3d314db4de/diesel/src/result.rs#L87) variants (`NotNullViolation`, `CheckViolation`), their respective Postgres and MySQL error codes, and extends the error matching function of each supported database.

### DIFF
--- a/diesel_tests/tests/errors.rs
+++ b/diesel_tests/tests/errors.rs
@@ -1,5 +1,7 @@
 use crate::schema::*;
-use diesel::result::DatabaseErrorKind::{ForeignKeyViolation, UniqueViolation};
+use diesel::result::DatabaseErrorKind::{
+    CheckViolation, ForeignKeyViolation, NotNullViolation, UniqueViolation,
+};
 use diesel::result::Error::DatabaseError;
 use diesel::*;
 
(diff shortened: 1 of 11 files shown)


weiznich:
We should have tests for this similar to [this one](https://github.com/diesel-rs/diesel/blob/c84388b803af6c92d29c0801161c07345669a2e6/diesel_tests/tests/errors.rs#L7-L18)

PhilipTrauner:
Will follow up with tests shortly, just been a bit busy 🙂

diesel_tests/tests/errors.rs
```diff
@@ -189,3 +191,84 @@ fn read_only_errors_are_detected() {
...
+        }
+        _ => panic!(
+            "{:?} did not match Err(DatabaseError(NotNullViolation, e))",
+            failure
+        ),
+    };
+}
+
+#[test]
+fn check_constraints_constraints_are_detected() {
```
weiznich: I think there is a constraints to much here

PhilipTrauner:
MySQL only started [enforcing check constraints](https://stackoverflow.com/questions/2115497/check-constraint-in-mysql-is-not-working) quite recently. Most of the CI targets are still running MySQL 5.7, with the notable exception being the macOS target (which is running 8.0). Should I upgrade the other targets ? (@weiznich)

weiznich:
I would prefer to run the test only if the mysql version actual supports that feature. Otherwise this could be a problem for possible new contributors running the test locally against an older mysql version. If there is no simple way to do that it should be fine to test this feature only on sqlite and postgres. In this case please add a comment why the test is disabled for mysql .

PhilipTrauner:
> If there is no simple way to do that it should be fine to test this feature only on sqlite and postgres.
I didn't feel comfortable adding version detection logic to a particular test, therefore I decided to disable it instead.

PR #2305, opened 2020-02-18
Some small refactorings

jplatte:
I first thought I couldn't remove `use diesel;` because some crates didn't have `edition = "2018"` and started working on that. Then I realized the code the macros generate have to work in the 2015 edition so I probably can't remove the redundant import. (I still don't understand why things break *inside diesel* when I remove it, but that's not that important)

If you want me to split this PR or remove the first commit altogether, I'm fine with that.

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -8,6 +8,12 @@ pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
     quote! {
         #[allow(unused_imports)]
         const _: () = {
+            // This import is not actually redundant. When using diesel_derives
+            // inside of diesel, `diesel` doesn't exist as an extern crate, and
+            // to work around that it contains a private
+            // `mod diesel { pub use super::*; }` that this import will then
+            // refer to. In all other cases, this imports refers to the extern
+            // crate diesel.
             use diesel;
 
             #item
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -52,7 +52,7 @@ fn option_ty_arg(ty: &Type) -> Option<&Type> {
 pub fn ty_for_foreign_derive(item: &DeriveInput, flags: &MetaItem) -> Result<Type, Diagnostic> {
     if flags.has_flag("foreign_derive") {
         match item.data {
-            Data::Struct(ref body) => match body.fields.iter().nth(0) {
+            Data::Struct(ref body) => match body.fields.iter().next() {
                 Some(field) => Ok(field.ty.clone()),
                 None => Err(flags
                     .span()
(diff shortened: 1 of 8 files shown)


weiznich:
I would prefer to have the 2018 edition change in a different PR.

> Then I realized the code the macros generate have to work in the 2015 edition so I probably can't remove the redundant import. (I still don't understand why things break inside diesel when I remove it, but that's not that important)

So the problem here is the following: We use the macros also in diesel itself, as they are useful for us too. For example `sql_function!` is quite widely used internally to define various sql functions. In the end all of those macros generate some code that relies on types from diesel being scope. That needs to be wrapped somehow by the macro, as we cannot assume that the user will import all of those types (Some of them are even not considered part of the public API). That implies we need something to import from. For cases where someone else uses diesel we just assume that there exists a crate named diesel that we can import. That `use diesel;` brings it into local scope. For our internal use things are a bit more complicated. We decided to workaround that by just having a `mod diesel { pub use super::*;'}` in our crate root. That allows proc macros to refer to diesel as like it was an external crate even inside diesel itself. In the end it's just a hack to workaround that there is nothing like `$crate` for proc macros.

jplatte:
Interesting! Should I add a comment to the `use diesel;` that explains this?

weiznich:
That's a good idea, otherwise I've forgotten that the next time I look at that code 🙈

jplatte:
So apparently CI does `cargo fmt` on 1.36.0 while you have a `rust-toolchain` file with 1.37.0? Not that it makes a difference here (they produce the same output, some later version changes things up in one place), I had a local rustup override to see newer clippy lints, but it does confuse me a bit.

diesel_derives/src/util.rs
```diff
@@ -8,6 +8,13 @@ pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
     quote! {
         #[allow(unused_imports)]
         const _: () = {
+            // This import is not actually redundant. When using diesel_derives
+            // inside of diesel, `diesel` doesn't exist as an extern crate, and
+            // to work around that it contains a private
+            // `mod diesel { pub use super::*; }` that this import will then
+            // refer to. In all other cases, this imports refers to the extern
+            // crate diesel.
+            #[allow(clippy::single_component_path_imports, clippy::unknown_clippy_lints)]
```
JohnTitor: I think this will be fixed next clippy release: rust-lang/rust-clippy#5183 (but we can remove this later.)

weiznich:
```suggestion
```

jplatte: Ah, nice. I don't particularly care whether it is silenced explicitly here if no future beta or stable clippy emits it from inside macros.

Should I remove the lint attribute?

jplatte: Given @weiznich's change suggestion, I've removed it.

jplatte:
CI failure seems unrelated.

PR #2313, opened 2020-02-27
GitHub actions

martell:
#2269

I don't think I can push to other peoples branches so I guess they can be cherrypicked from here if the commits are useful.

- Only Run Publish Docs on diesel-rs/diesel
- Fix actions on MacOS Latest
   The only way I could get this to work without running into sudo issues was to remove libpq
  

cc: @weiznich


weiznich:
Cherry-picked to #2269. Thanks.

PR #2314, opened 2020-02-28
update/nightly-2020-02-27

martell:
With some of the other PRs coming in like #2248 #2257 I have been trying to maintain a fork with both of them in the same branch to see what the world would be like.

In order to simplify some things I thought it would be a good time to upstream the update to `nightly-2020-02-27`

This currently depends on #2269 (  dc3dc23 is the only real commit here )
I'm not sure what the process for picking a nightly is but I hope this is reasonable.


martell:
hmm, what is interesting is that is passes when it contains #2248
I'll have to look at the specifics of why that is the case but I do believe that will be merged soon anyway?

weiznich:
#2248 will be merged as soon as we get at least another review on that PR, especially on the documentation.

martell:
super-seeded by #2339

PR #2315, opened 2020-02-28
Add revert capability to embed_migrations

cullenvandora:
I've added revert capability to embed_migrations macro.  This is useful to permit tearing down migrations when conducting unit tests.


cullenvandora:
I'm going to close this pull request as I've identified some errors that aren't covered by current unit tests.

PR #2316, opened 2020-02-29
Enable using this library from git directly

pksunkara:
I love living on the edge 😄 

Unfortunately, I am unable to install `diesel_cli` from github directly because it would want `git` and `branch` in the versions. I hacked my way around it by installing from path after cloning the repo. Maybe, also adding `git` to these items would fix the issue? Or maybe replace `path` with `git`? Any experts?

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -14,7 +14,6 @@ edition = "2018"
 
 [dependencies]
 byteorder = "1.0"
-diesel_derives = "~1.4.0"
 chrono = { version = "0.4", optional = true }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.8.0, <0.18.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
@@ -33,6 +32,10 @@ num-integer = { version = "0.1.32", optional = true }
 bigdecimal = { version = ">= 0.0.10, <= 0.1.0", optional = true }
 bitflags = { version = "1.0", optional = true }
 r2d2 = { version = ">= 0.8, < 0.9", optional = true }
+
+[dependencies.diesel_derives]
+version = "~1.4.0"
+path = "../diesel_derives"
 
 [dev-dependencies]
 cfg-if = "0.1.0"
--- a/diesel_derives/Cargo.toml
+++ b/diesel_derives/Cargo.toml
@@ -16,8 +16,11 @@ proc-macro2 = "1"
 
 [dev-dependencies]
 cfg-if = "0.1.0"
-diesel = "1.4.0"
 dotenv = "0.10.0"
+
+[dev-dependencies.diesel]
+version = "~1.4.0"
+path = "../diesel"
 
 [lib]
 proc-macro = true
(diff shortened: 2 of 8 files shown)


PhilipTrauner:
I've also run into this. I currently have to add `diesel_derives` to the `replace` table, which works around the issue.

```
[dependencies]
diesel = { git = "https://github.com/diesel-rs/diesel", package = "diesel", features = ["..."] }

[replace]
"diesel_derives:1.4.1" = {git = "https://github.com/diesel-rs/diesel", package = "diesel_derives" }
```

weiznich:
Thanks for submitting this PR but we are not interested in this change, because it complicates the release process greatly.

pksunkara:
It would not complicate the release process. When you release and do `cargo publish`, cargo sees the `version` is still there and uses it. Your release process would be the same as it is now. It doesn't change.

weiznich:
In my understanding a dependency constrain like
```
diesel_derives = { path = "../diesel_derives", version = "~1.4.0" }
```
reads like: use version ~1.4.0 of the crate diesel_derives that you find at the location `../diesel_derives`. That would make it impossible to do a cargo publish there because we use a local dependency here. If that's not the case please provide a link to the corresponding documentation. (If it's not documented behaviour I would rather not like to depend on it.)

paolobarbolini:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

weiznich:
Thanks for the clarification I didn't saw this clarification in earlier version.

diesel/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 byteorder = "1.0"
-diesel_derives = "~1.4.0"
+diesel_derives = { path = "../diesel_derives", version = "~1.4.0" }
```
weiznich: Can we change this to 
```toml
[dependencies.diesel_derives]
version = "~1.4.0"
path = "../diesel_derives"
```
as this makes it easier to change the version?
(Same for all other changes)

pksunkara:
Addressed your review. Regarding the link, I am pretty new to rust, so I read that new version and thought this is how it has always been.

weiznich:
CI says you need to update the Cargo.toml in `diesel_tests` too.

pksunkara:
CI is passing now.

diesel_derives/Cargo.toml
```diff
@@ -16,7 +16,7 @@ proc-macro2 = "1"
 
 [dev-dependencies]
 cfg-if = "0.1.0"
-diesel = "1.4.0"
+diesel = { path = "../diesel" }
```
weiznich: This will cause problems with `cargo publish` as this is a local only dependency. (It does not help that this is a dev-dependency…) You need to also add an explicit version here (and for the other dev-dependencies too)

diesel_tests/Cargo.toml
```diff
@@ -9,14 +9,14 @@ edition = "2018"
 
 [build-dependencies]
 diesel = { path = "../diesel", default-features = false }
-diesel_migrations = { version = "1.4.0" }
+diesel_migrations = { path = "../diesel_migrations" }
```
weiznich: Those dependencies should be fine, as we don't publish this crate.

pksunkara:
@weiznich It reminded me that you can probably use https://github.com/pksunkara/cargo-workspaces to simplify your release process.

I made it so that we could simply our release process at http://github.com/clap-rs/clap

PR #2321, opened 2020-03-06
Make ConnectionManager implement Sync automatically

mikeyhew:
By using `PhantomData<fn() -> T>` instead of `PhantomData<T>` for the type of the `_marker` field, this makes `ConnectionManager` implement `Sync` automatically, and means we can remove the `unsafe impl` of `Sync` for `ConnectionManager<T>`.

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -31,10 +31,8 @@ use crate::sql_types::HasSqlType;
 #[derive(Debug, Clone)]
 pub struct ConnectionManager<T> {
     database_url: String,
-    _marker: PhantomData<T>,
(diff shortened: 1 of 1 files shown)


diesel/src/r2d2.rs
```diff
@@ -31,11 +31,9 @@ use crate::sql_types::HasSqlType;
 #[derive(Debug, Clone)]
 pub struct ConnectionManager<T> {
     database_url: String,
-    _marker: PhantomData<T>,
+    _marker: PhantomData<fn() -> T>,
```
Mingun: I think, you should leave comment from PR description that explains, why not just `PhantomData<T>` is used

Issue #2191, opened 2019-10-09
Support custom aggregate functions using the sqlite backend


weiznich:
Diesel currently provides a quite nice interface for custom sql functions using the sqlite backend, but it is not possible to write custom aggregate functions yet. 

Starting points for an implementation:
* `sqlite::connection::raw::RawConnection` has [a function to register custom sql functions](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/connection/raw.rs#L67-L110). A similar function is required for aggregating functions. See the  [Sqlite documentation](https://www.sqlite.org/c3ref/create_function.html) for which parameters need to be passed. Both functions should share code where possible.
* `sqlite::connection::function` defines some boiler plate code that does setup/teardown before/after calling a user provided function inside of a sql function. Maybe this needs to be extended to handle some special cases for aggregating functions.
* `SqliteConnection` defines a function to [register custom sql functions]. A similar function is required for aggregating functions. 
* The `sql_function!` proc macro provides some code generation for defining the necessary types used to call such a function. The [already existing](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/sql_function.rs#L138-L196) function there needs to be extended to expose functions to register the aggregating function using the methods implemented in the previous steps. Depending on the [`is_aggregate`](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/sql_function.rs#L24) flag either methods to register aggregating or non aggregating functions should be generated.
* Write some test cases [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/connection/mod.rs#L326). See the already existing test cases for inspiration.

Open questions:
* How exactly does the user facing interface look like? It could be a good idea not to take two closure but a custom trait defining the two required methods

thekuom:
I would like to start on this if you're still looking to have this functionality

weiznich:
Yes we are still interested in this functionality. If you have any questions/problems while trying to implement this, be sure to ask for help here or in our gitter channel. Also opening a WIP PR to get some early feedback on the code could be a good idea.

weiznich:
Closed by #2325

PR #2325, opened 2020-03-10
Add ability to create custom aggregate functions in sqlite

thekuom:
Reference issue #2191


thekuom:
I've added some tests to help define the behavior we are expecting. There is a test with one arg and one with a workaround for multiargs, which takes a slice. 

Since we are going for the trait approach, my first approach for the multiarg was going to be to generate some trait `AggregatorN`, where `N` is the number of arguments in the function signature, but I don't think Rust can do that at this time.

I took inspiration from how Python does it here: https://wellsr.com/python/create-scalar-and-aggregate-functions-for-python-sqlite/

diesel/src/sqlite/connection/mod.rs
```diff
@@ -372,4 +372,105 @@ mod tests {
...
+
+    struct MySum<T: Foldable> {
+        sum: T::Sum,
+    }
+    impl<T: Foldable> MySum<T> {
+        fn new() -> Self {
+            Self { sum: 0 }
+        }
+    }
+    impl<T: Foldable> Aggregator<T> for MySum<T> {
```
weiznich: The same comment about `T: Foldable` from above applies here too.

Additionally we need to talk about the design of the `Aggregator` trait a bit more. First of all I think it's a good idea to have a septate trait for this. That trait should be something similar to the `Fn` traits in stdlib, but with two functions instead of one. So it might be useful to have a look at [their documentation](https://doc.rust-lang.org/core/ops/trait.Fn.html). Trait bounds around callbacks are normally written as `F: Fn(Arg1, Arg2) -> ReturnType`, which is syntactic sugar for the unstable `Fn` trait. Using the unstable trait this would look like `F: Fn<(Arg1, Arg2), Output = ReturnType>`.
Additional we need some way to setup a new instance of the `Aggregator` for each call. I think you want to use the `new` function for that, but that won't work in generic context, as it's not possible to access functions that are not part of a certain trait. So just let's use `Default` for this.

That leads to the following design of the `Aggregator` trait:
```rust
trait Aggregator<Args>: Default {
    type Output;

    fn step(&mut self, args: Args);
    fn finalize(self) -> Self::Output;
}
```
where I'm not sure if `finalize` should really consume `self` (It seems to be the most reasonable choice.)

diesel/src/sqlite/connection/mod.rs
```diff
@@ -372,4 +372,105 @@ mod tests {
...
+    #[test]
+    fn register_aggregate_function() {
+        use self::my_sum_example::dsl::*;
+
+        let connection = SqliteConnection::establish(":memory:").unwrap();
+        connection.execute("CREATE TABLE my_sum_example (id integer primary key autoincrement, value integer)").unwrap();
+        connection.execute("INSERT INTO my_sum_example (value) VALUES (1), (2), (3)").unwrap();
+
+        let my_sum_struct = MySum::new();
+        my_sum::register_impl(&connection, my_sum_struct).unwrap();
```
weiznich: Registering a custom aggregate function will need to only pass the corresponding type, not a specific instance, because otherwise you would use the same instance for all calls to that function, which would quite likely corrupt the results, due to invalid states.

So something like `my_sum::register_impl::<MySum>(&connection)`

diesel/src/sqlite/connection/mod.rs
```diff
@@ -372,4 +372,105 @@ mod tests {
...
+    struct RangeMax<T: SqlOrd> {
+        max_value: Option<T>,
+    }
+    impl<T: SqlOrd> RangeMax<T> {
+        fn new() -> Self {
+            Self { max_value: None }
+        }
+    }
+    impl<T: SqlOrd> Aggregator<T> for RangeMax<T> {
+        fn step(&mut self, exprs: [T; 3]) {
```
weiznich: This variant would not allow the arguments to have different types, which is a quite hard assumption for the general case.
Using the variant of the `Aggregator` trait proposed above would allow that by using a tuple here. Unfortunately we then need to decompose the tuple as argument here, as we cannot relay on the compiler magic happening for the `Fn` traits.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -372,4 +372,105 @@ mod tests {
...
         assert_eq!(Ok((2, 3, 4)), added);
     }
+
+    use crate::sql_types::Foldable;
+    sql_function! {
+        #[aggregate]
+        fn my_sum<ST: Foldable>(expr: ST) -> ST::Sum;
+    }
+
+    struct MySum<T: Foldable> {
```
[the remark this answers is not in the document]
thekuom: Got it :+1:

diesel/src/sqlite/connection/mod.rs
thekuom: That makes a lot of sense. Thanks for the insight!

diesel/src/sqlite/connection/mod.rs
```diff
@@ -372,4 +372,102 @@ mod tests {
...
+
+    #[derive(Default)]
+    struct RangeMax<T: Default> {
+        max_value: Option<T>,
+    }
+    impl<T: Default + Ord> Aggregator<(T, T, T)> for RangeMax<T> {
+        type Output = Option<T>;
+
+        fn step(&mut self, args: (T, T, T)) {
+            let (x0, x1, x2) = args;
```
weiznich: Just a small nit: This can also be written as
```suggestion
        fn step(&mut self, (x0, x1, x2): (T, T, T)) {
```

thekuom: Oh right! I will make sure to add that in my next commit

diesel_derives/src/sql_function.rs
```diff
@@ -123,7 +125,39 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic>
...
+                use diesel::sqlite::Aggregator;
+
+                #[allow(dead_code)]
+                pub fn register_impl<A, #(#arg_name),*, #(#type_args),*>(
+                    conn: &SqliteConnection
+                ) -> QueryResult<()>
+                where
+                     A: Aggregator<#(#arg_name),*, Output=ToSql<#return_type, Sqlite>> + Send + 'static,
+                     (#(#arg_name,)*): Queryable<(#(#arg_type,)*), Sqlite>,
+                     #(#type_params),*
```
thekuom: I've been having some difficulty with this. For example, in `diesel/src/expression/functions/aggregate_folding.rs`:
```
sql_function! {
    #[aggregate]
    fn sum<ST: Foldable>(expr: ST) -> ST::Sum;
}
```

What I've written does indeed generate `ST: Foldable` in the where clause. However, I'm receiving compile errors like this:
```
error[E0277]: the size for values of type `(dyn serialize::ToSql<<ST as sql_types::fold::Foldable>::Sum, sqlite::backend::Sqlite> + 'static)` cannot be known at compilation time
   --> diesel/src/expression/functions/mod.rs:167:9
    |
167 |           sql_function_proc! { $($args)* }
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
   ::: diesel/src/expression/functions/aggregate_folding.rs:3:1
    |
3   | / sql_function! {
4   | |     /// Represents a SQL `SUM` function. This function can only take types which are
5   | |     /// Foldable.
6   | |     ///
...   |
21  | |     fn sum<ST: Foldable>(expr: ST) -> ST::Sum;
22  | | }
    | |_- in this macro invocation
    |
    = help: the trait `std::marker::Sized` is not implemented for `(dyn serialize::ToSql<<ST as sql_types::fold::Foldable>::Sum, sqlite::backend::Sqlite> + 'static)`
    = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
note: required by `sqlite::connection::raw::Aggregator`
   --> diesel/src/sqlite/connection/raw.rs:14:1
    |
14  | pub trait Aggregator<Args>: Default {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0038]: the trait `serialize::ToSql` cannot be made into an object
   --> diesel/src/expression/functions/mod.rs:167:9
    |
167 |           sql_function_proc! { $($args)* }
    |           ^^^^^^^^^^^^^^^^^^ the trait `serialize::ToSql` cannot be made into an object
    |
   ::: diesel/src/expression/functions/aggregate_folding.rs:3:1
    |
3   | / sql_function! {
4   | |     /// Represents a SQL `SUM` function. This function can only take types which are
5   | |     /// Foldable.
6   | |     ///
...   |
21  | |     fn sum<ST: Foldable>(expr: ST) -> ST::Sum;
22  | | }
    | |_- in this macro invocation
    |
    = note: method `to_sql` has generic type parameters
```

Do you have any advice for this?

diesel_derives/src/sql_function.rs
```diff
@@ -123,7 +125,39 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic>
...
+            tokens = quote! {
+                #tokens
+
+                use diesel::sqlite::{Sqlite, SqliteConnection};
+                use diesel::serialize::ToSql;
+                use diesel::deserialize::Queryable;
+                use diesel::sqlite::Aggregator;
+
+                #[allow(dead_code)]
+                pub fn register_impl<A, #(#arg_name),*, #(#type_args),*>(
```
thekuom: Ideally, our `register_impl` would only require specifying the Aggregator trait implementation, but this requires a lot more type params. Is there any way around this that you can think of i.e. am I approaching this macro incorrectly?

diesel/src/sqlite/connection/functions.rs
```diff
@@ -53,6 +53,32 @@ where
...
+{
+    let fields_needed = Args::Row::FIELDS_NEEDED;
+    if fields_needed > 127 {
+        return Err(Error::DatabaseError(
+            DatabaseErrorKind::UnableToSendCommand,
+            Box::new("SQLite functions cannot take more than 127 parameters".to_string()),
+        ));
+    }
+
+    // TODO @thekuom: how do we convert this to ffi::sqlite3_value?
```
thekuom: In the `register` function, the function passed into the connection takes `ffi::sqlite3_value` as args. I am trying to do the same here. I haven't put too much time into this part yet, but I would appreciate any advice on approaching this function.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,6 +242,76 @@ extern "C" fn run_custom_function<F>(
...
+        aggregator.step(args);
+    }
+}
+
+extern "C" fn run_aggregator_final_function<'a, A>(
+    ctx: *mut ffi::sqlite3_context,
+)
+    where A: Aggregator<&'a [*mut ffi::sqlite3_value]> + Send + 'static,
+{
+    // TODO @thekuom: fill this out
```
weiznich: We need to do the following here:

* Grab the aggregator context
* Get a reference to the aggregator, using the code from above
* Move out the existing aggregator and replace it with `OptionalAggregator::None`. That guarantees that we call a potential `Drop` impl while calling `Aggregator::finalize` without having a existing reference left behind somewhere.
* Create a `SerializedValue` using the same method as in `functions::register` line 39 - 51
* Write back the result to the database using the same method as in `run_custom_function` line 234 - 238

diesel_derives/src/sql_function.rs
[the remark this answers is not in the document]
thekuom: Would we not have to include `type_params` somewhere to handle something like
```
sql_function! {
    #[aggregate]
    fn sum<ST: Foldable>(expr: ST) -> ST::Sum;
}
```

I was running into issues where it couldn't find `ST`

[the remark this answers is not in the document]
weiznich: I do not think that we can support  custom generic sql functions for sqlite, at least in the first design iteration. The current implementation for normal sql does also not support not support generic functions. See [the corresponding documentation](http://docs.diesel.rs/diesel/macro.sql_function.html#use-with-sqlite). 
That implies that we don't have to include `type_params` for now.

(Let's just get something working, then we could think about how to fix that in a future version)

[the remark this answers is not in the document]
thekuom: Ok, so for the existing aggregates in in `diesel/src/expression/functions/aggregate_folding.rs`, we will have to ignore those in `diesel_derives/src/sql_function.rs`?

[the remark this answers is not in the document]
weiznich: We would just handle this case as it's currently handled by the `sql_function!` implementation [here](https://github.com/diesel-rs/diesel/blob/00f7de703aea02659bddad8d168b9fe6d4aee276/diesel_derives/src/sql_function.rs#L138) (notice the `type_args.is_empty()` there?) by just not generating the corresponding `register_impl` function. As those functions are built-in we do not have need to register a implementation for them.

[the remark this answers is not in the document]
thekuom: Oh right. I do remember seeing that. Thanks!

thekuom:
There's still some TODOs around cleaning up the code. I ran both of the tests I wrote locally and they do pass 🎉

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,12 +251,150 @@ extern "C" fn run_custom_function<F>(
...
+                &OptionalAggregator::Some(ref agg) => agg,
+                &OptionalAggregator::None => {
+                    ffi::sqlite3_result_error(
+                        ctx,
+                        NULL_AG_CTX_ERR.as_ptr() as *const _ as *const _,
+                        NULL_AG_CTX_ERR.len() as _,
+                        );
+                    return;
+                }
+            },
```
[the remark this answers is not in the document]
weiznich: Forget what I've wrote above, that solution will leak memory.
We need to move our aggregator context out of the the sqlite context field and explicitly call drop on it. That means the code above should look like this:
```suggestion
            Some(ref a) => match std::mem::replace(a.as_mut(), OptionalAggregator::None {
                OptionalAggregator::Some(agg) => agg,
                OptionalAggregator::None => {
                    ffi::sqlite3_result_error(
                        ctx,
                        NULL_AG_CTX_ERR.as_ptr() as *const _ as *const _,
                        NULL_AG_CTX_ERR.len() as _,
                        );
                    return;
                }
            },
```

so that in the end of the function our context is cleared (== set to `None`) and the destructor of our aggregator, so that potential allocated memory/… is freed.

diesel/src/sqlite/connection/serialized_value.rs
```diff
@@ -5,6 +5,7 @@ use std::ptr::{self, NonNull};
 
 use crate::sqlite::SqliteType;
 
+#[allow(missing_debug_implementations)]
```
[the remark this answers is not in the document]
thekuom: It was giving me some error. I forgot to add a TODO to remove it

diesel_derives/src/sql_function.rs
```diff
@@ -123,7 +125,55 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic>
...
+                    ) -> QueryResult<()>
+                        where
+                        A: Aggregator<(#(#arg_name,)*)> + Send + 'static,
+                        A::Output: ToSql<#return_type, Sqlite>,
+                        (#(#arg_name,)*): Queryable<(#(#arg_type,)*), Sqlite>,
+                    {
+                        conn.register_aggregate_function::<(#(#arg_type,)*), #return_type, _, _, A>(#sql_name)
+                    }
+                };
+            } else if arg_name.len() == 1 {
```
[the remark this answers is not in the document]
thekuom: I included this because I thought this:
```
fn step(&self, args: (i32,)) {
  let x = args.0;
  ...
}
```
seemed a little unintuitive when the function signature is
```
fn my_sum(expr: Integer)
```

What are your specific concerns around having this be a special case?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,12 +251,150 @@ extern "C" fn run_custom_function<F>(
...
+                &mut OptionalAggregator::None => {
+                    // Not sure if we should use `write_unaligned` instead
+                    ptr::write(a.as_mut(), OptionalAggregator::Some(A::default()));
+                    if let &mut OptionalAggregator::Some(ref mut agg) = a.as_mut() {
+                        agg
+                    } else {
+                        unreachable!("We've written the aggregator above to that location, it must be there")
+                    }
+                }
+            },
```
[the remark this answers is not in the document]
thekuom: Wouldn't we need to keep the nested matches to capture the `a` that is referenced in the the `Some(&mut OptionalAggregator::None` arm?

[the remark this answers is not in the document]
weiznich: A right, that's a `NonNull` in there. I think we can make this work with one match nevertheless, by doing something like:

```rust
match aggregator_context.map(|a| &mut *a.as_ptr()) {
    Some(OptionalAggregator::Some(ref mut agg)) => agg,
    Some(OptionalAggregator::None) => {…}
    None => …
}
```

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,12 +251,150 @@ extern "C" fn run_custom_function<F>(
     }
 }
 
+// Need a custom option type here, because the std lib one does not have guarantees about the discriminate values?
```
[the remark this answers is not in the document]
weiznich: Ok I had a short chat with Ralf Jung in zulip. He says the following about the general approach:

> so given that RFC, I agree that it is okay to transmute a "bunch of zero bytes" to MyOption.
however, notice that MyOption contains padding, so if you transmute it back those padding bytes will be uninitialized.

so it seems to be fine what we are doing here.

Additionally I've found a section in the [corresponding RFC](https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md#guide-level-explanation) that specifically answers the question raised in this comment.

> It relies on the fact that the Some-ness of an Option (or the tag of any repr(Rust) enum) can't rely on the tag of a repr(C/Int) enum. Or in other words, repr(C/Int) enums have opaque tags. The cost of making this work is that Option<MyEnum> would have to be larger than MyEnum.

> It would be nice for this to work, but if you really need it, you can just define #[repr(u8)] COption<T> { ... } and use that.

diesel_derives/src/sql_function.rs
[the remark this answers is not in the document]
weiznich: My main concern is that it's an inconsistency. There is some case that is just handled differently. I see the reason why this is probably a good idea, but we need to document that accordantly. Another minor concern is that it would make it harder to write an macro that generates a similar function call for multiple tuple sizes starting by one element, but that's probably not relevant.

In that context: I just remember that we should probably update the documentation of `sql_function!` [here](https://github.com/diesel-rs/diesel/blob/5369a7a3d13b1323fcac9acf9d13256f19a970a0/diesel_derives/src/lib.rs#L558-L599) to mention support for aggregate functions and contain an simple example (just the sum test case or something like that?) Also the `Aggregator` trait needs at least a few lines of documentation pointing in principle to the `sql_function` example.

diesel/src/sqlite/connection/raw.rs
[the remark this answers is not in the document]
thekuom: I have this right now:
```
        let aggregator = match aggregate_context.map(|a| &mut *a.as_ptr()) {
            Some(&mut OptionalAggregator::Some(ref mut agg)) => agg,
            Some(mut a_ptr @ &mut OptionalAggregator::None) => {
                ptr::write_unaligned(a_ptr as *mut _, OptionalAggregator::Some(A::default()));
                if let &mut OptionalAggregator::Some(ref mut agg) = a_ptr {
                    agg
                } else {
                    unreachable!("We've written the aggregator above to that location, it must be there")
                }
            },
            None => ...
        };
```
which passes the tests. Do you have any good ways to test memory leaks? I tried (naively) just running `valgrind cargo test --no-default-features --features "sqlite" register_aggregate_function` but that didn't really work

diesel_derives/src/sql_function.rs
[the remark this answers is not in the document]
thekuom: I'll make sure to include the single and multiarg cases in the documentation :+1:

diesel/src/sqlite/connection/functions.rs
```diff
@@ -76,6 +62,35 @@ where
     Ok(())
 }
 
+pub(crate) fn build_sql_function_args<ArgsSqlType, Args>(args: &[*mut ffi::sqlite3_value]) -> Result<Args, Error>
```
weiznich, line 65: Can we use these functions in the existing non aggregate versions too?

diesel/src/sqlite/mod.rs
```diff
@@ -12,5 +12,19 @@ pub mod query_builder;
...
 pub use self::backend::{Sqlite, SqliteType};
 pub use self::connection::SqliteConnection;
-pub use self::connection::raw::Aggregator;
 pub use self::query_builder::SqliteQueryBuilder;
+
+/// The implementation details for a SQLite aggregate function
+///
+/// This trait is to be used in conjunction with the `sql_function!`
+/// macro for defining a custom SQLite aggregate function.
+pub trait SqliteAggregateFunction<Args>: Default {
```
weiznich, line 21: Some small suggestions on the wording, but I'm not 100% happy with my formulations either.

```suggestion
/// Trait for the implementation of a SQLite aggregate function
///
/// This trait is to be used in conjunction with the `sql_function!`
/// macro for defining a custom SQLite aggregate function. See
/// the documentation [there](link to sql_function!) for details. 
pub trait SqliteAggregateFunction<Args>: Default {
```

diesel/src/sqlite/mod.rs
```diff
@@ -12,5 +12,19 @@ pub mod query_builder;
...
+    /// The result type of the SQLite aggregate function
+    type Output;
+
+    /// The `step()` method is called once for every record of the query
+    fn step(&mut self, args: Args);
+    /// After the last row has been processed, the `finalize()` method is
+    /// called to compute the result of the aggregate function
+    fn finalize(&self) -> Self::Output;
```
weiznich, line 29: As we now move out the aggregator context in the finalize function we can just do 
```suggestion
    fn finalize(self) -> Self::Output;
```
here, right?
This should lower the overhead for cases where we would return a `String` for example, as we don't need to clone it anymore.

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,154 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
 /// #     Ok(())
 /// # }
 /// ```
+///
+/// ## Custom Aggregate Functions
+/// Custom aggregate functions can be created in SQLite by adding the `aggregate`
+/// attribute and implementing the
+/// [SqliteAggregateFunction](../diesel/sqlite/trait.SqliteAggregateFunction.html) trait.
```
weiznich, line 604: Again suggestions on better wording:
```suggestion
/// ## Custom Aggregate Functions
/// Custom aggregate functions can be created in SQLite by adding an `#[aggregate]`
/// attribute inside of `sql_function`.  `register_impl` needs to be called on the generated function with 
/// a type implementing the  [SqliteAggregateFunction](../diesel/sqlite/trait.SqliteAggregateFunction.html) trait 
/// as type parameter as shown in the examples below.
```

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,154 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+///         .get_result::<i32>(&connection)?;
+///
+///     println!("The total score of all the players is: {}", result);
+///
+/// #    assert_eq!(60, result);
+///     Ok(())
+/// }
+/// ```
+///
+/// With multiple parameters
```
weiznich, line 667: Some more details here would be great, so something like
```suggestion
/// With multiple function arguments the arguments are passed as tuple to `SqliteAggregateFunction`
```

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,154 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+/// #
+/// # #[cfg(not(feature = "sqlite"))]
+/// # fn main() {
+/// # }
+/// use diesel::sql_types::Integer;
+/// use diesel::sqlite::SqliteAggregateFunction;
+///
+/// sql_function! {
+///     #[aggregate]
+///     fn range_max(x0: Integer, x1: Integer, x2: Integer, x3: Integer) -> Integer;
```
weiznich, line 686: For the sake of a simple example, can we just use two parameters here? It should simplify the implementation a bit and teaches the same lesson.

diesel/src/sqlite/connection/raw.rs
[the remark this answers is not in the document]
weiznich: You can try with [the leak sanitizer](https://github.com/rust-lang/rustc-dev-guide/blob/master/src/sanitizers.md) provided by rustc. Otherwise the unsafe code looks sound for me. I will try to get a second opinion from someone else in the reviewer team on this, so cc @diesel-rs/reviewers: It would be great if someone could have a look at the unsafe code involved here.

thekuom:
Whoops, didn't see your comments. Was working on fixing the build error. Expect another commit here soon

diesel/src/sqlite/connection/functions.rs
thekuom: Are you referring to `register()` in `src/sqlite/connection/functions.rs` or is there another place I am missing?

weiznich: Oh, I've missed that `register` already uses this function, so forget what I've wrote above 🤦‍♂

diesel/src/sqlite/connection/mod.rs
```diff
@@ -370,4 +385,118 @@ mod tests {
...
+        }
+    }
+    table! {
+        range_max_example {
+            id -> Integer,
+            value1 -> Integer,
+            value2 -> Integer,
+            value3 -> Integer,
+        }
+    }
```
weiznich, line 476: Can you add some a single new line between each of the blocks, so that for example `MySum` and the corresponding `SqliteAggregateFunction` impl are seperate by an empty line?
(This applies to the whole block)

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,6 +243,115 @@ extern "C" fn run_custom_function<F>(
     }
 }
 
+// Need a custom option type here, because the std lib one does not have guarantees about the discriminate values
+// See: https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md#opaque-tags
+#[repr(u8)]
+enum OptionalAggregator<A> {
+    // Discriminant is 0
+    #[allow(dead_code)]
```
weiznich, line 251: Is this needed anymore? I think we cover this case somewhere below in one of the match statements.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,6 +243,115 @@ extern "C" fn run_custom_function<F>(
...
+    Ret: ToSql<RetSqlType, Sqlite>,
+    Sqlite: HasSqlType<RetSqlType>,
+{
+    unsafe {
+        let aggregate_context = ffi::sqlite3_aggregate_context(ctx, 0);
+        let mut aggregate_context = NonNull::new(aggregate_context as *mut OptionalAggregator<A>);
+        let aggregator = match aggregate_context {
+            Some(ref mut a) => match std::mem::replace(a.as_mut(), OptionalAggregator::None) {
+                OptionalAggregator::Some(agg) => agg,
+                OptionalAggregator::None => {
```
weiznich, line 322: If I think about this case more in detail, this happens if you call your aggregate function on an empty set. So we should probably just return a null value in this cases? 
That implies a few more things:
* We should have a test case covering applying an aggregate function to an empty select.
* If we decide to return null in such cases, that means we need to enforce that we only allow to register custom aggregate functions for function definitions where the return type is `Nullable<…>`.
* We should probably indicate this also in the definition of `SqliteAggregateFunction` by making `finalize` explicitly return a `Option<Self::Output>`

(Sorry for not noticing that any earlier :see_no_evil:)

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,151 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+/// sql_function! {
+///     #[aggregate]
+///     fn my_sum(x: Integer) -> Integer;
+/// }
+///
+/// #[derive(Default)]
+/// struct MySum<T> { sum: T }
+///
+/// # #[cfg(feature = "sqlite")]
+/// impl<T: Default + Copy + AddAssign> SqliteAggregateFunction<T> for MySum<T> {
```
weiznich, line 633: Probably a small thing but I think I would prefer having at least one example without a generic implementation of `SqliteAggregateFunction` to show it's totally fine to not use generics here.
(In the end that's the example people will copy and modify as first step. The simple at least one example, the better I would say)

diesel/src/sqlite/connection/raw.rs
thekuom: The linter was telling me something about this case never being constructed. I will remove it and check again 👍

diesel/src/sqlite/connection/raw.rs
thekuom: Sounds good! I should've thought about that as well early on

diesel/src/sqlite/mod.rs
```diff
@@ -13,3 +13,21 @@ pub mod query_builder;
 pub use self::backend::{Sqlite, SqliteType};
 pub use self::connection::SqliteConnection;
 pub use self::query_builder::SqliteQueryBuilder;
+
+/// Trait for the implementation of a SQLite aggregate function
+///
+/// This trait is to be used in conjunction with the `sql_function!`
+/// macro for defining a custom SQLite aggregate function. See
+/// the documentation [there](../../diesel_derives/macro.sql_function_proc.html) for details.
```
thekuom, line 21: Here is the link I mentioned in Gitter by the way

diesel/src/sqlite/mod.rs
```diff
@@ -13,3 +13,21 @@ pub mod query_builder;
...
+    /// The result type of the SQLite aggregate function
+    type Output;
+
+    /// The `step()` method is called once for every record of the query
+    fn step(&mut self, args: Args);
+    /// After the last row has been processed, the `finalize()` method is
+    /// called to compute the result of the aggregate function. If no rows
+    /// were processed, `finalize()` is not called and `None` is returned
+    /// instead
+    fn finalize(self) -> Self::Output;
```
thekuom, line 32: I decided to have `finalize` not return an `Option` since it made more sense to me not to return one. However, I did add an extra sentence in the documentation addressing when `None` would be returned

diesel/src/sqlite/mod.rs
weiznich: I think 
```suggestion
/// the documentation [there](../prelude/macro.sql_function.html) for details.
```

should also link to the corresponding documentation, but in such a way that it will not break on docs.rs

thekuom: I don't think it is working with
```
cargo doc --no-deps --features "sqlite extras"
```

thekuom: This is what is in `target/doc/diesel/prelude`

image

weiznich: Try building with a newer rustdoc version :wink:
`cargo +1.42.0 doc --features "sqlite"` leads to different results, including a `macro.sql_function.html` file.
(We will bump to rustdoc version before the next release to make this work)

diesel/src/sqlite/mod.rs
weiznich: :+1:

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,174 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+/// # use diesel::*;
+/// # use diesel::sql_types::Integer;
+/// #
+/// # fn main() {
+/// # }
+/// #
+/// # #[cfg(feature = "sqlite")]
+/// sql_function! {
+///     #[aggregate]
+///     fn my_sum(x: Integer) -> Integer;
```
thekuom, line 762: Reminder to myself to call this wont_work

diesel/src/sqlite/connection/raw.rs
```diff
@@ -194,6 +243,111 @@ extern "C" fn run_custom_function<F>(
...
+        // See: https://www.sqlite.org/c3ref/aggregate_context.html
+        let aggregate_context = ffi::sqlite3_aggregate_context(ctx, 0);
+        let mut aggregate_context = NonNull::new(aggregate_context as *mut OptionalAggregator<A>);
+        let aggregator = match aggregate_context {
+            Some(ref mut a) => match std::mem::replace(a.as_mut(), OptionalAggregator::None) {
+                OptionalAggregator::Some(agg) => agg,
+                OptionalAggregator::None => unreachable!("We've written to the aggregator in the xStep callback. If xStep was never called, then ffi::sqlite_aggregate_context() would have returned a NULL pointer")
+            },
+            None => return,
+        };
```
weiznich, line 327: One last simplification/improvement:
This should remove the nested match statement and make it a bit clearer that a null value is returned by the none case, by explicitly setting the corresponding value, instead of implicitly doing that because of sqlite default behaviour:
```suggestion
        let aggregator = match aggregate_context.map(|a| std::mem::replace(a.as_mut(), OptionalAggregator::None)) {
            Some(OptionalAggregator::Some(agg)) => agg,
            Some(OptionalAggregator::None) => unreachable!("We've written to the aggregator in the xStep callback. If xStep was never called, then ffi::sqlite_aggregate_context() would have returned a NULL pointer"),
            None => {
                // explicitly set a null value as return value here, just to make it obvious what's returned in this case
                SerializedValue {
                    ty: Sqlite::metadata(&()),
                    data: None,
                }.result_of(ctx);
                return;
            },
        };
```

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,174 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+/// sql_function! {
+///     #[aggregate]
+///     fn my_sum(x: Integer) -> Nullable<Integer>;
+/// }
+///
+/// #[derive(Default)]
+/// struct MySum<T> { sum: T }
+///
+/// # #[cfg(feature = "sqlite")]
+/// impl SqliteAggregateFunction<i32> for MySum<i32> {
```
weiznich, line 632: Any reason not to just have 
```suggestion
/// struct MySum { sum: i32 }
///
/// # #[cfg(feature = "sqlite")]
/// impl SqliteAggregateFunction<i32> for MySum {
```

here?

diesel_derives/src/lib.rs
```diff
@@ -597,6 +597,174 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
...
+///     if let Some(max_semeseter_avg) = result {
+///         println!("The largest semester average is: {}", max_semeseter_avg);
+///     }
+///
+/// #    assert_eq!(Some(90f32), result);
+///     Ok(())
+/// }
+/// ```
+///
+/// Custom aggregate functions must return a `Nullable`.
```
weiznich, line 749: Maybe we should include some reasoning why `Nullable` needs to be returned, so something like 
```suggestion
/// Custom aggregate functions must return a `Nullable`, to handle cases where an empty set is passed to the aggregating function.
```

weiznich:
I've just noticed that this should probably have a changelog entry. Could you add one as part of the PR?

CHANGELOG.md
```diff
@@ -26,10 +26,12 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * The `MacAddr` SQL type can now be used without enabling the `network-address`
   feature.
-
+
```
thekuom, line 29: github why 🤦‍♂

thekuom:
Hey @weiznich just pinging to see if this is ready to go :)

weiznich:
@thekuom Due to eastern on the weekend I would like to wait another few days for a potential review, otherwise this is fine from my side.

diesel/src/sqlite/mod.rs
```diff
@@ -13,3 +13,21 @@ pub mod query_builder;
...
```
weiznich, line 32: I had recently another thought about this, I think explicitly returning `Option<Self::Output>` here as actually the better solution, as there could be reasons not to return `NULL` for an empty set. A trivial example is the `count(something)` sql function, that returns `0` for empty sets. Defining such a function would be not possible with the current implementation. Based on that I propose the following change:

```suggestion
    fn finalize(aggregator: Option<Self>) -> Option<Self::Output>;
```

this allows that users set a custom default value or just return `None` for the `None` case by using the `?` operator on the `aggregator`

thekuom: Makes sense to me! I'll try to get to that this weekend.

thekuom: @weiznich I ran into some trouble with types. Please take a close look at commit [44ffaa4](https://github.com/diesel-rs/diesel/pull/2325/commits/44ffaa46b780ff90c9b91f4e0e6b442d0e27c762). The change I'm most curious about is in `src/type_impls/option.ts`. This change was needed so I could pass `Option<Ret>` to `process_sql_function_result`, since at the time, only `Nullable` implemented `ToSql` for `Option`, but there wasn't a way for me to tell the compiler that `RetSqlType` is `Nullable`. If I don't make this change, I would need to do this in `src/sqlite/connection/raw.ts` in the `run_aggregator_final_function`:

```rust
let result = A::finalize(aggregator);

match result {
  Some(r) => match process_sql_function_result::<RetSqlType, Ret>(r) {
    ...
  },
  None => {
    // The same code that exists in process_sql_function_result
  }
}
```

This felt like something I shouldn't have to do, but I was running into compile errors. I think the change I made is safe because only `Nullable<ST>` satisfies `T: IntoNullable<Nullable = T>`. Please let me know if there is something or some approach I am missing!

thekuom:
Unable to reproduce the nightly linting checks that are failing in the CI, but it appears they are working on it: https://github.com/rust-lang/rust/pulls?q=is%3Apr+unused_braces

diesel/src/sqlite/mod.rs
weiznich: I think it should be possible to define a `count` function as following:

```rust
sql_function!(#[aggregate] fn count(x: Integer) -> Integer);
```

This implies that my proposed change does not completely make sense. We need to return `Self::Output` instead of `Option<Self::Output>` as we may want to return a plain value. 

I've pushed this as https://github.com/diesel-rs/diesel/pull/2325/commits/5ffc26bb5e4fa84ecd3899238ff82e158add3a6c to this branch. Would be great if you could have a final look at that change, otherwise I think we are finished here. I will also try to do a final review in the next few days and after that merge this PR.

thekuom: This looks good to me! Thanks for all your help 🎉

weiznich:
@thekuom Thanks again for your great work here :+1:

PR #2329, opened 2020-03-16
infer_schema: Allow configuration of column sort order

dbrgn:
Current situation: The sort order of the fields in the schema must match the order of fields in the model struct. When the ordering of fields in the database changes and the schema is re-generated, compilation fails if the user doesn't also update the order of the model fields.

Current default sort order is ordinal position in the table. The problem arises if a migration removes some fields that are not at the end of the table. If a table has the fields "a, b, c" and a migration removes column b, then the reverse migration that re-adds this column will lead to a table with the fields "a, c, b".

This means that simply running a migration, reverting it and then re-generating the schema will result in non-compiling code.

This PR adds a configuration field called `column_sorting` and a CLI parameter called `column-sorting`. It can be set to either `ordinal_position` (default) or `name`. If set to `name`, then the column fields will be sorted by name. This ensures stable sorting even if columns are removed and re-added.

You might also want to consider making the name based sorting the default, although that would be a breaking change.


diesel_cli/src/main.rs
```diff
@@ -405,6 +405,14 @@ fn run_infer_schema(matches: &ArgMatches) -> Result<(), Box<dyn Error>> {
         config.with_docs = true;
     }
 
+    if let Some(sorting) = matches.value_of("column-sorting") {
+        match sorting {
```
weiznich: We probably want to use claps buildin enum support here, instead of doing a manual matching?

diesel_cli/tests/print_schema.rs
```diff
@@ -31,8 +31,16 @@ fn run_infer_schema_exclude() {
...
-    test_print_schema("print_schema_order", vec!["--with-docs"]);
+fn run_infer_schema_table_order() {
+    test_print_schema("print_schema_table_order", vec!["--with-docs"]);
+}
+
+#[test]
+fn run_infer_schema_column_order() {
+    test_print_schema(
+        "print_schema_column_order",
+        vec!["--column-sorting", "name"],
```
weiznich: We pass sorting related arguments here and set them also in the config file below. I think we should have two separate tests for those cases.

dbrgn: That's how all the other tests do it as well. We would need to refactor `test_print_schema` to be able to split up config and argument tests (since `test_print_schema` calls the config test after testing the arguments).

dbrgn:
> That said I'm not finally sold to this yet, as it adds another option.

Do you have another suggestion on how to solve the problem described in the PR description? Right now I have the situation that simply applying and reverting a migration makes my code not compile anymore.

diesel_cli/src/main.rs
dbrgn: How would that work? `possible_values` [only takes strings](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.possible_values). For other commands that currently use `possible_values` (like MIGRATION_FORMAT) we currently also match on the string.

Clap 3 will have `structopt` integrated, that could become handy for less stringly typed argument parsing. But it's not released yet.

weiznich: https://docs.rs/clap/2.33.0/clap/macro.arg_enum.html

weiznich:
@TaKO8Ki If you look for another PR that could get a rebase, then have a look at this one.

TaKO8Ki:
@weiznich 
Ok! I'll tackle this PR 👍

PR #2332, opened 2020-03-18
Schema unmappable chars (cont.)

### LINKED ISSUE #2168: Generated schema fails to compile when tables and field names contain spaces
### Versions

- **Rust:** 1.37.0
- **Diesel:** 1.4.2
- **Database:** Postgres 11.5 (postgres:latest as of the time of writing)
- **Operating System** macOS 10.14.6

### Feature Flags

- **diesel:** `"postgres"` and `"chrono"`.

## Problem Description

I am working with a legacy production database for an existing product within which there are a number of tables and columns with names which don't map to rust identifiers, eg:

```sql
CREATE TABLE IF NOT EXISTS "user-has-role" (
    "user" integer NOT NULL,
    role integer NOT NULL,
    id integer NOT NULL,
    "created at" timestamp without time zone DEFAULT now() NOT NULL,
    "expiry date" timestamp without time zone
);
```

The schema that diesel generates fails to compile because of these invalid identifiers.

Access and manipulate a legacy database.

The generated schema should include underscored table and field names, with the corresponding `#[sql_name=...]` attribute, eg:

```
table! {
    #[sql_name="user-has-role"]
    user_has_role (id) {
        user -> Int4,
        role -> Int4,
        id -> Int4,
        #[sql_name="created at"]
        created_at -> Timestamp,
        #[sql_name="expiry date"]
        expiry_date -> Nullable<Timestamp>,
    }
}
```

The generated schema.rs contains the following:

```
table! {
    user-has-role (id) {
        user -> Int4,
        role -> Int4,
        id -> Int4,
        created at -> Timestamp,
        expiry date -> Nullable<Timestamp>,
    }
}
```

Which results in the following compile error:

```
error: Invalid `table!` syntax. Please see the `table!` macro docs for more info. `https://docs.diesel.rs/diesel/macro.table.html`
   --> graphene-repo/src/schema.rs:924:1
    |
924 | / table! {
925 | |     user-has-role (id) {
926 | |         user -> Int4,
927 | |         role -> Int4,
...   |
931 | |     }
932 | | }
    | |_^
    |
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

I see similar errors for the `joinable!` and `allow_tables_to_appear_in_same_query!` macros.

### Steps to reproduce

1. Create an empty crate.
2. Add diesel dependency.
4. Run `diesel migration generate broken_schema` and add the above SQL snippet to the migration's `up.sql`.
5. Run `diesel setup && diesel migration run`
6. Inspect `schema.rs`.

## Checklist

abreis:
I'm opening this PR to help bring #2173 over the finish line, as that PR has stalled.

The goal of the PR is to make `diesel_cli` generate valid schemas when a database's table or column names contain characters that don't map well to [Rust identifier strings](https://doc.rust-lang.org/reference/identifiers.html).

* I've expanded the scope of #2173 to also map every non-alphanumeric, non-underscore character, which will let `diesel_cli` support less-common table names such as "some::table::name". 
* It now merges adjoining pairs of underscores (`__`) that may appear after the mapping, so that the resulting identifier is more rust-like.

Credit goes to @zoewithabang for the original implementation.
Resolves #2168, closes #2173.


abreis:
I pushed a fix for the (unrelated) test failures in #2333.

weiznich:
Test failures should been fixed now on master, so can you rebase this PR?

abreis:
Thanks for the heads-up. Tests look good!

diesel_cli/src/infer_schema_internals/inference.rs
```diff
@@ -23,6 +23,38 @@ fn is_reserved(name: &str) -> bool {
         )
 }
 
+fn contains_unmappable_chars(name: &str) -> bool {
+    // Rust identifier names are restricted to [a-zA-Z0-9_].
+    !name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_')
```
weiznich: I'm probably micro optimizing this, but wouldn't be 
```suggestion
    name.chars().any(|c| !(c.is_ascii_alphanumeric() || c == '_'))
```

more performant, as it does potential do an early return in the case of an invalid identifier?

diesel_cli/src/infer_schema_internals/inference.rs
```diff
@@ -23,6 +23,38 @@ fn is_reserved(name: &str) -> bool {
...
+}
+
+pub fn rust_name_for_sql_name(sql_name: &str) -> Option<String> {
```
weiznich: Can we return a `Cow<str>` here instead of the option? That would remove all that `.unwrap_or(…)` below.

abreis: Unfortunately, a None variant returned here will be checked in `print_schema.rs` (L206) to stop diesel_cli from adding the `#[sql_name = "..."]` attribute macro header in case the SQL name is a valid Rust identifier. I'm not sure it'd be straightforward to swap out the Option here.

diesel_cli/src/infer_schema_internals/inference.rs
abreis: I believe `Iterator::all()` is also short-circuiting, so it should return early on the first invalid character as well.

diesel_cli/src/infer_schema_internals/inference.rs
abreis: (I suppose we could use `Cow::is_borrowed()` for that check, but unfortunately it's nightly-only for now)

diesel_cli/src/infer_schema_internals/inference.rs
weiznich: Oh right, that's true. (Not sure why I missed that 🤦‍♂ )

diesel_cli/src/infer_schema_internals/inference.rs
weiznich: We should be able to detect such cases by just comparing `sql_name` with `rust_name`. If they are not equal we need the attribute.

abreis: Of course, you're right. I'll make the changes!

abreis: So, my thinking was that `rust_name: Cow<str>` would either point to an owned string, meaning `sql_name` had been modified, or to `&sql_name`, meaning it was unchanged. Then it'd be a simple ` if let Cow::Owned(_) = rust_name {}` to know whether we needed to add the attribute header to the schema or not.

However, storing `rust_name` as Cow in `data_structures::ColumnDefinition` forces me to add a lifetime to it, which ends up polluting _a lot_ of the code in diesel_cli with lifetime annotations. It's not very elegant.

I then tried to store `rust_name` as a String and do the comparison as you suggested ([see this commit](https://github.com/abreis/diesel/commit/964d8e62aa86963fffe18759ee1c9944d1ed5bf0)), converting from Cow\<str\> where needed. But now I'm doing string comparisons instead of checking for an Option variant, so I'm not sure this is better.

Thoughts?

weiznich: I think that variant look better than using `Option<String>`. As `Cow<str>` would introduce lifetimes I would say go for `String` instead in all places. In the end this is no performance critical code, so just allocating the string there for having a cleaner code is fine.

diesel_cli/src/infer_schema_internals/inference.rs
```diff
@@ -23,6 +24,38 @@ fn is_reserved(name: &str) -> bool {
         )
 }
 
+fn contains_unmappable_chars(name: &str) -> bool {
+    // Rust identifier names are restricted to [a-zA-Z0-9_].
+    !name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_')
+}
+
+pub fn rust_name_for_sql_name(sql_name: &str) -> Cow<str> {
```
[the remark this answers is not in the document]
abreis: Yeah, you're right, thanks. I went a bit on auto-pilot there with the `to_string()`s.

PR #2339, opened 2020-03-24
Update to nightly-2020-04-01

martell:


JohnTitor:
Seems 2020-04-01 is missing rustfmt, I don't think compiletests requires it though. Should we set rustup profile to minimal?

weiznich:
Merged with #2363

PR #2347, opened 2020-03-31
New developer experience

Kononnable:
* Due to recent changes to postgres docker image it is now mandatory to specify db user password(https://github.com/docker-library/postgres/issues/681).
* There was a problem with `travis.yml` file indentation.
* Running tests locally required shell `BACKEND` variable configured which I didn't see anywhere mentioned in the docs. I added it to `./bin/test/`, but I'm not sure if it needs to be documented - `./bin/diesel` is also dependent on that variable.

I didn't find any recent travis builds, so I'm not sure If `travis.yml` file should still exists(or are travis builds just disabled temporary). In any way build looks green now(https://travis-ci.com/github/Kononnable/diesel/builds/157316166).

<br>

During my local environment setup I encountered SIGABORT a couple of times during test runs.
```
test types_roundtrip::mysql_types::naive_time_roundtrips ... free(): invalid next size (fast)
error: process didn't exit successfully: `/home/kononnable/repos/github/kononnable/diesel/target/debug/deps/integration_tests-92db1114560af468` (signal: 6, SIGABRT: process abort signal)
```
Is it a known problem or should I investigate this further?


bin/test
```diff
@@ -40,5 +40,6 @@ else
   (cd diesel_compile_tests && cargo test $*)
   (cd diesel_migrations/migrations_internals && cargo test $*)
   (cd diesel_migrations/migrations_macros && cargo test $*)
+  export BACKEND="mysql sqlite postgres"
```
weiznich, line 43: This is something that should explicitly set by the developer as other parts of the test script also depend on that variable. There is no default choice there. 
(And yes this could probably have some better documentation somewhere...)

Kononnable: I have to disagree. This is part of the script which test every available backend - there is no way to test only single backend at this point and there is no dependency on this variable before or after(in this script branch). Script will fail on previous commands If some backends aren't configured and since `export` statement won't override variable outside of script(and it's children) scope it's safe to use it here. Anyhow, any value without every backend included will fail on testing examples.

Anyway, I'm ok with removing this line if we decide to document this behavior(force developer to set every backend in the variable manually), or make changes in script which will allow it to run full test, but only for specified backends. I can make the changes when we agree which way to go.

weiznich: Seems like I should read the whole script and not only parts of it. Anyway that's right it does make sense to set `BACKEND` there. Another thing I've noticed while digging into this is that we should probably also check if `BACKEND` is set [here](https://github.com/diesel-rs/diesel/blob/cb333673449211e4c18cebe94b54fcbd8f73565c/examples/sqlite/test_all), [here](https://github.com/diesel-rs/diesel/blob/cb333673449211e4c18cebe94b54fcbd8f73565c/examples/postgres/test_all) and [here](https://github.com/diesel-rs/diesel/blob/cb333673449211e4c18cebe94b54fcbd8f73565c/examples/mysql/test_all) before calling `bin/diesel`. If it's not set it should be set to the backend corresponding directory. 
Would be great to add this to the PR, mind fixing that?

Kononnable: Done :)

Issue #2331, opened 2020-03-17
impl To/FromSql<Text, Sqlite> for chrono::DateTime<FixedOffset>


boustrophedon:
Would you accept PRs for
`impl FromSql<Text, Sqlite> for DateTime<FixedOffset>` and the corresponding `ToSql` where the DateTime is stored as either [rfc2822](https://docs.rs/chrono/0.4.11/chrono/struct.DateTime.html#method.parse_from_rfc2822) or (probably better) [rfc3339](https://docs.rs/chrono/0.4.11/chrono/struct.DateTime.html#method.parse_from_rfc3339) ?

weiznich:
I would be open to such an addition. RFC3339 seems to be the better variant, as it seems to be supported by sqlite. ([Documentation](https://www.sqlite.org/lang_datefunc.html))

boustrophedon:
I'm sorry, I don't understand the context of the sqlite datetime functions wrt what I am proposing.

I am proposing allowing/using a TEXT column to store chrono::DateTime serialized/deserialized in rfc3339 format. (I do agree with you that 3339 is better, in particular it's sortable) This is desirable because the sqlite DATETIME datatype is a chrono::NaiveDateTime with no timezone data.

For testing, I was planning on just copying the tests [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/types/date_and_time/chrono.rs#L322). Maybe also a few tests that create multiple FixedOffset dts from the same Utc time, and checking that they deserialize back to distinct Tzs but are equivalent to the original Utc time.

Does this make sense? If so I'll work on it this weekend.

weiznich:
That means I should probably explain a bit more in detail what I've meant above. `Sqlite` fundamentally supports only 5 data types (`Null`, `Integer`, `Real`, `Text` and `Blob`). The date time functions I've linked above all work on `Text` fields, only the encoding of required to be one of that variants listed there. That's the main reason why I've linked that page. It basically says RFC3339 style datetime values are supported, while RFC2822 style values are not supported, which makes using the RFC3339 format the logical choice here.

So for the implementation strategy: Adding tests at the linked location and providing impls above is the right solution.

boustrophedon:
Oh, you mentioned it just for compatibility purposes, not that I would actually use it in the implementation.

[...]

PR #2351, opened 2020-04-09
[draft] Implement To/FromSql for chrono::DateTime with SQLite

boustrophedon:
This PR is not complete, see discussion from #2331 regarding SelectDsl trait implementation.

### DIFF
--- a/diesel/src/sqlite/types/date_and_time/chrono.rs
+++ b/diesel/src/sqlite/types/date_and_time/chrono.rs
@@ -1,6 +1,6 @@
 extern crate chrono;
 
-use self::chrono::{NaiveDate, NaiveDateTime, NaiveTime};
+use self::chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, NaiveTime};
 use std::io::Write;
 
 use crate::backend;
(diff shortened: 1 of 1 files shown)


diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -375,4 +394,216 @@ mod tests {
...
+        let query = select(sql::<Text>("'9999-01-08t00:00:00.000000-01:00'"));
+        let res = query.get_result::<DateTime<FixedOffset>>(&connection);
+        assert!(res.is_ok(), "Error selecting datetime: {}", res.unwrap_err());
+
+        let dt = res.unwrap();
+        assert_eq!(dt, distant_future);
+    }
+
+    // #[test]
+    // /// Test that the sql select...where ordering functionality interops properly with rfc3339 datetimes.
```
weiznich, line 585: Those two tests should just "work" if we use `Timestamptz` as sql type. Making them work with `Text` as sql type is basically impossible.

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -103,12 +103,31 @@ impl ToSql<Timestamp, Sqlite> for NaiveDateTime {
     }
 }
 
+// chrono::DateTime<FixedOffset> impls
+
+impl FromSql<Text, Sqlite> for DateTime<FixedOffset> {
```
weiznich, line 108: I should have made that probably clearer in #2331, but we definitively don't want to have literally `impl FromSql<Text, Sqlite> DateTime<FixedOffset>` here, because that would:

* Remove any type safety 
* Interact poorly with the strongly typed query builder

So to clarify that a bit more, I think a better approach to handle this is to use a distinct SQL type for this. I'm not 100% sold to reusing postgres `Timestamptz` type here as I'm not sure if the semantics are similar, but that could be a solution. Alternatively we could introduce a new sql type. So the strategy should be basically similar to that one above used to implement support for `NaiveDateTime`.
Other things I've noticed while thinking about this:
* We are using [here](https://github.com/diesel-rs/diesel/pull/2351/files#diff-eff2a353ad004b64b07f479329a5076cR63-R75) some guessing for the actual timestamp format. It's probably a good idea to do that here as well with all formats that support timezone information. So we can be compatible with all sqlite supported formats and not only rfc3339 style timestamps.
* We probably want to support `DateTime<TZ: Timezone>` instead of only `DateTime<FixedOffset>`, similar to the postgres variant.

PR #2354, opened 2020-04-09
Fix a broken link to Uuid

abreis:
Fixes a broken link in https://docs.diesel.rs/diesel/pg/types/sql_types/struct.Uuid.html.


weiznich:
Thanks :+1:

PR #2361, opened 2020-04-15
Fix a socket descriptor leak after a PostgreSQL connection attempt fails

### LINKED ISSUE #2360: Failed connection attempts result in leaked file descriptors
### Versions

- **Rust:** 1.42
- **Diesel:** 1.4.4
- **Database:** Any. I've been testing with a PostgreSQL database, but I believe this affects any database type.
- **Operating System** CentOS 8

### Feature Flags

- **diesel:** ["r2d2", "mysql", "postgres", "network-address", "serde_json"]

## Problem Description
Diesel seems to leak file descriptors whenever it fails to connect to a database. If faced with a outage that is long enough or anything else that would cause failed attempts to occur, eventually, Diesel may exhaust the system's 'open files' limit (which defaults to 1024 in CentOS 8). This puts the program into an irrecoverable state, as it can no longer create new socket file descriptors and thereby never make or receive a new connection, and requires the program to be restarted to resolve the issue.

I'd like my program to be able to gracefully recover from database or network outages without manual intervention.

From my testing, if, for example, the network goes offline, and then comes back online before the open file limit is hit, everything operates as normally and connections can be established again. If the program hits the open file limit due to leaking a file descriptor every failed attempt before the network comes back online, however, then it won't ever be able to create new sockets, until its process is killed and a new one brought up.

The file descriptors should be released immediately after the connection attempt.

The file descriptors for failed connection attempts remain acquired for the entirety of the program instance's life. Nothing will get them to be released, other than restarting the program itself.

### Steps to reproduce

To preface: I'm using Diesel and r2d2 within an Actix-based web service, running on a freshly installed CentOS 8 server. r2d2 is imported into my project via Diesel's `r2d2` feature. The databases that this program talks to are on separate servers, and share a network.

- Start up a new instance of a program that uses Diesel
- Grab its process ID via `ps aux`
- Get an initial count of open files with `lsof -p <pid> | wc -l`. My program in particular starts with 90 open files.
- Here, I issued a few normal requests to get r2d2's pool to acquire some database connections, and remeasured the open file count. In my case, with the handful of different pools that would get used while processing those requests, it went up to 92 - 94, and won't go any higher than that regardless of how many requests come in. If I stop sending requests for long enough for r2d2 to release connections, the open file count will go down, as expected. The program always operates in this open file count range under normal conditions.
- Sever the connection between the program and a PostgreSQL database that it talks to. Anything that causes Diesel to throw errors seems to work; in my case I removed PostgreSQL's entry from the database server's FirewallD zone. This made the program immediately start logging "No route to host" errors from r2d2, like so:
```
ERROR r2d2 - could not connect to server: No route to host
	Is the server running on host "192.168.0.3" and accepting
	TCP/IP connections on port 5432?
```
- Use that same `lsof` command from above every few minutes, monitoring the open file count. It seems like it goes up once per failed attempt, starting from 92 open files right when the firewall was modified:
    - After the first five attempts: 97
    - 2 minutes of failed attempts: 110 open files
    - 5 minutes: 147
    - 10 minutes: 215
    - 30 minutes: 567
    - 35 minutes: 627

- Inspecting the output of `lsof` at this point shows many entries that are near identical to the following:

```
COMMAND     PID     USER   FD      TYPE     DEVICE SIZE/OFF     NODE    NAME
xx          25972    xx   245u      sock      0,9      0t0     406107 protocol: TCP
```
These entries are the leaked socket file descriptors.

- At this point, I re-added PostgreSQL's port to the firewall, which allowed connections to be created successfully. I ran the `lsof` command again after waiting 20 minutes, and it showed that the count hadn't changed from 627.

From what I saw after that network outage, Diesel will never release those open files - I have about six hours worth of logs of nothing but the following error every time it'd try to acquire a new connection every few seconds, wherein the network came back online about two hours into those logs:
```
ERROR r2d2 - could not create socket: Too many open files
```

## Checklist

kevinpoitra:
Fixes #2360.

### DIFF
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -32,6 +32,9 @@ impl RawConnection {
             }
             _ => {
                 let message = last_error_message(connection_ptr);
+				
+                unsafe { PQfinish(connection_ptr) }
+				
                 Err(ConnectionError::BadConnection(message))
             }
         }
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -32,14 +32,14 @@ impl RawConnection {
             }
             _ => {
                 let message = last_error_message(connection_ptr);
-				
+
                 if !connection_ptr.is_null() {
-                     // Note that even if the server connection attempt fails (as indicated by PQstatus), 
-                     // the application should call PQfinish to free the memory used by the PGconn object.
-                     // https://www.postgresql.org/docs/current/libpq-connect.html
-                     unsafe { PQfinish(connection_ptr) }
+                    // Note that even if the server connection attempt fails (as indicated by PQstatus),
+                    // the application should call PQfinish to free the memory used by the PGconn object.
+                    // https://www.postgresql.org/docs/current/libpq-connect.html
+                    unsafe { PQfinish(connection_ptr) }
                 }
-				
+
                 Err(ConnectionError::BadConnection(message))
             }
         }
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -33,7 +33,12 @@ impl RawConnection {
             _ => {
                 let message = last_error_message(connection_ptr);
 				
-                unsafe { PQfinish(connection_ptr) }
+                if !connection_ptr.is_null() {
+                     // Note that even if the server connection attempt fails (as indicated by PQstatus), 
+                     // the application should call PQfinish to free the memory used by the PGconn object.
+                     // https://www.postgresql.org/docs/current/libpq-connect.html
+                     unsafe { PQfinish(connection_ptr) }
+                }
 				
                 Err(ConnectionError::BadConnection(message))
             }


diesel/src/pg/connection/raw.rs
```diff
@@ -32,6 +32,9 @@ impl RawConnection {
             }
             _ => {
                 let message = last_error_message(connection_ptr);
+				
+                unsafe { PQfinish(connection_ptr) }
```
weiznich, line 36: This should either have a comment pointing to some postgres documentation stating that `PQfinish` handles null pointers on it's own or a check for a null pointer. Also adding a comment why we need to call `PQfinish` here would be a good idea (maybe just copy the documentation snipped from the issue?)
```suggestion
                if !connection_ptr.is_null() {
                     // Note that even if the server connection attempt fails (as indicated by PQstatus), 
                     // the application should call PQfinish to free the memory used by the PGconn object.
                     // https://www.postgresql.org/docs/current/libpq-connect.html
                     unsafe { PQfinish(connection_ptr) }
                }
```

kevinpoitra: Makes sense to me, committed the change

PR #2364, opened 2020-04-15
Fix transactions not closing properly when failing on commit

### LINKED ISSUE #2346: Transaction does not close properly when failing on commit
### Versions

- **Rust:** latest stable
- **Diesel:** master
- **Database:** postgres
- **Operating System** linux

### Feature Flags

- **diesel:** postgres

## Problem Description
Connection gets released to the connection pool with transaction not being closed.
Then triggering "AlreadyInTransaction: Cannot perform this operation while a transaction is open" errors.

Serializable transaction fails on transaction commit:
https://github.com/diesel-rs/diesel/blob/5369a7a3d13b1323fcac9acf9d13256f19a970a0/diesel/src/pg/transaction.rs#L276
"DatabaseError: could not serialize access due to read/write dependencies among transactions"
Connection is released to the pool (r2d2) and should be usable again.

Next requests successfully start their own transaction.

AlreadyInTransaction: Cannot perform this operation while a transaction is open

## Related issues
https://github.com/diesel-rs/diesel/issues/2123
In my case, transactions are not nested but I suspect this is the same issue.

## Solving
I suspect catching the `DatabaseError` on `commit_transaction` and calling `rollback_transaction` would solve this issue.

## Checklist

Ten0:
Resolves #2346

Connection could be made available again with a still-open transaction due to transaction commit fail (typically `SerializationFailure` or `ReadOnlyTransaction`).

This eventually triggered "AlreadyInTransaction: Cannot perform this operation while a transaction is open" errors in unrelated places.

This commit fixes this issue by catching `SerializationFailure` or `ReadOnlyTransaction` on commit and attempting a rollback.


diesel/src/connection/mod.rs
```diff
@@ -95,10 +99,19 @@ pub trait Connection: SimpleConnection + Sized + Send {
         let transaction_manager = self.transaction_manager();
         transaction_manager.begin_transaction(self)?;
         match f() {
-            Ok(value) => {
-                transaction_manager.commit_transaction(self)?;
-                Ok(value)
-            }
+            Ok(value) => match transaction_manager.commit_transaction(self) {
```
weiznich, line 102: Is there any specific reason to duplicate the error handling code here and in the postgres specific transaction implementation instead of handling it directly inside of `AnsiTransactionManager::commit_transaction`?

Ten0: No you're right that was a mistake

Ten0:
Do you think we should only use it on `COMMIT` and not on savepoints?

diesel/src/connection/transaction_manager.rs
```diff
@@ -110,12 +110,21 @@ where
...
+    /// `ReadOnlyTransaction` rollback will be attempted. If it succeeds,
+    /// original error will be returned. If it fails, rollback error will be returned.
     fn commit_transaction(&self, conn: &Conn) -> QueryResult<()> {
         let transaction_depth = self.transaction_depth.get();
         self.change_transaction_depth(
             -1,
             if transaction_depth <= 1 {
-                conn.batch_execute("COMMIT")
+                conn.batch_execute("COMMIT").or_else(|err| match err {
+                    Error::DatabaseError(DatabaseErrorKind::SerializationFailure, _)
```
weiznich, line 122: Maybe add a small note here why we have choosen to only handle those two error kinds? It should probably also state that it is fine to add other cases if it turns out to be a problem later.

diesel/src/connection/mod.rs
```diff
@@ -45,6 +45,10 @@ pub trait Connection: SimpleConnection + Sized + Send {
     /// If there is already an open transaction,
     /// savepoints will be used instead.
     ///
+    /// If the transaction fails to commit due to a `SerializationFailure` or a
+    /// `ReadOnlyTransaction` rollback will be attempted. If it succeeds,
+    /// original error will be returned. If it fails, rollback error will be returned.
+    ///
```
weiznich, line 51: Just some small wording improvements. I think they apply in the other comments too

```suggestion
    /// If the transaction fails to commit due to a `SerializationFailure` or a
    /// `ReadOnlyTransaction` a rollback will be attempted. If the rollback succeeds,
    /// the original error will be returned, otherwise the error generated by the rollback
    /// will be returned. In the second case the connection should be considered broken 
    /// as it contains a uncommitted unabortable open transaction.
```

diesel/src/connection/transaction_manager.rs
Ten0: done

PR #2367, opened 2020-04-18
add select_by

### LINKED ISSUE #2037: Pass struct to select
Can a `struct` type be used in `select` to avoid duplicating fields declared in a `struct`?

Taking a `User` as a typical use case, only some fields should be retrieved from the database table in some cases (i.e. to omit `password`, etc). In my use case this is a different `struct` from the one used to create the user, so I can't use `find` as that'll retrieve all of the fields. Instead of listing all the fields again (as this is quite verbose) it would be more succinct to pass the `struct` type to `select` to avoid repetition like so:

```
pub fn read(connection: &Connection, id: &Uuid) -> Result<User, Error> {
  Ok(users::table.select(User).find(id).first::<User>(connection)?)
}
```

Rather than:
```
pub fn read(connection: &Connection, id: &Uuid) -> Result<User, Error> {
  Ok(users::table.select((users::id, users::name, users::email)).find(id).first::<User>(connection)?)
}
```

Unless there is already a DRYer approach to this?

clouds56:
fixes #2037

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -5,6 +5,7 @@ use std::result;
 
 use crate::backend::{self, Backend};
 use crate::row::{NamedRow, Row};
+use crate::expression::Expression;
 
 /// A specialized result type representing the result of deserializing
 /// a value from the database.
@@ -254,6 +255,150 @@ where
 
 #[doc(inline)]
 pub use diesel_derives::QueryableByName;
+
+/// Trait indicating that a record can be selected and queried from the database.
+///
+/// Types which implement `TableQueryable` represent the result of a SQL query. This
+/// does not necessarily mean they represent a single database table.
+///
+/// Diesel represents the return type of a query as the struct. The purpose of this
+/// trait is to provide name of select statement and use with QueryableByName.
+///
+/// This trait can be [derived](derive.QueryableByName.html)
+///
+/// # Examples
+///
+/// If we just want to map a query to our struct, we can use `derive`.
+///
+/// ```rust
+/// # include!("doctest_setup.rs");
+/// #
+/// #[derive(Queryable, PartialEq, Debug)]
+/// struct User {
+///     id: i32,
+///     name: String,
+/// }
+///
+/// # fn main() {
+/// #     run_test();
+/// # }
+/// #
+/// # fn run_test() -> QueryResult<()> {
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// let first_user = users.first(&connection)?;
+/// let expected = User { id: 1, name: "Sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+///
+/// If we want to do additional work during deserialization, we can use
+/// `deserialize_as` to use a different implementation.
+///
+/// ```rust
+/// # include!("doctest_setup.rs");
+/// #
+/// # use schema::users;
+/// # use diesel::backend::{self, Backend};
+/// # use diesel::deserialize::Queryable;
+/// #
+/// struct LowercaseString(String);
+///
+/// impl Into<String> for LowercaseString {
+///     fn into(self) -> String {
+///         self.0
+///     }
+/// }
+///
+/// impl<DB, ST> Queryable<ST, DB> for LowercaseString
+/// where
+///     DB: Backend,
+///     String: Queryable<ST, DB>,
+/// {
+///     type Row = <String as Queryable<ST, DB>>::Row;
+///
+///     fn build(row: Self::Row) -> Self {
+///         LowercaseString(String::build(row).to_lowercase())
+///     }
+/// }
+///
+/// #[derive(Queryable, PartialEq, Debug)]
+/// struct User {
+///     id: i32,
+///     #[diesel(deserialize_as = "LowercaseString")]
+///     name: String,
+/// }
+///
+/// # fn main() {
+/// #     run_test();
+/// # }
+/// #
+/// # fn run_test() -> QueryResult<()> {
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// let first_user = users.first(&connection)?;
+/// let expected = User { id: 1, name: "sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+///
+/// Alternatively, we can implement the trait for our struct manually.
+///
+/// ```rust
+/// # include!("doctest_setup.rs");
+/// #
+/// use schema::users;
+/// use diesel::deserialize::Queryable;
+///
+/// # /*
+/// type DB = diesel::sqlite::Sqlite;
+/// # */
+///
+/// #[derive(PartialEq, Debug)]
+/// struct User {
+///     id: i32,
+///     name: String,
+/// }
+///
+/// impl Queryable<users::SqlType, DB> for User {
+///     type Row = (i32, String);
+///
+///     fn build(row: Self::Row) -> Self {
+///         User {
+///             id: row.0,
+///             name: row.1.to_lowercase(),
+///         }
+///     }
+/// }
+///
+/// # fn main() {
+/// #     run_test();
+/// # }
+/// #
+/// # fn run_test() -> QueryResult<()> {
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// let first_user = users.first(&connection)?;
+/// let expected = User { id: 1, name: "sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+pub trait TableQueryable {
+    /// The Rust type you'd like to map from.
+    ///
+    /// This is typically a tuple of all of your struct's fields.
+    type Columns: Expression;
+
+    /// Construct an instance of this type
+    fn columns() -> Self::Columns;
+}
+
+pub trait TableQueryableStmt {
+    type Columns;
+}
 
 /// Deserialize a single field of a given SQL type.
 ///
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -396,10 +396,6 @@ pub trait TableQueryable {
     fn columns() -> Self::Columns;
 }
 
-pub trait TableQueryableStmt {
-    type Columns;
-}
-
 /// Deserialize a single field of a given SQL type.
 ///
 /// When possible, implementations of this trait should prefer to use an
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -27,6 +27,7 @@ mod order_clause;
 mod returning_clause;
 mod select_clause;
 mod select_statement;
+mod select_by_statement;
 mod sql_query;
 mod update_statement;
 pub(crate) mod upsert;
@@ -43,6 +44,7 @@ pub use self::insert_statement::{
 pub use self::query_id::QueryId;
 #[doc(hidden)]
 pub use self::select_statement::{BoxedSelectStatement, SelectStatement};
+pub use self::select_by_statement::SelectByStatement;
 pub use self::sql_query::{BoxedSqlQuery, SqlQuery};
 #[doc(inline)]
 pub use self::update_statement::{
--- a/diesel/src/query_builder/select_by_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_by_statement/dsl_impls.rs
@@ -1,0 +1,369 @@
+// use super::BoxedSelectByStatement;
+use crate::associations::HasTable;
+// use crate::backend::Backend;
+use crate::dsl::AsExprOf;
+// use crate::expression::nullable::Nullable;
+use crate::expression::*;
+use crate::insertable::Insertable;
+use crate::query_builder::insert_statement::InsertFromSelect;
+use crate::query_builder::{Query, SelectQuery, SelectByStatement};
+// use crate::query_dsl::boxed_dsl::BoxedDsl;
+use crate::query_dsl::methods::*;
+use crate::query_dsl::*;
+// use crate::query_source::joins::{Join, JoinOn, JoinTo};
+use crate::query_source::joins::JoinTo;
+// use crate::query_source::QuerySource;
+use crate::sql_types::{BigInt, Bool};
+
+impl<S, STMT, Rhs, Kind, On> InternalJoinDsl<Rhs, Kind, On>
+    for SelectByStatement<S, STMT>
+where
+    STMT: InternalJoinDsl<Rhs, Kind, On>,
+{
+    type Output = STMT::Output;
+
+    fn join(self, rhs: Rhs, kind: Kind, on: On) -> Self::Output {
+        self.inner.join(rhs, kind, on)
+    }
+}
+
+impl<S, STMT, Selection> SelectDsl<Selection>
+    for SelectByStatement<S, STMT>
+where
+    Selection: Expression,
+    STMT: SelectDsl<Selection>,
+{
+    type Output = STMT::Output;
+
+    fn select(self, selection: Selection) -> Self::Output {
+        self.inner.select(selection)
+    }
+}
+
+impl<ST, S, STMT> DistinctDsl for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery<SqlType = ST>,
+    STMT: DistinctDsl,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn distinct(self) -> Self::Output {
+        SelectByStatement::new(self.inner.distinct())
+    }
+}
+
+impl<S, STMT, Predicate> FilterDsl<Predicate>
+    for SelectByStatement<S, STMT>
+where
+    Predicate: Expression<SqlType = Bool> + NonAggregate,
+    STMT: FilterDsl<Predicate>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn filter(self, predicate: Predicate) -> Self::Output {
+        SelectByStatement::new(self.inner.filter(predicate))
+    }
+}
+
+impl<S, STMT, Predicate> OrFilterDsl<Predicate>
+    for SelectByStatement<S, STMT>
+where
+    Predicate: Expression<SqlType = Bool> + NonAggregate,
+    STMT: OrFilterDsl<Predicate>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn or_filter(self, predicate: Predicate) -> Self::Output {
+        SelectByStatement::new(self.inner.or_filter(predicate))
+    }
+}
+
+use crate::query_source::Table;
+
+impl<S, STMT, PK> FindDsl<PK> for SelectByStatement<S, STMT>
+where
+    STMT: FindDsl<PK>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn find(self, id: PK) -> Self::Output {
+        SelectByStatement::new(self.inner.find(id))
+    }
+}
+
+impl<ST, S, STMT, Expr> OrderDsl<Expr> for SelectByStatement<S, STMT>
+where
+    Expr: Expression,
+    Self: SelectQuery<SqlType = ST>,
+    STMT: OrderDsl<Expr>,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn order(self, expr: Expr) -> Self::Output {
+        SelectByStatement::new(self.inner.order(expr))
+    }
+}
+
+impl<ST, S, STMT, Expr> ThenOrderDsl<Expr> for SelectByStatement<S, STMT>
+where
+    Expr: Expression,
+    Self: SelectQuery<SqlType = ST>,
+    STMT: ThenOrderDsl<Expr>,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn then_order_by(self, expr: Expr) -> Self::Output {
+        SelectByStatement::new(self.inner.then_order_by(expr))
+    }
+}
+
+#[doc(hidden)]
+pub type Limit = AsExprOf<i64, BigInt>;
+
+impl<ST, S, STMT> LimitDsl for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery<SqlType = ST>,
+    STMT: LimitDsl,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn limit(self, limit: i64) -> Self::Output {
+        SelectByStatement::new(self.inner.limit(limit))
+    }
+}
+
+#[doc(hidden)]
+pub type Offset = Limit;
+
+impl<ST, S, STMT> OffsetDsl for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery<SqlType = ST>,
+    STMT: OffsetDsl,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn offset(self, offset: i64) -> Self::Output {
+        SelectByStatement::new(self.inner.offset(offset))
+    }
+}
+
+impl<S, STMT, Expr> GroupByDsl<Expr> for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery,
+    Expr: Expression,
+    STMT: GroupByDsl<Expr>,
+{
+    type Output = STMT::Output;
+
+    fn group_by(self, expr: Expr) -> Self::Output {
+        self.inner.group_by(expr)
+    }
+}
+
+impl<ST, S, STMT, Lock> LockingDsl<Lock> for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery<SqlType = ST>,
+    STMT: LockingDsl<Lock>,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn with_lock(self, lock: Lock) -> Self::Output {
+        SelectByStatement::new(self.inner.with_lock(lock))
+    }
+}
+
+impl<ST, S, STMT, Modifier> ModifyLockDsl<Modifier>
+    for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery,
+    STMT: ModifyLockDsl<Modifier>,
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn modify_lock(self, modifier: Modifier) -> Self::Output {
+        SelectByStatement::new(self.inner.modify_lock(modifier))
+    }
+}
+
+// impl<'a, F, S, D, W, O, L, Of, G, DB> BoxedDsl<'a, DB>
+//     for SelectStatement<F, SelectClause<S>, D, W, O, L, Of, G>
+// where
+//     Self: AsQuery,
+//     DB: Backend,
+//     S: QueryFragment<DB> + SelectableExpression<F> + Send + 'a,
+//     D: QueryFragment<DB> + Send + 'a,
+//     W: Into<BoxedWhereClause<'a, DB>>,
+//     O: Into<Option<Box<dyn QueryFragment<DB> + Send + 'a>>>,
+//     L: QueryFragment<DB> + Send + 'a,
+//     Of: QueryFragment<DB> + Send + 'a,
+//     G: QueryFragment<DB> + Send + 'a,
+// {
+//     type Output = BoxedSelectStatement<'a, S::SqlType, F, DB>;
+
+//     fn internal_into_boxed(self) -> Self::Output {
+//         BoxedSelectStatement::new(
+//             Box::new(self.select.0),
+//             self.from,
+//             Box::new(self.distinct),
+//             self.where_clause.into(),
+//             self.order.into(),
+//             Box::new(self.limit),
+//             Box::new(self.offset),
+//             Box::new(self.group_by),
+//         )
+//     }
+// }
+
+// impl<'a, F, D, W, O, L, Of, G, DB> BoxedDsl<'a, DB>
+//     for SelectStatement<F, DefaultSelectClause, D, W, O, L, Of, G>
+// where
+//     Self: AsQuery,
+//     DB: Backend,
+//     F: QuerySource,
+//     F::DefaultSelection: QueryFragment<DB> + Send + 'a,
+//     D: QueryFragment<DB> + Send + 'a,
+//     W: Into<BoxedWhereClause<'a, DB>>,
+//     O: Into<Option<Box<dyn QueryFragment<DB> + Send + 'a>>>,
+//     L: QueryFragment<DB> + Send + 'a,
+//     Of: QueryFragment<DB> + Send + 'a,
+//     G: QueryFragment<DB> + Send + 'a,
+// {
+//     type Output = BoxedSelectStatement<'a, <F::DefaultSelection as Expression>::SqlType, F, DB>;
+//     fn internal_into_boxed(self) -> Self::Output {
+//         BoxedSelectStatement::new(
+//             Box::new(self.from.default_selection()),
+//             self.from,
+//             Box::new(self.distinct),
+//             self.where_clause.into(),
+//             self.order.into(),
+//             Box::new(self.limit),
+//             Box::new(self.offset),
+//             Box::new(self.group_by),
+//         )
+//     }
+// }
+
+impl<S, STMT> HasTable for SelectByStatement<S, STMT>
+where
+    STMT: HasTable,
+{
+    type Table = STMT::Table;
+
+    fn table() -> Self::Table {
+        STMT::table()
+    }
+}
+
+// no IntoUpdateTarget: if you'd like to `.into_update_target`, you should not first `.select_by`
+// impl<F, W> IntoUpdateTarget for SelectStatement<F, DefaultSelectClause, NoDistinctClause, W>
+// where
+//     SelectStatement<F, DefaultSelectClause, NoDistinctClause, W>: HasTable,
+//     W: ValidWhereClause<F>,
+// {
+//     type WhereClause = W;
+
+//     fn into_update_target(self) -> UpdateTarget<Self::Table, Self::WhereClause> {
+//         UpdateTarget {
+//             table: Self::table(),
+//             where_clause: self.where_clause,
+//         }
+//     }
+// }
+
+// FIXME: Should we disable joining when `.group_by` has been called? Are there
+// any other query methods where a join no longer has the same semantics as
+// joining on just the table?
+impl<S, STMT, Rhs> JoinTo<Rhs> for SelectByStatement<S, STMT>
+where
+    STMT: JoinTo<Rhs>,
+{
+    type FromClause = STMT::FromClause;
+    type OnClause = STMT::OnClause;
+
+    fn join_target(rhs: Rhs) -> (Self::FromClause, Self::OnClause) {
+        STMT::join_target(rhs)
+    }
+}
+
+impl<S, STMT> QueryDsl for SelectByStatement<S, STMT> {}
+
+impl<S, STMT, Conn> RunQueryDsl<Conn>
+    for SelectByStatement<S, STMT>
+{
+}
+
+impl<S, STMT, Tab> Insertable<Tab>
+    for SelectByStatement<S, STMT>
+where
+    Tab: Table,
+    Self: Query,
+{
+    type Values = InsertFromSelect<Self, Tab::AllColumns>;
+
+    fn values(self) -> Self::Values {
+        InsertFromSelect::new(self)
+    }
+}
+
+impl<'a, S, STMT, Tab> Insertable<Tab>
+    for &'a SelectByStatement<S, STMT>
+where
+    Tab: Table,
+    Self: Query,
+{
+    type Values = InsertFromSelect<Self, Tab::AllColumns>;
+
+    fn values(self) -> Self::Values {
+        InsertFromSelect::new(self)
+    }
+}
+
+// impl<'a, F, S, D, W, O, L, Of, G> SelectNullableDsl
+//     for SelectStatement<F, SelectClause<S>, D, W, O, L, Of, G>
+// {
+//     type Output = SelectStatement<F, SelectClause<Nullable<S>>, D, W, O, L, Of, G>;
+
+//     fn nullable(self) -> Self::Output {
+//         SelectStatement::new(
+//             SelectClause(Nullable::new(self.select.0)),
+//             self.from,
+//             self.distinct,
+//             self.where_clause,
+//             self.order,
+//             self.limit,
+//             self.offset,
+//             self.group_by,
+//             self.locking,
+//         )
+//     }
+// }
+
+// impl<'a, F, D, W, O, L, Of, G> SelectNullableDsl
+//     for SelectStatement<F, DefaultSelectClause, D, W, O, L, Of, G>
+// where
+//     F: QuerySource,
+// {
+//     type Output =
+//         SelectStatement<F, SelectClause<Nullable<F::DefaultSelection>>, D, W, O, L, Of, G>;
+
+//     fn nullable(self) -> Self::Output {
+//         SelectStatement::new(
+//             SelectClause(Nullable::new(self.from.default_selection())),
+//             self.from,
+//             self.distinct,
+//             self.where_clause,
+//             self.order,
+//             self.limit,
+//             self.offset,
+//             self.group_by,
+//             self.locking,
+//         )
+//     }
+// }
--- a/diesel/src/query_builder/select_by_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_by_statement/dsl_impls.rs
@@ -1,7 +1,6 @@
 // use super::BoxedSelectByStatement;
 use crate::associations::HasTable;
 // use crate::backend::Backend;
-use crate::dsl::AsExprOf;
 // use crate::expression::nullable::Nullable;
 use crate::expression::*;
 use crate::insertable::Insertable;
@@ -13,7 +12,7 @@ use crate::query_dsl::*;
 // use crate::query_source::joins::{Join, JoinOn, JoinTo};
 use crate::query_source::joins::JoinTo;
 // use crate::query_source::QuerySource;
-use crate::sql_types::{BigInt, Bool};
+use crate::sql_types::Bool;
 
 impl<S, STMT, Rhs, Kind, On> InternalJoinDsl<Rhs, Kind, On>
     for SelectByStatement<S, STMT>
@@ -120,9 +119,6 @@ where
     }
 }
 
-#[doc(hidden)]
-pub type Limit = AsExprOf<i64, BigInt>;
-
 impl<ST, S, STMT> LimitDsl for SelectByStatement<S, STMT>
 where
     Self: SelectQuery<SqlType = ST>,
@@ -135,9 +131,6 @@ where
         SelectByStatement::new(self.inner.limit(limit))
     }
 }
-
-#[doc(hidden)]
-pub type Offset = Limit;
 
 impl<ST, S, STMT> OffsetDsl for SelectByStatement<S, STMT>
 where
--- a/diesel/src/query_builder/select_by_statement/mod.rs
+++ b/diesel/src/query_builder/select_by_statement/mod.rs
@@ -1,0 +1,84 @@
+use super::{AstPass, QueryFragment};
+use crate::deserialize::{TableQueryable, TableQueryableStmt};
+use crate::backend::Backend;
+// use crate::expression::subselect::ValidSubselect;
+use crate::expression::*;
+use crate::query_builder::{QueryId, SelectQuery};
+// use crate::query_source::joins::{AppendSelection, Inner, Join};
+use crate::query_source::*;
+use crate::result::QueryResult;
+
+mod dsl_impls;
+
+/// it is like `SelectStatement` but without
+/// `Query`, `ValidSubselect` and `AppendSelection`
+#[derive(Debug, Clone, Copy, QueryId)]
+#[doc(hidden)]
+#[must_use = "Queries are only executed when calling `load`, `get_result` or similar."]
+pub struct SelectByStatement<Selection, Statement> {
+    pub(crate) select: std::marker::PhantomData<Selection>,
+    pub(crate) inner: Statement,
+}
+
+impl<S, ST> SelectByStatement<S, ST> {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(inner: ST) -> Self {
+        SelectByStatement {
+            inner, select: Default::default()
+        }
+    }
+}
+
+impl<DB, S, STMT> QueryFragment<DB> for SelectByStatement<S, STMT>
+where
+    DB: Backend,
+    STMT: QueryFragment<DB>,
+{
+    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
+        self.inner.walk_ast(out.reborrow())?;
+        Ok(())
+    }
+}
+
+impl<S, STMT> SelectQuery for SelectByStatement<S, STMT>
+where
+    S: TableQueryable,
+    STMT: SelectQuery,
+{
+    type SqlType = STMT::SqlType;
+}
+
+/// Allow `SelectStatement<S, Statement>` to act as if it were `Statement`.
+impl<S, STMT, T> AppearsInFromClause<T> for SelectByStatement<S, STMT>
+where
+    STMT: AppearsInFromClause<T>,
+{
+    type Count = STMT::Count;
+}
+
+impl<S, STMT> QuerySource for SelectByStatement<S, STMT>
+where
+    S: TableQueryable,
+    STMT: QuerySource,
+    S::Columns: SelectableExpression<Self>,
+{
+    type FromClause = STMT::FromClause;
+    type DefaultSelection = S::Columns;
+
+    fn from_clause(&self) -> Self::FromClause {
+        self.inner.from_clause()
+    }
+
+    fn default_selection(&self) -> Self::DefaultSelection {
+        S::columns()
+    }
+}
+
+// not implement AppendSelection
+
+impl<S, STMT> TableQueryableStmt for SelectByStatement<S, STMT>
+where
+    S: TableQueryable,
+{
+    type Columns = S::Columns;
+}
--- a/diesel/src/query_builder/select_by_statement/mod.rs
+++ b/diesel/src/query_builder/select_by_statement/mod.rs
@@ -1,9 +1,11 @@
 use super::{AstPass, QueryFragment};
-use crate::deserialize::{TableQueryable, TableQueryableStmt};
+use crate::deserialize::{QueryableByName, TableQueryable};
 use crate::backend::Backend;
+use crate::connection::Connection;
 // use crate::expression::subselect::ValidSubselect;
 use crate::expression::*;
 use crate::query_builder::{QueryId, SelectQuery};
+use crate::query_dsl::LoadQuery;
 // use crate::query_source::joins::{AppendSelection, Inner, Join};
 use crate::query_source::*;
 use crate::result::QueryResult;
@@ -76,9 +78,13 @@ where
 
 // not implement AppendSelection
 
-impl<S, STMT> TableQueryableStmt for SelectByStatement<S, STMT>
+impl<Conn, S, ST> LoadQuery<Conn, S> for SelectByStatement<S, ST>
 where
-    S: TableQueryable,
+    Conn: Connection,
+    S: QueryableByName<Conn::Backend> + TableQueryable,
+    Self: QueryFragment<Conn::Backend> + SelectQuery + QueryId,
 {
-    type Columns = S::Columns;
+    fn internal_load(self, conn: &Conn) -> QueryResult<Vec<S>> {
+        conn.query_by_name(&self)
+    }
 }
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -41,6 +41,8 @@ mod order_dsl;
 mod save_changes_dsl;
 #[doc(hidden)]
 pub mod select_dsl;
+#[doc(hidden)]
+pub mod select_by_dsl;
 mod single_value_dsl;
 
 pub use self::belonging_to_dsl::BelongingToDsl;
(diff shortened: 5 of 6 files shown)


diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,362 @@
...
+    type Output = STMT::Output;
+
+    fn select(self, selection: Selection) -> Self::Output {
+        self.inner.select(selection)
+    }
+}
+
+impl<ST, S, STMT> DistinctDsl for SelectByStatement<S, STMT>
+where
+    Self: SelectQuery<SqlType = ST>,
```
clouds56, line 44: 1. Should we check `SelectQuery` in every dsl?
2. Is there any better solution to check inner `SelectClause` not changed (internally).

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,362 @@
...
+    STMT: HasTable,
+{
+    type Table = STMT::Table;
+
+    fn table() -> Self::Table {
+        STMT::table()
+    }
+}
+
+// no IntoUpdateTarget: if you'd like to `.into_update_target`, you should not first `.select_by`
```
clouds56, line 257: Should we implement `IntoUpdateTarget`

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,362 @@
...
+//             table: Self::table(),
+//             where_clause: self.where_clause,
+//         }
+//     }
+// }
+
+// FIXME: Should we disable joining when `.group_by` has been called? Are there
+// any other query methods where a join no longer has the same semantics as
+// joining on just the table?
+impl<S, STMT, Rhs> JoinTo<Rhs> for SelectByStatement<S, STMT>
```
clouds56, line 276: Should we implement `JoinTo`, also `GroupBy` and `InternalJoin`, now `GroupBy` and `InternalJoin` would automatically unboxed `SelectByStatement` to `SelectStatement`

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,362 @@
...
+    Self: Query,
+{
+    type Values = InsertFromSelect<Self, Tab::AllColumns>;
+
+    fn values(self) -> Self::Values {
+        InsertFromSelect::new(self)
+    }
+}
+
+// impl<'a, F, S, D, W, O, L, Of, G> SelectNullableDsl
```
clouds56, line 321: How we interact with nullable? unboxed as `SelectStatement`?

diesel/src/deserialize.rs
```diff
@@ -255,6 +256,146 @@ where
...
+/// # fn run_test() -> QueryResult<()> {
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// let first_user = users.first(&connection)?;
+/// let expected = User { id: 1, name: "sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+pub trait TableQueryable {
```
Mingun, line 389: I think, the best name will be `SelectResult`, `QueryResult` or even `Selectable`, because as you noted, implementors of this trait not necessary represents a table

clouds56: I also think TableQueryable is not a good name.
I think `Selectable` is better, would there be any conflict (or conflict in the future)?

diesel/src/query_builder/select_by_statement/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::result::QueryResult;
+
+mod dsl_impls;
+
+/// it is like `SelectStatement` but without
+/// `Query`, `ValidSubselect` and `AppendSelection`
+#[derive(Debug, Clone, Copy, QueryId)]
+#[doc(hidden)]
+#[must_use = "Queries are only executed when calling `load`, `get_result` or similar."]
+pub struct SelectByStatement<Selection, Statement> {
```
clouds56, line 20: Should require `QueryId` for `Selection`?

diesel/src/deserialize.rs
```diff
@@ -255,6 +256,153 @@ where
...
+/// does not necessarily mean they represent a single database table.
+///
+/// Diesel represents the return type of a query as the struct. The purpose of this
+/// trait is to provide name of select statement and use with QueryableByName.
+///
+/// This trait can be [derived](derive.QueryableByName.html)
+///
+/// # Examples
+///
+/// If we just want to map a query to our struct, we can use `derive`.
```
weiznich, line 271: These examples does not match what this trait is supposed to do. (Seems like they are just copied from `Queryable`?)

Same applies for nearly all documentation below.

diesel/src/deserialize.rs
weiznich: I agree that this should have a better name, `Selectable` is not bad, but I'm not 100% sold to it yet.

diesel/src/deserialize.rs
```diff
@@ -255,6 +256,153 @@ where
...
+/// # }
+/// ```
+pub trait TableQueryable {
+    /// The Rust type you'd like to map from.
+    ///
+    /// This is typically a tuple of all of your struct's fields.
+    type Columns: Expression;
+
+    /// Construct an instance of this type
+    fn columns() -> Self::Columns;
```
weiznich, line 396: Documentation does not match the actual functionality

diesel/src/deserialize.rs
```diff
@@ -255,6 +256,153 @@ where
...
+/// let expected = User { id: 1, name: "sean".into() };
+/// assert_eq!(expected, first_user);
+/// #     Ok(())
+/// # }
+/// ```
+pub trait TableQueryable {
+    /// The Rust type you'd like to map from.
+    ///
+    /// This is typically a tuple of all of your struct's fields.
+    type Columns: Expression;
```
weiznich, line 393: Should probably be called `SelectExpression` or something, because it does not necessarily only need to contain a list of columns. For a manual implementation this could be anything in theory.

(Same for the corresponding method)

diesel/src/deserialize.rs
```diff
@@ -255,6 +256,153 @@ where
...
+
+    /// Construct an instance of this type
+    fn columns() -> Self::Columns;
+}
+
+impl<T: TableQueryable> TableQueryable for Option<T> {
+    type Columns = T::Columns;
+    fn columns() -> Self::Columns {
+        T::columns()
```
weiznich, line 402: This should probably return `T::columns().nullable()` instead? Otherwise you wont be able to load the result into a `Option<T>`

diesel/src/query_builder/mod.rs
```diff
@@ -25,6 +25,7 @@ pub mod nodes;
 mod offset_clause;
 mod order_clause;
 mod returning_clause;
+mod select_by_statement;
```
weiznich, line 28: I would rather like to have that as module in `select_statement`

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+use crate::insertable::Insertable;
+use crate::query_builder::insert_statement::InsertFromSelect;
+use crate::query_builder::{Query, SelectByStatement, SelectQuery};
+// use crate::query_dsl::boxed_dsl::BoxedDsl;
+use crate::query_dsl::methods::*;
+use crate::query_dsl::*;
+// use crate::query_source::joins::{Join, JoinOn, JoinTo};
+use crate::query_source::joins::JoinTo;
+// use crate::query_source::QuerySource;
+use crate::sql_types::Bool;
```
weiznich, line 16: Those commented out imports needs to be removed.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+
+impl<S, STMT, Rhs, Kind, On> InternalJoinDsl<Rhs, Kind, On> for SelectByStatement<S, STMT>
+where
+    STMT: InternalJoinDsl<Rhs, Kind, On>,
+{
+    type Output = STMT::Output;
+
+    fn join(self, rhs: Rhs, kind: Kind, on: On) -> Self::Output {
+        self.inner.join(rhs, kind, on)
```
weiznich, line 25: Is there any reason not to reapply the already existing select here? So return `SelectByStatement<S, STMT::Output>`?

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+use crate::query_builder::{Query, SelectByStatement, SelectQuery};
+// use crate::query_dsl::boxed_dsl::BoxedDsl;
+use crate::query_dsl::methods::*;
+use crate::query_dsl::*;
+// use crate::query_source::joins::{Join, JoinOn, JoinTo};
+use crate::query_source::joins::JoinTo;
+// use crate::query_source::QuerySource;
+use crate::sql_types::Bool;
+
+impl<S, STMT, Rhs, Kind, On> InternalJoinDsl<Rhs, Kind, On> for SelectByStatement<S, STMT>
```
weiznich, line 18: I think `Stmt` instead of `STMT` would be a more idiomatic rust style. (Applies for all similar cases in this file)

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: > 1. Should we check SelectQuery in every dsl?

I don't think that's required.

> 2. Is there any better solution to check inner SelectClause not changed (internally).

I do not think that adding a distinct "clause" to the query will change the fields returned by the query, so I'm not sure we even need to check that.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn offset(self, offset: i64) -> Self::Output {
+        SelectByStatement::new(self.inner.offset(offset))
+    }
+}
+
+impl<S, STMT, Expr> GroupByDsl<Expr> for SelectByStatement<S, STMT>
```
weiznich, line 157: Reminder for myself: I need to think a bit more about the interaction of this feature with the newly merged foundation for true group by support.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+    SelectByStatement<S, STMT::Output>: SelectQuery<SqlType = ST>,
+{
+    type Output = SelectByStatement<S, STMT::Output>;
+
+    fn modify_lock(self, modifier: Modifier) -> Self::Output {
+        SelectByStatement::new(self.inner.modify_lock(modifier))
+    }
+}
+
+// impl<'a, F, S, D, W, O, L, Of, G, DB> BoxedDsl<'a, DB>
```
weiznich, line 196: Any reason this is not implemented? I would just assume that this returns something like `SelectByStatement<S, BoxedSelectStatement<'a, S::Columns::SqlType, F, DB>>`

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+//             self.where_clause.into(),
+//             self.order.into(),
+//             Box::new(self.limit),
+//             Box::new(self.offset),
+//             Box::new(self.group_by),
+//         )
+//     }
+// }
+
+// impl<'a, F, D, W, O, L, Of, G, DB> BoxedDsl<'a, DB>
```
weiznich, line 225: This impl is not required as we have a explicit select clause in all cases.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: No, we do only implement `IntoUpdateTarget` for cases without explicit select clause to allow using it as "where clause" in update/delete statements.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: As commented above: 
* For `InternalJoinDsl` I do not see any reason to "unbox" it
* For `GroupByDsl` I need to think about that case a bit more, especially how it interacts with the new system to handle group by clauses.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,365 @@
...
+
+impl<S, STMT, Tab> Insertable<Tab> for SelectByStatement<S, STMT>
+where
+    Tab: Table,
+    Self: Query,
+{
+    type Values = InsertFromSelect<Self, Tab::AllColumns>;
+
+    fn values(self) -> Self::Values {
+        InsertFromSelect::new(self)
```
weiznich, line 308: I think its sufficient here to just pass `self.inner` to the `InsertFromSelect` clause, as it's the same query and noone can modify the select clause there.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: Unboxing would be one solution, the other would be to return `SelectByStatement<Option<S>, Stmt::Output>` here.

diesel/src/query_builder/select_by_statement/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub(crate) inner: Statement,
+}
+
+impl<S, STMT> QueryId for SelectByStatement<S, STMT>
+where
+    STMT: QueryId,
+{
+    type QueryId = SelectByStatement<(), STMT::QueryId>;
+    const HAS_STATIC_QUERY_ID: bool = STMT::HAS_STATIC_QUERY_ID;
+}
```
weiznich, line 31: Is this is literally the same sql query as `STMT` we should just forward both to the impls of `STMT`, so something like this should be OK:
```suggestion
impl<S, STMT> QueryId for SelectByStatement<S, STMT>
where
    STMT: QueryId,
{
    type QueryId = STMT::QueryID;
    const HAS_STATIC_QUERY_ID: bool = STMT::HAS_STATIC_QUERY_ID;
}
```

To explain that a bit more: `QueryId` is used for the internal stmt caching mechanism. For static queries in some cases a static compile time id is used as stmt cache key. As `STMT` and `SelectByStatemet<S, STMT>` are the same query at SQL level they can share the same key in the statement cache. This allows to use the already cashed statement if somecone constructs the query normally via a explicit `select` in one case, and with a `select_by` in the other case. As the generate sql + statement is the same, this is a quite nice optimization in my opinion. 

(We should have probably something along this lines as comment here, to explain why this makes sense.)

diesel/src/query_builder/select_by_statement/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl<ST, S, STMT> SelectQuery for SelectByStatement<S, STMT>
+where
+    S: TableQueryable,
+    S::Columns: Expression<SqlType = ST>,
+    STMT: SelectQuery<SqlType = ST>,
+{
+    // TODO: check SelectByClause
+    type SqlType = ST;
+}
```
weiznich, line 61: Just noting: That's the super explicit variant. Just forwarding it to `STMT::SqlType` would also be fine.

diesel/src/query_builder/select_by_statement/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        S::columns()
+    }
+}
+
+// not implement AppendSelection
+
+impl<Conn, S, ST> LoadQuery<Conn, S> for SelectByStatement<S, ST>
+where
+    Conn: Connection,
+    S: QueryableByName<Conn::Backend> + TableQueryable,
```
weiznich, line 94: We should require `Queryable<ST::SqlType, Conn::Backend>` here instead of `QueryableByName` as mapping should be done by index and not by field name.

diesel/src/query_dsl/mod.rs
```diff
@@ -287,6 +291,127 @@ pub trait QueryDsl: Sized {
         methods::SelectDsl::select(self, selection)
     }
 
+    // SELECTBY_TODO: Needs usage example and doc rewrite
```
weiznich, line 294: Just for a later review: Doc changes are required.

diesel_derives/src/queryable_by_name.rs
```diff
@@ -38,8 +38,36 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
 
     let (impl_generics, _, where_clause) = generics.split_for_impl();
 
+    let impl_table_queryable: Option<syn::Item> = if model.has_table_name_attribute() {
```
weiznich, line 41: The more I think about this, the more I'm the opinion that this should be a separate derive for the following reasons:

* We do also need a `derive(Queryable)` on the same type otherwise (by having a separate derive, we can just also implement `Queryable` there)
* How would we handle the case of deriving `QueryableByName` where only `#[sql_type = "Foo"]` is given? I think just ignoring everything without table name attribute creates a wired edge case.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: I use `SelectByQuery` to ensure inner Columns invariant. We don't force inner STMT to be SelectStatement

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: Sorry, I just read the word `Join` not realize it is **Internal**

diesel/src/deserialize.rs
clouds56: Yes, I haven't start documents stuff. Would review all documents later

diesel/src/deserialize.rs
clouds56: last time I tried to use Nullable it blames Nullable is not Expression, I'm not sure where's wrong and was not clear about how it interacts with Option, would give another attempt.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: Could you explain a little more about what BoxedDsl is and why we need it?
In general I'd like not to restrict `Stmt` to `SelectStatement`, unless there's a good reason.

diesel/src/query_builder/select_by_statement/mod.rs
clouds56: Thanks for kindly explanation!

diesel_derives/src/queryable_by_name.rs
clouds56: Fair enough, we could also benifit from separate `derive` for join scenario.

clouds56:
Is there any suggestion for the trait/derive name instead of `TableQueryable`?
@Mingun suggests `Selectable` and `SelectResult`
I just comes around mind `QueryableWith` (and maybe we should use `select_with`), but it seems in rust `with` implies a closure [`RefCell::replace_with`](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.replace_with), or `QueryableByStruct`?

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: What I worried about happens, I tried my best to avoid `impl AsQuery for SelectByStatement`, but `InternalJoinDsl` requires it.
Once I impl the `AsQuery` (even with where clause), the 2 impls for LoadQuery would conflict
https://github.com/diesel-rs/diesel/blob/be0854094c79ae99ba3c893242a4be358ac0b0c5/diesel/src/query_dsl/load_dsl.rs#L21-L32
https://github.com/diesel-rs/diesel/blob/e3b5faf344b004052326460b2ebafe0d0b66664a/diesel/src/query_builder/select_by_statement/mod.rs#L102-L115
I think any solution would requires a breaking change here.

diesel/src/deserialize.rs
clouds56: Why not just `Expression`? Would `SelectExpression` be confused with trait `SelectableExpression`?

clouds56: BTW, I think any expression could be called "column" in `SELECT` statement, at least in [postgres](https://www.postgresql.org/docs/current/sql-select.html)
> Just as in a table, every output column of a SELECT has a name.
> An output column's name can be used to refer to the column's value in ORDER BY and GROUP BY clauses
> In more complex cases a function or type name may be used, or the system may fall back on a generated name such as ?column?.

Just like every item returned from a query called rows.

diesel/src/deserialize.rs
weiznich: There are two `Nullable` types in diesel, one in `diesel::expressions` and one in `diesel::sql_types`. Just try the other one. 
(I tend to use the wrong one quite often by myself.)

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: `BoxedDsl` is a way to concert a static query ast into something that could be chaged dynamically, i.e. where you can conditionally append a filer or something like that. Given the normal `SelectStatement` this is not possible, because each `QueryDsl` method return a new slightly changed type. For a `BoxedSelectStatement` this is different, as most methods will not change the type returned by such a method. This comes at a certain cost:
- `BoxedSelectstatement` requires a heap allocation per `QueryDsl` method call, whereas the normal `SelectStatement` uses zero sized static types in the most cases
- Building the corresponding sql will require a dynamic dispatch per query clause for `BoxedSelectStatements`, whereas it is mostly done at compile time for normal `SelectStatement`
- For normal `SelectStatement` we can do quite clever static compile time statement cashing based on the type ids, as for some/most cases the SQL will not change without a type change. This allows to skip building the sql at all if the query is already in the statement cache. `BoxedSelectStatement` cannot do that, as the same type could represent different sql. Therefore it needs to build the sql in all cases, as it is used as key into the statement cache. 

So for implementing `BoxedDsl` only for `Stmt` = `SelectStatement`: There are just no other cases in diesel that are useful, but we can probably also go with an generic impl here if possible.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: See the other comment, the resolution would be
1. has another maker `SelectByIndex` and change the wildcard `LoadQuery` implement depends on `SelectByIndex`.
2. looser the `Output` of `InternalJoinDsl` not depends on `AsQuery`
3. leave it as is until the [specialization](https://doc.rust-lang.org/unstable-book/language-features/specialization.html) stable.

diesel/src/deserialize.rs
clouds56: Changed to `QueryableByColumn`.
If there's some better name, welcome to propose.

diesel/src/deserialize.rs
clouds56: Basically done.

diesel/src/query_builder/mod.rs
clouds56: Why? I think `select_by_statement` is independent with `select_statement`, there is not any restriction that inner `Stmt` should be a `select_statement`.
And I think it is also large enough to have its own folder. If as a module, maybe we would have `diesel/src/query_builder/select_statement/select_by_statement/dsl_impls.rs`? This is unlike `BoxedSelectStatement` which is the same thing as `SelectStatement`.
If you still think subdir is better I could transfer it.

clouds56:
I resolved some comments, not necessarily mean they are already **resolved**, feel free to reopen any if you'd like to.

I've done some basically documents writing, but since I'm not native English speaker, please carefully review it and make suggestion.

I'm also thinking about the possibility of making `SelectByQuery` the default implementation of `LoadQuery` (which is more type safe as it contains column information) and the legacy `SelectStatement` via a new method `.select_by_index()`. This is a breaking change and I hope we could see it in diesel 3.0.
* both `users.select((id, name)).load::<User>()` and `users.select_by::<User>().load::<User>()` also requires `User` have `QueryableByColumn` with `Columns = (users::id, users::name)`
* the legacy `load_by_index` becomes `users.select_by_index((id, name)).load::<(u32, String)>()` or `users.select((id, name)).load_by_index::<(u32, String)>()`
* `.select()` and `.select_by()` would output `SelectStatment` while `.select_by_index()` would output `SelectByIndexStatement` which is a wrapper of `SelectStatment`
* migrations to 3.0 would be like add derive `QueryableByColumn` to the struct or change `select` to `select_by_index` or change `load` to `load_by_index`, both would work.

clouds56:
How it interacts with complex sql?
Maybe it could be like this:
```rust
#[derive(QueryableByColumn)]
#[table_name = "posts", args = "Args"]
struct Post {
  #[table_name = "users", column_name = "id"]
  user_id: i32, // users::id
  #[column_name = "id"]
  post_id: i32, // posts::id
  #[table_name = "users"]
  name: String, // users::name
  title: String, // posts::title
  #[sql_type = "Bool", expression = "users::name.eq(posts::title)", tag = "name_eq_title"]
  is_self: bool,
  #[sql_type = "Text", expression = "users::hair_color.concat(lit_ish)"]
  hair_color: String,
  #[expression = "dsl::now"]
  current_time: NativeDateTime, // dsl::now
}

#[derive(QueryableByColumnArgs)]
struct PostArgs {
  #[sql_type = "Text", tag = "lit_ish"]
  suffix: String
}
```
generates select clause like this
```sql
SELECT
  users.id as user_id,
  posts.id as post_id,
  users.name as name,
  posts.title as title,
  users.name = posts.title as name_eq_title,
  CONCAT(user, $1) as hair_color,
  NOW() as current_time,
FROM ...
[$1 = `suffix.to_sql()`]
```

1. the tag name is default field name, which could be used as subquery/insert_into_table.
2. `current_time` has expression `dsl::now` it would be used as expression as is
3. `is_name` has expression `users::name.eq(posts::title)`, but **I have no idea how to get the type of the expression** (in order to get `SelectableExpression` and `ValidGroup` information).
4. `hair_color` has external args `lit_ish`, which is given from `PostArgs`, we could use `select_by_with_args` to build it.

```rust
user.select_by_with_args::<Post>(PostArgs { suffix: "ish".into() }).load::<Post>()
```

Maybe we have to write:
```rust
#[derive(AsColumn)]
#[expression = "users::name.eq(posts::title)"]
#[expression_type = "diesel::dsl::Eq<users::name, posts::title>"]
struct name_eq_title;
```

diesel/src/deserialize.rs
Mingun: I don't like that name because it's completely obscure. What do `ByColumn` mean? Looks as if usual `Queryable` doesn't query columns.

Maybe `AutoQueryable`? `Auto` means, that struct fields automatically maps to table columns. Although there may be other, unwanted connotations. `Selectable` in this sense a much better. It is similar to `Queryable`, but different.

One other variant is `Mapped`, that means that struct is mapped to some table or view.

Or even just `TableOrView`.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: It's not clear for me why `AsQuery` is required for `InternalJoinDsl`. Could you elaborate that a bit?

diesel/src/deserialize.rs
weiznich: I do not like `QueryableByColumn` either, `Selectable` sounds better for me.

That said: This is a discussion that does not really matter till the actual feature implementation is finished.

weiznich:
I would like to add some general comments about my vision of this feature here as response to @clouds56 general comments above.

> I'm also thinking about the possibility of making SelectByQuery the default implementation of LoadQuery (which is more type safe as it contains column information) and the legacy SelectStatement via a new method .select_by_index(). This is a breaking change and I hope we could see it in diesel 3.0.

I would rather like to finish the implementation of diesel 2.0 first. After that we do not have any planes to release diesel 3.0 in near future. At least for me including such a change is not a option for diesel 2.0 and everything after that is unclear now. So I wouldn't want to make any planes about this yet. Basically, if this feature turns out to be as useful as some expect, we can adjust the documentation in such a way that using this combination is the preferred way, while other variants via raw select clauses + `Queryable` are options for advanced use cases. 


About the proposal for complex queries:

I would really like to keep this feature as compact as possible. Allowing to specify complex expressions as part of this derive feels like way to much magic for me. In my opinion diesel itself is not the right place for such a beast, especially since this could then simply be done by a third party crate. The only option that could be useful is probably allowing to specify also a table name per field for queries involving joins.

diesel/src/query_dsl/mod.rs
```diff
@@ -287,6 +291,134 @@ pub trait QueryDsl: Sized {
...
+    /// let all_data = join.load::<(User, Option<Post>)>(&connection)?;
+    /// let expected_data = vec![
+    ///     (User::new(1, "Sean"), Some(Post::new(post_id, 1, "Sean's Post"))),
+    ///     (User::new(2, "Tess"), None),
+    /// ];
+    /// assert_eq!(expected_data, all_data);
+    ///
+    /// // Since `PostTitle` is on the right side of a left join, `Option` is
+    /// // needed.
+    /// let names_and_titles = join.select_by::<(UserName, Option<PostTitle>)>()
```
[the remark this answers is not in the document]
clouds56: Yes, you have to type for both if it could not be inferred by context, theoretically, you could `.select_by::<A>()` while `.load::<B>()` if and only if `A` has the same `SelectExpression` as `B`.

diesel/src/query_builder/select_by_statement/dsl_impls.rs
clouds56: https://github.com/diesel-rs/diesel/blob/00f7de703aea02659bddad8d168b9fe6d4aee276/diesel/src/query_dsl/join_dsl.rs#L5-L11
I don't know why either, but it is required by the trait.

diesel/src/deserialize.rs
clouds56: It is not "Auto", the column name could be specified by `#[column_name = "id"]` while `Mapped` and `TableOrView` seems a little unrelated.
I think `Queryable` do not query columns, they query index. `Queryable` query index, `QueryableByName` query names, and `QueryableByColumn` query columns. (but we would use term "SelectExpression" instead of "Column", so it seems `Selectable` more close to `SelectExpression`)
While I worry about `Selectable` would be confused with `SelectableExpression`. And it might also have more opportunity to conflict with other top level ident.

diesel/src/deserialize.rs
clouds56: Sure. But just curious, I've done a search `/select.*expression/` and it only shows `SelectableExpression` which I think have a different meaning. Where is "other place"?
I never consist on naming, but I just want some name with consistence.

diesel/src/deserialize.rs
clouds56: Anyway, I'd like to change the trait name and change `Columns` to `SelectExpression` at the last second, I don't have a refactor tools and it is much easier to have `QueryableByColumn` to be `sed` than `Selectable` :)

diesel/src/query_builder/select_by_statement/dsl_impls.rs
```diff
@@ -0,0 +1,273 @@
...
+    Stmt::Output: SelectByQuery<Columns = CL>,
+{
+    type Output = SelectByStatement<S, Stmt::Output>;
+
+    fn or_filter(self, predicate: Predicate) -> Self::Output {
+        SelectByStatement::new(self.inner.or_filter(predicate))
+    }
+}
+
+use crate::query_source::Table;
```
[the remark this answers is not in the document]
clouds56: just note, this also happens in `select_statement/dsl_impls.rs`

diesel/src/query_dsl/mod.rs
```diff
@@ -287,6 +291,134 @@ pub trait QueryDsl: Sized {
...
+    ///
+    /// // By default, all columns from both tables are selected
+    /// let all_data = join.load::<(User, Option<Post>)>(&connection)?;
+    /// let expected_data = vec![
+    ///     (User::new(1, "Sean"), Some(Post::new(post_id, 1, "Sean's Post"))),
+    ///     (User::new(2, "Tess"), None),
+    /// ];
+    /// assert_eq!(expected_data, all_data);
+    ///
+    /// // Since `PostTitle` is on the right side of a left join, `Option` is
```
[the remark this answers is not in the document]
clouds56: [select](http://docs.diesel.rs/diesel/query_dsl/trait.QueryDsl.html#method.select) does so.

diesel/src/query_builder/mod.rs
clouds56: Just a note: I don't see why we couldn't have `InsertStatement`, `UpdateStatement` and even `DeleteStatement` behind the scenes thanks to `returning` clause. Or shall we create `InsertReturningByStatment`, `UpdateReturningByStatement` and `DeleteReturningByStatement`?

diesel/src/deserialize.rs
```diff
@@ -258,17 +258,23 @@ pub use diesel_derives::QueryableByName;
...
 /// This trait can be [derived](derive.QueryableByColumn.html)
 ///
+/// [`SelectByQuery`]: ..//query_builder/trait.SelectByQuery.html
+/// [`load`]: ../query_dsl/trait.RunQueryDsl.html#method.load
+/// [`select_by`]: ../query_dsl/methods/trait.SelectByDsl.html#tymethod.select_by
+///
 /// # Examples
 ///
-/// If we just want to map a query to our struct, we can use `derive`.
+/// If we just want to construct a query using our struct, we can use `derive`.
```
weiznich, line 277:
```suggestion
/// If you just want to construct a select clause using an existing struct, you can use `#[derive(QueryableByName)]`.
```
(Eventually with link to the actual derive macro.)

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
weiznich: I would be quite surprised if something like `.select_by::<A>().load::<B>()` works as the corresponding `LoadDsl` impl clearly states that
```rust
impl<Conn, S, Stmt> LoadQuery<Conn, S> for SelectByStatement<S, Stmt>
```
`A` (The first type parameter for `SelectByStatement`) and `B` (the second type parameter for `LoadQuery`) needs to be the same. The only question is: Is rustc's type inference able to figure that out through the layer of `RunQueryDsl` and other abstractions or does it require both annotations.

diesel/src/query_dsl/mod.rs
```diff
@@ -287,6 +291,133 @@ pub trait QueryDsl: Sized {
         methods::SelectDsl::select(self, selection)
     }
 
+    /// Adds a `SELECT` clause to the query using based on a struct implementing [QueryableByColumn].
```
weiznich, line 294: [`QueryableByColumn`] should be an link in that case :wink:

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
weiznich: That's an argument :+1:

diesel/src/query_builder/select_by_statement/dsl_impls.rs
weiznich: I cannot answer why it's there, because removing it does not have any big influences ([That one](https://github.com/diesel-rs/diesel/blob/00f7de703aea02659bddad8d168b9fe6d4aee276/diesel/src/query_dsl/join_dsl.rs#L29) needs to be removed too, then)

diesel/src/deserialize.rs
weiznich: In my understanding `SelectableExpression` says that something can be part of the select clause, where `SelectExpression` says it will be part of the select clause. So both refer to the same clause and both are expressions, while one states a possibility, while the other states a fact. 
That said: I'm also no native speaker, so this could be my non-prefect English, so if someone has other interpretations here I'm happy to be corrected.

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
clouds56: I'm wrong, `.select_by::<A>().load::<B>()` would not work, I think compiler could infer `A` that based by my test (i.e. `let _ = users.select_by::<A>().load()` compiles).

clouds56:
I'm going to change `diesel::deserialize::QueryByColumn` to `diesel::expression::Selectable`
```rust
pub trait Selectable {
    /// The expression you'd like to select.
    ///
    /// This is typically a tuple of all of your struct's fields.
    type SelectExpression: Expression;

    /// Construct an instance of the expression
    fn select_expression() -> Self::SelectExpression;
}
```
Is there any suggestions? Or if `Selectable<SelectExpression = SE>` seems too verbose shall we
```rust
pub trait Selectable {
    /// The expression you'd like to select.
    ///
    /// This is typically a tuple of all of your struct's fields.
    type Expression: Expression;

    /// Construct an instance of the expression
    fn new_expression() -> Self::Expression;
}
```
Considering `::Output` is used everywhere.
A quick search "Expression" in type def: `/type [a-zA-Z]*Expression[a-zA-Z]*[:;]/`
* `ValidGroupByClause::Expressions` (note "s"),
* `AsArrayExpression::Expression` (with `as_expression`), `AsExpressionList::Expression` (with `as_expression_list`), `AsExpression::Expression` (with `as_expression`)
* `AsInExpression::InExpression` (with `as_in_expression`),

Any other name also welcome.

weiznich:
Using `Expression` as associated type name seems like a good idea :+1:

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
weiznich: I mean we could make `.select_by::<A>().load::<B>()` working at the cost of requiring both type annotations in all cases, but I think the simplicity of just having one type annotation there is a good argument for not doing this.
Beside of that I think it should even work if you don't specify both type arguments as long as the return type can be inferred (for example by returning the result from a function, or by just having a type annotation on the let binding). It would probably be a good idea to have a "test case" for all those different variants in the existing test cases, to see if those work and to ensure they won't break in the future.

clouds56:
Trying to restart CI but fails. It seems fetch repo in CI failed hours ago.

weiznich:
I think github itself has issues today. I will restart the build manually tomorrow.

diesel/src/query_builder/select_statement/select_by/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn from_clause(&self) -> Self::FromClause {
+        self.inner.from_clause()
+    }
+
+    fn default_selection(&self) -> Self::DefaultSelection {
+        S::new_expression()
+    }
+}
+
+impl<Conn, S, Stmt> LoadQuery<Conn, S> for SelectByStatement<S, Stmt>
```
weiznich, line 128: I'm not sure if it would be meaningful to also implement `ExecuteDsl`?

diesel_derives/src/lib.rs
```diff
@@ -393,6 +394,51 @@ pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
...
+///
+/// If the name of a field on your struct is different than the column in your
+/// `table!` declaration, or if you are deriving this trait on a tuple struct,
+/// you can annotate the field with `#[column_name = "some_column"]`. For tuple
+/// structs, all fields must have this annotation.
+///
+/// If a field is another struct which implements `Selectable`,
+/// instead of a column, you can annotate that struct with `#[diesel(embed)]`.
+/// Then all fields contained by that inner struct are loaded into
+/// the embedded struct (not necessary the same table).
```
weiznich, line 414:
```suggestion
/// If a field is another struct which implements `Selectable`,
/// instead of a column, you can annotate that struct with `#[diesel(embed)]`.
/// Then all fields contained by that inner struct are selected as separate tuple. Fields 
/// from a inner struct can come from a different table, as long as the select clause is valid 
/// current query.
```

As `Selectable` does not load anything we should find another formulation here. Maybe just having one or two examples here how the select clause would look like can be helpful?

diesel/src/query_builder/mod.rs
```diff
@@ -127,6 +127,17 @@ pub trait SelectQuery {
     type SqlType;
 }
 
+/// Indicates that a type has specific `SELECT` clause, and could load by columns.
+///
+/// This trait is implemented for select_by statements, with columns information rather
+/// than only SQL type compared to `SelectQuery`. A statement implementing `SelectByQuery`
+/// is not necessary a select statements, it might also requires `SelectQuery` in order to
+/// be able to be executed.
+pub trait SelectByQuery {
```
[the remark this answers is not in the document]
clouds56: It is required by `LoadQuery` (in my implement), it is ensuring the `Expression` of `SelectByStatement` is the same as underlying `SelectStatement`.

diesel/src/query_builder/select_statement/select_by/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+/// When using `SelectDsl`/`SelectByDsl` would intendedly unbox/rebox
+/// Other operation should keep Selection invariant.
+#[doc(hidden)]
+#[must_use = "Queries are only executed when calling `load`, `get_result` or similar."]
+pub struct SelectByStatement<Selection, Statement> {
+    pub(crate) select: PhantomData<Selection>,
+    pub(crate) inner: Statement,
+}
+
+impl<S, Stmt> fmt::Debug for SelectByStatement<S, Stmt>
```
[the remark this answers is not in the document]
clouds56: I'm not sure, but the `inner` derives this.
https://github.com/diesel-rs/diesel/blob/bebdce636532465f2d6f6d0a2a0d5913f6d7b0d3/diesel/src/query_builder/select_statement/mod.rs#L39-L42
It's just like `derive` but I'd like not to let `PhantomData` hurts `Debug`.
If actually no `SelectStatement` actually impls Debug, that's fine, but as long as it derives `Debug` I'd like to forward it.

diesel/src/query_builder/select_statement/select_by/mod.rs
clouds56: I'm not sure but I think now `SelectByStatement` implements `QueryFragment` and  `QueryId`, it would automatically implement `ExecuteDsl`
https://github.com/diesel-rs/diesel/blob/bebdce636532465f2d6f6d0a2a0d5913f6d7b0d3/diesel/src/query_dsl/load_dsl.rs#L48-L57

weiznich: That seems to be correct :+1:

diesel/src/query_builder/select_statement/select_by/mod.rs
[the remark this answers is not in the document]
weiznich: Oh, I've missed that the derive is there. From that point of view it is fine to leave that impl there. 

(In practice I think there are not much actual usage for this feature as most concrete types used there do not implement `Debug`)

weiznich:
I just want to add a comment here that I've not forgot about this PR. I've just not as much time as I would like to work on open source currently. I will try to come back on this as soon as the situation is a bit better again. In the meanwhile: @diesel-rs/reviewers Feel free to suggests improvements here. 
For my point of few the following points are should be checked again: (Most of them likely need some experimentation)
* Is `select_by::<Type>` really a good API for this or should we try to do something like `.select(Type::some_static_method())` instead?
* How does this interact with `.group_by` clauses? I think that should be fine, but someone should try to write at least a compile_fail tests for some trivial cases.
* Can we avoid some of the code duplication by reusing `SelectStatement` more directly? For example by having another generic parameter?

weiznich:
With #2182 merged now we can avoid much of the code duplication here. The new basic idea is as following:

* `Selectable` remains the same
* We introduce a new select clause type, based on `Selectable` and implement the required traits. We are using a newly introduced type as `Expression::SqlType`. Somehow this type needs to reference the `Selectable` type. Let's name it `SelectBy<Q>`
* That will require that we have some  wild card impl `FromSqlRow<SelectBy<Q>, DB> for Q where Q: Selectable + Queryable<…>`

This should reduce the complete code to something like
```rust
 pub trait Selectable {
     type Selection: crate::expression::Expression;
 }

 pub struct SelectBy<ST>(std::marker::PhantomData<ST>);

 impl<ST> TypedExpressionType for SelectBy<ST> {}

 impl<ST, DB> QueryMetadata<SelectBy<ST>> for DB
 where
     DB: Backend,
 {
     fn row_metadata(_: &Self::MetadataLookup) -> Option<Vec<Self::TypeMetadata>> {
         None
     }
 }

 impl<Q, DB, ST> FromSqlRow<SelectBy<Q>, DB> for Q
 where
     ST: crate::expression::Expression,
     DB: Backend,
     Q: Selectable<Selection = ST>,
     Q: FromSqlRow<SqlTypeOf<ST>, DB>,
 {
     fn build_from_row<'a>(row: &impl Row<'a, DB>) -> Result<Self> {
         <Self as FromSqlRow<SqlTypeOf<ST>, DB>>::build_from_row(row)
     }
 }

 use crate::expression::SelectableExpression;
use crate::query_builder::select_clause::*;

 impl<QS, S> SelectClauseExpression<QS> for SelectBy<S>
 where
     S: Selectable,
     S::Selection: SelectableExpression<QS>,
 {
     type Selection = S::Selection;
     type SelectClauseSqlType = SqlTypeOf<S::Selection>;
 }

```

+ Some code to expose a API to actually use this.

This should greatly simplify the implementation and remove the blockers from my list above.

clouds56:
Thanks for response and not forget about this PR.

What about GroupBy? I think the only blocker was that you said you'd like to rethink `GroupBy`.
I know we are here to make diesel better, but after several months, if someone has to read the latest inner API again, rewrite all the code and fix all comments of review, would he feel tired and *hopeless* especially no idea about when could it be merged?
From my point of view, code at last commit is ready for merge, the things about `GroupBy` could be postponed to future PR. You said you have not much time on this PR and doing a refactoring of code without any discussion (I think it would be better if I'm informed before writing the PR and take it into consideration when design `select_by`).
My feeling is that the feature would never be accepted and I have to do refactoring again and again to chase after master or any other new ideas. And I'm already lost all the context about how diesel internally works.

> For my point of few the following points are should be checked again: (Most of them likely need some experimentation)
> * Is select_by::<Type> really a good API for this or should we try to do something like .select(Type::some_static_method()) instead?
> * How does this interact with .group_by clauses? I think that should be fine, but someone should try to write at least a compile_fail tests for some trivial cases.
> * Can we avoid some of the code duplication by reusing SelectStatement more directly? For example by having another generic parameter?

These are still not answered.

clouds56:
I'm not saying I'm giving this PR up. But I'd like to do nothing until things went clear.

weiznich:
> What about GroupBy? I think the only blocker was that you said you'd like to rethink GroupBy.
I know we are here to make diesel better, but after several months, if someone has to read the latest inner API again, rewrite all the code and fix all comments of review, would he feel tired and hopeless especially no idea about when could it be merged?
From my point of view, code at last commit is ready for merge, the things about GroupBy could be postponed to future PR. You said you have not much time on this PR and doing a refactoring of code without any discussion (I think it would be better if I'm informed before writing the PR and take it into consideration when design select_by).
My feeling is that the feature would never be accepted and I have to do refactoring again and again to chase after master or any other new ideas. And I'm already lost all the context about how diesel internally works.


I can fully understand that it is really frustrating that this PR is not merged yet, but as this is a really large change I do want to make sure that we get it somewhat right before merging it. In the end it's up to us diesel maintainers to maintain and live with all code we have merged. So better make sure those code is fine, otherwise we run into large trouble pretty soon (and probably will lose any interest in continue maintaining diesel). And yes working on OpenSource can sometimes be really frustrating. (Not only as contributor, but also as maintainer. Just check the issue tracker about what some comments request from us, because someone cannot use diesel for their business case...)

Additionally this feature is not something I personally need, therefore it's something where I accept contributions but will not spend much time on my own to figure out remaining open points. In the end my day has only a fixed amount of time and I cannot do everything. (And I'm not willing to spend all my free time on maintaining a open source project for other people. It's more like I fix the things I personally care about and spend some time to help/guide other people to fix/implement their features.) Therefore the list of open questions, as nobody has spend time to work on those it seems like there is not that much interest in this feature.

As for the concrete open points:

> Is select_by:: really a good API for this or should we try to do something like .select(Type::some_static_method()) instead?

This remains a open question. In the end it just needs someone who sits down tries out a few possibilities and provide a summary write up what worked well and what not. In the end this is a cosmetic small scale change, which would not change much of the underlying implementation. Just trying to get a nice API.

> How does this interact with .group_by clauses? I think that should be fine, but someone should try to write at least a compile_fail tests for some trivial cases.

As mentioned here: I would like to see and compile_fail test for some trivial case. Those needs to be written.

> Can we avoid some of the code duplication by reusing SelectStatement more directly? For example by having another generic parameter?

#2182 Allows to remove the code duplication. As mentioned here we can now implement this feature by "just" having a custom select clause instead. This should allow to remove basically all of `diesel/src/query_builder/select_statement/select_by/`.

This was not something that was clear in the beginning of this PR, but some of my motivation while implementing #2182 was to simplify approaches like this one and remove the code duplication here.


> I'm not saying I'm giving this PR up. But I'd like to do nothing until things went clear.

To write that again: I'm willing to help here with advice + code review as far as I find some spare time, but I do not have the time capacity or motivation to work on this feature by myself. If someone is interested in getting this merged they need to spend time and address the points above.

clouds56:
I think we doesn't need `impl<QS, S> SelectClauseExpression<QS> for SelectBy<S>`.
I'm not sure what `QueryMetadata` is but provide it as is.
updated and plz review.

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -70,6 +71,31 @@ where
     }
 }
 
+// FIXME: Should we disable select_by when `.group_by` has been called?
```
weiznich, line 74: I don't think we need to disable `select_by` when `group_by` has been called, as this is basically the same situation as for normal `select`.

diesel/src/query_builder/select_statement/dsl_impls.rs
```diff
@@ -70,6 +71,31 @@ where
...
+impl<SE, F, S, D, W, O, LOf, G, LC, Selection> SelectByDsl<Selection>
+    for SelectStatement<F, S, D, W, O, LOf, G, LC>
+where
+    SE: Expression,
+    Selection: Selectable<Expression = SE>,
+    Self: SelectDsl<SE>,
+    <Self as SelectDsl<SE>>::Output: SelectByQuery<Expression = SE>,
```
weiznich, line 81: I'm not sure why the `SelectByQuery` bound here is required. Instead of that bound there should be some bounds along the lines of `SelectBy<Selection>: SelectableExpression<F>` to verify that the select clause is fine for the given from clause.

diesel_derives/tests/selectable.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[cfg(feature = "mysql")]
+type IntSql = ::diesel::sql_types::BigInt;
+#[cfg(feature = "mysql")]
+type IntRust = i64;
+
+#[cfg(not(feature = "mysql"))]
+type IntSql = ::diesel::sql_types::Integer;
+#[cfg(not(feature = "mysql"))]
+type IntRust = i32;
```
weiznich, line 13: This is not required anymore. We can now just use `i64`/`BigInt` for all backends

weiznich (commented):
Thanks for picking this up again. It looks now nearly finished. I left a bunch of small stylistic comments, otherwise this is fine from my side. 
What I would like to see are 2 or 3 simple compile fail tests showing that this feature correctly rejects malformed select clauses (something like table appears not in from clause + maybe mixed aggregates).

diesel/src/expression/select_by.rs
```diff
@@ -0,0 +1,84 @@
...
+impl<T> SelectBy<T> {
+    pub fn new() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<T, ST> Expression for SelectBy<T>
+where
+    T: Selectable<Expression = ST>,
+    ST: QueryId + Expression,
```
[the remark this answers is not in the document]
clouds56: I once thought it not needed as well until the compiler complains

diesel/src/query_builder/select_statement/dsl_impls.rs
clouds56: It's just what SelectDsl write.

clouds56:
No, it doesn't work.
I realize that `impl<Q, DB, E> FromSqlRow<SelectBy<Q>, DB> for Q` has no effect, since when we are doing `.load::<Id>()` but not `.load::<SelectBy<Id>>()`. In this example, we have
```rust
#[derive(Selectable)]
#[table_name = "users"]
struct Id {
  id: i32
}
```
Now the problem is, there's no way to make `load::<U>()` only accept `Id`.

1. use `Expression` other than `SqlType` in trait `AsQuery` and `struct AnyExpression<SqlType>()` for original `.select()` dsl.
```rust
impl<T: Query> AsQuery for T {
    type SqlType = <Self as Query>::SqlType;
    type Query = Self;

    fn as_query(self) -> Self::Query {
        self
    }
}
```

2. another api named `.load_by`

3. not implement `Query` for `SelectStatement<SelectClause<SelectBy<T>>>`
   * would need carefully treatment and may introduce new select clause like `SelectByClause`
   * might make inner join of `SelectBy` statement much more complex, since it might relies on `AsQuery`

weiznich:
@clouds56 I've put together a minimal version that works [here](https://github.com/weiznich/diesel/commit/92c31a954bbf813d6ed30180e807928f715b7f7a). That obviously needs some cleanup + refactoring + finishing the implementation, but I think the basic idea should be clear.

clouds56:
That's an interesting point to introduce constraints with `CompatibleType`, while we cannot handle things like `(ST, Untyped)` as required in example of diesel_dynamic_schema.
I tried to `impl<A: CompatibleType, CompatibleType> CompatibleType for (A, B)` via `impl_tuples`, but things doesn't goes well, any suggestions?
Still investigating.

weiznich:
@clouds56 Thanks for updating the PR :+1: I'm quite occupied with work currently, but will try to find some time for looking into that.

weiznich:
@clouds56 I've found some time to look at this. I think it should be fixed by the impl added with my last commit.

diesel/src/query_dsl/mod.rs
```diff
@@ -287,6 +290,135 @@ pub trait QueryDsl: Sized {
...
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::users::dsl::*;
+    /// #     let connection = establish_connection();
+    /// // By default, all columns will be selected
+    /// let all_users = users.load::<(i32, String)>(&connection)?;
+    /// assert_eq!(vec![(1, String::from("Sean")), (2, String::from("Tess"))], all_users);
+    ///
+    /// let all_names = users.select_by::<UserName>().load::<UserName>(&connection)?;
```
pickfire, line 328: Is it necessary to make this distinct between `select` and `select_by`? Is it possible for `select` to have this feature?

weiznich: @pickfire That's a PR, so that's where development happens. This means this is not the right place to ask questions. If you want to contribute something to this feature please make a fully fleshed out proposal how to implement that or at least how the API would look like, otherwise this is only unnecessary noise for all contributors.

pickfire: Yes, this is a PR. So can I know where should I ask question for the PR if I should not ask in the PR itself? Or am I not able to give a comment if I did a review or maybe I should not review a PR at all?

weiznich: Asking questions is about the PR itself is totally fine, but that's not what you've written above. You proposed something without a second tought (It cannot work that way, because how would you deal with having a variable amount of arguments to that function (one for the old style select and none for the new style one?)) and then you asked a completely unrelated  question about something different.
From comments on PR's I expect that they are about the PR itself and not about some unrelated problem you currently have. I'm happy to get proposals anywhere, as long as the person writing the proposal has spend some time to actually thing about how this would look like and how to implement that. (It can even be fine to just raise a problem like: "I've though about implementing it that way, but hit a problem there") Otherwise such things are just a huge waste of time for me.

PR #2376, opened 2020-04-24
Many Nightly Fixes

martell:
Update to nightly that has clippy and rustfmt
Fix nightly builds https://github.com/rust-lang/rust/issues/70814
Fix an intra doc link for diesel as a dev-dependency


diesel_compile_tests/rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-04-01
+nightly-2020-03-19
```
weiznich, line 1: Why do we go back from `nightly-2020-04-01` to `nightly-2020-03-19`?

martell: I was trying to rollback to a version that does not have the following issue https://github.com/rust-lang/rust/issues/70814
but this version seems to still have the issues so it is probably easier to allow unused_braces.

martell: This was also the last build that had clippy and rustfmt before 04-01, I picked the most recent one that had both and did a work around for the nightly regression

martell:
@weiznich I think this should be good now, and ready for review :)

diesel_migrations/migrations_macros/Cargo.toml
```diff
@@ -24,3 +24,4 @@ proc-macro = true
 
 [features]
 default = []
+nightly = []
```
weiznich, line 27: We should not introduce a new feature flag here, as removing feature flags is quite hard (Would require a major version bump).
Also it seems like the corresponding rustc bug is "fixed" now, so maybe just using a newer nightly is sufficient?
https://github.com/rust-lang/rust/pull/71517

diesel_derives/src/lib.rs
```diff
@@ -18,6 +18,8 @@
     clippy::items_after_statements,
     clippy::used_underscore_binding
 )]
+// Nightly Regression https://github.com/rust-lang/rust/issues/70814
+#![cfg_attr(feature = "nightly", allow(unused_braces))]
```
JohnTitor, line 22: If you want to allow this lint only in nightly CI, you can pass the flag on CI config instead of adding new feature.

martell:
@weiznich I think I addressed all the concerns by updating :)

PR #2381, opened 2020-05-01
Add an example for Mysql transaction and lock

pymongo:
An example for Mysql transaction and lock using `for_update()` API.

I put Mysql general log to explain how it works.

By the way, I want to know how to use `with_lock()` API? 

I can't find any `with_lock()` example on the document.


weiznich:
Can you please elaborate why this needs a distinct example in the example directory, instead of just improving the API documentation [here](https://github.com/diesel-rs/diesel/blob/0b7bb3c7775b091ce9e48df748b75f2103d36c43/diesel/src/query_dsl/mod.rs#L773-L788), especially as this example applies to multiple backends, but is only provided for one.

PR #2399, opened 2020-05-16
Rewrite `embed_migrations!` to be a real proc macro

weiznich:
Remove the proc-macro-hack for `embed_migrations!` as it is not
required anymore. This enables importing the macro via use.
Also migrates `migrations_macros` to the 2018 edition.


diesel_migrations/migrations_macros/Cargo.toml
```diff
@@ -18,9 +18,24 @@ path = "../migrations_internals"
...
+path = "../"
+default-features = false
 
 [lib]
 proc-macro = true
 
 [features]
 default = []
+sqlite = []
+postgresql = []
```
JohnTitor, line 40: Should be?

```suggestion
postgres = []
```

weiznich: Yes :+1:

PR #2400, opened 2020-05-18
Properly decrement transaction_depth on a successful rollback from a serialization/read-only failure

chanks:
This iterates on PR #2364 - the logic in that PR wasn't decrementing the transaction_depth on a successful rollback, because change_transaction_depth was always using the Ok/Err value of the initial COMMIT, which was always Err. This PR changes it to use the Ok/Err value of the ROLLBACK, but still return an Err from the function if the commit failed.

This logic is a little thorny and repetitive in the use of change_transaction_depth, there may be a better way to represent it.

cc @Ten0

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -38,6 +38,7 @@ version = "~2.0.0"
 path = "../diesel_derives"
 
 [dev-dependencies]
+assert_matches = "1.0.1"
 cfg-if = "0.1.0"
 dotenv = ">=0.8, <0.11"
 quickcheck = "0.4"
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -38,7 +38,6 @@ version = "~2.0.0"
 path = "../diesel_derives"
 
 [dev-dependencies]
-assert_matches = "1.0.1"
 cfg-if = "0.1.0"
 dotenv = ">=0.8, <0.11"
 quickcheck = "0.4"
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -144,3 +144,93 @@ where
         self.transaction_depth.get() as u32
     }
 }
+
+#[cfg(test)]
+mod test {
+
+    #[test]
+    #[cfg(feature = "postgres")]
+    fn transaction_depth_is_tracked_properly_on_commit_failure() {
+        use crate::result::DatabaseErrorKind::SerializationFailure;
+        use crate::result::Error::DatabaseError;
+        use crate::*;
+        use assert_matches::assert_matches;
+        use std::sync::{Arc, Barrier};
+        use std::thread;
+
+        table! {
+            #[sql_name = "transaction_depth_is_tracked_properly_on_commit_failure"]
+            serialization_example {
+                id -> Serial,
+                class -> Integer,
+            }
+        }
+
+        let conn = crate::test_helpers::pg_connection_no_transaction();
+
+        sql_query("DROP TABLE IF EXISTS transaction_depth_is_tracked_properly_on_commit_failure;")
+            .execute(&conn)
+            .unwrap();
+        sql_query(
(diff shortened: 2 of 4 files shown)


Ten0:
Right, my mistake!
I had noticed there still was an issue there as I had built a wrapper around R2-D2's Connection that would report non-closed transactions and it would still report that, but I didn't have time to seriously dig into why yet.
This should probably also be backported (as we backported the previous fix attempt).

chanks:
I haven't looked at this crate's test suite yet, I can try to get to it in the next day or two.

chanks:
Do either of you have any suggestions on how to test this? I'm coming at this as someone who knows Postgres very well, SQLite only a bit and MySQL not at all. So I guess I have three options here:

1) Set up a test case where a real database returns a serialization failure on COMMIT (or maybe a read-only failure, but I'm not sure how to get those to trigger on COMMIT - I believe in Postgres at least they trigger at the point when the INSERT/UPDATE/DELETE is issued in the transaction). To accomplish this I'd need two connections writing to the same DB at the same time in different threads, and probably use some kind of concurrency primitives to synchronize their actions, which is a lot of overhead. This would also be a database-specific test, which, I don't know whether that's a faux pas for logic that's supposed to be database-agnostic.

2) Somehow stub out a fake connection object to return specific results/errors when sent certain queries. I've only looked around a bit but this doesn't seem to be done anywhere else in the codebase?

3) Rework the TransactionManager implementation to try to separate this commit/error-checking/rollback logic from the actual connection handling, and then unit-test the former.

Thoughts? All of these seem like pretty big lifts to fix a bug, unless there's a better option I'm missing.

weiznich:
I would go for the first solution, as this is the simplest one without changing large parts in diesels code base. We already have some [test case](https://github.com/diesel-rs/diesel/blob/45da0109a8b085e4aad629e64c0bb2d17f84c486/diesel_tests/tests/errors.rs#L109) that generates a `SerializationFailure` so the only thing that needs to be done is creating a new testcase out of this code, moving it to be a module local test case in `transaction_manager.rs` (because we want to access the internals of the specific transaction manager implementation), wrap up the relevant code into a transaction and add a few asserts that the `transaction_depth` is incremented/decremented as expected. I think it's also fine to have this a postgres specific test for now, as this should just verify that the general strategy is working now.

chanks:
Added a test case, please take a look @weiznich

diesel/Cargo.toml
```diff
@@ -38,6 +38,7 @@ version = "~2.0.0"
 path = "../diesel_derives"
 
 [dev-dependencies]
+assert_matches = "1.0.1"
```
weiznich, line 41: I would prefer not to add another dependency here, especially a simple on like that one.

This macro + a ordinary `assert!` should work well enough for that test case.
```rust
macro_rules! matches {
    ($expression:expr, $( $pattern:pat )|+ $( if $guard: expr )?) => {
        match $expression {
            $( $pattern )|+ $( if $guard )? => true,
            _ => false
        }
    }
}
```

diesel/src/connection/transaction_manager.rs
```diff
@@ -117,31 +117,120 @@ where
...
+                    assert_eq!(0, <AnsiTransactionManager as TransactionManager<PgConnection>>::get_transaction_depth(&conn.transaction_manager));
+
+                    let result =
+                    conn.build_transaction().serializable().run(|| {
+                        let _ = serialization_example::table
+                            .filter(serialization_example::class.eq(i))
+                            .count()
+                            .execute(&conn)?;
+
+                        assert_eq!(1, <AnsiTransactionManager as TransactionManager<PgConnection>>::get_transaction_depth(&conn.transaction_manager));
```
weiznich, line 210: That check should be better done at the beginning of the transaction callback, otherwise it seems to be a bit out of place here, in between the different queries.

chanks:
Updated for feedback @weiznich

Issue #2095, opened 2019-06-19
[diesel_cli] Support redo and revert all applied migrations


viniciusd:
It would be very helpful if it were possible to run both `diesel migration redo all` and `diesel migration revert all`, those would be very welcome on some CI setups.
For example, on testing both upgrade and downgrade migrations.

As a side enhancement, both redo and revert could support an integer number N of migrations in order to acting on the latest N migrations.

However, my main goal here is supporting an `all` option.

I am not familiar with diesel's code, yet I would be willing to take it as initial contact with both its code base and Rust itself. Some mentoring would be welcomed, but this look likes trivial enough for me to endeavour by myself.

weiznich:
The main entry point for the migration command is in [`diesel_cli`](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/main.rs#L79).  If you have any specific questions or need help somewhere, just ask here or on our [gitter channel](https://gitter.im/diesel-rs/diesel). Also opening a WIP PR is creating a good place for discussions about implementation related things.

viniciusd:
Is that a green flag? May I proceed with it?

weiznich:
:+1:

theredfish:
I can take this issue if nobody is working on it. I'm interested by these features.
@weiznich / @viniciusd  is it ok?

weiznich:
@theredfish Feel free to work on this feature.

viniciusd:
Go for it @theredfish , I will love to go over it. A lot has happened since then and I couldn't start working at this issue.

weiznich:
Fixed by #2401

PR #2401, opened 2020-05-18
Feat: migration revert and redo all

theredfish:
This pull request is related to #2095 .

This is a work in progress feature to track the work and open the discussion if needed.

- [x] revert all
- [x] revert the n last migrations
- [x] redo all
- [x] redo the n last migrations

### DIFF
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -18,11 +18,37 @@ pub fn build_cli() -> App<'static, 'static> {
         .setting(AppSettings::VersionlessSubcommands)
         .arg(migration_dir_arg())
         .subcommand(SubCommand::with_name("run").about("Runs all pending migrations"))
-        .subcommand(SubCommand::with_name("revert").about("Reverts the latest run migration"))
-        .subcommand(SubCommand::with_name("redo").about(
-            "Reverts and re-runs the latest migration. Useful \
-             for testing that a migration can in fact be reverted.",
-        ))
+        .subcommand(
+            SubCommand::with_name("revert")
+                .about("Reverts the latest run migration")
+                .arg(
+                    Arg::with_name("REVERT_ALL")
+                        .long("all")
+                        .help("Reverts all migrations")
+                        .long_help(
+                            "When this option is specified all migrations \
+                            will be reverted. Useful for testing your revert \
+                            scripts only.",
+                        )
+                )
+        )
+        .subcommand(
+            SubCommand::with_name("redo")
+                .about(
+                    "Reverts and re-runs the latest migration. Useful \
+                    for testing that a migration can in fact be reverted.",
+                )
+                .arg(
+                    Arg::with_name("REDO_ALL")
+                        .long("all")
+                        .help("Reverts and re-runs all migrations.")
+                        .long_help(
+                            "When this option is specified all migrations \
+                            will be reverts and re-runs. Useful for testing \
+                            that your migrations can be reverted and applied."
+                        )
+                )
+            )
         .subcommand(
             SubCommand::with_name("list")
                 .about("Lists all available migrations, marking those that have been applied."),
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -32,7 +32,8 @@ pub fn build_cli() -> App<'static, 'static> {
                         )
                         .takes_value(true)
                         .required(true)
-                        .default_value("1"),
+                        .default_value("1")
+                        .validator(migration_revert_number_validator),
                 ),
         )
         .subcommand(
@@ -232,4 +233,22 @@ fn migration_dir_arg<'a, 'b>() -> Arg<'a, 'b> {
         )
         .takes_value(true)
         .global(true)
+}
+
+fn migration_revert_number_validator(val: String) -> Result<(), String> {
+
+    if val == "all" {
+        return Ok(());
+    }
+
+    let number = match val.parse::<i64>() {
+        Ok(number) => number,
+        Err(_) => return Err("Cannot parse <REVERT_NUMBER>. The input must be an integer or 'all'.".to_string())
+    };
+
+    if number < 1 {
+        return Err("<REVERT_NUMBER> must be at least equal to 1.".to_string());
+    }
+
+    Ok(())
 }
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -22,21 +22,24 @@ pub fn build_cli() -> App<'static, 'static> {
             SubCommand::with_name("revert")
                 .about("Reverts the latest run migration")
                 .arg(
-                    Arg::with_name("REVERT_ALL")
-                        .long("all")
-                        .help("Reverts all migrations")
+                    Arg::with_name("REVERT_NUMBER")
+                        .long("number")
+                        .short("n")
+                        .help("Reverts the last migration files")
                         .long_help(
-                            "When this option is specified all migrations \
-                            will be reverted. Useful for testing your revert \
-                            scripts only.",
+                            "When this option is specified the last `n` migration files \
+                             will be reverted. By default revert the last one.",
                         )
-                )
+                        .takes_value(true)
+                        .required(true)
+                        .default_value("1"),
+                ),
         )
         .subcommand(
             SubCommand::with_name("redo")
                 .about(
                     "Reverts and re-runs the latest migration. Useful \
-                    for testing that a migration can in fact be reverted.",
+                     for testing that a migration can in fact be reverted.",
                 )
                 .arg(
                     Arg::with_name("REDO_ALL")
@@ -44,11 +47,11 @@ pub fn build_cli() -> App<'static, 'static> {
                         .help("Reverts and re-runs all migrations.")
                         .long_help(
                             "When this option is specified all migrations \
-                            will be reverts and re-runs. Useful for testing \
-                            that your migrations can be reverted and applied."
-                        )
-                )
-            )
+                             will be reverted and re-runs. Useful for testing \
+                             that your migrations can be reverted and applied.",
+                        ),
+                ),
+        )
         .subcommand(
             SubCommand::with_name("list")
                 .about("Lists all available migrations, marking those that have been applied."),
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -56,12 +56,31 @@ pub fn build_cli() -> App<'static, 'static> {
                 .arg(
                     Arg::with_name("REDO_ALL")
                         .long("all")
+                        .short("a")
                         .help("Reverts and re-runs all migrations.")
                         .long_help(
                             "When this option is specified all migrations \
                              will be reverted and re-runs. Useful for testing \
                              that your migrations can be reverted and applied.",
-                        ),
+                        )
+                        .takes_value(false)
+                        .conflicts_with("REDO_NUMBER"),
+                )
+                .arg(
+                    Arg::with_name("REDO_NUMBER")
+                        .long("number")
+                        .short("n")
+                        .help("Redo the last `n` migration files")
+                        .long_help(
+                            "When this option is specified the last `n` migration files \
+                             will be reverted and re-runs. By default redo the last migration.",
+                        )
+                        // TODO : when upgrading to clap 3.0 add default_value("1").
+                        // Then update code in main.rs for the revert subcommand.
+                        // See https://github.com/clap-rs/clap/issues/1605
+                        .takes_value(true)
+                        .validator(is_positive_int)
+                        .conflicts_with("REDO_ALL"),
                 ),
         )
         .subcommand(
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -236,15 +236,13 @@ fn migration_dir_arg<'a, 'b>() -> Arg<'a, 'b> {
 }
 
 fn migration_revert_number_validator(val: String) -> Result<(), String> {
-
     if val == "all" {
         return Ok(());
     }
 
-    let number = match val.parse::<i64>() {
-        Ok(number) => number,
-        Err(_) => return Err("Cannot parse <REVERT_NUMBER>. The input must be an integer or 'all'.".to_string())
-    };
+    let number = val.parse::<i64>().map_err(|_| {
+        "Cannot parse <REVERT_NUMBER>. The input must be an integer or 'all'.".to_string()
+    })?;
 
     if number < 1 {
         return Err("<REVERT_NUMBER> must be at least equal to 1.".to_string());
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -1,3 +1,4 @@
+use crate::validators::num::*;
 use clap::{App, AppSettings, Arg, Shell, SubCommand};
 
 pub fn build_cli() -> App<'static, 'static> {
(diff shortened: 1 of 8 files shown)


diesel_cli/src/cli.rs
```diff
@@ -18,11 +18,37 @@ pub fn build_cli() -> App<'static, 'static> {
...
-        .subcommand(SubCommand::with_name("revert").about("Reverts the latest run migration"))
-        .subcommand(SubCommand::with_name("redo").about(
-            "Reverts and re-runs the latest migration. Useful \
-             for testing that a migration can in fact be reverted.",
-        ))
+        .subcommand(
+            SubCommand::with_name("revert")
+                .about("Reverts the latest run migration")
+                .arg(
+                    Arg::with_name("REVERT_ALL")
```
pksunkara, line 25: Why is this in SCREAMING_SNAKE_CASE? `Arg::with_name("all")` sounds better. Args are unique to only the SubCommands. So, you don't need `REVERT_` here. Similarly below.

pksunkara:
It's probably out of scope, but I was also thinking that these subcommands need `-n 3` kind of arg so that they only do the latest 3 migrations etc..

diesel_cli/src/cli.rs
theredfish: Ok thank you, I will make the changes.

theredfish:
> It's probably out of scope, but I was also thinking that these subcommands need `-n 3` kind of arg so that they only do the latest 3 migrations etc..

Well I think I can take a look, but after the `all` options. Maybe in another pull request :)

theredfish:
I've been really busy these last weeks fyi. I'm back and available, so I will spend some time on this PR.

diesel_cli/src/cli.rs
theredfish: @pksunkara in fact I followed the naming convention previously used in this file. For example : 

```rust
.subcommand(
    SubCommand::with_name("generate")
        .about(
            "Generate a new migration with the given name, and \
             the current timestamp as the version",
        )
        .arg(
            Arg::with_name("MIGRATION_NAME")
                .help("The name of the migration to create")
                .required(true),
        )
        .arg(
            Arg::with_name("MIGRATION_VERSION")
                .long("version")
                .help(
                    "The version number to use when generating the migration. \
                     Defaults to the current timestamp, which should suffice \
                     for most use cases.",
                )
                .takes_value(true),
        )
```

To answer to your question.

pksunkara: Yeah, I would say so. I am maintainer of clap so I know how people use it. But @weiznich might have some preferences here.

weiznich: I would prefer to just keep using the old style. Generally we try to avoid bigger cosmetic code refactoring, as this will "destroy" the history of that code. (It makes it harder to use something like `git blame` to get information about when the code was last changed, which allows to see the reason in the corresponding commit message)

theredfish:
> 
> 
> It's probably out of scope, but I was also thinking that these subcommands need `-n 3` kind of arg so that they only do the latest 3 migrations etc..

@pksunkara  @weiznich 

I'm trying to find a good way to avoid a lot of API changes. So i'm considering adding something like `migration revert -n 2` for example, to revert the last two migrations.

In this case, I think `migration revert` should revert all migrations by default, and if we need to rollback only some migrations we can use the `-n` argument.
However it's an major breaking change. 

On the other hand, I'm considering using a default value (like "all") for the `-n` argument when no value is provided. But in my opinion it's less natural.

What do you think?

pksunkara:
Default value for n should be 1. That way it will not be a breaking change. If n is not provided and all is provided then you can revert all.

theredfish:
I added a commit where I can revert the n last migrations. It's a work in progress, so I haven't tried to revert all commits, and test limits for now. There are `todo` comments, it's where I need help / have some questions. If you're available.

theredfish:
@weiznich Should I avoid to revert the migration `00000000000000_diesel_initial_setup`? Currently it's possible but I wondering if this is something users expect? Since this is not their own migration.

For now I made the choice to exclude this migration from the `revert -n <REVERT_NUMBER>` command.

weiznich:
I think its completly fine to revert that migration to. The current version of `diesel migration run/revert` do all handle this migration like any other user provided migration.

theredfish:
@weiznich are you available for a first review about the revert feature? There are two TODOs, where I'm not sure how to manage the things. You can check the tests in `diesel_cli>tests>migration_revert`, they show how the API can be used and test the different use cases.

diesel_cli/src/cli.rs
```diff
@@ -222,3 +252,19 @@ fn migration_dir_arg<'a, 'b>() -> Arg<'a, 'b> {
         .takes_value(true)
         .global(true)
 }
+
+fn migration_revert_number_validator(val: String) -> Result<(), String> {
```
weiznich, line 256: The help should somehow mention that `all` is also a valid value for `diesel revert -n`

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -200,6 +200,55 @@ where
...
+    setup_database(conn)?;
+
+    let latest_migration_versions = match conn.latest_run_migration_versions(number) {
+        Ok(migration_versions) => migration_versions,
+        Err(_) => {
+            return Err(RunMigrationsError::MigrationError(
+                MigrationError::NoMigrationRun,
+            ))
+        }
+    };
```
weiznich, line 234: Maybe write this as
```suggestion
    let latest_migration_versions = conn.latest_run_migration_versions(number)
        .map_err(|_| RunMigrationsError::MigrationError(MigrationError::NoMigrationRun))?;
```

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -200,6 +200,55 @@ where
...
+
+    if latest_migration_versions.is_empty() {
+        return Err(RunMigrationsError::MigrationError(
+            MigrationError::NoMigrationRun,
+        ));
+    }
+
+    for migration_version in &latest_migration_versions {
+        // TODO see how to handle the errors... even if these errors
```
weiznich, line 243: I think the correct behavior here is to return the error and stop reverting anything at this point. We cannot be sure that the next migrations will be fine and will not damage anything.

diesel_migrations/migrations_internals/src/lib.rs
```diff
@@ -200,6 +200,55 @@ where
...
+) -> Result<Vec<String>, RunMigrationsError>
+where
+    Conn: MigrationConnection,
+{
+    let number = migrations_in_directory(path)?.len() as i64;
+
+    revert_latest_migrations_in_directory(conn, path, number)
+}
+
+// TODO : see if we need a function named revert_latest_migrations
```
weiznich, line 215: As long as those functions are not part of the public API I don't think it's required to add a `revert_latest_migrations` function, otherwise this would be desirable

diesel_cli/src/cli.rs
```diff
@@ -18,11 +18,41 @@ pub fn build_cli() -> App<'static, 'static> {
         .setting(AppSettings::VersionlessSubcommands)
         .arg(migration_dir_arg())
         .subcommand(SubCommand::with_name("run").about("Runs all pending migrations"))
-        .subcommand(SubCommand::with_name("revert").about("Reverts the latest run migration"))
-        .subcommand(SubCommand::with_name("redo").about(
-            "Reverts and re-runs the latest migration. Useful \
-             for testing that a migration can in fact be reverted.",
-        ))
+        .subcommand(
+            SubCommand::with_name("revert")
```
pksunkara, line 22: This subcommand should have another argument `all` which is exclusive with `number` arg. The `revert all` should revert all the migrations.

diesel_cli/src/cli.rs
```diff
@@ -18,11 +18,41 @@ pub fn build_cli() -> App<'static, 'static> {
...
+                             will be reverted. By default revert the last one.",
+                        )
+                        .takes_value(true)
+                        .required(true)
+                        .default_value("1")
+                        .validator(migration_revert_number_validator),
+                ),
+        )
+        .subcommand(
+            SubCommand::with_name("redo")
```
pksunkara, line 40: This subcommand should have a `number` argument which should default to `1` and should be exclusive with `all` arg.

diesel_cli/src/cli.rs
pksunkara, line 256: You don't need this whole function when you add the `all` argument. And you don't need the docs either since CLI help will show it.

theredfish:
Thank you for your reviews!

I've been really busy these last weeks. I'm back on the subject.

diesel_cli/src/cli.rs
theredfish: Hello @pksunkara 

I tried to use `conflicts_with` and a `default_value`, but it doesn't work since the default value will "activate" the `Arg`. 

```rust
.subcommand(
            SubCommand::with_name("revert")
                .about("Reverts the specified migrations")
                .arg(
                    Arg::with_name("REVERT_ALL")
                        .long("all")
                        .short("a")
                        .help("Reverts previously run migration files.")
                        .takes_value(false)
                        .conflicts_with("REVERT_NUMBER"),
                )
                .arg(
                    Arg::with_name("REVERT_NUMBER")
                        .long("number")
                        .short("n")
                        .help("Reverts the last `n` migration files")
                        .long_help(
                            "When this option is specified the last `n` migration files \
                             will be reverted. By default revert the last one.",
                        )
                        .takes_value(true)
                        .default_value("1")
                        .validator(is_positive_int)
                        .conflicts_with("REVERT_ALL"),
                ),
```

This will throw an error : 

```
diesel migration revert --all
stderr: error: The argument '--all' cannot be used with '--number <REVERT_NUMBER>'
```

If you need it, I can push my current code with the failing test.

diesel_cli/src/cli.rs
theredfish: Hello @pksunkara 

I tried to use `conflicts_with` and a `default_value`, but it doesn't work since the default value will "activate" the `Arg`. 

```rust
.subcommand(
            SubCommand::with_name("revert")
                .about("Reverts the specified migrations")
                .arg(
                    Arg::with_name("REVERT_ALL")
                        .long("all")
                        .short("a")
                        .help("Reverts previously run migration files.")
                        .takes_value(false)
                        .conflicts_with("REVERT_NUMBER"),
                )
                .arg(
                    Arg::with_name("REVERT_NUMBER")
                        .long("number")
                        .short("n")
                        .help("Reverts the last `n` migration files")
                        .long_help(
                            "When this option is specified the last `n` migration files \
                             will be reverted. By default revert the last one.",
                        )
                        .takes_value(true)
                        .default_value("1")
                        .validator(is_positive_int)
                        .conflicts_with("REVERT_ALL"),
                ),
```

This will throw an error : 

```
diesel migration revert --all
stderr: error: The argument '--all' cannot be used with '--number <REVERT_NUMBER>'
```

If you need it, I can push my current code with the failing test.

pksunkara: That issue had been fixed but it's only available in 3.0 prereleases. https://github.com/clap-rs/clap/issues/1605

theredfish: Ok thank you for the info, so I suppose I will need to handle the default value by myself or the conflict.

pksunkara: It's probably the best way to proceed and please add a comment saying that the logic should be removed when upgrading to  clap 3.0

diesel_cli/src/cli.rs
theredfish: I added a `validators` module. For now it's used in `REVERT_NUMBER` to check if the value of the argument is a positive number.

pksunkara (commented):
Looks great. Only [this](https://github.com/diesel-rs/diesel/pull/2401/files#r467845310) left

theredfish:
I noticed that the migration `00000000000000_diesel_initial_setup` is displayed when reverting migrations but not when running migrations.

```rust
fn run_migration<Conn>(
    conn: &Conn,
    migration: &dyn Migration,
    output: &mut dyn Write,
) -> Result<(), RunMigrationsError>
where
    Conn: MigrationConnection,
{
    conn.transaction(|| {
        if migration.version() != "00000000000000" {
            writeln!(output, "Running migration {}", name(&migration))?;
        }

       ...
}
```

```rust
fn revert_migration<Conn: Connection>(
    conn: &Conn,
    migration: &dyn Migration,
    output: &mut dyn Write,
) -> Result<(), RunMigrationsError> {
    conn.transaction(|| {
        writeln!(output, "Rolling back migration {}", name(&migration))?;
        ...
}
```

I suggest to keep the same behavior for both `migration revert` and `migration run`. Removing the output from `revert_migration` will avoid confusions for users.

theredfish:
Apparently the conflicting files block the Travis jobs. Should I rebase master?

pksunkara:
You have to do it anyway to get this merged, right? 😄

theredfish:
> You have to do it anyway to get this merged, right? 😄

It depends of the projects, so I wasn't so sure since it changes the history 😄. I will do that

pksunkara:
Yeah, this is your branch in your repo. Github comments are smart, so it should be not issue to understand what the others were talking about before you force push.

diesel_cli/src/main.rs
```diff
@@ -71,19 +72,43 @@ fn run_migration_command(matches: &ArgMatches) -> Result<(), Box<dyn Error>> {
...
+                    None => "1",
+                    Some(number) => number,
+                };
+
+                call_with_conn!(
+                    database_url,
+                    migrations::revert_latest_migrations_in_directory(
+                        &dir,
+                        // We can unwrap since the argument is validated from the cli module.
+                        number.parse::<u64>().unwrap()
```
weiznich, line 100: This should use `.expect` with an meaningful error message instead.

diesel_cli/src/main.rs
```diff
@@ -327,27 +352,54 @@ fn search_for_directory_containing_file(path: &Path, file: &str) -> DatabaseResu
...
+            let number = match args.value_of("REDO_NUMBER") {
+                None => "1",
+                Some(number) => number,
+            };
+
+            migrations::revert_latest_migrations_in_directory(
+                conn,
+                migrations_dir,
+                // We can unwrap since the argument is validated from the cli module.
+                number.parse::<u64>().unwrap(),
```
weiznich, line 382: Use `.expect` instead.

weiznich (approved):
Looks good modulo the two small nits regarding to the usage of `unwrap`.

theredfish:
Ok thank you for the review @weiznich . I replaced the `unwrap` calls by `expect` .

pksunkara (approved):

weiznich (approved):

weiznich:
@theredfish Thanks for working on this. This is really a well written PR :+1:

theredfish:
> 
> 
> @theredfish Thanks for working on this. This is really a well written PR 👍

Oh thank you ! It was really interesting to work on these features. Thanks again for the review and your time @weiznich and @pksunkara .

JasterV:
How can I update the diesel tool to support these new features? Thanks!

theredfish:
Hello @JasterV  one way would be to reference the commit hash in your `Cargo.toml`, by specifying `rev`. Or you can target the master / main branch (more info on the [cargo book](https://doc.rust-lang.org/cargo/index.html)). Another way would be to help to release a new version of Diesel but I can't find the next steps anymore.

weiznich:
@JasterV Please be aware that the master branch contains breaking changes, therefore you cannot simple use the newly build diesel-cli with an old project without doing an update first.
Otherwise this feature will be part of our next release.

ilslv:
@weiznich is there anything blocking release of the new version of `diesel_cli` containing features from this PR?

weiznich:
@ilslv We generally do not answer random requests for release time lines of certain features.

kajmagnus:
>  I think migration revert should revert all migrations by default [...] What do you think?

I wouldn't have liked that — it'd made me a bit afraid, even. It would have erased everything in the database (dropped all tables) and caused data loss (everything since the last backup), if anyone expected "revert" to, just like "undo", affect the last change only. Or if one forgot "-n N", if trying to fix some unexpected migration problem in a database with real data.

Just mentioning, in case thinking along those lines (being "afraid" of data loss) could be helpful some other time in the future.

Diesel, b.t.w., seems like a really nice migration tool, thinking about using it :- )  (and I was reading about down migrations and stumbled upon this issue).

weiznich:
The implemented default is to revert just 1 migration. Everything else needs to be specified explicitly via either `-n number` or `--all`

kajmagnus:
@weiznich yes I know (if you replied to me), that's why I wrote "would",  and "Just mentioning in case ... ", in case what I wrote would be helpful to the Original Poster in the future if writing more Diesel code

PR #2408, opened 2020-05-22
Upgrade to the latest bigdecimal version

### LINKED ISSUE #2402: bigdecimal version requirement
`diesel`'s `Cargo.toml` (both `master` and `1.4.x`) supports a somewhat awkward range of versions of `bigdecimal`:

https://github.com/diesel-rs/diesel/blob/45da0109a8b085e4aad629e64c0bb2d17f84c486/diesel/Cargo.toml#L32

Ever since `bigdecimal` `0.1.2` was released on April 26, I have seen a few questions about this because cargo now selects `0.1.2` if a user's `Cargo.toml` has `0.1.0`. It's not obvious (especially to new users) that they have to depend on bigdecimal `<=0.1.0` (or `=0.1.0`) in order to restrict their own dependency to the one that `diesel` has.

`0.1.x` should never break compatibility, so would `>=0.0.10, <0.2` be acceptable for this dependency instead?

dtulig:
I followed the same convention of limiting the patch version so this won't upgrade automatically. It doesn't look like bigdecimal has frequent releases anyway.

Here is the diff for bigdecimal: https://github.com/akubera/bigdecimal-rs/compare/v0.1.0...v0.1.2

fixes #2402

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -29,7 +29,7 @@ ipnetwork = { version = ">=0.12.2, <0.17.0", optional = true }
 num-bigint = { version = ">=0.1.41, <0.3", optional = true }
 num-traits = { version = "0.2", optional = true }
 num-integer = { version = "0.1.32", optional = true }
-bigdecimal = { version = ">= 0.0.10, <= 0.1.2", optional = true }
+bigdecimal = { version = ">= 0.0.10, < 0.2.0", optional = true }
 bitflags = { version = "1.0", optional = true }
 r2d2 = { version = ">= 0.8, < 0.9", optional = true }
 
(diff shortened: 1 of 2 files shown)


diesel/Cargo.toml
```diff
@@ -29,7 +29,7 @@ ipnetwork = { version = ">=0.12.2, <0.17.0", optional = true }
 num-bigint = { version = ">=0.1.41, <0.3", optional = true }
 num-traits = { version = "0.2", optional = true }
 num-integer = { version = "0.1.32", optional = true }
-bigdecimal = { version = ">= 0.0.10, <= 0.1.0", optional = true }
+bigdecimal = { version = ">= 0.0.10, <= 0.1.2", optional = true }
```
weiznich, line 32: As pointed out in #2402 its fine to allow all versions < 0.2.0

PR #2409, opened 2020-05-24
Send connection to migration library to allow multiple databases

pksunkara:
I have the barrel related code [here](https://github.com/reign-rs/barrel/commit/2043dd81cd66a0743318c94d72c5c5b6ccc358be).

Looks like these libraries are interdependent on each other and is going to complicate code changes and releases.


diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -54,9 +54,15 @@ impl<'a> fmt::Display for MigrationFileName<'a> {
...
+pub fn migration_from<Conn>(
+    conn: &Conn,
+    path: PathBuf,
+) -> Result<Box<dyn Migration>, MigrationError>
+where
+    Conn: MigrationConnection,
+{
     #[cfg(feature = "barrel")]
-    match ::barrel::integrations::diesel::migration_from(&path) {
+    match ::barrel::integrations::diesel::migration_from(conn, &path) {
```
pksunkara, line 65: The current `barrel` doesn't support this `conn` param. I can't update the version of barrel because barrel's version depends on this change being published.

diesel_cli/Cargo.toml
```diff
@@ -46,10 +46,14 @@ url = { version = "2.1.0" }
...
-sqlite = ["diesel/sqlite"]
 mysql = ["diesel/mysql", "url", "uses_information_schema"]
-barrel-migrations = ["migrations_internals/barrel", "barrel"]
+sqlite = ["diesel/sqlite"]
 sqlite-bundled = ["sqlite", "libsqlite3-sys/bundled"]
+barrel-migrations = ["migrations_internals/barrel", "barrel"]
+barrel-postgres = ["postgres", "barrel-migrations", "barrel/pg"]
+barrel-mysql = ["mysql", "barrel-migrations", "barrel/mysql"]
+barrel-sqlite = ["sqlite", "barrel-migrations", "barrel/sqlite3"]
+barrel-sqlite-bundled = ["sqlite-bundled", "barrel-migrations", "barrel/sqlite3"]
```
weiznich, line 56: There is really no way to prevent duplicating all those features here? I do not really like this.

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -54,9 +54,15 @@ impl<'a> fmt::Display for MigrationFileName<'a> {
     }
 }
 
-pub fn migration_from(path: PathBuf) -> Result<Box<dyn Migration>, MigrationError> {
+pub fn migration_from<Conn>(
```
weiznich, line 57: As this is a function in our public API I would like to see some reasoning why this is change is required and what other possible solutions could be implemented.
(Basically this needs a reasoning why we need to break our API here)

diesel_cli/Cargo.toml
pksunkara: We don't actually need them here. I added them to make it easy for people install diesel_cli. Instead of running,

```
cargo install diesel_cli --no-default-features --features "postgres, barrel-migrations, barrel/pg"
```

They can now just run:

```
cargo install diesel_cli --no-default-features --features barrel-postgres
```

Since they are just mapping to other features, they don't cost us anything.

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -54,9 +54,15 @@ impl<'a> fmt::Display for MigrationFileName<'a> {
...
+    _conn: &Conn,
```
teymour-aldridge, line 58:
```suggestion
    conn: &Conn,
```

diesel_migrations/migrations_internals/src/migration.rs
```diff
@@ -54,9 +54,15 @@ impl<'a> fmt::Display for MigrationFileName<'a> {
...
+    path: PathBuf,
+) -> Result<Box<dyn Migration>, MigrationError>
+where
+    Conn: MigrationConnection,
+{
     #[cfg(feature = "barrel")]
-    match ::barrel::integrations::diesel::migration_from(&path) {
+    match ::barrel::integrations::diesel::migration_from(_conn, &path) {
```
teymour-aldridge, line 65:
```suggestion
    match ::barrel::integrations::diesel::migration_from(conn, &path) {
```

diesel_migrations/migrations_internals/src/migration.rs
pksunkara: No, because `conn` is only used in barrel, so if barrel is not enable, your change will give warning

teymour-aldridge: Oh - my bad (I didn't see the feature flag)!

diesel_migrations/migrations_internals/src/migration.rs
pksunkara: So, the explanation is that migration libraries would need to know the connection type when building the migration even before running it because they might want to do different things based on the type of database. This change provides that.

Alternative is that the migration libraries are compiled for just one database kind and then the migration builder and runner would just use logic for that connection kind. (This is what is happening with current barrel).

I discarded the alternative because I envision the future where production database might be different from test/dev database and we want to run migrations on both kind of connections with the same installations.

diesel_cli/Cargo.toml
weiznich: Just going through all the old PR, so to continue here I would like to see this features to be removed. I think it's fine to just pass the 3 feature flags as shown above.

diesel_migrations/migrations_internals/src/migration.rs
weiznich: Instead of passing in the whole connection I would abstract that away behind an `enum` as each backend could have more than one concrete connection type (for example the pooled variant, or a custom logger wrapper, or a future postgres connection based on rust-postgres, …)

pksunkara:
I think this can be closed since I have decided not to work on this. This PR while fixes things but doesn't make it satisfactory enough. Also, barrel is pretty much abandoned.

PR #2410, opened 2020-05-25
Update `dotenv` to 0.15

JohnTitor:
Since we head 2.0 release, I think we could allow losing older dotenv support here.


weiznich:
As we don't export `dotenv` as public dependency anywhere I think that can be done unrelated to the 2.0 release, therefore I'm marking this as `maybe backport` so that we would potentially include it in a 1.4.x release (if we issue another one)

PR #2413, opened 2020-05-26
Custom types draft

l4l:
[Original source](https://github.com/l4l/l4l.github.io/blob/develop/public/posts/2020_05_24_custom_types_in_diesel.md). [Related discussion](https://github.com/l4l/diesel-custom-types/issues/1), and in particular stated problems so far:

> To note points that I've noticed while reading the post again:
> 
> # Major Points
> * Skip the experimenting part at the beginning. The general structure should be: 
>     1. Explain SqlType <-> Rust Type.
>     2. Show the sql side type (migration)
>     3. Implement the diesel sql type
>     4. Configure diesel cli (maybe mention patch option here?)
>     5. Implement `FromSql`/`ToSql` (mention that this depends on the actual use case)
>     6. Other stuff (like `diesel-derive-enum`, but that should mention shortly that this is a third party crate and diesel does not expose this kind of API because of flexibility concerns)
> * It should be mentioned somewhere that the process of adding support for arbitrary types is basically the same, as long as you own the rust side type. (Maybe linking to [`WriteTuple`](http://docs.diesel.rs/diesel/serialize/trait.WriteTuple.html)/[`Record`](http://docs.diesel.rs/diesel/pg/types/sql_types/struct.Record.html)?)
> 
> # Minor Points
> * Link relevant doc's/guides for derives/configs/traits…
> * `ToSql` impl should also be database specific, because that variant only works for postgres
> * Shortly mention alternative strategies to implement `FromSql`/`ToSql` by forwarding to an existing impl (`<String as ToSql<Text, Sqlite>::to_sql("enum_value_as_string")`
> * `AsExpression` is required for binding a value as expression of the specified `SqlType` in a query (just clarify that a bit)
> * `ToSql` implements the conversation to the database specific transmit over wire format. It's a bit hard to write much about that, because that's at least for postgres/mysql not documented much... (Maybe find at least something here to link that's not libpq source code…)
> * I'm not sure if the section about the upcoming change of the `RawValue` type is required. We would update the guide to match the 2.0 release later.

Should discussion also be specified in the draft itself, so it would be easier to find what is left to be fixed?

### DIFF
--- a/guide_drafts/custom_types.md
+++ b/guide_drafts/custom_types.md
@@ -1,0 +1,155 @@
+# Custom types in Diesel
+
+Recently, in a personal project, I decided to use [diesel](http://diesel.rs/), an ORM for Rust language that supports some popular relational databases. My hands-on experience with ORM was quite limited at that time, and I first made an SQL schema without even analysing for supported features. Unfortunately, my beliefs in seamless type integration between an app and schema were broken quite soon. Consider the following SQL schema:
+
+```sql
+CREATE TYPE Language AS ENUM (
+    'en', 'de', 'ru'
+);
+
+CREATE TABLE translations (
+    word_id        INTEGER NOT NULL,
+    translation_id INTEGER NOT NULL,
+    language       Language NOT NULL,
+
+    PRIMARY KEY (word_id, translation_id)
+)
+```
+
+And after `diesel migration run` the following _schema.rs_ will be generated:
+
+```rust
+// src/schema.rs
+
+table! {
+    translations (word_id, translation_id) {
+        word_id -> Int4,
+        translation_id -> Int4,
+        language -> Language,
+    }
+}
+```
+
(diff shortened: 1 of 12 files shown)


guide_drafts/custom_types.md
```diff
@@ -0,0 +1,155 @@
...
+
+#[derive(Debug, PartialEq, DbEnum)]
+pub enum Language {
+    En,
+    Ru,
+    De,
+}
+```
+
+You can find complete sources used in this post [here](https://github.com/l4l/diesel-custom-types) and leave a comment on [reddit](https://www.reddit.com/r/rust/comments/gptvej/custom_types_in_diesel/).
```
weiznich, line 155: Any chance to add this code as example [here](https://github.com/diesel-rs/diesel/tree/master/examples)?

l4l: You mean repo with [sources](https://github.com/l4l/diesel-custom-types)? Yep, sure, add a commit to this pr?

weiznich: Yes exactly this repo. Adding it as subfolder to `examples/postgresql` should be everything that is required. 
Adding it as additional commit would be great.

l4l: Done

PR #2412, opened 2020-05-26
Migrate examples to edition 2018

JohnTitor:

PR #2415, opened 2020-05-28
Use macros via full path when generating a `schema.rs` file

weiznich:


weiznich:
Cherry picked to #2412

PR #2416, opened 2020-05-28
GitHub actions/backport

weiznich:
Back port of #2269 as proposed there


JohnTitor:
This is related to #2269, but I think we can also remove AppVeyor config, cannot we? Seems it's totally outdated and Actions now check on Windows as well.

diesel_compile_tests/Cargo.lock
```diff
@@ -0,0 +1,678 @@
+# This file is automatically @generated by Cargo.
```
JohnTitor, line 1: I think this is introduced unintentionally.

weiznich: I've added this lock file because otherwise some dependencies won't build with the old nightly version used by compile_test. The alternative would be to update specific crates manually as done for the MSRV ci run. As the compile-test live in their own workspace that should be fine.

JohnTitor: Ah, I see! Then it should be fine.

PR #2425, opened 2020-06-08
Backport to 1.4.x some PRs marked as "maybe backport"

alfonso-eusebio:
These are the backports to branch 1.4.x of some straightforward PRs that had been marked as "maybe backport".

I've adapted the fix to "properly decrement transaction_depth" so that it doesn't take into account error `DatabaseErrorKind::ReadOnlyTransaction`, which doesn't exist in 1.4.x. But I have included some minor visibility and helper function changes, even though they probably don't make a big different in this branch.

I haven't been able to run the tests locally, but I've compiled the changes against a local binary to make sure there were no compilation issues.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -16,7 +16,7 @@ byteorder = "1.0"
 diesel_derives = "~1.4.0"
 chrono = { version = "0.4", optional = true }
 libc = { version = "0.2.0", optional = true }
-libsqlite3-sys = { version = ">=0.8.0, <0.18.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
+libsqlite3-sys = { version = ">=0.8.0, <0.19.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
 mysqlclient-sys = { version = ">=0.1.0, <0.3.0", optional = true }
 pq-sys = { version = ">=0.3.0, <0.5.0", optional = true }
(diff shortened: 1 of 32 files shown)


weiznich:

Test failures are caused by the dotenv update. I think the simplest way to resolve this is to remove the version bump from this backport PR. I do not think it is worth to backport all necessary changes to fix this to 1.4.x as this is solely an internal/dev-dependency for the released crates.

alfonso-eusebio:
Not sure why tests are failing with deprecation errors on unchanged files. Yesterday's PR went through without issues.
Not sure if it's worth looking into getting rid of that Error::description() deprecation problem as part of this PR (?).

alfonso-eusebio:
@weiznich We were writing at the same time. I'm happy to remove the `dotenv` update or try to get rid of the `description()` usage, whatever you guys think is best.

weiznich:
I don't think `Error::description` is the problem here. See for example [this](https://github.com/diesel-rs/diesel/runs/749654567?check_suite_focus=true#step:28:129) test failure. It fails because there is some issue with `error-chain` which is an internal dependency of an really old `dotenv` version, which is used because of the usage of `diesel-infer-schema` in `diesel_tests`. It seems like we just cannot use two different `dotenv` versions here.

weiznich:
I will try to issue a release today or tomorrow

PR #2428, opened 2020-06-09
Fixed build error installing the diesel_cli with barrel-migrations

0xc0deface:
I found that `cargo install diesel_cli --no-default-features --features="barrel-migrations sqlite"` was not working due to something on crates using a very old version. So i tried installing from git master directly with this command which also failed: 
`cargo install --git https://github.com/diesel-rs/diesel.git diesel_cli --no-default-features --features="barrel-migrations sqlite"`

This was because barrel didnt have a specific feature for the requested backend selected. Unfortunately cargo wont let us conditionally add feature `barrel/sqlite3` et al from the `sqlite` feature when `barrel-migrations` is enabled. however, as barrel only supports one backend at a time, this isnt a huge issue and we can just make a `barrel-migrations` feature for each of the backends. This means that rather than specifying `--features="barrel-migrations sqlite"` you would just specify `--features="barrel-migrations-sqlite"`

To test this installs you can try installing directly from my fork with:
`cargo install --git https://github.com/0xc0deface/diesel.git diesel_cli --no-default-features --features=barrel-migrations-sqlite`

Cheers.


weiznich:
Duplicate of #2409?

0xc0deface:
Sure looks like they are doing similar things, without the baggage of the rest of the change though. Also interesting to see how people working separately can arrive at very similar results. 

There is an issue with this PR where it isn't activating the barrel feature of migration_internals. This is why its building but subsequently doesn't work when you try to run a migration.

Barrel will need to change the version number of diesel they use to be compatible with 2.0.0 before that error goes away, so this patch is less broken, but still broken.

I can fix that, and also make a PR on barrel to bump the diesel version, unless you would prefer to let this get handled by the existing PR?

weiznich:
Please coordinate those with @pksunkara

pksunkara:
@0xc0deface What other changes are you planning?

pksunkara:
You can see [this commit](https://github.com/reign-rs/barrel/commit/2043dd81cd66a0743318c94d72c5c5b6ccc358be) for the barrel related changes that need to make it work for my diesel PR.

0xc0deface:
@pksunkara well none really. The only reason to pursue this PR rather than yours is if yours will take longer to get approved and merged, as my change is more minimal. Do you have an ETA on when your PR will be merged?

pksunkara:
My PR is more for diesel 2.0. if you want something fast, you might backport the features change to 1.4.x branch

diesel_cli/Cargo.toml
```diff
@@ -49,7 +49,10 @@ default = ["postgres", "sqlite", "mysql"]
 postgres = ["diesel/postgres", "url", "uses_information_schema"]
 sqlite = ["diesel/sqlite"]
 mysql = ["diesel/mysql", "url", "uses_information_schema"]
-barrel-migrations = ["migrations_internals/barrel", "barrel"]
+barrel-migrations = ["barrel"]
```
pksunkara, line 52: Let's add `migrations_internals/barrel` here and merge this PR.

weiznich:
Closed as the barrel integration is removed.

PR #2443, opened 2020-06-28
Uses libpq

sanpii:
Hello,

I wrote this safe crate for `pq-sys`.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -18,7 +18,7 @@ chrono = { version = "0.4.1", optional = true }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.8.0, <0.19.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
 mysqlclient-sys = { version = "0.2.0", optional = true }
-pq-sys = { version = "0.4.0", optional = true }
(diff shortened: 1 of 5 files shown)


weiznich:
Without wanting to criticize your work I see several major points that are against this change:
* It replace a lot of reviewed unsafe code that we known to be in a good state with third party crate that contains bigger amounts of unreviewed unsafe code. (According to a quick grep diesel uses 29 `unsafe` blocks as part of the postgres connection implementation, while `libpq.rs` contains 133 unsafe blocks`). Additionally with only looking roughly over the corresponding code, I've found already [one](https://github.com/sanpii/libpq.rs/blob/35a7acb014c11b5485ec7d83948fb58f896564e0/src/connection/mod.rs#L21) safety critical soundness issue. (libpq [does not allow](https://www.postgresql.org/docs/current/libpq-threading.html) issuing commands from multiple threads at the same time. Implementing `Sync` for the `Connection` type would exactly allow that.
* The libpq crate has a bus factor of 1 and we do not control this code 
* It enlarges our dependency tree by adding another bigger dependency
* Unknown impact on the performance of `PgConnection` 

To summarize the points above: I'm not in favor of this change

pksunkara:
I like the intention behind this but I agree with the unsafe issue pointed out above.

PR #2455, opened 2020-07-12
Fix documentation for importing any and all in pg

### LINKED ISSUE #2447: Documentation import incorrect `diesel::expression`
On the page: https://docs.rs/diesel/1.4.5/diesel/pg/expression/dsl/fn.any.html
It notes: 
> You can import it specifically from `diesel::expression::any`, or glob import `diesel::dsl::*`

I tried using the `diesel::expression::any` but this does not exist.
```
error[E0432]: unresolved import `diesel::expression::any`
 --> main.rs:8:5
  |
8 | use diesel::expression::any;
  |     ^^^^^^^^^^^^^^^^^^^^^^^ no `any` in `expression`

error: aborting due to previous error
```
`use diesel::pg::expression::dsl::any;` does work.

So maybe the documentation has to be updated here.

ropottnik:
@weiznich I have now updated the documentations of importing `any` and `all` from `diesel/src/pg/expression/array_comparison.rs`. I wrote it as @ralpha suggested in the issue description. 

However, wouldn't it make more sense to offer `diesel::dsl::any` for the specific import? I'm asking because the original import documentation seemed unaware of `pg` as specific db backend. 

closes #2447

### DIFF
--- a/diesel/src/pg/expression/array_comparison.rs
+++ b/diesel/src/pg/expression/array_comparison.rs
@@ -9,7 +9,7 @@ use crate::sql_types::Array;
 ///
 /// As with most bare functions, this is not exported by default. You can import
 /// it specifically from `diesel::pg::expression::dsl::any`, or glob import
-/// `diesel::dsl::*`.
+/// `diesel::dsl::any`.
 ///
 /// # Example
 ///
--- a/diesel/src/pg/expression/array_comparison.rs
+++ b/diesel/src/pg/expression/array_comparison.rs
@@ -8,7 +8,7 @@ use crate::sql_types::Array;
 /// Creates a PostgreSQL `ANY` expression.
 ///
 /// As with most bare functions, this is not exported by default. You can import
-/// it specifically from `diesel::expression::any`, or glob import
+/// it specifically from `diesel::pg::expression::dsl::any`, or glob import
 /// `diesel::dsl::*`
 ///
 /// # Example
@@ -37,7 +37,8 @@ where
 /// Creates a PostgreSQL `ALL` expression.
 ///
 /// As with most bare functions, this is not exported by default. You can import
-/// it specifically as `diesel::dsl::all`.
+/// it specifically as `diesel::pg::expression::dsl::all`, or glob import
+/// `diesel::dsl::*`.
 ///
 /// # Example
 ///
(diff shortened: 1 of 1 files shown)


diesel/src/pg/expression/array_comparison.rs
```diff
@@ -8,7 +8,7 @@ use crate::sql_types::Array;
 /// Creates a PostgreSQL `ANY` expression.
 ///
 /// As with most bare functions, this is not exported by default. You can import
-/// it specifically from `diesel::expression::any`, or glob import
+/// it specifically from `diesel::pg::expression::dsl::any`, or glob import
 /// `diesel::dsl::*`
```
JohnTitor, line 12: nit:

```suggestion
/// `diesel::dsl::*`.
```

weiznich:
@ropottnik Thanks for working on this :+1:

> However, wouldn't it make more sense to offer `diesel::dsl::any` for the specific import?

Yes, that would make more sense in my opinion. (I will add this as suggestions).

> I'm asking because the original import documentation seemed unaware of pg as specific db backend.

It is quite likely that this documentation was written back then when diesel only supported postgres and didn't change since then.

diesel/src/pg/expression/array_comparison.rs
```diff
@@ -8,8 +8,8 @@ use crate::sql_types::Array;
 /// Creates a PostgreSQL `ANY` expression.
 ///
 /// As with most bare functions, this is not exported by default. You can import
-/// it specifically from `diesel::expression::any`, or glob import
-/// `diesel::dsl::*`
+/// it specifically from `diesel::pg::expression::dsl::any`, or glob import
+/// `diesel::dsl::*`.
```
weiznich, line 12:
```suggestion
/// `diesel::dsl::any`.
```

diesel/src/pg/expression/array_comparison.rs
```diff
@@ -37,7 +37,8 @@ where
 /// Creates a PostgreSQL `ALL` expression.
 ///
 /// As with most bare functions, this is not exported by default. You can import
-/// it specifically as `diesel::dsl::all`.
+/// it specifically as `diesel::pg::expression::dsl::all`, or glob import
+/// `diesel::dsl::*`.
```
weiznich, line 41:
```suggestion
/// `diesel::dsl::all`.
```

PR #2462, opened 2020-07-20
chore: add @generated header to schema file

### LINKED ISSUE #2454: Include @generated in generated schema to help rustfmt skip it
A feature was [recently merged into `rustfmt` that makes it aware of `@generated`](https://github.com/rust-lang/rustfmt/pull/4296). This is an informal convention that tools can use to recognise generated files. [This issue in the rustfmt repo](https://github.com/rust-lang/rustfmt/issues/3958) discusses other tools that include or recoginse `@generated`. Notably Cargo includes it in a comment at the top of `Cargo.lock`:

```toml
# This file is automatically @generated by Cargo.
# It is not intended for manual editing.
```

It would be convenient if the `schema.rs` generated by `diesel_cli` included this text in a comment at the top of the file so that `rustfmt` will skip it, avoiding churn/manually reverting the changes made by `cargo fmt`.

## Checklist

- [x] This issue can be reproduced on Rust's stable channel.

pleshevskiy:
Closes #2454

File should start with `// @generated` prefix

see [rustfmt config docs](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#format_generated_files)


diesel_cli/src/print_schema.rs
```diff
@@ -7,6 +7,8 @@ use std::error::Error;
 use std::fmt::{self, Display, Formatter, Write};
 use std::io::Write as IoWrite;
 
+const SCHEMA_HEADER: &str = "// @generated automatically by Diesel CLI.\r\n\r\n";
```
JohnTitor, line 10: I don't think `\r\n` is reasonable here. Should use `writeln!` or just use `\n` like other places?

pleshevskiy: removed additional carriage return char

PR #2463, opened 2020-07-20
Update macOS MySQL version

JohnTitor:
[The latest homebrew formula pulls 8.0.21](https://formulae.brew.sh/formula/) and it breaks our CI. This just updates the number to unblock other PRs but it'd be great if we could stop hard-coding it as a follow-up work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ on:
       - '0.[0-9]+.x'
       - '1.[0-9]+.x'
 
-
 name: CI Tests
 
 jobs:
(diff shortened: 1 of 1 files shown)


weiznich:
As I'm not a MacOS user I'm not sure how to write the script otherwise, but if there is a possibility to change that I'm in favor of such a change.

JohnTitor:
> but if there is a possibility to change that I'm in favor of such a change.

The problem here is formula has the latest v8.x MySQL only. I'll try to find a (hacky) workaround if I get some time.

PR #2466, opened 2020-07-26
Use version number from installed lib

JohnTitor:
Follow-up of #2463
Tested on my fork: https://github.com/JohnTitor/diesel/actions/runs/183373156


weiznich:
Thanks :+1:

PR #2478, opened 2020-08-07
Merge/diesel dynamic schema

weiznich:
I propose merging `diesel_dynamic_schema` into the diesel main repository as this makes it much easier to coordinate the necessary changes between those two crates to implement dynamic select clause support.


diesel_dynamic_schema/src/column.rs
```diff
@@ -0,0 +1,55 @@
...
+    table: T,
+    name: U,
+    _sql_type: PhantomData<ST>,
+}
+
+impl<T, U, ST> Column<T, U, ST> {
+    pub(crate) fn new(table: T, name: U) -> Self {
+        Self {
+            table,
+            name: name,
```
Razican, line 21: This should remove the _clippy_ warning.
```suggestion
            name,
```

weiznich:
@Razican Thanks for the review. I will try to address the open points after my holidays (so not for before September). 

> I also noticed that in the Cargo.toml file, the diesel dependency is shown as 1.0.0, but in the rest of the crates in the workspace, it's shown like this:

Good point. Can you add some comment in the corresponding `Cargo.toml`?

> Finally, as far as I can tell this does not include the changes in diesel-rs/diesel-dynamic-schema#10, right? That would personally be a really nice feature to have. I've been trying to make it work with the latest 2.0.0 diesel branch, but I'm too far from understanding the inner workings of diesel to make it compile.

That's intentionally as I first want to merge both repositories. After that we can start adding new features. I will likely do a followup PR with those changes later.

Issue #1669, opened 2018-04-29
I think that for an option library like uuid, its feature can be reexported at the same time.


driftluo:
Issue:

Diesel 1.1.* only support  uuid 0.5.*, and I use uuid need its v4 feature, so on my Cargo.toml will be:

```toml
[dependencies.uuid]
features = ["serde", "v4"]
version = "^0.5.0"

[dependencies.diesel]
features = ["postgres", "chrono", "uuid", "r2d2"]
version = "^1.1.1"
```

But diesel 1.2.* support to uuid 0.6.*, when I `cargo update`, uuid doesn't update to 0.6, then will Rustc will complain like follow:

```
error[E0277]: the trait bound `uuid::Uuid: diesel::Expression` is not satisfied
 --> src\models\article_tag_relation.rs:9:10
  |
9 | #[derive(Insertable, Debug, Clone, Deserialize, Serialize)]
  |          ^^^^^^^^^^ the trait `diesel::Expression` is not implemented for `uuid::Uuid`
  |
  = note: required because of the requirements on the impl of `diesel::Expression` for `&'insert uuid::Uuid`
  = note: required because of the requirements on the impl of `diesel::expression::AsExpression<diesel::sql_types::Uuid>` for `&'insert uuid::Uuid`
```

So, I think that for an option library like uuid, its feature can be reexported at the same time. In this way, you do not need to pay attention to the version of the corresponding library, let it upgrade at the same time with the diesel.

If you accept this view, I will submit a pull request to complete this function.

sgrif:
This wouldn't solve your original problem, as the `v4` feature would not be enabled since the versions don't match up.

driftluo:
Ok,  I looked at the source code, uuid is converted from `trait FromSql IntoSql`, not the corresponding type in the uuid library, there seems to be no way to inherit the feature, this is my idealization

sgrif:
I do not want to re-export every feature of every optional dependency we support

sgrif:
While I can appreciate the sentiment here, I don't think this approach makes sense in Diesel. For most applications, they will likely want to enable at least one feature on the other crates (usually `serde`). This requires syncing versions. There's really no way around it.

PR #2487, opened 2020-08-17
re-export the `chrono` crate from diesel

r-52:
This allows the user to use the same version of `chrono` that Diesel ships with. The user avoids another direct dependency that could be another version or a not-compatible version of `chrono`. On the other hand, re-exporting could cause a breaking change through `chrono` to the user.

The re-export is guarded with a feature flag and uses the already known `chrono` feature flag.


weiznich:
We don't want to reexport any of our dependencies. See #1669 for some reasoning.

PR #2488, opened 2020-08-21
Ran cargo-diet to reduce download size from crates.io

Razican:
This just runs `cargo-diet` in order to remove tests and examples from the crates.io packages. They are not needed to compile, and removing them reduces download time (and therefore compile time). This is especially the case on places with a bad or slow Internet connection.

The results of the diet are the following:

**diesel_cli:**
```
Saved 47% or 117.8 KB in 118 files (of 248.4 KB and 141 files in entire crate)
```

**diesel_derives:**
```
Saved 23% or 35.6 KB in 9 files (of 156.7 KB and 32 files in entire crate)
```

**diesel_dynamic_schema:**
```
Saved 21% or 6.4 KB in 4 files (of 29.6 KB and 13 files in entire crate)
```

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -11,6 +11,7 @@ repository = "https://github.com/diesel-rs/diesel"
 keywords = ["diesel", "migrations", "cli"]
 autotests = false
 edition = "2018"
+include = ["src/**/*", "LICENSE-*", "README.md"]
 
 [[bin]]
 name = "diesel"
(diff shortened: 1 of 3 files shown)


pksunkara (approved):

Razican:
Something that was brought to my attention is that removing tests from crates.io prevents crater from running them, and therefore finding regressions in the compiler that could affect diesel. It's something to be taken into account before merging this. We could try to also include tests in crates.io and remove everything else.

pksunkara:
The download time is probably in the scale of milliseconds. With the new information I would rather not merge this PR.

weiznich:
As diesels test require a special setup (running database, setting up a DATABASE_URL environment variable) I would assume that they are not run by crater and similar tools (or that they fail)

Issue #786, opened 2017-03-08
Add SSL support for mysql


Kintaro:
At work our SQL database uses SSL certificates for the connection and I'd love to use diesel, but for the time being I have to fall back to the normal mysql crate. Adding SSL would be really beneficial.

sgrif:
We should be preferring SSL connections already, but I agree that we should add the ability to specify `?sslmode=require` in the URL, similar to PG.

Kintaro:
I meant more in terms of providing client certificates and keys.

sgrif:
Yes, I will add support for that as well

Kintaro:
Nice. Thanks a bunch. Would really help. We're building a prototype for
some financial systems and it worms at the moment with the mysql crate. But
the added readability abf exchanfabimity for pg and sqlite would really
help to put it into prod.

marti1125:
Hello I can't connect with Heroku its required: 
All connections require SSL: sslmode=require.

yerke:
@sgrif Hi. Thank you for creating and maintaining Diesel. Do you think you will have time to finish this soon? I do understand that MySQL is not a priority, but it would be nice to have some clarity on whether this will be solved. Thank you.

p-alik:
@sgrif, if no one works on this issue yet, I'm willing to take over.

weiznich:
Closed by #2491

PR #2491, opened 2020-08-28
Issue 786 mysql ssl support

p-alik:
This PR aims to add ssl connection support for MySQL backend, which is requested in the issue #786 .
At this stage it allow setting of mysql option `MYSQL_OPT_SSL_MODE` only.
The connection could be established  with `MYSQL_OPT_SSL_MODE` set to `SSL_MODE_DISABLED`, `SSL_MODE_PREFERRED` and `SSL_MODE_REQUIRED`.

Tested with
`DATABASE_URL='mysql://***?ssl_mode=(disabled|required|preferred)' cargo test --manifest-path diesel/Cargo.toml --features mysql --no-default-features mysql`

https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html
https://dev.mysql.com/doc/c-api/8.0/en/c-api-encrypted-connections.html

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -17,7 +17,7 @@ byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.17.2, <0.24.0", optional = true, features = ["bundled_bindings"] }
-mysqlclient-sys = { version = "0.2.0", optional = true }
+mysqlclient-sys = { version = "0.2.2", optional = true }
 pq-sys = { version = "0.4.0", optional = true }
 quickcheck = { version = "1.0.3", optional = true }
 serde_json = { version = ">=0.8.0, <2.0", optional = true }
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -17,7 +17,7 @@ byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.17.2, <0.24.0", optional = true, features = ["bundled_bindings"] }
-mysqlclient-sys = { version = "0.2.2", optional = true }
+mysqlclient-sys = { version = "0.2.5", optional = true }
 pq-sys = { version = "0.4.0", optional = true }
 quickcheck = { version = "1.0.3", optional = true }
 serde_json = { version = ">=0.8.0, <2.0", optional = true }
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -48,13 +48,8 @@ impl RawConnection {
         let unix_socket = connection_options.unix_socket();
         let client_flags = connection_options.client_flags();
 
-        #[cfg(not(windows))]
-        // allow single_match because compiler complains about if after cfg
-        // error: attributes are not yet allowed on `if` expressions
-        #[allow(clippy::single_match)]
-        match connection_options.ssl_mode() {
-            Some(ssl_mode) => self.set_ssl_mode(ssl_mode),
-            _ => (),
+        if let Some(ssl_mode) = connection_options.ssl_mode() {
+            self.set_ssl_mode(ssl_mode)
         }
 
         unsafe {
@@ -190,7 +185,6 @@ impl RawConnection {
         self.did_an_error_occur()
     }
 
-    #[cfg(not(windows))]
     fn set_ssl_mode(&self, ssl_mode: mysqlclient_sys::mysql_ssl_mode) {
         let v = ssl_mode as u32;
         let v_ptr: *const u32 = &v;
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -48,11 +48,13 @@ impl RawConnection {
         let unix_socket = connection_options.unix_socket();
         let client_flags = connection_options.client_flags();
 
-        if connection_options.ssl_mode().is_some() {
-            let ssl_mode = connection_options
-                .ssl_mode()
-                .expect("Failure on unwrapping ssl_mode");
-            self.set_ssl_mode(ssl_mode)
+        #[cfg(not(windows))]
+        // allow single_match because compiler complains about if after cfg
+        // error: attributes are not yet allowed on `if` expressions
+        #[allow(clippy::single_match)]
+        match connection_options.ssl_mode() {
+            Some(ssl_mode) => self.set_ssl_mode(ssl_mode),
+            _ => (),
         }
 
         unsafe {
@@ -188,6 +190,7 @@ impl RawConnection {
         self.did_an_error_occur()
     }
 
+    #[cfg(not(windows))]
     fn set_ssl_mode(&self, ssl_mode: mysqlclient_sys::mysql_ssl_mode) {
         let v = ssl_mode as u32;
         let v_ptr: *const u32 = &v;
--- a/diesel/src/mysql/connection/raw.rs
+++ b/diesel/src/mysql/connection/raw.rs
@@ -48,6 +48,13 @@ impl RawConnection {
         let unix_socket = connection_options.unix_socket();
         let client_flags = connection_options.client_flags();
 
+        if connection_options.ssl_mode().is_some() {
+            let ssl_mode = connection_options
+                .ssl_mode()
+                .expect("Failure on unwrapping ssl_mode");
+            self.set_ssl_mode(ssl_mode)
+        }
+
         unsafe {
             // Make sure you don't use the fake one!
             ffi::mysql_real_connect(
@@ -179,6 +186,19 @@ impl RawConnection {
     fn next_result(&self) -> QueryResult<()> {
         unsafe { ffi::mysql_next_result(self.0.as_ptr()) };
         self.did_an_error_occur()
+    }
+
+    fn set_ssl_mode(&self, ssl_mode: mysqlclient_sys::mysql_ssl_mode) {
+        let v = ssl_mode as u32;
+        let v_ptr: *const u32 = &v;
+        let n = ptr::NonNull::new(v_ptr as *mut u32).expect("NonNull::new failed");
+        unsafe {
+            mysqlclient_sys::mysql_options(
+                self.0.as_ptr(),
+                mysqlclient_sys::mysql_option::MYSQL_OPT_SSL_MODE,
+                n.as_ptr() as *const std::ffi::c_void,
+            )
+        };
     }
 }
 
--- a/diesel/src/mysql/connection/url.rs
+++ b/diesel/src/mysql/connection/url.rs
@@ -8,7 +8,6 @@ use std::ffi::{CStr, CString};
 
 use crate::result::{ConnectionError, ConnectionResult};
 
-#[cfg(not(windows))]
 use mysqlclient_sys::mysql_ssl_mode;
 
 bitflags::bitflags! {
@@ -49,7 +48,6 @@ pub struct ConnectionOptions {
     port: Option<u16>,
     unix_socket: Option<CString>,
     client_flags: CapabilityFlags,
-    #[cfg(not(windows))]
     ssl_mode: Option<mysql_ssl_mode>,
 }
 
@@ -78,7 +76,6 @@ impl ConnectionOptions {
             _ => None,
         };
 
-        #[cfg(not(windows))]
         let ssl_mode = match query_pairs.get("ssl_mode") {
             Some(v) => {
                 let ssl_mode = match v.to_lowercase().as_str() {
@@ -125,7 +122,6 @@ impl ConnectionOptions {
             port: url.port(),
             unix_socket: unix_socket,
             client_flags: client_flags,
-            #[cfg(not(windows))]
             ssl_mode: ssl_mode,
         })
     }
@@ -158,7 +154,6 @@ impl ConnectionOptions {
         self.client_flags
     }
 
-    #[cfg(not(windows))]
     pub fn ssl_mode(&self) -> Option<mysql_ssl_mode> {
         self.ssl_mode
     }
@@ -298,7 +293,6 @@ fn unix_socket_tests() {
     );
 }
 
-#[cfg(not(windows))]
 #[test]
 fn ssl_mode() {
     let ssl_mode = |url| ConnectionOptions::parse(url).unwrap().ssl_mode();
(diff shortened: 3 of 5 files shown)


p-alik:
I've to improve this PR for sake of Windows and I suppose to do that next week.

p-alik:
Due to https://github.com/sgrif/mysqlclient-sys/issues/27 there is no way to set ssl-mode for MySQL connection on Windows in same way as on Linux/Unix yet.

Would it be appropriate to restrict setting of MySQL option `MYSQL_OPT_SSL_MODE` for `#[cfg(not(windows))]`? Any other suggestions?

ELD:
That would probably be a suitable alternative for now, since the corresponding Windows bindings don’t work right now.

diesel/Cargo.toml
```diff
@@ -17,7 +17,7 @@ byteorder = "1.0"
 chrono = { version = "0.4.1", optional = true }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.8.0, <0.19.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
-mysqlclient-sys = { version = "0.2.0", optional = true }
+mysqlclient-sys = { version = "0.2.4", optional = true }
```
weiznich, line 20: `mysqlclient-sys 0.2.2` is the first version that supports the necessary functions, therefore I do not see any reason to bump the version here to 0.2.4.

diesel/src/mysql/connection/raw.rs
```diff
@@ -179,6 +188,20 @@ impl RawConnection {
         unsafe { ffi::mysql_next_result(self.0.as_ptr()) };
         self.did_an_error_occur()
     }
+
+    #[cfg(not(windows))]
```
weiznich, line 192: Is this a fundamental restriction of `libmysqlclient` that ssl is not supported on windows or is this a restriction of the `libmysqlclient-sys` crate?

diesel/src/mysql/connection/raw.rs
```diff
@@ -47,6 +47,15 @@ impl RawConnection {
         let port = connection_options.port();
         let unix_socket = connection_options.unix_socket();
 
+        #[cfg(not(windows))]
+        // allow single_match because compiler complains about if after cfg
+        // error: attributes are not yet allowed on `if` expressions
+        #[allow(clippy::single_match)]
+        match connection_options.ssl_mode() {
```
weiznich, line 54: Can we make `ConnectionOptions::ssl_mode()` just unconditionally return `None` on windows instead of having `cfg`'s all over the code?

weiznich (commented):
I've left some comments, otherwise this is looking good :+1:

(Sorry for taking some time to do a review here, it took some time to catch up with everything after holiday.)

diesel/src/mysql/connection/raw.rs
p-alik: It is a restriction of `libmysqlclient-sys`. There is a PR https://github.com/sgrif/mysqlclient-sys/pull/28
which would make all `not(windows)` in this once checks obsolete.

weiznich: I would rather not like to introduce such special case handling then. The current situation with mysqlclient seems to be quite unfortunate as I cannot merge PR's there. As we've hit that problem already more than once we should definitively seek a solution for this. Maybe @sgrif can grant me the corresponding rights or transfer the repositories to the diesel organization?

Beside of that as a long term goal it would probably be great to replace libmysqlclient with a pure rust implementation.

p-alik: > Beside of that as a long term goal it would probably be great to replace libmysqlclient with a pure rust implementation.

You've mentioned [mysql](https://docs.rs/mysql/19.0.1/mysql/) crate in https://github.com/sgrif/mysqlclient-sys/issues/25#issuecomment-623016021 in regard to this matter. But is performance of the crate comparable to `libmysqlclient-sys`?

weiznich: I'm not sure if the performance is comparable. That's one of the many open questions that need to be answered before such an implementation can be merged. As starting point have a look at the benchmarks I've pushed [here](https://github.com/weiznich/diesel/tree/more/benches/diesel_bench) (Improvements are welcome)

biot023:
Hi -- I guess nothing happened with this?
Are there any workarounds for SSL-secured access?
I'm going to talk to our DBA service about maybe SSH tunnelling but I'm not confident that will work out...

weiznich:
@biot023 Commenting on old issues/PR's just to ask for updates is something that is highly discouraged. Please stop doing that if you have nothing meaningful (like how to resolve the issues outlined above) to contribute.

biot023:
> @biot023 Commenting on old issues/PR's just to ask for updates is something that is highly discouraged. Please stop doing that if you have nothing meaningful (like how to resolve the issues outlined above) to contribute.

It was a genuine question. Never mind, consider me suitably discouraged.

biot023:
Maybe not entirely discouraged.
I've been able to set up ProxySQL (https://proxysql.com) in a docker instance to handle the SSL bit of the connection.
I can just point any `diesel` services at that image.
Hope that's of some help to somebody else struggling with this.

Neo-Zhixing:
Maybe as a stopgap measure we should fork `libmysqlclient-sys` into the diesel org for now and merge all pending PRs there? There are no new commits in that repo since 2019 and @sgrif has been inactive since July 2021.  MySQL SSL support is a real use case, and doing that would unblock us on multiple issues.

weiznich:
@Neo-Zhixing There is no need to fork mysqlclient-sys. I have all merge and publish rights for that crate. Unfortunately I do not have the time to do a lot active work there while maintaining diesel at the same time, so this is basically waiting for someone laying out a basic plan about the future of mysqlclient-sys and someone who implements it.

thomasmost:
> I've been able to set up ProxySQL (https://proxysql.com) in a docker instance to handle the SSL bit of the connection.
> I can just point any diesel services at that image.
> Hope that's of some help to somebody else struggling with this.

@biot023 this is I think exactly what I'll need to do—you don't happen to have an example Dockerfile/diesel config that you'd feel comfortable sharing do you (scrubbed of anything sensitive, obviously)?

biot023:
> @biot023 this is I think exactly what I'll need to do—you don't happen to have an example Dockerfile/diesel config that you'd feel comfortable sharing do you (scrubbed of anything sensitive, obviously)?

Hi -- I really struggled with this. I got it working, as I recall, but then had an issue with tons of connections being established to the main server and never dropped -- took the whole site down in about 4 minutes, when I tested it!
It's probable that you could get it working but in our case I went past the point of sunken cost fallacy and we abandoned diesel for that particular project.
That's not throwing shade on diesel, btw -- I still use it whenever I can, it just didn't work out in this scenario for us.
Sorry I'm not more help, man.
Cheers,
   Doug.

weiznich:
@biot023  and @thomasmost Just to reiterate what I've written above: I would be more than happy to merge support for mysql ssl connections to diesel, but that requires work from someone. I'm happy to provide some pointers + discuss design, but implementation needs to be done by someone else.

thomasmost:
@weiznich Let's do it! What is your schedule like/what forum would you prefer for such a discussion?

weiznich:
@thomasmost I had a quick look at open questions here: The interface proposed in this PR looks good to me. There are the following open points from my point of view:

* [ ] Windows support
    * This is the largest open point
    * Requires changes to [`mysqlclient-sys`](https://github.com/sgrif/mysqlclient-sys), which may require restructuring the code there. That's the point that is quite fuzzy for me now.
* [ ] Rebasing the PR to the current master branch
* [ ] Tests?

thomasmost:
Got it! I will see if I can get in there and work this out.

diesel/Cargo.toml
weiznich: We definitively need to update the minimal supported mysqlclient-sys version here to something newer than even released. My old comment was written with the assumption that the old release also contains all necessary functions. 

Generally speaking: We try not to update the minimal supported mysqlclient-sys version as that is considered a breaking change, due to the fact that it links a native library and cargo only allows one version of `mysqlclient-sys` in your dependency tree.

p-alik:
Will rebase the feature branch.

diesel/Cargo.toml
p-alik: 04c4dc482d8cf3291213a74f810b9030078dd192 brings 0.2.2

diesel/src/mysql/connection/raw.rs
p-alik: Is it convenient to do that this way?
3b9fd78aa9f878bb7552bb1ffa831c89365bfa82

thomasmost:
p-alik how can I help get these tests passing?

diesel/src/mysql/connection/raw.rs
weiznich: @thomasmost That's the most important open issue. Someone with access to a windows development environment needs to update mysqlclient-sys.

thomasmost:
@weiznich now this is lookin pretty good

weiznich:
@thomasmost Yes the diesel side of this change looks good, beside the fact that this only supports linux/macos. I would like to solve the windows issue first before finally merging this. As already mentioned that would require updating the bindings in mysqlclient-sys, which requires access to bindgen on windows. That's unfortunately something I cannot help with other than merging PR's and cutting releases.
I image something like following here:
- [ ] Update the generated bindings in mysqlclient-sys via a PR there (that should likely update bindings for both variants)
- [ ] Update this PR to use the mysqlclient-sys version based on the previous PR to check everything works fine
- [ ] I will merge the mysqlclient-sys PR + cut a release there
- [ ] Update this PR again to use the now newly released mysqlclient-sys version
- [ ] Merge this PR

thomasmost:
So we are waiting on this PR then: https://github.com/sgrif/mysqlclient-sys/pull/28/files ?

weiznich:
I've totally forgot about that PR 🙈  Merged it and cut a new [`mysqlclient-sys` release](https://crates.io/crates/mysqlclient-sys) containing the corresponding changes.

That means the following points are open before this PR can be merged:
- [ ] Update the PR to use the new `mysqlclient-sys` release + update the code to work on all platforms
- [ ] Add documentation about this new feature to `MysqlConnection::establish`
- [ ] Add a changelog entry.

As a warning note for @thomasmost: This PR targets the master branch, which contains quite a few backward incompatible  changes compared to the 1.4.x release. Additionally we won't immediately cut a new release after this PR is merged, just to be clear about that upfront.

thomasmost:
Understood. Totally understand. Thanks for the update/attention to this @weiznich and @p-alik 🙏🏼

diesel/Cargo.toml
```diff
@@ -17,7 +17,7 @@ byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.17.2, <0.24.0", optional = true, features = ["bundled_bindings"] }
-mysqlclient-sys = { version = "0.2.0", optional = true }
+mysqlclient-sys = { version = "0.2.2", optional = true }
```
thomasmost, line 20:
```suggestion
mysqlclient-sys = { version = "0.2.5", optional = true }
```

diesel/src/mysql/connection/url.rs
```diff
@@ -8,6 +8,9 @@ use std::ffi::{CStr, CString};
 
 use crate::result::{ConnectionError, ConnectionResult};
 
+#[cfg(not(windows))]
```
thomasmost, line 11: all of these `cfg` flags can be retired now, yeah?

p-alik: yes.

weiznich (approved):
Looks good for me if the following things are added:

* A changelog entry [here](https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md#added) + adjusting the minimal supported mysqlclient-sys version under ["REMOVED"](https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md#removed)
* A note in the documentation of `MysqlConnection::establish` on the available options.

p-alik:
Added requested notes to the changelog.

thomasmost (approved):
LGTM!

thomasmost:
@weiznich one last follow-up; just so I understand, are we talking days, weeks, or months before this gets included in a release of the crate?

weiznich:
> one last follow-up; just so I understand, are we talking days, weeks, or months before this gets included in a release of the crate?

We generally do not give any estimates when a certain release happens other than: When it's done. Additionally as this PR is not merged yet (due to missing documentation) it's not certain if the next release will contain this feature or not.

> Until that time, any reason I can't get my app working with an (admittedly, unstable) build using the master branch as its dependency, i.e.

Please open an discussion thread about that with more details about what you understand under "can't get working". A PR is definitively not the right place to discuss such issues.

p-alik:
> as this PR is not merged yet (due to missing documentation) it's not certain if the next release will contain this feature or not.
24b612d08d2826627c98a17819aaef834feab711 aims to add missed documentation.

thomasmost:
> > Until that time, any reason I can't get my app working with an (admittedly, unstable) build using the master branch as its dependency, i.e.
> 
> Please open an discussion thread about that with more details about what you understand under "can't get working". A PR is definitively not the right place to discuss such issues.


I think you misunderstood me @weiznich — I wasn't asking for support lol — nvm

Thanks again for your work on this crate

yerke:
@p-alik @weiznich Thank you very much for working on this!

ToniIltanen:
Im having problems with the ssl_mode=DISABLED, it seems diesel is not passing it to the actual connection, thus the connections ends up with `SSL connection error: error:0A000102:SSL routines::unsupported protocol`, no matter if the ssl_mode=DISABLED is added on the connection string.

weiznich:
@ToniIltanen Please don't post the same issue in multiple places. You won't get an faster answer by doing that, you just burn time of other people by doing that. At least for me that signals that this issue can be moved down on my priority list.

If you really need a timely answer from this project you should consider a support contract otherwise you need to wait till someone has time to investigate your problem.

ToniIltanen:
@weiznich I myself dont need this fixed, it was just an observation i made when i tested diesel with my clients db. I already fixed the issua by upgrading the db. But if you wish, I can remove the issue if that is a better solution for you.

weiznich:
@ToniIltanen Again: That's no reason to post to an old semi-related issue. If you feel that this issue is related to your issue you can link it from your issue instead of posting here and pinging 7 other people as well. Stop doing that.

libsilverwolf:
Is this feature also usable with `diesel-cli` ?

I tried

```
DATABASE_URL="mysql://***:3306/port?ssl_mode=required" diesel print-schema
```

but got 

```
Could not connect to database via `mysql://***:3306/port?ssl_mode=required`: Connections using insecure transport are prohibited while --require_secure_transport=ON.
```

I'm using MariaDB.

PR #2495, opened 2020-09-04
Expose sqlite's sqlite3_create_collation()

z33ky:
This PR adds `diesel::sqlite::SqliteConnection::register_collation()`, which wraps `sqlite3_create_collation_v2()` (ref https://sqlite.org/c3ref/create_collation.html).
The registered collations can be used via the `COLLATE`-constraint on table columns, as is demonstrated in the tests.

### DIFF
--- a/diesel/src/sqlite/connection/functions.rs
+++ b/diesel/src/sqlite/connection/functions.rs
@@ -17,7 +17,7 @@ pub fn register<ArgsSqlType, RetSqlType, Args, Ret, F>(
     mut f: F,
 ) -> QueryResult<()>
 where
-    F: FnMut(&RawConnection, Args) -> Ret + Send + 'static,
+    F: FnMut(&RawConnection, Args) -> Ret + Send + 'static + std::panic::RefUnwindSafe,
     Args: FromSqlRow<ArgsSqlType, Sqlite> + StaticallySizedRow<ArgsSqlType, Sqlite>,
     Ret: ToSql<RetSqlType, Sqlite>,
     Sqlite: HasSqlType<RetSqlType>,
--- a/diesel/src/sqlite/connection/functions.rs
+++ b/diesel/src/sqlite/connection/functions.rs
@@ -45,8 +45,14 @@ pub fn register_aggregate<ArgsSqlType, RetSqlType, Args, Ret, A>(
     fn_name: &str,
 ) -> QueryResult<()>
 where
-    A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send,
-    Args: FromSqlRow<ArgsSqlType, Sqlite> + StaticallySizedRow<ArgsSqlType, Sqlite>,
+    A: SqliteAggregateFunction<Args, Output = Ret>
+        + 'static
+        + Send
+        + std::panic::UnwindSafe
+        + std::panic::RefUnwindSafe,
+    Args: FromSqlRow<ArgsSqlType, Sqlite>
+        + StaticallySizedRow<ArgsSqlType, Sqlite>
+        + std::panic::UnwindSafe,
     Ret: ToSql<RetSqlType, Sqlite>,
     Sqlite: HasSqlType<RetSqlType>,
 {
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -215,8 +215,10 @@ impl SqliteConnection {
         mut f: F,
     ) -> QueryResult<()>
     where
-        F: FnMut(Args) -> Ret + Send + 'static,
-        Args: FromSqlRow<ArgsSqlType, Sqlite> + StaticallySizedRow<ArgsSqlType, Sqlite>,
+        F: FnMut(Args) -> Ret + Send + 'static + std::panic::RefUnwindSafe,
+        Args: FromSqlRow<ArgsSqlType, Sqlite>
+            + StaticallySizedRow<ArgsSqlType, Sqlite>
+            + std::panic::UnwindSafe,
         Ret: ToSql<RetSqlType, Sqlite>,
         Sqlite: HasSqlType<RetSqlType>,
     {
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -242,6 +242,48 @@ impl SqliteConnection {
         functions::register_aggregate::<_, _, _, _, A>(&self.raw_connection, fn_name)
     }
 
+    /// Register a collation function.
+    ///
+    /// `collation` must always return the same answer given the same inputs.
+    /// If `collation` panics and unwinds the stack, the process is aborted, since it is used
+    /// across a C FFI boundary, which cannot be unwound across.
+    ///
+    /// If the name is already registered it will be overwritten.
+    ///
+    /// This method will return an error if registering the function fails, either due to an
+    /// out-of-memory situation or because a collation with that name already exists and is
+    /// currently being used in parallel by a query.
+    ///
+    /// The collation needs to be specified when creating a table:
+    /// `CREATE TABLE my_table ( str TEXT COLLATE MY_COLLATION )`,
+    /// where `MY_COLLATION` corresponds to `collation_name`.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     let conn = SqliteConnection::establish(":memory:").unwrap();
+    /// // sqlite NOCASE only works for ASCII characters,
+    /// // this collation allows handling UTF-8 (barring locale differences)
+    /// conn.register_collation("RUSTNOCASE", |rhs, lhs| {
+    ///     rhs.to_lowercase().cmp(&lhs.to_lowercase())
+    /// })
+    /// # }
+    /// ```
+    pub fn register_collation<F>(&self, collation_name: &str, collation: F) -> QueryResult<()>
+    where
+        F: Fn(&str, &str) -> std::cmp::Ordering + Send + 'static,
+    {
+        self.raw_connection
+            .register_collation_function(collation_name, collation)
+    }
+
     fn register_diesel_sql_functions(&self) -> QueryResult<()> {
         use crate::sql_types::{Integer, Text};
 
@@ -513,5 +555,76 @@ mod tests {
             .get_result::<Option<i32>>(&connection)
             .unwrap();
         assert_eq!(Some(3), result);
+    }
+
+    table! {
+        my_collation_example {
+            id -> Integer,
+            value -> Text,
+        }
+    }
+
+    #[test]
+    fn register_collation_function() {
+        use self::my_collation_example::dsl::*;
+
+        let connection = SqliteConnection::establish(":memory:").unwrap();
+
+        connection
+            .register_collation("RUSTNOCASE", |rhs, lhs| {
+                rhs.to_lowercase().cmp(&lhs.to_lowercase())
+            })
+            .unwrap();
+
+        connection
+            .execute(
+                "CREATE TABLE my_collation_example (id integer primary key autoincrement, value text collate RUSTNOCASE)",
+            )
+            .unwrap();
+        connection
+            .execute("INSERT INTO my_collation_example (value) VALUES ('foo'), ('FOo'), ('f00')")
+            .unwrap();
+
+        let result = my_collation_example
+            .filter(value.eq("foo"))
+            .select(value)
+            .load::<String>(&connection);
+        assert_eq!(
+            Ok(&["foo".to_owned(), "FOo".to_owned()][..]),
+            result.as_ref().map(|vec| vec.as_ref())
+        );
+
+        let result = my_collation_example
+            .filter(value.eq("FOO"))
+            .select(value)
+            .load::<String>(&connection);
+        assert_eq!(
+            Ok(&["foo".to_owned(), "FOo".to_owned()][..]),
+            result.as_ref().map(|vec| vec.as_ref())
+        );
+
+        let result = my_collation_example
+            .filter(value.eq("f00"))
+            .select(value)
+            .load::<String>(&connection);
+        assert_eq!(
+            Ok(&["f00".to_owned()][..]),
+            result.as_ref().map(|vec| vec.as_ref())
+        );
+
+        let result = my_collation_example
+            .filter(value.eq("F00"))
+            .select(value)
+            .load::<String>(&connection);
+        assert_eq!(
+            Ok(&["f00".to_owned()][..]),
+            result.as_ref().map(|vec| vec.as_ref())
+        );
+
+        let result = my_collation_example
+            .filter(value.eq("oof"))
+            .select(value)
+            .load::<String>(&connection);
+        assert_eq!(Ok(&[][..]), result.as_ref().map(|vec| vec.as_ref()));
     }
 }
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -234,8 +234,14 @@ impl SqliteConnection {
         fn_name: &str,
     ) -> QueryResult<()>
     where
-        A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send,
-        Args: FromSqlRow<ArgsSqlType, Sqlite> + StaticallySizedRow<ArgsSqlType, Sqlite>,
+        A: SqliteAggregateFunction<Args, Output = Ret>
+            + 'static
+            + Send
+            + std::panic::UnwindSafe
+            + std::panic::RefUnwindSafe,
+        Args: FromSqlRow<ArgsSqlType, Sqlite>
+            + StaticallySizedRow<ArgsSqlType, Sqlite>
+            + std::panic::UnwindSafe,
         Ret: ToSql<RetSqlType, Sqlite>,
         Sqlite: HasSqlType<RetSqlType>,
     {
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -15,6 +15,19 @@ use crate::result::*;
 use crate::serialize::ToSql;
 use crate::sql_types::HasSqlType;
 
+/// For use in FFI function, which cannot unwind.
+/// Print the message, ask to open an issue at Github and [`abort`](std::process::abort).
+macro_rules! assert_fail {
+    ($fmt:expr $(,$args:tt)*) => {
+        eprint!(concat!(
+            $fmt,
+            "If you see this message, please open an issue at https://github.com/diesel-rs/diesel/issues/new.\n",
+            "Source location: {}:{}\n",
+        ), $($args,)* file!(), line!());
+        std::process::abort()
+    };
+}
+
 #[allow(missing_debug_implementations, missing_copy_implementations)]
 pub struct RawConnection {
     pub(crate) internal_connection: NonNull<ffi::sqlite3>,
@@ -328,8 +341,7 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
                 if let &mut OptionalAggregator::Some(ref mut agg) = a_ptr {
                     agg
                 } else {
-                    eprintln!("We've written the aggregator to the aggregate context, but it could not be retrieved");
-                    std::process::abort();
+                    assert_fail!("We've written the aggregator to the aggregate context, but it could not be retrieved.");
                 }
             }
             None => {
@@ -382,8 +394,7 @@ extern "C" fn run_aggregator_final_function<ArgsSqlType, RetSqlType, Args, Ret, 
         match std::mem::replace(a, OptionalAggregator::None) {
             OptionalAggregator::Some(agg) => agg,
             OptionalAggregator::None => {
-                eprintln!("We've written to the aggregator in the xStep callback. If xStep was never called, then ffi::sqlite_aggregate_context() would have returned a NULL pointer");
-                std::process::abort();
+                assert_fail!("We've written to the aggregator in the xStep callback. If xStep was never called, then ffi::sqlite_aggregate_context() would have returned a NULL pointer.");
             }
         }
     });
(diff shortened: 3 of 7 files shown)


diesel/src/sqlite/connection/raw.rs
```diff
@@ -379,6 +400,40 @@ unsafe fn null_aggregate_context_error(ctx: *mut ffi::sqlite3_context) {
...
+) -> libc::c_int
+where
+    F: FnMut(&str, &str) -> std::cmp::Ordering + Send + 'static,
+{
+    unsafe {
+        let user_ptr = user_ptr as *mut F;
+        let f = match user_ptr.as_mut() {
+            Some(f) => f,
+            None => {
+                //FIXME
```
z33ky, line 419: I'm not really sure what to do here if sqlite passes us null pointers. Can it just be assumed that this will not happen (i.e. `unwrap()` here)?

z33ky:
I should mention, my use-case is allowing for proper case-insensitive matching, since the `NOCASE`-collation available with sqlite only works with ASCII (ref §7 https://www.sqlite.org/datatype3.html).
Perhaps this use-case can be mentioned in the method-documentation?

diesel/src/sqlite/connection/raw.rs
p-alik: It seems to me `unreachable!` in case of `None` could be appropriate. 
https://github.com/diesel-rs/diesel/blob/44d0096ae99d135a644a2cb26da11b6a5135de90/diesel/src/sqlite/connection/raw.rs#L307-L309

z33ky: It wouldn't be, since "A panic! across an FFI boundary is undefined behavior." (https://doc.rust-lang.org/nomicon/ffi.html#ffi-and-panics).

diesel/src/sqlite/connection/functions.rs
```diff
@@ -66,6 +66,13 @@ where
     Ok(())
 }
 
+pub fn register_collation<F>(conn: &RawConnection, fn_name: &str, f: F) -> QueryResult<()>
```
weiznich, line 69: It is not clear for me why we need a separate function here instead of calling `RawConnectionn::register_collation_function()` directly in the implementation of `SqliteConnection`

diesel/src/sqlite/connection/mod.rs
```diff
@@ -242,6 +242,39 @@ impl SqliteConnection {
...
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     let conn = SqliteConnection::establish(":memory:").unwrap();
+    /// conn.register_collation("RUSTNOCASE", |rhs, lhs| {
+    ///     rhs.to_lowercase().cmp(&lhs.to_lowercase())
+    /// })
+    /// # }
+    /// ```
+    pub fn register_collation<F>(&self, fn_name: &str, f: F) -> QueryResult<()>
+    where
+        F: FnMut(&str, &str) -> std::cmp::Ordering + Send + 'static,
```
weiznich, line 273: I think a `Fn` instead of `FnMut` communicates the assumption of the callback being deterministic much better.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -242,6 +242,39 @@ impl SqliteConnection {
...
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     let conn = SqliteConnection::establish(":memory:").unwrap();
+    /// conn.register_collation("RUSTNOCASE", |rhs, lhs| {
+    ///     rhs.to_lowercase().cmp(&lhs.to_lowercase())
+    /// })
+    /// # }
+    /// ```
+    pub fn register_collation<F>(&self, fn_name: &str, f: F) -> QueryResult<()>
```
weiznich, line 271:
```suggestion
    pub fn register_collation<F>(&self, collation_name: &str, collation: F) -> QueryResult<()>
```

as naming things matters

diesel/src/sqlite/connection/mod.rs
```diff
@@ -242,6 +242,39 @@ impl SqliteConnection {
         functions::register_aggregate::<_, _, _, _, A>(&self.raw_connection, fn_name)
     }
 
+    /// Register a collation function.
+    ///
+    /// `f` must always return the same answer given the same inputs.
+    ///
+    /// This method will return an error if registering the function fails.
```
weiznich, line 249: Either skip that or mention exactly how it could fail.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -242,6 +242,39 @@ impl SqliteConnection {
...
+    ///
+    /// The collation can be specified when creating a table:
```
weiznich, line 251:
```suggestion
    /// The collation needs to be specified when creating a table:
```

diesel/src/sqlite/connection/raw.rs
```diff
@@ -388,3 +443,11 @@ where
     let ptr = data as *mut F;
     unsafe { Box::from_raw(ptr) };
 }
+
+extern "C" fn destroy_boxed_collation_fn<F>(data: *mut libc::c_void)
```
weiznich, line 447: Is there any reason why we can't reuse the `destroy_boxed_fn` function here. Technically this function does not need to know anything about `F`, therefore we can just remove the constraints there and use it everywhere.

diesel/src/sqlite/connection/raw.rs
weiznich: According to https://www.sqlite.org/c3ref/create_collation.html we can be sure that we get our pouter to the callback here. Following that reasoning I would use `unreachable!` or something similar + a comment explaining why this cannot happen to indicate that this is guaranteed by other things.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -379,6 +400,40 @@ unsafe fn null_aggregate_context_error(ctx: *mut ffi::sqlite3_context) {
...
+        let f = match user_ptr.as_mut() {
+            Some(f) => f,
+            None => {
+                //FIXME
+                return 0;
+            }
+        };
+
+        //FIXME: check args
+        let rhs =
```
weiznich, line 425: We should have a comment here indicating that the provided strings are `utf8` because we specified that above and otherwise that would be a bug in sqlite.

weiznich (changes requested):
In general I'm in favor of this change. I've left some comments at places where I think things can be improved. Additionally such a change requires an entry to our changelog.

diesel/src/sqlite/connection/raw.rs
z33ky: Then I'd either go to "full undefined behavior" territory and use `std::hint::unreachable_unchecked()`, or stay on the safe route and call `std::process::abort()` (with an `eprintln!()` in front).
IMO `unreachable!()` gives the impression that we can just panic and it's fine. Which it may eventually be (rust-lang/rust#74990), but at the moment it is not.

diesel/src/sqlite/connection/functions.rs
z33ky: I was merely following the lead of `register()` and `register_aggregate()`.
Would this also apply to them? I can refactor that in this PR if desired.

diesel/src/sqlite/connection/mod.rs
z33ky: It will not safe-guard against violating that requirement, but that does not impair my use-case so I'll change it.

z33ky:
`sqlite3_create_collation()` can also remove collation function by passing `NULL` as callback. Shall I add a `SqliteConnection::unregister_collation()` for this?

diesel/src/sqlite/connection/mod.rs
weiznich: I should word it a bit differently: A callback that is `FnMut` but not `Fn` is quite likely violating the requirement. A callback that is `Fn` needs to actively do bad things (like using random numbers) to violate the requirement.

diesel/src/sqlite/connection/functions.rs
weiznich: We are doing a bunch of other stuff (checking and converting formats for example) for normal functions, therefore the wrapper functions here. That is not required for registering a collation. Therefore I would only remove the `register_collation` function here.

diesel/src/sqlite/connection/raw.rs
weiznich: Using `eprintln!` + `abort()` sounds like the better solution for me. Can you have a quick lock at the other callbacks for normal and aggregate functions and change that there as well?

weiznich:
> sqlite3_create_collation() can also remove collation function by passing NULL as callback. Shall I add a SqliteConnection::unregister_collation() for this?

The question is how often would such a function be used?

diesel/src/sqlite/connection/raw.rs
z33ky: Yes, will do!

diesel/src/sqlite/connection/mod.rs
z33ky: Also with a `Cell` (or friends), but I guess one usually doesn't use that so the point still mostly stands.

z33ky:
> The question is how often would such a function be used?

It guess it would only be of potential use when dropping tables or some such. It certainly won't find use from me (at least at the moment).
Can I gather from the inquisition that I shell leave it absent then?

z33ky:
I hit a snag while making `run_aggregator_step_function()` unwind-safe: `SqliteAggregateFunction::step()` takes a `&mut self`. The general idea is that when `step()` panics, little to no care may be taken that `self` restores temporary invalid state (see https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html#what-is-unwindsafe).
We can communicate that we expect the `SqliteAggregateFunction` to be `UnwindSafe`, but apparently `&mut T` is never considered `UnwindSafe`, even if `T: UnwindSafe`, (I'm guesing) since the object continues to exist after the panic.

We can wrap it in a [`AssertUnwindSafe`](https://doc.rust-lang.org/std/panic/struct.AssertUnwindSafe.html), which means that we assure the type system about `SqliteAggregateFunction`'s continue existence after panics on the user's behalf, or we can `std::process::abort()` as well if it panics.
We could add our own marker trait which the user has to implement. Implementing this trait would be a promise that it's fine to wrap in `AssertUnwindSafe` for the purpose of calling `step()`.

weiznich:
I think `AssertUnwindSafe` + a comment mentioning that the implementation is responsible for cleaning up stuff after a panic should be fine there.

z33ky:
That should address all requested changes.
On top it also does some light refactoring of the aggregator function handling.
I don't know if reducing the scope of `unsafe` is desired since it adds a bit of "noise" to open a few more, but smaller, `unsafe` scopes than before. That's its own commit though and I can easily remove it if not desired.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +313,160 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
             Some(&mut OptionalAggregator::Some(ref mut agg)) => agg,
             Some(mut a_ptr @ &mut OptionalAggregator::None) => {
                 ptr::write_unaligned(a_ptr as *mut _, OptionalAggregator::Some(A::default()));
                 if let &mut OptionalAggregator::Some(ref mut agg) = a_ptr {
                     agg
                 } else {
-                    unreachable!(
-                        "We've written the aggregator above to that location, it must be there"
-                    )
+                    eprintln!("We've written the aggregator to the aggregate context, but it could not be retrieved");
```
weiznich, line 327: Could you improve that error message saying something like: If you ever see this message, please open an issue at the diesel github repository?


(Maybe with an indication on which line this error happened?)

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +313,160 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
-            Err(e) => {
-                let msg = e.to_string();
-                ffi::sqlite3_result_error(ctx, msg.as_ptr() as *const _, msg.len() as _);
+    let mut aggregate_context = NonNull::new(aggregate_context as *mut OptionalAggregator<A>);
+    let aggregator = aggregate_context.as_mut().map(|a| {
+        let a = unsafe { a.as_mut() };
+        match std::mem::replace(a, OptionalAggregator::None) {
+            OptionalAggregator::Some(agg) => agg,
+            OptionalAggregator::None => {
+                eprintln!("We've written to the aggregator in the xStep callback. If xStep was never called, then ffi::sqlite_aggregate_context() would have returned a NULL pointer");
```
weiznich, line 381: Could you improve that error message saying something like: If you ever see this message, please open an issue at the diesel github repository?


(Maybe with an indication on which line this error happened?)

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +313,160 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
-            Err(e) => {
-                let msg = e.to_string();
-                ffi::sqlite3_result_error(ctx, msg.as_ptr() as *const _, msg.len() as _);
-            }
-            _ => (),
-        };
-    }
+    let args = unsafe { slice::from_raw_parts(value_ptr, num_args as _) };
+    let args = build_sql_function_args::<ArgsSqlType, Args>(args);
+    let mut aggregator = std::panic::AssertUnwindSafe(aggregator);
```
weiznich, line 340: I had some time to think a bit more about this, so here some reasoning why this actually should be OK and users do not need to care about this:

* We catch the panic and get a `Err`
* We return an error message to sqlite, so that sqlite will abort executing this aggregate, so the old instance will not be used anymore.
* On the next call to this aggregate function we will create a new instance of `A`, so we cannot have compromised state here either.

(Maybe add that as comment here)

From that point of view: I'm not sure if we really need to have bounds on `UnwindSafe` on the corresponding types.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +313,160 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
 where
-    F: FnMut(&RawConnection, &[*mut ffi::sqlite3_value]) -> QueryResult<SerializedValue>
-        + Send
-        + 'static,
+    F: Fn(&str, &str) -> std::cmp::Ordering + Send + 'static,
 {
+    let user_ptr = user_ptr as *const F;
+    let f = unsafe { user_ptr.as_ref() }.unwrap_or_else(|| {
+        eprintln!(
+            "An unknown error occurred. user_ptr is a null pointer. This should never happen."
```
weiznich, line 427: Could you improve that error message saying something like: If you ever see this message, please open an issue at the diesel github repository?

(Maybe with an indication on which line this error happened?)

diesel/src/sqlite/mod.rs
```diff
@@ -24,12 +24,19 @@ pub trait SqliteAggregateFunction<Args>: Default {
     /// The result type of the SQLite aggregate function
     type Output;
 
-    /// The `step()` method is called once for every record of the query
+    /// The `step()` method is called once for every record of the query.
+    ///
+    /// This is called through a C FFI, as such panics do not propagate to the caller. Panics are
+    /// caught and cause a return with an error value. The implementation must still ensure that
```
weiznich, line 30: As explained above I think we do not need to have that actually here, because I see no way how the state could be reused here.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -230,19 +255,15 @@ extern "C" fn run_custom_function<F>(
...
-                ffi::sqlite3_result_error(
-                    ctx,
-                    NULL_DATA_ERR.as_ptr() as *const _ as *const _,
-                    NULL_DATA_ERR.len() as _,
-                );
+                context_error_str(ctx, NULL_DATA_ERR);
                 return;
             }
         };
         match f(&conn, args) {
```
weiznich, line 262: We should also have an `std::panic::catch_panic` here if we refactor this at all.

weiznich (commented):
Looks really good, modulo some small stylistic things. Thanks for working on this :+1:

diesel/src/sqlite/connection/raw.rs
z33ky: Ah, I didn't really follow how the aggregator is initialized.
Unwind-safetly is still a useful property though because `finalize()` will be called whether or not `step()` errored out.

I'm not certain if having the `UnwindSafe`-bound actually does much in any case, since this `AssertUnwindSafe` is where the unwind-safety rubber hits the road. Though without it we'd need another `AssertUnwindSafe`-wrapping in `run_aggregator_final_function()` if the bound does not exist. And that may be relevant `Drop` impementations, even if we don't use the aggregator any further.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -115,8 +129,12 @@ impl RawConnection {
         num_args: usize,
     ) -> QueryResult<()>
     where
-        A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send,
-        Args: FromSqlRow<ArgsSqlType, Sqlite>,
+        A: SqliteAggregateFunction<Args, Output = Ret>
+            + 'static
+            + Send
+            + std::panic::UnwindSafe
```
z33ky, line 135: I've changed it to both `RefUnwindSafe` and `UnwindSafe`, since `::step()` passes by reference and `::terminate()` by value. As [stated before](https://github.com/diesel-rs/diesel/pull/2495#discussion_r487028556), I'm not sure if the `RefUnwindSafe`-bound for `::step()` provides any practical safety since we just use `AssertUnwindSafe` instead to pass the `mut&` across and `RefUnwindSafe` is only for non-`mut` references.

z33ky:
The failed CI job looks to me like a transient failure; it seems MySQL didn't start up correctly there.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -204,7 +251,8 @@ extern "C" fn run_custom_function<F>(
 ) where
     F: FnMut(&RawConnection, &[*mut ffi::sqlite3_value]) -> QueryResult<SerializedValue>
         + Send
-        + 'static,
+        + 'static
+        + std::panic::RefUnwindSafe,
```
weiznich, line 255: As this is a `FnMut` and not a `Fn` I'm would guess that `RefUnwindSafe` is not the right trait here. We do potentially mutate `F` here.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -230,23 +274,30 @@ extern "C" fn run_custom_function<F>(
...
+        match result {
+            Ok(Ok(value)) => value.result_of(ctx),
+            Ok(Err(e)) => {
                 let msg = e.to_string();
-                ffi::sqlite3_result_error(ctx, msg.as_ptr() as *const _, msg.len() as _);
+                context_error_str(ctx, &msg);
+            }
+            Err(_) => {
+                let msg = format!("{} panicked", std::any::type_name::<F>());
+                unsafe { context_error_str(ctx, &msg) };
```
weiznich, line 298: Is there any reason why we need a nested unsafe block here?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +343,155 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
 
-    ffi::sqlite3_result_error(
-        ctx,
-        NULL_AG_CTX_ERR.as_ptr() as *const _ as *const _,
-        NULL_AG_CTX_ERR.len() as _,
-    );
+    context_error_str(ctx, NULL_AG_CTX_ERR)
+}
+
+unsafe fn context_error_str(ctx: *mut ffi::sqlite3_context, error: &str) {
```
weiznich, line 437: We should cite https://sqlite.org/c3ref/result_blob.html here, stating that sqlite will copy our error message, so it's fine to free it afterwards. (Just to make that clear for any future person looking at this code.) 
Also we maybe can take a `NonNull<ffi::sqlite3_context>` here to make this function not `unsafe` to use anymore?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -265,12 +316,12 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
     num_args: libc::c_int,
     value_ptr: *mut *mut ffi::sqlite3_value,
 ) where
-    A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send,
-    Args: FromSqlRow<ArgsSqlType, Sqlite>,
+    A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send + std::panic::RefUnwindSafe,
+    Args: FromSqlRow<ArgsSqlType, Sqlite> + std::panic::UnwindSafe,
```
weiznich, line 320: I'm not sure why the `Args` need to be `UnwindSafe`. They get reconstructed everytime the aggregator is called, so a potential panic cannot corrupt anything there.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +343,155 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
+    };
 }
 
 extern "C" fn run_aggregator_final_function<ArgsSqlType, RetSqlType, Args, Ret, A>(
     ctx: *mut ffi::sqlite3_context,
 ) where
-    A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send,
-    Args: FromSqlRow<ArgsSqlType, Sqlite>,
+    A: SqliteAggregateFunction<Args, Output = Ret> + 'static + Send + std::panic::UnwindSafe,
+    Args: FromSqlRow<ArgsSqlType, Sqlite> + std::panic::UnwindSafe,
```
weiznich, line 390: We do not even construct the arguments here, so `UnwindSafe` is not required here.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +343,155 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
+    lhs_len: libc::c_int,
+    lhs_ptr: *const libc::c_void,
+    rhs_len: libc::c_int,
+    rhs_ptr: *const libc::c_void,
+) -> libc::c_int
 where
-    F: FnMut(&RawConnection, &[*mut ffi::sqlite3_value]) -> QueryResult<SerializedValue>
-        + Send
-        + 'static,
+    F: Fn(&str, &str) -> std::cmp::Ordering + Send + 'static,
```
weiznich, line 450: I think this should be restricted to `RefUnwindSafe` as we cannot mutate the closure.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -292,99 +343,155 @@ extern "C" fn run_aggregator_step_function<ArgsSqlType, RetSqlType, Args, Ret, A
...
+        // Depending on the eTextRep-parameter to sqlite3_create_collation_v2() the strings can
+        // have various encodings. register_collation_function() always selects SQLITE_UTF8, so the
+        // pointers point to valid UTF-8 strings (assuming correct behavior of libsqlite3).
+        (
+            str::from_utf8_unchecked(slice::from_raw_parts(rhs_ptr as *const u8, rhs_len as _)),
+            str::from_utf8_unchecked(slice::from_raw_parts(lhs_ptr as *const u8, lhs_len as _)),
+        )
+    };
+
+    // It doesn't matter if f is UnwindSafe, since we abort on panic.
```
weiznich, line 484: I think similar to the sql function handler we should just return an error here, instead of abort the program.

weiznich:
This looks almost good, noted some minor improvements otherwise I'm fine with this now.

weiznich (approved):

weiznich:
@TaKO8Ki This is another candidate for rebasing + fixing the remaining comments if your are interested

TaKO8Ki:
@weiznich 
I'm going to tackle this!

weiznich:
Likely relevant: https://github.com/rusqlite/rusqlite/pull/839/

weiznich:
Closed by #2564

PR #2500, opened 2020-09-08
Add support for dynamic select clauses to diesel_dynamic_schema

weiznich:
This commit adds support for constructing dynamic select clauses and
receiving values into a dynamically sized container. I've opted for
requiring a user provided type as value type in the dynamically sized
`DynamicRow<I>` container, because that type needs to handle all
possible values that can appear in the loaded row. The requirements for
such type are normally highly dependent on the actual use case,
therefore we do not provide a "covers all cases" container type yet. The
implementation leave open the possibility to add such a container type
by just setting a default type for the generic paramater there.

See the added test cases and documentation for examples how to use this feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,12 +253,11 @@ jobs:
           command: test
           args: --manifest-path diesel_tests/Cargo.toml --no-default-features --features "${{ matrix.backend }}"
 
-      - name: Run diesel_dynamic_schema tests"
-        if: matrix.backend == 'sqlite'
+      - name: Run diesel_dynamic_schema tests
         uses: actions-rs/cargo@v1
         with:
           command: test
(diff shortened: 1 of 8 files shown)


pksunkara (approved):

Razican:
I'm having small issues integrating this with my codebase. It seems that `Column<Any>` doesn't implement `Expression`, and therefore, I can't run `column.desc()` for example, to order things in descending order.

It seems related to `Any` not implementing `SingleValue`. I might just be a bit confused about the new traits.

weiznich:
Seems like we need to improve the documentation then. (Any help here would be welcome)
`SingleValue` is only implemented for sql fragments with a at compile time known type. So any expression with the type `Any` will never implement that trait.
You probably want to set the correct SQL type for your columns as this does not restrict the construction of a `DynamicSelectClause`.

Issue #2307, opened 2020-02-18
Add support for `CURRENT_DATE`


sgrif:
Given that we support `CURRENT_TIMESTAMP`, the lack of `CURRENT_DATE` feels like an omission. While you can hack it with `date(now)`, I keep finding myself wishing there was a direct way to reference this. Given that `now` means `CURRENT_TIMESTAMP`, I think there's precedent for `today` meaning `CURRENT_DATE`. This is supported by all backends we support (and is in the ISO standard I believe).

The code for `now` can be pretty much directly re-used for this

Darth-Revan:
Seems to be a reasonable addition to the API. I'd like to give it shot. I already looked into the code of `now` and it should be quite easy to adapt it.

weiznich:
Sure go. If there is anything unclear feel free to ping me on gitter or here.

Darth-Revan:
An implementation of `today` has been added in #2317

TaKO8Ki:
@weiznich @Darth-Revan 
Can I take over this PR #2317?

weiznich:
@TaKO8Ki Sure

PR #2502, opened 2020-09-08
Add support for CURRENT_DATE

TaKO8Ki:
I've taken over https://github.com/diesel-rs/diesel/pull/2317.
Closes: #2307

### DIFF
--- a/diesel_tests/tests/expressions/date_and_time.rs
+++ b/diesel_tests/tests/expressions/date_and_time.rs
@@ -230,6 +230,32 @@ fn today_executes_sql_function_current_date() {
 }
 
 #[test]
+#[cfg(feature = "mysql")]
+fn now_executes_sql_function_now() {
+    use self::has_timestamps::dsl::*;
+
+    let connection = connection();
+    setup_test_table(&connection);
+    connection
+        .execute(
+            "INSERT INTO has_timestamps (created_at) VALUES
+                       (NOW() - interval 1 day), (NOW() + interval 1 day)",
+        )
+        .unwrap();
+
+    let before_today = has_timestamps
+        .select(id)
+        .filter(created_at.lt(now))
+        .load::<i32>(&connection);
+    let after_today = has_timestamps
+        .select(id)
+        .filter(created_at.gt(now))
+        .load::<i32>(&connection);
+    assert_eq!(Ok(vec![1]), before_today);
+    assert_eq!(Ok(vec![2]), after_today);
+}
+
+#[test]
 #[cfg(feature = "sqlite")]
 fn now_executes_sql_function_now() {
     use self::has_timestamps::dsl::*;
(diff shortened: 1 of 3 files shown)


diesel_tests/tests/expressions/date_and_time.rs
```diff
@@ -134,6 +141,120 @@ fn now_can_be_used_as_nullable() {
...
+        .select(id)
+        .filter(created_at.lt(now))
+        .load::<i32>(&connection);
+    let after_today = has_timestamps
+        .select(id)
+        .filter(created_at.gt(now))
+        .load::<i32>(&connection);
+    assert_eq!(Ok(vec![1]), before_today);
+    assert_eq!(Ok(vec![2]), after_today);
+}
```
TaKO8Ki, line 256: Since we didn't have a test for `now` in MySQL, I've added this test.

pksunkara (approved):

diesel_tests/tests/expressions/date_and_time.rs
```diff
@@ -333,7 +428,7 @@ fn adding_interval_to_nullable_things() {
     assert_eq!(expected_data, actual_data);
 }
 
-#[cfg(not(feature = "mysql"))] // FIXME: Figure out how to handle tests that modify schema
+#[cfg(any(feature = "postgres", feature = "sqlite", feature = "mysql"))]
```
weiznich, line 431: Can you explain how you addressed the comment about tests modifying the schema on mysql? In detail: How does this prevent failing tests for cases where tables already exist?

TaKO8Ki: I'm sorry. I removed the comment by mistake.

weiznich: While that adds back the comment it also allows this function to run on mysql, therefore it does not really address my remark above.

TaKO8Ki: I've added this comment again!

weiznich: Again: It's not about the comment, it's about the behavior. If you try to run the tests twice against a mysql database they will fail because tables already exists.

TaKO8Ki: Ok, I misunderstood.

diesel_tests/tests/schema_dsl/structures.rs
```diff
@@ -84,7 +84,7 @@ where
 {
     fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
         out.unsafe_to_cache_prepared();
-        out.push_sql("CREATE TABLE ");
+        out.push_sql("CREATE TABLE IF NOT EXISTS ");
```
TaKO8Ki, line 87: @weiznich 
I think this is the best way to prevent failing tests for cases where tables already exist. If there's something wrong with this, I'd like to know what it is.

TaKO8Ki:
I'd appreciate it if someone could re-review this pull request.

diesel_tests/tests/schema_dsl/structures.rs
weiznich: That should work :+1:

Issue #2393, opened 2020-05-12
Update dev-dependencies


ignatenkobrain:
Hello,

I'm trying to package diesel for Fedora, but found that many dev-dependencies are outdated. Would it be possible to upgrade them?

Thanks in advance!

https://github.com/diesel-rs/diesel/blob/fe87242595ef70abe9a1f4a745f40d3c7110b1bb/diesel/Cargo.toml#L42-L43

https://github.com/diesel-rs/diesel/blob/master/diesel_cli/Cargo.toml#L43

https://github.com/diesel-rs/diesel/blob/master/diesel_derives/Cargo.toml#L19

https://github.com/diesel-rs/diesel/blob/master/diesel_migrations/Cargo.toml#L19

weiznich:
I'm open to some contribution here, otherwise that's not on top of my todo list now, so feel free to submit corresponding PR's.

ignatenkobrain:
I'll try to find time for that.

weiznich:
Fixed by #2405 and #2509

PR #2509, opened 2020-09-19
Update dev-dependencies

TaKO8Ki:
update `cfg-if`, `difference`, `tempfile` and `regex`. cc #2393


pksunkara (approved):

JohnTitor (approved):
Generally, we don't need to update some deps (e.g. cfg-if or tempfile) when we follow caret requirements. Anyway it's a minor issue on dev-deps, looks fine to me.

PR #2512, opened 2020-09-21
Generate `.keep` instead of `.gitkeep`. 

TaKO8Ki:
Since https://github.com/diesel-rs/diesel/pull/1878 is not completed and has conflicts, I've taken over it.
closes https://github.com/diesel-rs/diesel/issues/1399,  closes https://github.com/diesel-rs/diesel/pull/1878

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -230,12 +230,13 @@ fn create_migrations_dir(matches: &ArgMatches) -> DatabaseResult<PathBuf> {
         // older version of diesel_cli that set a `.gitkeep` instead of a `.keep` file.
         // TODO: remove this after a few releases
         if let Ok(read_dir) = fs::read_dir(&dir) {
-            if let Some(dir_entry) = read_dir
-                .filter_map(|x| x.ok())
-                .find(|x| match x.file_type() {
-                    Ok(file_type) => file_type.is_file() && x.file_name() == ".gitkeep",
-                    Err(_) => false,
-                })
+            if let Some(dir_entry) =
+                read_dir
+                    .filter_map(|entry| entry.ok())
+                    .find(|entry| match entry.file_type() {
+                        Ok(file_type) => file_type.is_file() && entry.file_name() == ".gitkeep",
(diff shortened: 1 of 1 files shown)


pksunkara (approved):

diesel_cli/src/main.rs
```diff
@@ -225,7 +225,24 @@ fn create_migrations_dir(matches: &ArgMatches) -> DatabaseResult<PathBuf> {
...
     };
 
-    if !dir.exists() {
+    if dir.exists() {
+        // This is a cleanup code for migrating from an
+        // older version of diesel_cli that set a `.gitkeep` instead of a `.keep` file.
+        // TODO: remove this after a few releases
+        if let Some(dir_entry) = fs::read_dir(&dir)
+            .unwrap()
+            .filter(|x| x.is_ok())
```
weiznich, line 234: This should probably use `.filter_map` instead.

diesel_cli/src/main.rs
```diff
@@ -225,7 +225,24 @@ fn create_migrations_dir(matches: &ArgMatches) -> DatabaseResult<PathBuf> {
             .join("migrations"),
     };
 
-    if !dir.exists() {
+    if dir.exists() {
+        // This is a cleanup code for migrating from an
+        // older version of diesel_cli that set a `.gitkeep` instead of a `.keep` file.
+        // TODO: remove this after a few releases
+        if let Some(dir_entry) = fs::read_dir(&dir)
```
weiznich, line 232: As this is purly there because we do want to do some cleanup here we should never fail here. So do not use any code that may panic (so no `unwrap`'s) in the following code.

weiznich (changes requested):

diesel_cli/src/main.rs
TaKO8Ki: I've fixed this!

diesel_cli/src/main.rs
TaKO8Ki: I've just removed `unwrap`!

PR #2514, opened 2020-09-23
Implement combination queries (`UNION`, `INTERSECT` & `EXCEPT`)

gobanos:
A basic impl for #33


diesel/src/query_builder/combination_clause.rs
```diff
@@ -0,0 +1,206 @@
...
+pub struct Combination<Combinator, Rule, Source, Rhs> {
+    combinator: Combinator,
+    duplicate_rule: Rule,
+    source: Source,
+    rhs: Rhs,
+}
+
+impl<Combinator, Rule, Source, Rhs> Combination<Combinator, Rule, Source, Rhs> {
+    /// Create a new combination
+    pub fn new(combinator: Combinator, duplicate_rule: Rule, source: Source, rhs: Rhs) -> Self {
```
weiznich, line 25:
```suggestion
    pub(crate) fn new(combinator: Combinator, duplicate_rule: Rule, source: Source, rhs: Rhs) -> Self {
```

diesel/src/query_builder/combination_clause.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(())
+    }
+}
+
+impl<Combinator, Rule, Source, Rhs, Conn> RunQueryDsl<Conn>
+    for Combination<Combinator, Rule, Source, Rhs>
+{
+}
+
+impl<Combinator, Rule, Source, Rhs> QueryDsl for Combination<Combinator, Rule, Source, Rhs> {}
```
weiznich, line 66: I don't think that is currently useful as it does not provide any actual method from `QueryDsl` yet. For that we would need to implement some of the traits from `query_dsl::methods`

diesel/src/query_builder/combination_clause.rs
```diff
@@ -0,0 +1,206 @@
...
+    type Values = InsertFromSelect<Self, T::AllColumns>;
+
+    fn values(self) -> Self::Values {
+        InsertFromSelect::new(self)
+    }
+}
+
+/// Extension trait to combine queries using a combinator like `UNION`, `INTERSECT` or `EXPECT`
+/// with or without `ALL` rule for duplicates
+pub trait SelectCombinationQueryMethods: Query + Sized {
```
weiznich, line 82: Probably name this `CombineDsl` and move it to `diesel::query_dsl`?

diesel/src/query_builder/combination_clause.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Combine two queries using a SQL `EXCEPT ALL`
+    fn except_all<Rhs>(self, rhs: Rhs) -> Combination<Except, All, Self, Rhs>
+    where
+        Rhs: Query<SqlType = Self::SqlType>,
+    {
+        Combination::new(Except, All, self, rhs)
+    }
+}
+
+impl<T: Query> SelectCombinationQueryMethods for T {}
```
weiznich, line 157: We probably want to implement that for `T: Table`, `SelectStatement` and `BoxedSelectStatement` explicitly.

diesel/src/query_builder/mod.rs
```diff
@@ -11,6 +11,7 @@ mod clause_macro;
 
 mod ast_pass;
 pub mod bind_collector;
+pub mod combination_clause;
```
weiznich, line 14: We do not want to expose that as public API. The necessary traits to constructs such queries should be exposed via `diesel::query_dsl::QueryDsl`. Additionally there should be helper type defs in `diesel::dsl` to refer to the constructed types.

weiznich (commented):
Left some initial comments. Thanks for working on this :+1:

Beside of fixes for the actual comments I would like to see at least some test cases and maybe one compile fail tests demonstrating that everything works as expected.

diesel/src/query_dsl/combine_dsl.rs
```diff
@@ -0,0 +1,246 @@
...
+/// with or without `ALL` rule for duplicates
+pub trait CombineDsl {
+    /// What kind of query does this type represent?
+    type Query: Query;
+
+    /// Combine two queries using a SQL `UNION`
+    ///
+    /// # Examples
+    /// ```rust
+    /// # #[macro_use] extern crate diesel;
```
weiznich, line 17: Minor nit: `#[macro_use]` is not required on master anymore.

diesel/src/query_dsl/combine_dsl.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// let expected_data = vec![
+    ///     None,
+    ///     Some(String::from("Jack")),
+    ///     Some(String::from("Tess")),
+    ///     Some(String::from("Sean")),
+    /// ];
+    /// assert_eq!(Ok(expected_data), data);
+    /// # }
+    /// ```
+    fn union<Rhs>(self, rhs: Rhs) -> Combination<Union, Distinct, Self::Query, Rhs::Query>
```
weiznich, line 38: `diesel::dsl` should reexport helper types for each of those methods, so that we can write the return type here as `diesel::dsl::Union<Self, Rhs>` or `diesel::dsl::UnionAll<Self, Rhs>` (This allows potential users to refer easier to those types)

weiznich (commented):
This looks almost good now. I've left a few small suggestions. Beside of that the following things need to be done before merging this PR:

- [ ] Add a changelog entry [here](https://github.com/diesel-rs/diesel/blob/a4b8031f6a096c8ea4032e7d8f05711d7f0dea1a/CHANGELOG.md#added)
- [ ] Add a few test cases to diesel_tests
- [ ] Add at least one test case to diesel_compile_tests, showing that a mismatch in where clause types will fail to compile.

diesel/src/query_builder/combination_clause.rs
gobanos: I hit a limitation when trying to implement an `ORDER BY` for combination queries, as explained in [Pg docs](https://www.postgresql.org/docs/13/sql-select.html#SQL-ORDERBY) : 
> A limitation of this feature is that an ORDER BY clause applying to the result of a UNION, INTERSECT, or EXCEPT clause can only specify an output column name or number, not an expression.

That means the following query is invalid (`ERROR:  42P01: missing FROM-clause entry for table "users"`) : 
```sql
SELECT "users"."name"
FROM "users"
UNION
SELECT "animals"."name"
FROM "animals"
WHERE "animals"."name" IS NOT NULL
ORDER BY "users"."name"; -- "name" alone or 1 would work
```
Using the column name only doesn't look like a good idea because that can lead to conflicts, so that leave us with the numerical ordering. Any clues on how to get the ordering of a column in a select clause ?

weiznich: What about just not implement `OrderDsl` for the initial implementation? I mean the first version supporting this in diesel must not be complete, but can have some unpolished cases.
That written, I think correctly implementing that would require something like complete aliasing support for column names, in such a way that this columns can also be used in order/where clauses. At least the same infrastructure enabling something like that would be useful here. I would assume that this is something that will not implemented in the next few version, so I would just go with an initial version that does not support ordering for now.

gobanos: Ordering would be really handy in tests, what do you think of an undocumented `.positional_order_by(1.asc())` ?
If I can make if work, of course :wink:

weiznich: That would be a good solution, as long as it is clear that this method is not part of the public API. Maybe just something like done with `group_by` on the 1.x branch here: https://github.com/diesel-rs/diesel/blob/bec389a1ea329c161fc7a8fb8df8bea29a3432bf/diesel/src/query_dsl/group_by_dsl.rs
(So `#[doc(hidden)]` + an explicit comment in the source code that this is not part of the public API.)

diesel_compile_tests/tests/compile-fail/combinations_require_same_sql_types.rs
```diff
@@ -0,0 +1,39 @@
+#[macro_use]
```
weiznich, line 1: I would prefer to have this as `ui` test, as we plan to migrate all our compile tests to `ui`/`trybuild` at some point in time

weiznich (commented):
Implementation wise this looks really good now. I've left a bunch of suggestions how to improve some minor things. Shouldn't be hard to fix them

diesel/src/query_builder/combination_clause.rs
```diff
@@ -0,0 +1,229 @@
...
+    {
+        fn walk_ast(&self, mut out: AstPass<Sqlite>) -> QueryResult<()> {
+            self.source.walk_ast(out.reborrow())?;
+            self.combinator.walk_ast(out.reborrow())?;
+            self.duplicate_rule.walk_ast(out.reborrow())?;
+            self.rhs.walk_ast(out.reborrow())?;
+            Ok(())
+        }
+    }
+}
```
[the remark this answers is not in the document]
gobanos: > Have exactly one shared QueryFragment<DB> impl for CombanationClause. It should be possible to share this, as CombinationSupportedBy restricts which clauses are possible.
I've split the impls to support parenthesis around `Rhs` in supported backend (Pg & Mysql). Should I impl a trait on each backend to optionally add parenthesis ?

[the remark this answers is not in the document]
weiznich: What about wrapping `rhs` internally into some wrapper that implements `QueryFragment` in such a way that for `mysql` + `postgres` the argument is wrapped into parenthesis, but not for `sqlite`?

weiznich (approved):

weiznich (approved):

weiznich:
Thanks again :+1:

PR #2520, opened 2020-09-28
Fix #2140

### LINKED ISSUE #2140: Missing parenthesis lead to broken SQL queries
### Versions

- **Rust:** rustc 1.36.0 (a53f9df32 2019-07-03)
- **Diesel:** latest master (0cacfa080183076669112a5ccb192d9c30b8ac15)
- **Database:** psql (11.1, server 11.4 (Debian 11.4-1.pgdg90+1))
- **Operating System** Debian 11.4-1.pgdg90+1

### Feature Flags

- **diesel:** postgres

## Problem Description

Missing parentheses around equality operator lead to malformed SQL queries in the best case and wrong SQL queries silently returning incorrect results in the worst case.

Build a query which corresponds to something like this:

```sql
SELECT (1 > 0) = true
```

Translating this to diesel, the code will look like this:

```rust
numbers.select((n.gt(0)).eq(true));
```

Full runnable test can be found in my `operator-precedence-bug` branch of diesel. Here is the full test https://github.com/DNNX/diesel/compare/master...operator-precedence-bug which can be run with `(cd diesel_tests && cargo test --features "postgres" --no-default-features expressions::test_operator_precedence)`.

Green test.

```
➜  diesel git:(operator-precedence-bug) ✗ (cd diesel_tests && cargo test --features "postgres" --no-default-features expressions::test_operator_precedence)
   Compiling diesel_tests v0.1.0 (/Users/dnnx/work/diesel/diesel_tests)
    Finished dev [unoptimized + debuginfo] target(s) in 12.26s
     Running /Users/dnnx/work/diesel/target/debug/deps/integration_tests-514980afda54a7bd

running 1 test
test expressions::test_operator_precedence ... FAILED

failures:

---- expressions::test_operator_precedence stdout ----
thread 'expressions::test_operator_precedence' panicked at 'assertion failed: `(left == right)`
  left: `Ok(true)`,
 right: `Err(DatabaseError(__Unknown, "syntax error at or near \"=\""))`', diesel_tests/tests/expressions/mod.rs:469:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    expressions::test_operator_precedence

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 309 filtered out

error: test failed, to rerun pass '--test integration_tests'
```

No.

### Steps to reproduce

1. Check our my branch or apply the diff below to latest diesel master.
2. Run `(cd diesel_tests && cargo test --features "postgres" --no-default-features expressions::test_operator_precedence).

```diff
diff --git a/diesel_tests/tests/expressions/mod.rs b/diesel_tests/tests/expressions/mod.rs
index e6cd0b1e..2dbd584c 100644
--- a/diesel_tests/tests/expressions/mod.rs
+++ b/diesel_tests/tests/expressions/mod.rs
@@ -454,3 +454,17 @@ fn test_arrays_b() {
 
     assert_eq!(value, vec![7, 14]);
 }
+
+#[test]
+fn test_operator_precedence() {
+    use self::numbers::columns::*;
+    use self::numbers::table as numbers;
+
+    let connection = connection();
+    connection
+        .execute("INSERT INTO numbers (n) VALUES (2)")
+        .unwrap();
+    let source = numbers.select((n.gt(0)).eq(true));
+
+    assert_eq!(Ok(true), source.first(&connection));
+}
```

### Additional notes

I already raised this issue in https://github.com/diesel-rs/diesel/issues/2133#issuecomment-517470349, but I probably failed to express myself very clearly. 

The big problem with this bug is that it may result in incorrectly working SQL queries, in particular in Mysql. Mysql is less strict about queries, so it parses things like `select 2=2=2=2` just fine, whereas Postgres rejects such queries.

I tried to fix the bug myself, but I don't even know where to start. I think adding parentheses around every operator might be a solution, but there's one subtle issue with it: diesel also uses `eq` to generate `SET` part of the `UPDATE` statements - and there having parentheses around `x = <expr>` in `UPDATE foo SET x = <expr>` will be a problem.

## Checklist

weiznich:
Include parenthesis around every infix operation provided by one of the
`ExpressionMethod` traits.

It is not possible to include this directly into the `infix_operator!`
macro due several SQL constructs:

* `a LIKE b` can have optional `ESCAPE` clauses (same for `NOT LIKE`,
    `ILIKE` eg). Unconditionally putting a parenthesis around `(a LIKE b)` is no valid SQL
* `a BETWEEN l AND u` is implemented a set of two infix operators,
    `BETWEEN` and `AND`, so putting the parenthesis directly into the
    operator macro would lead to `x BETWEEN (l AND u)` which is not valid
    SQL

Instead I've opted for adding explicit parenthesis in the corresponding
`ExpressionMethod` (and similar traits implementations. As part of this
change I've changed all the return types there to use some type from
`diesel::dsl`, so that potential users have an easier time to figure out
what types to use if they have to write some where clause or something
like that.

While changing the return types to `diesel::dsl` types I noticed  and
fixed the following issues:

* Not all methods had an equivalent type in `diesel::dsl`, sometimes
    because it was missing, sometimes because the postgres specific types
    weren't correctly exported there. Both is fixed now
* Some types were not anymore compatible with the actual methods. A
    notable example is `dsl::And` and `BoolExpressionMethods::and` where the
    later returned a potentially nullable type depending on the input types,
    while the first was only valid for not nullable right sides. Fixing that
    was a bit more complicated:
* I've opted for making the return type of `BoolExpressionMethods::and`
          (and `or`) unconditonally nullable. This prevents having an explicit
         `ST` parameter on `dsl::And`
* This in turn requires that we accept a non nullable expression in
          places where a nullable expression is expected. Technically this
          was already possible, but would require explicitly calling
          `.nullable()` on affected expressions. Given that `.and()` and
          `.or()` are really common, that seemed not to be a good solution.
          As result I've changed the `AsExpression` impl for `T: Expression`
          in such a way that for any `T` with `T::SqlType: SqlType<IsNull =
          is_nullable::NotNull>` there is now an impl
          `AsExpression<Nullable<T::SqlType>>` so that such expressions are
          also accepted in nullable contexts.

### DIFF
--- a/diesel/src/expression/functions/date_and_time.rs
+++ b/diesel/src/expression/functions/date_and_time.rs
@@ -1,7 +1,8 @@
 use crate::backend::Backend;
-use crate::expression::coerce::Coerce;
 use crate::expression::functions::sql_function;
-use crate::expression::{AsExpression, Expression, ValidGrouping};
+#[cfg(feature = "postgres")]
+use crate::expression::{coerce::Coerce, AsExpression};
+use crate::expression::{Expression, ValidGrouping};
 use crate::query_builder::*;
 use crate::result::QueryResult;
 use crate::sql_types::*;
@@ -46,14 +47,6 @@ sql_function! {
     fn date(expr: Timestamp) -> Date;
 }
 
-impl AsExpression<Nullable<Timestamp>> for now {
-    type Expression = Coerce<now, Nullable<Timestamp>>;
-
-    fn as_expression(self) -> Self::Expression {
-        Coerce::new(self)
-    }
-}
-
 #[cfg(feature = "postgres")]
 impl AsExpression<Timestamptz> for now {
     type Expression = Coerce<now, Timestamptz>;
@@ -92,11 +85,3 @@ impl_selectable_expression!(today);
 
 operator_allowed!(today, Add, add);
 operator_allowed!(today, Sub, sub);
-
-impl AsExpression<Nullable<Date>> for today {
-    type Expression = Coerce<today, Nullable<Date>>;
-
-    fn as_expression(self) -> Self::Expression {
-        Coerce::new(self)
-    }
-}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,9 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   so non generic code does not require any change. For generic code you likely need to
   replace a trait bound on `Queryable<ST, DB>` with a trait bound on `FromSqlRow<ST, DB>`
   and a bound to `QueryableByName<DB>` with `FromSqlRow<Untyped, DB>`.
+  
+* Diesel's dsl now accept not nullable expressions in positions where nullable expressions 
+  are expected, without needing to call `.nullable()` explicitly
 
 
 ### Fixed
@@ -159,6 +162,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 * We've refactored our type level representation of nullable values. This allowed us to
   fix multiple long standing bugs regarding the correct handling of nullable values in some
   corner cases (#104, #2274)
+  
+* Parenthesis are now inserted around all infix operations provided by diesel's `ExpressionMethods` traits
 
 ### Deprecated
 
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -108,7 +108,10 @@ where
     type InExpression = Many<T::Expression>;
 
     fn as_in_expression(self) -> Self::InExpression {
-        let expressions = self.into_iter().map(AsExpression::as_expression).collect();
+        let expressions = self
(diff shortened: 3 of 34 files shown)


weiznich:
@diesel-rs/reviewers This is ready for review. If CI shows any more failing tests because changed debug_query output I will fix them on monday.

diesel/src/expression/functions/date_and_time.rs
```diff
@@ -46,29 +45,21 @@ sql_function! {
...
-
-    fn as_expression(self) -> Self::Expression {
-        Coerce::new(self)
-    }
-}
-
 #[cfg(feature = "postgres")]
-impl AsExpression<Timestamptz> for now {
-    type Expression = Coerce<now, Timestamptz>;
+impl crate::expression::AsExpression<Timestamptz> for now {
```
pksunkara, line 49: Can't we also add the `cfg` in front of one import line that imports these?

pksunkara (approved):
Looks great! 😄

PR #2522, opened 2020-09-29
`table!` with primitive type names

TaKO8Ki:
closes https://github.com/diesel-rs/diesel/issues/1204


weiznich:
I'm not sure if that's the correct way to fix the problem. We now introduce a new name collision with the new names. I think it's a better solution to just add those "names" to the reserved key-word list in [diesel-cli](https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel_cli/src/infer_schema_internals/inference.rs#L9) instead. So that we just generate valid names and annotate those columns with the correct `#[sql_name = "foo"]` attribute.

As an unrelated side node: We should remove the imports [here](https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L611-L621) and [here](https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L798-L804). Instead of doing the imports there we should use full paths everywhere where those items are use to prevent at least the name collisions from them.

TaKO8Ki:
@weiznich 
I've added `bool`, `columns` and so on to the reserved key-word list and removed as much imports as possible. Now in [this file](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/tests/print_schema/print_schema_column_renaming/mysql/expected.rs), we have tests for some of `RESERVED_NAMES`. Should I write tests for the other `RESERVED_NAMES`?

diesel/src/macros/mod.rs
```diff
@@ -613,12 +613,7 @@ macro_rules! __diesel_table_impl {
                 Table,
                 JoinTo,
             };
-            use $crate::associations::HasTable;
-            use $crate::insertable::Insertable;
-            use $crate::query_builder::*;
-            use $crate::query_builder::nodes::Identifier;
-            use $crate::query_source::{AppearsInFromClause, Once, Never};
-            use $crate::query_source::joins::{Join, JoinOn};
+            use $crate::query_builder::AsQuery;
```
weiznich, line 616: Any reason why those imports need to stay there?

diesel/src/macros/mod.rs
```diff
@@ -796,12 +791,7 @@ macro_rules! __diesel_table_impl {
             /// Contains all of the columns of this table
             pub mod columns {
                 use super::table;
-                use $crate::{Expression, SelectableExpression, AppearsOnTable, QuerySource};
-                use $crate::backend::Backend;
-                use $crate::query_builder::{QueryFragment, AstPass, SelectStatement};
-                use $crate::query_source::joins::{Join, JoinOn, Inner, LeftOuter};
-                use $crate::query_source::{AppearsInFromClause, Once, Never, Pick};
-                use $crate::result::QueryResult;
+                use $crate::QuerySource;
```
weiznich, line 794: Is there a specific reason why we need to import `QuerySource` here?

TaKO8Ki: That's why `from_clause` defined in `QuerySource` is used in the following lines. However, since we don't need to import `QuerySource` globally, I've imported it in functions that need it.

https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L38

https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L831

diesel/src/macros/mod.rs
TaKO8Ki: That's why `as_query` defined in `AsQuery` is used in the following lines. However, since we don't need to import `AsQuery` globally, I've imported it in functions that need it.

https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L713

https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/macros/mod.rs#L781

weiznich: That's a good fix. To just mention the other possible solution somewhere for the case you don't know this:

In rust it's possible to write `#crate::AsQuery::as_query(self).values()` instead, which allows to completely remove the import statement. (That's not required here, I'm just writing that as educational comment)

TaKO8Ki: That’s new for me! Thank you!

diesel/src/macros/mod.rs
```diff
@@ -613,12 +614,6 @@ macro_rules! __diesel_table_impl {
                 Table,
```
weiznich, line 614: Any reason why this imports remain here? I think they all should be removed.

TaKO8Ki: Since I missed them, I've removed them!

PR #2525, opened 2020-10-02
Postgres: BigDecimal with negative scale

### LINKED ISSUE #2385: Inserting BigDecimal with scale into Postgres is broken
### Versions

- **Rust:**
1.43

- **Diesel:**
current master, commit 968e7f79c

- **Database:**
Postgres

- **Operating System**
Linux

### Feature Flags

- **diesel:**
postgres, numeric

## Problem Description
First, thanks for continuing to maintain Diesel.

Inserting a BigDecimal as a numeric in Postgres is and probably has always been broken for any number with power of ten > 0 (or scale < 0 in BigDecimal terms). Looking at the current code https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/types/numeric.rs#L90, it's full of unchecked casts and arithmetic which are vulnerable to silent overflows.

Given this, I'd recommend to anyone inserting BigDecimals into Postgres that you avoid using the Diesel support and instead cast text values to numeric.

Insert a BigDecimal with a scale.

It works.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
DatabaseError(__Unknown, "invalid scale in external \"numeric\" value")', src/main.rs:31:5
```

No.

### Steps to reproduce
Run this https://github.com/leoyvens/diesel-numeric-bug.

## Checklist

alessandroasm:
Properly handle BigDecimals with negative scales.

Fixes #2385.

### DIFF
--- a/diesel/src/pg/types/numeric.rs
+++ b/diesel/src/pg/types/numeric.rs
@@ -89,7 +89,17 @@ mod bigdecimal {
         #[allow(clippy::assign_op_pattern, clippy::redundant_closure)]
         fn from(decimal: &'a BigDecimal) -> Self {
             let (mut integer, scale) = decimal.as_bigint_and_exponent();
-            let scale = scale as u16;
+
+            // Handling of negative scale
+            let scale = if scale < 0 {
+                for _ in 0..(-scale) {
+                    integer = integer * 10;
+                }
+                0
+            } else {
+                scale as u16
+            };
+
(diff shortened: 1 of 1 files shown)


weiznich (approved):

alessandroasm:
@weiznich Hi! If it's not much trouble could you tag this PR with "hacktoberfest-accepted"?
Thanks in advance!

PR #2529, opened 2020-10-07
feat: Add embed field attribute for AsChangeset

### LINKED ISSUE #2527: Annotation for setting item to null in AsChangeset
Realted issue: #26

I'd like to build a struct that implements `AsChangeset`, but it is not possible to set some values to `NULL` while keeping the struct ignoring some `None` fields. Using `Option<Option<T>>` or another type e.g. `enum { Value(T), Null, Unset }` will add extra overhead to the struct and may limit its usability.

Instead of adding an extra type, I'd suggest adding a field annotation `#[diesel(parse_with = "parse_func")]` similar to what `serde` use to `AsChangeset`, so that users themselves can determine whether to set null or to ignore the field. `parse_func` is a function of type `FnMut(T) -> Option<Option<impl AsChangset>>`. This will not break the current behavior and is relatively simple to implement IMHO.

e.g.

```rust
#[derive(AsChangeset)]
struct UpdateUser {
  #[diesel(parse_with = "update_user_hobby_set")]
  hobby: Option<Option<String>>,
  #[diesel(parse_with = "Country::update_set")]
  country: Country,
}

fn<T> update_user_hobby_set(hobby: Option<Option<String>>) -> Option<Option<T>>
where:
  T: AsChangeset
{
  // Returns None to ignore the field, Some(None) to set null, Some(Some(T)) to set value to T.
  // Making the return value an enum is also fine for me.
  hobby
}

struct Country(JsonValue)

impl Country {
  fn update_set(&self) -> Option<Option<i32>> {
    match self {
      JsonValue::Null => Some(None),
      JsonValue::Undefined => None,
      JsonValue::Integer(v) => *v as i32,
      _ => None,
    }
  }
}
```

Let me know if you have any other workarounds other than manually implementing `AsChangeset`. I can work on this when I have some spare time.

heyrict:
Closes #2527

### DIFF
--- a/diesel_derives/src/as_changeset.rs
+++ b/diesel_derives/src/as_changeset.rs
@@ -92,13 +92,18 @@ fn field_changeset_ty(
     treat_none_as_null: bool,
     lifetime: Option<proc_macro2::TokenStream>,
 ) -> syn::Type {
-    let column_name = field.column_name();
-    if !treat_none_as_null && is_option_ty(&field.ty) {
-        let field_ty = inner_of_option_ty(&field.ty);
-        parse_quote!(std::option::Option<diesel::dsl::Eq<#table_name::#column_name, #lifetime #field_ty>>)
-    } else {
+    if field.has_flag("embed") {
         let field_ty = &field.ty;
-        parse_quote!(diesel::dsl::Eq<#table_name::#column_name, #lifetime #field_ty>)
+        parse_quote!(#lifetime #field_ty)
+    } else {
+        let column_name = field.column_name();
+        if !treat_none_as_null && is_option_ty(&field.ty) {
+            let field_ty = inner_of_option_ty(&field.ty);
+            parse_quote!(std::option::Option<diesel::dsl::Eq<#table_name::#column_name, #lifetime #field_ty>>)
+        } else {
+            let field_ty = &field.ty;
+            parse_quote!(diesel::dsl::Eq<#table_name::#column_name, #lifetime #field_ty>)
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


weiznich (changes requested):
Thanks for working on this. This requires at least the following changes:

- [ ] Add the corresponding documentation on top of the `derive_as_changeset` method
- [ ] Add at least one test case showing that this works

EDIT: I try to find some time to have a look at the CI errors. It seems like a regression in rustc 😞 
EDIT2: Opened #2531 to fix the CI issues.

diesel_tests/tests/update.rs
```diff
@@ -1,5 +1,9 @@
 use crate::schema::*;
 use diesel::*;
+use diesel::{
+    expression::{bound::Bound, grouped::Grouped, operators},
+    sql_types::{Nullable, Text},
+};
```
weiznich, line 6: I would prefer to keep that in line with all other import in diesel. That means: No nested import, but better use the following:
```suggestion
use diesel::expression::{bound::Bound, grouped::Grouped, operators};
use diesel::sql_types::{Nullable, Text};
```

diesel_tests/tests/update.rs
```diff
@@ -301,3 +305,93 @@ fn upsert_with_sql_literal_for_target() {
...
+        hair_color: Some("Black".to_string()),
+    };
+
+    let updated_user: User = diesel::update(users::table)
+        .filter(users::id.eq(1))
+        .set(UpdateUser {
+            name: None,
+            hair_color: HairColor::Black,
+        })
+        .get_result(&connection)
```
weiznich, line 371: Better use a separate query here to load the results, as both `Sqlite` and `Mysql` do not support queries containing `RETURNING` clauses. (Or use the `UpdateAndFetchResult` trait impls.)

weiznich (changes requested):
Thanks for the update. Left 1 minor style comments + CI should pass.

heyrict:
Well, I'm a little confused about the `syntax error at or near "="` error, which seems to be an error originated from sql generation. I may take some more time looking into this.

weiznich:
@heyrict I would probably try to dump the generate SQL via `diesel::debug_query` and than look there what's wrong.

heyrict:
Thanks for the hint!

I didn't touch the implementation of `Eq<Left, Right>` though. I'm not sure why it is generating an invalid sql. I'll try to fix that later.

```sql
UPDATE "users" SET ("users"."hair_color" = $1) WHERE ("users"."id" = $2)
-- instead of
UPDATE "users" SET "hair_color" = $1  WHERE ("users"."id" = $2)
```

diesel_derives/src/as_changeset.rs
```diff
@@ -109,14 +114,19 @@ fn field_changeset_expr(
     lifetime: Option<proc_macro2::TokenStream>,
 ) -> syn::Expr {
     let field_access = field.name.access();
-    let column_name = field.column_name();
-    if !treat_none_as_null && is_option_ty(&field.ty) {
-        if lifetime.is_some() {
-            parse_quote!(self#field_access.as_ref().map(|x| #table_name::#column_name.eq(x)))
+    if field.has_flag("embed") {
+        parse_quote!(#lifetime self#field_access)
```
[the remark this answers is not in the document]
heyrict: Thanks! I'll try that this weekend.

weiznich:
That's implemented now since a while, so this PR can be closed.

PR #2532, opened 2020-10-07
Fix #1638

### LINKED ISSUE #1638: Introduce a `without-deprecated` feature
We have a lot of deprecations which do not give a warning due to various bugs in Rust. We've been recommending that people try compiling their apps with default features turned off if they're concerned about relying on deprecated code. However, this isn't always enough. If any dependency in the entire graph depends on Diesel without `default-features = false`, you cannot get rid of deprecated code.

We should introduce an explicit `without-deprecated` feature meant specifically for "just make sure I'm not depending on deprecated code" testing. Any place where we have `feature = "with-deprecated"` will need to change to `all(feature = "with-deprecated", not(feature = "without-deprecated"))`.

In an ideal world we shouldn't need this, and nobody should have to wonder if they're relying on deprecated code. However, the number of deprecations we have that do not warn are at a volume where I think this is warranted.

For reference, a few of the bugs in question:

https://github.com/rust-lang/rust/issues/49912
https://github.com/rust-lang/rust/issues/47237
https://github.com/rust-lang/rust/issues/47236
https://github.com/rust-lang/rust/issues/47238

weiznich:
Add a `without-deprecated` feature flag that really disables all
deprecated items


TaKO8Ki:
Don't you need to replace [this line](https://github.com/diesel-rs/diesel/blob/ba77345c3164edf588aaf32206f07c783e1ac3a9/diesel/src/pg/mod.rs#L26) with `#[cfg(all(feature = "with-deprecated", not(feature = "without-deprecated")))]`?

weiznich:
> Don't you need to replace this line with #[cfg(all(feature = "with-deprecated", not(feature = "without-deprecated")))]?

I've missed that one. Thanks for catching that :+1: 
I've pushed a new version fixing that + adding a changelog entry.

TaKO8Ki (approved):

PR #2538, opened 2020-10-12
Add docs to sql_function macro for functions without arguments

film42:
I was looking to port ordering by random from v1.4 to master branch. I left a comment for others and it was recommended I add the `random` example to the docs: https://github.com/diesel-rs/diesel/pull/1987#issuecomment-703584356

Please let me know if you would like more detail, and if so, where you would like any! Thanks for the great project! :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -213,6 +213,10 @@ Key points:
   NonAggregate` no longer implies `(OtherType, T): NonAggregate`.
   - With `feature = "unstable"`, `(T, OtherType): NonAggregate` is still implied.
 
+- In diesel v1.4 sql functions without arguments used the `no_arg_sql_function!` macro,
+  which has since been deprecated. The new `sql_function!` macro supports functions without
+  arguments.
+
 [2-0-migration]: FIXME write a migration guide
 
 ## [1.4.5] - 2020-06-09
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -863,6 +863,25 @@ pub fn derive_valid_grouping(input: TokenStream) -> TokenStream {
 /// # }
 /// ```
 ///
+/// # SQL Functions without Arguments
+///
+/// A common example is ordering a query using the `RANDOM()` sql function,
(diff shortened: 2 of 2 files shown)


diesel_derives/src/lib.rs
```diff
@@ -863,6 +863,29 @@ pub fn derive_valid_grouping(input: TokenStream) -> TokenStream {
 /// # }
 /// ```
 ///
+/// # SQL Functions without Arguments
+///
+/// In diesel v1.4 sql functions without arguments used the `no_arg_sql_function!` macro,
+/// which has since been deprecated. The new `sql_function!` macro supports functions without
+/// arguments.
```
weiznich, line 870: I think the documentation should not mention old diesel versions, as this can (and probably will) stick around much longer than this is relevant for anyone anymore.

film42: Will do! Where do you recommend filling in the blanks for the change log docs around migrating? https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md#removed

weiznich: I think adding a section to [this part](https://github.com/diesel-rs/diesel/blob/1d17019315f868b3ec6bbdef2c1a98972e0f5de7/CHANGELOG.md#upgrade-notes) of the changelog would be great.

CHANGELOG.md
```diff
@@ -213,6 +213,10 @@ Key points:
   NonAggregate` no longer implies `(OtherType, T): NonAggregate`.
   - With `feature = "unstable"`, `(T, OtherType): NonAggregate` is still implied.
 
+- In diesel v1.4 sql functions without arguments used the `no_arg_sql_function!` macro,
```
film42, line 216: I moved what I had written before to this set of key points. This will probably need to be expanded but I figured it's at least a good note.

weiznich:
Thanks for improving the documentation :+1:

film42:
@weiznich if it's not too much trouble, would you be willing to tag this PR with `hacktoberfest-accepted`? If not, no big deal! Thanks!

PR #2542, opened 2020-10-13
Replace `set-env` and `add-path` with secure ways on GHA

JohnTitor:
Replace `set-env` and `add-path` with the recommended ways since they're deprecated due to moderate security vulnerability (https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).


weiznich:
Thanks :+1:

PR #2547, opened 2020-10-23
Fix brew fails on macos-latest

TaKO8Ki:
Since all `brew update` fails in CI, I've fixed this.

workaround: https://github.com/actions/virtual-environments/issues/1811#issuecomment-708480190gi


weiznich:
Thanks :+1:

Issue #2344, opened 2020-03-30
[Feature Request] #[serialize_as] for Insertable, like #[deserialize_as] for Queryable


pheki:
#1852 added `#[deserialize_as]` for `Queryable` and `QueryableByName`, which I found to be very useful. It is stated in that PR that it was intended to follow that PR with an equivalent `#[serialize_as]` attribute for `Insertable` and `AsChangeset`. I was looking for it and I'd find it really helpful. Is that still on the roadmap?

Thanks!

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

weiznich:
It is still on the roadmap in terms of we would accept a PR implementing this feature. As far as I know none of the regular contributors is currently working on this.

pheki:
That's great! unfortunately I don't have the bandwith to understand diesel internals + implement it right now

thomaseizinger:
If I were to give this a shot, would I target current master or the 1.4 branch?

weiznich:
@thomaseizinger New features always targeting the master branch.

PR #2553, opened 2020-11-02
Support `#[diesel(serialize_as)]` on fields of `Insertable` structs

thomaseizinger:
Similar to how `#[diesel(deserialize_as)]` can be used for `Queryable`
structs to customize which type is being constructed, we add support
for doing the same thing for `Insertable` structs.

The conversion happens using the std trait `Into` which takes ownership
during conversion. As such, as soon as any field within an `Insertable`
struct has a `#[diesel(serialize_as)]` attribute, we no longer generate
an implementation of `Insertable` for borrowed versions of this struct.

Fields annotated with `#[diesel(serialize_as)]` are also incompatible
with `#[diesel(embed)]`. Using both on the same field will result in
a compile-error.

Resolves https://github.com/diesel-rs/diesel/issues/2344.

### DIFF
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -26,33 +26,46 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
     impl_generics.params.push(parse_quote!('insert));
     let (impl_generics, ..) = impl_generics.split_for_impl();
 
-    let (direct_field_ty, direct_field_assign): (Vec<_>, Vec<_>) = model
-        .fields()
-        .iter()
-        .map(|f| {
-            (
-                (field_ty(f, table_name, None)),
-                (field_expr(f, table_name, None)),
-            )
-        })
-        .unzip();
+    let mut generate_borrowed_insert = true;
 
-    let (ref_field_ty, ref_field_assign): (Vec<_>, Vec<_>) = model
-        .fields()
-        .iter()
-        .map(|f| {
-            (
-                (field_ty(f, table_name, Some(quote!(&'insert)))),
-                (field_expr(f, table_name, Some(quote!(&)))),
-            )
-        })
-        .unzip();
+    let mut direct_field_ty = Vec::with_capacity(model.fields().len());
+    let mut direct_field_assign = Vec::with_capacity(model.fields().len());
+    let mut ref_field_ty = Vec::with_capacity(model.fields().len());
+    let mut ref_field_assign = Vec::with_capacity(model.fields().len());
 
-    Ok(wrap_in_dummy_mod(quote! {
-        use diesel::insertable::Insertable;
-        use diesel::query_builder::UndecoratedInsertRecord;
-        use diesel::prelude::*;
+    for field in model.fields() {
+        let serialize_as = field.ty_for_serialize()?;
+        let embed = field.has_flag("embed");
 
+        match (serialize_as, embed) {
+            (None, true) => {
+                direct_field_ty.push(field_ty_embed(field, None));
+                direct_field_assign.push(field_expr_embed(field, None));
+                ref_field_ty.push(field_ty_embed(field, Some(quote!(&'insert))));
+                ref_field_assign.push(field_expr_embed(field, Some(quote!(&))));
+            }
+            (None, false) => {
+                direct_field_ty.push(field_ty(field, table_name, None));
+                direct_field_assign.push(field_expr(field, table_name, None));
+                ref_field_ty.push(field_ty(field, table_name, Some(quote!(&'insert))));
+                ref_field_assign.push(field_expr(field, table_name, Some(quote!(&))));
+            }
+            (Some(ty), false) => {
+                direct_field_ty.push(field_ty_serialize_as(field, table_name, &ty));
+                direct_field_assign.push(field_expr_serialize_as(field, table_name, &ty));
+
+                generate_borrowed_insert = false; // as soon as we hit one field with #[diesel(serialize_as)] there is no point in generating the impl of Insertable for borrowed structs
+            }
+            (Some(_), true) => {
+                return Err(field
+                    .flags
+                    .span()
+                    .error("`#[diesel(embed)]` cannot be combined with `#[diesel(serialize_as)]`"))
+            }
+        }
+    }
+
+    let insert_owned = quote! {
         impl #impl_generics Insertable<#table_name::table> for #struct_name #ty_generics
             #where_clause
         {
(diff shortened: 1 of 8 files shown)


diesel_derives/src/insertable.rs
```diff
@@ -86,41 +70,85 @@ fn field_ty(
...
+                panic!("unsupported combination")
+            } else {
+                let column_name = field.column_name();
+                let column: syn::Expr = parse_quote!(#table_name::#column_name);
+
+                if is_option_ty(&ty) {
+                    if lifetime.is_some() {
+                        unimplemented!()
+                    } else {
+                        parse_quote!(self#field_access.map(|x| #column.eq(#ty::from(x))))
```
weiznich, line 126: We should not depend on the standard prelude here, but instead use UFCS otherwise this will break as soon as someone does something fancy…
```suggestion
                        parse_quote!(self#field_access.map(|x| #column.eq(<#ty as std::convert::From<#field_type>>::from(x))))
```
I think we should call `From<Type>` for owned types, and `From<&Type>` for types that are stored as reference in the original insertable struct.

diesel_derives/src/lib.rs
```diff
@@ -237,6 +237,12 @@ pub fn derive_identifiable(input: TokenStream) -> TokenStream {
 /// example, an enum that maps to a label and a value column). Add
 /// `#[diesel(embed)]` to any such fields.
 ///
+/// To provide custom serialization behavior for a field, you can use
+/// `#[diesel(serialize_as = "SomeType")]`. If this attribute is present, Diesel
+/// will call `.into` on the corresponding field and serialize the instance of `SomeType`,
```
weiznich, line 242: We are calling `from` not `into`. Either change the implementation or the documentation. (I would say change the implementation as calling `into` is generally preferred as it is more generically)

diesel_tests/tests/serialize_as.rs
```diff
@@ -0,0 +1,65 @@
+use crate::schema::*;
```
weiznich, line 1: This example should be somewhere in the documentation

diesel_tests/tests/serialize_as.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn from(s: String) -> Self {
+        UppercaseString(s.to_uppercase())
+    }
+}
+
+impl From<&String> for UppercaseString {
+    fn from(s: &String) -> Self {
+        UppercaseString(s.to_uppercase())
+    }
+}
```
weiznich, line 21: Why do we need both impls?

weiznich (commented):
I left some initial notes.

diesel_derives/src/insertable.rs
```diff
@@ -26,27 +26,11 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
-                (field_ty(f, table_name, Some(quote!(&'insert)))),
-                (field_expr(f, table_name, Some(quote!(&)))),
-            )
-        })
-        .unzip();
+    let direct_field_ty = model.fields().iter().map(|f| field_ty(f, table_name, None)).collect::<Result<Vec<_>, _>>()?;
+    let direct_field_assign = model.fields().iter().map(|f| field_expr(f, table_name, None)).collect::<Result<Vec<_>, _>>()?;
+
+    let ref_field_ty = model.fields().iter().map(|f| field_ty(f, table_name, Some(quote!(&'insert)))).collect::<Result<Vec<_>, _>>()?;
+    let ref_field_assign = model.fields().iter().map(|f| field_expr(f, table_name, Some(quote!(&)))).collect::<Result<Vec<_>, _>>()?;
```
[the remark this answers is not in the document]
thomaseizinger: To my knowledge, `unzip` doesn't support the `Vec<Result>` -> `Result<Vec>` conversion. I didn't actually try though to be honest because I wanted to leave optimizations like this up to the end when there is a general ACK on the concept.

diesel_derives/src/insertable.rs
```diff
@@ -86,41 +70,85 @@ fn field_ty(
...
-        if is_option_ty(&field.ty) {
-            if lifetime.is_some() {
-                parse_quote!(self#field_access.as_ref().map(|x| #column.eq(x)))
+) -> Result<syn::Expr, Diagnostic> {
+    let expr = match field.ty_for_serialize()? {
+        Some(ty) => {
+            let field_access = field.name.access();
+
+            if field.has_flag("embed") {
+                panic!("unsupported combination")
```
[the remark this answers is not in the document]
thomaseizinger: Yep, that is planned. I first wanted to get some feedback on whether or not the assumption that `embed` and `serialize_as` are incompatible is actually correct :)

diesel_derives/src/insertable.rs
```diff
@@ -86,41 +70,85 @@ fn field_ty(
...
+                        parse_quote!(self#field_access.map(|x| #column.eq(#ty::from(x))))
+                    }
+                } else {
+                    parse_quote!(std::option::Option::Some(#column.eq(#ty::from(#lifetime self#field_access))))
+                }
+            }
+        },
+        None => {
+            let field_access = field.name.access();
+            if field.has_flag("embed") {
```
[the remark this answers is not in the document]
thomaseizinger: Agreed, will optimize once we settle on required functionality :)

diesel_derives/src/lib.rs
thomaseizinger: Yes, I started with the documentation but then ran into problems with `.into` because rustc couldn't resolve the types for cases where `serialize_as` is **not** used. Will update the documentation in the end!

diesel_tests/tests/serialize_as.rs
thomaseizinger: I'd love to not have the 2nd one because I don't think it makes sense but it didn't compile without it :(

thomaseizinger:
> I left some initial notes.

Thank you @weiznich !

How do we want to deal with references in general?

Currently, I am emitting this:

```rust
impl<'insert> Insertable<users::table> for InsertableUser {
    type Values = <(std::option::Option<diesel::dsl::Eq<users::name, UppercaseString>>, ) as Insertable<users::table>>::Values;
    fn values(self) -> Self::Values {
        (std::option::Option::Some(
            users::name.eq(UppercaseString::from(self.name)),
        ), )
            .values()
    }
}

impl<'insert> Insertable<users::table> for &'insert InsertableUser {
    type Values = <(std::option::Option<diesel::dsl::Eq<users::name, UppercaseString>>, ) as Insertable<users::table>>::Values;
    fn values(self) -> Self::Values {
        (std::option::Option::Some(
            users::name.eq(UppercaseString::from(&self.name)),
        ), )
            .values()
    }
}
```

Passing `InsertableUser` as a reference is pretty pointless now because we always allocate in `from` anyway. It also forces users to have a 2nd `From` impl for a reference.

Should I just omit the `impl<'insert>` block in case the struct is using `serialize_as`?

diesel_derives/src/insertable.rs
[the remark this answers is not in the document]
weiznich: Good point about `unzip()` doesn't support that. So maybe just go back to a normal loop here?

diesel_derives/src/lib.rs
weiznich: You probably need to use UFCS to specify all generic parameters there.

weiznich:
> Should I just omit the impl<'insert> block in case the struct is using serialize_as?

I think that's an reasonable thing to do as long as it is documented probably.

thomaseizinger:
> > Should I just omit the impl<'insert> block in case the struct is using serialize_as?
> 
> I think that's an reasonable thing to do as long as it is documented probably.

Cool thank you! I will roll with that then!

Is is also reasonable to say this doesn't work with `#[diesel(embed)]`?

thomaseizinger:
@weiznich this is ready for another review!

I think I've covered most of the comments. Tests are hopefully passing now! The failing compiletests also fail on master currently.

diesel_derives/src/insertable.rs
```diff
@@ -26,33 +26,45 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
+                ref_field_assign.push(field_expr(field, table_name, Some(quote!(&))));
+            }
+            (Some(ty), false) => {
+                direct_field_ty.push(field_ty_serialize_as(field, table_name, &ty));
+                direct_field_assign.push(field_expr_serialize_as(field, table_name, &ty));
+
+                generate_borrowed_insert = false; // as soon as we hit one field with #[diesel(serialize_as)] there is no point in generating the impl of Insertable for borrowed structs
+            }
+            (Some(_), true) => {
+                return Err(field
```
weiznich, line 60: can we use here the span of the attribute only? I think that should give us a nicer error message.

diesel_derives/src/lib.rs
```diff
@@ -237,6 +237,19 @@ pub fn derive_identifiable(input: TokenStream) -> TokenStream {
...
+/// rather than the actual field on your struct. This can be used to add custom behavior for a
+/// single field, or use types that are otherwise unsupported by Diesel.
+/// Using `#[diesel(serialize_as)]` is **incompatible** with `#[diesel(embed)]`.
+/// Normally, Diesel produces two implementations of the `Insertable` trait for your
+/// struct using this derive: one for an owned version and one for a borrowed version.
+/// Using `#[diesel(serialize_as)]` implies a conversion using `.into` an as such does
+/// not work on references.
```
weiznich, line 249:
```suggestion
/// Using `#[diesel(serialize_as)]` implies a conversion using `.into` which consumes the underlying value
```

weiznich (approved):
This PR looks good now modulo that two small things. Good work :+1:

> The failing compiletests also fail on master currently.

That should be fixed now. I somehow manged to merge two PR's that changed similar things and lead combined to a failing compile test 🙈

thomaseizinger:
> This PR looks good now modulo that two small things. Good work +1

Thank you!

> > The failing compiletests also fail on master currently.
> 
> That should be fixed now. I somehow manged to merge two PR's that changed similar things and lead combined to a failing compile test see_no_evil

That sounds like a job for [bors](https://bors.tech/) :)
From the documentation:

> Bors is a GitHub bot that prevents merge skew / semantic merge conflicts, so when a developer checks out the main branch, they can expect all of the tests to pass out-of-the-box. 

We use it extensively at work and it works really well!

weiznich:
I've just noticed that this probably should have a Changelog entry [here](https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md#added). Otherwise it's ready for merging now. (Sorry for finding just another thing that needs to be changed)

> That sounds like a job for bors :)

I'm well aware of bors, but I hesitated to setup and maintain that for diesel yet.

thomaseizinger:
> I've just noticed that this probably should have a Changelog entry [here](https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md?rgh-link-date=2020-11-05T14%3A59%3A34Z#added). Otherwise it's ready for merging now. (Sorry for finding just another thing that needs to be changed)

Done!

PR #2554, opened 2020-11-04
Fix a test-failure caused by semi-conflicting PR's

weiznich:


pksunkara (approved):
Just wanted to let you know that bors prevents these merge skews.

weiznich:
@pksunkara I know, but I think given the current amount of PR's and frequency of such things happening in the past it is not worth to setup and maintain bors just for this.

PR #2557, opened 2020-11-07
Moved error ui to trybuild tests

pksunkara:
Move ui compilation tests with errors to use trybuild

compiletest-rs has a lot of issues as described [here](https://github.com/dtolnay/trybuild/issues/13#issuecomment-507077406)

clap, structopt and other popular dervies use this.


weiznich (commented):
I'm generally in favor of this change, but I think the following points should be clarified or even fixed before merging:

- [x] The `ui` test from the `compiletests_rs` test-suite [here](https://github.com/diesel-rs/diesel/blob/master/diesel_compile_tests/tests/compile_tests.rs#L29-L32) should be removed, as it is not used anymore
- [ ] Having to maintain `trybuild` + `compiletests_rs` compile-fail tests seems to be just one dependency more to maintain. Therefore there should at least be a clear path what needs to happen for the test cases covered by the `compile-fail` test. Ideally this PR would port them to the corresponding `trybuild` variant as well.

pksunkara:
> The ui test from the compiletests_rs test-suite here should be removed, as it is not used anymore

They are still being used [here](https://github.com/pksunkara/diesel/tree/trybuild/diesel_compile_tests/tests/ui) to test the warning output because trybuild doesn't support warnings yet. https://github.com/dtolnay/trybuild/issues/64

> Therefore there should at least be a clear path what needs to happen for the test cases covered by the compile-fail test. Ideally this PR would port them to the corresponding trybuild variant as well.

The clear path is that we will be porting them to trybuild in the future. With `TRYBUILD=overwrite` it should be really easy to do that but I didn't want to do it in this PR, because it would make the change really huge and I wasn't sure if the error output I was seeing was correct or not.

weiznich:
> The clear path is that we will be porting them to trybuild in the future. With TRYBUILD=overwrite it should be really easy to do that but I didn't want to do it in this PR, because it would make the change really huge and I wasn't sure if the error output I was seeing was correct or not.

I'm fine with having one big PR for this in exchange of having a stable compile-fail test setup.

> They are still being used here to test the warning output because trybuild doesn't support warnings yet. dtolnay/trybuild#64

It's definitively a blocker for me having to maintain both test suites in parallel. As it stands this PR would only increase maintenance burden as we now have to maintain all test as before + the newly added `trybuild` tests. That's not something I'm willing to accept. From a short look at the tests those warnings seem to be worth of keeping them tested. Therefore I do not really see a way forward here without having a `trybuild` version which support warnings.

pksunkara:
This is almost ready. But some of the error messages in rust seem to use the full path of the file, as you can see here, https://github.com/diesel-rs/diesel/pull/2557/files#diff-698425b64863efdf87cb7f45db5b7f79f73242c05768ba9d6b68aa11b96dafd4R7

weiznich:
@pksunkara That sounds like a bug in trybuild, might be worth to report it there?

pksunkara:
Fixed the final issue. This PR is now ready. 

The issue was occurring because `diesel_compile_tests` isn't in the diesel workspace (probably because of `compile_tests`) which I have now changed.

.gitignore
```diff
@@ -1,4 +1,5 @@
 target
 Cargo.lock
 !diesel_cli/Cargo.lock
-.env
\ No newline at end of file
+.env
+.vscode
```
weiznich, line 5: This should live in your own global `.gitignore` config, not in the diesel specific one.

diesel_compile_tests/Cargo.toml
```diff
@@ -3,8 +3,6 @@ name = "diesel_compile_tests"
 version = "0.1.0"
 authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
 
-[workspace]
-
 [dependencies]
 diesel = { version = "2.0.0", default-features = false, features = ["extras", "sqlite", "postgres", "mysql", "unstable"], path = "../diesel" }
-compiletest_rs = "=0.4"
+trybuild = "1.0.35"
```
weiznich, line 8: Just to verify that: We do handle the warnings correctly now?

diesel_compile_tests/tests/fail/as_changeset_deprecated_column_name.rs
```diff
@@ -16,4 +16,6 @@ struct UserForm {
     name: String,
 }
 
-fn main() {}
+fn main() {
+    compile_error!("warnings");
```
weiznich, line 20: Could we add a comment here that this is required to actually check for the warnings? Maybe even link the upstream `trybuild` issue?

diesel_compile_tests/tests/fail/belongs_to_second_parent.rs
```diff
@@ -32,4 +32,6 @@ struct Foo {
     bar_id: i32,
 }
 
-fn main() {}
+fn main() {
+    compile_error!("warnings");
```
weiznich, line 36: Add a comment here why we use `compile_error!`.

diesel_compile_tests/tests/fail/select_requires_valid_grouping.stderr
```diff
@@ -0,0 +1,88 @@
+warning: unused `#[macro_use]` import
```
weiznich, line 1: As this warning is not relevant, could we remove the corresponding `#[macro_use]`?

weiznich (commented):
This looks nearly done. Great work :+1:. I've left a few small comments on minor things I've noted, but they all should be easy to fix. Additionally we need to have a look at the test error in [this CI run](https://github.com/diesel-rs/diesel/pull/2557/checks?check_run_id=1707600032)

diesel_compile_tests/Cargo.toml
pksunkara: Yes.

pksunkara:
The CI run is failing because we added `diesel_compile_tests` as member of workspace and it uses `nightly` `feature`. But our `cargo clippy` runs on `stable` channel. We should probably change the clippy to run on `1.40.0` nightly.

weiznich:
> The CI run is failing because we added diesel_compile_tests as member of workspace and it uses nightly feature. But our cargo clippy runs on stable channel. We should probably change the clippy to run on 1.40.0 nightly.

I'm quite sure that's not a good idea. `nightly` is a moving target, where things break randomly (see how much regressions I've already reported against rustc, because diesel is broken due to a change), where new lints get introduce (so the CI breaks for contributors without reason) and where `clippy` gets new lint every day or so (again which may break the CI for contributors without anything they can do). I would rather like to exclude `diesel_compile_tests` from all CI runs outside of the `compile_test` run.

pksunkara:
I think you misunderstood. What I meant is using 1.40.0 nightly version. Whenever stable is released, there is a nightly release too with the same code. 

https://static.rust-lang.org/dist/2019-12-19/

As you can see there's both stable and nightly releases on that day. For the clippy run, we can set the toolchain to `nightly-2019-12-19`, that way there's no difference. (I actually confirmed this locally).

pksunkara:
But I actually ended up doing what you suggested (Excluding diesel_compile_tests from some of the tests)

.github/workflows/ci.yml
```diff
@@ -356,7 +356,7 @@ jobs:
       - name: Run clippy
         uses: actions-rs/cargo@v1
         with:
-          command: clippy
+          command: clippy --all --exclude=diesel_compile_tests
```
weiznich, line 359:
```suggestion
          command: clippy 
          args: --all --exclude=diesel_compile_tests
```

Issue #2523, opened 2020-10-01
No way to express inconsistency when deserializing for `Queryable` data since #2182


Ten0:
### Versions

- **Rust:** Latest stable (1.46.0)
- **Diesel:** Latest master (d7ce43d2130bca434cd1acc2943a0c944d1deba4)
- **Database:** PostgreSQL
- **Operating System:** Linux

## Problem Description
Since #2480 was fixed, I've just been giving another try at updating our codebase to the latest master (#2182).
It turns out I still couldn't because I can't define custom `FromSqlRow` impls, due to conflicting impls with *downstream crates may implement trait*, where Rust claims that Diesel could go ahead and implement `Queryable` or `QueryableByName` themselves on my type, which would create conflicts because of the following impls:
https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/deserialize.rs#L359-L367
https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/deserialize.rs#L388-L399

Btw, same error when trying to implement `StaticallySizedRow` because of
https://github.com/diesel-rs/diesel/blob/d7ce43d2130bca434cd1acc2943a0c944d1deba4/diesel/src/deserialize.rs#L435-L442

### What are you trying to accomplish?

I want to customize a `FromSqlRowImpl` impl because I want to be allowed to fail on deserialization when seeing a data inconsistency, as well as easily reorganize my data in different sub-structures at the same time, while performing validation (which `Queryable` does not allow me to do).

### What is the expected output?
*compiles*

### What is the actual output?
```
error[E0119]: conflicting implementations of trait `diesel::deserialize::FromSqlRow<diesel::sql_types::Untyped, _>` for type `MyType`:
  --> lib.rs:30:1
   |
30 | / impl<ST, DB> diesel::deserialize::FromSqlRow<ST, DB> for MyType
31 | | where
32 | |     DB: diesel::backend::Backend,
33 | |     (Option<i32>, (Option<i32>, Option<i32>), Option<f64>): diesel::deserialize::FromSqlRow<ST, DB>,
...  |
54 | |     }
55 | | }
   | |_^
   |
   = note: conflicting implementation in crate `diesel`:
           - impl<DB, T> diesel::deserialize::FromSqlRow<diesel::sql_types::Untyped, DB> for T
[...]

PR #2559, opened 2020-11-09
Faillible Queryable - Non-generic Queryable version experiment

Ten0:
One approach to fixing #2523


weiznich:
Just to drop a note here: I'm currently busy with other work related stuff, so I will likely not be able to look at this for the next week or so.

Ten0:
So the test fails there seems to highlight a compatibility break with type inference where the compiler cannot infer "oh there's a single type I know for deserializing BigInt, that's `i64`, let me just use that" anymore, though I'm not sure why it has this impact.

If going for this approach, `readme`s and diesel_cli would still have to be updated.

weiznich:
Another important change here is that we cannot anymore write different impls depending on the type returned by some query or the backend. Doing that is basically required to port `diesel_cli`. As consequence I would like to see how that could be solved before looking any further into this.

Ten0:
> Another important change here is that we cannot anymore write different impls depending on the type returned by some query or the backend. Doing that is basically required to port `diesel_cli`. As consequence I would like to see how that could be solved before looking any further into this.

From https://github.com/diesel-rs/diesel/issues/2523#issuecomment-703781510:
> it would probably always be implemented without an additional layer of genericity, so it could probably use `FromSqlRow` instead of `Queryable`, as the _conflicting impl_ error wouldn't trigger.

As it turns out, we're actually using a bit of genericity here:
https://github.com/diesel-rs/diesel/blob/a312528c7cce6b25ae6f1f5987f784b635b18ed6/diesel_cli/src/infer_schema_internals/data_structures.rs#L78-L82

but for some reason it still works. (Maybe thanks to the local `UsesInformationSchema` trait.)

Ten0:
One more issue: rust error messages suggest less correct approaches to fixing `FromSqlRow<ST, DB> is not implemented for T`:
https://github.com/diesel-rs/diesel/pull/2559/checks?check_run_id=1617158750#step:6:258

weiznich:
Closed as I would like to go with #2599

PR #2563, opened 2020-11-11
Using chrono without the time 0.1 dependency

Razican:
Chrono has recently removed the time 0.1 dependency ([changelog](https://github.com/chronotope/chrono/blob/main/CHANGELOG.md#0416)). This makes it easy for diesel to remove the dependency too, by not using chrono's default, deprecated `oldtime` feature.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 byteorder = "1.0"
-chrono = { version = "0.4.1", optional = true }
+chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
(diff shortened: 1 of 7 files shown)


pksunkara (approved):

TaKO8Ki (approved):

sno2 (approved):

diesel/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 byteorder = "1.0"
-chrono = { version = "0.4.1", optional = true }
+chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
```
weiznich, line 17: As this bumps our minimal supported chrono version this requires a changelog entry.

weiznich (changes requested):
I would like to see an changelog entry for this as this is a potential breaking change.

diesel/Cargo.toml
Razican: Done, let me know if this is the correct place :)

PR #2565, opened 2020-11-16
pg: Add support for INET/CIDR-specific operators and functions

Tuetuopay:
This PR adds support for postgresql-specific INET and CIDR types. In this PR, support is added in the query builder for:
- `<<`, `>>=`, etc operators
- `host`, `masklen`, ... functions

PR #2566, opened 2020-11-17
pg: Add support for INET/CIDR-specific operators and functions

Tuetuopay:
This PR adds support for postgresql-specific INET and CIDR types. In this PR, support is added in the query builder for:
- `<<`, `>>=`, etc operators
- `host`, `masklen`, ... functions

This is #2565 but for Diesel master.

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -4,7 +4,7 @@ use super::operators::*;
 use crate::dsl;
 use crate::expression::grouped::Grouped;
 use crate::expression::{AsExpression, Expression, IntoSql, TypedExpressionType};
-use crate::sql_types::{Array, Bigint, Cidr, Inet, Nullable, Range, SqlType, Text};
+use crate::sql_types::{Array, Cidr, Inet, Nullable, Range, SqlType, Text};
 
 /// PostgreSQL specific methods which are present on all expressions.
 pub trait PgExpressionMethods: Expression + Sized {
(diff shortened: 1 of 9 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -683,3 +683,592 @@ where
     T::SqlType: RangeOrNullableRange,
 {
 }
+
+/// PostgreSQL specific methods present between CIDR/INET expressions
+pub trait PgNetExpressionMethods: Expression + Sized {
+    /**
```
weiznich, line 689: Any reason why this does not use the standard rust-doc comments (`///`)? 
(The same applies for all methods below)

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -683,3 +683,592 @@ where
...
+     * Creates a PostgreSQL `>>` expression.
+     *
+     * This operator returns wether a subnet strictly contains another subnet or address.
+     *
+     * # Example
+     *
+     * ```rust
+     * # #[macro_use] extern crate diesel;
+     * # extern crate ipnetwork;
```
weiznich, line 698: We don't need any of this anymore here. Master already targets rust-2018 where this just works.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -683,3 +683,592 @@ where
...
+
+impl<T> PgNetExpressionMethods for T
+where
+    T: Expression,
+    T::SqlType: InetOrCidr,
+{
+}
+
+/// PostgreSQL specific methods present between CIDR/INET expressions and Bigint expressions
+pub trait PgNetAddExpressionMethods: Expression + Sized {
```
weiznich, line 1091: Should this not just implement the corresponding add operator? So that `column_a + column_b` just works there.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -683,3 +683,592 @@ where
...
+
+impl<T> PgNetAddExpressionMethods for T
+where
+    T: Expression,
+    T::SqlType: InetOrCidr,
+{
+}
+
+/// PostgreSQL specific methods present between CIDR/INET expressions and Bigint expression
+pub trait PgNetSubExpressionMethods: Expression + Sized {
```
weiznich, line 1150: The comment from add also applies here.

diesel/src/pg/expression/mod.rs
```diff
@@ -9,6 +9,8 @@ pub(crate) mod array;
 pub mod array_comparison;
 pub(crate) mod expression_methods;
 pub mod extensions;
+#[cfg(not(feature = "sqlite"))]
```
weiznich, line 12: Why `not(feature = "sqlite")` and not `feature = "postgres")` as the corresponding feature is postgres specific.

weiznich (commented):
Thanks for working on this. I left a few questions I would like to see answered before deciding how to continue on this. Generally I'm in favor of this change.

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: No specific reason, just a personal habit of mine of using `/* ... */` style for multiline comments regardless of the language. Will change should you prefer `///` (`cargo fmt` does not change it which is surprising).

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Ah, nice! I forgot to remove those when porting from the 1.x PR. Will remove.

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Yup, definitely, I actually was not aware that Diesel supported this.

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Same as above.

diesel/src/pg/expression/mod.rs
Tuetuopay: This stems from the way the `sql_function!` works, and to make the library compile with both the `postgres` and `sqlite` features enabled. Without, the following error is thrown out:

```
error[E0277]: the trait bound `sqlite::backend::Sqlite: sql_types::HasSqlType<pg::types::sql_types::Cidr>` is not satisfied
  --> diesel/src/pg/expression/functions.rs:43:1
   |
43 | / sql_function! {
44 | |     /// Returns the network part of the address, zeroing out whatever is to the right of the
45 | |     /// netmask. (This is equivalent to casting the value to cidr.)
46 | |     fn network(addr: Inet) -> Cidr;
47 | | }
   | | ^
   | | |
   | |_the trait `sql_types::HasSqlType<pg::types::sql_types::Cidr>` is not implemented for `sqlite::backend::Sqlite`
   |   in this macro invocation
   |
             <sqlite::backend::Sqlite as sql_types::HasSqlType<sql_types::BigInt>>
             <sqlite::backend::Sqlite as sql_types::HasSqlType<sql_types::Binary>>
             <sqlite::backend::Sqlite as sql_types::HasSqlType<sql_types::Bool>>
             <sqlite::backend::Sqlite as sql_types::HasSqlType<sql_types::Date>>
           and 8 others
```

The gist of it is, `sql_function!` will try to define the function for the SQLite backend regardless of wether it supports it when the `sqlite` feature flag is turned on. When both features are defined (such as in the `bin/test` script), those functions are defined by the Pg-specific code and installed on the SQLite driver.

I am not using `feature = "postgres"` here as the `pg` mod is already behind this feature gate. A more correct way to say it would be "when only postgres is turned on", but I did not bother since this poses no build error with mysql enabled (though it may allow for some erroneous queries at runtime, so it may be a good idea to `not(feature = "sqlite", feature = "mysql")`).

As for the rationale of "not compiling stuff in for multiple backends", I would say it is OK to disable some backend-specific stuff is enabling more than one backend, as you most likely target some cross-backend compatibility by turning multiple ones on.

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: I tried the `Sub` out, and there is an issued regarding the fact that the `-` operator works for two Rhs types:
- `Inet - Bigint -> Inet` (`.sub()`)
- `Inet - Inet -> Biging` (`.diff()`)

As per the doc of Diesel's ops traits, it is not possible to have two Rhs types (and indeed I could not make it work), so one of them will need to be kept the current way (either `.sub()` or `.diff()`).

Personally, I'd rather keep `.diff()` and transition `.sub()` to the `Sub` trait, as to keep the consistency with `Add` which has a `Bigint` Rhs. Any preference?

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Changed in 486373b22

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Removed in bc1bd4a61

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Done in d486715c2

diesel/src/pg/expression/expression_methods.rs
Tuetuopay: Done in d486715c2. I did it for the previous `sub()` and kept `.diff()`

weiznich: :+1:

diesel/src/pg/expression/mod.rs
weiznich: Ok that look interesting. Compiling diesel with more than one backend enabled is definitively something we need to support. In my opinion it is not a solution to just provide don't provide all postgres specific functions in that case, is it is bad design to disable features that should be supported when more backends are enabled. I consider that more a bug of our `sql_function!` implementation that should be fixed there. I've looked a bit into why this is happening, but found no fast fix yet. Therefore I propose the following (as fixing this as part of this PR is probably a lot of work):
* You open a new issue stating that defining that function in a crate enabling the `sqlite` feature and the `postgres` feature will result in the mentioned error.
* We remove this single function (`network`) from the implementation of this PR for now, as this is the only remaining blocker for merging. After all it's quite easy to just define that function in any downstream crate if required. 
* I will try to look into the underlying issue with `sql_function!` and try to come up with a way to fix that behavior.

Tuetuopay: I think this hides a much more deep implementation issue in diesel's `sql_function!` macro. As you noted, the issue only arises for the `network` function, and not the others. The only difference this one has is it takes a concrete type (`Inet`) rather than a generic type (`T: InetOrCird + SingleValue`). Indeed, when using a generic type parameter, it compiles just fine.

Indeed, without the generic parameters, `sql_function!` does generates the `register_impl` and `register_nondeterministic_impl` functions which raise the error. Either this is wanted behaviour (no generics for sqlite), or this is an oversight in the macro.

For the time being, I'll just change `network` to accept `InetOrCidr` like other functions (since the actual postgres function accepts both, this is more correct), and the build should pass fine now.

However, I will still open the issue as this is clearly a bug and the "generics don't work for sqlite" is likely another one.

weiznich: Not accepting generic sql functions for sqlite was an explicit design decision as we need the the concrete types for deserializing the arguments internally. (Custom sql functions in sqlite are implemented by providing a c compatible callback via ffi)

Tuetuopay: Great! Means that switching the function to a generic is a hacky way, but future proof way, to get this to _not_ be installed on the sqlite driver. Seems like we have a solution then (for this PR at least).

I think we are good now.

PR #2579, opened 2020-11-27
Allow re-exporting diesel in other libs

pksunkara:
This change while verbose will allow diesel to be re-exported in other libs to conform with the Rust API guidelines.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -23,8 +23,14 @@ macro_rules! __diesel_column {
     ) => {
         $($meta)*
         #[allow(non_camel_case_types, dead_code)]
-        #[derive(Debug, Clone, Copy, $crate::query_builder::QueryId, Default)]
+        #[derive(Debug, Clone, Copy, Default)]
         pub struct $column_name;
+
+        #[allow(non_camel_case_types)]
+        impl $crate::query_builder::QueryId for $column_name {
+            type QueryId = $column_name;
+            const HAS_STATIC_QUERY_ID: bool = true;
+        }
 
         impl $crate::expression::Expression for $column_name {
             type SqlType = $($Type)*;
@@ -653,12 +659,18 @@ macro_rules! __diesel_table_impl {
             pub const all_columns: ($($column_name,)+) = ($($column_name,)+);
 
             #[allow(non_camel_case_types)]
-            #[derive(Debug, Clone, Copy, $crate::query_builder::QueryId)]
+            #[derive(Debug, Clone, Copy)]
             /// The actual table struct
             ///
             /// This is the type which provides the base methods of the query
             /// builder, such as `.select` and `.filter`.
             pub struct table;
+
+            #[allow(non_camel_case_types)]
+            impl $crate::query_builder::QueryId for table {
+                type QueryId = table;
+                const HAS_STATIC_QUERY_ID: bool = true;
+            }
 
             impl table {
                 #[allow(dead_code)]
@@ -803,12 +815,18 @@ macro_rules! __diesel_table_impl {
                 $($imports)*
 
                 #[allow(non_camel_case_types, dead_code)]
-                #[derive(Debug, Clone, Copy, $crate::query_builder::QueryId)]
+                #[derive(Debug, Clone, Copy)]
                 /// Represents `table_name.*`, which is sometimes needed for
                 /// efficient count queries. It cannot be used in place of
                 /// `all_columns`, and has a `SqlType` of `()` to prevent it
                 /// being used that way
                 pub struct star;
+
+                #[allow(non_camel_case_types)]
+                impl $crate::query_builder::QueryId for star {
+                    type QueryId = star;
+                    const HAS_STATIC_QUERY_ID: bool = true;
+                }
 
                 impl<__GB> $crate::expression::ValidGrouping<__GB> for star
                 where


pksunkara:
Shout out to @flodiebold for helping me figure this issue out yesterday!

diesel/src/macros/mod.rs
```diff
@@ -23,9 +23,15 @@ macro_rules! __diesel_column {
     ) => {
         $($meta)*
         #[allow(non_camel_case_types, dead_code)]
-        #[derive(Debug, Clone, Copy, $crate::query_builder::QueryId, Default)]
+        #[derive(Debug, Clone, Copy, Default)]
         pub struct $column_name;
 
+        #[allow(non_camel_case_types)]
```
weiznich, line 29: Can you elaborate why a manual impl here is required? For me that seems just like a stylistic change.

pksunkara: As explained, this is needed for other libs to allow re-exporting of diesel.

Let's assume a crate lib called `utils` which re exports `diesel` like following:

```rust
pub use diesel;
```

When these macros are used currently like:

```rust
utils::diesel::table! {
    articles (id) {
        id -> Integer,
    }
}
```

the following code gets generated:

```rust
impl ::diesel::query_builder::QueryId for $column_name {
    type QueryId = $column_name;
    const HAS_STATIC_QUERY_ID: bool = true;
}
```

which complains about `diesel` not being found in dependencies (since we are using `utils` as dependency)

```
 --> src/main.rs:1:1
  |
1 | / b::diesel::table! {
2 | |     articles (id) {
3 | |         id -> Integer,
4 | |     }
5 | | }
  | |_^ could not find `diesel` in `{{root}}`
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

This PR changes the generated code to be:

```rust
impl ::utils::diesel::query_builder::QueryId for $column_name {
    type QueryId = $column_name;
    const HAS_STATIC_QUERY_ID: bool = true;
}
```

which ends up letting us compile the program successfully.

weiznich: That makes sense. Thanks for the explanation. Another problem: This change will only fix the `table!` macro. Every `derive`/`proc_macro` will be broken, as all of them assume that `diesel` is some direct dependency. I would rather like to fix that there, instead of having a special fix for `table!` here.

pksunkara: This change fixes the non proc-macros and not the derives because anything which is exporting diesel will probably not rely on the derives.

If people want to use derives, then I would say that they need to explicitly add diesel as dependency. At least this was the consensus I understood talking to people on rust-lang.zulipchat.com

pksunkara: Also fixing this issue on proc-macros is a much more bigger effort and does not look good ergonomic wise which is also a reason any libs that export diesel will not prefer it.

--

The way to fix it (from looking at how `inventory` crate solved it) would be to add a `crate` attr.

```rust
#[derive(Identifiable)]
#[crate = ::utils::diesel]
pub struct User {
  pub id: i32,
}
```

But as I said, it becomes cumbersome for the end user which is why libs do not prefer it.

--

I think we should table the proc-macro correction part until someone needs it.

pksunkara: And I am not even going into the realm of dealing with renaming of diesel dependency. The best way forward to fix the proc macros is for rust to provide us the same `$crate` functionality for them too.

weiznich: I thought a bit more about the consequences of this change: We would basically commit to that `diesel::table!` is reexportable by any other crate without issues, which would basically mean we would commit in terms of semver guarantees that `diesel::table!` is a macro-rule macro and not a proc macro. That's currently true, for mostly historic reasons. Depending on how much features we add to the `table!` macro this likely will not remain true as the implementation is already quite complicated and it is probably much easier to just port the current implementation to a function like proc macro. (Basically similar to `sql_function!` in #1983). At that point we would break the reexportability if we are not able to provide some `$crate` like feature for proc macros.

pksunkara: A different way to make the proc-macros support re-exportability is to remove the `::` in front of `diesel` in the proc macro tokens. That way when `diesel` is used as dependency, it would work as it is now. But when it is used through re-export, the end user would only need to make sure `diesel` is in scope and not do anything else.

```rust
use utils::diesel;

#[derive(Identifiable)]
pub struct User {
  pub id: i32,
}
```

This I think feels like a much better compromise.

weiznich: I think most of the derives in diesel_derive already do this as we use them inside of diesel and they also work in third party crates.

pksunkara: It looks like the `wrap_in_dummy_mod` is the wrong approach and ends up giving issues for re-exporting. The correct approach seems to be https://github.com/rust-lang/rust/issues/56409. I am trying to get it working now in diesel.

weiznich: `wrap_in_dummy_mod` is there to prevent poluting the outer namespace with identifiers imported internally by the derives. I don't think that's something that need to be changed. The imports done inside of `wrap_in_dummy_mod` are likely the problem.

pksunkara: After spending a lot of time debugging, experimenting etc.. this is my final solution.

All *non proc_macros* currently work correctly with re-exporting

All *proc_macros* also work correctly with re-exporting but the end user needs to add this single line to their crate

```rust
use utils::diesel;
```

The issue seems to be that using a *proc_macro* in a *non proc_macro* leads to the end user needing that line again. So, I think this PR is still valid to reduce the instances where the end user needs that line. And re-exportability is not going to break if we change any of these *macros* to *proc_macros* later in the future. This is now just a convenience PR.

weiznich: > And re-exportability is not going to break if we change any of these macros to proc_macros later in the future. 

I don't want to be pedantic here, but how wouldn't re-exportability not change if we change the `table!` macro to a proc macro in a later version. Given your change it's possible to reexport `table!` requirements for down stream users on how to use that macro. If we would change now `table!` to a proc macro that would be a breaking change because down stream code using a reexport would stop to compile as users now would need to add this `use utils::diesel;` line. 

That written I can totally understand that this feature is desirable but I really do not see any good way to implement that without having language support for this.

PR #2580, opened 2020-11-27
Remove some extern crate refs

pksunkara:
Remove some extern crate references where they are not needed

### DIFF
--- a/diesel_migrations/migrations_macros/src/embed_migrations.rs
+++ b/diesel_migrations/migrations_macros/src/embed_migrations.rs
@@ -22,11 +22,8 @@ pub fn expand(path: String) -> proc_macro2::TokenStream {
     quote! {
         #[allow(dead_code)]
         mod embedded_migrations {
-            extern crate diesel;
-            extern crate diesel_migrations;
-
-            use self::diesel_migrations::*;
-            use self::diesel::connection::SimpleConnection;
+            use ::diesel_migrations::*;
+            use ::diesel::connection::SimpleConnection;
             use std::io;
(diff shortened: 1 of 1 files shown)


diesel_migrations/migrations_macros/src/embed_migrations.rs
```diff
@@ -22,11 +22,8 @@ pub fn expand(path: String) -> proc_macro2::TokenStream {
     quote! {
         #[allow(dead_code)]
         mod embedded_migrations {
-            extern crate diesel;
```
weiznich, line 25: The point of those references are that a user does not necessarily have `diesel` imported as diesel. They can have something like `extern crate diesel as diesel_something` or have renamed diesel via a `Cargo.toml` option.

weiznich (changes requested):
I think this would break potential usecases, therefore I would like to know why this is "required" to be changed.

diesel_migrations/migrations_macros/src/embed_migrations.rs
pksunkara: > have renamed diesel via a Cargo.toml option.

Correct me if I am wrong, but in that case, wouldn't this line give a compiler error?

weiznich: Just to write that somewhere: I try to avoid random changes to just adopt some newer pattern there as long as they are only stylistic and don't improve the code in other ways as you loose the ability to just to a `git blame` to see why something was implemented that way.

PR #2585, opened 2020-11-30
[docs] Don't display has_many as an attribute.

przprz:
The docs reads 
```
//! You can declare an association between two records with `#[belongs_to]`.
//! Unlike other ORMs, Diesel has no concept of `#[has_many`]
```
If there's no `has many` in Diesel, why use the argument syntax here?
I found it a little bit confusing when reading for the first time.

### DIFF
--- a/diesel/src/associations/mod.rs
+++ b/diesel/src/associations/mod.rs
@@ -2,7 +2,7 @@
 //!
 //! Associations in Diesel are always child-to-parent.
 //! You can declare an association between two records with `#[belongs_to]`.
-//! Unlike other ORMs, Diesel has no concept of `has_many`
+//! Unlike other ORMs, Diesel has no concept of `has many`
 //!
 //! ```rust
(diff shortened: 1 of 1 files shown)


diesel/src/associations/mod.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! Associations in Diesel are always child-to-parent.
 //! You can declare an association between two records with `#[belongs_to]`.
-//! Unlike other ORMs, Diesel has no concept of `#[has_many`]
+//! Unlike other ORMs, Diesel has no concept of `has_many`
```
przprz, line 5:
```suggestion
//! Unlike other ORMs, Diesel has no concept of `has many`
```

weiznich:
Filled https://github.com/rust-lang/rust/issues/79560 for the nightly failure (it's a upstream rustc bug)

PR #2587, opened 2020-12-03
Add test for inserting null in nullable column with non null default

pksunkara:
It looks like `Insertable` does support `DEFAULT` as you can see in the test [here]( https://github.com/diesel-rs/diesel/blob/master/diesel_tests/tests/insert.rs#L176-L182). It's  happening because of [this](https://github.com/diesel-rs/diesel/blob/master/diesel/src/insertable.rs#L246) line.

Also, `NULL` is inserted when `Option<Option<T>>` is set as `Some(None)` for a column with a non-null default. I have added a test to make sure this doesn't break.


weiznich (approved):
This looks good, but I think it would be great to add another example to the `diesel::insert_into` documentation showing how to do an insert with default.

pksunkara:
We would need to add two examples because we have to show both the cases where a non null is default.

PR #2597, opened 2020-12-18
Make BoolExpressionMethods not always return a Nullable<bool>

### LINKED ISSUE #2589: master: a.and(b) is nullable even though a and b are non nullable
### Versions

- **Rust:** latest stable - 1.48.0 (7eac88abb 2020-11-16)
- **Diesel:** latest master (ff8e5b15ea0dfaa22e9c12c5ed2a0e23992b0617)
- **Database:** postgresql
- **Operating System** Linux

### Feature Flags

- **diesel:** postgres, chrono

## Problem Description
a.and(b) is nullable even though a and b are non nullable

Writing an `and` condition of which I expect the output to be non-nullable.

*compiles*

```
error[E0271]: type mismatch resolving `<Grouped<diesel::expression::operators::And<diesel::expression::nullable::Nullable<Grouped<diesel::expression::operators::Gt<C, diesel::expression::bound::Bound<Timestamptz, DateTime<Utc>>>>>, diesel::expression::nullable::Nullable<Grouped<diesel::expression::operators::Lt<C, diesel::expression::bound::Bound<Timestamptz, DateTime<Utc>>>>>>> as diesel::Expression>::SqlType == Bool`
  --> src/lib.rs:49:29
   |
49 |             InInterval { gt, lt } => Box::new(column.gt(gt).and(column.lt(lt)))
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `Bool`, found struct `diesel::sql_types::Nullable`
   |
   = note: expected struct `Bool`
              found struct `diesel::sql_types::Nullable<Bool>`
   = note: required for the cast to the object type `dyn diesel::BoxableExpression<T, Pg, SqlType = Bool>`
```

no

### Steps to reproduce
```
git clone https://github.com/Ten0/diesel_2589_repro
cd diesel_2589_repro
cargo check
```

Code for reference in case that repo is removed later:
```rust
use diesel::prelude::*;

pub enum TimestampFilter {
	LowerThan(chrono::DateTime<chrono::Utc>),
	GreaterThan(chrono::DateTime<chrono::Utc>),
	InInterval {
		gt: chrono::DateTime<chrono::Utc>,
		lt: chrono::DateTime<chrono::Utc>,
	},
}


pub trait IntoDieselExpr<SqlType> {
	type OutSqlType: diesel::sql_types::SingleValue;
	fn into_diesel_expr<T, C>(
		self,
		column: C,
	) -> Box<dyn BoxableExpression<T, diesel::pg::Pg, SqlType = Self::OutSqlType>>
	where
		T: diesel::Table + 'static,
		SqlType: diesel::sql_types::SingleValue,
		C: Expression<SqlType = SqlType>
			+ diesel::query_builder::QueryFragment<diesel::pg::Pg>
			+ diesel::expression::ValidGrouping<(), IsAggregate = diesel::expression::is_aggregate::No>
			+ diesel::expression::SelectableExpression<T>
			+ Copy
			+ 'static;
}

impl IntoDieselExpr<diesel::sql_types::Timestamptz> for TimestampFilter {
	type OutSqlType = diesel::sql_types::Bool;
	fn into_diesel_expr<T, C>(
		self,
		column: C,
	) -> Box<dyn BoxableExpression<T, diesel::pg::Pg, SqlType = Self::OutSqlType>>
	where
		T: diesel::Table + 'static,
		C: Expression<SqlType = diesel::sql_types::Timestamptz>
			+ diesel::query_builder::QueryFragment<diesel::pg::Pg>
			+ diesel::expression::ValidGrouping<(), IsAggregate = diesel::expression::is_aggregate::No>
			+ diesel::expression::SelectableExpression<T>
			+ Copy
			+ 'static,
	{
		use TimestampFilter::*;
		match self {
			LowerThan(date) => Box::new(column.lt(date)),
			GreaterThan(date) => Box::new(column.gt(date)),
			InInterval { gt, lt } => Box::new(column.gt(gt).and(column.lt(lt)))
				as Box<dyn BoxableExpression<T, diesel::pg::Pg, SqlType = Self::OutSqlType>>,
		}
	}
}

impl IntoDieselExpr<diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>> for TimestampFilter {
	type OutSqlType = diesel::sql_types::Nullable<diesel::sql_types::Bool>;
	fn into_diesel_expr<T, C>(
		self,
		column: C,
	) -> Box<dyn BoxableExpression<T, diesel::pg::Pg, SqlType = Self::OutSqlType>>
	where
		T: diesel::Table + 'static,
		C: Expression<SqlType = diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>>
			+ diesel::query_builder::QueryFragment<diesel::pg::Pg>
			+ diesel::expression::ValidGrouping<(), IsAggregate = diesel::expression::is_aggregate::No>
			+ diesel::expression::SelectableExpression<T>
			+ Copy
			+ 'static,
	{
		use TimestampFilter::*;
		match self {
			LowerThan(date) => Box::new(column.lt(date)),
			GreaterThan(date) => Box::new(column.gt(date)),
			InInterval { gt, lt } => Box::new(column.gt(gt).and(column.lt(lt))),
		}
	}
}

```

## Checklist

Ten0:
Fixes #2589

With new ideas from #2589 and specifically https://github.com/diesel-rs/diesel/issues/2589#issuecomment-748238464, it turns out we can solve the issue encountered when writing 7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd of "not all methods had an equivalent type in `diesel::dsl`" in a simpler way, without making "`BoolExpressionMethods::and` (and `or`) unconditonally nullable" nor being a breaking change in most places.

As a consequence, this mostly reverts the part of 7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd that was focusing on this change, while introducing typing suggested in https://github.com/diesel-rs/diesel/issues/2589#issuecomment-748238464, reducing a bit the overall amount of code.


diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -37,17 +37,14 @@ pub trait BoolExpressionMethods: Expression + Sized {
     /// assert_eq!(expected, data);
     /// #     Ok(())
     /// # }
-    fn and<T>(self, other: T) -> dsl::And<Self, T>
+    fn and<T, ST>(self, other: T) -> dsl::MaybeNullableAnd<Self, T, ST>
```
Ten0, line 40: @weiznich How would you write this if not creating such another wrapper? (https://github.com/diesel-rs/diesel/issues/2589#issuecomment-748304471)

diesel/src/expression_methods/eq_all.rs
```diff
@@ -29,14 +28,14 @@ macro_rules! impl_eq_all {
...
-                Nullable<<$Left1 as EqAll<$Right1>>::Output>,
+                <$Left1 as EqAll<$Right1>>::Output,
                 <($($Left,)+) as EqAll<($($Right,)+)>>::Output,
             >>;
 
             fn eq_all(self, rhs: ($Right1, $($Right,)+)) -> Self::Output {
                 let ($Left1, $($Left,)+) = self;
                 let ($Right1, $($Right,)+) = rhs;
-                $Left1.eq_all($Right1).and(($($Left,)+).eq_all(($($Right,)+)))
+                $Left1.eq_all($Right1).and::<_, Bool>(($($Left,)+).eq_all(($($Right,)+)))
```
Ten0, line 38: So this is the part that is not solved: apparently now compiler is not able to infer which `ST` to use there, which I suspect is because of [this change](https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-1b2dc4703512cd6842e04129b8eab9628b1c555158c7a4cfa332f220b81e5da4) (tests are exploding in a lot of places for the same reason).

Now that probably means we'd have to revert this change, but then it looks like changes from https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-1b2dc4703512cd6842e04129b8eab9628b1c555158c7a4cfa332f220b81e5da4 to https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-04a3838f33b9f0c5cdb21bfaf94920c0bec33b778c02ce092c23ded8d97f2700 as well as https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-7c94f0bbca60b1d1a8174a803d1699d766b1d7365598f296e0879a8f4a5764bb depend on that change, and I'm not sure what these imply and what reverting them would imply.

@weiznich can you clarify what these changes imply?

Ten0: Ok so next commit does revert these and seems to pass test suite.

diesel/src/expression/helper_types.rs
```diff
@@ -87,13 +87,39 @@ pub type Asc<Expr> = super::operators::Asc<Expr>;
...
+/// [`lhs.and(rhs)`](../expression_methods/trait.BoolExpressionMethods.html#method.and)  
+/// when resulting expression is a `Nullable<Bool>`
+pub type NullableAnd<Lhs, Rhs> = MaybeNullableAnd<Lhs, Rhs, Nullable<sql_types::Bool>>;
+
 /// The return type of
 /// [`lhs.and(rhs)`](../expression_methods/trait.BoolExpressionMethods.html#method.and)
-pub type And<Lhs, Rhs> = Grouped<super::operators::And<Nullable<Lhs>, AsExpr<Rhs, Nullable<Lhs>>>>;
+///
+/// If you are attempting to express a query return type explicitly, you may want to use
+/// [And] or [NullableAnd] instead.
```
Mingun, line 104: You forget to add targets for references

diesel/src/expression/helper_types.rs
```diff
@@ -87,13 +87,39 @@ pub type Asc<Expr> = super::operators::Asc<Expr>;
...
+/// [`lhs.or(rhs)`](../expression_methods/trait.BoolExpressionMethods.html#method.or)  
+/// when resulting expression is a `Nullable<Bool>`
+pub type NullableOr<Lhs, Rhs> = MaybeNullableOr<Lhs, Rhs, Nullable<sql_types::Bool>>;
 
 /// The return type of
 /// [`lhs.or(rhs)`](../expression_methods/trait.BoolExpressionMethods.html#method.or)
-pub type Or<Lhs, Rhs> = Grouped<super::operators::Or<Lhs, AsExpr<Rhs, Nullable<Lhs>>>>;
+///
+/// If you are attempting to express a query return type explicitly, you may want to use
+/// [And] or [NullableAnd] instead.
```
Mingun, line 121: You forget to add targets for references

diesel/src/expression/helper_types.rs
```diff
@@ -87,13 +87,39 @@ pub type Asc<Expr> = super::operators::Asc<Expr>;
 /// [`expr.nullable()`](../expression_methods/trait.NullableExpressionMethods.html#method.nullable)
 pub type Nullable<Expr> = super::nullable::Nullable<Expr>;
 
+/// The return type of
+/// [`lhs.and(rhs)`](../expression_methods/trait.BoolExpressionMethods.html#method.and)  
+/// when resulting expression is a non-nullable `Bool`
+pub type And<Lhs, Rhs> = MaybeNullableAnd<Lhs, Rhs, sql_types::Bool>;
```
weiznich, line 93: Instead of having multiple variants here I would like to have just one:

```suggestion
pub type And<Lhs, Rhs, ST = sql_types::Bool> = Grouped<super::operators::And<Lhs, AsExprOf<Rhs, ST>>>;
```

pksunkara (commented):
I might be misunderstanding something, but don't we need to revert the following?

* File https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-cb055b4bb9e7efa6083a07a27019f0d52abeedb051162b636867fe309730d854
* Line https://github.com/diesel-rs/diesel/commit/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd#diff-354105a6b67774b9eeb51dd06d81db7a1b30232369b26c872385803da0f7698eR449

Otherwise looks good.

pksunkara (approved):

pksunkara:
I am still having trouble using this PR.

If there's a column `email VARCHAR` (it is nullable), why does `email.eq(...)` produce `Nullable<Bool>` for boxed expression? Shouldn't `.eq` always result in `Bool`?

```
error[E0271]: type mismatch resolving `<Grouped<diesel::expression::operators::And<Box<dyn BoxableExpression<table, Pg, SqlType = Bool>>, Grouped<diesel::expression::operators::Eq<columns::email, diesel::expression::bound::Bound<diesel::sql_types::Nullable<Text>, Option<&str>>>>>> as diesel::Expression>::SqlType == Bool`
  --> src/main.rs:25:17
   |
25 |     statement = Box::new(statement.and(users::email.eq(Some("John"))));
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `Bool`, found struct `diesel::sql_types::Nullable`
   |
   = note: expected struct `Bool`
              found struct `diesel::sql_types::Nullable<Bool>`
   = note: required for the cast to the object type `dyn BoxableExpression<table, Pg, SqlType = Bool>`
```

Ten0:
> If there's a column `email VARCHAR` (it is nullable), why does `email.eq(...)` produce `Nullable<Bool>` for boxed expression? Shouldn't `.eq` always result in `Bool`?

In SQL, if one of the values of a boolean operator is null, the result may be null, so this is the correct mapping.

pksunkara:
> In SQL, if one of the values of an operator such as AND or OR is null, the result may be null, so this is the correct mapping.

But we are not doing `AND` or `OR` here. I am doing `eq` which is checking for equality.

weiznich:
> I understand eq doesn't give me true when comparing two NULLs, but it still gives me a boolean and not null.

Thats not correct. `1 = NULL` resolves to `NULL`.

Ten0:
> > In SQL, if one of the values of an operator such as AND or OR is null, the result may be null, so this is the correct mapping.
> 
> But we are not doing `AND` or `OR` here. I am doing `eq` which is checking for equality.
> 
> I understand `eq` doesn't give me true when comparing two `NULL`s, but it still gives me a boolean and not null.

I meant `=` as well yes, which is why I pointed to `IS DISTINCT FROM`.

diesel/src/expression/not.rs
```diff
@@ -25,7 +25,7 @@ use crate::sql_types::{Bool, Nullable};
 /// ```
 pub fn not<T>(expr: T) -> not<T>
 where
-    T: AsExpression<Nullable<Bool>>,
+    T: AsExpression<Bool>,
```
weiznich, line 28: I'm not sure if that should only accept non nullable values. I think accepting both `Bool` and `Nullable<Bool>` here is more meaningful. On the other hand that could potentially lead to situations where rustc cannot infer the correct type for `ST` from `AsExpression` here.

weiznich (approved):
Looks good but we should talk about that minor nit regarding to the `not` function.

diesel/src/expression/not.rs
Ten0: Hmm that looks like kind of the same problem as `and`/`or` no? Where if we enforce that it takes an `AsExpression<Nullable<Bool>>` we'll never be able to make it return a non-nullable bool?

Additionally, not all expressions of SqlType `Bool` implement `AsExpression<Nullable<Bool>>` anymore, because otherwise this creates type inference issues with `and`/`or`: https://github.com/diesel-rs/diesel/pull/2597#discussion_r546080041
Hence the following does not compile anymore if we do this: `not(name.eq("Sean")` (from tests)

Finally, I believe the behavior of `not` on `Nullable<Bool>` is a bit of a trap in that `NULL` values result in `NULL`, so they do not change "truthiness" with regards to e.g. a `filter`, which may lead to nasty bugs in case [this kind of confusion](https://github.com/diesel-rs/diesel/pull/2597#issuecomment-748494079) is made and the compiler does not prevent it.

So overall I believe we should keep this behavior where we only accept non-nullable values here.

pksunkara: > I think accepting both Bool and Nullable<Bool>

I had encountered a situation where I want to store the same. Is there a way we can specify that?

weiznich: We must accept both variants (`Bool` and `Nullable<Bool>`) here otherwise it's not possible to write queries like `not(foo.and(bar))` or even `not(foo.eq(bar))` anymore in cases one of the arguments is `Nullable<Bool>`

Ten0: > I had encountered a situation where I want to store the same

Not sure what you mean by "the same".

> We must accept both variants (`Bool` and `Nullable<Bool>`) here otherwise it's not possible to write queries like `not(foo.and(bar))` or even `not(foo.eq(bar))` anymore in cases one of the arguments is `Nullable<Bool>`

So I thought it maybe wasn't a good idea to have `not` support this implicitly, as this may easily lead to nasty bugs:

> I believe the behavior of `not` on `Nullable<Bool>` is a bit of a trap in that `NULL` values result in `NULL`, so they do not change "truthiness" with regards to e.g. a `filter`, which may lead to nasty bugs in case [this kind of confusion](https://github.com/diesel-rs/diesel/pull/2597#issuecomment-748494079) is made and the compiler does not prevent it.

Example:
```rust
table.filter(not(foo.eq(bar)))
```
Where `foo` = `true` and `bar` = `NULL`, the lines won't show, because `not` won't change the "truthiness" of the expression, which is a bit counter-intuitive.

Also, our codebase is rather large and we have zero such patterns.

Given that, do you still believe it's important for `not` to support this? (and we couldn't be happy with a separate `nullable_not`?)

weiznich: I still believe that `not` should support that as this is just the behavior of the underlying SQL operator. I feel that we had this discussion already about a few operators (for example `eq` in another context). Every time the final answer was: We follow the semantic of the underlying SQL operator as we do not try to hide complexity at this level of abstraction. Hiding this complexity can be done in a crate build on top of diesel, providing it owns set of operators that is not based on SQL in my opinion.

Ten0: I agree. Done in https://github.com/diesel-rs/diesel/pull/2597/commits/df8b5e59ef232cd2d517f2e764caab207b39088b.

PR #2598, opened 2020-12-19
Add bind documentation to sql and sql_query functions

EclipsedSolari:
This is a followup from a discussion in the gitter.im channel.

* `sql` and `sql_query` each get a link to their respective `bind` functions and an example containing usage of `bind`.
* `SqlLiteral::bind`'s `UncheckedBind` struct had a broken link back to `SqlLiteral::bind` so I fixed that.
* Since `sql` already links to `sql_query`, I crosslinked `sql_query` back to `sql`.
* I'm not sure why `SqlQuery::bind`'s `UncheckedBind` struct doesn't get documented by cargo, but I added documentation comments to it anyway.

### DIFF
--- a/diesel/src/expression/sql_literal.rs
+++ b/diesel/src/expression/sql_literal.rs
@@ -238,7 +238,7 @@ impl<ST, T, GB> ValidGrouping<GB> for SqlLiteral<ST, T> {
 /// # }
 /// #
 /// # fn run_test_2() -> QueryResult<()> {
-/// #     use self::users::dsl::*;
+/// #     use crate::schema::users::dsl::*;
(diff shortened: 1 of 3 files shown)


weiznich (commented):
From my point of view this change looks fine, but it seems to introduce something that breaks the `query_builder::functions::sql_query` doc tests. (By putting it in some endless loop or something like this, it does not finish on my machine + on CI). That needs to be fixed before merging.

EclipsedSolari:
@weiznich This looks better. Not sure why the CI tests were breaking before, but splitting the doctests on `sql_query` into two functions instead of one fixed it.

weiznich:
Looks good now, thanks :+1:

PR #2599, opened 2020-12-21
Faillible Queryable

### LINKED ISSUE #2523: No way to express inconsistency when deserializing for `Queryable` data since #2182

Ten0:
Fixes #2523

**TODO:**
 - [X] Implement
 - [ ] OPTIONAL: Add support for `#[diesel(deserialize_as)]` at top-level when deriving `Queryable` (as described at https://github.com/diesel-rs/diesel/issues/2523#issuecomment-702348583)
 - [x] Add corresponding entries to the diesel 2.0 update guide


weiznich:
As with #2559 I would like to see a complete port of the repo before deciding anything else here. Generally speaking I think that one may be a better solution than the other PR, but I'm not entirely sure yet.

Ten0:
I beleive that's it :)

weiznich:
@Ten0 Could you please cherry-pick cba105b788b5d783197b964bda7ba8340702eb8d to fix the benchmarks? (Also for the other variant)

diesel/src/deserialize.rs
```diff
@@ -153,7 +153,7 @@ where
     type Row: FromStaticSqlRow<ST, DB>;
 
     /// Construct an instance of this type
-    fn build(row: Self::Row) -> Self;
+    fn build(row: Self::Row) -> Result<Self>;
```
weiznich, line 156: Instead of changing this signature, I suggest the following:

```rust
// Maybe even mark this as deprecated?
fn build(row: Self::Row) -> Self;

fn try_build(row: Self::Row) -> Result<Self> {
    Ok(Self::build(row))
}
```

Using this way less downstream code is broken by this change.

weiznich (commented):
After reviewing both PR's I find that one the better solution. With the suggested change we could even make that not-breaking for downstream code with a minimal maintenance burden on our side.
I still would like to get a benchmark run on this (seems like the CI was not fixed with the last commit, I will push a new fix to master.).

diesel/src/deserialize.rs
Ten0: I'm not convinced this would be the best approach.

It looks like with the `build` function being deprecated, it would indeed almost-prevent from using `build` in other `Queryable` impls, so being forced to write the following when build is faillible would be ok (if we forget how disgusting this is ^^):
```rust
fn build(row: Self::Row) -> Self {
    unreachable!()
    // or to preserve even more backwards compatibility
    Self::try_build(row).expect("you should never call build, but try_build instead")
}

fn try_build(row: Self::Row) -> Result<Self> {
    ...
}
```

However, this would not raise a warning or mention `try_build` in any error when implementing the trait, so it would leave the following part of the original issue un-addressed:
https://github.com/diesel-rs/diesel/issues/2523#issuecomment-703507764
> > Do you mean Queryable is manually implemented in several places in external crates ?
> > In which case I'm surprised because that is not at all something we've been using precisely because it wouldn't be allowed to fail to build these structures, so it doesn't extend our capabilities compared to deriving Queryable and selecting the appropriate fields in the right place.
> 
> There are other reasons that doing actual failable deserialization work why a manual `Queryable` impl can be useful. To just to name a few:
> 
>     * You want to map the results of several queries into a common `enum`
> 
>     * You want to just set some kind of default value for a specific field. (Or any other kind of operation that cannot fail, but adds or removes a field)
> 
> 
> Just to link at least one: [Example from `crates.io`](https://github.com/rust-lang/crates.io/blob/5a08887d4b531e034d01386d3e5997514f3c8ee5/src/models/badge.rs#L101)

https://github.com/diesel-rs/diesel/issues/2523#issuecomment-703781510
> I love this example because I could not have given a better one advocating for my suggestion: this is _precisely_ the kind of problem my suggestion is meant to solve. What happened there seems to be:
> 
>     * We wanted to recompose a structure from an already-deserializable row tuple, which Queryable is meant for
> 
>     * It could fail if the database had unexpected contents, but the function does not allow returning an inconsistency error -> well... I don't really want to query into `RawBadge` (or `(i32, String, serde_json::Value)`) instead of Badge everytime I `.load::<T>(db)` that type of data and then attempt the conversion after, so let's bet it's never going to happen, and we'll `.expect("unexpected data in db")`. => Now it could panic the server if the database had unexpected contents.
> 
> 
> Because we are looking for high availability and reliability even if there happened to be different levels of experience among coders, it's not acceptable for us to panic in these cases. In fact, it looks like a large majority of the case, panicking is not going to be the appropriate answer, while an `Err` on `.load(db)` (that could be sent to an error reporting system with additional contextual data for easy debug while not affecting the rest of program, in particular interacting properly with regards to open transactions getting released to connection pool and not breaking worker threads) would be. So in the end it looks like **if this suggestion forces users to go and update that kind of piece of code, it's actually going reduce the odds of them (and crates.io!) unexpectedly panicking, which is exactly the outcome I would expect and love to see from it!** 😃
> In my opinion, a few of these is very quickly sufficient to outweigh a large number of "oh well I'm going to have to add `Ok(` here".

So it looks like:
1. It would fail to encourage users to fix existing code that panics instead of erroring ([Example from `crates.io`](https://github.com/rust-lang/crates.io/blob/5a08887d4b531e034d01386d3e5997514f3c8ee5/src/models/badge.rs#L106), and also actually one or two places where we'd been lazy (didn't workaround with `FromSqlRow`) in our own codebase, which I'm very happy the current impl had me fix)
1. If the user attempts to implement `Queryable`, the compiler errors would tell them to implement `build` but not mention they may want to implement `try_build` instead.
1. It leaves us with a kind-of-disgusting syntax for use cases where `Queryable` is faillible, I beleive diesel 2.0 would be the best moment to get rid of.

Ten0:
> After reviewing both PR's I find that one the better solution.

That's also what I thought after experimenting with the other one. 🙂

diesel/src/deserialize.rs
weiznich: We could just do what the rust team had done with `std::error::Error`. That would mean deprecate the existing `build` method (so existing users get a warning), provide a default impl for that method (that panics, so new users dont need do handle it) and create a new method with an default impl forwarding to the old one.
Now I'm not really sure if that overhead is it really worth. On the other hand it is not that much overhead, so its probably just fine.

Unrelated to that: I think the name `try_build` is a bettet fit for the name of the new method as it indicates that it could fail.

Ten0: > deprecate the existing `build` method, provide a default impl for that method

Hmm providing a default impl seems to address 2 and most of 3 ("most of" because it's still not super clean because compilation wouldn't require implementing any method, even though it would actually be mandatory to implement one. `Error` didn't have this issue because it already required `Display`, and because `source()` returning None would usually be ok.)

However it seems to not address 1:
>  this would not raise a warning or mention `try_build` in any error when implementing the trait

The warning only triggers when *using* the method, not when implementing it ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=fn%20main()%20%7B%0A%20%20%20%20%0A%7D%0A%0Atrait%20Trait%20%7B%0A%20%20%20%20%23%5Bdeprecated%5D%0A%20%20%20%20fn%20f()%20%7B%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%7D%0A%7D%0A%0Aimpl%20Trait%20for%20()%20%7B%0A%20%20%20%20fn%20f()%20%7B%7D%0A%7D)).

weiznich: > The warning only triggers when using the method, not when implementing it (playground).

That's not good. So going down the route of just having both methods and deprecating one is not really an option :disappointed: 
Ok, that makes changing the method signature a better option. Additionally if we do a breaking change here anyway, it think we should go forward and rename the method to `try_build` as this name indicates that it could fail much more well.

Ten0: That makes sense, however my impression is that prefixing fallible functions by `try_` is pretty rare across the rust ecosystem (unless there's also a non-faillible version it would otherwise be confused with), as well as within diesel ([`FromSql`](https://docs.diesel.rs/master/diesel/deserialize/trait.FromSql.html), [`FromSqlRow`](https://docs.diesel.rs/master/diesel/deserialize/trait.FromSqlRow.html)), probably because in rust, error handling always has to be explicit anyway, so it's not super useful.

Not doing so may ease the migration as well, as it would mention that the return type doesn't match that of the trait, rather than complaining that the function is unknown. ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=trait%20Queryable1%20%7B%0A%20%20%20%20fn%20build()%20-%3E%20Result%3Ci32%2C()%3E%3B%0A%7D%0A%0Atrait%20Queryable2%20%7B%0A%20%20%20%20fn%20try_build()%20-%3E%20Result%3Ci32%2C()%3E%3B%0A%7D%0A%0Aimpl%20Queryable1%20for%20()%20%7B%0A%20%20%20%20fn%20build()%20-%3E%20i32%20%7B%0A%20%20%20%20%20%20%20%20unimplemented!()%0A%20%20%20%20%7D%0A%7D%0A%0Aimpl%20Queryable2%20for%20()%20%7B%0A%20%20%20%20fn%20build()%20-%3E%20i32%20%7B%0A%20%20%20%20%20%20%20%20unimplemented!()%0A%20%20%20%20%7D%0A%7D%0A%0Afn%20main%20()%0A%7B%0A%7D))

That being said given the low count of manual `Queryable` impls in our codebase, it's not something I'm super attached to.

weiznich: The error message is a good argument to keep the name as it is :+1:

weiznich (approved):

Ten0:
Fork dropped, we're finally back to following diesel master, woohoo! 😊
Thanks a lot!

PR #2603, opened 2020-12-21
Know that PG's `IS DISTINCT FROM` may never return null

### LINKED ISSUE #2602: Postgres's IsDistinctFrom incorrectly considers it may return NULL if either of the arguments is Nullable
### Versions

- **Rust:** latest stable
- **Diesel:** latest master (06494e8bdfe774bac9be5cab488b3d0cbec97f4e)
- **Database:** PostgreSQL
- **Operating System** Linux

### Feature Flags

- **diesel:** postgres extras

## Problem Description

As identified at https://github.com/diesel-rs/diesel/pull/2597#issuecomment-748500843, `IS DISTINCT FROM` for PostgreSQL (doc: [postgres](https://www.postgresql.org/docs/12/functions-comparison.html#FUNCTIONS-COMPARISON-PRED-TABLE), [diesel](https://docs.diesel.rs/master/diesel/expression_methods/trait.PgExpressionMethods.html#method.is_distinct_from)) can never return NULL, however, Diesel is not aware of this as it's treated as a standard infix operator:
https://github.com/diesel-rs/diesel/blob/06494e8bdfe774bac9be5cab488b3d0cbec97f4e/diesel/src/pg/expression/operators.rs#L5
https://github.com/diesel-rs/diesel/blob/06494e8bdfe774bac9be5cab488b3d0cbec97f4e/diesel/src/expression/operators.rs#L229-L237

Ten0:
Resolves #2602

- Refactor infix_operator to only express the expression SQL type and bounds once
- Add capability to have a constant (or different nullability constraints) to infix_operator (undocumented for now)
- Use new feature of infix_operator in `IS DISTINCT FROM` and `IS NOT DISTINCT FROM`

### DIFF
--- a/diesel/src/expression/operators.rs
+++ b/diesel/src/expression/operators.rs
@@ -181,6 +181,15 @@ macro_rules! __diesel_operator_to_sql {
 ///
 /// // Specifying both the return types and the backend
 /// infix_operator!(And, " && ", TsQuery, backend: Pg);
+///
+/// // When nullability does not depend on whether one of the arguments
+/// // is nullable:
+/// infix_operator!(
+///     IsDistinctFrom,
+///     " IS DISTINCT FROM ",
+///     ConstantNullability diesel::sql_types::Bool,
+///     backend: Pg
+/// );
 /// ```
 ///
 /// ## Example usage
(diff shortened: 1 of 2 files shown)


weiznich:
While I really appreciate that you work on fixing all this issues, I'm quite sure that opening all those PR's that fast is a good way to burn me out on maintaining diesel. Each time my backlog grows I'm asking myself How I should ever handle that on my own and unfortunately there is no one else I can share that work with. It's probably a good idea to give my some time to work through all those PR's and issues before opening any new, even if that likely means waiting a few weeks.

Ten0:
> While I really appreciate that you work on fixing all this issues, I'm quite sure that opening all those PR's that fast is a good way to burn me out on maintaining diesel. Each time my backlog grows I'm asking myself How I should ever handle that on my own and unfortunately there is no one else I can share that work with. It's probably a good idea to give my some time to work through all those PR's and issues before opening any new, even if that likely means waiting a few weeks.

You have all my moral support with regards to the amount of open stuff. ^^
If you want to add me to the reviewers ping so I can try and help review some of the PRs as well that's fine by me.

I had to implement these anyway because I wanted to finally get rid of #2274, which we still had as we were pinned to the commit just before #2182 was merged because of the follow-up issues. (Now we're about to move to a hopefully temporary fork of the latest master which just has these few PRs merged in advance.)

That being said, once they are implemented I can think of no better place to tidy them than in the open-awaiting-for-review PRs list, should they not be reviewed right away because we don't have enough time to allocate to them (which is completely understandable, thanks a lot for all the time you are putting in Diesel in any case <3).

weiznich (changes requested):
While I'm generally in favor of this change I'm do not see why the `infix_operator!` macro needs to be changed that way or better: 
Why we cannot write this change as
```rust
infix_operator!(IsDistinctFrom, " IS DISTINCT FROM ", Bool, backend: Pg);
```

by just specifying a return type?

weiznich:
> If you want to add me to the reviewers ping so I can try and help review some of the PRs as well that's fine by me.

I've added you to the corresponding team. Any help reviewing open active PR's is definitively welcome.

pksunkara:
I am actually relying on the return type of this being billable now. 😓

weiznich:
@pksunkara I'm not sure how that would be an argument to not fix this issue. That's just the cost of using a unreleased (and therefore not finished) version of diesel.

Ten0:
> Why we cannot write this change as
> 
> ```rust
> infix_operator!(IsDistinctFrom, " IS DISTINCT FROM ", Bool, backend: Pg);
> ```
> 
> by just specifying a return type?

Because the way `infix_operator` is currently written, whatever the input type, it will be made Nullable based on the whether the input arguments are nullable:
https://github.com/diesel-rs/diesel/blob/06494e8bdfe774bac9be5cab488b3d0cbec97f4e/diesel/src/expression/operators.rs#L229-L237
Not setting the return type is just like setting it to `Bool`
https://github.com/diesel-rs/diesel/blob/06494e8bdfe774bac9be5cab488b3d0cbec97f4e/diesel/src/expression/operators.rs#L220-L222

I beleive that was a smart choice in general, because most of the SQL operators match that behavior, e.g.
https://github.com/diesel-rs/diesel/blob/06494e8bdfe774bac9be5cab488b3d0cbec97f4e/diesel/src/pg/expression/operators.rs#L20-L22

Which is why my approach was to add support for specifying "please don't do that".

weiznich:
> Which is why my approach was to add support for specifying "please don't do that".
> 
> Another approach would have been to write what the macro generates manually for both these operators and any other one that also needs this and we might not have noticed yet, reducing the complexity of that macro but increasing that of defining such an operator. I would probably have gone for that approach if there was only one, but because there were at least 2 I was considering writing a small macro for implementing both, so I thought I might as well add support for this in the original macro.

OK, that makes sense now, should have looked at this macro a bit more in detail. Nevertheless I do not thing it's that common that a operator has this behavior, therefore I think we should not mention this feature as part of the documentation and we should not consider that feature to be part of the public API. (Mainly because I feel that the current implementation does not really fit into the API provided by the macro + I do not see a clear way to improve that API.)

diesel/src/expression/operators.rs
```diff
@@ -222,29 +231,65 @@ macro_rules! infix_operator {
...
+
+    ($name:ident, $operator:expr, ConstantNullability $($return_ty:tt)::*, backend: $backend:ty) => {
+        $crate::infix_operator!(
+            name = $name,
+            operator = $operator,
+            return_ty = ($($return_ty)::*),
+            backend_ty_params = (),
+            backend_ty = $backend,
+        );
+    };
```
weiznich, line 271: Following up on my last comment: We should hide this variants from rustdoc.

weiznich (approved):

pksunkara (approved):

pksunkara:
Can we expand this same usage to `IS NULL` and `IS NOT NULL` too? 

https://github.com/Ten0/diesel/blob/fix_2602_is_distinct_from_is_never_null/diesel/src/expression/operators.rs#L485-L486

Ten0:
> Can we expand this same usage to `IS NULL` and `IS NOT NULL` too?
> 
> https://github.com/Ten0/diesel/blob/fix_2602_is_distinct_from_is_never_null/diesel/src/expression/operators.rs#L485-L486

`IS NULL`/`IS NOT NULL` already has the correct output, as it's declared as a `postfix_operator`, whose output's nullability does not depend on the nullability of the arguments:
https://github.com/diesel-rs/diesel/blob/e8c8d167d56ad12a271535f9027a7e9768ce93f8/diesel/src/expression/operators.rs#L331-L355
(It's not an `infix_operator` so it doesn't go through the same macro.)

Ten0:
Considering this is deemed finished and was validated twice a week ago, is it ok to merge it, or are we waiting for additional input?

weiznich:
@Ten0 It's just that I've taken some days off. As mentioned already too often: This is a free time project for me, so if I don't have time or motivation to work on it I'm not doing anything which includes merging PR's.

Ten0:
> @Ten0 It's just that I've taken some days off. As mentioned already too often: This is a free time project for me, so if I don't have time or motivation to work on it I'm not doing anything which includes merging PR's.

Sorry, I didn't mean to put any pressure: it's completely ok if this has to wait. I just wanted to make sure this wouldn't stale due to a misunderstanding on who should click merge, given it's approved and I now have the button.

weiznich:
>  I just wanted to make sure this wouldn't stale due to a misunderstanding on who should click merge, given it's approved and I now have the button.

I would be surprised if you have merge rights already. As far as I know you are part of the `diesel/reviewers` team. Persons in that team get pinged by requesting a review from the corresponding team, otherwise there are no other rights. Merging PR's is a right that is granted to people in the `diesel/contributors` and `diesel/core` team. The normal process to get into one of those teams is by showing via review/contributions that you understand larger parts of the code base and have gained some trust of the members of the core team.

Ten0:
> I would be surprised if you have merge rights already. As far as I know you are part of the `diesel/reviewers` team.

That makes sense semantically, which is why I was surprised I now have the button. ^^
I haven't tried clicking it though, so maybe it would tell me I don't have rights if I did.

weiznich:
@Ten0 I've rechecked that now twice. That sounds strange, as you are not in the correct team, but let's just try that out on the next merge or so (I will just ping you there :shrug:)

PR #2610, opened 2020-12-26
Show examples of sqlite connnection urls

pickfire:


Ten0 (approved):

weiznich:
I think this is a nicer way to present the same information as before, but it does not add an `example of sqlite connection urls` as the PR description claims. Nerveless it's worth merging this change.

PR #2629, opened 2021-01-22
Fix downloading the source code for running our benchmarks

weiznich:


pksunkara:
I think this can be merged as soon as it's rebased.

pksunkara (approved):

Issue #2475, opened 2020-08-03
How to use table name different from the default?


rainyday:
How can I manually specify a table name for a Queryable item? I've looked around the documentation and github and it looks like this feature was added a few years ago in https://github.com/diesel-rs/diesel/pull/1233? Unfortunately, I'm unable to find any documentation on how to actually use this feature.
## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)

weiznich:
`Queryable` is not related to an table in any way. `Queryable` is a trait that maps the results of an query to a user provided type, where a query does not necessarily contain exactly one table. Therefore there is no way to "use a different table name" as this does not even make sense for what `Queryable` stand for.

Closed as not an issue.

rainyday:
@weiznich Ok, let me rephrase using the getting started example:

Given these contents of models.rs:

```rust
#[derive(Queryable)]
pub struct Post {
    pub id: i32,
    pub title: String,
    pub body: String,
    pub published: bool,
}
```

Diesel generates this schema.rs

```rust
table! {
    posts (id) {
        id -> Nullable<Integer>,
        title -> Text,
        body -> Text,
        published -> Bool,
    }
}
```

My question is: how is the table name `posts` determined and how can I configure it to be something else instead?

Issue #2169, opened 2019-09-20
sqlite: Queryable struct field order can silently swap columns


kpcyrd:
### Versions

- **Diesel:** 1.4.2
- **Database:** sqlite

## Problem Description

I've noticed a bug in my program that, after some debugging, turned out to be due to diesel assigning a column to the wrong field. I've resolved this issue by checking the order of the sqlite columns and defining my struct fields in the exact same order.

I think this issue should be detected both at compile time (the expected database schema is known to generate `src/schema.rs`) and at runtime in case a certain combination of database migrations/rollbacks messes up the order.

### Steps to reproduce

- create a table and a `Queryable struct`
- make all fields/columns of the same type
- define the struct fields in a different order than the sqlite table (`a, b, c` and `a, c, b`)
- insert a row and query it
- notice that the value of the `c` column is in the `b` field and vice versa.

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.
- [X] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)

weiznich:
That's [documented behaviour](http://docs.diesel.rs/diesel/deserialize/trait.Queryable.html#deriving).

> When this trait is derived, it will assume that the order of fields on your struct match the order of the fields in the query. This means that field order is significant if you are using #[derive(Queryable)]. Field name has no effect.

> I think this issue should be detected both at compile time (the expected database schema is known to generate src/schema.rs) and at runtime in case a certain combination of database migrations/rollbacks messes up the order.

Unfortunately not, because:
a) Informations about the requested column names are not available at runtime
b) how would you handle complex select clauses at compile time. 
(Side note: The field ordering in a auto generated `table!` call is deterministic, so that won't be a issue)

Duplicate to #2011 (and some other issues)

Could probably be resolved by introducing something like #2038

Issue #2011, opened 2019-03-13
the trait bound `*const str: diesel::deserialize::FromSql<diesel::sql_types::Timestamp, diesel::pg::Pg>` is not satisfied


GopherJ:
### Versions

- **Rust:** rustc 1.33.0-nightly (daa53a52a 2019-01-17)
- **Diesel:** 1.4.1
- **Database:** postgresql@10
- **Operating System** Ubuntu 18.10

### Feature Flags

- **diesel:** "r2d2", "postgres", "chrono", "uuid"

## Problem Description

This error message shows when compile.

### What is the expected output?

No error output

### What is the actual output?

Error message
![Screenshot from 2019-03-13 11-33-15](https://user-images.githubusercontent.com/33961674/54272376-dfd7b980-4583-11e9-861f-6911da7a7f0f.png)

### Are you seeing any additional errors?

![Screenshot from 2019-03-13 11-34-12](https://user-images.githubusercontent.com/33961674/54272412-f3832000-4583-11e9-9c2b-4437398e45c3.png)

##Other useful code

`src/schema.rs`
```rust
#![allow(proc_macro_derive_resolution_fallback)]
table! {
    access_tokens (id) {
        id -> Varchar,
        client_id -> Varchar,
        grant_type -> Varchar,
        issued_at -> Timestamp,
        scope -> Varchar,
        expires_at -> Timestamp,
        user_id -> Varchar,
    }
}

table! {
    acls (id) {
        id -> Varchar,
        allow -> Int4,
        ipaddr -> Nullable<Varchar>,
        username -> Nullable<Varchar>,
        clientid -> Nullable<Varchar>,
        access -> Int4,
        topic -> Varchar,
        created_at -> Timestamp,
        created_by -> Varchar,
    }
}

table! {
    auth_codes (id) {
        id -> Varchar,
        client_id -> Varchar,
        scope -> Varchar,
        expires_at -> Timestamp,
        redirect_uri -> Varchar,
        user_id -> Varchar,
    }
}

table! {
    clients (id) {
        id -> Varchar,
        name -> Varchar,
        secret -> Varchar,
        redirect_uri -> Varchar,
        response_type -> Varchar,
    }
}

table! {
    dashboards (id) {
        id -> Varchar,
        application_namespace -> Varchar,
        config -> Text,
        created_at -> Timestamp,
        user_id -> Varchar,
    }
}

table! {
    grant_types (name) {
        name -> Varchar,
    }
}

table! {
[...]

Issue #1926, opened 2018-11-27
table!() allows for swapping two fields of the same type


drozdziak1:
### Versions

- **Rust:**`rustc 1.30.1 (1433507eb 2018-11-07)`
- **Diesel:** `1.3.3`
- **Database:** SQLite 
- **Operating System**Arch Linux, 4.19.4-arch1-1-ARCH

### Feature Flags

- **diesel:** N/A

## Problem Description
This struct:
```rust
#[derive(Queryable, Debug, Clone, PartialEq, Eq, Default)]
pub struct ChannelStateRecord {
    pub id: i64,
    pub channel_id: Vec<u8>,
    pub address_a: Vec<u8>,
    pub address_b: Vec<u8>,
    pub nonce: Vec<u8>,

    pub balance_a: Vec<u8>,
    pub balance_b: Vec<u8>,

    pub sig_a_v: Option<Vec<u8>>,
    pub sig_a_r: Option<Vec<u8>>,
    pub sig_a_s: Option<Vec<u8>>,

    pub sig_b_v: Option<Vec<u8>>,
    pub sig_b_r: Option<Vec<u8>>,
    pub sig_b_s: Option<Vec<u8>>,
}
```
...works with this schema:
```rust
table! {
    states (id) {
        id -> Bigint,
        address_a -> Binary,
        address_b -> Binary,
        channel_id -> Binary,
        nonce -> Binary,
        balance_a -> Binary,
        balance_b -> Binary,
        sig_a_v -> Nullable<Binary>,
        sig_a_r -> Nullable<Binary>,
        sig_a_s -> Nullable<Binary>,
        sig_b_v -> Nullable<Binary>,
        sig_b_r -> Nullable<Binary>,
        sig_b_s -> Nullable<Binary>,
    }
}
```

### What are you trying to accomplish?
Hook my struct up with the database

### What is the expected output?
The schema has a couple fields swapped (`channel_id` comes before the addresses in `ChannelStruct`). And yet it compiles. Everything should be golden. **BUT** it's not, even though it compiles, the `table!()` macro appears to rely on field order and swapping any two of the same type seems to be enough to fool compile-time checks. As a result, my program reads/writes wrong fields off/to wrong columns. I'd expect Diesel to either be smart enough to point the problem out or to be able to use the schema correctly anyway. **I realize this looks an awful lot like forgetting to , but I did write a script that reproduces this **

### What is the actual output?
It builds and then uses what it built wrong.

### Steps to reproduce
[...]

PR #2631, opened 2021-01-22
Explain that not only the order of fields needed

pickfire:
Queryable need to match all the fields, not just the order but the number
**in some cases** (I don't know what case).

@weiznich @Mingun https://github.com/diesel-rs/diesel/pull/2627#issuecomment-765284022

I am confused myself about the behavior too, I only know in some cases it needs to be the same, after I read the docs I became more confused (since it works even when the number of fields is not the same but at least the order is the same), I tried my best to reduce the confusion saying that all the items needs to be the same.

At first I even thought this will do a smart job extracting the fields needed, like the struct being part of the schema.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -266,7 +266,7 @@ pub trait QueryDsl: Sized {
     /// let join = users::table.left_join(posts::table);
     ///
     /// // By default, all columns from both tables are selected.
-    /// // Note, the structs must contain all the fields from original schemas in order.
+    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
     /// let all_data = join.load::<(User, Option<Post>)>(&connection)?;
     /// let expected_data = vec![
     ///     (User::new(1, "Sean"), Some(Post::new(post_id, 1, "Sean's Post"))),
@@ -460,8 +460,8 @@ pub trait QueryDsl: Sized {
     ///     .execute(&connection)
     ///     .unwrap();
     ///
-    /// // It can load into struct as long as the struct contains all the
-    /// // fields in the schema in the same order without explicit select.
+    /// // By default, all columns from both tables are selected.
+    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
     /// let data = users
     ///     .inner_join(posts.on(title.like(name.concat("%"))))
     ///     .load::<(User, Post)>(&connection); // type could be elided
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -265,7 +265,8 @@ pub trait QueryDsl: Sized {
     /// #         .first::<i32>(&connection)?;
     /// let join = users::table.left_join(posts::table);
     ///
-    /// // By default, all columns from both tables are selected
+    /// // By default, all columns from both tables are selected.
+    /// // Note, the structs must contain all the fields from original schemas in order.
     /// let all_data = join.load::<(User, Option<Post>)>(&connection)?;
     /// let expected_data = vec![
     ///     (User::new(1, "Sean"), Some(Post::new(post_id, 1, "Sean's Post"))),
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -266,7 +266,8 @@ pub trait QueryDsl: Sized {
     /// let join = users::table.left_join(posts::table);
     ///
     /// // By default, all columns from both tables are selected.
-    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
+    /// // If no explicit select clause is used this means that the result
+    /// // type of this query must contain all fields from the original schema in order.
     /// let all_data = join.load::<(User, Option<Post>)>(&connection)?;
     /// let expected_data = vec![
     ///     (User::new(1, "Sean"), Some(Post::new(post_id, 1, "Sean's Post"))),
@@ -461,7 +462,9 @@ pub trait QueryDsl: Sized {
     ///     .unwrap();
     ///
     /// // By default, all columns from both tables are selected.
-    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
+    /// // If no explicit select clause is used this means that the
+    /// // result type of this query must contain all fields from the
+    /// // original schema in order.
     /// let data = users
     ///     .inner_join(posts.on(title.like(name.concat("%"))))
     ///     .load::<(User, Post)>(&connection); // type could be elided
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -423,6 +423,60 @@ pub trait QueryDsl: Sized {
     /// assert_eq!(Ok(expected_data), data);
     /// # }
     /// ```
+    ///
+    /// ### With explicit `ON` clause (struct)
+    ///
+    /// ```rust
+    /// # include!("../doctest_setup.rs");
+    /// # use schema::{users, posts};
+    /// #
+    /// # /*
+    /// allow_tables_to_appear_in_same_query!(users, posts);
+    /// # */
+    ///
+    /// # fn main() {
+    /// #     use self::users::dsl::{users, name};
+    /// #     use self::posts::dsl::{posts, user_id, title};
+    /// #     let connection = establish_connection();
+    /// #[derive(Debug, PartialEq, Queryable)]
+    /// struct User {
+    ///     id: i32,
+    ///     name: String,
+    /// }
+    ///
+    /// #[derive(Debug, PartialEq, Queryable)]
+    /// struct Post {
+    ///     id: i32,
+    ///     user_id: i32,
+    ///     title: String,
+    /// }
+    ///
+    /// diesel::insert_into(posts)
+    ///     .values(&vec![
+    ///         (user_id.eq(1), title.eq("Sean's post")),
+    ///         (user_id.eq(2), title.eq("Sean is a jerk")),
+    ///     ])
+    ///     .execute(&connection)
+    ///     .unwrap();
+    ///
+    /// // It can load into struct as long as the struct contains all the
+    /// // fields in the schema in the same order without explicit select.
+    /// let data = users
+    ///     .inner_join(posts.on(title.like(name.concat("%"))))
+    ///     .load::<(User, Post)>(&connection); // type could be elided
+    /// let expected_data = vec![
+    ///     (
+    ///         User { id: 1, name: String::from("Sean") },
+    ///         Post { id: 4, user_id: 1, title: String::from("Sean's post") },
+    ///     ),
+    ///     (
+    ///         User { id: 1, name: String::from("Sean") },
+    ///         Post { id: 5, user_id: 2, title: String::from("Sean is a jerk") },
+    ///     ),
+    /// ];
+    /// assert_eq!(Ok(expected_data), data);
+    /// # }
+    /// ```
     fn inner_join<Rhs>(self, rhs: Rhs) -> InnerJoin<Self, Rhs>
     where
         Self: JoinWithImplicitOnClause<Rhs, joins::Inner>,
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -72,16 +72,16 @@ use diagnostic_shim::*;
 ///
 /// # Attributes
 ///
-/// ## Optional type attributes
+/// ## Optional container attributes
 ///
-/// * `#[table_name = "some_table"]`, specifies the table for which the
-/// current type is a changeset. Requires that `some_table` is in scope.
-/// If this attribute is not used, the type name converted to
-/// `snake_case` with an added `s` is used as table name
 /// * `#[changeset_options(treat_none_as_null = "true")]`, specifies that
 /// the derive should threat `None` values as `NULL`. By default
 /// `Option::<T>::None` is just skipped. To insert a `NULL` using default
 /// behavior use `Option::<Option<T>>::Some(None)`
+/// * `#[table_name = "some_table"]`, specifies the table for which the
+/// current type is a changeset. Requires that `some_table` is in scope.
+/// If this attribute is not used, the type name converted to
+/// `snake_case` with an added `s` is used as table name
 ///
 /// ## Optional field attributes
 ///
@@ -113,14 +113,18 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes:
 ///
-/// ## Required type attributes
+/// ## Required container attributes
 ///
 /// * `#[sql_type = "SqlType"]`, to specify the sql type of the
 ///  generated implementations. If the attribute exists multiple times
 ///  impls for each sql type are generated.
 ///
-/// ## Optional type attribute
+/// ## Optional container attributes
 ///
+/// * `#[table_name = "some_table"]`, specifies the table for which the
+/// current type is a changeset. Requires that `some_table` is in scope.
+/// If this attribute is not used, the type name converted to
+/// `snake_case` with an added `s` is used as table name
 /// * `#[diesel(not_sized)]`, to skip generating impls that require
 ///   that the type is `Sized`
 #[proc_macro_derive(AsExpression, attributes(diesel, sql_type))]
@@ -135,7 +139,7 @@ pub fn derive_as_expression(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// # Required type attributes
+/// # Required container attributes
 ///
 /// * `#[belongs_to(User)]`, to specify a child-to-parent relation ship
 /// between the current type and the specified parent type (`User`).
@@ -147,7 +151,7 @@ pub fn derive_as_expression(input: TokenStream) -> TokenStream {
 /// appended `_id` is used as foreign key name. (`user_id` in this example
 /// case)
 ///
-/// # Optional type attributes
+/// # Optional container attributes
 ///
 /// * `#[table_name = "some_table_name"]` specifies the table this
 ///    type belongs to. Requires that `some_table_name` is in scope.
@@ -203,7 +207,7 @@ pub fn derive_from_sql_row(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// ## Optional type attributes
+/// ## Optional container attributes
 ///
 /// * `#[table_name = "some_table_name"]` specifies the table this
 ///    type belongs to. Requires that `some_table_name` is in scope.
@@ -212,8 +216,6 @@ pub fn derive_from_sql_row(input: TokenStream) -> TokenStream {
 /// * `#[primary_key(id1, id2)]` to specify the struct field that
 ///    that corresponds to the primary key. If not used, `id` will be
 ///    assumed as primary key field
-///
-///
 #[proc_macro_derive(Identifiable, attributes(table_name, primary_key, column_name))]
 pub fn derive_identifiable(input: TokenStream) -> TokenStream {
     expand_proc_macro(input, identifiable::derive)
@@ -251,7 +253,7 @@ pub fn derive_identifiable(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// ## Optional type attributes
+/// ## Optional container attributes
 ///
 /// * `#[table_name = "some_table_name"]`, specifies the table this type
 /// is insertable into. Requires that `some_table_name` is in scope.
@@ -395,30 +397,9 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 /// This trait can only be derived for structs, not enums.
 ///
 /// **When this trait is derived, it will assume that __all fields on your
-/// struct__ is the same as __all fields in the query__, including the order.
-/// This means that field order is significant if you are using
-/// `#[derive(Queryable)]`. Field name has no effect. For example,
-///
-/// ```rust
-/// # extern crate diesel;
-/// # use diesel::prelude::*;
-/// table! {
-///     posts {
-///         id -> Integer,
-///         user_id -> Integer,
-///         title -> VarChar,
-///     }
-/// }
-///
-/// #[derive(Queryable, Debug)]
-/// pub struct Post {
-///     id: i32,
-///     user_id: i32,
-///     title: String,
-///     // this cannot be id, user_id (missing title)
-///     // this cannot be id, title, user_id (incorrect order)
-/// }
-/// ```
+/// struct__ is the same as __all fields in the query__, including the order
+/// and count. This means that field order is significant if you are using
+/// `#[derive(Queryable)]`. Field name has no effect.
 ///
 /// To provide custom deserialization behavior for a field, you can use
 /// `#[diesel(deserialize_as = "SomeType")]`. If this attribute is present, Diesel
@@ -432,14 +413,20 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// ## Optional field attributes:
+/// ## Optional container attributes
+///
+/// * `#[table_name = "some_table_name"]`, specifies the table this type
+/// is insertable into. Requires that `some_table_name` is in scope.
+/// If this attribute is not used, the type name converted to
+/// `snake_case` with an added `s` is used as table name
+///
+/// ## Optional field attributes
 ///
 /// * `#[diesel(deserialize_as = "Type")]`, instead of deserializing directly
 ///   into the field type, the implementation will deserialize into `Type`.
 ///   Then `Type` is converted via
 ///   [`.try_into`](https://doc.rust-lang.org/stable/std/convert/trait.TryInto.html#tymethod.try_into)
 ///   into the field type. By default this derive will deserialize directly into the field type
-///
 ///
 /// # Examples
 ///
@@ -614,6 +601,7 @@ pub fn derive_queryable(input: TokenStream) -> TokenStream {
 ///   the derive will use the sql type of the corresponding column.
 ///
 /// ## Field attributes
+///
 /// * `#[column_name = "some_column"]`, overrides the column name for
 ///    a given field. If not set, the name of the field is used as column
 ///    name. This attribute is required on tuple structs, if
@@ -842,7 +830,7 @@ pub fn derive_sql_type(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// ## Optional type attributes
+/// ## Optional container attributes
 ///
 /// * `#[diesel(aggregate)]` for cases where the type represents an aggregating
 ///   SQL expression
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -394,11 +394,31 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 ///
 /// This trait can only be derived for structs, not enums.
 ///
-/// **When this trait is derived, it will assume that __all fields on your struct__
-/// matches __all fields in the query__, including the order. This means that field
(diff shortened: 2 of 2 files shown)


diesel_derives/src/lib.rs
```diff
@@ -397,7 +397,8 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 /// **When this trait is derived, it will assume that the order of fields on your
 /// struct match the order of the fields in the query. This means that field
 /// order is significant if you are using `#[derive(Queryable)]`. Field name has
-/// no effect.**
+/// no effect. In some cases, all the fields in the struct (including the order)
+/// must be exactly the same (like `Eq`) as the fields in the query.**
```
Ten0, line 401: I'm not sure what cases you're referring to that don't respect that rule.
How about:
```suggestion
/// no effect.** There should always be as many fields to your structure as
/// elements in the tuple used as in the `select` part of the query.
```

pickfire: In some cases like when join it needs the whole items, not as many fields but all the fields. But I am not sure if there are any other cases besides that, I only encountered it once (I rarely use diesel).

Ten0: > not as many fields but all the fields
I don't understand. Can you explain along with an example of a code that exhibits that behavior?

pickfire: https://github.com/diesel-rs/diesel/pull/2627#issuecomment-764347892

weiznich: @pickfire As already explained there: This is not related to joins. The behaviour is the same for every query. To word it really simply: Each query has a at compile time know SQL type, which is based on the expressions in your select clause (note that expressions are not necessarily columns…). Let's name that `ST`. If you now call `.load()` (or a similar method) the result struct needs to implement `Queryable<ST, _>` for your return type. The `#[derive(Queryable)]` now implements `Queryable` by just using mapping each return field by index to one of your struct fields, so that `ST` corresponds to each rust type. There is really no magic going on here and the behaviour is just the same for all queries.

Ten0: > https://github.com/diesel-rs/diesel/pull/2627#issuecomment-764347892

As I had read that topic, this addition is meant to specifically explain that case, as the issue was that through the inner join, the query implicitly selected all the fields of the `comments` table (4 of them), while the struct that was built only held two.
```rust
/// There should always be as many fields in your structure as
/// elements in the tuple used as in the `select` part of the query. This also
/// applies if the `select` is implicit.
```

Ten0: It looks like what might be missing from the doc is the recursive aspect. (Tuple of tuple expecting a struct of `Queryable` which themselves have the proper amount of fields.)

pickfire: Yeah, I think what I am saying here is the implicit recursive selection, I am not aware of the distinction between implicit and explicit selection, maybe this should be explained in `load` in which an explicit `select` is needed if a join is used, since it became a tuple.

I believe the usual case is that one can `load` a single struct with implicit select if there are less fields in the struct but not in the case of join.

weiznich: Ok, the whole discussion/problem seems to be more about what's the default select clause in which case, then how `#[derive(Queryable)]` work.
For me it does not seem like a good idea to document that all on `fn load()` is this function is not related to constructing the select clause.
There are exactly three cases for how the select clause looks like:
* A user passes an explicit select clause via [select](http://docs.diesel.rs/1.4.x/diesel/query_dsl/trait.QueryDsl.html#method.select). It think the examples and the explanation there cover this case quite well. Otherwise that's the right location to improve documentation about this case.
* A user does not call `.select(…)` and does not perform a join. This is the case with the weakest current documentation, as this is done implicitly and there is no "right" place to document this. It is somewhat covered in the [Getting Started](https://diesel.rs/guides/getting-started/) but that does not go in depth there.
* A user does not call `.select(…)` but performs one or more joins. This case is documented via [inner_join](http://docs.diesel.rs/1.4.x/diesel/query_dsl/trait.QueryDsl.html#method.inner_join) (All other join methods like this, it's basically the same for them). Now this documentation is quite detailed about the return types, but could mentioned a bit more what's the behaviour if there is already an existing select clause and how the new default select clause is constructed.

The right place to cover all that would probably be to write a `All about selects` guide (See https://github.com/diesel-rs/diesel/issues/1108), but at least I do not have the time to do that.

> I believe the usual case is that one can load a single struct with implicit select if there are less fields in the struct but not in the case of join. Maybe we need to explain the need to have explicit select in some parts?

Again: The rules for calling `load()` are the same, regardless if there is an join or not. There is no such thing like a implicit select that just drops fields and that's an explicit design decision of diesel. The field could and structure needs to match with that data structure you are loading data into. (At least as long as you use the derive…)

Ten0: > I believe the usual case is that one can `load` a single struct with implicit select if there are less fields in the struct but not in the case of join.

I don't think there is such a case even when there are no joins: it's always mandatory when deriving `Queryable` that there are `as many fields in your structure as elements in the tuple used as in the 'select' part of the query`, even with an implicit select.

pickfire: Is there a way to prevent footguns like these where we can check the original schema to see if it matches the struct?

pickfire:
I have updated it to say that all fields needs to be the same.

diesel_derives/src/lib.rs
weiznich: The main point of having something like `Queryable` is that it does **not** require the schema to work, because otherwise you wouldn't be able to represent the result of a more complex query with it. How would you handle the following SQL with only checking the schema? `SELECT departments.id, 42, max(employees.salary) FROM departments INNER JOIN employees ON departments.id = employees.department GROUP BY departments.id `? Such  query does not necessarily contain references to only one table, could mix columns through arbitrarily complex expressions and can even contain expressions without reference to columns. You cannot check that with "the schema".

diesel_derives/src/lib.rs
```diff
@@ -394,10 +394,11 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
...
 /// This trait can only be derived for structs, not enums.
 ///
-/// **When this trait is derived, it will assume that the order of fields on your
-/// struct match the order of the fields in the query. This means that field
+/// **When this trait is derived, it will assume that __all fields on your struct__
+/// matches __all fields in the query__, including the order. This means that field
 /// order is significant if you are using `#[derive(Queryable)]`. Field name has
-/// no effect.**
+/// no effect. Think `Eq`, if you have `a, b, c` in your schema, you need to have
+/// `a, b, c` in the struct derived, not `a, b` or `a, c, b`.**
```
weiznich, line 401: If we write something like that in the documentation, I would like to see a concrete example here, not something generic like this. This is not really easy to understand for anyone else.

diesel_derives/src/lib.rs
```diff
@@ -394,10 +394,11 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 ///
 /// This trait can only be derived for structs, not enums.
 ///
-/// **When this trait is derived, it will assume that the order of fields on your
-/// struct match the order of the fields in the query. This means that field
+/// **When this trait is derived, it will assume that __all fields on your struct__
+/// matches __all fields in the query__, including the order. This means that field
```
weiznich, line 398: We should maybe be clearer on what `matches` exactly means in this context.

weiznich (changes requested):

diesel_derives/src/lib.rs
```diff
@@ -394,10 +394,31 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
...
+/// #[derive(Queryable, Debug)]
+/// pub struct Post {
+///     id: i32,
+///     user_id: i32,
+///     title: String,
+///     // this cannot be id, user_id (missing title)
+///     // this cannot be id, title, user_id (incorrect order)
+/// }
+/// ```
 ///
```
Mingun, line 422: I think that the example should also show calculated fields, because for me it is still unclear why diesel can't at compile time calculate for each field in struct correct index of it in the `table!` macro. this should be as simple as get field name and find its index in the list of fields declared in the `table!` macro

diesel_derives/src/lib.rs
pickfire: Added

diesel_derives/src/lib.rs
pickfire: I changed it to "is the same as"

diesel_derives/src/lib.rs
pickfire: I didn't know how calculated fields works, maybe you can just provide a suggestion?

diesel_derives/src/lib.rs
```diff
@@ -394,11 +394,31 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
...
+/// }
+///
+/// #[derive(Queryable, Debug)]
+/// pub struct Post {
+///     id: i32,
+///     user_id: i32,
+///     title: String,
+///     // this cannot be id, user_id (missing title)
+///     // this cannot be id, title, user_id (incorrect order)
+/// }
```
weiznich, line 420: I'm sorry to write that, but this whole example is missing a few really important points:

* Just having a `table!` and a struct deriving `Queryable` means nothing for diesel. There is no connection between those two data structures as long as you don't have a query loading the result for the table into the data structure.
* Field names do not matter as written in the block above. So just renaming some field to something different will be fine as long as the types match
* It is really important to point out it is totally fine to use a struct implementing/deriving `Queryable` with one or more queries, which could reference as lot tables as you want.

diesel_derives/src/lib.rs
weiznich: @Mingun Please stop making random claims without having something to show. As you seem to be sure that this is just easy, it shouldn't be to hard for you to explain on the following example how this would work:

```rust
table! {
    users (id) {
        id -> Integer,
        name -> Text,
    }
}

table! {
    posts(id) {
        id -> Integer,
        title -> Text,
        body -> Nullable<Text>,
        user_id -> Integer,
    }
}

joinable!(…);
allow_tables_to_appear_in_same_query!(…);

#[derive(Queryable)]
struct MyResult {
    user_id: i32,
    post_id: Option<i32>,
    some_function_result: i32,
    some_integer_constant: i32
}

sql_function!(fn foo(a: Text, b: Integer) -> Integer);

fn some_not_so_complex_query(conn: &PgConnection) -> QueryResult<MyResult> {
    users::table.left_join(posts::table)
        .select((users::id, posts::id.nullable(), foo(users::name, users::id), 42.into_sql::<Integer>()))
        .load(conn)
}
```

Especially I want to know how you would address the following points:
* Handle that a column coming via a left join is nullable
* Handle the results of sql functions (or any other expression that is not "just" a table column)
* Handle constants
* Handle unnamed fields, like the result of the sql function or the constant.

diesel_derives/src/lib.rs
weiznich: Sorry for being so pedantic here, but I wouldn't say the "the field are the same as", as this would imply that they are the same fields. The query contains different types then the type deriving `Queryable`.

pickfire: But I thought they need to have all the fields?

diesel_derives/src/lib.rs
Mingun: @weiznich, that's exactly what I'm trying to convey to you. Why do you report this example only here in the comments, but you cannot understand that it should be in the documentation? The very fact of the presence of this PR, [as](https://github.com/diesel-rs/diesel/issues/2475) [well](https://github.com/diesel-rs/diesel/issues/2169) [as](https://github.com/diesel-rs/diesel/issues/2011) [many](https://github.com/diesel-rs/diesel/issues/1926) [and](https://github.com/diesel-rs/diesel/issues/1518) [many](https://github.com/diesel-rs/diesel/issues/1472) [other](https://github.com/diesel-rs/diesel/issues/573) [issues](https://github.com/diesel-rs/diesel/issues/421), confirms that the behavior of `Queryable` is incomprehensible not only to me.

Besides, in the [getting started guide](http://diesel.rs/guides/getting-started/) (which will be one of the first places to get information of the idiomatic diesel code) we see:

> Using `#[derive(Queryable)]` assumes that the order of fields on the `Post` struct matches the columns in the `posts` table, so make sure to define them in the order seen in the `schema.rs` file.

and in the API docs:
1.4.x: https://github.com/diesel-rs/diesel/blob/bec389a1ea329c161fc7a8fb8df8bea29a3432bf/diesel/src/deserialize.rs#L27-L30
master: https://github.com/diesel-rs/diesel/blob/8f149f9cdc085974283dc6e997390ed51bf7ec19/diesel_derives/src/lib.rs#L397-L400

If `Queryable` is not connected to the `table!` macro at all how order can be significant? And order of what? Can you specify it for your own example? But if `Queryable` is connected to `table!`, why you should keep order in sync manually? Compiler have enough information to do that for you.

weiznich: Ok, again: I belive both of you @Mingun and @pickfire are missing a few really important point here:

* First of all, diesel never claimed to work like other ORM's. If people believe that and try to apply their knowledge from say active record here, they will fail. (In the end that's similar to rust itself, if you try to squeeze your object orientated knowledge in there you will run into the borrow checker really fast). So please stop to transfer some knowledge that you have about ORM models to `Queryable`. If it where the same we would just have named it `Model` instead.
* Our documentation is a place to lookup things for all users, not only for beginners. Therefore we cannot simplify things here, as this would block more advanced usages of our API. All formulations give so far only target a single "problem" you've hit while starting diesel. It's great that you try to improve the documentation there, but please accept that I want to keep things written in our documentation true for all cases, not only for a single edge case.
* As pointed out multiple times above: Both of you seem to belief that `Queryable` is something magical. It is not, its doing something really simple. For me it seems like you attribute nearly all behaviour coming from a query dsl method call chain to `.load()` or `Queryable`, while in practice each of the parts have a separate influence. In the end that's the reason why I'm so hasty to add all this to the documentation of `Queryable`/`#[derive(Queryable)]`. It's just not the right place for this in my opinion. 
(I've already pointed this out [here](https://github.com/diesel-rs/diesel/pull/2631#discussion_r564534126))

Now I could say, technically it's clearly documented how `Queryable` works, as this is part of the trait bounds of our [public API](https://docs.diesel.rs/1.4.x/diesel/query_dsl/methods/trait.LoadQuery.html#implementors) (that's just on click away from `RunQueryDsl::load()`). Technically you just need to read those bounds and everything should be clear. I totally understand that this can be challenging and adding some more explicit documentation won't hurt, therefore here a short explanation on how I understand that query able works:

Let `Q` be some select query, that has a select clause `(S1, S2, …, Sn)`. Each select clause is required to be a valid `Expression`, therefore we can determine the SQL type of that select clause as `(ST1, ST2, …, Sn)` where `S1: Expression<SqlType = ST1>` and so on. For `load` it is now required to have a `Queryable` impl of the form `Queryable<(ST1, ST2, …, Sn), DB>` for your type where `DB` is just the used backend.  Now let's assume that you have the you want to derive `Queryable` for the following type:

```rust
struct MyType {
    first_field: T1,
    second_field: T2,
    …
    last_field: Tn,
}
```

All `#[derive(Queryable)]` does now is providing the following impl:
(using the `Queryable` definition from the master branch here, but the idea is the same for the 1.4.x releases.
```rust
impl<ST1, ST2, …, STn, DB> Queryable<(ST1, ST2, …, STn), DB> for MyType {
    type Row = (T1, T2, …, Tn);
    
    fn row((first_field, second_field, …, last_field): Self::Row) -> Result<Self> {
        Ok(Self { first_field, second_field, …, last_field } )
    }
}
```

As you easily can see does this impl nothing magical. It's just a stupid simple field mapping, that does not even know about `table!` or whatever. In my opinion most of the problems you've listed above are coming from people that do not understand how the select clause is composed, and not from how `#[derive(Queryable)]` work. That's just where the misunderstand surfaces, not where the actual issue lies. It is certainly possible to fix the underlying problem with documentation, but as pointed out above already: The API documentation on the relevant functions is already good in my opinion, what is missing is a overall guide that shows how this functions interact with each other. As already pointed out multiple times: I would be happy if someone writes such a guide, but as this is really a lot of work and I'm doing most of the work on diesel in my free time I'm not willing to spend that time on something that I cannot use for my own projects. 

So this is likely my last comment to this PR, at least as long as the changes do not reflect what I've written above.

pickfire: Oh, I think I get it now after your explanation but I think things are still missing in the documentation, at least not in the `Queryable` trait part since that is pretty clear.

https://docs.diesel.rs/1.4.x/diesel/query_dsl/trait.RunQueryDsl.html#returning-a-struct

Because of that example didn't mention that the struct needs to have all the fields in the table so that's why I am confused.

> Types which implement Queryable represent the result of a SQL query. This does not necessarily mean they represent a single database table.

Combine that with `Queryable` trait doc stuff become more confusing. From my point of view, the example above to load struct thinks of the struct as a "model" rather than something "queryable". A "model" needs to have all the fields but "queryable" does not need to have all the fields.

Maybe within the code we can mention that that only works if the `Queryable` struct have all the fields in the table in the same order. Do you think that is good to remind others a second time?

weiznich: > Because of that example didn't mention that the struct needs to have all the fields in the table so that's why I am confused.

Because again: For the general case that's just not true. You cannot write there that the return type will need to have all fields of an table, because that opens a lot of questions about which table exactly, how this works with select clauses and so on.

>  From my point of view, the example above to load struct thinks of the struct as a "model" rather than something "queryable". A "model" needs to have all the fields but "queryable" does not need to have all the fields.

Can you point at some of our documentation mentioning the term "model". I'm quite sure that we explicitly avoid this term for this reason. For me that seems more like you are trying to apply your knowledge from different ORM's to diesel.

> Maybe within the code we can mention that that only works if the `Queryable` struct have all the fields in the table in the same order.
 
Within which code? Also I see using the formulation " all the fields in the table in the same order" critically, as I already know that someone will then try to use literally `users::name` as struct field type or so.

pickfire:
```rust
#[derive(Queryable, PartialEq, Debug)]
struct User {
    id: i32,
    name: String,
}

let data = users
    .load::<User>(&connection)?;
let expected_data = vec![
    User { id: 1, name: String::from("Sean"), },
    User { id: 2, name: String::from("Tess"), },
];
assert_eq!(expected_data, data);
```
The example only shows that, people may not notice that `User` includes every item from `users` table, to be used this way. I mean maybe the example is lacking some context. It did not mention the word "model" but it feels like a model to me, maybe just my view, so I did the same thing (plus less fields) and it didn't work.

weiznich: In my understanding this example just points out that instead of receiving a tuple from the database (as shown in the example above) you could also map that to a struct containing the same fields in the same order. Having written that I can at least understand why you got that impression, so maybe that could be pointed out there explicitly?

pickfire: Yes, if we could have a comment saying that it is a **`Queryable` struct containing the same fields in the same order** that would be very good. At least others know that if it is just a normal struct or does not have all the fields or not in the same order would not work. It could give readers a clear context and the conditions required to do so.

Another suggestion I have is we could show an example on using `(User, Post)` in either `load` or `inner_join` docs since struct load is shown in `load` (without `explict` select), and we could mention the same thing in the above paragraph. I could send a pull request for this if you think it is a good idea (like the example above but with join).

weiznich: > Yes, if we could have a comment saying that it is a Queryable struct containing the same fields in the same order that would be very good. At least others know that if it is just a normal struct or does not have all the fields or not in the same order would not work. It could give readers a clear context and the conditions required to do so.

Adding such a statement to the documentation of `load` seems fine as long as "same fields" refers explicitly to tuple variant above.

> Another suggestion I have is we could show an example on using (User, Post) in either load or inner_join docs since struct load is shown in load (without explict select), and we could mention the same thing in the above paragraph. I could send a pull request for this if you think it is a good idea (like the example above but with join).

I believe that should be added to the examples section in the documentation of `inner_join`.

diesel_derives/src/lib.rs
pickfire: > Just having a table! and a struct deriving Queryable means nothing for diesel. There is no connection between those two data structures as long as you don't have a query loading the result for the table into the data structure.

Okay, will amend it with `load`.

> It is really important to point out it is totally fine to use a struct implementing/deriving Queryable with one or more queries, which could reference as lot tables as you want.

Ah, I am not aware of this.

diesel_derives/src/lib.rs
pickfire: I have updates the examples for this.

pickfire:
@weiznich Do we still need the `FIXME` in `QueryDsl::select`?

diesel_derives/src/lib.rs
pickfire: This part is moved.

diesel_derives/src/lib.rs
pickfire: Used `matches`

diesel_derives/src/lib.rs
```diff
@@ -394,10 +396,10 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
...
 /// This trait can only be derived for structs, not enums.
 ///
-/// **When this trait is derived, it will assume that the order of fields on your
-/// struct match the order of the fields in the query. This means that field
-/// order is significant if you are using `#[derive(Queryable)]`. Field name has
-/// no effect.**
+/// **Note**: When this trait is derived, it will assume that __all fields on
+/// your struct__ matches __all fields in the query__, including the order and
+/// count. This means that field order is significant if you are using
+/// `#[derive(Queryable)]`. Field name has no effect.
```
weiznich, line 402: I would like to keep the part with the field names bold here. That's the point people miss quite often.

diesel_derives/src/lib.rs
```diff
@@ -411,15 +413,21 @@ pub fn derive_query_id(input: TokenStream) -> TokenStream {
 ///
 /// # Attributes
 ///
-/// ## Optional field attributes:
+/// ## Optional container attributes
+///
+/// * `#[table_name = "some_table_name"]`, specifies the table this type
+/// is insertable into. Requires that `some_table_name` is in scope.
+/// If this attribute is not used, the type name converted to
+/// `snake_case` with an added `s` is used as table name
```
weiznich, line 421: We do not accept a `#[table_name]` attribute for `Queryable`.

weiznich (changes requested):
Thanks for the update. This sounds much better now :+1:
I've left a few comments with some minor formulation improvements + 2 places where the change incorrectly mentions that a `#[table_name]` attribute is allowed while it isn't.
If that's changed I'm fine with merging this.

diesel/src/query_dsl/mod.rs
```diff
@@ -265,7 +265,8 @@ pub trait QueryDsl: Sized {
     /// #         .first::<i32>(&connection)?;
     /// let join = users::table.left_join(posts::table);
     ///
-    /// // By default, all columns from both tables are selected
+    /// // By default, all columns from both tables are selected.
+    /// // Note, the structs must contain all the fields from original schemas in order.
```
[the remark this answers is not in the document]
pickfire:
```suggestion
    /// // By default, all columns from both tables are selected.
    /// // For this case the loading target type (struct, enum) must contain all the fields from original schemas in order.
```

Would this be better, not sure if "loading target type" is a good terminology because I don't understand it on first glance.

diesel/src/query_dsl/mod.rs
```diff
@@ -423,6 +424,60 @@ pub trait QueryDsl: Sized {
...
+    /// diesel::insert_into(posts)
+    ///     .values(&vec![
+    ///         (user_id.eq(1), title.eq("Sean's post")),
+    ///         (user_id.eq(2), title.eq("Sean is a jerk")),
+    ///     ])
+    ///     .execute(&connection)
+    ///     .unwrap();
+    ///
+    /// // It can load into struct as long as the struct contains all the
+    /// // fields in the schema in the same order without explicit select.
```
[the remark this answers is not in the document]
pickfire: See above comment

diesel_derives/src/lib.rs
```diff
@@ -113,14 +113,18 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
...
 ///  generated implementations. If the attribute exists multiple times
 ///  impls for each sql type are generated.
 ///
-/// ## Optional type attribute
+/// ## Optional container attributes
 ///
+/// * `#[table_name = "some_table"]`, specifies the table for which the
+/// current type is a changeset. Requires that `some_table` is in scope.
+/// If this attribute is not used, the type name converted to
+/// `snake_case` with an added `s` is used as table name
```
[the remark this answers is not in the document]
pickfire:
```suggestion
```

diesel_derives/src/lib.rs
pickfire: I personally find bolding everything hard to read, so that is why I bold only the first **Note** part. Because I could also miss it when the full text is bold but it doesn't seemed like something important without the special **Note** text. Do you want me to change it back?

diesel_derives/src/lib.rs
pickfire: Ops, I thought that was `Insertable`, my mistake, I got confused a bit.
```suggestion
```

diesel_derives/src/lib.rs
weiznich: That's correct, therefore I would only change the last sentence to bold.
```suggestion
/// **Note**: When this trait is derived, it will assume that __all fields on
/// your struct__ matches __all fields in the query__, including the order and
/// count. This means that field order is significant if you are using
/// `#[derive(Queryable)]`. __Field name has no effect__.
```

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
weiznich: What's about
```suggestion
    /// // By default, all columns from both tables are selected.
    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
```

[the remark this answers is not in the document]
pickfire: "result type of this query" does not seemed that good but I guess still okay, even though it is a bit confusing to me at first glance.

diesel/src/query_dsl/mod.rs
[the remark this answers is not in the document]
pickfire: Applied

diesel_derives/src/lib.rs
pickfire: I believe this is resolved right?

diesel/src/query_dsl/mod.rs
```diff
@@ -265,7 +265,8 @@ pub trait QueryDsl: Sized {
     /// #         .first::<i32>(&connection)?;
     /// let join = users::table.left_join(posts::table);
     ///
-    /// // By default, all columns from both tables are selected
+    /// // By default, all columns from both tables are selected.
+    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
```
weiznich, line 269:
```suggestion
    /// // If no explicit select clause is used this means that the result 
    /// // type of this query must contain all fields from the original schema in order.
```

diesel/src/query_dsl/mod.rs
```diff
@@ -423,6 +424,60 @@ pub trait QueryDsl: Sized {
...
+    /// diesel::insert_into(posts)
+    ///     .values(&vec![
+    ///         (user_id.eq(1), title.eq("Sean's post")),
+    ///         (user_id.eq(2), title.eq("Sean is a jerk")),
+    ///     ])
+    ///     .execute(&connection)
+    ///     .unwrap();
+    ///
+    /// // By default, all columns from both tables are selected.
+    /// // If no explicit select clause is used this means that the result type of this query must contain all fields from the original schema in order.
```
weiznich, line 464:
```suggestion
    /// // If no explicit select clause is used this means that the 
    /// // result type of this query must contain all fields from the 
    /// // original schema in order.
```

weiznich (approved):

PR #2637, opened 2021-01-29
Fix support for `pg_temp` in `PgMetadataLookup`

Ten0:
Lacking support for `pg_temp` breaks a variety of tests (in particular that of https://github.com/adwhit/diesel-derive-enum/pull/58)

### DIFF
--- a/diesel/src/pg/backend.rs
+++ b/diesel/src/pg/backend.rs
@@ -77,7 +77,7 @@ impl PgTypeMetadata {
     /// The [OID] of `T[]`
     ///
     /// [OID]: https://www.postgresql.org/docs/current/static/datatype-oid.html
-    pub fn array_oid(self) -> Result<u32, impl std::error::Error + Send + Sync> {
+    pub fn array_oid(&self) -> Result<u32, impl std::error::Error + Send + Sync> {
         self.0.as_ref().map(|i| i.array_oid).map_err(Clone::clone)
     }
 }
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -54,26 +54,44 @@ fn lookup_type(
     conn: &PgConnection,
 ) -> QueryResult<InnerPgTypeMetadata> {
     let search_path: String;
+    let mut search_path_has_temp_schema = false;
 
-    let search_schema = if let Some(schema) = cache_key.schema.as_ref() {
-        vec![schema.as_ref()]
-    } else {
-        search_path = crate::dsl::sql("SHOW search_path").get_result::<String>(conn)?;
+    let search_schema = match cache_key.schema.as_deref() {
+        Some("pg_temp") => {
+            search_path_has_temp_schema = true;
+            Vec::new()
+        }
+        Some(schema) => vec![schema],
+        None => {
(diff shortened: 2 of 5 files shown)


diesel/src/pg/metadata_lookup.rs
```diff
@@ -53,10 +53,17 @@ fn lookup_type(
...
     conn: &PgConnection,
 ) -> QueryResult<InnerPgTypeMetadata> {
+    const TEMP_SCHEMA_NAME: &str = "pg_temp";
     let search_path: String;
-
-    let search_schema = if let Some(schema) = cache_key.schema.as_ref() {
-        vec![schema.as_ref()]
+    let mut search_path_has_temp_schema = false;
+
+    let search_schema = if let Some(schema) = cache_key.schema.as_deref() {
```
weiznich, line 60: I would prefer to use a match statement for this instead, so that we don't have that nested if here.

diesel/src/pg/metadata_lookup.rs
```diff
@@ -65,15 +72,27 @@ fn lookup_type(
...
-        .filter(pg_namespace::nspname.eq(crate::dsl::any(search_schema)))
         .filter(pg_type::typname.eq(&cache_key.type_name))
-        .select((pg_type::oid, pg_type::typarray))
-        .first(conn)?;
+        .select((pg_type::oid, pg_type::typarray));
+    let nspname_filter = pg_namespace::nspname.eq(crate::dsl::any(search_schema));
+
+    let metadata = if search_path_has_temp_schema {
+        metadata_query
+            .filter(nspname_filter.or(pg_namespace::oid.eq(crate::dsl::sql("pg_my_temp_schema()"))))
```
weiznich, line 91: Could we define that function as ordinary sql function. `dsl::sql` will break the prepared statement caching here.

weiznich (changes requested):

diesel/src/pg/metadata_lookup.rs
Ten0: hmm as far as I'm concerned I like the clear distinction between "infer schema" and "provided schema".
I see "provided schema" possibly being `pg_temp` as a sub-part of that.

That's fine for me if you still want me to change though.

diesel/src/pg/backend.rs
```diff
@@ -78,7 +78,7 @@ impl PgTypeMetadata {
     ///
     /// [OID]: https://www.postgresql.org/docs/current/static/datatype-oid.html
     pub fn array_oid(self) -> Result<u32, impl std::error::Error + Send + Sync> {
```
weiznich, line 80: Seems i've missed that now twice. That should match the oid function above.

```suggestion
     pub fn array_oid(&self) -> Result<u32, impl std::error::Error + Send + Sync> {
```

diesel/src/pg/metadata_lookup.rs
weiznich: I think that distinction remains clear with the following pattern:

```rust
match cache_key.schema.as_deref() {
   Some(pg_temp) => {...}
   Some(schema) => {...}
   None => { infer schema stuff here}
}
```

weiznich (approved):

PR #2647, opened 2021-02-15
Implement `To/FromSql` for `IpAddr` enum

Turbo87:
https://github.com/diesel-rs/diesel/pull/749 implemented `Inet` support for Postgres using the `ipnetwork` crate. This PR adds support for `std::net::IpAddr` too.

The `Inet` type in Postgres supports including a subnet mask (see https://www.postgresql.org/docs/current/datatype-net-types.html), which the `IpAddr` enum does not support. For the simple use case of saving plain IP addresses to the database that does not seem necessary though, since we can assume a subnet mask of 32 for IPv4 addresses and 128 for IPv6, which corresponds to a full IP address without a subnet. When reading from the database I've added `assert_or_error!()` calls that check if the subnet mask length corresponds to the two values above.

Ideally this would mean that we could avoid the extra `network-address` feature flag when only supporting `std::net::IpAddr`, but unfortunately we still need the `libc` dependency to get access to `AF_INET`.

### DIFF
--- a/diesel/src/pg/types/network_address.rs
+++ b/diesel/src/pg/types/network_address.rs
@@ -3,7 +3,7 @@ extern crate libc;
 
 use self::ipnetwork::{IpNetwork, Ipv4Network, Ipv6Network};
 use std::io::prelude::*;
-use std::net::{Ipv4Addr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use crate::deserialize::{self, FromSql, FromSqlRow};
 use crate::pg::{Pg, PgValue};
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR.

We normally only provide `FromSql`/`ToSql` impls for types in diesel if the following rules are full filled:

1. Is this a commonly used type. (Yes because there is already support for `Inet` in diesel).
2. Could we provide a FromSql and a ToSql implementation for that type. (In this case Yes)
3. Could the conversion go somehow wrong because the sql type or the rust type could hold values that are not re presentable in the other type. (In this case the conversion from `Inet` to `std::net::IpAddr` can fail since it is missing subnet support. )

Given those points I do not feel that this impls belong into diesel, especially as you write that you need to assume things to make this work.

I'm leaving this open for discussion for other members of @diesel-rs/contributors for now, if no new argument comes up I will close this PR in a few days/weeks.

Turbo87:
I guess that sounds reasonable.

> Given those points I do not feel that this impls belong into diesel

Are you implying that this should be implemented in a dedicated crate instead? Since `IpAddr` is from the stdlib and `To/FromSql` are from diesel itself, wouldn't that make it hard to implement the trait for `IpAddr` from a third-party crate?

weiznich:
> Are you implying that this should be implemented in a dedicated crate instead? Since IpAddr is from the stdlib and To/FromSql are from diesel itself, wouldn't that make it hard to implement the trait for IpAddr from a third-party crate?

I'm only saying: It does not belong into diesel for the reasons listed above. I know that you cannot implement it in a third party crate either, but diesel already provides hooks for customizing the (de)serialization behaviour for unsupported types. See `#[diesel(deserialize_as = "Type")]`/`#[diesel(serialize_as = "Type")]` for the corresponding derives for details.

diesel/src/pg/types/network_address.rs
```diff
@@ -56,6 +56,76 @@ macro_rules! assert_or_error {
...
 
+impl FromSql<Inet, Pg> for IpAddr {
+    fn from_sql(value: PgValue<'_>) -> deserialize::Result<Self> {
+        // https://github.com/postgres/postgres/blob/55c3391d1e6a201b5b891781d21fe682a8c64fe6/src/include/utils/inet.h#L23-L28
+        let bytes = value.as_bytes();
+        assert_or_error!(4 <= bytes.len(), "input is too short.");
+        let af = bytes[0];
+        let prefix = bytes[1];
+        let net_type = bytes[2];
+        let len = bytes[3];
```
henryboisdequin, line 67:
```suggestion
        let (af, prefix, net_type, len) = (bytes[0], bytes[1], bytes[2], bytes[3]);
```

henryboisdequin (changes requested):
IMHO, this looks a little cleaner

weiznich:
Closed for the reasons listed above. I think one week is a reasonable time to hear other opinions here. If something comes up later on we can always reopen this PR.

stormshield-gt:
Just for clarification the workaround with `#[diesel(deserialize_as = "Type")]/#[diesel(serialize_as = "Type")]` don't work for `IpAddr` because `From<String>` is not implemented for `IpAddr`

weiznich:
@stormshield-gt  Otherwise posting questions at old closed PR's is discouraged. If you feel that something is unclear or needs clarification you are welcome to open a new discussion thread detailing your problem, including all necessary information to understand what exactly you want to know.

PR #2653, opened 2021-02-24
Improve general code

henryboisdequin:
This PR:

1. Improves `malformed_sql_query` test code

### DIFF
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -130,7 +130,7 @@ mod tests {
     use std::env;
 
     fn connection() -> MysqlConnection {
-        let _ = dotenv::dotenv();
+        dotenv::dotenv().ok();
         let database_url = env::var("MYSQL_UNIT_TEST_DATABASE_URL")
             .or_else(|_| env::var("MYSQL_DATABASE_URL"))
             .or_else(|_| env::var("DATABASE_URL"))
--- a/diesel/src/mysql/types/date_and_time.rs
+++ b/diesel/src/mysql/types/date_and_time.rs
@@ -164,7 +164,7 @@ mod tests {
     use crate::sql_types::{Date, Datetime, Time, Timestamp};
 
     fn connection() -> MysqlConnection {
-        dotenv().ok();
+        let _ = dotenv();
 
         let connection_url = ::std::env::var("MYSQL_UNIT_TEST_DATABASE_URL")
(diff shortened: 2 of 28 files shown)


diesel/src/pg/connection/mod.rs
```diff
@@ -193,15 +193,11 @@ mod tests {
...
         let query =
             crate::sql_query("SELECT not_existent FROM also_not_there;").execute(&connection);
+
         if let Err(err) = query {
-            match err {
-                DatabaseError(_, string) => {
-                    assert_eq!(Some(26), string.statement_position());
-                }
-                _ => unreachable!(),
+            if let DatabaseError(_, string) = err {
```
weiznich, line 198: This test should fail if something other than that error is returned. The old variant ensured this, while this one will silently accept other error messages.

diesel/src/mysql/types/date_and_time.rs
```diff
@@ -164,7 +164,7 @@ mod tests {
     use crate::sql_types::{Date, Datetime, Time, Timestamp};
 
     fn connection() -> MysqlConnection {
-        dotenv().ok();
+        let _ = dotenv();
```
weiznich, line 167: I would wait with applying this change till the documentation of `dotenv()` itself recommends this: https://github.com/dotenv-rs/dotenv/pull/57

weiznich (changes requested):

diesel/src/mysql/types/date_and_time.rs
henryboisdequin: @weiznich I've addressed your first comment. Now, I'll remove all the `dotenv()` changes, I will make another PR once that lands.

diesel/src/mysql/connection/mod.rs
```diff
@@ -130,7 +130,7 @@ mod tests {
     use std::env;
 
     fn connection() -> MysqlConnection {
-        let _ = dotenv::dotenv();
```
henryboisdequin, line 133: Should I keep this? I think it's better to use `.ok()` to keep consistency for now.

diesel_cli/src/infer_schema_internals/information_schema.rs
```diff
@@ -359,7 +359,7 @@ mod tests {
     use std::env;
 
     fn connection() -> PgConnection {
-        let _ = dotenv();
```
henryboisdequin, line 362: Same here

PR #2658, opened 2021-02-28
Format Markdown

henryboisdequin:
This PR formats markdown to have consistency in the codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,208 +8,209 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 ### Added
 
-* `Connection` and `SimpleConnection` traits are implemented for a broader range
-  of `r2d2::PooledConnection<M>` types when the `r2d2` feature is enabled.
+-   `Connection` and `SimpleConnection` traits are implemented for a broader range
+    of `r2d2::PooledConnection<M>` types when the `r2d2` feature is enabled.
 
-* Added `DatabaseErrorKind::ReadOnlyTransaction` to allow applications to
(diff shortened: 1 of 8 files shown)


weiznich:
We avoid reformating large pieces of the code base just for the sake of reformating. The reason is that such changes hide a lot of context when certain things where introduced and for which reasons. Additionally the exact markdown format is a matter of taste. For example I personally prefer to use `*italic*` instead of `_italic_`. Given those points I'm not in favor with such a change.

henryboisdequin:
@weiznich Understood but it does make me paranoid seeing inconsistent code in the codebase.

weiznich:
@henryboisdequin While I understand the general idea behind this change, let me just write some words about the experience collected over the past years here:
* Just changing something to fit a certain style is not really meaningful, as it soon will be back to a new inconsistent state due to new changes introduced. The only way to prevent something like this from happening is having some automatic check as part of the CI. That's what we do for our rust code by running rustfmt on the CI.
* Stylistic changes should first have a policy how something should look like. This should be documented as part of the repository and it should be discussed before doing actual changes to the code base. Open Source work normally means seeking some kind of consent between all contributors, especially for things like code styles as they tend to affect all of them. The normal way of doing things like this is to write a proposal like "how about changing this or that for the following reasons" and see how others response to this. After there is a consent and a  it should be straight to actual apply these changes to the code base.

As additional node: Style changes normally mean large scale changes to the whole code base. As consequence it is normally required for others to fix their PR's to follow the new policy and to resolve merge conflicts. Because of this I consider such changes as quite high cost, that should have a clear justification.

PR #2660, opened 2021-03-02
Add jsonb merge

### LINKED ISSUE #2036: Feature request: jsonb merge operator for postgres
Postgres supports merging two jsonb objects with the `||` operator. I recently needed that and implemented it [here](https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/jsonb.rs) but haven't been able to turn this into a proper PR.

If it's easy to integrate in its current form, I'd love to see that. If not, I am perfectly fine with "please make a PR" as a response to this issue.

Either way, I also need Postgres key deletion operator (`data::jsonb - key::text`) but haven't been able to work out the types for that one. Anything that you could do to include that in Diesel would be hugely appreciated!

henryboisdequin:
Closes #2036

TODO:

- [ ] Fix tests

This PR:

- Adds jsonb merge
- Has a doc test

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -4,7 +4,7 @@ use super::operators::*;
 use crate::dsl;
 use crate::expression::grouped::Grouped;
 use crate::expression::{AsExpression, Expression, IntoSql, TypedExpressionType};
-use crate::sql_types::{Array, Cidr, Inet, Nullable, Range, SqlType, Text};
+use crate::sql_types::{Array, Cidr, Inet, Jsonb, Nullable, Range, SqlType, Text};
 
 /// PostgreSQL specific methods which are present on all expressions.
 pub trait PgExpressionMethods: Expression + Sized {
@@ -1111,3 +1111,18 @@ impl InetOrCidr for Inet {}
 impl InetOrCidr for Cidr {}
 impl InetOrCidr for Nullable<Inet> {}
 impl InetOrCidr for Nullable<Cidr> {}
+
+/// PostgreSQL specific expression methods using the `Jsonb` datatype.
+pub trait PgJsonbExpressionMethods: Expression<SqlType = Jsonb> + Sized {
+    fn merge<T: AsExpression<Jsonb>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression> {
+        Grouped(JsonbMerge::<Self, T::Expression>::new(
+            self,
+            other.as_expression(),
+        ))
+    }
+}
+
+impl<T: Expression<SqlType = Jsonb>> PgJsonbExpressionMethods for T where
+    T: Expression<SqlType = Jsonb>
+{
+}
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -1173,7 +1173,7 @@ pub trait PgJsonbExpressionMethods: Expression + Sized {
     /// #     })
     /// #     diesel::insert_into(d_e_f).values(d_e_f_data);
     /// #
-    /// #     let projected_merged_json = json!({
+    ///       let projected_merged_json = json!({
     ///               "a": 1,
     ///               "b": 2,
     ///               "c": 3,
@@ -1198,10 +1198,10 @@ pub trait PgJsonbExpressionMethods: Expression + Sized {
     /// # #[cfg(not(feature = "serde_json"))]
     /// # fn main() {}
     /// ```
-    fn merge<T: AsExpression<Self::SqlType>>(
-        self,
-        other: T,
-    ) -> dsl::JsonbMerge<Self, T::Expression> {
+    fn merge<T: AsExpression<Self::SqlType>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression>
+    where
+        <Self as Expression>::SqlType: SqlType,
+    {
         Grouped(JsonbMerge::<Self, T::Expression>::new(
             self,
             other.as_expression(),
@@ -1209,8 +1209,8 @@ pub trait PgJsonbExpressionMethods: Expression + Sized {
     }
 }
 
-// Support for nullable jsonb
-trait JsonbOrNullableJsonb {}
+/// Hack for adding support for nullable `jsonb`.
+pub trait JsonbOrNullableJsonb {}
 impl JsonbOrNullableJsonb for Jsonb {}
 impl JsonbOrNullableJsonb for Nullable<Jsonb> {}
 
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -1113,7 +1113,7 @@ impl InetOrCidr for Nullable<Inet> {}
 impl InetOrCidr for Nullable<Cidr> {}
 
 /// PostgreSQL specific expression methods using the `Jsonb` datatype.
-pub trait PgJsonbExpressionMethods: Expression<SqlType = Jsonb> + Sized {
+pub trait PgJsonbExpressionMethods: Expression + Sized {
     /// Creates a PostgreSQL `||` operator
     ///
     /// This operator merges two `Jsonb` objects.
@@ -1209,7 +1209,10 @@ pub trait PgJsonbExpressionMethods: Expression<SqlType = Jsonb> + Sized {
     /// # #[cfg(not(feature = "serde_json"))]
     /// # fn main() {}
     /// ```
-    fn merge<T: AsExpression<Jsonb>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression> {
+    fn merge<T: AsExpression<Self::SqlType>>(
+        self,
+        other: T,
+    ) -> dsl::JsonbMerge<Self, T::Expression> {
         Grouped(JsonbMerge::<Self, T::Expression>::new(
             self,
             other.as_expression(),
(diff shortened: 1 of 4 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
...
+    /// #     );
+    /// #
+    /// #     let merged_json = a_b_c_json.merge(d_e_f_json);
+    /// #     assert_eq!(projected_merged_json, merged_json);
+    /// #
+    /// # }
+    /// #
+    /// # #[cfg(not(feature = "serde_json"))]
+    /// # fn main() {}
+    /// ```
```
Ten0, line 1211: looks like this doctest is entirely commented, I suppose we want to have some part of it not hidden from the doc.

Ten0 (approved):

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
...
+    /// #             "e": 2,
+    /// #             "f": 3,
+    /// #             "letters": [
+    /// #                 "d",
+    /// #                 "e",
+    /// #                 "f",
+    /// #             ]
+    /// #         }
+    /// #     "#;
+    /// #     diesel::insert_into(d_e_f).values(from_str(d_e_f_data));
```
weiznich, line 1179: Use the [`json!`](https://docs.serde.rs/serde_json/macro.json.html) macro + batch inserts here. This should simplify the code quite a lot and fix the errors thrown by CI.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
...
+    ///                   "c",
+    ///                   "d",
+    ///                   "e",
+    ///                   "f",
+    ///               ]
+    ///           }
+    ///       "#
+    ///       );
+    ///
+    ///       let merged_json = a_b_c_json.merge(d_e_f_json);
```
weiznich, line 1204: This loads first both values and calls `.merge()` afterwards, where the merge method does not really exist. Instead we want to call the merge method at database side.
The corresponding query would look like this:

```rust
let merged_json = characters::table.select(characters::a_b_c.merge(characters::d_e_f)).load(&conn);
```

Also using the `json!` macro linked above should simplify constructing the expected value. Additionally I'm not sure if we should only show the expected value and the final query in our documentation. I think it's better to either show also the inserts (so that the user know what's in the database) or only show the query.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
...
+            self,
+            other.as_expression(),
+        ))
+    }
+}
+
+impl<T: Expression<SqlType = Jsonb>> PgJsonbExpressionMethods for T where
+    T: Expression<SqlType = Jsonb>
+{
+}
```
weiznich, line 1223: We likely want to support this also for `Nullable<Jsonb>`, therefore we must do a small trick here:

```rust
trait JsonbOrNullableJsonb {}
impl JsonbOrNullableJsonb for Jsonb {}
impl JsonbOrNullableJsonb for Nullable<Jsonb> {}

impl<T> PgJsonbExpressionMethods for T 
where 
    T: Expression,
    T::SqlType: JsonbOrNullableJsonb 
{}
```

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
...
+    ///
+    ///       let merged_json = a_b_c_json.merge(d_e_f_json);
+    ///       assert_eq!(projected_merged_json, merged_json);
+    /// #
+    /// # }
+    /// #
+    /// # #[cfg(not(feature = "serde_json"))]
+    /// # fn main() {}
+    /// ```
+    fn merge<T: AsExpression<Jsonb>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression> {
```
weiznich, line 1212: To allow nullable types here this must be changed to 
```suggestion
    fn merge<T: AsExpression<Self::SqlType>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression> {
```

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,113 @@ impl InetOrCidr for Inet {}
 impl InetOrCidr for Cidr {}
 impl InetOrCidr for Nullable<Inet> {}
 impl InetOrCidr for Nullable<Cidr> {}
+
+/// PostgreSQL specific expression methods using the `Jsonb` datatype.
+pub trait PgJsonbExpressionMethods: Expression<SqlType = Jsonb> + Sized {
```
weiznich, line 1116: To allow nullable types here we must lift that constraint a bit:
```suggestion
pub trait PgJsonbExpressionMethods: Expression + Sized {
```

weiznich (changes requested):
This already looks good implementation wise. I've added a few comments about also allowing to use this method on nullable jsonb expressions.
Beside of that the documentation needs some work. I've added suggestions there how to improve things and fix the CI errors.
If you need any help doing that just send me a message

henryboisdequin:
@weiznich Thanks for the comments, I have addressed them. How do I fix the CI?

weiznich:
@henryboisdequin Fixing the CI means to fix those compiler errors. That is normally done by trying to compile (and test the crate) locally and fix the errors there, as this gives much faster turn around times.
Now the CI gives two errors:

```
   Compiling diesel v2.0.0 (D:\a\diesel\diesel\diesel)
error[E0271]: type mismatch resolving `<<T as expression::AsExpression<<Self as expression::Expression>::SqlType>>::Expression as expression::Expression>::SqlType == types::sql_types::Jsonb`
Error:     --> diesel\src\pg\expression\expression_methods.rs:1201:5
     |
1201 | /     fn merge<T: AsExpression<Self::SqlType>>(
1202 | |         self,
1203 | |         other: T,
1204 | |     ) -> dsl::JsonbMerge<Self, T::Expression> {
1205 | |         Grouped(JsonbMerge::<Self, T::Expression>::new(self, other.as_expression()))
1206 | |     }
     | |_____^ expected associated type, found struct `types::sql_types::Jsonb`
     |
     = note: expected associated type `<Self as expression::Expression>::SqlType`
                         found struct `types::sql_types::Jsonb`
     = help: consider constraining the associated type `<Self as expression::Expression>::SqlType` to `types::sql_types::Jsonb` or calling a method that returns `<Self as expression::Expression>::SqlType`
     = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
     = note: required because of the requirements on the impl of `expression::AsExpression<types::sql_types::Jsonb>` for `<T as expression::AsExpression<<Self as expression::Expression>::SqlType>>::Expression`
```

Which is caused because of [this line](https://github.com/diesel-rs/diesel/pull/2660/files#diff-57d1a76b94fe1cf64983a084f54fc80081ddeabd55f3b0ee0856656f7bedf023R83), where we force the right side of the operation to `Jsonb`, which won't work for the nullable variant. This can be fixed by using `SqlTypeOf<Lhs>` there instead of `Jsonb`.

```
error[E0277]: the trait bound `<Self as expression::Expression>::SqlType: sql_types::SqlType` is not satisfied
Error:     --> diesel\src\pg\expression\expression_methods.rs:1201:17
     |
1201 |     fn merge<T: AsExpression<Self::SqlType>>(
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `sql_types::SqlType` is not implemented for `<Self as expression::Expression>::SqlType`
     | 
    ::: diesel\src\expression\mod.rs:195:11
     |
195  | pub trait AsExpression<T>
     |           ------------ required by a bound in this
196  | where
197  |     T: SqlType + TypedExpressionType,
     |        ------- required by this bound in `expression::AsExpression`
     |
help: consider further restricting the associated type
     |
1204 |     ) -> dsl::JsonbMerge<Self, T::Expression> where <Self as expression::Expression>::SqlType: sql_types::SqlType {
     |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Which simply can be fixed by doing what the compiler suggests.

henryboisdequin:
```
error[E0433]: failed to resolve: use of undeclared type or module `expression`
    --> diesel/src/pg/expression/expression_methods.rs:1203:18
     |
1203 |         <Self as expression::Expression>::SqlType: SqlType,
     |                  ^^^^^^^^^^ use of undeclared type or module `expression`
```

@weiznich Sorry for all of the inconvenience but where do we get `expression` from?

weiznich:
@henryboisdequin It's a module defined [here](https://github.com/diesel-rs/diesel/blob/d79455dd339657ab7bce2056fcab8ac184d2d62c/diesel/src/lib.rs#L154). You can import it via `crate::expression::Expression` (or just skip the module name as it is already in scope because of [this](https://github.com/diesel-rs/diesel/blob/d79455dd339657ab7bce2056fcab8ac184d2d62c/diesel/src/pg/expression/expression_methods.rs#L6) import)

henryboisdequin:
@weiznich Woah, some CI failure's are unrelated to this PR

weiznich:
@henryboisdequin Yes some are. That unfortunately happens from time to time. I normally just retrigger the build in such cases, but here are [other](https://github.com/diesel-rs/diesel/pull/2660/checks?check_run_id=2022824789#step:19:342) valid errors, so that's not needed in this case.
You should really try to fix them locally. To reproduce those errors run `cargo test --manifest-path "diesel/Cargo.toml" --features "postgres serde_json"` in your clone.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1111,3 +1111,114 @@ impl InetOrCidr for Inet {}
...
+    fn merge<T: AsExpression<Self::SqlType>>(self, other: T) -> dsl::JsonbMerge<Self, T::Expression>
+    where
+        <Self as Expression>::SqlType: SqlType,
+    {
+        Grouped(JsonbMerge::<Self, T::Expression>::new(
+            self,
+            other.as_expression(),
+        ))
+    }
+}
```
Ten0, line 1212: In order to allow handling cases where one is nullable and the other is not, we could do something similar to what is done with `And`:
https://github.com/diesel-rs/diesel/blob/9edd8acb51e0dcd44ccdbc7aae38c323005a3481/diesel/src/expression_methods/bool_expression_methods.rs#L41-L49
(That is, being generic on an additional ST and putting a bound on `JsonbMerge<Self, T::Expression>: Expression`.)

weiznich:
Closed by #3035

PR #2663, opened 2021-03-04
Fix a use-after-free bug in our Sqlite backend

### LINKED ISSUE #2662: sqlite's `LIKE` clause causes `DeserializationError`
## Setup

SQl, table and data.
```sql
DROP TABLE IF EXISTS tasks;

CREATE TABLE tasks  (
  uuid TEXT NOT NULL PRIMARY KEY,
  parent_task TEXT,
  name TEXT NOT NULL,
  status TEXT NOT NULL DEFAULT 0,
  description TEXT,
  CONSTRAINT fk_parent,
  FOREIGN KEY(parent_task)
      REFERENCES tasks(uuid)
)

INSERT INTO tasks (uuid, parent_task, name, status, description)
VALUES ('03eec7f5-3163-496b-8ab3-8cf97e7860a0', NULL, 'A_root', 'Pending', 'a root');
```

```rust
table! {
    tasks (uuid) {
        uuid -> Text,
        parent_task -> Nullable<Text>,
        name -> Text,
        status -> Text,
        description -> Nullable<Text>,
    }
}
```

### Versions

- **Rust:**
```
$rustc --version
rustc 1.50.0 (cb75ad5db 2021-02-10)
```

- **Diesel:**
```toml
 diesel = { version = "1.4.5", features = ["sqlite", "uuidv07", "uuid"] }
```
- **Database:**
```
$sqlite3 --version                                                                                        nix-shell
3.34.1 2021-01-20 14:10:07 10e20c0b43500cfb9bbc0eaa061c57514f715d87238f4d835880cd846b9ebd1f
```

- **Operating System**
```
$nixos-version --json                                                                                     nix-shell
{"nixosVersion":"21.03.20210209.29a15e2","nixpkgsRevision":"29a15e2c1f5ad64c235d78ba5e29a56fe520ad45"}
```

### I want to use the LIKE clause with the function `sql_query`.

```rust
let req = sql_query("SELECT *, 0 as depth FROM tasks WHERE uuid LIKE ?;")
  .bind::<Text, _>("43eec7f5%")
  .get_results(self.connection.as_ref().unwrap())
  .expect("res");;
```

The code above generates the error:
```
thread 'main' panicked at 'res: DeserializationError("Columnuuidwas not present in query")', src/store.rs:135:14
```

weiznich:
We've missused `sqlite3_column_name`. The
[SQLite](https://www.sqlite.org/c3ref/column_name.html) documentation
states that the following:

> The returned string pointer is valid until either the prepared statement
> is destroyed by sqlite3_finalize() or until the statement is automatically
> reprepared by the first call to sqlite3_step() for a particular
> run or until the next call to sqlite3_column_name()
> or sqlite3_column_name16() on the same column.

As part of our `query_by_name` infrastructure we've first received all
field names for the prepared statement and stored them as string slices
for later use. After that we called `sqlite3_step()` for the first time,
which invalids the pointer and therefore the stored string slice.
I've opted to fix this by just populating the field name map after the
first call to `sqlite3_step()` as a minimal fix for this issue for the
1.x release series. We need to investigate further if and how a similar
issue can occur with the current master branch. The corresponding code
changed quite a lot and at least this particular behaviour is not
possible anymore.

cc @diesel-rs/contributors and @diesel-rs/core: I've plan to release this change tomorrow. Additionally I see if we need to get a RUSTSEC advisory for this or not.

Fixes #2662

### DIFF
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -42,28 +42,35 @@ where
 
 pub struct NamedStatementIterator<'a, T> {
     stmt: StatementUse<'a>,
-    column_indices: HashMap<&'a str, usize>,
+    column_indices: Option<HashMap<&'a str, usize>>,
     _marker: PhantomData<T>,
 }
 
 impl<'a, T> NamedStatementIterator<'a, T> {
     #[allow(clippy::new_ret_no_self)]
     pub fn new(stmt: StatementUse<'a>) -> QueryResult<Self> {
-        let column_indices = (0..stmt.num_fields())
+        Ok(NamedStatementIterator {
+            stmt,
+            column_indices: None,
+            _marker: PhantomData,
+        })
+    }
+
+    fn populate_column_indices(&mut self) -> QueryResult<()> {
+        let column_indices = (0..self.stmt.num_fields())
             .filter_map(|i| {
-                stmt.field_name(i).map(|column| {
-                    let column = column
-                        .to_str()
+                dbg!(i);
+                dbg!(self.stmt.field_name(i)).map(|column| {
+                    let column = dbg!(column
+                        .to_str())
                         .map_err(|e| DeserializationError(e.into()))?;
                     Ok((column, i))
                 })
             })
             .collect::<QueryResult<_>>()?;
-        Ok(NamedStatementIterator {
-            stmt,
-            column_indices,
-            _marker: PhantomData,
-        })
+
+        self.column_indices = Some(column_indices);
+        Ok(())
     }
 }
 
@@ -78,8 +85,13 @@ where
             Ok(row) => row,
             Err(e) => return Some(Err(e)),
         };
+        if self.column_indices.is_none() {
+            if let Err(e) = self.populate_column_indices() {
+                return Some(Err(e));
+            }
+        }
         row.map(|row| {
-            let row = row.into_named(&self.column_indices);
+            let row = row.into_named(self.column_indices.as_ref().expect("it's there because we populated it above"));
             T::build(&row).map_err(DeserializationError)
         })
     }
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -95,7 +95,11 @@ where
             }
         }
         row.map(|row| {
-            let row = row.into_named(self.column_indices.as_ref().expect("it's there because we populated it above"));
+            let row = row.into_named(
+                self.column_indices
+                    .as_ref()
+                    .expect("it's there because we populated it above"),
+            );
             T::build(&row).map_err(DeserializationError)
         })
     }
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -42,6 +42,11 @@ where
 
 pub struct NamedStatementIterator<'a, T> {
     stmt: StatementUse<'a>,
+    // The actual lifetime of the stored column name is
+    // not really `'a`, but it's impossible to have a better
+    // fitting lifetime here.
+    // See the `Statement::field_name` method for details
+    // how long the underlying livetime is valid
     column_indices: Option<HashMap<&'a str, usize>>,
     _marker: PhantomData<T>,
 }
@@ -59,10 +64,9 @@ impl<'a, T> NamedStatementIterator<'a, T> {
     fn populate_column_indices(&mut self) -> QueryResult<()> {
         let column_indices = (0..self.stmt.num_fields())
             .filter_map(|i| {
-                dbg!(i);
-                dbg!(self.stmt.field_name(i)).map(|column| {
-                    let column = dbg!(column
-                        .to_str())
+                self.stmt.field_name(i).map(|column| {
+                    let column = column
+                        .to_str()
                         .map_err(|e| DeserializationError(e.into()))?;
                     Ok((column, i))
                 })
--- a/diesel/src/sqlite/connection/stmt.rs
+++ b/diesel/src/sqlite/connection/stmt.rs
@@ -67,7 +67,7 @@ impl Statement {
     /// > or sqlite3_column_name16() on the same column.
     /// 
     /// https://www.sqlite.org/c3ref/column_name.html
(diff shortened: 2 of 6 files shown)


diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -42,28 +42,35 @@ where
...
+        })
+    }
+
+    fn populate_column_indices(&mut self) -> QueryResult<()> {
+        let column_indices = (0..self.stmt.num_fields())
             .filter_map(|i| {
-                stmt.field_name(i).map(|column| {
-                    let column = column
-                        .to_str()
+                dbg!(i);
```
Ten0, line 62: dbg

diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -42,28 +42,35 @@ where
...
+                dbg!(self.stmt.field_name(i)).map(|column| {
```
Ten0, line 63: dbg

diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -42,28 +42,35 @@ where
...
+                    let column = dbg!(column
```
Ten0, line 64: dbg

diesel/src/sqlite/connection/stmt.rs
```diff
@@ -58,8 +58,15 @@ impl Statement {
...
     }
 
-    /// The lifetime of the returned CStr is shorter than self. This function
-    /// should be tied to a lifetime that ends before the next call to `reset`
+    /// The lifetime of the returned CStr is shorter than self.
+    ///
+    /// > The returned string pointer is valid until either the prepared statement
+    /// > is destroyed by sqlite3_finalize() or until the statement is automatically
+    /// > reprepared by the first call to sqlite3_step() for a particular
+    /// > run or until the next call to sqlite3_column_name()
```
Ten0, line 66: I don't understand what "for a particular run" means in this context. I'd be able to review more usefully if I understood precisely what this means :)

diesel/src/sqlite/connection/stmt.rs
```diff
@@ -58,8 +58,15 @@ impl Statement {
...
+    /// > or sqlite3_column_name16() on the same column.
+    /// 
+    /// https://www.sqlite.org/c3ref/column_name.html
     unsafe fn field_name<'a>(&self, idx: usize) -> Option<&'a CStr> {
```
Ten0, line 70: The doc states that "The lifetime of the returned `CStr` is shorter than self.", but it looks like this signature actually allows *extending* the lifetime.
```rust
field_name(&self, idx: usize) -> Option<&CStr>
```
or (if we want to be more explicit)
```rust
field_name<'a, 'b: 'a>(&'a self, idx: usize) -> Option<&'b CStr>
```
would solve part of the issue.

diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -78,8 +85,13 @@ where
...
             Err(e) => return Some(Err(e)),
         };
+        if self.column_indices.is_none() {
+            if let Err(e) = self.populate_column_indices() {
+                return Some(Err(e));
+            }
+        }
         row.map(|row| {
-            let row = row.into_named(&self.column_indices);
+            let row = row.into_named(self.column_indices.as_ref().expect("it's there because we populated it above"));
```
Ten0, line 94: So this seems to not break memory safety specifically because this call to `into_named` enforces downgrading the lifetime of `'a` in `HashMap<&'a str, usize>` to the lifetime of `self`.
Maybe we could explicit it there in order to contain why this works closer? (It looks that currently it wouldn't be hard to change the signature of `into_named`, see it compile and call it a day.)
```rust
impl<'a, T> Iterator for NamedStatementIterator<'a, T>
where
    T: QueryableByName<Sqlite>,
{
    type Item = QueryResult<T>;

    fn next<'s>(&'s mut self) -> Option<Self::Item> {
        let row = match self.stmt.step() {
            Ok(row) => row,
            Err(e) => return Some(Err(e)),
        };
        if self.column_indices.is_none() {
            if let Err(e) = self.populate_column_indices() {
                return Some(Err(e));
            }
        }
        // explicitly downgrade the lifetime to a correct one
        let column_indices: &'s HashMap<&'s str, usize> = self.column_indices.as_ref().expect("it's there because we populated it above");
        row.map(|row| {
            let row = row.into_named(column_indices);
            T::build(&row).map_err(DeserializationError)
        })
    }
}
```

diesel/src/sqlite/connection/stmt.rs
weiznich: That's the exact wording of the SQLite documentation for this function. I interpret this in such a way that the pointer for a given column name can change if you execute a prepared statement and after the executing call `sqlite3_step()` the first time for this newly executed statement. (`sqlite3_step()` is basically `Iterator::next()` to get a new row from the result set). So we should fine as long as we call `sqlite3_column_name()` after the first call to `sqlite3_step()` for a particular run.

diesel/src/sqlite/connection/statement_iterator.rs
weiznich: This is not possible, it results in the following error message:
```
error[E0597]: `self` does not live long enough
  --> diesel/src/sqlite/connection/statement_iterator.rs:94:34
   |
83 |     fn next<'s>(&'s mut self) -> Option<Self::Item> {
   |             -- lifetime `'s` defined here
...
93 |         row.map(|row| {
   |                 ----- value captured here
94 |             let indices: &'s _ = self.column_indices.as_ref().expect("it's there because we populated it above");
   |                          -----   ^^^^ borrowed value does not live long enough
   |                          |
   |                          type annotation requires that `self` is borrowed for `'s`
...
98 |     }
   |     - `self` dropped here while still borrowed
```

That written: I'm quite sure no one will actually touch this code again, as this is already rewritten on the master branch in such a way that this cannot happen there. (Though we should check if we could trigger there the multiple calls to `sqlite3_column_name` thing there.)

diesel/src/sqlite/connection/stmt.rs
Ten0: Hmm so if I understand correctly "for a particular run" means for us "for an instance of `StatementUse`".

weiznich: Thats correct :+1:

PR #2664, opened 2021-03-06
Use compact encoding for PgNumeric with neg weight

mwilliammyers:
This continues #1992, but adds support for using the compact encoding for `BigDecimals` with negative weights.

While not supporting the more compact encoding for numbers with negative weights only causes issues with (all versions of) cockroach, I felt like we should fully support this encoding since we switched to it in #1992.


henryboisdequin (approved):

TaKO8Ki (approved):

weiznich:
@henryboisdequin There is really no need to request a review from me personally here. If I don't respond to a PR in a few days that does mostly mean I do not have the capacity to do something about it now. Requesting a review in such cases only raises my frustration level, so please don't do that.

PR #2614, opened 2020-12-31
Fix broken links in doc

TaKO8Ki:
I found broken links in doc, so I've fixed them.

PR #2669, opened 2021-03-08
Link QueryDsl::select in group_by

pickfire:


weiznich:
Blocked on #2614 as the current minimal supported rust version does not support intra-doc links.

PR #2670, opened 2021-03-08
Mention the order of select and group_by

pickfire:
select can only be used after group_by, but not before it

```rust
let query = comments::table
    .group_by((comments::user_id, comments::post_id))
    .select((comments::user_id, comments::post_id, count(comments::id)));
```

I tried flipping `group_by` and `select` and found that it does not work (I expected this) but it wasn't mentioned in the docs.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -836,8 +836,8 @@ pub trait QueryDsl: Sized {
 
     /// Sets the `group by` clause of a query.
     ///
-    /// **Note:** Queries having a `group by` clause require a custom select clause.
-    /// Use `QueryDsl::select()` to specify one
+    /// **Note:** Queries having a `group by` clause require a custom select clause _after_ it.
+    /// Use `QueryDsl::select()` to specify one.
     ///
     /// If there was already a group by clause, it will be overridden.
     /// Ordering by multiple columns can be achieved by passing a tuple of those


henryboisdequin (approved):

diesel/src/query_dsl/mod.rs
```diff
@@ -836,8 +836,8 @@ pub trait QueryDsl: Sized {
 
     /// Sets the `group by` clause of a query.
     ///
-    /// **Note:** Queries having a `group by` clause require a custom select clause.
-    /// Use `QueryDsl::select()` to specify one
+    /// **Note:** Queries having a `group by` clause require a custom select clause _after_ it.
```
weiznich, line 839: Sorry for being pedantic, but that's just not true. The real think is a bit more complex than this simplified statement. 
Each time you call `.select()` it checks that the select clause is valid for the current `.group_by()` clause, so technically speaking it's possible to write queries that set the select clause before calling `.group_by()` as long as the select clause is valid for both, no group by clause and the group by clause you set afterwards.
The simplest example for such a query is the following:
```rust
some_table::table.select(1.into_sql::<Integer>()).group_by(some_table::name)
```
as all literal expressions are valid regardless for the group by clause.

pickfire: That's what I tried, I just swap the position and it does not work. Maybe you can take a look at what I was doing https://github.com/pickfire/diesel-group-by

pickfire: > Please stop to make conclusions about how the whole system works from a single example, especially if the author tells you that your assumptions are not correct.

I understand, but I wanted to try to understand more about it so I did a few more examples, does that mean every example that I do is incorrect?

>  For your example that means user_id, post_id, count(id) would be required to be valid for no group by clause. As this mixes aggregate and non aggregate expressions that's not the case. It is important to note that this is really depended on the expressions used as select clause so we cannot generally say that you need to call select after group_by.

How does the user know in what cases what behavior are there and in what cases what behavior are there? Do they need to experience themselves not being able to compile or we could put some note up front to tell them about their behavior? Maybe the behavior is just too complicated to be documented?

diesel/src/query_dsl/mod.rs
```diff
@@ -836,8 +836,8 @@ pub trait QueryDsl: Sized {
...
+    /// Use `QueryDsl::select()` to specify one.
```
pickfire, line 840:
```suggestion
    /// **Note:** Queries having a `group by` clause require a custom select clause.
    /// Use `QueryDsl::select()` to specify one. `select` needs to be before `group_by`
    /// if there are group by clause within select clause.
```

diesel/src/query_dsl/mod.rs
weiznich: > I understand, but I wanted to try to understand more about it so I did a few more examples, does that mean every example that I do is incorrect?

That's all fine. I only ask that you don't try to extrapolate from one example the whole behaviour of a complex generic function complex. I have the feeling that this happened here and that this also happened in #2631. I mean our documentation needs to cover all edge cases, not only that one that you personally hit. I personally have the opinion that it's better to have documentation not mentioning some things, than have documentation mentioning things as impossible that are in fact possible.

> How does the user know in what cases what behavior are there and in what cases what behavior are there? Do they need to experience themselves not being able to compile or we could put some note up front to tell them about their behavior? Maybe the behavior is just too complicated to be documented?

So what are your actual suggestions here? Do you want to see the API changed? Have you concrete proposals for this beside "the current way is bad"? In my opinion this is nothing that can "just" be fixed by adding some half true statement to some random location in our documentation.
The main point here is that SQL itself is complex, especially the interactions between different features. Now diesel chooses to provide a lot guarantees at compile time, that means most of this is already documented as part of the type system.

diesel/src/query_dsl/mod.rs
pickfire: Oops, I accidently typed the wrong thing.

```suggestion
    /// **Note:** Queries having a `group by` clause require a custom select clause.
    /// Use `QueryDsl::select()` to specify one. `select` needs to be after `group_by`
    /// if some fields depends on the fields within `group_by`.
```

`s/before/after` which is a typo, and the select needs to be after it if some fields is depending on the group by fields, is that correct and understandable?

weiznich: I do not feel that this is correct either. I mean a select clause does not consist of fields (`users.name`), but of expressions  (e.g. `avg(users.age) + users.id` to have something that does not make much sense). Also as noted above this is not really clear on when you require a select clause after `group_by` and when not. Unfortunately that's something that's quite complex to get really right in one sentence. Again as mentioned already several times: A select clause must always be valid for the current `group by` clause (for what this means look up the postgres docs on interaction between group by and select). So if a select comes before `group_by` it must be valid for no group by clause as well. I do not find that very surprising as this is only the type state equivalent of the normal builder pattern.

pickfire: I guess maybe this patch is not necessary, it isn't something that can be put into words.

PR #2672, opened 2021-03-09
Fix QueryableByName derive link in Untyped doc

pickfire:
image

This will probably fix it. Not tested.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -119,7 +119,6 @@ pub mod expression_types {
     /// using [`#[derive(QueryableByName)]`](../deserialize/derive.QueryableByName.html)
     /// on the corresponding result type.
     ///
-    /// [`QueryableByName`]: ../deserialize/derive.QueryableByName.html
     #[derive(Clone, Copy, Debug)]
     pub struct Untyped;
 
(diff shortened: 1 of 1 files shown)


diesel/src/expression/mod.rs
```diff
@@ -116,9 +116,9 @@ pub mod expression_types {
     /// we do not know which fields are returned from such a query at compile time.
     ///
     /// For loading values from queries returning a type of this expression, consider
-    /// using [`#[derive(QueryableByName)]`] on the corresponding result type.
+    /// using [`QueryableByName` derive] on the corresponding result type.
```
weiznich, line 119:
```suggestion
    /// using [`#[derive(QueryableByName)]`](../deserialize/derive.QueryableByName.html)
    /// on the corresponding result type.
```

diesel/src/expression/mod.rs
```diff
@@ -116,9 +116,9 @@ pub mod expression_types {
...
     ///
-    /// [`#[derive(QueryableByName)]`]: ../deserialize/derive.QueryableByName.html
+    /// [`QueryableByName`]: ../deserialize/derive.QueryableByName.html
```
weiznich, line 121:
```suggestion
```

weiznich (commented):
I would rather like to fix just the link instead of removing the `#[derive()]` syntax there. In that way we stay in line with other places in the documentation linking derives.

PR #2674, opened 2021-03-10
It's 2021

henryboisdequin:
Hasn't been updated in a while...

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -175,7 +175,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   Copyright 2015-2021 Sean Griffin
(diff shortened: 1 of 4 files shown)


LICENSE-APACHE
```diff
@@ -175,7 +175,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   Copyright 2015-2018 Sean Griffin
+   Copyright 2015-2021 Sean Griffin
```
weiznich, line 178: We should likely update that to something like
```suggestion
   Copyright 2015-2018 Sean Griffin, 2018-2021 Diesel Core Team 
```
or something like that. 
cc @diesel-rs/core

thecooldrop (approved):
This seems appropriate, and quite timely as well :smile:

pksunkara (approved):

PR #2676, opened 2021-03-12
Bug/fix potential use after free

weiznich:
This is basically #2663 but targeting the master branch

> The returned string pointer is valid until either the prepared statement
> is destroyed by sqlite3_finalize() or until the statement is automatically
> reprepared by the first call to sqlite3_step() for a particular
> run or until the next call to sqlite3_column_name()
> or sqlite3_column_name16() on the same column.

As concequence this means we cannot just call `sqlite3_column_name` in
`Field::field_name()`. To fix this issue I decided to get all column
names pointers once after calling `step()` the first time and cache them
as part of the specific Statement instance. This introduces a lot of
lifetimes to be very explicit about what is valid for which time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,6 +243,14 @@ Key points:
 
 [2-0-migration]: FIXME write a migration guide
 
+## [1.4.6] - 2021-03-05
+
+### Fixed
+
+* Fixed a use-after-free issue in the `QueryableByName` implementation
+  of our `Sqlite` backend
+* Updated several dependencies
+
 ## [1.4.5] - 2020-06-09
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


pksunkara (approved):

weiznich:
As expected this has an impact of performance
* Insert statements seem to take ~1.15 times as long as before 
* Query statements with only one result seem to take up to 1.3 times as long as before (this is kind of expected as we do a bit more work on the first call to `step()`
* `QueryableByName` get's a significant speedup for larger result sets, something like master is taking up to 1.7 times as long as the code in this PR.

The second and third change is kind of expected, while the first one is rather surprising.

PR #2677, opened 2021-03-12
Report an error with incorrect sql_type

pickfire:
@weiznich This is the one that I told you, the other method that I can think of is have an internal `impl FromSqlType<BigInt> for i64` which not only could be reused between different database but we could have a helper to like in the heapsize example to check if i64 implements BigInt. Is it feasible?

What do you think?

Before

```rust
error[E0277]: the trait bound `CommentCount: FromSqlRow<Untyped, Sqlite>` is not satisfied
  --> src/main.rs:48:9
   |
48 |     "#).get_results(&conn);
   |         ^^^^^^^^^^^ the trait `FromSqlRow<Untyped, Sqlite>` is not implemented for `CommentCount`
   |
   = note: required because of the requirements on the impl of `LoadQuery<SqliteConnection, CommentCount>` for `SqlQuery`

error: aborting due to previous error; 1 warning emitted

error: could not compile `hello`
```

After

```rust
error: i32 is not implemented for BigInt
 --> $DIR/ensures_derive_queryable_by_name_valid_type.rs:9:6
  |
9 |     id: i32,
  |         ^^^
  |
help: BigInt implements "i64"
 --> $DIR/ensures_derive_queryable_by_name_valid_type.rs:8:18
  |
8 |     #[sql_type = "BigInt"]
  |                  ^^^^^^^^
```


diesel_compile_tests/tests/fail/ensures_derive_queryable_by_name_valid_type.rs
```diff
@@ -0,0 +1,13 @@
+extern crate diesel;
+
+use diesel::*;
+use diesel::sql_types::*;
+
+#[derive(QueryableByName)]
+struct User {
+    #[sql_type = "BigInt"]
+	id: i32,
```
henryboisdequin, line 9:
```suggestion
struct User {
    #[sql_type = "BigInt"]
    id: i32,
```

diesel_compile_tests/tests/fail/ensures_derive_queryable_by_name_valid_type.rs
```diff
@@ -0,0 +1,13 @@
...
+}
+
+fn main() {
+}
```
henryboisdequin, line 13:
```suggestion
fn main() {}
```

diesel_derives/src/queryable_by_name.rs
```diff
@@ -118,3 +121,38 @@ fn sql_type(field: &Field, model: &Model) -> syn::Type {
         }
     }
 }
+
+fn check_sql_type(field_ty: &syn::Type, st: &syn::Type) {
+    // syn heapsize example trick won't work here because there is an additional database generic
+    // unless we create another trait without the DB part
+    //
+    //     quote_spanned! {field_ty.span()=>
+    //         function(st)
```
henryboisdequin, line 130:
```suggestion
    // syn heapsize example trick won't work here because there is an additional database generic
    // unless we create another trait without the DB part.
    //
    //     quote_spanned! { field_ty.span() =>
    //         function(st)
```

diesel_compile_tests/tests/trybuild.rs
```diff
@@ -3,5 +3,5 @@ extern crate trybuild;
 #[test]
 fn trybuild() {
     let t = trybuild::TestCases::new();
-    t.compile_fail("tests/fail/*.rs");
+    t.compile_fail("tests/fail/ensures_derive_queryable_by_name_valid_type.rs");
```
henryboisdequin, line 6:
```suggestion
    t.compile_fail("tests/fail/*.rs");
```

henryboisdequin (approved):
Just a few nits, other than that looks good

diesel_compile_tests/rust-toolchain
```diff
@@ -1 +0,0 @@
-nightly-2020-05-01
```
pksunkara, line 1: Why is this removed?

diesel_derives/src/diagnostic_shim.rs
```diff
@@ -80,6 +91,15 @@ impl Diagnostic {
     }
 }
 
+#[cfg(not(feature = "nightly"))]
+impl DiagnosticExt for Diagnostic {
+    fn span_help<T: Into<String>>(mut self, _span: Span, msg: T) -> Diagnostic {
```
henryboisdequin, line 96:
```suggestion
    fn span_help<T: Into<String>>(mut self, _span: Span, msg: T) -> Self {
```

weiznich:
While I can see the benefit of having something like this, I'm not in line with this particular implementation.
A potential implementation needs to fulfill at least the following requirements:
* There should be no hard errors for any valid code. This explicitly needs to include custom third party types.
* It needs to handle more that a single type combination
* It should not require duplicating all type combinations that implement `FromSql` in diesel itself as such a list quickly can gets out of date.
* If we choose to emit some kind of warning, then there should be a way to suppress this warning.

diesel_compile_tests/rust-toolchain
pickfire: I accidentally removed it.

diesel_derives/src/diagnostic_shim.rs
pickfire: Err, do we do this?

pickfire:
> There should be no hard errors for any valid code. This explicitly needs to include custom third party types.

It may have issue when user rename their types, which very very likely they won't, why would someone rename a primitive type?

> It needs to handle more that a single type combination

Can you please show me an example of a single type combination? I never seen that within a struct.

> It should not require duplicating all type combinations that implement FromSql in diesel itself as such a list quickly can gets out of date.

I haven't been able to figure this one out yet, both method seemed to require such changes, the other method can be reused for parts unlike this one.

> If we choose to emit some kind of warning, then there should be a way to suppress this warning.

Ah, this could be done for this method but not the other method. The other method is on type level so it probably couldn't show it as a warning. Now I see the pros in this method.

Both methods I can think of doesn't seemed to be able to fulfill of of these, maybe at most 3 since the duplicate thing, I can't figure that out with the way `FromSql` is working since the struct is not able to relate what database the users are using. If we can do that, that is a different story but it would be a breaking change. Do you have any other ideas?

weiznich:
> It may have issue when user rename their types, which very very likely they won't, why would someone rename a primitive type?

It will return a hard error as soon as a user passes in a custom enum type that implements `FromSql<BigInt, DB>`, which is not something that I consider as very very unlikely, but as valid use case.

> Can you please show me an example of a single type combination? I never seen that within a struct.

You code only handles `BigInt` + `i64` not any other supported type mapping in diesel. Why should we check only for this specific case and not for `Text` and `String` or `Integer` and `i32` or …

diesel_derives/src/diagnostic_shim.rs
henryboisdequin: This results in an error?

pickfire:
> It will return a hard error as soon as a user passes in a custom enum type that implements FromSql<BigInt, DB>, which is not something that I consider as very very unlikely, but as valid use case.

Ah, that is true. I think if they implements that then it would be a false positive which is incorrect. If one implemented their own date type they could hit the error.

> You code only handles BigInt + i64 not any other supported type mapping in diesel. Why should we check only for this specific case and not for Text and String or Integer and i32 or …

BigInt + i64 is just the proof of concept. I could include other stuff too.

weiznich:
> >   It will return a hard error as soon as a user passes in a custom enum type that implements FromSql<BigInt, DB>, which is not something that I consider as very very unlikely, but as valid use case.
> 
> Ah, that is true. I think if they implements that then it would be a false positive which is incorrect. If one implemented their own date type they could hit the error.

That's exactly my point here. Emitting a false positive in this case is not really acceptable, because this pattern is something we want to support. Additionally I do not see a good way to work that around at the level of derive code without explicitly listing type combination which should error, which is obviously something that does also not work well.

pickfire:
I get it now. Do you have other suggestion that we could improve the error on this part?

weiznich:
@pickfire The only way I'm aware of is to improve the generic code is such a way that rustc itself emits a better error message, but that's really no simple task.

pickfire:
@weiznich Is there any reason why we don't do `diesel-mysql` like what serde did, then we don't have to do `SqlType` and when the users pull in the dependencies, they only get to pull in the specific database, that way we can cross check if the fields in the struct can be used correctly since we don't have to have an extra `DB` generic, it may also help in making the code compiles faster I guess?

weiznich:
@pickfire As far as I'm aware there are no technical reasons to have all backends in one crate. This is just something that has involved historically. With the current design it's already possible to implement custom backends as third party crate. See [diesel-oci](https://github.com/GiGainfosystems/diesel-oci) for an example. I think it may even be meaningful to move the current backends into separate crates in long term, but that's quite a bit of work and at least for me other things are currently more important.
That all written: Even if every backend would be a separate crate removing the `DB` parameter would mean that we drop support for using more than one database at a time. That's something we explicitly support, even if that requires a bit more work than people expect. In the end the `DB` parameter is there to check for exactly those consistency issues you've mentioned above at compile time. We just reject code/sql that is not compatible with a certain database (at least as long as it is written using our dsl).

Issue #2679, opened 2021-03-12
Document more about the expected behavior of Connection::establish


keepsimple1:
https://github.com/diesel-rs/diesel/blob/c2366e8a0ba7d97c6824b6e0cd5705c9710e0cc1/diesel/src/connection/mod.rs#L41

The current doc does not mention what happens if the `database_url` does not exist. Should the method create a new database automatically? Or, is it implementation-dependent? It seems like SQLite3 driver does create a new database if not exists yet, mostly because of its C FFI. But there is no doc for SQLite3 `establish` implementation: https://github.com/diesel-rs/diesel/blob/c2366e8a0ba7d97c6824b6e0cd5705c9710e0cc1/diesel/src/sqlite/connection/mod.rs#L57 .

It will be helpful to spell out what is the expected behavior in this case.

weiznich:
We are happy to receive a PR for this. The behaviour is implementation specific, so there is no way to document this in a general way.

PR #2682, opened 2021-03-14
add doc comments

keepsimple1:
Fix https://github.com/diesel-rs/diesel/issues/2679.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -54,10 +54,12 @@ impl Connection for SqliteConnection {
     type Backend = Sqlite;
     type TransactionManager = AnsiTransactionManager;
 
-    /// Establish a connection to the `database_url`.
+    /// Establish a connection to the database specified by `database_url`.
     ///
-    /// If the `database_url` does not exist, this method will
-    /// try to create a new file and then establish a connection to it.
+    /// See [SqliteConnection] for supported `database_url`.
+    ///
+    /// If the database does not exist, this method will try to
+    /// create a new database and then establish a connection to it.
     fn establish(database_url: &str) -> ConnectionResult<Self> {
         use crate::result::ConnectionError::CouldntSetupConfiguration;
 
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -54,9 +54,9 @@ impl Connection for SqliteConnection {
     type Backend = Sqlite;
     type TransactionManager = AnsiTransactionManager;
 
-    /// Establish a connection to `database_url`.
+    /// Establish a connection to the `database_url`.
     ///
-    /// If the file `database_url` does not exist, this method will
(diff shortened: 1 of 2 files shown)


pksunkara (approved):

diesel/src/sqlite/connection/mod.rs
```diff
@@ -54,6 +54,10 @@ impl Connection for SqliteConnection {
     type Backend = Sqlite;
     type TransactionManager = AnsiTransactionManager;
 
+    /// Establish a connection to `database_url`.
```
henryboisdequin, line 57:
```suggestion
    /// Establish a connection to the `database_url`.
```

diesel/src/sqlite/connection/mod.rs
```diff
@@ -54,6 +54,10 @@ impl Connection for SqliteConnection {
...
+    ///
+    /// If the file `database_url` does not exist, this method will
```
henryboisdequin, line 59: Do I not understand this or isn't the `database_url` not a file and in the `establish` function aren't we trying to establish a connection to the `database_url` and that the `establish` function has nothing to do with files?

keepsimple1: SQLite3 databases are files,  its database_url are `file://` . I can remove the word `file` as I don't want to distract readers from the main point of this doc comment.

henryboisdequin: @keepsimple1 Yes, that would be great. Thanks for the clarification!

diesel/src/sqlite/connection/mod.rs
weiznich:
```suggestion
    /// Establish a connection to the database specified by `database_url`.
    ///
    /// Supported connection URLs are:
    ///
    /// - File paths (`test.db`)
    /// - [URIs](https://sqlite.org/uri.html) (`file://test.db`)
    /// - Special identifiers (`:memory:`)
    ////
```

(Just including the information from the type itself here as well)

diesel/src/sqlite/connection/mod.rs
```diff
@@ -54,6 +54,10 @@ impl Connection for SqliteConnection {
...
+    /// try to create a new file and then establish a connection to it.
```
weiznich, line 60:
```suggestion
    /// If the database does not exist, this method will
    /// try to create a database and then establish a connection to it.
```

As sqlite does not only allow file based databases (but also in memory databases) be a bit more generic here about what is created and how it work.

weiznich (commented):

I've left a few small suggestions on how the wording could be improved a bit to include some more information.

diesel/src/sqlite/connection/mod.rs
keepsimple1: Thanks for your suggestions. I've updated the diff, and to avoid repeating the same info, I've included a link to `SqliteConnection` https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/connection/mod.rs#L29-L36

PR #2695, opened 2021-04-01
Add some `#[doc(alias = "…")] attributes to commonly missed functions

weiznich:


JohnTitor (approved):
We need some tweaks to CI but otherwise 👍

PR #2705, opened 2021-04-07
examples/mysql/all_about_inserts/src/lib.rs:  Add an extra space to a few asserts; this gets three tests passing that weren't previously

mcronce:


weiznich:
Thanks for the fix :+1: 
(I'm not sure why this was not catched by our ci before 🙈 )

PR #2724, opened 2021-04-13
Adding features reporting to diesel_cli

alexxroche:
As outlined in https://github.com/diesel-rs/diesel/discussions/2717


diesel_cli/src/cli.rs
```diff
@@ -121,6 +121,10 @@ pub fn build_cli() -> App<'static, 'static> {
                         .help("The format of the migration to be generated."),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("features")
+                .about("List which features have been compiled into this version of diesel_cli"),
+        )
```
pksunkara, line 127: This change is meaningless. You already added the subcommand below.

diesel_cli/src/cli.rs
```diff
@@ -162,6 +166,16 @@ pub fn build_cli() -> App<'static, 'static> {
                 .possible_values(&Shell::variants()),
         );
 
+    let features_subcommand = SubCommand::with_name("features")
+        .about("list support for which database engines have been compiled into this version of diesel")
```
pksunkara, line 170: Needs better phrasing of the description.

diesel_cli/src/cli.rs
```diff
@@ -162,6 +166,16 @@ pub fn build_cli() -> App<'static, 'static> {
...
+        .arg(
+            Arg::with_name("features")
+                .long("features")
+                .short("f")
+                .takes_value(false)
+                .help("Which engines supported."),
```
pksunkara, line 176: Needs better phrasing of the description.

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,31 @@ fn regenerate_schema_if_file_specified(
     Ok(())
 }
 
+// List which features have been compiled into this version of diesel_cli
+#[cfg(any(feature = "postgres", feature = "mysql", feature = "sqlite"))]
```
pksunkara, line 611: Don't need this line.

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,31 @@ fn regenerate_schema_if_file_specified(
...
+    features.push("postgres");
+    #[cfg(feature = "mysql")]
+    features.push("mysql");
+    #[cfg(feature = "sqlite")]
+    features.push("sqlite");
+
+    for f in &features {
+        match f as &str {
+            "postgres" | "sqlite" | "mysql" | "sqlite-bundled" => print!("{} ", f),
+            other => print!("'{}' ", other),
```
pksunkara, line 624: Not sure why we need this.

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,31 @@ fn regenerate_schema_if_file_specified(
...
+    #[cfg(feature = "postgres")]
+    features.push("postgres");
+    #[cfg(feature = "mysql")]
+    features.push("mysql");
+    #[cfg(feature = "sqlite")]
+    features.push("sqlite");
+
+    for f in &features {
+        match f as &str {
+            "postgres" | "sqlite" | "mysql" | "sqlite-bundled" => print!("{} ", f),
```
pksunkara, line 623: Where is `sqlite-bundled` coming from?

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,31 @@ fn regenerate_schema_if_file_specified(
...
+            other => print!("'{}' ", other),
+        };
+    }
+    if features.is_empty() {
+        println!("[warning] No features detected");
+    } else {
+        println!();
```
pksunkara, line 630: Why do we need this `match` statement? We can just do `println!("Features: {}", features.join(", "))`

pksunkara (changes requested):

diesel_cli/src/cli.rs
```diff
@@ -162,6 +166,16 @@ pub fn build_cli() -> App<'static, 'static> {
                 .possible_values(&Shell::variants()),
         );
 
+    let features_subcommand = SubCommand::with_name("features")
+        .about("list support for which database engines have been compiled into this version of diesel")
+        .arg(
+            Arg::with_name("features")
```
pksunkara, line 172: I don't understand why you have an arg here. Do you want `features` as subcommand or arg?

diesel_cli/src/main.rs
alexxroche: None of the features have spaces, so the output can be parsed as space delimited by other programs that depend on the output. By adding "Features: " that just increases the complexity of processing the output. I can see that adding "Features: " provides clarity, but in practical use the command run is `diesel features` which should lead the user to expect that the output would be features.

Also by having this part rather than the nifty .join(", ") means that adding a standard `-n     do not output the trailing newline` in the future becomes trivial.

diesel_cli/src/cli.rs
alexxroche: I agree. Any suggestions?

diesel_cli/src/cli.rs
alexxroche: I will remove this.

diesel_cli/src/cli.rs
alexxroche: I was trying to get `diesel -f` to work but realised that is probably not desirable. I will remove this part.

diesel_cli/src/main.rs
alexxroche: https://github.com/diesel-rs/diesel/blob/master/diesel_cli/README.md#installation
Line 33 `cargo install diesel_cli --no-default-features --features "sqlite-bundled"`

diesel_cli/src/main.rs
alexxroche: In the spirit for rust's match I though that this would provide completeness while also providing a method of highlighting any future features that were implemented and enabled.

diesel_cli/src/main.rs
alexxroche: I will remove `#[cfg(any(feature = "postgres", feature = "mysql", feature = "sqlite"))]`

alexxroche (commented):
Helpful review from pksunkara.

diesel_cli/src/main.rs
pksunkara: Yeah, but it's meaningless here, isn't it? Since you are not pushing it into `features`.

diesel_cli/src/main.rs
pksunkara: Just because there's `match` in rust doesn't mean we have to use it. What does this subcommand need to do? Print all the features that were pushed into the `features` vec. Why do we care about `other`? It's not someone else giving us this `features` value. It's us who are building it.

diesel_cli/src/main.rs
pksunkara: Okay, but we can still do `println!("{}", features.join(" "))`. If you want newline you can manipulate this string inside `join`.

diesel_cli/src/cli.rs
```diff
@@ -162,6 +162,10 @@ pub fn build_cli() -> App<'static, 'static> {
                 .possible_values(&Shell::variants()),
         );
 
+    let features_subcommand = SubCommand::with_name("features").about(
+        "list support for which database engines have been compiled into this version of diesel"
```
henryboisdequin, line 166:
```suggestion
        "lists all the database engines available with the version of diesel the user is using"
```

Maybe better?

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,30 @@ fn regenerate_schema_if_file_specified(
...
+    #[cfg(feature = "postgres")]
+    features.push("postgres");
+    #[cfg(feature = "mysql")]
+    features.push("mysql");
+    #[cfg(feature = "sqlite")]
+    features.push("sqlite");
+
+    for f in &features {
+        match f as &str {
+            "postgres" | "sqlite" | "mysql" | "sqlite-bundled" => print!("{} ", f),
```
henryboisdequin, line 622:
```suggestion
            "postgres" | "sqlite" | "mysql" => print!("{} ", f),
```

Fix for @pksunkara's comment

diesel_cli/src/main.rs
```diff
@@ -606,6 +607,30 @@ fn regenerate_schema_if_file_specified(
...
+            other => print!("'{}' ", other),
```
henryboisdequin, line 623:
```suggestion
            _ => {}
```

Fix for @pksunkara's comment

henryboisdequin (changes requested):

diesel_cli/src/main.rs
alexxroche: Applied.

diesel_cli/src/main.rs
pksunkara: Please remove the match and do this and it will be good for merging.

diesel_cli/src/cli.rs
```diff
@@ -162,6 +162,10 @@ pub fn build_cli() -> App<'static, 'static> {
                 .possible_values(&Shell::variants()),
         );
 
+    let features_subcommand = SubCommand::with_name("features").about(
+        "lists all the database engines available with this version of diesel"
```
pksunkara, line 166:
```suggestion
        "Lists all the database engines available with this version of diesel"
```

pksunkara (changes requested):
Also, please squash the commits.

diesel_cli/src/cli.rs
alexxroche: I agree. Thank you for the suggestion. Your sentence is much easier to parse and sounds more vernacular. I've tweaked it a little to make it shorter.

diesel_cli/src/main.rs
alexxroche: Changed as written by henryboisdequin

Ten0:
> Also, please squash the commits.

We could merge as "Squash and merge".

pksunkara:
Yeah, that's not the main blocker but [this comment chain](https://github.com/diesel-rs/diesel/pull/2724#discussion_r612709593) is.

pksunkara (approved):
Looks good. But I am wondering if a separate subcommand is needed. Wouldnt having it in version info be enough?

PR #2728, opened 2021-04-15
Parse error from sqlite exec instead of returning `DatabaseErrorKind::Unknown`

m-rots:
Currently the `exec` function on SQLite only returns `DatabaseErrorKind::Unknown` and does not try to parse this error as seen in the `last_error` function from the [`diesel::sqlite::connection::stmt`](https://github.com/m-rots/diesel/blob/177290d82aa5bb5d7376389e7de5c886986ed11d/diesel/src/sqlite/connection/stmt.rs#L75) module.

With this pull request I simply re-use `ensure_sqlite_ok` from the `stmt` module to retrieve the extended result code. Retrieving the extended result code is needed as [SQLite only returns the primary result code for historic compatibility](https://www.sqlite.org/rescode.html#primary_result_codes_versus_extended_result_codes).

### Background
I stumbled upon the `DatabaseErrorKind::Unknown` while using `PRAGMA foreign_keys = ON` on the connection together with `DEFERRABLE INITIALLY DEFERRED` on a foreign key in the schema. Using this branch locally solved the issue for me, as Diesel now correctly returns `DatabaseErrorKind::ForeignKeyViolation`.

### Famous last words
- As I do not know your preferences regarding scoping, I simply added `pub(super)` to the `ensure_sqlite_ok` function.
- I'm unfamiliar with unsafe in Rust. Therefore, I do not know whether the `err_msg` in the exec function is freed when just using `ptr::null_mut()`.

### DIFF
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -1,6 +1,6 @@
 extern crate libsqlite3_sys as ffi;
 
-use std::ffi::{CStr, CString, NulError};
+use std::ffi::{CString, NulError};
 use std::io::{stderr, Write};
 use std::os::raw as libc;
 use std::ptr::NonNull;
@@ -8,6 +8,7 @@ use std::{mem, ptr, slice, str};
 
 use super::functions::{build_sql_function_args, process_sql_function_result};
 use super::serialized_value::SerializedValue;
+use super::stmt::ensure_sqlite_ok;
 use super::{Sqlite, SqliteAggregateFunction};
 use crate::deserialize::FromSqlRow;
 use crate::result::Error::DatabaseError;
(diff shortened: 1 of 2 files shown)


TaKO8Ki (approved):

diesel/src/sqlite/connection/raw.rs
```diff
@@ -62,27 +63,21 @@ impl RawConnection {
...
 
-        if err_msg.is_null() {
-            Ok(())
-        } else {
-            let msg = convert_to_string_and_free(err_msg);
-            let error_kind = DatabaseErrorKind::Unknown;
-            Err(DatabaseError(error_kind, Box::new(msg)))
-        }
+        ensure_sqlite_ok(result, self.internal_connection.as_ptr())?;
+        Ok(())
```
weiznich, line 80:
```suggestion
        ensure_sqlite_ok(result, self.internal_connection.as_ptr())
```

weiznich (approved):
Looks good, beside that minor simplification.

diesel/src/sqlite/connection/raw.rs
m-rots: Thanks for the suggestion! I changed it around :)

PR #2729, opened 2021-04-16
Backport Postgres custom type OID caching to 1.4.x

garywang:


weiznich:
Thanks for opening this PR. We don't backport features to previous releases. That's something something we only do for bug fixes.

PR #2732, opened 2021-04-16
allow paths in #[table_name]

lukas-code:
This pull request allows any path as the argument for the `#[table_name]` attribute instead of just identifiers.
This way the table module no longer needs to be imported to derive traits like `Insertable`.
```rust
mod schema {
    diesel::table! {
        users {
            id -> Integer,
            name -> Text,
        }
    }
}

use diesel::Insertable;

#[derive(Insertable)]
#[table_name = "schema::users"]
pub struct User {
    pub id: i32,
    pub name: String,
}
```

### DIFF
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -76,7 +76,7 @@ use diagnostic_shim::*;
 /// `Option::<T>::None` is just skipped. To insert a `NULL` using default
 /// behavior use `Option::<Option<T>>::Some(None)`
 /// * `#[table_name = "path::to::table"]`, specifies a path to the table for which the
-/// current type is a changeset.
+/// current type is a changeset. The path is relative to the current module.
 /// If this attribute is not used, the type name converted to
 /// `snake_case` with an added `s` is used as table name.
 ///
@@ -147,7 +147,7 @@ pub fn derive_as_expression(input: TokenStream) -> TokenStream {
 /// # Optional container attributes
 ///
 /// * `#[table_name = "path::to::table"]` specifies a path to the table this
-///    type belongs to.
+///    type belongs to. The path is relative to the current module.
 ///    If this attribute is not used, the type name converted to
 ///    `snake_case` with an added `s` is used as table name.
 ///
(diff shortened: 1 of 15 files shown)


henryboisdequin (approved):
LGTM nice implementation

Might conflict with #2626

diesel_derives/src/lib.rs
```diff
@@ -78,7 +76,7 @@ use diagnostic_shim::*;
 /// `Option::<T>::None` is just skipped. To insert a `NULL` using default
 /// behavior use `Option::<Option<T>>::Some(None)`
 /// * `#[table_name = "some_table"]`, specifies the table for which the
-/// current type is a changeset. Requires that `some_table` is in scope.
```
weiznich, line 81: This remains somewhat true in my opinion. For `#[table_name = "some::releative::path"]` it is required that `some` is in scope. 
That written I'm not sure on the actual wording here.

weiznich (changes requested):
Generally I'm in favor with this change :+1:.
The actual implementation should: 
* Add a few more test cases for the other derives using `#[table_name]` (like `QueryableByName`)
* Add a changelog entry
* Use a better wording in the derive documentation to make clear that `#[table_name]` expects a path and outline what this places as requirement onto the caller.

diesel_derives/src/lib.rs
```diff
@@ -77,10 +75,10 @@ use diagnostic_shim::*;
 /// the derive should threat `None` values as `NULL`. By default
 /// `Option::<T>::None` is just skipped. To insert a `NULL` using default
 /// behavior use `Option::<Option<T>>::Some(None)`
-/// * `#[table_name = "some_table"]`, specifies the table for which the
-/// current type is a changeset. Requires that `some_table` is in scope.
+/// * `#[table_name = "path::to::table"]`, specifies a path to the table for which the
+/// current type is a changeset.
```
lukas-code, line 79: Should I add "Requires that `path` is in scope." here? It seems kinda obvious to me that the path is relative to the current scope.

weiznich: Stating the obvious here cannot hurt, therefore I believe adding something like: "The path is relative to the current module" is a good idea.

lukas-code: done

weiznich (approved):

PR #2736, opened 2021-04-19
Updated ipnetwork dependency to 0.18.0

Razican:
The `ipnetwork` crate just released a new version (0.18.0). By looking at the changes, there shouldn't be anything that is not backwards compatible, so this adds the new version to the list of supported versions by `diesel`.


henryboisdequin (approved):
Yep, looks good

Issue #2513, opened 2020-09-22
SQlite database 'is' and 'is not' operators are missing


zdenek-crha:
### Versions

- **Rust:**  1.40
- **Diesel:** 1.4.5
- **Database:** SQLite
- **Operating System** GNU/Linux (Debian)

### Feature Flags

- **diesel:**  [ "sqlite" ]

## Problem Description

The sqlite operators `is` and `is not` are not available in expression dsl.

### What are you trying to accomplish?

I have a database with two nullable columns which both hold uuid (last uuid seen on remote and last uuid archived localy). To see which rows in the table are elegible for archiving, I have to select all rows where `last_seen` and `last_archived` are not equal. Ie either one of them is `NULL` or both are set and their values differ.

There are rows with all combinations of those two columns in database. 

Using  `filter(dsl::last_seen.ne(dsl::last_archived))` which generates SQL query using `!=` operator and **returns empty result set**. I've run the SQL directly on  database directly and the behaviour is exactly same. 

I have been pointed to `is` and `is not` sqlite operators that allow comparison with null (https://sqlite.org/lang_expr.html, section 2) and when I replace `!=` with `is not` the query returns expected results.

Unfortunately, `diesel` does not support those operators for `sqlite` database.

weiznich:
As a official supported workaround you can just use [`diesel_infix_operator!`](http://docs.diesel.rs/1.4.x/diesel/macro.diesel_infix_operator.html) to define your own version of this operators. 
Given that workaround this issue is for me not high priority, which means I would happily accept a PR adding support for this features as long as it is well written, but I will not work on support for this myself.

zdenek-crha:
@weiznich Thanks for pointing me to the `diesel_infix_operator!`, I will give it a try.

Voodlaz:
@weiznich I want to work on this

weiznich:
@Voodlaz Sure feel free to submit a PR

hi-rustin:
I am working on this.

hi-rustin:
@weiznich Can you provide some help please? I'm not quite sure how to support these two operators, is there an example I can refer to?

weiznich:
@hi-rustin Those two operators are very similar to `IS DISTINCT FROM` and `IS NOT DISTINCT FROM` for the postgresql backend. This means you can use the implementation of those operators as general orientation which parts are required:

[...]

PR #2738, opened 2021-04-21
Added support for SQLite's `IS` and `IS NOT`

hi-rustin:
close https://github.com/diesel-rs/diesel/issues/2513

### DIFF
--- a/diesel/src/sqlite/expression/expression_methods.rs
+++ b/diesel/src/sqlite/expression/expression_methods.rs
@@ -1,0 +1,72 @@
+//! Sqlite specific expression methods.
+
+use super::operators::*;
+use crate::dsl;
+use crate::expression::grouped::Grouped;
+use crate::expression::{AsExpression, Expression};
+use crate::sql_types::SqlType;
+
+/// Sqlite specific methods which are present on all expressions.
+pub trait SqliteExpressionMethods: Expression + Sized {
+    /// Creates a Sqlite `IS` expression.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use schema::animals::dsl::*;
+    /// #     let connection = establish_connection();
+    /// let jack_is_a_dog = animals
+    ///     .select(name)
+    ///     .filter(species.is("dog"))
+    ///     .get_results::<Option<String>>(&connection)?;
+    /// assert_eq!(vec![Some("Jack".to_string())], jack_is_a_dog);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn is<T>(self, other: T) -> dsl::Is<Self, T>
+    where
+        Self::SqlType: SqlType,
+        T: AsExpression<Self::SqlType>,
+    {
+        Grouped(Is::new(self, other.as_expression()))
+    }
+
+    /// Creates a Sqlite `IS NOT` expression.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use schema::animals::dsl::*;
+    /// #     let connection = establish_connection();
+    /// let jack_is_not_a_spider = animals
+    ///     .select(name)
+    ///     .filter(species.is_not("spider"))
+    ///     .get_results::<Option<String>>(&connection)?;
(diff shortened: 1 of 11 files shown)


hi-rustin:
@weiznich Chocolatey's service seems to be broken.

henryboisdequin (approved):
Implementation looks good.

diesel/src/sqlite/expression/operators.rs
```diff
@@ -0,0 +1,5 @@
+use crate::sql_types::Bool;
+use crate::sqlite::Sqlite;
+
+__diesel_infix_operator!(Is, " IS ", ConstantNullability Bool, backend: Sqlite);
```
pksunkara, line 4: What happens if you compare a null value to a text value?

hi-rustin: This behaves identically to the `=` operator, except that `NULL` is treated as a normal value.

I will add the document.

hi-rustin: @pksunkara Added.

hi-rustin: The IS and IS NOT operators work like = and != except when one or both of the operands are NULL. In this case, if both operands are NULL, then the IS operator evaluates to 1 (true) and the IS NOT operator evaluates to 0 (false). If one operand is NULL and the other is not, then the IS operator evaluates to 0 (false) and the IS NOT operator is 1 (true). It is not possible for an IS or IS NOT expression to evaluate to NULL.

hi-rustin: The IS and IS NOT operators work like = and != except when one or both of the operands are NULL. In this case, if both operands are NULL, then the IS operator evaluates to 1 (true) and the IS NOT operator evaluates to 0 (false). If one operand is NULL and the other is not, then the IS operator evaluates to 0 (false) and the IS NOT operator is 1 (true). It is not possible for an IS or IS NOT expression to evaluate to NULL.

weiznich:
The [compile-test CI errors](https://github.com/diesel-rs/diesel/pull/2738/checks) seems to be valid and should be fixed before merging.
As a side node: We should probably report this as "noisy" diagnostic bug to the trybuild crate.

diesel/src/sqlite/expression/operators.rs
weiznich: I would prefer to include something more in line with this description to the documentation, so that it explicitly states the handling of `NULL` values.

hi-rustin: @weiznich Added.

diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -10,8 +10,10 @@ use crate::sql_types::SqlType;
 pub trait SqliteExpressionMethods: Expression + Sized {
     /// Creates a Sqlite `IS` expression.
     ///
-    /// This behaves identically to the `=` operator, except that `NULL` is
-    /// treated as a normal value.
+    /// The `IS` operator work like = except when one or both of the operands are NULL.
+    /// In this case, if both operands are NULL, then the `IS` operator evaluates to 1 (true).
+    /// If one operand is NULL and the other is not, then the `IS` operator evaluates to 0 (false).
```
weiznich, line 15: As diesel handles the `0`-> `false`, `1` -> `true` mapping here I would prefer to just use the boolean values here:
```suggestion
    /// In this case, if both operands are NULL, then the `IS` operator evaluates to true.
    /// If one operand is NULL and the other is not, then the `IS` operator evaluates to false.
```

diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -43,8 +45,10 @@ pub trait SqliteExpressionMethods: Expression + Sized {
 
     /// Creates a Sqlite `IS NOT` expression.
     ///
-    /// This behaves identically to the `!=` operator, except that `NULL` is
-    /// treated as a normal value.
+    /// The `IS NOT` operator work like != except when one or both of the operands are NULL.
+    /// In this case, if both operands are NULL, then the `IS NOT` operator evaluates to 0 (false).
+    /// If one operand is NULL and the other is not, then the `IS NOT` operator is 1 (true).
```
weiznich, line 50: As diesel handles the `0`-> `false`, `1` -> `true` mapping here I would prefer to just use the boolean values here:
```suggestion
    /// In this case, if both operands are NULL, then the `IS NOT` operator evaluates to false.
    /// If one operand is NULL and the other is not, then the `IS NOT` operator is true.
```

weiznich (commented):
One last small improvement for the documentation.

weiznich:
@hi-rustin You need to fix the compile_tests failure. To do that you need to run `TRYBUILD=overwrite cargo test` in the diesel_compile_tests crate.

hi-rustin:
> @hi-rustin You need to fix the compile_tests failure. To do that you need to run `TRYBUILD=overwrite cargo test` in the diesel_compile_tests crate.

Fixed. TThanks for the heads up, I missed your last reply.

pksunkara (approved):

PR #2590, opened 2020-12-08
[WIP] Add support for `HAVING` clauses

dnamsons:
Resolves #2581

### TO DO 
- [ ] Add tests

Issue #2581, opened 2020-11-27
Add support for `HAVING` clauses


weiznich:
With #2571 merged diesel now officially supports `GROUP BY` causes. As mentioned in #210 we have no concrete planes for supporting `HAVING` clauses in the next release. I've opening this issue to just have one for `HAVING` clauses + write down some notes for potential contributors. So if anyone is interested in adding support for such clauses feel free to open a PR.

Steps to add support for `HAVING` clauses:
* Define a `HavingDsl` trait similar to the existing [`FilterDsl`](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_dsl/filter_dsl.rs#L12) trait
* Add a generic impl for all tables, similar to [this impl](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_dsl/filter_dsl.rs#L20) for `FilterDsl`.
* Add support for `HAVING` clauses to our query builder:
    + Add a `NoHavingClause` similar to [`NoWhereClause`](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_builder/where_clause.rs#L31-L37)
    + Add a `HavingClause<Expr>` similar to [`WhereClause<Expr>`](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_builder/where_clause.rs#L71-L83)
    + Add another generic field to [`SelectStatement`](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_builder/select_statement/mod.rs#L40-L58) containing the having clause. Default that generic parameter to `NoHavingClause`. Add this generic parameter to all impls in that file.
    + Make all of the impls in [`diesel/src/query_builder/select_statement/dsl_impl.rs](https://github.com/diesel-rs/diesel/blob/6cd920184782745a46ee77d29ce20215228b9ed6/diesel/src/query_builder/select_statement/dsl_impls.rs) generic over the new generic paramater.
    + Fix all compilation errors coming from the newly added field by forwarding the having clause to the newly constructed select statement structs.
    + Implement `HavingDsl` for `SelectStatement`
[...]

PR #2740, opened 2021-04-23
Add support for HAVING clauses

hi-rustin:
close https://github.com/diesel-rs/diesel/issues/2581
close https://github.com/diesel-rs/diesel/pull/2590
close https://github.com/diesel-rs/diesel/issues/2745
This PR is based on #2590, thanks dnamsons for the great work.


hi-rustin:
/hold

- [ ] Better documentation
- [ ] changelog
- [ ]  a few test cases for this as part of diesel_tests + maybe a compile test (ui) showing that having collides with locking clauses

hi-rustin:
@weiznich CI passed. Could you please take a look?

diesel/src/pg/query_builder/distinct_on.rs
```diff
@@ -33,16 +34,18 @@ where
     }
 }
 
-impl<ST, F, S, D, W, O, LOf, G, Selection> DistinctOnDsl<Selection>
-    for SelectStatement<F, S, D, W, O, LOf, G>
+impl<ST, F, S, D, W, O, LOf, G, H, Selection> DistinctOnDsl<Selection>
```
pksunkara, line 37: The order here seems different than `SelectStatement` which is why you had to write `NoLockingClause` all the time below. You don't need to do that if you change the `SelectStatement`.

weiznich:
I will try to have a look in the next few days, but I cannot promise anything due to real life work.

@diesel-rs/reviewers if anyone is interested, feel free to have a look at this.

diesel/src/query_builder/select_statement/mod.rs
```diff
@@ -47,6 +49,7 @@ pub struct SelectStatement<
     LimitOffset = LimitOffsetClause<NoLimitClause, NoOffsetClause>,
     GroupBy = NoGroupByClause,
     Locking = NoLockingClause,
+    Having = NoHavingClause,
```
pksunkara, line 52: As said above, let's move this up before `Locking`

pksunkara (changes requested):

diesel/src/pg/query_builder/distinct_on.rs
hi-rustin: Fixed.

diesel/src/query_builder/select_statement/mod.rs
hi-rustin: Fixed.

pksunkara (commented):
I don't see any other issue. But I might have missed some nuances here and there which is why I am not approving this and instead say 👍

hi-rustin:
> I don't see any other issue. But I might have missed some nuances here and there which is why I am not approving this and instead say 👍

Thanks you for your review!

diesel/src/query_dsl/having_dsl.rs
```diff
@@ -0,0 +1,29 @@
+use crate::dsl::Having;
+use crate::query_source::*;
+
+/// The `having` method
+///
+/// This trait should not be relied on directly by most apps. Its behavior is
+/// provided by [`QueryDsl`]. However, you may need a where clause on this trait
+/// to call `having` from generic code.
+///
+/// [`QueryDsl`]: ../trait.QueryDsl.html
```
JohnTitor, line 10: We can use the intra-doc link here:

```suggestion
/// [`QueryDsl`]: crate::QueryDsl
```

weiznich (changes requested):
I've added a few comments about mixed generic usage for `SelectStatement`, those should be easy to fix. Otherwise we should make `SelectStatement` and `BoxedSelectStatement` behave in the same way if there is no `GROUP BY` clause. If you choose to forbid this case, you should add a compile fail test demonstrating that for both `SelectStatement` and `BoxedSelectStatement`.

diesel/src/query_builder/select_statement/boxed.rs
```diff
@@ -347,11 +355,28 @@ where
...
             limit_offset: self.limit_offset,
             group_by: self.group_by,
+            having: self.having,
             _marker: PhantomData,
         }
     }
 }
 
+impl<'a, ST, QS, DB, Predicate, GB> HavingDsl<Predicate>
+    for BoxedSelectStatement<'a, ST, QS, DB, GB>
```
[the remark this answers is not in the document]
hi-rustin: Now they both require group by. But the test ran into some problems, as I said on gitter.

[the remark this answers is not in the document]
hi-rustin: I removed the invalid test.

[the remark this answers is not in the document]
hi-rustin: I added some boxed tests.

weiznich:
https://github.com/diesel-rs/diesel/commit/4ef8c77248c80458a31d3ca6c7daf0c0fbd34579 Seems to be relevant to work around the overflow for missing impls?

hi-rustin:
> [4ef8c77](https://github.com/diesel-rs/diesel/commit/4ef8c77248c80458a31d3ca6c7daf0c0fbd34579) Seems to be relevant to work around the overflow for missing impls?

I tried, but it still doesn't seem to work.

hi-rustin:
@weiznich I've created an issue to track this [issue](https://github.com/diesel-rs/diesel/issues/2745), so maybe I can follow up with some help.

hi-rustin:
> As another blind guess: Maybe the overflow error can be fixed by adding some having based restrictions here: https://github.com/diesel-rs/diesel/blob/master/diesel/src/query_builder/select_statement/mod.rs#L105 in a similar way than this is already be done for other clauses there.

Thanks for your help. But it still not work.

diesel/src/query_dsl/having_dsl.rs
```diff
@@ -0,0 +1,35 @@
...
+    SelectStatement<T>: HavingDsl<Predicate>,
+    T::DefaultSelection: Expression<SqlType = T::SqlType> + ValidGrouping<()>,
+    T::SqlType: TypedExpressionType,
+{
+    type Output = dsl::Having<SelectStatement<T>, Predicate>;
+
+    fn having(self, predicate: Predicate) -> dsl::Having<Self, Predicate> {
+        self.as_query().having(predicate)
+    }
+}
```
weiznich, line 35: I've spend some time to investigate the overflow error. Seems like it is happening because of this impl. That made me think about why we even have this impl in the first place, as a select statement created directly from a table will always have no group by clause. This means this impl is not needed and can be removed. After that the compile test emits a more reasonable error message.

weiznich (changes requested):
See the inline comment for a fix for the overflow issue. Other than that I noticed a few more things that should be fixed:

- [ ] `HAVING` clauses should be checked if they are valid for a specific from clause, similar to how this is done for `WHERE` clauses. In fact something like `users::table.select(users::name).group_by(users::id).having(posts::id.eq(42))` just compiles fine. (There should also a compile test for this case)
- [ ] `HAVING` clauses for boxed select statements do not work. Something like `users::table.select(users::name).group_by(users::id).into_boxed().having(users::id.gt(1))` generates a compiler error. We should have at least one test in `diesel_tests` demonstrating that this works.

diesel/src/query_dsl/having_dsl.rs
hi-rustin: Fixed. Thanks for your help.

hi-rustin:
>     * `HAVING` clauses should be checked if they are valid for a specific from clause, similar to how this is done for `WHERE` clauses. In fact something like `users::table.select(users::name).group_by(users::id).having(posts::id.eq(42))` just compiles fine. (There should also a compile test for this case)

Done.

hi-rustin:
>     * `HAVING` clauses for boxed select statements do not work. Something like `users::table.select(users::name).group_by(users::id).into_boxed().having(users::id.gt(1))` generates a compiler error. We should have at least one test in `diesel_tests` demonstrating that this works.

I don't think it will work because we require having to have a group by clauses.

diesel_tests/tests/having.rs
```diff
@@ -0,0 +1,154 @@
...
+        .select((users::id, users::name, posts::title));
+
+    let expected_data = vec![(3, "Nick".to_string(), "Hi Nick".to_string())];
+    let data: Vec<(i32, String, String)> = source.load(&connection).unwrap();
+
+    assert_eq!(expected_data, data);
+}
+
+#[test]
+fn boxed_multi_condition_having_with_group_by() {
```
hi-rustin, line 125: @weiznich Is this test not a valid test for boxed?

weiznich: This test first sets the having clause and the boxes the query. I'm looking for another test that calls `.into_boxed()` before calling `.having()`, so that we call into the `HavingDsl` of `BoxedSelectStatement`. (Note that you need to call `.into_boxed()` after calling `.group_by`, because changing the grouping is not supported on boxed queries)

hi-rustin:
> This test first sets the having clause and the boxes the query. I'm looking for another test that calls .into_boxed() before calling .having(), so that we call into the HavingDsl of BoxedSelectStatement. (Note that you need to call .into_boxed() after calling .group_by, because changing the grouping is not supported on boxed queries)

If so is it impossible to use having, because it always requires group by.

weiznich:
> If so is it impossible to use having, because it always requires group by.

Its possible to use having on (and have group by clauses on) boxed queries, you just need to call `.group_by` before calling `.into_boxed`. The only restriction there is that you cannot change the group by clause after boxing. On the other hand its fine to have an group by clause.

hi-rustin:
> Its possible to use having on (and have group by clauses on) boxed queries, you just need to call `.group_by` before calling `.into_boxed`. The only restriction there is that you cannot change the group by clause after boxing. On the other hand its fine to have an group by clause.

```rust
impl<'a, ST, QS, DB, GB, Predicate> HavingDsl<Predicate>
    for BoxedSelectStatement<'a, ST, QS, DB, GroupByClause<GB>>
```
But our constraint makes it impossible to use having after into_boxed without group by, right?

weiznich:
Thats correct and thats the reason why you need to set the group by clause before boxing.
Basically this query needs to work:
 users::table.select(users::name).group_by(users::id).into_boxed().having(users::id.gt(1))

hi-rustin:
> Thats correct and thats the reason why you need to set the group by clause before boxing.
> Basically this query needs to work:
> users::table.select(users::name).group_by(users::id).into_boxed().having(users::id.gt(1))

Yes, I see what you mean, I tried this query, but due to the limitation of the having clause, using having alone after into_boxed does not compile. Because we require the having clause to have a groub by clause.

weiznich:
This query has a group by clause. Its set before boxing, but boxing keeps that clause.

hi-rustin:
> Thats correct and thats the reason why you need to set the group by clause before boxing.
> Basically this query needs to work:
> users::table.select(users::name).group_by(users::id).into_boxed().having(users::id.gt(1))

I tried it and it works when we modify the constraint to 

```rust
impl<'a, ST, QS, DB, GB, Predicate> HavingDsl<Predicate>
    for BoxedSelectStatement<'a, ST, QS, DB, GB>
```

hi-rustin:
> I've added a few comments about mixed generic usage for `SelectStatement`, those should be easy to fix. Otherwise we should make `SelectStatement` and `BoxedSelectStatement` behave in the same way if there is no `GROUP BY` clause. If you choose to forbid this case, you should add a compile fail test demonstrating that for both `SelectStatement` and `BoxedSelectStatement`.

But our original idea was to restrict it to use only with group by, right?

hi-rustin:
I think if we want the having only work with group by, the trait `HavingDsl<_>` **shouldn't** implemented for `BoxedSelectStatement<'_, (diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text), JoinOn<diesel::query_source::joins::Join<schema::users::table, SelectStatement<JoinOn<diesel::query_source::joins::Join<schema::posts::table, schema::comments::table, Inner>, Grouped<diesel::expression::operators::Eq<diesel::expression::nullable::Nullable<schema::comments::columns::post_id>, diesel::expression::nullable::Nullable<schema::posts::columns::id>>>>>, Inner>, Grouped<diesel::expression::operators::Eq<diesel::expression::nullable::Nullable<schema::posts::columns::user_id>, diesel::expression::nullable::Nullable<schema::users::columns::id>>>>, Sqlite, (schema::users::columns::id, schema::posts::columns::id)>` right?

weiznich:
It should be possible to constrain the `HavingDsl` impl for `BoxedSelectStatement` in a similar way than the corresponding statement for `SelectStatement` as the actual expression of the group by clause is available via the `GB` generic parameter. It's just that `GB` is not `NoGroupByClause` or `GroupByClause<T>`, but the actual expression that is the group by clause or `()` for no group by clause. So you need to check if `GB` is not `()`, by for example checking that `GB` is a `Expression`, because `()` does not implement `Expression` but any other valid group by clause does.

hi-rustin:
Try to reopen it.

> It should be possible to constrain the `HavingDsl` impl for `BoxedSelectStatement` in a similar way than the corresponding statement for `SelectStatement` as the actual expression of the group by clause is available via the `GB` generic parameter. It's just that `GB` is not `NoGroupByClause` or `GroupByClause<T>`, but the actual expression that is the group by clause or `()` for no group by clause. So you need to check if `GB` is not `()`, by for example checking that `GB` is a `Expression`, because `()` does not implement `Expression` but any other valid group by clause does.

Fixed. @weiznich Can you help me open my PR? diesel-rs/diesel#2740 It seems that GitHub doesn't recognize my new commit, so I'm trying to close and reopen it, and it seems that now I can't reopen it.

hi-rustin:
In fact, I have 8 commits on my branch, but only 7 in the PR.

weiznich:
Test

PR #2741, opened 2021-04-23
Feature/selectable v2

### LINKED ISSUE #1435: Consider `#[derive(QueryAll)]`

### LINKED ISSUE #2037: Pass struct to select

weiznich:
This is a rather large PR. It builds on both #2709 and #2367.

This provides a `Selectable` trait that allows to attach a select clause construction to types. Additionally a corresponding derive is provided, so that this can be used to construct select clauses matching types that implement `Queryable`, so that we side step the "famous" type errors you get if you mix up the field order in such a struct. 
This version requires explicit calls to `.select()` to set that select clause. In contrast to #2367 we don't add an additional method for that, but instead use a type constructor provided by the `Selectable` trait to construct the select clause explicitly. I feel that this way this feature composes better with other parts of `QueryDsl`. So is it for example possible to mix select clause parts coming from `Selectable` and other explicit parts. As soon as you use `as_select()` for some part of your select clause that type is locked in and only that type will be accepted at the corresponding position using `.load()` or similar methods. This is to prevent errors coming from accidentally mixing structs or something like that.
In contrast to `#2709` this does not provide explicit `select_and_load()` methods as those had really bad error messages. It remains possible for third party crates to provide such methods.

Closes #2709
Closes #2367
fixes #2037
fixes #1435

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -40,6 +40,8 @@ pub mod nullable;
 #[macro_use]
 pub mod operators;
 #[doc(hidden)]
+pub mod select_by;
+#[doc(hidden)]
 pub mod sql_literal;
 #[doc(hidden)]
 pub mod subselect;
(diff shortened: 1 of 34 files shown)


Ten0 (approved):

diesel_compile_tests/tests/fail/selectable.rs
```diff
@@ -0,0 +1,212 @@
...
+    //
+    // plain queries
+    let _  = posts::table.select(Post::as_select()).load(&conn).unwrap();
+
+    // boxed queries
+    let _  = posts::table.into_boxed().select(Post::as_select()).load(&conn).unwrap();
+    let _  = posts::table.select(Post::as_select()).into_boxed().load(&conn).unwrap();
+
+    // mixed clauses
+    let _ = posts::table.select((Post::as_select(), posts::title)).load::<(_, String)>(&conn).unwrap();
```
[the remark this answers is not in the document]
weiznich: All queries till line 144 are expected to pass. There are no error messages for those lines in the corresponding .stderr file.

diesel/src/expression/mod.rs
```diff
@@ -318,6 +320,126 @@ where
 {
 }
 
+/// Trait indicating that a record can be selected and queried from the database.
+///
+/// Types which implement `Selectable` represent the select clause of a SQL query.
+/// Use [`SelectableHelper::as_select()`] to construct the select clause. Once you
+/// called `.select(YourType::as_select())` we enforce at type system level that you
```
henryboisdequin, line 327:
```suggestion
/// called `.select(YourType::as_select())` we enforce at the type system level that you
```

diesel/src/expression/mod.rs
```diff
@@ -318,6 +320,126 @@ where
...
+/// use the same type to load the query result into.
+///
+/// The constructed select clause can contain abitary expressions comming from different
```
henryboisdequin, line 330:
```suggestion
/// The constructed select clause can contain arbitrary expressions coming from different
```

henryboisdequin (approved):
Just some doc changes, the implementation looks good :+1:

PR #2748, opened 2021-04-30
Rewrite diesel for each tuple as proc macro

weiznich:
This is an experiment to see if we could reduce the compile times a bit by using a proc macro instead of the old large recursive macro_rules monster.

That written I'm opening this now as WIP PR, to gather some initial number and some feedback on the actual implementation. I feel there should be quite a bit room for improvements especially for the proc macro code here, so if someone feels like profiling some code: Take a look at that macro :wink:

@diesel-rs/reviewers Please try to measure the following times for diesel itself:

* `cargo check --no-default-features`
* `cargo check --no-default-features --features "32-column-table"`
* * `cargo check --no-default-features --features "64-column-table"`
* * `cargo check --no-default-features --features "128-column-table"`

for this branch + the current master branch. Be sure to use a clean build for all cases (so perform a `cargo clean` before each command). It would be great to get some number here in relation to the used operating system + rust compiler version.


diesel_dynamic_schema/Cargo.toml.orig
```diff
@@ -0,0 +1,45 @@
...
+name = "diesel-dynamic-schema"
+version = "1.0.0"
+authors = ["Sean Griffin <sean@seantheprogrammer.com>"]
+license = "MIT OR Apache-2.0"
+edition = "2018"
+<<<<<<< HEAD
+include = ["src/**/*", "LICENSE-*", "README.md"]
+=======
+autotests = false
+>>>>>>> Add support for dynamic select clauses to diesel_dynamic_schema
```
pksunkara, line 11: I guess a temporary file?

diesel_derives/src/for_each_tuple.rs
```diff
@@ -0,0 +1,977 @@
...
+}
+
+macro_rules! start_timed_block {
+    ($name: expr, $debug: expr) => {{
+        if $debug {
+            eprintln!("    {}", $name);
+        }
+        std::time::Instant::now()
+    }};
+}
```
pksunkara, line 48: I guess these are just temporary?

diesel_derives/src/for_each_tuple.rs
```diff
@@ -0,0 +1,977 @@
...
+
+struct ExprReplacer<'a> {
+    idx: u32,
+    repeated_types: &'a [TypeParam],
+}
+
+impl<'a> syn::visit_mut::VisitMut for ExprReplacer<'a> {
+    fn visit_expr_field_mut(&mut self, i: &mut syn::ExprField) {
+        if let syn::Member::Named(n) = i.member.clone() {
+            if n == "idx" {
```
pksunkara, line 901: Even though this is an internal macro, I don't like the fact that we override `.idx`. Can we do `._idx` or something to make sure it is different to a field?

diesel/src/query_dsl/positional_order_dsl.rs
```diff
@@ -111,22 +111,12 @@ impl Order for Desc<OrderColumn> {
...
-        )+
-    };
-}
+#[diesel_derives::__diesel_for_each_tuple]
+impl<#[repeat] T: Order> Order for (T,) {
+    type Fragment = (<T as Order>::Fragment,);
 
-__diesel_for_each_tuple!(impl_order_for_all_tuples);
+    fn into_fragment(self) -> Self::Fragment {
+        #[repeat]
```
pksunkara, line 119: I felt this was inconsistent and was thinking this `#[repeat]` should be inside the `()` on the item because we were using this to repeat whole statements. But then I saw `FromStaticSqlRow` where 2 items are supposed to repeat. I don't mind the exception now, but I would still suggest looking into making this a bit better for consistency.

diesel_derives/src/for_each_tuple.rs
weiznich: Yes they are temporary. I've left them there for now as I'm not that happy with the performance of this proc macro attribute. I would have guessed that this approach can be much faster for large tuple sizes than the old macro by example based implementation, but that's unfortunately not the case. This macros are there to have something to measure what takes that much time.

weiznich:
Closed as outdated experiment

PR #2759, opened 2021-05-07
Remove TODO

hi-rustin:


hi-rustin:
Introduced by https://github.com/diesel-rs/diesel/pull/2512.

weiznich:
That means this is not even released yet. Therefore I think we should keep this in place for at least one year after the 2.0 release.

Issue #2747, opened 2021-04-30
QueryableByName derive does not work for column names that are not idents


Ten0:
### Versions

- **Rust:** Latest stable
- **Diesel:** Latest master
- **Database:** Postgres
- **Operating System** Linux

### Feature Flags

- **diesel:**

## Problem Description
Deriving QueryableByName to parse the output of an `EXPLAIN` query is impossible.

### What are you trying to accomplish?
Deriving QueryableByName to parse the output of an `EXPLAIN` query:
```rust
	#[derive(QueryableByName)]
	struct Out {
		#[sql_type = "diesel::sql_types::Text"]
		#[column_name = "QUERY PLAN "]
		qp: String,
	}
```

### What is the expected output?
Compiles, generating ~
```rust
impl<DB: diesel::backend::Backend> QueryableByName<DB> for Out
where
	String: diesel::deserialize::FromSql<diesel::sql_types::Text, DB>,
{
	fn build<'__a>(row: &impl NamedRow<'__a, DB>) -> deserialize::Result<Self> {
		Ok(Self {
			qp: NamedRow::get(row, "QUERY PLAN")?,
		})
	}
}
```

### What is the actual output?
```
error: proc-macro derive panicked
   --> /home/ten/stockly/Main/stocks/Service/src/items/import.rs:146:20
    |
146 |             #[derive(Debug, QueryableByName)]
    |                             ^^^^^^^^^^^^^^^
    |
    = help: message: `"QUERY PLAN"` is not a valid identifier
```

This is due to https://github.com/diesel-rs/diesel/blob/cc8f0b7350347223205f22f40e1e8de1ca442906/diesel_derives/src/field.rs#L22

### Steps to reproduce
```rust
	#[derive(QueryableByName)]
	struct Out {
		#[sql_type = "diesel::sql_types::Text"]
		#[column_name = "QUERY PLAN "]
		qp: String,
	}
```

### Resolution
We should probably allow this to not be an ident, and only enforce it's an ident if reaching this point:
https://github.com/diesel-rs/diesel/blob/4e697281f1956899e995d23e24661fa02c310f64/diesel_derives/src/queryable_by_name.rs#L102

weiznich:
Seems reasonable :+1:

hi-rustin:
Let me try to enhance it.

hi-rustin:
> We should probably allow this to not be an ident

If it is not an ident, then what type should it be?

weiznich:
[...]

PR #2764, opened 2021-05-08
QueryableByName supoort non ident column name

hi-rustin:
Try to close https://github.com/diesel-rs/diesel/issues/2747

close https://github.com/diesel-rs/diesel/pull/2772

close https://github.com/diesel-rs/diesel/pull/2770

### DIFF
--- a/diesel_derives/src/field.rs
+++ b/diesel_derives/src/field.rs
@@ -13,13 +13,12 @@ pub struct Field {
     pub span: Span,
     pub sql_type: Option<syn::Type>,
     pub flags: MetaItem,
-    column_name_from_attribute: Option<syn::Ident>,
+    column_name_from_attribute: Option<MetaItem>,
 }
 
 impl Field {
     pub fn from_struct_field(field: &syn::Field, index: usize) -> Self {
-        let column_name_from_attribute =
-            MetaItem::with_name(&field.attrs, "column_name").map(|m| m.expect_ident_value());
+        let column_name_from_attribute = MetaItem::with_name(&field.attrs, "column_name");
         let name = match field.ident.clone() {
             Some(mut x) => {
                 // https://github.com/rust-lang/rust/issues/47983#issuecomment-362817105
@@ -51,7 +50,8 @@ impl Field {
 
     pub fn column_name(&self) -> syn::Ident {
         self.column_name_from_attribute
-            .clone()
+            .as_ref()
+            .map(|m| m.expect_ident_value())
             .unwrap_or_else(|| match self.name {
                 FieldName::Named(ref x) => x.clone(),
                 _ => {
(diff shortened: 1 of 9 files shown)


hi-rustin:
This PR is in a very rough stage.

/hold

- [ ] better name
- [ ] better tests

diesel_derives/src/field.rs
```diff
@@ -65,6 +65,23 @@ impl Field {
             })
     }
 
+    pub fn column_str_name(&self) -> String {
+        self.column_name_from_attribute
+            .as_ref()
+            .map(|m| {
+                m.str_value().unwrap_or_else(|e| {
+                    e.emit();
+                    "unknown_column".to_string()
```
weiznich, line 74: If its a named field we should use the actual name of the field here, similary that done in the existing `column_name` function.

diesel_derives/src/field.rs
```diff
@@ -65,6 +65,23 @@ impl Field {
...
+                })
+            })
+            .unwrap_or_else(|| {
+                self.span
+                    .error("All fields of tuple structs must be annotated with `#[column_name]`")
+                    .emit();
+                "unknown_column".to_string()
```
weiznich, line 81: Same as above

diesel_derives/src/queryable_by_name.rs
```diff
@@ -22,15 +22,25 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
+                            {
+                                let field = diesel::row::NamedRow::get(row, stringify!(#name))?;
+                                <#deserialize_ty as Into<#field_ty>>::into(field)
+                            }
+                    ))
+                } else {
+                    let name = f.column_str_name();
+                    Ok(quote!(
+                       {
+                           let field = diesel::row::NamedRow::get(row, stringify!(#name))?;
```
weiznich, line 39: As name already is a string, this stringify is not needed anymore.

hi-rustin:
/unhold

Ready for review.

hi-rustin:
https://github.com/diesel-rs/diesel/pull/2764/checks?check_run_id=2534309735 failed because I used the new API, but since this CI uses minimized packages, it will fail. But if I don't change it, other CIs will fail because of clippy. So maybe the solution is to allow the use of deprecated API in this place? Or is there another way to solve the problem?

weiznich:
In this specific case its fine to just bump the minimal version of url to some version that supports the new conversion.

hi-rustin:
> In this specific case its fine to just bump the minimal version of url to some version that supports the new conversion.

It was implemented on February 22, but was not released until yesterday. So I upgraded to 2.2.2.

diesel_derives/src/queryable_by_name.rs
```diff
@@ -22,15 +22,25 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
+                    ))
+                } else {
+                    let name = f.column_name_str();
+                    Ok(quote!(
+                       {
+                           let field = diesel::row::NamedRow::get(row, #name)?;
+                           <#deserialize_ty as Into<#field_ty>>::into(field)
+                       }
+                    ))
+                }
```
weiznich, line 43: I don't think we need to check if there is a `table_name` attribute or not. If there is a `column_name` attribute we should just unconditionally use the string value from there:

hi-rustin: Makes sense. Fixed.

weiznich (approved):

PR #2770, opened 2021-05-09
FIX - Warning deprecated function url::Url::into_string

Elrendio:


pksunkara (approved):

Elrendio:
The test doesn't pass because it loaded the version `2.1.0` of crate [url](https://docs.rs/url/2.1.0/url/index.html) and not the latest version `2.2.2`. I don't quite understand why because the `Cargo.toml` specifies `url = { version = "2.1.0" }` which is equivalent to `url = { version = "^2.1.0" }` and therefore should load the latest compatible version, aka `2.2.2`

weiznich:
This happens because we also check that everything builds with `-Z minimal-version`, so that the minimal specified version of some dependency is usable with diesel.
Other than that this fix is already included in #2764.

Ten0:
In that case it looks like we should increase the minimal version of that crate then no?

Ten0:
> In that case it looks like we should increase the minimal version of that crate then no?

Oh that's what's done in that other PR, great 🙂

PR #2772, opened 2021-05-10
Fixed URL deprecated method

Razican:
This commit fixes a deprecation in the `url` crate, that is making builds of `diesel_cli` fail (because the deprecated warnings are denied).


weiznich:
This is a duplicate of #2770 and this fix is also already included in #2764. I likely will merge #2764 as soon as the CI passes there.
(On top of that, as outlined in the other PR's this simple change does not fix the CI failure because of our `-Z minimal-versions` check.)

PR #2774, opened 2021-05-11
Try to speedup the diesel insert bench

weiznich:
Remove two unneeded allocations:
* Do not reallocate the string created by `format!`
* Do consume the user vector on insert


weiznich:
Closed as benchmarks do not show any improvements

PR #2777, opened 2021-05-12
COUNT(DISTINCT ...) support

oeed:
As mentioned in #2713, there is not currently support for select `COUNT(DISTINCT column)`. This PR adds the function `count_distinct`, with usage:

```rust
assert_eq!(Ok(2), posts.select(count_distinct(user_id)).first(&connection))
```

I've based in the implementation on the `exists` and `sql_function` implementations, so if there is a better way to implement this I'm happy to take guidance.

There are some failing CI tests, but the errors seem rather unrelated to my changes - so if anyone as any advice to as to how to fix those or if thy can be ignored that'd be great.

---

As an aside, I see the existing count queries use `type SqlType = BigInt`. Would it not make more sense to use `Unsigned<BigInt>`, as a presumably you can't have negative counts?

### DIFF
--- a/diesel/src/expression/count.rs
+++ b/diesel/src/expression/count.rs
@@ -1,11 +1,13 @@
 use std::marker::PhantomData;
 
-use super::{functions::sql_function, is_aggregate, AsExpression};
+use super::functions::sql_function;
+use super::{is_aggregate, AsExpression};
 use super::{Expression, ValidGrouping};
 use crate::query_builder::*;
 use crate::result::QueryResult;
 use crate::sql_types::{BigInt, DieselNumericOps, SingleValue, SqlType};
-use crate::{backend::Backend, AppearsOnTable, SelectableExpression};
+use crate::backend::Backend;
+use crate::{AppearsOnTable, SelectableExpression};
 
 sql_function! {
     /// Creates a SQL `COUNT` expression
--- a/diesel/src/expression/count.rs
+++ b/diesel/src/expression/count.rs
@@ -92,7 +92,8 @@ impl_selectable_expression!(CountStar);
 /// # fn main() {
 /// #     use schema::posts::dsl::*;
 /// #     let connection = establish_connection();
-/// assert_eq!(Ok(2), posts.select(count_distinct(user_id)).first(&connection));
+/// let unique_user_count = posts.select(count_distinct(user_id)).first(&connection);
+/// assert_eq!(Ok(2), unique_user_count);
 /// # }
 /// ```
 pub fn count_distinct<T, E>(expr: E) -> CountDistinct<T, E::Expression>
--- a/diesel/src/expression/count.rs
+++ b/diesel/src/expression/count.rs
@@ -3,10 +3,10 @@ use std::marker::PhantomData;
 use super::functions::sql_function;
 use super::{is_aggregate, AsExpression};
 use super::{Expression, ValidGrouping};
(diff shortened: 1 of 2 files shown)


diesel/src/expression/count.rs
```diff
@@ -1,9 +1,11 @@
-use super::functions::sql_function;
+use std::marker::PhantomData;
+
+use super::{functions::sql_function, is_aggregate, AsExpression};
```
weiznich, line 3: This should follow the general use statement style in diesel

```suggestion
use super::functions::sql_function;
use super::{is_aggregate, AsExpression};
```

diesel/src/expression/count.rs
```diff
@@ -1,9 +1,11 @@
...
 use super::{Expression, ValidGrouping};
-use crate::backend::Backend;
 use crate::query_builder::*;
 use crate::result::QueryResult;
 use crate::sql_types::{BigInt, DieselNumericOps, SingleValue, SqlType};
+use crate::{backend::Backend, AppearsOnTable, SelectableExpression};
```
weiznich, line 8: Same as above:

```suggestion
use crate::backend::Backend;
use crate::{AppearsOnTable, SelectableExpression};
```

diesel/src/expression/count.rs
```diff
@@ -71,3 +73,83 @@ impl<DB: Backend> QueryFragment<DB> for CountStar {
...
+///
+/// ```rust
+/// # #[macro_use] extern crate diesel;
+/// # include!("../doctest_setup.rs");
+/// # use diesel::dsl::*;
+/// #
+/// # fn main() {
+/// #     use schema::posts::dsl::*;
+/// #     let connection = establish_connection();
+/// assert_eq!(Ok(2), posts.select(count_distinct(user_id)).first(&connection));
```
weiznich, line 93: Can we separate the assert and the query for readability?
```suggestion
/// let unique_user_count = posts.select(count_distinct(user_id)).first(&connection);
/// assert_eq!(Ok(2), unique_user_count);
```

weiznich (commented):
Thanks for submitting this feature :+1: 
Looks good to me, modulo some small stylistic changes.

> As an aside, I see the existing count queries use `type SqlType = BigInt`. Would it not make more sense to use `Unsigned<BigInt>`, as a presumably you can't have negative counts?

We use `BigInt` instead of the unsigned variant, for the following reasons:

* Postgres and Sqlite do not support unsigned types. Making this database depended would hurt writing code targeting more than one DBMS
* unsigned for Mysql is really only a check constrained. You cannot have values larger than `i64::MAX` there, so this does not restrict the range of valid values.

oeed:
Those changes look great, have committed them.

Explanation about unsigned vs signed makes complete sense too.

hi-rustin (commented):
You need to fmt the code.
Also you need to fix the compile test issue by running `TRYBUILD=overwrite cargo test` in the compiletests directory.
Also could you add some test cases?

- Normal test
- Compilation failure test for fields not on the table for count_distinct

pksunkara (approved):

pksunkara:
Please update the compilation tests to fix them.

weiznich:
I've merged this manually to include a fix for the compile test failure.

Siceberg:
Thanks for the work guys, this is just the feature I need

SecretKeeper:
will this available in v2? if yes when v2 comes?

Issue #2694, opened 2021-03-31
IS NOT DISTINCT FROM ANY is not a valid PostgreSQL query


regar42:
### Versions

- **Rust:** 1.51.0
- **Diesel:** master (7e676025)
- **Database:** PostgreSQL 13.1
- **Operating System:** Arch Linux

### Feature Flags

- **diesel:** postgres

## Problem Description

`IS NOT DISTINCT FROM ANY` is not a valid PostgresSQL statement, however Diesel allows compiling it and outputs an invalid Postgres query.

### Steps to reproduce
```rust
use diesel::{
    dsl::any,
    pg::PgQueryBuilder,
    query_builder::{QueryBuilder, QueryFragment},
    NullableExpressionMethods, PgExpressionMethods, QueryDsl,
};

fn main() {
    let mut backend = PgQueryBuilder::new();
    a::table
        .select(
            a::id
                .nullable()
                .is_not_distinct_from(any(b::table.select(b::id))),
        )
        .to_sql(&mut backend)
        .unwrap();
    println!("{}", backend.finish());
}

diesel::table! {
        a (id) {
                id -> Int4,
        }
}

diesel::table! {
        b (id) {
                id -> Nullable<Int4>,
        }
}

```

### What is the expected output?
Diesel should not allow this query to compile.
### What is the actual output?
Diesel compiles and outputs this query, which is not actually supported:
```sql
SELECT ("a"."id" IS NOT DISTINCT FROM ANY(SELECT "b"."id" FROM "b")) FROM "a"
```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for opening this issue. I can reproduce this problem and would be happy to receive a fix for it.

Ten0:
Looks like this is due to this:
https://github.com/diesel-rs/diesel/blob/7ff5e6327d5e88095cb4b342da13bf06d6d0d1cd/diesel/src/pg/expression/array_comparison.rs#L75-L81
which seems not really correct (looks like `Any<Expr>` is not really an expression by itself, and the current writing would allow it to also appear in a select).

It looks like one way to fix it would be to introduce something like:
[...]

PR #2779, opened 2021-05-14
Add AsExpressionOrArrayComparison trait

hi-rustin:
Try to close https://github.com/diesel-rs/diesel/issues/2694

### DIFF
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -3,7 +3,7 @@
 //! a boxed trait object, these can be useful for writing concise return types.
 use super::array_comparison::{AsInExpression, In, NotIn};
 use super::grouped::Grouped;
-use super::{AsExpression, Expression};
+use super::Expression;
 use crate::sql_types;
 
 /// The SQL type of an expression
@@ -13,7 +13,7 @@ pub type SqlTypeOf<Expr> = <Expr as Expression>::SqlType;
 pub type AsExpr<Item, TargetExpr> = AsExprOf<Item, SqlTypeOf<TargetExpr>>;
 
 /// The type of `Item` when converted to an expression of `Type`
-pub type AsExprOf<Item, Type> = <Item as AsExpression<Type>>::Expression;
+pub type AsExprOf<Item, Type> = <Item as AsExpressionOrArrayComparison<Type>>::ExpressionOrArrayCmp;
 
 /// The return type of
 /// [`lhs.eq(rhs)`](crate::expression_methods::ExpressionMethods::eq())
@@ -120,6 +120,7 @@ pub use super::functions::helper_types::*;
 #[cfg(feature = "postgres")]
 pub use crate::pg::expression::helper_types::*;
 
+use crate::expression::AsExpressionOrArrayComparison;
 #[doc(inline)]
 #[cfg(feature = "sqlite")]
 pub use crate::sqlite::expression::helper_types::*;
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -13,11 +13,19 @@ pub type SqlTypeOf<Expr> = <Expr as Expression>::SqlType;
 pub type AsExpr<Item, TargetExpr> = AsExprOf<Item, SqlTypeOf<TargetExpr>>;
 
 /// The type of `Item` when converted to an expression of `Type`
-pub type AsExprOf<Item, Type> = <Item as AsExpressionOrArrayComparison<Type>>::ExpressionOrArrayCmp;
+pub type AsExprOf<Item, Type> = <Item as AsExpression<Type>>::Expression;
+
+/// TODO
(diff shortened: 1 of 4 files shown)


hi-rustin:
/hold

- [ ] docs
- [ ] Resolving impl conflicts
- [ ] tests
- [ ]  check the pg manual

diesel/src/expression/helper_types.rs
```diff
@@ -13,7 +13,7 @@ pub type SqlTypeOf<Expr> = <Expr as Expression>::SqlType;
 pub type AsExpr<Item, TargetExpr> = AsExprOf<Item, SqlTypeOf<TargetExpr>>;
 
 /// The type of `Item` when converted to an expression of `Type`
-pub type AsExprOf<Item, Type> = <Item as AsExpression<Type>>::Expression;
+pub type AsExprOf<Item, Type> = <Item as AsExpressionOrArrayComparison<Type>>::ExpressionOrArrayCmp;
```
weiznich, line 16: This should be done as separate type definition. The existing `AsExprOf` impl remains useful on its own.

weiznich (commented):
I would also like to see the query from the original issue added as compile fail test

hi-rustin:
> I would also like to see the query from the original issue added as compile fail test

Of course, it's already on my to-do list.

diesel/src/pg/expression/array_comparison.rs
```diff
@@ -94,6 +81,33 @@ where
...
+/// # Example
+///
+/// ```rust
+/// # include!("../../doctest_setup.rs");
+/// # use diesel::dsl::*;
+/// #
+/// # fn main() {
+/// #     use schema::users::dsl::*;
+/// #     let connection = establish_connection();
+/// #     connection.execute("INSERT INTO users (name) VALUES ('Jim')").unwrap();
```
TaKO8Ki, line 98: These comments are not displayed in the documentation, so I think it is not necessary.

@weiznich 
Is there any particular reason why they are commented out?

Ten0:
I have no idea why this conflicting impl error triggers considering `Any` does not implement `Expression` nor `AsExpression`.

diesel/src/expression/helper_types.rs
hi-rustin: Added a new type.

hi-rustin:
> I have no idea why this conflicting impl error triggers considering `Any` does not implement `Expression` nor `AsExpression`.
> Also, weird, when I generate the doc for Any, this implementation shows on its page as a blanket impl:
> image
> even though no `Expression` or `AsExpression` implementation shows there.

@weiznich Do you have any thoughts on this conflict? I tried to resolve it, but I also wasn't able to find out why it conflicted.

hi-rustin:
It looks like the GitHub actions service is out of service!

weiznich:
@hi-rustin @Ten0 I believe that the conflicting impl error is raised because a third party crate could provide the following impl: 
```rust
impl AsExpression<crate::MyCustomThirdPartyType> for Any<crate::any_other_type> 
{
// …
}
```

From diesels point of view such a impl wouldn't be logical, but from the compiler's point of view such a impl could possibly exists.

That written, I have no clue how to workaround this.

Ten0:
> @hi-rustin @Ten0 I believe that the conflicting impl error is raised because a third party crate could provide the following impl:
> 
> ```rust
> impl AsExpression<crate::MyCustomThirdPartyType> for Any<crate::any_other_type> 
> {
> // …
> }
> ```
> 
> From diesels point of view such a impl wouldn't be logical, but from the compiler's point of view such a impl could possibly exists.
> 
> That written, I have no clue how to workaround this.

That makes sense. One alternate way to solve this would be to introduce `.eq_any(` and `.eq_all(`.
That would be much simpler and modular from with regards to the trait-system, and I don't think there's really ever a scenario where it's going to be a blocker compared to having an `any` expression-like.
On the other hand that would break backwards-compatibility.

weiznich:
I think going doing the extension trait variant is the only real chance to fix this issue. Method names are up to discussion, but using `.eq_any` is not possible, as the `IN` equivalent is already named that way. We should also keep in mind that this not only affects `=` but also other operators (`!= ALL()`) is also useful :wink:

hi-rustin:
> I think going doing the extension trait variant is the only real chance to fix this issue. Method names are up to discussion, but using `.eq_any` is not possible, as the `IN` equivalent is already named that way. We should also keep in mind that this not only affects `=` but also other operators (`!= ALL()`) is also useful 😉
> Breaking things for making it more sound is fine in my opinion.

Then I understand we need to redefine the `= any()` operator for pg? Also there could be a lot of operators in this place? For example: `=` `! =` `<>` `>` `<`? Do I need to define a lot of extension traits for this to work? I'm not sure I understand you guys correctly. Thank you very much!

hi-rustin:
I didn't have time to continue modifying it recently and closed it for now. I will open it later when I retry to fix the issue.

PR #2791, opened 2021-05-21
Add support for no argument functions for Sqlite

### LINKED ISSUE #2790: sqlite: not possible to use SQL functions without Arguments
### Versions

- **Rust:** 1.52.1
- **Diesel:** master
- **Database:** Sqlite
- **Operating System:** Linux

### Feature Flags

- **diesel:** sqlite

## Problem Description

[This condition](https://github.com/diesel-rs/diesel/blob/7587a3bda93bcbe8d44a22f8537da157b953ae5e/diesel_derives/src/sql_function.rs#L245) skips the code generation of `register_impl` and `register_nondeterministic_impl` for SQL functions without arguments. Because the [documentation](https://docs.diesel.rs/master/diesel/prelude/macro.sql_function.html#sql-functions-without-arguments) doesn't mention any limitations for Sqlite SQL functions without arguments, I'm assuming this to be a bug. 

I've tracked down the [commit that introduced that regression]( https://github.com/diesel-rs/diesel/commit/dd05ee2588f80f240896bd8c95cd8539d706d7e7#diff-3dc6bb01ed73137c473d075bf6f9ae482857b07fdfd5aa7d95691705daaebc6bR235) but its commit message doesn't help to understand why the condition was changed.

I want to use a SQL function without arguments with Sqlite.

No errors.

> cannot find function `register_nondeterministic_impl` in module `foo`

No.

### Steps to reproduce

```rs
sql_function! {
    fn foo() -> Text;
}
fn main() {
    let conn = diesel::SqliteConnection::establish(":memory:");
    foo::register_impl(&conn, || "foo".to_string());
}
```

## Checklist

toxeus:
Closes #2790

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -241,6 +241,21 @@ impl SqliteConnection {
     }
 
     #[doc(hidden)]
+    pub fn register_noarg_sql_function<RetSqlType, Ret, F>(
+        &self,
+        fn_name: &str,
+        deterministic: bool,
+        f: F,
+    ) -> QueryResult<()>
+    where
+        F: FnMut() -> Ret + std::panic::UnwindSafe + Send + 'static,
+        Ret: ToSql<RetSqlType, Sqlite>,
+        Sqlite: HasSqlType<RetSqlType>,
+    {
+        functions::register_noargs(&self.raw_connection, fn_name, deterministic, f)
+    }
+
+    #[doc(hidden)]
     pub fn register_aggregate_function<ArgsSqlType, RetSqlType, Args, Ret, A>(
         &mut self,
         fn_name: &str,
@@ -411,6 +426,26 @@ mod tests {
 
         let added = crate::select(my_add(1, 2)).get_result::<i32>(connection);
         assert_eq!(Ok(3), added);
+    }
+
+    sql_function!(fn answer() -> Integer);
+
+    #[test]
+    fn register_noarg_function() {
+        let connection = &mut SqliteConnection::establish(":memory:").unwrap();
+        answer::register_impl(&connection, || 42).unwrap();
+
+        let answer = crate::select(answer()).get_result::<i32>(connection);
+        assert_eq!(Ok(42), answer);
+    }
+
+    #[test]
+    fn register_nondeterministic_noarg_function() {
+        let connection = &mut SqliteConnection::establish(":memory:").unwrap();
+        answer::register_nondeterministic_impl(&connection, || 42).unwrap();
+
+        let answer = crate::select(answer()).get_result::<i32>(connection);
+        assert_eq!(Ok(42), answer);
     }
 
     sql_function!(fn add_counter(x: Integer) -> Integer);
--- a/diesel_derives/src/sql_function.rs
+++ b/diesel_derives/src/sql_function.rs
@@ -302,6 +302,66 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic> 
                 }
             };
         }
+
+        if cfg!(feature = "sqlite")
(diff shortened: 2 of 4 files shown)


diesel_derives/src/sql_function.rs
```diff
@@ -302,6 +302,66 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic>
                 }
             };
         }
+
+        if cfg!(feature = "sqlite")
+            && type_args.is_empty()
+            && arg_name.is_empty()
+            && !sql_name.starts_with("pg")
```
toxeus, line 309: This line is quite hacky. It's here to skip

```rust
sql_function! { fn pg_my_temp_schema() -> Oid; }
```

which uses a return type only known to Postgres. It would be cleaner to skip functions that contain types unknown to Sqlite. Is there a mechanism to check that? Btw, this issue also exists for functions taking more than zero arguments as soon as they use types unknown to Sqlite.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -237,6 +237,21 @@ impl SqliteConnection {
...
+        fn_name: &str,
+        deterministic: bool,
+        f: F,
+    ) -> QueryResult<()>
+    where
+        F: FnMut() -> Ret + std::panic::UnwindSafe + Send + 'static,
+        Ret: ToSql<RetSqlType, Sqlite>,
+        Sqlite: HasSqlType<RetSqlType>,
+    {
+        functions::register_noargs(&self.raw_connection, fn_name, deterministic, f)
```
toxeus, line 252: This could be inlined.

diesel_derives/src/sql_function.rs
toxeus: > I would be fine to implement it that way, as long as it's the same for functions with all arities.

I would include such a fix for the newly introduced functions here and fix the existing functions in a separate PR. I like to work with a narrow scope because it's more efficient for the reviewer and for me. Does that sound good for you?

toxeus: @weiznich, fixed.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -413,6 +428,17 @@ mod tests {
         assert_eq!(Ok(3), added);
     }
 
+    sql_function!(fn answer() -> Integer);
+
+    #[test]
+    fn register_noarg_function() {
+        let connection = SqliteConnection::establish(":memory:").unwrap();
+        answer::register_nondeterministic_impl(&connection, || 42).unwrap();
```
weiznich, line 436: I would like to see a similar test for the `register_impl` variant here for completeness.

diesel_derives/src/sql_function.rs
```diff
@@ -392,3 +452,21 @@ impl ToTokens for StrictFnArg {
         self.name.to_tokens(tokens);
     }
 }
+
+fn is_sqlite_return_type(ty: &syn::Type) -> bool {
```
weiznich, line 456: As we want to use that functions on sql functions with arguments to check the type of the argument too, I would prefer to just name it `is_sqlite_type`.

weiznich (commented):
Looks nearly good :+1:

I've left some minor comments + I would like to see a Changelog entry for this.

diesel/src/sqlite/connection/mod.rs
toxeus: Fixed.

diesel_derives/src/sql_function.rs
toxeus: Fixed.

weiznich (approved):

toxeus:
Nightly `rustup` fails for Windows. Looks like it needs more memory. Not related to this PR.

PR #2796, opened 2021-05-27
Increase robustness of Sqlite SQL functions

toxeus:
Increase the robustness of Sqlite SQL functions by skipping them if a type unknown to Sqlite is used in the function's arguments or return value.

As [agreed](https://github.com/diesel-rs/diesel/pull/2791#discussion_r637036494) earlier.

### DIFF
--- a/diesel_derives/src/sql_function.rs
+++ b/diesel_derives/src/sql_function.rs
@@ -142,6 +142,11 @@ pub(crate) fn expand(input: SqlFunctionDecl) -> Result<TokenStream, Diagnostic> 
         }
     };
 
+    let is_supported_on_sqlite = cfg!(feature = "sqlite")
+        && type_args.is_empty()
+        && is_sqlite_type(&return_type)
+        && arg_type.iter().all(|a| is_sqlite_type(a));
+
     if is_aggregate {
         tokens = quote! {
             #tokens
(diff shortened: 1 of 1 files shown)


diesel_derives/src/sql_function.rs
```diff
@@ -453,7 +454,12 @@ impl ToTokens for StrictFnArg {
     }
 }
 
-fn is_sqlite_type(ty: &syn::Type) -> bool {
+fn is_sqlite_type<T: ToTokens>(ty: &T) -> bool {
+    let ty = quote!(#ty).to_string();
+    let ty_stripped = ty
+        .strip_prefix("Nullable < ")
+        .and_then(|t| t.strip_suffix(" >"))
+        .unwrap_or(ty.as_str());
```
toxeus, line 462: Aggregations can have a nullable return type. If there's a way to handle that case without string manipulation then plz let know.

weiznich: We could try to deconstruct a concrete syn type here. I think `syn::Type` or `syn::TypePath` should be the correct one here. This would also handle fully qualified types correctly.

toxeus: I have very limited experience with the `syn` crate. Therefore, I don't understand what you mean by "try to deconstruct a concrete syn type". Can you please elaborate?

toxeus: Thanks, done. Sorry, I was busy the last two weeks and came around working on this just now.

weiznich (approved):
Looks fine for me :+1: Thanks for the update

PR #2797, opened 2021-05-29
1.4.x: Update 'libsqlite3-sys' to 0.22

SergioBenitez:
Update `libsqlite3-sys`. Without this update, Rocket's `rocket_contrib`, which depends on `diesel`, cannot update other dependencies, resulting in quite a few out of date dependencies.

Unless the CI does something special, this will almost certainly fail the CI. The 1.4.x branch seems to be in a rather broken state. `rust-toolchain` points to an MSRV that is incorrect, and the source doesn't build due to a `deny`. In any case, here's the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
 for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
 
+## [1.4.7] - 2021-06-03
+
+### Fixed
+
+* Updated 'libsqlite3-sys' to 0.22
+
 ## [1.4.6] - 2021-03-05
 
 ### Fixed
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -16,7 +16,7 @@ byteorder = "1.0"
 diesel_derives = "~1.4.0"
 chrono = { version = "0.4", optional = true }
 libc = { version = "0.2.0", optional = true }
-libsqlite3-sys = { version = ">=0.8.0, <0.21.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
+libsqlite3-sys = { version = ">=0.8.0, <0.23.0", optional = true, features = ["min_sqlite_version_3_7_16"] }
 mysqlclient-sys = { version = ">=0.1.0, <0.3.0", optional = true }
 pq-sys = { version = ">=0.3.0, <0.5.0", optional = true }
 quickcheck = { version = "0.4", optional = true }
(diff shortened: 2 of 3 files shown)


weiznich:
Looks good from my side, but this needs a Changelog entry before merging + a rebase/merge of the commit that fixes the mysql CI

> Unless the CI does something special, this will almost certainly fail the CI. The 1.4.x branch seems to be in a rather broken state. rust-toolchain points to an MSRV that is incorrect, and the source doesn't build due to a deny. In any case, here's the PR.

To clarify that neither is the branch in a "broken" state nor is the MSRV set incorrectly. (We don't define MSRV as diesel and any version of it's dependencies will built at that rust version, but only: There is a set of dependency versions where diesel and all dependencies build at that version. We check this by using `-Z minimal-versions` + some manual updates to fix broken dependencies). CI is expected to succeed. It set's `RUSTFLAGS=--cap-lints=warn` in a similar way than cargo handles denys in dependencies.

SergioBenitez:
> > Unless the CI does something special, this will almost certainly fail the CI. The 1.4.x branch seems to be in a rather broken state. rust-toolchain points to an MSRV that is incorrect, and the source doesn't build due to a deny. In any case, here's the PR.
> 
> To clarify that neither is the branch in a "broken" state nor is the MSRV set incorrectly. (We don't define MSRV as diesel and any version of it's dependencies will built at that rust version, but only: There is a set of dependency versions where diesel and all dependencies build at that version. We check this by using `-Z minimal-versions` + some manual updates to fix broken dependencies). CI is expected to succeed. It set's `RUSTFLAGS=--cap-lints=warn` in a similar way than cargo handles denys in dependencies.

Got it. So the CI _does_ do something special. This deviates significantly from the norm in Rust ecosystem, where, aside from system-level dependencies, crates are generally expected to compile with `cargo check`. I see no mention of this at all anywhere in the repository. Specifically, doing as [CONTRIBUTING](https://github.com/diesel-rs/diesel/blob/3e8fd4d7ce257830f1735e9a31c9a5f0fb0d777f/CONTRIBUTING.md) suggests results in the myriad errors I allude to.

> Looks good from my side, but this needs a Changelog entry before merging + a rebase/merge of the commit that fixes the mysql CI

Done, though I wasn't quite sure which date to use in the CHANGELOG, so I've just used today's.

SergioBenitez:
The CI is failing on nightly because `derive-error-chain` uses dependency versions like `0.11.x` which appear to no longer be accepted by Cargo. This seems to be caused by https://github.com/rust-lang/cargo/pull/9508. Polling there.

weiznich:
Thanks for the update. I've field https://github.com/rust-lang/rust/issues/85986 for the regression, let's see what's the response there.

> Got it. So the CI does do something special. This deviates significantly from the norm in Rust ecosystem, where, aside from system-level dependencies, crates are generally expected to compile with cargo check. I also see no mention of this at all anywhere in the repository. Specifically, doing as CONTRIBUTING suggests results in the myriad errors I allude to.

Right it does something special. I do not want to include that into `CONTRIBUTING` as we normally do not accept PR's to any other branch than master, beside special cases like this one. The next major release will likely fix that issue, by just removing `#[deny(warnings)]` before release.

weiznich:
Just want to leave a comment here that I haven't found the time to do a release today. I will try finding some time tomorrow.

weiznich:
Published as 1.4.7 :tada:

SergioBenitez:
Awesome. Thank you!

PR #2799, opened 2021-06-01
Return a iterator from `Connection::load`

weiznich:
This PR provides a prototypical implementation of `Connection::load` so
that a iterator of `Row`'s is returned. This is currently only
implemented for the postgres backend to checkout the performance impact
of this change.

@diesel-rs/reviewers Feel free to have a first look at this and comment on the general approach. I'm well aware that there are many missing pieces like documentation, implementation for sqlite and mysql, cleaning up some artifacts from refactoring, …


diesel/src/pg/connection/cursor.rs
```diff
@@ -24,13 +28,13 @@ impl<'a> ExactSizeIterator for Cursor<'a> {
 }
 
 impl<'a> Iterator for Cursor<'a> {
-    type Item = PgRow<'a>;
+    type Item = crate::QueryResult<PgRow<'a>>;
```
Ten0, line 31: It looks like we don't really need `Iterator` that returns `Row`s: all we ever do with it is call `.map(FromSqlRow::build_from_row)`. If we instead make an iterator that both creates the `Row<'a>` and calls `FromSqlRow` on it as part of a single iterator we don't need to add `Rc`s everywhere.

That being said in row-by-row mode there's no need to keep a reference to a global result object, the `PGresult` returned by each call to `PQgetResult` is enough so we could just store it in `Row`. ([doc](https://www.postgresql.org/docs/current/libpq-single-row-mode.html))

diesel/src/pg/connection/result.rs
```diff
@@ -2,21 +2,26 @@ extern crate pq_sys;
...
 use crate::result::{DatabaseErrorInformation, DatabaseErrorKind, Error, QueryResult};
+use crate::util::OnceCell;
 
-pub struct PgResult {
+pub(crate) struct PgResult<'a> {
     internal_result: RawResult,
     column_count: usize,
     row_count: usize,
+    column_name_map: OnceCell<Vec<Option<String>>>,
+    _marker: PhantomData<&'a super::PgConnection>,
```
Ten0, line 21: This marker seems to indicate that the connection should not be dropped or reused before the `PgResult` is dropped, but I haven't seen a reference of this in the Postgres doc. Did I miss it and the previous impl was not super safe because that was missing? Otherwise, why is this here?

Ten0 (commented):
If I understand correctly, the PG backend is currently using the entire-result mode of `libpq`, while for this to resolve [the original issue](https://github.com/diesel-rs/diesel/discussions/2707#discussion-3311186), we'd have to use [row-by-row mode](https://www.postgresql.org/docs/current/libpq-single-row-mode.html) and this is not yet implemented as part of this draft.

diesel/src/query_dsl/mod.rs
```diff
@@ -1404,7 +1404,7 @@ pub trait RunQueryDsl<Conn>: Sized {
     where
         Self: LoadQuery<Conn, U>,
     {
-        self.internal_load(conn)
+        self.internal_load(conn)?.collect()
```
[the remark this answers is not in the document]
weiznich: The point is even for implementations which fetch everything at once we must iterate over all elements as we would like to map every row to a tuple of rust values. Returning an iterator internally may allow future users to decide to collect into some different container than a vector. At least in my own personal projects I build sometimes a `HashMap` for random lookup out of the database result. That would be easier if I could just take a `Iterator<Item = (Key, Value)>` and collect that into the HashMap directly.

diesel/src/pg/connection/cursor.rs
weiznich: > It looks like we don't really need Iterator that returns Rows: all we ever do with it is call .map(FromSqlRow::build_from_row). If we instead make an iterator that both creates the Row<'a> and calls FromSqlRow on it as part of a single iterator we don't need to add Rcs everywhere. 

For diesel itself it's true that we never do anything else than calling `FromSqlRow::build_from_row` onto the returned row. The idea behind this change is that it's now possible to just skip diesels deserialization layer, in such a way that it is possible to interact with the raw result set. This is useful for cases where you don't know the schema at compile time and want to build for example select clauses at runtime. In the end this just tries to separate different diesel layers more strictly, so that in the future someone can decide to use diesels `Connection` but to skip the query builder entirely. 

That all written, I'm not sure if that's even worth it and if it will work with all backends.

diesel/src/pg/connection/result.rs
weiznich: It's just there to be really safe that nothing goes wrong 🙈

weiznich:
> If I understand correctly, the PG backend is currently using the entire-result mode of libpq, while for this to resolve the original issue, we'd have to use row-by-row mode and this is not yet implemented as part of this draft.

After having looked at the row-by-row mode I'm not sure if I really want to use that as part of the libpq based connection. In general it seems to be a shortcoming of libpq that there is no way to fetch single rows/blocks of rows while reusing existing buffers like this is possible in other client implementations. We may want to keep the fetch all solution for `PgConnection` and maybe provide a pure rust `Connection` implementation someday that solves this better.

diesel/src/sqlite/connection/row.rs
```diff
@@ -0,0 +1,155 @@
...
+            PrivateSqliteRow::Direct(stmt) => stmt.value(self.col_idx),
+            PrivateSqliteRow::Duplicated { values, .. } => {
+                values.get(self.col_idx as usize).and_then(|v| {
+                    v.as_ref()
+                        .and_then(|v| unsafe { SqliteValue::new(v.value.as_ptr()) })
+                })
+            }
+            PrivateSqliteRow::TemporaryEmpty => unreachable!(),
+        }
+    }
```
weiznich, line 154: I'm not sure if that function is really sound, especially if we could get surprising behaviour somewhere if someone manages to cache a single value somewhere + we change the duplicated representation internally. A review would be welcome.

weiznich:
I've updated all implementation now. Let's see how the benchmarks look. A more in depth review would be welcome.
Things I will address maybe next week:
- Add comments to all new unsafe blocks whey the are sound
- Add insightful messages to `unreachable!` calls prompting users to open github issues if they ever see that message.

diesel/src/sqlite/connection/row.rs
weiznich: As expected this was not sound at all. Replaced by a new better implementation that does not need to use unsafe.

weiznich:
@diesel-rs/reviewers This is now ready for a finial review.

Benchmark results:

The largest regression seems to be for `queryable_by_name` on the sqlite backend. Otherwise there are some slight regressions for large results sets for sqlite + mysql and some minor speedups for smaller results sets. I'm not sure where exactly they are coming from.

<details>
 <summary> <b>Postgres</b> </summary>

```
group                                                                                     changes                                 master
-----                                                                                     -------                                 ------
bench_insert/diesel/1                                                                     1.00   347.3±40.27µs        ? ?/sec     1.03  357.0±126.75µs        ? ?/sec
bench_insert/diesel/10                                                                    1.00   404.1±29.76µs        ? ?/sec     1.04   419.2±44.81µs        ? ?/sec
bench_insert/diesel/100                                                                   1.04  1274.3±145.39µs        ? ?/sec    1.00  1229.8±67.29µs        ? ?/sec
bench_insert/diesel/25                                                                    1.00   548.7±58.93µs        ? ?/sec     1.00   550.5±45.86µs        ? ?/sec
bench_insert/diesel/50                                                                    1.00   728.4±56.76µs        ? ?/sec     1.05   764.4±48.21µs        ? ?/sec
bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially    1.00     36.4±1.58ms        ? ?/sec     1.06     38.5±1.68ms        ? ?/sec
bench_medium_complex_query/diesel/1                                                       1.05    81.1±13.91µs        ? ?/sec     1.00    77.0±13.03µs        ? ?/sec
bench_medium_complex_query/diesel/10                                                      1.01    91.4±12.66µs        ? ?/sec     1.00    90.5±12.20µs        ? ?/sec
bench_medium_complex_query/diesel/100                                                     1.01   145.5±20.79µs        ? ?/sec     1.00   143.6±22.42µs        ? ?/sec
bench_medium_complex_query/diesel/1000                                                    1.00   624.3±76.15µs        ? ?/sec     1.05   656.4±74.72µs        ? ?/sec
bench_medium_complex_query/diesel/10000                                                   1.00      5.3±0.29ms        ? ?/sec     1.01      5.4±0.31ms        ? ?/sec
bench_medium_complex_query/diesel_boxed/1                                                 1.00    82.6±11.00µs        ? ?/sec     1.01    83.4±10.76µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10                                                1.00    92.4±14.17µs        ? ?/sec     1.00    92.8±13.10µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/100                                               1.05   144.9±26.42µs        ? ?/sec     1.00   138.3±21.43µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/1000                                              1.00   653.3±85.44µs        ? ?/sec     1.00   653.2±86.72µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10000                                             1.02      5.4±0.30ms        ? ?/sec     1.00      5.3±0.24ms        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1                                     1.02   227.8±16.94µs        ? ?/sec     1.00   223.5±19.59µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10                                    1.00   238.2±17.40µs        ? ?/sec     1.01   240.1±27.49µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/100                                   1.00   348.1±29.94µs        ? ?/sec     1.07   372.1±30.73µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1000                                  1.00  1277.6±105.32µs        ? ?/sec    1.12  1432.9±113.39µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10000                                 1.00     10.5±0.58ms        ? ?/sec     1.17     12.2±0.67ms        ? ?/sec
bench_trivial_query/diesel/1                                                              1.00     59.3±5.39µs        ? ?/sec     1.01     59.9±4.02µs        ? ?/sec
bench_trivial_query/diesel/10                                                             1.07     68.1±3.29µs        ? ?/sec     1.00     64.0±4.24µs        ? ?/sec
bench_trivial_query/diesel/100                                                            1.03    122.2±5.30µs        ? ?/sec     1.00   118.2±10.75µs        ? ?/sec
bench_trivial_query/diesel/1000                                                           1.00   605.9±34.08µs        ? ?/sec     1.03   621.3±51.12µs        ? ?/sec
bench_trivial_query/diesel/10000                                                          1.00      5.5±0.30ms        ? ?/sec     1.01      5.6±0.23ms        ? ?/sec
bench_trivial_query/diesel_boxed/1                                                        1.05     63.0±3.45µs        ? ?/sec     1.00     60.1±3.47µs        ? ?/sec
bench_trivial_query/diesel_boxed/10                                                       1.00     66.2±3.85µs        ? ?/sec     1.03     68.5±4.01µs        ? ?/sec
bench_trivial_query/diesel_boxed/100                                                      1.00   118.6±16.26µs        ? ?/sec     1.05   124.9±11.99µs        ? ?/sec
bench_trivial_query/diesel_boxed/1000                                                     1.01   626.1±40.92µs        ? ?/sec     1.00   621.0±37.46µs        ? ?/sec
bench_trivial_query/diesel_boxed/10000                                                    1.00      5.4±0.28ms        ? ?/sec     1.06      5.8±0.38ms        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1                                            1.00   136.6±10.98µs        ? ?/sec     1.01    137.4±5.61µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10                                           1.00    137.9±6.76µs        ? ?/sec     1.01    138.6±8.84µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/100                                          1.00   197.4±10.68µs        ? ?/sec     1.09   214.8±13.71µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1000                                         1.00   682.2±33.94µs        ? ?/sec     1.15   783.7±39.84µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10000                                        1.00      5.4±0.25ms        ? ?/sec     1.19      6.4±0.36ms        ? ?/sec
```

</details>

<details>
<summary><b>Sqlite:</b></summary>

```
group                                                                                     changes                                master
-----                                                                                     -------                                ------
bench_insert/diesel/1                                                                     1.01      9.1±0.05µs        ? ?/sec    1.00      9.0±0.04µs        ? ?/sec
bench_insert/diesel/10                                                                    1.06     57.6±2.34µs        ? ?/sec    1.00     54.6±1.28µs        ? ?/sec
bench_insert/diesel/100                                                                   1.00    514.5±2.41µs        ? ?/sec    1.00    515.0±2.72µs        ? ?/sec
bench_insert/diesel/25                                                                    1.00    136.5±1.24µs        ? ?/sec    1.01    137.3±1.53µs        ? ?/sec
bench_insert/diesel/50                                                                    1.00    264.2±1.52µs        ? ?/sec    1.00    264.4±1.63µs        ? ?/sec
bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially    1.06    532.6±2.78µs        ? ?/sec    1.00    500.2±3.52µs        ? ?/sec
bench_medium_complex_query/diesel/1                                                       1.00      3.4±0.05µs        ? ?/sec    1.07      3.7±0.03µs        ? ?/sec
bench_medium_complex_query/diesel/10                                                      1.00      6.1±0.05µs        ? ?/sec    1.02      6.2±0.06µs        ? ?/sec
bench_medium_complex_query/diesel/100                                                     1.03     34.2±0.27µs        ? ?/sec    1.00     33.3±0.21µs        ? ?/sec
bench_medium_complex_query/diesel/1000                                                    1.04    312.2±1.96µs        ? ?/sec    1.00    299.9±2.10µs        ? ?/sec
bench_medium_complex_query/diesel/10000                                                   1.05      3.1±0.02ms        ? ?/sec    1.00      3.0±0.02ms        ? ?/sec
bench_medium_complex_query/diesel_boxed/1                                                 1.00      5.9±0.42µs        ? ?/sec    1.00      5.9±0.28µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10                                                1.00      8.2±0.10µs        ? ?/sec    1.01      8.3±0.06µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/100                                               1.01     36.2±0.30µs        ? ?/sec    1.00     35.9±0.21µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/1000                                              1.02    316.3±2.21µs        ? ?/sec    1.00    309.9±2.08µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10000                                             1.02      3.1±0.01ms        ? ?/sec    1.00      3.0±0.02ms        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1                                     1.17     20.3±0.10µs        ? ?/sec    1.00     17.4±0.10µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10                                    1.07     24.1±0.19µs        ? ?/sec    1.00     22.5±0.14µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/100                                   1.19     81.7±0.49µs        ? ?/sec    1.00     68.8±0.35µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1000                                  1.24    658.2±5.11µs        ? ?/sec    1.00    531.5±3.27µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10000                                 1.26      6.5±0.05ms        ? ?/sec    1.00      5.1±0.04ms        ? ?/sec
bench_trivial_query/diesel/1                                                              1.00   812.7±10.89ns        ? ?/sec    1.12    911.1±6.73ns        ? ?/sec
bench_trivial_query/diesel/10                                                             1.01      3.8±0.04µs        ? ?/sec    1.00      3.8±0.06µs        ? ?/sec
bench_trivial_query/diesel/100                                                            1.05     34.0±0.42µs        ? ?/sec    1.00     32.3±0.62µs        ? ?/sec
bench_trivial_query/diesel/1000                                                           1.09    342.1±4.18µs        ? ?/sec    1.00    312.9±2.47µs        ? ?/sec
bench_trivial_query/diesel/10000                                                          1.09      3.4±0.02ms        ? ?/sec    1.00      3.1±0.01ms        ? ?/sec
bench_trivial_query/diesel_boxed/1                                                        1.00   1595.4±9.29ns        ? ?/sec    1.03  1645.0±26.47ns        ? ?/sec
bench_trivial_query/diesel_boxed/10                                                       1.04      4.6±0.04µs        ? ?/sec    1.00      4.4±0.06µs        ? ?/sec
bench_trivial_query/diesel_boxed/100                                                      1.09     34.7±0.27µs        ? ?/sec    1.00     31.9±0.15µs        ? ?/sec
bench_trivial_query/diesel_boxed/1000                                                     1.11    348.8±6.12µs        ? ?/sec    1.00    314.4±1.51µs        ? ?/sec
bench_trivial_query/diesel_boxed/10000                                                    1.09      3.4±0.03ms        ? ?/sec    1.00      3.1±0.03ms        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1                                            1.00      5.4±0.05µs        ? ?/sec    1.01      5.5±0.09µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10                                           1.07      8.7±0.07µs        ? ?/sec    1.00      8.2±0.10µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/100                                          1.21     39.9±0.41µs        ? ?/sec    1.00     33.0±0.28µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1000                                         1.24    359.8±1.78µs        ? ?/sec    1.00   289.9±26.22µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10000                                        1.28      3.5±0.01ms        ? ?/sec    1.00      2.8±0.01ms        ? ?/sec
```

</details>

<details>
<summary><b>Mysql:</b></summary>


```
group                                                                                     changes                                 master
-----                                                                                     -------                                 ------
bench_insert/diesel/1                                                                     1.01   718.3±96.91µs        ? ?/sec     1.00   714.2±94.25µs        ? ?/sec
bench_insert/diesel/10                                                                    1.04   862.8±72.87µs        ? ?/sec     1.00  828.3±102.29µs        ? ?/sec
bench_insert/diesel/100                                                                   1.00      2.0±0.12ms        ? ?/sec     1.03      2.1±0.16ms        ? ?/sec
bench_insert/diesel/25                                                                    1.10  1106.8±143.39µs        ? ?/sec    1.00  1007.7±88.46µs        ? ?/sec
bench_insert/diesel/50                                                                    1.00  1423.4±165.37µs        ? ?/sec    1.01  1441.6±191.62µs        ? ?/sec
bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially    1.00     92.2±4.03ms        ? ?/sec     1.03     94.8±5.33ms        ? ?/sec
bench_medium_complex_query/diesel/1                                                       1.04   103.0±11.69µs        ? ?/sec     1.00     99.0±9.42µs        ? ?/sec
bench_medium_complex_query/diesel/10                                                      1.03   146.4±16.58µs        ? ?/sec     1.00   142.7±15.57µs        ? ?/sec
bench_medium_complex_query/diesel/100                                                     1.04   241.9±27.91µs        ? ?/sec     1.00   232.9±26.16µs        ? ?/sec
bench_medium_complex_query/diesel/1000                                                    1.12  1098.2±191.08µs        ? ?/sec    1.00   980.1±98.75µs        ? ?/sec
bench_medium_complex_query/diesel/10000                                                   1.07      9.3±0.51ms        ? ?/sec     1.00      8.7±0.51ms        ? ?/sec
bench_medium_complex_query/diesel_boxed/1                                                 1.03   103.9±12.31µs        ? ?/sec     1.00   100.8±11.65µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10                                                1.04   151.0±19.24µs        ? ?/sec     1.00   145.7±16.69µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/100                                               1.08   243.8±38.68µs        ? ?/sec     1.00   225.5±25.55µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/1000                                              1.16  1119.0±166.28µs        ? ?/sec    1.00   963.8±79.69µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10000                                             1.08      9.3±0.55ms        ? ?/sec     1.00      8.6±0.45ms        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1                                     1.04   147.2±15.53µs        ? ?/sec     1.00   141.5±13.87µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10                                    1.07   207.9±20.62µs        ? ?/sec     1.00   193.6±22.22µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/100                                   1.07   330.3±34.14µs        ? ?/sec     1.00   308.5±28.77µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1000                                  1.01  1439.0±122.89µs        ? ?/sec    1.00  1421.7±116.62µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10000                                 1.05     12.9±0.87ms        ? ?/sec     1.00     12.3±0.59ms        ? ?/sec
bench_trivial_query/diesel/1                                                              1.00     53.9±2.94µs        ? ?/sec     1.06     57.4±6.60µs        ? ?/sec
bench_trivial_query/diesel/10                                                             1.01     62.1±5.43µs        ? ?/sec     1.00     61.6±4.37µs        ? ?/sec
bench_trivial_query/diesel/100                                                            1.08    143.8±9.79µs        ? ?/sec     1.00   133.5±11.52µs        ? ?/sec
bench_trivial_query/diesel/1000                                                           1.09   899.8±56.38µs        ? ?/sec     1.00   827.5±61.76µs        ? ?/sec
bench_trivial_query/diesel/10000                                                          1.20      9.0±0.48ms        ? ?/sec     1.00      7.5±0.40ms        ? ?/sec
bench_trivial_query/diesel_boxed/1                                                        1.00     55.8±5.43µs        ? ?/sec     1.05     58.9±4.47µs        ? ?/sec
bench_trivial_query/diesel_boxed/10                                                       1.02     63.6±5.61µs        ? ?/sec     1.00     62.1±5.96µs        ? ?/sec
bench_trivial_query/diesel_boxed/100                                                      1.09    144.6±9.43µs        ? ?/sec     1.00    132.6±9.15µs        ? ?/sec
bench_trivial_query/diesel_boxed/1000                                                     1.12   913.3±70.87µs        ? ?/sec     1.00   815.3±63.91µs        ? ?/sec
bench_trivial_query/diesel_boxed/10000                                                    1.18      8.9±0.49ms        ? ?/sec     1.00      7.5±0.43ms        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1                                            1.00    107.3±7.62µs        ? ?/sec     1.01   108.0±10.37µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10                                           1.00    117.2±8.85µs        ? ?/sec     1.00   117.0±58.14µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/100                                          1.04   195.2±18.02µs        ? ?/sec     1.00   187.3±12.88µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1000                                         1.10   999.1±59.64µs        ? ?/sec     1.00   905.4±46.01µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10000                                        1.05      8.7±0.50ms        ? ?/sec     1.00      8.2±0.52ms        ? ?/sec
```

</details>

pksunkara:
I had observed that anything around 15% change in clap benchmarks are noise. GitHub seems to use shared runners or something, or maybe they use different capacity machines.

weiznich:
> I had observed that anything around 15% change in clap benchmarks are noise. GitHub seems to use shared runners or something, or maybe they use different capacity machines.

The results from the benchmarks above are in line with what I've measured locally. It's quite likely that diesels values are more stable than claps because Diesels Benchmarks tend do more work, as we are measuring database roundtrips. More work == more time == minor differences do not matter that much. Different runner configurations should not matter that much here, as we run each benchmark on the same runner for master + the changeset.

Ten0 (commented):
I haven't finished wrapping my head around all of this yet, but I now have a clearer view of how this interacts with other backends.

Anyway, here are some more thoughts :)

Given some more time I''ll end up being able to do a full in-depth review of this somewhere within the next 5 weeks.

Ten0:
> I had observed that anything around 15% change in clap benchmarks are noise.

On that topic, here's an enlightening talk: https://youtu.be/r-TLSBdHe1A (In particular the anecdotes until the 15th minute highlight well how random benchmarks can be, and why.)

diesel/src/pg/connection/result.rs
Ten0: It's actually documented that it's not needed: https://www.postgresql.org/docs/13/libpq-exec.html#LIBPQ-PQCLEAR

Ten0 (commented):
Ok so the only pending question for me is do we want to:
1. As is now, provide a single `load_iter()` function that provides an iterator which, in some implementations case, starts providing results before the query is even fully executed on the server, and in others does not, and we're responsible to choose for users which is best depending on the backend
2. Separate the two cases with an API that specifies whether it is a lazy load, or something that loads everything and then deserializes it, leaving that complexity to the user. (e.g. `load_iter()` vs `load_then_deserialize_iter()` where we would currently implement `load_iter()` for mysql and sqlite, and `load_then_deserialize_iter()` for pg)

Pros & Cons:
 - 1 is simpler than 2
 - 2 is more extensible than 1 (we can add support for both row-by-row mode and non-row-by-row mode in parallel in a backwards-compatible manner)
   + That might be useless if the lack of buffer reuse in libpq's row-by-row mode is actually not a significant performance issue.
 - Downside for 2: Unless I'm mistaken, we will never implement non-lazy version for mysql and sqllite

It looks like we may want to do 2 if libpq's row-by-row mode is significantly less performant than getting the single result in a batch, because that will enable us to solve [the original issue](https://github.com/diesel-rs/diesel/discussions/2707#discussion-3311186).

However if that's not the case, it looks like doing 1 and moving to always-row-by-row mode would be a better approach.

Note: I employed the term "performance" above as if it was something with total ordering but did not specify it. It may actually be more costly in terms of CPU resources to the server to make more allocations, but at the same time actually faster to answer to the caller if the server is not overloaded because it starts deserializing before having received everything. (Let's call these better ordered properties *load* and *speed*).
I feel like overall with a server coded in Rust, typically being faster than the database itself, and cloud technologies enabling us to easily have several stateless instances running, *speed* is typically more relevant than *load*.

With that in mind, it doesn't look so unlikely that row-by-row mode would be faster (thus more performant with regards to our main criteria) than default mode. So I think we want to give a go at implementing it to resolve that question.

That being said I don't think resolving this question should be a blocker to merge this with solution 1, as long as we take care to keep this in mind before releasing 2.0.

diesel/src/connection/mod.rs
```diff
@@ -27,8 +25,25 @@ pub trait SimpleConnection {
     fn batch_execute(&mut self, query: &str) -> QueryResult<()>;
 }
 
+/// This trait describes which cursor type is used by a given connection
+/// implementation. This trait is only useful in combination with [`Connection`].
+///
+/// Implementation wise this is a workaround for GAT's
+pub trait IterableConnection<'a, DB: Backend> {
```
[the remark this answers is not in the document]
weiznich: Because otherwise we cannot put the right trait constraints on the `Row` associated type.

weiznich:
Thanks for the in depth review :+1:. I will try to address the comments next week.

About the general discussion about lazy loading or not for postgresql: What about the following approach: Instead of deciding that for each call to `load_iter` we could provide methods to configure that on connection level. This would allow us to decouple that "implementation specific detail" from the general API design. In the end that would result into having something like `PgConnection::set_loading_behavior` to switch between row by row and the fetch-all API.

diesel/src/sqlite/connection/functions.rs
```diff
@@ -117,34 +124,90 @@ where
...
 
 impl<'a> FunctionRow<'a> {
-    fn new(args: &'a [*mut ffi::sqlite3_value]) -> Self {
-        Self { args }
+    fn new(args: &mut [*mut ffi::sqlite3_value]) -> Self {
+        let lenghts = args.len();
+        let args = unsafe {
+            Vec::from_raw_parts(
+                // This cast is safe because:
+                // * Casting from a pointer to an arry to a pointer to the first array
```
weiznich, line 157:
```suggestion
                // * Casting from a pointer to an array to a pointer to the first array
```

Issue #2810, opened 2021-06-19
How to check if a transaction returned a rollback error?


keepsimple1:
### Discussed in https://github.com/diesel-rs/diesel/discussions/2801

<div type='discussions-op-text'>

<sup>Originally posted by **keepsimple1** June  3, 2021</sup>
https://github.com/diesel-rs/diesel/blob/3e8fd4d7ce257830f1735e9a31c9a5f0fb0d777f/diesel/src/connection/mod.rs#L52-L54

Referring to the above doc comment, how do we check if a transaction hit the second case, i.e. returned an error from the rollback?
</div>

PR #2811, opened 2021-06-19
Add RollbackError for failed transaction rollbacks.

keepsimple1:
This is to resolve issue #2810 .

Test case is constructed using sqlite rollback specifics but it verifies the general logic of the patch.  (Also because I don't have PostgreSQL / MySQL locally.)


diesel_tests/tests/transactions.rs
```diff
@@ -55,6 +55,34 @@ fn transaction_is_rolled_back_when_returned_an_error() {
     drop_test_table(connection, test_name);
 }
 
+#[test]
+#[cfg(feature = "sqlite")]
```
hi-rustin, line 59:
```suggestion
```

Should we run this test on both mysql and pg?

hi-rustin (commented):
Thanks for open the PR!

rest LGTM.

diesel_tests/tests/transactions.rs
keepsimple1: Based on the [CI result](https://github.com/diesel-rs/diesel/actions/runs/953354716) of the first commit, sqlite and mysql are good. `pg` failed the test. 

- for sqlite, I am confident that the test works as expected because of [sqlite3 doc](https://www.sqlite.org/lang_transaction.html) (see the section near the bottom).

- for mysql, I didn't find clear doc about rollback behavior, but the CI result shows the test works.

- for pg, I think the test didn't work because `pg` probably not treat it as error when a transaction is already rollback, based on my understanding of [their doc](https://www.postgresql.org/docs/9.4/sql-rollback.html). 

If we know for sure that mysql ROLLBACK command behaves the same as sqlite3, in the regard of executing 2nd time would trigger an error, then I can add mysql into the feature list here, which will be better than only having sqlite.  Or if there is another (better) way to cause a rollback fail for all of three backends,  I will certainly adopt.

keepsimple1: I've added more comments to describe the rationale for the test.

weiznich: Seems reasonable for me :+1:

Issue #2814, opened 2021-06-25
Generate SQL type definitions for unknown types generate one sql_types module per sql type


Ten0:
https://github.com/diesel-rs/diesel/commit/b048d6edc4abefc771163469191701cad2d32af4#diff-aa753cc8b4b14a2c300e8369ae4fc79342ecc199af968eeb17c684d7f5c417b5R259-R280

This generates multiple modules when there are several missing types, making compilation fail.

PR #2815, opened 2021-06-25
Only generate one sql_types module

weiznich:
Fixes  https://github.com/diesel-rs/diesel/issues/2814


diesel_cli/src/print_schema.rs
```diff
@@ -246,17 +246,30 @@ impl Display for CustomTypeList {
         if self.types.is_empty() {
             return Ok(());
         }
+        let mut out = None;
```
Ten0, line 249: How about we just create the module here since we just checked that `!self.types.is_empty()`?
That would remove the need for `out` being an Option and for the `first` stuff.
That changes the behaviour for non-postgres backends but they `unreachable!()` anyway, and I'm not even sure we can really say it's most likely they wouldn't need an `sql_types` module should we ever give them some form of support for custom types.

Ten0 (commented):
Otherwise LGTM :)

PR #2820, opened 2021-07-08
AssumeNotNull expression support

Ten0:
Resolves #2819

### DIFF
--- a/diesel/src/expression/assume_not_null.rs
+++ b/diesel/src/expression/assume_not_null.rs
@@ -34,8 +34,6 @@ where
     }
 }
 
-/// AssumeNotNull can be used in where clauses everywhere, but can only be used in
-/// select clauses for outer joins.
 impl<T, QS> AppearsOnTable<QS> for AssumeNotNull<T>
 where
     T: AppearsOnTable<QS>,
--- a/diesel/src/expression/assume_not_null.rs
+++ b/diesel/src/expression/assume_not_null.rs
@@ -1,0 +1,60 @@
+use crate::backend::Backend;
+use crate::expression::TypedExpressionType;
+use crate::expression::*;
+use crate::query_builder::*;
+use crate::query_source::joins::ToInnerJoin;
+use crate::result::QueryResult;
+use crate::sql_types::{DieselNumericOps, IntoNotNullable};
+
+#[derive(Debug, Copy, Clone, DieselNumericOps, ValidGrouping)]
+pub struct AssumeNotNull<T>(T);
+
+impl<T> AssumeNotNull<T> {
+    pub fn new(expr: T) -> Self {
+        AssumeNotNull(expr)
+    }
+}
+
+impl<T> Expression for AssumeNotNull<T>
+where
+    T: Expression,
+    T::SqlType: IntoNotNullable,
+    <T::SqlType as IntoNotNullable>::NotNullable: TypedExpressionType,
+{
+    type SqlType = <T::SqlType as IntoNotNullable>::NotNullable;
+}
+
+impl<T, DB> QueryFragment<DB> for AssumeNotNull<T>
+where
+    DB: Backend,
+    T: QueryFragment<DB>,
+{
+    fn walk_ast(&self, pass: AstPass<DB>) -> QueryResult<()> {
+        self.0.walk_ast(pass)
+    }
+}
+
+/// AssumeNotNull can be used in where clauses everywhere, but can only be used in
+/// select clauses for outer joins.
+impl<T, QS> AppearsOnTable<QS> for AssumeNotNull<T>
+where
+    T: AppearsOnTable<QS>,
+    AssumeNotNull<T>: Expression,
+{
+}
+
+impl<T: QueryId> QueryId for AssumeNotNull<T> {
+    type QueryId = T::QueryId;
+
+    const HAS_STATIC_QUERY_ID: bool = T::HAS_STATIC_QUERY_ID;
+}
+
+impl<T, QS> SelectableExpression<QS> for AssumeNotNull<T>
+where
+    Self: AppearsOnTable<QS>,
+    QS: ToInnerJoin,
+    T: SelectableExpression<QS::InnerJoin>,
+{
+}
+
+impl<T> SelectableExpression<()> for AssumeNotNull<T> where Self: AppearsOnTable<()> {}
--- a/diesel/src/expression_methods/global_expression_methods.rs
+++ b/diesel/src/expression_methods/global_expression_methods.rs
@@ -442,9 +442,7 @@ pub trait ExpressionMethods: Expression + Sized {
     /// This is meant to cover for cases where you know that given the `WHERE` clause
     /// the field returned by the database will never be `NULL`.
     ///
-    /// This **will cause runtime errors** on `load()` if the "assume" turns out to be incorrect.  
-    /// If incorrect in `Nullable` tuples, this **will cause silent fails** (see second
-    /// example).
+    /// This **will cause runtime errors** on `load()` if the "assume" turns out to be incorrect.
     ///
     /// # Examples
     /// ## Normal usage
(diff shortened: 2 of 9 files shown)


Ten0:
Compile tests fail seems to be unrelated.

Mingun:
Because diesel tries to provide compile-time safety of all SQL created by builder, it can be worth to enable this feature only with special feature flag, disabled by default. Then nobody cannot use it accidentally

pksunkara (approved):

Ten0:
> Because diesel tries to provide compile-time safety of all SQL created by builder, it can be worth to enable this feature only with special feature flag, disabled by default. Then nobody cannot use it accidentally

I'd favor not doing this, for two reasons:
 1. [The use-case described in the original discussion](https://github.com/diesel-rs/diesel/discussions/2819#discussion-3449725) seems common enough that I expect this to become part of a common-enough usage that I wouldn't expect it to be hidden behind a feature.
 2. I feel like the naming of the function itself, as well as the documentation, already provide a good warning regarding what the behavior of that function is.

I believe Serde did what you suggested for their `impl Deserialize for Rc<T>` as its behavior may be surprising to some users, but I feel they were far from being as covered by point 2 as we are (because it was just the standard `Deserialize` impl).

diesel/src/expression/assume_not_null.rs
```diff
@@ -0,0 +1,60 @@
...
+    DB: Backend,
+    T: QueryFragment<DB>,
+{
+    fn walk_ast(&self, pass: AstPass<DB>) -> QueryResult<()> {
+        self.0.walk_ast(pass)
+    }
+}
+
+/// AssumeNotNull can be used in where clauses everywhere, but can only be used in
+/// select clauses for outer joins.
```
weiznich, line 38: Is there any reason for this restriction? I mean it's totally fine and valid to have something like this:
```rust
some_table::table.filter(some_table::some_nullable_column.is_not_null()).select((some_table::some_nullable_column.assume_not_null()).load(…)?
```

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -434,6 +434,80 @@ pub trait ExpressionMethods: Expression + Sized {
...
+    /// Converts this potentially nullable expression into one which will be **assumed**
+    /// to be not-null. This method has no impact on the generated SQL, however it will
+    /// enable you to attempt deserialization of the returned value in a non-`Option`.
+    ///
+    /// This is meant to cover for cases where you know that given the `WHERE` clause
+    /// the field returned by the database will never be `NULL`.
+    ///
+    /// This **will cause runtime errors** on `load()` if the "assume" turns out to be incorrect.  
+    /// If incorrect in `Nullable` tuples, this **will cause silent fails** (see second
+    /// example).
```
weiznich, line 447: I would skip this sentence, as:

* This is not a silent fail in my opinion, but a correct deserialisation
* That does not only happen for tuples

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -434,6 +434,80 @@ pub trait ExpressionMethods: Expression + Sized {
...
+    ///     use self::posts;
+    ///     use self::users;
+    ///     let connection = &mut establish_connection();
+    ///
+    ///     let data = posts::table.left_join(users::table)
+    ///         .select((users::id, posts::author_name.assume_not_null()).nullable())
+    ///         .load::<Option<(i32, String)>>(connection);
+    ///     // This will cause the `Option` to be `None` where the `left_join` succeeded but the
+    ///     // `author_name` turned out to be `NULL`, because `Option` deserialization relies
+    ///     // on encountering an unexpected `NULL` while deserializing the inner fields.
```
weiznich, line 504: I believe this documentation belongs to the `FromSqlRow` impl of `Option<T>` instead, as this can also happen in other constructed cases. I would rather include an example that really returns an error from `load()`

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -434,6 +434,80 @@ pub trait ExpressionMethods: Expression + Sized {
...
+    ///
+    /// fn main() {
+    ///     use self::posts::dsl::*;
+    ///     let connection = &mut establish_connection();
+    ///
+    ///     let data = posts
+    ///         .filter(author_name.is_not_null())
+    ///         .select(author_name.assume_not_null())
+    ///         .load::<String>(connection);
+    ///     println!("{:?}", data);
```
weiznich, line 472: Can we add an `assert!(data.ok())` here?

weiznich (commented):
I've added a few small comments on points I would like to see improved, otherwise this looks good :+1:

> Because diesel tries to provide compile-time safety of all SQL created by builder, it can be worth to enable this feature only with special feature flag, disabled by default. Then nobody cannot use it accidentally

As there are already a number of other methods in the query dsl that allow you to trigger a runtime error I see no reason to handle this method differently. (This already can happen for `sql_query`, `sql`, or as soon as you use a feature that is only supported on a newer version of that database system you are using)

> Compile tests fail seems to be unrelated.

Those are "releated" to this PR due to https://github.com/dtolnay/trybuild/issues/116

diesel/src/expression/assume_not_null.rs
Ten0: That's a mistake, it comes from the copy-pasting of `Nullable`.
I don't understand what it means in the context of `Nullable` though.

diesel/src/expression_methods/global_expression_methods.rs
Ten0: I'm not sure where we could put that comment about `FromSqlRow`.
Also it seems important to me to underline this behavior implied by the usage of `assume_not_null()`, as it's the one that feels the most of a trap to me: if you were wrong about "assuming" that `author_name` (unrelated to the left join) was not null, then instead of having the option set iff the left join succeeded, it will be `None` when it was "assume"d wrong.

diesel/src/expression/assume_not_null.rs
weiznich: I believe the original comment meant to express that you only can use `.select(some::column.nullable())`  on a query that actually does a left join. I'm not sure if that statement remains true after the `SqlType` rewrite.

diesel/src/expression_methods/global_expression_methods.rs
weiznich: > I'm not sure where we could put that comment about FromSqlRow... ?
Also it seems important to me to underline this behavior implied by the usage of assume_not_null(), as it's the one that feels the most of a trap to me: if you were wrong about "assuming" that author_name (unrelated to the left join) was not null, then instead of having the option set iff the left join succeeded, it will be None when it was "assume"d wrong.

We could put that directly on the implementation of `FromSqlRow for Option<T>` and link that from here. Would that be an option?

diesel/src/expression/assume_not_null.rs
Ten0: I don't think that is actually true anymore.

diesel/src/expression_methods/global_expression_methods.rs
Ten0: So if I understand correctly that would actually be [here](https://docs.diesel.rs/master/diesel/deserialize/trait.FromStaticSqlRow.html#impl-FromStaticSqlRow%3CNullable%3C(SA%2C)%3E%2C%20__DB%3E-for-Option%3C__T%3E) with the documentation repeated for every tuple size, as that is actually defined here:
https://github.com/diesel-rs/diesel/blob/cfb95067153ba55dff556c0b580cbc0e47f3fd28/diesel/src/type_impls/tuples.rs#L259-L287

Or maybe we want to add it directly on the `FromStaticSqlRow` or `Queryable` trait?

weiznich: That's a good point. Adding it to the general documentation of `Queryable` seems a good solution. (I would add it there, as that's the trait most users interact with.)

Ten0: a7fae660e9

PR #2828, opened 2021-07-14
Call rollback_transaction instead of begin_transaction when rolling back

dchenk:
In https://github.com/diesel-rs/diesel/commit/cfb95067153ba55dff556c0b580cbc0e47f3fd28 in the file `diesel/src/r2d2.rs` the `rollback_transaction` function implementing `TransactionManager` was changed to call `T::begin_transaction` instead of `T::rollback_transaction`. This appears to be a copy-paste error, and it's causing problems because errors returned by the function (closure) passed to `transaction` are being ignored and, instead of the transaction being rolled back, a new savepoint is being created.

To reproduce, start a new project with just the following dependency:
```
diesel = { git = "https://github.com/diesel-rs/diesel", rev = "0aa0da7941289fc0a4f6acd866e0a6913c87979d", features = ["chrono", "postgres", "r2d2", "serde_json"] }
```

And the following `main.rs` file:
```
#[macro_use]
extern crate diesel;

use diesel::{Connection, PgConnection, r2d2, RunQueryDsl};

mod schema;

const DB_URL: &str = "<REDACTED>";

#[derive(Debug)]
enum TxnError {
    Internal(diesel::result::Error),
    IDExists,
}

impl From<diesel::result::Error> for TxnError {
    fn from(err: diesel::result::Error) -> Self {
        TxnError::Internal(err)
    }
}

fn main() {
    let db_pool = r2d2::Pool::builder()
        .max_size(5)
        .build(r2d2::ConnectionManager::<PgConnection>::new(DB_URL))
        .unwrap();

    let mut conn = db_pool.get().unwrap();

    conn.transaction::<(), TxnError, _>(|conn| {
        use diesel::{insert_into, prelude::*};
        use schema::a::dsl::{a, id};

        insert_into(a)
            .values(&id.eq(123))
            .execute(conn)?;

        Ok(())
    })
        .unwrap();

    let _ignored = conn.transaction::<(), TxnError, _>(|conn| {
        Err(TxnError::IDExists)
    });
}
```

And this migration:
```
create table a (
  id int primary key
);
```

With the following for `diesel.toml`:
```
[print_schema]
file = "src/schema.rs"
```

You get a `src/schema.rs` file generated when you run the migration.

Start a Postgres server configured to _log all queries/statements_.

Do `cargo run`.  Check the Postgres query logs, and notice that we have the following queries:
- `BEGIN`
- `INSERT INTO "a" ("id") VALUES ($1)`
- `COMMIT`
- `BEGIN`
- `SAVEPOINT diesel_savepoint_1`

But there is no `ROLLBACK` for the second transaction.

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -196,7 +196,7 @@ where
     }
 
     fn rollback_transaction(conn: &mut PooledConnection<M>) -> QueryResult<()> {
-        T::begin_transaction(&mut **conn)
+        T::rollback_transaction(&mut **conn)
     }
 
     fn commit_transaction(conn: &mut PooledConnection<M>) -> QueryResult<()> {


weiznich (approved):
Thanks for submitting this PR. :+1:

dchenk:
@weiznich no problem. Could you make a crate release after merging this?

weiznich:
> Could you make a crate release after merging this?

We plan to release all changes on the master branch as breaking 2.0 release someday. As some of the changes we want to do are missing (see the [2.0 milestone](https://github.com/diesel-rs/diesel/milestone/16/) for details) we likely won't do a release now. I cannot give you any ETA here, as this is a free time project for me. That means the release will be done as soon as we've already in place.

mattdowdell:
Came to say big thanks for finding and fixing this; it saved me a bunch of time when I hit this yesterday.

For anyone else who hits this and like me was on a commit between the breakage and fix, this manifests with a `RollbackError`, wrapping the internal `Unknown` error, with a message of "current transaction is aborted, commands ignored until end of transaction block" set in the accompanying info.

For me that came from hitting a `UniqueViolaton` on an INSERT which correctly produced an error but then got morphed into an Unknown error due to the bug identified here. I'd imagine it would also occur for any error originating from PostgreSQL that was returned from the transaction.

weiznich:
@mattdowdell Just to clarify a few points here:
* That's definitively the risk of using a unreleased version. There is an reason why we don't have cut a release yet.
* Any help to prevent such bugs in the future are welcome. This includes:
    * Joining the diesel-rs/reviewers team and review PR's
    * Help expanding the test suite to cover more parts of the API
    * Somehow measure the test coverage to find missing spots like this one

mattdowdell:
Completely agree with the implied risk of using an unreleased version, I'm using this for a personal project so I don't mind finding little issues like this. My intention with my previous message was to add an error message which can be found via search for those who are taking on that risk, and aren't religiously upgrading to latest.

I had semi-automated integration test that found this in my code,  but I'll see if I can find some time to contribute a test to prove this fix works. Out of interest, does diesel have any policies/practices around adding tests to prove bug fixes in a TDD-like fashion?

weiznich:
> Completely agree with the implied risk of using an unreleased version, I'm using this for a personal project so I don't mind finding little issues like this. My intention with my previous message was to add an error message which can be found via search for those who are taking on that risk, and aren't religiously upgrading to latest.

:+1:

> I had semi-automated integration test that found this in my code, but I'll see if I can find some time to contribute a test to prove this fix works. 

That would be great. That said I'm sure that this PR fixes the problem, as this is a quite obvious bug :wink:

> Out of interest, does diesel have any policies/practices around adding tests to prove bug fixes in a TDD-like fashion?

We have a quite large integration test suite in `diesel-tests`, some unit tests in diesel itself and quite a lot documentation tests. This means large parts of the API are completely covered with tests, but unfortunately not everything. I generally only accept PR's with new features if they contain at least some tests for the feature.

PR #2832, opened 2021-07-21
CLI - Don't silent fail when failing to read migration directory

Ten0:
The `migrations_directories` from `migration_internals` silently dropped failing to read the migrations directory by calling `into_iter()` on the result of `read_dir()`.

The error is now propagated as an extra `Result` level on these functions, because keeping the same signature would require putting these errors inside the iterator, which would have required adding a dependency on `either` for the implementation to not be unreasonable.


weiznich (approved):
Looks good for me if CI errors are fixed :+1:

PR #2835, opened 2021-07-24
Supporting Const-Generics on derives for FromSqlRow and AsExpression

### LINKED ISSUE #2834: Compilation error with const-generic in custom sql type
### Versions

- **Rust:** 1.53.0
- **Diesel:** 1.4.7
- **Database:** Postgres
- **Operating System** Ubuntu x86_64

### Feature Flags

- **diesel:** [ "postgres" ]

## Problem Description
Compilation error occurs when trying to build custom sql type with a const-generic.

Using diesel to derive a rust generic sized array stored in postgres backed by a postgres array.

Compilation occurs successfully

Compilation fails when trying to derive `AsExpression`.

No. I can implement `ToSql<Array<Float8>, Pg>` and `FromSql<Array<Float8>, Pg>` without errors as well.

### Steps to reproduce

Create a new project with the following files:

`**src/main.rs**`
```rust
use diesel::{
    sql_types::{Array, Float8},
    AsExpression,
};

#[derive(AsExpression)]
#[sql_type = "Array<Float8>"]
struct DoubleArray<const N: usize>([f64; N]);

fn main() {
    println!("Hello, world!");
}

```

`**Cargo.toml**`
```toml
[package]
name = "test-rs"
version = "0.1.0"
edition = "2018"

[dependencies]
diesel = { version = "1.4" , features = ["postgres"] }
```

Execute: `cargo check` and see the following errors

> error[E0412]: cannot find type `N` in this scope

> error[E0747]: unresolved item provided when a constant was expected

I can also provide additional files like `schema.rs` if needed, but it doesn't seem necessary to produce the bug.

## Checklist

- [ x ] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [discussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
- [ x ] This issue can be reproduced on Rust's stable channel. (Your issue will be
- [ x  ] This issue can be reproduced without requiring a third party crate

msdrigg:
This pull request updates the diesel_derives macros for deriving the FromSqlRow and the AsExpression traits to allow for const-generics. In the FromSqlRow, there were issues with the order of standard generics and const generics. In the AsExpression derivation, there was a const-generic missing totally. 

This pull request does not build with the standard rust-toolchain of 1.48 but it does build with 1.51 and the test succeeds.

This won't be able to merge until diesel upgrades to rust 1.51, so it is currently blocked. I spoke with @weiznich about this in issue #2834. This  pr should also totally resolve #2834.

### DIFF
--- a/diesel_derives/tests/as_expression.rs
+++ b/diesel_derives/tests/as_expression.rs
@@ -1,0 +1,89 @@
+use diesel::backend::Backend;
+use diesel::deserialize::FromSql;
+use diesel::expression::AsExpression;
+use diesel::serialize::{Output, ToSql};
+use diesel::*;
+use std::convert::TryInto;
+use std::io::Write;
+
+use helpers::connection;
+
+table! {
+    my_structs (foo) {
+        foo -> Integer,
+        bar -> Text,
+    }
+}
+use diesel::sql_types::Text;
+#[derive(Debug, AsExpression, FromSqlRow, Clone, Copy, PartialEq)]
+#[sql_type = "Text"]
+struct StringArray<const N: usize>(pub [u8; N]);
+
+impl<DB, const N: usize> FromSql<Text, DB> for StringArray<N>
+where
+    DB: Backend,
+    String: FromSql<Text, DB>,
+{
+    fn from_sql(bytes: backend::RawValue<DB>) -> deserialize::Result<Self> {
+        let string = <String as FromSql<Text, DB>>::from_sql(bytes)?;
+        let bytes_array: [u8; N] = string.into_bytes().try_into().unwrap();
+        Ok(StringArray(bytes_array))
+    }
+}
+
+impl<DB, const N: usize> ToSql<Text, DB> for StringArray<N>
+where
+    DB: Backend,
+    String: ToSql<Text, DB>,
+{
+    fn to_sql<W: Write>(&self, out: &mut Output<W, DB>) -> serialize::Result {
+        let string = std::str::from_utf8(&self.0).unwrap().to_owned();
+
+        string.to_sql(out)
+    }
+}
+
+#[test]
+fn struct_with_sql_type() {
+    #[derive(Debug, Clone, PartialEq, Queryable, Selectable)]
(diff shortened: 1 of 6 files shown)


msdrigg:
Also, @weiznich, I only included one test where I derived AsExpression and FromSqlRow in a struct and then used a parent struct in a select command. This doesn't seem like it is the simplest way to test these derivations, but I don't have a good idea for how to test the AsExpression and FromSqlRow behavior directly. Do you have any ideas?

diesel_derives/tests/tests.rs
```diff
@@ -15,3 +15,5 @@ mod insertable;
 mod queryable;
 mod queryable_by_name;
 mod selectable;
+
+mod as_expression;
```
Ten0, line 19: why is this separated from the rest?

Ten0 (approved):

diesel_derives/tests/as_expression.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    let conn = &mut connection();
+    let data = my_structs::table
+        .select(MyStruct::as_select())
+        .get_result(conn);
+    assert!(data.is_err());
+}
+
+// #[test]
```
weiznich, line 62: Any reason why this code is commented out?

msdrigg: I left that code in while waiting for feedback on my test case. 

But it is unnecessary to the pr, so I have removed it for the time being.

diesel_derives/tests/tests.rs
msdrigg: There is no reason. I have fixed this.

diesel_derives/tests/tests.rs
```diff
@@ -15,3 +15,4 @@ mod insertable;
 mod queryable;
 mod queryable_by_name;
 mod selectable;
+mod as_expression;
```
JohnTitor, line 18: Seems we sort modules alphabetically, it should be a good idea to follow it.

PR #2846, opened 2021-08-09
Updated CLI readme with additional instructions

tim-harding:
Added a `toml` snippet to help Windows users with installation problems. This difficulty with the linker has resurfaced many times in the project bug reports and I think it would be suitable to add to the documentation.

### DIFF
--- a/diesel_cli/README.md
+++ b/diesel_cli/README.md
@@ -12,7 +12,15 @@ The diesel cli, by default, requires [`openssl`][openssl], [`libpq`][postgres],
 you can run `cargo install diesel_cli`.
 
 > Note: Make sure that both the `bin` and `lib` directories for
-> postgres are added to your PATH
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for submitting this PR.
We generally try not to have operating system specific installation instructions somewhere for the following reasons:
* They tend to be out of date quite soon (especially for windows)
* There always will be another operating system we don't provide instructions for 
* In the end they are really specific to the environment the author used. (For this example: What is about other database systems, other install locations and so on).

Additionally all these things are documented in a general way in the readme's of the crates actually linking the corresponding libraries:
* [pq-sys](https://github.com/sgrif/pq-sys)
* [mysqlclient-sys](https://github.com/sgrif/mysqlclient-sys)
* [libsqlite3-sys](https://github.com/rusqlite/rusqlite/tree/master/libsqlite3-sys)

PR #2849, opened 2021-08-12
Using GitHub actions concurrency mode to reduce CI queue time

hi-rustin:
In my past contributions, I've found that GitHub actions are heavily queued because they are under-resourced. This PR will use concurrency mode to ensure that only one commit on each PR will run tests. This way when we do multiple commits, the tests from the previous commit will automatically break and only the tests from the last commit will run. This will greatly save our testing resources. Especially if the contributor is unable to stop the GitHub actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,11 @@ on:
 
 name: CI Tests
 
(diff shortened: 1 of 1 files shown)


hi-rustin:
My PR itself is an example of this. I added two commits in a short period of time and the first one would be quickly cancelled.

image

image

weiznich (approved):
Sounds like a great addition :+1:

PR #2853, opened 2021-08-15
support `serialize_as` for `AsChangeset`

flisky:
model after `Insertable`'s `serialize_as` in #2553 , cc @thomaseizinger .

### DIFF
--- a/diesel_derives/src/as_changeset.rs
+++ b/diesel_derives/src/as_changeset.rs
@@ -1,6 +1,7 @@
 use proc_macro2::TokenStream;
-use syn::{DeriveInput, Path};
+use syn::{DeriveInput, Expr, Path, Type};
 
+use attrs::AttributeSpanWrapper;
 use field::Field;
 use model::Model;
 use util::{inner_of_option_ty, is_option_ty, wrap_in_dummy_mod};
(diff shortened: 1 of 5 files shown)


diesel_derives/src/as_changeset.rs
```diff
@@ -58,21 +40,58 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
+                &ty,
+            ));
+            direct_changeset_expr.push(field_changeset_expr_serialize_as(
+                field,
+                &table_name,
+                treat_none_as_null,
+                &ty,
+            ));
+
+            generate_borrowed_changeset = false; // as soon as we hit one field with #[diesel(serialize_as)] there is no point in generating the impl of AsChangeset for borrowed structs
```
pksunkara, line 65: Why not?

diesel_derives/src/as_changeset.rs
```diff
@@ -58,21 +40,58 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
...
-            type Changeset = <(#(#ref_changeset_ty,)*) as AsChangeset>::Changeset;
+    let mut direct_changeset_ty = Vec::with_capacity(fields_for_update.len());
+    let mut direct_changeset_expr = Vec::with_capacity(fields_for_update.len());
+    let mut ref_changeset_ty = Vec::with_capacity(fields_for_update.len());
+    let mut ref_changeset_expr = Vec::with_capacity(fields_for_update.len());
 
-            fn as_changeset(self) -> Self::Changeset {
-                (#(#ref_changeset_expr,)*).as_changeset()
-            }
+    for field in fields_for_update {
```
pksunkara, line 50: Don't like the fact that we moved away from `map`.

thomaseizinger (commented):
LGTM!

FWIW, I wouldn't bother trying to reuse code between the macros for now.

diesel_derives/src/as_changeset.rs
flisky: The conversion(`Into`) takes ownership. It would ask for a `Clone` on `serialize_as` fields if we generate the borrowed implementation. Any thought?

thomaseizinger: We could require `AsRef` instead of `Into` as a trait bound.

flisky: So we would require **two** traits to be implemented: `Into` for owned `AsChangeset`, and `AsRef` for borrowed one?

weiznich: @pksunkara Let me just summarize the options here shortly:

* We could require only an impl like `Into<Ty> for &MyType`, but that wouldn't allow users to reuse allocations 
* We could require both `Into<Ty> for MyType` and `Into<Ty> for &MyType`, that would require the user to provide both impls as there is no way to detect which impls actually exist from inside of a custom derive
* We could require that `Ty: Clone`, which would potentially hide allocations.
* We could require only `Into<Ty> for MyType` and only generate `Insertable<…> for Struct` (and not the reference variant). This pushes potential users into the "right" direction to reuse allocations + makes potential clones explicit

At least I feel that the last option is that one most in line with rusts design ideals.

flisky: There's no way to check whether trait is implemented in runtime, right? So we need user pass the flag in (maybe through `serialize_as`'s attribute)?

Oh, yes. It seems **`AsRef` doesn't work**:( we cannot returns an owned object's reference.

thomaseizinger: > We could require both `Into<Ty> for MyType` and `Into<Ty> for &MyType`, that would require the user to provide both impls as there is no way to detect which impls actually exist from inside of a custom derive

You can't emit code based on which type implements which trait, yes. But you can emit code that is itself conditional on whether a trait is implemented. We can emit something like:

```rust
impl<'a> AsChangeset for &'a T where Foo: AsRef<Bar> {

}
```

That should allow us to call `foo.as_ref()` inside the impl block. If a user doesn't need this conversion and hence doesn't implement `AsRef`, things should still work, but they will get a compile error of unresolved trait bounds in case they try to pass a reference of `T` in.

I am not sure we need it though.

weiznich: @thomaseizinger Restricting the impl in that way is unfortunately not possible. Rustc is too clever here and complains that `Foo` does not implement `AsRef<Bar>`. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d1f5277fc1baa7f5c98d530df73ae448) for an example.

thomaseizinger: Bugger, thanks for clearing it up.

pksunkara:
This needs rebase on the new derive code that's merged recently

flisky:
@pksunkara rebased.

Very weird the CI keeps failling compile, but I cannot reproduce locally and using `act -j compile_tests`, and I cannot find where `AttributeSpanWrapper` comes from. I really appreciate if someone would take a look.

pksunkara:
That's the latest master we merged recently again.

flisky:
@pksunkara thanks for the hint, and I didn't realize that CI tests on the after-merge-master code.

diesel_derives/src/lib.rs
```diff
@@ -86,6 +97,10 @@ mod valid_grouping;
 /// * `#[diesel(column_name = some_column_name)]`, overrides the column name
 ///    of the current field to `some_column_name`. By default the field
 ///    name is used as column name.
+/// * `#[diesel(serialize_as = "SomeType")]`, instead of serializing the actual
```
weiznich, line 100:
```suggestion
/// * `#[diesel(serialize_as = SomeType)]`, instead of serializing the actual
```

diesel_derives/src/lib.rs
```diff
@@ -61,6 +61,17 @@ mod valid_grouping;
 /// from the name of the corresponding column, you can annotate the field with
 /// `#[diesel(column_name = some_column_name)]`.
 ///
+/// To provide custom serialization behavior for a field, you can use
+/// `#[diesel(serialize_as = "SomeType")]`. If this attribute is present, Diesel
```
weiznich, line 65:
```suggestion
/// `#[diesel(serialize_as = SomeType)]`. If this attribute is present, Diesel
```

diesel_derives/tests/as_changeset.rs
```diff
@@ -218,6 +218,57 @@ fn with_explicit_column_names() {
     assert_eq!(Ok(expected), actual);
 }
 
+#[test]
+fn with_serialize_as() {
+    #[derive(Debug, FromSqlRow, AsExpression)]
+    #[sql_type = "sql_types::Text"]
```
weiznich, line 224:
```suggestion
    #[diesel(sql_type = sql_types::Text)]
```

weiznich (approved):

diesel_derives/src/as_changeset.rs
weiznich: At least for me that looks ok. The corresponding block would be quite complex if put into a `.map()`

Issue #2322, opened 2020-03-08
Add `DatabaseErrorKind` variants for connection errors


LLFourn:
On 1.4.3, if you lose your connection to a postgres database and try to make a query you will get:

```
Err(
    DatabaseError(
        __Unknown,
        "server closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n",
    ),
)
```

or

```
Err(
    DatabaseError(
        UnableToSendCommand,
        "no connection to the server\n",
    ),
)
```

For obvious reasons, It would be useful to be able to identify a bad connection.

weiznich:
We are happy to accept a PR adding this error variant

henryboisdequin:
I would like to work on this issue.

weiznich:
@henryboisdequin Feel free to go for it :+1:

weiznich:
@henryboisdequin Feel free to go for it :+1:

hi-rustin:
@weiznich @LLFourn How can I reproduce this problem? Because I want to determine the error code of it. Can you help me with the steps to reproduce it?

weiznich:
@hi-rustin I think one way to reproduce the problem is to kill/stop the postgres server while some query is running. That's probably something that's not that easy to time, though.

hi-rustin:
> @hi-rustin I think one way to reproduce the problem is to kill/stop the postgres server while some query is running. That's probably something that's not that easy to time, though.

This results in the following error: [08006] An I/O error occurred while sending to the backend.

LLFourn:
> 
> 
> @hi-rustin I think one way to reproduce the problem is to kill/stop the postgres server while some query is running. That's probably something that's not that easy to time, though.

I was stopping the server after a connection was established and then try and make a query which is easy to do. Doing it during a query might also yield a different error code.

hi-rustin:
@weiznich I reproduced the issue locally, but the error does not have ResultField::SqlState, maybe we need to match the message to add the error variant?

weiznich:
@hi-rustin Matching the error message is a possible solution, but only as long as the error message remains stable over various postgres versions.

hi-rustin:
> @hi-rustin Matching the error message is a possible solution, but only as long as the error message remains stable over various postgres versions.

I'll check tomorrow. Thanks for your reply!

hi-rustin:
Tested 13.4, 12.8, 11.13, 10.18, 9.6, all with the same error.

PR #2860, opened 2021-08-18
 Add ClosedConnection error kind

hi-rustin:
close https://github.com/diesel-rs/diesel/issues/2322

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -124,6 +124,9 @@ pub enum DatabaseErrorKind {
     /// A check constraint was violated.
     CheckViolation,
 
+    /// The connection to the server was unexpectedly closed.
+    ClosedConnection,
+
     #[doc(hidden)]
     Unknown, // Match against _ instead, more variants may be added in the future
 }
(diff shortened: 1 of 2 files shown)


hi-rustin:
Tested locally.

```sh
➜  diesel_demo git:(master) ✗ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/show_posts`
thread 'main' panicked at 'Error loading posts: DatabaseError(ClosedConnection, "server closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n")', src/bin/show_posts.rs:19:10
➜  diesel_demo git:(master) ✗ 
```

diesel/src/result.rs
```diff
@@ -124,6 +124,9 @@ pub enum DatabaseErrorKind {
     /// A check constraint was violated.
     CheckViolation,
 
+    /// The connection to the server was unexpectedly closed.
```
weiznich, line 127: I would like to have a comment here stating that we emit this error message on best effort basis, so that it is clear that we may miss situations where this occurs.

weiznich (approved):
Looks fine module the small documentation comment change.

diesel/src/result.rs
hi-rustin: Added. Thanks for your review!

PR #1990, opened 2019-02-18
Add Send bound to BoxedSelectStatement

mehcode:
Fixes #1625 

---

Motivation is to enable conditional queries to be executed asynchronously _now_. The current version of my async adapter ( see https://github.com/mehcode/actix-diesel ) requires the query to be `Send`.

Hoping that the incoming 2.0 bump will let us make these breaking changes.

PR #2881, opened 2021-08-27
[1.4.x] Backport "Send" bounds for BoxedSelectStatements

smklein:
Cherry picks https://github.com/diesel-rs/diesel/pull/1990 onto the 1.4.x branch.

I believe this is non-breaking - and already merged in main - but it helps Diesel users writing async adapters on the latest stable release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Updated `libsqlite3-sys` to allow version 0.22
 * Updated `ipnetwork` to allow version 0.18
+* Boxed queries (constructed from `.into_boxed()`) are now `Send`.
 
 ## [1.4.6] - 2021-03-05
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,6 @@ members = [
     "examples/mysql/getting_started_step_1",
     "examples/mysql/getting_started_step_2",
     "examples/mysql/getting_started_step_3",
-    # Dependency on out-dated bcrypt
-    #"examples/postgres/advanced-blog-cli",
     "examples/postgres/all_about_inserts",
     "examples/postgres/all_about_updates",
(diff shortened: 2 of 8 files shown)


smklein:
Hrm. So I'm noticing a lot of things breaking, which also seem to be breaking on the unchanged 1.4.x branch due to third party dependencies falling over.

... do you have advice for finding "what was the version of everything at the time of publishing 1.4.7", before I try guessing + setting upper bounds on various Cargo.toml files?

weiznich:
We cannot do this on 1.4.x as this is definitively a breaking change.
This PR adds additional trait bounds to public usable trait impl, so after this any type that does not fulfill these new bounds wont implement the trait anymore. Specifically this is the case for the `BoxedDsl` impl on `SelectStatement`. Any generic parameter ther could be `!Send` for existing code and is controlled by crate users. Releasing this change would stop these queries from compiling. That means we cannot release that as non major version bump.
On top of that I consider this as "adding a new feature". Our policy there is that we don't backport new features as we don't have the manpower to do development on two separate versions. The only things that I accept as backports are:
* Critical bugfixes
* Version bumps for important dependencies.

PS: The last time I've checked the CI was passing fine, so failures seems to be caused by this change.

smklein:
Ack, gotcha. Apologies for the noise; I wasn't clear on the policy. Sounds like I should try upgrading to 2.0 - even for non-native async support, this feature is basically mandatory for "make a query and send it to a synchronous thread for processing".

BTW, as a heads up, I checked out origin/1.4.x, I'm still seeing the following error (which was related to the "advancec-blog-cli", and I'm still fairly certain unrelated to this change):

```
 $ cargo build
    Updating crates.io index
error: no matching package named `block-cipher-trait` found                                                            
location searched: registry `https://github.com/rust-lang/crates.io-index`
required by package `blowfish v0.3.0`
    ... which is depended on by `bcrypt v0.2.0`
    ... which is depended on by `advanced-blog-cli v0.1.0 (/home/smklein/repos/diesel/examples/postgres/advanced-blog-cli)`
```

PR #2883, opened 2021-09-01
Upgraded num-bigint and bigdecimal dependencies

Razican:
This updates `num-bigint` to allow also for version 0.4, which has small changes and doesn't impact the build, and `bigdecimal`, which only updates its dependency on `num-bigint`.


JohnTitor:
CI complains of a trait bound failure.

Razican:
> CI complains of a trait bound failure.

It seems I forgot to update the dependency in diesel tests, should be working now.

xrl:
Can we get this cherry-picked to a `1.4.X` release?

weiznich:
@xrl Feel free to do that, if that's no breaking change. If we do another patch release I would include something like that, but I will likely spend no time to do the backport on my own or issue an release just for this.

Issue #1868, opened 2018-10-01
Support mysql's CLIENT_FOUND_ROWS flag


pjenvey:
This is a common flag to enable w/ mysql's unusual default behavior of returning the number of rows actually changed during an UPDATE (vs how many were matched). 

Some ORMs even default this flag to on for more consistency across dialects (e.g. [sqlalchemy](https://docs.sqlalchemy.org/en/latest/dialects/mysql.html#rowcount-support) and [activerecord](https://github.com/rails/rails/blob/aa3dcab/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L16)).

blackghost1987:
Thanks, this was bugging me for a while! This silly default behavior of MySQL can be really surprising when changing Diesel backends.

I think the relevant code is the last parameter of the `ffi::mysql_real_connect` call in `RawConnection::connect()`.
The big problem here is that Client Flags can't be encoded in the URL as it seems 😞 So we either put it in as a default, or we have to modify the public API (not just for MySql) to somehow pass these flags to a `Connection`.

The easiest way would be to make it a default, but I'm worried this would be a breaking change for some applications, maybe somebody is taking advantage of this behavior. I'm not saying it can't be done, but it probably requires a major version release?

I couldn't find a way to make this configurable, but here are the steps I came up with, maybe somebody better versed in Diesel could take a look and finish this plan:
1. create a bitflag for MySql Capability Flags as per https://dev.mysql.com/doc/internals/en/capability-flags.html
2. add the bitflag to `mysql::connection::ConnectionOptions` (make it optional, because it can't be parsed from the URL, so it had to be set later on)
3. use the bitflag in `RawConnection::connect()` as the mysql_real_connect() call's last parameter
4. create a way to somehow modify the `ConnectionOptions` instance AFTER parsing the URL but BEFORE calling `connect()`

The last part is the real catch here, because `ConnectionOptions::parse()` and `raw_connection.connect()` are both done in `establish(database_url: &str)`, which is defined in the `Connect` trait. So we can't add a parameter to `establish()` without adding it to every backend. Also it's a static constructor, so we cannot set a value in the instance previously either....

Any more ideas?

weiznich:
@blackghost1987 would be fine with setting this flag by default. Feel free to open a PR for this against the master branch.

blackghost1987:
@weiznich thanks for merging the PR into master! This issue probably should be closed as well

PR #2887, opened 2021-09-02
add CLIENT_FOUND_ROWS flag to mysql connect

blackghost1987:
Solves #1868 

This change enables the CLIENT_FOUND_ROWS flag by default, which makes MySQL UPDATES work like most other SQL DBs, and return the number of found rows, even if the UPDATE doesn't change any values in the row. This change could make it easier for people to change SQL dialects if they are unaware of this peculiarity of MySQL. Although if an application relies on MySQL's default behavior, then this change can cause issues, so I think this should be considered a breaking change (although very minor), and should be documented on release. @weiznich encouraged me to open this PR anyway (see [here](https://github.com/diesel-rs/diesel/issues/1868#issuecomment-911723565))

I've also added all possible CapabilityFlag values, not just CLIENT_FOUND_ROWS, but they are unused for now. Maybe they'll be useful in the future if we want to change the defaults again, or if we can come up with a way to make the flags configurable.


weiznich:

I would like to see a test case for this behavior so that we don't accidentally regress on that. Additionally this requires an entry in the "Changed" section of the  Changelog.

diesel/src/mysql/connection/url.rs
```diff
@@ -8,13 +8,44 @@ use std::ffi::{CStr, CString};
...
+}
+
 pub struct ConnectionOptions {
     host: Option<CString>,
     user: CString,
     password: Option<CString>,
     database: Option<CString>,
     port: Option<u16>,
     unix_socket: Option<CString>,
+    client_flags: Option<CapabilityFlags>,
```
Ten0, line 48: It looks like this is never `None`, and even if it was, it could probably be represented as not-an-option with the default value being that you'd unwrap to when sending it to `mysql_real_connect`, which always requires one
=> I don't think this should be an `Option`.
```suggestion
    client_flags: CapabilityFlags,
```

diesel/src/mysql/connection/url.rs
```diff
@@ -59,13 +90,17 @@ impl ConnectionOptions {
             Some(segment) => Some(CString::new(segment.as_bytes())?),
         };
 
+        // this is not present in the database_url, using a default value
```
Ten0, line 93: Maybe some people have a use for the old behavior though.
Perhaps ultimately we want to make the connection open-able from `ConnectionOptions` and implement [`FromStr`](https://doc.rust-lang.org/stable/std/str/trait.FromStr.html) on it, as well as some things to update the flags that don't break the typing, such as this one?
May be slightly drifting off-topic but I feel like the requirement of [`establish`](https://docs.diesel.rs/master/diesel/prelude/trait.Connection.html#tymethod.establish) on `Connection` is a bit too much: it's not super flexible as it requires that the url is the only input, and it has to be something that always errors on r2d2 connections... It feels like it's something each connection should implement separately.

Not a blocker for me though.

Ten0 (approved):
LGTM modulo the `Option` thing

diesel/src/mysql/connection/url.rs
blackghost1987: You're totally right, I wasn't sure if ConnectionOptions is constructed anywhere else in the codebase when I added this. Now I realize all the fields are private and the only constructor is `parse()`, so it's really unnecessary. If we want to disable this in the future we can just use CapabilityFlags::empty(), so I'll get rid of the Option.

diesel/src/mysql/connection/url.rs
blackghost1987: Yes, that was my impression as well, `establish()` is too restrictive, there's no way to manipulate ConnectionOptions outside of `parse()` currently. This is understandable of course, because ConnectionOptions is MySQL specific currently, so there could not be any ways to manipulate it through the Connection trait which is not database engine specific.
I'm not sure if there's a use for something like ConnectionOptions for other database engines, but if so, maybe there could be a non-db-specific ConnectionOptions trait in the future?

blackghost1987:
@weiznich I've added a test and edited the Changelog. The CI failure seems unrelated to me, any ideas?

weiznich:
@blackghost1987 Seems like a rustc issue: https://github.com/rust-lang/rust/issues/88617 + there should be a fix soon: https://github.com/rust-num/num-bigint/pull/219

weiznich:
@blackghost1987 Thanks for the update :+1:

PR #2901, opened 2021-09-16
Dev test

unitsoftrover:
test


weiznich:
Thanks for opening this PR. Can you add a description what the change want to achieve?

unitsoftrover:
I am making a backend to support odbc to sqlserver. I need to modify some in diesel. Now it just can run. this is a test PR.

weiznich:
I'm definitively open to include changes required to enable more third party backends. Its fine to open a test PR here, but in general I would prefer to have an separate PR per feature you want to see added. Such a PR should include some information why this change is required + needs to only contain that change and no other changes to the project setup + should be formated with rustc + proper documentation.

weiznich:
Closed as the author never resolved the necessary change requests. Feel free to update the PR, we can reopen it after that.

Issue #2753, opened 2021-05-02
Diesel migrations breaks diesel with the new feature resolver


ThouCheese:
### Versions

- rustc --version: `rustc 1.51.0 (2fd73fabe 2021-03-23)`
- **diesel**: `1.4.6`
- **diesel_migrations**: `1.4.0`
- **Postgres**: `psql (PostgreSQL) 13.2`
- **Linux**: 5.11.16 (Arch)

### Feature Flags

- **diesel**: `features = ["postgres"]`
- **diesel_migrations**:  No special features
- **cargo**: resolver = "2"

## Problem Description
When enabling the new resolver in `Cargo.toml` by specifying `resolver = "2"`, compiling `diesel` and `diesel_migrations` fails. In my problem I list postgres as the database, but it fails for `sqlite` and `mysql` too. I tried fiddling with the feature flags of diesel and diesel_migrations, but I couldn't pinpoint the exact problem.

### What are you trying to accomplish?
I'm trying to compile diesel and diesel migrations with the new resolver.

### What is the expected output?
I expect this to compile.

### What is the actual output?
It fails to compile with the following output:
```
error[E0432]: unresolved import `diesel`
  --> /home/me/.cargo/registry/src/github.com-1ecc6299db9ec823/diesel-1.4.6/src/sql_types/mod.rs:37:48
   |
37 | #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
   |                                                ^^^^^^^ could not find `pg` in `diesel`
   |
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Are you seeing any additional errors?
This error is repeated 26 times for various types in `sql_types/mod.rs`.

### Steps to reproduce
For easy repro, I have created a standalone repository: https://github.com/ThouCheese/diesel-issue. It contains mostly this:

#### `main.rs`
```rust
fn main() {}
```

#### `Cargo.toml`
```toml
[package]
name = "tmp"
version = "0.1.0"
edition = "2018"
resolver = "2"

[dependencies]
diesel = { version = "*", features = ["postgres"] }
diesel_migrations = { version = "*" }
```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
[...]

PR #2902, opened 2021-09-17
Ensure compatibility with 2021 edition by working around cargo featur…

weiznich:
…e resolver related breaking changes

Addresses https://github.com/diesel-rs/diesel/issues/2753 as this will unfortunately become the new default in the 2021 edition (regardless that this seems to be in conflict with any of the guarantees given which changes are allowed at edition boundary)
CC https://github.com/rust-lang/rust/issues/88903

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
           brew services start mysql@5.7 &&
           brew services stop mysql@5.7;sleep 3;brew services start mysql@5.7 &&
           sleep 2 &&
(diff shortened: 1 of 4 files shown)


pksunkara:
Am I right in understanding that we don't need this change for master?

weiznich:
The underlying problem happens because diesel is a direct and a build time dependency for diesel_migrations. The new feature resolver doesn't unify any features between those two diesel versions anymore, but it does unify features for diesel_derives. This leads to a situation where diesel_derives is generating code for the build time diesel version which depends on features that are disabled. 
This cannot happen on master anymore as we removed the build time dependency on diesel from diesel_migrations.

pksunkara (approved):

PR #2900, opened 2021-09-16
Allow to specialize `QueryFragment` impls easily

weiznich:
The main motivation of this change is to allow third party backends like
diesel-oci to speciaize `QueryFragment` impls without using the unstable
`specialization` feature. This is done by introducing a new marker trait
`SqlDialect` that has an associated type for each specializable
`QueryFragment` impl.

Beside of that this allows us to fix a few long standing issues:
* `eq_any` can now use `= ANY()` on postgres
* We can now support batch inserts for sqlite as long as there is no
default value in the insert values

Fixes #2898
Fixes #2694 (by deprecating the corresponding functions)

PR #2909, opened 2021-09-23
feat: Implement Clone for OwnedBatchInsert

frol:
This allows to clone queries that insert a vector of records (useful when you want to implement a retriable query).

P.S. Before this fix, I can clone a query which inserts a single value, but unable to clone a query when I insert a vec of values.

### DIFF
--- a/diesel/src/insertable.rs
+++ b/diesel/src/insertable.rs
@@ -293,7 +293,7 @@ where
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct OwnedBatchInsert<V, Tab> {
     pub(crate) values: Vec<V>,
     _marker: PhantomData<Tab>,


weiznich:
Thanks for opening this PR.

We generally do not accept PR's against the 1.4.x branch, especially if they contain "new" features. The only thing accepted in this branch are critical bug fixes + version bumps of dependencies.

frol:
@weiznich This is a minor change, so I hoped you might consider including it. I would consider moving to alpha releases, but there are none :disappointed:

weiznich:
Even if the change is only minor, we won't include it. We do not have the capacity to do development on more than one branch nor do we plan to issue another 1.4.x patch release any time soon.
If you want to have pre releases from the 2.0 branch consider contributing there by for example reviewing open PR's like #2799, #2900 or #2626.

PR #2911, opened 2021-09-24
[postgres] Match more postgres errors explicitly

smklein:
This PR originally attempted to avoid "string matching" for connection-based errors. This didn't work, and I documented my findings below.

Instead, this PR now simply attempts to handle more connection-based errors explicitly,
and provides some documentation about "why we're still doing string matching" as a catch-all for network-based errors.

-------

(These instructions are specific to the commit at https://github.com/diesel-rs/diesel/pull/2911/commits/1af6c15d520d06e333192937cdb0f14db114ea9c )

To repro, after checking out diesel and applying this PR:

- Run `$ docker-compose up`
- Run `$ cargo test --features "postgres"`. This starts the tests, including a test that just hangs out sleeping (I don't want to submit this test, but it's easy for repro-ing).
- Run `$ docker network disconnect diesel_default diesel.postgres`. This causes the sleeping test to fail, and shows what the result of a network disconnect through Diesel.

After doing this, I see the following output:

```
test pg::connection::tests::wait_a_looooong_time ... FAILED

failures:

---- pg::connection::tests::wait_a_looooong_time stdout ----
Unknown error code: None
thread 'pg::connection::tests::wait_a_looooong_time' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(ClosedConnection, "server closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n")', diesel/src/pg/connection/mod.rs:353:10
```

Indicating:
- The actual error code is none
- We only see a `ClosedConnection` error because of the equality check on the "server closed the connection..." message.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -343,6 +343,17 @@ mod tests {
     }
 
     #[test]
+    fn wait_a_looooong_time() {
+        let connection = &mut connection();
+
+        crate::sql_query(
+            "SELECT pg_sleep(100.0);",
+        )
+        .execute(connection)
+        .unwrap();
+    }
+
+    #[test]
     fn static_batch_inserts_are_cached() {
         let connection = &mut connection();
         connection.begin_test_transaction().unwrap();
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -343,17 +343,6 @@ mod tests {
     }
 
     #[test]
-    fn wait_a_looooong_time() {
-        let connection = &mut connection();
-
-        crate::sql_query(
-            "SELECT pg_sleep(100.0);",
-        )
-        .execute(connection)
-        .unwrap();
-    }
-
-    #[test]
     fn static_batch_inserts_are_cached() {
         let connection = &mut connection();
         connection.begin_test_transaction().unwrap();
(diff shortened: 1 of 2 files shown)


smklein:
This is not yet ready for review - I'm trying to use structured errors, but noticing that `PG_DIAG_SQLSTATE` sometimes returns `None` here.

This is odd - it's documented as [always present](https://www.postgresql.org/docs/current/libpq-exec.html#:~:text=PG_DIAG_SEVERITY,these.%20Always%20present.) in the postgres docs!

weiznich:
A quick search brought up [this](https://www.postgresql.org/message-id/E1ZPY5q-0000Ng-Oy@gemulon.postgresql.org) which sounds like it may be possible that there is just no SqlState associated with certain errors.

smklein:
> A quick search brought up [this](https://www.postgresql.org/message-id/E1ZPY5q-0000Ng-Oy@gemulon.postgresql.org) which sounds like it may be possible that there is just no SqlState associated with certain errors. That would mean that the documentation is wrong

Oof, that's a bummer. I updated the codebase to at least make this more obvious to the next person coming along, and handled some of the connection errors explicitly regardless, if Diesel gets the chance to encounter them.

pksunkara (approved):

PR #2932, opened 2021-10-27
Upgraded dependencies

Razican:
This PR upgrades all dependencies to their latest available versions. It removes all duplicate dependencies except for those coming from the "quickcheck" upgrade, tracked in #2930, and those coming from Clap 2.x, since 3.0 has not been released yet (https://github.com/clap-rs/clap/issues/2869).


Razican:
I have removed the update of the rand dependency and added it to the quickcheck update. This should fix the tests.

PR #2933, opened 2021-10-27
Support skipping fields in Queryable derive (falls back to Default impl)

Ten0:
Not sure if there was a feature request for this, but I happened to need it and felt it wouldn't be too hard to implement.

Does this seem reasonable? (We perhaps just want this also be available on the other derives?)

### DIFF
--- a/diesel_derives/src/field.rs
+++ b/diesel_derives/src/field.rs
@@ -91,6 +91,10 @@ impl Field {
         self.flags.has_flag(flag)
     }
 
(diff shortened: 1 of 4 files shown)


AJTJ:
Would love if this were to be picked up again :)

Ten0:
The problem here is not workforce but the decision on the relevant discussion. If you have *new* insights on the topic feel free to add them there. 🙂

Issue #2123, opened 2019-07-17
Connection state gets corrupted when an SQL error occurred in nested transactions.


KeenS:
### Versions

- **Rust:** rustc 1.36.0 (a53f9df32 2019-07-03)
- **Diesel:** 1.4.2
- **Database:** PostrgreSQL, psql (11.4 (Ubuntu 11.4-0ubuntu0.19.04.1), server 9.6.13)
- **Operating System** Ubuntu 19.04

### Feature Flags

- **diesel:** `"postgres"`

## Problem Description

When an SQL fails in nested transactions, the connection gets corrupted and all the other transaction running as SERIALIZABLE fails in `AlreadyInTransaction`.

### What are you trying to accomplish?

Just writing an application. Transaction is nested in an accident

### What is the expected output?

`Ok(1)`

### What is the actual output?

`Err(AlreadyInTransaction)`

### Are you seeing any additional errors?

postgresql output:

```
postgresql_1     | ERROR:  column "foo" does not exist at character 8
postgresql_1     | STATEMENT:  SELECT foo
postgresql_1     | ERROR:  current transaction is aborted, commands ignored until end of transaction block
postgresql_1     | STATEMENT:  RELEASE SAVEPOINT diesel_savepoint_1
```

### Steps to reproduce

run code below

Cargo.toml
```toml
[package]
name = "diesel-bug-report"
version = "0.1.0"
edition = "2018"

[dependencies]
diesel = {version = "1.4.0", features = ["postgres"]}
```

src/main.rs
```rust
use diesel::pg::PgConnection;
use diesel::prelude::*;
use diesel::result::Error;

fn main() {
    let conn = PgConnection::establish("postgres://user:password@localhost/database").unwrap();
    // fail once
    let ret = conn.transaction(|| {
        conn.transaction(|| {
            // handling error
            match conn.execute("SELECT foo") {
                // do nothing
                Ok(_) => (),
                // ignore the error
                Err(e) => eprintln!("error occurred: {}", e),
            };
            Ok::<_, Error>(())
        })
    });
    println!("{:?}", ret);
    // other transaction
    let ret = conn
        .build_transaction()
        .serializable()
        .run(|| conn.execute("SELECT 1"));
    // must be Ok(1), but get Err(AlreadyInTransaction)
    println!("{:?}", ret);
}
```

Then, gets output

```console
$ cargo run
error occurred: column "foo" does not exist
[...]

PR #2935, opened 2021-10-31
Proper depth tracking and robust broken transaction detection

Emm:
* Fixes the detection of errored transaction blocks in Postgres, without relying on the English version of a localized error message (related to #2123)
* Fixes connections unable to open new transactions in Postgres after a commit fails due to an `INITIALLY DEFERRED` constraint or trigger (which is quite dangerous as this breaks connection pools and is NOT detected by r2d2's `SELECT 1` check)


diesel/src/connection/mod.rs
```diff
@@ -39,8 +39,15 @@ pub trait ConnectionGatWorkaround<'a, DB: Backend> {
     type Row: crate::row::Row<'a, DB>;
 }
 
+/// Methods necessary for the transaction manager
+pub trait TransactionalConnection {
```
weiznich, line 43: I feel that adding a whole new trait for this is to complicated. I would like to see possible alternatives here. Optimally this should be handled inside of the transaction manager implementation only.

weiznich (commented):
I can understand your general approach on how to fix this issue, but I would like to not introduce an additional trait for this, especially if it just does nothing for 2 of 3 existing backends. Maybe we can find another way to just implement this in the scope of `AnsiTransactionManager` itself only?

Another way to fix the issue with the localized error messages is to just using `SET lc_messages TO 'en_US.UTF-8';` as part of the connection setup to force the error messages to be in English.

diesel/src/connection/mod.rs
Emm: @weiznich , relying on localized error messages is really fragile and [discouraged](https://www.postgresql.org/docs/14/errcodes-appendix.html) by the Postgres documentation:

> All messages emitted by the PostgreSQL server are assigned five-character error codes that follow the SQL standard's conventions for “SQLSTATE” codes. Applications that need to know which error condition has occurred should usually test the error code, rather than looking at the textual error message. The error codes are less likely to change across PostgreSQL releases, and also are not subject to change due to localization of error messages. Note that some, but not all, of the error codes produced by PostgreSQL are defined by the SQL standard; some additional error codes for conditions not defined by the standard have been invented or borrowed from other databases.

Forcing the error messages to be in English would also be potentially a breaking change for Diesel consumers, without compilation errors.

I can suggest two approaches if you don't like the additional trait:
- move `is_transaction_broken()` to the `Connection` trait
- strongly-typed "raw" error codes in `Error`, ie `RawError::Postgres(SomePostgresErrorCodeEnum)` - this would be a good idea in general, but it's also a lot more work

Emm: As a side note, do you know why the CI tests have no output? It's somewhat disturbing

Emm: We could also change the bounds like so:
`trait TransactionalConnection: Connection`, and use `TransactionalConnection` instead of `Connection` as bound for `AnsiTransactionManager`'s parameter. This would make a lot of sense, and `TransactionalConnection`' implementations could move closer to `AnsiTransactionManager`.

weiznich:
@Emm Sorry for the late answer.

> As a side note, do you know why the CI tests have no output? It's somewhat disturbing

At least for the the CI has output. A potential issue is that github action requires you to login to have access to the logs, so maybe that's the problem for you?

I've copied the output of one failing run below:

> ```
>  ---- connection::transaction_manager::test::nested_postgres_transaction_is_rolled_back_upon_deferred_constraint_failure > stdout ----
> thread 'connection::transaction_manager::test::nested_postgres_transaction_is_rolled_back_upon_deferred_constraint_failure' panicked at 'assertion failed: result.is_ok()', diesel/src/connection/transaction_manager.rs:692:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```


> Forcing the error messages to be in English would also be potentially a breaking change for Diesel consumers, without compilation errors.

Having a breaking change there is currently acceptable, as the next release will be 2.0. We should likely change that anyway, as  we have some other locations that need to check the error message to get information that are not exposed in any other way.

> * move is_transaction_broken() to the Connection trait

Having it in `Connection` also feels wrong, as connections do not really handle transactions at all. This functionality is provided by the transaction manager implementation.

> * strongly-typed "raw" error codes in Error, ie RawError::Postgres(SomePostgresErrorCodeEnum) - this would be a good idea in general, but it's also a lot more work

One option here would be to somehow integrate this information into [DatabaseErrorInformation](https://docs.diesel.rs/master/diesel/result/trait.DatabaseErrorInformation.html).

Generally speaking I think the most clearest option here is to just use a different connection manager implementation for the postgres connection instead of the generic `AnsiTransactionManager`. At least for me that seems to be the right place to fix this issue, as you would have direct access to the underlying connection there and it would easily be possible to check any required condition there for the postgres specific cases. This should be implemented in such a way that code is shared between the two `TransactionManager` implementations, maybe even by just calling the `AnsiTransactionManager` internally.

Emm:
@weiznich ,

> > As a side note, do you know why the CI tests have no output? It's somewhat disturbing
> 
> At least for the the CI has output. A potential issue is that github action requires you to login to have access to the logs, so maybe that's the problem for you?

Could be, looks like I can see the logs now...

> > Forcing the error messages to be in English would also be potentially a breaking change for Diesel consumers, without compilation errors.
> 
> Having a breaking change there is currently acceptable, as the next release will be 2.0. We should likely change that anyway, as we have some other locations that need to check the error message to get information that are not exposed in any other way.

I would really recommend against it. Changing the language in the connection is hostile to non-anglophone consumers of Diesel, and in any case, error message checking is an anti-pattern. If that's used elsewhere in the codebase, I would really recommend switching to the error code approach, even something unsafe like `enum ErrorCode { Postgres(String), MySql(String), Sqlite(String) }` would be better than checking error messages.

> One option here would be to somehow integrate this information into [DatabaseErrorInformation](https://docs.diesel.rs/master/diesel/result/trait.DatabaseErrorInformation.html).
> 
> Generally speaking I think the most clearest option here is to just use a different connection manager implementation for the postgres connection instead of the generic `AnsiTransactionManager`. At least for me that seems to be the right place to fix this issue, as you would have direct access to the underlying connection there and it would easily be possible to check any required condition there for the postgres specific cases. This should be implemented in such a way that code is shared between the two `TransactionManager` implementations, maybe even by just calling the `AnsiTransactionManager` internally.

Ok, this could work. I'll have a look at it this week.

weiznich:
> I would really recommend against it. Changing the language in the connection is hostile to non-anglophone consumers of Diesel, and in any case, error message checking is an anti-pattern. If that's used elsewhere in the codebase, I would really recommend switching to the error code approach, even something unsafe like enum ErrorCode { Postgres(String), MySql(String), Sqlite(String) } would be better than checking error messages.

That all sounds really nice and right in theory. In practice there are things without any error code where it is really useful to provide at least some indication on what has gone wrong. One example here is detecting that the database connection was closed somehow. See https://github.com/diesel-rs/diesel/pull/2911 for details. Feel free to suggest a better way to implement this functionality.

Emm:
@weiznich, I have put the PR in draft mode. I have also fixed the tests (for some reason they were working fine in docker...). I'd like to clarify the desired implementation before moving forward.

Currently, we have a `struct AnsiTransationManager` + a `TransactionManager` trait. We could probably implement most of the common code as default implementation of `TransactionManager` methods, and then replace `AnsiTransactionManager` with `PostgresTransactionManager`, `MysqlTransactionManager` and `SqliteTransactionManager`. Would this be acceptable?

> One example here is detecting that the database connection was closed somehow. See #2911 for details. Feel free to suggest a better way to implement this functionality.

In this particular case, `PQStatus(raw_conn)` will give you `CONNECTION_BAD`, I believe.

weiznich:
I had another look at this and came to the following conclusion: 

I think we can and should fix that without adding another `TransactionManager` implementation. I guess adding another trait for this is fine, but it shouldn't be the supertrait of connection. It should only be required for using `AnsiTransactionManager`, so connection implementations using their own transaction manager don't need to bother with this.  We can do that by constraining the impl corresponding `TransactionManager` for `AnsiTransactionManager` like following:

```rust
impl<Conn> TransactionManager<Conn> for AnsiTransactionManager
where
    Conn: Connection<TransactionManager = Self> + TransactionStateCheck,
{
…
}
```

Further more I would like to suggest changing the definition of `TransactionStateCheck` (your former `TransactionalConnection` trait, feel free to suggest another name here) in the following way, so that possible other backends have access to more information about the error:

```rust
// Not sure about the variants here
#[non_exhaustive]
enum TransactionState {
    // Decrement the counter, as this will be "fixed" by the final rollback
    BrokenWillBeFixedByRollback,
    // Maybe panic for this case
    UnfixableBroken,
    // Ignore it?
    NotBroken, 
    ConnectionBroken,
}

trait TransactionStateCheck {
    fn is_transaction_broken(depth: u32, error: crate::result::Error, conn: &Self) -> TransactionState;
}
```

Emm:
Re. `TransactionStateCheck`, I'm good with the general idea. Any particular reason for `#[non_exhaustive]` and for not using `&self` in `TransactionCheck`

For the `PQStatus` thing, I'll be happy to fill an issue.

weiznich:
> Any particular reason for #[non_exhaustive] and for not using &self in TransactionCheck?

`#[non_exhaustive]` because I'm not sure if we need other cases in the future. Using `&self` sounds like a good idea.

Emm:
So:
- I have opened #2957 to track the connection error improvement
- I have a prototype based on your suggestion, with significant modifications

Instead of a "transaction state checker", we get a "commit error processor", which I think makes more sense, since it takes an error in parameter. It also consumes (and possibly returns) the error instead of taking it as parameter. It has a smaller return type, and it is only concerned with the action to take following the error and nothing else, which is either `RollbackAndThrow` or `Throw`. IMO, this makes `non_exhaustive` not necessary. This also does not mix concerns between the action to take, the state of the transaction and the state of the connection.

Side note, is there a good reason to `panic()` upon a nested commit error in the transaction manager rather than simply returning the error? IMO, it's up to the caller to know whether this warrants a panic.

Which brings me to an additional suggestion. IMO, an error upon commit should be its own `DatabaseErrorKind` variant (eg, `DatabaseErrorKind::CommitError`) containing the original error. Right now, it's very hard to use `transaction()` and figure out programmatically whether this occurred upon commit and hosed the entire transaction or not (obviously, this is mostly relevant for non-Postgres platforms)

diesel/src/connection/commit_error_processor.rs
```diff
@@ -0,0 +1,27 @@
+//! A module to evaluate what to do when a commit triggers an error.
+use crate::result::{DatabaseErrorKind, Error};
+
+/// Transaction status returned upon error on commit.
+#[derive(Debug)]
+pub enum CommitErrorOutcome {
```
weiznich, line 6: Maybe we should just use [`std::ops::ControlFlow`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) here instead? On the other hand a custom enums makes it much clearer what happens for each variant.

diesel/src/connection/transaction_manager.rs
```diff
@@ -1,5 +1,8 @@
-use crate::connection::Connection;
-use crate::result::{DatabaseErrorKind, Error, QueryResult};
+use crate::connection::{
+    commit_error_processor::{CommitErrorOutcome, CommitErrorProcessor},
+    Connection,
+};
```
weiznich, line 4: Diesel prefers to not use nested imports.

diesel/src/connection/transaction_manager.rs
```diff
@@ -114,43 +117,57 @@ where
         let transaction_depth = conn.transaction_state().transaction_depth;
         if transaction_depth <= 1 {
             match conn.batch_execute("COMMIT") {
-                // When any of these kinds of error happen on `COMMIT`, it is expected
-                // that a `ROLLBACK` would succeed, leaving the transaction in a non-broken state.
-                // If there are other such errors, it is fine to add them here.
```
weiznich, line 119: We should add this comment to the new `process_commit_error` function, because it makes sense there.

diesel/src/pg/connection/mod.rs
```diff
@@ -48,6 +49,36 @@ impl<'a> ConnectionGatWorkaround<'a, Pg> for PgConnection {
...
+                | Error::DatabaseError(DatabaseErrorKind::ReadOnlyTransaction, _) => {
+                    CommitErrorOutcome::RollbackAndThrow(error)
+                }
+                _ => CommitErrorOutcome::Throw(error),
+            }
+        } else {
+            let transaction_status = unsafe { self.raw_connection.transaction_status() };
+            match error {
+                Error::DatabaseError(DatabaseErrorKind::Unknown, _)
+                    if transaction_status == pq_sys::PGTransactionStatusType::PQTRANS_INERROR =>
```
weiznich, line 72: I would prefer to not expose `pq_sys` types outside of `pg/connection/raw.rs`

diesel/src/pg/connection/raw.rs
```diff
@@ -98,6 +98,10 @@ impl RawConnection {
         );
         RawResult::new(ptr, self)
     }
+
+    pub unsafe fn transaction_status(&self) -> PGTransactionStatusType {
```
weiznich, line 102: There is no need to mark this function as unsafe. As far as I see it's valid to call this function as long as `PgConnection` exists, because `internal_connection` is guaranteed to be not null.

diesel/src/pg/transaction.rs
```diff
@@ -1,6 +1,9 @@
 #![allow(dead_code)]
 use crate::backend::Backend;
-use crate::connection::{AnsiTransactionManager, TransactionManager};
+use crate::connection::{
+    commit_error_processor::CommitErrorProcessor,
+    {AnsiTransactionManager, TransactionManager},
+};
```
weiznich, line 6: Prefer flat imports here
```suggestion
use crate::connection::commit_error_processor::CommitErrorProcessor;
use crate::connection::{AnsiTransactionManager, TransactionManager};
```

diesel/src/r2d2.rs
```diff
@@ -17,7 +17,10 @@ use std::fmt;
 use std::marker::PhantomData;
 
 use crate::backend::Backend;
-use crate::connection::{ConnectionGatWorkaround, SimpleConnection, TransactionManager};
+use crate::connection::{
+    commit_error_processor::{CommitErrorOutcome, CommitErrorProcessor},
+    ConnectionGatWorkaround, SimpleConnection, TransactionManager,
+};
```
weiznich, line 23:
```suggestion
use crate::connection::commit_error_processor::CommitErrorProcessor;
use crate::connection::{AnsiTransactionManager, TransactionManager};
```

weiznich (approved):
This looks really good now :+1: Thanks for your great work here.

I've left a few comments on minor things that could be improved, otherwise this should be ready to be merged.

weiznich:
> Side note, is there a good reason to panic() upon a nested commit error in the transaction manager rather than simply returning the error? IMO, it's up to the caller to know whether this warrants a panic.

> Which brings me to an additional suggestion. IMO, an error upon commit should be its own DatabaseErrorKind variant (eg, DatabaseErrorKind::CommitError) containing the original error. Right now, it's very hard to use transaction() and figure out programmatically whether this occurred upon commit and hosed the entire transaction or not (obviously, this is mostly relevant for non-Postgres platforms)

That's both a good idea.

Emm:
I'll see what I can do about all that this weekend, then.

Emm:
I have addressed the comments in the PR. I prefer to address the `panic()` and the dedicated error variant in another PR.

Ten0 (commented):
Overall great, good thing we're attempting to fix this.

Currently we have a wrapper around pooled connections that checks transaction depths and attempts rollback and logs errors there before releasing to the pool.
Even with this, we're going to have to keep it in order to still attempt rollback on panic. But it should prove useful less often with `INITIALLY DEFERRED` constraints being properly handled :)

There's one thing that bugs me though, where it seems we may decrement transaction depth without being sure that actually succeeded.

diesel/src/connection/transaction_manager.rs
```diff
@@ -114,43 +115,57 @@ where
...
+                            let r = conn.batch_execute("ROLLBACK");
+                            conn.transaction_state().change_transaction_depth(-1, r)?;
+                            error
+                        }
+                        CommitErrorOutcome::Throw(error) => {
+                            let _r = conn
+                                .transaction_state()
+                                .change_transaction_depth(-1, Ok(()));
+                            error
+                        }
```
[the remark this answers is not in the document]
Emm: I think you make some good points wrt decrementing the transaction depth upon failure.

I'm less sanguine about the variant names you suggest. IMO, the role of the processor should really be to specify the action to take following the commit failure, because this depends so much on the database implementation, but I'm open to changing these if you make a good argument.

> When implementor of CommitErrorProcessor cannot guarantee that TransactionAlreadyRolledBack is true, then it shall return ConnectionBroken. This guarantees that we don't ever release a connection to the pool with transaction_depth=0 when that is in fact not true at the database level.

Well, returning `ConnectionBroken` / `TransactionStatusUnknown` doesn't guarantee anything of the sort, unfortunately, since we don't have a switch at the moment to turn the connection into a broken connection. We can return `DatabaseErrorKind::ClosedConnection` from the manager, but this is not going to do anything about the connection. It's tricky to get right: what if a particular connection type could recover from a transport error (I don't think Postgres can).

> With regards to process_commit_error's default impl, I think we should either:
>
> * Remove the default impl completely

I could go along with it as long as we have a utility function which is called for the mysql and sqlite implementations, since I don't believe they need to be different at this time (AFAIK, we don't have anything to test the status of the connection in either case)

>  For the same reason, I think the impl for postgres should not do _ => CommitErrorOutcome::TransactionAlreadyRolledBack either, and should instead return that only for specified errors, or check transaction_status if it doesn't know.

I'd be happy to document that, but my understanding is that as long as a deferred trigger fired upon commit (and failed), any kind of exception could have been thrown by the trigger (and automatically rollback the connection). That said, enumerating the errors in the match rather than doing a default match would be safer.

> check transaction_status if it doesn't know

That's not very useful for if the transaction depth is > 1.

diesel/src/connection/transaction_manager.rs
```diff
@@ -259,4 +472,274 @@ mod test {
...
+            assert!(result.is_ok());
+            Ok(())
+        });
+        assert_eq!(
+            0,
+            <AnsiTransactionManager as TransactionManager<PgConnection>>::get_transaction_depth(
+                conn
+            )
+        );
+        assert!(result.is_err());
```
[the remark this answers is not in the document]
Emm: Fair enough.

diesel/src/connection/transaction_manager.rs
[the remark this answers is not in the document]
Ten0: > Well, returning `ConnectionBroken` doesn't guarantee anything of the sort, unfortunately, since we don't have a switch at the moment to turn the connection into a broken connection

What I'm trying to guarantee is that "we don't ever release a connection to the pool **with `transaction_depth`=0** when that is in fact not true".
What I mean by that is that specifically, i'd like to maintain the behavior that if we're unsure, we leave the transaction depth higher, so that the caller can take additional action with the knowledge that stuff was not properly closed.

IIUC, this property was here before, and wouldn't be if we don't introduce the option to throw and not decrement transaction depth.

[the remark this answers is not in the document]
Emm: Yeah, but thinking about it, and after some diving into Postgres' code (the only database we need to be concerned about wrt broken connections/transactions since we don't/can't check this for Mysql/Sqlite), connections do **NOT** appear to be able to recover from a `CONNECTION_BAD` status (unless you reset them by reopening them). So, operationally, once the connection to the Postgres server is lost, the transaction depth of the manager is irrelevant, since any future operation will fail with an error.

Now, from a correctness point of view, there is no telling the state of the server, since the connection may have been lost before or after the commit, which may or may not have processed correctly. The correct thing to do, IMO, would be to introduce something like 

```rust
enum TransactionManagerStatus {
   Valid { transaction_depth: i32 } // Not sure that I like an `i32` that `get_transaction_depth()` returns as an u32, but that's not the point here....
   Invalid // Hosed, you can't play with transactions with this connection again
}
```

> What I mean by that is that specifically, i'd like to maintain the behavior that if we're unsure, we leave the transaction depth higher, so that the caller can take additional action with the 
knowledge that stuff was not properly closed.

But that's misleading as well. IMO, the most robust behaviour in case of lost connection would to break the transaction manager and return a specific error wrapping the original cause, but at least for the "specific error" part, I intended to submit that in a specific PR.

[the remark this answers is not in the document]
Emm: @weiznich , any thought regarding changing the API of transaction manager to potentially make it "broken" in case we get `PQTRANS_UNKNOWN` from PG (or equivalent for Mysql/Sqlite)?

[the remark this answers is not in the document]
weiznich: @Emm I hadn't the time to look at this yet. Will try to do that this week.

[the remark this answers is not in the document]
weiznich: Changing the transaction data from a simple integer that tracks only the depth to an enum that tracks also if the connection contains a broken connection is a great ideas. Given that we could easily implement behavior that just returns an error from any called function as soon as the connection contains a broken connection. This should drastically improve the behavior around panicing in transactions in combination with the usage of connection pools. So :+1: from me here. 
(Its totally fine to implement not all of that in this pr.)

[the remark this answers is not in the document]
Emm: All right. I'll give a go at updating the transaction manager in this PR, changing to a "broken" transaction manager, based on getting `PQTRANS_UNKNOWN`. We can implement a similar mechanism to implement a broken state for the connection based on getting `CONNECTION_BAD` at the connection level, which is important if people get a broken connection outside of the transaction manager.

Thanks for your feedback

weiznich:
@Emm As you marked this as draft but the CI seems to pass: Is this ready for review? Or is there something missing where you could need help?

Emm:
@weiznich , sorry for the late reply. The one piece missing was test coverage.

The PR is now ready for review. 

I started by removing `CommitErrorProcessor::process_commit_error()`'s default implementation, to ensure that additional connection implementations explicitly implement it.

I removed catch-all `_` clauses to add safety to the error handling code (if additional errors / `ErrorKind` variants are added), the code will need to be updated.

Transitioning from a simple `transaction_depth` to an enum is quite expensive. The boilerplate to update the transaction status is expensive, all the more because we need to match the transaction manager's status twice, once at the beginning to ensure that we can actually execute the operation, and a second time after running the operation, since doing anything requires `&mut conn` and we cannot prove that the operation did not change the status of the linked transaction manager. In order to reduce the duplication of this logic, operations in the transaction manager internally run as callbacks return an `OperationOutcome` type, and are run with a `run_if_valid()` internal method. This `run_if_valid()` is also responsible for processing the outcome, and updating the transaction depth or marking the transaction manager as broken.

A notable operational change is that we no longer panic when a rollback in a nested transaction fails (I had originally intended that for a different pull request, but given the amount of changes in the transaction manager already, might as well go for broke). Another change is that attempting to commit a transaction when the transaction depth is 0 throws a `NotInTransaction` error.

We also expose `transaction_depth()` and `change_transaction_depth()` publicly, since the state of `ValidTransactionManagerStatus` is hidden.

I'm not really happy about the verbosity of this solution, but it seems reasonably safe.

I also added a `MockConnection` to test the behaviour of the transaction manager (although it uses `Pg` as `Backend`, because implementing a new backend seemed like a lot of effort).

weiznich:
@Emm Thanks for the summary :+1:. I will try to do a review this week.

weiznich (commented):
Thanks for the great update :+1: Implementation wise this PR looks really great now. I've noticed a few minor things that could be improved. The only thing I really would like to see fixed before merging this is the simplification of the `run_if_valid` function.

Emm:
@weiznich , thanks for the review. I'll have a look next week.

diesel/src/connection/transaction_manager.rs
```diff
@@ -31,78 +33,222 @@ pub trait TransactionManager<Conn: Connection> {
...
+            |conn, transaction_depth| match transaction_depth {
+                0 => Ok(OperationOutcome {
+                    result: conn.batch_execute(sql),
+                    update: ManagerUpdate::UpdateTransaction(TransactionUpdate {
+                        update_condition: UpdateCondition::AfterResultCheck,
+                        transaction_depth_change: TransactionDepthChange::IncreaseDepth,
+                    }),
+                }),
+                depth if depth > 0 => Err(Error::AlreadyInTransaction),
+                _depth => panic!("Transaction depth < 0"),
```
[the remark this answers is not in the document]
Emm: @weiznich , absolutely, but maybe in a separate commit

diesel/src/connection/transaction_manager.rs
```diff
@@ -31,78 +33,222 @@ pub trait TransactionManager<Conn: Connection> {
...
+            };
+            let result = conn.batch_execute(&*start_transaction_sql);
+            Ok(OperationOutcome {
+                result,
+                update: ManagerUpdate::UpdateTransaction(TransactionUpdate {
+                    update_condition: UpdateCondition::AfterResultCheck,
+                    transaction_depth_change: TransactionDepthChange::IncreaseDepth,
+                }),
+            })
+        })
```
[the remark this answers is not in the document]
Emm: I see where you're going with that, but it means nesting calls to `run_if_valid`, which, on balance, is best avoided if the cost is four lines of duplicated code (IMO).

diesel/src/connection/transaction_manager.rs
```diff
@@ -31,78 +33,222 @@ pub trait TransactionManager<Conn: Connection> {
...
+                            result
+                        }
+                    }
+                }
+                ManagerUpdate::MarkManagerAsInError => {
+                    *status = TransactionManagerStatus::InError;
+                    result
+                }
+            },
+        }
```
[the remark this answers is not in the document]
Emm: I'm happy making `change_transaction_depth` take `TransactionDepthChange` instead of `i32` (that's also safer). I'm less sold on avoid the nested match. IMO, having the same variant on multiple lines is best avoided (unless you use `if` in your match condition), as this obfuscates the fact that these three branches are actually the same case. I have however shortened the code via destructuring. I will implement the des-nestification if you think it's best, but please have a look at the new version first.

weiznich:
@Emm I've rebased your branch [here](https://github.com/weiznich/diesel/tree/Emm-master) to fix the CI errors caused by the latest merge to master. Additionally I've experimented a bit with an alternative approach of handling all the different states [here](https://github.com/weiznich/diesel/commit/906939eeb088f56ef42e9914f7dd479414405e22). I feel that is quite simpler to read due to much less scope nesting. I would be happy to hear your opinion on that.

Emm:
@weiznich, brilliant. This looks just as safe, but less cumbersome. The use of `NonZeroU32` in particular is inspired, I didn't even know this was a thing. I have put your changes on my branch, and as far as I'm concerned, this is mergeable.

diesel/src/connection/commit_error_processor.rs
```diff
@@ -0,0 +1,64 @@
...
+/// `DatabaseErrorKind::SerializationFailure` or `DatabaseErrorKind::ReadOnlyTransaction`
+pub fn default_process_commit_error(
+    transaction_state: &ValidTransactionManagerStatus,
+    error: Error,
+) -> CommitErrorOutcome {
+    let transaction_depth = transaction_state.transaction_depth();
+    match error {
+        Error::DatabaseError(DatabaseErrorKind::ReadOnlyTransaction, _)
+        | Error::DatabaseError(DatabaseErrorKind::SerializationFailure, _)
+            if transaction_depth.is_none() =>
```
weiznich, line 40: After thinking again about this I believe that's not correct. 
```suggestion
            if transaction_depth == NonZeroU32::new(1) =>
```

(Will fix that next week + add an test for this case)

weiznich:
I've update this PR again to

* Fix one issue I've found in my previous implementation of `commit_processor`
* Add some additional code to improve the handling of r2d2 and returned connections with open transactions (this should now not happen anymore)
* Add an additional error variant to indicate the outcome of a failed `commit` attempt. 


Comments especially from @diesel-rs/reviewers  or @Emm  are welcome. Otherwise I want to merge this soon.

Emm:
@weiznich, I've started a review, I hope to be able to complete it tomorrow.

Emm:
I'm not going to be done today, but I'm leaving a few comments for now. I want to add tests for the read only case (also nested), I'm uncomfortable with the fact that this behaviour is not covered.

weiznich:
@Emm  No worries, take your time ❤️ . I've applied your suggestions (everything but the tests). That written I will likely take a break over the holidays, so don't expect answers in timely manner here.

Emm:
@weiznich , noted. Thanks for the quick update on the PR. I have some bad news, though... the MySQL backend doesn't handle serialization errors correctly when nested, which is due to the fact that the deadlock error is not thrown at commit time (and thus, not caught by the manager, which does not decrement the transaction level). I'll work on a solution during the holidays.

Emm:
@weiznich, to recap:
- in Mysql, a statement triggers a serialization error (error 1213), which rolls back the transaction
- as this is not intercepted by the transaction manager, the transaction depth is not updated

The obvious solution would be to add an error parameter to `rollback_transaction()`. Unfortunately, the error we could feed to `rollback_transaction()` is `E: From<Error>`, which doesn't let us retrieve the original error, so we can't check if we got a `SerializationFailure`.

Options:
- don't do anything, but document that serialization failures in MySql will break the manager (I think it's the same for RO errors?)
- see if we also get a serialization failure upon commit, in which case we should be good
- change the trait bound from `From<Error>` to something which would let us retrieve the original error
- add a `last_error() -> Option<DatabaseError>` to `Connection` to retrieve the last error (very ugly)

In any case, I think the match in `mysql/connection/stmt/mod.rs` should be amended to return `SerializationFailure` when receiving an error with code 1213.

weiznich:
@Emm Thanks for the summary. I'm not sure if I understand that correctly. 
So you say that for `MySql` the `SerializationError` is returned as soon as you do something that would cause this and not only if the connection is committed, right? Assuming that the user bubbles up the error from the transaction callback that would mean that we roll back the transaction at this point. Now as far as I can see any rollback will decrease the connection depth as long as the rollback succeeds. That leaves me wondering what's the problem here that you try to describe. I cannot see how this could cause a broken transaction manager 🤔 

> In any case, I think the match in mysql/connection/stmt/mod.rs should be amended to return SerializationFailure when receiving an error with code 1213.

:+1:

Emm:
@weiznich , it goes like this:

| Step | Mysql state | Manager depth |
| -- | -- | -- |
| Pre-transaction | Not in transaction | 0 |
| Begin transaction | In transaction | 1 |
| Statement throws a serialization error and Mysql rolls back the transaction | Not in transaction | 1 |
| The manager has its `rollback_transaction()` method invoked | Not in transaction | Attempt to switch from 1 to 0, but rollback fails since we're no longer in a transaction |

This also applies for nested transactions, obviously.

weiznich:
@Emm Thanks for the nice summary. I will need some days to think about potential solutions. I will report back as soon as I've had some time to try a few things.

Emm:
@weiznich , no worries. We could also put that in another issue and merge the existing code, since this does solve the issue for PG (note that, also I didn't try, I don't think the serialization failure would ever come up for Sqlite, since it locks the file for writing).

weiznich:
@Emm That sounds good. I will merge the PR now so that we have the already implemented parts merged. 
Can you share the additional test cases that you've already written? That would help us to test different solutions for a possible mysql fix.

Emm:
@weiznich , you're welcome, and thanks for doing all the review/modification work. I'll add a PR with the MySql test ignored this week.

PR #2944, opened 2021-11-03
FROM ONLY <table> clause

Ploppz:
Compile fail test don't pass yet due to https://github.com/dtolnay/trybuild/issues/140 but I would like to already request feedback on, for example, the placement of my implementation and how to expose it in the API.

### DIFF
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -29,6 +29,24 @@ macro_rules! __diesel_fix_sql_type_import {
 
 #[macro_export]
 #[doc(hidden)]
+#[cfg(feature = "postgres_backend")]
+macro_rules! __diesel_internal_table_backend_specific_impls {
+    ($table:ident, $column_name:ident) => { 
+        impl $crate::expression::AppearsOnTable<$crate::query_builder::Only<$table>>
+            for $column_name {}
+        impl $crate::expression::SelectableExpression<$crate::query_builder::Only<$table>>
+            for $column_name {}
+    };
+}
+#[macro_export]
+#[doc(hidden)]
+#[cfg(not(feature = "postgres_backend"))]
+macro_rules! __diesel_internal_table_backend_specific_impls {
+    ($table:ident, $column_name:ident) => {};
+}
+
+#[macro_export]
+#[doc(hidden)]
 macro_rules! __diesel_column {
     (
         table = $table:ident,
@@ -134,6 +152,9 @@ macro_rules! __diesel_column {
                 self.eq(rhs)
             }
         }
+
+
+        $crate::__diesel_internal_table_backend_specific_impls!($table, $column_name);
 
         $crate::__diesel_generate_ops_impls_if_numeric!($column_name, $($Type)*);
         $crate::__diesel_generate_ops_impls_if_date_time!($column_name, $($Type)*);
--- a/diesel/src/pg/expression/extensions/only_dsl.rs
+++ b/diesel/src/pg/expression/extensions/only_dsl.rs
@@ -23,7 +23,7 @@ use crate::query_source::Table;
 /// tables.
 pub trait OnlyDsl: AsQuery + Sized {
     /// See the trait-level docs.
-    fn only(self) -> SelectStatement<Only<Self>> {
+    fn only(self) -> crate::dsl::SelectFromOnly<Self> {
         SelectStatement::simple(Only { query: self })
     }
 }
--- a/diesel/src/pg/expression/extensions/only_dsl.rs
+++ b/diesel/src/pg/expression/extensions/only_dsl.rs
@@ -1,0 +1,31 @@
+use crate::query_builder::Only;
+use crate::query_builder::{AsQuery, SelectStatement};
+use crate::query_source::Table;
+
+/// The `only` method
+///
+/// This is only implemented for the Postgres backend.
+/// The `ONLY` clause is used to select only from one table and not any inherited ones.
+///
+/// Calling this function on a table (`mytable.only()`) will result in the SQL `FROM ONLY mytable`.
+///
+/// Example:
+///
+/// ```ignore
+/// let n_users_in_main_table = users::table
+///     .only()
+///     .select(count(users::id))
+///     .first::<i64>(connection)
+///     .unwrap();
+///
+/// ```
+/// Selects the number of entries in the `users` table excluding any rows found in inherited
+/// tables.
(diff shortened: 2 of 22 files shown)


diesel/src/macros/mod.rs
```diff
@@ -810,6 +814,9 @@ macro_rules! __diesel_table_impl {
                 }
             }
 
+            // impl $crate::expression::AppearsOnTable<$crate::query_builder::Only<table>> for <table as $crate::query_source::QuerySource>::DefaultSelection {}
```
Ploppz, line 817: This resulted in an impl conflict error, but apparently was not needed after all.

diesel/src/query_builder/only_clause.rs
```diff
@@ -0,0 +1,85 @@
...
+};
+
+/// Represents a query with an `ONLY` clause.
+#[derive(Debug, QueryId)]
+pub struct Only<T> {
+    query: T,
+}
+
+/// Add `ONLY` before a query source.
+pub trait OnlyDsl: AsQuery + Sized {
```
weiznich, line 19: The dsl should be moved to the querydsl module (or the corresponding submodule of the pg module)

diesel/src/query_builder/only_clause.rs
```diff
@@ -0,0 +1,85 @@
...
+    query_source::Table,
+};
+
+/// Represents a query with an `ONLY` clause.
+#[derive(Debug, QueryId)]
+pub struct Only<T> {
+    query: T,
+}
+
+/// Add `ONLY` before a query source.
```
weiznich, line 18: This is the main entry point for using `from only ...` clauses. I feel the documentation here could be a bit more detailed.

diesel/src/macros/mod.rs
weiznich: Yes this is not needed if we implement the corresponding traits for columns, as this is then implied by a more general impl.

weiznich (commented):
Thanks for submitting this pr :+1:
I've added a few comments on how to improve the general structure of the changed code.

diesel/src/query_builder/only_clause.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn walk_ast(&self, mut out: AstPass<DB>) -> QueryResult<()> {
+        out.push_sql("ONLY ");
+        self.query.walk_ast(out.reborrow())?;
+        Ok(())
+    }
+}
+
+impl<T> QueryDsl for Only<T> {}
+
+impl<T, S> SelectDsl<S> for Only<T>
```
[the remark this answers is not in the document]
Ploppz: What do you mean?

diesel/src/query_builder/only_clause.rs
Ploppz: I can't find any dsl module in `src/pg`

weiznich: I would likely add it [here](https://github.com/diesel-rs/diesel/tree/master/diesel/src/pg/expression/extensions) as additional sub module + reexport it from `diesel::query_dsl`

diesel/src/query_builder/only_clause.rs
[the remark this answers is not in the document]
weiznich: The current implementation only allows users to call `.select()` on any table you've called `.only()` on. To allow calling other `QueryDsl` methods, like for example filter, we need to implement more traits. Basically every trait reexported [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/query_dsl/mod.rs#L63-L78) needs to be implemented.

diesel/src/query_builder/only_clause.rs
Ploppz: I guess it could go directly into the `pg` module?

Ploppz:
I updated the commit. Is the structure good now? I put `Only` and `OnlyDsl` in different places, not sure about that though.

weiznich:
Thanks for the update :+1:

The general structure looks much nicer now. I left a comment on a minor improvement and a comment about how to fix the compilation errors. Beside of that I would like to see the following things implemented/added

- [ ] A bit more documentation on the `OnlyDsl` trait
- [ ] A changelog entry
- [ ] Adding the missing query dsl impls.

diesel/src/macros/mod.rs
```diff
@@ -135,6 +135,10 @@ macro_rules! __diesel_column {
             }
         }
 
+
+        impl $crate::expression::AppearsOnTable<$crate::pg::Only<$table>> for $column_name {}
```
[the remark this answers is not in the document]
Ploppz: ah indeed I ran into problems when I tried that `#[cfg(feature = "postgres_backend")]`. Thanks, will do that

Ploppz:
Does it look good now?

diesel_compile_tests/tests/fail/only_only_on_table.rs
```diff
@@ -0,0 +1,18 @@
...
+        id -> Int8,
+    }
+}
+
+fn main() {
+    let mut connection = PgConnection::establish("").unwrap();
+
+    foo::table.select(foo::id).only();
+    foo::table.select(foo::id).filter(foo::id.eq(1)).only();
+    foo::table.select(foo::id.only());
```
weiznich, line 17: I would like to see here the following additional checks:

```rust
let mut conn = SqliteConnection::establish("").unwrap();

foo::table.only().load(connection).unwrap();

let mut conn = MysqlConnection::establish("").unwrap();

foo::table.only().load(connection).unwrap();
```

weiznich (approved):
This looks good now, beside the three minor things mentioned in the review comments.
Thanks for your work here :+1:

diesel/src/pg/expression/extensions/only_dsl.rs
```diff
@@ -0,0 +1,31 @@
...
+///     .select(count(users::id))
+///     .first::<i64>(connection)
+///     .unwrap();
+///
+/// ```
+/// Selects the number of entries in the `users` table excluding any rows found in inherited
+/// tables.
+pub trait OnlyDsl: AsQuery + Sized {
+    /// See the trait-level docs.
+    fn only(self) -> SelectStatement<Only<Self>> {
```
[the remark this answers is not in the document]
weiznich:
```suggestion
    fn only(self) -> crate::dsl::SelectFromOnly<Self> {
```

weiznich (approved):

PR #2948, opened 2021-11-09
Coerce pg types to lowercase untils wrapped in quotes

zephraph:
## Background

I lost a _lot_ of time to this today. This might not be the best way to go about solving it, but _some_ mechanism to inform users what's going on would be good. Essentially I was implementing a custom enum. This is roughly the code I have:

```rust
#[derive(SqlType, Debug)]
#[postgres(type_name = "MyEnum")]
pub struct MyEnumType;

#[derive(Clone, Debug, AsExpression, FromSqlRow)]
#[sql_type = "MyEnumType"]
enum MyEnum { ... }
```

Notice that `type_name` is PascalCase. Something I didn't recognize (that I should've) is that pg coerces all identifiers to lowercase unless they're explicitly wrapped in quotes. When I created the enum in the database it was via a statement like 

```sql
CREATE TYPE MyEnum AS ENUM (
    'state1',
    'state2'
);
```

Notice again that `MyEnum` is pascal case. Then, when I was trying to make a call to serialize something in the DB I got this error:

```
Unknown diesel error: SerializationError(FailedToLookupTypeError(PgMetadataCacheKey { schema: None, type_name: \"MyEnum\" }))
```

Everything looked right! I could not for the life of me figure out what was up. This `FailedToLookupTypeError` is happening because the key that the `PgMetadata` struct actually has is all lowercase. 

## Solution

I've went for the simplest possible solution here. Assuming the type isn't wrapped in quotes (which would cause postgres to preserve the casing) then we `to_lowercase()` the type name such that it matches the `PgMetadataCacheKey`. There may be more sophisticated solutions here, but it's a bit beyond the time I have to invest in it. Also, I've obviously only made this change for pg given I'm somewhat familiar with its behavior. 

If you're good w/ this level of change I can push it forward further. Perhaps, if we know the casing behavior for MySql and sqlite, we can put fixes in place for those as well.

### DIFF
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -47,7 +47,21 @@ where
                     .store_type(cache_key, type_metadata);
                 PgTypeMetadata(Ok(type_metadata))
             }
-            Err(_e) => PgTypeMetadata(Err(FailedToLookupTypeError::new(cache_key.into_owned()))),
+            Err(_e) => {
+                let metadata_cache = self.get_metadata_cache();
+                let lower_schema = schema.map(|schema| schema.to_lowercase()).unwrap();
+                if let Some(_) = metadata_cache.lookup_type(&PgMetadataCacheKey {
+                    schema: Some(Cow::Borrowed(lower_schema.as_str())),
+                    type_name: Cow::Borrowed(type_name.to_lowercase().as_str()),
+                }) {
+                    return PgTypeMetadata(Err(FailedToLookupTypeError::InvalidCasing(Box::new(
+                        cache_key.into_owned(),
+                    ))));
+                }
+                PgTypeMetadata(Err(FailedToLookupTypeError::TypeNotFound(Box::new(
+                    cache_key.into_owned(),
+                ))))
+            }
         }
     }
 }
--- a/diesel/src/pg/backend.rs
+++ b/diesel/src/pg/backend.rs
@@ -29,28 +29,37 @@ impl From<(u32, u32)> for InnerPgTypeMetadata {
 /// This error indicates that a type lookup for a custom
 /// postgres type failed
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct FailedToLookupTypeError(Box<PgMetadataCacheKey<'static>>);
-
-impl FailedToLookupTypeError {
-    /// Construct a new instance of this error type
-    /// containing information about which type lookup failed
-    pub fn new(cache_key: PgMetadataCacheKey<'static>) -> Self {
-        Self(Box::new(cache_key))
-    }
+pub enum FailedToLookupTypeError {
+    /// Error used when a type error lookup fails
+    TypeNotFound(Box<PgMetadataCacheKey<'static>>),
+    /// Error used when the lookup fails but there is a lowercase version of the given typename
+    InvalidCasing(Box<PgMetadataCacheKey<'static>>),
 }
 
 impl std::error::Error for FailedToLookupTypeError {}
 
 impl std::fmt::Display for FailedToLookupTypeError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        if let Some(schema) = self.0.schema.as_ref() {
-            write!(
-                f,
-                "Failed to find a type oid for `{}`.`{}`",
-                schema, self.0.type_name
-            )
-        } else {
-            write!(f, "Failed to find a type oid for `{}`", self.0.type_name)
+        match self {
+            FailedToLookupTypeError::TypeNotFound(cache_key) => match cache_key.schema.as_ref() {
+                Some(schema) => write!(
+                    f,
+                    "Failed to find a type oid for `{}`.`{}`",
+                    schema, cache_key.type_name
+                ),
+                _ => write!(f, "Failed to find a type oid for `{}`", cache_key.type_name),
+            },
+            FailedToLookupTypeError::InvalidCasing(cache_key) => match cache_key.schema.as_ref() {
+                Some(schema) => write!(
+                    f,
+                    "Failed to find a type oid for `{}`.`{}` but did find `{}`.`{}`. Is the casing correct?",
+                    schema,
+                    cache_key.type_name,
+                    schema.to_lowercase(),
+                    cache_key.type_name.to_lowercase()
+                ),
+                _ => write!(f, "Failed to find a type oid for `{}` but did find `{}`. Is the casing correct?", cache_key.type_name, cache_key.type_name.to_lowercase())
+            }
         }
     }
 }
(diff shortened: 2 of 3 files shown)


weiznich:
Thanks for opening this PR.

Instead of adding hidden behaviour to the derive, I would prefer to add either documentation to this error type stating what could be possibly wrong or add a improved debug/display implementation that points out a potential problem. 
The problem I see with doing the lowercasing just all the time is that it can lead to really hard to spot bugs for people using quoted identifiers.

zephraph:
That's fair. I'll think about it a bit and see if I can update the error message. Perhaps on the error path we can do a second lookup for the lowercase version and specifically mention that may be the error if it's found.

weiznich:
> Perhaps on the error path we can do a second lookup for the lowercase version and specifically mention that may be the error if it's found.

That sounds like a good idea. The corresponding place to perform this check is here: https://github.com/diesel-rs/diesel/blob/b2c58897c39c519a946314bd5b63765d3af56204/diesel/src/pg/metadata_lookup.rs#L50
Basically we only need to call [`lookup_type`](https://github.com/diesel-rs/diesel/blob/b2c58897c39c519a946314bd5b63765d3af56204/diesel/src/pg/metadata_lookup.rs#L64) again with the lower cased type name and see if that query succeeds. Additionally we likely need to adjust the error type here: https://github.com/diesel-rs/diesel/blob/b2c58897c39c519a946314bd5b63765d3af56204/diesel/src/pg/backend.rs#L32 to include this additional information? (And maybe switch that `Debug` impl to a manual one to have a better control about the emitted error message. )

zephraph:
Okay @weiznich, it may fail CI, but I made a first pass at updating it to just be an error message. I'm still _really_ new to the rust ecosystem so forgive me if I made any obvious errors. Thanks for the hint about where to make updates, that was helpful.

diesel/src/pg/metadata_lookup.rs
```diff
@@ -47,7 +47,21 @@ where
                     .store_type(cache_key, type_metadata);
                 PgTypeMetadata(Ok(type_metadata))
             }
-            Err(_e) => PgTypeMetadata(Err(FailedToLookupTypeError::new(cache_key.into_owned()))),
+            Err(_e) => {
+                let metadata_cache = self.get_metadata_cache();
+                let lower_schema = schema.map(|schema| schema.to_lowercase()).unwrap();
+                if let Some(_) = metadata_cache.lookup_type(&PgMetadataCacheKey {
```
weiznich, line 53: `metadata_cache.lookup_type(…)` looks up existing cached metadata, without touching the database. That's not what we want in this case. We want to make an explicit query here, by calling the standalone `lookup_type` function from line 64. (Just noticed that both have the same name, which can be irritating)

Additionally the CI suggestion from below is correct in terms of simplifying the code. As the "other" `lookup_type` function returns an `Result<>` the corresponding code looks like:

```suggestion
                if lookup_type(&PgMetadataCacheKey { … }, self).is_ok()
```

diesel/src/pg/metadata_lookup.rs
```diff
@@ -47,7 +47,21 @@ where
                     .store_type(cache_key, type_metadata);
                 PgTypeMetadata(Ok(type_metadata))
             }
-            Err(_e) => PgTypeMetadata(Err(FailedToLookupTypeError::new(cache_key.into_owned()))),
+            Err(_e) => {
+                let metadata_cache = self.get_metadata_cache();
+                let lower_schema = schema.map(|schema| schema.to_lowercase()).unwrap();
```
weiznich, line 52: We shouldn't call `unwrap()` here as the schema name could be empty. Just doing the following things should be fine:
```suggestion
                let lower_schema = schema.map(|schema| schema.to_lowercase());
```

This will transform the schema name to lower case if there is a schema name, otherwise it will just set `lower_schema` to `None`

diesel/src/pg/metadata_lookup.rs
```diff
@@ -47,7 +47,21 @@ where
...
+                if let Some(_) = metadata_cache.lookup_type(&PgMetadataCacheKey {
+                    schema: Some(Cow::Borrowed(lower_schema.as_str())),
+                    type_name: Cow::Borrowed(type_name.to_lowercase().as_str()),
+                }) {
```
weiznich, line 56: There is no need to use `Cow::Borrowed` here, as we don't need neither the lower cased type name nor the lower cased schema name later on. Constructing `PgMetadataCacheKey` can be simplified to:
```rust
PgMetadataCacheKey {
    schema: lower_schema.map(Cow::Owned),
    type_name: Cow::Owned(type_name.to_lowercase()),
}
```

`Cow` is some kind of "magic" type that either accepts a borrowed value (`&str` in this case) or the owned counter part (`String` in this case). It can be used to reduce unnecessary calls to clone. Generally speaking it should be only used in hot code paths, otherwise it mostly just complicate things.

diesel/src/pg/metadata_lookup.rs
```diff
@@ -47,7 +47,21 @@ where
...
+                    return PgTypeMetadata(Err(FailedToLookupTypeError::InvalidCasing(Box::new(
+                        cache_key.into_owned(),
+                    ))));
+                }
+                PgTypeMetadata(Err(FailedToLookupTypeError::TypeNotFound(Box::new(
+                    cache_key.into_owned(),
+                ))))
```
weiznich, line 63: I personally prefer to write such either or conditions explicitly by using a else clause. This makes it much clearer what are possible variants. So for this case something like this:

```rust
if lookup_type(…).is_ok() {
    PgTypeMetadata(Err(FailedToLookupTypeError::InvalidCasing(Box::new(
       cache_key.into_owned(),
     ))))
} else {
      PgTypeMetadata(Err(FailedToLookupTypeError::TypeNotFound(Box::new(
          cache_key.into_owned(),
      ))))
}
```

Note that we can now skip the explicit return there as the whole `if` statement is the last expression in our function.

weiznich (commented):
Thanks for the update. The general approach looks already good :+1: I've left some comments on how to improve things. Hopefully they are helpful.

As an additional addition I would like to see a small test case showing that this works. Basically just add another test case to this file (https://github.com/diesel-rs/diesel/blob/19db25af30259f4b127933a77aff607f36e79957/diesel_tests/tests/custom_types.rs#L1) using a "wrong" type definition + a test function using it in some query. After that check that the error matches the expected value.

>  I'm still really new to the rust ecosystem so forgive me if I made any obvious errors. 

Thanks for mentioning that :+1:. I've tried to include a bit more context in each comment to explain some reasoning there, so feel free to skip that if you already familiar with that. On the other hand if it remains unclear just ask and I will try to clarify that.

zephraph:
I _think_ https://github.com/diesel-rs/diesel/pull/2993 may have ultimately fixed this. Regardless it's been a long time and I don't have the capacity to pick this back up so I'm just going to close. Thanks for your help back then @weiznich! Time flies, ha.

PR #2951, opened 2021-11-10
Add a test case to investigate #2947

weiznich:
Just a quick PR to see if I can reproduce the issue using the CI.


weiznich:
Seems to be a libpq issue

PR #2960, opened 2021-11-16
Upgradded dependencies after #2930

Razican:
It seems that on #2930 some dependencies were reverted, so this updates then again.

### DIFF
--- a/diesel_compile_tests/Cargo.lock
+++ b/diesel_compile_tests/Cargo.lock
@@ -355,9 +355,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.60"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1500e84d27fe482ed1dc791a56eddc2f230046a040fa908c08bda1d9fb615779"
+checksum = "e277c495ac6cd1a01a58d0a0c574568b4d1ddf14f59965c6a58b8d96400b54f3"
 dependencies = [
  "itoa",
  "ryu",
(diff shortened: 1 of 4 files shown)


diesel_compile_tests/Cargo.lock
```diff
@@ -355,9 +355,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.60"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1500e84d27fe482ed1dc791a56eddc2f230046a040fa908c08bda1d9fb615779"
+checksum = "e277c495ac6cd1a01a58d0a0c574568b4d1ddf14f59965c6a58b8d96400b54f3"
```
weiznich, line 360: This is causing the compile tests to fail as this serde_json version does only support newer rust versions than that one used for the compile tests. Compile tests depend on the exact compiler output. This output does change between compiler versions, so I've choose to pin the serde_json version for now.

I would accept a separate PR here that updates the used rust version so that we can update serde_json as well. Such a update would need to:
* Update the rust version in diesel_compile_tests/rust-toolchain
* Update the rust version for the corresponding CI job 
* Update the generated compiler output by running `TRYBUILD=overwrite cargo test` in the compile test directory

weiznich (commented):
Everything beside the compile tests change is fine for me :+1:

diesel_compile_tests/Cargo.lock
Razican: Makes sense :) I will do the changes later today

Razican: Done :)

pksunkara (approved):

PR #2962, opened 2021-11-21
More flexible Selectable derive

### LINKED ISSUE #2950: Selectable derive does not work with generic embed sub-fields
### Versions

- **Diesel:** Latest master (b2c58897c39c519a946314bd5b63765d3af56204)

```rust
#[derive(Queryable, Selectable)]
#[table_name = "schema::xxx"]
pub struct WithField<T> {
	pub field: i32,
	#[diesel(embed)]
	pub inner: T,
}
```

```rust 
impl<T, __DB> Selectable<__DB> for WithField<T>
where T: Selectable<__DB> {}
```

```rust 
impl<T, __DB> Selectable<__DB> for WithField<T> {}
```
-> fails to compile with "the trait bound `T: diesel::Selectable<__DB>` is not satisfied".

### Solving

Looks like if a field has `embed`, the `Selectable` derive should add bounds `FieldType: diesel::Selectable<__DB>`, regardless of whether the type is generic or not.

This would also allow the `Selectable` derive to interact as expected with manual implementations of `Selectable` on a single backend (non-generic).

Ten0:
Resolves #2950

Previously, if fields had `embed`, the `Selectable` derive supposed that they implemented `Selectable<DB>` for every `DB: Backend`.

If that wasn't the case (generic subfield or backend-specific impl) the derive would fail.

This commit fixes this issue by introducing the appropriate additional `where` bounds.

### DIFF
--- a/diesel_derives/src/selectable.rs
+++ b/diesel_derives/src/selectable.rs
@@ -9,10 +9,19 @@ pub fn derive(item: syn::DeriveInput) -> Result<proc_macro2::TokenStream, Diagno
     let model = Model::from_item(&item)?;
 
     let (_, ty_generics, _) = item.generics.split_for_impl();
+
     let mut generics = item.generics.clone();
     generics
         .params
         .push(parse_quote!(__DB: diesel::backend::Backend));
+    for embed_field in model.fields().iter().filter(|f| f.has_flag("embed")) {
+        let embed_ty = &embed_field.ty;
+        generics
+            .where_clause
+            .get_or_insert_with(|| parse_quote!(where))
+            .predicates
+            .push(parse_quote!(#embed_ty: Selectable<__DB>));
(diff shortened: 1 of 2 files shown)


pksunkara (commented):
Would this work when the `embed` is on a non-generic struct? I don't think rust supports `struct: trait` syntax in `where`, right?

weiznich:
@pksunkara [This does work for non-generic structs](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6a11c0529fa558f3dd621c7d0575fba4). The trick there is that the `Selectable` trait itself is generic over `DB`, which means that rustc cannot assume that it is implemented for all types, so putting up a where clause constraint is fine.

weiznich (approved):

PR #2967, opened 2021-12-02
No unnecessary boxing in load_iter

Ten0:
This recently slipped through my review of #2799, but was pointed out by rust-analyzer.
I'm not sure why we'd want to force boxing here. I'd rather avoid the dynamic dispatch by default and let the user do it if necessary (which seems very unlikely).

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -376,6 +376,11 @@ pub mod helper_types {
     #[cfg(feature = "postgres_backend")]
     pub type SelectFromOnly<T> =
         crate::query_builder::SelectStatement<crate::pg::query_builder::only_clause::Only<T>>;
+
+    /// [`Iterator`](std::iter::Iterator) of [`QueryResult<U>`](crate::result::QueryResult)
+    ///
+    /// See [`RunQueryDsl::load_iter`] for more information
+    pub type LoadIter<'a, Q, Conn, U> = <Q as load_dsl::LoadQueryGatWorkaround<'a, Conn, U>>::Ret;
 }
 
 pub mod prelude {
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -51,7 +51,7 @@ pub use self::join_dsl::{InternalJoinDsl, JoinOnDsl, JoinWithImplicitOnClause};
 #[doc(hidden)]
 pub use self::load_dsl::CompatibleType;
 #[doc(hidden)]
-pub use self::load_dsl::{LoadQuery, LoadQueryGatWorkaround};
+pub use self::load_dsl::LoadQuery;
 pub use self::save_changes_dsl::{SaveChangesDsl, UpdateAndFetchResults};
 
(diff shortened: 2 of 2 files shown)


weiznich:
It was an explicit decision by me to use  `Box<dyn Iterator<Item = QueryResult<U>>` there. My main motivation is to make the return type clear, as the opaque GAT workaround trait thing is not really obvious. This is especially true if we remove the `*GatWorkaround` traits from our public API and replace the return type with a simple type def. (That's something we want to do before cutting a release, as this gives us the possibility to remove the `*GatWorkaround` traits as soon as real GAT are available in stable rust without breaking changes.)
See https://github.com/diesel-rs/diesel/commit/69876a1892644f1dddbf4889763fc741f1f3faca for the commit that introduced this. 

In practice I don't think there will be a noticeable performance impact in nearly all use cases. Any use case that really requires that last 0.1% of performance should maybe just use `LoadQuery::internal_load()` instead, as this function returns the same iterator as unboxed version. So maybe instead of changing the return type we should just change to documentation to state that if someone really needs a unboxed iterator they should use `LoadQuery::internal_load()` instead?

Ten0:
Would having a `LoadIter<'a, Q, Conn, U>` helper type as return type for this function solve this issue? This would enable us to update it. 
I initially wanted to do that but wasn't sure where I could put it.

weiznich:
> Would having a LoadIter<'a, Q, Conn, U> helper type as return type for this function solve this issue? This would enable us to update it.

That would solve the GAT update issue. I've have a work in process branch locally that applies this change (+ a lot of other stability related things) to prepare the 2.0 release. 
My current problem with this approach is that this makes it really hard to understand that this function returns an iterator, because that cannot be seen from the typedef and the underlying trait is "private". That's the main reason why I think the boxing there is the better solution.

Ten0:
> this makes it really hard to understand that this function returns an iterator, because that cannot be seen from the typedef and the underlying trait is "private"

I think it should be made pretty clear by the naming and the documentation of the `load_iter` function, and we could also have some documentation on the type alias itself.
I don't see how someone using the `load_iter` function could avoid knowing that this is an iterator.

weiznich:
> (I don't know where we should put that type alias.)

I think `diesel::dsl` should be the right place for this.

weiznich:
CI seems to be broken due to a breaking change in the yaml definitions accepted by github :(

weiznich (approved):
:+1: Feel free to merge this as soon as the CI passes

Ten0:
Hmm I don't see the updated doc at https://docs.diesel.rs/master/diesel/prelude/trait.RunQueryDsl.html#method.load_iter
Is the CI supposed to deploy this automatically, or is there something else to do?

weiznich:
@Ten0  seems like the corresponding workflow failed. I will try to have a look later today or so.

weiznich:
Fixed with https://github.com/diesel-rs/diesel/commit/9a5f3346b05d2c54915a0c9ec5edba2a3d450df6. Hopefully that broke nothing else.

Issue #1636, opened 2018-04-12
Figure out a better way to deprecate old feature names


sgrif:
The current way we're deprecating the various x-column-tables feature spams builds with an unreasonable number of warnings, and causes our crate to fail to compile if we have `#![deny(warnings)]`. Ideally we show the warning in the actual invocation of `table!`, but https://github.com/rust-lang/rust/issues/49912 prevents us from just doing the obvious thing there.

We should find a better solution though, that doesn't spam logs, and in particular, generates no warnings from inside Diesel itself when compiled with these features turned on. The solution of "remember to remove `#[deny(warnings)]` when we release isn't going to work long term.

npapapietro:
What is the status of this with https://github.com/rust-lang/rust/pull/62042 fixing https://github.com/rust-lang/rust/issues/49912

weiznich:
@npapapietro Yes those fix would help here. We are happy to receive a PR that fixes this. Based on the linked rustc issues it should be possible to just add a conditional `#[deprecated]` attribute to the `table!` macro. This can be done via `#[cfg_attr]` based on the deprecated feature flags (`large-table`, `huge-table`, ...)

npapapietro:
I'll try and take this up. As I'm just diving into diesel I'd like some more info as this will be a 1st time pr. 

Let me see if if I understand this issue correctly

* You have a set of features that you would like to deprecate. Where does this list live? I see `with-deprecated` feature but it is an empty list at the moment.
* If diesel is build with one of these features we should add the `#[deprecated]` attr to the `table!` macro. Would this be injected into the `$meta` macro arg in `__diesel_parse_table` or at the `impl` of each feature case? This is where I am slightly confused.

weiznich:
> You have a set of features that you would like to deprecate. Where does this list live? I see with-deprecated feature but it is an empty list at the moment.

We would like to deprecate the `huge-tables` feature in favor of the `64-column-tables` feature flag. Also the `large-tables` feature in favor of the `32-column-tables` feature. 

> If diesel is build with one of these features we should add the #[deprecated] attr to the table! macro. Would this be injected into the $meta macro arg in __diesel_parse_table or at the impl of each feature case? This is where I am slightly confused

Adding the deprecated attribute inside the expanded `table!` macro does not work, as code in those context does not have access to the corresponding flag. (That's because the expanded code is parsed in the context of a third party crate.) 
[...]

PR #2968, opened 2021-12-05
Deprecating old feature names

npapapietro:
PR addressing https://github.com/diesel-rs/diesel/issues/1636

* Had to add `#![allow(deprecated)]` or builds will fail
* Adding listed features with deprecated message but allow build
* Life cycle of this, messages should be updated when the release is known for removing the deprecated feature all together.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -93,6 +93,7 @@
 
 #![cfg_attr(feature = "unstable", feature(trait_alias))]
 // Built-in Lints
+#![allow(deprecated)]
 #![deny(warnings)]
 #![warn(
     missing_debug_implementations,
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -93,7 +93,6 @@
 
 #![cfg_attr(feature = "unstable", feature(trait_alias))]
 // Built-in Lints
-#![allow(deprecated)]
 #![deny(warnings)]
 #![warn(
     missing_debug_implementations,
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -2,6 +2,10 @@
 #![cfg_attr(rustfmt, rustfmt_skip)] // https://github.com/rust-lang-nursery/rustfmt/issues/2755
 
 pub(crate) mod prelude {
+    #[cfg_attr(
+        any(feature = "huge-tables", feature = "large-tables"),
+        allow(deprecated)
+    )]
     #[doc(inline)]
     pub use crate::{
         allow_columns_to_appear_in_same_group_by_clause,
--- a/diesel/src/macros/mod.rs
+++ b/diesel/src/macros/mod.rs
@@ -319,6 +319,10 @@ macro_rules! __diesel_column {
 /// ```ignore
 /// pub type BoxedQuery<'a, DB, ST = SqlType> = BoxedSelectStatement<'a, ST, table, DB>;
 /// ```
+
+#[allow(deprecated)]
+#[cfg_attr(feature="huge-tables", deprecated="`huge-tables` is deprecated in favor of `64-column-tables`")]
+#[cfg_attr(feature="large-tables", deprecated="`large-tables` is deprecated in favor of `32-column-tables`")]
 #[macro_export]
 macro_rules! table {
     ($($tokens:tt)*) => {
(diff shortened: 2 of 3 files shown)


pksunkara (approved):

diesel/src/lib.rs
```diff
@@ -93,6 +93,7 @@
 
 #![cfg_attr(feature = "unstable", feature(trait_alias))]
 // Built-in Lints
+#![allow(deprecated)]
```
weiznich, line 96: This shouldn't be required. I would prefer to add a conditional (`#[cfg_attr]`) scoped allow attribute to the exact locations where the error is emitted instead.

npapapietro: I'm unsure if this is allowed.

```rust
#[cfg_attr(feature="huge-tables", allow(deprecated), deprecated="`huge-tables` is deprecated in favor of `64-column-tables`")]
#[cfg_attr(feature="large-tables", allow(deprecated), deprecated="`large-tables` is deprecated in favor of `32-column-tables`")]
#[macro_export]
macro_rules! table {
...
```

Still throws 
```
error: use of deprecated macro `table`: `large-tables` is deprecated in favor of `32-column-tables`
   --> diesel/src/pg/metadata_lookup.rs:153:1
    |
153 | table! {
    | ^^^^^
    |
note: the lint level is defined here
   --> diesel/src/lib.rs:97:9
    |
97  | #![deny(warnings)]
    |         ^^^^^^^^
    = note: `#[deny(deprecated)]` implied by `#[deny(warnings)]`
```

weiznich: I think this should be fixable by putting `#[cfg_attr(any(feature="huge-tables" feature="large-tables"), allow(deprecated)]` directly on the `table!` call mentioned in the error message. (If that does not work, just putting it as inner attribute into the corresponding (metadata_lookup) module should work)

npapapietro: To expand, I've tried

`diesel/src/lib.rs`
```rust
#[macro_use]
#[doc(hidden)]
#[allow(deprecated)]
pub mod macros;
```

`diesel/src/macros/mod.rs`
```rust
#![allow(warnings)]
#![allow(unused_parens)] // FIXME: Remove this attribute once false positive is resolved.
#![cfg_attr(rustfmt, rustfmt_skip)] // https://github.com/rust-lang-nursery/rustfmt/issues/2755
```

`diesel/src/macros/mod.rs`
```rust
#[cfg_attr(any(feature="huge-tables" feature="large-tables"), allow(deprecated)]
#[cfg_attr(feature="huge-tables", deprecated="`huge-tables` is deprecated in favor of `64-column-tables`")]
#[cfg_attr(feature="large-tables", deprecated="`large-tables` is deprecated in favor of `32-column-tables`")]
#[macro_export]
macro_rules! table {
```

And I've gotten the same build error due to deny on warnings in `src/lib.rs`

weiznich: Deprecated warnings are issued on places where a deprecated item is **used**, not where it is defined. That means we need to suppress the warning at the usage location, not at the definition side in `src/macros/mod.rs`
I think the only usage of the `table!` macro in diesel is here:
 https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/metadata_lookup.rs#L153-L167

Adding the corresponding allow annotation there should fix the compilation error.

npapapietro: This approach doesn't work either. I've tried tons of placement options. 

I'm pretty sure the `#![deny(warnings]` blocks all warnings even if allowed as well turns unblocked warnings into build errors.  As an experiment I removed it in favor of all the warnings we would normally block
```rust
#![deny(
    missing_debug_implementations,
    missing_copy_implementations,
    missing_docs,
    clippy::unwrap_used,
    clippy::print_stdout,
    clippy::wrong_pub_self_convention,
    clippy::mut_mut,
    clippy::non_ascii_literal,
    clippy::similar_names,
    clippy::unicode_not_nfc,
    clippy::enum_glob_use,
    clippy::if_not_else,
    clippy::items_after_statements,
    clippy::used_underscore_binding,
   // rest of general rust lint warnings
)]
```

and this produces the intended result. Additionally I propose adding a void, empty function that is called in `metadata_lookup.rs` that is well named so the warnings are direct and not scattered.

npapapietro:
I've pushed an alternate proposal

diesel/src/lib.rs
weiznich: Surprising  that putting the `#[allow(deprecated)]` did not work as expected. I've experimented a bit with this and pushed a more minimal version that only applies `#[allow(deprecated)]` where it is really needed and put that behind the deprecated feature flags itself.

PR #2972, opened 2021-12-10
Better error message span

pksunkara:


pksunkara:
It looks like any other approaches I can come up with are similarly bad, therefore I went with your commit. The `postgres_type` had an easier implementation though since we validate before we build the type itself.

This PR is ready.

weiznich (approved):

PR #2975, opened 2021-12-13
Improve error messages for `table!`

weiznich:
We now use an explicit `compile_error!` call to produce a meaningful
error message if a user provided an unsupported number of columns for a
specific table. Before this change we would fail with a large error
message about unimplemented trait bounds.


pksunkara (approved):

weiznich:
@pksunkara I had to do some fundamental changes to make this work for all feature sizes. You may want to have another look here.

pksunkara (approved):
Looks good!

PR #2993, opened 2021-12-20
FIX metadata lookup with no schema failing on non-lowercase types

### LINKED ISSUE #2992: Metadata lookup with no schema fails on non-lowercase types
### Versions

- **Rust:** latest stable
- **Diesel:** latest master
- **Database:** postgres
- **Operating System:** Linux

## Problem Description
I'm [currently updating diesel-derive-enum](https://github.com/adwhit/diesel-derive-enum/pull/66) to support the latest diesel master version.
It fails tests, because, probably since https://github.com/diesel-rs/diesel/commit/887be77536ae855ade3a85aa3c0875c597d23db6, the method employed to resolve these types decides to lowercase when looking up.

https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/metadata_lookup.rs#L91-L97
```
SELECT "pg_type"."oid", "pg_type"."typarray" FROM "pg_type" WHERE ("pg_type"."oid" = $1::regtype::oid) -- binds: ["Stylized_External_Type"]
```

*resolves to the existing type*

```
    DatabaseError(
        Unknown,
        "type \"stylized_external_type\" does not exist",
    ),
```

### Steps to reproduce

I've created #2993 so that the tests make the issue obvious.
We need to add the fix<s> - but I'm not sure what it should be</s> EDIT: Found a fix.

## Possibly linked

#2948

Ten0:
Fix #2992

### DIFF
--- a/diesel/src/pg/metadata_lookup.rs
+++ b/diesel/src/pg/metadata_lookup.rs
@@ -90,9 +90,9 @@ fn lookup_type<T: Connection<Backend = Pg>>(
         // and let postgres figure out the name resolution on it's own.
         metadata_query
             .filter(
-                pg_type::oid.eq(crate::dsl::sql("")
+                pg_type::oid.eq(crate::dsl::sql("quote_ident(")
                     .bind::<crate::sql_types::Text, _>(&cache_key.type_name)
-                    .sql("::regtype::oid")),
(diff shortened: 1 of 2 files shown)


weiznich (approved):
:+1: from my side if the CI passes.

PR #2998, opened 2021-12-22
Introduce a public API to set values for `Output` directly

weiznich:
This satifies two use cases:

* Using temporary values as part of the `ToSql<T, Sqlite>` implemenation
* Allowing third parties to provide their own custom `BindCollector`
impl that does not implement `Write`.

### DIFF
--- a/diesel/src/sqlite/connection/bind_collector.rs
+++ b/diesel/src/sqlite/connection/bind_collector.rs
@@ -18,7 +18,7 @@ impl SqliteBindCollector<'_> {
 /// This type represents a value bound to
 /// an sqlite prepared statement
 ///
-/// It can be constructed via the various `From<T>` implemenations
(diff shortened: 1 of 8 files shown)


Ten0 (approved):
LGTM :)

diesel/src/sqlite/connection/bind_collector.rs
```diff
@@ -15,8 +15,87 @@ impl SqliteBindCollector<'_> {
     }
 }
 
+/// This type represents a value bound to
+/// an sqlite prepared statement
+///
+/// It can be constructed via the various `From<T>` implemenations
```
JohnTitor, line 21:
```suggestion
/// It can be constructed via the various `From<T>` implementations
```

JohnTitor (approved):
Looks reasonable to me with one nit 👍

PR #3004, opened 2021-12-26
Add Sqlite support for the RETURNING clause

toxeus:
Add Sqlite support for the `RETURNING` clause by leveraging
the implementation for Postgres. So this change is essentially
just adjusting the build system to enable the support if a
Sqlite version which supports the `RETURNING` clause [1] is
complied against.

The "insert record with returning" unit tests for Postgres
couldn't be re-used because Sqlite doesn't support batch insert.
Because those tests provide value for Postgres, new versions
for Sqlite have been created which don't use batch insert.

[1] https://sqlite.org/releaselog/3_35_0.html

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -61,6 +61,7 @@ network-address = ["ipnetwork", "libc"]
 numeric = ["num-bigint", "bigdecimal", "num-traits", "num-integer"]
 postgres_backend = ["diesel_derives/postgres", "bitflags", "byteorder"]
 mysql_backend = ["diesel_derives/mysql", "byteorder"]
+returning_clauses_for_sqlite_3_35 = []
 
 [package.metadata.docs.rs]
 features = ["postgres", "mysql", "sqlite", "extras"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,8 +263,15 @@ jobs:
           command: test
           args: --manifest-path diesel_tests/Cargo.toml --no-default-features --features "${{ matrix.backend }} unstable"
 
+      - name: Run diesel_tests (beta)
+        if: matrix.run == 'beta'
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --manifest-path diesel_tests/Cargo.toml --no-default-features --features "${{ matrix.backend }},returning_clauses_for_sqlite_3_35"
+
       - name: Run diesel_tests
-        if: matrix.run != 'nightly'
+        if: matrix.run == 'stable'
         uses: actions-rs/cargo@v1
         with:
           command: test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Added support for all the derive attributes being inside `#[diesel(...)]`
 
+* Added support for `RETURNING` expressions for Sqlite via the `returning_clauses_for_sqlite_3_35` feature
+
 ### Removed
 
 * All previously deprecated items have been removed.
--- a/diesel/src/sqlite/backend.rs
+++ b/diesel/src/sqlite/backend.rs
@@ -54,7 +54,10 @@ impl TypeMetadata for Sqlite {
 }
 
 impl SqlDialect for Sqlite {
+    #[cfg(not(feature = "returning_clauses_for_sqlite_3_35"))]
     type ReturningClause = sql_dialect::returning_clause::DoesNotSupportReturningClause;
+    #[cfg(feature = "returning_clauses_for_sqlite_3_35")]
+    type ReturningClause = sql_dialect::returning_clause::PgLikeReturningClause;
 
     type OnConflictClause = SqliteOnConflictClaues;
 
--- a/diesel_tests/Cargo.toml
+++ b/diesel_tests/Cargo.toml
@@ -27,6 +27,7 @@ unstable = ["diesel/unstable"]
 postgres = ["diesel/postgres"]
 sqlite = ["diesel/sqlite"]
 mysql = ["diesel/mysql"]
+returning_clauses_for_sqlite_3_35 = ["diesel/returning_clauses_for_sqlite_3_35"]
 
 [[test]]
 name = "integration_tests"
--- a/diesel_tests/tests/delete.rs
+++ b/diesel_tests/tests/delete.rs
@@ -29,7 +29,10 @@ fn delete_single_record() {
 }
 
 #[test]
-#[cfg(not(any(feature = "sqlite", feature = "mysql")))]
+#[cfg(not(any(
+    all(feature = "sqlite", not(feature = "returning_clauses_for_sqlite_3_35")),
+    feature = "mysql"
+)))]
 fn return_deleted_records() {
     use crate::schema::users::dsl::*;
     let connection = &mut connection_with_sean_and_tess_in_users_table();
--- a/diesel_tests/tests/insert.rs
+++ b/diesel_tests/tests/insert.rs
@@ -152,6 +152,26 @@ fn insert_records_as_boxed_static_array() {
 }
 
 #[test]
+#[cfg(all(feature = "sqlite", feature = "returning_clauses_for_sqlite_3_35"))]
+fn insert_record_using_returning_clause() {
+    use crate::schema::users::table as users;
+    let connection = &mut connection();
+    let new_user = &NewUser::new("Sean", Some("Black"));
+
+    let inserted_user = insert_into(users)
+        .values(new_user)
+        .get_result::<User>(connection)
+        .unwrap();
+    let expected_user = User {
+        id: inserted_user.id,
+        name: "Sean".to_string(),
+        hair_color: Some("Black".to_string()),
+    };
+
+    assert_eq!(expected_user, inserted_user);
+}
+
+#[test]
 #[cfg(not(any(feature = "sqlite", feature = "mysql")))]
 fn insert_records_using_returning_clause() {
     use crate::schema::users::table as users;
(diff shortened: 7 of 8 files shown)


diesel/build.rs
```diff
@@ -0,0 +1,11 @@
+fn main() {
+    #[cfg(feature = "sqlite")]
+    {
+        #[cfg(feature = "libsqlite3-sys/with-asan")]
+        println!("cargo:rustc-cfg=sqlite_supports_returning");
+        #[cfg(not(feature = "libsqlite3-sys/with-asan"))]
```
toxeus, line 6: The ASAN feature is causing linker errors for the build script. Unfortunately, it doesn't seem possible to compile code conditionally based on features of a dependency. I think this is a dead end unless someone has an idea how to work around this.

I see two alternative approaches that should work:

1. I could implement a proc macro attribute instead which would eliminate the build script and, as a consequence, the ASAN linker errors, too. But the proc macro would then depend on `libsqlite3-sys`. Also, it would require the introduction of a new crate because it doesn't really fit in `diesel_derives` nor in `diesel_migrations`.
2. We add an additional feature flag as [discussed here](https://github.com/diesel-rs/diesel/discussions/2684#discussioncomment-664045) along with the disadvantages it comes along.

@weiznich what do you think?

weiznich: The first solution would have the same problem as the build script based approach. You would need to build sqlite for the host target with different flags than for the target system. The asan feature passes the the corresponding flags to the compiler unconditionally for both targets. That will remain the same if its a proc macro instead. That in turn means that a feature flag based solution is the only possible solution, at least as long as we dont emulate the returning clause on older sqlitr versions somehow...

toxeus: > at least as long as we dont emulate the returning clause on older sqlitr versions somehow...

I was pondering the same. Old sqlite version could implicitly implement `explicit_returning()` from `examples/sqlite/all_about_inserts/src/lib.rs`. But I see two major downsides:

1. We're hiding from the user that we're opening a transaction and that we're hitting the db twice. Users can end up nesting transactions unknowingly. 
2. Users can choose `id` to be whatever. So we cannot rely it to be an incrementing integer. There could be tables that don't have an `id` column at all. This renders the `explicit_returning()` code too fragile for a general use case.

Old sqlite users have no other choice but to implement something a la `explicit_returning()` I'm afraid. 

I'll refactor the PR to use a feature flag.

weiznich: > We're hiding from the user that we're opening a transaction and that we're hitting the db twice. Users can end up nesting transactions unknowingly

I haven't tried that yet, but maybe a CTE could help there to avoid the transaction.

> Users can choose id to be whatever. So we cannot rely it to be an incrementing integer. There could be tables that don't have an id column at all. This renders the explicit_returning() code too fragile for a general use case.

We could use the "hidden" rowid column for that.

Both combined would result in the following query:

```sql
WITH 
INSERT ... AS insert
SELECT returning FROM table ORDER BY rowid DESC LIMIT count
```

toxeus: I've already checked that earlier. `rowid` is not guaranteed to exist :(

image

toxeus:
The tests currently pass but they never run with the `returning_clauses_for_sqlite_3_35` feature enabled. What is a good test candidate to enable that feature?

diesel/build.rs
weiznich: I know. The important point for me is that it exists by default. Thant means people that are unaware about the details will not disable it. I would expect that only a minority disables this feature and that this minority is very aware of technical details. That means documenting this behavior may be enough. There are already a few other places in diesel that will not work if you use a too old db version, that wouldn't be that different in my opinion.

weiznich: I should probably add that I would like to experiment with this a bit, but that will likely need to wait till mid of January.

weiznich:
> The tests currently pass but they never run with the returning_clauses_for_sqlite_3_35 feature enabled. What is a good test candidate to enable that feature?

That's exactly the issue with adding more features. Theoretically that expands test matrix with an additional dimension. That would mean we need to run all tests with this feature enabled and with the feature disabled. That will not work practically. 
The best workaround that I see is to use one of the rust channel, for example the beta channel and enable the flag only there.

diesel/build.rs
toxeus: Alright, thx for the heads up!

toxeus:
> That's exactly the issue with adding more features.

Agreed. Sqlite's commitment to backwards compatibility is a blessing and a curse at the same time. I wish they would at least define which minor versions can be treated as EOL. For example the suggested workaround above using CTE also runs into a similar issue because CTE is only available from a certain version on.

> The best workaround that I see is to use one of the rust channel

Sometimes, the best solution is a pragmatic one :)

mainrs:
I get compilation errors when working on your branch:

```$console
$ cargo build
error[E0277]: the trait bound `ReturningClause<(columns::id, columns::name, columns::hashed_password)>: QueryFragment<Sqlite, DoesNotSupportReturningClause>` is not satisfied
  --> src/routes/auth.rs:46:14
   |
46 |             .get_result(conn)
   |              ^^^^^^^^^^ the trait `QueryFragment<Sqlite, DoesNotSupportReturningClause>` is not implemented for `ReturningClause<(columns::id, columns::name, columns::hashed_password)>`
   |
             <ReturningClause<Expr> as QueryFragment<DB, PgLikeReturningClause>>
             <ReturningClause<Expr> as QueryFragment<DB>>
   = note: required because of the requirements on the impl of `QueryFragment<Sqlite>` for `ReturningClause<(columns::id, columns::name, columns::hashed_password)>`
   = note: 1 redundant requirement hidden
   = note: required because of the requirements on the impl of `QueryFragment<Sqlite>` for `InsertStatement<table, ValuesClause<(DefaultableColumnInsertValue<ColumnInsertValue<columns::name, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>, DefaultableColumnInsertValue<ColumnInsertValue<columns::hashed_password, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>), table>, diesel::query_builder::insert_statement::Insert, ReturningClause<(columns::id, columns::name, columns::hashed_password)>>`
   = note: required because of the requirements on the impl of `LoadQuery<'_, SqliteConnection, _>` for `InsertStatement<table, ValuesClause<(DefaultableColumnInsertValue<ColumnInsertValue<columns::name, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>, DefaultableColumnInsertValue<ColumnInsertValue<columns::hashed_password, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>), table>>`

error[E0277]: the trait bound `(i64, std::string::String, std::string::String): FromStaticSqlRow<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text), Sqlite>` is not satisfied
  --> src/routes/auth.rs:46:14
   |
46 |             .get_result(conn)
   |              ^^^^^^^^^^ the trait `FromStaticSqlRow<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text), Sqlite>` is not implemented for `(i64, std::string::String, std::string::String)`
   |
             <(T1, T2, T0) as FromStaticSqlRow<(ST1, ST2, ST0), __DB>>
note: required because of the requirements on the impl of `Queryable<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text), Sqlite>` for `User`
  --> src/db/auth.rs:11:10
   |
11 | #[derive(Queryable)]
   |          ^^^^^^^^^
12 | pub struct User {
   |            ^^^^
   = note: required because of the requirements on the impl of `FromSqlRow<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text), Sqlite>` for `User`
   = note: required because of the requirements on the impl of `CompatibleType<User, Sqlite>` for `(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text)`
   = note: required because of the requirements on the impl of `LoadQuery<'_, SqliteConnection, User>` for `InsertStatement<table, ValuesClause<(DefaultableColumnInsertValue<ColumnInsertValue<columns::name, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>, DefaultableColumnInsertValue<ColumnInsertValue<columns::hashed_password, diesel::expression::bound::Bound<diesel::sql_types::Text, std::string::String>>>), table>>`
   = note: this error originates in the derive macro `Queryable` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `blanch-server` due to 2 previous errors
```

toxeus:
@mainrs it's impossible to tell what your issue is when you only share the error output. You should provide a *complete and preferably minimal* example if you want to receive meaningful help. See [here](https://stackoverflow.com/help/minimal-reproducible-example) to learn more.

weiznich:
@toxeus The currently minimal supported sqlite version is 3.20.0. This is defined by the binary interface we use (In detail the `sqlite3_prepare_v3` function). CTE's are supported since 3.8.3, so much longer. That wouldn't be a problem. 
That written: I've tried the statement I've suggested above and I didn't manage to get sqlite to accept the corresponding SQL at all. It seems like you cannot have a insert statement in a CTE expression as part of the `WITH` clause.
That means the only option that is left is adding a feature flag for this feature.

diesel/Cargo.toml
```diff
@@ -61,6 +61,7 @@ network-address = ["ipnetwork", "libc"]
 numeric = ["num-bigint", "bigdecimal", "num-traits", "num-integer"]
 postgres_backend = ["diesel_derives/postgres", "bitflags", "byteorder"]
 mysql_backend = ["diesel_derives/mysql", "byteorder"]
+returning_clauses_for_sqlite_3_35=[]
```
weiznich, line 64:
```suggestion
returning_clauses_for_sqlite_3_35 = []
```

diesel_compile_tests/tests/fail/delete_statement_does_not_support_returning_methods_on_sqlite.rs
```diff
@@ -1,22 +0,0 @@
-extern crate diesel;
```
weiznich, line 1: I'm not sure if we should remove this tests. We don't enable the new feature as part of the `diesel_compile_tests` build, right?

weiznich (approved):
Looks good for me if the small nits are solved + a changelog entry is added

diesel/Cargo.toml
toxeus: done

diesel_compile_tests/tests/fail/delete_statement_does_not_support_returning_methods_on_sqlite.rs
toxeus: yes, right. done

toxeus:
As discussed above, I have enabled the feature on the beta channel so we have some test coverage.

PR #3005, opened 2021-12-30
Tighten minimum required version for libsqlite3-sys

toxeus:
`libsqlite3-sys` is built with the `bundled_bindings`
feature enabled which was introduced in commit [1]
which in turn is part of `libsqlite3-sys` 0.17.2.

Versions below 0.17.2 will hence fail to compile.
Therefore, tighten the minimum required version.

[1] https://github.com/rusqlite/rusqlite/commit/c70d14854282ccddcf9c42d2078eebad54eb7c86<br>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,8 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
   with the database. Beside of the changed signature, existing impls of this trait should remain unchanged in almost 
   all cases.
 
+* The minimal supported version of libsqlite3-sys is now 0.17.2.
+
 ### Fixed
 
 * Many types were incorrectly considered non-aggregate when they should not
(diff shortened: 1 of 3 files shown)


weiznich (approved):
:+1: From me with an added changelog entry as this is a potentially breaking change.

toxeus:
@weiznich I don't see how this is a breaking change. You can test it yourself; just set `version = "<0.17.2"` and you can observe the following for `cargo c -p diesel`
image

(While taking the screenshot I have noticed that the minimum required version is 0.17.2 and not 0.17.3 as suggested by the tags in the rusqlite repo. I should have tried to infer the libsqlite3-sys version from Cargo.toml instead. I fixed the PR accordingly)

weiznich:
Thats true for the current master version. The 1.4.x release didn't use the `bundled_bindings` feature and therefore supported the older versions as well. Its now a breaking change because any crate can depend only on one version of libsqlite3-sys as it links a c library. So if down stream crate depends on 0.8 it will work with diesel 1.4.x but not with 2.0.
That is all no large problem as we do a major version bump anyway. I just want to state this fact as part of the changelog, probably as 

> The minimal supported version of libsqlite3-sys is now 0.17.2.

toxeus:
Oic, you meant breaking change with respect to version 1.x. I understood you meant that this was a breaking change on master. Makes sense then :+1:

PR #3011, opened 2022-01-04
Update benchmark dependencies

weiznich:
* Allow to use sqlx 0.5.10 which hopefully should fix the sqlite
benchmarks
* Update sea-orm to 0.5
* Remove quaint/sqlite as they use a old libsqlite3 version

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -76,7 +76,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: bench
-          args: --manifest-path diesel_bench/Cargo.toml --no-default-features --features "${{matrix.backend}} sqlx-bench sqlx/${{matrix.backend}} tokio rusqlite futures sea-orm sea-orm/sqlx-${{matrix.backend}} criterion/async_tokio quaint quaint/sqlite quaint/serde-support serde"
+          args: --manifest-path diesel_bench/Cargo.toml --no-default-features --features "${{matrix.backend}} sqlx-bench sqlx/${{matrix.backend}} tokio rusqlite futures sea-orm sea-orm/sqlx-${{matrix.backend}} criterion/async_tokio"
 
       - name: Run Benchmarks (Mysql)
         if: matrix.backend == 'mysql'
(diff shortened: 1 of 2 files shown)


.github/workflows/metrics.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: bench
-          args: --manifest-path diesel_bench/Cargo.toml --no-default-features --features "${{matrix.backend}} sqlx-bench sqlx/${{matrix.backend}} tokio rusqlite futures sea-orm sea-orm/sqlx-${{matrix.backend}} criterion/async_tokio quaint quaint/sqlite quaint/serde-support serde"
```
BobbyByrne, line 79: Is serde only used in the quaint dependency here, which is why it's also removed?

weiznich: Thats correct. Quaint uses serde to deserialize database values.

BobbyByrne: is the intent to move the other benchmarks away from quaint as well or just sqlite due to the old libsqlite3?

weiznich: I only want to remove quaint from the sqlite benchmark as all other benchmarks work fine.
The sqlite benchmark does not compile if the sqlx or rusqulite feature is enabled + quaint is used because quiant uses an older version of libsqlite3-sys. Cargo does not allow multiple versions of the same crate in the dependency tree if that crate links a c library.

BobbyByrne: ah, gotcha. I think that's all my questions answered. What else would you like me to do as a reviewer on this?

Ten0 (approved):

.github/workflows/metrics.yml
weiznich: The general review flow is as following:
1. You look at the code and ask anything that you want to know about the change
2. If you see any possible improvements/bugs/whatever… note them as comment (as you already have done)
3. The "File changes" view has a "Review changes" button on the top right that allows you to put in a final rating of the PR. You can either approve the changes (if there are no open/only trivial questions left) or request changes (if there are things that you want to clarify)
4. Depending on the outcome of 3 the PR get's merged (approved) or updated (requested changes). In the second case we start with 1 again.

PR #3017, opened 2022-01-13
Fix #3016

### LINKED ISSUE #3016: R2d2 connection pool may hang on droped packages
### Versions

- **Rust:** rustc 1.56.1 (likely irrelevant)
- **Diesel:** 1.4.8 
- **Database:** postgresql
- **Operating System** Ubuntu 21.10 (likely irrelevant)

### Feature Flags

- **diesel:** postgres r2d2

## Problem Description

Cargo.toml
```toml
[package]
name = "crates-io-hang-reproduction"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
diesel = { version = "1.4.8", features = ["r2d2", "postgres"] }
dotenv = "0.15.0"
parking_lot = "0.11.2"
thiserror = "1.0.30"
url = "2.2.2"

```

main.rs: 
```rust
use std::time::{Instant, Duration};

use diesel::RunQueryDsl;

#[macro_use]
extern crate diesel;

pub mod db {
    use diesel::prelude::*;
    use diesel::r2d2::{self, ConnectionManager, CustomizeConnection};
    use parking_lot::{ReentrantMutex, ReentrantMutexGuard};
    use std::sync::Arc;
    use std::{ops::Deref, time::Duration};
    use thiserror::Error;
    use url::Url;

    #[derive(Clone)]
    pub enum DieselPool {
        Pool {
            pool: r2d2::Pool<ConnectionManager<PgConnection>>,
        },
        Test(Arc<ReentrantMutex<PgConnection>>),
    }

    impl DieselPool {
        pub(crate) fn new(
            url: &str,
            config: r2d2::Builder<ConnectionManager<PgConnection>>,
        ) -> Result<DieselPool, PoolError> {
            let manager = ConnectionManager::new(connection_url(url));

            // For crates.io we want the behavior of creating a database pool to be slightly different
            // than the defaults of R2D2: the library's build() method assumes its consumers always
            // need a database connection to operate, so it blocks creating a pool until a minimum
            // number of connections is available.
            //
            // crates.io can actually operate in a limited capacity without a database connections,
            // especially by serving download requests to our users. Because of that we don't want to
            // block indefinitely waiting for a connection: we instead need to wait for a bit (to avoid
            // serving errors for the first connections until the pool is initialized) and if we can't
            // establish any connection continue booting up the application. The database pool will
            // automatically be marked as unhealthy and the rest of the application will adapt.
            let pool = DieselPool::Pool {
                pool: config.build_unchecked(manager),
            };
            match pool.wait_until_healthy(Duration::from_secs(5)) {
                Ok(()) => {}
                Err(PoolError::UnhealthyPool) => {}
                Err(err) => return Err(err),
            }

            Ok(pool)
        }

        pub(crate) fn new_test(url: &str) -> DieselPool {
            let conn = PgConnection::establish(&connection_url(url))
                .expect("failed to establish connection");
            conn.begin_test_transaction()
                .expect("failed to begin test transaction");
            DieselPool::Test(Arc::new(ReentrantMutex::new(conn)))
        }

        pub fn get(&self) -> Result<DieselPooledConn<'_>, PoolError> {
            match self {
                DieselPool::Pool { pool } => {
                    if let Some(conn) = pool.try_get() {
                        Ok(DieselPooledConn::Pool(conn))
                    } else if !self.is_healthy() {
                        Err(PoolError::UnhealthyPool)
                    } else {
                        Ok(DieselPooledConn::Pool(pool.get()?))
                    }
                }
                DieselPool::Test(conn) => Ok(DieselPooledConn::Test(conn.lock())),
            }
        }

        pub fn state(&self) -> PoolState {
            match self {
                DieselPool::Pool { pool, .. } => {
                    let state = pool.state();
                    PoolState {
                        connections: state.connections,
                        idle_connections: state.idle_connections,
                    }
                }
                DieselPool::Test(_) => PoolState {
                    connections: 0,
                    idle_connections: 0,
                },
            }
        }

        pub fn wait_until_healthy(&self, timeout: Duration) -> Result<(), PoolError> {
            match self {
                DieselPool::Pool { pool, .. } => match pool.get_timeout(timeout) {
                    Ok(_) => Ok(()),
                    Err(_) if !self.is_healthy() => Err(PoolError::UnhealthyPool),
                    Err(err) => Err(PoolError::R2D2(err)),
                },
                DieselPool::Test(_) => Ok(()),
            }
        }

        fn is_healthy(&self) -> bool {
            self.state().connections > 0
        }
    }

    #[derive(Debug, Copy, Clone)]
    pub struct PoolState {
        pub connections: u32,
        pub idle_connections: u32,
    }

    pub enum DieselPooledConn<'a> {
        Pool(r2d2::PooledConnection<ConnectionManager<PgConnection>>),
        Test(ReentrantMutexGuard<'a, PgConnection>),
    }

    impl Deref for DieselPooledConn<'_> {
        type Target = PgConnection;

        fn deref(&self) -> &Self::Target {
            match self {
                DieselPooledConn::Pool(conn) => conn.deref(),
                DieselPooledConn::Test(conn) => conn.deref(),
            }
        }
    }

    pub fn connect_now() -> ConnectionResult<PgConnection> {
        let url = connection_url(&std::env::var("DATABASE_URL").unwrap());
        PgConnection::establish(&url)
    }

    pub fn connection_url(url: &str) -> String {
        let mut url = Url::parse(url).expect("Invalid database URL");
        if dotenv::var("HEROKU").is_ok() && !url.query_pairs().any(|(k, _)| k == "sslmode") {
            url.query_pairs_mut().append_pair("sslmode", "require");
        }
        url.into()
    }

    #[derive(Debug, Clone, Copy)]
    pub struct ConnectionConfig {
        pub statement_timeout: u64,
        pub read_only: bool,
    }

    impl CustomizeConnection<PgConnection, r2d2::Error> for ConnectionConfig {
        fn on_acquire(&self, conn: &mut PgConnection) -> Result<(), r2d2::Error> {
            use diesel::sql_query;

            sql_query(format!(
                "SET statement_timeout = {}",
                self.statement_timeout * 1000
            ))
            .execute(conn)
            .map_err(r2d2::Error::QueryError)?;
            if self.read_only {
                sql_query("SET default_transaction_read_only = 't'")
                    .execute(conn)
                    .map_err(r2d2::Error::QueryError)?;
            }
            Ok(())
        }
    }

    #[cfg(test)]
    pub(crate) fn test_conn() -> PgConnection {
        let conn = PgConnection::establish(&crate::env("TEST_DATABASE_URL")).unwrap();
        conn.begin_test_transaction().unwrap();
        conn
    }

    #[derive(Debug, Error)]
    pub enum PoolError {
        #[error(transparent)]
        R2D2(#[from] r2d2::PoolError),
        #[error("unhealthy database pool")]
        UnhealthyPool,
    }
}

table! {
    users {
        id -> Integer,
        name -> Text,
    }
}

fn main() {
    let db_url = std::env::var("DATABASE_URL").unwrap();
    let pool = self::db::DieselPool::new(&db_url, Default::default()).unwrap();
    loop {
        println!("Start loop");
        let loop_start = Instant::now();
        let conn = pool.get().unwrap();
        println!("Took {} ms to get connection", loop_start.elapsed().as_millis());

        let _: (i32, String) = users::table.first(&*conn).unwrap();
        println!("Sleep 1s");
        std::thread::sleep(Duration::from_secs(1));
        println!("Finish sleep");
    }
}

```

Using the following reproduction steps produces an application freeze:

1. `DATABASE_URL=… cargo run`
2. `sudo iptables -A OUTPUT -p tcp --dport 5432 -j DROP`

I expect a panic due to the unwrap in line 199. 

The output stops + the application seems to hang

### Steps to reproduce

Using the following reproduction steps produces an application freeze:

1. `DATABASE_URL=… cargo run`
2. `sudo iptables -A OUTPUT -p tcp --dport 5432 -j ACCEPT`

## Checklist

weiznich:
This demonstrate a potential fix for #3016 by using the async libpq
interface + busy polling. This allows us to abort polling as soon as we
reach the timeout.
I see multiple issues with this approach:
* Add's new methods to `Connection` (`get_timeout`/`set_timeout`). This
  could likely be avoided by exposing them directly on `PgConnection`.
On the other hand that would, at least for 1.4.x imply that we cannot
use those timeout only for the `is_valid` function. It would be all or
nothing then.
* This uses busy polling, which could be quite inefficient. I'm not sure
  what better posibilities there are without pulling in
tokio/mio/async_std.

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -190,4 +190,12 @@ pub trait Connection: SimpleConnection + Sized + Send {
 
     #[doc(hidden)]
     fn transaction_manager(&self) -> &Self::TransactionManager;
+
+    #[doc(hidden)]
+    fn set_timeout(&mut self, _duration: Option<std::time::Duration>) {}
+
+    #[doc(hidden)]
+    fn get_timeout(&self) -> Option<std::time::Duration> {
+        None
+    }
 }
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -105,6 +105,14 @@ impl Connection for PgConnection {
     fn transaction_manager(&self) -> &Self::TransactionManager {
         &self.transaction_manager
     }
+
+    fn set_timeout(&mut self, duration: Option<std::time::Duration>) {
+        self.raw_connection.timeout = duration;
+    }
+
+    fn get_timeout(&self) -> Option<std::time::Duration> {
+        self.raw_connection.timeout
+    }
 }
 
 impl PgConnection {
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -13,6 +13,7 @@ use result::*;
 #[allow(missing_debug_implementations, missing_copy_implementations)]
 pub struct RawConnection {
     internal_connection: NonNull<PGconn>,
+    pub(crate) timeout: Option<std::time::Duration>,
 }
 
 impl RawConnection {
@@ -28,6 +29,7 @@ impl RawConnection {
                 let connection_ptr = unsafe { NonNull::new_unchecked(connection_ptr) };
                 Ok(RawConnection {
                     internal_connection: connection_ptr,
+                    timeout: None,
                 })
             }
             _ => {
@@ -72,7 +74,10 @@ impl RawConnection {
         param_formats: *const libc::c_int,
         result_format: libc::c_int,
     ) -> QueryResult<RawResult> {
-        let ptr = PQexecPrepared(
+        if self.timeout.is_some() {
+            PQsetnonblocking(self.internal_connection.as_ptr(), 1);
+        }
+        let res = PQsendQueryPrepared(
             self.internal_connection.as_ptr(),
             stmt_name,
             param_count,
@@ -81,7 +86,17 @@ impl RawConnection {
             param_formats,
             result_format,
         );
-        RawResult::new(ptr, self)
+        if res != 1 {
+            return Err(Error::DatabaseError(
+                DatabaseErrorKind::UnableToSendCommand,
+                Box::new(self.last_error_message()),
+            ));
+        }
+        let res = self.pq_exec_finish();
+        if self.timeout.is_some() {
+            PQsetnonblocking(self.internal_connection.as_ptr(), 0);
+        }
+        res
     }
 
     pub unsafe fn prepare(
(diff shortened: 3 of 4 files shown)


Ten0:
Haven't read in detail but not fond of active waiting either ^^
If PQexec manages to not wait actively perhaps it would be possible through the async implementation too?

In the end, isn't this a bug of [`pqWait`](https://github.com/postgres/postgres/blob/27b77ecf9f4d5be211900eda54d8155ada50d696/src/interfaces/libpq/fe-misc.c#L1003) that should be fixed by the postgres team if the timeout settings don't work?

weiznich:
> Haven't read in detail but not fond of active waiting either ^^ 

I'm not fond of it either. Basically I've just put this together to have some basis to see how to fix the underlying problem. A real solution should likely use `poll` or something similar there to wait. Before spending the time to implement that I would like to wait for a response from the crates.io team here to better understand their actual needs for a fix here.
To be honest here: I would likely dismiss this issue as following if this was not raised by the crates.io team: That's just doing what blocking io is designed to do. If you care about such things consider using the async prototype. 
As this is raised by the crates.io team I consider this to be something we should support for them as crates.io is somewhat critical for rust :see_no_evil: 

To be a bit more concrete about possible poll strategies here:

* Pull in `mio` (or a similar crate) and use their poll functions to wait with a timeout there.
* Pull in `tokio` to use the async ecosystem for timeouts (large dependency impact :disappointed: )
* Don't pull in anything other than `libc` and use the exposed unsafe `poll`/`select` functions  (small dependency impact, harder to implement)

As a side note: I've tried a similar setup with the postgres crate with a timeout set there at connection level and it had exactly the same problem :disappointed: 

> In the end, isn't this a bug of pqWait/pqWaitTimed that should be fixed by the postgres team if the timeout settings don't work?

That's correct, but unfortunately it seems like the postgres team dismissed feature requests for doing that in the past, while explicitly pointing to the async interface for those cases.. See the linked issues for details. Even that would not be that bad if `pqWait`/`pqWaitTimed` would be public functions, but unfortunately they are private to libpq so we cannot use them.

pietroalbini:
Hi, and sorry for the delay!

We discussed what approaches crates.io should take in the last crates.io team meeting, and switching the whole application to async is not something we'll have the human bandwidth for in the foreseeable future. If you could implement this (or something like this) in the non-async API it'd be awesome :heart:

weiznich:
@pietroalbini I totally understand that a change to use the fully async implementation would be quite a major change.

There are a few other things I would like to know:
* Does the crates.io team prefer a timely fix (something available in the next 2 weeks) using the 1.4.x release or would you be fine with having this fixed in the upcoming 2.0 release which should hopefully have a first pre release in the march or so?
* If you prefer a fix based on 1.4.x, do you want a release for that or would you be fine with having a branch with the fixed version somewhere?
* How urgent is the fix for you? Is it important for the crates.io team to see a fix in hours or months?
* Does the crates.io team have opinions on how the polling should be implemented? Via busy polling? Via epoll, mio or tokio?

pietroalbini:
Replying to the rest of the comment in the coming days (in a rush now), but

> How urgent is the fix for you? Is it important for the crates.io team to see a fix in hours or months?

Weeks/months. This is not happening during normal operations, but in case of infrastructure outages: I don't think there's a need to drop everything to work on this.

weiznich:
Closed as this can be handled by libpq via the `tcp_user_timeout` paramater.

PR #3021, opened 2022-01-16
Add ignored mysql serialization issue tests.

Emm:


weiznich:
Thanks for providing this tests. I will try to have a look in the next few weeks.

weiznich:
I've added a "fix" here that allows both tests to pass. This "fix" just tracks at connection level if we hit a `SerializationError` or not and special cases the rollback correspondingly. At least from my point of view that seems like a hacky solution. Additionally it only works with `Connection::execute_returning_count` but not with `Connection::load` due to borrowcheck issues. (That's maybe even not a problem because `SerializationFailure` should only happen with insert/update/delete statements which will use `.execute()` on the mysql backend as there is no `RETURNING` clause support.)

@diesel-rs/reviewers Would like to hear your opinions here.

diesel/src/connection/transaction_manager.rs
```diff
@@ -200,6 +223,34 @@ where
                 transaction_depth.get() - 1
             )),
         };
+
+        if let Some(msg) = transaction_state.previous_serialization_error.take() {
+            // we can safely ignore the result here
+            // as the only other error than the "rollback error" variants
+            // is failing to load the transaction state, but that's something
+            // we have already done above
```
chusteven, line 231: > This "fix" just tracks at connection level if we hit a SerializationError or not and special cases the rollback correspondingly. At least from my point of view that seems like a hacky solution.

Curious what you mean by hacky? Or at least how the same logic could be achieved alternately? Mostly asking because this makes sense to me (basically an early return of sorts), and especially in the context of transactions/isolation levels -- any violation of which I would assume the right "correction" would be to immediately rollback what could be rolled back and then just bail on the rest, returning to the caller an error

chusteven (commented):
Wow these tests are pretty neat :D 

Generally speaking, the tests look legit (seem to follow the same pattern as the Postgres ones?) and as far as the feedback requested re: whether or not the fix in commit https://github.com/diesel-rs/diesel/pull/3021/commits/e91d048e3a75559312a0c8fe66fac7f0a03f1ebc is hacky or not, I think it's fine :) -- in the future, if more error types start necessitating lots of early returns then I think that the `rollback_transaction` function would still be a natural place to take responsibility for these errors

diesel/src/connection/transaction_manager.rs
```diff
@@ -40,6 +40,28 @@ pub trait TransactionManager<Conn: Connection> {
     /// inside of a transaction, and that operations are not run in a `InError`
     /// transaction manager.
     fn transaction_manager_status_mut(conn: &mut Conn) -> &mut TransactionManagerStatus;
+
+    /// Executes the given function inside of a database transaction
+    ///
+    /// Each implementation of this function needs to fullfill the documented
+    /// behaviour of [`Connection::transaction`]
+    fn transaction<F, R, E>(conn: &mut Conn, callback: F) -> Result<R, E>
```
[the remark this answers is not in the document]
weiznich: Basically to give third party `TransactionManager` implementations an starting point. They may not need to use custom behaviour to implement this function at all.

diesel/src/connection/transaction_manager.rs
```diff
@@ -237,55 +288,91 @@ where
...
+            Err(Error::CommitTransactionFailed {
+                commit_error: Box::new(error),
+                rollback_result: Box::new(Ok(())),
+            })
+        }
+        CommitErrorOutcome::ThrowAndMarkManagerAsBroken(error) => {
+            // The connection contains an unrecoverable broken transaction
+            // so mark the transaction state as broken and return the error
+            *AnsiTransactionManager::transaction_manager_status_mut(conn) =
+                TransactionManagerStatus::InError;
```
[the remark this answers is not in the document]
chusteven: Ah! Yeah that makes sense

> A connection pool will just not use a "broken" connection and remove it from the pool.

Neat! I'm poking around the `r2d2` module now and will see what I can discover there

diesel/src/connection/transaction_manager.rs
chusteven: I think I see what you mean. Yeah on one hand I want to say that most DBs are aware of isolation levels and especially the serializable one; but on the other hand, generic implementations are probably best left generic.

> Maybe I should just generalize that to something that would allow other connections to register "rollback" error variants in a similar manner

I'll have to study the code related to transactions to get a better idea of where this new behavior should live, but do agree with taking this approach

diesel/src/connection/transaction_manager.rs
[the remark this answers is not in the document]
weiznich: [This](https://github.com/diesel-rs/diesel/blob/7184f128a3c1435d12b32f413b14b6632e2f753c/diesel/src/r2d2.rs#L78) trait controls if a connection is considered broken or not. See [here](https://github.com/diesel-rs/diesel/blob/7184f128a3c1435d12b32f413b14b6632e2f753c/diesel/src/pg/connection/mod.rs#L174-L186) for the postgres specific impl.

diesel/src/connection/transaction_manager.rs
chusteven: The code from commit 553e5b575596a6d9d592bf244fcd149f7fce196d I think indeed makes the error more  generalizable but now I have one question that I feel bad for not asking earlier.

For the code below this line, I see that special handling is required to set the `transaction_state.previous_error_relevant_for_rollback` to `Some`; but otherwise, I'm a bit confused when you write "the result can be safely ignored". You're talking about ignoring the result from the new util method `process_commit_error `, right? If so, is it fair to say that we aren't ignoring the result so much as "overwriting" it to ensure that callers get this specific `kind` of error (as opposed to one of the `CommitErrorOutcome` one)  -- is that about right? If so, curious why we'd need to do this. But mostly just making sure I understand what's going on -- apologies if this is an ill-formed question

chusteven (commented):
Looks legit! The code in the `mysql` module helps to see too. Not sure what the approval process is. Happy to give this code an approve assuming tests pass (looks like merge conflicts need be resolved first to kick off CI tests) but probably would be more comfortable letting someone else approve and merge it as I'm so new to this codebase

diesel/src/connection/transaction_manager.rs
```diff
@@ -237,55 +285,92 @@ where
...
+                Self::rollback_transaction(conn).map_err(|e| {
+                    if is_serialization_error {
+                        if let Error::DatabaseError(
+                            crate::result::DatabaseErrorKind::SerializationFailure,
+                            _,
+                        ) = e
+                        {
+                            return e;
+                        }
+                    }
```
chusteven, line 322: More for my own understanding that anything, since we decided to introduce a generic error relevant for rollback I wonder two things
1. Should `is_serialization_error` be named something else like `is_error_relevant_for_rollback`?
2. Should the matching on line 315-321 just forward the error kind of the `Some(transaction_state. previous_error_relevant_for_rollback)` (I think it's a tuple so maybe can just return the `.0`th element of that `Some()`)

Asking because it seems to me like we should just forward along anything inside the `if` check on line 314 regardless of whether it matches a `crate::result::DatabaseErrorKind::SerializationFailure` or not

weiznich: > 1. Should is_serialization_error be named something else like is_error_relevant_for_rollback?

Yes that should be named different. That's a left over part of the old verison.

> 2. Should the matching on line 315-321 just forward the error kind of the Some(transaction_state. previous_error_relevant_for_rollback) (I think it's a tuple so maybe can just return the .0th element of that Some())

This error mapping here is likely not even required as `rollback_transaction` will handle that internally.

chusteven: Ah I see! In that case it seems like this could all be simplified by just calling:
```
Err(e) => {
    let transaction_state = Self::get_transaction_state(conn)?;
    Self::rollback_transaction(conn).map_err(|e| Error::RollbackError(Box::new(e)))?;
    Err(e)
}
```

Anyway thanks for fielding my questions!

weiznich: So I had a look at the corresponding code again. Seems like we cannot simplify that anymore, beside the name change. ` Self::rollback_transaction(conn).map_err(|e| Error::RollbackError(Box::new(e)))?;` does not work as we want to return a `Error::DatabaseError(crate::result::DatabaseErrorKind::SerializationFailure,  _)` for special cases, not a `RollbackError`. Both, the `is_serialization_error` and the actual error checking are there as a safeguard, so that we do only not wrap the inner error in those cases.

I've filed https://github.com/diesel-rs/diesel/pull/3128 for the variable name.

chusteven: Ah! You know what it was.. I was thrown off by the `SerializationFailure` as the hardcoded error type (albeit within a guard clause) because I recall originally you had asked whether this error type should be made more generic.

So I guess my question now is why not just return the error `e` without the guard clause since we already know it's relevant for the rollback (generic `DatabaseErrorKind ` error)?
```rust
Err(e) => {
    let transaction_state = Self::get_transaction_state(conn)?;
    let is_error_relevant_for_rollback = transaction_state
        .previous_error_relevant_for_rollback
        .is_some();

    Self::rollback_transaction(conn).map_err(|e| {
        if is_error_relevant_for_rollback {
            return e;
        }

        Error::RollbackError(Box::new(e))
    })?;
    Err(e)
}
```

Issue #3026, opened 2022-01-21
Update the CLI to use clap 3.0


weiznich:
With the resent [clap 3.0 release ](https://epage.github.io/blog/2021/12/clap3/) we should update our usage of clap in diesel CLI to the newest version as this hopefully fixes some of the existing issues/workarounds there. Any help doing that would be welcome. 

The code in question is mainly located here:
https://github.com/diesel-rs/diesel/blob/bdd008c97b53957329ef2cae0504739e76dd3860/diesel_cli/src/cli.rs

The update needs to study the clap migration guide and update removed/changed/deprecated function calls to their new equivalent. The corresponding code is covered by tests in the same crate, so verifying that everything works as expected after the migration should be quite straightforward.

There is an additional use of structopt/clap in one of the example crate.

https://github.com/diesel-rs/diesel/blob/bdd008c97b53957329ef2cae0504739e76dd3860/examples/postgres/advanced-blog-cli/src/cli.rs

Updating this is not a priority, but contributions there are welcome as well.

As a note for contributors: I will cross post this issue in the "This week in Rust" Call for Participation thread to hopefully get some contributions here.

czotomo:
I'd like to work on that, @weiznich 
There's an issue, though.
Clap 
[pins os_str_bytes crate to 6.0](https://github.com/clap-rs/clap/commit/b531964f46e9fc84dc8bec9205018f4931e97d83)
which
[dropped support for Rust 1.51](https://github.com/dylni/os_str_bytes/commit/6acb7e6e4233948ffe855af4079ad3355350d0db)

weiznich:
@czotomo Thanks for your interest in working on this :+1:

Its no problem to increase the msrv, as long as we mention that in our changelog (there is already a entry for that, just update the version there) and as long as we update the version used in our ci (.github/workflows/ci.yaml) + rust-toolchain file.

czotomo:
@weiznich 
Is 1.54 a reasonable MSRV or too recent? Turns out clap itself uses a 1.54-stabilised feature :(
```
error[E0658]: arbitrary expressions in key-value attributes are unstable
 --> clap-3.0.10/src/lib.rs:8:39
  |
8 | #![cfg_attr(feature = "derive", doc = include_str!("../README.md"))]
```

weiznich:
1.54 is fine as well. Those "checks" are in that we are aware of bumping the MSRV.

weiznich:
Closed by #3028 3028

PR #3028, opened 2022-01-23
[#3026] Bump clap to version 3

czotomo:
Clap changelog for reference:
https://github.com/clap-rs/clap/releases/tag/v3.0.0


czotomo:
Oh, Clippy issues. 
I'll work on them but I already see this is not going to be an easy journey. 
This is probably due to the rustc version dump.

pksunkara (approved):

PR #3031, opened 2022-01-26
Fix AsExpression doc link

Ten0:


diesel/src/serialize.rs
```diff
@@ -186,7 +186,8 @@ where
 /// database, you should use `i32::to_sql(x, out)` instead of writing to `out`
 /// yourself.)
 ///
-/// Any types which implement this trait should also [`#[derive(AsExpression)]`].
+/// Any types which implement this trait should also
+/// [`#[derive(AsExpression)]`](crate::AsExpression).
```
weiznich, line 190:
```suggestion
/// [`#[derive(AsExpression)]`](derive@crate::AsExpression).
```

Just to be sure that does not "break" again

weiznich:
I'm not sure why the CI is failing here, probably because my suggestion was broken? 🙈

Ten0:
No mine was broken too, I was hoping yours would fix it ^^'
Maybe that's a feature problem or something that prevents it from resolving this derive? Perhaps I should just leave the link as a standard link (the issue came from the `;` I believe).

weiznich:
I've pushed a fixed version. Seems like just the used path was not correct :shrug:

Ten0:
LGTM 🙂

PR #3035, opened 2022-02-01
implement jsonb concat operator

czotomo:


czotomo:
@weiznich am I supposed to push all the changes generated by compiletests locally even if Cargo.lock now contains bumped version of mysqlclient-sys? 🤔

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1167,3 +1167,90 @@ impl InetOrCidr for Inet {}
...
+    /// #         name VARCHAR NOT NULL,
+    /// #         address JSONB NOT NULL
+    /// #     )").unwrap();
+    /// #
+    /// let santas_address: serde_json::Value = serde_json::from_str(r#"{
+    ///     "street": "Article Circle Expressway 1",
+    ///     "city": "North Pole",
+    ///     "postcode": "99705",
+    ///     "state": "Alaska"
+    /// }"#).unwrap();
```
weiznich, line 1210: Could we use the `serde_json::json!` macro here? That should make it much easier to read the code + remove the `.unwrap()`

(Same for the other json strings)

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1167,3 +1167,90 @@ impl InetOrCidr for Inet {}
...
+where
+    T: Expression,
+    T::SqlType: JsonB,
+{
+}
+#[doc(hidden)]
+/// Marker trait used to implement `PgJsonbExpressionMethods` on the appropriate types.
+pub trait JsonB {}
+
+impl JsonB for Jsonb {}
```
weiznich, line 1256: We should allow nullable `jsonb` values here too. That's possible by the following change:

```suggestion
   T::SqlType: JsonbOrNullableJsonb,
{
}
#[doc(hidden)]
/// Marker trait used to implement `PgJsonbExpressionMethods` on the appropriate types.
pub trait JsonbOrNullableJsonb {}

impl JsonbOrNullableJsonb for Jsonb {}
impl JsonbOrNullableJsonb for Nullable<Jsonb> {}
```

weiznich (commented):
Looks good from my side with those small suggestions fixed :+1:

About the compile tests: Yes those need an update too, as some error messages change due to the fact that there are now additional trait impls that rustc could suggest. You can do that by running `TRYBUILD=overwrite cargo test` in the compile tests directory and commit the changes afterwards. I will add that to the general instructions, as this may happen for other operators as well.

weiznich (approved):

weiznich:
Thanks for your contribution ❤️

PR #3036, opened 2022-02-03
implement jsonb key existence check expression

czotomo:


czotomo:
@weiznich ready for review, I believe

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1241,6 +1241,66 @@ pub trait PgJsonbExpressionMethods: Expression + Sized {
     {
         Grouped(ConcatJsonb::new(self, other.as_expression()))
     }
+    /// Creates a PostgreSQL `?` expression.
+    ///
+    /// This operator checks if the right hand side string exists as a top-level key within the JSON
+    ///
+    /// # Example
```
Mingun, line 1248: I think it also will be worth to copy examples from https://www.postgresql.org/docs/devel/functions-json.html

pksunkara (approved):

PR #3041, opened 2022-02-07
Implement jsonb any-key and all-keys check expressions

mrbuzz:
Implemented JSONB  `?|` and `?&` operators.


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1302,8 +1302,129 @@ pub trait PgJsonbExpressionMethods: Expression + Sized {
...
+    /// #     Ok(())
+    /// # }
+    /// ``````
+    fn has_any_key<T>(self, other: T) -> dsl::HasAnyKeyJsonb<Self, T>
+    where
+        T: AsExpression<Array<VarChar>>,
+    {
+        Grouped(HasAnyKeyJsonb::new(self, other.as_expression()))
+    }
+    /// Creates a PostgreSQL `?&` expression.
```
pksunkara, line 1365: Empty line before this please

pksunkara (changes requested):
Looks good except for minor nitpick. Can merge immediately after it's addressed 

Also ci is failing.

diesel/src/pg/expression/expression_methods.rs
mrbuzz: Thanks! There were just a couple of minor format mistakes. They should be fixed now

pksunkara (approved):

PR #3048, opened 2022-02-17
Fix clap 3.1 deprecation warnings in Diesel CLI

Ten0:
Those would prevent from compiling at all due to `#![deny(warnings)]`.
Also tests broke due to an error message update.

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -20,8 +20,8 @@ doc = false
 
 [dependencies]
 chrono = { version = "0.4.19", default-features = false, features = ["clock", "std"] }
-clap = "3.0"
-clap_complete = "3.0"
+clap = "3.1"
+clap_complete = "3.1"
 dotenv = "0.15"
 heck = "0.3.1"
 serde = { version = "1.0.0", features = ["derive"] }
--- a/diesel_cli/src/cli.rs
+++ b/diesel_cli/src/cli.rs
@@ -1,12 +1,12 @@
 use crate::validators::num::*;
-use clap::{App, AppSettings, Arg};
+use clap::{Arg, Command};
 use clap_complete::Shell;
 
 fn str_as_char(str: &str) -> char {
     str.chars().next().unwrap()
 }
 
-pub fn build_cli() -> App<'static> {
+pub fn build_cli() -> Command<'static> {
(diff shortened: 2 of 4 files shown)


Ten0:
The CI fails seem unrelated

pksunkara:
> Testing bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "Commands out of sync; you can't run this command now")', benches/diesel_benches.rs:378:14

That doesn't look related to sqlite though. 🤔

Ten0:
> > Testing bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
> > thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "Commands out of sync; you can't run this command now")', benches/diesel_benches.rs:378:14
> 
> That doesn't look related to sqlite though. thinking

Unless I'm mistaken this only appears in the MacOS/Mysql tests, not sqlite.
It's also a database error, not a CLI error.

weiznich:
Merged as both CI errors are not related to this PR.

The sqlite windows failure seems to be related to the update to the new windows-22 images.

The mysql file could be related to an updated mysql version as well? 

Might be worth to track them down separately.

Ten0:
> The mysql file could be related to an updated mysql version as well?
> 
> Might be worth to track them down separately.

Yes I suspect the MySQL one highlights a legitimate usage issue - from googling the error message, which just wasn't highlighted in previous versions.

weiznich:
I had a short look at the corresponding CI logs. The [last passing run of this CI job](https://github.com/diesel-rs/diesel/actions/runs/1834472425) did use mariadb 10.6.4, while these runs here did use mariadb 10.6.7. 

Changelogs for:

* [10.6.5](https://mariadb.com/kb/en/mariadb-1065-changelog/)
* [10.6.6](https://mariadb.com/kb/en/mariadb-1066-changelog/)
* [10.6.7](https://mariadb.com/kb/en/mariadb-1067-changelog/)

I'm quite unsure what may cause this as the corresponding benchmark executes quite basic queries. The only notable thing there may be a large amount of bind parameters for a `IN` expression.

PR #3049, opened 2022-02-18
add ssl_ca to accepted connection url query params

thomasmost:
While #2491 allowed setting ssl_mode in the DATABASE_URL, this is insufficient to successfully initiate an SSL connection in some instances. For example, while running a dockerized service on Debian, the Certificate Authority roots need to be manually set. [See PlanetScale's docs on the subject](https://docs.planetscale.com/concepts/secure-connections#why-do-i-need-to-care-about-a-ca-when-connecting-to-planetscale?)

This adds optional support for `ssl_ca`, following the existing architecture for setting ssl_mode and unix_port in the connection string

Note that I've confirmed this allows for a successful connection to PlanetScale from a debian container


thomasmost:
Tests are failing in `master` right now but once that's fixed I'll merge and clean up any remaining failures

weiznich (approved):
Looks good for me as soon as the rustfmt + clippy CI run passes.

thomasmost:
@weiznich fixed!

PR #3052, opened 2022-02-21
Fix/ci 2022

weiznich:
Fixes two out of three failing CI runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,16 +128,15 @@ jobs:
         if: runner.os == 'macOS' && matrix.backend == 'mysql'
         run: |
           brew update
-          brew install mariadb
-          mysql_install_db
-          mysql.server start
+          brew install mariadb@10.5
+          /usr/local/opt/mariadb@10.5/bin/mysql_install_db
+          /usr/local/opt/mariadb@10.5/bin/mysql.server start
           sleep 3
-          mysql -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'runner'@'localhost';" -urunner
-          macos_mysql_version="$(ls /usr/local/Cellar/mariadb)"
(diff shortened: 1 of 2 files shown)


weiznich:
I'm still totally unsure what's wrong with the mysql benchmark on macos. [The mysql documentation](https://dev.mysql.com/doc/refman/8.0/en/commands-out-of-sync.html#:%7E:text=If%20you%20get%20Commands%20out,you%20have%20called%20mysql_free_result()%20) indicates that this is an error in our client side code. This raises multiple questions:

* Why does it work for all other CI targets?
* Why does it work for all other queries on that CI target?

Reducing the number of entries in the corresponding table seems to "fix" that error, but that's not a real solution.

Ten0:
> Why does it work for all other CI targets?

Wasn't there something like, on MacOS we are using the MariaDB libs while on other targets we are using the MySQL libs, because they sopped working at some point?
https://github.com/diesel-rs/diesel/blob/8825808ebac4999514f731f3220903fff60bb2aa/.github/workflows/ci.yml#L131

Perhaps the error is visible more or less aggressively depending on the backend and the amount of memory allocated (which could explain why the number of entries has an impact).
I guess the alternative would be a bug in the MariaDB libs, but that seems more unlikely than a mistake on our side?

weiznich:
Yes the macos ci target is the only one using mariadb server and client side, so that could be the reason for this issue. Nevertheless the question is why does this happen after an update between mariadb patch versions. (10.6.4 -> 10.6.7)

Ten0:
> Nevertheless the question is why does this happen after an update between mariadb patch versions. (10.6.4 -> 10.6.7)

I only see two possibilities:
 - There is an actual mistake in the way we handle these, and they just became more likely to error than they were before (I don't think that would be considered a breaking change)
 - They bugged it

Ten0 (approved):

PR #3053, opened 2022-02-21
Fix alias visibility

Ten0:
Alias declared at the module level would not be visible out of the module due to the lack of "pub".
This fixes it by allowing to specify a visibility modifier in the macro.

Aliases declared with the table-like module-level syntax are always `pub` (like tables).

### DIFF
--- a/diesel/src/query_source/aliasing/macros.rs
+++ b/diesel/src/query_source/aliasing/macros.rs
@@ -37,7 +37,7 @@
 /// diesel::alias!(schema::users as users_alias: UsersAlias);
 /// // or
 /// diesel::alias!{
-///     const USERS_ALIAS_2: Alias<UsersAlias2> = schema::users as users_alias_2;
+///     pub const USERS_ALIAS_2: Alias<UsersAlias2> = schema::users as users_alias_2;
 /// }
 ///
 /// fn some_function_that_returns_a_query_fragment(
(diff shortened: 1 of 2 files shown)


diesel/src/query_source/aliasing/macros.rs
```diff
@@ -79,23 +79,23 @@ macro_rules! alias {
...
         $crate::alias! {
             $(
-                const $alias_name: Alias<$alias_ty> = $($table)::+ as $alias_name;
+                pub const $alias_name: Alias<$alias_ty> = $($table)::+ as $alias_name;
             )*
         }
     };
-    ($(const $const_name: ident: Alias<$alias_ty: ident> = $($table: ident)::+ as $alias_sql_name: ident);* $(;)?) => {
-        $crate::alias!(NoConst $($($table)::+ as $alias_sql_name: $alias_ty,)*);
+    ($($vis: vis const $const_name: ident: Alias<$alias_ty: ident> = $($table: ident)::+ as $alias_sql_name: ident);* $(;)?) => {
```
weiznich, line 86: Just to make that sure: We have a 
```
diesel::alias!{
     const USERS_ALIAS_2: Alias<UsersAlias2> = schema::users as users_alias_2;
}
```
like case somewhere to verify that we could create private variants too?

Ten0: almost: we've got the in-function version of the macro:
```rust
($($($table: ident)::+ as $alias: ident),* $(,)?) => {{
        $crate::alias!(NoConst $($($table)::+ as $alias: $alias,)*);
        ($($crate::query_source::Alias::<$alias>::default()),*)
}};
```
that does not pass the visibility modifier arg ([which is always optional](https://doc.rust-lang.org/reference/macros-by-example.html#metavariables)), so we know the macro is able to parse that properly.

Ten0: I'm going to add that to the visibility test anyway, that's quickly done

weiznich (approved):
Looks good from my side, but I would like to fix the CI issues fist before merging new stuff.

PR #3056, opened 2022-02-23
Generate Postgres Arrays with Nullable elements by default

Justice4Joffrey:
It's not possible to enforce non-nullability of Postgres array elements.

The previous behaviour of mapping to `Array<ST>` works fine in practice because Diesel would never _itself_ insert arrays with `NULL` elements. This was at the risk of deserialization errors though, should `NULL` values ever be made present (e.g. by some other process/manually).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -248,7 +248,7 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Implementations of custom SQLite SQL functions now check for panics
 
-* `infer_schema!` now generates `Array<Nullable<ST>>` rather than `Array<ST>` for Postgres Array types. Existence of 
+* `diesel print-schema` now generates `Array<Nullable<ST>>` rather than `Array<ST>` for Postgres Array types. Existence of 
   `NULL` values in database arrays would previously result in deserialization errors. Non-nullable arrays are now opt
   in (by schema patching).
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -248,6 +248,10 @@ for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/
 
 * Implementations of custom SQLite SQL functions now check for panics
 
+* `infer_schema!` now generates `Array<Nullable<ST>>` rather than `Array<ST>` for Postgres Array types. Existence of
```
weiznich, line 251:
```suggestion
* `diesel print-schema` now generates `Array<Nullable<ST>>` rather than `Array<ST>` for Postgres Array types. Existence of 
```

weiznich (approved):
Thanks for this PR :+1:

Looks good with the small suggested adjustment for the Changelog text. I will merge the PR after I figured out whats wrong with the failing CI runs. (Those are not caused by this PR, but by environment updates on the corresponding runners.)

chrislearn:
This is a break change and it make code verbose.

chrislearn:
Or is it possible to add an option to allow generate array without Nullable  wrapper?

weiznich:
@chrislearn We are aware that this is a breaking change. That's why it is included in a major version update. We do not have any planes to make this behaviour configurable as this is essentially a correctness fix. Checkout the [relevant part](https://github.com/diesel-rs/diesel/blob/master/guide_drafts/migration_guide.md#changed-nullability-of-array-elements) of the migration guide for possible ways to "workaround" your problem.

PR #3066, opened 2022-03-03
Add mutable downcast for BoxableConnection

Sytten:
This is necessary when doing code based migrations where you need to query the database.


weiznich (approved):
:+1: Thanks for this PR.

I would like to add a small test for this somewhere, so that we don't break it again. That test does not need to  be super complex. Just something that creates a `BoxableConnection` and downcasts it to execute a really simple query.

Sytten:
On it I will push something today

Issue #1681, opened 2018-05-04
Support Updating Individual Array Elements


philip-alldredge:
## Problem Description
As far as I am aware, Diesel does not provide a mechanism for referencing individual array elements inside of the SQL queries.

### What are you trying to accomplish?
Updating a single element in an array column when working with a Postgres database. I'd like to generate a query like:

```
UPDATE table_with_array
SET my_array[2] = 5.0;
```
Using syntax like:
`update(table_with_array.set(my_array.index(1).eq(5.0)).execute(db_connection);`

sgrif:
This is a reasonable feature request, but isn't likely to get added in the near future. I'd recommend using [`sql_query`](http://docs.diesel.rs/diesel/fn.sql_query.html) for this.

philip-alldredge:
Okay. Thanks! It's only a minor inconvenience. Everything else works great.

sgrif:
Glad to hear it. :smile:

sgrif:
I actually don't think this would be super hard to implement if someone were interested in tackling it (being able to select a single element of an array would be a good first step. I think this might make sense as an issue for @galuhsahid or @yamishi13 in a few weeks.

weiznich:
Just to write it somewhere:
I think this should be done in a similar manner like the implementation of  numeric operations in diesel for various types. This means in detail that instead of providing a index method we should use the `Index` trait from the std-lib and then return a custom wrapper for codegen from there. (Just like for Add for example)

Justice4Joffrey:
I've just had a look at how this might be implemented. 

The docs suggest arrays containing `NULL` aren't supported. Afaict that's a bad fit for Postgres arrays which can always contain `NULL` elements.

e.g. demonstrated by this somewhat surprising behaviour -

```
db=# CREATE TABLE posts (id SERIAL PRIMARY KEY, tags TEXT[] NOT NULL);
db=# insert into posts (tags) values (array['cool', 'great']);
db=# update posts set tags[5] = 'incredible';
db=# select * from posts;
+----+-----------------------------------+
| id |               tags                |
+----+-----------------------------------+
|  1 | {cool,great,NULL,NULL,incredible} |
+----+-----------------------------------+
```

Any Array which ends up containing `NULL`s will break deserialization with `DeserializationError(UnexpectedNullError)`. Therefore, offering the `index()` method is only going to make this footgun a lot easier to trigger.

[...]

PR #3075, opened 2022-03-08
selectable/filterable array.index(n)

Justice4Joffrey:
@weiznich just want to get your views on this.

A couple of questions I had -

- should I create an `extract_operator!` macro or just define `ArrayIndex` as is in `pg/operators.rs`
- it would be nice to convey 1-indexing to the user, some other way than a doc comment. In Postgres, `foo[-1]` is totally legal though (and accesses a negatively-indexed value, rather than starting from the back e.g. Python).
- The index type should arguably be Nullable<Integer>, or accept both. Would be good to know your thoughts on that.

#1681

### DIFF
--- a/diesel/src/pg/expression/operators.rs
+++ b/diesel/src/pg/expression/operators.rs
@@ -32,13 +32,16 @@ infix_operator!(IsContainedByJsonb, " <@ ", backend: Pg);
 #[derive(Debug, Clone, Copy, QueryId, DieselNumericOps, ValidGrouping)]
 #[doc(hidden)]
 pub struct ArrayIndex<L, R> {
-    pub(crate) left: L,
-    pub(crate) right: R,
+    pub(crate) array_expr: L,
+    pub(crate) index_expr: R,
 }
 
 impl<L, R> ArrayIndex<L, R> {
-    pub fn new(left: L, right: R) -> Self {
-        Self { left, right }
+    pub fn new(array_expr: L, index_expr: R) -> Self {
+        Self {
+            array_expr,
+            index_expr,
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


diesel/src/pg/expression/operators.rs
```diff
@@ -26,3 +28,44 @@ infix_operator!(HasAnyKeyJsonb, " ?| ", backend: Pg);
 infix_operator!(HasAllKeysJsonb, " ?& ", backend: Pg);
 infix_operator!(ContainsJsonb, " @> ", backend: Pg);
 infix_operator!(IsContainedByJsonb, " <@ ", backend: Pg);
+
+#[derive(Debug, Clone, Copy, QueryId, DieselNumericOps, ValidGrouping)]
+#[doc(hidden)]
+pub struct ArrayIndex<L, R> {
+    pub(crate) left: L,
+    pub(crate) right: R,
```
weiznich, line 36: I would prefer other names than `left` and `right` here.

```suggestion
    pub(crate) array_expr: L,
    pub(crate) index_expr: R,
```

weiznich (commented):

This already looks good, beside that minor naming suggestion. Also it does likly require a rebase, due to recent merges to the master branch.

To answer your questions:

> * should I create an extract_operator! macro or just define ArrayIndex as is in pg/operators.rs

It's fine to have a adhoc definition for this. The existing macros are mostly for cases where third party crates may want to define an similar operator as well. That's not the case here.

> *  it would be nice to convey 1-indexing to the user, some other way than a doc comment. In Postgres, foo[-1] is totally legal though (and accesses a negatively-indexed value, rather than starting from the back e.g. Python).

I would prefer to stay as close as possible here to the postgres semantics. So no index conversion.

 > *  The index type should arguably be Nullable, or accept both. Would be good to know your thoughts on that.

I do not think `null` as index value has a valid usecase. As of this I would stay with the current version.

weiznich:
Thanks :+1:

PR #3074, opened 2022-03-07
Remove time dependency from chrono

AldaronLau:
I know this is changed in the master branch, but it would be nice to have on 1.4.x as well, considering that 2.0.0 isn't out yet.

PR #3077, opened 2022-03-10
Backporting sqlite read_blob fix to version 1.4.X to unblock the build with address sanitizer

kgrech:
Original fix in master: [PR](https://github.com/diesel-rs/diesel/pull/2941)


weiznich:
Strictly speaking we do only backport critical bugfixes to the 1.4.x release. That change is something I would not classify as critical bug fix as it requires manually building rust-stdlib. Similar to https://github.com/diesel-rs/diesel/pull/3074 I would be fine with including this into a release which would happen nevertheless. That means I will leave this PR open till we either issue another 1.4.x release or we release a first 2.0.0 version.

That written: I prerequisite for including this into a release would be that the CI passes.

weiznich:
Closed as 2.0.0 is finally released. It does contain this fix.

PR #3090, opened 2022-03-18
Revert PR #2944 (FROM ONLY <table> clause)

Ploppz:
Reverts #2944.
Reason: We need a more general `ONLY` implementation that can be applied to any table, so that we can for example support something along the lines of `.inner_join(my_table.only())` (SQL: `INNER JOIN ONLY my_table`)`

When I ran `./bin/test`, a lot of tests passed (maybe all or most?) before I got this error 

```
+ cd diesel_bench
+ cargo bench --features sqlite --list
error: Found argument '--list' which wasn't expected, or isn't valid in this context

USAGE:
    cargo bench --features <FEATURES>...

For more information try --help
```

I don't know if it got to compile error testing yet, I would expect changes in stderr to have to be committed there because PR #2944 did require that too (due to extra suggestions by the compiler that appeared).


weiznich (approved):

weiznich:
Looks good, thanks for filling this pr :+1:. I will fix the merge conflicts tomorrow and merge it afterwards.

PR #3092, opened 2022-03-19
Missing json operators

### LINKED ISSUE #3032: Add support for currently unsupported JSON/JSONB operators
Diesel currently supports the postgres json and jsonb types. We do not provide built-in support for [various operators](https://www.postgresql.org/docs/current/functions-json.html) available for these types. This is a tracking issue for adding support for these operators. 

The general strategy for adding support for new operators is as following:

1. Define the operator via [`infix_operator!()`](https://docs.diesel.rs/master/diesel/macro.infix_operator.html). The contains operator  would be defined as following: `infix_operator!(KeyExists, " ? ", backend: Pg);` These operators can be defined [here](https://github.com/diesel-rs/diesel/blob/dd5f41dc93c574077646185393cd064c7103000b/diesel/src/pg/expression/operators.rs). If there is already an existing definition, this step could be skipped.
2. Define a new helper type [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/expression/helper_types.rs#L30). Helper types are a tool for diesel users to simplify type definitions. Again if there is already a matching type definition this could be skipped. 
3. Add a new `ExpressionMethods` extension trait for the corresponding types. This only needs to happen once for `jsonb`   and once for `jsonb` + `json`.  See the [`PgRangeExpressionMethods` trait](https://github.com/diesel-rs/diesel/blob/dd5f41dc93c574077646185393cd064c7103000b/diesel/src/pg/expression/expression_methods.rs#L621-L690) here as example.
4. Add a new method to the corresponding `ExpressionMethods` trait. See [here](https://github.com/diesel-rs/diesel/blob/dd5f41dc93c574077646185393cd064c7103000b/diesel/src/pg/expression/expression_methods.rs#L143-L200) for an example for a different operator/type. This definition should:
    * Use the previously defined helper type
    * Contain a documentation with an example
5. Submit a PR with the change

Operator list:

- [ ] `json/jsonb -> integer` (indexing by index, returns `json/jsonb`) **
- [ ] `json/jsonb -> text` (indexing by key name, returns `json/jsonb`) **
- [ ] `json/jsonb ->> integer` (indexing by index, returns field as `text`) **
- [ ] `json/jsonb ->> text` (indexing by key name, returns field as `text`) **
- [ ] `json/jsonb #> text[]` (indexing by path, returns field as `json/jsonb`) *
- [ ] `json/jsonb #>> text[]` (indexing by path, returns field as `text`) *
- [ ] `jsonb @> jsonb` (checks if the second value is contained by the first value, returns a `boolean`) *
- [ ] `jsonb <@ jsonb` (checks if the first value is contained by the second value, returns a `boolean`) *
- [ ] `jsonb ? text` (checks if a string exists as top level key in the given json, returns a `boolean`) *
- [ ] `jsonb ?| text[]` (checks if any of the strings exists as top level key in the given json, returns a `boolean`) *
- [ ] `jsonb ?& text[]` (checks if all of the strings exist as top level key in the given json, returns a `boolean`) *
- [ ] `jsonb || jsonb` (concats two json values, returns `jsonb`) *
- [ ] `jsonb - text` (remove the key given as string from the json, returns `jsonb`) **
- [ ] `jsonb - text[]` (removes all keys given as string from the json, returns `jsonb`) **
- [ ] `jsonb - integer` (removes the field with the given index from the json, returns `jsonb`) **
- [ ] `jsonb #- text[]` (removes the given path from the json, returns `jsonb`) *
- [ ] `jsonb @? jsonpath` (Does JSON path return any item for the specified JSON value?, returns a `boolean`) ***
- [ ] `jsonb @@ jsonpath (Returns the result of a JSON path predicate check for the specified JSON value. Only the first item of the result is taken into account. If the result is not Boolean, then NULL is returned., returns a `boolean`) ***

Items marked with one `*` can follow the instructions directly, items marked with `**` require additional work to unify the different input types behind a common operator/function call (See [here](https://github.com/diesel-rs/diesel/blob/dd5f41dc93c574077646185393cd064c7103000b/diesel/src/pg/expression/expression_methods.rs#L551-L561) for a general example on how to abstract over expressions of different types), items marked with `***` likely require additional design work.

Please add a comment to this issue if you plan to work on a specific operator.

If there is anything unclear about how to add a support for a specific operator just ask and we will try to answer your questions.

weiznich:
This PR adds the missing json/jsonb operators as listed in #3032. It skips the last both operators as they use jsonpath as input type and it's not clear how to represent that on rusts side.

Fixes #3032 
Closes #3044

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -1782,15 +1782,248 @@ pub trait PgAnyJsonExpressionMethods: Expression + Sized {
     fn retrieve_as_object<T>(
         self,
         other: T,
-    ) -> dsl::RetrieveAsObjectJsonb<Self, T::Expression, <T::Expression as Expression>::SqlType>
+    ) -> dsl::RetrieveAsObjectJson<Self, T::Expression, <T::Expression as Expression>::SqlType>
     where
         T: JsonIndex,
         <T::Expression as Expression>::SqlType: SqlType,
     {
-        Grouped(RetrieveAsObjectJsonb::new(
+        Grouped(RetrieveAsObjectJson::new(
             self,
             other.as_json_index_expression(),
         ))
+    }
+
+    /// Creates a PostgreSQL `->>` expression.
+    ///
+    /// This operator extracts the value associated with the given key, that is provided on the
+    /// Right Hand Side of the operator.
+    ///
+    /// Extracts n'th element of JSON array (array elements are indexed from zero, but negative integers count from the end).
+    /// Extracts JSON object field as Text with the given key.
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #    contacts {
+    /// #        id -> Integer,
+    /// #        name -> VarChar,
+    /// #        address -> Jsonb,
+    /// #    }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    ///
+    /// # #[cfg(feature = "serde_json")]
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::contacts::dsl::*;
+    /// #     let conn = &mut establish_connection();
+    /// #     diesel::sql_query("DROP TABLE IF EXISTS contacts").execute(conn).unwrap();
+    /// #     diesel::sql_query("CREATE TABLE contacts (
+    /// #         id SERIAL PRIMARY KEY,
+    /// #         name VARCHAR NOT NULL,
+    /// #         address JSONB NOT NULL
+    /// #     )").execute(conn)
+    /// #        .unwrap();
+    /// #
+    /// let santas_address: serde_json::Value = serde_json::json!({
+    ///     "street": "Article Circle Expressway 1",
+    ///     "city": "North Pole",
+    ///     "postcode": "99705",
+    ///     "state": "Alaska"
+    /// });
+    /// diesel::insert_into(contacts)
+    ///     .values((name.eq("Claus"), address.eq(&santas_address)))
+    ///     .execute(conn)?;
+    ///
+    /// let santas_postcode = contacts.select(address.retrieve_as_text("postcode")).get_result::<String>(conn)?;
+    /// assert_eq!(santas_postcode, "99705");
+    ///
+    ///
+    /// let robert_downey_jr_addresses: serde_json::Value = serde_json::json!([
+    ///     {
+    ///         "street": "Somewhere In La 251",
+    ///         "city": "Los Angeles",
+    ///         "postcode": "12231223",
+    ///         "state": "California"
+    ///     },
+    ///     {
+    ///         "street": "Somewhere In Ny 251",
+    ///         "city": "New York",
+    ///         "postcode": "3213212",
+    ///         "state": "New York"
+    ///     }
+    /// ]);
+    ///
+    /// diesel::insert_into(contacts)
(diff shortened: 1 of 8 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1581,8 +1688,358 @@ where
...
+    ///     "street": "Article Circle Expressway 1",
+    ///     "city": "North Pole",
+    ///     "postcode": "99705",
+    ///     "state": "Alaska"
+    /// });
+    /// diesel::insert_into(contacts)
+    ///     .values((name.eq("Claus"), address.eq(&santas_address)))
+    ///     .execute(conn)?;
+    ///
+    /// let santas_postcode = contacts.select(address.retrieve_by_path_as_object(vec!["postcode"])).get_result::<serde_json::Value>(conn)?;
```
pksunkara, line 1944: Can you use RDJ example data here and use `vec!["1", "postcode"]` here? Or is that not supported by postgres?

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1581,8 +1688,358 @@ where
...
+    ///     "street": "Article Circle Expressway 1",
+    ///     "city": "North Pole",
+    ///     "postcode": "99705",
+    ///     "state": "Alaska"
+    /// });
+    /// diesel::insert_into(contacts)
+    ///     .values((name.eq("Claus"), address.eq(&santas_address)))
+    ///     .execute(conn)?;
+    ///
+    /// let santas_postcode = contacts.select(address.retrieve_by_path_as_text(vec!["postcode"])).get_result::<String>(conn)?;
```
pksunkara, line 2009: Same question here as the above comment.

diesel/src/pg/expression/helper_types.rs
```diff
@@ -105,3 +105,23 @@ pub type IsContainedByJsonb<Lhs, Rhs> =
 
 /// The return type of `lhs.index(rhs)`
 pub type ArrayIndex<Lhs, Rhs> = super::operators::ArrayIndex<Lhs, AsExprOf<Rhs, Integer>>;
+
+/// The return type of `lhs.remove(rhs)`
+pub type RemoveFromJsonb<Lhs, Rhs, St> =
+    Grouped<super::operators::RemoveFromJsonb<Lhs, AsExprOf<Rhs, St>>>;
```
pksunkara, line 111:
```suggestion
    Grouped<super::operators::RemoveFromJsonb<Lhs, AsExprOf<Rhs, ST>>>;
```

also in the template.

diesel/src/pg/expression/helper_types.rs
```diff
@@ -105,3 +105,23 @@ pub type IsContainedByJsonb<Lhs, Rhs> =
...
+pub type RetrieveAsObjectJson<Lhs, Rhs, ST> =
+    Grouped<super::operators::RetrieveAsObjectJson<Lhs, AsExprOf<Rhs, ST>>>;
+
+/// The return type of `lhs.retrieve_as_text(rhs)`
+pub type RetrieveAsTextJson<Lhs, Rhs, ST> =
+    Grouped<super::operators::RetrieveAsTextJson<Lhs, AsExprOf<Rhs, ST>>>;
+
+/// The return type of `lhs.retrieve_by_path_as_object(rhs)`
+pub type RetrieveByPathAsObjectJson<Lhs, Rhs> =
+    Grouped<super::operators::RetrieveByPathAsObjectJson<Lhs, AsExprOf<Rhs, Array<Text>>>>;
```
pksunkara, line 123: I am wondering why this is not `Array<VarChar>`? We seem to use it like that in above JSONB operators.

pksunkara (commented):
`jsonb #- text[]` seem to be missing.

PR #3093, opened 2022-03-22
Enable expressions on the left-hand side of update.set()

Justice4Joffrey:
It looks like there are a bunch of strategies one can use to implement this behaviour

```
diesel::update(my_table::table).set(array_col.index(1).eq(true))
```

My line of reasoning is -

- There should be a trait between `Column` and `Assign` which `Column` implements which reduces to the current implementation
- This allows the definition of `AssignmentTarget` for arbitrary types which will be picked up by the `AsChangeset` `impl`.

I hit a wall updating the definition for `AsChangeset` as I have an unbounded type parameter (changset.rs:44). I was exploring a few workarounds but none that made me particularly happy.

Would be great to know your thoughts on this direction and whether it's possible to complete this implementation without something heavy like introducing a new param to `AsChangeset`.


weiznich:

> Would be great to know your thoughts on this direction 

Abstracting away the column bound seems sensible for me. 

> whether it's possible to complete this implementation without something heavy like introducing a new param to AsChangeset.

I cannot really comment on that without sitting down and basically implementing the thing on my own. What I can do is to is outlining a few constraints on potential solutions and give a few ideas on how to workaround that unbound parameter problem

So the basic constraint first: We do not want to change the public trait/function signature there, so anything needs to fit into the current `AsChangeset` impl. This means adding a generic parameter to `AsChangeset` is out of scope.

So how could it work anyway? Basically the answer there is that we don't need the `AssignmentTarget::walk_ast` function at all. This removes the need to have a generic parameter for `DB` there. This raises the question how would we get access to the customized `walk_ast` implementation? We can solve that by having a `AssignmentTarget::get_target()` function that returns a type with a matching `QueryFragment` impls for each implementation. This function would be called inside of the `AsChangeset` impl for `Eq<Left, Right>` so that we store the custom query node in `Assign`. For colums this would return a new type wrapper that implements `QueryFragment` like the current implementation. Put together that would look somewhat like this:

```rust
trait AssignmentTarget {
    type Table;
    type QueryAstNode;

    fn get_target(self) -> Self::QueryAstNode;
}

impl AssignmentTarget for C where C: Column {
    type Table = C::Table;
    type QueryAstNode = ColumnWrapperForUpdate<C>;

    fn get_target(self) -> Self::QueryAstNode {
        ColumnWrapperForUpdate(self)
    }
}

struct ColumnWrapperForUpdate<C>(C);

impl<DB, C> QueryFragment<DB> for ColumnWrapperForUpdate<C> where DB: Backend + DieselReserveSpecialization, C: Column {
    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
        out.push_identifier(C::NAME)?;
    }
}

```

Justice4Joffrey:
Ok - thanks for the feedback. Makes sense.

I ran in to a couple of other issues implementing this. Most frustrating is being unable to use the fully qualified column names that `QueryFragment` generates.

Invalid:
`UPDATE "posts" SET "posts"."tags"[$1] = $2 WHERE ("posts"."id" = $3) RETURNING "posts"."id", "posts"."tags"`
Valid:
`UPDATE "posts" SET "tags"[$1] = $2 WHERE ("posts"."id" = $3) RETURNING "posts"."id", "posts"."tags"`

I'm assuming this will need to leave `QueryFragment` alone so will most likely end up with some duplication, although I'm going to have another think about it when I get the time.

Justice4Joffrey:
Thanks - that was helpful.

This looks to be working barring some test output clearup, but I had a few points to raise -

- I wanted to blanket implement `AssignmentTarget` for all `T where T: Column` couldn't as `AssignableExpression` has an overlapping impl (missing speciailzation), so changed the table macro instead for all Column definitions.
- I wasn't sure about the best strategy for exposing/hiding internal types, such as `ColumnWrapperForUpdate`. Afaict, not making this `pub` makes Rust unhappy as it's leaked elsewhere.
- I don't love the name `UncorrelatedColumn`, but that seemed the most reasonable given the PG docs https://www.postgresql.org/docs/14/sql-expressions.html

diesel/src/query_builder/uncorrelated_column.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Provide a wrapper type which will render a `Column` without its
+/// 'correlation' (i.e. table name).
+pub(crate) trait Uncorrelated {
+    /// The `Column` type to wrap.
+    type Column: Column;
+
+    /// `Column` wrapped to signify it shouldn't be rendered with its
+    /// correlation.
+    fn uncorrelated(self) -> UncorrelatedColumn<Self::Column>;
+}
```
weiznich, line 14: As far as I can tell, we don't need this trait at all. We can just all `AssignableExpression::into_assignable_expression` for the inner type of `ArrayIndex`.

diesel/src/query_builder/mod.rs
```diff
@@ -55,7 +56,11 @@ pub use self::sql_query::{BoxedSqlQuery, SqlQuery};
 pub use self::upsert::on_conflict_target_decorations::DecoratableTarget;
 
 #[doc(inline)]
-pub use self::update_statement::changeset::AsChangeset;
+pub use self::uncorrelated_column::UncorrelatedColumn;
+#[doc(inline)]
+pub use self::update_statement::changeset::{
+    AsChangeset, AssignableExpression, AssignmentTarget, ColumnWrapperForUpdate,
+};
```
weiznich, line 63: None of those types should be part of our public API, as they are only internal helper types. 
For those types that are required for implementing the `table!` macro: Reexport them via `diesel::internal::table_macro` as `#[doc(hidden)]` item.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -68,20 +69,95 @@ where
...
+
+impl<Expr> AssignmentTarget for Expr
+where
+    Expr: AssignableExpression,
+{
+    type Table = <<Expr as AssignableExpression>::Column as Column>::Table;
+    type QueryAstNode = ExpressionWrapperForUpdate<Expr::Expression>;
+
+    fn into_target(self) -> Self::QueryAstNode {
+        ExpressionWrapperForUpdate(self.into_assignable_expression())
```
weiznich, line 161: Is there a specific reason here why we implement this for generic expressions, instead of just for `ArrayIndex<C, Expr>`? Fixing this would likely remove the conflicting impl error with the generic column impl, which in turn would simplify the whole implementation as well.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -68,20 +69,95 @@ where
...
+    type Column: Column;
+    /// The expression to assign to. It's normally required to transform
+    /// `Column`s using [crate::query_builder::UncorrelatedColumn]. This
+    /// removes the table identifier.
+    type Expression;
+
+    /// Return the expression in a form which can be used in an update
+    /// statement.
+    fn into_assignable_expression(self) -> Self::Expression;
+}
```
weiznich, line 138: It's not obvious for me why this needs to be a distinct trait than `AssignmentTarget`. I believe the whole transformation should work with a single trait layer.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -68,20 +69,95 @@ where
...
+pub trait AssignmentTarget {
+    /// Table the assignment is to
+    type Table: Table;
+    /// A wrapper around a type to assign to (this wrapper should implement
+    /// [QueryFragment]).
+    type QueryAstNode;
+
+    /// Move this in to the AST node which should implement [QueryFragment].
+    fn into_target(self) -> Self::QueryAstNode;
+}
```
weiznich, line 107: This likely should be defined in a `mod private {}` + then be rexported as `pub(crate)` item from this module to make this diesel private.

See here: https://github.com/diesel-rs/diesel/blob/c237a74b63900b734c659a201fc1d6b87108033b/diesel/src/backend.rs#L390 for an example of this strategy.

weiznich (commented):
Thanks for working on this. This is already in a good state as the implementation seems to work. I feel that we can simplify the implementation quite a bit by removing some layers of internal traits. That should make the change much easier to understand.

I've left a few comments on the corresponding lines. Hopefully the explain the idea sufficient details. If that's not the case please ask.

The general idea is to just have `AssignmentTarget` and provide two implementations:
* One generic for all columns
* One specific for `ArrayIndex`

Those impl should not cause  a conflicting impl problem. Both can directly construct the corresponding `QueryAstNode` return type.
We want to remove the generic `Expr` impl as it covers any expression, also those that are not valid in that context.

diesel_tests/tests/update.rs
```diff
@@ -307,3 +307,30 @@ fn upsert_with_sql_literal_for_target() {
...
+        .execute(connection)
+        .unwrap();
+    update(posts)
+        .set(tags.eq(vec!["programming", "rust"]))
+        .execute(connection)
+        .unwrap();
+    update(posts)
+        .set(tags.index(1).eq("postgres"))
+        .execute(connection)
+        .unwrap();
```
chusteven, line 331: Apologies that I can't contribute much to this review, but just want to understand that these two update statements are unblocked by this PR? If I understand correctly right now `diesel` doesn't provide any way to update records in the DB based on literals and this PR allows it to happen? If so, very neat!

If that's too reductive, then the only feedback I might be is to add tests for other types of updates (maybe a few from the PG documentation like an operator invocation (a + b), type cast (lit as i32), etc.)

pksunkara (commented):
Can you please add some documentation blurb? I am not too sure either what this PR unlocks without digging deep into the diesel code.

weiznich:
@pksunkara @chusteven This PR is about enabling queries like

```rust
diesel::update(my_table::table).set(array_col.index(1).eq(true))
```

```sql
UPDATE my_table SET array_col[1] = true
```

Especially about addressing a single array element as part of the `SET` clause of the update statement. That was not possible yet, as the left side of such a `SET` statement was always expected to be a column.

> If that's too reductive, then the only feedback I might be is to add tests for other types of updates (maybe a few from the PG documentation like an operator invocation (a + b), type cast (lit as i32), etc.)

These kinds of updates require flexibility on the right side of the `SET` statement, as everything there is on the right side of the assignment operator. Expressions like `col.eq(col + 1)` (or any other diesel expression) are already  supported there. 

Essentially this change is about allowing some flexibility on the left side of the assignment operator as well.

Justice4Joffrey:
I had assumed the update syntax provided would be standard (e.g. json `update foo set my_column->'a' = 1234`) but that doesn't seem to be the case. It's mentioned on the [arrays](https://www.postgresql.org/docs/current/arrays.html) page, but not as part of the syntax in [update](https://www.postgresql.org/docs/current/arrays.html).

The main motivation for providing `Uncorrelated` column and `AssignableExpression` was to make extending this for other expressions slightly less obscure (e.g. implement the `AssignableExpression` trait and rewrite the expression with `column.uncorrelate()`). In either case, though, I do agree that it's adding unecessary indirection.

Should get some time this week to address the above.

diesel/src/query_builder/update_statement/changeset.rs
Justice4Joffrey: Yes - that's a fair point. This doesn't need 2 trait layers.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -68,20 +69,76 @@ where
...
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct ExpressionWrapperForUpdate<C>(C);
+
+impl<DB, Expr> QueryFragment<DB> for ExpressionWrapperForUpdate<Expr>
+where
+    DB: Backend + DieselReserveSpecialization,
+    Expr: QueryFragment<DB>,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        QueryFragment::walk_ast(&self.0, out.reborrow())
```
weiznich, line 142: This is not needed anymore, right?

diesel/src/query_builder/mod.rs
```diff
@@ -55,7 +55,7 @@ pub use self::sql_query::{BoxedSqlQuery, SqlQuery};
 pub use self::upsert::on_conflict_target_decorations::DecoratableTarget;
 
 #[doc(inline)]
-pub use self::update_statement::changeset::AsChangeset;
+pub use self::update_statement::changeset::{AsChangeset, AssignmentTarget};
```
weiznich, line 58: I would prefer to not expose `AssignmentTarget` as part of the public API yet. Can we use a crate local reexport instead?

weiznich (approved):
This looks much cleaner now. Thanks for the update :+1:

I've found some now unused code that should be removed + I would like to keep `AssignmentTarget` "private" for now. With those minor things fixed this is ready for merging. Thanks again for working on this.

weiznich:
CI failures are caused by https://github.com/clap-rs/clap/issues/3648 / https://github.com/clap-rs/clap/pull/3646

Justice4Joffrey:
Referencing issue #1681

PR #3101, opened 2022-03-25
Make `RawBindCollector` fields public

weiznich:
Otherwise it's not possible to use this bind collector type from a third
party crate (like `diesel_async` or any pure rust mysql/postgres implementation)

### DIFF
--- a/diesel/src/expression_methods/mod.rs
+++ b/diesel/src/expression_methods/mod.rs
@@ -21,7 +21,7 @@ pub use self::global_expression_methods::{ExpressionMethods, NullableExpressionM
 #[doc(inline)]
 pub use self::text_expression_methods::TextExpressionMethods;
 
-#[cfg(feature = "postgres")]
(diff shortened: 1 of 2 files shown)


Ten0:
Should we make these pub based on the third-party-backend feature?

weiznich:
@Ten0 Good point :+1:

Another thing I would like to ask you: Are you interested in joining the diesel core team? I personally appreciate your technical comments here and started to see them as a good input for technical changes. So adding you to the core team would formalize this + allow you to publish new crate versions as well. (Which in turn would increase the bus factor of diesel to at least two active core maintainers).

Ten0:
> Another thing I would like to ask you: Are you interested in joining the diesel core team? I personally appreciate your technical comments here and started to see them as a good input for technical changes. So adding you to the core team would formalize this + allow you to publish new crate versions as well. (Which in turn would increase the bus factor of diesel to at least two active core maintainers).

Sure! :)

Issue #2342, opened 2020-03-27
Remove Trait bound From<Error> from Connection::transaction


gliderkite:
I am working on a crate that makes use of an Error type defined in a dependency library, which I use as return type of a method that needs to be run as part of a [transaction](https://docs.rs/diesel/1.4.4/diesel/connection/trait.Connection.html#method.transaction), such as for example:

```rust
use my_lib::Error;

fn foo() -> Result<(), Error>;

conn.transaction(|| { foo() });
```

This results in the following error when compiling:

> error[E0277]: the trait bound `my_lib::Error: std::convert::From<diesel::result::Error>` is not satisfied

AFAIK it's it not possible for me to create the `From<Error>` implementation required by this method, because the `my_lib::Error` type is defined in another crate. And I have to create my own `TransactionError` dummy type just to map the `my_lib::Error` (and map it back at the end of the transaction).

Is there a strong reason to have this Trait bound in place?
Would it be possible to remove it and make the API more flexible?

weiznich:
Please make a concrete suggestions how the API would be more flexible if we remove that bound. Also keep in mind that every use case that is supported today should be supported in the future, that includes using a custom error type as return type of the transaction callback.

I'm closing this issue now, because it is currently not actionable for the diesel team. If you have any concrete suggestions we can talk about reopening this issue.

gliderkite:
I wasn't aware of this policy in `Diesel` (this was the first issue I opened); in fact I see many open tickets that are not actionable today but still open, for example https://github.com/diesel-rs/diesel/issues/399#issuecomment-603491154, that by chance redirected me to this comment, that could be applied here as well https://github.com/diesel-rs/diesel/issues/399#issuecomment-603491154. I'll reopen this issue in case I'll find the time to think to an alternative, and leave to you the job to redirect other people opening similar issues to this closed one.

weiznich:
I probably should explain why I closed this issue a bit more in detail. 
The corresponding code for the [transaction function is here](https://github.com/diesel-rs/diesel/blob/5369a7a3d13b1323fcac9acf9d13256f19a970a0/diesel/src/connection/mod.rs#L90-L107) , to show what we are talking about. We have the following constraints there:

* We need to return some error type from that function because a transaction may fail
* A potential user should be able to provide domain specific callbacks there, this implies they may return their own error type
* We need to have some way to handle an internal error (like connection was lost, or something like that) wile doing the transaction handling.

[...]

PR #3102, opened 2022-03-26
WIP/Prototype  - Remove From<Error> bound from Connection::transaction

jacob-pro:
A possible solution to https://github.com/diesel-rs/diesel/issues/2342

### DIFF
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -2,6 +2,7 @@ use crate::connection::commit_error_processor::{CommitErrorOutcome, CommitErrorP
 use crate::connection::Connection;
 use crate::result::{DatabaseErrorKind, Error, QueryResult};
 use std::borrow::Cow;
+use std::fmt::{Debug, Display, Formatter};
 use std::num::NonZeroU32;
 
 /// Manages the internal transaction state for a connection.
@@ -43,7 +44,7 @@ pub trait TransactionManager<Conn: Connection> {
 
     /// Executes the given function inside of a database transaction
     ///
-    /// Each implementation of this function needs to fullfill the documented
+    /// Each implementation of this function needs to fulfill the documented
     /// behaviour of [`Connection::transaction`]
     fn transaction<F, R, E>(conn: &mut Conn, callback: F) -> Result<R, E>
     where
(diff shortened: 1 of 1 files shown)


Ten0:
It looks like this would definitely not be something we'd use, so I'm wondering to what extent this is worth the added complexity in the codebase and for users to generally consider. 

Our use case is using an `anyhow`-like error type to generate backtraces and return internal errors such as the ones encountered when using Diesel, which are then reported to an error monitoring system before being converted into errors destined to the end user. (so using this would decrease the quality of the backtrace, and have our functions juggle between unnecessary conversions) 

Can you:
- Provide an example of what your use-case is and how this would increase its ergonomics
- Show how it is not solved by defining this same enum you add in this PR in one of your own crates and using it as the error type for the already existing transaction function (specifically, it looks like it would even enable you to have an `From<TransactionError> for SomeOtherExternalError` impl which you couldn't have here because you wouldn't own any of the types nor the trait)

(NB: Possibly the issue is a better place for this than the PR?)

weiznich:
@jacob-pro If you expect this to  discussed you need to provide answers to the question above. Otherwise I likely will close this PR next week due to inactivity.

jacob-pro:
1. So my usecase is when building an API (using actix), I want to return a standardised `http_api_problem` error (which is defined in an external crate) from each of the route handlers.

3. Defining the enum in my own crate does work, its just that then I have this boilerplate code that I either: (a) have to copy and paste everywhere, or (b) create a very tiny library just containing the code I have included in this PR, it would in my opinion be preferable to have this as part of the diesel API itself...

4. So this is a valid concern, but it the usecase I described the same problem exists even without transactions, i.e. say I want to perform just a simple database query, I already have to have my own function trait for converting the disel error to the `http_api_problem` error in the firstplace since I can't implement From. 

Example: https://github.com/jacob-pro/calpol/blob/525fba9cce33cdb07460deaa34e748ff35d874a3/src/api/v1/users.rs#L171-L176

Ten0:
https://github.com/diesel-rs/diesel/issues/2342#issuecomment-1086631629

Ten0:
https://github.com/diesel-rs/diesel/issues/2342#issuecomment-1086649534

PR #2827, opened 2021-07-14
Various insert improvments

weiznich:
* Cache insert statments from queries, as those are safe to cache if the
  query itself is safe to cache
* Add a way to directly insert from a statically sized array
* Decouple default values from insert expressions and allow to generate
  insertable implementations that do not use the default values.
   + This unblocks real batch inserts on sqlite
   + This allows cache the prepared statements for insert from statically sized arrays as we know the size at compile time here
* Update the benchmarks to use the statically compile time variant.
  Other implementations also use prepared statements here, so this is
just fair

PR #2507, opened 2020-09-17
Rework benchmarks

weiznich:
Create a distinct benchmark suite that tests various different rust database connection crates: Those benchmarks are created with the following goals:

a) To track diesels performance and find potential regressions
b) To evaluate potential alternatives for the currently used C-dependencies
c) To compare diesel with the competing crates

It currently supports the following database systems:

* PostgreSQL
* MySQL/MariaDB
* SQLite

and the following crates:

* Diesel
* [SQLx](https://github.com/launchbadge/sqlx)
* [Rustorm](https://github.com/ivanceras/rustorm)
* [Quaint](https://github.com/prisma/quaint)
* [Postgres](https://github.com/sfackler/rust-postgres)
* [Rusqlite](https://github.com/rusqlite/rusqlite)
* [Mysql](https://github.com/blackbeam/rust-mysql-simple)

I've included other crates as well into this setup as there are at least 2 pure rust database connection implementations for postgres and mysql. As an long term goal it can be interesting for diesel to switch away from using the corresponding c based libraries for those database to simplify the setup + replace potential unsafe C code with mostly safe rust code. In my opinion such a replacement should at least full fill the following requirements:

* Provide a comparable level of performance to the platform native c libraries
* Be a mature actively maintained crate

Additionally a comparison with those crates provides in my opinion at least a crude way to estimate how large the overhead introduced by diesel is compared to manually handling everything.

By default only the diesel benchmarks are build and executed. For all other benchmarks you need to specify their corresponding feature flags (depending on the crate, more than one, see the Readme for details)

# Open Questions:

* How to integrate this into our CI? (I think we should check that the diesel benchmarks still compile during normal CI and have some special `@github-thing run-benches` or so to get a diesel benchmark comparison between the current PR and the master branch?)
* Are the benchmarks meaningful? Do we miss some important case?

cc @mehcode @pimeys @ivanceras @gwenn @sfackler @blackbeam as the corresponding benchmarks could also be interesting for your crates (If not, sorry for bother you in advance)

PR #3103, opened 2022-03-28
Benchmarks with instruction counts

weiznich:
Hopefully this works also on the CI, otherwise this is only an option for local runs.

### DIFF
--- a/diesel_bench/Cargo.toml
+++ b/diesel_bench/Cargo.toml
@@ -51,6 +51,7 @@ sqlite = ["diesel/sqlite"]
 mysql = ["diesel/mysql"]
 sqlx-bench = ["sqlx", "tokio", "sqlx/runtime-tokio-rustls"]
 instruction_count = ["criterion-perf-events", "perfcnt"]
+fast_run = []
 
 [replace]
 "quaint:0.2.0-alpha.13" = {git = "https://github.com/prisma/quaint", rev = "e077df3"}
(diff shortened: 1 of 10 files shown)


chusteven (commented):
Unrelated to this change, didn't realize `diesel` had actions that ran benchmarks against other ORMs! What a neat way to develop a library 😊 -- curious if there have been any past conversations/issues/etc. where I can see the outcome of past benchmark results and how that's impacted development?

weiznich:
We use the diesel part of these benchmarks to check if changes introduce regression/improvements in terms of performance. If we think that we can run the benchmarks via the `run-benchmarks` label. That means [any PR that has this label](https://github.com/diesel-rs/diesel/pulls?q=is%3Apr+label%3Arun-benchmarks+) used this to make a decision. I found it helpful for https://github.com/diesel-rs/diesel/pull/2799, https://github.com/diesel-rs/diesel/pull/2827 but also with https://github.com/diesel-rs/diesel/pull/2774 and https://github.com/diesel-rs/diesel/pull/2788 (to name two PR's where the change did not result in the expected performance change.)

As for comparing with the other frameworks: That's mostly there to get a general feeling where diesel preforms well and where it does not perform well. Although the maintainers of the other  frameworks are free to use those benchmarks too. (https://github.com/diesel-rs/diesel/pull/2507 has some history on that). The [metrics](https://github.com/diesel-rs/metrics/) repository contains some statistics for the full runs. Additionally I normally point to these numbers to support the claim that diesel performs quite will. On the other hand those numbers show quite good where diesel could potentially be improved. (Basically everywhere where something else is faster as diesel should be designed that way that it only adds minimal or even non overhead compared against executing queries manually in the best potentially way).

PR #3107, opened 2022-03-31
Add a codeowners file to automatically request reviews from the revie…

weiznich:
…wers team

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @diesel-rs/reviewers


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @diesel-rs/reviewers
```
JohnTitor, line 1: Seems we have to give the write access to the team (looking at https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax)?

weiznich: That's unfortunate. We should not need to do that. I will close this PR as this means it won't work that way.

PR #3109, opened 2022-03-31
Optimize sqlite `StatementIterator`

weiznich:
This change remove a temporary state from the sqlite StatementIterator
which showed up in some profiles for our benchmarks. The benchmarks show
a 1-6% improvement in construction count for the query benchmarks and
no changes for the insert benchmarks.

<details>
<summary> Relevant benchmark results </summary>

```
bench_trivial_query/diesel/1
                        time:   [6541.7627 cycles 6542.1171 cycles 6542.5357 cycles]
                        change: [-3.0210% -2.8542% -2.7324%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  6 (6.00%) high mild
  7 (7.00%) high severe
                        time:   [13976.9443 cycles 13981.9545 cycles 13986.5513 cycles]
                        change: [-1.4962% -1.4095% -1.3296%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  2 (2.00%) high severe
                        time:   [40414.1384 cycles 40456.7169 cycles 40500.9642 cycles]
                        change: [-0.7671% -0.6327% -0.5016%] (p = 0.00 < 0.05)
                        time:   [23882021.3000 cycles 23882214.7229 cycles 23882406.7944 cycles]
                        change: [-6.0633% -6.0461% -6.0331%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  2 (2.00%) high mild
                        time:   [23890404.5122 cycles 23890792.3608 cycles 23891254.8076 cycles]
                        change: [-6.0062% -6.0030% -5.9998%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe
                        time:   [27181116.5123 cycles 27189714.1840 cycles 27198788.2537 cycles]
                        change: [-5.4711% -5.4298% -5.3904%] (p = 0.00 < 0.05)
                        Performance has improved.

                        time:   [22495.0324 cycles 22496.0806 cycles 22497.3718 cycles]
                        change: [-1.5580% -1.4632% -1.3608%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe
                        time:   [41632.5260 cycles 41681.6865 cycles 41730.2439 cycles]
                        change: [-1.0619% -0.9198% -0.7813%] (p = 0.00 < 0.05)
  1 (1.00%) high mild
                        time:   [126045.5130 cycles 126164.1230 cycles 126292.1045 cycles]
                        change: [-0.4169% -0.2920% -0.1611%] (p = 0.00 < 0.05)
  2 (2.00%) high mild
                        time:   [23650424.9443 cycles 23653430.5380 cycles 23656376.8618 cycles]
                        change: [-3.0016% -2.9868% -2.9741%] (p = 0.00 < 0.05)
                        Performance has improved.
                        time:   [23666598.7259 cycles 23673235.8710 cycles 23683469.4276 cycles]
                        change: [-2.9908% -2.9637% -2.9209%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
                        time:   [50436656.7219 cycles 50448976.0640 cycles 50461415.4847 cycles]
                        change: [-3.1256% -3.0928% -3.0606%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.2s, enable flat sampling, or reduce sample count to 60.
Benchmarking bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially: Collecting 100 samples in estimated 5.1724 s (5050 iter                                                                                                                                                            bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
                        time:   [4011704.3305 cycles 4012211.9494 cycles 4012708.0785 cycles]
                        change: [-3.6397% -3.0779% -1.9779%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  4 (4.00%) low severe
  4 (4.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe

<details>

### DIFF
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -23,7 +23,7 @@ impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
         // a user stored the row in some long time container before calling next another time
         // In this case we copy out the current values into a temporary store and advance
         // the statement iterator internally afterwards
-        let mut last_row = {
+        let last_row = {
             let mut last_row = match outer_last_row.try_borrow_mut() {
                 Ok(o) => o,
                 Err(_e) => {
@@ -38,7 +38,7 @@ impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
             let duplicated = last_row.duplicate(column_names);
             std::mem::replace(last_row, duplicated)
         };
-        if let PrivateSqliteRow::Direct(ref mut stmt) = last_row {
+        if let PrivateSqliteRow::Direct(mut stmt) = last_row {
             let res = unsafe {
                 // This is actually safe here as we've already
                 // performed one step. For the first step we would have
@@ -46,18 +46,23 @@ impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
                 // have access to `PrivateSqliteRow` at all
                 stmt.step(false)
             };
+            *outer_last_row = Rc::new(RefCell::new(PrivateSqliteRow::Direct(stmt)));
             match res {
                 Err(e) => Some(Err(e)),
                 Ok(false) => None,
                 Ok(true) => {
                     let field_count = field_count;
                     Some(Ok(SqliteRow {
-                        inner: Rc::clone(&outer_last_row),
+                        inner: Rc::clone(outer_last_row),
                         field_count,
                     }))
                 }
             }
         } else {
+            match last_row {
+                PrivateSqliteRow::Direct(_) => dbg!("direct"),
+                PrivateSqliteRow::Duplicated { .. } => dbg!("duplicated"),
+            };
             // any other state than `PrivateSqliteRow::Direct` is invalid here
             // and should not happen. If this ever happens this is a logic error
             // in the code above
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -59,10 +59,6 @@ impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
                 }
             }
         } else {
-            match last_row {
-                PrivateSqliteRow::Direct(_) => dbg!("direct"),
-                PrivateSqliteRow::Duplicated { .. } => dbg!("duplicated"),
-            };
             // any other state than `PrivateSqliteRow::Direct` is invalid here
             // and should not happen. If this ever happens this is a logic error
             // in the code above
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -12,16 +12,74 @@ pub struct StatementIterator<'stmt, 'query> {
     field_count: usize,
 }
 
+impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
+    #[cold]
+    fn handle_duplicated_row_case(
+        outer_last_row: &mut Rc<RefCell<PrivateSqliteRow<'stmt, 'query>>>,
+        column_names: &mut Option<Rc<[Option<String>]>>,
+        field_count: usize,
+    ) -> Option<QueryResult<SqliteRow<'stmt, 'query>>> {
+        // We don't own the statement. There is another existing reference, likly because
+        // a user stored the row in some long time container before calling next another time
+        // In this case we copy out the current values into a temporary store and advance
+        // the statement iterator internally afterwards
+        let mut last_row = {
+            let mut last_row = match outer_last_row.try_borrow_mut() {
+                Ok(o) => o,
+                Err(_e) => {
+                    return Some(Err(crate::result::Error::DeserializationError(
+                                    "Failed to reborrow row. Try to release any `SqliteField` or `SqliteValue` \
+                                     that exists at this point"
+                                        .into(),
+                                )));
+                }
+            };
+            let last_row = &mut *last_row;
+            let duplicated = last_row.duplicate(column_names);
+            std::mem::replace(last_row, duplicated)
+        };
+        if let PrivateSqliteRow::Direct(ref mut stmt) = last_row {
+            let res = unsafe {
+                // This is actually safe here as we've already
+                // performed one step. For the first step we would have
+                // used `PrivateStatementIterator::NotStarted` where we don't
+                // have access to `PrivateSqliteRow` at all
+                stmt.step(false)
+            };
+            match res {
+                Err(e) => Some(Err(e)),
+                Ok(false) => None,
+                Ok(true) => {
+                    let field_count = field_count;
+                    Some(Ok(SqliteRow {
+                        inner: Rc::clone(&outer_last_row),
+                        field_count,
+                    }))
+                }
+            }
+        } else {
+            // any other state than `PrivateSqliteRow::Direct` is invalid here
+            // and should not happen. If this ever happens this is a logic error
+            // in the code above
+            unreachable!(
+                "You've reached an impossible internal state. \
+                             If you ever see this error message please open \
+                             an issue at https://github.com/diesel-rs/diesel \
+                             providing example code how to trigger this error."
+            )
+        }
+    }
+}
+
 enum PrivateStatementIterator<'stmt, 'query> {
-    NotStarted(StatementUse<'stmt, 'query>),
+    NotStarted(Option<StatementUse<'stmt, 'query>>),
     Started(Rc<RefCell<PrivateSqliteRow<'stmt, 'query>>>),
-    TemporaryEmpty,
 }
 
 impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
     pub fn new(stmt: StatementUse<'stmt, 'query>) -> StatementIterator<'stmt, 'query> {
         Self {
-            inner: PrivateStatementIterator::NotStarted(stmt),
+            inner: PrivateStatementIterator::NotStarted(Some(stmt)),
             column_names: None,
             field_count: 0,
         }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: diesel/src/sqlite/connection/statement_iterator.rs [hunk 1]
```diff
@@ -12,16 +12,79 @@ pub struct StatementIterator<'stmt, 'query> {
...
+                        inner: Rc::clone(outer_last_row),
+                        field_count,
+                    }))
+                }
+            }
+        } else {
+            match last_row {
+                PrivateSqliteRow::Direct(_) => dbg!("direct"),
+                PrivateSqliteRow::Duplicated { .. } => dbg!("duplicated"),
+            };
```
weiznich, line 65:
```suggestion
```

diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -12,16 +12,75 @@ pub struct StatementIterator<'stmt, 'query> {
     field_count: usize,
 }
 
+impl<'stmt, 'query> StatementIterator<'stmt, 'query> {
+    #[cold]
+    fn handle_duplicated_row_case(
+        outer_last_row: &mut Rc<RefCell<PrivateSqliteRow<'stmt, 'query>>>,
+        column_names: &mut Option<Rc<[Option<String>]>>,
+        field_count: usize,
+    ) -> Option<QueryResult<SqliteRow<'stmt, 'query>>> {
```
chusteven, line 21: Curious what this kind of duplication handling is needed for. Is this a SQLite or SQLite driver bug?

diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -12,16 +12,75 @@ pub struct StatementIterator<'stmt, 'query> {
...
+                    return Some(Err(crate::result::Error::DeserializationError(
+                                    "Failed to reborrow row. Try to release any `SqliteField` or `SqliteValue` \
+                                     that exists at this point"
+                                        .into(),
+                                )));
+                }
+            };
+            let last_row = &mut *last_row;
+            let duplicated = last_row.duplicate(column_names);
+            std::mem::replace(last_row, duplicated)
```
chusteven, line 39: Unrelated to this PR, I would've assume based solely on the method name that this is an `unsafe` method?

chusteven (commented):
Curious only about how come such code handling "duplicated" rows was needed in the first place? Or maybe I'm misunderstanding this module entirely 😅

diesel/src/sqlite/connection/statement_iterator.rs
weiznich: [`std::mem::replace`](https://doc.rust-lang.org/std/mem/fn.replace.html) is actually a safe function. It replaces the value at the location of the first argument with the value from the second argument, while returning the old value.

diesel/src/sqlite/connection/statement_iterator.rs
weiznich: The reasoning here is a bit "complicated". The main issue here is that `SQLite` interface returns something like an iterator, but not exactly an iterator according to rusts definition. The main difference here is that `Iterator::next` is defined as `fn next(&mut self) -> Option<Self::Item>`. That implies that each item is independent from the underlying iterator, so it cannot hold a reference to the original iterator at all.

Now sqlites interface for prepared statements requires multiple functions to setup a statement. It boils down to the following steps:
1. Create a prepared statement itself by using [`sqlite3_prepare_*`](https://sqlite.org/c3ref/prepare.html)
2. Bind input via [`sqlite3_bind_*`](https://sqlite.org/c3ref/bind_blob.html)
3. Call [`sqlite3_step`](https://sqlite.org/c3ref/step.html) to start processing the statement
4. Call for each column [`sqlite3_column_value`](https://sqlite.org/c3ref/column_blob.html) to get a pointer to the corresponding column inside of the sqlite statement.
5. Repeat 3. + 4. for all rows 

The problem here is that step 5. will invalid the pointers received in step 4. They now will point to the corresponding values for the new row. But if we use the `Iterator` interface here users could hold a previous row and try to access those values. Now we could say we just cannot use `Iterator` here as that's done by the other database libraries. Or we could do some clever stuff. Basically we return an `Rc<…>` as row. That allows us to hold an reference to the previous row inside of our iterator implementation. We then can check if someone holds the previous row or not by just checking the strong count of the `Rc<…>`. If that count is 1 our internal instance is the only reference to the previous row. That means  we can just overwrite the content there and reuse the allocation. That's the happy path.
If the strong count is not 1, someone else has a reference to the previous row. Now as the row is internally `Rc<RefCell<_>>` we are able to modify the content of their row behind the scenes. We do that by just cloning the corresponding column values for this last row and injecting the cloned values in the last row. In that way we can forward the statement iterator another step as the value ptrs are now not used by the previous row. That's the duplicated_row_case. I assume that this is the unlikely path, therefore the separate function + `#[cold]`. This assumption is based on the observation that diesels own `RunQueryDsl` basically always hits the happy path, as we don't hold an reference to the last row at all. To hit the other path users must directly call `Connection::load()` and store the rows somewhere.

chusteven: Thank you so much for your answer! I learned a little bit more about unsafe code and use cases for `Rc` and `RefCell` structs. I'm still a bit confused overall especially on this point

> That allows us to hold an reference to the previous row inside of our iterator implementation. We then can check if someone holds the previous row or not by just checking the strong count of the Rc<…>.

Like... who else would hold this value? If `diesel` is the only thing responsible for iterating through the returned data from the sqlite driver, then why would this be needed? EDIT: Now actually I see "To hit the other path users must directly call Connection::load() and store the rows somewhere." I think I may have answered my own question 😅 


Also I think I don't understand statements very well; my understanding was always that a prepared statement was something the client sent to the database to say "hey expect that the signature of this query I'm about to want to run is X" and then when we actually send along the parameters we're basically delegating to the database to do sanitization and data type assertions (to avoid SQL injection); so I'm a bit confused when it seems like we actually use [the statement struct itself to store the values (by values I mean the actual integers, strings, etc. that are returned for each row) themselves](https://github.com/diesel-rs/diesel/blob/c237a74b63900b734c659a201fc1d6b87108033b/diesel/src/sqlite/connection/statement_iterator.rs#L49-L60)

chusteven: Oh and by the way, now that I've sat on this some more I think I see where the complexity is; and that's in the value of each row needing to be stored via some pointer that is perpetually written and re-written -- I guess that's very memory efficient but also potentially disastrous in case multiple tasks are trying to read the value at that point in a concurrent manner and have race conditions (which it seems like `diesel` is handling intelligently via the reference counting... although I wonder now why `Rd` and not `Arc` but that's a separate issue -- I've learned enough today :D)

Ten0: > why `Rc` and not `Arc`

`Rc<RefCell>` would have to be replaced by `Arc<Mutex>` to allow the user to send it across threads, the performance overhead of which seems not worth considering how unlikely it is that someone actually needs this feature:

> diesels own `RunQueryDsl` basically always hits the happy path, as we don't hold an reference to the last row at all

chusteven: Thanks! Still trying to pick up pieces of diesel and I'll look into the `RunQueryDsl` trait next :)

weiznich: > Also I think I don't understand statements very well; my understanding was always that a prepared statement was something the client sent to the database to say "hey expect that the signature of this query I'm about to want to run is X" and then when we actually send along the parameters we're basically delegating to the database to do sanitization and data type assertions (to avoid SQL injection); so I'm a bit confused when it seems like we actually use the statement struct itself to store the values (by values I mean the actual integers, strings, etc. that are returned for each row) themselves

A prepared statement does more than just checking and serializing bind value. It's a way to tell the server that we want to execute a certain query with the following variables. What a clever optimizing server now could do is to generate an executable optimized representation of that query internally and link it to the prepared statement. In the end SQL is just an interpreted programming language, so the server may prepare that early on. As a client library we can use this behaviour in various ways:

1. We can delegate bind serialization + sanitizing to the server, as prepared statements allow us to use place holder values
2. We can store references to the prepared statements that are likely to be reexecuted later on. This saves the amount of time needed to prepare the statement on query reexecution. (Checkout the [`diesel::connection::statement_cache`](https://docs.diesel.rs/master/diesel/connection/statement_cache/index.html) module in diesel for details about how this is handled in diesel)

Depending on the actual database implementation prepared statements could bundle more functionality. For SQLite a statement could be executed and the execution is tracked inside the statement itself. That seems to be meaningful if you think about a prepared statement as a function/state combination that represents some SQL query. A SQLite prepared statements is basically a program that waits to be executed. You can start the execution at any time and for a stared execution it will track the current state. If required you can reset it to the starting point by using [`sqlite3_reset`](https://sqlite.org/c3ref/reset.html)

> > why Rc and not Arc
>
> Rc<RefCell> would have to be replaced by Arc<Mutex> to allow the user to send it across threads, the performance overhead of which seems not worth considering how unlikely it is that someone actually needs this feature:

The main reason to not using a `Arc<Mutex<_>>` there is that it is really hard to use `SqliteConnection` in multiple threads at the same time at all. The type implements only `Send`, but not `Sync`. That means you cannot execute queries from different threads using the same connection at the same time. This also applies to the returned iterator. You cannot call next from different threads at the same time anyway. That's the main reason to use `Rc<RefCell<_>>` instead of `Arc<Mutex<_>>`.

PR #3111, opened 2022-04-01
Bench updates

weiznich:
* Fix the diesel_async benchmarks
* Bump some dependencies
* Use batch inserts for all crates supporting sqlite so that this is a fair benchmark

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2018"
 byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.19", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
-libsqlite3-sys = { version = ">=0.17.2, <0.24.0", optional = true, features = ["bundled_bindings"] }
+libsqlite3-sys = { version = ">=0.17.2, <0.25.0", optional = true, features = ["bundled_bindings"] }
 mysqlclient-sys = { version = "0.2.5", optional = true }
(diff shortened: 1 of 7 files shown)


chusteven (commented):
Looks good. And on the topic of async, one thing I noticed when trawling some old message boards is that `diesel_async` depends on not the native drivers like `libpq` but rather `tokio-postgres`. Perhaps this gets into the weeds of DB drivers and the like (for instance, I only recently learned that most DBs ship with native C libraries for other programs to take care of SQL parsing, wire protocols, network calls, etc. 😅 ), but is having two separate dependencies a concern for diesel as far as diesel+async is concerned?

weiznich:
>  Perhaps this gets into the weeds of DB drivers and the like (for instance, I only recently learned that most DBs ship with native C libraries for other programs to take care of SQL parsing, wire protocols, network calls, etc. 😅 ), but is having two separate dependencies a concern for diesel as far as diesel+async is concerned?

That's the reason for the distinction of `Backend` and `Connection` in diesel. A `Backend` describes a specific sql database implementation. This includes things like the specific sql syntax or how the database serializes values. A `Connection` describes how should diesel interacts with the database, so essentially which low level driver is utilized to talk with the database. The design explicitly allows that a `Backend` has more than one `Connection` implementation.

For diesel we have different feature flags for the `Pg` backend (`postgres_backend`) and for the `PgConnection` implementation (`postgres`, which implies `postgres_backend`).  This allows users to better opt in what should be included and what not. For the diesel_async use case I just enable the `postgres_backend` feature to have access to the backend implementation. That removes the dependency to libpq as we don't need `PgConnection` there.

PR #3117, opened 2022-04-05
ONLY <table> postgres keyword (second attempt)

Ploppz:
`Only<S>` (the result of `mytable.only()`) implements `Table`, and only works on `Pg` backend.

I had to add some new macros with long names `__diesel_internal_backend_specific_*`. For example for a postgres-only part of the `allow_tables_to_appear_in_same_query` macro.

### DIFF
--- a/diesel/src/pg/expression/extensions/only_dsl.rs
+++ b/diesel/src/pg/expression/extensions/only_dsl.rs
@@ -1,0 +1,54 @@
+use crate::query_builder::Only;
+use crate::Table;
+
+/// The `only` method
+///
+/// This is only implemented for the Postgres backend.
+/// The `ONLY` clause is used to select only from one table and not any inherited ones.
+///
+/// Calling this function on a table (`mytable.only()`) will result in the SQL `ONLY mytable`.
+/// `mytable.only()` can be used just like any table in diesel since it implements
+/// [Table](crate::Table).
+///
+/// Example:
+///
+/// ```rust
+/// # include!("../../../doctest_setup.rs");
+/// # use schema::{posts, users};
+/// # use diesel::dsl::*;
+/// # fn main() {
+/// # let connection = &mut establish_connection();
+/// let n_sers_in_main_table = users::table
+///     .only()
+///     .select(count(users::id))
+///     .first::<i64>(connection);
+/// # }
+/// ```
+/// Selects the number of entries in the `users` table excluding any rows found in inherited
+/// tables.
+///
+/// It can also be used in inner joins:
+///
+/// ```rust
+/// # include!("../../../doctest_setup.rs");
+/// # use schema::{posts, users};
+/// # use diesel::dsl::*;
+/// # fn main() {
+/// # let connection = &mut establish_connection();
+/// # let _ =
+/// users::table
+///     .inner_join(posts::table.only())
+///     .select((users::name, posts::title))
+///     .load::<(String, String)>(connection);
+/// # }
+/// ```
+/// That query excludes any posts that reside in any inherited table.
+///
+pub trait OnlyDsl: Table {
+    /// See the trait-level docs.
+    fn only(self) -> Only<Self> {
+        Only { source: self }
+    }
+}
+
+impl<T: Table> OnlyDsl for T {}
--- a/diesel/src/pg/query_builder/only.rs
+++ b/diesel/src/pg/query_builder/only.rs
@@ -1,0 +1,95 @@
+use crate::expression::{Expression, ValidGrouping};
+use crate::pg::Pg;
+use crate::query_builder::{AsQuery, AstPass, FromClause, QueryFragment, QueryId, SelectStatement};
+use crate::query_source::QuerySource;
+use crate::result::QueryResult;
+use crate::{JoinTo, SelectableExpression, Table};
+
+/// Represents a query with an `ONLY` clause.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Only<S> {
+    pub(crate) source: S,
+}
+
+impl<S> QueryId for Only<S>
+where
+    Self: 'static,
+    S: QueryId,
+{
+    type QueryId = Self;
+    const HAS_STATIC_QUERY_ID: bool = <S as QueryId>::HAS_STATIC_QUERY_ID;
+}
+
+impl<S> QuerySource for Only<S>
+where
+    S: Table + Clone,
+    <S as QuerySource>::DefaultSelection: ValidGrouping<()> + SelectableExpression<Only<S>>,
+{
+    type FromClause = Self;
+    type DefaultSelection = <S as QuerySource>::DefaultSelection;
+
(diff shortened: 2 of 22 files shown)


Ten0:
Where's the first attempt?

weiznich:
Relevant context:
* https://github.com/diesel-rs/diesel/pull/2944
* https://gitter.im/diesel-rs/diesel?at=622b2914c435002500e726fd
* https://github.com/diesel-rs/diesel/pull/3090

weiznich:
I just want to leave a comment here that I likely do not have the time to review that this week.

diesel/src/pg/expression/extensions/only_dsl.rs
```diff
@@ -0,0 +1,41 @@
...
+/// That query excludes any posts that reside in any inherited table.
+///
+pub trait OnlyDsl: AsQuery + Sized + QuerySource + Clone {
+    /// See the trait-level docs.
+    fn only(self) -> Only<Self> {
+        Only { source: self }
+    }
+}
+
+impl<T: QuerySource + AsQuery + Sized + Clone> OnlyDsl for T {}
```
weiznich, line 41: Shouldn't that be bound to `Table` only, instead of query source? I mean calling `.only()` on for example an alias is not meaningful, right?

diesel/src/pg/query_builder/only.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{
+    expression::Expression,
+    expression::ValidGrouping,
+    pg::Pg,
+    query_builder::{AsQuery, AstPass, FromClause, QueryFragment, QueryId, SelectStatement},
+    query_source::QuerySource,
+    result::QueryResult,
+    JoinTo, SelectableExpression, Table,
+};
```
weiznich, line 9: Diesel does use unnested use statements everywhere else. I would prefer keeping it that way.

diesel/src/pg/query_builder/only.rs
```diff
@@ -0,0 +1,96 @@
...
+    <S as QuerySource>::DefaultSelection: ValidGrouping<()> + SelectableExpression<Only<S>>,
+{
+    type PrimaryKey = <S as Table>::PrimaryKey;
+    type AllColumns = <S as Table>::AllColumns;
+    fn primary_key(&self) -> Self::PrimaryKey {
+        self.source.primary_key()
+    }
+    fn all_columns() -> Self::AllColumns {
+        S::all_columns()
+    }
```
weiznich, line 95:
```suggestion
    type AllColumns = <S as Table>::AllColumns;
    
    fn primary_key(&self) -> Self::PrimaryKey {
        self.source.primary_key()
    }
    
    fn all_columns() -> Self::AllColumns {
        S::all_columns()
    }
```

diesel_tests/tests/only.rs
```diff
@@ -0,0 +1,236 @@
+use crate::schema::connection;
+// use crate::schema_dsl::*;
```
weiznich, line 2:
```suggestion
```

diesel_tests/tests/only.rs
```diff
@@ -0,0 +1,236 @@
...
+use diesel::dsl::*;
+use diesel::pg::expression::dsl::OnlyDsl;
+use diesel::*;
+
```
weiznich, line 6: Can we add a short comment here why we need different table definition than the default ones?

(I know that's because of inheritance, but I stop every time here and think: Why are they here?)

diesel_tests/tests/only.rs
```diff
@@ -0,0 +1,236 @@
+use crate::schema::connection;
+// use crate::schema_dsl::*;
+use diesel::dsl::*;
+use diesel::pg::expression::dsl::OnlyDsl;
+use diesel::*;
```
weiznich, line 5: I would prefer using `diesel::prelude::*;` instead.
```suggestion
use diesel::*;
```

diesel_tests/tests/only.rs
```diff
@@ -0,0 +1,236 @@
...
+fn inner_join_only() {
+    // Test that it's possible to call `.inner_join(X::table.only().on(...)`
+    let connection = &mut connection();
+    setup_tables(connection);
+    test_scenario(connection);
+
+    println!(
+        "{}",
+        diesel::query_builder::debug_query(&users::table.only().inner_join(posts::table.only()))
+    );
```
weiznich, line 197:
```suggestion
```

diesel_tests/tests/only.rs
```diff
@@ -0,0 +1,236 @@
...
+            .count()
+            .get_result::<i64>(connection)
+            .unwrap(),
+        1
+    );
+}
+
+#[test]
+fn inner_join_only() {
+    // Test that it's possible to call `.inner_join(X::table.only().on(...)`
```
weiznich, line 189: Specifically this test is missing. I suggest that we either remove that comment or add a corresponding test :wink:

weiznich (approved):
Thanks for doing this implementation. The code looks already quite good :+1:

I've left a few small notes about minor stylistic things that I would like to see fixed before merging this PR. Other that that this feature should have a changelog entry as well.

Ploppz:
@weiznich I think I have addressed all your comments in the most recent changes

diesel/src/pg/expression/extensions/only_dsl.rs
```diff
@@ -0,0 +1,41 @@
...
+/// This is only implemented for the Postgres backend.
+/// The `ONLY` clause is used to select only from one table and not any inherited ones.
+///
+/// Calling this function on a table (`mytable.only()`) will result in the SQL `ONLY mytable`.
+/// `mytable.only()` can be used just like any table in diesel since it implements
+/// [Table](crate::Table).
+///
+/// Example:
+///
+/// ```ignore
```
weiznich, line 15: I would prefer not to introduce new `ignore` doc tests. We should at least check that any code in our documentation compiles.

(Same applies for the test below)

weiznich (approved):
Thanks for the update :+1: 
With fixing the doc tests so that the actually check that the example is at least compiling I'm fine with this PR.

diesel/src/pg/expression/extensions/only_dsl.rs
Ploppz: Is there a module from which I can import the `users` and `posts` tables? I think it's only in `diesel_tests::schema` - what do I do, define them anew?

weiznich: https://github.com/diesel-rs/diesel/blob/c237a74b63900b734c659a201fc1d6b87108033b/diesel/src/doctest_setup.rs 

which is included via `include!()` in other doc tests. See here for example https://github.com/diesel-rs/diesel/blob/c237a74b63900b734c659a201fc1d6b87108033b/diesel/src/insertable.rs#L42
(You likely need to adjust the path, as it's relative to the current source file.)

Issue #2963, opened 2021-11-26
Document what to do when pooled connection rollback fails


mleonhard:
[`diesel::connection::Connection::transaction`](https://docs.rs/diesel/1.4.8/diesel/connection/trait.Connection.html#method.transaction) docs say:
> ```rust
> fn transaction<T, E, F>(&self, f: F) -> Result<T, E> where
>     F: FnOnce() -> Result<T, E>,
>     E: From<Error>,
> ```
> Executes the given function inside of a database transaction
> 
> If there is already an open transaction, savepoints will be used instead.
> 
> If the transaction fails to commit due to a `SerializationFailure`, a rollback will be attempted. If the rollback succeeds, the original error will be returned, otherwise the error generated by the rollback will be returned. In the second case the connection should be considered broken as it contains a uncommitted unabortable open transaction.

I'm using [`diesel::r2d2::Pool`](https://docs.rs/diesel/1.4.8/diesel/r2d2/struct.Pool.html) with [`diesel::r2d2::ConnectionManager::<PgConnection>`](https://docs.rs/diesel/1.4.8/diesel/r2d2/struct.ConnectionManager.html).  I need the docs to answer these two questions:
- How can I close a connection with an uncommited and unabortable transaction and avoid returning it to the pool?
- Is this even necessary when using the connection pool?

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

weiznich:
Generally speaking: If something is not documented we do not guarantee a certain behavior. That's in most cases not because the docs are missing but because we haven't decided whats the right behavior there.
In this specific case we would be happy if someone starts a discussion what behavior should be guaranteed here. If that's decided we happily accept a PR updating the implementation and the documentation.
The discussion in #2935 might be relevant here as well.

chusteven:
[...]

PR #3122, opened 2022-04-06
Add documentation to diesel's r2d2 module on what happens when connections fail inside a pool

chusteven:
Addresses questions from https://github.com/diesel-rs/diesel/issues/2963

I wasn't sure if this was the best place to put this documentation as this is a module level docstring, though only concerns itself with errors, so... 🤷‍♂️  perhaps [here](https://github.com/diesel-rs/diesel/blob/a12f95f70d8f32eaefb42ec9d4e4b93904a51c96/diesel/src/r2d2.rs#L78) is a better place to put it? In any case, wanted to push up this documentation first for a check on accuracy

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -1,6 +1,20 @@
 //! Connection pooling via r2d2.
 //!
 //! Note: This module requires enabling the `r2d2` feature
+//!
+//! When used inside a pool, if an individual connection becomes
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks, that looks good to me :+1:

I would say that this is the correct place to document this as the `R2D2Connection` trait is a bit more specific there.

Issue #1951, opened 2019-01-10
Document proper usage of r2d2


thsowers:
Visiting https://github.com/diesel-rs/r2d2-diesel has a deprecation notice at the top pointing towards the r2d2 module powered by: https://github.com/sfackler/r2d2, which has great documentation on the package. However on our end there is no documentation on how integration with diesel works. Should we document how to properly use this?

The source in diesel specifies "See the [r2d2 documentation] for usage examples." and then points "[r2d2 documentation]" to the same file. Of the three tests in this file, only one actually utilizes diesel, and very minimally: `select("foo".into_sql::<Text>());`. 

Could we add realistic example and perhaps a README for r2d2 functionality, particularly for new users?

Related: #1874

mehcode:
This is one of the main reasons I want to do #1958 

---

@thsowers See: https://github.com/diesel-rs/r2d2-diesel/blob/master/examples/postgres.rs -- adapting it is simple. You just use `diesel::r2d2` instead of `r2d2`.

kswope:
This is badly needed.  I just spent a few hours figuring out that I have to add r2d2 to the "features" in Cargo.toml.  If I hadn't found some something about it on the forums I would have never figured it out.

Maybe a small section in http://diesel.rs/guides/

vladinator1000:
I started making an example project with Actix but got stuck on this compile error
```
error: the trait bound `i32: diesel::deserialize::FromSql<diesel::sql_types::Uuid, diesel::pg::Pg>` is not satisfied
label: the trait `diesel::deserialize::FromSql<diesel::sql_types::Uuid, diesel::pg::Pg>` is not implemented for `i32`
```

Has anyone seen this before?

I'm trying to compile this code.

```rust
use diesel::prelude::*;

use crate::schema::player;

#[derive(Queryable)]
struct Player {
    id: i32,
    name: String,
}

#[derive(Insertable)]
#[table_name = "player"]
struct NewPlayer<'a> {
    name: &'a str,
}

impl Player {
    fn create(conn: &PgConnection, new_player: NewPlayer) -> Self {
        diesel::insert_into(player::table)
            .values(new_player)
            .get_result(conn)
            .expect("Error creating player")
    }
}

#[cfg(test)]
mod test {
    use super::*;

    #[test]
    fn create_player() {
        let pool = get_pool();
        let conn: &PgConnection = pool.get().unwrap();

        let new_player = NewPlayer { name: "Jon" };

[...]

PR #3123, opened 2022-04-07
Adding postgres example app with connection pooling

chusteven:
This was requested in https://github.com/diesel-rs/diesel/issues/1951 and alongside https://github.com/sgrif/diesel.rs-website/pull/149 I think should be enough for folks to get started.

Tested by running some cargo commands as follows:
```
@ getting_started_with_connection_pooling [stevenchu]$ cargo build
   Compiling diesel_demo_pg_with_conn_pooling v0.1.0 (/Users/stevenchu/personal_projects/rust/diesel/examples/postgres/getting_started_with_connection_pooling)
    Finished dev [unoptimized + debuginfo] target(s) in 1.34s
@ getting_started_with_connection_pooling [stevenchu]$ cargo run --bin show_posts
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `/Users/stevenchu/personal_projects/rust/diesel/target/debug/show_posts`
Displaying 2 posts
Title 1
-----------

Entry 1

Title 2
-----------

Entry 2

@ getting_started_with_connection_pooling [stevenchu]$
```


weiznich:
Thanks for working on this :+1: Unfortunately do I believe that's the wrong way to present this information.

Instead of adding this to the getting started guide I would like to add such an example to the `diesel::r2d2` module documentation. 
The getting started guide is meant for getting users started with a minimal working example, we should try not to put everything in there.

In addition to that, I feel that the example code in this PR could easily lead to cases where users misuse a connection pool. `establish_connection` seems to suggest that you should create a pool and checkout connection once. After that the pool is droped. That's not how pools are normally used. The normal usage would construct  a connection pool once at the start of a long running process and then store the pool in some kind of shared state so that all of the application can repeatably checkout connections from the pool again and again. This safes the overhead of establishing a new connection every time you need a connection. I believe this fact is something that should be mentioned in the corresponding documentation as well.

chusteven:
> Instead of adding this to the getting started guide I would like to add such an example to the `diesel::r2d2` module documentation. The getting started guide is meant for getting users started with a minimal working example, we should try not to put everything in there.

Makes a lot of sense 😊 

> `establish_connection` seems to suggest that you should create a pool and checkout connection once. After that the pool is droped. That's not how pools are normally used.

Ahhhh yeah I forgot about how this works! Automatic memory management for the win :D 

> The normal usage would construct a connection pool once at the start of a long running process and then store the pool in some kind of shared state so that all of the application can repeatably checkout connections from the pool again and again. 

Ahhh makes sense; okay I threw together an example [here](https://gist.github.com/chusteven/1065f1414318a270f7d590e770b4bfcc) -- does it look okay to you? If so I can take a stab at adding it to the docstring (still new to how Rust has code that can run in docstrings but think it's neat and happy to see how it's done :) )

> I believe this fact is something that should be mentioned in the corresponding documentation as well.

👍

chusteven:
> > The normal usage would construct a connection pool once at the start of a long running process and then store the pool in some kind of shared state so that all of the application can repeatably checkout connections from the pool again and again.
> 
> ~Makes sense; okay I threw together an example [here](https://gist.github.com/chusteven/1065f1414318a270f7d590e770b4bfcc) -- does it look okay to you? If so I can take a stab at adding it to the docstring (still new to how Rust has code that can run in docstrings but think it's neat and happy to see how it's done :) )~
> 
> Oh lol just realized that in this gist the calls to the DB will be blocking. Anyway mostly what I was trying to do was pull in an entirely new dependency (like `actix-web` with which I am used to working), but I guess I could emulate that usage with starting a new thread inside every iteration of the for loop.

Okay took another stab and now the example application looks like [this](https://gist.github.com/chusteven/3902810acc31566fca59475b4b9cf540) with output [here](https://gist.github.com/chusteven/18898d5544f12bb5bd61b490a07d8713) suggesting that indeed only 5 connections are getting creating and used at any given moment.

I'll go ahead and discard this PR and open a new one where I'll attempt to port this over into a doc test (seems like I can take [this doctest](https://github.com/diesel-rs/diesel/blob/6a5ab5610322c952408c6e9b91d36d0f8d88f31b/diesel/src/connection/mod.rs#L247-L288) as inspiration)

PR #3126, opened 2022-04-08
Add documentation in `r2d2` module of example application using connection pooling

chusteven:
# Description

Addresses https://github.com/diesel-rs/diesel/issues/1951 by adding a small application as a doctest to the `r2d2` module

In addition to that, I've notably had to do a few things to the `doctest_setup.rs` module:
- Moved all the `DROP TABLE` statements into a `clean_tables()` function
- And all the `CREATE` and `INSERT` ones into a `create_tables_with_data()` one
- Both of which explicitly take `&mut DbConnection` as their single argument

# Testing

`bin/test` passes all the way up until [this test](https://github.com/diesel-rs/diesel/blob/cb9ac5e5329391b780127beea7a1be07590122f0/bin/test#L44) at which point I get the following errors which seem unrelated?
```
+ cd diesel_tests
+ cargo test --features mysql --no-default-features -- --test-threads 1
   Compiling diesel_derives v2.0.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel_derives)
   Compiling diesel v2.0.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel)
   Compiling diesel_migrations v2.0.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel_migrations)
   Compiling diesel_tests v0.1.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel_tests)
error[E0599]: the method `execute` exists for mutable reference `&mut diesel::MysqlConnection`, but its trait bounds were not satisfied
   --> diesel_tests/tests/expressions/ops.rs:209:10
    |
209 |         .execute("INSERT INTO unsigned_table VALUES (1,1), (2,2)")
    |          ^^^^^^^ method cannot be called on `&mut diesel::MysqlConnection` due to unsatisfied trait bounds
    |
   ::: /Users/stevenchu/personal_projects/rust/diesel/diesel/src/mysql/connection/mod.rs:26:1
    |
26  | pub struct MysqlConnection {
    | --------------------------
    | |
    | doesn't satisfy `diesel::MysqlConnection: diesel::RunQueryDsl<_>`
    | doesn't satisfy `diesel::MysqlConnection: diesel::Table`
    |
    = note: the following trait bounds were not satisfied:
            `&mut diesel::MysqlConnection: diesel::Table`
            which is required by `&mut diesel::MysqlConnection: diesel::RunQueryDsl<_>`
            `&&mut diesel::MysqlConnection: diesel::Table`
            which is required by `&&mut diesel::MysqlConnection: diesel::RunQueryDsl<_>`
            `&mut &mut diesel::MysqlConnection: diesel::Table`
            which is required by `&mut &mut diesel::MysqlConnection: diesel::RunQueryDsl<_>`
            `diesel::MysqlConnection: diesel::Table`
            which is required by `diesel::MysqlConnection: diesel::RunQueryDsl<_>`
            `&diesel::MysqlConnection: diesel::Table`
            which is required by `&diesel::MysqlConnection: diesel::RunQueryDsl<_>`
...
```

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -2,6 +2,92 @@
 //!
 //! Note: This module requires enabling the `r2d2` feature
 //!
+//! # Example
+//!
+//! The below snippet is a contrived example emulating a web application,
+//! where one would first initialize the pool in the `main()` function
+//! (at the start of a long-running process). One would then pass this
+//! pool struct around as shared state, which, here, we've emulated using
+//! threads instead of routes.
+//!
+//! ```rust
+//! # include!("doctest_setup.rs");
+//! use diesel::result::Error;
+//! use diesel::pg::PgConnection;
+//! use diesel::prelude::*;
+//! use diesel::r2d2::ConnectionManager;
+//! use diesel::r2d2::Pool;
+//! use std::env;
+//! use std::sync::Arc;
+//! use std::sync::Mutex;
+//! use std::thread;
+//!
+//! pub type PostgresPool = Pool<ConnectionManager<PgConnection>>;
+//!
+//! pub fn get_connection_pool(pool_size: u32) -> PostgresPool {
+//!     let url = database_url_for_env();
+//!     let manager = ConnectionManager::<PgConnection>::new(url);
+//!     // Refer to the `r2d2` documentation for more methods to use
+//!     // when building a connection pool
+//!     Pool::builder()
+//!         .max_size(pool_size)
+//!         .test_on_check_out(true)
+//!         .build(manager)
+//!         .expect("Could not build connection pool")
+//! }
+//!
+//! pub fn create_user(conn: &mut PgConnection, user_name: &str) -> Result<usize, Error> {
+//!     use schema::users::dsl::*;
+//!
+//!     diesel::insert_into(users)
+//!         .values(name.eq(user_name))
+//!         .execute(conn)
+//! }
+//!
+//! pub fn setup_user_table(conn: &mut PgConnection) {
+//!     diesel::sql_query("DROP TABLE IF EXISTS users CASCADE").execute(conn).unwrap();
+//!     diesel::sql_query("CREATE TABLE users (
+//!         id SERIAL PRIMARY KEY,
+//!         name VARCHAR NOT NULL
+//!     )")
+//!         .execute(conn)
+//!         .unwrap();
+//! }
+//!
+//! pub fn delete_user_table(conn: &mut PgConnection) {
+//!     diesel::sql_query("DROP TABLE IF EXISTS users CASCADE").execute(conn).unwrap();
+//! }
+//!
+//! fn main() {
+//!     let pool_size = 1;
+//!     let connection_pool = Arc::new(Mutex::new(get_connection_pool(pool_size)));
+//!     setup_user_table(&mut connection_pool.lock().unwrap().get().unwrap());
+//!
+//!     let mut threads = vec![];
+//!     let max_users_to_create = 1;
+//!
+//!     for i in 0..max_users_to_create {
+//!         let connection_pool = Arc::clone(&connection_pool);
+//!         threads.push(thread::spawn({
+//!             move || {
+//!                 let connection_pool = connection_pool.lock().unwrap();
+//!                 let conn = &mut connection_pool.get().unwrap();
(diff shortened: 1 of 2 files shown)


diesel/src/r2d2.rs
```diff
@@ -2,6 +2,92 @@
...
+//!         .unwrap();
+//! }
+//!
+//! pub fn delete_user_table(conn: &mut PgConnection) {
+//!     diesel::sql_query("DROP TABLE IF EXISTS users CASCADE").execute(conn).unwrap();
+//! }
+//!
+//! fn main() {
+//!     let pool_size = 1;
+//!     let connection_pool = Arc::new(Mutex::new(get_connection_pool(pool_size)));
```
chusteven, line 63: Now that I look at [the documentation](https://docs.diesel.rs/master/r2d2/index.html#structs) (which doesn't put the pool behind a `Mutex` and `Arc`) I'm not sure this is entirely needed and so don't wanna mislead anyone who uses this module

diesel/src/r2d2.rs
```diff
@@ -2,6 +2,89 @@
...
+//!
+//! pub fn setup_user_table(conn: &mut PgConnection) {
+//!     diesel::sql_query("DROP TABLE IF EXISTS users CASCADE").execute(conn).unwrap();
+//!     diesel::sql_query("CREATE TABLE users (
+//!         id SERIAL PRIMARY KEY,
+//!         name VARCHAR NOT NULL
+//!     )")
+//!         .execute(conn)
+//!         .unwrap();
+//! }
```
chusteven, line 53: Request to reviewers: whether or not I should try and emulate [this logic](https://github.com/diesel-rs/diesel/blob/c6dcd6ba116b86758febf5594c03bfed13aadb3c/diesel/src/doctest_setup.rs#L16) -- i.e. that nothing is ever actually committed.

On one hand, I believe that makes a lot of sense; on the other, since we're using a connection pool (even in a contrived example here where there's only 1 connection), I'm not sure how best to do this -- and furthermore, it's only really a problem if tests run in parallel which I don't think they do but I'm not sure if we want to make that assumption (assuming the Rust ecosystem/some crate one day adds support for running tests in parallel)

Another thing I could do is just run raw SQL queries with `diesel::sql_query` and create and entirely new table `users_table_for_r2d2_module_level_doctest` or something

diesel/src/r2d2.rs
```diff
@@ -2,6 +2,89 @@
...
+//! # include!("doctest_setup.rs");
+//! use diesel::result::Error;
+//! use diesel::pg::PgConnection;
+//! use diesel::prelude::*;
+//! use diesel::r2d2::ConnectionManager;
+//! use diesel::r2d2::Pool;
+//! use std::env;
+//! use std::thread;
+//!
+//! pub type PostgresPool = Pool<ConnectionManager<PgConnection>>;
```
weiznich, line 23:
```suggestion
//! pub type DatabasePool = Pool<ConnectionManager<DbConnection>>;
```

+ Defining `type DbConnection = PgConnection/MysqlConnection/…` in `doctest_setup.rs` should allow you to run this test for all backends.

diesel/src/r2d2.rs
weiznich: I would probably choose the following strategy:

1. Move the corresponding sql commands in `doctests_setup.rs` in a distinct `setup_schema` function.
2. Use [`r2d2::Builder::connection_customizer`](https://docs.diesel.rs/master/diesel/r2d2/struct.Builder.html#method.connection_customizer) to provide a callback that gets executed once for each connection in the pool + calls the corresponding setup function (that should not be visible to the user, so place a `#` in front of the corresponding lines)
3. Set the connection pool size to 1 (that should not be visible to the user, so place a `#` in front of the corresponding line)

weiznich (commented):

I've added a few small comments that should hopefully answer your questions. The general direction looks good :+1:

diesel/src/r2d2.rs
chusteven: Neat! Didn't realize (2) was a thing `r2d2` provided but yeah it's definitely useful :D 

Took a stab at the other two -- open to feedback if it wasn't quite how you envisioned it for sure

weiznich (approved):
Perfect, thanks for writing this example :+1:

PR #3134, opened 2022-04-16
Reduced the required dependencies for diesel_dynamic_schema

Razican:
Now that we have the `postgres_backend` and `mysql_backend` features (would it make sense to also have the `sqlite_backend` feature?), I think it should be possible to use these dependencies so that we don't need to compile the system features. This can reduce compilation time of crates that provide their own PostgreSQL and MySQL connections (such as the new async connections).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,7 +287,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --manifest-path diesel_dynamic_schema/Cargo.toml --no-default-features --features "${{ matrix.backend }}"
+          args: --manifest-path diesel_dynamic_schema/Cargo.toml --no-default-features --features "${{ matrix.backend }} diesel/${{ matrix.backend }}"
 
       - name: Run diesel_benches
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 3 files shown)


Razican:
So, the issue is that while the crate itself doesn't need the feature, the tests need it (you need a connection to test Diesel). So, I see two options:
 - Depend on the connections for the full crate (no change with the `master` branch), which increases the number of dependencies
 -  Have a separate `diesel_connection` crate with the default implementations for connections for MySQL, PostgreSQL and SQlite, and then use this crate as a dependency in `diesel` (for `postgres`, `mysql` and `sqlite` features), but not use it for the `*_backend` features. This would allow us to add this dependency as a dev-dependency only. We could even re-export these connections in `diesel` and make it backwards-compatible.

I'm personally more in favour of the second option, and I'm happy to adapt the dependency to do that, but what do you think, @weiznich?

weiznich:
There is a third much simpler solution here: We can just pass the corresponding manually via the command line, like this is done for `diesel_derives` here: https://github.com/diesel-rs/diesel/blob/50c2c85032857a65e4a8d68dc851fa7ab91d65b2/.github/workflows/ci.yml#L234

That would mean we just require that tests are called with `cargo test --features "postgres diesel/postgres"`. You should update `bin/tests` + the ci script in that case.

Razican:
> There is a third much simpler solution here: We can just pass the corresponding manually via the command line, like this is done for `diesel_derives` here:
> 
> https://github.com/diesel-rs/diesel/blob/50c2c85032857a65e4a8d68dc851fa7ab91d65b2/.github/workflows/ci.yml#L234
> 
> That would mean we just require that tests are called with `cargo test --features "postgres diesel/postgres"`. You should update `bin/tests` + the ci script in that case.

Thanks for the suggestion! I did the change. The CI failure seems unrelated, so I would say this is passing the CI :)

weiznich:
CI failures are caused by https://github.com/clap-rs/clap/issues/3648 / https://github.com/clap-rs/clap/pull/3646

tomasz-spoton:
@weiznich should we pin the patch version just in case? 🙈

czotomo:
@weiznich should we pin the patch version just in case? 🙈

weiznich:
@czotomo No need to do that, the clap team already released a fixed version :wink:

Issue #2957, opened 2021-11-14
Do not use the error message to figure out whether the connection was closed


Emm:
Following the discussion in #2935, it appears that Diesel use string matching to figure if the Postgres connection to the server was closed, in `src/pg/connection/result.rs` (`CLOSED_CONNECTION_MSG`)

This is problematic for two reasons, as:
- matching on error messages is inherently fragile and discouraged by Postgres
- Postgres error messages are localized

The correct solution involves using `PQStatus()` on the raw connection and checking whether this returns `CONNECTION_BAD`

weiznich:
Closed by #3135

PR #3135, opened 2022-04-17
Determining closed connections using `PQstatus` rather than matching on message strings

chusteven:
Addresses https://github.com/diesel-rs/diesel/issues/2957

## Description

This PR makes it so that errors reported for the `PgResult` need to match on message strings directly from libpq. Instead we call `PQstatus` directly and attempt to match against the `ConnStatusType::CONNECTION_BAD` variant.

## Testing

`cargo test --no-default-features --features "extras postgres"` passes

### DIFF
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -15,6 +15,12 @@ pub(super) struct RawConnection {
     internal_connection: NonNull<PGconn>,
 }
 
+impl From<&RawConnection> for *const PGconn {
+    fn from(raw_conn: &RawConnection) -> *const PGconn {
+        raw_conn.internal_connection.as_ptr()
+    }
+}
+
 impl RawConnection {
     pub(super) fn establish(database_url: &str) -> ConnectionResult<Self> {
         let connection_string = CString::new(database_url)?;
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -104,7 +104,7 @@ impl RawConnection {
     }
 
     pub(super) fn get_status(&self) -> ConnStatusType {
-        unsafe { PQstatus(self.internal_connection.as_ptr() as *const PGconn) }
+        unsafe { PQstatus(self.internal_connection.as_ptr()) }
     }
 }
 
--- a/diesel/src/pg/connection/raw.rs
+++ b/diesel/src/pg/connection/raw.rs
@@ -15,12 +15,6 @@ pub(super) struct RawConnection {
     internal_connection: NonNull<PGconn>,
 }
 
-impl From<&RawConnection> for *const PGconn {
-    fn from(raw_conn: &RawConnection) -> *const PGconn {
-        raw_conn.internal_connection.as_ptr()
-    }
-}
-
 impl RawConnection {
     pub(super) fn establish(database_url: &str) -> ConnectionResult<Self> {
         let connection_string = CString::new(database_url)?;
@@ -107,6 +101,10 @@ impl RawConnection {
 
     pub(super) fn transaction_status(&self) -> PgTransactionStatus {
         unsafe { PQtransactionStatus(self.internal_connection.as_ptr()) }.into()
+    }
+
+    pub(super) fn get_status(&self) -> ConnStatusType {
+        unsafe { PQstatus(self.internal_connection.as_ptr() as *const PGconn) }
     }
 }
 
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -70,8 +70,8 @@ impl PgResult {
                         _ => DatabaseErrorKind::Unknown,
                     };
                 let error_information = Box::new(PgErrorInformation(internal_result));
-                let pc: *const PGconn = conn.into();
-                if unsafe { PQstatus(pc) } == ConnStatusType::CONNECTION_BAD {
+                let pg_connection: *const PGconn = conn.into();
+                if unsafe { PQstatus(pg_connection) } == ConnStatusType::CONNECTION_BAD {
                     error_kind = DatabaseErrorKind::ClosedConnection;
                 }
                 Err(Error::DatabaseError(error_kind, error_information))
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -7,14 +7,10 @@ use std::os::raw as libc;
 use std::rc::Rc;
 use std::{slice, str};
 
-use super::raw::RawResult;
+use super::raw::{RawConnection, RawResult};
 use super::row::PgRow;
 use crate::result::{DatabaseErrorInformation, DatabaseErrorKind, Error, QueryResult};
 use crate::util::OnceCell;
-
-// Message after a database connection has been unexpectedly closed.
-const CLOSED_CONNECTION_MSG: &str = "server closed the connection unexpectedly\n\t\
-This probably means the server terminated abnormally\n\tbefore or while processing the request.\n";
 
 pub(crate) struct PgResult {
     internal_result: RawResult,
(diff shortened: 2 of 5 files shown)


chusteven:
Note: I read through the context on (in order of most recent to oldest):
- https://github.com/diesel-rs/diesel/pull/2935
- https://github.com/diesel-rs/diesel/pull/2911
- https://github.com/diesel-rs/diesel/pull/2860

And tried to determine why string matching was performed in the first place but couldn't so for full transparency I'm only doing what was suggested in [this comment](https://github.com/diesel-rs/diesel/pull/2935#issuecomment-964431318). I did a little research by trying to pull down the `postgres` repo but didn't get far trying to grok C code 😅 to the point I was particularly confident about what I was reading

https://github.com/postgres/postgres/blob/42dbbca58e8e87e461bb0a4fe48a450e90e1e932/src/interfaces/libpq/fe-misc.c#L753-L764

diesel/src/pg/connection/raw.rs
```diff
@@ -15,6 +15,12 @@ pub(super) struct RawConnection {
     internal_connection: NonNull<PGconn>,
 }
 
+impl From<&RawConnection> for *const PGconn {
+    fn from(raw_conn: &RawConnection) -> *const PGconn {
+        raw_conn.internal_connection.as_ptr()
+    }
+}
+
```
weiznich, line 23: I would prefer using the longer explicit variant here instead of having this impl in place.

diesel/src/pg/connection/result.rs
```diff
@@ -74,14 +70,8 @@ impl PgResult {
...
-                // NOTE: Ideally, we'd remove this check, in favor of matching
-                // on an error code (instead of the more brittle description
-                // string).
-                //
-                // Unfortunately, this particular error is often propagated in
-                // cases - such as network disconnect - without any accompanying
-                // "SqlState", and therefore without an error code..
-                if error_information.message() == CLOSED_CONNECTION_MSG {
+                let pg_connection: *const PGconn = conn.into();
+                if unsafe { PQstatus(pg_connection) } == ConnStatusType::CONNECTION_BAD {
```
weiznich, line 74: We should likely add a method to `RawConnection` to wrap the call to `PQstatus` there, so something like `RawConnection::get_status() -> i32` or so. This has multiple advantages:
* We can only call `PQstatus` with a valid connection object
* We have most connection related unsafe code in one module.

weiznich (commented):
Thanks for working on this. Generally this seems like the correct solution, but I would like to structure the code a bit differently. I've left some comments at the corresponding location.

diesel/src/pg/connection/result.rs
chusteven: Thanks for the review. And this feedback makes a ton of sense. Wasn't quite sure if this next commit addresses things quite the way you envisioned but will push it up now anyway :)

chusteven:
Okay as far as testing that this is exhaustive, I'm still not entirely sure as I think there are a lot of failure cases, but I did add one test, inspired by [this PR](https://github.com/diesel-rs/diesel/pull/2911). Steps to reproduce:

1/ Add following code to this commit:
```diff
@ diesel [stevenchu]$ git diff
diff --git a/diesel/src/pg/connection/mod.rs b/diesel/src/pg/connection/mod.rs
index c1a6cf3c8b..58f26e236d 100644
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -264,6 +264,15 @@ mod tests {
     use crate::sql_types::{Integer, VarChar};
     use std::num::NonZeroU32;

+    #[test]
+    fn wait_a_looooong_time() {
+        let connection = &mut connection();
+
+        crate::sql_query("SELECT pg_sleep(100.0);")
+            .execute(connection)
+            .unwrap();
+    }
+
     #[test]
     fn malformed_sql_query() {
         let connection = &mut connection();
diff --git a/diesel/src/pg/connection/result.rs b/diesel/src/pg/connection/result.rs
index f751b861ad..689d720f36 100644
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -71,8 +71,14 @@ impl PgResult {
                     };
                 let error_information = Box::new(PgErrorInformation(internal_result));
                 let conn_status = conn.get_status();
+                println!("The conn_status object is: {:?}", conn_status);
+                println!(
+                    "And error_information.message() is: {:?}",
+                    error_information.message()
+                );
                 if conn_status == ConnStatusType::CONNECTION_BAD {
                     error_kind = DatabaseErrorKind::ClosedConnection;
+                    println!("Yes we're inside here! {:?}", error_kind);
                 }
                 Err(Error::DatabaseError(error_kind, error_information))
             }
```

2/ Run the test

```
@ diesel [stevenchu]$ cargo test --no-default-features --features "postgres" wait_a_looooong_time
   Compiling diesel v2.0.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel)
    Finished test [unoptimized + debuginfo] target(s) in 2.48s
     Running unittests (/Users/stevenchu/personal_projects/rust/diesel/target/debug/deps/diesel-56765496a50953ec)

running 1 test
...
```

3/ Shell into the postgres DB you are running locally (either on your machine directly or via docker)

4/ Find the `pid`

```
diesel_test=# SELECT pid, usename, application_name, query FROM pg_stat_activity;
  pid  |  usename  | application_name |                                query
-------+-----------+------------------+---------------------------------------------------------------------
   873 |           |                  |
   875 | stevenchu |                  |
 41521 | postgres  |                  | SELECT pg_sleep(100.0);
 41252 | postgres  | psql             | SELECT pid, usename, application_name, query FROM pg_stat_activity;
   871 |           |                  |
   870 |           |                  |
   872 |           |                  |
(7 rows)
```

5/ Kill it

```
diesel_test=# select pg_terminate_backend(41521);
 pg_terminate_backend
----------------------
 t
(1 row)
```

6/ Watch for test output:
```
@ diesel [stevenchu]$ cargo test --no-default-features --features "postgres" wait_a_looooong_time
   Compiling diesel v2.0.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel)
    Finished test [unoptimized + debuginfo] target(s) in 2.48s
     Running unittests (/Users/stevenchu/personal_projects/rust/diesel/target/debug/deps/diesel-56765496a50953ec)

running 1 test
test pg::connection::tests::wait_a_looooong_time ... FAILED

failures:

---- pg::connection::tests::wait_a_looooong_time stdout ----
The conn_status object is: CONNECTION_BAD
And error_information.message() is: "FATAL:  terminating connection due to administrator command\nserver closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n"
Yes we're inside here! ClosedConnection
thread 'pg::connection::tests::wait_a_looooong_time' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(ClosedConnection, "FATAL:  terminating connection due to administrator command\nserver closed the connection unexpectedly\n\tThis probably means the server terminated abnormally\n\tbefore or while processing the request.\n")', diesel/src/pg/connection/mod.rs:273:14


failures:
    pg::connection::tests::wait_a_looooong_time

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 58 filtered out; finished in 7.32s

error: test failed, to rerun pass '--lib'
```

Note that this is a different error message from the one originally being matched against (perhaps because there's a difference when an admin PG user kills a process vs a network disconnect via the Docker example), but either way seems promising since I'd hope that this too would get caught as an unexpected close (and now does!)

https://github.com/diesel-rs/diesel/blob/d285ebd7768330e5abc31abc1d0c3a29f5261e09/diesel/src/pg/connection/result.rs#L15-L17

weiznich (approved):

weiznich:
Thanks for the update + doing the necessary tests. Generally speaking: If you are unsure if something addresses my remarks just push it anyways. It's always good to know how the code looks like to give specific comments + my remarks are sometimes neither clear nor the best solution. That means if you feel that you could implement a better solution just do it :wink:

PR #3137, opened 2022-04-20
Add support for uuid 1.0.0

weiznich:


KodrAus:
Looks like the build failure here is that `Uuid::from_fields` is now infallible: https://docs.rs/uuid/latest/uuid/struct.Uuid.html#method.from_fields

diesel/src/pg/types/uuid.rs
```diff
@@ -33,15 +33,24 @@ fn uuid_to_sql() {
     use crate::query_builder::bind_collector::ByteWrapper;
 
     let mut buffer = Vec::new();
-    let test_uuid = uuid::Uuid::from_fields(0xFFFF_FFFF, 0xFFFF, 0xFFFF, b"abcdef12").unwrap();
+    let bytes = [
+        0xFF_u8, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66,
+        0x31, 0x32,
+    ];
+
+    let test_uuid = uuid::Uuid::from_slice(&bytes).unwrap();
```
chusteven, line 41: Is there any value in continuing to test the `from_fields` method? Looks like that's still supported

chusteven (approved):
lgtm!

diesel/src/pg/types/uuid.rs
weiznich: It is still supported, but the signature changed between with uuid 1.0.0 release. As we support any uuid version since 0.7 I've just used another method to construct the test values so that those tests support all uuid versions supported by diesel as well.

PR #3140, opened 2022-04-21
Add support for slices of refs with BelongingToDsl

valkum:
See https://github.com/diesel-rs/diesel/discussions/3138 for discussions.

Add support for slices of refs with `BelongingToDsl`

To support this usage we need to implement Identifiable for references of references as well, thus we add a blanket impl here.

Furthermore, `BelongsTo` needs to be implemented for `&Parent`.
As blanked implementation for this is only possible with workarounds to avoid
overlapping implementations, we instead add an additional implementation to
the derive macro.

I am currently looking into the recent feedback from the discussion and will update this PR once I am satisfied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,8 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 * Added support for table aliasing via the `alias!` macro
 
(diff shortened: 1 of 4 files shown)


valkum:
Closed in favor of #3141

PR #3141, opened 2022-04-21
Add support for slices of refs with BelongingToDsl

weiznich:
See #3138 for discussions.

We cannot add a blanked impl like
```rust
impl<'a, T> Identifiable for &'a &'a T
where
    &'a T: Identifiable
{
    // …
}
```

here as this results in type inference issues with `.belonging_to` and
`.grouped_by`. Rustc tries to infinitely recurse on `&&&&&…&&& T as
Identifiable` then. As of this I've opted into just generating concrete
impls via `#[derive(Identifiable)]` as this works around this type
inference issues.

Also we cannot use a wild card impl for the added `BelongsTo<&Parent>`
case as this causes a conflicting impl error with our
`BelongsTo` impls for tuples. So again we use concrete impls here.

cc #3140 

@valkum Sorry to open this parallel PR, but I had done some experimentation with the blanked `Identifiable` impl and noticed that the code is basically finished at that state of experimentation. So I've just pushed this version.

### DIFF
--- a/diesel_derives/src/identifiable.rs
+++ b/diesel_derives/src/identifiable.rs
@@ -44,5 +44,15 @@ pub fn derive(item: DeriveInput) -> TokenStream {
                 (#(&self.#field_name),*)
             }
         }
+
+        impl #ref_generics Identifiable for &'ident &'ident #struct_name #ty_generics
+            #where_clause
+        {
+            type Id = (#(&'ident #field_ty),*);
+
+            fn id(self) -> Self::Id {
+                (#(&self.#field_name),*)
+            }
+        }
     })
 }
--- a/diesel_derives/src/identifiable.rs
+++ b/diesel_derives/src/identifiable.rs
@@ -45,7 +45,7 @@ pub fn derive(item: DeriveInput) -> TokenStream {
             }
         }
 
-        impl #ref_generics Identifiable for &'ident &'ident #struct_name #ty_generics
+        impl #ref_generics Identifiable for &'_ &'ident #struct_name #ty_generics
             #where_clause
         {
             type Id = (#(&'ident #field_ty),*);
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -163,6 +163,10 @@ pub fn derive_as_expression(input: TokenStream) -> TokenStream {
 /// This derive implement support for diesel's associations api. Check the
 /// module level documentation of the `diesel::associations` module for details.
 ///
+/// This derive generates the following impls:
+/// * `impl BelongsTo<Parent> for YourType`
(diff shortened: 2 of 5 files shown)


diesel_derives/src/identifiable.rs
```diff
@@ -44,5 +44,15 @@ pub fn derive(item: DeriveInput) -> TokenStream {
                 (#(&self.#field_name),*)
             }
         }
+
+        impl #ref_generics Identifiable for &'ident &'ident #struct_name #ty_generics
```
Ten0, line 48: Looks like this impl is not as general as it could be: the outer reference is unnecessarily constrained.
```suggestion
        impl #ref_generics Identifiable for &'_ &'ident #struct_name #ty_generics
```

diesel_derives/src/lib.rs
```diff
@@ -234,6 +238,10 @@ pub fn derive_from_sql_row(input: TokenStream) -> TokenStream {
 /// Our rules for inferring table names is considered public API.
 /// It will never change without a major version bump.
 ///
+/// This derive generates the following impls:
+/// * `impl Identifiable for &'a YourType`
+/// * `impl Identifiable for &'a &'a YourType`
```
chusteven, line 243: Couple semi-related things:
1/ Does this need to be updated as per previous review's comment?
```suggestion
/// * `impl Identifiable for &'_ &'a YourType`
```

2/ I'm scared of lifetimes but this still seems odd to me; why would one ever need to specify the same lifetime `'a` twice in a row? (Assuming that this constraint _was_ needed?

weiznich: > 2/ I'm scared of lifetimes but this still seems odd to me; why would one ever need to specify the same lifetime 'a twice in a row? (Assuming that this constraint was needed?)

That's a question not that simple to answer. Generally speaking I think different lifetimes for nested references like that one are only relevant if you deal with multiple different reference types at once and one of them is mutable. Otherwise rustc should be able to just use the shortest of all involved lifetimes as "union" of all involved lifetimes. As soon a as mutable reference is involved that becomes tricky, because you cannot hold multiple mutable references to the same object at the same time.

For this specific case: If you have two nested references like that one you can write that as `&'a &'b Type`. From there you can infer that the lifetime `'b` needs to be at least as long as `'a` or to word it differently: `'a` could be a shorter lifetime than `'b'. If now due to some external need `'a` and `'b` are coupled to mutable references this could cause issues if we would require that both are the same lifetime.

EDIT: 

(It the buttons here in the wrong order, so the PR was already merged before I could apply your suggestion :man_facepalming: . Pushed that to 5b9c3bd4c0fb2b63733d2b8baaafce23c8f3ba78)

PR #3145, opened 2022-04-22
Fix links in migration guide

Heliozoa:
Hi,
thanks for your work on diesel.

I noticed the links in the migration docs weren't working properly, fixed in this PR. I also moved the a-elements to the same line as the header, without the change the header itself was cut off when clicking the link.


weiznich:
Thanks :+1:

Issue #3127, opened 2022-04-08
RFE: Allow `LIKE` for binary types


ensc:
### Versions

- **Diesel:**  master
- **Database:**  postgres

## Problem Description

`LIKE` and related operators (https://github.com/diesel-rs/diesel/blob/master/diesel/src/expression_methods/text_expression_methods.rs) can be applied to binary types.  E.g. I have a table where the filenames are of `BYTEA` type:

```
\d file
 fname  | bytea                       |           | not null | 
```

Queries like

```
# select count(*) from file where fname like '%.tgz';
 count  
--------
 107889
```

are working fine and I would like to use them in diesel.

In stable branch, it was possible to implement this manually by something like

```rust
diesel::expression::operators::Like::new(self.as_expression(), other.as_expression())
```

But `expressions::operators` became non-public recently and expression above fails to compile now.

weiznich:
First of all: I do not consider that to be a bug. This should likely be filled as feature request, so a [thread in our discussion forum.](https://github.com/diesel-rs/diesel/discussions/new?category=ideas)  https://github.com/diesel-rs/diesel/discussions/3039 seems to be related.

The corresponding item was never part of the public API, which means we did not guarantee anything there. That written: I can totally understand that this might break existing users, and that you want to have a replacement. So as a workaround you can just use [`infix_operator!()`](https://docs.diesel.rs/master/diesel/macro.infix_operator.html) to define the corresponding operator in your crate. Other than that I might be fine with adding official support for this to diesel, if someone can bring up some documentation for this operator on the postgres side + fills a PR that adds support (+ some tests).

weiznich:
Closed by #3150

PR #3150, opened 2022-04-28
Add like, not_like and concat for PostgreSQL Bytea

Justice4Joffrey:
I saw a couple of issues requestinhg support for this (e.g. #3127).

This is mostly a copy of `TextExpressionMethods`.

Would love to know if there's a more concise way of creating a `Vec<u8>` from a readable literal other than `"S%".as_bytes().to_owned())`.

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -10,7 +10,8 @@ use crate::dsl;
 use crate::expression::grouped::Grouped;
 use crate::expression::operators::{Asc, Desc};
 use crate::expression::{AsExpression, Expression, IntoSql, TypedExpressionType};
-use crate::sql_types::{Array, Inet, Integer, Jsonb, SqlType, Text, VarChar};
+use crate::pg::expression::expression_methods::private::BinaryOrNullableBinary;
+use crate::sql_types::{Array, Binary, Inet, Integer, Jsonb, SqlType, Text, VarChar};
 use crate::EscapeExpressionMethods;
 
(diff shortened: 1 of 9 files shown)


Mingun:
I believe, the shortest way is `b"S%".to_vec()`

weiznich (commented):

The implementation itself looks good. I would like to see a compile test that demonstrates that code using `like`/`concat`/`not_like` with a binary column only compiles using a postgres connection.

Justice4Joffrey:
In the current implementation, all of these pg-specific expression methods will be available for all `T`'s when the "postgres_backend" feature flag is set. Probably works in practice but is clearly incorrect.

E.g.

``` rust
impl<T> PgTextExpressionMethods for T
where
    T: Expression,
    T::SqlType: TextOrNullableText,
{
}
```

weiznich:
That's what I assumed :wink:.

So the correct way to fix this is to introduce postgres specific variants of the corresponding operators. It should be easy to define them via the `diesel::infix_operation!` macro.

Justice4Joffrey:
Ok I see. I think the take-home here is you should never return a type other than those defined in `pg/helper_types.rs`.

IMO it would be good to make that slightly more obvious. All methods in that file specifically use `helper_types` through the crate-level `dsl` module, which is misleading.

weiznich (approved):

weiznich:
Thanks for the update :+1:

Yes that requirement could made more clear. I would definitively accept a PR adding corresponding comments. Otherwise I try to remember that and will add a comment there myself as soon as I touch that part of the code again.

PR #3154, opened 2022-04-29
Allow `DISTINCT ON` to be used with queries containing an explicit sort order

weiznich:
This fixes the issue reported here: https://github.com/diesel-rs/diesel/discussions/3144#discussioncomment-2658922

### DIFF
--- a/diesel/src/pg/query_builder/distinct_on.rs
+++ b/diesel/src/pg/query_builder/distinct_on.rs
@@ -23,6 +23,33 @@ where
     T::SqlType: SingleValue,
 {
 }
+impl<T> ValidOrderingForDistinct<DistinctOnClause<T>> for OrderClause<crate::helper_types::Desc<T>>
+where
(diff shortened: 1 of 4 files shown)


chusteven (approved):
Looks legit -- but re: below comment

> I will have a look at the the simple case with only a single `col.desc()`

I'm curious what else needs be done to support this across multiple columns?

weiznich:
>  I'm curious what else needs be done to support this across multiple columns?

We would somehow need to extend these impls:
https://github.com/diesel-rs/diesel/blob/cfc0d42cf6d4ac7ffb480ffacc56561ae61d02c4/diesel/src/type_impls/tuples.rs#L373-L374

so that they do not just work with a single column but match a whole set of prefixes. Theoretically that might be possible somehow by generating much more concrete impls, practically I've decided that this is not worth the work for now. That's maybe something we can fix if specialization ever becomes stable.

PR #3155, opened 2022-04-29
Fix #3151

### LINKED ISSUE #3151: Invalid SQL when using union on a query that uses order_by or limit
### Versions

- **Rust:** 1.60.0
- **Diesel:** 2.0.0-rc.0
- **Database:** postgresql 13.6, mysql 8.0.28, sqlite 3.35.5
- **Operating System:** Ubuntu 21.10

### Feature Flags

- **diesel:** postgres, mysql, sqlite

## Problem Description
Diesel emits invalid SQL when using `union` on a query that uses `order_by` or `limit` 

```rust
let users = users::table
        .order_by(users::id.asc())
        .limit(1)
        .union(users::table.order_by(users::id.desc()).limit(1))
        .load::<User>(conn)?;
```

Up to 2 entries, with both the lowest id in table and the largest one

 - Postgres 
 > DatabaseError(Unknown, "syntax error at or near \"UNION\"")

 - Mysql: 
 > DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UNION (SELECT `users`.`id`, `users`.`email` FROM `users` ORDER BY `users`.`id` D' at line 1")

 - Sqlite: 
 > DatabaseError(Unknown, "ORDER BY clause should come after UNION not before")

No

### Steps to reproduce

[Reproduction repository](https://github.com/gobanos/diesel-union-bug)

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
 
weiznich:
Allow `ORDER BY` clauses to be used in any combination subquery by
either put parenthesis around the subqueries (Postgres/Mysql) or adding
a wrapping subquery (Sqlite).

### DIFF
--- a/diesel/src/query_builder/combination_clause.rs
+++ b/diesel/src/query_builder/combination_clause.rs
@@ -26,8 +26,8 @@ where
 pub struct CombinationClause<Combinator, Rule, Source, Rhs> {
     combinator: Combinator,
     duplicate_rule: Rule,
-    source: Source,
-    rhs: RhsParenthesisWrapper<Rhs>,
+    source: ParenthesisWrapper<Source>,
+    rhs: ParenthesisWrapper<Rhs>,
 }
 
 impl<Combinator, Rule, Source, Rhs> CombinationClause<Combinator, Rule, Source, Rhs> {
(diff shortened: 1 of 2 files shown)


chusteven (approved):
Nice; code is generic enough to only necessitate a few lines of changes 👍

PR #3159, opened 2022-05-02
Fix #3156

### LINKED ISSUE #3156: Performance regression when getting multiple connections with `r2d2::Pool`
### Versions

- **Rust:** rustc 1.62.0-nightly (e85edd9a8 2022-04-28)
- **Diesel:** v2.0.0-rc0
- **Database:** postgresql
- **Operating System** Linux 5.17.3-arch1-1

### Feature Flags

- **diesel:** ["postgres", "r2d2"]

## Problem Description

As is reported in #3144, performance regression is observed when upgrading `diesel` from `v1.4.8` to `v2.0.0-rc0`. Aggregation queries (e.g. `count(*)` are particularly affected, while the performance hit on normal queries is minimal.

The performance of `v2.0.0-rc0` should be on par with `v1.4.8`.

`cargo bench` output of `table.filter(id.eq(1)).count().get_result()`:

| hash | setup | perf |
|---|---|---|
| [d0bbda8](https://github.com/heyrict/diesel-aggr-sample/tree/d0bbda8a7066cf9b310e3beed2eece6418faeec9) |diesel 1.4 + r2d2| 185,936 ns/iter (+/- 15,282) |
| [599c492](https://github.com/heyrict/diesel-aggr-sample/tree/599c4923f1832fbafaaac33e6df75f6db8815e84) |diesel 2.0 + r2d2| 1,388,410 ns/iter (+/- 158,035) |
| [d3c23ba](https://github.com/heyrict/diesel-aggr-sample/tree/d3c23ba13a3202e59e28e0525a5cc7f0f40f461c) |diesel 1.4 - r2d2| 103,562 ns/iter (+/- 2,997) |
| [58d1210](https://github.com/heyrict/diesel-aggr-sample/tree/58d1210e65ac3ba2dc7e52b5b77853d323c23f9f) |diesel 2.0 - r2d2| 109,238 ns/iter (+/- 5,887) |

### Steps to reproduce

The minimal reproduction sample is [here](https://github.com/heyrict/diesel-aggr-sample).

## Checklist

weiznich:
Our `R2D2Connection::is_broken` implementations where broken in such a
way that they marked any valid connection as broken as soon as it was
checked into the pool again. This resulted in the pool opening new
connections everytime a connection was checked out of the pool, which
obviously removes the possibility of reusing the same connection again
and again. This commit fixes that issue and adds some tests to ensure
that we do not break this again in the future.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -177,11 +177,15 @@ impl crate::r2d2::R2D2Connection for PgConnection {
     }
 
     fn is_broken(&mut self) -> bool {
-        self.transaction_state
-            .status
-            .transaction_depth()
-            .map(|d| d.is_some())
-            .unwrap_or(true)
+        match self.transaction_state.status.transaction_depth() {
+            // all transactions are closed
+            // so we don't consider this connection broken
+            Ok(None) => false,
+            // The transaction manager is in an error state
+            // or contains an open transaction
+            // Therefore we consider this connection broken
+            Err(_) | Ok(Some(_)) => true,
+        }
     }
 }
 
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -180,7 +180,7 @@ impl crate::r2d2::R2D2Connection for PgConnection {
         self.transaction_state
             .status
             .transaction_depth()
-            .map(|d| d.is_none())
+            .map(|d| d.is_some())
             .unwrap_or(true)
     }
 }
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -327,6 +327,28 @@ where
     }
 }
 
+#[derive(QueryId)]
+pub(crate) struct CheckConnectionQuery;
+
+impl<DB> QueryFragment<DB> for CheckConnectionQuery
+where
+    DB: Backend,
+{
+    fn walk_ast<'b>(
+        &'b self,
+        mut pass: crate::query_builder::AstPass<'_, 'b, DB>,
+    ) -> QueryResult<()> {
+        pass.push_sql("SELECT 1");
+        Ok(())
+    }
+}
+
+impl Query for CheckConnectionQuery {
+    type SqlType = crate::sql_types::Integer;
+}
+
+impl<C> RunQueryDsl<C> for CheckConnectionQuery {}
+
 #[cfg(test)]
 mod tests {
     use std::sync::mpsc;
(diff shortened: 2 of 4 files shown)


diesel/src/r2d2.rs
```diff
@@ -394,26 +416,110 @@ mod tests {
...
+            assert_eq!(checkout_count.load(Ordering::Relaxed), 1);
+            std::mem::drop(conn);
+        }
+
+        assert_eq!(acquire_count.load(Ordering::Relaxed), 1);
+        assert_eq!(release_count.load(Ordering::Relaxed), 0);
+        assert_eq!(checkin_count.load(Ordering::Relaxed), 1);
+        assert_eq!(checkout_count.load(Ordering::Relaxed), 1);
+
+        // check that we remove a connection with open transactions from the pool
```
czotomo, line 488: Unrelated question: is there a way to roll back and return the connection to the pool in this scenario?

czotomo (approved):

diesel/src/pg/connection/mod.rs
```diff
@@ -180,7 +180,7 @@ impl crate::r2d2::R2D2Connection for PgConnection {
         self.transaction_state
             .status
             .transaction_depth()
-            .map(|d| d.is_none())
+            .map(|d| d.is_some())
             .unwrap_or(true)
```
chusteven, line 184: Curious how come this couldn't be something like:
```
match self.transaction_state.status {
    TransactionManagerStatus::Valid(_) => false,
    TransactionManagerStatus::InError => true,
}
```

instead?

Ten0: We also want to check the transaction depth - if we failed/forgot to rollback the current transaction (e.g. in case of panic) before releasing to the pool, any new use of the connection would be within a transaction while it would not expect to.

So it looks like that piece of documentation you linked should be the one that gets updated.

Side note: I think this function may be easier to read (mistake would have been harder to make) if it was written in the following form:
```rust
match self.transaction_state.status.transaction_depth() {
    Ok(None) => false,
    Err(_) | Ok(Some(_)) => true,
}
```

chusteven: Ahh I guess I was confused because I thought there may be a legitimate reason for the transaction depth to be non-0 at the point when the connection gets put back. But now that I think about it, that's not likely to happen since the flow would _usually_ be (a) task acquires connection from pool, (b) does all operations including transactions, (c) hopefully transactions complete successfully at which point the transaction depth returns to 0 or, if they don't, hopefully rollbacks reset transaction depth appropriately to 0, (d) release back to pool. But as you say in edge cases such as panics, then yeah the transaction depth not being 0 at point (d) entails that the connection is "dirty" so to speak.

Thanks for the note! And yeah the `match` clause you've provided I think seems quite nice :) Perhaps the documentation can be updated as part of this PR, but if not I'll be happy to do it since I initially wrote it myself 😅

heyrict:
Confirmed that the performance is back to normal (even better than 1.4!). Thank you for looking into this!

| hash | setup | perf |
|---|---|---|
| [d0bbda8](https://github.com/heyrict/diesel-aggr-sample/tree/d0bbda8a7066cf9b310e3beed2eece6418faeec9) |diesel 1.4 + r2d2| 185,936 ns/iter (+/- 15,282) |
| [599c492](https://github.com/heyrict/diesel-aggr-sample/tree/599c4923f1832fbafaaac33e6df75f6db8815e84) |diesel 2.0 + r2d2| 1,388,410 ns/iter (+/- 158,035) |
| [d3c23ba](https://github.com/heyrict/diesel-aggr-sample/tree/d3c23ba13a3202e59e28e0525a5cc7f0f40f461c) |diesel 1.4 - r2d2| 103,562 ns/iter (+/- 2,997) |
| [58d1210](https://github.com/heyrict/diesel-aggr-sample/tree/58d1210e65ac3ba2dc7e52b5b77853d323c23f9f) |diesel 2.0 - r2d2| 109,238 ns/iter (+/- 5,887) |
| [1235a70](https://github.com/heyrict/diesel-aggr-sample/tree/1235a70e472c8a31e548eb5f2006f44f6247587d) |diesel 2.0 (this PR) + r2d2| 124,368 ns/iter (+/- 2,900) |

Ten0 (approved):
Other than what is mentionned at https://github.com/diesel-rs/diesel/pull/3159#discussion_r863139142 LGTM :)
Thanks!

diesel/src/r2d2.rs
weiznich: Depends on what you expect here. Generally speaking the `TransactionManager` trait is nothing that should be used much. I expect that most users use `Connection::transaction` which handles this by just accepting a closure that should be executed inside of the transaction. There we try really hard to roll back things if something goes wrong. This leaves the following options where someone can actually hit this:
* There is an panic inside of the user provided closure of `Connection::transaction` .  Likely that's an user error.
* A user explicitly used the `TransactionManager` API, which allows to create such cases, but also provides functions to rollback/commit transactions. So that's a user mistake there, better drop that connection.
* Something goes wrong with our rollback code in `Connection::transaction`. In this case we cannot do anything with the connection at all, it is likely just broken at a network level.

For the first two cases we could likely try harder to rollback the transaction, but both are on the unhappy path. That means it's for now easier to just mark the connection as broken and tell the pool to get a new one in such cases.

czotomo: Thank you!

PR #3165, opened 2022-05-03
`ipnet-address` feature, providing conversions to/from the `ipnet` crate

smklein:
NOTE: this is one possible prototype of https://github.com/diesel-rs/diesel/discussions/3164

Details:
- This adds an `ipnet-address` feature, which acts exactly like `network-address`, but using the `ipnet` crate instead of `ipnetwork`.
- This copies a fair amount of code that is also in-use by `network-address`, but allows the features to remain functional simultaneously.


weiznich (approved):
Looks fine to me if a changelog entry is added.
To note this somewhere: This likely won't be part of the initial 2.0 release, as it's not part of the RC release.

weiznich:
@smklein Just a remainder that this is only waiting for the Changelog entry to be added.

smklein:
> @smklein Just a remainder that this is only waiting for the Changelog entry to be added.

This is done; added it under "Unreleased", as you mentioned this would not be part of the 2.0 release.

weiznich:
The clippy CI failure seems to be relevant.

smklein:
> The clippy CI failure seems to be relevant.

Okay, I believe this has been fixed - mysql failures still look present, but those also look present on the master branch

weiznich:
> mysql failures still look present, but those also look present on the master branch

Those are caused by an incompatibility between the visual studio version available on github actions and the installed libmysqlclient version. I'm not sure yet how to fix that.

PR #3167, opened 2022-05-04
Optimise build times

weiznich:
This commit replaces some recursive macros with simpler implementations
by generating non recursive where clauses by just forwarding to the next
smaller impl on the corresponding tuple.
According to my measurements this reduces the build time for a `cargo
check --no-default-features --features "sqlite 64-column-table"` build
from ~1min 20s to ~1min 05s.


Ten0:
Maybe relevant: that's weird, it looks like we had the opposite issue with #2480 (Rust issue https://github.com/rust-lang/rust/issues/75542) which we eventually worked around with 4db0e1d75f2fcfdecc733c3979ffd499cd1fa403 which goes the opposite way from this.

weiznich:
Yes that was surprising for me as well. Anyway see #3163 for more context. It seems like some rustc developers are aware of the issue regarding to `SqlType`

PR #3170, opened 2022-05-05
Update documentation for r2d2 to be more complete

chusteven:
Inspired by [this comment](https://github.com/diesel-rs/diesel/pull/3159#discussion_r863139142) I wanted to be sure that the note on error handling (btw, I think we can close https://github.com/diesel-rs/diesel/issues/2963?) was correct

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -78,16 +78,18 @@
 //!
 //! When used inside a pool, if an individual connection becomes
 //! broken (as determined by the [R2D2Connection::is_broken] method)
-//! then `r2d2` will close and return the connection to the DB.
+//! then, when the connection goes out of scope, `r2d2` will close
+//! and return the connection to the DB.
 //!
 //! `diesel` determines broken connections by whether or not the current
 //! thread is panicking or if individual `Connection` structs are
 //! broken (determined by the `is_broken()` method). Generically, these
 //! are left to individual backends to implement themselves.
 //!
-//! For SQLite, PG, and MySQL backends, specifically, `is_broken()`
-//! is determined by whether or not the `TransactionManagerStatus` (as a part
-//! of the `AnsiTransactionManager` struct) is in an `InError` state.
+//! For SQLite, PG, and MySQL backends `is_broken()` is determined
+//! by whether or not the `TransactionManagerStatus` (as a part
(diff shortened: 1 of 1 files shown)


diesel/src/r2d2.rs
```diff
@@ -78,16 +78,18 @@
 //!
 //! When used inside a pool, if an individual connection becomes
 //! broken (as determined by the [R2D2Connection::is_broken] method)
-//! then `r2d2` will close and return the connection to the DB.
+//! then, when the connection goes out of scope, `r2d2` will close
+//! and return the connection to the DB.
```
chusteven, line 82: Actually now that I read the `r2d2` code a little more, I'm not sure this is correct? It looks like [these lines](https://github.com/sfackler/r2d2/blob/c572a8250923f7f157f73b5700e64d21d598ccb6/src/lib.rs#L500-L502) are what's relevant but if I follow the code to the `drop_conns` method, it seems like the default `on_release` is actually a noop, which now has me wondering whether `diesel` really does return the connection to the DB (and how even is that done normally?).

I may be misunderstanding something too though because I can't say I quite understand [these lines](https://github.com/sfackler/r2d2/blob/c572a8250923f7f157f73b5700e64d21d598ccb6/src/lib.rs#L192-L194) either and maybe that's doing what I would expect it to (actually kill the connection, from the perspective of the DB server too).

Anyway thought to take this opportunity to ask some questions 😅

weiznich: This statement is quite correct in my opinion. R2d2 will move the connection out of the pool, that means at some point `Drop` is called for the various wrapper types. This eventually leads to calling `Drop` on the inner `RawConnection` type in diesel. For example that one for sqlite https://github.com/diesel-rs/diesel/blob/0c6219f5bd9641eea09c6fa7dca4995aefda30a0/diesel/src/sqlite/connection/raw.rs#L212 (just picked on of the types, there is a similar impl for postgres/mysql as well). Those `Drop` impls then handle whatever is required to close the database connection.

The corresponding pattern is called [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) and originally was a C++ thing. Rust by Example as [a section](https://doc.rust-lang.org/rust-by-example/scope/raii.html) about this as well.

PR #3175, opened 2022-05-10
Make TransactionDepthChange public

weiznich:
This is kind of required as otherwise custom connection implementations
cannot provide their own transaction manager implementation.


chusteven (approved):
👌

Issue #2489, opened 2020-08-25
Feature Request: MySQL Update on duplicate key


Dolphindalt:
MySQL insert on duplicate key support such as the following query:
```sql
INSERT INTO table (id, title, message, likes) VALUES(1, "A", "bruh", 21) ON DUPLICATE KEY UPDATE    
title="A", message="bruh", likes=21
```
I could take a shot at implementing this feature if it should be a feature of diesel.

weiznich:
A well written PR would be welcome.

starccy:
If I am not mistaken, there is already such a feature:
```rust
diesel::insert_into(some_table)
    .values(value)
    .on_conflict(some_table::id)
    .do_update()
    .set(some_value);
```

weiznich:
@starccy That's only available for sqlite and postgres yet. Mysql did not support this at the time of implementing this feature.

weiznich:
Closed by #3177

PR #3177, opened 2022-05-12
[WIP] Add support for ON CONFLICT for MySQL backend

chusteven:
Addresses https://github.com/diesel-rs/diesel/issues/2489

# Description

Working off the example [here](https://gitter.im/diesel-rs/diesel?at=627cfc65a687887f8d9ff2dc) I shuffled around just a few things (note in the `diesel/src/pg/backend.rs`, I moved the outside `PgOnConflictClaues` struct inside the `sql_dialect` module making it behave more similar to the `ReturningClause` struct)

Otherwise, I'm not entirely sure if I've written the code as it should be but thought maybe it'd be easier to put it up for review and ask for feedback. Some thing I'm uncertain about:
- For testing... I'm thinking it is sufficient to just add to the same doctests ([here](https://github.com/diesel-rs/diesel/blob/42413462fbb67d6b4b612f1f62ba9f8af465d920/diesel/src/upsert/on_conflict_extension.rs#L26), for example) with appropriate `cfg` flags for MySQL specific setup/logic
- What to do about the MySQL "DO NOTHING" action -- on one hand, the [workarounds suggested online](https://stackoverflow.com/questions/4596390/insert-on-duplicate-key-do-nothing) definitely make sense and are somewhat straightforward; on the other hand, I don't know exactly how we would accomplish this given the AST walk (by this point) has no access to any surrounding context needed like column names; I can think of two possible solutions
    - 1/ Make it so this doesn't compile for MySQL; changes would likely be similar to adding backend-specific types like this PR already does for the on conflict clause struct/trait; users who desire this behavior could use the "DO UPDATE" action instead and implement a trivial `id = id` there (I think anyway)
    - 2/ Amend the AST to take in some arbitrary column name and do a noop set as the link above suggests

# Testing

Currently in flux since this is a WIP


diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -90,6 +91,25 @@ where
...
+    for DoUpdate<T>
+where
+    DB: Backend,
+    T: QueryFragment<DB>,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        out.unsafe_to_cache_prepared();
+        if self.changeset.is_noop(out.backend())? {
+            out.push_sql(" DO NOTHING");
+        } else {
```
chusteven, line 105: Wasn't exactly sure what this was for but opted to just copy pasta in case

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -132,8 +152,9 @@ where
     T: Column,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
-        out.push_sql("new.");
+        out.push_sql("VALUES(");
         out.push_identifier(T::NAME)?;
+        out.push_sql(")");
```
chusteven, line 157: I liked what existed previously [as per docs](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html)

> Beginning with MySQL 8.0.19, it is possible to use an alias for the row, with, optionally, one or more of its columns to be inserted, following the VALUES or SET clause, and preceded by the AS keyword. Using the row alias new, the statement shown previously using VALUES() to access the new column values can be written in the form shown here:

```sql
INSERT INTO t1 (a,b,c) VALUES (1,2,3),(4,5,6) AS new
  ON DUPLICATE KEY UPDATE c = new.a+new.b;
```

But just went with the current approach since it was fastest 😅

diesel/src/query_builder/upsert/on_conflict_clause.rs
```diff
@@ -95,7 +95,7 @@ where
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
         self.values.walk_ast(out.reborrow())?;
         out.push_sql(" ON DUPLICATE KEY");
-        self.target.walk_ast(out.reborrow())?;
+        // MySQL doesn't require a target
```
chusteven, line 98: Happy to take out this comment but leaving it for purposes of review now; odd to me but I suppose this keyword "ON DUPLICATE KEY" implicitly required for some kind of primary key- or unique-ness constraint on the table

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -279,25 +277,26 @@ impl<Stmt, Target> IncompleteOnConflict<Stmt, Target> {
     ///     .do_update()
     ///     .set(name.eq("I DONT KNOW ANYMORE"))
     ///     .execute(conn);
+    /// # #[cfg(any(feature = "sqlite", feature = "postgres"))]
     /// assert_eq!(Ok(1), insert_count);
+    /// # #[cfg(feature = "mysql")]
+    /// assert_eq!(Ok(2), insert_count);
```
chusteven, line 283: This was a little funky to me but I found documentation at the MariaDB website [here](https://mariadb.com/kb/en/insert-on-duplicate-key-update/#:~:text=ON%20DUPLICATE%20KEY%20UPDATE%20is,API's%20CLIENT_FOUND_ROWS%20flag%20is%20set.)

> The row/s affected value is reported as 1 if a row is inserted, and 2 if a row is updated, unless the API's CLIENT_FOUND_ROWS flag is set.

A little bit odd to me, so happy to either see if we can set this flag somewhere or perhaps just document this behavior everywhere in the tests?

chusteven (commented):
Amended tests to make the `ON CONFLICT DO UPDATE` case work; still brainstorming the `DO NOTHING` case

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -238,8 +233,11 @@ impl<T: QuerySource, U, Op, Ret, Target>
     /// [`on_conflict`]: crate::query_builder::InsertStatement::on_conflict()
     pub fn do_nothing(self) -> InsertStatement<T, OnConflictValues<U, Target, DoNothing>, Op, Ret> {
         let target = self.target;
-        self.stmt
-            .replace_values(|values| OnConflictValues::new(values, target, DoNothing))
+        self.stmt.replace_values(|values| {
+            println!("values is {:?}", values);
+            println!("target is {:?}", target);
+            OnConflictValues::new(values, target, DoNothing)
+        })
```
chusteven, line 240: Oh whoops didn't mean to commit; was doing this to try and figure out what needed to be done for the `DO NOTHING` case (trying to figure out if we can pass a single column to the downstream AST to do a trivial SET like `ON DUPLICATE KEY UPDATE id=id`

chusteven:
Okay with commit , I can run the affected doctests and will now be poking around to see how I can perhaps add a column to the AST somehow

```
@ diesel [stevenchu]$ cargo test --no-default-features --features "mysql" -- upsert::on_conflict_extension --test-threads 1
   Compiling diesel v2.0.0-rc.0 (/Users/stevenchu/personal_projects/rust/diesel/diesel)
    Finished test [unoptimized + debuginfo] target(s) in 1.79s
     Running unittests (/Users/stevenchu/personal_projects/rust/diesel/target/debug/deps/diesel-04f791a1b8ab85e3)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 29 filtered out; finished in 0.00s

   Doc-tests diesel

running 7 tests
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::IncompleteOnConflict<Stmt, Target>::do_update (line 286) ... ok
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::IncompleteOnConflict<Stmt, Target>::do_update (line 320) ... ok
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::IncompleteOnConflict<Stmt, Target>::do_update (line 354) ... ok
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 153) ... FAILED
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 99) ... FAILED
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 23) ... FAILED
test src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 52) ... FAILED

failures:

---- src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 153) stdout ----
Test executable failed (exit code 101).

stdout:
Hey I'm inside of do_nothing!
values is ValuesClause { values: (Expression(ColumnInsertValue { expr: Bound { item: 2, _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "Sean", _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "brown", _marker: PhantomData }, p: PhantomData })), _marker: PhantomData }
target is ConflictTarget((name, hair_color))

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(1)`,
 right: `Err(DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DO NOTHING' at line 1"))`', src/upsert/on_conflict_extension.rs:44:1


---- src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 99) stdout ----
Test executable failed (exit code 101).

stdout:
Hey I'm inside of do_nothing!
values is ValuesClause { values: (Expression(ColumnInsertValue { expr: Bound { item: 2, _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "Sean", _marker: PhantomData }, p: PhantomData })), _marker: PhantomData }
target is ConflictTarget(name)

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(0)`,
 right: `Err(DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DO NOTHING' at line 1"))`', src/upsert/on_conflict_extension.rs:43:1


---- src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 23) stdout ----
Test executable failed (exit code 101).

stdout:
Hey I'm inside of on_conflict_do_nothing!
values is ValuesClause { values: (Expression(ColumnInsertValue { expr: Bound { item: 1, _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "Sean", _marker: PhantomData }, p: PhantomData })), _marker: PhantomData }

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(1)`,
 right: `Err(DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DO NOTHING' at line 1"))`', src/upsert/on_conflict_extension.rs:19:1


---- src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 52) stdout ----
Test executable failed (exit code 101).

stdout:
Hey I'm inside of on_conflict_do_nothing!
values is BatchInsert { values: [ValuesClause { values: (Expression(ColumnInsertValue { expr: Bound { item: 1, _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "Sean", _marker: PhantomData }, p: PhantomData })), _marker: PhantomData }, ValuesClause { values: (Expression(ColumnInsertValue { expr: Bound { item: 1, _marker: PhantomData }, p: PhantomData }), Expression(ColumnInsertValue { expr: Bound { item: "Sean", _marker: PhantomData }, p: PhantomData })), _marker: PhantomData }], _marker: PhantomData }

stderr:
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Ok(1)`,
 right: `Err(DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DO NOTHING' at line 1"))`', src/upsert/on_conflict_extension.rs:20:1



failures:
    src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 153)
    src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict (line 99)
    src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 23)
    src/upsert/on_conflict_extension.rs - upsert::on_conflict_extension::InsertStatement<T, U, Op, Ret>::on_conflict_do_nothing (line 52)

test result: FAILED. 3 passed; 4 failed; 0 ignored; 0 measured; 156 filtered out; finished in 4.06s

error: test failed, to rerun pass '--doc'
```

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -95,16 +97,34 @@ where
...
     /// #     let conn = &mut establish_connection();
-    /// #     #[cfg(feature = "postgres")]
-    /// #     diesel::sql_query("TRUNCATE TABLE users").execute(conn).unwrap();
-    /// #     #[cfg(feature = "sqlite")]
-    /// #     diesel::sql_query("DELETE FROM users").execute(conn).unwrap();
+    /// #     diesel::sql_query("DROP TABLE users").execute(conn).unwrap();
+    /// #[cfg(any(feature = "sqlite", feature = "postgres"))]
+    /// #     diesel::sql_query("CREATE TABLE users (id SERIAL PRIMARY KEY, name TEXT)").execute(conn).unwrap();
+    /// #[cfg(feature = "mysql")]
+    /// #     diesel::sql_query("CREATE TABLE users (id SERIAL PRIMARY KEY, name VARCHAR(255))").execute(conn).unwrap();
```
chusteven, line 127: Note that this (and the custom doctest setup above) is needed because MySQL doesn't allow unique indexes on `TEXT` columns -- ref [here](https://stackoverflow.com/questions/1827063/mysql-error-key-specification-without-a-key-length) on stackoverflow

diesel/src/backend.rs
```diff
@@ -264,9 +265,23 @@ pub mod sql_dialect {
         /// `QueryFragment` implementations
         pub trait SupportsOnConflictClause {}
 
+        /// Indicates that a backend provides support for `ON CONFLICT` clauses
+        /// using the postgresql `ON CONFLICT` syntax
+        #[derive(Debug, Copy, Clone)]
+        pub struct PgLikeOnConflictClause;
+
+        impl SupportsOnConflictClause for PgLikeOnConflictClause {}
+
```
weiznich, line 274: As this and the `MysqlLikeOnConflictClause` type is backend specific they should be part of the corresponding backend module. If any `QueryFragment` implementation refers to those types, that's an indication that this impl should moved to `backend/query_builder/…` as well.

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -18,11 +18,32 @@ where
...
+
+impl<DB> QueryFragment<DB, sql_dialect::on_conflict_clause::MysqlLikeOnConflictClause>
+    for DoNothing
+where
+    DB: Backend<
+        OnConflictClause = crate::backend::sql_dialect::on_conflict_clause::MysqlLikeOnConflictClause,
+    >,
 {
+    ///
+    /// FIXME: This is super wrong right now and will likely cause a panic; MySQL doesn't support
```
weiznich, line 41: I would go for just not supporting `.do_nothing()` on mysql for now. That's possible by just not providing this `QueryFragment` impl. If you choose that way, it would be great to have a compile tests for this behaviour.

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -18,11 +18,32 @@ where
     }
 }
 
-impl<DB, T> QueryFragment<DB, T> for DoNothing
+impl<DB> QueryFragment<DB, sql_dialect::on_conflict_clause::PgLikeOnConflictClause> for DoNothing
```
weiznich, line 21: This breaks support for `DO NOTHING` clauses on sqlite. We should likely restrict `T` there to some new trait that's only implemented for `PgLikeOnConflictClause` and `SqliteOnConflictClause`.

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -69,6 +91,25 @@ where
     }
 }
 
+impl<DB, T>
+    QueryFragment<DB, crate::backend::sql_dialect::on_conflict_clause::MysqlLikeOnConflictClause>
```
weiznich, line 95: This impl should be moved to `mysql/query_builder/upsert.rs` or something like that, as it's mysql specific.

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -51,10 +72,11 @@ where
     }
 }
 
-impl<DB, T, SP> QueryFragment<DB, SP> for DoUpdate<T>
+impl<DB, T>
+    QueryFragment<DB, crate::backend::sql_dialect::on_conflict_clause::PgLikeOnConflictClause>
```
weiznich, line 76: Again this breaks the sqlite implementation. (Fixing this can be done by restricting `SP` to the same trait as above)

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -89,10 +130,11 @@ where
     }
 }
 
-impl<DB, T, SP> QueryFragment<DB, SP> for Excluded<T>
+impl<DB, T>
+    QueryFragment<DB, crate::backend::sql_dialect::on_conflict_clause::PgLikeOnConflictClause>
```
weiznich, line 134: Again, breaks sqlite.

diesel/src/query_builder/upsert/on_conflict_actions.rs
weiznich: The `AsChangeset` derives allows to ignore fields that are set to `None`. If all fields are set to `None` the corresponding action is empty. I think in this case we might be able to work around that by getting the corresponding primary key from here: https://github.com/diesel-rs/diesel/blob/983209a6188392552efa51fe76284eb3097b012a/diesel/src/upsert/on_conflict_extension.rs#L388 (likely requires changing `T: QuerySource` to `T: Table`)

diesel/src/upsert/on_conflict_extension.rs
weiznich: I think it might be fine to just use `VARCHAR` for the other backends as well.

diesel/src/upsert/on_conflict_extension.rs
weiznich: Technically we should be able to get the corresponding information from the generic type `T`, which is basically the table. That would require changing the bound above to `T: Table` instead of `T: QuerySource`. That would allow us to get the primary key, but that's not that easy because that key could be a composite key. Adding just a bound there to prevent that is not a great idea as that would prevent the usage of composite primary keys for the other backends as well. It's likely possible to work that around by just moving the type information down to the corresponding `QueryFragment` impl and adding the restrictions there to the mysql specific impl. This likely requires including these information somehow into `DoNothing`.

Docs for `Table`: https://docs.diesel.rs/master/diesel/prelude/trait.Table.html

diesel/src/upsert/on_conflict_extension.rs
weiznich: This flag already set here: https://github.com/diesel-rs/diesel/blob/983209a6188392552efa51fe76284eb3097b012a/diesel/src/mysql/connection/url.rs#L121

weiznich (commented):
This already looks quite good. I've left a few comments where things needs some improvements. Otherwise I've added some basic ideas how to solve the `DoNothing` problem. Hopefully that helps, otherwise I will try to write down that a bit more detailed tomorrow or the day after.

diesel/src/upsert/on_conflict_extension.rs
chusteven: 🤔  maybe this isn't set in the doctest setup then; will look again!

chusteven:
> Thanks for working on this 👍 This already looks quite good. I've left a few comments where things needs some improvements. Otherwise I've added some basic ideas how to solve the `DoNothing` problem. Hopefully that helps, otherwise I will try to write down that a bit more detailed tomorrow or the day after.

Thank you! And much appreciated on the inline notes -- I'm still wrapping my head around Rust's type system and traits so having specific pointers is a great help :) As far as the `DoNothing` problem yeah... I think if there's something we _can_ do, I'd like to at least give it a good honest try -- so, for the moment, will keep poking at the problem

diesel/src/backend.rs
chusteven: Yeah I'm definitely still a bit confused on the matter, but I was following [the code](https://github.com/diesel-rs/diesel/blob/42413462fbb67d6b4b612f1f62ba9f8af465d920/diesel/src/query_builder/returning_clause.rs#L30-L54) you had previously pointed me to as an example to work off (though I am not sure if it's the best example as I'm still kind of in just-get-it-to-compile mode 😅 , rather than fully understanding how such generic code is organized)

Where
`ReturningClause = crate::backend::sql_dialect::returning_clause::PgLikeReturningClause`

So what I was trying to do is basically something along the lines of
`OnConflictClause = ...::PgLikeOnConflictClause`

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: Yeah definitely. I think I was/am struggling with both:
- `do_nothing`
- `on_conflict_do_nothing`

as per [this comment](https://github.com/diesel-rs/diesel/pull/3177#issuecomment-1128415843). And was just trying to understand what the generics represented and if it was possible to their existing bounds to get access to what was needed (essentially any arbitrary column) or amend the trait bound without breaking too much downstream to do so.

But yeah if I can't figure out one or the other, then maybe easiest just to save it for later. Will keep digging in any case

diesel/src/upsert/on_conflict_extension.rs
chusteven: This seems very promising! I am less worried about a primary key being a single column or multiple and more about how to grab the actual name of the columns since ultimately that seems to be [the canonical way to "do nothing"](https://stackoverflow.com/questions/4596390/insert-on-duplicate-key-do-nothing); but since `PrimaryKey` looks to be a trait (bound to other traits) I'm not sure how to actually grab a column name (ultimately would be represented by a string, I would think) 😞

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: Yeah... that's I think what I'm trying to do; see my response to [your earlier comment](https://github.com/diesel-rs/diesel/pull/3177/files#r875224280)

chusteven (commented):
Thank you so much for the review! I've come back to this and will keep poking around :)

diesel/src/upsert/on_conflict_extension.rs
chusteven: Oh no... it totally is set in the doctests setup... I think I'm just misunderstanding something

> For [INSERT ... ON DUPLICATE KEY UPDATE](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html) statements, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the CLIENT_FOUND_ROWS flag, the affected-rows value is 1 (not 0) if an existing row is set to its current values.

diesel/src/upsert/on_conflict_extension.rs
weiznich: > but since PrimaryKey looks to be a trait (bound to other traits) I'm not sure how to actually grab a column name (ultimately would be represented by a string, I would think) 😞

The idea there is not to directly grap the column name, but to forward it to an internal helper. Something like the following could work:

```rust
struct MysqlDoNothingHelper<PK>(PK);

// impl for a single column primary key
impl<C: Column + QueryFragment<Mysql>> QueryFragment<Mysql> for MysqDoNothingHelper<C> {
     fn walk_ast<'b>(&'b self, pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
        self.0.walk_ast(out.reborrow())?;
        out.push_sql("=");
        self.0.walk_ast(out)?;
        Ok(())
    }
}

// impl for composite primary key impl with two columns
// (checkout `diesel_for_each_tuple!` to automate these impls for all supported tuple sizes
impl<C1: Column + QueryFragment<Mysql>, C2: Column + QueryFragment<Mysql>> QueryFragment<Mysql> for MysqDoNothingHelper<(C1, C2)> {
     fn walk_ast<'b>(&'b self, pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
        // whatever sql is required for more than one column.
        Ok(())
    }
}
```

diesel/src/query_builder/upsert/on_conflict_actions.rs
weiznich: So what we want here is something like:
```rust 
impl<DB, S> QueryFragment<DB, S> for DoNothing 
where S: PostgresLikeDoNothing
{…}
```
with a newly introduced `PostgresLikeDoNothing` (feel free to choose a better name :wink:) trait, that's implemented for `PgLikeOnConflictClause` and `SqliteLikeOnConflictClause`.

This allows to provide another impl for the mysql case, as long as the mysql `OnConflict` marker type does not implement `PostgresLikeDoNothing`, while the other two backends can share the corresponding implementation.

chusteven: Ugh I couldn't figure out how to do this and was flailing a little too much 😅 . Instead I'm opting to not support `DO NOTHING` on MySQL for now -- I figure any users can workaround the issue by performing an effective noop update if they need to

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: So sorry but I think I'm misunderstanding this. By breaking do you mean failing to compile? Because I at least see that 
```
cargo test --no-default-features --features "extras sqlite"
```

finishes successfully for me 😅

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: I think I see what you're saying but, again, I'm working off [this example](https://github.com/diesel-rs/diesel/blob/42413462fbb67d6b4b612f1f62ba9f8af465d920/diesel/src/query_builder/returning_clause.rs#L30-L54) provided earlier which appears to have `Pg*` traits/structs in a general module

Sorry if I'm misunderstanding though

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: Sigh. Even without modifying `T: QuerySource` to `T: Table`, I think the tricky thing here is related to the difficulties I'm having with getting `DO NOTHING` to work at all. So I've gotten rid of this `if/else` block and added a TODO for the moment

diesel/src/query_builder/upsert/on_conflict_actions.rs
chusteven: Same question re: what is meant by "break" since it still compiles (for me, at least!)

diesel/src/upsert/on_conflict_extension.rs
chusteven: This is 🤩  but I'm afraid I couldn't quite figure out where I would even use this struct. At least with the code I've got so far it's hard for me to see where I would call this code. Like... in a less strongly typed language, I could do something like `if isinstance(backend, mysql)` I think... but here it looks like you envision having yet another query fragment be built conditionally for only a specific backend (which I think, here, is done via marker traits). But I'm having some trouble wrapping my head around how to do that :(

chusteven (commented):
I may have bitten off more than I can chew. I'm pushing up a commit now that I think should at least support MySQL's equivalent of `ON CONFLICT DO UPDATE` but not the `DO NOTHING` clause. And, hopefully, not break Sqlite too 😅 

Thanks for your help and time so far though!

weiznich:
I've tried to allocate some time for this this week, but failed to do that. Hopefully I'm able to add a few more in depth comments next week here.

chusteven:
> I've tried to allocate some time for this this week, but failed to do that. Hopefully I'm able to add a few more in depth comments next week here.

Thank you for looking :) But if you can't find the time or if the feedback is too involved, I'm happy to just let this PR sit for a while until I can wrap my head around generic code/traits a little more

weiznich:
@chusteven I think I might find some time to finally drive that PR to completion. Let me know if you are interested to continue to work on this and we will figure out something to talk about the problems described above.

chusteven:
> @chusteven I think I might find some time to finally drive that PR to completion. Let me know if you are interested to continue to work on this and we will figure out something to talk about the problems described above.

Yes that would be fantastic! I will look again at the discussion to try and remind myself of where I was and then look forward to hearing back from you as well. Fortunately looks like the code itself hasn't diverged too far from the main branch :)

weiznich:
As far as I remember the main problem here was to explain how `QueryFragment` works and how certain impls can be specialized for specific backends/dialect, right?

I've sat down and wrote some minimal playground examples trying to explain that, as most of that stuff is quite unique to diesel. (Also I'm not sure if the specialization pattern we use there, is even documented somewhere…) The main motivation of that system is to have as much information as possible available at compile time. This enables diesel to check whether a certain expression is supported by a specific backend (== there is an impl) and also provides a really good performance, as most of that stuff is optimized away at compile time (due to the massive usage of zero sized structs, which cannot appear in the compilation result).

A simplistic version of `QueryFragment` would look something like [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=30dbb8393df86400b1d5f349f611436a). This already demonstrates how to provide implementations for specific backend. That's just done by using an concrete backend type for the first generic parameter. This approach works ok, but it does not allow other backends to reuse the same implementation easilyj

Now it would be possible to say, fine, let's just write a generic `QueryFragment` impl for this and restrict the number of allowed backends via another (marker) trait. That works, as demonstrated in [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7f3f60cfcb91b850f726b4ba333d2065). In fact that's how diesel 1.x was implemented. Now this approach has another problem: You can have only one generic impl, as any other would conflict with the first impl as there could be a potential backend type that implements both support traits at the same time. Such a type wouldn't be meaningful in the context of diesel, but rustc cannot know that. 

This raises the question: How to solve that problem. There is the unstable (and unsound…) `specialization` nightly feature. My answer involves associated types, as a associated type can only have one specific type as "value". This requires introducing another generic parameter on the `QueryFragment` trait that uses the associated type as argument. Combined with defaults for generic types this allows some limited form of specializations. [This playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a936ab3939972916d5224feb5aaa3ef3) contains an adjusted version of the example above.

Hopefully that clears up how the general specialization works. For the "on conflict" impl we now must adjust the corresponding generic types in `SqlDialect` + possibly add some additional generic impls of `QueryFragment` in the right place.

chusteven:
> As far as I remember the main problem here was to explain how QueryFragment works and how certain impls can be specialized for specific backends/dialect, right?

Yeah I believe so. I was generally struggling because I haven't worked too much with interfaces and traits before `diesel` so I had been largely trying to work off existing code and copy/paste Postgres -> for MySQL. But your playground links and writeup look to be really helpful so I'm going to return to this when I get a bit of time and try and take a more systematic approach.

Thanks for your response! I will hope to have some more commits or at least continue the discussion within a week or two :)

chusteven:
Your playground links helped me greatly in understanding how diesel works generically at a very high level. But I am still having trouble wrapping my head around this particular issue.

The thing that is giving me the most difficulty is exactly what you said below

> Hopefully that clears up how the general specialization works. For the "on conflict" impl we now must adjust the corresponding generic types in SqlDialect + possibly add some additional generic impls of QueryFragment in the right place.

For some specifics, here are the things that I am thinking:
- [This file](https://github.com/diesel-rs/diesel/blob/master/diesel/src/query_builder/upsert/on_conflict_clause.rs) looks to be the "generic impls" as you mention
- But since the MySQL syntax is slightly different -- instead of "ON CONFLICT" it's "ON DUPLICATE KEY" -- I'm still trying to figure out how to have a default implementation **and** provide a Mysql-specific one in the `../mysql/query_builder/...` directory; I believe my first attempt was following in the lines of the `PgLikeReturningClause` to achieve as much, though I got the sense that [this wasn't necessarily desirable](https://github.com/diesel-rs/diesel/pull/3177/files#r875224280)
- An even tougher problem for me to understand is that MySQL additionally does not make use of the `Target` parameter

All that to say, I will continue playing around with the issue but don't expect to drive this to completion as I just don't have a solid handle on the matter now :/ So if this feature is something you would like to release soon, then I'm happy to close this PR or let someone else take over the branch. Otherwise, we can keep it open and I will continue to chip away at it when I find the time :)

weiznich:
I've pushed an update here that mostly fixes the implementation and all doc tests. Before merging I want to add a few additional tests + some compile fail tests to see whether certain invariants hold for all supported backends.

diesel/src/backend.rs
```diff
@@ -345,6 +345,13 @@ pub(crate) mod sql_dialect {
         /// `QueryFragment` implementations
         pub trait SupportsOnConflictClause {}
 
+        /// A marker trait indicating if a `ON CONFLICT (...) DO UPDATE ... [WHERE ...]` clause is supported or not
+        pub trait SupportsOnConflictClauseWhere {}
+
+        /// A marker trait indicating whether the on conflict clause implementation
+        /// is mostly like postgresql
+        pub trait PgLikeOnConflictClause: SupportsOnConflictClause {}
```
chusteven, line 353: Does this syntax with the `:` character mean that `PgLikeOnConflictClause` is the same as `SupportsOnConflictClause`? Like a marker trait or something?

diesel/src/mysql/query_builder/query_fragment_impls.rs
```diff
@@ -63,3 +69,71 @@ where
         Ok(())
     }
 }
+
+impl<T> QueryFragment<Mysql, crate::mysql::backend::MysqlOnConflictClause> for DoNothing<T>
+where
+    T: Table,
+    T::PrimaryKey: Column,
```
chusteven, line 76: Potentially unrelated question: the fact that we are restricting this implementation to just tables that have primary keys (at least that's what I think this is doing), does that mean that `diesel` doesn't require tables to have primary keys?

diesel/src/mysql/query_builder/query_fragment_impls.rs
```diff
@@ -63,3 +69,71 @@ where
...
+impl<T, Tab> QueryFragment<Mysql, crate::mysql::backend::MysqlOnConflictClause> for DoUpdate<T, Tab>
+where
+    T: QueryFragment<Mysql>,
+    Tab: Table,
+    Tab::PrimaryKey: Column,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Mysql>) -> QueryResult<()> {
+        out.unsafe_to_cache_prepared();
+        out.push_sql(" UPDATE ");
+        if self.changeset.is_noop(out.backend())? {
```
chusteven, line 96: Neat! I can't remember why it wasn't something I thought we could do back then: https://github.com/diesel-rs/diesel/blob/e01a9beb2c111d42c46a4343b6444b69db281cda/diesel/src/query_builder/upsert/on_conflict_actions.rs#L83-L85

diesel/src/mysql/query_builder/query_fragment_impls.rs
```diff
@@ -63,3 +69,71 @@ where
...
+    Target: QueryFragment<Mysql>,
+    Action: QueryFragment<Mysql>,
+    NoWhereClause: QueryFragment<Mysql>,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Mysql>) -> QueryResult<()> {
+        self.values.walk_ast(out.reborrow())?;
+        out.push_sql(" ON DUPLICATE KEY");
+        self.target.walk_ast(out.reborrow())?;
+        self.action.walk_ast(out.reborrow())?;
+        self.where_clause.walk_ast(out)?;
```
chusteven, line 120: Because this implementation is marked with `NoWhereClause` means that this like effectively does nothing?

diesel/src/mysql/query_builder/query_fragment_impls.rs
```diff
@@ -63,3 +69,71 @@ where
...
+where
+    Values: QueryFragment<Mysql>,
+    Target: QueryFragment<Mysql>,
+    Action: QueryFragment<Mysql>,
+    NoWhereClause: QueryFragment<Mysql>,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Mysql>) -> QueryResult<()> {
+        self.values.walk_ast(out.reborrow())?;
+        out.push_sql(" ON DUPLICATE KEY");
+        self.target.walk_ast(out.reborrow())?;
```
chusteven, line 118: Oh neat! Is the target always going to be the `DuplicatedKeys` as a marker trait b/c MySQL doesn't actually need targets?

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -1,14 +1,23 @@
...
+use crate::backend::Backend;
 use crate::expression::{AppearsOnTable, Expression};
 use crate::query_builder::*;
 use crate::query_source::*;
 use crate::result::QueryResult;
 
 #[doc(hidden)]
 #[derive(Debug, Clone, Copy, QueryId)]
-pub struct DoNothing;
+pub struct DoNothing<T>(PhantomData<T>);
```
chusteven, line 12: What's the need for the `PhantomData` here?

diesel/src/query_builder/upsert/on_conflict_actions.rs
```diff
@@ -1,14 +1,23 @@
...
+
+impl<T> DoNothing<T> {
+    pub(crate) fn new() -> Self {
+        Self(PhantomData)
+    }
+}
 
-impl<DB> QueryFragment<DB> for DoNothing
+impl<T, DB> QueryFragment<DB> for DoNothing<T>
```
chusteven, line 20: For this and the rest of the implementations in this file... why the need to introduce the new generic parameter(s)?

diesel/src/upsert/mod.rs
```diff
@@ -4,6 +4,7 @@
 //!
 //! * PostgreSQL version 9.5 or newer
 //! * Sqlite3 version 3.24.0 or newer
+//! * MySQL version 8.0 or newer
```
chusteven, line 7: It appears as though this is supported on MySQL 5.7 as well! https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html

chusteven (commented):
> I've pushed an update here that mostly fixes the implementation and all doc tests. Before merging I want to add a few additional tests + some compile fail tests to see whether certain invariants hold for all supported backends.

Thank you so much for taking the time to work on this. The code changes are still tough for me to follow exactly because I generally struggle with generics still :/ but at very least, it can serve as a basis for a working feature I understand from the perspective of the client at least.

Looking at the tests now as I think they will also help me wrap my head around how this will work. Thank you so much again -- I use MySQL at work so will be really thrilled to get to see this in action!

diesel/src/mysql/query_builder/query_fragment_impls.rs
weiznich: Diesel requires tables to have primary keys. 
In fact this is just an "artificial" restriction because rustc cannot see that the `PrimaryKey` associated type of the `Table` trait implements `Column`

diesel/src/mysql/query_builder/query_fragment_impls.rs
weiznich: Yes, although I prefer to call that function nevertheless for the so that if we ever add something in that `QueryFragment` impl we get it here as well.

diesel/src/mysql/query_builder/query_fragment_impls.rs
weiznich: Yes, the target will always be `DuplicatedKeys`

diesel/src/query_builder/upsert/on_conflict_actions.rs
weiznich: Because we need to know the table type in the `QueryFragment` impl for `DoNothing (and `DoUpdate`) to emit the `id = id` workaround there. (Specifically we need to know the name of the primary key there.)

chusteven: Thank you this was really helpful!

weiznich (approved):

diesel/src/lib.rs
```diff
@@ -547,3 +547,99 @@ pub use crate::result::Error::NotFound;
...
+//         query.execute(&mut conn).unwrap();
+
+//         let user = users::table
+//             .find(42)
+//             .select(users::name)
+//             .get_result::<String>(&mut conn)
+//             .unwrap();
+//         assert_eq!(user, "Other John");
+//     }
+// }
```
JeremyLARDENOIS, line 645: Hello, is there a reason why this test is commented on?

weiznich: Good catch. That's a test that I locally used for faster turn around times. It's not supposed to be there at all. I've removed it from the PR.

PR #3178, opened 2022-05-14
Updated ipnetwork to 0.19

Razican:


Razican:
It seems that the issue is the MSRV, that the new ipnetwork crate requires the 2021 edition (1.56), while the current MSRV is 1.54.

An option would be to update the MSRV to 1.56 (which also includes nice stuff, such as setting the MSRV directly in Cargo.toml). This would also bring the opportunity to use the 2021 edition in Clippy.

weiznich:
This does not require a MSRV bump according to our MSRV policy: 

> For any named minimal supported Rust version we guarantee that it is possible to build Diesel with the 
default features enabled using some set of dependencies. Those set of dependencies is not necessarily 
an up to date version of the specific dependency. We check this by using the unstable `-Z minimal-version` cargo flag. 

(That's because older ipnetwork versions work just fine with the current MSRV.)

JohnTitor (approved):
LGTM with or without nit fix

Issue #2535, opened 2020-10-09
SupportsReturningClause mysql


zhiburt:
## Problem Description

Following the get started guide step by step I was willing to run insert code with `mysql`. 

```rust
        // ...
        let inserted = diesel::insert_into(entities::table)
            .values(&ext)
            .get_result::<Entity>(&conn)
            .unwrap();
```

But I was constantly getting a compile error.

```log
the trait bound `diesel::mysql::Mysql: diesel::backend::SupportsReturningClause` is not satisfied
required because of the requirements on the impl of `diesel::query_builder::QueryFragment<diesel::mysql::Mysql>` for `diesel::query_builder::returning_clause::ReturningClause<schema::extension::columns::id>`
required because of the requirements on the impl of `diesel::query_builder::QueryFragment<diesel::mysql::Mysql>` for 
...
```

And eventually (after figuring out that `execute` doesn't return id :smile: ) I took a look at the doc and it seems to me currently it's impossible to run such code with `mysql` since its supported only by `postgress` right?

https://docs.diesel.rs/diesel/backend/trait.SupportsReturningClause.html

### What are you trying to accomplish?

If it's true I would consider adding some information somewhere that it's not work with mysql could reduce peoples time spent on trying to resolve it. Probably in get_result doc comment.

### Comment

I may be incorrect in my assumptions in which case I am sorry :disappointed: 

And thank you for the crate.

thecooldrop:
I would like to take this issue as first contribution to Diesel. 

May I ask you @weiznich to assign it to me?

weiznich:
@thecooldrop Feel free to submit a PR for this. I should probably add that I don't think that `get_result` is the right place for this as `get_result` is available for queries using the `mysql` backend, just not for insert statements.

thecooldrop:
@weiznich Your comment is really confusing me since according to my analysis, given that `MySql` does not implement `SupportsReturningClause` we should not be able to have get_results for any of the statement types with MySql backend. Here is why:

1. `RunQueryDsl::get_result<U>(self, conn: &Conn ) -> QueryResult<U>` is defined only for those implementors of the trait which also implement the `LoadQuery<Conn,U>`
2. `LoadQuery<Conn,U>` is implemented only for those types which implement `AsQuery` whose associated `AsQuery::Query` associated type implements the `QueryFragment<DB: Backend>`.
[...]

Issue #1882, opened 2018-10-12
Is MariaDB supported? 


johnkarasev:
Just wondering if diesel supports MariaDB.

weiznich:
`MariaDB` is more or less supported through the `MySql` backend. 

In the long term we want to add a distinct `MariaDB` backend that  shares most of it implementation with the existing `MySql` backend. 

If anyone is willing working on this feel free to claim that issue or drop a message on gitter.

ian-p-cooke:
@virome are you still working on this?  I might want to do something similar to support TiDB and would like to see how you refactor things.

virome:
@ian-p-cooke Unfortunately, I ended up not having much time to work on this. If no one else wants to take it, I still would like to finish it at some point.
If you're ready to go with your work, I wouldn't wait for me.

p-alik:
I used a workaround to build diesel with mariadb
https://stackoverflow.com/questions/61505861/crate-with-diesel-and-mysql-dependencies-no-longer-compiles-against-mariadb-afte

saresend:
Is this still available? I'd be interested to give it a go

weiznich:
@saresend This is still an open issue, so feel free to work on this. Be warned that this requires likely a complex large change to diesel itself.

saresend:
Yeah that's fine, just want to double check you'd be willing to help answer questions / support this work

weiznich:
@tgross35 Let's move the discussion about a custom mariadb backend here:

> Thanks for the response, seems like there's a couple reasons for a dedicated mariadb backend. What would be the bare minimum PR to get support moving in the right direction? It seems like duplicating diesel/diesel/src/mysql and updating the connection would be needed, but I'm not sure how many of the other sections of the repo would need to be touched just to get something going, or what a prioritized todo list would look like.

It's hard to tell what's the exact minimum PR for something like that, because as already written: We do not even have a specific design for that in mind. Keep that in mind for anything else that follows. I would prefer something that does not duplicate that much code, as that all is code that needs to be maintained. I would guess that there is no need to duplicate all of `diesel/diesel/src/mysql`. For example the connection/serialization/deserialization implementation can and should be shared between a potential mariadb backend and the existing mysql backend. I would likely start with the following approach:

[...]

PR #3186, opened 2022-05-24
Allow mysql schemas that contain either a json or a datatime field

weiznich:
Unfortunately we cannot add a test for json, as the behaviour there
diverges between mariadb and mysql.


diesel_cli/src/print_schema.rs
```diff
@@ -144,6 +144,8 @@ fn mysql_diesel_types() -> HashSet<&'static str> {
 
     types.insert("TinyInt");
     types.insert("Tinyint");
+    types.insert("Datetime");
```
JohnTitor, line 147: Does it make sense to add `TIMESTAMP` along with it?

chusteven:
> Unfortunately we cannot add a test for json, as the behaviour there diverges between mariadb and mysql.

Curious what `diesel`'s stance on MySQL vs. MariaDB; I'm guessing `diesel` supports both?

weiznich:
> Curious what diesel's stance on MySQL vs. MariaDB; I'm guessing diesel supports both?

It's more like we officially support MySQL. MariaDB just works due to the fact that it is mostly the same thing. (In fact I personally use MariaDB locally for testing as this is just that what is provided by my distro.)
In the long term we would like to have a separate backend for MariaDB (https://github.com/diesel-rs/diesel/issues/1882) as this would allow us to support some SQL features that are not supported for MySQL yet without causing issues (`RETURNING` support https://github.com/diesel-rs/diesel/issues/2535)

diesel_cli/src/print_schema.rs
weiznich: It's already there: https://github.com/diesel-rs/diesel/blob/b51e6584ac4c6dc4495010e6e2f282c6de327e1f/diesel_cli/src/print_schema.rs#L79

PR #3190, opened 2022-05-26
Make rustdoc JSON not cause ICE

Enselic:
To tro to help figure out what is causing https://github.com/rust-lang/rust/issues/93588

### DIFF
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -332,6 +332,15 @@ pub trait AsQuery {
     fn as_query(self) -> Self::Query;
 }
 
+impl<T: Query> AsQuery for T {
+    type SqlType = <T as Query>::SqlType;
+    type Query = T;
+
+    fn as_query(self) -> <T as AsQuery>::Query {
+        self
+    }
+}
+
 /// Takes a query `QueryFragment` expression as an argument and returns a type
 /// that implements `fmt::Display` and `fmt::Debug` to show the query.
 ///
--- a/diesel/src/query_builder/mod.rs
+++ b/diesel/src/query_builder/mod.rs
@@ -332,15 +332,6 @@ pub trait AsQuery {
     fn as_query(self) -> Self::Query;
 }
 
-impl<T: Query> AsQuery for T {
-    type SqlType = <Self as Query>::SqlType;
-    type Query = Self;
-
-    fn as_query(self) -> Self::Query {
-        self
-    }
-}
-
 /// Takes a query `QueryFragment` expression as an argument and returns a type
 /// that implements `fmt::Display` and `fmt::Debug` to show the query.
 ///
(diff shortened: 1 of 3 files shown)


diesel/src/query_builder/mod.rs
```diff
@@ -332,15 +332,6 @@ pub trait AsQuery {
...
 
-impl<T: Query> AsQuery for T {
-    type SqlType = <Self as Query>::SqlType;
-    type Query = Self;
-
-    fn as_query(self) -> Self::Query {
-        self
-    }
-}
-
```
weiznich, line 343: Thanks for working on this ❤️ 

This change would unfortunately break most of diesel's query builder. As a blind guess I would expect that this is also the main fix for the ICE, as this impl implies quite a lot of other concrete impls. (Just have a look what else implements `AsQuery` or `Query` to get a feeling for that). If I can help with concrete information here just reach out.

Enselic: Thanks for taking a look! I was just being a bit lazy. I pushed a commit where I did the same change as I did for `AsExpression`, namely replacing `Self` with `T`, and now the diff should look less problematic.

Enselic: @weiznich If you are familiar with the diesel code base, the best way you can help is probably to remove more and more code from diesel/diesel until the ICE stops from happening. Do it "one line at a time". Then we will know what line that is crucial to making the ICE happen. I have tried this myself but I found the diesel code base to be riddled with dependencies spread across the code base via macros and such, and I wasn't tenacious enough to follow through. But finding a minimal way to reproduce the problem we see would certainly help. That is what https://github.com/rust-lang/rust/issues/93588#issuecomment-1139124795 was talking about too.

Once have such an [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example) it will probably not be hard to fix the root problem.

Enselic:
Sorry! I realized just now that I accidentally created a PR in this repo instead of my fork.

I created a PR in my fork now: https://github.com/Enselic/diesel/pull/1, so please refer to that instead.

weiznich:
To be sure: If it's as simple as changing these three impls to prevent this ICE from happening I would be totally fine with just changing diesel here.

Enselic:
If you asses that to be a good near-term solution, feel free to merge this code. That would be totally fine with me. However, I think at least some other crates in the diesel workspace still causes ICE when generating rustdoc JSON, but maybe it is just as easy to fix them as it is to fix the main crate.

Long-term, rustdoc of course needs to fix the ICE for diesel and other crates that has the same code structure. It should't be necessary change client code in order to make rustdoc JSON work.

BTW, I am very curious, so I hope you don't mind me asking: What do you use the diesel rustdoc JSON for?

weiznich:
> What do you use the diesel rustdoc JSON for?

I've played around with something like your [`cargo-public-api`](https://github.com/Enselic/cargo-public-api#diff-the-public-api) tool, especially the diff mode. That's nothing that worked at the time due to the ICE issues.

PR #3195, opened 2022-06-02
Put the `backend::sql_dialect` module behind the

weiznich:
`i-implement-a-third-party-backend-and-opt-into-breaking-changes`

This will allow us to change some of the associated types on existing
`SqlDialect` impls in later diesel versions. We might want to do this
for the following reasons:

* The backend gained support for previously unsupported SQL operations
* The backend fixed/introduced a bug that requires special handling
* We got some edge case wrong with sharing the implementation between
backends

Additionally this commit fixes a few types in type and module names in
the `sql_dialect` module.


chusteven (commented):
Looks like CI failed behind some public/private stuff?

JohnTitor (approved):
Makes sense to me

PR #3196, opened 2022-06-03
Use windows 2019 instead of windows latest

Sytten:
Lets see if it helps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         rust: ["stable", "beta", "nightly"]
         backend: ["postgres", "sqlite", "mysql"]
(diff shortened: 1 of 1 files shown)


weiznich:
https://github.com/diesel-rs/diesel/commit/d9d46ad11b2d37c4ee43b123a77241c2a06a584d might be relevant for fixing the sqlite build failure.

Sytten:
It seems like the issue is the lib command

weiznich:
Yes, the linked commit fixed that for the windows 2022 image. I think we just need to revert that one to make it work with the older image again.

Sytten:
Reverting the fix didn't seem to help

Sytten:
@weiznich seems to work!

weiznich:
Thanks ❤️

PR #3197, opened 2022-06-07
Explictly sort migrations on migration redo to ensure that we revert

weiznich:
them from newest to oldest, but apply them in the revers order


pksunkara (approved):

czotomo (approved):

JohnTitor (approved):
LGTM

PR #3202, opened 2022-06-10
Fix #1535

### LINKED ISSUE #1535: sqlite3 foreign keys which don't name columns can't be inferred
### Versions

- **Rust:** `rustc 1.25.0-nightly (616b66dca 2018-02-02)`
- **Diesel:** `diesel 1.1.0`
- **Database:** `sqlite3 3.16.0 2016-11-04 19:09:39 0e5ffd9123d6d2d2b8f3701e8a73cc98a3a7ff5f`
- **Operating System** macOS 10.12.6 (16G1114)

### Feature Flags

- **diesel:** `sqlite postgres`

## Problem Description

Running `diesel print-schema` fails, with very terse output:

```
$ RUST_BACKTRACE=full RUST_LOG=trace diesel print-schema
Unexpected null for non-null column
```

### Steps to reproduce

```
sqlite> .schema accounts
CREATE TABLE accounts (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  account TEXT NOT NULL,
  data_center_id INTEGER NOT NULL,
  auth_key BLOB NOT NULL,
  UNIQUE (account, data_center_id),
  FOREIGN KEY (data_center_id) REFERENCES data_centers
);
sqlite> .schema data_centers
CREATE TABLE data_centers (
  id INTEGER PRIMARY KEY
);
```

This produces: 

```
sqlite> PRAGMA FOREIGN_KEY_LIST('accounts');
id          seq         table         from            to          on_update   on_delete   match
----------  ----------  ------------  --------------  ----------  ----------  ----------  ----------
0           0           data_centers  data_center_id  -NULL-      NO ACTION   NO ACTION   NONE
```

`to` is `NULL` because `REFERENCES data_centers` doesn't name any columns; it's implicitly the same as `REFERENCES data_centers(data_center_id)`. I'm pretty sure this is due to the `pragma_foreign_key_list` pseudo-table doesn't expect a `NULL`.

https://github.com/diesel-rs/diesel/blob/07319d363aff713dbe52914aa2f512431f559864/diesel_infer_schema/infer_schema_internals/src/sqlite.rs#L27-L38

## Checklist

weiznich:
For implicit foreign key constraint on sqlite no explicit primary key is given.
This means `PRAGMA FOREIGN_KEY_LIST` will return a null value instead.
For this cases we just look up the primary key of the referenced table
explicitly.

### DIFF
--- a/diesel_cli/src/infer_schema_internals/sqlite.rs
+++ b/diesel_cli/src/infer_schema_internals/sqlite.rs
@@ -31,7 +31,7 @@ table! {
         seq -> Integer,
         _table -> VarChar,
         from -> VarChar,
-        to -> VarChar,
+        to -> Nullable<VarChar>,
         on_update -> VarChar,
         on_delete -> VarChar,
         _match -> VarChar,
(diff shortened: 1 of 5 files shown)


chusteven (approved):
Lgtm. But one question: when the reference is explicit, diesel doesn't have this problem -- is that right?
```
FOREIGN KEY (data_center_id) REFERENCES data_centers(id)
```

weiznich:
>  But one question: when the reference is explicit, diesel doesn't have this problem -- is that right?

That's correct.

PR #3206, opened 2022-06-13
Numeric implements SqlOrd

joel-u410:
For postgres at least (untested on other databases), `Numeric` should implement `SqlOrd` to allow use of aggregation functions like `max()`.


weiznich:
Thanks for opening this PR. Generally speaking it is fine to add such an impl, but we cannot accept a `SqlOrder` impl that is conditionally applied if a specific feature is passed as this would enabled for the sqlite/mysql backend as well as soon as a user want to use more than one backend at once. This means I would like to see tests for all backends. Those should be part of the PR.

scvalex:
I just ran into this as well.

@joel-u410 Any chance you could add the requested tests so that this gets merged?  Otherwise, do you mind if I fork your branch, add the tests, and open a new PR?

scvalex:
In case anyone else runs into this before it gets merged, the simple workaround is to define our own `min` and `max` functions:

```
use diesel::sql_types::Numeric;

sql_function! { fn min(x: Numeric) -> Numeric; }
sql_function! { fn max(x: Numeric) -> Numeric; }
```

Issue #1612, opened 2018-04-04
Add `DatabaseErrorKind` variants for isolation failures


kestred:
## Feature Request

The `DatabaseErrorKind` enum is intended to container error kind variants describing errors which are commonly recoverable at the application level.

I'd like to have a variant or variants representing when an isolation failure occurs in the database (e.g. an otherwise valid request fails due to a conflicting transaction occurring at the same time; such as commit serializability, dead lock, etc).

In some applications these are frequently recoverable via simple retry logic (ideally implemented with randomized retry falloff), so it should make sense to have variants for them in the enum.

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.]

sgrif:
Seems reasonable. Can you point me at the specific error code and provide a minimal query/code to trigger the error?

kestred:
Error codes are `40001 | serialization_failure` and `40P01 | deadlock_detected`, often one or the other depending on whether the transaction is constructed with `build_transaction().serializable()` or not.

kestred:
This error occurs for my use case specifically when two different services were attempting to do something that conflicts, but I've not been successful building a simplified use case.

### Non-Working Test
Sourced from the [postgres documentation](https://www.postgresql.org/docs/current/static/transaction-iso.html#XACT-SERIALIZABLE) hoping to find a minimal test case --- I'd written this, but I must've gotten something a bit wrong 😓.

```rust
#[test]
#[cfg(feature = "postgres")]
fn isolation_error_is_detected() {
    // we need custom and multiple transactions, so we can't use default transaction
    let conn = connection_without_transaction();
    conn.execute(
            "CREATE TABLE IF NOT EXISTS isolation_example (
                id SERIAL PRIMARY KEY,
                class INTEGER NOT NULL,
                value INTEGER NOT NULL
            );",
        )
        .unwrap();
    conn.execute(
            "INSERT INTO isolation_example (class, value) VALUES
                (1, 10),
                (1, 20),
                (2, 100),
                (2, 200)
             ;",
        ).unwrap();

    let other_process = ::std::thread::spawn(|| {
        let other_conn = connection_without_transaction();
        other_conn.build_transaction().serializable().run::<_, result::Error, _>(|| {
            other_conn
                .execute("SELECT SUM(value) AS sum FROM isolation_example WHERE class = 1);")
                .unwrap();

[...]

PR #3208, opened 2022-06-16
Backport fixes to 2.0.x branch

weiznich:
This essentially prepares the final 2.0.0 release by:
* Backporting all relevant fixes
* Bumping the version numbers
* Adding a `Cargo.lock` file as this hopefully allows us to checkout this branch at any later point in time and it will continue to build. This should not affect the release to crates.io in any way.


weiznich:
@diesel-rs/core  I propose to do a final stable diesel 2.0 release next week. Please leave a comment here if you have any objections. I will fill a PR against the web repository next week so that we also update the webpage to diesel 2.0 + include the migration guide, the changelog and a small release post there.

Ten0:
> I propose to do a final stable diesel 2.0 release next week. Please leave a comment here if you have any objections.

The two things I have in mind that slightly scare me with regards to backwards compatibility upon solving are the following two comments:
1. https://github.com/diesel-rs/diesel/issues/1612#issuecomment-1096987553
2. https://github.com/diesel-rs/diesel/pull/2799#pullrequestreview-708943633
    I'm not super confident with the current approach of https://github.com/diesel-rs/diesel/pull/2799#issuecomment-903161724 because that introduces:
    
     **a.** Unnecessary lifetime bounds on the deserialization part
     **b.** A somewhat-global config at the connection level, whereas whether you want to use one or the other would more typically depend on the amount of data you're querying and what you're doing with it afterwards (unless row-by-row is always ~faster) - which appears to be fundamentally more fine-grained than connection-level. 
     
   If however row-by-row is always ~faster, having just that version is always ideal and consistent across all backends so we could probably keep just that.

Unfortunately I have been very busy with other topics and couldn't work on these yet.

 - I've commented at 1. on this topic.
 - Would you agree it would be useful to have the benchmark suggested in 2's comment and the decision would depend on it?
 - If so, what do you think would be a reasonable timing to attempt implementation?

weiznich:
> Return a iterator from Connection::load #2799 (review)
> I'm not super confident with the current approach of
> 
> Return a iterator from Connection::load #2799 (comment) because that introduces:
> 
> a. Unnecessary lifetime bounds on the deserialization part
> b. A somewhat-global config at the connection level, whereas whether you want to use one or the other would more typically depend on the amount of data you're querying and what you're doing with it afterwards (unless row-by-row is always ~faster) - which appears to be fundamentally more fine-grained than connection-level.
> 
> If however row-by-row is always ~faster, having just that version is always ideal and consistent across all backends so we could probably keep just that.

I remember that I've done some experimentation back then but I cannot find the corresponding code anymore. Will try to have another look in the next days. 
What I remember is that the row by row mode was significantly slower for all of our benchmarks. I think the main issue there was that it allocates things per row again and again, where the "normal" mode reuses buffers and other things.

To address the concrete points:

> a. Unnecessary lifetime bounds on the deserialization part

Yes the lifetime for `Cursor` is not optimal right now, but `Cursor` is not part of the public API. That means we can remove the lifetime if it turns out that we do not need it. That wouldn't work the other way around.

> b. A somewhat-global config at the connection level, whereas whether you want to use one or the other would more typically depend on the amount of data you're querying and what you're doing with it afterwards (unless row-by-row is always ~faster) - which appears to be fundamentally more fine-grained than connection-level.

The point is there is not really an other place to configure this than the connection, because as soon as you call `load_iter()` or whatever the query is already executed. What we could do is to provide the connection level configuration with some documentation that basically says: Do not use this if you don't call `Connection::load` directly. Then we could modify `RunQueryDsl` to do "runtime type checking" to see if the connection is a `PgConnection` and just enable that mode there transparently. That would be hacky but would result in that behaviour most people would expect.

Ten0:
> The point is there is not really an other place to configure this than the connection, because as soon as you call `load_iter()` or whatever the query is already executed

I was thinking about the design described in that comment : https://github.com/diesel-rs/diesel/pull/2799#pullrequestreview-708943633

weiznich:
I've put up a WIP PR for row by row mode in #3213 so that we could quickly check the performance implication of enabling it generally. 
That written: I would really prefer not to have backend specific methods there, but instead a way to enable it only for specific queries. Maybe even by wrapping the corresponding query into an "option" struct that signals this should be loaded via row-by-row mode. The WIP-PR contains some parts to just enable the row by row mode only for `load_iter()`.

weiznich:
Closed in favor of #3237

PR #3211, opened 2022-06-20
More straightforward transaction error handling

Ten0:
**Changes commit error strategy:**
 - Errors on commit will be returned directly, *unless* rollback is attempted and fails.
 - In that case, `Error::RollbackErrorOnCommit` will contain both errors. (`RollbackError` and `CommitTransactionFailed` variants are removed, and typical rollback errors are otherwise returned unpacked)
 - `BrokenTransaction` error is renamed to `BrokenTransactionManager`, as this is in effect when it is triggered, and "BrokenTransaction" could be confused with "top-level rollback necessary", which is a different concept.

This design stems from noticing that:
 - End-user having to match on original error as well as possibly
   recursively `CommitError` containing what they are looking for
   (typically, a possible serialization failure) is a hassle.
 - Rollback errors are far less common than commit errors, and typically
   result in abandonning to use this connection ever again, so it's fine
   if getting the details about those is a bit more complex.
 - It is still possible for the user to differentiate commit errors from
   in-closure errors if really necessary (backtrace will be properly
   collected, and user has control over the closure so they can check)

This allows for easier handling of e.g. serialization failures.

**Changes transaction error handling strategy:**
- Instead of providing API for custom error handling on commit, connection is free to register updates to the transaction manager status as it encounters errors.
- This transaction manager status is updated to be able to store that transaction is aborted and requires top-level rollback to be usable again.
- This information is used in `AnsiTransactionManager`'s commit/rollback logic. Notably, if top-level transaction is detected to be broken, the transaction manager will not even attempt rollback of savepoints.

This design seems to simplify error handling (number of lines goes down), while removing some mysql-specific code from the general `AnsiTransactionManager` and still allowing proper transaction depth tracking. (The tests updates mostly relate to the first part, about not packing rollback/commit errors.)


Ten0:
This is in draft pending answer to the following question (and non-pg tests fixes):

I'd need some enlightenment regarding what the purpose of this is:
https://github.com/diesel-rs/diesel/blob/46c366a08c2f64fa77eda9639ee15d50d21cb1f3/diesel/src/connection/transaction_manager.rs#L314
to be sure of what I'm doing.

Also, that stuff about `is_error_relevant_for_rollback` isn't done here:
https://github.com/diesel-rs/diesel/blob/7184f128a3c1435d12b32f413b14b6632e2f753c/diesel/src/pg/transaction.rs#L303
But I thought PG's transactions were a superset of ANSI, so why is it not here?

Somewhat independent but still slightly related: shouldn't we mark the transaction manager as broken if we reach this point:
https://github.com/diesel-rs/diesel/blob/df7ca0ae54a20e4c0d59c13e00d1c07667d61723/diesel/src/connection/transaction_manager.rs#L351
and `rollback_result` is an error?

Ten0:
other question: currently, when is `previous_error_relevant_for_rollback` cleaned after a successful commit or rollback?

weiznich:
> I'd need some enlightenment regarding what the purpose of this is:

This is required because mysql returns serialization errors not at the end of the transaction call, but on the corresponding statement call inside of the transaction. So to make the transaction return a serialization error we somehow need to remember that this was the likely reason why the transaction failed in the first place. This is essentially a hack to somehow get this information to the right place. See https://github.com/diesel-rs/diesel/pull/3021 for more details and discussion.

> Also, that stuff about is_error_relevant_for_rollback isn't done here:
> But I thought PG's transactions were a superset of ANSI, so why is it not here?

As noted above: That's mainly for the reason that this is currently a mysql specific hack. Sometimes I think that we we should just ditch `AnsiTransactionManager` at all and provide a tailored implementation for each backend. That would make such details much easier. In the end there are quite a lot non-obvious details there that differ from backend to backend.

> Somewhat independent but still slightly related: shouldn't we mark the transaction manager as broken if we reach this point:
> and rollback_result is an error?

Seems reasonable :+1:

> other question: currently, when is previous_error_relevant_for_rollback cleaned after a successful commit or rollback? (so that it doesn't affect subsequent transactions)

Opps, seems like I've missed that 🙈 
I would probably handle that here: https://github.com/diesel-rs/diesel/blob/df7ca0ae54a20e4c0d59c13e00d1c07667d61723/diesel/src/connection/transaction_manager.rs#L143

diesel/src/connection/transaction_manager.rs
```diff
@@ -56,10 +56,16 @@ pub trait TransactionManager<Conn: Connection> {
                 Self::commit_transaction(conn)?;
                 Ok(value)
             }
-            Err(e) => {
-                Self::rollback_transaction(conn).map_err(|e| Error::RollbackError(Box::new(e)))?;
-                Err(e)
-            }
+            Err(user_error) => Err(match Self::rollback_transaction(conn) {
+                Ok(()) => user_error,
+                Err(rollback_error) => E::from(match rollback_error {
```
weiznich, line 61: Can we unnest the match and Err wrapping here? That would make it much easier to read.

diesel/src/connection/transaction_manager.rs
```diff
@@ -228,15 +234,11 @@ where
             .previous_error_relevant_for_rollback
             .take()
         {
-            // we can safely ignore the result of process_commit_error here
-            // as the only other error than the "rollback error" variants
-            // is failing to load the transaction state, but that's something
-            // we have already done above
-            let _ = process_commit_error(
+            let _: Error = process_commit_error(
```
weiznich, line 237: We likely want to keep/add some comment here why this error is ignored.

- [ ]

diesel/src/result.rs
```diff
@@ -66,7 +66,13 @@ pub enum Error {
     ///
     /// An example of when this error would be returned is if a rollback has
     /// already be called on the current transaction.
-    RollbackError(Box<Error>),
+    RollbackError {
+        /// The error that caused the rollback failure
+        rollback_error: Box<Error>,
```
weiznich, line 71: I would like to improve the name of this field, so that it is clearer that this is the error that **caused** the rollback instead of the error that was returned by an failed rollback. Maybe `rollback_cause` or something like that

weiznich (commented):
I really like the general direction of this change :+1:
I've left a few small notes where things could be improved slightly.

diesel/src/connection/transaction_manager.rs
Ten0: I typically find this kind of writing easier to read because it quickly reads "whatever happens, we're gonna return an error - then here here's what error it will be".
That being said I can change that to wrap each error in err individually if you have a strong preference for this.

diesel/src/pg/connection/mod.rs
```diff
@@ -169,6 +116,50 @@ impl GetPgMetadataCache for PgConnection {
     }
 }
 
+#[inline(always)]
+fn update_transaction_manager_status<T>(
+    query_result: QueryResult<T>,
+    conn: &mut PgConnection,
+) -> QueryResult<T> {
+    if let Err(Error::DatabaseError { .. }) = query_result {
+        /// avoid monomorphizing for every result type - this part will not be inlined
```
weiznich, line 125: This comment seems to disagree with the `#[inline(always)]` above.

weiznich (commented):
I've left two minor comments, otherwise this looks great :+1:

diesel/src/pg/connection/mod.rs
Ten0: I want to always inline the above function, but not the `non_generic_inner` one

weiznich: I totally missed that inner function. Maybe explicitly put it outside the outer one to make this more clear?

Ten0: That doesn't seem worth the scoping pollution.
I think the "this part", the function name and call below, the fact it's a doc comment... should make enough clues for this to still be the best writing.

Ten0:
Note to self: there may be some documentation update to do still

Ten0:
@Emm as you had worked on this originally maybe you'll want to have a look on that I didn't forget anything. ;)

diesel/src/connection/transaction_manager.rs
```diff
@@ -854,10 +904,12 @@ mod test {
...
                             let r = q.execute(conn);
                             after_barrier.wait();
-                            r
-                        })
+                            dbg!(r)
+                        });
+                        dbg!(r2)
                     });
 
+                    dbg!(&result);
```
weiznich, line 912: These leftover `dbg!` calls can probably be removed, right?

diesel/src/mysql/connection/bind.rs
```diff
@@ -846,8 +846,9 @@ mod tests {
             ).execute(conn)
             .unwrap();
 
-        let stmt = conn.prepared_query(&crate::sql_query(
-            "SELECT
+        let stmt = super::super::prepared_query(
```
weiznich, line 849: I would prefer `crate::mysql::connection` or something like that here instead of `super::super`.

weiznich (approved):
Looks good to me beside the two minor nits. Feel free to merge it :+1:

PR #3213, opened 2022-06-24
Add a WIP postgres row-by-row implementation

weiznich:


weiznich:
Benchmark results:

```
group                                                                                     changes                                 master
-----                                                                                     -------                                 ------
bench_insert/diesel/1                                                                     1.00   286.4±81.22µs        ? ?/sec     1.16   332.4±59.28µs        ? ?/sec
bench_insert/diesel/10                                                                    1.00   363.1±55.26µs        ? ?/sec     1.06   386.0±52.89µs        ? ?/sec
bench_insert/diesel/100                                                                   1.00   973.9±91.46µs        ? ?/sec     1.00   978.4±93.72µs        ? ?/sec
bench_insert/diesel/25                                                                    1.00   432.2±60.34µs        ? ?/sec     1.17   505.9±58.65µs        ? ?/sec
bench_insert/diesel/50                                                                    1.00   614.6±61.10µs        ? ?/sec     1.05  643.6±101.84µs        ? ?/sec
bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially    1.00      9.5±0.59ms        ? ?/sec     1.13     10.7±0.37ms        ? ?/sec
bench_medium_complex_query/diesel/1                                                       1.00    82.6±13.14µs        ? ?/sec     1.02    84.0±17.28µs        ? ?/sec
bench_medium_complex_query/diesel/10                                                      1.01    91.3±17.38µs        ? ?/sec     1.00    90.2±15.46µs        ? ?/sec
bench_medium_complex_query/diesel/100                                                     1.03   156.2±50.52µs        ? ?/sec     1.00   151.1±24.29µs        ? ?/sec
bench_medium_complex_query/diesel/1000                                                    1.17  753.7±117.65µs        ? ?/sec     1.00  645.4±101.25µs        ? ?/sec
bench_medium_complex_query/diesel/10000                                                   1.12      5.9±0.28ms        ? ?/sec     1.00      5.3±0.37ms        ? ?/sec
bench_medium_complex_query/diesel_boxed/1                                                 1.00    91.5±18.43µs        ? ?/sec     1.06    97.0±18.25µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10                                                1.00    93.3±16.25µs        ? ?/sec     1.06    98.9±22.49µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/100                                               1.09   163.6±26.70µs        ? ?/sec     1.00   150.2±26.24µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/1000                                              1.13   766.4±82.67µs        ? ?/sec     1.00   680.5±93.23µs        ? ?/sec
bench_medium_complex_query/diesel_boxed/10000                                             1.17      6.4±0.52ms        ? ?/sec     1.00      5.4±0.32ms        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1                                     1.00   233.0±21.35µs        ? ?/sec     1.16   269.3±32.18µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10                                    1.00   252.8±23.87µs        ? ?/sec     1.15   290.8±29.68µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/100                                   1.04   416.5±35.81µs        ? ?/sec     1.00  400.5±119.85µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/1000                                  1.14  1611.3±118.79µs        ? ?/sec    1.00  1419.5±96.49µs        ? ?/sec
bench_medium_complex_query/diesel_queryable_by_name/10000                                 1.23     14.5±0.65ms        ? ?/sec     1.00     11.8±0.85ms        ? ?/sec
bench_trivial_query/diesel/1                                                              1.00     63.5±4.30µs        ? ?/sec     1.07     67.8±6.83µs        ? ?/sec
bench_trivial_query/diesel/10                                                             1.00     67.2±4.93µs        ? ?/sec     1.09     72.9±7.17µs        ? ?/sec
bench_trivial_query/diesel/100                                                            1.11    139.2±7.93µs        ? ?/sec     1.00   125.4±10.54µs        ? ?/sec
bench_trivial_query/diesel/1000                                                           1.19   726.8±50.42µs        ? ?/sec     1.00   608.7±25.52µs        ? ?/sec
bench_trivial_query/diesel/10000                                                          1.21      6.7±0.26ms        ? ?/sec     1.00      5.5±0.27ms        ? ?/sec
bench_trivial_query/diesel_boxed/1                                                        1.00     68.1±8.07µs        ? ?/sec     1.06     72.2±9.00µs        ? ?/sec
bench_trivial_query/diesel_boxed/10                                                       1.00     67.6±6.14µs        ? ?/sec     1.09     73.4±4.35µs        ? ?/sec
bench_trivial_query/diesel_boxed/100                                                      1.11    142.7±9.23µs        ? ?/sec     1.00    129.0±7.34µs        ? ?/sec
bench_trivial_query/diesel_boxed/1000                                                     1.24   789.1±34.02µs        ? ?/sec     1.00   634.1±37.49µs        ? ?/sec
bench_trivial_query/diesel_boxed/10000                                                    1.10      6.3±0.38ms        ? ?/sec     1.00      5.7±0.24ms        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1                                            1.00    135.7±6.38µs        ? ?/sec     1.19   161.9±59.53µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10                                           1.00    150.1±9.53µs        ? ?/sec     1.02    152.6±8.17µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/100                                          1.10   230.3±16.52µs        ? ?/sec     1.00   209.0±10.26µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/1000                                         1.35  1002.2±56.68µs        ? ?/sec     1.00   740.5±48.23µs        ? ?/sec
bench_trivial_query/diesel_queryable_by_name/10000                                        1.31      7.6±0.37ms        ? ?/sec     1.00      5.8±0.22ms        ? ?/sec

```

The results are mixed. There are regressions for large return set sizes, and improvements for smaller return set sizes.

weiznich:
@Ten0 Any comments here?

Ten0:
I find it surprising that we observe improvement for small query sets: I wouldn't expect the libpq internals to be significantly slower than what we have here. (I would expect the difference in number of operations to be less than 5us~=5000 CPU cycles).

It looks like row-by-row mode indeed introduces significant overhead for large queries.
I'm curious how this would look like for a costly query though (one that would filter a subset over a large number of lines in a table instead of just getting the full contents of the table) - maybe the fact we can start deserializing before the DB is done filtering would then show improvements. (I'm not sure what portion of actual query plans that would represent though.)

In any case, given the possibly negative impact of row-by-row, it looks like we ideally would want to make both interfaces available.

I'm going to meditate on the pros and cons of https://github.com/diesel-rs/diesel/pull/2799#pullrequestreview-708943633 vs https://github.com/diesel-rs/diesel/pull/3208#issuecomment-1165859477 some more. :)

weiznich:
For the sake of completeness I've run the benchmarks locally using the instruction mode. Overall this is a regression for all load benchmarks in term of instruction count.

```
bench_trivial_query/diesel/1
                        time:   [47597.2320 cycles 47695.8461 cycles 47808.7797 cycles]
                        change: [+3.8631% +4.2555% +4.6305%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  2 (2.00%) high severe
                        time:   [54309.3457 cycles 54415.0789 cycles 54530.7138 cycles]
                        change: [+2.9628% +3.3120% +3.6446%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
  1 (1.00%) high severe
                        time:   [90264.2711 cycles 90431.0609 cycles 90610.6913 cycles]
                        change: [+1.5527% +1.7965% +2.0554%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
bench_trivial_query/diesel/10
                        time:   [79298.5560 cycles 79399.3846 cycles 79510.0819 cycles]
                        change: [+28.389% +28.971% +29.478%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  2 (2.00%) high severe
bench_trivial_query/diesel_boxed/10
                        time:   [85518.3680 cycles 85702.6821 cycles 85898.4498 cycles]
                        change: [+26.170% +26.532% +26.952%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  2 (2.00%) high severe
bench_trivial_query/diesel_queryable_by_name/10
                        time:   [129795.8086 cycles 129938.2929 cycles 130082.5921 cycles]
                        change: [+21.670% +22.025% +22.370%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe
bench_trivial_query/diesel/100
                        time:   [401181.9958 cycles 402656.1712 cycles 404099.1215 cycles]
                        change: [+84.775% +85.258% +85.845%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
bench_trivial_query/diesel_boxed/100
                        time:   [408520.1064 cycles 409364.0049 cycles 410375.7279 cycles]
                        change: [+83.799% +84.372% +84.947%] (p = 0.00 < 0.05)
                        Performance has regressed.
bench_trivial_query/diesel_queryable_by_name/100
                        time:   [533650.6759 cycles 533915.7405 cycles 534229.5046 cycles]
                        change: [+82.257% +82.480% +82.701%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  7 (7.00%) high mild
  1 (1.00%) high severe
bench_trivial_query/diesel/1000
                        time:   [3700857.4415 cycles 3719000.9006 cycles 3737656.1750 cycles]
                        change: [+103.72% +104.51% +105.16%] (p = 0.00 < 0.05)
                        Performance has regressed.
bench_trivial_query/diesel_boxed/1000
                        time:   [3728010.4318 cycles 3749026.2192 cycles 3768316.7382 cycles]
                        change: [+103.71% +104.49% +105.33%] (p = 0.00 < 0.05)
                        Performance has regressed.
bench_trivial_query/diesel_queryable_by_name/1000
                        time:   [4726384.5034 cycles 4736202.3426 cycles 4744591.8467 cycles]
                        change: [+115.36% +115.86% +116.39%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) low mild
                        time:   [37410302.7121 cycles 37419484.3000 cycles 37428510.7305 cycles]
                        change: [+109.52% +109.59% +109.65%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) low mild
                        time:   [37309005.7529 cycles 37319840.3086 cycles 37330868.6770 cycles]
                        change: [+108.90% +108.96% +109.02%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
                        time:   [46428695.4627 cycles 46436181.9871 cycles 46443436.0002 cycles]
                        change: [+119.47% +119.63% +119.79%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) low mild
  1 (1.00%) high mild

                        time:   [52560.0971 cycles 52606.9583 cycles 52662.6288 cycles]
                        change: [+3.1916% +3.5610% +3.9727%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
  1 (1.00%) high severe
                        time:   [67212.1310 cycles 67298.2557 cycles 67390.0599 cycles]
                        change: [+2.2071% +2.6874% +3.1342%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
                        time:   [94080.4016 cycles 94171.7581 cycles 94276.1093 cycles]
                        change: [+0.4806% +0.9399% +1.3573%] (p = 0.00 < 0.05)
  3 (3.00%) high mild
  2 (2.00%) high severe
bench_medium_complex_query/diesel/10
                        time:   [72779.4448 cycles 72912.1282 cycles 73055.2279 cycles]
                        change: [+17.724% +18.511% +19.302%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  1 (1.00%) high severe
bench_medium_complex_query/diesel_boxed/10
                        time:   [86814.1199 cycles 86897.8617 cycles 86994.6888 cycles]
                        change: [+14.470% +15.003% +15.521%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
bench_medium_complex_query/diesel_queryable_by_name/10
                        time:   [157878.3881 cycles 158142.6598 cycles 158392.6842 cycles]
                        change: [+25.868% +26.130% +26.379%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
bench_medium_complex_query/diesel/100
                        time:   [306595.0948 cycles 306844.5870 cycles 307122.1109 cycles]
                        change: [+63.483% +63.789% +64.074%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  3 (3.00%) high severe
bench_medium_complex_query/diesel_boxed/100
                        time:   [316086.6459 cycles 316272.2656 cycles 316455.5422 cycles]
                        change: [+56.803% +57.097% +57.382%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
bench_medium_complex_query/diesel_queryable_by_name/100
                        time:   [802072.8662 cycles 802365.1052 cycles 802668.2263 cycles]
                        change: [+74.344% +74.505% +74.692%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  2 (2.00%) high severe
bench_medium_complex_query/diesel/1000
                        time:   [2627144.8122 cycles 2635477.3334 cycles 2643940.6887 cycles]
                        change: [+78.211% +78.695% +79.129%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
bench_medium_complex_query/diesel_boxed/1000
                        time:   [2629716.4933 cycles 2635236.3455 cycles 2641233.3298 cycles]
                        change: [+76.738% +77.050% +77.397%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
bench_medium_complex_query/diesel_queryable_by_name/1000
                        time:   [7297458.5960 cycles 7300823.1368 cycles 7304293.9112 cycles]
                        change: [+89.181% +89.336% +89.487%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
                        time:   [26266563.4646 cycles 26278191.4520 cycles 26289506.5685 cycles]
                        change: [+84.615% +84.722% +84.837%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
                        time:   [26279791.1679 cycles 26288650.8567 cycles 26297794.7088 cycles]
                        change: [+84.058% +84.206% +84.313%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
                        time:   [73886043.2784 cycles 73902999.1980 cycles 73920138.7056 cycles]
                        change: [+93.905% +94.011% +94.110%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially: Collecting 100 samples in estimated 7.2667 s (200 itera                                                                                                                                                            bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
                        time:   [49708148.5605 cycles 49723760.7950 cycles 49740166.4571 cycles]
                        change: [+79.917% +80.068% +80.221%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  1 (1.00%) high severe

bench_insert/diesel/1   time:   [52515.0953 cycles 52891.5925 cycles 53311.4130 cycles]
                        change: [-2.8378% -1.7171% -0.6849%] (p = 0.00 < 0.05)
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
bench_insert/diesel/10  time:   [81408.0281 cycles 81876.4082 cycles 82379.2810 cycles]
                        change: [-1.7340% -0.3680% +0.6967%] (p = 0.60 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  9 (9.00%) high mild
  2 (2.00%) high severe
bench_insert/diesel/25  time:   [133455.4938 cycles 133861.3460 cycles 134308.2095 cycles]
                        change: [-1.1346% -0.3288% +0.3541%] (p = 0.42 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe
bench_insert/diesel/50  time:   [214988.8688 cycles 215821.6580 cycles 216722.8772 cycles]
                        change: [-0.7615% -0.3394% +0.1251%] (p = 0.13 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
bench_insert/diesel/100 time:   [373649.9564 cycles 374837.2927 cycles 376075.5133 cycles]
                        change: [+0.3659% +0.7042% +1.0702%] (p = 0.00 < 0.05)
  2 (2.00%) high mild
```

weiznich:
I've opened https://github.com/diesel-rs/diesel/pull/3221 for the future proofing

Ten0:
> I propose to just land both changes

Both these of changes seem fine. 👍
They may be useful to us in the future and I don't think they would be very limiting to users right now (and if they turn out to be we can revert).

> * This would result in having `RunQueryDsl::load_iter()` and `RunQueryDsl::load_then_deserialize_iter()` as part of the public API. Implementation wise we will likely want to add another parameter to `LoadQuery::internal_load()` to enable this change.

I think ideally then these would be different return types, so two different functions, to avoid the lifetime bound on the connection in the `load_then_deserialize_iter` case and avoid the internal requirement for a single iterator to support both cases.

---
> * this PR suggests that we just want to default to row-by-row mode for `RunQueryDsl::load_iter()` but not for `RunQueryDsl::load()`, so that this "just" works for most cases where users may want to use an iterator.

So it seems there are in fact 3 options (I hadn't thought of n°2):
1. two functions (`load_iter()` loading as we iterate, `load_then_deserialize_iter()` loading first and returning an iterator for deserialization that is not bound to the lifetime of the connection. For now, the latter would only be made available on PG - until somebody comes up with a need for benefit 4 below)
2. one function, defaulting to row-by-row
3. one function, defaulting to full-load

So I have meditated some.
I believe in the following 3 statements:
1. General case if using Postgres is you're loading a reasonable amount of data, so ideal performance to load into a HashMap would be achieved by full-load into deserialize iter (as proven significant by this PR).
2. Cases where people are stream-loading very large amounts of data are less common. It would be fine to have to manually enable it then, through a less direct API.
3. In Rust, the most performant pattern is typically also the simplest code (e.g. using references instead of owned data is faster because you don't allocate, and is also simpler because you have `Copy`, so you don't have to write `.clone()`, unnecessary `collect()`s are unnecessary code and unnecessary performance decrease...).
This is one of the things I enjoy a lot about Rust, and I would want to preserve this as much as possible for Diesel as well.
4. People using `.load_iter` are likely to expect that data would be stream-processed from the database, especially if that is the case for all other backends.

So:
 - If we do option 2, we lose statement 3 (because statements 1 and 2).
 - If we do option 3, we lose statement 4, so we'd need to strongly document that but people are likely to still not read it.
 - If we do option 1, we increase the complexity of the global API for usage that would typically be required in a single backend.

Statements 1,2 and 3 would tend to have me favor option 3 over option 2, although that is detrimental to 4:
- I wouldn't want, in order to obtain the best perf., to have to make a call on my connections to set them to "best performance mode".
- If users miss the "you may want to disable row-by-row to achieve the best performance" documentation, the mis-usage is likely to go unnoticed. (also if good practice for perf. end up being having to globally switch off row by row every time, we should probably make that the default)

However option 3, due to the requirement to activate row-by-row manually, actually also would require us to add documentation on [`RunQueryDsl::load_iter`](https://docs.diesel.rs/master/diesel/query_dsl/trait.RunQueryDsl.html#method.load_iter), so that ends up also having an similar impact on the general complexity of the API (measured in the number of concepts you need to assimilate to use it properly) as doing option 1.
While the impact if missed wouldn't be as permanent, it may be large, as it may result in breaking software that would otherwise work.

So overall I would go for option 1, which does satisfy 1 to 4, where the following benefits seem worth the added complexity for the public api (that I estimate small) of an additional function in the [`RunQueryDsl`](https://docs.diesel.rs/master/diesel/prelude/trait.RunQueryDsl.html) trait (pretty much only?):
1. Compared to option 2, general case if using Postgres is you're loading a reasonable amount of data, so ideal performance to load into a HashMap would be easily achieved by `load_then_deserialize_iter().collect::<Result<HashMap<_,_>,_>>()?`
2. Compared to option 3, it would be harder to miss "actually load while iterating please Postgres - different behavior from other backends" than if it's just `load_iter()` with a somewhat-hidden switch on the connection. (Deserialization may even be done after closing transaction 🏎️)
3. Compared to option 2, it would be harder to miss `load_then_deserialize_iter()` as the most performant option if it's a separate function than if it's `load_iter()` with a somewhat-hidden switch on the connection (because it would show [here](https://docs.diesel.rs/master/diesel/prelude/trait.RunQueryDsl.html))
4. `load_then_deserialize_iter()` is not necessarily completely nonsensical for other backends if you really really want to release the connection (close the transaction) ASAP. (it would provide an iterator whose lifetime is not bound to that of the connection)
5. It's a more direct/natural implementation when the goal is to expose both functionalities, making it possible to have even different iterator types for each mode, making it possible to implement both rather independently in different modules, instead of needing to have a larger iterator type able to store properties for both cases as well as a discriminant, and having to `match` on the mode (or other morally equivalent ifs) at every call to `next()`, making each function of the iterator individually more complex. This kind of pattern is generally less likely to cause issues where constraints on one side would prevent from improving behavior on the other (like would be the case here for lifetime in one case and size of extra fields in the other). The usual benefit of using dynamically-dispatched behavior is that it enables for building a type that can be dynamically one or the other, but in that case it seems:
   - Unlikely to be needed, as whether you need to row-by-row would almost always be a property of the query, which is described at the same place as the function would be called at, and the iterator entirely consumed.
   - Easy to achieve anyway if ever needed using [`Either`](https://docs.rs/either/1.7.0/either/enum.Either.html#impl-Iterator) with no downside compared to having it embedded in Diesel

---
(Going for option 1 would probably imply separating the interfaces now as that would otherwise be a somewhat-breaking change - I could probably work on that soon.)

weiznich:
@Ten0 I've pushed https://github.com/diesel-rs/diesel/pull/3213/commits/75e5d88641876f1857783b19c7b331dad9634dd8 which contains more or less a complete version of the generic type approach.
The short version is that `RunQueryDsl::load_iter()` is now generic over a loading mode, so that users can specify that mode on case by case basis for each query. For the `Sqlite` and `Mysql` backend only a default loading mode is provided, for the `Pg` backend a default mode (the current implementation) and a `RowByRowMode` implementation is provided. Users can call this by passing the corresponding type as argument e.g. `query.load_iter::<_, RowByRowMode>(&mut conn)?`. The lifetime of the returned cursor depends on the mode, which means for the default mode no lifetime dependency to the connection is added, while the row by row mode "blocks" the connection. I think this might be an interesting middle ground between the other solutions as it does not result in API duplication, but allows fine grained control over all parts. 
There are a few disadvantages/rough edges of the current implementation:

* This requires to always specify the loading mode for `load_iter()`, at least if a `PgConnection` is used. For `SqliteConnection` and `MysqlConnection` rustc infers that correctly to the only type that is usable there. We might be able to fix that someday, if rust ever supports defaulted generic for functions or if we find another clever workaround for that.
* There are sometimes type inference issues with other `Connection` methods as well. We might workaround that by restructuring the way the `Connection` trait is currently structured. The current implementation will for example fail to infer the correct loading mode type for `PgConnection::establish` .
 

In addition to the general issues pointed out above, the implementation is missing tests and documentation and we might need to fix other small things as well (r2d2, for exampler

Ten0:
> For `SqliteConnection` and `MysqlConnection` rustc infers that correctly to the only type that is usable there.

Aha that's a nice trick! I think this solves our issues :)

> * There are sometimes type inference issues with other `Connection` methods as well. We might workaround that by restructuring the way the `Connection` trait is currently structured. The current implementation will for example fail to infer the correct loading mode type for `PgConnection::establish` .

Shouldn't the genericity be on the `load_iter` function instead of the `Connection` trait?
It looks like it would avoid this issue.
https://github.com/diesel-rs/diesel/commit/75e5d88641876f1857783b19c7b331dad9634dd8#diff-0506f1581eee95db62d34d7529dca6387d95ff8c416c026fd5d8265368b9bc30R197
https://github.com/diesel-rs/diesel/commit/75e5d88641876f1857783b19c7b331dad9634dd8#diff-0506f1581eee95db62d34d7529dca6387d95ff8c416c026fd5d8265368b9bc30R370

weiznich:
> Shouldn't the genericity be on the load_iter function instead of the Connection trait?
It looks like it would avoid this issue.
75e5d88#diff-0506f1581eee95db62d34d7529dca6387d95ff8c416c026fd5d8265368b9bc30R197
75e5d88#diff-0506f1581eee95db62d34d7529dca6387d95ff8c416c026fd5d8265368b9bc30R370

That's unfortunately not that simple. If I add the generic type there, rustc seems not to be able to infer `DefaultLoadingMode` for `SqliteConnection`/`MysqlConnection` anymore. I've pushed https://github.com/diesel-rs/diesel/pull/3213/commits/80d621a96464b59a91f29789f40adac089df30c8 which moves `Connection::load` into a different trait, so that only that method uses a trait which is generic over the loading mode. This fixes the type inference issues. That commit is almost complete, it only misses some additional tests.  

In my opinion that might be the right solution for this problem. As we cannot introduce that later on, we need to decide now if we want to go this way or not.
If we choose this way I suggest that we release a 2.0.0-rc.1 version + wait maybe 2 weeks before doing a final 2.0 release. I'm happy to hear opinions on this.

diesel/src/pg/connection/mod.rs
```diff
@@ -110,6 +197,28 @@ impl Connection for PgConnection {
...
+        T: Query + QueryFragment<Self::Backend> + QueryId + 'query,
+        Self::Backend: QueryMetadata<T::SqlType>,
+    {
+        self.with_prepared_query(&source, |stmt, params, conn| {
+            use self::private::PgLoadingMode;
+            let result = stmt.execute(conn, &params, Self::USE_ROW_BY_ROW_MODE)?;
+            let cursor = Self::get_cursor(conn, result);
+
+            cursor
+        })
```
weiznich, line 218: @Ten0 We need to somehow integrate your `update_transaction_manager_status` function here, but I'm not sure on the best way to actually do this due to lifetime issues. The problem is that now the cursor may depend on the lifetime of the connection itself, which restricts the possibility to call something else with dependency on the connection lifetime itself.

diesel/src/pg/connection/mod.rs
```diff
@@ -25,6 +25,92 @@ use crate::RunQueryDsl;
...
+///
+/// for r in iter1 {
+///     let (id, name) = r?;
+///     println!("Id: {} Name: {}", id, name);
+/// }
+///
+/// for r in iter2 {
+///     let (id, name) = r?;
+///     println!("Id: {} Name: {}", id, name);
+/// }
```
weiznich, line 73: This currently fails to compile for whatever reason. I need to figure out where the additional lifetime comes from.

weiznich:
This is now ready to review. I've already noted two points that would need to be fixed before merging, otherwise the implementation is complete.

diesel/src/pg/connection/mod.rs
weiznich: I've pushed https://github.com/diesel-rs/diesel/pull/3213/commits/2880f99ccea5912b9d019c26142bd6abd5a4d9a1 to address this, but I'm not sure if that's the "correct" solution for row by row mode.

weiznich: https://www.postgresql.org/docs/current/libpq-single-row-mode.html Can be read in such a way that we would need to call `update_transaction_manager_status` each time we get a result back from the database for row by row mode.

Ten0: > https://www.postgresql.org/docs/current/libpq-single-row-mode.html Can be read in such a way that we would need to call `update_transaction_manager_status` each time we get a result back from the database for row by row mode.

That's what I thought too. Looks like your implementation does that :)

In order to pass around a single reference everywhere instead of two we could pack both transaction_manager and raw_connection in a sub-struct in PgConnection.

weiznich:
@Ten0 I've merged this now and will try to release a 2.0.0-rc.1 version these days. Is there anything else that should be included?

Ten0 (commented):
Sorry for the late review.
This seems to work for us :)

diesel/src/pg/connection/cursor.rs
```diff
@@ -66,112 +57,348 @@ impl Iterator for Cursor<'_> {
...
+            Some(Ok(row))
+        } else {
+            None
+        }
+    }
+}
+
+impl Drop for RowByRowCursor<'_> {
+    fn drop(&mut self) {
+        loop {
```
Ten0, line 118: Is there no way to cancel the rest of the execution of the query and not load any more?

diesel/src/pg/connection/mod.rs
```diff
@@ -119,33 +237,26 @@ impl GetPgMetadataCache for PgConnection {
...
-    conn: &mut PgConnection,
+    conn: &mut ConnectionAndTransactionManager,
 ) -> QueryResult<T> {
     if let Err(Error::DatabaseError { .. }) = query_result {
         /// avoid monomorphizing for every result type - this part will not be inlined
-        fn non_generic_inner(conn: &mut PgConnection) {
-            if AnsiTransactionManager::transaction_manager_status_mut(conn)
-                .is_not_broken_and_in_transaction()
-            {
+        fn non_generic_inner(raw_conn: &mut RawConnection, tm: &mut AnsiTransactionManager) {
```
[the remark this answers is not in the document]
weiznich: I've opted for two arguments here as this makes accessing the relevant parts in this function easier in my opinion.

diesel/src/pg/connection/cursor.rs
weiznich: https://www.postgresql.org/docs/13/libpq-single-row-mode.html states that you have to call `get_next_result()` till it returns a null pointer. 

The only way to cancel server side operations is:
https://www.postgresql.org/docs/current/libpq-cancel.html
If I read that correctly, this pages says that this operation is best effort and may change nothing. This would imply that we need to call get_next_result()` in a loop nevertheless just to be sure.

weiznich:
I will try to open an PR addressing the other two points later on today.

diesel/src/pg/connection/cursor.rs
Ten0: That is probably desirable behavior still if the iterator is not already entirely consumed: if we are looping on a large dataset (which will generally be the use-case for row-by-row loading) and break in the middle, I think we want to tell the database that it doesn't need to send the entire rest of its results: If it is large, data is not already on the client, so we can try to prevent the database from sending them, or even compute them.

weiznich: https://stackoverflow.com/questions/18015203/postgresql-cancel-query-from-c-c-program indicates that canceling such queries won't do much.

diesel/src/pg/connection/mod.rs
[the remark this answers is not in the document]
weiznich: As this code is only hit in case of error, I would rather like to optimize for readability instead of performance as performance won't matter that much here.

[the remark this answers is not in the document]
weiznich: :+1:

PR #3221, opened 2022-07-06
Future proofing `LoadQuery` and `RunQueryDsl` for possible row-by-row

weiznich:
libpq implementation

### DIFF
--- a/diesel/src/query_dsl/load_dsl.rs
+++ b/diesel/src/query_dsl/load_dsl.rs
@@ -25,6 +25,9 @@ where
     for<'a> Self: LoadQueryGatWorkaround<'a, 'query, Conn, U>,
 {
     /// Load this query
+    #[diesel_derives::__diesel_public_if(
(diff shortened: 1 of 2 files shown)


diesel/src/query_dsl/load_dsl.rs
```diff
@@ -25,6 +25,9 @@ where
     for<'a> Self: LoadQueryGatWorkaround<'a, 'query, Conn, U>,
 {
     /// Load this query
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
```
Ten0, line 29: do we want to even make the whole trait private?

Ten0 (approved):

diesel/src/query_dsl/load_dsl.rs
weiznich: The trait itself is required as bound for generic code, so that users can call `RunQueryDsl::load` (and `load_iter`) in a generic context.

PR #3228, opened 2022-07-14
Improve code generated by the insertable derive

weiznich:
This commit changes the code generated by the insertable derive in such
a way that error messages about field mismatches now point to the fields
itself, instead to the derive.

See here for the old error message: https://github.com/weiznich/rust-foundation-community-grant/blob/77f131de301a4a113c45db15c66e6247a3e4bee9/test_cases/tests/diesel/bad_insertable_field.stderr

### DIFF
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -1,9 +1,9 @@
-use proc_macro2::TokenStream;
-use syn::{DeriveInput, Expr, Path, Type};
-
 use attrs::AttributeSpanWrapper;
 use field::Field;
 use model::Model;
+use proc_macro2::TokenStream;
+use quote::quote_spanned;
+use syn::{DeriveInput, Expr, Path, Type};
 use util::{inner_of_option_ty, is_option_ty, wrap_in_dummy_mod};
 
 pub fn derive(item: DeriveInput) -> TokenStream {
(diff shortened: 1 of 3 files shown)


diesel_derives/src/insertable.rs
```diff
@@ -87,7 +87,7 @@ pub fn derive(item: DeriveInput) -> TokenStream {
         {
             type Values = <(#(#direct_field_ty,)*) as Insertable<#table_name::table>>::Values;
 
-            fn values(self) -> Self::Values {
+            fn values(self) -> <(#(#direct_field_ty,)*) as Insertable<#table_name::table>>::Values {
```
pksunkara, line 90: This doesn't need to change.

diesel_derives/src/insertable.rs
```diff
@@ -105,8 +105,8 @@ pub fn derive(item: DeriveInput) -> TokenStream {
             {
                 type Values = <(#(#ref_field_ty,)*) as Insertable<#table_name::table>>::Values;
 
-                fn values(self) -> Self::Values {
-                    (#(#ref_field_assign,)*).values()
+                fn values(self) -> <(#(#ref_field_ty,)*) as Insertable<#table_name::table>>::Values {
```
pksunkara, line 108: This doesn't need to change.

diesel_derives/src/insertable.rs
```diff
@@ -105,8 +105,8 @@ pub fn derive(item: DeriveInput) -> TokenStream {
...
+                   (#(#ref_field_assign,)*).values()
```
pksunkara, line 109: Bad indentation

pksunkara (approved):
Minor nitpicks

diesel_derives/src/insertable.rs
weiznich: It does need to change as otherwise rustc emits for whatever reason 2 error message, where one is pointing to the derive itself, and the other one to the corresponding field.

diesel_derives/src/insertable.rs
weiznich:
```suggestion
                    (#(#ref_field_assign,)*).values()
```

PR #3235, opened 2022-07-16
Updated ipnetwork to 0.20

Razican:
This PR updates the ipnetwork dependency to also allow ipnetwork 0.20.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Increasing the minimal supported Rust version will always be coupled at least wi
   may be enabled concurrently with the previously existing `network-address`
   feature.
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Razican:
It seems that `clap` now requires 1.56.1, instead of 1.56.0, and this breaks the build.

weiznich (approved):
This looks fine to me. Before merging a changelog entry should be added.

Razican:
> This looks fine to me. Before merging a changelog entry should be added.

I'm adding that change. Should I update the minimum supported Rust version to 1.56.1 too?

weiznich:
> Should I update the minimum supported Rust version to 1.56.1 too?

I've already pushed that fix.

PR #3238, opened 2022-07-19
CLI: Allow running without a `down.sql` migration

Ten0:
Some people don't like to write down migrations, as depending on your workflow, they take a lot of time to write, may be *never* used (esp. when you have CD pipelines and just reset the database from scratch when developing), and consequently are typically largely untested.

I wish to support this for those (like us) who would rather write a new `up` migration to revert unwanted changes through CI/CD if necessary.

(NB: The original issue is that since we typically wouldn't fill `down.sql`, we would have to go edit that file to add a newline for formatting everytime we create a migration, and then leave it there even though it doesn't mean anything, and we would be unsure which ones were properly filled and which ones weren't.)

### DIFF
--- a/diesel_migrations/migrations_macros/src/embed_migrations.rs
+++ b/diesel_migrations/migrations_macros/src/embed_migrations.rs
@@ -47,7 +47,7 @@ fn migration_literal_from_path(path: &Path) -> proc_macro2::TokenStream {
     if version_from_string(&name).is_none() {
         panic!(
             "Invalid migration directory, the directory's name should be \
-             <timestamp>_<name_of_migration>, and it should only contain \
+             <timestamp>_<name_of_migration>, and it should contain \
              up.sql and down.sql."
         );
     }
--- a/diesel_migrations/migrations_macros/src/embed_migrations.rs
+++ b/diesel_migrations/migrations_macros/src/embed_migrations.rs
@@ -46,21 +46,28 @@ fn migration_literal_from_path(path: &Path) -> proc_macro2::TokenStream {
         .to_string_lossy();
     if version_from_string(&name).is_none() {
         panic!(
-            "Invalid migration directory, the directory's name should be \
+            "Invalid migration directory: the directory's name should be \
              <timestamp>_<name_of_migration>, and it should contain \
-             up.sql and down.sql."
+             up.sql and optionally down.sql."
         );
     }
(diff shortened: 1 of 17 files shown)


diesel_migrations/migrations_macros/src/embed_migrations.rs
```diff
@@ -47,7 +47,7 @@ fn migration_literal_from_path(path: &Path) -> proc_macro2::TokenStream {
     if version_from_string(&name).is_none() {
```
weiznich, line 47: Should we handle the migrations without down migration for `embed_migrations` as well? I mean we could just convert accept an `Option` as argument there + error at runtime similarly to what you've implemented for file based migrations.

weiznich (commented):
This looks fine for me, but I would like to see the following issues addressed before merging:

* A test in diesel cli that just checks that no down migration is created and common subcommands work with this setup
* Decide how how want to handle such migrations for `embed_migrations!`

weiznich (approved):
Looks good for me :+1:
Feel free to merge it as soon as the CI passes

diesel_migrations/src/file_based_migrations.rs
```diff
@@ -185,11 +185,12 @@ impl<DB: Backend> Migration<DB> for SqlFileMigration {
...
-            conn,
-            &self.base_path.join("down.sql"),
-            &self.name,
-        )?)
+        let down_path = self.base_path.join("down.sql");
+        if !matches!(down_path.metadata(), Err(e) if e.kind() == std::io::ErrorKind::NotFound) {
+            Ok(run_sql_from_file(conn, &down_path, &self.name)?)
+        } else {
+            Err(MigrationError::NoMigrationRevertFile.into())
+        }
```
weiznich, line 193:
```suggestion
        if matches!(down_path.metadata(), Err(e) if e.kind() == std::io::ErrorKind::NotFound) {
            Err(MigrationError::NoMigrationRevertFile.into())
        } else {
            Ok(run_sql_from_file(conn, &down_path, &self.name)?)
        }
```

To fix the clippy failure

weiznich:
@Ten0 you want to include this into the rc.1 release as well?

Ten0:
No opposition but also we don't personally need it as we are following master automatically anyway.

PR #3240, opened 2022-07-21
Improve the documentation of `QueryDsl::inner_join`

weiznich:
This commit adds some notes about the used join syntax and how this can
affect query planing on postgres based on the discussion in #3216.

Additionally it removes the restricting that the same table cannot be
joined twice. Instead we point to `diesel::alias!` now.

### DIFF
--- a/diesel/src/query_dsl/mod.rs
+++ b/diesel/src/query_dsl/mod.rs
@@ -331,10 +331,9 @@ pub trait QueryDsl: Sized {
     /// [`.on`]: JoinOnDsl::on()
     ///
     /// You can join to as many tables as you'd like in a query, with the
-    /// restriction that no table can appear in the query more than once. The reason
-    /// for this restriction is that one of the appearances would require aliasing,
-    /// and we do not currently have a fleshed out story for dealing with table
-    /// aliases.
+    /// restriction that no table can appear in the query more than once. For
+    /// tables that appear more than once in a single query the usage of [`alias!`](crate::alias!)
+    /// is required.
     ///
     /// You will also need to call [`allow_tables_to_appear_in_same_query!`].
     /// If you are using `diesel print-schema`, this will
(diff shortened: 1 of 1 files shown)


diesel/src/query_dsl/mod.rs
```diff
@@ -363,6 +365,15 @@ pub trait QueryDsl: Sized {
     ///     INNER JOIN comments ON comments.user_id = users.id
     /// ```
     ///
+    /// The exact generated SQL may change in future diesel version as long as the
+    /// generated query continue to produce same results. The currently generated
+    /// SQL is refered as ["explicit join"](https://www.postgresql.org/docs/current/explicit-joins.html)
+    /// by the PostgreSQL documentation and may have implications on the choosen query plan
+    /// for large numbers of joins in the same query. Checkout the documentation of the
+    /// [`join_collapse_limit` paramater](https://www.postgresql.org/docs/current/runtime-config-query.html#GUC-JOIN-COLLAPSE-LIMIT)
+    /// to control this behaviour.
```
JohnTitor, line 374:
```suggestion
    /// The exact generated SQL may change in future diesel version as long as the
    /// generated query continues to produce same results. The currently generated
    /// SQL is referred as ["explicit join"](https://www.postgresql.org/docs/current/explicit-joins.html)
    /// by the PostgreSQL documentation and may have implications on the chosen query plan
    /// for large numbers of joins in the same query. Checkout the documentation of the
    /// [`join_collapse_limit` paramater](https://www.postgresql.org/docs/current/runtime-config-query.html#GUC-JOIN-COLLAPSE-LIMIT)
    /// to control this behaviour.
```

diesel/src/query_dsl/mod.rs
```diff
@@ -363,6 +365,15 @@ pub trait QueryDsl: Sized {
...
+    ///
+    ///
```
JohnTitor, line 376: nit:
```suggestion
```

JohnTitor (approved):
LGTM with a few nits

PR #3242, opened 2022-07-22
Allow custom select expressions in `Selectable`

Ten0:
It's is a recurring use-case that 90% of the fields of a `Selectable` struct could be selected directly, but just one of them needs a bit more complex expression.

For this use case, this commit adds support for specifying fully custom select expression at the field level when deriving `Selectable`, by specifying the `select_expression` and `select_expression_type` attributes.

e.g: 
https://github.com/diesel-rs/diesel/blob/0d6395dd74dbf5ed1c0621082a40d5544a4fd224/diesel_derives/tests/selectable.rs#L131-L135

This could probably be later improved by getting able to infer `select_expression_type` for simple cases.

### DIFF
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -855,7 +855,7 @@ pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
 ///   It should be used in conjunction with `select_expression_type` (described below)
 /// * `#[diesel(select_expression_type = the_custom_select_expression_type]`, to be used
 ///   in conjunction with `select_expression` (described above).
-///   For example: `#[diesel(select_expression = dsl::IsNotNull<my_table::some_field>)]`
+///   For example: `#[diesel(select_expression_type = dsl::IsNotNull<my_table::some_field>)]`
 #[proc_macro_error]
 #[proc_macro_derive(Selectable, attributes(diesel))]
 pub fn derive_selectable(input: TokenStream) -> TokenStream {
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -848,6 +848,14 @@ pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
 /// * `#[diesel(embed)]`, specifies that the current field maps not only
 ///    single database column, but is a type that implements
 ///    `Selectable` on it's own
+/// * `#[diesel(select_expression = some_custom_select_expression)]`, overrides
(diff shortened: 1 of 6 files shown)


Ten0:
That thing I was talking about at https://github.com/orgs/diesel-rs/teams/core/discussions/1/comments/1 ;)

diesel_derives/src/lib.rs
```diff
@@ -848,6 +848,14 @@ pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
...
 ///    single database column, but is a type that implements
 ///    `Selectable` on it's own
+/// * `#[diesel(select_expression = some_custom_select_expression)]`, overrides
+///   the entire select expression for the given field. It may be used to select with
+///   custom tuples, or specify `select_expression = my_table::some_field.is_not_null()`,
+///   or separate tables...
+///   It should be used in conjunction with `select_expression_type` (described below)
+/// * `#[diesel(select_expression_type = the_custom_select_expression_type]`, to be used
+///   in conjunction with `select_expression` (described above).
+///   For example: `#[diesel(select_expression = dsl::IsNotNull<my_table::some_field>)]`
```
pksunkara, line 858:
```suggestion
///   For example: `#[diesel(select_expression_type = dsl::IsNotNull<my_table::some_field>)]`
```

diesel_derives/tests/selectable.rs
```diff
@@ -114,3 +114,28 @@ fn embedded_option_with_nullable_field() {
...
+        foo: i32,
+        #[diesel(
+            select_expression = my_structs::bar.is_not_null(),
+            select_expression_type = dsl::IsNotNull<my_structs::bar>
+        )]
+        bar_is_set: bool,
+    }
+
+    let conn = &mut connection();
+    let data = my_structs::table.select(A::as_select()).get_result(conn);
```
pksunkara, line 139: I don't understand the SQL query that's generated here. Is it something like `SELECT bar IS NOT NULL FROM my_structs;`?

Ten0: `SELECT foo, bar IS NOT NULL FROM my_structs`, as foo is also present in the struct.

pksunkara: Oh, so we are getting boolean instead of bar value. Makes sense.

pksunkara (approved):

weiznich (approved):
Looks great :+1:

PR #3246, opened 2022-07-22
Add `Timestamptz` support for the `time` crate v0.3, to Diesel's 2.0 branch

coriolinus:
As requested in https://github.com/diesel-rs/diesel/discussions/3241

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -35,7 +35,7 @@ r2d2 = { version = ">= 0.8.2, < 0.9.0", optional = true }
 itoa = { version = "1.0.0", optional = true }
 # newer versions of `time` have a MSRV of 1.57.0, which is newer than
 # diesel's own MSRV of 1.56.0
-time = { version = "=0.3.9", optional = true, features = ["macros"] }
+time = { version = "=0.3.9", optional = true, features = ["macros", "formatting", "parsing"] }
 
 [dependencies.diesel_derives]
 version = "~2.0.0-rc.0"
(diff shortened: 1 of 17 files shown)


weiznich:
I was just about to look into the test failures. Glad to see that this is already fixed. Seems to be a subtle difference in how time and chrono formates timestamps with millisecond resolution.

coriolinus:
Yes, chrono gives you a format flag `%.f` which emits subseconds only if non-zero, whereas with the time crate you have to emulate that manually. It's not a huge deal once I figured out what the issue actually was; most of the debugging involved was just in figuring out what was actually going wrong.

diesel/Cargo.toml
```diff
@@ -33,6 +33,9 @@ bigdecimal = { version = ">=0.0.13, < 0.4.0", optional = true }
 bitflags = { version = "1.2.0", optional = true }
 r2d2 = { version = ">= 0.8.2, < 0.9.0", optional = true }
 itoa = { version = "1.0.0", optional = true }
+# newer versions of `time` have a MSRV of 1.57.0, which is newer than
+# diesel's own MSRV of 1.56.0
+time = { version = "=0.3.9", optional = true, features = ["macros", "formatting", "parsing"] }
```
weiznich, line 38: It is fine to also allow newer time versions here, as we define our MSRV not by using the latest variant of all dependencies, but the oldest one. Or better: There must be a dependency version that builds fine with the given MSRV, we do not require that to be true for all dependency versions.

weiznich: Additionally we want to include `time` in the `extras` feature in line 52

weiznich (commented):
This looks good for me. I've noted a few minor things regarding to how the optional dependency is setup and tested. If those are fixed, this is fine for merging.

PR #3253, opened 2022-07-26
Reduce chrono features

Guara92:
Hi,

I know that 2.0 is near, but the upgrade is not easy and probabily for some time there will still be users of 1.4 so I done this in order to avoid the dependency from time 0.1, this should avoid a security problem and reduce the compile time.


weiznich:
We do currently not have the capacity to backport such things to older releases. I should probably add here that I do not see that as a security issue as diesel does not use any of the affected code at all, even if it might be contained in our dependencies.
If you are interested in avoiding this dependency you need to update to diesel 2.0. The corresponding fix is available there + there are already multiple existing (pre-)releases.

PR #3254, opened 2022-07-27
Fix #3252

### LINKED ISSUE #3252: begin_test_transaction behavior changed in Diesel 2.0
### Versions

- **Rust:** 1.62.1
- **Diesel:** 2.0.0-rc.1
- **Database:** PG 11
- **Operating System** macOS 12.4

### Feature Flags

- **diesel:**  ["postgres", "r2d2", "chrono"]

## Problem Description

In my 1.4.8 diesel usage I had a connection customizer that set the begin_test_transaction

Which I injected into my "test-only" connection pool which has a max-size of 1.

Then in my tests I was able to grab the DB connection perform inserts, then drop the connection and pass the pool to the code being tested (e.g. API requests) and it would use the SAME Connection & transaction block.

Happy testing with a single-connection pool.

The tests to run correctly.

The transaction is rolled back once the connection is returned to the pool and thus breaking the tests.  as the code being tests creates a new separate transaction.

### Steps to reproduce

Connection customizer to call begin_test_transaction

```rust
    #[derive(Debug, Clone, Copy)]
    struct TestConnectionCustomizer;

    impl<C, E> CustomizeConnection<C, E> for TestConnectionCustomizer
    where
        C: diesel::Connection,
    {
        fn on_acquire(&self, conn: &mut C) -> Result<(), E> {
            conn.begin_test_transaction()
                .expect("Failed to start test transaction");

            Ok(())
        }
    }
```

Sample test

```rust
#[tokio::test]
async fn my_test() {
    let pool  = test_pool();
    let mut db = pool.get().unwrap();
    let user = UserFactory::default().insert(&mut db);
    drop(db);

    let obj = ObjectToTest::new(&pool);
    obj.some_method(user.id);

    let mut db = pool.get().unwrap();
    let updated_user = users::table.filter(users::id.eq(user.id)).first(&mut db).unwrap();
    drop(db);

    assert_eq!(user.name, new_user.name);
}
```

This code works under Diesel 1.4.8 but does not in Diesel 2.0.0-rc.1

Here is raw SQL queries that were run

- Diesel 1.4.8

```
execute <unnamed>: SET TIME ZONE 'UTC'
execute <unnamed>: SET CLIENT_ENCODING TO 'UTF8'
statement: BEGIN
execute <unnamed>: SELECT 1
execute <unnamed>: INSERT INTO "users" ("name") VALUES ($1) RETURNING "users"."id", "users"."name", "users"."created_at", "users"."updated_at"
execute <unnamed>: SELECT 1
execute <unnamed>: UPDATE "users" SET "name" = $1 WHERE "users"."id" = $2 RETURNING "users"."id", "users"."name", "users"."created_at", "users"."updated_at"
execute <unnamed>: SELECT 1
execute __diesel_stmt_0: SELECT "users"."id", "users"."name", "users"."created_at", "users"."updated_at" FROM "users" WHERE "users"."id" = $1 LIMIT $2
-- implicit close & rollback
```

- Diesel 2.0.0-rc.1

```
execute <unnamed>: SET TIME ZONE 'UTC'
execute <unnamed>: SET CLIENT_ENCODING TO 'UTF8'
statement: BEGIN
execute <unnamed>: SELECT 1
execute <unnamed>: INSERT INTO "users" ("name") VALUES ($1) RETURNING "users"."id", "users"."name", "users"."created_at", "users"."updated_at"
-- implicit close & rollback
execute <unnamed>: SET TIME ZONE 'UTC'
execute <unnamed>: SET CLIENT_ENCODING TO 'UTF8'
statement: BEGIN
execute <unnamed>: SELECT 1
execute <unnamed>: UPDATE "users" SET "name" = $1 WHERE "users"."id" = $2 RETURNING "users"."id", "users"."name", "users"."created_at", "users"."updated_at"
-- implicit close & rollback
execute <unnamed>: SET TIME ZONE 'UTC'
execute <unnamed>: SET CLIENT_ENCODING TO 'UTF8'
statement: BEGIN
execute <unnamed>: SELECT 1
execute __diesel_stmt_0: SELECT "users"."id", "users"."name", "users"."created_at", "users"."updated_at" FROM "users" WHERE "users"."id" = $1 LIMIT $2
-- implicit close & rollback
```

## Checklist

weiznich:
This commit fixes an issue where test transactions caused connection to
be marked as broken when returned to a r2d2 connection pool. It is a
quite common pattern to create a connection pool with 1 connection, and
call `begin_test_transaction` on the connection before putting it into
the pool. This allows users to easily test applications using connection
pools without modifying an existing schema. The recent changes to our
transaction manager implementation + our r2d2 implementation broke this
pattern.

This commit introduces an additional flag inside of the transaction
manager state, which tracks whether `begin_test_transaction` was called
or not. This flag is later used to decide whether a connection should be
considered broken or not.

In addition this commit centralises the `is_broken`  implementation as
part of the transaction manager itself + marks
`transaction_manager_status_mut` as part of the unstable public
API (this method is only useful for third party connection or
transactionmanager implementations)

### DIFF
--- a/diesel/src/connection/mod.rs
+++ b/diesel/src/connection/mod.rs
@@ -299,7 +299,13 @@ where
             }
             TransactionManagerStatus::InError => panic!("Transaction manager in error"),
         };
-        Self::TransactionManager::begin_transaction(self)
+        Self::TransactionManager::begin_transaction(self)?;
+        // set the test transaction flag
+        // to pervent that this connection gets droped in connection pools
+        // Tests commonly set the poolsize to 1 and use `begin_test_transaction`
+        // to prevent modifications to the schema
+        Self::TransactionManager::transaction_manager_status_mut(self).set_test_transaction_flag();
+        Ok(())
     }
 
     /// Executes the given function inside a transaction, but does not commit
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -38,6 +38,9 @@ pub trait TransactionManager<Conn: Connection> {
     /// Used to ensure that `begin_test_transaction` is not called when already
     /// inside of a transaction, and that operations are not run in a `InError`
     /// transaction manager.
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
     fn transaction_manager_status_mut(conn: &mut Conn) -> &mut TransactionManagerStatus;
 
     /// Executes the given function inside of a database transaction
@@ -66,6 +69,35 @@ pub trait TransactionManager<Conn: Connection> {
             },
         }
     }
+
+    /// This methods checks if the connection manager is considered to be broken
+    /// by connection pool implementations
+    ///
+    /// A connection manager is considered to be broken by default if it either
+    /// contains an open transaction (because you don't want to have connections
+    /// with open transactions in your pool) or when the transaction manager is
+    /// in an error state.
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
+    fn is_broken_transaction_manager(conn: &mut Conn) -> bool {
+        match Self::transaction_manager_status_mut(conn).transaction_state() {
+            // all transactions are closed
+            // so we don't consider this connection broken
+            Ok(ValidTransactionManagerStatus {
+                in_transaction: None,
+            }) => false,
+            // The transaction manager is in an error state
+            // Therefore we consider this connection broken
+            Err(_) => true,
+            // The transaction manager contains a open transaction
+            // we do consider this connection broken
+            // if that transaction was not opened by `begin_test_transaction`
+            Ok(ValidTransactionManagerStatus {
+                in_transaction: Some(s),
+            }) => !s.test_transaction,
+        }
+    }
 }
 
 /// An implementation of `TransactionManager` which can be used for backends
@@ -77,6 +109,9 @@ pub struct AnsiTransactionManager {
 }
 
 /// Status of the transaction manager
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
 #[derive(Debug)]
 pub enum TransactionManagerStatus {
     /// Valid status, the manager can run operations
@@ -157,6 +192,15 @@ impl TransactionManagerStatus {
             TransactionManagerStatus::InError => Err(Error::BrokenTransactionManager),
         }
     }
+
+    pub(crate) fn set_test_transaction_flag(&mut self) {
+        if let TransactionManagerStatus::Valid(ValidTransactionManagerStatus {
+            in_transaction: Some(s),
+        }) = self
+        {
+            s.test_transaction = true;
+        }
+    }
 }
 
 /// Valid transaction status for the manager. Can return the current transaction depth
@@ -171,6 +215,7 @@ pub struct ValidTransactionManagerStatus {
 struct InTransactionStatus {
     transaction_depth: NonZeroU32,
     top_level_transaction_requires_rollback: bool,
+    test_transaction: bool,
 }
 
 impl ValidTransactionManagerStatus {
@@ -209,6 +254,7 @@ impl ValidTransactionManagerStatus {
                 self.in_transaction = Some(InTransactionStatus {
                     transaction_depth: NonZeroU32::new(1).expect("1 is non-zero"),
                     top_level_transaction_requires_rollback: false,
+                    test_transaction: false,
                 });
                 Ok(())
             }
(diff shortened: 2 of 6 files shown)


Ten0:
It's generally better to not complexify general code with test code so I would tend to not favor this approach if possible.

How about an approach where we would create a `TestTransactionConnection<T>` wrapper around any [`R2D2Connection`](https://docs.diesel.rs/master/diesel/r2d2/trait.R2D2Connection.html) that would also have `R2D2Connection` implemented but would automatically open a test transaction and override the `is_broken` behavior and prevent from touching transaction status (in particular prevent from committing)?

Or how about we propose an update to [`CustomizeConnection`](https://docs.diesel.rs/master/r2d2/trait.CustomizeConnection.html) so that it can hook on `is_valid` and/or (`take_from_pool` and `release_to_pool`), possibly opening transactions and rolling them back everytime they are taken from the pool/released?

Otherwise this looks correct. :)

weiznich:
To be clear about this upfront: I do not claim that the presented solution is the best way to solve this issue. Everything that follows is just trying to collect some potential problems with other suggested variants. 

The largest issue with the presented approach is as already pointed out the fact that it "infects" internal code with test-specific stuff.

> How about an approach where we would create a TestTransactionConnection<T> wrapper around any R2D2Connection that would also have R2D2Connection implemented but would automatically open a test transaction and override the is_broken behavior and prevent from touching transaction status (in particular prevent from committing)?

For this approach I see a major usability issue, as that requires users to use different pool/connection types based on whether they run tests or not. Technically that could be solved by accepting a `&mut impl LoadConnection<Backend = Pg>` or similar as function argument, but I assume that most existing code uses a fixed type at this position. Introducing a "special" connection type for this would require changing the type to either the generic variant (may affect compile times?) or conditionally on whether tests are run or not (hard to do in a multi-crate setup). The same argument more or less applies to the suggested alternative `ConnectionManager` impl specifically tailored to tests as suggested in the issue as well.

> Or how about we propose an update to CustomizeConnection so that it can hook on is_valid and/or (take_from_pool and release_to_pool), possibly opening transactions and rolling them back everytime they are taken from the pool/released

This might be an solution, but would require changes to upstream code. I'm not sure if they are open to such a change. Also what would an API look like, that would allow to conditionally disable the `has_broken` check?

weiznich:
> Hmm I think if we call the release_to_pool hook before checking for has_broken that would allow the caller to attempt closing any open transactions, releasing an effectively non-broken connection to the pool (also releasing locks on the DB and not polluting other tests) so we wouldn't need to really hook is_broken.

I'm not sure if that solves the problem outlined in #3252, as that specifically would require to leave the connection open even if it's checked in to the pool again.

Ten0:
> I'm not sure if that solves the problem outlined in #3252, as that specifically would require to leave the connection open even if it's checked in to the pool again.

I imagined that opening the transaction automatically and rolling it back automatically at every fetch/release from/to the pool would be fit for testing purposes.
@urkle wdyt?

jtgeibel:
> I imagined that opening the transaction automatically and rolling it back automatically at every fetch/release from/to the pool would also be fit for testing purposes.

I don't think that would work for our crates.io tests. Some of our tests require a pool with a single connection, where the transaction persists even if the connection is returned to the pool and then taken again elsewhere in the codebase. If the transaction is automatically rolled back, then our (test) background worker would never see the jobs enqueued earlier (in the now rolled back transaction).

weiznich:
@jtgeibel Thanks for reaching out here. That's what I had expected as well. Our internal code base uses a similar patterns. This PR would allow this pattern in the future, while other solutions likely require more changes.

That written: I definitively see the point that it would be great to have more hook points in r2d2. That could be implemented as well in parallel in my opinion as those features are independent.

urkle:
> I imagined that opening the transaction automatically and rolling it back automatically at every fetch/release from/to the pool would also be fit for testing purposes.

@Ten0 that was the start of opening the original bug report as Diesel 2.0 changed the behavior to roll it back when returning to the pool which breaks 99% of my tests as they are testing code that only can accept a Pool not a DB connection. Thus in my tests I need to 
1. grab a connection
2. create factories
3. release the connection (if I don't do this then the code under test cannot obtain the connection)
4. run the tests passing the cloned pool to the object under test
5. grab a connection
6. inspect the changed models (assert!s)
7. fin!

for examples
```rust
#[tokio::test]
async fn  updated_user() {
   let pool = test_pool();
   
   let db = pool.get().expect("failed to obtain DB connection");
   let user = UserFactory::default().insert(&db);
   drop(db);

   let schema = build_schema(&pool); // this method internally creates a new async-graphql schema for the object under test and binds a clone of the pool into it's context.
   let response = build_request(...).run(&schema).await;
  
   let db = pool.get().expect("failed to obtain DB connection");
   let record = users::table
           .filter(users::id.eq(user.id))
           .first::<User>(&db)
           .unwrap();
    drop(db);

     // assertions go here against user vs record etc..
}
```

weiznich:
@diesel-rs/reviewers Any final opinion here? Otherwise I suggest to merge this PR next week.

PR #3257, opened 2022-07-29
Allow custom backends to overwrite `QueryFragment` impls for `SelectStatement` and `BoxedSelectStatement`

weiznich:
This commit introduces the necessary associated type to `SqlDialect` so
that custom backends can provide custom `QueryFragment` implementations
for `SelectStatement` and `BoxedSelectStatement`. This is required as
some database systems require a different order of clauses there. (The
`firebird` backend is using a custom fork for this currently).
All existing backends in diesel will just use the existing
implementations.
In addition this commit extends the `__diesel_public_if` macro to
conditionally make struct fields public as well. As it's not possible to
apply proc-macro attributes directly to struct fields, this is solved by
having an explicit list of "public" fields in the macro call on the
struct itself. Such structs are also marked as `#[non_exhaustive]` so
that we can add additional fields later on.

See https://github.com/fernandobatels/rsfbclient/pull/127 for a possible use case


weiznich:
Nightly builds fail due to https://github.com/rust-lang/rust/issues/100620

PR #3258, opened 2022-07-29
Support tuples in custom select expressions in selectable

Ten0:
That didn't work before (see test).


Ten0:
It looks like I don't have enough expertise on `syn` to solve this last part quickly indeed, so that would be of great help! 😃

My beginning of investigation seems to show that `syn` when [`Parse`](https://docs.rs/syn/latest/syn/parse/trait.Parse.html)ing `(my_structs::bar.is_not_null(), my_structs::some_int),` into an [`Expr`](https://docs.rs/syn/latest/syn/enum.Expr.html), first identifies it as a [`Paren`](https://docs.rs/syn/latest/syn/enum.Expr.html#variant.Paren) instead of [`Tuple`](https://docs.rs/syn/latest/syn/enum.Expr.html#variant.Tuple) and then fails to parse the thing inside as an expression `my_structs::bar.is_not_null(), my_structs::some_int` (as indeed that is not an expression).
But then that's `syn`'s parse so I'm not sure how I can affect that...

Thanks! 🙏

weiznich:
I would likely try to parse it directly as `ExprTuple` first. Use the result if it succeeds, otherwise fallback to parsing it via `Expr` and error out if that fails too.

Ten0:
> I would likely try to parse it directly as `ExprTuple` first. Use the result if it succeeds, otherwise fallback to parsing it via `Expr` and error out if that fails too.

So fun fact I just tried that, which required enabling the `full` syn feature so that the `Parse` impl would exist for that type (https://docs.rs/syn/1.0.98/src/syn/expr.rs.html#2280-2282), and it worked, but then I noticed when looking at that parse impl that it actually delegated to `Expr as Parse`... Turns out all that was needed was to enable the feature named `full` ^^'

weiznich:
Good that the fix was that easy. That written: I'm not that happy with requiring the "full" feature for syn, as that increases compile times quite a bit and at least for my projects the dependency chain involving syn -> diesel_derives -> diesel is already the limiting factor. That means I would really prefer a solution that works without this feature flag.

Ten0:
Hmm maybe there would be a way to achieve that by actually not parsing the expression there, but I think ATM we're blocked for this by the fact that the input contains the current ~attribute and the next one (see my dbg output above).
Somehow calling parse on it only parses the first expression and does not use anything after:
```
Punct {
        ch: ',',
        spacing: Alone,
        span: #0 bytes(46879..46880),
    },
    Ident {
        ident: "select_expression_type",
        span: #0 bytes(46893..46915),
    },
    Punct {
        ch: '=',
        spacing: Alone,
        span: #0 bytes(46916..46917),
    },
```
and in addition the next calls actually process this properly, but I have no idea how this works.

Maybe if we cut this better we can avoid parsing it as an expr and just keep it as a tokens stream and insert that in the generated code? (Errors wouldn't be as good then though.)

NB: I find syn's compile time increase there to be pretty low: 0.4s on my machine. That seems completely neglictable on any CI run (and doesn't really matter when developing as this is only compiled once).

weiznich:
I've pushed https://github.com/weiznich/diesel/commit/16b7c4f13e6210d8ace7e2fc9ba804aca04b1e5b to demonstrate how this could be done without `feature = "full"`. That commit also adds some tests for error messages.

diesel_derives/src/field.rs
```diff
@@ -145,11 +144,52 @@ pub enum FieldName {
...
         match *self {
             FieldName::Named(ref x) => x.to_tokens(tokens),
             FieldName::Unnamed(ref x) => x.to_tokens(tokens),
         }
     }
 }
+
+/// We use this instead of directly `syn::Expr` to reduce compilation time
+///
+/// `syn::Expr` does not properly support tuples when `syn/full` feature is enabled,
```
weiznich, line 158:
```suggestion
/// `syn::Expr` does not properly support tuples when `syn/full` feature is not enabled,
```

PR #3274, opened 2022-08-14
Move the CONCAT change out of 2.0.0 Rc1 changelog

thomasmost:
Just a small change since I realized I put the changelog update under the wrong release, technically


thomasmost:
Thanks 🙏

PR #3277, opened 2022-08-17
Implement Display for PgMoney

crlsktr:
Serialization via serde is difficult due to the inability to implement traits for diesel types in projects that use diesel.
Alternatively, providing the Display trait in the package would allow serialization of the type PgMoney/Cents by using serde_with DisplayFromStr.

### DIFF
--- a/diesel/src/pg/types/money.rs
+++ b/diesel/src/pg/types/money.rs
@@ -38,7 +38,7 @@ impl ToSql<Money, Pg> for PgMoney {
 }
 
 impl fmt::Display for PgMoney {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 write!(f, "{}", self.0)
     }
 }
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. Both `PgMoney` and `Cents` are designed as intermediate types, which can be used to implement more advanced types on top of them in a third party crate. Based on this I would like not to add more functionality to those types, especially as the inner value is publicly accessible.

For the serde specific use-case I see 3 options:
* Use serde's [`#[serde(serialize_with = "…")]`](https://serde.rs/field-attrs.html#serialize_with) and provide the serialization there
* Use diesel's [`#[diesel(deserialize_as = "…")]`](https://docs.diesel.rs/1.4.x/diesel/deserialize/trait.Queryable.html#deriving)  to deserialize into a type that implements the corresponding serde traits
* Provide your own type implementation on top of `PgMoney` by implementing `FromSql` + deriving `FromSqlRow`.

Neither of the solutions is hard to implement in my opinion.

crlsktr:
That makes sense thank you for taking the time.

PR #3280, opened 2022-08-20
Shell script fixes

Heliozoa:
I had issues running bin/test due to having diesel in a path that contains whitespace and some variables not being quoted, so the scripts tried to cd to `some/path/with` where diesel was at `some/path/with whitespace/diesel` in a couple places.

Instead of figuring out each spot where this be an issue I just ran the scripts through https://www.shellcheck.net/ and fixed any issues that popped up. While I was at it I also made the indentation consistent at 4 spaces. I can make a more minimal PR or one that has everything indented with 2 spaces if that's preferred though.


Heliozoa:
Oops, things worked fine locally so I assumed the changes were OK, but looks like they cause issues during CI...

JohnTitor (approved):

Issue #2005, opened 2019-03-06
Emitted schema could use comments from database


eropple:
So I'm new to Diesel, and I apologize if this has already been suggested, but I couldn't find it after some searching. :)

One of the things about Diesel that I really like--as somebody who is generally suspicious of generated code--is how clear and straightforward the generated `schema.rs` file is. However, there's one thing that's bugging me: any comments in the database are ignored. Comments are not in the SQL standard, but both Postgres and MySQL support some flavor of them (and Postgres's can be [deep](https://www.postgresql.org/docs/9.1/sql-comment.html)!). Historically I've always been a big fan of their use, but I'm also the sort of developer who always has a Postgres console open anyway. It'd be handy purely as reference, and might encourage a little bit more clarity, to surface these somehow in `schema.rs`. I can see a couple ways this could be useful right off the top of my head:

- tagging tables and columns for display in documentation tools (both Postgres and MySQL, at least, support column and table comments)
- surfacing comments on constraints/unique indexes (Postgres supports these) when handling a constraint violation error

Obviously this is not a super-high priority item (and I might give it a try myself as I get more comfortable with Rust, I am a novice at this) but given the way Diesel seems happy to get comfortable with Postgres features--and others I'm sure, but I'm a Postgres user by default--this could be a handy way to help treat the database as the arbiter of truth and better sync up one's data layer with one's schema.

Anyway, Diesel seems awesome so far. Thanks!

weiznich:
Thats implemented in https://github.com/diesel-rs/diesel/pull/3281

PR #3281, opened 2022-08-20
Turn postgres/mysql table/column comments into Rust doc comments

Heliozoa:
I took a stab at implementing the easy part of #2005.

Sqlite doesn't seem to have (proper) support for table/column comments, so the functionality is added only for postgres and mysql.

As implemented the comments are always queried from the db but only written with `--with-docs`. The interaction with the autogenerated docs such that it will prefer the schema comment if one exists and use the automatic one otherwise. Might be better to make this more configurable?

I renamed `UsesInformationSchema` -> `DefaultSchema` and removed `type_column`/`type_schema` as those were only used in `get_table_data` which is no longer generic, but `default_schema` was still in use elsewhere.

I wasn't able to figure out how to make `null_if_text` more generic, but maybe it's fine to keep it simple since it's only used in one place here?

I'm still not too familiar with diesel's code so there might be some oddities in here, but hopefully this is at least in the right direction.


diesel_cli/src/infer_schema_internals/mysql.rs
```diff
@@ -173,3 +256,85 @@ fn unsigned_types_are_supported() {
...
+            get_table_data(&mut connection, &table_1, &ColumnSorting::OrdinalPosition)
+        );
+        assert_eq!(
+            Ok(vec![id_without_comment]),
+            get_table_data(&mut connection, &table_2, &ColumnSorting::OrdinalPosition)
+        );
+    }
+
+    #[test]
+    #[ignore = "FIXME: Figure out how to handle tests that modify schema"]
```
weiznich, line 316: This can be solved by using a temporary table: https://dev.mysql.com/doc/refman/8.0/en/create-temporary-table.html

weiznich (commented):
Thanks for opening this PR. This looks already quite good. I've added some notes on minor details that should be improved before merging this. Otherwise this feature should have an entry in our changelog.

weiznich:
Nightly CI failures are caused by https://github.com/rust-lang/rust/issues/100620

diesel_cli/src/infer_schema_internals/mysql.rs
Heliozoa: This doesn't seem to work unfortunately. It appears that info on temporary tables is not stored in `information_schema.tables` where the comment information is fetched from but in [information_schema.innodb_temp_table_info](https://dev.mysql.com/doc/mysql-infoschema-excerpt/8.0/en/information-schema-innodb-temp-table-info-table.html) which doesn't provide the comment info.

Would it be acceptable to do something like
```sql
DROP TABLE IF EXISTS diesel_test.get_table_data_loads_column_information; # named after the test to avoid conflicts with other tests
CREATE TABLE diesel_test.get_table_data_loads_column_information # ...
```
optionally with another drop table at the end as cleanup?

weiznich: Yes that would be also fine. There are other tests in diesel_cli that modify the schema as well so I do not see any reason for not doing that here as well.

Heliozoa:
Weird test failures on the mysql tests on macos, for some reason the column data for `id SERIAL ...` comes back with the type `bigint(20) unsigned` where it's `bigint unsigned` on other platforms...apparently this is not a meaningful difference but it caused the comparison to fail nevertheless. I tried changing the type to `BIGINT(20)` hoping to get `bigint(20) unsigned` on all platforms but that didn't work. I opted to just use something else as the id instead (`VARCHAR(255)`).

weiznich (approved):
Looks good to me :+1: Will merge this next week.

PR #3288, opened 2022-08-23
Speedup compile times for joins

weiznich:
Addresses #3223


diesel/src/query_source/joins.rs
```diff
@@ -106,41 +106,59 @@ impl<Left, Right> QuerySource for Join<Left, Right, Inner>
...
     type FromClause = Self;
-    type DefaultSelection = Left::Output;
+    // combining two valid selectable expressions for both tables will always yield a
+    // valid selectable expressions for the whole join, so no need to check that here
+    // again. These checked turned out to be quite expensive in terms of compile time
+    // so we use a wrapper type to just skip the check and forward other more relevant
+    // trait implementations to the inner type
+    //
+    // See https://github.com/diesel-rs/diesel/issues/3223 for details
+    type DefaultSelection = self::private::SkipSelectableExpressionWrapper<Left::Output>;
```
Mingun, line 120: Interesting... How did you trick the compiler into disabling check? For me, this looks like an insoluble task

weiznich: The relevant line is here: https://github.com/diesel-rs/diesel/pull/3288/files#diff-d0e8a449e498610baa9a5f3086200726d02bd5e1612f5f1d5587c4418baddf0eR482
This impl does not check that the trait holds true for the inner type, so it effectively skips the recursive trait bound check on large tuples, which is the underlying problem.

Mingun: You mean, that the most important thing there that `T` is not bounded, so the compiler does not do expensive recursive checks as it did before?

weiznich: Yes that's the main optimization.

Ten0 (approved):
Apart from these minor suggestions, LGTM for merge :)

diesel/src/query_source/joins.rs
```diff
@@ -106,41 +106,59 @@ impl<Left, Right> QuerySource for Join<Left, Right, Inner>
...
+        )
     }
 }
 
 impl<Left, Right> QuerySource for Join<Left, Right, LeftOuter>
 where
     Left: QuerySource + AppendSelection<Nullable<Right::DefaultSelection>>,
     Right: QuerySource,
-    Left::Output: SelectableExpression<Self>,
+    Left::Output: AppearsOnTable<Self>,
```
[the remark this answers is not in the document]
weiznich: Seems like the closest test is this one for aliases: https://github.com/diesel-rs/diesel/blob/596cdd21cc4aa4cefeccdcb3988c14f9a01ed6de/diesel_compile_tests/tests/fail/derive/aliases.rs

I will add some compile fail tests for invalid joins as well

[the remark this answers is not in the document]
weiznich: https://github.com/diesel-rs/diesel/pull/3288/commits/60594b4ff7e42e53da74d149e168d5e0a0345743

Ten0 (approved):

Issue #1306, opened 2017-11-16
foo.is_null() is not the same as foo.eq(None)


icefoxen:
## Setup

I have a model something like this:

```sql
CREATE TABLE things (
   id INTEGER PRIMARY KEY NOT NULL,
   data INTEGER NOT NULL,
   optional_data VARCHAR
);
```

```rust
pub struct Thing {
    pub id: i32,
    pub data: i32,
    pub optional_data: Option<String>,
}
```

### Versions

- **Rust:** 1.21
- **Diesel:** 0.16
- **Database:** SQLite3 (haven't tried on Postgres)
- **Operating System** Linux

## Problem Description

I want to get all the `Thing`s with a particular data value and no optional data, so I wrote:

```rust
transactions.filter(                                                                                              
            data.eq(target_data)
                .and(optional_data.eq(None as Option<String>)                                                 
        )
```

But, it never succeeds even when there is data I would expect to match.  Instead, to do this query I have to write:

```rust
transactions.filter(                                                                                              
            data.eq(target_data)
                .and(optional_data.is_null())                                                 
        )
```

It does appear that this is perfectly consistent with SQL (though if there's implementation-defined variations here I don't know anything about it), since SQL NULL is not equal to any value: https://stackoverflow.com/questions/1833949/why-is-null-not-equal-to-null-false#1833989

But, it is extremely inconsistent with Rust and its representation of `Option` for "missing data".  SQL says NULL is never equal to anything including itself, similar to floating point NaN's.  But Rust says that an `Option<T>::None` is equal to any other `Option<T>::None`.  I sort of feel that if you're going to use SQL semantics, you should define a separate type than `Option` with explicitly different behavior for this, so it doesn't look like it SHOULD obey Rust semantics.  Given that `optional_data.eq(None)` is NEVER true, it seems like it should be impossible to say.

Thank you!

sgrif:
Making this change would mean that `Option<T>` no longer generates consistent SQL, which would break our prepared statement caching. Ultimately I don't think it's a bad thing for Diesel to be very close to SQL, which is what we always err towards. Note that we do support the PostgreSQL `IS NOT DISTINCT FROM` operator with [`.is_not_distinct_from`](http://docs.diesel.rs/diesel/expression_methods/trait.PgExpressionMethods.html#method.is_not_distinct_from), which has the semantics you want.

joelgallant:
[...]

PR #3294, opened 2022-08-26
Explain None/null convention in `column.eq` docstring

sebasv:
After running into the same issue as described in #1306 I thought I'd add to the docstring explaining that the `.eq` method follows SQL semantics, not Rust semantics when it comes to `None`.

### DIFF
--- a/diesel/src/expression_methods/global_expression_methods.rs
+++ b/diesel/src/expression_methods/global_expression_methods.rs
@@ -10,7 +10,23 @@ pub trait ExpressionMethods: Expression + Sized {
     /// Creates a SQL `=` expression.
     ///
     /// Note that this function follows SQL semantics around `None`/`null` values,
-    /// so `column.eq(None)` will never match. Use [`column.is_null()`](ExpressionMethods::is_null()) instead.
+    /// so `eq(None)` will never match. Use [`is_null`](ExpressionMethods::is_null()) instead.
+    ///
+    ///
+    #[cfg_attr(
+        any(feature = "sqlite", feature = "postgres"),
+        doc = "To get behavior that is more like the Rust `=` operator you can also use the"
+    )]
+    #[cfg_attr(
+        feature = "sqlite",
+        doc = "sqlite-specific [`is`](crate::SqliteExpressionMethods::is())"
+    )]
+    #[cfg_attr(all(feature = "sqlite", feature = "postgres"), doc = "or the")]
+    #[cfg_attr(
+        feature = "postgres",
+        doc = "postgres-specific [`is_not_distinct_from`](crate::PgExpressionMethods::is_not_distinct_from())"
+    )]
+    #[cfg_attr(any(feature = "sqlite", feature = "postgres"), doc = ".")]
     ///
     /// # Example
     ///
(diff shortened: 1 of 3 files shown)


diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -9,6 +9,9 @@ use crate::sql_types::{SingleValue, SqlType};
 pub trait ExpressionMethods: Expression + Sized {
     /// Creates a SQL `=` expression.
     ///
+    /// Note that this function follows SQL semantics around `None`/`null` values,
```
weiznich, line 12: It might be useful to link the sqlite specific [`IS`](https://docs.diesel.rs/master/diesel/expression_methods/trait.SqliteExpressionMethods.html#method.is) operator and the postgres specific [`IS NOT DISTINCT FROM`](https://docs.diesel.rs/master/diesel/expression_methods/trait.PgExpressionMethods.html#method.is_not_distinct_from) here, as both are more in line with rusts `=` operator.

sebasv: Great suggestion!

weiznich:
Thanks for the update. However the CI errors look valid and need to be resolved before merging this.

sebasv:
@weiznich the doc builds fail when a limited set of features are included because the docs now reference sqlite-specific and pg-specific functions in the shared codebase. 

I can try to solve this, but it is not clear to me why the docs are being built with features disabled. Shouldn't docs always reflect all features? If you agree I will de-parametrize the features from the doc build. Otherwise I would love a suggestion on how to solve this, I expected `#[cfg(doc)]` to do the trick but apparently that does not work with features.

weiznich:
I would prefer allowing to keep that documentation backend specific as not all users build diesel with all backends enabled and that will lead to dead links otherwise. See here for examples of conditional documentation: 
https://github.com/diesel-rs/diesel/blob/cb185429d1ed8550f25615e84c14534692da8bda/diesel/src/backend.rs#L145

sebasv:
Sorry if this is an open door, but any reason why Diesel doesn't use cached environments? Seems a bit of a waste of expensive runner minutes to re-install three databases, N times over for each push to a branch with an open PR

sebasv:
@weiznich I've pushed a change that resolves the failed trybuild on my local machine but I don't know if you want this fix to be in this PR, because the problem seems unrelated to my change

weiznich:
> Sorry if this is an open door, but any reason why Diesel doesn't use cached environments? Seems a bit of a waste of expensive runner minutes to re-install three databases, N times over for each push to a branch with an open PR

Installing the database takes normally less than 1 minute, while compiling and running all tests takes normally at least 10 minutes. So yes having some cached environment would likely help a bit here, but it will not bring any great speedup, especially as runner minutes are "free" for open source projects. The hard truth here is that nobody put the time in to fix this correctly. At least I have been happy that it works at all and does not require much maintenance. That all written: We would likely accept PR's improving our CI setup. 

> I've pushed a change that resolves the failed trybuild on my local machine but I don't know if you want this fix to be in this PR, because the problem seems unrelated to my change

These build failures are surprising as they do not happen on other branches. Anyway it's fine to include this fix here.

PR #3297, opened 2022-08-30
Add QueryId derivation for diesel_cli generated custom sql types

alordash:
This PR adds ``#[derive(..., diesel::query_builder::QueryId)]`` for custom sql types, generated with diesel cli.


Ten0:
What is this for?

alordash:
To allow using custom sql types in ``WHERE`` clauses. [This example](https://github.com/diesel-rs/diesel/tree/master/examples/postgres/custom_types) won't work if we add filters to this [load query](https://github.com/diesel-rs/diesel/blob/cb185429d1ed8550f25615e84c14534692da8bda/examples/postgres/custom_types/src/main.rs#L28-L35):
```rs
let t = translations::table
    .select((
        translations::word_id,
        translations::translation_id,
        translations::language,
    ))
    .filter(translations::language.eq(Language::En))    // error: QueryId is not implemented for schema::sql_types::Language
    .get_results::<Translation>(conn)
    .expect("select");
```

Ten0:
Ah I see. As it turns out these are [implemented by the macro](https://github.com/adwhit/diesel-derive-enum/blob/110d36b2ec119df34b519ec5d6b25571bc64a55d/src/lib.rs#L284) if you use diesel_derive_enum. (that's probably my code ^^')

Ten0 (approved):

weiznich:
@Ten0 Might be worth to plan a 2.0.1 patch release for diesel_cli in a few weeks including this "fix"?

Ten0:
That looks like technically a breaking change but if we do it asap and synchronize on the release with one on diesel_derive_enum it looks like impact should only be on devs and not CIs and people who manually implement stuff on these types.

Ten0:
Ah it looks like diesel-derive-enum is missing a release though so people are probably using the master of that 🤔

alordash:
Perhaps this feature could be provided via new setting in diesel_cli config file?

weiznich:
Using the config file for this would be the backward compatible solution. We should choose that way and not pretend that noone would implement `QueryId` manually. I will revert this PR for now even if that means that @alordash work did not help here.

For a complete solution I would imaging something like follows:

* Add a new key to our `diesel.toml` file for configuring additional derives 
* Change the default [`diesel.toml`](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/default_files/diesel.toml) to include  the newly created config key + default to include `diesel::query_builder::QueryId` (That way new projects automatically get the correct derive there)
* Change the code to include any derive listed in the new config key
* Add a few tests
* (Maybe add a CLI flag for this as well)

alordash:
I will try to add new config setting. I think this feature could eventually be used to allow users to attach all kinds of trait derivations by listing them (like it was done in ``import_types`` setting).

SanchithHegde:
@alordash Have you started any work on the suggested approach? If not, I'd like to give it a try.

alordash:
@SanchithHegde Hi, unfortunately no, I am a little bit busy these days. I don't mind if you try to implement this feature.

SanchithHegde:
@alordash No problem, I'll take this up.

PR #3308, opened 2022-09-06
diesel_cli: Add config option for custom type derives

SanchithHegde:
Please let me know if anything has to be changed. Open to suggestions on a better name for the config option and any more test cases that can be added.

Supersedes #3297 and addresses #3306.

EDIT: The [Configuring Diesel CLI](https://github.com/sgrif/diesel.rs-website/blob/01755d528c5aa7343bdc58e04e529543e7f8b3f7/src/guides/configuring-diesel-cli.md) guide is yet to be updated.


diesel_cli/src/print_schema.rs
```diff
@@ -230,13 +231,14 @@ pub fn output_schema(
     Ok(out)
 }
 
-struct CustomTypeList {
+struct CustomTypeList<'a> {
     backend: Backend,
     types_sorted: Vec<ColumnType>,
     with_docs: bool,
+    derives: &'a [String],
```
SanchithHegde, line 238: Should this be a `Vec<String>` instead?

diesel_cli/src/cli.rs
```diff
@@ -229,6 +229,15 @@ pub fn build_cli() -> Command<'static> {
...
                 .help("Generate SQL type definitions for types not provided by diesel"),
+        )
+        .arg(
+            Arg::new("custom-type-derives")
+                .long("custom-type-derives")
+                .takes_value(true)
+                .multiple_values(true)
+                .action(clap::ArgAction::Append)
+                .number_of_values(1)
+                .help("A list of derives to implement for every custom type, separated by commas."),
```
Ten0, line 240:
```suggestion
                  .help("A list of derives to implement for every automatically generated SqlType in the schema, separated by commas."),
```

diesel_cli/src/print_schema.rs
```diff
@@ -275,7 +279,14 @@ impl Display for CustomTypeList {
                         writeln!(out, "///")?;
                         writeln!(out, "/// (Automatically generated by Diesel.)")?;
                     }
-                    writeln!(out, "#[derive(diesel::sql_types::SqlType)]")?;
+                    if self.derives.is_empty() {
+                        writeln!(
+                            out,
+                            "#[derive(diesel::sql_types::SqlType, diesel::query_builder::QueryId)]"
+                        )?;
```
Ten0, line 286: Didn't we say it was breaking if we were doing `QueryId` by default and that was why we wanted to have a config that people would instead update manually?

diesel_cli/src/print_schema.rs
```diff
@@ -275,7 +279,14 @@ impl Display for CustomTypeList {
...
+                    } else {
+                        writeln!(out, "#[derive({})]", self.derives.join(", "))?;
```
Ten0, line 288: I think it would be reasonable to always put the `SqlType` derive by default. I can't think of a scenario where people would want to not put it.

Ten0 (changes requested):
Other than that lgtm

SanchithHegde:
@Ten0 I've addressed the review comments and fixed tests.

I'm unsure about the tests though: they pass, and I only had to update the `diesel.toml` files and not the CLI args, so I'm confused as to how tests pass.

diesel_cli/src/print_schema.rs
```diff
@@ -280,12 +280,13 @@ impl Display for CustomTypeList {
                         writeln!(out, "/// (Automatically generated by Diesel.)")?;
                     }
                     if self.derives.is_empty() {
+                        writeln!(out, "#[derive(diesel::sql_types::SqlType)]")?;
```
Mingun, line 283: I think, it would be cleaner if `self.derives` would be augmented with `diesel::sql_types::SqlType` if that type not already presented there

diesel_cli/tests/print_schema/print_schema_custom_types_overriding_derives_works/diesel.toml
```diff
@@ -0,0 +1,4 @@
+[print_schema]
+file = "src/schema.rs"
+with_docs = true
+custom_type_derives = ["diesel::sql_types::SqlType", "core::fmt::Debug"]
```
Ten0, line 4: > I'm unsure about the tests though: they pass, and I only had to update the `diesel.toml` files and not the CLI args, so I'm confused as to how tests pass.

Isn't it that they use both the corresponding diesel.toml and CLI args? (So in order to see that CLI args work we could remove these and see whether the args specified at https://github.com/diesel-rs/diesel/pull/3308/files#diff-4f7007ed0625d981d1a3230723b3a57050fe8909d83510011656a2693c5304afR225-R227 hit in such a way that the output doesn't change?)

weiznich (commented):
Looks good from my side. I would like to see an additional test for the command line flags (see @Ten0 comment about this) + a Changelog entry.

diesel_cli/tests/print_schema/print_schema_custom_types_overriding_derives_works/diesel.toml
SanchithHegde: > Isn't it that they use both the corresponding diesel.toml and CLI args? (So in order to see that CLI args work we could remove these and see whether the args specified at [#3308 (files)](https://github.com/diesel-rs/diesel/pull/3308/files#diff-4f7007ed0625d981d1a3230723b3a57050fe8909d83510011656a2693c5304afR225-R227) hit in such a way that the output doesn't change?)

I was able to trace and figure out how the tests are run: first the CLI args are used as configuration for generating the schema, and if it generates the expected output, the configuration from the corresponding `diesel.toml` is used for generating the schema. When I remove this line from the `diesel.toml` file but keep the CLI args as is:

```toml
[print_schema]
# ...
custom_type_derives = ["diesel::sql_types::SqlType", "core::fmt::Debug"]
```

the test as a whole fails (as expected), since using the CLI args as config generates the expected output, but using the `diesel.toml` file doesn't include `Debug` in the list of derives. So, please ignore my (outdated) comment about me not being able to understand how the tests passed.

@weiznich This test verifies that overrides work for both the config file and CLI args. Please let me know if you'd like me to include a test for only the CLI args, or if any additional tests are necessary. I will be including a changelog entry soon.

weiznich: :+1: That's fine for me. We don't need an additional test then.

weiznich (approved):

Ten0 (approved):

51yu:
wondering when we publish cli binary (2.0.2) so that `custom-type-derives` option is available ? : )

weiznich:
@51yu We generally do not provide ETAs for the diesel release. Please don't ask these questions.

SanchithHegde:
Is there anything blocking this from being included in an upcoming release? Would you prefer that I update the "Configuring Diesel CLI" guide?

weiznich:
@SanchithHegde There is nothing blocking this to be included in the next feature release. It's planed to include that. It's just that other parts are not ready for the release yet. Opening a PR to update the configuring "Configuring Diesel CLI" guide will at least be helpful.

SanchithHegde:
> Opening a PR to update the configuring "Configuring Diesel CLI" guide will at least be helpful.

Sure, I'll do that.

PR #3327, opened 2022-09-19
Fix "Wrong postgres_type naming"

### LINKED ISSUE #3314: Diesel 2.0: Wrong `postgres_type` naming
### Versions
- **Rust:** Latest
- **Diesel:** 2.0
- **Database:** PostgreSQL
- **Operating System:** Linux

### Feature Flags
- **diesel:** postgres

## Problem Description
When there is a custom type, which it's name contains capital letters. The generated name in the `schema.rs` file will be in lower case, like the following example:

```
pub mod sql_types {
    #[derive(diesel::sql_types::SqlType)]
    #[diesel(postgres_type(name = "poststatus"))]
    pub struct PostStatus;
}
```

In this case, inserting a row (it needs to contain the custom type), will return an error like this:
```
thread 'main' panicked at 'Error saving new post: SerializationError(FailedToLookupTypeError(PgMetadataCacheKey { schema: None, type_name: "poststatus" }))', src/lib.rs:30:10
```

postgres also reports this:
```
ERROR:  type "poststatus" does not exist
STATEMENT:  SELECT "pg_type"."oid", "pg_type"."typarray" FROM "pg_type" WHERE ("pg_type"."oid" = quote_ident($1)::regtype::oid) LIMIT $2
```

The same can happen for other queries, like select by the custom type, or update or so on.

### Steps to reproduce
- Clone this sample project: https://github.com/omid/diesel_demo
- Run postgresql
- Configure `.env` file
- Run `make`, then you will see the error
- It will work if you go to `schema.rs` and change `poststatus` to `PostStatus`, and run `cargo run --bin write_post`

omid:
Fixes #3314


diesel_cli/tests/print_schema.rs
```diff
@@ -229,6 +229,12 @@ fn print_schema_sqlite_implicit_foreign_key_reference() {
     test_print_schema("print_schema_sqlite_implicit_foreign_key_reference", vec![]);
 }
 
+#[test]
+#[cfg(feature = "postgres")]
+fn print_schema_respect_type_name_case() {
+    test_print_schema("print_schema_respect_type_name_case", vec!["--with-docs"])
```
Ten0, line 235:
```suggestion
fn print_schema_respects_type_name_case() {
    test_print_schema("print_schema_respects_type_name_case", vec!["--with-docs"])
```

and of course corresponding file names

Ten0 (approved):

weiznich:
Thanks :+1:

PR #3331, opened 2022-09-21
Make `InsertStatement` fields public behind feature flag so that third party backends can use these information more easily

weiznich:
This is required for third party backends to implement similar batch inserts in a similar way than sqlite 

See https://github.com/GiGainfosystems/diesel-oci/pull/45/commits/f2d117c92ca8e2a517611b80761876a35f55ede2 for details.


JohnTitor (approved):
Looks reasonable to me.

PR #3341, opened 2022-09-25
Verify that group_by expression is valid for the given table

weiznich:
This commit fixes a bug where diesel would accept group by expressions that contain columns of tables not even appearing in the current from clause. The provided compile test contains some examples that would have been accepted before this commit (not all of them where accepted before, only those with `.select(…).group_by(…)`).

This is technically a breaking change as this rejects previously accepted queries. However I would argue that this is at the same time a critical bugfix as this turns invalid queries in a compiler error.

I explicitly request a review from @diesel-rs/core here how to handle this fix. I propose that we call this a bug fix even if it's an breaking change (those are allowed in the semver spec). I would assume that this either does not break working code (as the generated queries where broken anyway) or at least point the code author at potential issues with their code.


weiznich:
Compile tests fail as the pushed output assumes https://github.com/diesel-rs/diesel/pull/3340 is already merged.

sgrif (approved):

sgrif:
Calling this a bugfix seems fine to me

Issue #3330, opened 2022-09-21
Upsert with where clause Postgres implementation is wrong


bowenyang007:
## Problem Description

Using `on_conflict(...).with_target(...)` is supposed to limit the rows updated during an upsert, but the implementation is slightly wrong. As a result no row is actually being limited. (code here: https://github.com/diesel-rs/diesel/blob/v2.0.0/diesel/src/query_builder/upsert/on_conflict_target_decorations.rs#L19)

Note, the current implementation actually solves a different problem around index predicate but I don't believe that it was the intention of the original issue (https://github.com/diesel-rs/diesel/issues/2430). 

## Details
For example, looking at the test file (https://github.com/diesel-rs/diesel/blob/master/diesel_tests/tests/debug/mod.rs#L158), we would potentially get back an sql statement that doesn't work as intended. 
```
INSERT INTO "users" ("name", "hair_color") VALUES ($1, $2), ($3, $4) ON CONFLICT ("hair_color") WHERE "users"."hair_color" = $5 DO NOTHING
-- binds ...
```

The `WHERE` clause should actually be moved to the end, otherwise it does nothing. 

For example, according to the postgres documentation (https://www.postgresql.org/docs/current/sql-insert.html), `WHERE` needs to happen after `DO UPDATE SET...`. 
```
-- Don't update existing distributors based in a certain ZIP code
INSERT INTO distributors AS d (did, dname) VALUES (8, 'Anvil Distribution')
    ON CONFLICT (did) DO UPDATE
    SET dname = EXCLUDED.dname || ' (formerly ' || d.dname || ')'
    WHERE d.zipcode <> '21201';
```

## Repro
I tested in a postgresql DB with a trivial example: 
```
create table test(
    name text primary key,
    version int    
);

insert into test values ('a', 5)

-- this works as intended, row is not updated
insert into test ("name", "version") values ('a', 3) on conflict ("name") 
do update set "version" = excluded."version" where test."version" < 1;


-- moving where right after on conflict is wrong. row will be updated regardless what the condition is
insert into test ("name", "version") values ('a', 3) on conflict ("name") where test."version" < 1
do update set "version" = excluded."version"

weiznich:
Thanks for the bug report. To clarify that: This was correct with the 1.4 release? Or was this always broken?

The relevant impl is here:

[...]

PR #3344, opened 2022-09-26
Clear db connection after database error

### LINKED ISSUE #3342: DatabaseError "another command is already in progress\n" upon using the same db connection multiple times
### Versions

- **Rust:** 1.63.0
- **Diesel:** v2.0.0
- **Database:** PostgreSQL
- **Operating System** Linux

### Feature Flags

- **diesel:** ["postgres", "chrono"]

## Problem Description
It looks like while making two queries using the same connection (not concurrently), from time to time, first query left the connection (or channel to the db) in "not usable state" and subsequent query failed with the following error,
`DatabaseError(UnableToSendCommand, "another command is already in progress\n")`

> A null pointer is returned when the command is complete and there will be no more results.

From [libpq docs](https://www.postgresql.org/docs/current/libpq-async.html#LIBPQ-PQGETRESULT). As per my understanding, before using the same connection for another query, we need to "clean" connection/channel by fetching result repeatedly until we get NULL. I managed to test this hypothesis and it worked, but I'm not sure whether it's the right solution. For more information about my changes, see "Steps to reproduce".

The use-case is, 
1. create a connection 
2. try to insert a record 
3. upon failing on 'duplicate violation' error, fetch record.

The steps 2 and 3 can be done multiple times using the same connection. [think pool manger provides an existing connection] 
FYI: the problem is reproducible without any pool manager.

The same use-case flow works first time using newly created connection. Upon using the same connection to do the same flow next time, database error is being thrown on 3rd step (fetch record).
The expected output is to be able to use the same connection for as many queries as possible, sequentially. In this case, fetch record without throwing error.

Throws `DatabaseError` whereas it should fetch the record.

### Steps to reproduce
Here is the public repo, https://github.com/heyrutvik/diesel-error-demo, which reproduces the problem. It uses [my branch of diesel](https://github.com/heyrutvik/diesel/commit/1a8e2d84a672d3d8f5bee4a217cbc896314950e8) that I created to debug the issue but one can change `Cargo.toml` to uncomment a line to use diesel `2.0.0` version. Please read README which has "extended" logs and explanation. I also tried to explain the scenario in its `main.rs` file. There is also a possible solution explained in the README.

## Checklist

I believe README and comments in `main.rs` are enough to understand the issue but feel free to ping me your questions.

heyrutvik:
Closes #3342


heyrutvik:
The build is failed because of [this](https://github.com/diesel-rs/diesel/actions/runs/3126043831/jobs/5071080266). I'm not seeing any clippy errors in local but will make changes as per the build log.

diesel_tests/tests/connection.rs
```diff
@@ -2,22 +2,21 @@ use crate::schema::*;
 use diesel::connection::BoxableConnection;
 use diesel::*;
 
-table! {
```
seeekr, line 5: Moving around & changing code that's not necessary as part of the new or updated test code should be avoided, i.e. usually the minimal set of changes to resolve an issue is preferable.
If this is indeed updating some test code to follow newer and better patterns that are already in use elsewhere, that's of course acceptable and probably desirable. I don't know the specifics of this case, but I wanted to leave this general note so you could act on it based on your best judgment.

heyrutvik: @seeekr the PR has been updated to keep only ticket related changes. Thanks for your feedback.

heyrutvik:
@weiznich There is a [table definition](https://github.com/diesel-rs/diesel/blob/master/diesel_tests/tests/connection.rs#L5-L11) in the connection.rs file which could be moved to the relevant test method. I don't see a reason to keep it outside. The test I added [encloses table definition within](https://github.com/diesel-rs/diesel/pull/3344/files#diff-fe8a7db272b54bc3a035d79dca6f87144fbeb215cd460a481afcabee27b98557R127-R133) it which is inspired from other test cases, [example](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/connection/mod.rs#L736).

@seeekr has made a suggestion that we can move that table definition to the relevant test method (and other similar pattern in the codebase, if any) using following changeset to keep the changes minimal,
```rust
fn managing_updated_at_for_table() {

    table! {
        #[sql_name = "auto_time"]
            auto_time_table {
            id -> Integer,
            n -> Integer,
            updated_at -> Timestamp,
        }
    }

    use auto_time_table::dsl::{auto_time_table as auto_time, n, updated_at};
    ...
```

Initially, I made changes to enclose that table definition, different from this, but it was cluttering the actual changes. I have reverted it to push only changes relevant to the ticket. Let me know your thoughts about it, whether we want to include such changes in current PR or in a separate one.

diesel/src/pg/connection/result.rs
```diff
@@ -48,6 +48,12 @@ impl PgResult {
                 ))
             }
             _ => {
+                // "clearing" the connection by polling result till we get a null.
+                // this will indicate that the previous command is complete and
+                // the same connection is ready to process next command.
+                // https://www.postgresql.org/docs/current/libpq-async.html
+                while conn.get_next_result()?.is_some() {}
```
weiznich, line 55: We want to return the orginal error here, so we want to avoid using `?` at all. Something like

```suggestion
                while conn.get_next_result().map(Option::is_some).unwrap_or(true) {}
```

(although this will recursively call the same function (`PgResult::new`) again and potentially run this loop there again. I believe that should be fine as this should stop at some layer wen `get_next_result()` returns `None`)

weiznich (commented):
Thanks for filling this PR :+1:
This looks already good, but I feel that we should act a bit more careful here in terms of error propagation.

diesel/src/pg/connection/result.rs
heyrutvik: @weiznich Is the default value `true` intended? Wouldn't it keep the loop running even after some error? Am I missing something?

weiznich: Yes `true` is the correct default value as far as I can tell. The libpq documentation states that `PQgetResult` (which is called internally by this function) must be called till it returns `null`. It does not talk about any errors there, so I assume that means we literally need to call it again till it returns null.

That by itself seems reasonable as [`get_next_result()`](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/connection/raw.rs#L119) calls `PQgetResult` (which cannot return an error) and then `PgResult::new()` (which can return an error). So the only way `get_next_result()` would ever return an error is by hitting basically the current code block. We would like to ignore that as the first error message is relevant.

heyrutvik:
Added separate commit to move table definition to its relevant test method. I haven't looked into all the tests file because of the time crunch but updated the same file where I added the test for the ticket. Also, updated the polling code as per your suggestion. Thanks, @weiznich.

weiznich (approved):

weiznich:
Thanks again for providing the example + working on the fix :+1: I will try to fix https://github.com/diesel-rs/diesel/issues/3330 as well and then put all the recent fixes into a 2.0.1 patch release. I cannot give an ETA on the exact release date.

PR #3349, opened 2022-09-29
Fix #3330

### LINKED ISSUE #3330: Upsert with where clause Postgres implementation is wrong

weiznich:
This commits fixes an issue with our dsl generating an invalid query for `UPSERT` statements with `WHERE` clauses. As described in https://www.postgresql.org/docs/current/sql-insert.html `WHERE` is only valid in combination with `DO UPDATE` here. In detail this commit fixes the following issues:

* `WHERE` appears now in the correct place in the generated query (after `ON UPDATE SET …` instead of before this clause
* We now reject queries that use `filter_target` but `DO NOTHING` as action. Those are not valid as outlined in the postgresql documentation.

In addition a test case for both issues is added.

Similarly to #3341 this contains a technically breaking change, as it turns more broken queries in compile-errors. Similar to that issue I would classify that as bug fix.


diesel/src/pg/query_builder/query_fragment_impls.rs
```diff
@@ -98,15 +100,20 @@ where
...
-    for DecoratedConflictTarget<T, U>
+impl<Values, Target, Action, Where> QueryFragment<Pg, crate::pg::backend::PgOnConflictClaues>
+    for OnConflictValues<Values, DecoratedConflictTarget<Target, Where>, DoUpdate<Action>>
 where
-    T: QueryFragment<Pg>,
-    U: QueryFragment<Pg>,
+    Values: QueryFragment<Pg>,
+    Target: QueryFragment<Pg>,
+    DoUpdate<Action>: QueryFragment<Pg>,
+    Where: QueryFragment<Pg>,
```
chusteven, line 109: Curious about the change to what look like generics... Is this to make the code more readable? I've always been confused by the plethora of U, V, T, etc. letters all over the place :)

weiznich: Some of them are generics. Generics in rust must always be defined explicitly before usage. That happens by the `impl<Define, Generics, Here>` block at the beginning of the impl. For this case that means `Values`, `Target`, `Action` and `Where` are generics, anything else is a concrete type. 

As for what this change want's to achieve: It changes what queries are accepted or rejected at compile time. Before this change any generic action was accepted (by another impl than this one) in place of the last generic argument to `OnConflictValues`. This impl does not restrict that type parameter to the specific `DoUpdate<Action>` type. It's basically that which allows the added compile test to reject the invalid query now. That did not happen before.

Ten0 (approved):

Issue #3313, opened 2022-09-08
Insertable derive macro causes struct properties to fail to compile with `error: unnecessary qualification` error when `-Dunused_qualifications` rustc flag used


mikemcdermottredjack:
## Setup
.cargo/config:
[target.'cfg(all())']
rustflags = [
    "-Dunused_qualifications"
]

lib.rs: 
use diesel::Insertable;

#[derive(Insertable)]
struct MyTable {
    my_column: i32
}

cargo.toml:
[dependencies]
diesel = { version = "~2.0", features = ["postgres", "postgres_backend", "chrono", "r2d2", "uuid"], default-features = false }

### Versions

Rust: 1.63
Diesel: 2.0.0
Database: PgSQL
OS: MacOS

## Problem Description
Building with rustc -Dunused_qualifications causes properties of structs using the Insertable derive macro to fail with an unnecessary qualification error, regardless of whether or not the property of the struct is used.

### What is the expected output?
No error

### What is the actual output?
Error

### Are you seeing any additional errors?
No

### Steps to reproduce
See attached test project for minimal reproducible example
[test_diesel_bug.tar.gz](https://github.com/diesel-rs/diesel/files/9529720/test_diesel_bug.tar.gz)

Run cargo build on the project. 

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
Yes

saethlin:
It looks like in the provided example the `#[derive(Insertable)]` expands to:
```rust
#[allow(unused_imports)]
const _: () = {
    use diesel;
    use diesel::insertable::Insertable;
    use diesel::internal::derives::insertable::UndecoratedInsertRecord;
    use diesel::prelude::*;
    impl Insertable<my_tables::table> for MyTable {
        type Values = <(
            std::option::Option<diesel::dsl::Eq<my_tables::my_column, i32>>,
        ) as Insertable<my_tables::table>>::Values;
        fn values(
            self,
        ) -> <(
            std::option::Option<diesel::dsl::Eq<my_tables::my_column, i32>>,
        ) as Insertable<my_tables::table>>::Values {
            (std::option::Option::Some(my_tables::my_column.eq(self.my_column)),)
                .values()
        }
    }
    impl<'insert> Insertable<my_tables::table> for &'insert MyTable {
        type Values = <(
[...]

PR #3353, opened 2022-09-30
Fix Unused Qualifications Issue in Insertable

mikemcdermottredjack:
Suppresses the lint so that using -Dunused_qualifications does not throw an error for structs using the Insertable macro. 

Fixes https://github.com/diesel-rs/diesel/issues/3313

### DIFF
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -122,7 +122,7 @@ pub fn derive(item: DeriveInput) -> TokenStream {
         #[allow(unused_qualifications)]
         #insert_owned
 
-    #[allow(unused_qualifications)]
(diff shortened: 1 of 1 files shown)


diesel_derives/src/insertable.rs
```diff
@@ -119,8 +119,10 @@ pub fn derive(item: DeriveInput) -> TokenStream {
         use diesel::internal::derives::insertable::UndecoratedInsertRecord;
         use diesel::prelude::*;
 
+        #[allow(unused_qualifications)]
         #insert_owned
 
+    #[allow(unused_qualifications)]
```
Ten0, line 125: Indent seems incorrect here. Otherwise lgtm

Ten0 (approved):

PR #3366, opened 2022-10-09
Replace useless call to `ok` with `unwrap`

hkBst:
Replace useless call to `ok` with `unwrap`. This is also what dotenvy recommends: https://docs.rs/dotenvy/latest/dotenvy/fn.dotenv.html.

### DIFF
--- a/examples/postgres/getting_started_step_1/src/lib.rs
+++ b/examples/postgres/getting_started_step_1/src/lib.rs
@@ -7,7 +7,7 @@ use dotenvy::dotenv;
 use std::env;
 
 pub fn establish_connection() -> PgConnection {
-    dotenv().ok();
+    dotenv().unwrap();
 
     let database_url = env::var("DATABASE_URL").expect("DATABASE_URL must be set");
     PgConnection::establish(&database_url)


Ten0:
Current workflow may be: people copy paste and don't know this fails and keep this in their final code - > dirty.
I would tend to agree this is indeed not ideal.

Now with the proposed workflow, people would get an error on copy pasting if they actually don't use dotenv files, and get to always figure out that they don't need dotenv.

I would argue that dotenv is in fact rarely the proper way to pass env variables to your program (docker, CI systems, direnv...) and that it may be reasonable to remove that opinionated code from the example (a simple mention in the guide that this is one way to do it would be fine?)

But that is just my personal opinion, it's probably better if we wait for more 🙂

NB: I think if we happen to decide keep the behavior as is, the idiomatic way to ignore the potential error is indeed not `.ok()`, but the following:
```
let _ = dotenv();
```

weiznich:
> I would argue that dotenv is in fact rarely the proper way to pass env variables to your program (docker, CI systems, direnv...) and that it may be reasonable to remove that opinionated code from the example (a simple mention in the guide that this is one way to do it would be fine?)

I would be fine with removing `dotenv` from all the examples, as long as such a PR updates all examples at once and a separate PR is filled against the web-page repository to update the code there as well.

hkBst:
This seems to have gotten stuck over the past month. The immediate question is whether to ignore missing dotenv files or to propagate such errors. For the longer term the use of dotenv was questioned, but I'll leave that for someone else.

Jaffa-Cakes (approved):

PR #3370, opened 2022-10-11
Rewrite table! as proc_macro

weiznich:
This commits rewrites `diesel::table!` as proc marco. The rewrite is done for the following reasons:

* Allow to reuse the syn parser of `table!` for other functionality as well (That's the main reason, might be really useful for some ideas regarding to for example generating migrations by diffing `schema.rs` + the actual database in `diesel-cli`)
* Simpler to read, as the old `macro_rules!` based implementation was quite hard to follow (That should make it easier to implement future changes to `table!` as well)
* Better error reporting, as we now can control the exact spans of items.


weiznich:
@diesel-rs/reviewers This is ready for a final review

weiznich:
@diesel-rs/reviewers Please leave a comment here if you plan to review this PR. Otherwise I would like to move forward and merge it next week.

PR #3373, opened 2022-10-12
First draft of the dynamic schema customization

Razican:
This is a first draft related to [this idea](https://github.com/diesel-rs/diesel/discussions/3372). It's also my first contribution to anything related to the query builder, and therefore, I expect it to have major flaws, but let's hope I can learn and also fix them :)

The idea here is to have an API similar to the update / insert statements. For example:

```rust
let columns = vec![
    ColumnDefinition::new("id", "serial"),
    ColumnDefinition::new("name", "text").not_null()
].into().primary_key("id");
create_table("users").columns(columns).run(&mut connection)?;
```

Something that was mentioned in the idea was that it should work with the strongly typed DSL as well, but I'm not sure how to achieve this. Since this changes the schema definition, it's tricky to re-use existing types. Nevertheless, I guess I could use some of it to define column types.

I'm opening this draft PR in order to get some feedback and guidance, to make sure I'm not spending time on something that doesn't make much sense.


weiznich:
Thanks for opening this PR.

I think `create_database`/`drop_database` looks fine, although they can be moved to `diesel` itself.

For `CREATE TABLE` I imagine the following interface: `diesel::create_table(some_table_type).execute(…)` where `some_table_type` is either a type from the generated `schema.rs` file or a `diesel_dynamic_schema::Table` instance. In both cases all the necessary information are accessible via the [`diesel::Table` trait](https://docs.diesel.rs/master/diesel/prelude/trait.Table.html), which lists all columns and the primary key. Each column also contains information about the corresponding field type via [`diesel::expression::Expression::SqlType`](https://docs.diesel.rs/master/diesel/expression/trait.Expression.html) although that likely needs some helper code to translate the corresponding compile time types in something that could be used for create table. We might need a new trait for that.

PR #3374, opened 2022-10-12
Add "delete" ability to `diesel_dynamic_schema::Table`

Razican:
This draft PR is meant to add "delete" capabilities to `diesel_dynamic_schema::Table`. Once it's working fine, this code should work:

```rust
    let users = diesel_dynamic_schema::table("users");
    let id = users.column::<Integer, _>("id");

    diesel::delete(users.filter(id.eq(42)))
        .execute(connection)
        .unwrap();
```

Also, I would like to implement the truncate table statements, but it's currently not working (I need to implement `diesel::query_builder::IntoUpdateTarget` for `diesel_dynamic_schema::Table`).

I would like to see if the CI complains with the new changes.

### DIFF
--- a/diesel_dynamic_schema/src/table.rs
+++ b/diesel_dynamic_schema/src/table.rs
@@ -1,3 +1,4 @@
+use diesel::associations::HasTable;
 use diesel::backend::Backend;
 use diesel::expression::expression_types;
 use diesel::internal::table_macro::{FromClause, SelectStatement};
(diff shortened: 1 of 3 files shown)


diesel_dynamic_schema/src/table.rs
```diff
@@ -112,3 +113,18 @@ impl<T, U> QueryId for Table<T, U> {
...
+impl<T, U> HasTable for Table<T, U>
+where
+    T: Default,
+    Self: QuerySource + AsQuery,
+{
+    type Table = Self;
+
+    fn table() -> Self {
+        Self {
+            name: T::default(),
```
weiznich, line 126: There is no way that this will work, as the delete statement needs to have access to the underlying table name.

It be possible to fix this by either looking at the existing `HasTable` impl for `SelectStatement` or at the `IntoUpdateTarget` impls for `SelectStatement` or `Table`

weiznich (changes requested):
This approach will unfortunately not work.

PR #3376, opened 2022-10-19
Add comment on relevance of order for Queryable

pinpox:
As suggested by @weiznich over on Matrix

### DIFF
--- a/examples/sqlite/getting_started_step_1/src/models.rs
+++ b/examples/sqlite/getting_started_step_1/src/models.rs
@@ -1,5 +1,8 @@
 use diesel::prelude::*;
 
+// Order matters here!
+// Queryable assumes that the order of field on the struct matches the columns in the corresponding table.
+// See https://docs.diesel.rs/diesel/deserialize/trait.Queryable.html#derivingfor details
 #[derive(Queryable)]
 pub struct Post {
     pub id: i32,


Ten0 (approved):

Ten0:
This looks like an sqlite-specialized example. Do we have equivalents for the other backends? If that's the case we probably want to add this comment here as well no?

weiznich:

The relevant examples for are [here for mysql](https://github.com/diesel-rs/diesel/blob/5dc6b2a9e681c2bf0880b9b75c84edd784e8f221/examples/mysql/getting_started_step_1/src/models.rs#L3) and [here for postgresql](https://github.com/diesel-rs/diesel/blob/5dc6b2a9e681c2bf0880b9b75c84edd784e8f221/examples/postgres/getting_started_step_1/src/models.rs#L3). I agree that we should add the comment there as well.

The CI currently fails due to a change in clap. That's unrelated to this PR. I pushed a fix for that [here](https://github.com/diesel-rs/diesel/pull/3370/commits/c3daebe0932197955e019a7fb21bbb0474aca070), but that's not merged to master yet.

examples/sqlite/getting_started_step_1/src/models.rs
```diff
@@ -1,5 +1,8 @@
 use diesel::prelude::*;
 
+// Order matters here!
+// Queryable assumes that the order of field on the struct matches the columns in the corresponding table.
+// See https://docs.diesel.rs/diesel/deserialize/trait.Queryable.html#derivingfor details
```
weiznich, line 5: We should point to the current documentation, instead to the old diesel 1.4 documentation here.
```suggestion
// See https://docs.diesel.rs/master/diesel/deserialize/derive.Queryable.html for details
```

Jaffa-Cakes (approved):
Looks good to me so long as [this](https://github.com/diesel-rs/diesel/pull/3376#discussion_r999822910) is changed!

PR #3391, opened 2022-11-04
Allow the `pagination.rs` to accept the optional page parameters.

Moicen:
In the `advanced-blog-cli` example, it would be more practical for the `pagination.rs` has the ability to accept optional page parameters. When the page parameters are none, then it will return all the entities from the database. In addition, returning the total number of entities regardless of the page parameters are often used in real projects.

### DIFF
--- a/examples/postgres/advanced-blog-cli/src/pagination.rs
+++ b/examples/postgres/advanced-blog-cli/src/pagination.rs
@@ -3,18 +3,21 @@ use diesel::prelude::*;
 use diesel::query_builder::*;
 use diesel::query_dsl::methods::LoadQuery;
 use diesel::sql_types::BigInt;
+use std::fmt::Debug;
 
 pub trait Paginate: Sized {
-    fn paginate(self, page: i64) -> Paginated<Self>;
+    fn paginate(self, page: Option<i64>) -> Paginated<Self>;
 }
 
 impl<T> Paginate for T {
-    fn paginate(self, page: i64) -> Paginated<Self> {
+    fn paginate(self, page: Option<i64>) -> Paginated<Self> {
+        let offset = if page.is_some() { (page.clone().unwrap() - 1) * DEFAULT_PER_PAGE } else { -1 };
         Paginated {
             query: self,
+            some_per_page: Some(DEFAULT_PER_PAGE),
             per_page: DEFAULT_PER_PAGE,
             page,
-            offset: (page - 1) * DEFAULT_PER_PAGE,
+            offset,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


examples/postgres/advanced-blog-cli/src/pagination.rs
```diff
@@ -58,16 +81,22 @@ impl<T: Query> Query for Paginated<T> {
...
+        T: QueryFragment<Pg>,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
-        out.push_sql("SELECT *, COUNT(*) OVER () FROM (");
-        self.query.walk_ast(out.reborrow())?;
-        out.push_sql(") t LIMIT ");
-        out.push_bind_param::<BigInt, _>(&self.per_page)?;
-        out.push_sql(" OFFSET ");
-        out.push_bind_param::<BigInt, _>(&self.offset)?;
+        if self.page.is_some() && self.some_per_page.is_some() {
```
weiznich, line 88: This `QueryFragment` implementation is incompatible with the declared prepared statement caching. We need to call [`AstPass::unsafe_to_cache_prepared`](https://docs.diesel.rs/master/diesel/query_builder/struct.AstPass.html#method.unsafe_to_cache_prepared) here as otherwise diesel might reuse an existing prepared statement for the "other" variant.

weiznich (changes requested):
Thanks for opening this PR. Beside the mentioned issue with the implementation I'm not really in favor for merging this change. The `pagination` example is explicitly not a full fleshed pagination implementation that needs to satisfy all use cases. It is designed to demonstrate how diesel can be extend by third party crates. For this use case the code is more than sufficient, especially as it is coupled to the corresponding guide.

Moicen:
> Thanks for opening this PR. Beside the mentioned issue with the implementation I'm not really in favor for merging this change. The `pagination` example is explicitly not a full fleshed pagination implementation that needs to satisfy all use cases. It is designed to demonstrate how diesel can be extend by third party crates. For this use case the code is more than sufficient, especially as it is coupled to the corresponding guide.

Okay! I guess this PR itself can be a reference to the people who need this optional parameters.

examples/postgres/advanced-blog-cli/src/pagination.rs
Moicen: @weiznich  Could you please provide some examples on how to use this [AstPass::unsafe_to_cache_prepared](https://docs.diesel.rs/master/diesel/query_builder/struct.AstPass.html#method.unsafe_to_cache_prepared)?

Moicen: Do you mean the additional query is not correct to write like this?

image

weiznich: > Could you please provide some examples on how to use this AstPass::unsafe_to_cache_prepared?

You literally only need to call this method inside your `QueryFragment::walk_ast` method. So an example on how to use this would be literally just `out.unsafe_to_cache_prepared()`.


> Do you mean the additional query is not correct to write like this?

The problem here is that the following problem can happen:

1. You call your paginate setup with `Some(_)` as `page`. 
2. Diesel creates a prepared statement for the query generated by the `if` branch
3. You call your paginate setup with `None` as `page`
4. Diesel just sees that this is the same type based on the `QueryId` impl and skips building the query string at all as it just uses the type id to identify the already created prepared statement.

To prevent that from happening you can use `AstPass::unsafe_to_cache_prepared` to completely opt out the caching of prepared statements for this query, or you could provide a custom `QueryId` impl that sets [`QueryId::HAS_STATIC_QUERY_ID`](https://docs.diesel.rs/master/diesel/query_builder/trait.QueryId.html#associatedconstant.HAS_STATIC_QUERY_ID) to `false`, which forces the query builder to run every time the query is executed and which uses the query string as cache key.

Moicen: @weiznich Thanks for teaching!

PR #3395, opened 2022-11-06
Drop GAT workarounds

Elrendio:
**What it does:**
- Migrates GatWorkarounds to actual GATs

**Good to know:**
- Trait `ConnectionSealed` is kept to keep the feature gate around the Connection implementation.
- I had to keep the associated type on `RowFieldLifetimeHelper` in order to express `FieldRet` without the lifetime of `Row`
- This increases the minimal rust version from `1.60.0` to `1.65.0`


Elrendio:
Tests seems to fail because of GAT not stable :thinking:

Ten0:
> Tests seems to fail because of GAT not stable thinking

https://github.com/diesel-rs/diesel/blob/f560c29b4226d8651b6946de19e1cfcaea479309/.github/workflows/ci.yml#L298

diesel/src/row.rs
```diff
@@ -154,15 +154,15 @@ mod private {
...
+    pub trait RowSealed<DB: Backend> {
         /// Field type returned by a `Row` implementation
         ///
         /// * Crates using existing backend should not concern themself with the
         ///   concrete type of this associated type.
         ///
         /// * Crates implementing custom backends should provide their own type
         ///   meeting the required trait bounds
-        type Field: Field<'a, DB>;
+        type Field<'a>: Field<'a, DB>;
```
Ten0, line 165: why doesn't this move on the `Row` trait?

diesel/src/row.rs
```diff
@@ -29,15 +29,15 @@ pub trait RowIndex<I> {
...
+pub type FieldRet<'a, R, DB> = <R as RowSealed<DB>>::Field<'a>;
 
 /// Represents a single database row.
 ///
 /// This trait is used as an argument to [`FromSqlRow`].
 ///
 /// [`FromSqlRow`]: crate::deserialize::FromSqlRow
 pub trait Row<'a, DB: Backend>:
-    RowIndex<usize> + for<'b> RowIndex<&'b str> + for<'b> RowGatWorkaround<'b, DB> + Sized
+    RowIndex<usize> + for<'b> RowIndex<&'b str> + RowSealed<DB> + Sized
```
Ten0, line 40: Looks like RowIndex here might also be a Gat workardound

diesel/src/row.rs
Elrendio: I had to keep the associated type on RowSealed in order to express FieldRet without the lifetime of Row

diesel/src/row.rs
Ten0: it isn't

diesel/src/row.rs
Ten0: explained in the commit message and additional doc

diesel/src/pg/connection/mod.rs
```diff
@@ -364,14 +360,14 @@ mod private {
 
     pub trait PgLoadingMode<B>
     where
-        for<'conn, 'query> Self: ConnectionGatWorkaround<'conn, 'query, Pg, B>,
+        for<'conn, 'query> Self: ConnectionSealed<Pg, B>,
```
Ten0, line 363:
```suggestion
        Self: ConnectionSealed<Pg, B>,
```

Ten0:
Likely needs some msrv upgrade, compile tests output regeneration and some clippy stuff

Elrendio:
Clippy warnings seems to be mostly unrelevant. For the elided lifetime, it is required and  I really don't think `'_` is  clearer than `'conn`

Elrendio:
The compile error comes from a warning that's also generated and seems to be a bug from Rust

.github/workflows/ci.yml
```diff
@@ -295,7 +295,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2022-08-12
+          toolchain: nightly-2022-11-05
```
weiznich, line 298: This likely requires rerunning the compile tests with `TRYBUILD=overwrite` + committing all the changes.

.github/workflows/ci.yml
```diff
@@ -423,11 +423,11 @@ jobs:
         run: cargo +nightly -Z build-std test --manifest-path diesel/Cargo.toml --no-default-features --features "sqlite extras libsqlite3-sys libsqlite3-sys/bundled libsqlite3-sys/with-asan" --target x86_64-unknown-linux-gnu
 
   minimal_rust_version:
-    name: Check Minimal supported rust version (1.60.0)
+    name: Check Minimal supported rust version (1.65.0)
```
weiznich, line 426: This requires an change log entry.

diesel/src/connection/mod.rs
```diff
@@ -40,7 +40,17 @@ pub trait SimpleConnection {
 ///
 /// Users should threat this type as `impl Iterator<Item = QueryResult<impl Row<DB>>>`
 pub type LoadRowIter<'conn, 'query, C, DB, B = DefaultLoadingMode> =
-    <C as ConnectionGatWorkaround<'conn, 'query, DB, B>>::Cursor;
+    <C as LoadRowIterDiesel2Compat<DB, B>>::Cursor<'conn, 'query>;
+
+#[doc(hidden)]
```
weiznich, line 45: This should be moved to the `private` module.

diesel/src/connection/mod.rs
```diff
@@ -379,10 +388,16 @@ where
...
 /// different loading modes via the generic paramater.
-pub trait LoadConnection<B = DefaultLoadingMode>: Connection
-where
-    for<'conn, 'query> Self: ConnectionGatWorkaround<'conn, 'query, Self::Backend, B>,
-{
+pub trait LoadConnection<B = DefaultLoadingMode>: Connection {
+    /// The cursor type returned by [`LoadConnection::load`]
+    ///
+    /// Users should handle this as opaque type that implements [`Iterator`]
+    type Cursor<'conn, 'query>: Iterator<Item = QueryResult<Self::Row<'conn, 'query>>>;
```
weiznich, line 395: The `LoadConnection::load` method should be changed so that it returns `QueryResult<Self::Cursor<'conn, 'query>` now. I think that's much easier to understand for potential users + it's the same type as before.

diesel/src/connection/mod.rs
```diff
@@ -40,7 +40,17 @@ pub trait SimpleConnection {
 ///
 /// Users should threat this type as `impl Iterator<Item = QueryResult<impl Row<DB>>>`
 pub type LoadRowIter<'conn, 'query, C, DB, B = DefaultLoadingMode> =
```
weiznich, line 42: This type can be deprecated. (It should be behind the `with-deprecated` feature flag + removed for the `without-deprecated` flag)

diesel/src/row.rs
```diff
@@ -29,15 +29,15 @@ pub trait RowIndex<I> {
 /// Return type of [`Row::get`]
 ///
 /// Users should threat this as opaque [`impl Field<DB>`](Field) type.
-pub type FieldRet<'a, R, DB> = <R as RowGatWorkaround<'a, DB>>::Field;
+pub type FieldRet<'a, R, DB> = <R as RowFieldLifetimeHelper<DB>>::Field<'a>;
```
weiznich, line 32: This type should be deprecated as well

diesel/src/row.rs
weiznich: Can you check if `pub type FieldRet<'a, R, DB> = <R as Row<'a, DB>>::Field<'a>;` works for all provided cases? If that's the case I would rather move the `Field` GAT to `Row` instead of having another trait for this.

weiznich (changes requested):
I've left a number of comments about which parts should be changed so that it is clearer for potential users what the the corresponding types look like.

For the clippy/ci stuff: I've opened https://github.com/diesel-rs/diesel/pull/3396 to address that.

diesel/src/connection/mod.rs
```diff
@@ -490,30 +505,11 @@ impl<DB: Backend + 'static> dyn BoxableConnection<DB> {
 // because we want to replace them by with an associated type
```
Ten0, line 505: this comment block should be updated

weiznich:
Yes `HasBindCollector` and also [`HasRawValue`](https://docs.diesel.rs/master/diesel/backend/trait.HasRawValue.html) are  GAT workarounds. Both can be added directly to `Backend`.

weiznich:
@Elrendio Do you plan to follow up onto the comments above?

weiznich:
This is currently blocked on https://github.com/rust-lang/rust/issues/106417 and on a trybuild release containing https://github.com/dtolnay/trybuild/pull/213

PR #3396, opened 2022-11-07
Rust toolchain update

weiznich:


Ten0 (approved):
Simple curiosity, why does this not go with updating `.github/workflows/ci.yml` ?

PR #3400, opened 2022-11-08
Implement Default on tables

Elrendio:
**Good to know:**
- Useful when building generic code around tables (same as columns)

### DIFF
--- a/diesel_derives/src/table.rs
+++ b/diesel_derives/src/table.rs
@@ -201,7 +201,7 @@ pub(crate) fn expand(input: TableDecl) -> TokenStream {
             pub const all_columns: (#(#column_names,)*) = (#(#column_names,)*);
 
             #[allow(non_camel_case_types)]
-            #[derive(Debug, Clone, Copy, self::diesel::query_builder::QueryId)]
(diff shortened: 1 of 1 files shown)


weiznich (approved):

diesel_derives/src/table.rs
```diff
@@ -201,7 +201,7 @@ pub(crate) fn expand(input: TableDecl) -> TokenStream {
             pub const all_columns: (#(#column_names,)*) = (#(#column_names,)*);
 
             #[allow(non_camel_case_types)]
-            #[derive(Debug, Clone, Copy, self::diesel::query_builder::QueryId)]
+            #[derive(Debug, Clone, Copy, self::diesel::query_builder::QueryId, Default)]
```
Ten0, line 204: we should do the same trick as was done for QueryId

Ten0:
We should have in mind that this is *technically* breaking, although it's unlikely to significantly break stuff.

Ten0 (approved):

weiznich:
For reference: https://github.com/diesel-rs/diesel/pull/1559 introduced the same derive for column structs. We didn't tread that as breaking back then.

Ten0:
If nobody came back for this I'd be in favor of just merging it as-is then. 🙂

PR #3402, opened 2022-11-09
Add a `#[derive(MultiConnection)]` proc-macro

weiznich:
This proc macro implements `Connection` and related traits for an enum of connections for different database backends. This allows to easily support multiple database systems at the same time. I've chosen to implement this feature as a derive, instead of providing just one `AnyConnection` enum with all diesel connection types, to allow third party connection implementations to be used there as well.

The following use:
```rust
#[derive(diesel::MultiConnection)]
pub enum MultiConnection {
    Postgres(PgConnection),
    Mysql(diesel::MysqlConnection),
    Sqlite(diesel::SqliteConnection),
}
```

generates [this implementation](https://gist.github.com/weiznich/19406dde61f5225cd2dc2927a6d33338). It then can be used like this:

```rust
fn main() {
    use diesel::prelude::*;

    diesel::table! {
        users(id) {
            id -> Integer,
            name -> Text,
        }
    }

    // all connection types share the same type
    let mut pg_connection: MultiConnection =
        MultiConnection::establish("postgres://localhost/diesel_test").unwrap();
    let mut sqlite_connection: MultiConnection = MultiConnection::establish(":memory:").unwrap();
    let mut mysql_connection: MultiConnection =
        MultiConnection::establish("mysql://weiznich:diesel@localhost/diesel_tests").unwrap();

    assert!(matches!(&pg_connection, MultiConnection::Postgres(_)));
    assert!(matches!(&sqlite_connection, MultiConnection::Sqlite(_)));
    assert!(matches!(&mysql_connection, MultiConnection::Mysql(_)));

    pg_connection.begin_test_transaction().unwrap();
    sqlite_connection.begin_test_transaction().unwrap();
    mysql_connection.begin_test_transaction().unwrap();

    diesel::sql_query(
        "CREATE TABLE IF NOT EXISTS users (id SERIAL PRIMARY KEY, name TEXT NOT NULL)",
    )
    .execute(&mut pg_connection)
    .unwrap();

    diesel::sql_query("CREATE TABLE IF NOT EXISTS users (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT NOT NULL)")
        .execute(&mut sqlite_connection)
        .unwrap();

    diesel::sql_query(
        "CREATE TEMPORARY TABLE users (id INTEGER PRIMARY KEY AUTO_INCREMENT, name TEXT NOT NULL)",
    )
    .execute(&mut mysql_connection)
    .unwrap();

    let insert_statement = diesel::insert_into(users::table).values(users::name.eq("John"));

    insert_statement.execute(&mut pg_connection).unwrap();
    insert_statement.execute(&mut mysql_connection).unwrap();
    insert_statement.execute(&mut sqlite_connection).unwrap();

    // we can execute queries again all of them
    let select_statement = users::table.order_by(users::id).limit(20);
    let res = select_statement
        .load::<(i32, String)>(&mut pg_connection)
        .unwrap();
    dbg!(res);

    let res = select_statement
        .load::<(i32, String)>(&mut mysql_connection)
        .unwrap();
    dbg!(res);

    let res = select_statement
        .load::<(i32, String)>(&mut sqlite_connection)
        .unwrap();
    dbg!(res);

    // execute backend specific queries on the underlying connection
    if let MultiConnection::Postgres(ref mut pg_conn) = pg_connection {
        let res = users::table.load::<(i32, String)>(pg_conn).unwrap();
        dbg!(res);
    }
}
```

After landing this I suggest we use this to reimplement `diesel_cli`, which should remove quite a bit of generic code there. In addition that would allow us to test this feature internally.


diesel/src/query_builder/bind_collector.rs
```diff
@@ -31,7 +31,7 @@ pub trait BindCollector<'a, DB: TypeMetadata>: Sized {
     ) -> QueryResult<()>
     where
         DB: Backend + HasSqlType<T>,
-        U: ToSql<T, DB> + 'a;
```
weiznich, line 34: We need to decide if we consider that a breaking change or not. In the former case we need to find another solution for how to handle unsized bind paramaters like `str` and `[u8]`. This is currently required for [handling `str` and `[u8]` in the custom bind collector implementation](https://gist.github.com/weiznich/19406dde61f5225cd2dc2927a6d33338#file-impl-rs-L613-L722)

Ten0: I think this is only a breaking change for people implementing `BindCollector`, which is:
- People who are implementing a third-party connection, so they are likely relying on `i-opt-into-breaking-changes` anyway
- Very very few people.

-> The cost/ratio seems to be in favor of not considering this a breaking change - unless I'm mistaken on the number of third-party connection implementations out there.

diesel_derives/src/multiconnection.rs
```diff
@@ -0,0 +1,1365 @@
...
+                );
+        }
+    });
+
+    let populate_into_bind_value_fields = connection_types
+        .iter()
+        .map(|c| {
+            let ident = syn::Ident::new(&c.name.to_string().to_lowercase(), c.name.span());
+            quote::quote! {
+                #ident: Some(#ident)
```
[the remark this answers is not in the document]
Ten0: Actually we could even write the following impls in static code:
```rust
trait PushBoundValueToCollectorDB<DB: Backend> {
    fn push_bound_value(&self, v: (), collector: PgCollector, lookup: ()) {
        use diesel::query_builder::BindCollector;
        if let InnerBindValue::Sized(v) = v {
            let v = v
                .downcast_ref::<T>()
                .expect("We know the type statically here");
            collector.push_bound_value::<ST, _>(v, lookup)
        } else {
            unreachable!("We set the value to `InnerBindValue::Sized`")
        }
    }
}

struct PushBoundValueToCollectorImpl<ST, T> {
    _spooky: std::marker::PhantomData<(ST, T)>,
}

impl<DB: Backend, ST, T> PushBoundValueToCollectorDB<DB> for PushBoundValueToCollectorImpl<ST, T> {}
```

and then the macro would only have to generate:
```rust
trait PushBoundValueToCollector:
    PushBoundValueToCollectorDB<Pg> + PushBoundValueToCollectorDB<Mysql>
{
}

struct InnerBindValue {
    value: BindValueKind,
    push_bound_value_to_collector: &'static dyn PushBoundValueToCollector,
}
```

(That being said even if we put the code that can be static in the macro that is probably a more elegant writing than what I suggested before)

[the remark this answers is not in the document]
weiznich: That's a great idea :+1:. I've implemented that in https://github.com/diesel-rs/diesel/pull/3402/commits/a3624cc2badd8e458d0dfb2100d1d35871917aed.

diesel/src/query_builder/bind_collector.rs
weiznich: I'm aware of 2 third party connection implementations:
* [diesel-oci](https://github.com/GiGainfosystems/diesel-oci), which is essentially maintained by myself so that will not be an issue
* [rsbfclient](https://github.com/fernandobatels/rsfbclient), which uses the ` i-opt-into-breaking-changes` flag anyways.

PR #3419, opened 2022-11-21
register is_nullable as a reserved name due to the clash with a modul…

### LINKED ISSUE #3404: Name colission in `table!` macro with `is_nullable` column
### Discussed in https://github.com/diesel-rs/diesel/discussions/3403

<div type='discussions-op-text'>

<sup>Originally posted by **forest1102** November 11, 2022</sup>
Hello, I am trying to update diesel 2.0, and I have a problem with a table that has an `is_nullable` column.  
Even though by adding `sql_name` attribute helps, I have to modify the schema file every time I changed the schema file. 
Suppose I made a migration file like below. 

```sql
create table db_definition_items
(
    id bigserial primary key,
    is_nullable boolean not null
);
```
and after `diesel database reset`, it will produce table schema below. 

```rs
diesel::table! {
    db_definition_items (id) {
        id -> Int8,
        is_nullable -> Bool,
    }
}
```

But this code has an error saying, 
```
`is_nullable` is ambiguous
ambiguous because of multiple glob imports of a name in the same module
consider adding an explicit import of `is_nullable` to disambiguate
consider adding an explicit import of `is_nullable` to disambiguate
```

How can I suppress this error?

## versions

```bash
$ rustc -V         
rustc 1.65.0 (897e37553 2022-11-02)

$ diesel --version
diesel 2.0.1
```

### Cargo.toml

```Cargo.toml
[package]
name = "serve"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
diesel = "2.0.2"
```</div>

czotomo:
…e with the same name

Fixes #3404

### DIFF
--- a/diesel_cli/src/infer_schema_internals/inference.rs
+++ b/diesel_cli/src/infer_schema_internals/inference.rs
@@ -13,7 +13,7 @@ static RESERVED_NAMES: &[&str] = &[
     "macro", "match", "mod", "move", "mut", "offsetof", "override", "priv", "proc", "pub", "pure",
     "ref", "return", "Self", "self", "sizeof", "static", "struct", "super", "trait", "true",
     "type", "typeof", "unsafe", "unsized", "use", "virtual", "where", "while", "yield", "bool",
-    "columns",
+    "columns", "is_nullable",
 ];
 
 fn is_reserved_name(name: &str) -> bool {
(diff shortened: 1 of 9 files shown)


czotomo:
Ugh, 
https://github.com/bitvecto-rs/funty/issues/7

weiznich (approved):

The change looks good, but requires a rustfmt run. I think it should be fine to bump the mysql dependency of `diesel_bench` in this PR, as that's  requires only changing the version in the corresponding `Cargo.toml`. I would prefer having a separate commit for that.

weiznich:
:+1:

Issue #3368, opened 2022-10-10
Postgres upsert regression in diesel 2.0.1


jgallagher:
I believe there's a regression in diesel 2.0.1 coming from 2.0.0 related to https://github.com/diesel-rs/diesel/issues/3330. That issue has a couple notes about the placement of WHERE clauses in relation to DO UPDATE upserts, but the linked postgres docs (https://www.postgresql.org/docs/current/sql-insert.html) show the conflict target can _also_ include a WHERE clause; that previously worked with diesel 2.0.0 but no longer compiles in 2.0.1.

I put an attempt at a minimal reproducer here: https://github.com/jgallagher/diesel-2-0-1-regression. The salient details are that we have a table with a unique index that itself involves another column:

```sql
CREATE TABLE items (
    id SERIAL PRIMARY KEY,
    name VARCHAR NOT NULL,
    time_deleted TIMESTAMPTZ
);

CREATE UNIQUE INDEX ON items (
    name
) WHERE
    time_deleted IS NULL;
```

that restrict item names to unique _if_ their `time_deleted` column IS NULL, but names with a non-NULL `time_deleted` are not required to be unique. We can then upsert with this query in diesel 2.0.0:

```rust
    let query = diesel::insert_into(dsl::items)
        .values(NewItem { name, time_deleted: None })
        .on_conflict(dsl::name)
        .filter_target(dsl::time_deleted.is_null())
        .do_nothing()
        .returning(Item::as_returning());
```

which produces a query with a correct WHERE clause as part of the conflict target:

```sql
INSERT INTO "items" ("name", "time_deleted") VALUES ($1, DEFAULT)
    ON CONFLICT ("name") WHERE ("items"."time_deleted" IS NULL)
    DO NOTHING RETURNING "items"."id", "items"."name", "items"."time_deleted"
```

However, this same query now fails to compile in diesel 2.0.1:

```
error[E0277]: the trait bound `query_builder::upsert::on_conflict_clause::OnConflictValues<query_builder::insert_statement::ValuesClause<(DefaultableColumnInsertValue<ColumnInsertValue<columns::name, diesel::expression::bound::Bound<Text, &str>>>, Defaultab
[...]

Issue #2430, opened 2020-06-10
Feature request: support where clause for upsert


s97712:
Postgresql upsert support where clause, such as below query: 
```sql
INSERT INTO table VALUES(...)  ON CONFLICT (id) DO UPDATE SET ... WHERE ...
```
Consider adding this feature for diesel?

weiznich:
We would be happy to accept a PR for this, but at least I do not have any plan to invest much time in this feature anytime soon.

ropottnik:
I'd like to give it a try. Can you assign me?

ropottnik:
thanks @JohnTitor

PR #3440, opened 2022-12-04
Add Conditionally InsertStatement

neogenie:
Allow to apply filters on `ON CONFLICT ... DO UPDATE ... WHERE ...`


weiznich:

Could you please clarify/address the following points:

* What exactly is supposed to work after this PR, what did not work before. Can you reference an existing issue/discussion?
* Failing CI tests (They obviously are expected to pass)
* Does this contain a breaking change? It looks like, as the return type of the `set` method is changed

neogenie:
Hi @weiznich!

Thank for fast feedback!

At least Postgres has the ability to specify a WHERE condition for a construct:

```sql
-- Don't update existing distributors based in a certain ZIP code
INSERT INTO distributors AS d (did, dname) VALUES (8, 'Anvil Distribution')
    ON CONFLICT (did) DO UPDATE
    SET dname = EXCLUDED.dname || ' (formerly ' || d.dname || ')'
    WHERE d.zipcode <> '21201';
```
There are several discussions of this topic in: #3330 and one failed Implementation attempt: #2430.

I've already fixed the tests, also added a test for the new functionality.

There are no breaking changes, a new return type `ConditionallyInsertStatement` has been added:

```rust
#[derive(Debug, Clone)]
pub struct ConditionallyInsertStatement<Stmt, WhereClause = NoWhereClause> {
    insert: Stmt,
    where_clause: WhereClause,
}
```
as a return type of `set()` method for `do_update().set()` instead of `InsertStatement`:

```rust
pub fn set<Changes>(
        self,
        changes: Changes,
    ) -> ConditionallyInsertStatement<
        InsertStatement<T, OnConflictValues<U, Target, DoUpdate<Changes::Changeset>>, Op, Ret>,
    >
    where
        T: QuerySource,
        Changes: AsChangeset<Target = T>,
    {
        let target = self.target;
        let insert = self.stmt.replace_values(|values| {
            OnConflictValues::new(values, target, DoUpdate::new(changes.as_changeset()))
        });
        ConditionallyInsertStatement::new(insert)
    }
```

Its behavior is identical to the `InsertStatement` but adds a filterable behavior: 

```rust
impl<Stmt, WhereClause, Predicate> FilterDsl<Predicate>
    for ConditionallyInsertStatement<Stmt, WhereClause>
where
    Stmt: AsQuery,
    WhereClause: WhereAnd<Predicate>,
{
    type Output = ConditionallyInsertStatement<Stmt, WhereClause::Output>;

    fn filter(self, predicate: Predicate) -> Self::Output {
        ConditionallyInsertStatement {
            insert: self.insert,
            where_clause: self.where_clause.and(predicate),
        }
    }
}
```

This allows the construction to be used:

```rust
insert_into(distributors)
        .values(&new_distributors)
        .on_conflict(did)
        .do_update()
        .set(dname.eq(excluded(dname)))
        .filter(zipcode.ne("21201".to_string()))
        .execute(connection)
        .unwrap();
```

weiznich:
Thanks for answering my questions

> There are several discussions of this topic in: https://github.com/diesel-rs/diesel/issues/3330 and one failed Implementation attempt: https://github.com/diesel-rs/diesel/issues/2430.

To be clear here: #2430 implements a **different** feature. It only looks related but adds the where clause in a different position with a different semantics. See https://github.com/diesel-rs/diesel/issues/3368 for details.

> as a return type of set() method for do_update().set() instead of InsertStatement:

I'm sorry, but that's a breaking change. We do not want to issue another major diesel version anytime soon. Cases like this one are the reason why I prefer to talk about potential implementations first, before people open PR's. This prevents frustration on both sides.

neogenie:
@weiznich

neogenie:
@weiznich 

Okay, I agree that this is may be a breaking change but it doesn't seem like it should affect most of the users, because the interface of `ConditionallyInsertStatement` is backward compatible with `InsertStatement`.

Whatever it was, I do not mind discussing alternative ways of implementation. The current solution seems the least invasive to me, but I'm willing to change the code if there are alternative implementations.

weiznich:
> Okay, I agree that this is may be a breaking change but it doesn't seem like it should affect most of the users, because the interface of ConditionallyInsertStatement is backward compatible with InsertStatement.

It does not matter if that would affect users or not. It's a breaking change and therefore not allowed.

As for the "backward compatibility": The following code would break:

```rust
fn foo<T, U>(stmt: InsertStatement<T, U>) {}

let stmt = insert_into(distributors)
        .values(&new_distributors)
        .on_conflict(did)
        .do_update()
        .set(dname.eq(excluded(dname)));

foo(stmt);
```

I'm sorry about that as you already invested some time to implement this solution, but we cannot consider this implementation due to the reasons outlined before.

As for alternative backward compatible ways to implement this: I currently cannot say how that should be done, and I do currently not have the capacity to do the required design work for such a feature.

neogenie:
@weiznich 
I replaced the implementation, now it does not contain breaking changes.

Please take a look.

neogenie:
@weiznich up

weiznich:
@neogenie I do currently not have the capacity to review that in a timely manner. Pinging me again and again does reduce my capacity to do diesel stuff even more and likely moves this specific PR down on my priority list. Please consider what you expect from an open source project run by volunteers.

diesel/src/query_builder/upsert/conditionally_insert_statement.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::{Expression, QueryResult, QuerySource, RunQueryDsl};
+
+#[derive(Debug, Clone)]
+pub struct ConditionallyInsertStatement<Stmt, WhereClause = NoWhereClause> {
+    insert: Stmt,
+    where_clause: WhereClause,
+}
+
+impl<Stmt, WhereClause> ConditionallyInsertStatement<Stmt, WhereClause> {
+    pub fn new(
```
weiznich, line 20: I would prefer keeping that as `pub(crate)` (or even more restricted) function

diesel/src/query_builder/upsert/conditionally_insert_statement.rs
```diff
@@ -0,0 +1,108 @@
...
+        where_clause: WhereClause,
+    ) -> ConditionallyInsertStatement<Stmt, WhereClause> {
+        ConditionallyInsertStatement {
+            insert,
+            where_clause,
+        }
+    }
+}
+
+impl<Stmt, WhereClause> QueryId for ConditionallyInsertStatement<Stmt, WhereClause>
```
weiznich, line 31: We should likely just add a `#[derive(QueryId)]` to the struct itself, as the `QueryId` implementation should depend on both struct fields.

diesel/src/query_builder/upsert/conditionally_insert_statement.rs
```diff
@@ -0,0 +1,108 @@
...
+where
+    Stmt: QueryId,
+{
+    type QueryId = Stmt::QueryId;
+    const HAS_STATIC_QUERY_ID: bool = Stmt::HAS_STATIC_QUERY_ID;
+}
+
+impl<Stmt, WhereClause, DB> QueryFragment<DB> for ConditionallyInsertStatement<Stmt, WhereClause>
```
weiznich, line 39: As far as I understand that's a postgres only feature, right? In that case we should provide only an `QueryFragment` implementation for `Pg` instead of a generic one for all database backend.

diesel/src/query_builder/upsert/conditionally_insert_statement.rs
```diff
@@ -0,0 +1,108 @@
...
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        self.insert.walk_ast(out.reborrow())?;
+        out.push_sql(" ");
+        self.where_clause.walk_ast(out.reborrow())?;
+        Ok(())
+    }
+}
+
+impl<Stmt, WhereClause> AsQuery for ConditionallyInsertStatement<Stmt, WhereClause>
```
weiznich, line 53: This currently drops the `filter` part of the query. I would assume that we want to preserve this part, which means that we would need to modify this impl such that it changes the `insert` field to the corresponding `as_query` variant + implement all required traits (`Query`, support for calling `returning`, …).

The other alternative would be to change the general implementation strategy to not introduce a new outer wrapper type, but rather introduce an new internal type in the existing `InsertStatement` struct. So basically add another field to `OnConflictClause` here: https://github.com/diesel-rs/diesel/blob/eff47cc11983bb93adf00afc3b11fd024bdbc541/diesel/src/query_builder/upsert/on_conflict_clause.rs#L11
or any of the inner types. That's possible without breaking change because `OnConflictClause` cannot be named outside of diesel.

diesel/src/query_builder/upsert/conditionally_insert_statement.rs
```diff
@@ -0,0 +1,108 @@
...
+{
+    type SqlType = Stmt::SqlType;
+    type Query = Stmt::Query;
+
+    fn as_query(self) -> Self::Query {
+        Stmt::as_query(self.insert)
+    }
+}
+
+impl<Stmt, WhereClause, Predicate> FilterDsl<Predicate>
```
weiznich, line 65: We likely want to implement `OrFilterDsl` as well?

weiznich (changes requested):
Thanks for the update :+1:

This is now not a breaking change but I believe it's missing quite a bit of functionality that should be there. While going over the change I've noticed that it is probably easier to not introduce a new wrapper type like `ConditionallyInsertStatement` for this but rather store this information inside of for example `OnConflictClause` (or any child type).

Other than that I would like to see an changelog entry for such a feature + some sort of mentioning is some documentation. (By adding documentation to some of the entry points or by adding an example to `.on_confict` or similar methods)

neogenie:
> Thanks for the update 👍
> 
> While going over the change I've noticed that it is probably easier to not introduce a new wrapper type like `ConditionallyInsertStatement` for this but rather store this information inside of for example `OnConflictClause` (or any child type).

@weiznich 

Your idea seemed to me not bad and certainly cleaner and simpler than the introduction of a new wrapper type.
And changed the PR. Please take a look.

If everything is ok according to the concept, then I will add the documentation and the necessary restrictions on types

diesel/src/query_builder/upsert/on_conflict_clause.rs
```diff
@@ -3,41 +3,62 @@ use super::on_conflict_target::*;
...
 use crate::backend::Backend;
 use crate::insertable::*;
+use crate::query_builder::where_clause::NoWhereClause;
 use crate::query_builder::*;
 use crate::result::QueryResult;
 
 #[doc(hidden)]
 #[derive(Debug, Clone, Copy)]
-pub struct OnConflictValues<Values, Target, Action> {
+pub struct OnConflictValues<Values, Target, Action, WhereClause> {
```
weiznich, line 12:
```suggestion
pub struct OnConflictValues<Values, Target, Action, WhereClause = NoWhereClause> {
```

That makes it backward compatible with the old definition

diesel/src/query_builder/upsert/on_conflict_clause.rs
```diff
@@ -3,41 +3,62 @@ use super::on_conflict_target::*;
...
     type QueryId = ();
 
     const HAS_STATIC_QUERY_ID: bool = false;
 }
 
-impl<Values> OnConflictValues<Values, NoConflictTarget, DoNothing> {
+impl<Values> OnConflictValues<Values, NoConflictTarget, DoNothing, NoWhereClause> {
     pub(crate) fn do_nothing(values: Values) -> Self {
-        Self::new(values, NoConflictTarget, DoNothing)
+        Self::new(values, NoConflictTarget, DoNothing, NoWhereClause {})
```
weiznich, line 30:
```suggestion
        Self::new(values, NoConflictTarget, DoNothing, NoWhereClause)
```

as that's the idiomatic way to instance zero sized structs

diesel/src/query_builder/upsert/on_conflict_clause.rs
```diff
@@ -48,19 +69,22 @@ where
     }
 }
 
-impl<DB, Values, Target, Action> QueryFragment<DB> for OnConflictValues<Values, Target, Action>
+impl<DB, Values, Target, Action, WhereClause> QueryFragment<DB>
```
weiznich, line 72: This needs to be split into two seperate impls. One for for `NoWhereClause` that's there for all supported backends and one for `WhereClause{}` that's only there for postgres

In addition I would like to see an additional compile fail test that demonstrates that this does not compile with non-postgres backends.

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -76,8 +79,12 @@ where
     /// ```
     pub fn on_conflict_do_nothing(
         self,
-    ) -> InsertStatement<T, OnConflictValues<U::ValueClause, NoConflictTarget, DoNothing>, Op, Ret>
-    {
+    ) -> InsertStatement<
+        T,
+        OnConflictValues<U::ValueClause, NoConflictTarget, DoNothing, NoWhereClause>,
```
weiznich, line 84: With the added default value for the new generic argument this change is not necessary anymore.

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -236,15 +243,18 @@ impl<T: QuerySource, U, Op, Ret, Target>
     ///
     /// [`on_conflict_do_nothing`]: crate::query_builder::InsertStatement::on_conflict_do_nothing()
     /// [`on_conflict`]: crate::query_builder::InsertStatement::on_conflict()
-    pub fn do_nothing(self) -> InsertStatement<T, OnConflictValues<U, Target, DoNothing>, Op, Ret> {
+    pub fn do_nothing(
+        self,
+    ) -> InsertStatement<T, OnConflictValues<U, Target, DoNothing, NoWhereClause>, Op, Ret> {
```
weiznich, line 248: Neither this change should be necessary given the default parameter

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -236,15 +243,18 @@ impl<T: QuerySource, U, Op, Ret, Target>
...
         let target = self.target;
-        self.stmt
-            .replace_values(|values| OnConflictValues::new(values, target, DoNothing))
+        self.stmt.replace_values(|values| {
+            OnConflictValues::new(values, target, DoNothing, NoWhereClause {})
```
weiznich, line 251:
```suggestion
            OnConflictValues::new(values, target, DoNothing, NoWhereClause)
```

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -383,14 +430,56 @@ impl<T: QuerySource, U, Op, Ret, Target>
     pub fn set<Changes>(
         self,
         changes: Changes,
-    ) -> InsertStatement<T, OnConflictValues<U, Target, DoUpdate<Changes::Changeset>>, Op, Ret>
+    ) -> InsertStatement<
+        T,
+        OnConflictValues<U, Target, DoUpdate<Changes::Changeset>, NoWhereClause>,
```
weiznich, line 435: Shouldn't be necessary with the default parameter

diesel/src/upsert/on_conflict_extension.rs
```diff
@@ -383,14 +430,56 @@ impl<T: QuerySource, U, Op, Ret, Target>
...
         Changes: AsChangeset<Target = T>,
     {
         let target = self.target;
         self.stmt.replace_values(|values| {
-            OnConflictValues::new(values, target, DoUpdate::new(changes.as_changeset()))
+            OnConflictValues::new(
+                values,
+                target,
+                DoUpdate::new(changes.as_changeset()),
+                NoWhereClause {},
```
weiznich, line 449:
```suggestion
                NoWhereClause,
```

weiznich (commented):
Thanks for the update. This approach looks much better :+1:

I've left a number of stylistic comments + I would like to see an compile fail test demonstrating that such queries do not compile for non-postgres backends. In addition it would be great to have a changelog entry for this feature

weiznich (approved):
:+1: Looks good now

PR #3451, opened 2022-12-15
Fix #3446

### LINKED ISSUE #3446: Infer_schema! generates joinable for composite Foreign Key
## Setup
The following repository was used to setup and reproduce the problem: https://github.com/lochetti/diesel_joinable_composite_fk

The problem is reproduced just by running the migration with the print_schema enabled.

### Versions

- **Rust**: 1.65
- **Diesel**: 2.0.2
- **Database**: Postgresql
- **Operating System**: Fedora 37

### Feature Flags

- **diesel:**["32-column-tables", "chrono", "numeric", "postgres", "r2d2", "without-deprecated"]

## Problem Description
According to [the documentation](https://docs.diesel.rs/diesel/macro.joinable.html)  we didn't expect to have a joinable being created between the second and third tables, since the only relationship between them is the composite foreign key.

A more dangerous implication of this problem is that it seems that diesel chooses the first column to be the joinable which could lead to wrong joining strategies. This can be verified by changing the order of the columns in the definition of the foreign key, in the migration file, which generates a different join strategy.

I'm trying to be sure that the infer_schema only creates joinables that are correct, and in that case the joinable created could not be correct.

I expect to don't have a joinable between those tables, according to the documentation provided.

See the schema.rs file in the [setup repository](https://github.com/lochetti/diesel_joinable_composite_fk/blob/main/src/schema.rs).

No.

### Steps to reproduce
1. clone the repository
2. configure a postgres database_url in the .env file
3. Run `diesel setup` .
4. See the outputed schema.rs file

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues. 

weiznich:
This commit fixes an issue where diesel print-schema incorrectly emitted a `joinable!` call for compound foreign keys by filtering out such foreign key constraints more explicitly.

### DIFF
--- a/diesel_cli/tests/print_schema.rs
+++ b/diesel_cli/tests/print_schema.rs
@@ -270,6 +270,17 @@ fn print_schema_reserved_names() {
     test_print_schema("print_schema_reserved_name_mitigation_issue_3404", vec![])
 }
 
+#[test]
+#[cfg(feature = "postgres")]
+fn print_schema_regression_3446_ignore_compound_foreign_keys() {
+    test_print_schema("print_schema_regression_3446_compound_keys", vec![])
+}
+
+#[test]
+fn print_schema_several_keys_with_compound_key() {
+    test_print_schema("print_schema_several_keys_with_compound_key", vec![])
+}
+
 #[cfg(feature = "sqlite")]
 const BACKEND: &str = "sqlite";
 #[cfg(feature = "postgres")]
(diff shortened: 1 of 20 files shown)


diesel_cli/tests/print_schema.rs
```diff
@@ -270,6 +270,12 @@ fn print_schema_reserved_names() {
     test_print_schema("print_schema_reserved_name_mitigation_issue_3404", vec![])
 }
 
+#[test]
+#[cfg(feature = "postgres")]
+fn print_schema_regression_test_for_3446() {
```
JeremyLARDENOIS, line 275: Is it common to name a function by his issue? Shouldn't be better to name it something like `test_print_schema_with_empty_vec()` or something like this ?

weiznich: It's a good suggestion to include also what's tested in the name. Will change that :+1:

PR #3453, opened 2022-12-15
Disable default features from chrono

### LINKED ISSUE #3452: Diesel 1.4.8 requries Chrono with `oldtime`
### Versions

- **Diesel:** 1.4.8

### Feature Flags

- **diesel:** `chrono`

## Problem Description

Diesel 1.4.8 requires `chrono` as:
```
chrono = { version = "0.4", optional = true }
```

While Diesel 2.0.2 requiers `chrono` as:
```
chrono = { version = "0.4.20", optional = true, default-features = false, features = ["clock", "std"] }
```

The issue is that the default features from `chrono` contain the `oldtime` feature, which in turn requires a version of `time` that has [known security issues](https://github.com/time-rs/time/issues/293). Is there a chance we can get a new patch release for the 1.4 series disabling the default features from `chrono`?

gahag-cw:
Fixes #3452

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["database"]
 [dependencies]
 byteorder = "1.0"
 diesel_derives = "~1.4.0"
-chrono = { version = "0.4", optional = true }
+chrono = { version = "0.4", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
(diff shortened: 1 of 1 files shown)


weiznich:
This is essentially a duplicate of https://github.com/diesel-rs/diesel/pull/3253.

As written there: We do not have the capacity to maintain old releases in our free time. The 2.0 release contains the required fix. Please update your code to the new diesel version to get access to this (and other) fixes.

PR #3455, opened 2022-12-16
Clippy fixes

Ten0:


weiznich:
I've submitted and merged #3456 which contained an additional fix to make all of the clippy checks passing.

PR #3457, opened 2022-12-19
Use `#[derive(MultiConnection)]`  in the implementation of diesel-cli

weiznich:


diesel_cli/src/infer_schema_internals/foreign_keys.rs
```diff
@@ -6,13 +6,10 @@ use super::table_data::TableName;
 use crate::database::InferConnection;
 
 pub fn remove_unsafe_foreign_keys_for_codegen(
-    database_url: &str,
+    conn: &mut InferConnection,
```
JeremyLARDENOIS, line 9: Hello, just a question for me. I don't understand the meaning of &mut and why you need it here. Can you explain it to me?

weiznich: `&mut Type` represents a mutable reference to value of `Type`. Methods on the [`Connection` trait](https://docs.diesel.rs/master/diesel/connection/trait.Connection.html) require mutable access to the corresponding type instance. We do use a mutable reference here to being able to:
* Call connection methods inside of this function (requires mutable access)
* Reuse the connection after calling this function (requires the reference)

Hopefully that gives you at least some idea what that's supposed to mean. See [this section of the rust book](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#mutable-references) for more details. Feel free to ask for more details if something there is unclear.

JeremyLARDENOIS: Thank you for your answer, it was very helpful. I think I will keep learning rust :smile: .

PR #3474, opened 2023-01-13
Fix #3470

### LINKED ISSUE #3470: A `ReadOnlyTransaction` error within an inner transaction doesn't rollback the savepoint
### Versions

- **Rust:** 1.66
- **Diesel:** 2.0.2
- **Database:** PostgreSQL 14.6
- **Operating System:** Ubuntu 22.10

### Feature Flags

- **diesel:** ["postgres"]

## Problem Description

Starting with diesel `2.0.0-rc.1` (commit 53bad857f30f2f32c14ae295487a88da278fb2fc of #3211) when within a nested transaction, if the inner transaction aborts because it attempts to mutate within a read-only transaction, then the connection is left in a bad state even outside of the inner transaction. Any further commands result in `current transaction is aborted, commands ignored until end of transaction block` until the outermost transaction is also rolled back. It appears that the SAVEPOINT for the inner transaction is not actually rolled back, as the equivalent series of SQL commands (including a `ROLLBACK TO SAVEPOINT ...`) does not result in such an error in `psql`.

I am trying to port crates.io to diesel 2.0. Our tests are run within an outer transaction that rolls back when the test completes. In production the code will continue to work as expected as this code path has just the one transaction, but the additional transaction wrapping the entire test causes this test to fail.

It is expected that if a nested transaction encounters a `ReadOnlyTransaction` error then it should still rollback to the appropriate SAVEPOINT so that future read-only commands (such as SELECT) can be executed.

The observed output is an error stating `current transaction is aborted, commands ignored until end of transaction block`.

No additional errors are seen.

### Steps to reproduce

Create a simple test table with `cargo run` and then run `cargo test`. In diesel releases starting with `2.0.0-rc.1` the 2nd test fails with `Error: DatabaseError(Unknown, "current transaction is aborted, commands ignored until end of transaction block")`.

```rust
use diesel::{prelude::*, sql_query};

const DATABASE_URL: &str = "postgres:///";

fn connect() -> PgConnection {
    PgConnection::establish(DATABASE_URL).unwrap()
}

fn main() {
    sql_query("CREATE TABLE rollback_test (id INT PRIMARY KEY, value INT NOT NULL)")
        .execute(&mut connect())
        .unwrap();
}

#[cfg(test)]
mod tests {
    use super::*;
    use diesel::update;

    diesel::table! {
        rollback_test (id) {
            id -> Int4,
            value -> Int4,
        }
    }

    fn update_with_fallback_query(conn: &mut PgConnection) -> QueryResult<()> {
        conn.transaction(|conn| {
            sql_query("SET TRANSACTION READ ONLY").execute(conn)?;
            dbg!(update(rollback_test::table)
                .set(rollback_test::value.eq(0))
                .execute(conn))
        })
        .or_else(|_| sql_query("SELECT 1").execute(conn))
        .map(|_| ())
    }

    #[test]
    fn test_rollback_of_readonly() -> QueryResult<()> {
        update_with_fallback_query(&mut connect())
    }

    /// Test that start failing at rev="53bad857f30f2f32c14ae295487a88da278fb2fc"
    #[test]
    fn test_rollback_of_readonly_within_outer_transaction() -> QueryResult<()> {
        connect().transaction(|conn| update_with_fallback_query(conn))
    }
}
```

For comparison, here are similar SQL commands run via `psql`:

```sql
BEGIN;
SAVEPOINT rollback_test;
SET TRANSACTION READ ONLY;
UPDATE rollback_test SET value=1;
-- ERROR:  cannot execute UPDATE in a read-only transaction
SELECT 1;
-- ERROR:  current transaction is aborted, commands ignored until end of transaction block
ROLLBACK TO SAVEPOINT rollback_test;
SELECT 1;
-- No longer an error, as we've left the inner transaction.
-- So diesel must not be issuing the expected savepoint rollback.
COMMIT;
```

## Checklist

weiznich:
This commit tries to address an issue where we do not issue rollback to savepoint commands due to libpq reporting a transaction as broken. It turns out that transactions are sometimes recoverable by just rolling back to a previous savepoint instead of rolling back the outer transaction.

We address this issue by always executing the `ROLLBACK TO SAVEPOINT` command even if a connection is required as needing a top level rollback. This was previously skipped. We do ignore any errors from that attempt if the requires top level rollback flag is set. If the `ROLLBACK TO SAVEPOINT` command succeeds, we consider the transaction as recovered and disable the requires top level rollback flag.

cc @jtgeibel

### DIFF
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -1017,9 +1017,9 @@ mod test {
             .execute(conn)?;
             conn.transaction(|conn| {
                 sql_query("SET TRANSACTION READ ONLY").execute(conn)?;
-                dbg!(crate::update(rollback_test::table)
+                crate::update(rollback_test::table)
                     .set(rollback_test::value.eq(0))
-                    .execute(conn))
+                    .execute(conn)
             })
             .map(|_| {
                 panic!("Should use the `or_else` branch");
@@ -1060,6 +1060,7 @@ mod test {
         .execute(conn)
         .unwrap();
 
+        let start_barrier = Arc::new(Barrier::new(2));
         let commit_barrier = Arc::new(Barrier::new(2));
 
         let other_start_barrier = start_barrier.clone();
@@ -1080,7 +1081,7 @@ mod test {
                         conn
                     ).transaction_depth().expect("Transaction depth")
                 );
-                let _ = rollback_test2::table.load::<(i32, i32)>(conn)?;
+                rollback_test2::table.load::<(i32, i32)>(conn)?;
                 crate::insert_into(rollback_test2::table)
                     .values((rollback_test2::id.eq(1), rollback_test2::value.eq(42)))
                     .execute(conn)?;
@@ -1091,6 +1092,7 @@ mod test {
                             conn
                         ).transaction_depth().expect("Transaction depth")
                     );
+                    start_barrier.wait();
                     commit_barrier.wait();
                     let r = rollback_test2::table.load::<(i32, i32)>(conn);
                     assert!(r.is_err());
@@ -1102,9 +1104,14 @@ mod test {
                         conn
                     ).transaction_depth().expect("Transaction depth")
                 );
-                assert!(r.is_ok());
+                assert!(
+                    matches!(r, Err(crate::result::Error::RollbackTransaction)),
+                    "rollback failed (such errors should be ignored by transaction manager): {}",
+                    r.unwrap_err()
+                );
                 let r = rollback_test2::table.load::<(i32, i32)>(conn);
                 assert!(r.is_err());
+                // This should fail because of ser failure
                 Ok(())
             });
             assert!(r.is_err());
@@ -1117,6 +1124,7 @@ mod test {
         });
 
         let t2 = std::thread::spawn(move || {
+            other_start_barrier.wait();
             let conn = &mut crate::test_helpers::pg_connection_no_transaction();
             assert_eq!(
                 None,
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -333,39 +333,37 @@ where
         let transaction_state = Self::get_transaction_state(conn)?;
 
         let rollback_sql = match transaction_state.in_transaction {
-            Some(ref mut in_transaction) => {
-                match in_transaction.transaction_depth.get() {
-                    1 => Cow::Borrowed("ROLLBACK"),
-                    depth_gt1 => {
-                        if in_transaction.top_level_transaction_requires_rollback {
-                            // There's no point in *actually* rolling back this one
-                            // because we won't be able to do anything until top-level
-                            // is rolled back.
-
-                            // To make it easier on the user (that they don't have to really look
-                            // at actual transaction depth and can just rely on the number of
-                            // times they have called begin/commit/rollback) we don't mark the
-                            // transaction manager as out of the savepoints as soon as we
-                            // realize there is that issue, but instead we still decrement here:
-                            in_transaction.transaction_depth = NonZeroU32::new(depth_gt1 - 1)
-                                .expect("Depth was checked to be > 1");
-                            return Ok(());
-                        } else {
-                            Cow::Owned(format!(
-                                "ROLLBACK TO SAVEPOINT diesel_savepoint_{}",
-                                depth_gt1 - 1
-                            ))
-                        }
-                    }
-                }
-            }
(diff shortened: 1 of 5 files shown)


diesel/src/connection/transaction_manager.rs
```diff
@@ -333,39 +333,37 @@ where
...
 
         match conn.batch_execute(&rollback_sql) {
             Ok(()) => {
-                Self::get_transaction_state(conn)?
-                    .change_transaction_depth(TransactionDepthChange::DecreaseDepth)?;
+                let tm_status = &mut Self::get_transaction_state(conn)?;
+                if let ValidTransactionManagerStatus {
+                    in_transaction:
+                        Some(InTransactionStatus {
+                            ref mut top_level_transaction_requires_rollback,
```
Ten0, line 352: We should probably rename this to top_level_transaction_may_require_rollback to explicit the definition that wouldn't match https://github.com/diesel-rs/diesel/issues/3470#issuecomment-1381598316 anymore

diesel/src/connection/transaction_manager.rs
```diff
@@ -333,39 +333,37 @@ where
...
+                            ..
+                        }),
+                } = tm_status
+                {
+                    // rolling back the save point did work, even if the connection
+                    // was marked to require a top level rollback. This likely means that
+                    // we recovered the transaction state and we are able to perform
```
Ten0, line 359: So I've tested this and actually it's not true :(

Conn1:
```sql
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE ;
SELECT * FROM users;
SAVEPOINT sp1;
INSERT INTO users(name) VALUES ('u1');
SAVEPOINT sp2;
```

Conn2:
```sql
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE ;
SELECT * FROM users;
INSERT INTO users(name) VALUES ('u2');
COMMIT;
```

Conn1:
```sql
SELECT * FROM users; -- errors
ROLLBACK TO SAVEPOINT sp2; -- succeeds
SELECT * FROM users; -- still errors
```

Ten0: I think this approach still has potential - we can may be able to live with a connection on which we only know that it "may" require rollback, ignoring otherwise relevant errors and behaving nicely with regards to rollback attempt on commit fail - I don't think we need to know that we *for sure* require rollback.

I'll try to add a commit on that path here.

Also maybe pg gives us information about whether the connection actually was fixed through the transaction_status, which would enable us to be more accurate there.

weiznich: I've pushed your example as (failing) testcase. I need to stop working on that now, can probably continue on Monday.

Ten0: Ahaha fun fact: if hitting "commit" after receiving a serialization failure (so same commands as previously), PG returns that the commit has succeeded, but in fact it was actually rolled back.

Ten0:
CI won't run completely due to a Clap upgrade that changed the format of some errors.

Unfortunately when trying to test `diesel_cli` locally I'm getting `thread panicked while panicking. aborting.`, so that's hard for me to fix.

Otherwise I think the current implementation is likely to be alright.
Change compared to previous behavior:
When hitting a ser failure within a savepoint:
previous:
```sql
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
SELECT * FROM users;
INSERT INTO users(name) VALUES ('Ten');
SAVEPOINT sp1;
SELECT * FROM users; -- Error
ROLLBACK;
```

New:
```sql
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
SELECT * FROM users;
INSERT INTO users(name) VALUES ('Ten');
SAVEPOINT sp1;
SELECT * FROM users; -- ser failure error
ROLLBACK TO SAVEPOINT sp1; -- maybe this will fix it? (`requires_rollback_maybe_up_to_top_level = false` because PG will tell us again if that didn't fix it)
COMMIT; -- ser failure error (transaction will be rolled back at this point)
```

Note that as I mentioned at https://github.com/diesel-rs/diesel/pull/3474#discussion_r1069586954:
```sql
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
SELECT * FROM users;
INSERT INTO users(name) VALUES ('Ten');
SAVEPOINT sp1;
SELECT * FROM users; -- ser failure error
COMMIT; -- Returns a success but actually this did rollback
```
so we're walking in a mine field here ^^

weiznich (commented):
Thanks @Ten0 for these changes. Looks good from my side :+1:

PR #3477, opened 2023-01-14
mention possibility to bundle SQLite

schlamar:
See https://github.com/diesel-rs/diesel/discussions/3255


Mingun:
I think, that adding an example part of toml would be even more useful

schlamar:
https://github.com/diesel-rs/diesel/pull/3478

PR #3496, opened 2023-01-27
Introduce a `--diff-schema` flag for `diesel migration generate`

weiznich:
If that flag is set diesel-cli will compare the current `schema.rs` file against the provided database and generate a corresponding stub for up/down migrations. This feature is not designed to generate perfect migrations all the times, but to generate the "boring" stuff most of the time.

It currently handles:
* Adding/Droping tables
* Adding/Dropping columns

It mostly generates the correct sql side types in most of the cases, although that depends a bit on what users specify (see the test cases for examples). It is expected that users need to check and potentially modify the generated migrations. I would expect that the type mapping get some improvements over time as people start using this feature

It cannot handle more complex patterns like changing primary keys or adding unique indices or migrating the data stored in tables itself. This is by definition not part of the designed feature set and will always require manual input from the user.


diesel_cli/src/migrations/diff_schema.rs
```diff
@@ -0,0 +1,596 @@
...
+
+#[allow(clippy::enum_variant_names)]
+enum SchemaDiff {
+    DropTable {
+        table: TableName,
+        columns: TableData,
+        foreign_keys: Vec<ForeignKeyConstraint>,
+    },
+    CreateTable {
+        to_ceate: TableDecl,
```
Mingun, line 281: Probably
```suggestion
        to_create: TableDecl,
```

weiznich: Good point :+1:
(Fixed in one of the force pushes)

weiznich:
@diesel-rs/reviewers That is now ready for a final review

PR #3504, opened 2023-02-04
Fix `FileBasedMigrations` documentation

DianaNites:
The `run_in_transaction` example for `FileBasedMigrations` has most of its documentation hidden by rustdoc because a single `#` is used to [hide portions of the example](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example), while the intent seems to have been toml comments

This PR escapes them by using `##` instead so it renders correctly


JeremyLARDENOIS:
Hello @DianaNites, the file `20160107082941_create_posts/metadata.toml` doesn't appear in the file architecture, could you add it?

DianaNites:
Done

weiznich (approved):

weiznich:
Thanks :+1:

PR #3514, opened 2023-02-15
Fix copy/paste issue in group_by documentation.

AndrolGenhald:
It looks like the `group_by()` documentation was probably copied from `order()`, but this particular phrase was missed when tweaking it to apply to 'group_by' instead of `order()`.


weiznich:
Thanks :+1:

Issue #3386, opened 2022-10-30
print_schema.rs throws error for MySQL when enum present in schema


adwhit:
### Versions

- **Rust:**: 1.64
- **Diesel:** 2.0.1
- **Database:**: MySQL 8
- **Operating System**: MacOS

### Feature Flags

- **diesel:** postgres,mysql,sqlite

## Problem Description

Trying to test my crate https://github.com/adwhit/diesel-derive-enum.

I have a very simple schema to run against mysql (running in a docker container).

```sql
CREATE TABLE simple (
  id SERIAL PRIMARY KEY,
  some_value ENUM('foo', 'bar', 'baz_quxx') NOT NULL
);
```
This is valid mysql, see https://dev.mysql.com/doc/refman/8.0/en/enum.html.

When I execute `diesel-cli` I get an error:

`$ diesel --database-url mysql://root:mysql@127.0.0.1:3306/db migration run`
```
Encountered unknown type for Mysql: enum
thread 'main' panicked at 'internal error: entered unreachable code: Mysql only supports a closed set of types.
                         If you ever see this error message please open an issue at https://github.com/diesel-rs/diesel containing a dump of your schema definition.', diesel_cli-2.0.1/src/print_schema.rs:313:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The same problem came up here: https://github.com/diesel-rs/diesel/discussions/3378 -  unresolved.

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues)  and the [disussion forum](https://github.com/diesel-rs/diesel/discussions)  for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for opening this issue. This is expected behavior as we do not have any support for the mysql enum type in diesel yet. At least for me its not clear how exactly a type safe abstraction for such types would look like as there is one shared underlying type (`enum`) that is used for fundamentally different types, for example if you define different enum columns (with different possible values) on the same table. We would need a design for handling these situations first. I'm open for suggestions (+ implementations here), but that's likely nothing I personally will work on for the next few releases.

What definitively should be done is improving the error message, so that it states that `enum` is currently unsupported.

weiznich:
[...]

PR #3522, opened 2023-02-21
Add mysql enum and mysql set support to print-schema

peter-todorov:
Related issue https://github.com/diesel-rs/diesel/issues/3386

Clippy and rusfmt are happy. Tested this with:
```shell
cargo check --no-default-features --features mysql
cargo check --no-default-features --features postgres
cargo check --no-default-features --features sqlite
cargo check
```

Generated code in `sql_types`:
DeepinScreenshot_select-area_20230221155014

Generated code in table definition:
DeepinScreenshot_select-area_20230221154948

I'm not sure if there should be an `Enum` suffix on the generated types.

Tested this with a 57 table database, compiling the generated schema yields no errors or warnings, but I haven't tested using it. Also haven't tested generating a schema for sqlite or postgres.


diesel_cli/src/print_schema.rs
```diff
@@ -202,13 +202,21 @@ pub fn output_schema(
...
+                })
+            })
+            .filter(|t| !diesel_provided_types.contains(&t.column_type.rust_name as &str))
             .collect::<Vec<_>>();
 
-        all_types.sort_unstable_by(|a, b| a.rust_name.cmp(&b.rust_name));
-        all_types.dedup_by(|a, b| a.rust_name.eq(&b.rust_name));
+        all_types.sort_unstable_by(|a, b| a.column_type.rust_name.cmp(&b.column_type.rust_name));
+
+        #[cfg(feature = "postgres")]
```
weiznich, line 217: That should be done for all backends, as it simplifies the code and we do note expect to have duplicates for mysql

diesel_cli/src/print_schema.rs
```diff
@@ -326,16 +368,57 @@ impl Display for CustomTypeList {
...
+                            None => String::new(),
+                            Some(f) => f.to_uppercase().collect::<String>() + c.as_str(),
+                        }
+                    };
+
+                    if let Some(ref schema) = t.schema {
+                        writeln!(
+                            out,
+                            "#[diesel(mysql_type(name = \"{mysql_name}\", schema = \"{schema}\"))]"
+                        )?;
```
weiznich, line 412: This can not happen as we do not support schemas for mysql. Also the derive does not support that parameter. (That's mostly a postgres thing).

weiznich (commented):

This already looks good,. I've left two minor comments on how to improve some details.

Otherwise I would like to see an additional test in `diesel_cli/tests/print_schema.rs`: https://github.com/diesel-rs/diesel/blob/8d64a4011bed57851e3cd747df6d855d44ab9993/diesel_cli/tests/print_schema.rs

For writing such tests I usually:

* Copy an existing test case in `tests/print_schema` (for example that one: https://github.com/diesel-rs/diesel/tree/8d64a4011bed57851e3cd747df6d855d44ab9993/diesel_cli/tests/print_schema/print_schema_unsigned)
* Change the `schema.rs` file in the newly created test folder to whatever I would like to test
* Add the test in `tests/print_schema.rs` as additional test case (In that case: Restrict it to be mysql only)
* Run the test with `cargo test` 
* Review the output with `cargo insta review` (requires installing `cargo-insta`)

diesel_cli/src/print_schema.rs
peter-todorov: I think at this point in the code we can only do de-duplication for custom postgres types, since they have separate definitions and unique names. For my mysql database of 30-40 enums and one set this dedup reduces them to two elements: "enum" and "set" and we lose everything else.

weiznich: That's a good point. It's also a bug in your current implementation. It's quite common that users install diesel-cli with more than one feature enabled. In that case this code will be executed for the generating a schema from a mysql database as well. This would result in the issue you've described above. You can reproduce  that by building diesel-cli with the following command: `cargo build --no-default-features --features "postgres mysql"`.

So that's something that needs to be fixed. 

Maybe by just doing something like:

```rust
if backend == Backend::Pg {
    // reason why we only dedup for postgres
    all_types.dedup_by(|a, b| a.column_type.rust_name.eq(&b.column_type.rust_name));
}
```

diesel_cli/src/print_schema.rs
```diff
@@ -557,11 +666,32 @@ impl<'a> Display for ColumnDefinitions<'a> {
...
+                #[cfg(not(feature = "mysql"))]
+                let _ = self.custom_type_defs;
+
                 if column.rust_name == column.sql_name {
-                    writeln!(out, "{} -> {},", column.sql_name, column.ty)?;
+                    writeln!(out, "{} -> {},", column.sql_name, column_type)?;
                 } else {
                     writeln!(out, r#"#[sql_name = "{}"]"#, column.sql_name)?;
-                    writeln!(out, "{} -> {},", column.rust_name, column.ty)?;
+                    writeln!(out, "{} -> {},", column.rust_name, column_type)?;
```
weiznich, line 694: This change is causing the CI failures. The old code was calling the `Display` implementation of the `ColumnType` type here: https://github.com/diesel-rs/diesel/blob/8d64a4011bed57851e3cd747df6d855d44ab9993/diesel_cli/src/infer_schema_internals/data_structures.rs#L74-L97

That generates some important wrapper-types depending on the nullability (and other conditions) for the type. By just using the the type name itself we loose that information. It's likely a better idea to just modify the stored values for `ColumnType::rust_name` at some stage for all custom mysql type fields and leave that code here unchanged

weiznich (approved):
Thanks :+1:

weiznich:
I've rebased the branch to pull in the latest changes from master. That should fix the remaining CI failures.

Ten0:
This is broken - at least on PG.
It generates duplicate use statements for enums, and some non-nullable fields holding these enums are generated as nullable.

weiznich:
I will try to have a look at the bugs today or tomorrow. @Ten0 You don't have a schema that reproduces these bugs that you can easily share?

peter-todorov:
I can look at this later today. Alongside the fix I'll include a unit test for the PG case that got missed by the current suite.

Ten0:
I'll extract the minimal repro today but yesterday I really needed some sleep ;)

weiznich:
No worries. Take your time. It's just easier to start from something than needing to find a reproduction first :wink:

Ten0:
https://github.com/Ten0/tmp_issue_repro_diesel_3522/blob/dfaa227e03294799675d5cd26c78858ed58565c2/src/schema.rs#L15-L18

I didn't manage to reproduce the other bug, but I could identify that it doesn't only mess up nullable vs non-nullable: I also have unrelated types that are turned from `Nullable<Varchar>` into `SomeEnum`

weiznich:
Thanks :heart:

peter-todorov:
@weiznich can I have a crack at this first. I've just downloaded postgres and made a new fork

weiznich:
Sure, just let me know if you need some help somewhere.

PR #3523, opened 2023-02-22
Revert "Add mysql enum and mysql set support to print-schema"

Ten0:
Reverts diesel-rs/diesel#3522
It is broken - at least on PG.
It generates duplicate use statements for enums, and some non-nullable fields holding these enums are generated as nullable.

Reopens #3386

(@weiznich not sure whether you think I should insta merge-when-ready in this kind of scenario?)


weiznich:
@Ten0 It's totally fine to merge such PR's on your own. After all you are a member of the diesel core team, so technically there is no need to ask me (or someone else).

I personally try to go generally with the following rules:

* For small bugfixes (or reverts like that one): It's fine to just merge them if you are sure, otherwise wait one or two days for a review and merge them later
* For larger features: As soon as it is ready, I try to wait at least a week and and try to ask whether someone plans to have a look at the feature. If noone shows up on the PR I try to do another review on my own going over everything again and merge it afterwards.

PR #3530, opened 2023-03-03
A fixed version of #3525

weiznich:


Ten0:
I've tried unpinning my CLI this week and it seems I still have a non-nullable type that ends up being generated as nullable :'(
I should be able to investigate this today.

Ten0:
> Interestingly enough, it seems to mess-up types with fields that have the same name, but on different tables.

So this is the issue for both:
https://github.com/diesel-rs/diesel/blob/0f15f89b6b025c9a49bbf6d612a57aaa1a78452a/diesel_cli/src/print_schema.rs#L642-L652

1. It triggers not if we are generating for mysql, but if we have the mysql feature enabled. So if you have a CLI that happens to *support* mysql because it's compiled with that feature, then it attempts to perform some replacements.
2. This looks up by column name and not by column name + table name, so it picks any column with the same name even if it's for a different table.

In addition it seems the type clones in this area and unnecessary and those could use refs.

PR #3535, opened 2023-03-08
derive(Insertable) into multiple tables

### LINKED ISSUE #1314: Feature : derive(Insertable) into multiple tables
I'm currently trying to use the same struct in many tables and was looking for something like this:

#[derive(Insertable)]
#[table_name="table1", "table2"]

instead of having to make separate structs for each table. Keep up the good work!

6293:
Closes #1314

### DIFF
--- a/diesel_compile_tests/tests/fail/derive/multiple_table_names.stderr
+++ b/diesel_compile_tests/tests/fail/derive/multiple_table_names.stderr
@@ -1,23 +1,39 @@
 error: expected a single table name attribute
-  --> tests/fail/derive/multiple_table_names.rs:29:43
+  --> tests/fail/derive/multiple_table_names.rs:29:30
    |
 29 | #[diesel(table_name = users, table_name = users_)]
-   |                                           ^^^^^^
+   |                              ^^^^^^^^^^
+   |
+   = note: remove this attribute
 
 error: expected a single table name attribute
-  --> tests/fail/derive/multiple_table_names.rs:37:43
+  --> tests/fail/derive/multiple_table_names.rs:37:30
    |
 37 | #[diesel(table_name = users, table_name = users_)]
-   |                                           ^^^^^^
+   |                              ^^^^^^^^^^
+   |
+   = note: remove this attribute
 
 error: expected a single table name attribute
-  --> tests/fail/derive/multiple_table_names.rs:44:43
+  --> tests/fail/derive/multiple_table_names.rs:44:30
    |
 44 | #[diesel(table_name = users, table_name = users_)]
-   |                                           ^^^^^^
+   |                              ^^^^^^^^^^
+   |
+   = note: remove this attribute
 
 error: expected a single table name attribute
-  --> tests/fail/derive/multiple_table_names.rs:52:23
+  --> tests/fail/derive/multiple_table_names.rs:52:10
    |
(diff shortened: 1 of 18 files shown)


diesel_derives/src/model.rs
```diff
@@ -110,7 +111,18 @@ impl Model {
     }
 
     pub fn table_name(&self) -> &Path {
-        self.table_name.as_ref().unwrap_or(&self.name)
+        match self.table_names.len() {
+            0 => &self.name,
+            1 => &self.table_names[0],
+            _ => abort_call_site!("Multiple table name attributes are not supported"),
```
pksunkara, line 117:
```suggestion
            _ => abort_call_site!("expected a single table name attribute"),
```

Let's avoid erroring out on the call site. Let's try doing it on the spans for table names.

diesel_derives/src/model.rs
```diff
@@ -110,7 +111,18 @@ impl Model {
     }
 
     pub fn table_name(&self) -> &Path {
-        self.table_name.as_ref().unwrap_or(&self.name)
+        match self.table_names.len() {
+            0 => &self.name,
+            1 => &self.table_names[0],
```
JosueMolinaMorales, line 116: Is there a reason for just returning the first table name and not any other table name?

Ten0:
Haha I was just thinking about this today!

diesel_derives/src/model.rs
6293: While `Insertable` calls `table_names` method, other derives would call this to check if there is only one table name

diesel_derives/src/model.rs
6293: ty for the insight. applied

diesel_compile_tests/tests/fail/derive/multiple_table_names.stderr
```diff
@@ -0,0 +1,23 @@
+error: expected a single table name attribute
+  --> tests/fail/derive/multiple_table_names.rs:29:43
+   |
+29 | #[diesel(table_name = users, table_name = users_)]
+   |                                           ^^^^^^
```
Mingun, line 5: It would be even better, if the error would be reported in this way:
```suggestion
error: expected a single table name attribute
  --> tests/fail/derive/multiple_table_names.rs:29:43
   |
29 | #[diesel(table_name = users, table_name = users_)]
   |                              ^^^^^^^^^^^^^^^^^^^ note: remove this attribute
```

6293: agree. I updated the code to check the number of `table_name` attributes when building `Model`, instead of when retrieving them from `Model`. Now, the error will point to the `table_name` attribute, with a friendly note. 

Btw I'm not sure how to put the note right after the highlighting carets as you provided, using `proc_macro_error`. Would be nice if you had any advice on this.

weiznich: That's currently not possible, as the rust-lang proc macro crate  does not provide API's for that. See https://github.com/rust-lang/rust/issues/54140 for details.

weiznich (approved):
Thanks for opening this PR. This change looks fine for me :+1:

pksunkara (approved):

PR #3547, opened 2023-03-13
Fix print_schema

Ten0:
That was broken as of e90dfe0cf5549fa35eff81f3f4e01449911d3df4

https://github.com/diesel-rs/diesel/blob/0f15f89b6b025c9a49bbf6d612a57aaa1a78452a/diesel_cli/src/print_schema.rs#L642-L652 https://github.com/diesel-rs/diesel/blob/0f15f89b6b025c9a49bbf6d612a57aaa1a78452a/diesel_cli/src/print_schema.rs#L548

1. It triggers not if we are generating for mysql, but if we have the mysql feature enabled. So if you have a CLI that happens to *support* mysql because it's compiled with that feature, then it attempts to perform some replacements.
2. This looks up by column name and not by column name + table name, so it picks any column with the same name even if it's for a different table.
3. It's not ok to dedup pg types by name (https://github.com/diesel-rs/diesel/blob/0f15f89b6b025c9a49bbf6d612a57aaa1a78452a/diesel_cli/src/print_schema.rs#L227) if lookup for `use` statement generation is made by column - even if that were to be "fixed" to include the table (https://github.com/diesel-rs/diesel/blob/0f15f89b6b025c9a49bbf6d612a57aaa1a78452a/diesel_cli/src/print_schema.rs#L548)

-----

What this does:
- Represents the requirement of import and override for columns by a 2d array of `ColumnType` that zips with the definitions arrays of tables and columns, so that remapping is obvious and we don't have to search for it in a custom types array and filter on backend-specific conditions
- Backend-specific dedup/no-dedup of rust types with regards to the `sql_types` module is done in the code that generates the `sql_types` module


Ten0:
cc @peter-todorov

weiznich (approved):

weiznich:
:+1: Thanks for providing this fix

PR #3548, opened 2023-03-14
Allow binding unsized types in AstPass::push_bind_param

leoyvens:
This has the cost of some allocation overhead for `DebugQuery`, but it is convenient to be able to use `push_bind_param` with `str` and slices. In my particular use case this makes migrating to Diesel 2.0 easier.

### DIFF
--- a/diesel/src/query_builder/ast_pass.rs
+++ b/diesel/src/query_builder/ast_pass.rs
@@ -69,7 +69,10 @@ where
         }
     }
 
-    pub(crate) fn debug_binds(formatter: &'a mut Vec<&'b dyn fmt::Debug>, backend: &'b DB) -> Self {
+    pub(crate) fn debug_binds(
+        formatter: &'a mut Vec<Box<dyn fmt::Debug + 'b>>,
+        backend: &'b DB,
+    ) -> Self {
         AstPass {
             internals: AstPassInternals::DebugBinds(formatter),
             backend,
@@ -207,7 +210,7 @@ where
     pub fn push_bind_param<T, U>(&mut self, bind: &'b U) -> QueryResult<()>
     where
         DB: HasSqlType<T>,
-        U: ToSql<T, DB>,
+        U: ToSql<T, DB> + ?Sized,
     {
         match self.internals {
             AstPassInternals::ToSql(ref mut out, _) => out.push_bind_param(),
@@ -216,7 +219,7 @@ where
                 ref mut metadata_lookup,
             } => collector.push_bound_value(bind, metadata_lookup)?,
             AstPassInternals::DebugBinds(ref mut f) => {
-                f.push(bind);
+                f.push(Box::new(bind));
             }
             AstPassInternals::IsNoop(ref mut result) => **result = false,
             _ => {}
@@ -227,7 +230,7 @@ where
     pub(crate) fn push_bind_param_value_only<T, U>(&mut self, bind: &'b U) -> QueryResult<()>
     where
         DB: HasSqlType<T>,
-        U: ToSql<T, DB>,
+        U: ToSql<T, DB> + ?Sized,
     {
         match self.internals {
             AstPassInternals::CollectBinds { .. } | AstPassInternals::DebugBinds(..) => {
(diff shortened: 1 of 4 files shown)


weiznich:
Thanks for opening this PR. Can you provide an example where this might be useful? Also we need a test case for that somewhere in diesel_tests.

leoyvens:
@weiznich Thank you for your review, I have added a test. In our use case, we have many custom impls of `QueryFragment`, and are adapting that to the new lifetime requirements of `AstPass` and `fn walk_ast` in Diesel 2.0. [This commit](https://github.com/graphprotocol/graph-node/commit/b012cbe05a59e8705f544c3b035ef1d2d983e7ad) shows a few code locations that would take advantage of this change to the `push_bind_param` to make upgrading easier.

diesel/src/query_builder/ast_pass.rs
```diff
@@ -69,7 +69,10 @@ where
         }
     }
 
-    pub(crate) fn debug_binds(formatter: &'a mut Vec<&'b dyn fmt::Debug>, backend: &'b DB) -> Self {
+    pub(crate) fn debug_binds(
+        formatter: &'a mut Vec<Box<dyn fmt::Debug + 'b>>,
```
JosueMolinaMorales, line 73: Whats the benefit of using `Box<>` here?

diesel/src/query_builder/ast_pass.rs
```diff
@@ -207,7 +210,7 @@ where
     pub fn push_bind_param<T, U>(&mut self, bind: &'b U) -> QueryResult<()>
     where
         DB: HasSqlType<T>,
-        U: ToSql<T, DB>,
+        U: ToSql<T, DB> + ?Sized,
```
JosueMolinaMorales, line 213: Can you explain the reasoning for `?Sized`

JosueMolinaMorales (commented):
Cool stuff! I left some questions to get a better since of the PR. Thanks!

diesel/src/query_builder/ast_pass.rs
weiznich: That's required because we cannot store references there anymore due to the fact that it is not possible to (up)cast a trait object (something that's not sized) to another trait object (dyn Debug). That would be required for the changed signature of `push_bind_param`

diesel/src/query_builder/ast_pass.rs
weiznich: That's required so that we can have something like `AstPass:push_bind_param::<Text, str>(…, "abc")` as `str` is not `Sized` there. So this requirement relaxes the bounds for this type, by allowing more types there.

weiznich (approved):

diesel/src/query_builder/ast_pass.rs
Ten0: Ah I should have commented this yesterday... This probably means that accepting this change means we won't ever be able to use &dyn AnyHelperTraitForBindingPush anywhere in the `push_bind_param` interface except for debugging, because the impact on perf would be too large.

I suspect we probably wouldn't want to do that in most cases because likely that would prevent the compiler form const-optimizing it, but maybe for `push_bind_param_value_only` that would end up being acceptable... so I'm not sure if that isn't risky for the future...
I was curious as well to see what the concrete use-case that would be provided as answer to https://github.com/diesel-rs/diesel/pull/3548#issuecomment-1468680101 would look like to try and make an informed weighing of what it enables vs what it blocks.

weiznich: @Ten0 I'm not 100% sure that I can follow your argumentation there. Can you provide a few more details about your ideas around `AnyHelperTraitForBindingPush`?

Ten0: The idea is that since this prevents from using &dyn XX internally inside this function, this prevents any design where we would want to go with &dyn XXX at some point in some of the passes. Currently there is only `Debug` where that was the case, but I'm not sure if there's really no chance that we would want to do a design that uses this at some point, and I'm not sure if the gain of not having to add an extra `&` on the user side is worth it (if that part of the code did compile, otherwise it's definitely worth it).
That being said I don't have a strong opinion on it, just wanted to raise the point.

leoyvens: Unfortunately just adding a `&` wasn't sufficient, because then we had something like `&'temp &'b str` which fixed the sizedness but got us a lifetime error because the reference now lives for `'temp` instead of `'b` as required by `fn push_bind_param`.

Ten0: OK that was my question at https://github.com/diesel-rs/diesel/pull/3548#discussion_r1138872331
LGTM then

PR #3553, opened 2023-03-17
Deny unsafe code

weiznich:
This commit denies unsafe code. It allows the usage of unsafe code only in specific locations and add some comments why it's allowed there. I hope that this helps in the future to detect new unsafe code locations and discuss there need.

### DIFF
--- a/diesel/src/mysql/connection/stmt/iterator.rs
+++ b/diesel/src/mysql/connection/stmt/iterator.rs
@@ -1,5 +1,4 @@
 #![allow(unsafe_code)] // module uses ffi
-                       //
 use std::cell::{Ref, RefCell};
 use std::rc::Rc;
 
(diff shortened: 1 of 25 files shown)


diesel/src/mysql/connection/stmt/iterator.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unsafe_code)] // module uses ffi
+                       //
```
JohnTitor, line 2: nit: This line seems unnecessary.

JohnTitor (approved):
Sounds great to me 👍

diesel/src/mysql/connection/stmt/iterator.rs
weiznich:
```suggestion
```

PR #3556, opened 2023-03-21
Add some documentation about common compiler error message types that users can encounter if something goes wrong

weiznich:
There are a few quite common error cases that generate large complicated looking messages, but that are quite easy to break down if the error message does not scare you away. This documentation section tries to give users a potential starting point where they need to look for the issue that causes the error message.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -85,6 +85,57 @@
 //! [`ToSql`]: serialize::ToSql
 //! [`FromSql`]: deserialize::FromSql
 //!
+//! ## How to read diesels compile time error messages
+//!
+//! Diesel is known for generating large complicated looking errors. Usually
+//! most of these error messages can be break down easily. The following
+//! section tries to give an overview of common error messages and how to read them.
+//! As a general note it's always usefull to read the complete error message as emitted
+//! by rustc, including the `required because of …` part of the message.
+//! You IDE might hide important parts!
+//!
+//! If you use a nightly compiler you might want to enable the `nightly-error-messages`
+//! feature flag to automatically improve some error messages.
+//!
+//! The following error messages are common:
+//!
+//! * `the trait bound (diesel::sql_types::Integer, …, diesel::sql_types::Text): load_dsl::private::CompatibleType<YourModel, Pg> is not satisfied`
+//!    while trying to execute a query:
+//!    This error indicates a mismatch between what your query returns and what your model struct
+//!    expects the query to return. The fields need to match in terms of field order, field type
+//!    and field count. If your are sure that everything must match double check the enabled diesel
+//!    features (for support for types from other crates) and double check (via `cargo tree`)
+//!    that there is only one version of such a shared crate in your dependency tree.
+//!    Consider using [`#[derive(Selectable)]`](derive@crate::prelude::Selectable) +
+//!    `#[diesel(check_for_backend(diesel::pg::Pg))]`
+//!    to improve the generated error message.
+//! * `the trait bound i32: diesel::Expression is not satisfied` in the context of `Insertable`
+//!    model structs:
+//!    This error indicates a type mismatch between the field you are trying to insert into the database
+//!    and the actual database type. These error messages contain a line
+//!    like ` = note: required for i32 to implement AsExpression<diesel::sql_types::Text>`
+//!    that show both the provided rust side type (`i32` in that case) and the expected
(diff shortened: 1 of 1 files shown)


diesel/src/lib.rs
```diff
@@ -85,6 +85,57 @@
...
+//!    This error indicates a type mismatch between the field you are trying to update and the actual
+//!    database type. Double check your type mapping.
+//! * `the trait bound SomeLargeType: QueryFragment<Sqlite, SomeMarkerType> is not satisfied` while
+//!    trying to execute a query.
+//!    This error message indicates that a given query is not supported by your backend. This usually
+//!    means that you are trying to use SQL features from one SQL dialect on a different database
+//!    system. Double check your query that everything required is supported by the selected
+//!    backend. If that's the case double check that the relevant feature flags are enabled
+//!    (`returning_clauses_for_sqlite_3_35` for enabling support for returning clauses in newer
+//!    sqlite versions)
```
chusteven, line 130:
```suggestion
//!    (for example, `returning_clauses_for_sqlite_3_35` for enabling support for returning clauses in newer
//!    sqlite versions)
```

diesel/src/lib.rs
```diff
@@ -85,6 +85,57 @@
...
+//! * `the trait bound posts::title: SelectableExpression<users::table> is not satisfied` while
+//!    executing a query: This error message indicates that you try to select a field from a table
```
chusteven, line 132:
```suggestion
//!    executing a query: 
//!    This error message indicates that you're trying to select a field from a table
```

Not sure if a new line is needed or not 🤷‍♂️

chusteven (approved):
Looks great to me! Some nits on grammar, but otherwise I think this was a fantastic idea. One question, though: where will this appear? Like... I imagine a lot of ppl show up in the issues page or gitter and ask about "what does this scary compiler message mean?" and you'll have to redirect them to this file.

Is there a way to get ahead of these questions? Like... being in `lib.rs` does that mean it will show up on the home page of the docs for `diesel` [here](https://docs.rs/diesel/2.0.3/diesel/)? Or just generally wondering how ppl will even know that this documentation exists

weiznich:
Thanks for the review and the suggestions :+1:. Will include that later in the PR.

> Like... I imagine a lot of ppl show up in the issues page or gitter and ask about "what does this scary compiler message mean?" and you'll have to redirect them to this file.
>
> Is there a way to get ahead of these questions? Like... being in lib.rs does that mean it will show up on the home page of the docs for diesel here? Or just generally wondering how ppl will even know that this documentation exists

This will appear in the API documentation in basically the linked location (just above the feature flag section). Additionally it might be a good idea to put a FAQ section into the readme and/or the web page that links this section (and others that answer other commonly questions).

chusteven:
> This will appear in the API documentation in basically the linked location (just above the feature flag section). Additionally it might be a good idea to put a FAQ section into the readme and/or the web page that links this section (and others that answer other commonly questions).

Ah I see it'll show up on this page: https://docs.rs/diesel/latest/diesel/

I think including it in a FAQ section in the `README` and the web page is a decent idea, but probably not worth going too crazy before we actually see lots and lots of ppl asking about it :P

diesel/src/lib.rs
```diff
@@ -85,6 +85,57 @@
...
+//!
+//! * `the trait bound (diesel::sql_types::Integer, …, diesel::sql_types::Text): load_dsl::private::CompatibleType<YourModel, Pg> is not satisfied`
+//!    while trying to execute a query:
+//!    This error indicates a mismatch between what your query returns and what your model struct
+//!    expects the query to return. The fields need to match in terms of field order, field type
+//!    and field count. If your are sure that everything must match double check the enabled diesel
+//!    features (for support for types from other crates) and double check (via `cargo tree`)
+//!    that there is only one version of such a shared crate in your dependency tree.
+//!    Consider using [`#[derive(Selectable)]`](derive@crate::prelude::Selectable) +
+//!    `#[diesel(check_for_backend(diesel::pg::Pg))]`
```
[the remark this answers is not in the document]
weiznich: `value` in terms of `diesel::pg::Pg`? In that case it accepts a list of diesel backend types that are used for checking.
`value` in terms of improvements: It generates additional code that improves the compiler error message by pointing the the relevant field of the mismatch. See https://github.com/diesel-rs/diesel/pull/3359 for details

PR #3561, opened 2023-03-24
impl<'a, Parent, Child> BelongingToDsl<(&'a [Parent], &'a [Parent])> for Child

Bajix:
See [discussions#3559](https://github.com/diesel-rs/diesel/discussions/3559)

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -166,6 +166,28 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


Bajix:
Yes because a tuple of slices is all that's needed to express wrap around; if the first slice wraps around, then the second slice will start at the beginning or otherwise be empty. This design is taking a page from that of [std::collection::VecDeque::as_slices](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.as_slices).

weiznich (approved):

PR #3568, opened 2023-03-25
Update all the crates to the 2021 edition

weiznich:
This is mostly just running `cargo fix --edition` and `cargo fix --edition-idioms`

### DIFF
--- a/diesel_derives/Cargo.toml
+++ b/diesel_derives/Cargo.toml
@@ -9,11 +9,7 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
 autotests = false
 include = ["src/**/*", "LICENSE-*"]
 rust-version = "1.65.0"
-# we cannot update to newer editions
-# because we relay on a name resolution
-# hack to get a `diesel` item inside of diesel
-# as well as for using diesel-derives in external crates
-edition = "2015"
+edition = "2021"
 
 [dependencies]
 syn = { version = "2.0", features = ["derive", "fold", "full"] }
--- a/diesel_tests/tests/lib.rs
+++ b/diesel_tests/tests/lib.rs
@@ -2,13 +2,9 @@
 
 #[macro_use]
 extern crate assert_matches;
-extern crate chrono;
+
 #[macro_use]
 extern crate diesel;
-#[cfg(feature = "sqlite")]
-extern crate diesel_migrations;
-extern crate dotenvy;
-extern crate quickcheck;
 
 mod alias;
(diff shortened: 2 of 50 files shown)


diesel_derives/Cargo.toml
```diff
@@ -9,11 +9,7 @@ repository = "https://github.com/diesel-rs/diesel/tree/master/diesel_derives"
 autotests = false
 include = ["src/**/*", "LICENSE-*"]
 rust-version = "1.65.0"
-# we cannot update to newer editions
-# because we relay on a name resolution
-# hack to get a `diesel` item inside of diesel
-# as well as for using diesel-derives in external crates
-edition = "2015"
+edition = "2021"
```
pksunkara, line 12: We need this 2015, right?

diesel_tests/tests/lib.rs
```diff
@@ -2,13 +2,9 @@
...
 #[macro_use]
 extern crate assert_matches;
-extern crate chrono;
+
 #[macro_use]
 extern crate diesel;
-#[cfg(feature = "sqlite")]
-extern crate diesel_migrations;
-extern crate dotenvy;
-extern crate quickcheck;
```
pksunkara, line 11: Mistake?

diesel_derives/Cargo.toml
weiznich: We needed it before https://github.com/diesel-rs/diesel/pull/3564
I tried that migration already yesterday and noticed that dependencies. I did not think much about that and went on trying to minimize one of the rust-analyzer type issues with diesel, so that the rust-analyzer team had at least a minimal example without depending on diesel. While doing that someone from that team said that they cannot explain how name resolution even worked for code generated by diesel proc-macro. Because I hit that by accident a bit earlier I was able to explain that and they suggested the `extern crate self as diesel;` fix. That in turn remove the need to use the 2015 edition there, so after all that's all somehow related.

diesel_tests/tests/lib.rs
weiznich: That's fine. `extern crate …` statements are not required anymore, so we can remove them (even those with a `#[cfg]` on.)

diesel_derives/Cargo.toml
pksunkara: Weird, I remember trying this exact trick when I was refactoring the derives. But, it didn't work back then, so I thought maybe the issue was something else and ended up keeping the old hack.

pksunkara (approved):

PR #3571, opened 2023-03-28
Enable reflection on table type 'SYSTEM VERSIONED'

### LINKED ISSUE #3570: `WITH SYSTEM VERSIONING` (mariaDB) causes `print-schema` to fail
### Versions

- **Rust:** rustc 1.70.0-nightly (1db9c061d 2023-03-21)
- **Diesel:** diesel 2.0.1
- **Database:** mariadb
- **Operating System** macos

### Feature Flags

- **diesel:** "mysql_backend", "serde_json", "chrono", "uuid"

## Problem Description

Enabling tables with system versioning (`WITH SYSTEM VERSIONING`) means there is no output in the schema.rs file.

Generate the schema using `diesel print-schema`

Some invocation of the `diesel::table!` macro in `schema.rs`

No output (just the comment)

None

### Steps to reproduce

System versioned tables are supported in MariaDB (also in IBM DB2, and SQL server, current WIP with PG) but using them seems like Diesel can't reflect the tables. For example:

```sql
CREATE TABLE foo (
    x serial,
    primary key (x)
) WITH SYSTEM VERSIONING;
```

Creates no output. Dropping `WITH SYSTEM VERSIONING` generates create output

```rust
diesel::table! {
    foo (x) {
        x -> Unsigned<Bigint>,
    }
}
```

I haven't tried it, but I would have to imagine that application versioning doesn't work either

### Context

MySQL doesn't yet support system versioning, so the issue probably stems from some incompatibility there. There are a few small but growing incompatibilities, so it might make sense to have a `mariadb` connection string option (SQLAlchemy went this route for some features where autodetection didn't work well)

The clause `WITH SYSTEM VERSIONING` adds two columns hidden by default: `ROW_START` and `ROW_END`. It also creates a quasi-column ("period" type) called `SYSTEM_TIME` that can be selected  with a special syntax (`select * from my_table for SYSTEM_TIME all`). These column & period names can also be specified. When updating a row, a system versioned table just duplicates the row and changes the end date, to keep a history of all changes without triggers (and hides the historical rows by default).

There is also application versioning, which is a bit different but uses similar new syntax & temporal constructs.

Ideally, I think it would be good for Diesel to provide a way to work with system versioned tables (e.g. `.filter(foo).system_time(all)`) since they're slowly popping up in more database flavors. But for now, I'd just be happy if the tables reflect to Rust properly, ignoring any system versioning constructs (though adding the ROW_START and ROW_END columns to reflection would be quite awesome)

Links:

- MariaDB has some pretty good docs & examples on the topic https://mariadb.com/kb/en/temporal-tables/
- I had a pretty complete PR to SQLA for adding support of system versioning https://github.com/sqlalchemy/sqlalchemy/pull/7933 Just didn't get merged because of some clunky interworkings with SQLA, but there is good discussion on that PR. My docfile also has a pretty good writeup https://github.com/sqlalchemy/sqlalchemy/blob/cc8a25258e5effb8b40566c34719e325e49a02e6/doc/build/core/temporal.rst

If diesel would want to add some more specific SV/AV support, I'd be happy to help architect

## Checklist

tgross35:
Fixes #3570

There are no side effects to this PR since system versioned tables appear identical to base tables for any frontend.

I'm unsure if a test is possible since it requires a mariadb backend - but if it is possible, let me know.

### DIFF
--- a/diesel_cli/src/infer_schema_internals/information_schema.rs
+++ b/diesel_cli/src/infer_schema_internals/information_schema.rs
@@ -154,7 +154,7 @@ where
             >,
             NotLike<tables::table_name, &'static str>,
         >,
-        Like<tables::table_type, &'static str>,
+        Or<Eq<tables::table_type, &'static str>, Eq<tables::table_type, &'static str>>,
     >: QueryFragment<Conn::Backend>,
     Conn::Backend: QueryMetadata<sql_types::Text>,
 {
@@ -169,7 +169,11 @@ where
         .select(table_name)
         .filter(table_schema.eq(db_schema_name))
         .filter(table_name.not_like("\\_\\_%"))
-        .filter(table_type.like("BASE TABLE"))
+        .filter(
+            table_type
+                .eq("BASE TABLE")
+                .or(table_type.eq("SYSTEM VERSIONED")),
+        )
         .load::<String>(connection)?;
     table_names.sort_unstable();
     Ok(table_names


tgross35:
I have confirmed locally that this works with the following:

```sql
CREATE TABLE foo (
    x serial,
    primary key (x)
) WITH SYSTEM VERSIONING;
```

```
$ target/debug/diesel print-schema --database-url="mysql://root:example@0.0.0.0:4000/db"

// @generated automatically by Diesel CLI.

diesel::table! {
    foo (x, row_end) {
        x -> Unsigned<Bigint>,
    }
}
```

Without this patch, no output is generated

tgross35:
That makes sense too. I think that all that is needed is to ignore the last PK from the group for that table type, so I’ll take a look at how to do that.

I would expect there to be some edge cases but I think that simple change would at least handle the majority. And make it so the joinable macros also get generated, I recently realized they are missing.

tgross35:
Unfortunately I have not had a time to get back to this, anyone is welcome to pick this up if they are willing to

PR #3589, opened 2023-04-11
Change incorrect uses of mysql feature to mysql_backend

oeed:
As noted [here](https://github.com/weiznich/diesel_async/issues/70), some MySQL types are unavailable when compiling with `mysql_backend` when they should be, instead using `mysql`.

`InsertOrIgnore` is the only one I've personally run in to, but I did a full repo search and there were 3 other files that looked like they might also require changing too, but I am not certain – so please do check those closely.


weiznich (approved):
Thanks

PR #3591, opened 2023-04-12
docs(pg): unhide docs from some type aliases

NishantJoshi00:
Removed `#[doc(hidden)]` from 2 PostgreSQL type aliases, as discussed on matrix

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -236,7 +236,6 @@ pub mod sql_types {
     pub struct Uuid;
 
     /// Alias for `Binary`, to ensure `infer_schema!` works
-    #[doc(hidden)]
     pub type Bytea = crate::sql_types::Binary;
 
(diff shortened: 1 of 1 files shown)


diesel/src/pg/types/mod.rs
```diff
@@ -422,7 +421,6 @@ pub mod sql_types {
     #[diesel(postgres_type(oid = 829, array_oid = 1040))]
     pub struct MacAddr;
 
-    #[doc(hidden)]
     /// Alias for `MacAddr` to be able to use it with `infer_schema`.
```
weiznich, line 424:
```suggestion
    /// Alias for `MacAddr` to be able to use it with `diesel print-schema`.
```

weiznich (changes requested):
I think we should adjust the documentation slightly as `infer_schema!` is not a thing anymore.

JohnTitor (approved):

Issue #3566, opened 2023-03-25
MultiConnection not usable for queries containing `Selectable` or boxed queries.


weiznich:
### Versions

- **Rust:**
- **Diesel:** [e548447](https://github.com/diesel-rs/diesel/commit/e5484477b11eeae13054ab4c3b3a006d9eddfe82) 
- **Database:**
- **Operating System**

### Feature Flags

- **diesel: sqlite postgres**

### What are you trying to accomplish?

I've tried the following code:

```rust
use diesel::prelude::*;

table! {
    users {
        id -> Integer,
        name -> Text,
    }
}

#[derive(Selectable, Queryable)]
struct User {
    id: i32,
}

#[derive(diesel::MultiConnection)]
enum DbConnection {
    Sqlite(SqliteConnection),
    Pg(PgConnection)
}

fn main() {
    let conn = &mut DbConnection::establish("…").unwrap();

    let u = users::table.into_boxed().load::<(i32, String)>(conn);

    let u = users::table
        .select(User::as_select())
        .load(conn);
}

```

### What is the expected output?

Code compiles

### What is the actual output?

Code does not compile:

```rust
error[E0277]: the trait bound `BoxedSelectStatement<'_, (diesel::sql_types::Integer, diesel::sql_types::Text), FromClause<users::table>, multi_connection_impl::backend::MultiBackend>: QueryFragment<Sqlite>` is not satisfied
    --> src/main.rs:24:61
     |
24   |     let u = users::table.into_boxed().load::<(i32, String)>(conn);
     |                                       ----                  ^^^^ the trait `QueryFragment<Sqlite>` is not implemented for `BoxedSelectStatement<'_, (diesel::sql_types::Integer, diesel::sql_types::Text), FromClause<users::table>, multi_connection_impl::backend::MultiBackend>`
     |                                       |
     |                                       required by a bound introduced by this call
     |
     = help: the following other types implement trait `QueryFragment<DB, SP>`:
               <BoxedSelectStatement<'a, ST, QS, DB, GB> as QueryFragment<DB, AnsiSqlSelectStatement>>
               <BoxedSelectStatement<'a, ST, QS, DB, GB> as QueryFragment<DB>>
               <BoxedSelectStatement<'a, ST, QS, multi_connection_impl::backend::MultiBackend, GB> as QueryFragment<multi_connection_impl::backend::MultiBackend, MultiSelectStatementSyntax>>
[...]

PR #3597, opened 2023-04-14
Prepare a 2.0.4 Release

weiznich:
As suggested in https://github.com/diesel-rs/diesel/discussions/3594 it might be reasonable to cut a 2.0.4 release to bring the r-a fix faster to our users. 

Thanks for @adamchalmers for preparing the branch.

@diesel-rs/core I would like to cut a release on monday if noone as objections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ default features enabled using some set of dependencies. Those set of dependenci
 an up to date version of the specific dependency. We check this by using the unstable `-Z minimal-version` cargo flag. 
 Increasing the minimal supported Rust version will always be coupled at least with a minor release.
 
-## Unreleased 
+## Unreleased
+
+## [2.0.4] 2023-04-14
 
 ## Added
 
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diesel"
-version = "2.0.3"
+version = "2.0.4"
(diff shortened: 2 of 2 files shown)


diesel_derives/src/table.rs
weiznich, line 1: @adamchalmers Can you remove this file. It shouldn't be here, as the 2.0.x branch uses the `macro_rules` based `table!` macro.

adamchalmers: Done

adamchalmers:
Hmm, the "Check Minimal support rust version (1.56.0)" action is failing because libsqlite3-sys 0.26 requires 1.58 (because it uses `format!("{x}")` syntax).

But I don't know why it's even trying version 0.26, because it should use the minimal version.

nickelc:
Is it possible to pick up #3435 as well?

Ten0:
> Is it possible to pick up xxx as well?

Should we release a 2.1 and get everything in? Or are there breaking changes other ten bugfixes that I missed?

weiznich:
@Ten0 Its my plan to do a 2.1 release sometime soon, but its not ready to release it now due to #3566. I have at least a partial fix for that but I had now time to open a PR yet.

adamchalmers:
> Is it possible to pick up #3435 as well?

I've added this commit to the release. @weiznich let me know if you want me to remove it.

adamchalmers:
OK, all tests pass now!

PR #3617, opened 2023-05-02
Remove cargo's audit allowlist that is no longer needed

isimluk:
The allowlist item has been introduced by #2928 & #2955. The item is no longer needed after chrono has been fixed on 0.4.20 in #3264.


weiznich:
@isimluk I think it should be fine to just update the chrono version in all that dependencies. I would accept a PR for that.

weiznich (approved):

PR #3622, opened 2023-05-09
Fix #3506

### LINKED ISSUE #3506: `diesel migrations redo` doesn't respect `run_in_transaction` migration metadata
### Versions

- **Rust:** rustc 1.67.0 
- **Diesel:** 2.0.3
- **Database:** sqlite 3.40.1
- **Operating System** Arch Linux

### Feature Flags

- **diesel:** "sqlite", "time"
- **diesel_cli:** "sqlite"

## Problem Description

The `diesel migrations redo` command doesn't respect the `run_in_transaction` migration metadata, while `run` and `revert` separately do.

This seems to be because it is [hard-coded](https://github.com/diesel-rs/diesel/blob/8afb1bb47f31a292c7853f6cb48145231994ee7e/diesel_cli/src/main.rs#L452-L474) to use a transaction for anything except mysql, specifically for the redo command.

A migration that needs to handle transactions itself

Specifically temporarily turn off an [sqlite specific pragma](https://www.sqlite.org/pragma.html#pragma_foreign_keys) as dictated [here](https://www.sqlite.org/foreignkeys.html#fk_schemacommands), which cannot be done inside a transaction.

```
Rolling back migration <migration>
Running migration <migration>
```

```
Rolling back migration <migration>
Failed to run <migration> with: cannot start a transaction within a transaction
```

### Steps to reproduce

- `diesel migration generate example`
- Add the following

```toml
# example/metadata.toml
run_in_transaction = false
```

```sql
-- example/up.sql
BEGIN TRANSACTION;
COMMIT TRANSACTION;
```

```sql
-- example/down.sql
BEGIN TRANSACTION;
COMMIT TRANSACTION;
```

- Run `diesel migration run` and see it works as expected
- Run `diesel migration revert` and see it works as expected
- Run `diesel migration redo` and see it fail

## Checklist

weiznich:
This commit fixes #3506 by not running the diesel migration redo command in a transaction if any affected transaction specifies `run_in_transaction = false`.

It also adds some tests for this case.

### DIFF
--- a/diesel_cli/src/migrations/mod.rs
+++ b/diesel_cli/src/migrations/mod.rs
@@ -1,7 +1,7 @@
 use chrono::Utc;
 use clap::ArgMatches;
 use diesel::backend::Backend;
-use diesel::migration::MigrationSource;
+use diesel::migration::{Migration, MigrationSource};
 use diesel::Connection;
 use diesel_migrations::{FileBasedMigrations, HarnessWithOutput, MigrationError, MigrationHarness};
 use std::any::Any;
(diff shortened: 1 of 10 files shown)


diesel_cli/src/migrations/mod.rs
```diff
@@ -273,55 +273,75 @@ fn redo_migrations<Conn, DB>(
...
-        harness.run_migrations(&migrations)?;
+            harness.run_migrations(&migrations)?;
 
-        Ok(())
-    };
+            Ok(())
+        };
 
-    if should_redo_migration_in_transaction(conn) {
+    if !should_use_not_use_transaction && should_redo_migration_in_transaction(conn) {
```
hasnainzeenwala, line 344: since there are circumstances now when migrations won't be redone as a transaction, the comment on line 266 and 267 needs to be changed.

diesel_cli/src/migrations/mod.rs
```diff
@@ -273,55 +273,75 @@ fn redo_migrations<Conn, DB>(
...
+        let number = args.get_one::<u64>("REDO_NUMBER").unwrap();
+        let number = std::cmp::min(*number as usize, applied_migrations.len());
+        &applied_migrations[..number]
+    };
+    let should_use_not_use_transaction = versions_to_revert.iter().any(|v| {
+        migrations
+            .get(v)
+            .map(|m| !m.metadata().run_in_transaction())
+            .unwrap_or_default()
+    });
```
hasnainzeenwala, line 294: A single migration having `metadata().run_in_transaction()` as false can cause all the migrations to not be run as a transaction. So if there's a failure in the middle of the redo operation, how are we restoring the DB to a consistent state?

weiznich: We cannot really restore the database into a consistent state in that case. If one of the migrations fails it aborts the process there. It's then up to the user to fix the problems.
Now that's not optimal, but on the other hand we cannot do much more without a single transaction.
Other variants that I considered:
* Requiring a `--allow-without-transaction` flag to allow running these redo operations. I've decided not to do that, as some backends (mysql) never use transactions for migrations (because you cannot run data definition commands in transactions there). Having such a flag would be kind of surprising then, as it would imply something different.
* Just run those migrations outside of a transaction that do require to be executed outside of a transaction. That has the problem that you need to split down the migration list in a much finer manner + it also might result in a database where only a partial set of migrations was reapplied.

weiznich:
Thanks for the review:

> From what I understand, the idea behind the fix is that you go through all the migrations that need to be redone, take a look at their metadata, and if any one has run_in_transaction() set to true, you don't run the closure you created as a transaction.

That's correct :+1:

> I wanted to understand how is the metadata field for a migration is created. I followed how Metadata argument is set for SqlFileMigration and it seems to be done by reading some toml file. Can someone explain to me how and when this file is created? And how does it know if the SQL inside contains a transaction?

It's an explicit user decision to create that file: https://docs.diesel.rs/2.0.x/diesel_migrations/struct.FileBasedMigrations.html#

diesel_cli/src/migrations/mod.rs
weiznich: Will correct that

diesel_cli/src/migrations/mod.rs
hasnainzeenwala: right that makes sense.

diesel_cli/src/migrations/mod.rs
```diff
@@ -273,55 +273,75 @@ fn redo_migrations<Conn, DB>(
...
-        // Sort the migrations by version to apply them in order.
-        migrations.sort_by_key(|m| m.name().version().as_owned());
+            let mut migrations = reverted_versions
+                .into_iter()
+                .map(|v| {
+                    migrations
+                        .remove(&v)
+                        .ok_or_else(|| MigrationError::UnknownMigrationVersion(v.as_owned()))
+                })
+                .collect::<Result<Vec<_>, _>>()?;
```
hasnainzeenwala, line 335: can we create a new function for this. something like, `remove_reverted_migrations(migrations)`

diesel_cli/src/migrations/mod.rs
```diff
@@ -273,55 +273,75 @@ fn redo_migrations<Conn, DB>(
...
                         }
-                        Err(e) => {
-                            Some(Err(e))
-                        }
-                    }
-                })
-                .collect::<Result<Vec<_>, _>>()?
-        };
+                    })
+                    .collect::<Result<Vec<_>, _>>()?
```
hasnainzeenwala, line 319: This part is quite difficult to read. Perhaps it could be pulled into a separate function, `revert_last_n_migrations()`

hasnainzeenwala (changes requested):

diesel_cli/src/migrations/mod.rs
Ten0: As far as I'm concerned I typically don't like functions that are only called once, as you can achieve the same clarity by putting a comment just above the section, whereas having an extra function:
- complexifies understanding precisely what it does and making sure everything is correct (review) because you have to always keep track mentally of where you are in the call tree as you go to understand in practice exactly what will be linearly executed, instead of just scrolling up and down
- complexifies modifying the code because whenever you consider updating a node (function) you have to check for each of the calling nodes up to your entry point if they aren't called by anyone else to be sure of what else you may be affecting. When there's a single function, it's trivial to identify that that is not the case, and that makes edits easier

hasnainzeenwala: The comments are helpful, thanks.

hasnainzeenwala (approved):

hasnainzeenwala:
changes LGTM @weiznich

PR #3627, opened 2023-05-14
Add ability to specify `ssl_cert` and `ssl_key` for MySql

GorgWorgington:
Also updated docs for `MysqlConnection` per https://github.com/diesel-rs/diesel/discussions/3624#discussioncomment-5869196


weiznich (changes requested):
Thanks for opening this PR. The changes itself look good :+1:

Can you add two tests for the newly added url params? Something like the `ssl_ca_test` in `src/mysql/connection/url.rs`

PR #3628, opened 2023-05-15
Multi column `distinct_on` for Postgresql

omid:
- Add ability to define multiple columns in a single `distinct_on` for PostgreSQL, like: `.distinct_on((column_a, column_b))`. Main files related to this:
  - https://github.com/diesel-rs/diesel/pull/3628/files#diff-556fca1d890492102d32274fc8da5ebdf8759195ac66e9df5aec8ae849ea2c54
  - https://github.com/diesel-rs/diesel/pull/3628/files#diff-0e5dba8a60dbbbead443c0ba8d1120357269b3f38ff74225b4a9d4a856145800
  - https://github.com/diesel-rs/diesel/pull/3628/files#diff-556fca1d890492102d32274fc8da5ebdf8759195ac66e9df5aec8ae849ea2c54
- Fixed some tiny typos and grammar issues


diesel/src/pg/backend.rs
```diff
@@ -138,18 +138,18 @@ impl SqlDialect for Pg {
...
-    type ArrayComparison = PgStyleArrayComparision;
+    type ArrayComparison = PgStyleArrayComparison;
 }
 
 impl DieselReserveSpecialization for Pg {}
 impl TrustedBackend for Pg {}
 
 #[derive(Debug, Copy, Clone)]
-pub struct PgOnConflictClaues;
+pub struct PgOnConflictClause;
```
omid, line 148: I'm not sure if it's a public API change, if so, we need to inform people about it :)

diesel/src/pg/backend.rs
```diff
@@ -138,18 +138,18 @@ impl SqlDialect for Pg {
...
-impl sql_dialect::on_conflict_clause::SupportsOnConflictClause for PgOnConflictClaues {}
-impl sql_dialect::on_conflict_clause::SupportsOnConflictClauseWhere for PgOnConflictClaues {}
-impl sql_dialect::on_conflict_clause::PgLikeOnConflictClause for PgOnConflictClaues {}
+impl sql_dialect::on_conflict_clause::SupportsOnConflictClause for PgOnConflictClause {}
+impl sql_dialect::on_conflict_clause::SupportsOnConflictClauseWhere for PgOnConflictClause {}
+impl sql_dialect::on_conflict_clause::PgLikeOnConflictClause for PgOnConflictClause {}
 
 #[derive(Debug, Copy, Clone)]
-pub struct PgStyleArrayComparision;
+pub struct PgStyleArrayComparison;
```
omid, line 155: I'm not sure if it's a public API change, if so, we need to inform people about it :)

diesel/src/query_dsl/group_by_dsl.rs
```diff
@@ -18,7 +18,7 @@ pub trait GroupByDsl<Expr: Expression> {
     type Output;
 
     /// See the trait documentation.
-    fn group_by(self, expr: Expr) -> dsl::GroupBy<Self, Expr>;
+    fn group_by(self, expr: Expr) -> Self::Output;
```
weiznich: Can we change that back? The main reason for the old version is that [rustdoc renders that as link to the correct type alias](https://docs.diesel.rs/2.0.x/diesel/query_dsl/methods/trait.GroupByDsl.html#tymethod.group_by), which is what any user should use as return type there. `Self::Output` might point to an non-nameable or will cause otherwise confusion.

diesel/src/pg/backend.rs
weiznich: Yes that type is not part of the public API, so it should be fine to change it.

diesel/src/pg/backend.rs
weiznich: That type is private as well.

weiznich (commented):
This looks fine for me (beside the minor thing about `Self::Output` vs the dsl types)
Additionally I would like to see some compile test that verifies that we reject queries with multiple columns in `.distinct_on` calls without matching order clause.

diesel/src/query_dsl/group_by_dsl.rs
omid: Yep, already changed back

omid:
> Additionally I would like to see some compile test that verifies that we reject queries with multiple columns in `.distinct_on` calls without matching order clause.

@weiznich, actually it failed with simple correct orders :(
On my local tests, I didn't have any `order`s and it was fine. But in tests which has orders, it fails.

With simple orders (like `.order_by((id, name))` it's fine, but as soon as we have `.asc()` or `.desc()` it fails.
I'm not sure how I can solve it. I'm trying but couldn't find the proper way yet :( )

weiznich:
I do not have a solution for that yet, so I need to find some time to search for a solution. I cannot promise that this happens this week due to other priorities.

omid:
> I do not have a solution for that yet, so I need to find some time to search for a solution. I cannot promise that this happens this week due to other priorities.

I'm also working on this.

First try, to generate all impl cases failed. Because it needs to generate a lot of code, like this:
```
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Asc<__D1>, __D2)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Desc<__D1>, __D2)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(__D1, crate::helper_types::Asc<__D2>)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(__D1, crate::helper_types::Desc<__D2>)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Asc<__D1>, crate::helper_types::Asc<__D2>)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Asc<__D1>, crate::helper_types::Desc<__D2>)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Desc<__D1>, crate::helper_types::Asc<__D2>)>
{
}
impl<__D1, __D2,> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>>
for OrderClause<(crate::helper_types::Desc<__D1>, crate::helper_types::Desc<__D2>)>
{
}
```

if we have `128-column-tables` feature enabled, just for the 128 column scenario, it needs to generate 1.18e61 cases :))

weiznich:
> I need to learn how you did that, thanks!

There are just some patterns that keep coming up again and again in diesel. That's one of them. I should probably start writing more about those internals some day...

omid:
> There are just some patterns that keep coming up again and again in diesel. That's one of them. I should probably start writing more about those internals some day...

That would be awesome!

But... regarding the implementation, I found a tiny issue in your code. Here is the code for a tuple of two:
```
trait OrderDecorator {
    type Plain;
}

impl<C> OrderDecorator for C where C: Column {
   type Plain = C;
}

impl<C> OrderDecorator for Desc<C>  {
    type Plain = C;
}

impl<C> OrderDecorator for Asc<C>  {
    type Plain = C;
}

impl<__D1,__D2, __O1, __O2> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>> for OrderClause<(__O1, __O2)> 
where __O1: OrderDecorator<Plain = __D1>,
__O2: OrderDecorator<Plain = __D2>,
{}
```

The issue is that it doesn't work for a query like this:
```
let source = users
        .select((name, hair_color))
        .order((id, name.asc(), hair_color.desc()))
        .distinct_on((id, name));
```
because `order` is a tuple of three and `distinct_on` is tuple of two.

One solution could be to have many impls per `DistinctOnClause` for example have this also:
```
impl<__D1,__D2, __O1, __O2,__O3> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>> for OrderClause<(__O1, __O2,__O3)> 
where __O1: OrderDecorator<Plain = __D1>,
__O2: OrderDecorator<Plain = __D2>,
{}

impl<__D1,__D2, __O1, __O2,__O3,__O4> ValidOrderingForDistinct<DistinctOnClause<(__D1, __D2)>> for OrderClause<(__O1, __O2,__O3,__O4)> 
where __O1: OrderDecorator<Plain = __D1>,
__O2: OrderDecorator<Plain = __D2>,
{}

// and so on
```

weiznich:
Yes my implementation would only work with "matching" distinct and order by clauses. 

So the basic problem here is that the following combinations are fine:

* Distinct on clause contains less columns than order by clause and these columns match the begining of the order by clause, so `ValidOrderingForDistinct<DistinctOnClause<(__O1::Plain, __O2::Plain)>> for OrderClause<(__O1, __O2,__O3, __O4, …)>`
* Distinct on clause contains more columns than order by clause  where the order by clause "matches" the first columns of the distinct clause: `ValidOrderingForDistinct<DistinctOnClause<(__O1::Plain, __O2::Plain, D1, D2, …)>> for OrderClause<(__O1, __O2)>`
* The column count matches -> The case in my example above.

So we need way more impls to describe that. For each tuple size n we need m impls where m is the maximal supported tuple size. (That ignores one or two additional impls to allow single column cases without tuples) That means it's just m*m impls. So for a max tuple size of 32 we have 1024 impl, for a max size of 128 we have 16k impls. That does not seem to scale well.

I see the following options here:

1. Just allow exact matches
2. Maybe do not use the `max_table_columns` count to determine the upper limit here, but use something smaller. We could say that 5 (or any other reasonable small number) columns in order by/distinct on clauses should be enough for anyone.

I think variant 1 would make it hard to add support for the unsupported cases later on, because it might require removing these existing impls (which is a breaking change). Variant 2 would allow to add support for the unsupported cases, by just bumping the number. I personally would likely go with variant 2, although that will require some modifications to the `diesel_for_each_tuple!` macro. I think the easiest route there is to just allow an additional optional argument that specifies the tuple size. If that's not set, just fall back to the "default" size.

omid:
@weiznich I would also do #2.

But I'm not sure what exactly you suggested.

1. Do you mean, for example, for a _distinct on_ tuple of N members, should we implement _order by_ tuples of 1 to N+5? (actually, min(N+5, MAX_TUPLE_SIZE))?
2. Or for any _distinct on_ tuple, just _order by_ tuples of 1 to 5 only?
3. Or just impl to max 5 members, either _distinct on_ tuples and _order by_ tuples? (I think you meant this one, then see other items as new suggestions :D)

impl of the first one is harder :)

weiznich:
I've meant to suggest this variant: 

> Or just impl to max 5 members, either distinct on tuples and order by tuples? (I think you meant this one, then see other items as new suggestions :D)

With the restriction that the max size of the distinct on and order by tuples is 5 (or whatever other limit). So really just a "few" (25) impls for all the cases, but for a low column count.

I think that's the solution that should cover most of the use-cases and that's easy to communicate to the users.

omid:
@weiznich I need your help. I pushed last changes.

1. I couldn't make the macro smaller, it's a bit tricky! I need to find the minimum of two lists and define bounds based on that, which makes the recursion too complex, based on my knowledge.
2. Now there is a conflict between `impl<T> ValidOrderingForDistinct<DistinctOnClause<T>> for OrderClause<T> where T: Expression {}` and 
```
 #[allow(unused_parens)]
#[cfg(feature = "postgres")]
impl<_OT1, T1,> crate::query_dsl::order_dsl::ValidOrderingForDistinct<crate::pg::DistinctOnClause<(T1,)>,> for crate::query_builder::order_clause::OrderClause<(_OT1,)>
    where
        _OT1: crate::pg::OrderDecorator<Column = T1>,
        T1: Column,
{}
 ```
To solve it, I tried to put `T1, Column` here, but still has the conflict (But I think it makes sense to have `Column` there, no?). It seems like `Expression` is really general, I don't know what should I use here to limit the scope?
And if I remove that general impl, I'll get this error:
```
error[E0277]: the trait bound `SelectStatement<FromClause<schema::users::table>,
query_builder::select_clause::SelectClause<(schema::users::columns::name, schema::users::columns::hair_color)>,
query_builder::distinct_clause::NoDistinctClause, query_builder::where_clause::NoWhereClause,
query_builder::order_clause::OrderClause<schema::users::columns::name>>: DistinctOnDsl<_>` is not satisfied
```

I'll try to find it, but I think you can lead me and review my changes in the meanwhile 🙏🏼

**Comment by weiznich** [comment]
I will have a look at the macro later this week. As for the conflicting impls: You should be able to solve that with the `SingleValue` bound that was there before.

**Comment by omid** [comment]
There are still conflicts, after I put:
```
impl<T> ValidOrderingForDistinct<DistinctOnClause<T>> for OrderClause<T>
    where
        T: Expression,
        T::SqlType: SingleValue,
{
}
```
the conflicts are like before:
```
`conflicting implementations of trait `ValidOrderingForDistinct<DistinctOnClause<(_,)>>`
for type `order_clause::OrderClause<(_,)>`
```

**Comment by weiznich** [comment]
I've pushed a commit that hopefully should workaround these conflicting impls. It also restructures how these impls are generated so that the additional `paste` dependency is not required anymore. Can you verify that it actually works now?

**Comment by omid** [comment]
@weiznich 
```
impl<T, O> ValidOrderingForDistinct<DistinctOnClause<T>> for OrderClause<O> where O: OrderDecorator<Column = T> {}
```
solves the not implemented case.
I'll continue writing more tests.

**Comment by omid** [comment]
@weiznich I think there are no tests to check compile time errors, am I right?
And also, I think we cannot easily show a compile time warning about the limitation of having 5 cases only.

**Comment by weiznich** [comment]
Tests to check for compile time error are usually placed here: https://github.com/diesel-rs/diesel/tree/master/diesel_compile_tests%2Ftests
You basically add a file with code that is supposed to cause an compiler error to the fail folder. You execute these tests by running `cargo test` in the compile_test directory. You need to accept the output after running the new test for the first time (instructions are in the test failure)

**Review by weiznich — APPROVED** [review]

**Comment by weiznich** [comment]
Thanks for finishing this up. I think it's not ready for being merged :tada:

**Comment by omid** [comment]
> I think it's not ready for being merged

It's NOT or it's NOW ready? :D 
I'm confused between NOT and 🎉  :))

**Comment by weiznich** [comment]
It is supposed to be a "now" (I edit the post above as well)

**Comment by omid** [comment]
And... there is a problem. The old syntax doesn't work anymore.

`.order_by((dsl::sql::<Bool>("DESC col2")))`

with the following error:
```
expected `(SqlLiteral<Bool>,)`, found `SqlLiteral<Bool>`
```

Yet I'm not sure if it's related to this PR or not

weiznich:
@omid I think what I've written above is the actual issue. I've already tried to work on a fix and got something at works at least partially. See https://github.com/diesel-rs/diesel/pull/3643 for that. It would be helpful if you can contribute test case that worked with diesel 2.0 and broke with 2.1 there.
Also ideas how to solve the last missing part are also helpful.

PR #3629, opened 2023-05-15
Update advanced blog example

omid:
`structopt` is kinda deprecated

### DIFF
--- a/examples/postgres/advanced-blog-cli/Cargo.toml
+++ b/examples/postgres/advanced-blog-cli/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2018"
 publish = false
 
 [dependencies]
-bcrypt = "0.14.0"
+argon2 = "0.5.0"
 chrono = { version = "0.4.20", default-features = false, features = ["clock", "std"] }
 diesel = { version = "2.0.0", path = "../../../diesel", features = ["postgres", "chrono"] }
 dotenvy = "0.15"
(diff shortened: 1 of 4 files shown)


examples/postgres/advanced-blog-cli/Cargo.toml
```diff
@@ -5,11 +5,11 @@ edition = "2018"
 publish = false
 
 [dependencies]
-bcrypt = "0.10.1"
+bcrypt = "0.14.0"
```
weiznich, line 8: Could we switch that to [argon2](https://lib.rs/crates/argon2) as that seems to be the state of art password hashing function?

weiznich (commented):
Thanks for the PR :+1:
I was aware of that, I just had no time to update that. Would it be possible to switch the password hashing as well?

examples/postgres/advanced-blog-cli/Cargo.toml
omid: sure

examples/postgres/advanced-blog-cli/src/auth.rs
```diff
@@ -74,7 +82,11 @@ fn register_user(
     username: &str,
     password: &str,
 ) -> Result<User, AuthenticationError> {
-    let hashed_password = hash(password, DEFAULT_COST)?;
+    let salt = SaltString::from_b64(SALT_STRING)?;
```
weiznich, line 85:
```suggestion
    // In real applications you should never use a constant salt!!
    // Checkout https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#salting for details
    let salt = SaltString::from_b64(SALT_STRING)?;
```

weiznich (approved):

PR #3239, opened 2022-07-20
Allow for "bool" type from django

ianeli1:
Hey all.

Encountered this issue while attempting to modernize a Django app by running `diesel print-schema` on an already created SQLite3 db, which returned `Unsupported type: bool`.

Django appears to have [used `"bool"` for their types](https://github.com/django/django/blob/24effbceb871e71d3bc320b91252f743714722df/django/db/backends/sqlite3/base.py#L32) instead of `"boolean"`.

PR #3632, opened 2023-05-17
Allow for "bool" type from django with tests

social-anthrax:
Taken from PR #3239:
> Hey all.
> 
> Encountered this issue while attempting to modernize a Django app by running `diesel print-schema` on an already created SQLite3 db, which returned `Unsupported type: bool`.
> 
> Django appears to have [used `"bool"` for their types](https://github.com/django/django/blob/24effbceb871e71d3bc320b91252f743714722df/django/db/backends/sqlite3/base.py#L32) instead of `"boolean"`.

This PR is continued from #3239, with two additions

- `type_name.contains("bool")` has been changed to `type_name.starts_with("bool")` to match fewer cases. 
- test has been implemented in diesel_cli/tests/print_schema.rs


czotomo:
I see one problem, and that problem is sqlite's type system. This is perfectly legal:
`sqlite> create table foo (x booldozer, y boolbool);
sqlite> insert into foo values (1, 2);
sqlite> select * from foo;
1|2`

social-anthrax:
I see your point. Would it be better to strictly match bool and boolean?

czotomo:
> I see your point. Would it be better to strictly match bool and boolean?

I am not sure. Strict matching is safer, but it's also ... strict. 
Either do that, or maybe log a "you have such and such type and it's going to be treated as boolean" message?

social-anthrax:
I feel that at the moment only bool and boolean have been encountered, so it's easy enough to do a strict match. Maybe in the future if more issues like this arise it will be worth matching anything starting with bool and throwing a warning

weiznich (approved):
Thanks for including the tests :+1:

weiznich:
I believe it's fine to explicitly match these two cases for now. In the medium term I would suggest that we allow to configure that mapping via `diesel.toml`. That means that we only support the basic "real" sqlite types, and everything else is configured there via a list of regular expressions and replacements or something like that. That would allow to customize these mappings much easier based on specific usecases. I would be happy to mentor work in that direction.

PR #3635, opened 2023-05-19
Fix compile fails on target `aarch64-unknown-linux-gnu` with feature mysql

gwy15:
Current diesel has this conversion
```rust
    fn for_input((tpe, data): (MysqlType, Option<Vec<u8>>)) -> Self {
        let (tpe, flags) = tpe.into();
        let is_null = i8::from(data.is_none());
        let mut bytes = data.unwrap_or_default();
        let ptr = NonNull::new(bytes.as_mut_ptr());
        let len = bytes.len() as libc::c_ulong;
        let capacity = bytes.capacity();
        mem::forget(bytes);
        Self {
            tpe,
            bytes: ptr,
            length: len,
            capacity,
            flags,
            is_null,
            is_truncated: None,
        }
    }
```
Notice that `is_null` has type `mysqlclient_sys::my_bool = std::os::raw::c_char` but it receives an `i8`. On certain platforms, `aarch64-unknown-linux-gnu` in this case, `c_char` is `u8` thus compile failes.

This PR fix this.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -222,7 +222,7 @@ impl Drop for BindData {
 impl BindData {
     fn for_input((tpe, data): (MysqlType, Option<Vec<u8>>)) -> Self {
         let (tpe, flags) = tpe.into();
-        let is_null = i8::from(data.is_none());
+        let is_null = ffi::my_bool::from(data.is_none());
         let mut bytes = data.unwrap_or_default();
         let ptr = NonNull::new(bytes.as_mut_ptr());
         let len = bytes.len() as libc::c_ulong;


weiznich (approved):

I should maybe add some general words on the support story of diesel for architectures other than x86_64 here. Basically we can only ensure that diesel works on system that we do have access to. In my case thats only `x86_64` because that's the only hardware I own and that's the only hardware offered by github actions. As consequence that means we cannot give any guarantees for other architectures. It does not mean that we won't accept patches there (as long as they don't break those architectures that we can test), but it means that it might just be broken for whatever reason and we can not tell you about that. If you are interested in changing that we need to talk about ways to get the maintainers easy access to systems with those architectures you are interested in being supported. Ideally in such a way that we can run our CI tests there as well.

Issue #3646, opened 2023-06-01
`joinable!` not generated when FKs in different tables have the same name


tp-woven:
### Versions

- **Rust:** 1.69, 2021 edition
- **Diesel:** 2.1.0
- **Database:** Postgres (`postgres:latest` Docker image)
- **Operating System:** MacOS Ventura 13.4

### Feature Flags

- **diesel:** postgres

## Problem Description

Having two tables with a foreign key that has the same name causes `joinable` to not be generated.

### What are you trying to accomplish?

This is the "first" migration in a project, and I have the following table definitions:

```sql
CREATE TABLE "kinds"
(
    "kind_id"   SERIAL  NOT NULL,
    "kind_name" VARCHAR NOT NULL UNIQUE,
    PRIMARY KEY ("kind_id")
);

CREATE TABLE "data_mapping"
(
    "data_kind_id__kind_id" INT     NOT NULL,
    "data_attribute"        VARCHAR NOT NULL,
    PRIMARY KEY ("data_attribute"),
    CONSTRAINT fk_data_kind FOREIGN KEY (data_kind_id__kind_id) REFERENCES kinds (kind_id)
);

CREATE TABLE "other_table"
(
    "data_kind_id__kind_id" INT     NOT NULL,
    "data_attribute"        VARCHAR NOT NULL,
    PRIMARY KEY ("data_attribute"),
    CONSTRAINT fk_data_kind FOREIGN KEY (data_kind_id__kind_id) REFERENCES kinds (kind_id)
);
```

When I run the migration, the generated `schema.rs` contains the `table!` definitions but not the `joinable!`.

### What is the expected output?

```
diesel::joinable!(data_mapping -> kinds (data_kind_id__kind_id));
diesel::joinable!(other_table -> kinds (data_kind_id__kind_id));
```

### What is the actual output?

No `joinable!`.

### Are you seeing any additional errors?

Not an additional error, but some more information...

Changing the constraint name so that it doesn't match between the two tables _does_ generate the `joinable!`. Similarly, when switching to "inline" FKs on the column definitions (`"data_kind_id__kind_id" INT NOT NULL REFERENCES kinds (kind_id)`) works as well, because the generated constraint names do not match.

### Steps to reproduce

Fresh database, create a migration with the tables above, and apply the migration.

## Checklist

[...]

PR #3650, opened 2023-06-06
add test reproducing #3646

tp-woven:
This test reproduces #3646: Adding two foreign keys referencing the same table that have the same name "confuses" `load_foreign_key_constraints`.


weiznich:
Thanks for providing this test. That was really helpful. I've pushed https://github.com/diesel-rs/diesel/pull/3650/commits/0a252c36e9afc223dbe86dcd181e674b3a0119db to fix that issue. Can you confirm that it also fixes your usecase?

tp-woven:
Thanks for the quick fix! 🙏 
Confirmed that after checking-out this branch and installing diesel_cli from it, the issue is solved.

PR #3655, opened 2023-06-09
Added "All About Updates" sqlite examples

gernotpokorny:
I copy & pasted the Postgres examples and adapted them to sqlite. Note: The tests do not look ideal to me


kw217:
@gernotpokorny, can you rebase this to current? Or @weiznich is there a way you can do that from your side?

Issue #1412, opened 2017-12-19
Add support for array operations in the dsl.


Aaronepower:
It'd be nice to if the dsl supported array operations such as `append` or `remove`.

## Rust
```rust
struct Post {
     id: i32,
     authors: Vec<String>,
}

fn main() {
    use schema::posts::dsl::*;
    diesel::update(posts.find(&1))
      .array_append(authors, String::from("Jane Doe"));
}
```

## SQL
```sql
UPDATE posts SET authors = authors || '{Jane Doe}' WHERE id = 1;
```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.

sgrif:
Note that your SQL doesn't map to the operator or types you've specified. `||` is the "concatenate" operator, and the equivalent Rust would likely be `authors.eq(authors.concat(vec!["Jane Doe"]))`

dessalines:
Question, how do I concat the opposite way? IE `'some_text' || my_column` , or `"some_text".concat(my_column)` ?

weiznich:
@dessalines That's not the place to ask such questions. This is a feature request about extending the support of postgres array operations, not about string concats. Please use our support channels (gitter, forum) for asking question and do not add them as comments to unrelated issues.

PR #3660, opened 2023-06-14
Add support for the array concat operation in the dsl

sintemal:
This adds support for the ` || ` operator on arrays. This feature was already requested in #1412

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -370,6 +370,50 @@ pub trait PgArrayExpressionMethods: Expression + Sized {
     {
         ArrayIndex::new(self, other.as_expression())
     }
+
+    /// Concatenates two PostgreSQL Arrays using the `||` operator.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     posts {
+    /// #         id -> Integer,
+    /// #         tags -> Array<VarChar>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::posts::dsl::*;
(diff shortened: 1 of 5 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -370,6 +370,50 @@ pub trait PgArrayExpressionMethods: Expression + Sized {
     {
         ArrayIndex::new(self, other.as_expression())
     }
+
+    /// Concatenates two PostgreSQL Arrays using the `||` operator.
```
pksunkara, line 374: Please follow the same pattern of documentation from other methods. `contains` is a good example.

diesel/src/pg/expression/helper_types.rs
```diff
@@ -184,3 +184,7 @@ pub type LikeBinary<Lhs, Rhs> = Grouped<super::operators::LikeBinary<Lhs, AsExpr
 #[cfg(feature = "postgres_backend")]
 pub type NotLikeBinary<Lhs, Rhs> =
     Grouped<super::operators::NotLikeBinary<Lhs, AsExprOf<Rhs, Binary>>>;
+
+/// The return type of [`lhs.remove_by_path(rhs)`](super::expression_methods::PgArrayExpressionMethods::concat)
```
Ten0, line 188: documentation is wrong here

Ten0 (approved):
implementation-wise this looks correct

I'm confident for merge if this is tested (either manually or by adding an auto-test) (once all the comments are solved of course)

sintemal:
I think this should fix the documentation

pksunkara (approved):

pksunkara:
Can you please repush to trigger the CI?

sintemal:
One compiletest somehow fails, but i am not familiar enough with detailed typesystem to figure out why it does. Maybe you guys could take a look, if you can spot the bug. I could imagine that it has something to do with the nested type of `Array<ST>` and that `__diesel_internal_SameResultAsInput`in `operators.rs` doesn't support this...

sintemal:
This actually seemed to be the problem. Thank you very much!

PR #3662, opened 2023-06-15
Allow ~type inference on `And` and `Or` helper types

Ten0:
This is [particularly useful for query fragment functions type inference](https://github.com/diesel-rs/diesel/discussions/3661#discussioncomment-6180421) (you probably want to read that discussion first).

It works with types that are `Expression`s and have a `SqlType` that is either `Bool` or `Nullable<Bool>`, with `bool` (and `&bool` and `&&bool`), and with custom user types provided they implement this trait (otherwise they would just have to specify manually).

Latest design discussions on why we currently have this constraint: https://github.com/diesel-rs/diesel/issues/2589#issuecomment-748238464

Cases where an additional type parameter would still have to be specified in the helper type generic parameters are:
- If this trait isn't implemented for the `other` parameter of the expression - this is technically a breaking change, but a user would have to be using a custom type that has `AsExpression<Bool>` but **not** `Expression<SqlType=Bool>` as `other` **and** have written `dsl::And/Or<x, ThatCustomType>` without the explicit 3rd parameter somewhere for it to actually break. I would estimate that to be unlikely enough that we can fake the non-breaking change, especially in the light of its use for `diesel_auto_type`/`Selectable`.
- If the user actually was using the not-preferred implementation of `AsExpression` (e.g. towards `Nullable<Bool>` instead of `Bool`) - this is not a breaking change, because in that case the user is already overriding the 3rd generic parameter.

### DIFF
--- a/diesel/src/expression_methods/bool_expression_methods.rs
+++ b/diesel/src/expression_methods/bool_expression_methods.rs
@@ -126,29 +126,36 @@ pub trait PreferredBoolSqlType {
     /// That should be either `Bool` or `Nullable<Bool>`.
     type PreferredSqlType;
 }
+
 impl<E: Expression> PreferredBoolSqlType for E
 where
     E::SqlType: BoolOrNullableBool,
 {
     type PreferredSqlType = <E as Expression>::SqlType;
 }
+
 /// This impl has to live in Diesel because otherwise it would conflict with the blanket impl above
 /// because "diesel might add an implementation of Expression for bool"
 impl PreferredBoolSqlType for bool {
     type PreferredSqlType = sql_types::Bool;
 }
+
 impl PreferredBoolSqlType for &bool {
     type PreferredSqlType = sql_types::Bool;
 }
+
 impl PreferredBoolSqlType for &&bool {
     type PreferredSqlType = sql_types::Bool;
 }
+
 impl PreferredBoolSqlType for Option<bool> {
     type PreferredSqlType = sql_types::Nullable<sql_types::Bool>;
 }
+
 impl PreferredBoolSqlType for &Option<bool> {
     type PreferredSqlType = sql_types::Nullable<sql_types::Bool>;
 }
+
 impl PreferredBoolSqlType for &&Option<bool> {
     type PreferredSqlType = sql_types::Nullable<sql_types::Bool>;
 }
--- a/diesel/src/expression_methods/bool_expression_methods.rs
+++ b/diesel/src/expression_methods/bool_expression_methods.rs
@@ -112,7 +112,7 @@ where
 ///
 /// It works with types that are [Expression]s and have a [`SqlType`](Expression::SqlType) that is
 /// either [`Bool`](sql_types::Bool) or [`Nullable<Bool>`](sql_types::Nullable), and with [`bool`]
-/// (and `&bool` and `&&bool`).
+/// (and `Option<bool>` and references to those).
 ///
 /// Cases where an additional type parameter would still have to be specified in the helper type
 /// generic parameters are:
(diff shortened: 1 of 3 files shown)


diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+///   (in that case the user (you?) probably wants to implement it)
+/// - If the user actually was using the not-preferred implementation of `AsExpression`
+///   (e.g. towards `Nullable<Bool>` instead of `Bool`)
+pub trait PreferredBoolSqlType {
+    /// The preferred `Bool` SQL type for this AsExpression implementation.
+    ///
+    /// That should be either `Bool` or `Nullable<Bool>`.
+    type PreferredSqlType;
+}
+impl<E: Expression> PreferredBoolSqlType for E
```
weiznich, line 129:
```suggestion

impl<E: Expression> PreferredBoolSqlType for E
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+where
+    E::SqlType: BoolOrNullableBool,
+{
+    type PreferredSqlType = <E as Expression>::SqlType;
+}
+/// This impl has to live in Diesel because otherwise it would conflict with the blanket impl above
```
weiznich, line 135:
```suggestion

/// This impl has to live in Diesel because otherwise it would conflict with the blanket impl above
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+/// because "diesel might add an implementation of Expression for bool"
+impl PreferredBoolSqlType for bool {
+    type PreferredSqlType = sql_types::Bool;
+}
+impl PreferredBoolSqlType for &bool {
```
weiznich, line 140:
```suggestion

impl PreferredBoolSqlType for &bool {
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+    type PreferredSqlType = sql_types::Bool;
+}
+impl PreferredBoolSqlType for &&bool {
```
weiznich, line 143:
```suggestion

impl PreferredBoolSqlType for &&bool {
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+    type PreferredSqlType = sql_types::Bool;
+}
+impl PreferredBoolSqlType for Option<bool> {
```
weiznich, line 146:
```suggestion

impl PreferredBoolSqlType for Option<bool> {
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+    type PreferredSqlType = sql_types::Nullable<sql_types::Bool>;
+}
+impl PreferredBoolSqlType for &Option<bool> {
```
weiznich, line 149:
```suggestion

impl PreferredBoolSqlType for &Option<bool> {
```

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -103,3 +103,52 @@ where
...
+    type PreferredSqlType = sql_types::Nullable<sql_types::Bool>;
+}
+impl PreferredBoolSqlType for &&Option<bool> {
```
weiznich, line 152:
```suggestion

impl PreferredBoolSqlType for &&Option<bool> {
```

weiznich (approved):
This looks great :+1:

PR #3678, opened 2023-07-05
Icon for the generated doc

daladim:
This adds the Diesel icon on the pages generated by `cargo doc`.

That's a draft however, because:
* We may prefer an image that does not include the white "Diesel" text (especially to play nice with the light theme). However, I'm not sure such an image already exists.
* I'm not sure the URL I'm pointing to is guaranteed to be always valid in the future. Is it? Should we include e.g. a version number in its URL?


daladim:
Here is what the rendered doc now would look like:
image

weiznich:
I think it's fine to assume that the logo stays on the web page for some while. If someone submitts a PR there, we can also host only the logo without the text.

PR #3681, opened 2023-07-06
fix(diesel_cli): use separate lines for multiple annotations

DannyGoldberg:
We were running into a compilation error from diesel-derives where we had both a `sql_name` and `max_length` annotation for the same property, which would generate something like this:

```rust
diesel::table! {
    example (id) {
        id -> Uuid,
        #[sql_name = "type", max_length = 200]
        type_ -> Nullable<Varchar>,
    }
}
```

and would throw a compilation error like this:

```
error: Invalid `table!` syntax. Please see the `table!` macro docs for more info.
       Docs available at: `https://docs.diesel.rs/master/diesel/macro.table.html`
```

_note: this isn't yet tested locally, as I've had trouble getting this repo running locally_

This should result in something like the following, which wouldn't fail to compile:

```rust
diesel::table! {
    example (id) {
        id -> Uuid,
        #[sql_name = "type"]
        #[max_length = 200]
        type_ -> Nullable<Varchar>,
    }
}
```

### DIFF
--- a/diesel_cli/src/print_schema.rs
+++ b/diesel_cli/src/print_schema.rs
@@ -697,20 +697,11 @@ impl<'a> Display for ColumnDefinitions<'a> {
                 }
 
                 // Write out attributes
-                if column.rust_name != column.sql_name || column.ty.max_length.is_some() {
-                    let mut is_first = true;
-                    write!(out, r#"#["#)?;
-                    if column.rust_name != column.sql_name {
-                        write!(out, r#"sql_name = {:?}"#, column.sql_name)?;
-                        is_first = false;
-                    }
-                    if let Some(max_length) = column.ty.max_length {
-                        if !is_first {
-                            write!(out, ", ")?;
-                        }
-                        write!(out, "max_length = {}", max_length)?;
-                    }
-                    writeln!(out, r#"]"#)?;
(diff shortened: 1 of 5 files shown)


diesel_cli/src/print_schema.rs
```diff
@@ -697,20 +697,11 @@ impl<'a> Display for ColumnDefinitions<'a> {
...
-                        if !is_first {
-                            write!(out, ", ")?;
-                        }
-                        write!(out, "max_length = {}", max_length)?;
-                    }
-                    writeln!(out, r#"]"#)?;
+                if column.rust_name != column.sql_name {
+                    writeln!(out, r#"#[sql_name = "{}"]"#, column.sql_name)?;
+                }
+                if let Some(max_length) = column.ty.max_length {
```
DannyGoldberg, line 703: @Ten0 this intends to fix an issue I ran into from #3552, where multiple comma-separated annotations would cause compilation errors -- github actions tests pass, although I am having some trouble testing it locally

weiznich:
Thanks for opening this PR. Can you add a test for that in `diesel_cli/tests/print_schema.rs` (see the `print_schema` dir there for examples of existing test cases.)

Ten0 (approved):

PR #3683, opened 2023-07-07
Diesel query fragment type inference

Ten0:
As discussed in https://github.com/diesel-rs/diesel/discussions/3661

This introduces:
- The `#[dsl::auto_type]` attribute macro, allowing to infer type of query fragment functions
- Using the same type inference on `Selectable` derives, which allows skipping specifying `select_expression_type` most of the time, in turn enabling most queries to be written using just a `Selectable` derive.

### DIFF
--- a/diesel/src/query_source/aliasing/macros.rs
+++ b/diesel/src/query_source/aliasing/macros.rs
@@ -84,6 +84,10 @@ macro_rules! alias {
                 pub const $alias_name: Alias<$alias_ty> = $($table)::+ as $alias_name;
             )*
         }
+        $(
+            #[allow(non_camel_case_types)]
+            type $alias_name = $crate::query_source::Alias::<$alias_ty>;
+        )*
     };
     ($($vis: vis const $const_name: ident: Alias<$alias_ty: ident> = $($table: ident)::+ as $alias_sql_name: ident);* $(;)?) => {
         $crate::alias!(NoConst $($($table)::+ as $alias_sql_name: $vis $alias_ty,)*);
--- a/diesel/src/query_source/aliasing/macros.rs
+++ b/diesel/src/query_source/aliasing/macros.rs
@@ -84,10 +84,6 @@ macro_rules! alias {
                 pub const $alias_name: Alias<$alias_ty> = $($table)::+ as $alias_name;
             )*
         }
-        $(
-            #[allow(non_camel_case_types)]
-            type $alias_name = $crate::query_source::Alias::<$alias_ty>;
-        )*
     };
     ($($vis: vis const $const_name: ident: Alias<$alias_ty: ident> = $($table: ident)::+ as $alias_sql_name: ident);* $(;)?) => {
         $crate::alias!(NoConst $($($table)::+ as $alias_sql_name: $vis $alias_ty,)*);
@@ -95,6 +91,9 @@ macro_rules! alias {
             #[allow(non_upper_case_globals)]
             $vis const $const_name: $crate::query_source::Alias::<$alias_ty> =
                 $crate::query_source::Alias::new($alias_ty { table: $($table)::+::table });
+
+            #[allow(non_camel_case_types)]
+            $vis type $const_name = $crate::query_source::Alias::<$alias_ty>;
         )*
     };
     (NoConst $($($table: ident)::+ as $alias_sql_name: ident: $vis: vis $alias_ty: ident),* $(,)?) => {
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -22,8 +22,10 @@ extern crate proc_macro2;
 extern crate quote;
 extern crate syn;
 
-use proc_macro::TokenStream;
-use syn::parse_macro_input;
+use {
+    proc_macro::TokenStream,
+    syn::{parse_macro_input, parse_quote},
+};
 
 mod attrs;
 mod deprecated;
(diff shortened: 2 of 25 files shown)


Ten0:
So I've tested this on our codebase and it feels amazing :)
That makes it really easy to write most queries using `Selectable` derives.

diesel_derives/src/lib.rs
```diff
@@ -22,8 +22,10 @@ extern crate proc_macro2;
 extern crate quote;
 extern crate syn;
 
-use proc_macro::TokenStream;
-use syn::parse_macro_input;
+use {
+    proc_macro::TokenStream,
+    syn::{parse_macro_input, parse_quote},
+};
```
weiznich, line 28: I would prefer to stay with non-nested imports here, just to be consistent with the rest of the code base.

diesel_derives/src/lib.rs
```diff
@@ -880,10 +882,12 @@ pub fn derive_queryable_by_name(input: TokenStream) -> TokenStream {
 ///   the entire select expression for the given field. It may be used to select with
 ///   custom tuples, or specify `select_expression = my_table::some_field.is_not_null()`,
 ///   or separate tables...
-///   It should be used in conjunction with `select_expression_type` (described below)
-/// * `#[diesel(select_expression_type = the_custom_select_expression_type]`, to be used
-///   in conjunction with `select_expression` (described above).
-///   For example: `#[diesel(select_expression_type = dsl::IsNotNull<my_table::some_field>)]`
+///   It may be used in conjunction with `select_expression_type` (described below)
+/// * `#[diesel(select_expression_type = the_custom_select_expression_type]`, should be used
```
weiznich, line 886: Maybe list cases that continue to require this? (I would assume that if `select_expression` is a custom function call it might be required?)

diesel_derives/src/lib.rs
```diff
@@ -1627,3 +1631,151 @@ pub fn table_proc(input: TokenStream) -> TokenStream {
...
+///  - Assume that methods, unless otherwise annotated, have a type alias
+///    available as `diesel::dsl::PascalCaseOfMethodName` (e.g. for the
+///    `x.and(y)` call, it expects that there is a `diesel::dsl::And<X, Y>` type
+///    alias available)
+///
+/// The defaults can be changed by passing the following attributes to the
+/// macro:
+/// - `#[auto_type(no_type_alias)]` to disable the generation of the type alias.
+/// - `#[auto_type(dsl_path = "path::to::dsl")]` to change the path where the
+///   macro will look for type aliases for methods.
```
weiznich, line 1696: We might want to be explicit here that this is "required" if you mix your own custom query dsl extensions with diesel types. Maybe we should just document a pattern for that? Something like create your own `dsl` module somewhere as follows:

```rust
mod dsl {
    /// export all of diesel dsl
    pub use diesel::dsl::*;

    /// Export your extension types here
    pub use crate::your_extension::dsl::YourType;
}

```

diesel_derives/src/lib.rs
```diff
@@ -1627,3 +1631,151 @@ pub fn table_proc(input: TokenStream) -> TokenStream {
...
+///     )
+/// }
+///
+/// #[dsl::auto_type]
+/// fn users_with_posts_with_id_greater_than(id_greater_than: i32) -> _ {
+///     // If we didn't specify the type for this query fragment, the macro would infer it as
+///     // `user_has_post_with_id_greater_than<i32>`, which would be incorrect because there is
+///     // no generic parameter.
+///     let filter: user_has_post_with_id_greater_than =
+///         user_has_post_with_id_greater_than(id_greater_than);
```
weiznich, line 1741: I really like that pattern. That's really good work :heart:

diesel_derives/src/selectable.rs
```diff
@@ -136,9 +136,25 @@ fn to_field_ty_bound(field_ty: &syn::Type) -> Option<TokenStream> {
...
+        let (inferred_type, errors) = type_inference::infer_expression_type(
+            expr,
+            field.select_expression_type.as_ref().map(|t| &t.item),
+            &type_inference::InferrerSettings::builder()
+                .dsl_path(parse_quote!(diesel::dsl))
+                .function_types_case(crate::AUTO_TYPE_DEFAULT_FUNCTION_TYPE_CASE)
+                .method_types_case(crate::AUTO_TYPE_DEFAULT_METHOD_TYPE_CASE)
+                .build(),
+        );
+        let mut type_token_stream = quote!(#inferred_type);
```
weiznich, line 151: Can we use `quote_spanned!` here with a span of the original expression? That should result in the compiler pointing at the `#[diesel(select_expression = …)]` type in the relevant error message. (Also would be great to have a compile test for that)

weiznich (approved):
This looks great :tada:

I've left some minor stylistic comments otherwise I think the code is good. We should really add a changelog entry for this before merging.

diesel/src/query_source/aliasing/macros.rs
```diff
@@ -84,6 +84,10 @@ macro_rules! alias {
                 pub const $alias_name: Alias<$alias_ty> = $($table)::+ as $alias_name;
             )*
         }
+        $(
+            #[allow(non_camel_case_types)]
+            type $alias_name = $crate::query_source::Alias::<$alias_ty>;
```
Ten0, line 89: Took the option here to not put the implicit type alias in the explicit `const` syntax.
I'd like a second opinion on this. (I have no strong perference as I'm not using this syntax anyway.)

diesel_derives/src/lib.rs
Ten0: I have some more of that in the dsl_auto_type crate:
https://github.com/Ten0/diesel/blob/71a5a4236e4dd860b1e0dd4ccfe42481b4b11743/dsl_auto_type/src/auto_type/mod.rs#L6-L13

Is that ok there or do you have a strong preference for changing it there as well?

diesel_derives/src/selectable.rs
Ten0: Actually quote_spanned only puts the span on the tokens that are directly written in the macro, but `#inferred_type` just keeps its span.
There was however an issue with the generated `_` placeholders though, which were not ideally spanned, so that is fixed in the type inferrence code.

diesel_derives/src/lib.rs
weiznich: As long as it is consistent in a single crate I do not have strong preferences here.

diesel/src/query_source/aliasing/macros.rs
weiznich: I think I would prefer to have it always there, just to have the always the same generated code. That makes it probably easier to work with that kind of code in the context of macros.

weiznich (approved):
:tada:

PR #3691, opened 2023-07-13
Expose push_bind_param_value_only to backends

Sytten:
Hello!
Working on the sea-query integration with diesel. Since I use the builder of sea-query I need to be able to bind parameters without inserting the placeholder like what SqlQuery does.
I think it's fair to put it behind the feature flag since it should not really we used by most people.

### DIFF
--- a/diesel/src/query_builder/ast_pass.rs
+++ b/diesel/src/query_builder/ast_pass.rs
@@ -227,6 +227,14 @@ where
         Ok(())
     }
 
+    /// Push a value onto the given query to be sent separate from the SQL
+    ///
+    /// This method affects multiple AST passes. It should be called at the
+    /// point in the raw SQL is inserted. This assumes the parameter placeholder
+    /// (`$1` on PG, `?` on other backends) is already inserted.
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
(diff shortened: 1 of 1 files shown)


diesel/src/query_builder/ast_pass.rs
```diff
@@ -227,7 +227,20 @@ where
...
+    /// (`$1` on PG, `?` on other backends) is already inserted.
+    #[cfg_attr(
+        not(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"),
+        doc(hidden)
+    )] // This is used by raw SQL fragments
+    #[cfg_attr(
+        doc_cfg,
+        doc(cfg(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"))
+    )]
+    pub fn push_bind_param_value_only<T, U>(&mut self, bind: &'b U) -> QueryResult<()>
```
weiznich, line 243: No need to to this kind of complicated setup here. We have an proc_macro attribute for that which takes care of all the things.

```suggestion
    #[diesel_derives::__diesel_public_if(
        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
    )]
    pub(crate) fn push_bind_param_value_only<T, U>(&mut self, bind: &'b U) -> QueryResult<()>
```

weiznich (changes requested):
I'm ok with this change, but I would like to use the existing proc macro for making the publicity of that function dependent on the right feature flag. (That just makes it easier to read and ensures that we won't miss anything)

PR #3694, opened 2023-07-14
Fix sqlite timestamps

### LINKED ISSUE #3320: Wrong documentation around sqlite datetime / inference not taking tz into account
### Versions

- **Rust:1.61**
- **Diesel:2.0.0**
- **Database:Sqlite**
- **Operating System**

### Feature Flags

- **diesel: time**

## Problem Description

It looks like the documentation surrounding timestamps for Sqlite is wrong.
Specially https://github.com/diesel-rs/diesel/blob/e9a951f1dc8ae893c31002ea2e00690c91e86ce4/diesel/src/sql_types/mod.rs#L335 and https://github.com/diesel-rs/diesel/blob/e9a951f1dc8ae893c31002ea2e00690c91e86ce4/diesel/src/sql_types/mod.rs#L342

The `FromSql` to `ToSql` traits are only implemented for `TimestamptzSqlite` for `OffsetDateTime`. Which currently you have to change manually in the schema since there is no inference for it.

- Allow inference to automatically detect the TZ version
- Change the documentation to reflect that
- (Maybe) Rename `TimestamptzSqlite` since it's pretty ugly, but I guess there is a reason behind that choice

I will do the change if we agree on what the behaviour should be.

## Checklist

### LINKED ISSUE #3693: ToSql<TimestamptzSqlite, Sqlite> is inconsistent between NaiveDateTime and PrimitiveDateTime
### Versions

- **Rust: 1.60**
- **Diesel: 2.1.0**
- **Database: sqlite**
- **Operating System: All**

### Feature Flags

- **diesel:** `time`, `chrono`

## Problem Description

Currently the encoding of `DateTime` without a timezone in `chrono` and `time` are not consistent.
With `chrono` we add a `+00.00` and not with `time`, this breaks the parsing if someone switches from `time` to `chrono` because the `FromSql` of `chrono` expects to have a `Z` or a `+00.00`.

```rust
impl ToSql<TimestamptzSqlite, Sqlite> for NaiveDateTime {
    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
        out.set_value(self.format("%F %T%.f+00:00").to_string());
        Ok(IsNull::No)
    }
}
```

```rust
const ENCODE_PRIMITIVE_DATETIME_FORMAT_WHOLE_SECOND: &[FormatItem<'_>] =
    format_description!("[year]-[month]-[day] [hour]:[minute]:[second]");
const ENCODE_PRIMITIVE_DATETIME_FORMAT_SUBSECOND: &[FormatItem<'_>] =
    format_description!("[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:6]");

impl ToSql<TimestamptzSqlite, Sqlite> for PrimitiveDateTime {
    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
        let format = if self.microsecond() == 0 {
            ENCODE_PRIMITIVE_DATETIME_FORMAT_WHOLE_SECOND
        } else {
            ENCODE_PRIMITIVE_DATETIME_FORMAT_SUBSECOND
        };
        out.set_value(self.format(format).map_err(|err| err.to_string())?);
        Ok(IsNull::No)
    }
}
```

I am doing some testing of the sqlite as part of the sea-query/diesel bridge and I found that inserting records with a `PrimitiveDateTime` and reading them using a `NaiveDateTime` breaks.

It should be able to read both ways because people might switch their library and would not expect that part to break.
Either the `+00.00` is wrong in `ToSql<TimestamptzSqlite, Sqlite> for NaiveDateTime` or its missing in `ToSql<TimestamptzSqlite, Sqlite> for PrimitiveDateTime`.
This might be "hard" to change to though, so I would suggest to be more permissive in the `FromSql<TimestamptzSqlite, Sqlite> for NaiveDateTime` by supporting the format without a timezone (or a `Z`).

`DeserializationError("Invalid datetime 2020-01-01 02:02:02")`

No

### Steps to reproduce

```rust
diesel::table! {
    character (id) {
        id -> Integer,
        created -> TimestamptzSqlite,
    }
}

#[derive(QueryableByName, Debug)]
#[diesel(table_name = character)]
struct Character {
    id: i32,
    created: NaiveDateTime,
}

#[derive(Insertable, Debug)]
#[diesel(table_name = character)]
struct CharacterInsertSet {
    id: i32,
    created: PrimitiveDateTime,
}
```

1. Create a sqlite table character as set above
2. Insert using `CharacterInsertSet`
3. Try to read using `Character`
4. It fails

## Checklist

Sytten:
Fixes #3693 
Fixes #3320

This is a big change in the way we store and parse timestamp in sqlite.
- **(Major)** We don't ignore timezones anymore if they are present, I am unsure why this what put in place but it feels very very very wrong to me. The data written by diesel is always switched to UTC, but we can't assume the data in the database was written by us. We still fallback to a "naive" implementation if the timezone is missing. I had to remove "bad" tests since `1970-01-01 00:00+01:00` is not the same time as `1970-01-01 00:00Z` in my book. We still ignore the timezone if the user uses a "naive" datetime to read the data (I think that is fair, though we could verify with other ORM if they correct to UTC instead).
- Uniformity between time and chrono, the code looks more similar and the formats are the same so a user can switch library without any issue now (that was something I found in #3693)
- Fixed some documentation

### DIFF
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -332,12 +332,14 @@ pub struct Time;
 /// - [`std::time::SystemTime`][SystemTime] (PG only)
 /// - [`chrono::NaiveDateTime`][NaiveDateTime] with `feature = "chrono"`
 /// - [`time::PrimitiveDateTime`] with `feature = "time"`
+/// - [`time::OffsetDateTime`] with `feature = "time"` (MySQL only)
 ///
 /// ### [`FromSql`](crate::deserialize::FromSql) impls
 ///
 /// - [`std::time::SystemTime`][SystemTime] (PG only)
 /// - [`chrono::NaiveDateTime`][NaiveDateTime] with `feature = "chrono"`
 /// - [`time::PrimitiveDateTime`] with `feature = "time"`
+/// - [`time::OffsetDateTime`] with `feature = "time"` (MySQL only)
 ///
 /// [SystemTime]: std::time::SystemTime
 #[cfg_attr(
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -332,14 +332,12 @@ pub struct Time;
 /// - [`std::time::SystemTime`][SystemTime] (PG only)
 /// - [`chrono::NaiveDateTime`][NaiveDateTime] with `feature = "chrono"`
 /// - [`time::PrimitiveDateTime`] with `feature = "time"`
-/// - [`time::OffsetDateTime`] with `feature = "time"`
 ///
 /// ### [`FromSql`](crate::deserialize::FromSql) impls
 ///
 /// - [`std::time::SystemTime`][SystemTime] (PG only)
 /// - [`chrono::NaiveDateTime`][NaiveDateTime] with `feature = "chrono"`
 /// - [`time::PrimitiveDateTime`] with `feature = "time"`
-/// - [`time::OffsetDateTime`] with `feature = "time"`
 ///
 /// [SystemTime]: std::time::SystemTime
 #[cfg_attr(
--- a/diesel/src/sqlite/types/date_and_time/chrono.rs
+++ b/diesel/src/sqlite/types/date_and_time/chrono.rs
@@ -1,5 +1,4 @@
 extern crate chrono;
-use chrono::Timelike;
 
 use self::chrono::{DateTime, Local, NaiveDate, NaiveDateTime, NaiveTime, TimeZone, Utc};
 
@@ -9,10 +8,16 @@ use crate::serialize::{self, IsNull, Output, ToSql};
 use crate::sql_types::{Date, Time, Timestamp, TimestamptzSqlite};
 use crate::sqlite::Sqlite;
 
+// Warning to future editors:
+// Changes in the following formats need to be kept in sync
+// with the formats of the "time" module.
+// We do not need a distinction between whole second and
+// subsecond since %.f will only print the dot if needed.
+// We always print as many subsecond as his given to us,
+// this means the subsecond part can be 3, 6 or 9 digits.
 const DATE_FORMAT: &str = "%F";
 
-const ENCODE_TIME_FORMAT_WHOLE_SECOND: &str = "%T";
-const ENCODE_TIME_FORMAT_SUBSECOND: &str = "%T%.f";
+const ENCODE_TIME_FORMAT: &str = "%T%.f";
 
 const TIME_FORMATS: [&str; 9] = [
     // Most likely formats
@@ -20,11 +25,9 @@ const TIME_FORMATS: [&str; 9] = [
     "%R", "%RZ", "%R%:z", "%TZ", "%T%:z", "%T%.fZ", "%T%.f%:z",
 ];
 
-const ENCODE_NAIVE_DATETIME_FORMAT_WHOLE_SECOND: &str = "%F %T";
-const ENCODE_NAIVE_DATETIME_FORMAT_SUBSECOND: &str = "%F %T%.f";
+const ENCODE_NAIVE_DATETIME_FORMAT: &str = "%F %T%.f";
 
-const ENCODE_DATETIME_FORMAT_WHOLE_SECOND: &str = "%F %T%:z";
-const ENCODE_DATETIME_FORMAT_SUBSECOND: &str = "%F %T%.f%:z";
+const ENCODE_DATETIME_FORMAT: &str = "%F %T%.f%:z";
 
 const NAIVE_DATETIME_FORMATS: [&str; 18] = [
     // Most likely formats
@@ -110,12 +113,7 @@ impl FromSql<Time, Sqlite> for NaiveTime {
 #[cfg(all(feature = "sqlite", feature = "chrono"))]
 impl ToSql<Time, Sqlite> for NaiveTime {
     fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
-        let format = if self.nanosecond() == 0 {
-            ENCODE_TIME_FORMAT_WHOLE_SECOND
-        } else {
-            ENCODE_TIME_FORMAT_SUBSECOND
-        };
-        out.set_value(self.format(format).to_string());
+        out.set_value(self.format(ENCODE_TIME_FORMAT).to_string());
         Ok(IsNull::No)
     }
 }
@@ -144,12 +142,7 @@ impl FromSql<Timestamp, Sqlite> for NaiveDateTime {
 #[cfg(all(feature = "sqlite", feature = "chrono"))]
 impl ToSql<Timestamp, Sqlite> for NaiveDateTime {
     fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
-        let format = if self.nanosecond() == 0 {
-            ENCODE_NAIVE_DATETIME_FORMAT_WHOLE_SECOND
-        } else {
-            ENCODE_NAIVE_DATETIME_FORMAT_SUBSECOND
-        };
-        out.set_value(self.format(format).to_string());
+        out.set_value(self.format(ENCODE_NAIVE_DATETIME_FORMAT).to_string());
         Ok(IsNull::No)
     }
 }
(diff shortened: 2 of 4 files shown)


diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -7,21 +9,85 @@ use crate::serialize::{self, IsNull, Output, ToSql};
...
+    // Most likely formats
+    "%T%.f", "%T", // All other valid formats in order of increasing specificity
+    "%R", "%RZ", "%R%:z", "%TZ", "%T%:z", "%T%.fZ", "%T%.f%:z",
+];
+
+const ENCODE_NAIVE_DATETIME_FORMAT_WHOLE_SECOND: &str = "%F %T";
+const ENCODE_NAIVE_DATETIME_FORMAT_SUBSECOND: &str = "%F %T%.f";
+
+const ENCODE_DATETIME_FORMAT_WHOLE_SECOND: &str = "%F %T%:z";
+const ENCODE_DATETIME_FORMAT_SUBSECOND: &str = "%F %T%.f%:z";
```
Sytten, line 27: I need to change `%.f` to `%.3f` here to be consistent with `time` crate

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -150,14 +178,33 @@ impl FromSql<TimestamptzSqlite, Sqlite> for NaiveDateTime {
 #[cfg(all(feature = "sqlite", feature = "chrono"))]
 impl ToSql<TimestamptzSqlite, Sqlite> for NaiveDateTime {
     fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
-        out.set_value(self.format("%F %T%.f+00:00").to_string());
+        let format = if self.nanosecond() == 0 {
```
Sytten, line 181: And change here to only care about the micro seconds to match the `%.3f`

diesel/src/sql_types/mod.rs
```diff
@@ -332,14 +332,12 @@ pub struct Time;
 /// - [`std::time::SystemTime`][SystemTime] (PG only)
 /// - [`chrono::NaiveDateTime`][NaiveDateTime] with `feature = "chrono"`
 /// - [`time::PrimitiveDateTime`] with `feature = "time"`
-/// - [`time::OffsetDateTime`] with `feature = "time"`
```
weiznich, line 335: There are such impls for the mysql backend. Probably we should change it to (mysql only) instead of removing these from the list

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -636,22 +704,6 @@ mod tests {
             "1970-01-01T00:00:00+00:00",
             "1970-01-01T00:00:00.000+00:00",
             "1970-01-01T00:00:00.000000+00:00",
-            "1970-01-01 00:00+01:00",
```
weiznich, line 639: I assume these are removed because the serialization/deserialization of these values changed? Can you explain how and why they changed?

(We need to be very careful here to not introduce a breaking change for existing users)

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -7,21 +9,85 @@ use crate::serialize::{self, IsNull, Output, ToSql};
 use crate::sql_types::{Date, Time, Timestamp, TimestamptzSqlite};
 use crate::sqlite::Sqlite;
 
-const SQLITE_DATE_FORMAT: &str = "%F";
+const DATE_FORMAT: &str = "%F";
```
weiznich, line 12: Can we add a comment here that these formats need to be in sync with that one in the `time` support module (and vice versa)?

weiznich (commented):
Thanks for opening this PR. I've left a bunch of questions and notes for minor improvements. Beside of that I want to clarify an important point: You consider that a minor change, as it does only strictly extend which time formats are parsed and does not error out on formats that worked before?

Additionally I would like to see a few tests for time<->chrono roundtrips just to make sure that they stay consistent in the future.

diesel/src/sqlite/types/date_and_time/chrono.rs
Sytten: Agreed we need to be super careful, I consider this test to be a bug but it might be considered a breaking change depending on how you look at it.

Basically right now if you insert manually a timestamp with a timezone in sqlite and **read** it using diesel it will ignore the timestamp. So for example `1970-01-01T00:00:00-01:00` will be treated as if it is `1970-01-01T00:00:00Z` when it is in fact ` 1970-01-01T00:01:00Z`. So for me this is a bug, but it does mean that people might get different dates.

Why it works if you **write** with diesel is because it will always translate to UTC before writing it. I didnt change that behaviour (but we might want to).

weiznich: I think we can argue here that this is a bug fix. We should add it as potential breaking change to the changelog although.

Sytten:
@weiznich This is ready for another round of review, I changed a bit the encoding.

I realized we didn't need the differentiation for `subsecond` in `chrono` since the dot is handled by the formatter.
I changed `time` to now print all digits instead of just 3 or 6 like it currently done, this aligns it with `chrono` (and what `sqlx` is doing whatever that's worth).

I am still unsure how many digits we should really print, PG and Mysql cut at the microseconds but sqlite by default stops at the millisecond with `%f`. And if we should differentiate between a datetime with timezone and a "naive" datetime/time. Any external opinion would be appreciated, but it seems all over the place.

The case for printing they all is quite simple, it's still respecting the RFC3339 and the user can decide to not pass us the microseconds/nanoseconds if they don't want them in the database but the inverse (stopping at X digits) means we decide what loss of information is acceptable.

diesel/src/sqlite/types/date_and_time/chrono.rs
Mingun: Even better to make this comment doccomment and put a direct intradoc reference to the corresponding constant(?), then editors, like VS Code, will able to create a clickable link to it

diesel/src/sqlite/types/date_and_time/chrono.rs
```diff
@@ -7,21 +8,89 @@ use crate::serialize::{self, IsNull, Output, ToSql};
...
 use crate::sqlite::Sqlite;
 
-const SQLITE_DATE_FORMAT: &str = "%F";
+// Warning to future editors:
+// Changes in the following formats need to be kept in sync
+// with the formats of the "time" module.
+// We do not need a distinction between whole second and
+// subsecond since %.f will only print the dot if needed.
+// We always print as many subsecond as his given to us,
+// this means the subsecond part can be 3, 6 or 9 digits.
```
weiznich, line 17:
```suggestion
/// Warning to future editors:
/// Changes in the following formats need to be kept in sync
/// with the formats of the ["time"](super::time) module.
/// We do not need a distinction between whole second and
/// subsecond since %.f will only print the dot if needed.
/// We always print as many subsecond as his given to us,
/// this means the subsecond part can be 3, 6 or 9 digits.
```

[as suggested here: ](https://github.com/diesel-rs/diesel/pull/3694/files#r1275723428)

diesel/src/sqlite/types/date_and_time/mod.rs
```diff
@@ -92,3 +92,265 @@ impl ToSql<sql_types::TimestamptzSqlite, Sqlite> for String {
...
+            .execute(conn)
+            .unwrap();
+
+        crate::sql_query("CREATE TABLE table_time(id INTEGER PRIMARY KEY, time TEXT);")
+            .execute(conn)
+            .unwrap();
+    }
+
+    #[test]
+    fn time_to_chrono_date() {
```
weiznich, line 157: Can we actually check the results of that queries for whether or not they return the "right" values?

diesel/src/sqlite/types/date_and_time/time.rs
```diff
@@ -14,23 +14,30 @@ use crate::serialize::{self, IsNull, Output, ToSql};
 use crate::sql_types::{Date, Time, Timestamp, TimestamptzSqlite};
 use crate::sqlite::Sqlite;
 
+// Warning to future editors:
+// Changes in the following formats need to be kept in sync
+// with the formats of the "chrono" module.
+// We need a distinction between whole second and subsecond
+// since there is no format option to forgo the dot.
+// We always print as many subsecond as his given to us,
+// this means the subsecond part can be between 1 and 9 digits.
```
weiznich, line 23:
```suggestion
/// Warning to future editors:
/// Changes in the following formats need to be kept in sync
/// with the formats of the ["chrono"](super::chrono) module.
/// We need a distinction between whole second and subsecond
/// since there is no format option to forgo the dot.
/// We always print as many subsecond as his given to us,
/// this means the subsecond part can be between 1 and 9 digits.
```

weiznich (commented):
Thanks for the update and sorry for taking so long to do the review.

I think it's mostly fine now. I would like to change the comments to doc-comments as suggested my @Mingun 
I also would like to actually check the values in the compatibility tests to ensure that we load the right thing there.

Sytten:
@weiznich Ready!

weiznich (approved):
Thanks :+1:

PR #3714, opened 2023-07-25
Implementing ToSql for [u8;N]

snf:
Hi thanks for developing and supporting diesel.

This PR would allow to bind slices of arbitrary size to queries (my specific usecase is raw queries). It's mostly for convenience to prevent having to reshape types prior to sending them to diesel.

An alternative would be to implement it for all the `T: AsRef<[u8]>`.

### DIFF
--- a/diesel/src/type_impls/primitives.rs
+++ b/diesel/src/type_impls/primitives.rs
@@ -116,7 +116,7 @@ mod foreign_impls {
     struct BinarySliceProxy([u8]);
 
     #[derive(AsExpression)]
-    #[diesel(foreign_derive, not_sized)]
+    #[diesel(foreign_derive)]
     #[diesel(sql_type = Binary)]
     struct BinaryArrayProxy<const N: usize>([u8; N]);
 }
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. Could you add at least some minor text for the motivation of that PR.

weiznich (changes requested):
Thanks for the update. The change itself looks good. 

I would like to see a test for that in `diesel_tests/tests/types.rs` just to have something for the future to assert that this impl is there.

weiznich (approved):
Thanks

snf:
Thank you! Waiting for the tests though

weiznich:
> Waiting for the tests though

It's configured in such a way that it only merges the PR if the test passes. I just allows me to not coming back for merging in a day or two if everything is fine.

snf:
I'll come back to it this afternoon (failed test)

snf:
I'm not sure all of this is actually needed. I can also squash the commits into a single one to reduce noise.

weiznich (approved):

PR #3724, opened 2023-08-01
Allow `#[diesel(treat_none_as_{null, default_value} = …]` on fields

moulins:
If present on a field, this will override the corresponding container-level attribute.

This restores feature parity with the old `diesel 1.x` versions, which allowed these two attributes on struct fields.


pksunkara (commented):
Need tests, both happy and error cases.

Ten0 (approved):

moulins:
Something I just thought about: as implemented, putting an explicit `treat_none_as_null` on a non-`Option` field will do nothing; maybe it should be an error instead?

Ten0:
> maybe it should be an error instead

yes that makes sense, that would be better

Issue #3717, opened 2023-07-27
Filter table not applied on `diesel migration --diff-schema`


weiznich:
### Discussed in https://github.com/diesel-rs/diesel/discussions/3712

<div type='discussions-op-text'>

<sup>Originally posted by **forest1102** July 25, 2023</sup>
I am trying out a new option `diff-schema`.  
However, when I use the `filter` field in diesel.toml just omits tables used in other crates. `diesel migration generate edit_column_name --diff-schema` results in writing code to delete tables not listed in the filter fields. I expect migration do nothing about unlisted tables.

# Inputs

initial database info
```sql
CREATE TABLE "table_a"(
	"id" INT4 NOT NULL PRIMARY KEY,
	"code" VARCHAR NOT NULL
);
CREATE TABLE "table_b"(
	"id" INT4 NOT NULL PRIMARY KEY,
);
CREATE TABLE "table_c"(
	"id" INT4 NOT NULL PRIMARY KEY,
);
```

diesel.toml
```  :diesel.toml
[print_schema]



file = "schema_a/src/schema.rs"
filter = { only_tables = [
    "^table_a$",
] }
```

schema_a/src/schema.rs
``` diff
diesel::table! {
    table_a (id) {
        id -> Int8,
-       code -> Varchar,
+      script -> Varchar,
    }
}
```

command
```shell
diesel migration generate edit_column_name --diff-schema
```

## Actual Result

up.sql
```sql
DROP TABLE IF EXISTS "table_b";
ALTER TABLE "table_a" DROP COLUMN "code";
ALTER TABLE "table_a" ADD COLUMN "script" VARCHAR NOT NULL;
DROP TABLE IF EXISTS "table_c";
```

## Expected

up.sql
```sql
ALTER TABLE "table_a" DROP COLUMN "code";
ALTER TABLE "table_a" ADD COLUMN "script" VARCHAR NOT NULL;
```

## Version
```
diesel 
 Version: 2.1.0
 Supported Backends: postgres
```</div>

weiznich:
The issue here is that the `diff-schema` code currently just ignores these filters. That means we need to add support for these filters at the relevant location here:

https://github.com/diesel-rs/diesel/blob/f4d5f93ccfd039702361e3a5a1b832b7be4839b8/diesel_cli/src/migrations/diff_schema.rs#L46

See the `print-schema` code for how to apply the filtering:

[...]

PR #3725, opened 2023-08-02
Fix/3717 filter table on `diesel migration --diff-schema`

gmanninglive:
Fixes: #3717 

I have made changes to allow `diesel migration --diff-schema` to accept filter args matching `diesel print-schema`

I came across a few issues while working on this:
1. I can't think of how to move the `filter` attribute from `PrintSchema`  to `Config` without a breaking change
2. The documentation for `table-names` arg seems incorrect to me, It currently states `(default only-tables if not empty).` But I'm pretty sure it [defaults to `None` currently](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/main.rs#L236)

**TODO:** 
1. Update documentation where necessary

I'll add some more info to this PR description before requesting review


diesel_cli/src/migrations/mod.rs
```diff
@@ -88,7 +88,7 @@ pub(super) fn run_migration_command(
                 if let Some(diff_schema) = diff_schema {
                     self::diff_schema::generate_sql_based_on_diff_schema(
                         config,
-                        matches,
+                        args,
```
weiznich, line 91: Can you verify that it continues to work passing `--database-url` as global parameter after this change? So that `diesel --database-url migration generate --diff-schema` continues to work?

weiznich (approved):
Thanks for working on this. The implementation looks really good :+1: I have one minor question otherwise this is fine from my side.

diesel_cli/src/migrations/mod.rs
gmanninglive: ah good spot, I'll have a look into this

weiznich:
I've rebased this branch to pull in the changes from https://github.com/diesel-rs/diesel/pull/3729 to fix the broken tests.

diesel_cli/src/migrations/mod.rs
gmanninglive: @weiznich I can confirm, the global parameters are passed to the sub-commands. I can add a test case for this if needed but it is [claps api](https://docs.rs/clap/latest/clap/struct.Arg.html#method.global) which is unlikely to change

diesel_cli/src/cli.rs
```diff
@@ -145,6 +147,29 @@ pub fn build_cli() -> Command {
                         .default_missing_value("NOT_SET")
                         .num_args(0..=1)
                         .require_equals(true),
+                )
+                .arg(
+                    Arg::new("table-name")
+                        .index(2)
+                        .num_args(1..)
+                        .action(clap::ArgAction::Append)
+                        .help("Table names to filter (default only-tables if not empty)."),
```
gmanninglive, line 156: @weiznich would you like me to change this to:
```suggestion
                        .help("Table names to filter"),
```

Because as far as I can see, filtering so far has only worked by specifying `-o` or `-e`
the change would also be made for print-schema [below](https://github.com/diesel-rs/diesel/pull/3725/files#diff-324fe63179d6dfa0bc37bfbf038adadd1782c3448c9034c1142fef59b8984bf7R226)

gmanninglive:
> I've rebased this branch to pull in the changes from #3729 to fix the broken tests.

Thanks for rebasing this!, I've left a couple comments and will set ready for review. Thanks for approving already too :smile:

diesel_cli/src/cli.rs
weiznich: I'm in favor of this change (also for the print-schema) part.

diesel_cli/src/migrations/mod.rs
weiznich: I think it's fine to leave these tests to clap. It's just that I forget from time to time what's the behaviour there.

gmanninglive: Great! I agree with leaving it to clap.

weiznich (commented):
Thanks :+1:

diesel_cli/src/cli.rs
weiznich: No worries from taking some time. We all work on this in our free time, so it's fine to take as much time as needed. There is always real life, which is more important than this.

PR #3737, opened 2023-08-10
Tweak the creation of database connections in diesel_CLI

weiznich:
This commit changes how we establish database connections in diesel cli. Instead of relying ot the auto generated `InferConnection::establish` method we manually construct the right enum variant. This allows us to easily return the underlying error message instead of emitting a generic ` Invalid connection url for multiconnection` error.


chusteven (approved):
Good idea!

PR #3744, opened 2023-08-14
Fix missing type alias order by for auto type

### LINKED ISSUE #3736: Missing type alias OrderBy for dsl::auto_type
### Versions

- **Rust:** rustc --version = rustc 1.70.0 (90c541806 2023-05-31)
- **Diesel:** master
- **Database:** _
- **Operating System** Archlinux

### Feature Flags

- **diesel:** "postgres", "r2d2", "serde_json", "chrono",

## Problem Description
QueryDsl provides an alias for `order` named [`order_by`](https://docs.rs/diesel/latest/diesel/prelude/trait.QueryDsl.html#method.order_by). Therefore type generated using `#[dsl::auto_type]` is `diesel::dsl::OrderBy` which does not exists! I therefore think we should add a type alias `OrderBy` for `Order`.

I see two other alternatives:
1. Deprecate `order_by` on QueryDsl
2. Add documentation on `order_by` to indicate that it doesn't work with `dsl::auto_type` and that one should use `order` in that case.

I can provide a PR once solution has been chosen.

Use  `#[dsl::auto_type]`

I'd like a simplier solution to use dsl::auto_type

## Checklist

Elrendio:
Resolves #3736

**What it does:**
- Adds type alias [`OrderBy`](crate::dsl::OrderBy) pointing to [` Order`](crate::dsl::Order) and uses it as return type or [`.order_by(ordering)`](crate::prelude::QueryDsl::order_by)


weiznich (approved):
Thanks :+1:

PR #3747, opened 2023-08-15
Add `Bool` type to MultiConnection

BlackDex:
During trying to implement MultiConnection i found that the `Bool` type was missing.

This PR adds this type.

### DIFF
--- a/diesel_derives/src/multiconnection.rs
+++ b/diesel_derives/src/multiconnection.rs
@@ -582,6 +582,7 @@ fn generate_bind_collector(connection_types: &[ConnectionVariant]) -> TokenStrea
             quote::quote!(diesel::sql_types::Binary),
             quote::quote!([u8]),
         ),
+        (quote::quote!(diesel::sql_types::Bool), quote::quote!(bool)),
     ]
     .into_iter()
     .map(|t| generate_to_sql_impls(t, connection_types));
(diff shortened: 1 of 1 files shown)


weiznich:
I think just adding this impl would be a breaking change :(

Existing users of `#[derive(MultiConnection)]` could have already manually implemented that these traits. On the other hand "breaking" changes are allowed for bug fixes and we could argue that this is a bug fix. 
(I wouldn't expect any breakage here as otherwise we would have seen reports about the missing impl earlier)

BlackDex:
I would also say this is a bug fix, since this is a missing default type. It's not that this is a whole new type not previously existing on the separate backends.

weiznich:
cc @diesel-rs/core on whether or no we want to consider that as bug fix.

sgrif:
This seems backwards compatible afaict so treating it as a bug fix seems fine

Ten0:
IMO if somebody was competent enough to realize that impl was missing and implemented it manually without bothering to open an issue, breaking change is fair enough ^^

Ten0 (approved):

weiznich:
Thanks for the swift responses. Seems like we are all on line in calling this a bug fix :+1:

@pksunkara Your input is appreciated here as well :heart:

Issue #3745, opened 2023-08-14
`dsl::max` doesn't work with `dsl::auto_type`


Elrendio:
### Versions

- **Rust:** rustc --version = rustc 1.70.0 (90c541806 2023-05-31)
- **Diesel:** master
- **Database:** _
- **Operating System** Archlinux

### Feature Flags

- **diesel:** "postgres", "r2d2", "serde_json", "chrono",

## Problem Description
FYI @Ten0 knows and says it's a blocker for the release of `auto_type` so he'll try to fix.

When  deriving on a struct `Selectable` with one field having a `select_expression` containing `dsl::max` compilation breaks with error:
```
error[E0573]: expected type, found function `diesel::dsl::max`
   --> mod.rs:463:6
    |
463 |                     diesel::dsl::max<table::column>,
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not a type
    |
help: consider importing this type alias instead
    |
8   + use diesel::helper_types::max;
    |
help: if you import `max`, refer to it directly
    |
463 -                     diesel::dsl::max<table::column>,
463 +                     max<table::column>,
    |
```
### What are you trying to accomplish?
Use `Selectable` derive

### What is the expected output?
Code compiles

### What is the actual output?
Compilation fails

### Are you seeing any additional errors?
No

### Steps to reproduce
```rust
mod schema {

    diesel::table! {
	    table (id) {
		    id -> Int4,
                    other_id -> Int4,,
		    column -> Int4,
	    }
    }
    diesel::table! {
	    other_table (id) {
		    id -> Int4,
		    column -> Int4,
	    }
    }
}

#[derive(Queryable, Serialize, Selectable)]
#[diesel(table_name = table)]
pub(crate) struct Struct {
	#[diesel(
		select_expression = schema::table::table
			.filter(schema::table::other_id.eq(schema::other_table::id))
			.select(dsl::max(table::column))
			.single_value(),
	)]
	pub(crate) max_val: Option<i32>,
}

```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

[...]

PR #3751, opened 2023-08-16
Mention calling of SQL functions in `Selectable` docs

ISibboI:
In the custom select expression example for the `Selectable` trait, mention that this can be used to call arbitrary SQL functions as well. Mention the `CURRENT_TIMESTAMP` function explicitly.

Relates to #3739


ISibboI:
Seems like some test fails, which is strange, because I didn't change any code. Must be some random failure, because it talks about a deadlock...

diesel/src/expression/mod.rs
```diff
@@ -544,12 +544,18 @@ where
...
+/// This is useful for example to
+///
+///  * avoid nesting types, or to
+///  * populate fields with values other than table columns, such as
+///    the result of an SQL function like `CURRENT_TIMESTAMP()`
+///    or a custom SQL function.
+///
+/// The select expression is specified via the `select_expression` parameter.
+/// For more complex expressions, it may be required to also specify the result type
+/// of the expression with `select_expression_type` (most query fragments are supported for this).
```
weiznich, line 558: I feel the last part in the parenthesis needs to be updated now. Something like the following should work.

```suggestion
/// of the expression with `select_expression_type`. For most query fragments this can be inferred and is not needed to be specified.
```

weiznich (commented):
Thanks for this PR. I have some minor wording request otherwise this looks good :+1:

diesel/src/expression/mod.rs
```diff
@@ -544,12 +544,18 @@ where
...
-/// flatten the fields. If the `select_expression` is simple enough,
-/// it is not necessary to specify `select_expression_type`
-/// (most query fragments are supported for this).
+/// It is also possible to specify an entirely custom select expression
+/// for fields when deriving [`Selectable`](derive@Selectable).
+/// This is useful for example to
+///
+///  * avoid nesting types, or to
+///  * populate fields with values other than table columns, such as
+///    the result of an SQL function like `CURRENT_TIMESTAMP()`
```
Ten0, line 553: Until we generate the appropriate type from `sql_function` I don't think that currently works, does it? (#3745)

diesel/src/expression/mod.rs
ISibboI: Done.

diesel/src/expression/mod.rs
ISibboI: Thanks for the review! I have now added a field to the doctest below this comment that gets initialised with `diesel::dsl::now`. It seems to complain about some trait not being implemented:

```
error[E0277]: the trait bound `expression::select_by::SelectBy<UserPost, _>: load_dsl::private::CompatibleType<_, _>` is not satisfied
    --> src/expression/mod.rs:596:12
     |
37   |     .first(connection)?;
     |      ----- ^^^^^^^^^^ the trait `load_dsl::private::CompatibleType<_, _>` is not implemented for `expression::select_by::SelectBy<UserPost, _>`
     |      |
     |      required by a bound introduced by this call
     |
     = help: the trait `load_dsl::private::CompatibleType<U, DB>` is implemented for `expression::select_by::SelectBy<U, DB>`
     = note: required for `SelectStatement<FromClause<JoinOn<Join<table, table, Inner>, Grouped<Eq<Nullable<user_id>, Nullable<id>>>>>, ..., ..., ..., ..., ...>` to implement `LoadQuery<'_, _, _>`
     = note: the full type name has been written to '/tmp/rustdoctestBQRQaQ/rust_out.long-type-2905737123615719141.txt'
note: required by a bound in `first`
    --> /home/sibbo/git/diesel/diesel/src/query_dsl/mod.rs:1780:22
     |
1780 |         Limit<Self>: LoadQuery<'query, Conn, U>,
     |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `RunQueryDsl::first`
```

Is this what you mean?

Ten0: Actually for the particular case of `dsl::now` given how it's defined it would work, however if a user were to define a custom function that does `CURRENT_TIMESTAMP()` using `sql_function!` that wouldn't work.

I suspect your error comes from the type you are selecting to, and you may get a more explicit error message using `check_for_backend`.

pksunkara (approved):

diesel/src/expression/mod.rs
```diff
@@ -551,12 +551,19 @@ where
...
+///    or a custom SQL function.
+///
+/// The select expression is specified via the `select_expression` parameter.
+/// For more complex expressions, it may be required to also specify the result type
+/// of the expression with `select_expression_type`, However, for most query fragments
+/// this can be inferred and is not needed to be specified.
```
Ten0, line 566:
```suggestion
/// this can be inferred and does not need to be specified.
```

Ten0 (approved):

PR #3778, opened 2023-09-05
Monomorphization of 'cached_statement'

smklein:
For context, see: https://matklad.github.io/2021/09/04/fast-rust-builds.html#Keeping-Instantiations-In-Check

This PR attempts to preserve the API used by the statement_cache, but attempts to monomorphize over the underlying `QueryFragment + QueryId` type.

The generics in the statement cache appear to be:

1. The DB backend
2. The `QueryFragment + QueryId` type

Although it probably makes sense to preserve being generic over the backend, duplicating this code for each query type results in a noticeable binary bloat. In a debug build of https://github.com/oxidecomputer/omicron , this change reduces the binary size by ~620 KiB.

I'm putting forward this PR as a bit of an experiment -- I think there are many other spots (for example, over the connection types) where we're generic over the `QueryFragment + QueryId` types, and for which we generate a significant amount of code for each query. I think, in the limit, it should be possible to generate this code "once per backend", and monomorphize more of the "per-query" codepaths.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diesel"
-version = "2.1.0"
+version = "2.1.1"
 license = "MIT OR Apache-2.0"
 description = "A safe, extensible ORM and Query Builder for PostgreSQL, SQLite, and MySQL"
 readme = "README.md"
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "diesel"
-version = "2.1.1"
+version = "2.1.0"
 license = "MIT OR Apache-2.0"
 description = "A safe, extensible ORM and Query Builder for PostgreSQL, SQLite, and MySQL"
 readme = "README.md"
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -177,29 +177,9 @@ where
     /// parameter indicates if the constructed prepared statement will be cached or not.
     /// See the [module](self) documentation for details
     /// about which statements are cached and which are not cached.
+    // Note: This function is intentionally monomorphic over the "source" type.
     #[allow(unreachable_pub)]
-    pub fn cached_statement<T, F>(
-        &mut self,
-        source: &T,
-        backend: &DB,
-        bind_types: &[DB::TypeMetadata],
-        prepare_fn: F,
-    ) -> QueryResult<MaybeCached<'_, Statement>>
-    where
-        T: QueryFragment<DB> + QueryId,
-        F: FnOnce(&str, PrepareForCache) -> QueryResult<Statement>,
-    {
-        let mut f = Some(prepare_fn);
-        self.cached_statement_inner(T::query_id(), source, backend, bind_types, &mut |s, p| {
-            let f = f.take().unwrap();
-            f(s, p)
-        })
-    }
-
-    // Note: This function is a direct implementation of "cached_statement", but
-    // it is intentionally monomorphic over the "source" type.
-    #[inline(never)]
-    fn cached_statement_inner(
+    pub fn cached_statement(
         &mut self,
         maybe_type_id: Option<TypeId>,
         source: &dyn QueryFragment<DB>,
@@ -307,7 +287,6 @@ where
 {
     /// Create a new statement cache key for the given query source
     // Note: Intentionally monomorphic over source.
-    #[inline(never)]
     #[allow(unreachable_pub)]
     pub fn for_source(
         maybe_type_id: Option<TypeId>,
@@ -332,7 +311,6 @@ where
     /// This is an optimization that may skip constructing the query string
     /// twice if it's already part of the current cache key
     // Note: Intentionally monomorphic over source.
-    #[inline(never)]
     #[allow(unreachable_pub)]
     pub fn sql(&self, source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<Cow<'_, str>> {
         match *self {
@@ -342,7 +320,6 @@ where
     }
 
     // Note: Intentionally monomorphic over source.
-    #[inline(never)]
     fn construct_sql(source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<String> {
         let mut query_builder = DB::QueryBuilder::default();
         source.to_sql(&mut query_builder, backend)?;
--- a/diesel/src/connection/statement_cache.rs
+++ b/diesel/src/connection/statement_cache.rs
@@ -189,9 +189,27 @@ where
         T: QueryFragment<DB> + QueryId,
         F: FnOnce(&str, PrepareForCache) -> QueryResult<Statement>,
     {
+        let mut f = Some(prepare_fn);
+        self.cached_statement_inner(T::query_id(), source, backend, bind_types, &mut |s, p| {
+            let f = f.take().unwrap();
+            f(s, p)
+        })
+    }
+
+    // Note: This function is a direct implementation of "cached_statement", but
+    // it is intentionally monomorphic over the "source" type.
+    #[inline(never)]
+    fn cached_statement_inner(
+        &mut self,
+        maybe_type_id: Option<TypeId>,
+        source: &dyn QueryFragment<DB>,
+        backend: &DB,
+        bind_types: &[DB::TypeMetadata],
+        prepare_fn: &mut dyn FnMut(&str, PrepareForCache) -> QueryResult<Statement>,
+    ) -> QueryResult<MaybeCached<'_, Statement>> {
         use std::collections::hash_map::Entry::{Occupied, Vacant};
 
-        let cache_key = StatementCacheKey::for_source(source, bind_types, backend)?;
+        let cache_key = StatementCacheKey::for_source(maybe_type_id, source, bind_types, backend)?;
 
         if !source.is_safe_to_cache_prepared(backend)? {
             let sql = cache_key.sql(source, backend)?;
@@ -288,16 +306,16 @@ where
     DB::TypeMetadata: Clone,
 {
     /// Create a new statement cache key for the given query source
+    // Note: Intentionally monomorphic over source.
+    #[inline(never)]
     #[allow(unreachable_pub)]
-    pub fn for_source<T>(
-        source: &T,
+    pub fn for_source(
+        maybe_type_id: Option<TypeId>,
+        source: &dyn QueryFragment<DB>,
         bind_types: &[DB::TypeMetadata],
         backend: &DB,
-    ) -> QueryResult<Self>
-    where
-        T: QueryFragment<DB> + QueryId,
-    {
-        match T::query_id() {
+    ) -> QueryResult<Self> {
+        match maybe_type_id {
             Some(id) => Ok(StatementCacheKey::Type(id)),
             None => {
                 let sql = Self::construct_sql(source, backend)?;
@@ -313,15 +331,19 @@ where
     ///
     /// This is an optimization that may skip constructing the query string
     /// twice if it's already part of the current cache key
+    // Note: Intentionally monomorphic over source.
+    #[inline(never)]
     #[allow(unreachable_pub)]
-    pub fn sql<T: QueryFragment<DB>>(&self, source: &T, backend: &DB) -> QueryResult<Cow<'_, str>> {
+    pub fn sql(&self, source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<Cow<'_, str>> {
         match *self {
             StatementCacheKey::Type(_) => Self::construct_sql(source, backend).map(Cow::Owned),
             StatementCacheKey::Sql { ref sql, .. } => Ok(Cow::Borrowed(sql)),
         }
     }
 
-    fn construct_sql<T: QueryFragment<DB>>(source: &T, backend: &DB) -> QueryResult<String> {
+    // Note: Intentionally monomorphic over source.
+    #[inline(never)]
+    fn construct_sql(source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<String> {
         let mut query_builder = DB::QueryBuilder::default();
         source.to_sql(&mut query_builder, backend)?;
         Ok(query_builder.finish())
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -248,8 +248,10 @@ fn prepared_query<'a, T: QueryFragment<Mysql> + QueryId>(
     statement_cache: &'a mut StatementCache<Mysql, Statement>,
     raw_connection: &'a mut RawConnection,
 ) -> QueryResult<MaybeCached<'a, Statement>> {
-    let mut stmt = statement_cache
-        .cached_statement(source, &Mysql, &[], |sql, _| raw_connection.prepare(sql))?;
+    let mut stmt =
+        statement_cache.cached_statement(T::query_id(), source, &Mysql, &[], &mut |sql, _| {
+            raw_connection.prepare(sql)
+        })?;
     let mut bind_collector = RawBytesBindCollector::new();
     source.collect_binds(&mut bind_collector, &mut (), &Mysql)?;
     let binds = bind_collector
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -356,7 +356,7 @@ impl PgConnection {
         let cache_len = self.statement_cache.len();
         let cache = &mut self.statement_cache;
         let conn = &mut self.connection_and_transaction_manager.raw_connection;
-        let query = cache.cached_statement(source, &Pg, &metadata, |sql, _| {
+        let query = cache.cached_statement(T::query_id(), source, &Pg, &metadata, &mut |sql, _| {
             let query_name = if source.is_safe_to_cache_prepared(&Pg)? {
                 Some(format!("__diesel_stmt_{cache_len}"))
             } else {
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -308,9 +308,13 @@ impl SqliteConnection {
     {
         let raw_connection = &self.raw_connection;
         let cache = &mut self.statement_cache;
-        let statement = cache.cached_statement(&source, &Sqlite, &[], |sql, is_cached| {
-            Statement::prepare(raw_connection, sql, is_cached)
-        })?;
+        let statement = cache.cached_statement(
+            T::query_id(),
+            &source,
+            &Sqlite,
+            &[],
+            &mut |sql, is_cached| Statement::prepare(raw_connection, sql, is_cached),
+        )?;
 
         StatementUse::bind(statement, source)
     }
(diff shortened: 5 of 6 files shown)


weiznich:
Thanks for opening this PR :+1:. I think this might be a good idea, but we really need to make sure that we don't regress performance over that. I've triggered a benchmark run on the CI and the results there look fine. See for example https://github.com/diesel-rs/diesel/actions/runs/6097519380/job/16545313665?pr=3778 (The critcmp step has the results). Now that's running on the CI and therefore it's likely noisy. So it might be worth to run the benchmarks again locally and verify these results (maybe even with the instruction count mode?).

Other than that: I also rebased the PR to fix the unrelated CI failures. The remaining CI failure was introduced with the change (essentially clippy being pedantic about things).

diesel/src/connection/statement_cache.rs
```diff
@@ -189,9 +189,27 @@ where
         T: QueryFragment<DB> + QueryId,
         F: FnOnce(&str, PrepareForCache) -> QueryResult<Statement>,
```
weiznich, line 190: I think it should be fine to take a `&mut dyn Fn()` here already. It might be also fine to just change the API of `cached_statement` to what `cached_statement_inner` now looks like. This API is only public behind the "unstable" `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature flag.

weiznich:
I've pushed another commit that fixes the perf based instruction count benchmark metrics. 
You can run these benchmark from the `diesel_bench` folder via
```sh
# once for master
DATABASE_URL=/tmp/test.db cargo +stable bench --no-default-features --features="sqlite instruction_count fast_run" -- --save-baseline master
# for the change, should show changes
DATABASE_URL=/tmp/test.db cargo +stable bench --no-default-features --features="sqlite instruction_count fast_run" -- --baseline master
```
There are features for the other backends as well. `instruction_count` enables the perf based metrics, without that wall time is used. `fast_run` disables certain benchmark sizes.

These metrics show a slight regression:

<details>
<summary> Benchmark results </summary>


```
bench_trivial_query/diesel/1
                        time:   [6198.2000 cycles 6198.2246 cycles 6198.2550 cycles]
                        change: [+1.3840% +1.3970% +1.4146%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  5 (5.00%) high mild
  7 (7.00%) high severe
                        time:   [12299.9480 cycles 12303.6940 cycles 12309.0331 cycles]
                        change: [+0.8329% +0.8780% +0.9180%] (p = 0.00 < 0.05)
Found 21 outliers among 100 measurements (21.00%)
  12 (12.00%) high mild
  9 (9.00%) high severe
                        time:   [25007.8800 cycles 25042.6383 cycles 25057.8212 cycles]
                        change: [+0.5276% +0.6429% +0.6825%] (p = 0.00 < 0.05)
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low severe
  5 (5.00%) high mild
  3 (3.00%) high severe
Benchmarking bench_trivial_query/diesel/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.4s, enable flat sampling, or reduce sample count to 50.
                        time:   [23800830.4051 cycles 23800974.1808 cycles 23801127.5047 cycles]
                        change: [-0.0083% -0.0023% +0.0040%] (p = 0.50 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe
Benchmarking bench_trivial_query/diesel_boxed/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.4s, enable flat sampling, or reduce sample count to 50.
                        time:   [23808058.6217 cycles 23808192.7746 cycles 23808328.0193 cycles]
                        change: [+0.0002% +0.3030% +0.8429%] (p = 0.18 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe
Benchmarking bench_trivial_query/diesel_queryable_by_name/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.3s, enable flat sampling, or reduce sample count to 50.
                        time:   [27274391.3732 cycles 27424965.6166 cycles 27491082.4160 cycles]
                        change: [-1.1312% -0.3410% -0.0696%] (p = 0.18 > 0.05)
                        No change in performance detected.
Found 19 outliers among 100 measurements (19.00%)
  1 (1.00%) low severe
  18 (18.00%) high severe

                        time:   [23138.9929 cycles 23142.9749 cycles 23147.0767 cycles]
                        change: [+1.0488% +1.0920% +1.1364%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild
  2 (2.00%) high severe
                        time:   [38380.0018 cycles 38425.1353 cycles 38470.8968 cycles]
                        change: [+0.9967% +1.1475% +1.2910%] (p = 0.00 < 0.05)
                        time:   [81003.7369 cycles 81011.5802 cycles 81018.9495 cycles]
                        change: [+0.6097% +0.6162% +0.6236%] (p = 0.00 < 0.05)
Benchmarking bench_medium_complex_query/diesel/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.7s, enable flat sampling, or reduce sample count to 50.
                        time:   [22916615.1137 cycles 22984558.8302 cycles 23015564.6610 cycles]
                        change: [-0.4260% -0.1848% -0.0017%] (p = 0.09 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) low severe
  3 (3.00%) high mild
  7 (7.00%) high severe
Benchmarking bench_medium_complex_query/diesel_boxed/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.0s, enable flat sampling, or reduce sample count to 50.
                        time:   [23030128.6204 cycles 23030384.0231 cycles 23030632.6621 cycles]
                        change: [-0.0215% -0.0051% +0.0092%] (p = 0.54 > 0.05)
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) high mild
  9 (9.00%) high severe
Benchmarking bench_medium_complex_query/diesel_queryable_by_name/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.7s, enable flat sampling, or reduce sample count to 50.
                        time:   [27989809.7144 cycles 27990302.2189 cycles 27990831.0977 cycles]
                        change: [-0.3170% -0.2651% -0.2129%] (p = 0.00 < 0.05)
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe

bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
                        time:   [3869630.0091 cycles 3870933.5961 cycles 3872166.2909 cycles]
                        change: [-0.0627% -0.0260% +0.0085%] (p = 0.17 > 0.05)
                        No change in performance detected.
  1 (1.00%) low severe
  1 (1.00%) low mild

bench_insert/diesel/1   time:   [20369.7704 cycles 20371.3994 cycles 20373.5267 cycles]
                        change: [+0.9084% +0.9501% +0.9948%] (p = 0.00 < 0.05)
  2 (2.00%) high mild
bench_insert/diesel/100 time:   [464065.9954 cycles 464140.9285 cycles 464206.4359 cycles]
                        change: [+0.0345% +0.1594% +0.2846%] (p = 0.00 < 0.05)
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
```

</details>

I should probably add that the `bench_trivial_query/diesel/1` really stresses the query builder performance as that one runs with < 500ns per iteration for the sqlite backend. Nevertheless it would be great if we can slightly tweak the change to not regress that benchmark.

diesel/src/connection/statement_cache.rs
smklein: Sounds good, done. I went ahead and collapsed this extra function call (so `cached_statement_inner` no longer exists)

smklein:
Thanks for the tips on running the benchmarks. After collapsing the function `cache_statement` function I'm seeing the following:

```
bench_trivial_query/diesel/1                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
                        time:   [7513.5419 cycles 7513.5760 cycles 7513.6181 cycles]                                                                                                                                                                                                                                                                                                                                                                                                         
                        change: [+0.6409% +0.6560% +0.6727%] (p = 0.00 < 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
Found 14 outliers among 100 measurements (14.00%)                                                                                                                                                                                                                                                                                                                                                                                                                                            
  6 (6.00%) high mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  8 (8.00%) high severe                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
                        time:   [13601.5529 cycles 13606.8259 cycles 13612.3762 cycles]                                                                                                                                                                                                                                                                                                                                                                                                      
                        change: [+0.0768% +0.1424% +0.2082%] (p = 0.00 < 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
  1 (1.00%) high mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                        time:   [45657.2723 cycles 45721.3604 cycles 45783.4704 cycles]                                                                                                                                                                                                                                                                                                                                                                                                      
                        change: [+0.1898% +0.3402% +0.5006%] (p = 0.00 < 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
  2 (2.00%) low mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
                        time:   [25514648.0319 cycles 25514763.6448 cycles 25514877.8544 cycles]                                                                                                                                                                                                                                                                                                                                                                                             
                        change: [-0.0020% -0.0009% +0.0002%] (p = 0.13 > 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
                        No change in performance detected.                                                                                                                                                                                                                                                                                                                                                                                                                                   
  1 (1.00%) low mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
                        time:   [25520424.0499 cycles 25520545.0000 cycles 25520664.7091 cycles]                                                                                                                                                                                                                                                                                                                                                                                             
                        change: [-0.0045% -0.0038% -0.0031%] (p = 0.00 < 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
  3 (3.00%) low mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  1 (1.00%) high mild                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                        time:   [29056280.9355 cycles 29059091.6455 cycles 29062285.6926 cycles]                                                                                                                                                                                                                                                                                                                                                                                             
                        change: [-0.1017% -0.0762% -0.0520%] (p = 0.00 < 0.05)                                                                                                                                                                                                                                                                                                                                                                                                               
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild                                                                     
  10 (10.00%) high severe
                        time:   [26607.8522 cycles 26612.4192 cycles 26616.8591 cycles]
                        change: [+0.7455% +0.8051% +0.8669%] (p = 0.00 < 0.05)
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
                        time:   [41745.1819 cycles 41782.9852 cycles 41821.8198 cycles]
                        change: [+0.2140% +0.3290% +0.4513%] (p = 0.00 < 0.05)
                        time:   [151545.4151 cycles 151661.2229 cycles 151779.2761 cycles]
                        change: [+0.1531% +0.2241% +0.2975%] (p = 0.00 < 0.05)
  1 (1.00%) high mild
                        time:   [26583961.0174 cycles 26586573.8940 cycles 26589168.5516 cycles]
                        change: [+0.0561% +0.0669% +0.0786%] (p = 0.00 < 0.05)
                        time:   [26597067.0640 cycles 26599956.1780 cycles 26602904.5400 cycles]
                        change: [+0.0553% +0.0662% +0.0781%] (p = 0.00 < 0.05)
  3 (3.00%) high mild
                        time:   [31474268.8008 cycles 31477961.2672 cycles 31481803.1718 cycles]
                        change: [+0.0479% +0.0629% +0.0774%] (p = 0.00 < 0.05)
  2 (2.00%) high mild

bench_loading_associations_sequentially/diesel/bench_loading_associations_sequentially
                        time:   [4218133.4062 cycles 4218710.5240 cycles 4219253.4612 cycles]
                        change: [+0.5893% +0.6085% +0.6295%] (p = 0.00 < 0.05)
  1 (1.00%) low severe
  3 (3.00%) low mild

bench_insert/diesel/1   time:   [23336.2704 cycles 23339.1313 cycles 23341.4995 cycles]
                        change: [+0.4308% +0.5060% +0.5516%] (p = 0.00 < 0.05)
bench_insert/diesel/100 time:   [471449.7597 cycles 471555.5618 cycles 471649.6840 cycles]
                        change: [+0.9371% +1.1286% +1.3177%] (p = 0.00 < 0.05)
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) low mild
  3 (3.00%) high severe
```

I think these results make sense -- it's performing a very minor amount of dynamic dispatch, so for a "single query case" I wouldn't expect performance to be better. For a significant number of distinct queries, however, it should reduce code size, which could relieve some pressure from the instruction cache by simply producing less code.

weiznich (approved):
Thanks for the update. I think that's a reasonable change.
I will wait with merging till the next week to give other reviewers from @diesel-rs/reviewers  the chance to look at this as well.

diesel/src/connection/statement_cache.rs
```diff
@@ -313,15 +310,17 @@ where
...
+    pub fn sql(&self, source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<Cow<'_, str>> {
         match *self {
             StatementCacheKey::Type(_) => Self::construct_sql(source, backend).map(Cow::Owned),
             StatementCacheKey::Sql { ref sql, .. } => Ok(Cow::Borrowed(sql)),
         }
     }
 
-    fn construct_sql<T: QueryFragment<DB>>(source: &T, backend: &DB) -> QueryResult<String> {
+    // Note: Intentionally monomorphic over source.
+    fn construct_sql(source: &dyn QueryFragment<DB>, backend: &DB) -> QueryResult<String> {
```
Ten0, line 323: I fear that specifically this one dynamic dispatch here may prevent optimizer from optimizing some specific query fragments into simply functions that return a constant compile-time-computed String...

-> I would be more comfortable if this single part was monomorphized once per statement.

AFAICT that could be done by declaring a new dyn-able trait just here that has this additional construct_sql method inside and is implemented for all QueryFragments (or by adding a construct_sql method to QueryFragment but that seems dirty)

diesel/src/mysql/connection/mod.rs
```diff
@@ -248,8 +248,10 @@ fn prepared_query<'a, T: QueryFragment<Mysql> + QueryId>(
     statement_cache: &'a mut StatementCache<Mysql, Statement>,
     raw_connection: &'a mut RawConnection,
 ) -> QueryResult<MaybeCached<'a, Statement>> {
-    let mut stmt = statement_cache
-        .cached_statement(source, &Mysql, &[], |sql, _| raw_connection.prepare(sql))?;
+    let mut stmt =
+        statement_cache.cached_statement(T::query_id(), source, &Mysql, &[], &mut |sql, _| {
```
Ten0, line 252: in order to avoid having to update all these places and guarantee consistency between source and maybe_type_id everywhere, we can leave the signature of the cached_statement function as it is, and immediately call an inner function that is not generic.
The cached_statement function will always be inlined, so the generated code will be the same as this, but we keep the previous (safer) interface and abstract the "please only monomorphize once" internals.

diesel/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "diesel"
-version = "2.1.0"
+version = "2.1.1"
```
Ten0, line 3: shouldn't that be part of a separate pr/commit?

diesel/src/connection/statement_cache.rs
weiznich: I had the same fear, that's why I've requested to run [the benchmarks here](https://github.com/diesel-rs/diesel/pull/3778#issuecomment-1708901348) and they show essentially no regression after that change, even for trivial sqlite queries (selecting a single user from a table that contains a single user). It's important to note that this seems to only happen after the wrapper function around the monomorphize internal `cached_statement_internal` function was removed.

Ten0: > this seems to only happen after

I'm not sure I understand what the *this* refers to in this sencence

weiznich: "This" refers to the benchmark results linked in the sentence before. I've written that because there is another result [here](https://github.com/diesel-rs/diesel/pull/3778#issuecomment-1708721404), which were produced based on an older version that featured the `_internal` method.

weiznich: I had an in detail look at the generated code based on [this example](https://godbolt.org/z/r8MThc59v).  The short version is that we loss the ability to inline the whole ast building into the function that accepts the `&dyn QueryFragment`, but we don't loss the ability to inline `QueryFragments` to each other. In my opinion in does not really matter to have a single virtual call on top as long as the "heavy" work (ast building) is inlined.

This summary is based on the following observations in the linked `godbolt` snippet: The generated assembly for `impl_queryable` (line 308, line 370) and `dyn_queryable` (line 432) is different. Especially the snippet between setting up the fake `AstPass` and the `println!` is much shorter for `dyn_queryable` and contains only a `call` instruction. Also `dyn_queryable` appears only once while `impl_queryable` appears twice (so once for each query). If we now look for `Queryable` impl for `Select<I>` we only see two of these impls in line 174 and line 216. These impls contain two `moveabs` instructions, the first one in the same between both impls (line 185 and line 227), which indicates that this one corresponds to `SELECT *`, while the second one is different (line 193 and line 235). These correspond to the `movabs` instructions in the `Queryable` impls for `From1` and `From2` (line 268 and line 293), which indicates that these impls where inlined. Combined with the fact that there are only two `Queryable` impls for `Select<I>` I conclude that this means that the `dyn_queryable` impl also uses the inline `QueryFragment` impls (as no other one exists).

Based on these results (and the positive benchmark results) I propose to merge this PR as it is.

diesel/Cargo.toml
weiznich:
```suggestion
version = "2.1.0"
```

diesel/src/connection/statement_cache.rs
Ten0: Ah I was essentially scared about the cost of setting up the entire `QueryBuilder` when in fact we only need the output of its `finish`. (In the godbolt example QueryBuilder is only a String).
But it seems currently that is just at worst an integer increment: https://github.com/diesel-rs/diesel/blob/70420834f3831ce08ddf9005d148329105630163/diesel/src/pg/query_builder/mod.rs#L17-L20 and all of the other types used as QueryBuilder hold only a String, and `finish` always just gives back that contained `String`. As long as that is the case I agree that the single dynamic dispatch probably doesn't really matter.
This is what I had in mind though: #3842

PR #3780, opened 2023-09-05
Configurable migration table

kriswuollett:


weiznich:
To leave a comment here as well:
As [commented on the discussion thread](https://github.com/diesel-rs/diesel/discussions/3771#discussioncomment-6923958): This is the wrong approach. Please follow what was suggested there, otherwise this feature cannot be accepted. (Additionally the suggested approach will likely simplify the effort to implement such feature significantly).

PR #3792, opened 2023-09-14
SQLite: Deserialize database from in-memory buffer

retro-git:
I am creating a project that reads data from an sqlite database. I am simply using sqlite as a storage medium for a large amount of static, readonly data that my application needs to query.

For this type of use-case, it can be advantageous to embed the database file directly into the executable with the include_bytes! macro. Then the byte buffer can be deserialized into an in-memory sqlite database. This approach means the program can be distributed more easily as a single executable.

This PR adds support for doing that, using the "sqlite_deserialize" function from the sqlite C library.


diesel/src/sqlite/connection/mod.rs
```diff
@@ -406,6 +407,41 @@ impl SqliteConnection {
...
+    ///
+    /// # Errors
+    ///
+    /// This function will return `Err` if the deserialization fails.
+    pub fn deserialize_database_from_buffer(&mut self, data: &[u8]) -> QueryResult<()> {
+        let db_ptr = self.raw_connection.internal_connection.as_ptr();
+        let data_ptr = data.as_ptr() as *mut u8;
+        let data_len = data.len() as i64;
+
+        unsafe {
```
weiznich, line 424: This unsafe block should be in `sqlite/connection/raw.rs` instead. So we don't need to allow `unsafe_code` in the main connection module as well.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -406,6 +407,41 @@ impl SqliteConnection {
...
 
+    /// Deserialize an SQLite database from a byte buffer.
+    ///
+    /// This function takes a byte slice and attempts to deserialize it into a SQLite database.
+    /// If successful, the database is loaded into the connection. If the deserialization fails,
+    /// an error is returned.
+    ///
+    /// # Errors
+    ///
+    /// This function will return `Err` if the deserialization fails.
```
weiznich, line 418: I would prefer having a example on this function, even if is "only" a [`no_run` example](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes).

diesel/src/sqlite/connection/mod.rs
```diff
@@ -406,6 +407,41 @@ impl SqliteConnection {
             .register_collation_function(collation_name, collation)
     }
 
+    /// Deserialize an SQLite database from a byte buffer.
+    ///
+    /// This function takes a byte slice and attempts to deserialize it into a SQLite database.
```
weiznich, line 412: The documentation needs to mention that it will open the database in read only mode. It might be even a good idea to change the method name to `deserialize_read_only_database_from_buffer()`

diesel/src/sqlite/connection/mod.rs
```diff
@@ -436,6 +472,36 @@ mod tests {
     use crate::prelude::*;
     use crate::sql_types::Integer;
 
+    #[test]
+    fn database_deserializes_successfully() {
+        let connection = &mut SqliteConnection::establish(":memory:").unwrap();
+        let data = include_bytes!("sqlite_test.db");
```
weiznich, line 478: It would be nice to not include an actual database in the git repository. We could just add support for `sqlite3_serialize` and do a roundtrip here instead (so first serialize then deserialize).

weiznich (changes requested):
Thanks for opening this PR. I'm open to add such a function, but I have some minor suggestions about the documentation and code structure. Also I would like to support the serialize functionality there as well, to keep the functionality symmetrical.

retro-git:
> Thanks for opening this PR. I'm open to add such a function, but I have some minor suggestions about the documentation and code structure. Also I would like to support the serialize functionality there as well, to keep the functionality symmetrical.

Thanks very much for the guidance. I believe I implemented all of your suggestions. However, in the process of implementing serialization (and making the test do a round-trip as suggested), I implemented a RAII SerializedDatabase struct to wrap the allocated serialized database and drop it (the data needs to be cleaned up by calling sqlite3_free). I was unsure where to put this and so I believe that code needs restructuring - if you could take a look again and provide some further help that would be great :)

diesel/src/sqlite/connection/serialized_database.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl SerializedDatabase {
+    /// Creates a new `SerializedDatabase` with the given data pointer and length.
+    pub fn new(data: *mut u8, len: usize) -> Self {
+        Self { data, len }
+    }
+
+    /// Returns a slice of the serialized database.
+    pub fn as_slice(&self) -> &[u8] {
```
weiznich, line 19: I would like to see an additional comment on this unsafe block that the pointer is never null because we don't pass the `NO_COPY` flag.

diesel/src/sqlite/connection/serialized_database.rs
```diff
@@ -0,0 +1,32 @@
+#![allow(unsafe_code)]
+extern crate libsqlite3_sys as ffi;
+
+/// `SerializedDatabase` is a wrapper for a serialized database that is dynamically allocated by calling `sqlite3_serialize`.
+/// This RAII wrapper is necessary to deallocate the memory when it goes out of scope with `sqlite3_free`.
+#[derive(Debug)]
+pub struct SerializedDatabase {
```
weiznich, line 7: I think it might be meaningful to implement `Deref<Target = [u8]>` for this wrapper type so that it can be used as `&[u8]` directly.

weiznich (commented):
Thanks for the update. It now looks mostly good. I would like to see a small comment on one of the unsafe blocks for some invariant that needs to be upheld. The `Deref<Target = [u8]>` impl might be nice, but that's not strictly required for merging.

In addition it might be good to add an entry to the Changlog.md file for this feature.

weiznich (approved):
Thanks :+1:

PR #3643, opened 2023-05-26
Try to hack around a breaking change in diesel 2.1

weiznich:
cc @omid 

That change is not complete yet. It misses:

* [x] A regression test for the now broken variant to verify that it actually works and to prevent breaking it again
* [x] Figuring out to support the `DISTINCT ON (name, id) … ORDER BY name ASC, id DESC` cases.

PR #3793, opened 2023-09-15
Fix #3766

### LINKED ISSUE #3766: Regression in distinct_on: cannot use .nullable() on column
### Versions

- **Rust:** rustc 1.71.1
- **Diesel:** 2.1.1
- **Database:** Postgres
- **Operating System** Linux

## Problem Description

Code using `fn distinct_on` works in Diesel `v2.0.1` but not in `v2.1.1`.

Minimal reproducible example: https://gist.github.com/Ploppz/f8235d83f287bb7926655cd91dfba8fa
`Cargo.toml`: `diesel = { version = "=2.0.1", features = ["postgres", "r2d2"] }`

It compiles in `v2.0.1` but in `v2.1.1` there is the compiler error:
```
error[E0277]: the trait bound `NullableExpression<measurement::columns::id>: Column` is not satisfied
  --> src/main.rs:74:10
   |
74 |         .distinct_on(m::id.nullable())
   |          ^^^^^^^^^^^ the trait `Column` is not implemented for `NullableExpression<measurement::columns::id>`
   |
```

So I tried to remove the `nullable()` so the two lines become:
```rust
        .order_by((m::id, wj::submit_date.nullable().desc()))
        .distinct_on(m::id)
```
But then I get the error:
```
error[E0271]: type mismatch resolving `<Only<table> as AppearsInFromClause<table>>::Count == Never`
  --> src/main.rs:74:10
   |
74 |         .distinct_on(m::id)
   |          ^^^^^^^^^^^ expected `Once`, found `Never`
   |
```
Which makes sense I guess, because the `measurement` table is left-joined so `m::id` is not always present.

## Checklist

weiznich:
This commit fixes #3766 by adding all the missing impls. It uses the same strategy as #3643 for all implementations.

cc @Ploppz Can you double check that it fixes your example as well.

### DIFF
--- a/diesel/src/pg/query_builder/distinct_on.rs
+++ b/diesel/src/pg/query_builder/distinct_on.rs
@@ -37,31 +37,84 @@ where
 }
 
 macro_rules! valid_ordering {
-    // Special-case: rule out the all ident case:
+    // Special-case: for single tuple elements
+    // generate plain impls as well:
+    (
+        @optional_untuple:
+        [generics: $($T: ident)*]
+        [distinct: $D:ident]
+        [order: $O: ty,]
+    ) => {
+        // nothing if both a single tuple elements
+    };
+    (
+        @optional_untuple:
+        [generics: $($T: ident)*]
+        [distinct: $D:ident]
+        [order: $($O: ty,)*]
+    ) => {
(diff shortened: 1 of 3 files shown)


Ploppz:
It works in my project. Thanks for the fix!

Issue #1728, opened 2018-05-23
Support multiple schemas in diesel.toml


compressed:
### Versions

- **Diesel:** 1.3
- **Database:** Postgres

### What are you trying to accomplish?

In 1.3, the `diesel.toml` file has been introduced. In the current form, you can only read tables from one schema. I'd like a way to specify multiple schemas that can all be generated into a single `schema.rs` file.

Perhaps you could have multiple `[print-schema]` sections, each one belonging to a different schema, since you may want to apply different filtering depending on the schema. I think all the existing options are good to have available for each schema you want to load from, except probably `patch_file` since it applies to the whole file.

sgrif:
I'm not sure that multiple `print-schema` sections is the way to go, but I think we need to support this somehow.

compressed:
That sounds better to me because, looking again, `patch_file` and `with_docs` are universal options that would apply to all schemas. It would certainly make for a shorter and easier to understand file. I'm not sure if you'd want schema specific `import_types` (I don't use that feature currently).

sgrif:
I don't think we need `import_types` be schema specific. There's no harm in having the imports used by schema 2 in the tables from schema 1 99% of the time. I'm sure someone out there has a very specific case where this doesn't work, but... that's what `patch_file` is for.

trsh:
I like to keep my structure like /models/table1/model.rs, schema.rs, /models/table2/model.rs,schema.rs. So yes I would also appreciate some scenario, where I can split **Large** single schema.rs in to multiple ones.

tcmal:
Something like `[printschema.whatever]` with the usual properties could work if it inherited from the base `printschema`, I can't think of many other ways to do it & support filtering that isn't kinda messy imo.

huhlig:
Something like the following would likely be beneficial

`
# For documentation on how to configure this file,
# see diesel.rs/guides/configuring-diesel-cli

[print_schema]
include_schemas = ["schema1", "schema2"]

[print_schema.schema1]
file = "src/models/schema1.rs"
filter = { except_tables = ["spatial_ref_sys"] }

[print_schema.schema2]
file = "src/models/schema2.rs"
filter = { except_tables = ["spatial_ref_sys"] }
`

drusellers:
I would for sure love multiple schema support. If there is an approved direction I'd be happy to try and take it on.

weiznich:
[...]

PR #3796, opened 2023-09-17
splitting into multiple schema.rs

### LINKED ISSUE #3799: splitting into multiple schema.rs
Your project helps a lot.  
One feature I want is splitting into multiple schema.rs. 
For instance, making the following diesel.toml

```toml
[[print_schema]]
file = "db_company/src/schema.rs"
filter = { only_tables = ["^companies$"] }
[[print_schema]]
file = "db_user/src/schema.rs"
filter = { only_tables = ["^users$"] }
```

Then, the `companies` schema info is written in `db_company/src/schema.rs`, while the `users` schema info goes to `db_user/src/schema.rs`. 

## Current Behavior

print_schema field only allows only on schema info.  
So, what I do right now is select one print_schema and comment out others like the following files. 

```toml
[print_schema]
file = "db_company/src/schema.rs"
filter = { only_tables = ["^companies$"] }
# [print_schema]
# file = "db_user/src/schema.rs"
# filter = { only_tables = ["^users$"] }
```

## Motivation

- Readability
  - By splitting schema.rs, we can easily search definition of tables. 
- Build Performance
  - This feature will parallelize building schema.rs. which will result in better build performance. 

## How to do 

- make print_schma field in diesel.toml to either single or multiple print_schema 
- edit print_schema.rs to run multiple time.

_Originally posted by @forest1102 in https://github.com/diesel-rs/diesel/discussions/3715_

forest1102:
fixes #3799
I made an additional property `sub_print_schema` in `diesel.toml` to output multiple `schema.rs`. 
Tables included in the output code of `sub_print_schema` is first filtered by `print_schma`.


weiznich:
Thanks for opening this PR.

As written in the linked issue, we are generally interested in a solution to this problem, although I'm not sure if I personally would like to include a new `sub_print_schema` field for that opposed to just allowing several occurrences of `[print-schema]`. Can you explain what's the reasoning why we should want to use the proposed implementation instead?

In addition to the fundamental technical question, this PR would also need to have some tests if we decide to go this way.

forest1102:
There are two main reasons why I choose to add `sub_print_schema`

1. Simplicity of coding
   When making the type of print_schema one or more print_schema, there are lots of stuff to change. For instance, implementing diff-schema will be tough. Instead of doing that, I decided to make diff_schema have only interest in the main print_schema field. 
2. Consistency of CLI parameters
  When running `diesel migration run` with specified targets or other specifiers as a parameter, if allowing multiple print_schema, I wonder how those parameters are applied. When it comes to sub_print_schema, specifiers of tables interested by CLI will be applied to print_schema and print_sub_schema will filter the tables specified by print_schema. By that way, specified parameters of CLI and diesel.toml are not wasted. 

If this fits your thoughts too, I will add some tests.

forest1102:
@Sytten 
Sorry, I don't understand what you are telling. 
Print schema will be used for aggregating all tables used. 
Sub print schema will be used for splitting table schema, and parallelizing compiling.  
What do you mean by the error when using sqlite?

Sytten:
Unsure what your PR does, but if it allows printing multiple database schemas like the issue mentioned, then it should be possible to remove the following limitation in the sqlite inference: https://github.com/diesel-rs/diesel/blob/d4a8a6738fbfe3a00ecb16527db3f507edc2f464/diesel_cli/src/infer_schema_internals/sqlite.rs#L39

forest1102:
@Sytten 
I still don't understand your question.  
This PR is to print out multiple schema.rs files using `sub_print_schema`, and sqlite cannot make multiple schema names.

Sytten:
In your PR description is says `fixes #1728`, this issue is to support multiple database schema in **one** `schema.rs` file so I don't think this closes the issue. In Sqlite you can attach databases which act similar to multiple schemas in other databases. **If** your PR was to support multiple database schema (which again is the issue #1728), then we could likely remove the check I linked above. Since this PR **doesn't address this issue**, you should remove the `fixes #1728` and ignore my comments :)

forest1102:
@Sytten 
Thanks for the suggestion. I created issue and changed addressed issue.

weiznich:
> There are two main reasons why I choose to add sub_print_schema
>
> 1.    Simplicity of coding
    When making the type of print_schema one or more print_schema, there are lots of stuff to change. For instance, implementing a diff-schema will be tough. Instead of doing that, I decided to make diff_schema have only interest in the main print_schema field.
> 2.    Consistency of CLI parameters
    When running diesel migration run with specified targets or other specifiers as a parameter, if allowing multiple print_schema, I wonder how those parameters are applied. Regarding sub_print_schema, specifiers of tables interested by CLI will be used to print_schema and print_sub_schema will filter the tables specified by print_schema. In that way, specified parameters of CLI and diesel.toml are not wasted.

Thanks for writing up your reasoning. I do not really agree on these conclusions. I can follow your argumentation in so far that it's easier to implement it as new feature instead of extending an existing old feature and provide support for all the functionality the existing feature has, but for me that only sounds like getting around of  doing the work for implementing the proper way. As for the CLI flags: Those could appear more than once.

To summarize that: I would prefer having an implementation that uses several real print-schema instances instead of introducing a new striped down feature for this.

forest1102:
Another benefit of sub_print_schema is that searching tables and foreign keys runs only for print_schema and sub_print_schema are generated by reusing those results.

weiznich:
I'm sorry, but this still reads like an unnecessary duplication of features for me. I'm really not that sure that it is the right behavior ro share foreign key definitions between the different schemas.

As for the CLI: Yes, that would likely require new CLI parameters to control the new functionality.

forest1102:
@weiznich 
Thanks for the suggestions.  
Now, I rethought implementations. 
Can you give an opinion on my following specifications to make this feature?   

The New `diesel.toml` format allows a hashmap of a print_schema field with the name of the schema. 
- for print_schema command
  - when print_schema field is just a single print_schema
    - No changes
  - when it is a hashmap of print_schema
    - when no options for table or name specified
      - run each of print_schema in diesel.toml
    - when multiple name parameters and partia print_schema parameter are specified
      - matched name of print_schema will be modified
        e.g. `diesel migration run --name user_schema --only-tables user --name company_schema --only-tables company` will modify user_schema only_tables field and company company_schema field. 
      - no matching print_schema name will return an error. 
    - no name is specified and print_schema parameters are specified
      - all diesel.toml field is dismissed, and run with the specified parameter. 
- for diff_schema command
 - when print_schema field is just a single print_schema
    - No changes
  - when it is a hashmap of print_schema
    - with name parameter
      - run with matched name
    - without name parameter 
      - return error

weiznich:
I'm sorry but I cannot really follow what you are suggesting. Can you maybe try to provide some examples of the old and new (envisioned) `diesel.toml` format and corresponding CLI calls for these examples?

forest1102:
Suppose following diesel.toml  
```:toml
[print_schema.user]
file="user/src/schema.rs"
[print_schema.company]
file="company/src/schema.rs"
filter={only_tables=["companies"]}
```

`diesel migration run --name user --only-tables users` will override user's filter and run command. 

diff_schema requires to set name parameter if print_schema is a hashmap, and the specified name of print_schema will be applied

weiznich:
@forest1102 Thanks for spelling out that proposal. I think I would be fine with such a change, with some minor additions:

* I wouldn't want to use `--name` for it. I think something more specific like `--print-schema-key` or similar would fit better.
* It should also allow to set the actual database schema on each instance. So something like:
```toml
[print-schema.different-thing]
schema = "bar"
file = "src/schema_bar.rs" 
```

should be possible
* What about applying the corresponding CLI action to **all** defined schemas if no specific  `--print-schema-key` flag is given?
(* Likely we should disallow `[print-schema.default]` and keep that one for the existing `[print-schema]` entry?

forest1102:
@weiznich 
Thanks for the suggestion. 
I agree with your idea that `--print-schema-key` is more suitable.  
But, I didn't understand applying the corresponding CLI action when not specifying `--print-schema-key` flag, so can you give me some examples? And what is the meaning of disallowing `[print-schema.default]`? 

> should be possible 
> 
>     * What about applying the corresponding CLI action to **all** defined schemas if no specific  `--print-schema-key` flag is given?
> 
>     * ( Likely we should disallow `[print-schema.default]` and keep that one for the existing `[print-schema]` entry?

weiznich:
> But, I didn't understand applying the corresponding CLI action when not specifying --print-schema-key flag, so can you give me some examples? 

Let's assume the following config:

```toml
[print_schema]
file = "src/schema.rs"
filter={only_tables=["bar"]}

[print_schema.user]
file="user/src/schema.rs"
filter={only_tables=["users"]}

```

Then `diesel print-schema --print-schema-key user` should only perform the `print-schema` command defined for in the `print_schema.user` entry, so it should only change `user/src/schema.rs`. `diesel print-schema` (without any key) should act like you are executing `diesel print-schema --print-schema-key key` for all defined keys. So it should change both schema files. `diesel print-schema --print-schema-key default` should then refer to the `[print-schema]` entry itself, so it should only change the `src/schema.rs` file.

forest1102:
@weiznich 
Thanks for explanation and all of your suggestions. I will implement those features, however I don't know how to parse hashmap with empty key like you wrote as a example.  
```:toml
[print_schema]
file = "src/schema.rs"
filter={only_tables=["bar"]}

[print_schema.user]
file="user/src/schema.rs"
filter={only_tables=["users"]}
```

weiznich:
I would probably use an explicit insert into the HashMap after the parsing step using `default` as key.

forest1102:
@weiznich 
I realized cli is now incompatible by making cli parameters occurred multiple times for multiple print_schema.  
I can't no longer use `index` in arguments of parameters for table-names.  
So, one solution is to delete table-names arguments and make `only-table` have table-names arguments. But in this way, `diesel print_schema user1 user2 --only-table` is no longer valid.  
Do you have any idea to make compatible?

weiznich:
It's unfortunately not acceptable to break the existing CLI interface as that would require a major diesel release and we don't plan to do something like that anytime soon. That written: I'm not sure yet that I really understood what is the actual issue here. Is that something that's disallowed by clap? If so it might be worth to ask in their support channels for alternative ways to represent such a pattern.

forest1102:
@weiznich 

According to [this discussion](https://github.com/clap-rs/clap/discussions/5287#discussioncomment-8049185), I need to use latest version of clap. However, this requires us to update minimum version of rust.  
That's why one of CI fails. 
Is that possible?

forest1102:
@weiznich 
Are my tests sufficient? Do I need any other process to be merged?

diesel_cli/tests/print_schema.rs
```diff
@@ -415,6 +475,34 @@ fn test_print_schema_config(test_name: &str, test_path: &Path, schema: String) {
 
     insta::assert_snapshot!("expected", result);
 }
+fn test_multiple_print_schema_config(test_name: &str, test_path: &Path, schema: String) {
```
weiznich, line 478:
```suggestion

fn test_multiple_print_schema_config(test_name: &str, test_path: &Path, schema: String) {
```

diesel_cli/src/cli.rs
```diff
@@ -289,6 +302,12 @@ pub fn build_cli() -> Command {
                 .action(clap::ArgAction::Append)
                 .number_of_values(1)
                 .help("A list of derives to implement for every automatically generated SqlType in the schema, separated by commas."),
+        )
+        .arg(
+            Arg::new("print-schema-key")
+                .long("print-schema-key")
+                .action(ArgAction::Append)
+                .help("select schema key from diesel.toml, use 'default' for print_schema without key."),
```
weiznich, line 310: Can we default to the default schema key if that's not set. If that's already the case maybe indicate that in the help.

diesel_cli/src/config.rs
```diff
@@ -53,23 +76,325 @@ impl Config {
...
+
+            if matches.get_flag("generate-custom-type-definitions") {
+                config.generate_missing_sql_type_definitions = Some(false);
+            }
+
+            if let Some(derives) = matches.get_many("custom-type-derives") {
+                let derives = derives.cloned().collect();
+                config.custom_type_derives = Some(derives);
+            }
+            dbg!(config.clone());
```
weiznich, line 340:
```suggestion
```

Or replace it with an explicit log statement (`tracing::debug` or similar)

diesel_cli/src/errors.rs
```diff
@@ -58,4 +58,8 @@ pub enum Error {
     ColumnLiteralParseError(syn::Error),
     #[error("Failed to parse database url: {0}")]
     UrlParsingError(#[from] url::ParseError),
+    #[error("Failed to parse CLI parameter: {0}")]
+    ClapMatchesError(#[from] clap::parser::MatchesError),
+    #[error("No print schema key with the name `{0}` exists")]
```
weiznich, line 63: Can we point out more explicitly that this refers to `[print_schema.{0}]` entries in your `diesel.toml` file?

diesel_cli/src/migrations/mod.rs
```diff
@@ -93,7 +93,16 @@ pub(super) fn run_migration_command(matches: &ArgMatches) -> Result<(), crate::e
             {
                 let config = Config::read(matches)?;
                 let diff_schema = if diff_schema == "NOT_SET" {
-                    config.print_schema.file.clone()
+                    if config.print_schema.all_configs.len() != 1 {
+                        return Err(crate::errors::Error::UnsupportedFeature(
```
weiznich, line 97: This should default to the default schema if no other schema key is set explicity

diesel_cli/src/migrations/diff_schema.rs
```diff
@@ -89,7 +86,19 @@ pub fn generate_sql_based_on_diff_schema(
...
-
+    let table_names = load_table_names(&mut conn, None)?;
+    let tables_from_database = filter_table_names(
+        table_names.clone(),
+        &config
+            .print_schema
+            .all_configs
+            .first_key_value()
+            .ok_or(errors::Error::UnsupportedFeature(
+                "select exact one print schema key".to_string(),
```
weiznich, line 97: This should default to the default schema if no `print-schema-key` is set explicitly.

diesel_cli/src/main.rs
```diff
@@ -236,57 +236,15 @@ fn run_infer_schema(matches: &ArgMatches) -> Result<(), crate::errors::Error> {
     use crate::print_schema::*;
 
     tracing::info!("Infer schema");
-
     let mut conn = InferConnection::from_matches(matches)?;
-    let mut config = Config::read(matches)?.set_filter(matches)?.print_schema;
-
-    if let Some(schema_name) = matches.get_one::<String>("schema") {
```
weiznich, line 243: These configuration overwrites are still used in the new code?

diesel_cli/src/cli.rs
```diff
@@ -289,6 +302,12 @@ pub fn build_cli() -> Command {
                 .action(clap::ArgAction::Append)
                 .number_of_values(1)
                 .help("A list of derives to implement for every automatically generated SqlType in the schema, separated by commas."),
+        )
+        .arg(
+            Arg::new("print-schema-key")
```
weiznich, line 307: I wonder if just `schema-key` might be the better name here?

weiznich (commented):
Thank for the update. This looks now much more like what I had in mind :+1:

I've left a few more or less stylistic comments and some questions that I would like to see answered. Other than that I would like to see at least one test for the generate schema part as well.

Finally it would be great if you can open a PR for the web-page as well that documents this new configuration file extension there as well: https://github.com/sgrif/diesel.rs-website/blob/master/src/guides/configuring-diesel-cli.md

weiznich:
(CI failures are caused by the recent rust release, I'm working on that)

diesel_cli/src/main.rs
forest1102: Yes, those are applied by `set_filter` and `update_config`

diesel_cli/src/cli.rs
forest1102: Thanks, that would be better.

forest1102:
@weiznich 
I am working on the merge conflict, so I will re-request review when it is done

weiznich (approved):

hxk1633:
Is there any documentation/examples that explain these changes? I've been trying to figure out a way to handle multiple schemas in my diesel code and this looks like what I might need,

weiznich:
@hxk1633 See the [nested sub-schema section in the relevant guide](https://diesel.rs/guides/configuring-diesel-cli.html)

hxk1633:
I'm not sure if this feature is the right tool for what I'm doing. I have created a new schema in my migrations called "v2" and I have an old schema (public) that I need to query with diesel. My understanding is that this only works with schemas generated from the CLI command to generate migrations. This is my diesel.toml file:

hxk1633:
I'm not sure if this feature is the right tool for what I'm doing. I have created a new schema in my migration files called "v2" and I have an old default schema (public) that I need to query with diesel.  These are the diesel.toml and schema files.

```
# For documentation on how to configure this file,
# see https://diesel.rs/guides/configuring-diesel-cli

[print_schema]
file = "src/diesel_schema/public.rs"
custom_type_derives = ["diesel::query_builder::QueryId"]

[print_schema.v2]
file = "src/diesel_schema/v2.rs"
custom_type_derives = ["diesel::query_builder::QueryId"]

[migrations_directory]
dir = "migrations"
```

`src/diesel_schema/v2.rs`
```
// @generated automatically by Diesel CLI.

pub mod v2 {
    pub mod sql_types {
        #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
        #[diesel(postgres_type(name = "device_status", schema = "v2"))]
        pub struct DeviceStatus;
    }

    diesel::table! {
        v2.account (id) {
           ....
        }
    }

    diesel::table! {
        v2.contact (id) {
         ...
        }
    }

    diesel::table! {
        v2.device (id) {
             ....
        }
    }

    diesel::joinable!(device -> account (account_id));

        account,
        contact,
        device
    );
}
```

`src/diesel_schema/public.rs`
```
pub mod public {
    pub mod sql_types {
        #[derive(diesel::query_builder::QueryId, diesel::sql_types::SqlType)]
        #[diesel(postgres_type(name = "device_status", schema = "public"))]
        pub struct DeviceStatus;
    }

    diesel::table! {
        public.account (id) {
           ....
        }
    }

    diesel::table! {
        public.contact (id) {
         ...
        }
    }

    diesel::table! {
        public.device (id) {
             ....
        }
    }

    diesel::joinable!(device -> account (account_id));

        account,
        contact,
        device
    );
}
```

I'm not sure if the schema names "public" and "v2" have to be included in the schema files. Is the schema name after "print_schema" in diesel.toml enough for me to start querying from both public and v2 schema in my code?

weiznich:
@hxk1633 I won't provide any further support on an old closed PR, as that's clearly the wrong place to answer support questions. In fact I should have not responded to the first question already. 
If you want to ask questions I encourage you to open a new discussion thread with details about your problem.

PR #3812, opened 2023-09-29
Use `#[diagnostic::on_unimplemented]`

weiznich:
This PR changes our usage of `#[rustc_on_unimplemented]` to use the
`#[diagnostic::on_unimplemented]` attribute instead. The former
attribute is considered to be rustc internal and permanently unstable,
while the later one is hopefully stabilised at some point. This PR also
adds another instance of such attributes to the definition of the
`CompatibleType` trait to point users to `Selectable` + `#[diesel(check_for_backend(…))]`


AlexStorm1313 (commented):
If I'm looking at this correctly the new attribute is functionally the same, but is specifically created for use outside the internal rust library?

weiznich:
Yes, thats a good summary. It even calls the same code internally in rustc.

AlexStorm1313 (approved):

PR #3816, opened 2023-10-05
Add clippy::pedantic for star import

DannyGoldberg:
To fix:

```
warning: usage of wildcard import
   --> src/models/schema_gen.rs:174:9
    |
174 |     use diesel::sql_types::*;
    |         ^^^^^^^^^^^^^^^^^^^^ help: try: `diesel::sql_types::{Nullable, Text, Timestamptz, Uuid, Varchar}`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports
```

For now, I added a `patch_file` to my `diesel.toml` configuration, but I figured it'd be nicer for all if it's added automatically.

Or perhaps there's a better way that a reviewer will recommend :)

### DIFF
--- a/diesel_cli/src/print_schema.rs
+++ b/diesel_cli/src/print_schema.rs
@@ -555,6 +555,7 @@ impl<'a> Display for TableDefinition<'a> {
                 {
                     if already_imported_custom_types.insert(&ct.rust_name) {
                         if !has_written_import {
+                            writeln!(out, "#[allow(clippy::pedantic)]")?;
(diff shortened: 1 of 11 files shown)


weiznich:
Thanks for opening this PR.

Rather than adding this `#[allow()]` to each generated instance of the `table!` macro, I would go and allow it here: https://github.com/diesel-rs/diesel/blob/0377c258db72a31058f93d6c16c5175f5f08e139/diesel_derives/src/table.rs#L184 instead (so in the macro itself)

PR #3833, opened 2023-10-18
Try out cargo-dist

weiznich:


Gankra:
Looking into the windows issue with https://github.com/axodotdev/cargo-dist/issues/496

PR #3849, opened 2023-11-10
Track PR metrics with Bencher

epompeii:
This PR is a follow up from a discussion with @weiznich on adding the nightly `master` metrics to Bencher: https://github.com/diesel-rs/diesel/discussions/3848

A separate Testbed has been created for all three targets in order to distinguish their results (ie `ubuntu-latest-postgres`, `ubuntu-latest-sqlite`, and `ubuntu-latest-mysql`).
Currently only the `sqlite` metrics are being ingested by Bencher, [due to build failures in the other two](https://github.com/diesel-rs/diesel/actions/workflows/metrics.yml). Please, let me know if you would like for me to try to fix those in a follow up PR.

The results can be seen here: https://bencher.dev/perf/diesel
Note that it takes a hot second for all 99 results to be returned.
Those results are from a test run on `epompeii/diesel`: https://github.com/epompeii/diesel/actions/runs/6826611755/job/18566903071#step:10:3477

In order for this to work on `diesel-rs/diesel` a maintainer will need to add the `BENCHER_API_TOKEN` as a repository secret. I could either send over the key I setup or add one of the maintainers of `diesel` to the project on Bencher and you can use your own API token (preferred solution).


weiznich:
Thanks for opening this PR. For now I would like to use the `bench` workflow (that is used to check for regressions in PR's) as experimentation workflow as it contains much less benchmarks and might benefit more from getting comments on PR's and similar improvements. 

As a more general note for bencher itself: It would be great if the frontend could group benchmarks from the same group somewhat. We likely do not want to see all benchmarks for all sizes at the same time. Either we would be interested in seeing all `insert` benchmarks of a certain size for the different crates, or we would be interested in seeing all sizes for one specifc crate.

epompeii:
> For now I would like to use the bench workflow (that is used to check for regressions in PR's) 

Oh, my misunderstanding. I'm sorry! I'll make the change to the `bench` workflow.

> It would be great if the frontend could group benchmarks from the same group somewhat.

I totally agree! My current thinking is split between two ways of doing this:

1. Create a way to add tag particular dimensions (branch, testbed, benchmark). Then you could filter benchmarks based on a particular tag, say to view all `insert` benchmarks you would tag each `insert` benchmark with `operation: insert`etc. (Tracking issue: https://github.com/bencherdev/bencher/issues/240)
2. Add a way to search each dimension (branch, testbed, benchmark) in the perf plot view and then allow you to save/star particular plots (https://github.com/bencherdev/bencher/issues/231) and even view them in a dashboard like view (https://github.com/bencherdev/bencher/issues/232). So you could have an `insert` benchmarks plot and `trivial_query` plot, etc. Currently you have to do this with browser bookmarks and multiple windows.

I think either way, I will end up doing 2, so the question then becomes, would it be worth it to also have 1?
I'm very much open to suggestions here if you have any 😃

epompeii:
Okay, I have moved things over to `bench` workflow!
You can see an example PR where this branch is serving as the target branch: https://github.com/epompeii/diesel/pull/1

There are three comments on the PR, one for each testbed: `ubuntu-latest-sqlite`, `ubuntu-latest-mysql`, and `ubuntu-latest-postgres`.

I have an option to only start commenting once there is [an Alert is generated](https://bencher.dev/docs/explanation/thresholds) (using the `--ci-only-on-alert` flag). I didn't enable it so you could see the output, but let me know if you want that in the final version.

Currently the "tag to run benchmarks" mechanism is in place, but it is not meant as a security feature. Once the tag is added any subsequent revisions with that tag will also be benchmarked. I can setup GitHub Actions to require a reviewer before each run if you would prefer that. However, we way that I have things set up now _should_ be secure from pwn requests.

weiznich:
I just want to leave a comment here that this PR is not forgotten. I'm away for a few days. I try to have a closer look when I'm back.

weiznich:
Thanks for updating the PR to work with the comparing results on PR's. That look helpful.

There are a few things I would like clarify before moving forward:

* Current the summaries posted to the PR do not contain any indication by how much the performance changed. It would be really helpful to have this information in the PR comments.
* If I click on any of the benchmark result links I'm asked for a login. All the information there needs to be public, as we cannot expect from contributors to have an account there.

epompeii:
I've been doing a lot of thinking lately about the security implications of running benchmarks from fork PRs.

Was the original intention behind the "run on tag" semantics meant as a security feature (requiring review) or simply as a speed of CI optimization (on run benchmarks on PRs where it would matter)?

If the intention of the "run on tag" was meant as a security feature, I think we should move over to creating a couple of GitHub Environments and add Required Reviewers for fork PRs. I have a write up on this option here: https://bencher.dev/docs/how-to/github-actions/#benchmark-fork-pr-from-target-branch-with-required-reviewers

If the "run on tag" was meant more as a speed of CI optimization, I can work that into this much safer option: https://bencher.dev/docs/how-to/github-actions/#benchmark-fork-pr-and-upload-from-default-branch

Please, let me know what you think, and I will update this PR accordingly!

weiznich:
Thanks for the updates. It's good to see what it would look like. 

You mention that this approach now reuses "old" builds for the master branch as optimization. That might become problematic, as it seems like there is really a lot of noise between running benchmark jobs on different action runners. The old approach did run both the current PR and the master PR after each other on the same machine. That would enforce the all benchmarks have a similar environment. It reduces the amount of noise drastically in my opinion, which is the thing that it makes helpful for judging the performance impact of certain changes. I'm not sure that this behavior is still there with the new approach? Maybe you can clarify that? 

> Was the original intention behind the "run on tag" semantics meant as a security feature (requiring review) or simply as a speed of CI optimization (on run benchmarks on PRs where it would matter)?

The original reasoning behind the tag based solution is mostly that it is a speed optimization, as everything is "local" to the PR. The worst thing that could happened there was that someone submitted a PR that wast CI time. There are no results that would have been stored for a longer time, so it couldn't influence anything else.

epompeii:
> You mention that this approach now reuses "old" builds for the master branch as optimization. That might become problematic, as it seems like there is really a lot of noise between running benchmark jobs on different action runners. The old approach did run both the current PR and the master PR after each other on the same machine. That would enforce the all benchmarks have a similar environment. It reduces the amount of noise drastically in my opinion, which is the thing that it makes helpful for judging the performance impact of certain changes. I'm not sure that this behavior is still there with the new approach? Maybe you can clarify that?

I do need to clarify 😃
The main motivation behind splitting out the benchmarking of the `master` branch and the PR branches is to keep things more statistically sound (**not** as an optimization). That is, Bencher keeps track of the benchmark results for a _particular_ branch over time. Running the `master` branch every time there is a PR would lead to irregular over sampling of the `master` branch.

For example, under the old system if we had 3 PRs against `master` for `HEAD#1` and then later only 1 PR against `HEAD#2` we would have 4 samples for `HEAD#1` and only 2 samples for `HEAD#2` (1 initial branch `push` run + each PR run). This means we would be statistically over weighting `HEAD#1` in our calculations.

The statistical thresholds used by Bencher are what are meant to help deal with this noise, and we only want a single sample run against any branch per change. So yes,  being able to judge the performance impact of certain changes is still here in the new approach! It is relying on the statistics, powered by the data of previous runs.

The issue with relying on only a single run per version is you can't do such statistical analysis, as `N=1`. Even within a single run, there can be very high volatility between when one benchmark suite is run and another is run. I've found the statistical approach be more reliable than single shot relative benchmarking using wall clock time. 

> This has the added benefit of halving the PR benchmarking times.

This is very much a positive externality/benefit and not at all the motivation behind the change. I'm sorry for any confusion!

> The original reasoning behind the tag based solution is mostly that it is a speed optimization, as everything is "local" to the PR. The worst thing that could happened there was that someone submitted a PR that wast CI time. There are no results that would have been stored for a longer time, so it couldn't influence anything else.

Okay, that makes sense. That has me leaning towards adding "run on tag" to this much safer option: https://bencher.dev/docs/how-to/github-actions/#benchmark-fork-pr-and-upload-from-default-branch
This would mean you only have to tag a PR once as needing to be benchmarked and every subsequent run will automatically be safely benchmarked without needing any additional approval.

epompeii:
I have updated this PR to now use the `workflow_run` + "run on tag" setup.
In order for `workflow_run` to actually run, it needs to exist on the _base_ branch for the repository.

Here is a new example PR doing just that: https://github.com/epompeii/diesel/pull/3

Let me know what you think!

weiznich:
Thanks for the update.

> The issue with relying on only a single run per version is you can't do such statistical analysis, as N=1. Even within a single run, there can be very high volatility between when one benchmark suite is run and another is run. I've found the statistical approach be more reliable than single shot relative benchmarking using wall clock time.

Well, that true if we would have done a single run only, but we use `criterion` there, which on it's own performs quite a lot of stuff to ensure that we get statistically relevant results by essentially running a certain benchmark several times. I think for some of our benchmarks `criterion` performs [up to several millions individual runs](https://github.com/diesel-rs/diesel/actions/runs/7222127467/job/19678516655#step:8:146).

> For example, under the old system if we had 3 PRs against master for HEAD#1 and then later only 1 PR against HEAD#2 we would have 4 samples for HEAD#1 and only 2 samples for HEAD#2 (1 initial branch push run + each PR run). This means we would be statistically over weighting HEAD#1 in our calculations.

My main issue with this approach is that the old benchmark on HEAD#1 might have run on a completely different runner with different setup, etc. The [metrics data](https://github.com/diesel-rs/metrics/) indicate that there is quite a lot of difference between these different runs. I totally understand that you can do some statistics there to get a threshold that says: Anything above/below that is a change in the runtime, but my fear is that this threshold is really large. That would render such a system much less useful for me. 
My reasoning with the current approach is essentially: Both benchmarks run under very similar conditions (as the run after each other). That  removes one potential source of noise and allows us to reduce the threshold for consider something as "change" drastically. The statistic for that is provided by criterion.

epompeii:
> My reasoning with the current approach is essentially: Both benchmarks run under very similar conditions (as the run after each other). That removes one potential source of noise and allows us to reduce the threshold for consider something as "change" drastically. The statistic for that is provided by criterion.

Sounds good! I can move things over to using Relative Benchmarking then: https://bencher.dev/docs/how-to/track-benchmarks/#relative-benchmarking

epompeii:
Okay, I've moved things back to using relative benchmarking.
The PR branch and the diesel `master` branch are benchmarked one after the other, within the same job.

For security reasons, the results are uploaded as artifacts. Then in a separate `workflow_run` job they are stored in Bencher and the PR comment is generated.

You can see the results in this example pull request: https://github.com/epompeii/diesel/pull/3

weiznich:
Thanks for the update. I wont have the capacity to review this PR before my Christmas holidays, I will try to do this in the new year.

.github/workflows/run_benches.yml
```diff
@@ -47,42 +46,32 @@ jobs:
       - name: Install rust toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Install critcmp
```
weiznich, line 50: I would prefer to keep critcmp as second option for now.

.github/workflows/track_benches.yml
```diff
@@ -0,0 +1,95 @@
...
+            core.exportVariable("PR_HEAD", `${githubEvent.pull_request.head.ref}-${githubEvent.pull_request.head.sha.slice(0, 8)}`);
+            core.exportVariable("PR_ID", `${githubEvent.pull_request.head.ref}/${process.env.BENCHER_TESTBED}/${process.env.BENCHER_ADAPTER}`);
+            core.exportVariable("PR_NUMBER", githubEvent.number);
+      - uses: bencherdev/bencher@main
+      - name: Track base Benchmarks
+        run: |
+          bencher run \
+          --if-branch '${{ env.PR_HEAD }}' \
+          --else-branch \
+          --token "${{ secrets.BENCHER_API_TOKEN }}" \
```
weiznich, line 74: As a note to myself: We should to set our own API token here before merging, right?

.github/workflows/track_benches.yml
```diff
@@ -0,0 +1,95 @@
...
+          --file "$BASE_BENCHMARK_RESULTS"
+      - name: Create PR threshold
+        run: |
+          bencher threshold create \
+          --project "$BENCHER_PROJECT" \
+          --branch '${{ env.PR_HEAD }}' \
+          --testbed "$BENCHER_TESTBED" \
+          --measure latency \
+          --test t \
+          --upper-boundary 0.98 \
```
weiznich, line 84: That's essentially the p value for the test to reject the PR if the performance changes are to large. (So if it's outside 0.98 of the assumed distribution, right). Might be worth to add a short comment there just in case we want to touch that value later on.

weiznich (commented):
Thanks for continue to work on this. It look nearly good now. I left some minor comments about the github action definition. The results + comments look ok for me, although I personally would prefer if the table posted in the comment to the PR would contain the old (main branch) and new (PR branch) timing in an explicit manner + the absolute difference in a third column, but that's a shouldn't block merging this PR.

.github/workflows/track_benches.yml
epompeii: Yes! Add `BENCHER_API_TOKEN` as a Repository secret. (ex: Repo -> Settings -> Secrets and variables -> Actions -> New repository secret) 
I have already started using the `diesel` project on Bencher, so I can just invite you to the project there, or you can make an entirely new project. If you choose the latter then we can do a slug-swap if you want to have the `/perf/diesel` URL going forward.

.github/workflows/track_benches.yml
epompeii: Yes, and great point!
I pulled this out to an env var and added a comment to make it clearer 😃

.github/workflows/track_benches.yml
```diff
@@ -0,0 +1,100 @@
...
+      BENCHER_PROJECT: diesel
+      BENCHER_ADAPTER: rust_criterion
+      BENCHER_TESTBED: ubuntu-latest-${{ matrix.backend }}
+      PR_BENCHMARK_RESULTS: pr_${{ matrix.backend }}.txt
+      BASE_BENCHMARK_RESULTS: base_${{ matrix.backend }}.txt
+      GITHUB_EVENT: event_${{ matrix.backend }}.json
+      # This is the confidence interval for the t-test Threshold
+      # Adjust this value to lower to make the test more sensitive to changes
+      # Adjust this value to higher to make the test less sensitive to changes
+      # https://bencher.dev/docs/explanation/thresholds/#t-test-threshold-upper-boundary
```
weiznich, line 27: Unrelated note: I like that you presented the underlying statistics there in detail. That's great :heart:

.github/workflows/track_benches.yml
weiznich: If it's possible to remove you as owner from the existing project after I joined I would like to go that route.

.github/workflows/run_benches.yml
epompeii: Sounds great!
I've added `critcmp` back and it is now running in tandem with Bencher 😃
Example run here: https://github.com/epompeii/diesel/actions/runs/7767012585#summary-21183420265

.github/workflows/track_benches.yml
epompeii: Yes, it is! I sent you an invite, so you should see that in you inbox. Once you accept, I can just remove myself from the existing project.

epompeii:
Thanks for all the great feedback! 😃
Some reviewers prefer to be the one to hit `resolve` on their comments, so I have held off doing so.
With that said though, I think the only thing left is getting you added to the project!

weiznich (approved):

weiznich:
(I resolved the merge conflict, CI builds will likely fail due to the recent rustc API, will take care of that later)

epompeii:
@weiznich I just wanted to check in. What can I do to help get this PR across the finish line?

weiznich:
There is not much that can currently be done as this is blocked on passing CI, which in turn is blocked on fixing https://github.com/rust-lang/rust/issues/120830 which is essentially just needs some more time to get the PR with the fixed merged there.

PR #3854, opened 2023-11-17
Fix impl SqlOrd postgres > postgres_backend feature flag.

brocaar:
First of all, thanks for your excellent work!

I believe the `SqlOrd` should be implemented if the `postgres_backend` is enabled, not `postgres`. Else the `SqlOrd` trait would not be implemented for these types if using `diesel-asyc`.


JohnTitor (approved):
Seems reasonable 👍

brocaar:
@weiznich the last Diesel release has been quite some time ago and does not include this fix. Would it be possible to create a new release which includes this fix? Thanks in advance!

weiznich:
@brocaar The next feature release is still blocked on https://github.com/diesel-rs/diesel/issues/3745. This fix might be something for a patch release, but it would need someone to backport this and other relevant fixes to the 2.1.x branch.

PR #3862, opened 2023-11-29
add skip_insertion attribute for Insertable derive macro

### LINKED ISSUE #860: Allow to ignore field os struct
Actually I have this struct

```rust
#[derive(Clone, Debug, Queryable, Serialize, AsChangeset, Identifiable, Associations)]
#[has_many(authors)]
pub struct User {
	pub id: i32,
	pub name: String,
	pub email: String,
	#[serde(skip_serializing)]
	pub password: String,
	pub created_at: DateTime<UTC>,
	pub author: Option<Author>
}
```

The column `author` is not in the database, but when I return the model, I would like to put inside of the struct the `Author` (a user can or not to have an author). For that I want to tell diesel that ignores the field `author`. I'm searching if something like `#[ignore_column]`.

```rust
	#[ignore_column]
	pub author: Option<Author>
```

formlogic-kirk:
Address a subset of #860, specifically related to the `Insertable` derive macro. It's currently fairly painful to use `Insertable` for tables with generated columns, this pr allows fields to be skipped during insertion.

### DIFF
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -45,6 +45,10 @@ fn derive_into_single_table(
     let mut ref_field_assign = Vec::with_capacity(model.fields().len());
 
     for field in model.fields() {
+        // skip this field while generating the insertion
+        if let true = field.skip_insertion() {
+            continue;
+        }
         // Use field-level attr. with fallback to the struct-level one.
         let treat_none_as_default_value = match &field.treat_none_as_default_value {
             Some(attr) => {
--- a/diesel_derives/src/insertable.rs
+++ b/diesel_derives/src/insertable.rs
@@ -46,7 +46,7 @@ fn derive_into_single_table(
 
     for field in model.fields() {
         // skip this field while generating the insertion
-        if let true = field.skip_insertion() {
+        if field.skip_insertion() {
             continue;
         }
         // Use field-level attr. with fallback to the struct-level one.
--- a/diesel_tests/tests/insert.rs
+++ b/diesel_tests/tests/insert.rs
@@ -469,7 +469,7 @@ fn insert_with_generated_column() {
         "CREATE TABLE user_with_last_names (
         first_name VARCHAR NOT NULL PRIMARY KEY,
         last_name VARCHAR NOT NULL,
-        full_NAME VARCHAR GENERATED ALWAYS AS (first_name || ' ' || last_name) STORED
+        full_name VARCHAR GENERATED ALWAYS AS (first_name || ' ' || last_name) STORED
     )",
     )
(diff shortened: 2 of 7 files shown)


diesel_derives/src/insertable.rs
```diff
@@ -45,6 +45,10 @@ fn derive_into_single_table(
     let mut ref_field_assign = Vec::with_capacity(model.fields().len());
 
     for field in model.fields() {
+        // skip this field while generating the insertion
+        if let true = field.skip_insertion() {
```
Ten0, line 49:
```suggestion
        if field.skip_insertion() {
```

diesel_tests/tests/insert.rs
```diff
@@ -452,6 +452,47 @@ fn insert_returning_count_returns_number_of_rows_inserted() {
...
+        last_name VARCHAR NOT NULL,
+        full_NAME VARCHAR GENERATED ALWAYS AS (first_name || ' ' || last_name) STORED
+    )",
+    )
+    .execute(connection)
+    .unwrap();
+    let new_users: &[_] = &[UserWithLastName {
+        first_name: "Sean".to_string(),
+        last_name: "Black".to_string(),
+        ..Default::default()
```
Ten0, line 482: Looks like a usage of this feature would always end up containing a part that's going to look like this where you have dummy values...
What happens with the ID though? It doesn't have the attribute so it looks like you are somewhat-implicitly setting it to zero?
And that looks encouraged by this feature...

formlogic-kirk: I wasn't intending to show anything related to the id/primary key field here, so I'll remove that from the test as it's not useful. I really just want to be able to insert into tables with generated columns without having to convert into some intermediate type first.

weiznich (commented):
Thanks for opening this PR. I think conceptually these changes look fine.

I somewhat agree with @Ten0  that it might make it easier to accidentally insert default values. We probably should not use such example in the documentation/tests.

That brings me to the point of things missing from this PR that needs to be fixed before merging:

* The documentation needs to be updated: https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel_derives/src/lib.rs#L324 (It needs to mention that attribute)
* This requires an entry in the `Changelog.md` file

formlogic-kirk:
> I somewhat agree with @Ten0 that it might make it easier to accidentally insert default values. We probably should not use such example in the documentation/tests.

I replaced it with a string indicating that the value will not be used, but I'm not sure how else to handle it. I could try making it an `Option<String>` if you think that would be clearer maybe?

weiznich (approved):

PR #3873, opened 2023-12-11
Add support for syn 2.0.40's new Expr::Group to dsl::auto_type

Ten0:
This expression may get yielded when parsing `$a: expr` from other macros.

### DIFF
--- a/dsl_auto_type/src/auto_type/expression_type_inference.rs
+++ b/dsl_auto_type/src/auto_type/expression_type_inference.rs
@@ -82,6 +82,9 @@ impl TypeInferrer<'_> {
             expr,
             type_hint.filter(|h| !matches!(h, syn::Type::Infer(_))),
         ) {
+            (syn::Expr::Group(syn::ExprGroup { expr, .. }), type_hint) => {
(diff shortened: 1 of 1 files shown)


weiznich (approved):
The change itself looks good 👍

I would prefer to have a some sort of test somewhere that verifies that the code compiles. 

Also I would like to use the possibility to point again to #3783 for the PR that fixes most of the type aliases to work with auto_type.

Ten0:
There aren't tests for `#[dsl::auto_type]` besides the doctests so far and that doesn't feel worth putting in the doc, and unfortunately this slipped out of my mind and I can't remember the precise specifics of how to reproduce.
But that did fix my issue in a macro and matches the syn documentation so I'm confident enough that this is correct and we're unlikely to break it in the future.
I may add a test for this in the test file from the other PR but for now I think it would be unreasonable to dedicate that amount of investigation/merge time for such a small thing.

weiznich:
That sounds ok. I always forget that the https://github.com/diesel-rs/diesel/pull/3783 is not merged yet as that PR adds a bunch of "tests".

PR #3895, opened 2023-12-30
Fix BoxedExpression for having

obsoleszenz:


weiznich:
Thanks for opening this PR.

I've pushed https://github.com/diesel-rs/diesel/pull/3895/commits/2aa169dd5abf3ac2792503a7a0ae9940fcc16103 which basically applied the fix to another location + slightly modifies the test. After that everything compiles fine.

Ten0 (approved):

PR #3897, opened 2024-01-05
Refactor error handling in diesel-cli

weiznich:
This commit refactors the error handling in diesel-cli to use a central error type and to use errors wherever meaningful (in contrast to just panicing). This hopefully improves the quality of the error messages emitted by diesel-cli

In addition to the error handling changes this commit also includes some log statements for:

* Complex functionality
* Any query executed

This hopefully simplifies debugging any potential issues in the CLI tool more easily.


diesel_cli/src/database.rs
```diff
@@ -134,38 +136,63 @@ impl InferConnection {
...
+            url: database_url,
+        })?;
+
+        conn.set_instrumentation(|event: InstrumentationEvent<'_>| {
+            if let InstrumentationEvent::FinishQuery { query, error, .. } = event {
+                if let Some(err) = error {
+                    tracing::error!(?query, ?err, "Failed to execute query");
+                } else {
+                    tracing::debug!(?query);
+                }
```
chusteven, line 150: Should this be opt in? Sometimes queries can have sensitive data, so I wonder if clients using `diesel` should be able to configure the raw SQL being logged or not.

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -33,12 +32,12 @@ pub fn determine_column_type(
     // Postgres doesn't coerce varchar[] to text[] so print out a message to inform
     // the user.
     if let (true, Some(tpe)) = (is_array, diesel_alias_without_postgres_coercion) {
-        writeln!(
-            &mut stderr(),
+        tracing::info!("Cannot coerce varchar[] into text[]");
+        eprintln!(
```
chusteven, line 36: Just curious under what conditions this returns either an `Ok` or an `Err`? I see the `?` in the deleted code, and tried reading up on https://doc.rust-lang.org/std/macro.eprintln.html but couldn't quite understand

chusteven (commented):
Neat

diesel_cli/src/database.rs
weiznich: This change only affects diesel-cli. It contains quite a few hardcoded queries for loading schema information. The only user provided SQL that's executed there is the SQL in your migrations. Given that you run the client tool mostly manually I wouldn't expect that to leak any important information.

diesel_cli/src/infer_schema_internals/pg.rs
weiznich: The old code was from some time before `eprintln!` existed. It basically emulated what `eprintln` does internally by writing directly to stderr, with the difference that it would bubble up any potential error from write. I think `eprintln` just panics in that case, but given that this likely means that stderr is not there it implies that something is really broken. At that point there is not much we could do anyway, so I don't think that matters.

PR #3898, opened 2024-01-09
fix: fix glob resolution issue

bvanjoi:
Hi, we have recently made some changes related to glob name resolution in rustc and identified some regressions in this case. This PR aims to address these issues. You can find more background information at: https://github.com/rust-lang/rust/pull/114682

### DIFF
--- a/diesel_migrations/migrations_internals/src/connection.rs
+++ b/diesel_migrations/migrations_internals/src/connection.rs
@@ -41,7 +41,7 @@ where
     }
 
     fn latest_run_migration_version(&self) -> QueryResult<Option<String>> {
-        use diesel::dsl::max;
+        use diesel::expression::dsl::max;
         __diesel_schema_migrations.select(max(version)).first(self)
     }
 


diesel_migrations/migrations_internals/src/connection.rs
```diff
@@ -41,7 +41,7 @@ where
     }
 
     fn latest_run_migration_version(&self) -> QueryResult<Option<String>> {
-        use diesel::dsl::max;
```
bvanjoi, line 44: `diesel::dsl` has re-exported an ambiguous ﻿max, one exported from ﻿`dsl::helper_types::*` and another from ﻿`dsl::expression::dsl`, see: https://github.com/diesel-rs/diesel/blob/1.4.x/diesel/src/lib.rs#L220-L224

weiznich:
Thanks for opening this PR. What is the impact of not issuing a patch release for this issue for diesel 1.x. Would it break existing applications on newer rust version or would it only result in a new warning?

diesel_migrations/migrations_internals/src/connection.rs
weiznich: This pattern continues to exist in diesel 2.0. See https://github.com/diesel-rs/diesel/issues/3745 for some discussion around that. As it is hard to completely remove this bug without doing a breaking change (and therefore a major version bump) what's the impact on that?

bvanjoi:
> What is the impact of not issuing a patch release for this issue for diesel 1.x.

There are approximately over 200 regressions as a result of this case. For details, refer to `﻿migrations_internals-1.4.x` at the following link: https://crater-reports.s3.amazonaws.com/pr-114682-1/index.html

> Would it break existing applications on newer rust version

If the https://github.com/rust-lang/rust/pull/114682 gets merged, it will lead to a build error.

> or would it only result in a new warning

Initially, I intended to issue a warning regarding this matter. However, I've discovered a more complex case: `﻿max` has been used as a function in `﻿__diesel_schema_migrations.select(max(version))`, when it is actually intended to refer to a `type ﻿max = xxx`. For more details, please refer to https://github.com/rust-lang/rust/pull/114682#issuecomment-1881630225.

Therefore, resolving this issue simply as a warning won't truly address this regression. It requires additional work, such as ignoring the type namespace if there is ambiguity in the binding defined at the extern crate. However, implementing this change doesn't seem worthwhile as the problem can be resolved by submitting this PR. Once this PR is merged, we can advise users who encounter this regression to use `1.4.9`

diesel_migrations/migrations_internals/src/connection.rs
bvanjoi: > This pattern continues to exist in diesel 2.0

The regression test report doesn't indicate any regression in ﻿`diesel@2.0`, so I need to investigate further to determine if it's truly identical to 1.4.x.

weiznich:
> Once this PR is merged, we can advise users who encounter this regression to use 1.4.9

The main reason why I asked is because I consider diesel 1.4.x as not actively supported anymore. I fear that issuing another update will raise expectations from users that did not update yet that we will continue to provide fixes and other stuff for this old version. That's nothing I'm able to do on the side. For that reason I would prefer not having to issue an update that fixes an breaking change done by rustc.

PR #3917, opened 2024-01-26
Fix doc example to make it clear that ".returning()" does not preserve original order

greyblake:
The current doc example implies, that the order of returned values are always preserved:

```rs
let inserted_names = diesel::insert_into(users)
    .values(&vec![name.eq("Timmy"), name.eq("Jimmy")])
    .returning(name)
    .get_results(&connection);
assert_eq!(Ok(vec!["Timmy".to_string(), "Jimmy".to_string()]), inserted_names);
```

But unfortunately it's not guaranteed (at least definitely not by PostgreSQL DB, which is used for this doc test).

See more: https://stackoverflow.com/questions/5439293/is-insert-returning-guaranteed-to-return-things-in-the-right-order

P.S.
Thank you for the nice and reliable piece of software :heart:


weiznich (approved):
Thanks for opening this PR and improving the documentation :+1: 

(Tests will likely fail due to https://github.com/chronotope/chrono/issues/1389, I will take care of that later on)

greyblake:
@weiznich Thank you!

PR #3921, opened 2024-02-02
Add Postgres TABLESAMPLE support.

dvogel:
These changes add support for queries like so:

```
SELECT * FROM my_table TABLESAMPLE BERNOULLI(10)
```

The syntax, excerpt from [the PostgreSQL docs](https://www.postgresql.org/docs/16/sql-select.html):

```
TABLESAMPLE sampling_method ( argument [, ...] ) [ REPEATABLE ( seed ) ]
```

I'm skeptical that these changes are ready to merge. The basic functionality does work for me in a side project. However I have not been able to test this to a level I feel comfortable with.

Note the TODO comment regarding the `AppearsInFromClause` associated `Count` type. It is set to `Once` but I struggled to understand how this is used. If this means once per joinable fragment then it should be `Once`. However PostgreSQL does allow separate `TABLESAMPLE` clauses in each joinable fragment. Since it seems meant to pertain to the underlying table, perhaps it should be derived from the `S` type parameter? If so I'm afraid this is beyond my current understanding of the `diesel_derives` code :)

I'm not sure whether I've properly exported the `TablesampleDsl` trait correctly. In my application code, where I call `.tablesample(...)` I also need to `use diesel::dsl::TablesampleDsl`. That felt a little unexpected. I looked into re-exporting it into the prelude but I didn't see other backend-specific DSL traits re-exported there so I backed away.


weiznich (commented):
Thanks for working on this :+1: That's really helpful and the PR is a good starting point to iterate on.

> Note the TODO comment regarding the AppearsInFromClause associated Count type. It is set to Once but I struggled to understand how this is used. If this means once per joinable fragment then it should be Once. However PostgreSQL does allow separate TABLESAMPLE clauses in each joinable fragment. Since it seems meant to pertain to the underlying table, perhaps it should be derived from the S type parameter? If so I'm afraid this is beyond my current understanding of the diesel_derives code :)

`AppearsInFromClause::Count` there indicates how often a query source appears in your from clause. `Once` says, that a certain query source (that one that's given to `AppearsInFromClause` as generic argument) appears in your from clause, which later allows columns in select/filter/etc use that table. The added impl essentially just say that `Tablesample` is samewhat transparent there and can be treated just like `table` itself.

> I'm not sure whether I've properly exported the TablesampleDsl trait correctly. In my application code, where I call .tablesample(...) I also need to use diesel::dsl::TablesampleDsl. That felt a little unexpected. I looked into re-exporting it into the prelude but I didn't see other backend-specific DSL traits re-exported there so I backed away.

The prelude reexports methods from `expression_methods`: https://github.com/diesel-rs/diesel/blob/master/diesel/src/lib.rs#L648
which in turn reexports the backend specific variants: https://github.com/diesel-rs/diesel/blob/master/diesel/src/expression_methods/mod.rs. That written: That trait does not really fit into the expression_methods module as it's something different. I would likely reexport it directly from `diesel::prelude` (and add `OnlyDsl` as well there).

> It seems most of the DSL tests are in the doctests. I tried to follow that pattern but I wasn't able to run those. A naive cargo rustdoc --features postgres -- --test led to all tests failing. I'd very much appreciate any direction you can provide to run those tests.

There are also some tests in `diesel_tests`, but it should be fine use doc-tests for this functionality. After all we only care that it's covered somewhere.

For running the tests you need to:
* Set the `DATABASE_URL` environment variable to a database that should be used for tests
* Run tests via `cargo test --no-default-features --features "postgres"`

In addition to the doc tests, I would appreciate having at least one test for this in `diesel_compile_tests`, that show this can only be used with the postgres backend.

dvogel:
Thank you for all of your feedback. I found it quite helpful! I have updated this branch to:

* add a compile test
* use bind params
* export the dsl symbols more ergonomically
* encode the sql specialization into type parameters.

Regarding the type parameters, I'm hoping you can provide some feedback regarding how close my approach is to your suggestion. I don't use this approach much in my own code so I'm likely missing some of the nuance in the existing dsl code. The code works in my own project, however the second of the two doctests fails to compile and I'm struggling to figure out why. The error:

```
---- diesel/src/pg/expression/extensions/tablesample_dsl.rs - pg::expression::extensions::tablesample_dsl::TablesampleDsl (line 63) stdout ----
error[E0277]: the trait bound `query_source::joins::Join<Tablesample<users::table, diesel::dsl::BernoulliMethod>, posts::table, Inner>: AppearsInFromClause<posts::table>` is not satisfied
   --> diesel/src/pg/expression/extensions/tablesample_dsl.rs:73:6
    |
12  |     .inner_join(posts::table)
    |      ^^^^^^^^^^ the trait `AppearsInFromClause<posts::table>` is not implemented for `query_source::joins::Join<Tablesample<users::table, diesel::dsl::BernoulliMethod>, posts::table, Inner>`
    |
    = help: the trait `AppearsInFromClause<T>` is implemented for `query_source::joins::Join<Left, Right, Kind>`
note: required for `posts::columns::id` to implement `AppearsOnTable<query_source::joins::Join<Tablesample<users::table, diesel::dsl::BernoulliMethod>, posts::table, Inner>>`
   --> diesel/src/pg/expression/extensions/../../../doctest_setup.rs:285:13
    |
285 |             id -> Integer,
    |             ^^
    = note: 2 redundant requirements hidden
    = note: required for `((users::columns::id, users::columns::name), (posts::columns::id, posts::columns::user_id, posts::columns::title))` to implement `AppearsOnTable<query_source::joins::Join<Tablesample<users::table, diesel::dsl::BernoulliMethod>, posts::table, Inner>>`
    = note: required for `query_source::joins::Join<Tablesample<users::table, diesel::dsl::BernoulliMethod>, posts::table, Inner>` to implement `QuerySource`
    = note: 1 redundant requirement hidden
    = note: required for `JoinOn<Join<Tablesample<table, BernoulliMethod>, table, Inner>, Grouped<Eq<Nullable<user_id>, Nullable<id>>>>` to implement `QuerySource`
    = note: the full type name has been written to '/tmp/rustdoctestMSz9fp/rust_out.long-type-14004784566645490992.txt'
    = note: required for `SelectStatement<FromClause<Tablesample<users::table, diesel::dsl::BernoulliMethod>>>` to implement `InternalJoinDsl<posts::table, Inner, diesel::expression::grouped::Grouped<diesel::expression::operators::Eq<NullableExpression<posts::columns::user_id>, NullableExpression<users::columns::id>>>>`
```

The difference between the two examples is the join, as the error indicates. However I'm having trouble understanding how the type constraints are being applied here. In `diesel/src/query_source/joins.rs` there is an impl for `AppearsInFromClause`:

```
impl<T, Left, Right, Kind> AppearsInFromClause<T> for Join<Left, Right, Kind>
where
    Left: AppearsInFromClause<T> + QuerySource,
    Right: AppearsInFromClause<T> + QuerySource,
    Left::Count: Plus<Right::Count>,
{
    type Count = <Left::Count as Plus<Right::Count>>::Output;
}
```

Since this impl targets `Join`, then something about the constraints on the type parameters must cause a mismatch:

```
Left: Tablesample<users::table, diesel::dsl::BernoulliMethod>
Left: AppearsInFromClause<T> + QuerySource,
```

In `diesel_derives/src/table.rs` I added:

```
impl<TSM> diesel::query_source::AppearsInFromClause<diesel::query_builder::Tablesample<table, TSM>>
    for table
    where TSM:  diesel::query_builder::TablesampleMethod
{
	type Count = diesel::query_source::Once;
}
```

Since `BernoulliMethod` is an impl of `TablesampleMethod`, that constraint is satisfied. Let's look at `QuerySource`. In `src/pg/query_builder/tablesample.rs` I added:

```
impl<S, TSM> QuerySource for Tablesample<S, TSM>
where
    S: Table + Clone,
    TSM: TablesampleMethod,
    <S as QuerySource>::DefaultSelection:
        ValidGrouping<()> + SelectableExpression<Tablesample<S, TSM>>,
{
    type FromClause = Self;
    type DefaultSelection = <S as QuerySource>::DefaultSelection;
```

It's pretty obvious that `users` implements `Table`. In `diesel_derives/src/table.rs` I added:

```
impl<TSM> diesel::SelectableExpression<diesel::query_builder::Tablesample<super::table, TSM>>
    for #column_name
    where TSM:  diesel::query_builder::TablesampleMethod {}
```

So it seems my mistake is likely not satisfying this constraint:

```
    <S as QuerySource>::DefaultSelection:
        ValidGrouping<()> + SelectableExpression<Tablesample<S, TSM>>
```

However, with the exception of the TSM type parameter, this follows the `Only` impl of `QuerySource` that works with joins.

diesel/src/pg/expression/extensions/tablesample_dsl.rs
```diff
@@ -0,0 +1,96 @@
...
+///     .select((users::name, posts::title))
+///     .load::<(String, String)>(connection);
+/// # }
+/// ```
+/// That query selects all of the posts for a random 10 percent of users, returning the same
+/// results each time it is run due to the static seed of 42.0.
+///
+pub trait TablesampleDsl: Table {
+    /// See the trait-level docs.
+    fn tablesample<TSM: TablesampleMethod>(
```
weiznich, line 82: I would rather split up that method in the following parts:

```
trait TablesampleDsl: Table {
     fn tablesample_bernoulli(self, portion: i16) -> Tablesample<Self, BernoulliMethod>;
     fn tablesample_system(self, portion: i16) -> Tablesample<Self, BernoulliMethod>;
}

impl Tablesample<T, TSM> {
    pub fn with_seed(self, seed: f64) -> Self;
}

```

In that way it's much closer to the builder pattern used by the other DSL variants.

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/// Represents a query with a `TABLESAMPLE` clause.
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy)]
+pub struct Tablesample<S, TSM> {
+    pub source: S,
+    pub method: PhantomData<TSM>,
+    pub portion: i16,
+    pub seed: Option<f64>,
```
weiznich, line 24: I would like to keep these fields private

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,170 @@
+use crate::{
+    expression::{Expression, ValidGrouping},
+    pg::Pg,
+    query_builder::{AsQuery, AstPass, FromClause, QueryFragment, QueryId, SelectStatement},
+    query_source::QuerySource,
+    result::QueryResult,
+    sql_types::{Double, SmallInt},
+    JoinTo, SelectableExpression, Table,
+};
```
weiznich, line 9: Minor style nit: Diesel uses unnested imports everywhere else, it would be great to keep that consistent.

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,170 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::backend::Backend;
+    use crate::pg::Pg;
+    use crate::query_builder::QueryBuilder;
+    use diesel::dsl::*;
+    use diesel::*;
+
+    // TODO: Borrowed from src/pg/transaction.rs -- should this be extracted elsewhere?
```
weiznich, line 136: It's fine to duplicate it here.

diesel_compile_tests/tests/fail/pg_specific_tablesample_cannot_by_used_on_sqlite.rs
```diff
@@ -0,0 +1,26 @@
...
+fn main() {
+    use self::users::dsl::*;
+    let mut connection = SqliteConnection::establish(":memory:").unwrap();
+
+    let random_user_ids = users
+        .tablesample(
+            TablesampleMethod::System(10),
+            TablesampleSeed::Repeatable(42),
+        )
+        .load::<i32>(&mut connection);
```
weiznich, line 25: Can we repeat these lines with a `MysqlConnection`?

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,170 @@
...
+    type OnClause = <S as JoinTo<T>>::OnClause;
+
+    fn join_target(rhs: T) -> (Self::FromClause, Self::OnClause) {
+        <S as JoinTo<T>>::join_target(rhs)
+    }
+}
+
+impl<S, TSM> Table for Tablesample<S, TSM>
+where
+    S: Table + Clone + AsQuery,
```
weiznich, line 107: I think it should be fine to just skip these `Clone` bounds, as they are not needed, right?

weiznich (commented):
Thanks for the update. This is now much closer to what I had in mind :+1: 

I've left a bunch of comment for some changes to the exposed API, but the implementation itself, based on the generic marker types looks fine.

As for the errors with the joins: It looks like you might missing the equivalent of these impls: 

https://github.com/diesel-rs/diesel/blob/1c24bf64ea2c4961d7d042c21e93e69fdb56073a/diesel/src/macros/mod.rs#L230-L245

Also you likely need to update the output for the other compile tests as well as some of the error output is quite sensitive to what rustc emits (It contains a number of types that implement a certain trait, which obviously might change here)

diesel/src/pg/query_builder/tablesample.rs
dvogel: I believe these constraints are needed. Without them we run into issues with the `from_clause` method where the source is cloned. Without them we run into this compile error:

```
  --> diesel/src/pg/query_builder/tablesample.rs:91:9
   |
90 |     fn from_clause(&self) -> Self::FromClause {
   |                              ---------------- expected `Tablesample<S, TSM>` because of return type
91 |         self.clone()
   |         ^^^^^^^^^^^^ expected `Tablesample<S, TSM>`, found `&Tablesample<S, TSM>`
   |
   = note: expected struct `Tablesample<_, _>`
           found reference `&Tablesample<_, _>`
note: `Tablesample<S, TSM>` does not implement `Clone`, so `&Tablesample<S, TSM>` was cloned instead
  --> diesel/src/pg/query_builder/tablesample.rs:91:9
   |
91 |         self.clone()
   |         ^^^^
   = help: `Clone` is not implemented because the trait bound `S: Clone` is not satisfied
help: consider annotating `Tablesample<S, TSM>` with `#[derive(Clone)]`
   |
38 + #[derive(Clone)]
39 | pub struct Tablesample<S, TSM>
   |

For more information about this error, try `rustc --explain E0308`.
```

dvogel:
Thanks for pointing me toward `macros/mod.rs`! Adding those trait impls did allow my doctest to compile. The latest changes include the builder pattern approach you suggested. I believe all of the test failures are unrelated to my changes. I have [a CI run with a few extra unrelated fixes here](https://github.com/dvogel/diesel/actions/runs/7857108458) that show the postgresql tablesample tests all passing.

weiznich:
The fixes for the failing CI tests have been merged, so if you rebase your PR I can do a hopefully final review.

dvogel:
Thanks for the heads up. I've rebased this and marked it as ready for review.

diesel/src/pg/expression/extensions/tablesample_dsl.rs
```diff
@@ -0,0 +1,71 @@
...
+/// Calling this function on a table (`mytable.tablesample(...)`) will result in the SQL
+/// `FROM mytable TABLESAMPLE ...` --
+/// `mytable.tablesample(...)` can be used just like any table in diesel since it implements
+/// [Table](crate::Table).
+///
+/// The `BernoulliMethod` and `SystemMethod` types can be used to indicate the sampling method for
+/// a `TABLESAMPLE method(p)` clause where p is specified by the portion argument. The provided
+/// percentage should be an integer between 0 and 100.
+///
+/// If the seed argument is is Some(f) then f becomes the seed in `TABLESAMPLE ... REPEATABLE (f)`.
```
weiznich, line 22: This part is not correct anymore, it should likely be moved to the `with_seed()` function. We likely want to link that function from here as well?

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/// Represents a query with a `TABLESAMPLE` clause.
+#[doc(hidden)]
+#[derive(Debug, Clone, Copy)]
+pub struct Tablesample<S, TSM>
+where
+    TSM: TablesampleMethod,
+{
+    pub source: S,
```
weiznich, line 42:
```suggestion
    source: S,
```

We generally do not expose the internals of our query dsl structs, because that allows us to adjust them later if needed without issuing a new major release.

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,213 @@
...
+pub struct SystemMethod;
+
+impl TablesampleMethod for SystemMethod {
+    fn method_name_sql() -> &'static str {
+        "SYSTEM"
+    }
+}
+
+/// Represents a query with a `TABLESAMPLE` clause.
+#[doc(hidden)]
```
weiznich, line 36: We likely want to make this not `#[doc(hidden)]` to expose the documentation of the `with_seed()` function as part of the public API.

diesel/src/pg/query_builder/tablesample.rs
```diff
@@ -0,0 +1,213 @@
...
+    method: PhantomData<TSM>,
+    portion: i16,
+    seed: Option<f64>,
+}
+
+impl<S, TSM> Tablesample<S, TSM>
+where
+    TSM: TablesampleMethod,
+{
+    pub fn new(source: S, portion: i16) -> Tablesample<S, TSM> {
```
weiznich, line 52:
```suggestion
    pub(crate) fn new(source: S, portion: i16) -> Tablesample<S, TSM> {
```

There is no need for someone outside of diesel to call this function, as query dsl structs are generally constructed via the corresponding trait methods.

diesel/src/pg/query_builder/tablesample.rs
weiznich: :+1:

diesel/src/query_builder/mod.rs
```diff
@@ -120,6 +120,9 @@ pub(crate) use self::insert_statement::ColumnList;
 #[cfg(feature = "postgres_backend")]
 pub use crate::pg::query_builder::only::Only;
 
+#[cfg(feature = "postgres_backend")]
+pub use crate::pg::query_builder::tablesample::{Tablesample, TablesampleMethod};
```
weiznich, line 124:
```suggestion
pub use crate::pg::query_builder::tablesample::{Tablesample};
```

Exposing the `TablesampleMethod` trait shouldn't be required for the user, right?

For the macros we usually expose it via `diesel::internals::macro_name::TablesampleMethod` where macro_name corresponds to the macro requiring that trait.

weiznich (commented):
Thanks for the update :+1:

I'm sorry but I have another round of mostly minor comments about documentation and what to make public where. It would be great if you could address them, otherwise I might fix them myself in the next week. 

The other thing that would be great to have is an entry into the `Changelog.md` file mentioning this new feature.

weiznich:
No worries, take the time you need for this I would not have time to do this on my own in the next week or so.

weiznich (approved):
Thanks for the update. It looks good now 👍

I will take care of the CI failure tomorrow. It seems to be unrelated (and caused by a new deprecation in a new chrono version)

Issue #2149, opened 2019-08-17
sqlite rowid support


ivegotasthma:
## Setup
```
NixOS 19.03

[nix-shell:~/work]$ rustup --version
rustup 1.16.0 ( )

[nix-shell:~/work]$ rustc --version
rustc 1.38.0-nightly (c43d03a19 2019-08-14)

[nix-shell:~/work]$ cargo --version
cargo 1.38.0-nightly (e853aa976 2019-08-09)
```
### Versions

- **Rust:**
- **Diesel:** 1.4.2
- **Database:** sqlite
- **Operating System**

### Feature Flags

- **diesel:** sqlite, chrono

## Problem Description
diesel doesn't know that sqlite generates a PK for each table called rowid. I don't need to specify a PK.

### What are you trying to accomplish?
create a table

### What is the expected output?
migrations are ok

### What is the actual output?
`Diesel only supports tables with primary keys. Table poster has no primary key`

### Are you seeing any additional errors?
no

### Steps to reproduce
migration:

```
CREATE TABLE IF NOT EXISTS poster (
    key TEXT UNIQUE NOT NULL
);
```

running `diesel migration run` causes the error.
## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) for similar issues.
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)

sgrif:
> diesel doesn't know that sqlite generates a PK for each table called rowid. I don't need to specify a PK.

This is because SQLite doesn't expose this to us. Tables can be created without rowids, and we have no way of knowing whether that was the case or not. You should explicitly specify a primary key if you'd like to use Diesel to access that table, otherwise you can add it to the list of tables to ignore.

longsleep:
It would be nice to reopen this.

To figure out which tables are rowid table one can use something like this:

```sql
SELECT name, NOT EXISTS(SELECT 1 FROM pragma_index_xinfo(x.name)) AS 'hasRowid'
    FROM sqlite_schema AS x
    WHERE type='table'
    ORDER BY name;
```

With this information, diesel could implicitly add a primary key to the schema if there is none available explicitly.

longsleep:
> @longsleep Can you explain how that would be different to what's implemented in #3680?

Oh great thanks for pointing  to that MR. I did not check the "unreleased" changes. 

[...]

PR #3926, opened 2024-02-06
Add implicit sqlite rowid columns to print-schema output

longsleep:
Whenever sqlite uses an implicit rowid it also need to be added as column to the corresponding schema. This change fixes the implicit primary key generator added in #3680 which was never tested as the test added in this pull request never actually is run.

This change enables the test and enhances the schema printer to generate the missing column if the primary key is an implicit rowid column.

Related: #2149

### DIFF
--- a/diesel_cli/src/infer_schema_internals/sqlite.rs
+++ b/diesel_cli/src/infer_schema_internals/sqlite.rs
@@ -114,6 +114,13 @@ fn get_sqlite_version(conn: &mut SqliteConnection) -> QueryResult<SqliteVersion>
     Ok(SqliteVersion::new(parts[0], parts[1], parts[2]))
 }
 
+// In sqlite the rowid is a signed 64-bit integer.
+// See: https://sqlite.org/rowidtable.html
+// We should use BigInt here but to avoid type problems with foreign keys to
+// rowid columns this is for now not done. A patch can be used after the schema
+// is generated to convert the columns to BigInt as needed.
+const ROWID_TYPE_NAME: &str = "Integer";
+
 pub fn get_table_data(
     conn: &mut SqliteConnection,
     table: &TableName,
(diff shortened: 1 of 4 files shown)


diesel_cli/src/infer_schema_internals/sqlite.rs
```diff
@@ -134,6 +141,40 @@ pub fn get_table_data(
...
+            }
+        }) {
+            // Prepend implicit rowid column for the rowid implicit primary key.
+            result.insert(
+                0,
+                ColumnInformation {
+                    column_name: String::from(primary_key),
+                    type_name: String::from(ROWID_TYPE_NAME),
+                    type_schema: None,
+                    nullable: true,
```
weiznich, line 170:
```suggestion
                    nullable: false,
```

I think rowid cannot ever by null, right? (At least that what's stated [here](https://sqlite.org/rowidtable.html))

weiznich (commented):
Thanks, for the fast fix :+1:

This looks fine beside that minor thing with making `rowid` columns not nullable. (That will likely require updating the generated schema.rs file)

diesel_cli/src/infer_schema_internals/sqlite.rs
longsleep: Good point - this was an intentional choice to make the auto generated column the same as the minimal explicitly created rowid which usually would also be defined nullable even though that it can never return NULL I think it can be set to NULL (and the database will then generate a new value) if that makes sense.

In my testing so far the generated schemas with this patch work just fine and no additional post processing is required for a functional database access to tables with implicit rowid.

longsleep: @weiznich I would prefer to keep the rowid column nullable for now, mostly to avoid too big of a change for existing Diesel users (e.g. https://github.com/diesel-rs/diesel/blob/1c24bf64ea2c4961d7d042c21e93e69fdb56073a/diesel_cli/tests/print_schema/print_schema_simple/sqlite/expected.snap#L17 would change).

Imo time will tell if this decision should be revisited.

weiznich:
Can you rebase the branch to pull in the latest changes (to fix the CI)?

weiznich (approved):

longsleep:
> Can you rebase the branch to pull in the latest changes (to fix the CI)?

I have debased but the tests still fail with errors imo unrelated to the actual change. What needs to happen to get this merged?

weiznich:
Seems like we have hit exactly that time window where a new stable rustc version dropped. That usually means that new lints fire off and cause issues. If you don't mind you can push a change that fixes that, otherwise I will try to take care of that later today. 
(There is yet another issue in the nightly released today, so yet another build will fail for now :()

weiznich:
Can you rebase the PR again to pull in the fixes for the CI?

PR #3932, opened 2024-02-09
Fix typo

stormshield-kg:


JohnTitor (approved):

dvogel:
I ran into this issue too. I think the fix is a little more complicated. Since this typo had been preventing the CI job from running these tests, they have drifted a bit. This would normally be visible but the sqlite ASAN changes are preventing the full CI job from running. You can cherry-pick commit https://github.com/dvogel/diesel/commit/46349b912472bc6e126fa26a982db49381a9d3ad to get the fixes for the tests. You can see a CI run for those changes [here](https://github.com/dvogel/diesel/actions/runs/7857775494) (though that branch has some additional commits you won't want to allow the test job to be run).

weiznich:
This still needs to wait for https://github.com/rust-lang/rust/issues/120830 to being fixed. Otherwise this should work now.

PR #3933, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. Can you explain whether or not that will preserve currently specified minimal versions and how that will support version ranges (like those for the libsqlite3-sys crate)? Changing these minimal versions is not possible without issuing a major release.

Eveeifyeve:
Even tho I am not the author but based on my uses and what the commit does it's doesn't but you can specify versions as I think. But really all it does it opens pr with the version updated you could just close the prs that you don't want.

weiznich:
Given that we use version ranges in many places that would mean that we get a lot of PR's that need to be closed. I personally rather spend the time to do manual updates before any release instead of closing a never ending stream of PR.

cc @diesel-rs/reviewers for your opinions

kw217:
I agree. I've used dependabot previously in a large organisation with many repositories. It was a helpful way of ensuring the work got done and nothing was forgotten, but I also had a dread of arriving in the morning to a pile of PRs. As long as Diesel has a process that ensures this work is done, I don't see the need for change.

weiznich:
There seems to be a conses to not setup this integration.

PR #3939, opened 2024-02-19
cli setup improvements

dennybiasiolli:
- cli: using fs::create_dir_all in create_migrations_directory
    
    To recursively create a directory and all of its parent components if they are missing.

- cli: writing the custom migration dir to the config file
- cli: fixing windows paths in toml strings and adding a specific test for that

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -144,9 +144,11 @@ fn create_config_file(
     let path = Config::file_path(matches);
     if !path.exists() {
         let source_content = include_str!("default_files/diesel.toml").to_string();
+        // convert the path to a valid toml string (escaping backslashes on windows)
+        let migrations_dir_toml_string = migrations_dir.display().to_string().replace('\\', "\\\\");
         let modified_content = source_content.replace(
             "dir = \"migrations\"",
-            &format!("dir = \"{}\"", migrations_dir.display()),
+            &format!("dir = \"{}\"", migrations_dir_toml_string),
         );
         let mut file = fs::File::create(path)?;
         file.write_all(modified_content.as_bytes())?;
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -183,7 +183,7 @@ fn generate_completions_command(matches: &ArgMatches) {
 /// Returns a `DatabaseError::ProjectRootNotFound` if no Cargo.toml is found.
 fn create_migrations_directory(path: &Path) -> Result<PathBuf, crate::errors::Error> {
     println!("Creating migrations directory at: {}", path.display());
-    fs::create_dir(path)?;
+    fs::create_dir_all(path)?;
     fs::File::create(path.join(".keep"))?;
     Ok(path.to_owned())
 }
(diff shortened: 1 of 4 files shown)


dennybiasiolli:
@weiznich if you need me to rebase, just let me know

diesel_cli/src/main.rs
```diff
@@ -136,12 +136,20 @@ fn create_migrations_dir(matches: &ArgMatches) -> Result<PathBuf, crate::errors:
...
+    matches: &ArgMatches,
+    migrations_dir: &Path,
+) -> Result<(), crate::errors::Error> {
     use std::io::Write;
     let path = Config::file_path(matches);
     if !path.exists() {
+        let source_content = include_str!("default_files/diesel.toml").to_string();
+        let modified_content = source_content.replace(
+            "dir = \"migrations\"",
+            &format!("dir = \"{}\"", migrations_dir.display()),
```
weiznich, line 149: I'm not sure if that works on windows, as the relevant tests seem to fail.

weiznich (changes requested):
Thanks for opening this PR :+1: I'm generally fine with how this change is implemented and what it is supposed to change. The CI test failures on windows seem to be related to the change (and likely to the review comment left for the `create_config_file` function). That needs to be fixed before merging as we also need to support windows there.

diesel_cli/src/main.rs
dennybiasiolli: Oh you are right, let me setup a Windows virtual machine so I can fix that.

dennybiasiolli:
@weiznich I just found the reason for test errors: tests on windows are receiving an absolute path as `migrations_dir` and they crash on windows for that reason.
With [this workaround](https://github.com/dennybiasiolli/diesel/commit/c836378ca3e1d8f358b1a266414149a7830330c8) I managed to solve the problem for tests, but a cleaner solution would be to pass a relative path when testing.
Any idea?

weiznich:
As this might also be a potential problem if someone passes an absolute path manually to the CI, I would prefer an solution that just converts the path to an path relative to the `diesel.toml` file instead. Maybe just call [`Path::strip_prefix`](https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix)?

dennybiasiolli:
Let's try with another solution, the problem could be that the string wrote in the .toml file must be escaped and I'm not doing that.

I'm trying with
```rust
        let migrations_dir_toml_string = format!(
            "dir = \"{}\"",
            migrations_dir.display().to_string().replace('\\', "\\\\")
        );
```

Workflow here: https://github.com/dennybiasiolli/diesel/actions/runs/8005969731
Fingers crossed!

dennybiasiolli:
Ok now it works! https://github.com/dennybiasiolli/diesel/actions/runs/8007224538
Let me use the same fix in this branch

dennybiasiolli:
"STDERR: Failed to execute a database query: Identifier name 'diesel_setup_writes_migration_dir_by_arg_to_config_file_escaping_windows_paths' is too long"

This should be an easy fix 😰

dennybiasiolli:
The only failing check error now is this one:
`MBERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-8.0.31-winx64.zip'. Exception calling "Read" with "3" argument(s): "Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host." `

I think it shouldn't be. problem, but feel free to re-run the workflow just to be 100% sure 😉

weiznich (approved):

PR #3940, opened 2024-02-19
Use BigInt for integer primary key in sqlite for diesel-cli

### LINKED ISSUE #852: INTEGER PRIMARY KEY in SQLite
Diesel deduces the a column's type as `BigInt` if it's declared as `BIGINT` in the database. This is generally fine in SQLite because `INTEGER` and `BIGINT` are synonyms, but `INTEGER PRIMARY KEY` is treated differently and is always 64-bit according to https://sqlite.org/autoinc.html. Moreso, `BIGINT PRIMARY KEY AUTOINCREMENT` is not accepted.

It might be a good idea to detect these columns and use `BigInt` for them. Note that the SQLite documentation says that `INTEGER PRIMARY KEY` is special only for tables that aren't declared with `WITHOUT ROWID`, but `BIGINT PRIMARY KEY` is still not allowed even when using that.

stormshield-kg:
Fixes #852.

The `sqlite-integer-primary-key-is-bigint` config parameter is used to modify the generation.
With this option, the [`PRAGMA table_list`](https://www.sqlite.org/pragma.html#pragma_table_list) query is used to get the `WITHOUT ROWID` attribute for a table (available from SQLite 3.37).


diesel_cli/src/infer_schema_internals/sqlite.rs
```diff
@@ -176,6 +176,67 @@ impl QueryableByName<Sqlite> for PrimaryKeyInformation {
...
+    let table_list_results = sql_query(table_list_query).load::<WithoutRowIdInformation>(conn)?;
+
+    match table_list_results
+        .iter()
+        .find(|wr_info| wr_info.name == table.sql_name)
+        .map(|wr_info| wr_info.without_row_id)
+    {
+        Some(false) => Ok(true),
+        _ => Ok(false),
+    }
```
weiznich, line 237:
```suggestion
    let res = table_list_results
        .iter()
        .find(|wr_info| wr_info.name == table.sql_name)
        .map(|wr_info| !wr_info.without_row_id)
        .unwrap_or_default();
    Ok(res)
```

looks a bit cleaner.

diesel_cli/src/migrations/diff_schema.rs
```diff
@@ -514,11 +514,32 @@ where
...
-        generate_column_type_name(query_builder, &column.ty);
-        if primary_keys.contains(&column.rust_name) && primary_keys.len() == 1 {
+
+        if cfg!(all(
+            feature = "sqlite",
+            feature = "integer_primary_key_i64_for_sqlite_3_37"
+        )) && is_only_primary_key
+            && column.ty.sql_name.eq_ignore_ascii_case("BigInt")
+        {
+            let ty = ColumnType {
```
weiznich, line 529: Can we have a small comment here explaining that this is necessary to make this equivalent to the `ROWID` column? Maybe even with a link to the sqlite documentation?

diesel_cli/tests/migration_generate.rs
```diff
@@ -286,6 +286,29 @@ fn backend_file_path(test_name: &str, file: &str) -> PathBuf {
...
+        insta::assert_snapshot!("expected", contents);
+    }
+}
+
+fn schema_file() -> &'static str {
+    if cfg!(all(
+        feature = "sqlite",
+        feature = "integer_primary_key_i64_for_sqlite_3_37"
+    )) {
+        "schema_i64_pk.rs"
```
weiznich, line 306: This shouldn't be required if that is a config/cli option

weiznich (changes requested):
Thanks for working on this feature :+1:

I generally like the idea to address the underlying issue, but I strongly suggest not to use a feature flag for this as this makes it:

* harder to discover this feature
* harder to document this feature
* harder to test this feature (another entry in the feature matrix…)

Instead I would suggest making this a config option in the `diesel.toml` file and a CLI flag. This allows to control that feature at runtime.

diesel_cli/src/infer_schema_internals/sqlite.rs
```diff
@@ -204,6 +205,65 @@ impl QueryableByName<Sqlite> for PrimaryKeyInformation {
...
+pub fn column_is_row_id(
+    conn: &mut SqliteConnection,
+    table: &TableName,
+    column_name: &str,
+    type_name: &str,
+) -> Result<bool, crate::errors::Error> {
+    let sqlite_version = get_sqlite_version(conn)?;
+    if sqlite_version < SqliteVersion::new(3, 37, 0) {
+        return Err(crate::errors::Error::UnsupportedFeature(
+            "Parameter `sqlite_integer_primary_key_is_bigint` needs SQLite 3.37 or above".into(),
```
weiznich, line 231:
```suggestion
            format!("Parameter `sqlite_integer_primary_key_is_bigint` needs SQLite 3.37 or above. Your current SQLite version is {sqlite_version}"),
```

weiznich (changes requested):
Thanks for the update. This looks now much closer to something that is ready to be merged. I've left a few minor comments. The most critical thing is the loading of the primary keys while checking if the column is an alias for rowid. That needs to be fixed, the other comments are more suggestions.

It would also be great to have an entry in the `Changelog.md` file for this feature.

teohhanhui:
~~Sorry for commenting on a closed PR, but was there a reason why this was turned into a `BigInt` instead of `i64` to match [rowid](https://www.sqlite.org/lang_createtable.html#rowid) in SQLite?~~

**EDIT**: Just realized that this is [`diesel::BigInt`](https://docs.rs/diesel/latest/diesel/sql_types/struct.BigInt.html) - not [`num_bigint::BigInt`](https://docs.rs/num-bigint/latest/num_bigint/struct.BigInt.html). My apologies.

PR #3941, opened 2024-02-19
Implement Default for AssumeNotNull

dullbananas:


Ten0 (approved):

Ten0:
Just not sure to what extent this is backwards compatible - can't users implement that in their own crate if they own T?

weiznich:
That should be backward compatible. At least the following code does not compile in a downstream crate with an error message related to the fact that you cannot implement a foreign trait for a for foreign type. 

```rust
struct Foo;

impl Default for diesel::dsl::AssumeNotNull<Foo> {
    fn default() -> Self {
        todo!()
    }
}


```

This only works if the trait itself has a type parameter or if the type is marked at fundamental. (Beside of that downstream users wouldn't be able to provide a meaningful implementation of `Default` for that type, as it exposes no public fields and no constructor.)

PR #3942, opened 2024-02-20
updated quaint to pass bench tests

sabify:


sabify:
@weiznich Quaint is under very active development, and there are feature set inconsistencies in the crate that cause compile errors. It would be better if this crate were removed from diesel_bench, as it is a blocking feature for benchmarking postgres integration.

The postgres benchmarks of diesel metrics haven't received updates for months.

weiznich:
I know. I'm happy to accept a PR fixing that, otherwise there are just more important things for me to work on currently.

Issue #3868, opened 2023-12-07
Deserialization layer seems too permissive with regards to checking the actual types received


Ten0:
When the database sends a type that doesn't match the one declared in the schema, we don't check that it matches, and may attempt deserializing from a wire representation of a different type, possibly returning incorrect data without generating an error.

I encountered this scenario while changing the type of a column from `Int4` to `Int8`. I expected to see errors due to mismatches for the time it would take for my new service to deploy after the migration was done running, which I was fine with. (Or at least that it would deserialize the appropriate thing because everybody uses little endian, right?)

What actually happened however is the line went through here:
https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel/src/deserialize.rs#L538-L550
Then through here:
https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel/src/pg/types/integers.rs#L45-L63
Note that in this second part:
 - the length checks are `debug_assert!` (I wouldn't want it to panic anyway, I would want a deserialization error instead)
 - `NetworkEndian` is actually `BigEndian` (I just imagined that PG uses little endian since that's the one actually used on all machines so that's one more reason why I didn't expect this issue ^^')

Anyway so of course, with big endian and ignoring the size mismatch, diesel ended up deserializing only the leading zeroes of my `Int8` into an `Int4`, causing it to send a bunch of zeroes into my applicative code.

That doesn't seem like the ideal behavior. Instead, one would expect that if the Rust `schema.rs` of the currently deployed service doesn't match the actual database schema, corresponding queries error.

It seems that we could prevent that by checking the consistency between the database-provided OID:
https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel/src/pg/value.rs#L93
and our known OID for the type:
https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel/src/sql_types/mod.rs#L100-L101

[...]

PR #3944, opened 2024-02-23
Return Error From Postgres Float Deserialization

JonBoyleCoding:
I had an error where I was getting the debug assertion panic on a table and I had no idea what field was causing the issue. Here's a modification I made to ensure that any deserialization errors get passed up the chain with the field name without panicking in order to properly diagnose where my issue is happening.

I don't know if his is wanted or not, but thought I'd make the PR in case it is. Happy to do more modifications if necessary.

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -5,7 +5,6 @@ use std::result;
 
 use crate::backend::Backend;
 use crate::expression::select_by::SelectBy;
-use crate::result::DeserializeFieldError;
 use crate::row::{NamedRow, Row};
 use crate::sql_types::{SingleValue, SqlType, Untyped};
 use crate::Selectable;
@@ -543,13 +542,10 @@ where
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
         T::from_nullable_sql(field.value()).map_err(|e| {
             if e.is::<crate::result::UnexpectedNullError>() {
-                return e;
+                e
+            } else {
+                Box::new(crate::result::DeserializeFieldError::new(field, e))
             }
-
-            Box::new(DeserializeFieldError {
-                field_name: field.field_name().map(|s| s.to_string()),
-                error: e,
-            })
         })
     }
 }
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -542,6 +542,10 @@ where
 
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
         Ok(T::from_nullable_sql(field.value()).map_err(|e| {
+            if e.is::<crate::result::UnexpectedNullError>() {
+                return e;
+            }
+
             Box::new(DeserializeFieldError {
                 field_name: field.field_name().map(|s| s.to_string()),
                 error: e,
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -541,7 +541,7 @@ where
         use crate::row::Field;
 
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
-        Ok(T::from_nullable_sql(field.value()).map_err(|e| {
+        T::from_nullable_sql(field.value()).map_err(|e| {
             if e.is::<crate::result::UnexpectedNullError>() {
                 return e;
             }
@@ -550,7 +550,7 @@ where
                 field_name: field.field_name().map(|s| s.to_string()),
                 error: e,
             })
-        })?)
+        })
     }
 }
 
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -5,6 +5,7 @@ use std::result;
 
 use crate::backend::Backend;
 use crate::expression::select_by::SelectBy;
+use crate::result::DeserializeFieldError;
 use crate::row::{NamedRow, Row};
 use crate::sql_types::{SingleValue, SqlType, Untyped};
 use crate::Selectable;
@@ -540,9 +541,12 @@ where
         use crate::row::Field;
 
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
-        T::from_nullable_sql(field.value()).map_err(|e| {
-            format!("Error for field '{}': {:?}", field.field_name().unwrap(), e).into()
-        })
+        Ok(T::from_nullable_sql(field.value()).map_err(|e| {
+            Box::new(DeserializeFieldError {
+                field_name: field.field_name().map(|s| s.to_string()),
+                error: e,
+            })
+        })?)
     }
 }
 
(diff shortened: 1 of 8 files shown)


diesel/src/deserialize.rs
```diff
@@ -545,7 +545,9 @@ where
         use crate::row::Field;
 
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
-        T::from_nullable_sql(field.value())
+        T::from_nullable_sql(field.value()).map_err(|e| {
+            format!("Error for field '{}': {:?}", field.field_name().unwrap(), e).into()
```
weiznich, line 549: Given that this might contain a completely user defined error (as `FromSql` can be implemented by third party crates for custom types) I wouldn't want to just convert that error to a string. Rather I would prefer having an error type that wraps the inner error. So something like 

```rust
// part of `src/result.rs

#[non_exhaustive]
pub struct DeserializeFieldError {
    // use an option to not require unwrapping
    pub field_name: Option<String>,
    pub error: Box<dyn Error + …>
}

impl Error for DeserializeFieldError {
 // … (forward the inner error as source)
}
```

Afterwards we can change this line to use the new error type instead.

diesel/src/pg/types/floats/mod.rs
```diff
@@ -117,16 +117,18 @@ impl ToSql<sql_types::Numeric, Pg> for PgNumeric {
...
-            "Received more than 4 bytes while decoding \
-             an f32. Was a double accidentally marked as float?"
-        );
-        debug_assert!(
-            bytes.len() >= 4,
-            "Received less than 4 bytes while decoding \
-             an f32."
-        );
+
+        #[cfg(debug_assertions)]
```
weiznich, line 121: I would prefer returning a straight error here instead. I doubt that this will actually break anyone. (And if it does it uncovers bugs in their application)

weiznich (changes requested):
Thanks for opening this PR :+1:. Overall I'm happy with the suggestion to return a more useful error in this situation. I've made some suggestions how we could improve this in a more general way. With these suggestions applied + a test case for the new error I would be happy to merge this.

It would be great if that could also adjust the code in these impl:[ https://github.com/diesel-rs/diesel/blob/bc263afa1765f650d9c4bf36c40536ff95b9cade/diesel/src/pg/types/integers.rs#L45-L63 (](https://github.com/diesel-rs/diesel/blob/ad8ef479e1c6f926b940ad9a43877fc5a993ec5f/diesel/src/pg/types/integers.rs#L24-L83) (although that's not necessary for merging this PR)

There is a bit more discussion on this topic in https://github.com/diesel-rs/diesel/issues/3868

diesel/src/pg/types/floats/mod.rs
JonBoyleCoding: Done!

diesel/src/deserialize.rs
```diff
@@ -545,7 +546,12 @@ where
         use crate::row::Field;
 
         let field = row.get(0).ok_or(crate::result::UnexpectedEndOfRow)?;
-        T::from_nullable_sql(field.value())
+        Ok(T::from_nullable_sql(field.value()).map_err(|e| {
+            Box::new(DeserializeFieldError {
```
weiznich, line 550: The CI error messages indicate that I missed something. We use the error type to transport the information that the field is null:

https://github.com/diesel-rs/diesel/blob/ad8ef479e1c6f926b940ad9a43877fc5a993ec5f/diesel/src/deserialize.rs#L443

This means we need to change this mapping here to exclude this specific errors:

This should result in something like:

```rust
.map_err(|e| {
    if !e.is::<crate::result::UnexpectedNullError>() {
        Box::new(DeserializeFieldError {
                field_name: field.field_name().map(|s| s.to_string()),
                error: e,
          })
    } else {
        e
    }
}
```

This should hopefully fix the CI issues,

diesel/src/result.rs
```diff
@@ -414,3 +414,32 @@ impl fmt::Display for UnexpectedEndOfRow {
 }
 
 impl StdError for UnexpectedEndOfRow {}
+
+/// An error occurred while deserializing a field
+#[derive(Debug)]
+pub struct DeserializeFieldError {
```
weiznich, line 420:
```suggestion
#[non_exhaustive]
pub struct DeserializeFieldError {
```

We don't  want that users can construct this error type. We also want to be able to add fields later on.

weiznich (commented):
Thanks for updating the code. I've noticed an issue that I missed while proposing this solution. I've left a comment how to fix that. Otherwise this approach looks good now.

diesel/src/result.rs
JonBoyleCoding: Just to check - you want me to remove the `pub` here?

weiznich: `pub` should stay there, as you likely want to be able to access these fields from your application as wel.

weiznich (commented):
In terms of changes this looks fine now, so good work :+1:

I would like to have a test for this before merging (put it into `diesel_tests`, should create a table, and query as the wrong type + check for this error)  + this likely requires an entry in the changelog (See the `Changelog.md` file for this).

JonBoyleCoding:
Hi - sorry I haven't had a chance to work on this further. I saw that you have continued this on so thanks for that. I have been very busy at work and was unable to work on writing tests.

weiznich:
> Hi - sorry I haven't had a chance to work on this further. I saw that you have continued this on so thanks for that. I have been very busy at work and was unable to work on writing tests.

No worries about that. It's great that submitted that PR at all.

weiznich:
To just leave a comment here: I've found a performance regression in this commit regarding to certain benchmarks while running it locally. I'm currently investigating options to workaround this regression.

weiznich:
@Ten0 I've pushed my local experimentation here: https://github.com/weiznich/diesel/commit/5ef08d965c08565fc31e97e518d6cd67639f1753

As far as I remember there was a ~5% instruction count regression for queryable_by_name queries on postgres. Some of the added attributes makes things slightly better. It likely just needs another round of local benchmarking to validate the results, I just run out of time back then and forgot about it.

Ten0:
> @Ten0 I've pushed my local experimentation here: [weiznich@5ef08d9](https://github.com/weiznich/diesel/commit/5ef08d965c08565fc31e97e518d6cd67639f1753)
> 
> As far as I remember there was a ~5% instruction count regression for queryable_by_name queries on postgres. Some of the added attributes makes things slightly better. It likely just needs another round of local benchmarking to validate the results, I just run out of time back then and forgot about it.

Thanks! 😊
Do you have a quick pointer on how I could run the relevant instruction count benchmark(s)?

weiznich:
> Do you have a quick pointer on how I could run the relevant instruction count benchmark(s)?

You should be able to run the benchmarks via `cargo bench --features "instruction_count Postgres"` from the diesel_benches. It requires a Linux environment that has access to the performance counters (that can run perf). 

There is also a "fast_run" feature that disables some of the benchmarks. Criterion also allows to store and use explicit baselines, which is useful for comparing to a specific version.

weiznich:
I've done some more benchmarks locally and I'm now quite certain that the regression in instruction count does not make the code measurably slower. At least `perf diff` did not show anything but noise after comparing the base branch with the change (including the optimizations).

weiznich (approved):

PR #3949, opened 2024-02-28
feat: add `tokio-postgres` in `diesel_bench`

sabify:


weiznich (commented):
Can you explain why you would expect different results for this crate than for the `postgres` crate (which is a thin wrapper over `tokio-postgres`)?

sabify:
There's still marginal difference between these two crates performance in benchmark and I thought it would be nice to have that crate in bench feature set.

Please look at the following benchmark result (expect the bench_insert that has a bug in this specific run).

Benchmark result: https://github.com/sabify/diesel/actions/runs/8081091336/job/22083538884

PR #3953, opened 2024-03-07
Fix new deprecated functions from chrono 0.4.35

weiznich:
For the tests we switch to the suggested alternatives, for the real implementation we continue to use the deprecated functions as the suggested alternatives do not exist in our minimal supported chrono version.


pppp24 (approved):
looks good 👍

PR #3964, opened 2024-03-15
Update README to indicate that diesel supports MariaDB

lottaquestions:
Description
=========
Diesel was tested against MariaDB 11.4, and worked out of the box using the MySQL setup instructions and example.

Testing
======
1. Installing rust:

    ```
    curl --proto '=https' --tlsv1.3 https://sh.rustup.rs -sSf | sh
    ```

2. Ensuring rust can be found in the PATH for the terminal session:

    ```
    . "$HOME/.cargo/env"
    ```

3. Then follow the diesel instructions: https://diesel.rs/guides/getting-started. The following tweak had to be made since MariaDB was not installed in a standard location:

    ```
    RUSTFLAGS='-L /github/MariaDB/installations/11.4/lib' cargo install diesel_cli --no-default-features --features mysql
    ```
4. Run the MySQL example in https://github.com/lottaquestions/diesel/tree/master/examples/mysql/getting_started_step_3:
   4.1 Write  a post
    ```
    export RUSTFLAGS='-L /github/MariaDB/installations/11.4/lib'
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/github/MariaDB/installations/11.4/lib cargo run --bin write_post
    ```
   4.2 Publish a post
    ```
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/github/MariaDB/installations/11.4/lib cargo run --bin publish_post 1
    ```
   4.3 Get a post
    ```
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/github/MariaDB/installations/11.4/lib cargo run --bin get_post 1
    ```
   4.4 Show posts
    ```
    LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/github/MariaDB/installations/11.4/lib cargo run --bin show_posts
    ```
5. Query the MariaDB database directly to show that the schema was created correctly and that the data was correctly inserted in the database:
    ```
    MariaDB [(none)]> show databases;
    +--------------------+
    | Database           |
    +--------------------+
    | diesel_demo        |
    | information_schema |
    | mysql              |
    | performance_schema |
    | sys                |
    | test               |
    +--------------------+
    6 rows in set (0.002 sec)
     
    MariaDB [(none)]> use diesel_demo;
    Database changed
    MariaDB [diesel_demo]> show tables
        -> ;
    +----------------------------+
    | Tables_in_diesel_demo      |
    +----------------------------+
    | __diesel_schema_migrations |
    +----------------------------+
    1 row in set (0.002 sec)
     
    MariaDB [diesel_demo]> select * from __diesel_schema_migrations limit 10;
    +----------------+---------------------+
    | version        | run_on              |
    +----------------+---------------------+
    | 20240314204819 | 2024-03-14 16:49:18 |
    +----------------+---------------------+
    1 row in set (0.001 sec)
     
    MariaDB [diesel_demo]> show tables;
    +----------------------------+
    | Tables_in_diesel_demo      |
    +----------------------------+
    | __diesel_schema_migrations |
    | posts                      |
    +----------------------------+
    2 rows in set (0.002 sec)
     
    MariaDB [diesel_demo]> select * from posts;
    +----+------------------+-------------------------------------------------------+-----------+
    | id | title            | body                                                  | published |
    +----+------------------+-------------------------------------------------------+-----------+
    |  1 | Diesel test demo | This test proves that Diesel works with MariaDB 11.4
     |         1 |
    +----+------------------+-------------------------------------------------------+-----------+
    1 row in set (0.002 sec)
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Supported databases:
 1. [PostgreSQL](https://docs.diesel.rs/master/diesel/pg/index.html)
 2. [MySQL](https://docs.diesel.rs/master/diesel/mysql/index.html)
 3. [SQLite](https://docs.diesel.rs/master/diesel/sqlite/index.html)
+4. [MariaDB](https://docs.diesel.rs/master/diesel/mysql/index.html) - Uses the same setup as MySQL.
 
 You can configure the database backend in `Cargo.toml`:
 


weiznich:
Thanks for opening this PR. Please note that "supports" is something different than "works with". For the former we give more guarantees what's expected and what's not while the later just happens to work at the current time. We do not want to claim mariadb support at this point in time as there are quite a few unsupported features for that backend. Supporting these features would need another backend/connection in diesel. See https://github.com/diesel-rs/diesel/issues/1882 for more discussion around this topic. Contributions there are welcome.

I'm closing this PR as I'm not want to state official support for something that's not really officially supported by us.

ottok:
Isn't this a bit unreasonable? Instead of outright closing the PR, could you just give feedback to submitter on how you wish the MariaDB support would be mentioned?

You, @virome and @tgross35 discuss MariaDB compatibility in #1882 and clearly it is already known by you but not documented in docs. In this PR a new person who didn't know anything about it and was wondering about it ended up doing a bunch of testing to see if it actually works, and contributed to docs to capture it at least with one sentence. Could that be supported with at least some advice in _how_ you want MariaDB to be mentioned in the docs?

Right now I don't spot the docs saying anything about MariaDB compatibility at all (nor anything about libmariadb3 compatibility, which is slightly a different topic).

tgross35:
MDB keeps a list of differences from My, which is pretty comprehensive https://mariadb.com/kb/en/incompatibilities-and-feature-differences-between-mariadb-11-3-and-mysql-8-/

The most relevant things for me at the application or ORM level have been `RETURNING` (that one is nice for performance, saves a round trip), temporal tables (`WITH SYSTEM VERSINOING`), native UUID and INET6/INET4 types (usually this can be worked around but it can be ugly) and `OR REPLACE` syntax (just a convenience).

weiznich:
@ottok Can you clarify how exactly that would work? I already stated that we don't consider mariadb as supported backend, but merely only as a backend that happens to work with diesel. There are a few other database systems in a similar situation, we don't mention any of them, so why should we handle mariadb differently.

To repeat my main concern here: There is a huge difference between "I've tested this locally and it seems to work for the starter example" and "This is a officially supported backend". For the later one we would want to 
* have a CI setup that runs the whole test-suite against this backend to make sure it does not break
* a setup that allows us to support most/all features of that backend without compromising other backends (in that case you cannot support `RETURNING`  (or any of the types, or …) for mariadb only if you use the mysql backend for this, therefore the need for a distinct backend as stated in #1882)

Given these points I'm sorry but I really see no way forward how this PR can be accepted at all. I do not want to waste the time of the contributor by suggesting something different, so I closed this PR.

ottok:
What is the README was updated to mention that MariaDB support is partial, and that MariaDB users should contribute to #1882?

There might be a bit of a chicken-and-egg situation if the docs don't indicate anything about MariaDB support, turning away MariaDB users, decreasing the number of potential contributors to #1882.

weiznich:
Our readme is not the place the right place to point users to such issues. Our experience have shown that public links to issues rarely lead to getting more contributors but often lead to unhelpful comments (I want this too, why is this not implemented, +1, etc) on these issues. Thats not helpful for us maintainers. Anyone that is willing to contribute support for another database backend is able to find that issue by a quick search.

PR #3967, opened 2024-03-17
CLI: Better error reporting for missing patch file

dbrgn:
When a patch file - specified in the config - doesn't exist, the CLI error message is very opaque:

```
$ diesel print-schema
Encountered an IO error: No such file or directory (os error 2)
```

I had to clone the repo and add debug print statements until I found the culprit: The patch file.

This MR adds an explicit error message that can help to find the cause of the error.

```
$ diesel print-schema
Failed to read patch file at /.../src/schema.patch: No such file or directory (os error 2)
Encountered an IO error: No such file or directory (os error 2)
```

(By the way, it might be worthwhile to start using `anyhow` in the CLI project. It allows attaching context to errors, and printing them with a traceback. For CLI applications, this is a great improvement in my experience.)

### DIFF
--- a/diesel_cli/src/print_schema.rs
+++ b/diesel_cli/src/print_schema.rs
@@ -256,7 +256,17 @@ pub fn output_schema(
             "Found patch file to apply to the generated schema"
         );
         tracing::trace!(?out, "Schema before applying patch file");
-        let patch = std::fs::read_to_string(patch_file)?;
+        let patch = match std::fs::read_to_string(patch_file) {
+            Ok(patch) => patch,
+            Err(e) => {
+                eprintln!(
+                    "Failed to read patch file at {}: {}",
+                    patch_file.display(),
+                    e
+                );
+                return Err(e.into());
+            }
+        };
         let patch = diffy::Patch::from_str(&patch)?;
 
         out = diffy::apply(&out, &patch)?;


weiznich:
Thanks for submitting this PR. I think this addresses the particular issue, but does not cover the more general problem. (To be clear that's fine)

We recently refactored the error handling compared to the last release, so that we now have a central enum to record all the possible errors: https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/errors.rs#L22 I think a good approach here would be to just remove the `#[from]` annotation for `IoError` and enforce that we always set a path there for the corresponding file. I'm happy to accept a PR for that as well.

dbrgn:
> I think a good approach here would be to just remove the `#[from]` annotation for `IoError` and enforce that we always set a path there for the corresponding file. I'm happy to accept a PR for that as well.

That's a good approach too. Nevertheless, `anyhow` adds a lot of value for CLI crates, so that would be my recommendation, since you get proper tracebacks "for free" 🙂

weiznich:
My problem with that approach is that you still end up with `anyhow::Result` "everywhere" as that's what is returned buy the context method. That in turn makes it easy to forget calling context or to otherwise add additional information to the error as just using the ? operator is more convenient. 

That written: I wouldn't reject a PR that uses anyhow in a consistent manner throughout the diesel-cli crate.

PR #3977, opened 2024-04-05
Moveable binds and rows

momobel:
In the context of diesel_async, an async connection would also like to be used with sqlite. To enable creating an async connection wrapping a syn connection some types needs to be moveable and/or owned.

This PR adds support for moveable binds and owned row in core Diesel types as well as the sqlite connection and backend.

Support for postgres was tested at compile time but crashed due to invalid `Rc` constraints. That would need some deeper look to properly implement but doesn't seem to be blocking.

### DIFF
--- a/diesel/src/query_builder/bind_collector.rs
+++ b/diesel/src/query_builder/bind_collector.rs
@@ -46,6 +46,23 @@ pub trait BindCollector<'a, DB: TypeMetadata>: Sized {
     }
 }
 
+/// A movable version of the bind collector which allows it to be extracted, moved and refilled.
+///
+/// This is mostly useful in async context where bind data needs to be moved across threads.
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
+pub trait MoveableBindCollector<DB: TypeMetadata> {
+    /// The movable bind data of this bind collector
+    type BindData: Send + 'static;
+
+    /// Builds a movable version of the bind collector
+    fn moveable(&self) -> Self::BindData;
+
+    /// Refill the bind collector with its bind data
+    fn append_bind_data(&mut self, from: &Self::BindData);
+}
+
 #[derive(Debug)]
 /// A bind collector used by backends which transmit bind parameters as an
 /// opaque blob of bytes.
@@ -122,6 +139,26 @@ where
         self.metadata.push(metadata);
         self.binds.push(None);
         Ok(())
+    }
+}
+
+impl<DB> MoveableBindCollector<DB> for RawBytesBindCollector<DB>
+where
+    for<'a> DB: Backend<BindCollector<'a> = Self> + TypeMetadata + 'static,
+    <DB as TypeMetadata>::TypeMetadata: Clone + Send,
+{
+    type BindData = Self;
+
+    fn moveable(&self) -> Self::BindData {
+        RawBytesBindCollector {
+            binds: self.binds.clone(),
+            metadata: self.metadata.clone(),
+        }
+    }
+
+    fn append_bind_data(&mut self, from: &Self::BindData) {
+        self.binds.extend(from.binds.iter().cloned());
+        self.metadata.extend(from.metadata.clone());
     }
 }
 
--- a/diesel/src/query_builder/collected_query.rs
+++ b/diesel/src/query_builder/collected_query.rs
@@ -1,0 +1,51 @@
+use super::{AstPass, MoveableBindCollector, Query, QueryFragment, QueryId};
+use crate::backend::{Backend, DieselReserveSpecialization};
+use crate::result::QueryResult;
+use crate::sql_types::Untyped;
+
+#[derive(Debug)]
+#[must_use = "Queries are only executed when calling `load`, `get_result` or similar."]
+/// A SQL query variant with already collected bind data which can be moved
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
+pub struct CollectedQuery<T> {
+    sql: String,
+    safe_to_cache_prepared: bool,
+    bind_data: T,
+}
+
+impl<T> CollectedQuery<T> {
+    /// Builds a [CollectedQuery] with movable bind data
+    pub fn new(sql: String, safe_to_cache_prepared: bool, bind_data: T) -> Self {
+        Self {
+            sql,
+            safe_to_cache_prepared,
+            bind_data,
+        }
+    }
+}
+
+impl<DB, T> QueryFragment<DB> for CollectedQuery<T>
+where
+    DB: Backend + DieselReserveSpecialization,
+    for<'a> <DB as Backend>::BindCollector<'a>: MoveableBindCollector<DB, BindData = T>,
+{
+    fn walk_ast<'b>(&'b self, mut pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        if !self.safe_to_cache_prepared {
+            pass.unsafe_to_cache_prepared();
+        }
+        pass.push_sql(&self.sql);
+        pass.push_bind_collector_data::<T>(&self.bind_data)
+    }
+}
+
+impl<T> QueryId for CollectedQuery<T> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
+
+impl<T> Query for CollectedQuery<T> {
+    type SqlType = Untyped;
+}
--- a/diesel/src/row.rs
+++ b/diesel/src/row.rs
@@ -148,6 +148,18 @@ where
     }
 }
 
+/// A row that can be turned into an owned version
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
+pub trait IntoOwnedRow<'a, DB: Backend>: Row<'a, DB> {
+    /// The owned version of the row
+    type OwnedRow: Row<'a, DB> + Send + 'static;
+
+    /// Turn the row into its owned version
+    fn into_owned(self) -> Self::OwnedRow;
+}
+
 // These traits are not part of the public API
 // because:
 // * we want to control who can implement `Row` (for `RowSealed`)
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -28,7 +28,7 @@ use crate::expression::QueryMetadata;
 use crate::query_builder::*;
 use crate::result::*;
 use crate::serialize::ToSql;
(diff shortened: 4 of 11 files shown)


diesel/src/query_builder/bind_collector.rs
```diff
@@ -46,6 +46,23 @@ pub trait BindCollector<'a, DB: TypeMetadata>: Sized {
...
 
+/// Bind data from a BindCollector, that can be moved across threads
+pub trait MoveableBindData: Send + 'static {}
+
+/// A movable version of the bind collector which allows it to be extracted, moved and refilled.
+///
+/// This is mostly useful in async context where bind data needs to be moved across threads.
+pub trait MoveableBindCollector<DB: TypeMetadata> {
+    /// The movable bind data of this bind collector
+    type BindData: MoveableBindData;
```
weiznich, line 57: Can we just have `type BindData: Send + 'static;` instead here? Or do we need that `MovableBindData` trait for something else?

diesel/src/query_builder/bind_collector.rs
```diff
@@ -46,6 +46,23 @@ pub trait BindCollector<'a, DB: TypeMetadata>: Sized {
...
+
+    /// Builds a movable version of the bind collector
+    fn moveable(&self) -> Self::BindData;
+
+    /// Refill the bind collector with its bind data
+    fn fill_bind_data(&mut self, from: &Self::BindData);
```
weiznich, line 63: Maybe `append_bind_data` is a better name?

diesel/src/query_builder/bind_collector.rs
```diff
@@ -46,6 +46,23 @@ pub trait BindCollector<'a, DB: TypeMetadata>: Sized {
     }
 }
 
+/// Bind data from a BindCollector, that can be moved across threads
+pub trait MoveableBindData: Send + 'static {}
+
+/// A movable version of the bind collector which allows it to be extracted, moved and refilled.
+///
+/// This is mostly useful in async context where bind data needs to be moved across threads.
+pub trait MoveableBindCollector<DB: TypeMetadata> {
```
weiznich, line 55: I would like to gate that behind the `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature, as it's not useful for anyone outside of diesel-async/other crates doing really low level stuff.

diesel/src/query_builder/collected_query.rs
```diff
@@ -0,0 +1,48 @@
+use super::{AstPass, MoveableBindCollector, Query, QueryFragment, QueryId};
+use crate::backend::{Backend, DieselReserveSpecialization};
+use crate::result::QueryResult;
+use crate::sql_types::Untyped;
+
+#[derive(Debug)]
+#[must_use = "Queries are only executed when calling `load`, `get_result` or similar."]
+/// A SQL query variant with already collected bind data which can be moved
+pub struct CollectedQuery<T> {
```
weiznich, line 9: I would like to gate that behind the `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature, as it's not useful for anyone outside of diesel-async/other crates doing really low level stuff.

diesel/src/query_builder/collected_query.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl<T> QueryId for CollectedQuery<T> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
+
+impl<T> Query for CollectedQuery<T> {
+    type SqlType = Untyped; // FIXME set proper type of underlying original query
```
weiznich, line 47: We should be fine to just have a `p: PhantomData<ST>` field on the `CollectedQuery` struct which uses an additional generic type `ST`, which corresponds tho the sql type of the original query.

diesel/src/row.rs
```diff
@@ -148,6 +148,15 @@ where
     }
 }
 
+/// A row that can be turned into an owned version
+pub trait IntoOwnedRow<'a, DB: Backend> {
```
weiznich, line 152: I would like to gate that behind the `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature, as it's not useful for anyone outside of diesel-async/other crates doing really low level stuff.

Also maybe just make this a super trait of `Row`, so

```suggestion
pub trait IntoOwnedRow<'a, DB: Backend>: Row<'a, DB> {
```

diesel/src/sqlite/connection/row.rs
```diff
@@ -16,11 +17,19 @@ pub struct SqliteRow<'stmt, 'query> {
...
+        // starting point to create the OwnedSqliteRow
         values: Vec<Option<OwnedSqliteValue>>,
-        column_names: Rc<[Option<String>]>,
+        column_names: Rc<[Option<String>]>, // Switch to Box for the OwnedSqliteRow
     },
 }
 
+impl<'a> IntoOwnedRow<'a, Sqlite> for SqliteRow<'a, '_> {
+    type OwnedRow = OwnedSqliteRow;
+    fn into_owned(self) -> Self::OwnedRow {
```
weiznich, line 28: Minor style nit:
```suggestion

    fn into_owned(self) -> Self::OwnedRow {
```

diesel/src/sqlite/connection/row.rs
```diff
@@ -16,11 +17,19 @@ pub struct SqliteRow<'stmt, 'query> {
 pub(super) enum PrivateSqliteRow<'stmt, 'query> {
     Direct(StatementUse<'stmt, 'query>),
     Duplicated {
+        // starting point to create the OwnedSqliteRow
```
weiznich, line 20:
```suggestion
```

We don't need that work in process comment anymore, right?

diesel/src/sqlite/connection/row.rs
```diff
@@ -16,11 +17,19 @@ pub struct SqliteRow<'stmt, 'query> {
...
         values: Vec<Option<OwnedSqliteValue>>,
-        column_names: Rc<[Option<String>]>,
+        column_names: Rc<[Option<String>]>, // Switch to Box for the OwnedSqliteRow
```
weiznich, line 22:
```suggestion
        column_names: Rc<[Option<String>]>,
```

weiznich (commented):
Thanks for opening this PR and thanks for working on this. Overall, as already discussed the implementation looks good.

I left a round of feedback consisting of minor mostly stylistic things, otherwise this should be fine to merge.

diesel/src/query_builder/bind_collector.rs
momobel: It was a bit simplifying diesel_async trait bounds with `MD: MoveableBindData` instead of `MD: Send + 'static`. I went ahead and removed the `MoveableBindData` trait as suggested since it doesn't have much added value indeed.

diesel/src/query_builder/bind_collector.rs
momobel: Sure, renamed

diesel/src/sqlite/connection/row.rs
momobel: Cleaned

diesel/src/sqlite/connection/row.rs
momobel: Cleaned as well

diesel/src/query_builder/collected_query.rs
momobel: I am trying to figure out a solution here ([diesel branch `sql](https://github.com/WattSense/diesel/tree/sql_type_for_collected_query and https://github.com/WattSense/diesel_async/tree/collected_query_sqltype) but it needs bound on HasSqlType<T>.

weiznich: Right, that won't work. In that case: I think using `Untyped` there is fine, as it doesn't allow anything bad anyway. It just restricts that if you manage to construct that type you are responsible for deserializing results on your own (as we do in diesel-async) or that you use `QueryableByName`. 

So just removing that fixme should resolve this issue.

momobel: Thanks for the explanation, I was wondering how that works since it is possible to get structs from the load result.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -220,6 +224,13 @@ impl LoadConnection<DefaultLoadingMode> for SqliteConnection {
     }
 }
 
+impl WithMetadataLookup for SqliteConnection {
+    #[allow(unsafe_code)]
```
weiznich, line 228:
```suggestion
```

diesel/src/query_builder/bind_collector.rs
```diff
@@ -125,6 +142,26 @@ where
...
+    fn moveable(&self) -> Self::BindData {
+        RawBytesBindCollector {
+            binds: self.binds.clone(),
+            metadata: self.metadata.clone(),
+        }
+    }
+
+    fn append_bind_data(&mut self, from: &Self::BindData) {
+        self.binds = from.binds.clone();
+        self.metadata = from.metadata.clone();
```
weiznich, line 161:
```suggestion
        self.binds.extend(from.binds.cloned());
        self.metadata.extend(from.metadata.clone());
```

To make this behave similar than the sqlite implementation

diesel/src/sqlite/connection/row.rs
```diff
@@ -21,6 +22,14 @@ pub(super) enum PrivateSqliteRow<'stmt, 'query> {
     },
 }
 
+impl<'a> IntoOwnedRow<'a, Sqlite> for SqliteRow<'a, '_> {
```
weiznich, line 25:
```suggestion
impl<'stmt> IntoOwnedRow<'a, Sqlite> for SqliteRow<'stmt, '_> {
```

diesel/src/sqlite/connection/sqlite_value.rs
```diff
@@ -40,6 +42,19 @@ impl Drop for OwnedSqliteValue {
...
+
+impl From<&OwnedSqliteValue> for SqliteValue<'_, '_, '_> {
+    fn from(value: &OwnedSqliteValue) -> Self {
+        SqliteValue {
+            _row: None,
+            value: value.value,
+        }
+    }
+}
+
```
weiznich, line 57:
```suggestion
```

(It's not used anymore, right?)

weiznich (approved):
Looks good, beside the minor stylistic changes requested. 

(I will push them based on my suggestions directly after submitting this review)

diesel/src/sqlite/connection/sqlite_value.rs
momobel: Indeed, didn't catch this one!

diesel/src/query_builder/bind_collector.rs
momobel: Done, just minor change:

```
self.binds.extend(from.binds.iter().cloned())
```

weiznich (approved):

PR #3979, opened 2024-04-05
Alias op impls

formlogic-kirk:
Add derive of `DieselNumericOps` to `AliasedField` and a test.


weiznich (approved):
Thanks for putting together the PR

PR #3980, opened 2024-04-06
Help cargo to break dependency cycle caused by dev deps

kornelski:
Cargo can automatically remove `path` dependencies when publishing when they don't have a `version` specified. 

This way it should be possible to publish crates without editing their `Cargo.toml`, and therefore keep their [vcs metadata](https://github.com/rust-lang/cargo/issues/13695).

### DIFF
--- a/diesel_derives/Cargo.toml
+++ b/diesel_derives/Cargo.toml
@@ -23,7 +23,6 @@ cfg-if = "1"
 dotenvy = "0.15"
 
 [dev-dependencies.diesel]
-version = "~2.1.0"
 path = "../diesel"
 
 [lib]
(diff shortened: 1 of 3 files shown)


pksunkara (approved):

kornelski:
I don't see how it could break doc links. To have any undocumented types you'd have to use types from the dev dependency in a non-test build of the library, and that is not allowed and won't compile.

Ten0 (approved):
Ah right, I thought doclinks would be able to resolve through dev deps, but apparently they aren't (so cargo doc does yield an unresolved link warning)

okfm then ¯\_(ツ)_/¯

weiznich:
The failing CI is caused by a regression in rustc. I've filled https://github.com/rust-lang/rust/issues/123622 for that.

weiznich:
Would it be possible to rebase this PR on top of the latest master version to pull in the fixes for the CI setup? (Sorry for that…)

weiznich:
Thanks for submitting this PR ❤️

PR #3988, opened 2024-04-11
Fix Option field check in diesel_derives

### LINKED ISSUE #3987: Option field check in diesel_derives in incomplete
## Problem Description

The following code fails to compile:

```rust
diesel::table! {
    test_table (id) {
        id -> Int8,
        test -> Text,
    }
}

macro_rules! define {
    ($field_ty:ty) => {
        // Compilation error for S1
        #[derive(diesel::AsChangeset)]
        #[diesel(table_name = test_table)]
        pub struct S1 {
            pub test: $field_ty,
        }

        // Compilation is fine for S2
        #[derive(diesel::AsChangeset)]
        #[diesel(table_name = test_table)]
        pub struct S2 {
            pub test: Option<String>,
        }
    };
}

define!(Option<String>);
```
Playground link: https://www.rustexplorer.com/b/he8xgp

This is because the check for an `Option` field type in the proc-macro doesn't take into account the invisible delimiter introduced by the declarative macro:

https://github.com/diesel-rs/diesel/blob/5045fdb22307e3c7c8a8b9bd0b934dfa4736583b/diesel_derives/src/util.rs#L120-L138

stormshield-kg:
Fixes #3987.

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -117,8 +117,17 @@ pub fn is_option_ty(ty: &Type) -> bool {
     option_ty_arg(ty).is_some()
 }
 
-fn option_ty_arg(ty: &Type) -> Option<&Type> {
+fn option_ty_arg(mut ty: &Type) -> Option<&Type> {
     use syn::PathArguments::AngleBracketed;
+
+    // Check the inner equivalent type
+    loop {
+        match ty {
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. Would it be possible to add your reproduction case as test case to the diesel_derive tests?

weiznich (approved):
Thanks :+1:

PR #3989, opened 2024-04-12
Fix inserting values with `#[derive(Insertable)]` and `#[derive(MultiConnection)]`

weiznich:
This commit adds a missing impl that prevented inserts to `#[derive(MultiConnection)]` types if the corresponding struct used `#[derive(Insertable)]`


DavisRayM (approved):
Verified that this solves the missing `InsertValues<schema::users::table, multiconnection::multi_connection_impl::backend::MultiBackend>` implementation warning raised by the test when tested on the `master` branch. The derive macro seems to expand correctly IMO

PR #3990, opened 2024-04-12
Fix pq_sys::PQputCopyData call

stormshield-kg:
`c_char` is `u8` on `aarch64`, so this PR fixes a compilation error.


weiznich:
Thanks for providing this fix. I wonder why the M1 CI did not pick up that error. Do you build `pq-sys` with the bindgen feature enabled or do you use the bundled bindings?

weiznich (approved):

stormshield-kg:
I'm cross-compiling with `--target=aarch64-unknown-linux-gnu`.

PR #3991, opened 2024-04-12
sqlite-backend

amitu:
We are adding support for diesel/sqlite in wasm, but currently diesel's sqlite Backend implementation is tied to connection implementation, and have sqlite dependency. This PR is an attempt to split the sqlite backend from sqlite connection, as is done for PostgreSQL and MySQL already.


amitu:
I meant to open this PR in personal fork. Opened here by mistake. I will work on this first and once we are happy with it (and if we end up finishing anything) will open it here (assuming there is interest). Apologies for noise.

PR #4002, opened 2024-04-24
Add a way to skip generating some sql type definitions

Sagebati:
I run across a problem using some third party crates. with `generate_missing_sql_type_definitions`

- [pgvector](https://docs.rs/pgvector/latest/pgvector/) needs it set to false
- [diesel-derive-enum](https://github.com/adwhit/diesel-derive-enum) needs it set to true

So I added a `skip_missing_sql_type_definitions`, it takes a vec of regexes if any matches we ignore the type.

I hope this is valuable.


weiznich (commented):

I'm generally open to such a change but I would expect that it includes at least a few tests + passes the CI. Also we likely want to have a changelog entry for this addition as it might be useful for others as well?

Ten0:
Wouldn't updating PgVector be a simpler option that would avoid:
- sql type declarations in PgVector
- additional non-trivial options in Diesel
?

weiznich:
@Ten0 I fear it's not possible for crates like pgvetcor to do that in a helpful way. They need to define their own SQL type for:

* Implementing the necessary (de)serialization traits out of the box
* Provide custom DSL methods for expressions of that SQL type

I do not see how you could one or the other with arbitrary types provided by a downstream crate (from point of view of pgvector) without running into limitations due to the orphan rule.

For reference their [diesel implementation](https://docs.rs/crate/pgvector/latest/source/src/diesel_ext.rs). I would assume that this also affects other crates like [postgis-diesel](https://docs.rs/postgis_diesel/latest/postgis_diesel/)

Sagebati:
Thanks, I added a test and a CHANGELOG entry. but I struggle with the test. It feels like the diesel.toml in the test folder isn't read.

weiznich:
Ahh, right. I always forget this. The tests do 2 things:

* Test whether the expected code is generated via the option set in your `diesel.toml`. This part likely works fine
* Test whether the expected code is generated via the option set in the CLI interface. This part is likely missing and causes the test to fail

Sorry for not mentioning that earlier.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 ### Added
 
-* Support `[print_schema] skip_missing_sql_type_definitions=["Vector"]`. If a `missing type` matches one element on the list it's skipped.
+* Support `[print_schema] exclude_custom_sql_type_definitions=["Vector"]`. If a `missing type` matches one element on the list it's skipped.
```
Mingun: You name it `except_custom_sql_type_definitions`
```suggestion
* Support `[print_schema] except_custom_sql_type_definitions=["Vector"]`. If a `missing type` matches one element on the list it's skipped.
```

CHANGELOG.md
```diff
@@ -12,15 +12,22 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 ### Added
 
+* Support `[print_schema] exclude_custom_type_definitions=["Vector"]`. If a `custom type` matches one element on the list it's
+  skipped.
 * Added automatic usage of all sqlite `rowid` aliases when no explicit primary key is defined for `print-schema`
 * Added a `#[dsl::auto_type]` attribute macro, allowing to infer type of query fragment functions
-* Added the same type inference on `Selectable` derives, which allows skipping specifying `select_expression_type` most of the time, in turn enabling most queries to be written using just a `Selectable` derive.
-* Added an optional `#[diesel(skip_insertion)]` field attribute to the `Insertable` derive macro, allowing fields which map to generated columns to be skipped during insertion.
+* Added the same type inference on `Selectable` derives, which allows skipping specifying `select_expression_type` most
```
weiznich, line 19: I would rather like to not have all these unrelated changes in this file.

weiznich (commented):
Looks fine for me now beside the unrelated changes in the Changelog.md file. 

I will merge this likely next week to give @Ten0 the possibility to response to my answer above.

Ten0:
> to give @Ten0 the possibility to response to my answer above

Makes sense, didn't realize the types were so non-specific so yeah the resulting types are significantly more correct this way 👍😊

Sagebati:
Thanks, for your changes any clue why the CI is failing again ?

weiznich:
The CI errors are caused by https://github.com/rust-lang/rust/issues/124396. They are not related to this PR. I will fix them as soon as I'm back from holiday.

PR #4010, opened 2024-05-03
Improve error messages around AppearsInFromClause

Ten0:
Linked to #4008

Aliases to the same table not defined by the same `alias!` macro call turn out to actually not propagate up to `TableNotEqual`, instead they will lead to `AppearsInFromClause` erros directly.

This PR:
- Adds diagnostic notes to `AppearsInFromClause` unimplemented errors to account for the above
- Adds the corresponding compile test to `diesel_compile_tests/tests/fails/derive/aliases.rs`
- Removes mention of the aliases scenario from `TableNotEqual` (@weiznich does that look ok to you or did you actually encouter a case where it would propagate up to this?)
- Adds a bunch of `allow_tables_to_appear_in_same_query` in other compile tests where they were otherwise missing, but not the main source of error that we'd want to test there. This allows the error messages that we test to be more representative of what users would encounter, and otherwise makes the compilation not fail if something that should fail... fails to fail, successfully making the tests fail. 😁

### DIFF
--- a/diesel/src/query_source/mod.rs
+++ b/diesel/src/query_source/mod.rs
@@ -78,6 +78,10 @@ pub trait Table: QuerySource + AsQuery + Sized {
 /// context where referencing that table would be ambiguous (depending on the
 /// context and backend being used, this may or may not be something that would
 /// otherwise result in a runtime error).
+#[diagnostic::on_unimplemented(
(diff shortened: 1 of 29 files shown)


weiznich (approved):
Looks fine for me. Thanks 👍

As for the `on_unimplemented` messages: At least I consider them just as a starting point. We likely want to adjust dem later based on feedback from our users.

PR #4011, opened 2024-05-03
Add support of `dsl::select()` and `.into_sql()` to auto_type

Ten0:
This requires some tricky workarounds to get the type accessible where it should be due to name conflict with the `IntoSql` trait.


weiznich (approved):
Looks good, but we want to adjust the documentation to remove the statement that bare selects are not supported yet: https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/lib.rs#L1786

Issue #3996, opened 2024-04-19
The diesel 2.1.5 version cannot close the tls connection


12Saturn:
### Versions

- **Rust:*1.75.0*
- **Diesel:*2.1.5*
- **Database:*mysql 8.0.36*
- **Operating System:*mac*
- **Mysql client:*8.0.22*

## Problem Description
When using diesel to connect to mysql, I do not want to use SSL connection, but adding ‘ssl_mode=disabled’ does not work

conn mysql :
```rust
#[tokio::main]
async fn main() {
              let database_url = "mysql://root:123456@127.0.0.1:3306/test?ssl_mode=disabled";
              let mut tmp_conn = diesel::mysql::MysqlConnection::establish(&database_url)
                  .expect("Failed to connect db");
              tmp_conn.run_pending_migrations(MYSQL_MIGRATIONS).unwrap();
              let manager = AsyncDieselConnectionManager::<AsyncMysqlConnection>::new(database_url);
              let pool = MySqlPool::builder()
                  .build(manager)
                  .await
                  .expect("Failed to create db pool");
}
```

```cargo.toml
tokio = { version = "1.27.0", features = ["sync", "rt", "time", "macros", "rt-multi-thread"] }
diesel = { version = "2.1.5", features = ["mysql", "postgres", "serde_json", "chrono"] }
diesel-async = { version = "0.4.1", features = ["bb8", "postgres", "mysql"] }
diesel_migrations = "2.1.0"
```

 wifeshark Indicates the result of packet capture：

![sc_20240419163931](https://github.com/diesel-rs/diesel/assets/161667945/4221cc33-674c-4242-89b3-fa8aff6b5c6d)

weiznich:
Thanks for opening this bug report. Can you please complete the bug template as otherwise this report is not actionable for anyone due to missing details. Additionally you want to provide more details that just stating that "something" does not work. You might want to include additional details like a minimal reproducible example, information about your database system and your client libraries and your environment configuration.

12Saturn:
Thank you very much for your reply. I have added the test examples and test results

weiznich:
Thanks for adding the details. I'm sorry but there are still some important points missing until this becomes actionable:

* Your application uses connections from `diesel_async`  and `diesel`. Can you clarify if that problem happens with the connection from `diesel_async` or from `diesel`. In the former case that's the wrong issue tracker here.
[...]

Issue #3929, opened 2024-02-07
`Using unsupported buffer type: 253  (parameter: 1)` when trying to use the CLI


watsom27:
## Setup
- Latest version of `diesel_cli`
- MySql 8.3.0 running in docker

### Versions

- **Rust:** rustc 1.75.0 (82e1608df 2023-12-21)
- **Diesel:** 2.1.1
- **Database:** MySql 8.3.0
- **Operating System** MacOs

### Feature Flags

- **diesel:** mysql

## Problem Description
When running any commands using the `diesel` command, the error below is given:

```bash
$: diesel migration list
Using unsupported buffer type: 253  (parameter: 1)
```

Its worth noting that `diesel setup` _does_ create the database, but fails to run any migrations.

### What are you trying to accomplish?

Running migrations

### What is the expected output?

Migrations that run? Not sure what else to put here

### What is the actual output?

```bash
$: diesel setup
Creating migrations directory at: [my_directories]/migrations
Creating database: DieselTest
Failed to run migrations: Using unsupported buffer type: 253  (parameter: 1)
```

### Are you seeing any additional errors?
No

### Steps to reproduce
Repo: https://github.com/watsom27/diesel_bug

1. Spin up the MySql docker
```bash
docker compose up
```

2. Try diesel command
```bash
diesel setup
```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for opening this issue and adding these details. I would say it is expected that diesel works with the latest libmysqlclient version, therefore that's likely a bug. So I reopen this issue.

Can you verify a detail:  You can fix this by just using a older libmysqlclient version with the newer database? (So client 8.0, database 8.3)

watsom27:
> Can you verify a detail: You can fix this by just using a older libmysqlclient version with the newer database? (So client 8.0, database 8.3)

Correct, the 8.0 `libmysqlclient` seems to work with the 8.3 MySql docker image. Although I haven't done a full in depth test, but it creats the database and runs the migrations.

weiznich:
[...]

PR #4012, opened 2024-05-03
Add support for block expressions to #[dsl::auto_type]

Ten0:
This allows inline annotations of the form:
```rust
some_expression_that_auto_type_can_infer({ let x: TypeThatAutoTypeCantInfer = abc(); x })
```

That will work even with `select_expression` in `Selectable` implementations, which before didn't provide a way to provide type annotations.

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -558,13 +558,22 @@ where
 ///    or a custom SQL function.
 ///
 /// The select expression is specified via the `select_expression` parameter.
-/// For more complex expressions, it may be required to also specify the result type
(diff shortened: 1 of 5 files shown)


weiznich (approved):

Ten0:
auto_type can probably go in the next release, as it's unlikely that it needs significant changes 🙂

weiznich:
That's good to know. I plan to do a feature release "soon". The only thing left on my side is figuring out how to build a bundled libmysqlclient version on windows + update diesel to depend on that new version of mysqlclient to fix the bindings bugs (#3996 and #3929). After that I think we are ready to go for diesel 2.2 :tada:

PR #4014, opened 2024-05-05
Adds unsigned type for post id received from the terminal in examples

rnambaale:
The idea of this pull request is to let Rust's type system communicate that a valid ID can't be signed.

### DIFF
--- a/examples/mysql/getting_started_step_3/src/bin/publish_post.rs
+++ b/examples/mysql/getting_started_step_3/src/bin/publish_post.rs
@@ -9,7 +9,7 @@ fn main() {
     let id = args()
         .nth(1)
         .expect("publish_post requires a post id")
-        .parse::<i32>()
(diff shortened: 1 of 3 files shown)


rnambaale:
> I'm sorry but the ID is in fact allowed to be negative. There is no problem with that in the current code. I fact it's not allowed to be a `u32` as that does not fit in the corresponding database type (as that one is signed as well).

@weiznich Thanks for your comment.

Should I go ahead and modify the database type to make it unsigned and subsequently prevent negative ids downstream, is that a change you would be happy to have?

weiznich:
I don't think that's helpful either as most database system do not support unsigned integer types at all. The only supported database system that happens to support them is MySQL and even there it's kind of a niche feature.

rnambaale:
Thanks, I will close then.

PR #4015, opened 2024-05-06
diesel_derives: specify `derive_from_sql_row()` docs

bachmannscode:
This change is about a terminology detail that may slightly mislead some library users. The doc comment currently says that this derive implements `Queryable` for primitive types. "Primitive type" in that context is supposed to refer to a rust type that corresponds to a single SQL type in contrast to a whole database row.

Unfortunately the current term collides with actual Rust types that have different meaning [1] so we change it to the regular Rust term [2] instead.

[1] https://doc.rust-lang.org/rust-by-example/primitives.html
[2] https://doc.rust-lang.org/rust-by-example/generics/new_types.html


weiznich:
Thanks for opening this PR. I agree with you that "primitive type" could refer to the wrong thing there, but I fear that "newtype struct" is even more incorrect as the derive can also be used on different rust types like enums or struct (for representing a sql record type or jsonb type).

bachmannscode:
> Thanks for opening this PR. I agree with you that "primitive type" could refer to the wrong thing there, but I fear that "newtype struct" is even more incorrect as the derive can also be used on different rust types like enums or struct (for representing a sql record type or jsonb type).

Thank you for the review. True, I totally missed that. Do you think one of those could improve the docs:
```
Implements `Queryable` for types that correspond to a single SQL type. The type has to implement `FromSql`.
```
or
```
Implements `Queryable` for types that correspond to a single SQL type (for ones that implement `FromSql`).
```
?

weiznich:
I think the first variant sounds better.

bachmannscode:
> I think the first variant sounds better.

Ack, done.

weiznich (approved):

weiznich:
Thanks 👍

bachmannscode:
Thank you 💪

PR #4019, opened 2024-05-07
When using `.optional()`, convert a `QueryBuilderError` into `None`

dessalines:
This is my first diesel PR, I apologize in advance for any issues. Let me know of any documentation I missed, tests I should add, or anything I can do to improve it.

I also had to comment a few tests in `bin/test` to get this to pass, which leads me to believe the `main` branch has some unaddressed errors which might fail CI before reaching the test I added.

Related discussion: https://github.com/diesel-rs/diesel/discussions/4018

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -45,8 +45,6 @@ pub enum Error {
     /// An example of when this error could occur is if you are attempting to
     /// construct an update statement with no changes (e.g. all fields on the
     /// struct are `None`).
-    ///
-    /// When using `optional_empty_changeset`, this error is turned into `None`.
     QueryBuilderError(Box<dyn StdError + Send + Sync>),
 
     /// An error occurred deserializing the data being sent to the database.
@@ -441,7 +439,9 @@ impl fmt::Display for UnexpectedEndOfRow {
 
 impl StdError for UnexpectedEndOfRow {}
 
-/// Expected when an update has no changes to save
+/// Expected when an update has no changes to save.
+///
+/// When using `optional_empty_changeset`, this error is turned into `None`.
 #[derive(Debug, Clone, Copy)]
 pub struct EmptyChangeset;
 
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -45,6 +45,8 @@ pub enum Error {
     /// An example of when this error could occur is if you are attempting to
     /// construct an update statement with no changes (e.g. all fields on the
     /// struct are `None`).
+    ///
+    /// When using [`optional`], this error is turned into `None`.
     QueryBuilderError(Box<dyn StdError + Send + Sync>),
 
     /// An error occurred deserializing the data being sent to the database.
@@ -265,15 +267,21 @@ pub trait OptionalExtension<T> {
     /// [`get_result`]: crate::query_dsl::RunQueryDsl::get_result()
     /// [`first`]: crate::query_dsl::RunQueryDsl::first()
     ///
+    /// Also by default, Diesel treats an empty update as a `QueryBuilderError`. This method will
+    /// convert that error into `None`.
+    ///
     /// # Example
     ///
     /// ```rust
-    /// use diesel::{QueryResult, NotFound, OptionalExtension};
+    /// use diesel::{QueryResult, NotFound, OptionalExtension, QueryBuilderError};
     ///
     /// let result: QueryResult<i32> = Ok(1);
     /// assert_eq!(Ok(Some(1)), result.optional());
     ///
     /// let result: QueryResult<i32> = Err(NotFound);
+    /// assert_eq!(Ok(None), result.optional());
+
+    /// let result: QueryResult<i32> = Err(QueryBuilderError);
     /// assert_eq!(Ok(None), result.optional());
     /// ```
     fn optional(self) -> Result<Option<T>, Error>;
@@ -284,6 +292,7 @@ impl<T> OptionalExtension<T> for QueryResult<T> {
         match self {
             Ok(value) => Ok(Some(value)),
             Err(Error::NotFound) => Ok(None),
+            Err(Error::QueryBuilderError(_)) => Ok(None),
             Err(e) => Err(e),
         }
     }
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -46,7 +46,7 @@ pub enum Error {
     /// construct an update statement with no changes (e.g. all fields on the
     /// struct are `None`).
     ///
-    /// When using [`optional`], this error is turned into `None`.
+    /// When using `optional`, this error is turned into `None`.
     QueryBuilderError(Box<dyn StdError + Send + Sync>),
 
     /// An error occurred deserializing the data being sent to the database.
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -288,7 +288,7 @@ pub trait OptionalExtension<T> {
     /// ```rust
     /// use diesel::{QueryResult, OptionalExtension, result::Error::QueryBuilderError, result::EmptyChangeset};
     /// let result: QueryResult<i32> = Err(QueryBuilderError(Box::new(EmptyChangeset)));
-    /// assert_eq!(Ok(None), result.optional());
+    /// assert_eq!(Ok(None), result.optional_empty_changeset());
     /// ```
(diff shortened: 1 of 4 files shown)


diesel_tests/tests/update.rs
```diff
@@ -234,6 +234,28 @@ fn update_with_no_changes() {
         .unwrap();
 }
 
+#[test]
```
dessalines, line 237: Since this test is just copied from the one above, it might make sense to add the new line just above the panic one above. Or just keep it as its own test, up to you.

weiznich (changes requested):
The structure of the change looks ok, although I'm quite unhappy with the new matching there as it allows a potentially large number of different error. That's not something I would like to support there.

diesel/src/result.rs
```diff
@@ -284,6 +292,7 @@ impl<T> OptionalExtension<T> for QueryResult<T> {
         match self {
             Ok(value) => Ok(Some(value)),
             Err(Error::NotFound) => Ok(None),
+            Err(Error::QueryBuilderError(_)) => Ok(None),
```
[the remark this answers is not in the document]
Ten0: If we want to make that into an error that we may match on, then we probably want to turn it into a some form of typed error: it's less error prone, and does not imply the additional runtime cost that string comparisons would imply.

The type already allows that: 
image

So if it's currently a `String` (or `&'static str`?) that we then turn into a `Box<dyn StdError + ...>`, we probably want to create a dedicated type for it (with the same message as before in its Display impl) and match it with `.is::<_>()`, as is done with `UnexpectedNullError` in the `FromSql` code for `Option`.

Ten0 (commented):
Thanks for tackling this topic and opening this PR! 😊
It does indeed look like it would be useful to be able to catch this error in an ergonomic manner. 🙂

However mixing it within the existing `.optional()` extension feels odd to me: it feels like it might make it harder to use `.optional()` without mixing in potential unrelated/unexpected scenarios into `None`.
I find that it is typically better to avoid falling back from larger error categories than may really happen, as this may turn into silent fails, resulting in hard-to-debug issues.

As far as I'm concerned we have many use-cases where we use the `update(table.find(id)).filter(make_sure_that_state_is_ok()).optional()` combination, where `make_sure_that_state_is_ok()` is for us to get single-query happy path, and then if `None` re-query to know whether the ID was wrong or the state was wrong and return the appropriate user error.

However we have no use-case such as the one described by this issue.

This tends to make me believe that it's better to be able to guarantee that `.optional` really only catches the scenario I'm personally always looking for where `make_sure_that_state_is_ok()` is `FALSE`, to avoid potentially enforcing upon users the unexpected silent fail scenario described above, so I'd probably prefer if it stayed dedicated to the current meaning it has. I think a design that wouldn't have this downside of too-large-matching, but would still have the same ergonomics as the solution currently proposed by this PR, would be introducing a separate function similar to `.optional()` that catches precisely that *empty changeset* scenario.

However as weiznich pointed out [here](https://github.com/diesel-rs/diesel/discussions/4018#discussioncomment-9345516), `.optional()` is just an extension trait that is reasonably easy to re-create outside of Diesel. (We do have a few similar extension traits ourselves, such as `.get_only_result()` which is a `get_result` that errors if there are multiple results, which is an extremely common case for us, prevents bugs where one would forget to write the appropriate `.find` or `.filter` clause (again we prefer hard errors to silent fails)... and have found that having that in our `diesel_helpers` crate doesn't generate any friction.)
So overall I have no strong opinion on whether the extension trait itself should even live within Diesel itself - I guess it depends on really how common this use-case is, as this is a compromise between providing the feature and API&Doc size/maintenance cost, and all I can say is I personally haven't encountered this use-case.

However what seems pretty clear that we should do within Diesel is adding a typed error for this and exporting it in a similar way to [`UnexpectedNullError`](https://github.com/diesel-rs/diesel/blob/df896872d05a6bfb794c3435d9002bb710d7e327/diesel/src/result.rs#L393-L403), so that users *can* actually match this error and construct any extension trait related to this that they see fit, in an idiomatic and ~typed manner, (that is, not by matching on a string which the API does not guarantee may not change and is slow).

diesel/src/result.rs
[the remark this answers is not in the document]
dessalines: I was thinking the same, I'd be glad to do this if its desired.

Ten0:
Ok so next steps for this PR:
- Create the new error type and have the relevant code export it
- Get opinion from @diesel-rs/reviewers on whether it's worth including the extension trait in Diesel itself
- If so, add that as well

Ten0:
Sorry, missclick

diesel/src/result.rs
```diff
@@ -287,6 +301,13 @@ impl<T> OptionalExtension<T> for QueryResult<T> {
             Err(e) => Err(e),
         }
     }
+    fn optional_empty_changeset(self) -> Result<Option<T>, Error> {
```
dessalines, line 304: I just used the same trait here, hope that's okay.

diesel/src/result.rs
```diff
@@ -413,3 +434,18 @@ impl fmt::Display for UnexpectedEndOfRow {
 }
 
 impl StdError for UnexpectedEndOfRow {}
+
+/// Expected when an update has no changes to save
+#[derive(Debug, Clone, Copy)]
+pub struct EmptyChangeset;
```
dessalines, line 440: Not sure if set should be capitalized or not.

diesel_tests/tests/update.rs
```diff
@@ -228,10 +227,30 @@ fn update_with_no_changes() {
         name: None,
         hair_color: None,
     };
-    update(users::table)
+    let update_result = update(users::table).set(&changes).execute(connection);
+    assert!(update_result.is_err());
```
dessalines, line 231: Changed this from a panic check, to an `is_err` check, since this now returns `QueryBuilder(EmptyChangeset)`

diesel/src/result.rs
weiznich: I fear that this is a breaking change, as someone could implement `OptionalExtension` for their own types, which in turn would then not implement the new method.

(I doubt that anyone will do this, but better be safe than sorry)

diesel_tests/tests/update.rs
weiznich: It's better to have a separate test for this. These tests are quite fast so there is no point to worry about anything.

weiznich (commented):
Looks fine for me beside the comment about the breaking change to the `OptionalExtension` trait. I think it's safer to just introduce a new trait for that + reexport that from the prelude.

diesel/src/result.rs
dessalines: No problem, did this now.

diesel/src/result.rs
```diff
@@ -45,6 +45,8 @@ pub enum Error {
     /// An example of when this error could occur is if you are attempting to
     /// construct an update statement with no changes (e.g. all fields on the
     /// struct are `None`).
+    ///
+    /// When using `optional_empty_changeset`, this error is turned into `None`.
```
weiznich, line 49: Minor note: That comment is not correct anymore. It might be a got idea to move this comment to the new error type.

weiznich (approved):
Looks good for me with the small documentation fix applied.

diesel/src/result.rs
dessalines: K, will do.

diesel/src/result.rs
```diff
@@ -268,7 +268,7 @@ pub trait OptionalExtension<T> {
     /// # Example
     ///
     /// ```rust
-    /// use diesel::{QueryResult, NotFound, OptionalExtension};
+    /// use diesel::{QueryResult, NotFound, OptionalExtension, result::Error::QueryBuilderError};
```
weiznich, line 271:
```suggestion
    /// use diesel::{QueryResult, NotFound, OptionalExtension};
```

weiznich (approved):

JMLX42:
@weiznich can we please have a minor release for this?

weiznich:
@JMLX42 As this is a new feature it needs to wait until the next diesel feature release happens. As for the timeline for such a release: This is a free time project, so that will happen as soon as it is ready. We generally do not give ETA's there. You can contribute or sponsor the development to make such a release happen faster.

PR #4021, opened 2024-05-11
Postgres Citext Support

tommymcc:
Adds support for the [Postgres citext type](https://www.postgresql.org/docs/current/citext.html).

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -561,6 +561,27 @@ pub mod sql_types {
     #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
     #[diesel(postgres_type(oid = 18, array_oid = 1002))]
     pub struct CChar;
+
+    /// The [`Citext`] SQL type. This is a PostgreSQL specific type.
+    ///
+    /// Strings must be valid UTF-8.
+    ///
+    /// ### [`ToSql`] impls
+    ///
+    /// - [`String`]
+    /// - [`&str`][str]
+    ///
+    /// ### [`FromSql`] impls
+    ///
+    /// - [`String`]
+    ///
+    /// [`ToSql`]: crate::serialize::ToSql
+    /// [`FromSql`]: crate::deserialize::FromSql
+    /// [`Citext`]: https://www.postgresql.org/docs/current/citext.html
+    #[cfg(feature = "postgres_backend")]
+    #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
+    #[diesel(postgres_type(name = "citext"))]
+    pub struct Citext;
 }
 
 mod ops {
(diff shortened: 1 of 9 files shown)


tommymcc:
Apologies, I'd created this PR against the Diesel repository rather than my own fork by accident. I had meant to ask in Gitter whether it would be acceptable first.

diesel_tests/tests/types.rs
```diff
@@ -1508,3 +1512,65 @@ fn cchar_to_sql() {
...
+            non_null_ci -> Citext,
+            nullable_ci -> Nullable<Citext>,
+            null_value -> Nullable<Citext>,
+        }
+    }
+
+    let rows_inserted = insert_into(case_insensitive::table)
+        .values((
+            case_insensitive::non_null_ci.eq("UPPERCASE_VALUE".to_string()),
+            case_insensitive::nullable_ci.eq("lowercase_value".to_string()),
```
Ten0, line 1548: Maybe we can put not a string but an &str on this one so that both types are tested?

diesel/src/pg/types/mod.rs
```diff
@@ -561,6 +561,27 @@ pub mod sql_types {
...
+    ///
+    /// - [`String`]
+    ///
+    /// [`ToSql`]: crate::serialize::ToSql
+    /// [`FromSql`]: crate::deserialize::FromSql
+    /// [`Citext`]: https://www.postgresql.org/docs/current/citext.html
+    #[cfg(feature = "postgres_backend")]
+    #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
+    #[diesel(postgres_type(name = "citext"))]
+    pub struct Citext;
```
Ten0, line 584: I think we need to add this type to the list of types that are known by diesel in the cli when generating the schema, so that the cli doesn't generate the corresponding type in the schema.rs.

Im surprised you haven't encountered issues due to this during tests.
Did you run tests using the cli yet?

tommymcc: Thanks, I'll have a look into that. I started working backwards using a schema generated without Citext support, I still have to test starting with schema generation. I'll check that and add get back to you with the results.

As I mentioned in Gitter, I'd jumped the gun a little on creating the PR before I had a chance to test properly but I'll  make sure it all works before pushing any more changes.

diesel_tests/tests/types.rs
tommymcc: Sounds like a good call :+1:

diesel/src/pg/types/mod.rs
tommymcc: Nice catch, I've updated the list of known types to include `Citext` and added a test to validate it generates the correct schema. Let me know if there's a better place I could put it. 

I've also tested with our application in work and the test suite (which include case insensitively fetching a record by a `citext` field) passes.

Are there any other areas I need to consider?

weiznich (approved):

PR #4025, opened 2024-05-14
Support equality coercions, in particular in implicit joins

Ten0:
Not sure to what extent we want this because it's maybe a bit edge-case to support.

I've needed it because I'm joining on a partially-migrated schema that has FK types as `i64` and source table as `i32`, but I'd be fine with just having that merged in our fork until it's fully migrated.

Nonetheless it's implemented so I'm opening this draft PR to gather feedback on whether this is something that we want to support or whether it's not worth it.

### DIFF
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -25,6 +25,9 @@ pub type AsExprOf<Item, Type> = <Item as AsExpression<Type>>::Expression;
 /// [`lhs.eq(rhs)`](crate::expression_methods::ExpressionMethods::eq())
 pub type Eq<Lhs, Rhs> = Grouped<super::operators::Eq<Lhs, AsExpr<Rhs, Lhs>>>;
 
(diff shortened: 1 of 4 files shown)


weiznich:
I think I would rather prefer having a more general solution here. Something along the lines like 

```rust
impl<T> AsExpression<Int8> for T
where T: Expression<SqlType = Int4> {
 // …
}
```
That particular impl runs into a conflicting implementation issue, which might be possible to solve. We would also need to investigate whether that causes type inference issues in some situations.

Another alternative is to expose a better structured variant of the internal coerce expression:

https://github.com/diesel-rs/diesel/blob/16c4cf78dda73a7f72d5a7735bc3c79c7a39848a/diesel/src/expression/coerce.rs

Mingun:
Interesting... why a `Self: Expression` bound is needed here, the `Expression` implementation few lines above is not enough?
https://github.com/diesel-rs/diesel/blob/16c4cf78dda73a7f72d5a7735bc3c79c7a39848a/diesel/src/expression/coerce.rs#L42-L54

PR #4030, opened 2024-05-18
Allow raw identifiers for SqlIdentifier (column-name)

z33ky:
This allows using the `r#identifier` syntax for SqlIdentifier (column-name), which is used in derive macros.
```rust
table! {
    things {
        id -> Integer,
        r#identifier -> Integer,
    }
}

#[derive(diesel::Queryable)]
struct Thing {
    id: i32,
    r#identifier: i32
}
```
Previously the derive macros would panic when encountering such an identifier:
 >```   `"r#identifier"` is not a valid identifier```

### DIFF
--- a/diesel_derives/src/attrs.rs
+++ b/diesel_derives/src/attrs.rs
@@ -62,7 +62,7 @@ impl SqlIdentifier {
                 ident.set_span(self.span);
                 Ok(ident)
             }
-            Err(_e) if self.field_name.contains(" ") => Err(syn::Error::new(
+            Err(_e) if self.field_name.contains(' ') => Err(syn::Error::new(
                 self.span(),
                 format!(
                     "Expected valid identifier, found `{0}`. \
--- a/diesel_derives/src/attrs.rs
+++ b/diesel_derives/src/attrs.rs
@@ -62,6 +62,14 @@ impl SqlIdentifier {
                 ident.set_span(self.span);
                 Ok(ident)
             }
+            Err(_e) if self.field_name.contains(" ") => Err(syn::Error::new(
+                self.span(),
+                format!(
+                    "Expected valid identifier, found `{0}`. \
+                 Diesel does not support column names with whitespaces yet",
+                    self.field_name
+                ),
+            )),
             Err(_e) => Err(syn::Error::new(
                 self.span(),
                 format!(
--- a/diesel_derives/src/attrs.rs
+++ b/diesel_derives/src/attrs.rs
@@ -56,9 +56,13 @@ impl SqlIdentifier {
         self.span
     }
 
-    pub fn valid_ident(&self) -> Result<()> {
-        if syn::parse_str::<Ident>(&self.field_name).is_err() {
-            Err(syn::Error::new(
+    pub fn to_ident(&self) -> Result<Ident> {
+        match syn::parse_str::<Ident>(&format!("r#{}", self.field_name)) {
+            Ok(mut ident) => {
+                ident.set_span(self.span);
+                Ok(ident)
+            }
+            Err(_e) => Err(syn::Error::new(
                 self.span(),
                 format!(
                     "Expected valid identifier, found `{0}`. \
@@ -66,22 +70,28 @@ impl SqlIdentifier {
                  perhaps you meant to write `{0}_`?",
                     self.field_name
                 ),
-            ))
-        } else {
-            Ok(())
+            )),
         }
     }
 }
 
 impl ToTokens for SqlIdentifier {
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        Ident::new(&self.field_name, self.span).to_tokens(tokens)
+        if self.field_name.starts_with("r#") {
+            Ident::new_raw(&self.field_name[2..], self.span).to_tokens(tokens)
+        } else {
+            Ident::new(&self.field_name, self.span).to_tokens(tokens)
+        }
     }
 }
 
 impl Display for SqlIdentifier {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        f.write_str(&self.field_name)
+        let mut start = 0;
+        if self.field_name.starts_with("r#") {
+            start = 2;
+        }
+        f.write_str(&self.field_name[start..])
     }
 }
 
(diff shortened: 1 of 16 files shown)


weiznich (changes requested):
Thanks for opening this PR. I think the change itself looks fine, but we want to have some tests that show that everything is parsed correctly.

z33ky:
Sure. I added it for the `Queryable` test, since that probably the simplest. Is that sufficient?
Trying to run the tests locally didn't work, I'm getting lots of compiler errors (also without my change). Here's to hoping it runs just like that :D

weiznich:
I think I would prefer to have these kind of test for `Insertable`, `AsChangeset`, `Selectable` and `QueryableByName` as well.

z33ky:
Alright, I added raw identifiers to the respective test cases.
Sorry for the force-push spam due to my local setup being unable to execute the tests locally.
The spell-check complains about the Spanish "tipe", though Spanish is also used for the other names of that struct. I'm not sure what's up with that or how I would go about solving this.

weiznich:
No worries about the force pushes, that's fine.

For the spell checker: You can just add an exception here: https://github.com/diesel-rs/diesel/blob/master/.typos.toml#L8

As for running the tests: There is a bash script in `bin/test` that should allow to execute all tests as long as you've installed the build dependencies and configured the required database urls.

diesel_derives/tests/queryable_by_name.rs
```diff
@@ -18,11 +19,19 @@ fn named_struct_definition() {
     struct MyStruct {
         foo: i32,
         bar: i32,
+        r#type: i32,
     }
 
     let conn = &mut connection();
-    let data = sql_query("SELECT 1 AS foo, 2 AS bar").get_result(conn);
-    assert_eq!(Ok(MyStruct { foo: 1, bar: 2 }), data);
+    let data = sql_query("SELECT 1 AS foo, 2 AS bar, 3 AS `r#type`").get_result(conn);
```
z33ky, line 26: QueryableByName perhaps poses an interesting question. Should the query indeed contain `` `r#type` `` to match or should the SQL query contain `type` and map to the raw identifier `r#type`?

z33ky:
> As for running the tests: There is a bash script in bin/test that should allow to execute all tests as long as you've installed the build dependencies and configured the required database urls.

Ah thanks, that works.

diesel_derives/tests/queryable_by_name.rs
weiznich: I would expect `type` there.

diesel_derives/tests/schema.rs
```diff
@@ -11,5 +12,6 @@ table! {
         id -> Integer,
         name -> Text,
         hair_color -> Nullable<Text>,
+        r#type -> Nullable<Text>,
```
Ten0, line 15: Note that the CLI currently would generate `type_` in such scenarios:
```rust
table! {
	things(id) {
		id -> Int4,
		#[sql_name = "type"]
		type_ -> Int4,
	}
}
```

(I understand that this only adds support and does not change behavior of the CLI)

diesel_derives/tests/queryable_by_name.rs
```diff
@@ -132,15 +170,29 @@ fn embedded_option() {
...
 
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, QueryableByName)]
+    #[diesel(table_name = my_structs)]
+    struct T {
+        r#type: i32,
+    }
+
     let conn = &mut connection();
-    let data = sql_query("SELECT 1 AS foo, 2 AS bar").get_result(conn);
+    let data = sql_query("SELECT 1 AS foo, 2 AS bar, 3 AS \"r#type\"").get_result(conn);
```
Ten0, line 180: I feel like the above QueryableByName should try to read `type`, not `r#type`.

diesel_derives/tests/queryable_by_name.rs
```diff
@@ -32,11 +41,12 @@ fn tuple_struct() {
     struct MyStruct(
         #[diesel(column_name = foo)] i32,
         #[diesel(column_name = bar)] i32,
+        #[diesel(column_name = r#type)] i32,
     );
 
     let conn = &mut connection();
-    let data = sql_query("SELECT 1 AS foo, 2 AS bar").get_result(conn);
-    assert_eq!(Ok(MyStruct(1, 2)), data);
+    let data = sql_query("SELECT 1 AS foo, 2 AS bar, 3 AS \"r#type\"").get_result(conn);
```
Ten0, line 48: I think to be consistent with previous comment we probably should search for `type` as name if passing the `r#type` ident

and ideally also allow passing `column_name = "some arbitrary string"` where we would then parse `3 as "some arbitrary string"` even if `some arbitrary string` is not an ident to support other weird things users might want. (I'm not sure if that is already supported and it's not a blocker here)

z33ky: I have lifted the restriction on `column_name` insofar, that a raw identifier is now allowed, however it must still correspond to a valid Rust identifier.
The previous error-message for invalid names is retained, though it now emits an unhelpful
```
error: Expected valid identifier, found `ty pe`. Diesel automatically renames invalid identifiers, perhaps you meant to write `ty pe_`?
   --> diesel_derives/tests/as_changeset.rs:287:32
    |
287 |         #[diesel(column_name = "ty pe")]
    |                                ^^^^^^^
```
The error-message should probably use `inference::rust_name_for_sql_name()`?

diesel_derives/tests/schema.rs
z33ky: Should diesel_cli be changed to emit `r#type` in this case now?

Ten0: I think adding the underscore is typically more practical.

diesel_derives/src/attrs.rs
```diff
@@ -56,32 +56,50 @@ impl SqlIdentifier {
...
-    pub fn valid_ident(&self) -> Result<()> {
-        if syn::parse_str::<Ident>(&self.field_name).is_err() {
-            Err(syn::Error::new(
+    pub fn to_ident(&self) -> Result<Ident> {
+        match syn::parse_str::<Ident>(&format!("r#{}", self.field_name)) {
+            Ok(mut ident) => {
+                ident.set_span(self.span);
+                Ok(ident)
+            }
+            Err(_e) if self.field_name.contains(" ") => Err(syn::Error::new(
```
weiznich, line 65:
```suggestion
            Err(_e) if self.field_name.contains(' ') => Err(syn::Error::new(
```

Issue #4035, opened 2024-05-22
Investigate new nightly warning about a soundness hole in the type system


weiznich:
### Versions

- **Rust:** 1.80 (nightly)
- **Diesel:** master

## Problem Description

We get the following new warning for builds of `diesel-cli` on rust nightly:

```
warning[E0210]: type parameter `Col` must be covered by another type when it appears before the first local type (`database::multi_connection_impl::backend::MultiBackend`)
   --> diesel_cli/src/database.rs:109:10
    |
109 | #[derive(diesel::MultiConnection)]
    |          ^^^^^^^^^^^^^^^^^^^^^^^ type parameter `Col` must be covered by another type when it appears before the first local type (`database::multi_connection_impl::backend::MultiBackend`)
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #124559 <https://github.com/rust-lang/rust/issues/124559>
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
    = note: in this case, 'before' refers to the following order: `impl<..> ForeignTrait<T1, ..., Tn> for T0`, where `T0` is the first and `Tn` is the last
    = note: `#[warn(uncovered_param_in_projection)]` on by default
    = note: this warning originates in the derive macro `diesel::MultiConnection` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### What is the expected output?

Builds without warning

Related: https://github.com/rust-lang/rust/issues/124559

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [ ] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

PR #4043, opened 2024-05-24
Prepare a diesel 2.2 release

weiznich:
Release announcements: https://github.com/sgrif/diesel.rs-website/pull/219

We likely should address https://github.com/diesel-rs/diesel/issues/4035 before cutting the release, but I wanted to prepare everything else anyway as I'm waiting on feedback from the relevant compiler contributors there.

If everything goes well I would like to issue a release in the next two weeks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ concurrency:
 jobs:
   check_and_test:
     name: Check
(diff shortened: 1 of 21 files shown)


weiznich:
@diesel-rs/core Any opinions on this? I would rather release this sooner than later to address the mysql bindings issues.

Ten0:
Only had a quick glance but what's mentioned in the changelog looks OK to me 😊
Thanks! 🙏

PR #4054, opened 2024-05-31
Fix lifetimes with `#[auto_type]`

weiznich:
This commit fixes using `#[auto_type]` with functions with reference arguments. We now check the return type if any lifetime is contained there and if that's the case we require that it is a named lifetime from one of the function inputs. This lifetime is then correctly used in the generated return type.

cc @Ten0 

There is another issue with static functions:

For example using `Selectable` currently fails.

```rust
#[derive(Selectable)]
struct User {
    id: i32,
}

#[auto_type]
fn selectable() -> _ {
    users::table.select(User::as_select())
}

```


Ten0:
Currently working on this, I've got a slightly different design using `syn::visit` that supports the full tree and generics with no more code.

weiznich:
Right, I totally forgot that this already exists in syn :see_no_evil: 
But yes that should be the better solution.

Ten0:
Aah I meant I had the full implementation ~ready

Ten0:
https://github.com/Ten0/diesel/tree/fix/auto_type_lifetimes

weiznich:
Sorry, I misunderstood that. 
Yes please force push your version here as it seems to be more complete than my version.

PR #4061, opened 2024-06-07
Port the relations to example to sqlite

### LINKED ISSUE #4060: `belonging_to().inner_join()` and  `.select()` with two-item tuple causes build error on sqlite backend
### Versions

- **Rust: 1.78.0**
- **Diesel: 2.2.0**
- **Database: sqlite**
- **Operating System: debian trixie**

### Feature Flags

- **diesel: sqlite**

## Problem Description

i'm following along with the "relations" tutorial using an sqlite database, and i'm running into a compile error when using a tuple for `.select()` as shown here: https://github.com/diesel-rs/diesel/blob/2.2.x/examples/postgres/relations/src/main.rs#L147

NOTE: i was able to workaround this issue by manually specifying the return value of the `.load()` call, as in:

```rust
let attrs = BookAuthor::belonging_to(&author)
      .inner_join(authors::table)
      .select((BookAuthor::as_select(), Author::as_select()))
      .load::<(BookAuthor, Author)>(conn);
```

**build error**

```rust
error[E0283]: type annotations needed
    --> src/main.rs:155:18
     |
155  |                 .load(conn);
     |                  ^^^^ ---- type must be known at this point
     |                  |
     |                  cannot infer type of the type parameter `U` declared on the method `load`
     |
     = note: cannot satisfy `_: FromSqlRow<(expression::select_by::SelectBy<BookAuthor, Sqlite>, expression::select_by::SelectBy<Author, Sqlite>), Sqlite>`
     = help: the following types implement trait `FromSqlRow<ST, DB>`:
               <(T1, T0) as FromSqlRow<(ST1, Untyped), __DB>>
               <(T1, T2, T0) as FromSqlRow<(ST1, ST2, Untyped), __DB>>
               <(T1, T2, T3, T0) as FromSqlRow<(ST1, ST2, ST3, Untyped), __DB>>
               <(T1, T2, T3, T4, T0) as FromSqlRow<(ST1, ST2, ST3, ST4, Untyped), __DB>>
               <(T1, T2, T3, T4, T5, T0) as FromSqlRow<(ST1, ST2, ST3, ST4, ST5, Untyped), __DB>>
               <(T1, T2, T3, T4, T5, T6, T0) as FromSqlRow<(ST1, ST2, ST3, ST4, ST5, ST6, Untyped), __DB>>
               <(T1, T2, T3, T4, T5, T6, T7, T0) as FromSqlRow<(ST1, ST2, ST3, ST4, ST5, ST6, ST7, Untyped), __DB>>
               <(T1, T2, T3, T4, T5, T6, T7, T8, T0) as FromSqlRow<(ST1, ST2, ST3, ST4, ST5, ST6, ST7, ST8, Untyped), __DB>>
             and 23 others
     = note: required for `(SelectBy<BookAuthor, Sqlite>, SelectBy<Author, Sqlite>)` to implement `load_dsl::private::CompatibleType<_, Sqlite>`
     = note: required for `SelectStatement<FromClause<JoinOn<..., ...>>, ..., ..., ...>` to implement `LoadQuery<'_, diesel::SqliteConnection, _>`
note: required by a bound in `diesel::RunQueryDsl::load`
```

i'm trying to perform an `inner_join` on a many-to-many relationship with sqlite as the database engine.

### Steps to reproduce

port the `relations` postgres example directly to `SqliteConnection`, using the same schema and model and then try to build.

## Checklist

weiznich:
Closes #4060, as this issue cannot be reproduced (as demonstrated by this port)


examples/sqlite/relations/src/main.rs
```diff
@@ -0,0 +1,206 @@
+use diesel::prelude::*;
```
kw217, line 1: All these files look fine - I diffed them against the postgres versions and only the expected changes were present.

Not for this PR, but more generally - is there a way we can do better, maybe with macros or preprocessing, so that we can have all the code written once, and compilation is attempted against all the backends?

kw217 (approved):
LGTM.

examples/postgres/relations/migrations/2023-02-17-084617_add_books_authors/up.sql
```diff
@@ -1,6 +1,6 @@
 -- Your SQL goes here
 CREATE TABLE books_authors (
-  book_id INTEGER REFERENCES books(id),
-  author_id INTEGER REFERENCES authors(id),
+  book_id INTEGER NOT NULL REFERENCES books(id),
```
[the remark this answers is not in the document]
weiznich: I just noticed that the generated schema did not match what is checked in as schema, so I just fixed that. #4066 will then just adjust the test scripts to check that as well.

examples/sqlite/relations/src/main.rs
weiznich: It's quite hard to do that with macros as there are some subtle differences between each of the backends in terms of supported SQL features. There is the `#[derive(MultiConnection)]` macro that allows you to write backend "agnostic" code, but these examples should demonstrate that you can in fact optimize your code for a specific backend (and use features that are only supported by this backend).

kw217: OK, makes sense. I'll keep pondering this, though.

PR #4063, opened 2024-06-07
Allow Field and Row types with Queryable derive

### LINKED ISSUE #4062: derive(Queryable) doesn't work with a field of `Field` type
### Versions

- **Rust:** 1.78.0
- **Diesel:** 2.2.0
- **Database:** PostgreSQL
- **Operating System** Linux

### Feature Flags

- **diesel:** none

I'm trying use a type called `Field` in my `Queryable` data structure. In my case it has `FromSql` implementation from `Jsonb`, but in a minimal example for reproduction below I aliased `i32` to make things as simple as possible.

It should work.

```
error[E0106]: missing lifetime specifier
 --> src/main.rs:7:12
  |
7 |     field: Field,
  |            ^^^^^ expected named lifetime parameter
  |
  = note: for more information on higher-ranked polymorphism, visit https://doc.rust-lang.org/nomicon/hrtb.html

error[E0106]: missing lifetime specifier
 --> src/main.rs:7:12
  |
7 |     field: Field,
  |            ^^^^^ expected named lifetime parameter

error[E0107]: missing generics for trait `diesel::row::Field`
   --> src/main.rs:7:12
    |
7   |     field: Field,
    |            ^^^^^ expected 1 generic argument
    |
note: trait defined here, with 1 generic parameter: `DB`
   --> /home/xfix/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.0/src/row.rs:100:11
    |
100 | pub trait Field<'a, DB: Backend> {
    |           ^^^^^     --
help: add missing generic argument
    |
7   |     field: Field<DB>,
    |                 ++++

error[E0782]: trait objects must include the `dyn` keyword
 --> src/main.rs:7:12
  |
7 |     field: Field,
  |            ^^^^^
  |
help: add `dyn` keyword before this trait
  |
7 |     field: dyn Field,
  |            +++

Some errors have detailed explanations: E0106, E0107, E0782.
For more information about an error, try `rustc --explain E0106`.
```

### Steps to reproduce

Compile the following code.

```rust
use diesel::Queryable;

type Field = i32;

#[derive(Queryable)]
pub struct Example {
    field: Field,
}

fn main() {}
```

## Checklist

sugar700:
Fixes #4062.

### DIFF
--- a/diesel_derives/tests/queryable.rs
+++ b/diesel_derives/tests/queryable.rs
@@ -87,3 +87,16 @@ fn multiple_tables() {
         data
     );
 }
+
+#[test]
+fn name_conflict() {
+    type Field = i32;
+    type Record = i32;
+
+    #[derive(Debug, Clone, PartialEq, Eq, Queryable)]
+    struct MyStruct(Field, Record);
+
+    let conn = &mut connection();
+    let data = select(sql::<(Integer, Integer)>("1, 2")).get_result(conn);
+    assert_eq!(Ok(MyStruct(1, 2)), data);
+}
--- a/diesel_derives/src/queryable.rs
+++ b/diesel_derives/src/queryable.rs
@@ -46,7 +46,7 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
 
     Ok(wrap_in_dummy_mod(quote! {
         use diesel::deserialize::{self, FromStaticSqlRow, Queryable};
-        use diesel::row::{Row, Field};
+        use diesel::row::{Row as _, Field as _};
         use std::convert::TryInto;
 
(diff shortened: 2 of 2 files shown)


weiznich (approved):
Thanks for submitting this PR :heart:

diesel_derives/tests/queryable.rs
```diff
@@ -87,3 +87,16 @@ fn multiple_tables() {
         data
     );
 }
+
+#[test]
+fn name_conflict() {
+    type Field = i32;
+    type Record = i32;
```
Mingun, line 94: Probably this is should be `Row`?

weiznich: That would have introduced a test for the `Row` case, so yes that would have been the optimal solution, but it's not that bad to not have it.

PR #4064, opened 2024-06-07
Fix a potential regression with old mysqlclient-sys versions

weiznich:
This commit fixes a potential regression with old mysqlclient-sys version that just do not have a used constant. We fallback to using a workaround based on default values in this situation. Manual tests with all versions have verified that this seems to work I'm not 100% sure why this was not found earlier by our minimal version check CI.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -389,13 +389,13 @@ impl BindData {
             length,
             capacity,
             flags,
-            is_null: ffi::FALSE,
-            is_truncated: Some(ffi::FALSE),
+            is_null: super::raw::ffi_false(),
+            is_truncated: Some(super::raw::ffi_false()),
         }
     }
 
     fn is_truncated(&self) -> bool {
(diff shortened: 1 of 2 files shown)


kw217:
There are two other uses of `ffi::FALSE`, but they both appear in test code; I assume the intention is we always test with a current `mysqlclient-sys` so this is OK.
https://github.com/diesel-rs/diesel/blob/53dcd4643511ed2fb0301f174e460695677242a5/diesel/src/mysql/connection/bind.rs#L1273 and https://github.com/diesel-rs/diesel/blob/53dcd4643511ed2fb0301f174e460695677242a5/diesel/src/mysql/connection/bind.rs#L1289

kw217:
This looks good to me.

weiznich:
> There are two other uses of ffi::FALSE, but they both appear in test code; I assume the intention is we always test with a current mysqlclient-sys so this is OK.

Yes for test code that's fine as we always use the latest mysqlclient-sys version there.

PR #4067, opened 2024-06-10
Do not implicitly convert `std::io::Error` to our CLI Error type

weiznich:
This commit changes our CLI tool to always explicitly include paths in the IO related error messages by enforcing that you cannot implicitly construct our own Error type from a `io::Error` (e.g. via `?`). Instead an explicit conversion is required, which allows to specify the relevant path

### DIFF
--- a/diesel_cli/src/config.rs
+++ b/diesel_cli/src/config.rs
@@ -56,7 +56,8 @@ impl Config {
         let path = Self::file_path(matches);
 
         if path.exists() {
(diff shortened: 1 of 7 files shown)


weiznich:
@Ten0 Any chance to also review https://github.com/diesel-rs/diesel/pull/4064. I would like to issue a patch release with this change, the mysql fix, the auto_type fix and with https://github.com/sgrif/pq-sys/pull/64 to fix a bunch of issues/regressions with the 2.2 release.

Issue #777, opened 2017-03-04
Add integration tests for examples


killercup:
We currently have the app from the Getting Started guide implemented for all backends, but there are no tests to ensure they all behave the same.

The app implements several simple binaries to create/display/publish/delete posts. Writings some tests that interact with _all_ the binaries and assert their output is good would be great (it doesn't even have to cover all failure cases).

I wrote the [assert_cli](https://crates.io/crates/assert_cli) crate for a similar use-case. There might be better crates for that now, though. Or actually, it might be possible to do that in a shell script. (Don't know how well that works on Windows, though.)

cgm616:
I'm interested in this issue because I'm trying to figure out how to test my own application that uses diesel. Will I have to run my own CI server with a DB to connect to and verify after every command? There has to be a better way. 

Hopefully tests in the examples will give me an idea.

heyztb:
Since there hasn't been any activity on this since 2017, I'd like to tackle this issue 😄

killercup:
@heyztb sure, go for it! Maybe open a PR with a rough idea what you want to test and then iterate on that? This way, we can give you feedback as early as possible. Alternatively, feel free to ask questions here or (preferred) on Gitter.

heyztb:
@killercup - Sounds good. I'll open a PR once I've got a tentative test up and running. I'm going to work on the Postgres example first, mainly because I'm not sure exactly which library I need to install in order to work with MySQL (I'm on macOS Mojave, btw).

I do have a question, and I figured this would be an appropriate place to ask but I'm also going to link this issue to the Gitter channel and get some feedback there -

I understand that this issue asks for Integration tests, and I've taken a look at the assert_cmd crate, but I'm not too sure how I'd pass in arguments to the various binaries. I'd rather the tests not require actual user input, and I would like to simulate that instead. My question is, how do I go about passing in things like the title and body of a post in my tests? There's no straightforward way from what I can tell, so are we sure that unit tests aren't more appropriate for this? Thanks for the help. I look forward to contributing more!

killercup:
@heyztb cool!

Very good question regarding the interativity! I actually totally forgot about that. It would be nice to test that, but it's totally fine to start with unit tests: We are trying to make sure the diesel examples are fine, not that `stdin` works, after all :)

[...]

PR #4083, opened 2024-06-25
getting started integration tests

mwspracklen:
Working on adding integration tests per #777 . I've created a quick example of how I'm planning to implement this in examples/sqlite/getting_started_step_2 . Initial thought is to add a tests directory with a test that uses assert_cmd to run the relevant binaries for each step. Considering writing in some calls to diesel to examine the database as well, let me know if that's the correct route or if just verifying binary output is sufficient.

In examples/sqlite/getting_started_step_2 I ran the following to test
echo "DATABASE_URL=file:/tmp/test.db" > .env; rm -f /tmp/test.db ; diesel migration run; cargo test; rm -f .env

### DIFF
--- a/examples/sqlite/getting_started_step_2/Cargo.toml
+++ b/examples/sqlite/getting_started_step_2/Cargo.toml
@@ -10,6 +10,9 @@ diesel = { version = "2.1.0", path = "../../../diesel", features = ["sqlite", "r
 dotenvy = "0.15"
 libsqlite3-sys = { version = "0.28.0", features = ["bundled"] }
 
+[dev-dependencies]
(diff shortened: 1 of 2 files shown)


VicVerevita:
Hello, I have made the postgresql and mysql tests for getting started in #4087. Tell me if you will finish the sqlite tests or if I should add them as well in my PR.

mwspracklen:
I was planning to wait for feedback, but it's probably better if there's one PR for the whole issue, so you can take the whole thing and I'll delete this PR

weiznich:
It would have been better to at least pull in the commit that @mwspracklen has done, but now it is at it is.

PR #4085, opened 2024-06-26
Parenthesized expression support in #[dsl::auto_type]

Ten0:
For some reason rust-analyzer seems to pass this instead of invisible group delimiters in macro expansions sometimes, resulting in errors in rust-analyzer but not rustc if this is not supported.
(Also that is trivial to support and may be something useful in other scenarios anyway.)


kw217 (approved):
LGTM.

PR #4093, opened 2024-07-02
Bump libsqlite3-sys and diffy to newer versions

weiznich:


Cargo.toml
```diff
@@ -35,6 +35,9 @@ members = [
 rust-version = "1.78.0"
 include = ["src/**/*.rs", "tests/**/*.rs", "LICENSE-*", "README.md"]
 
+[workspace.dependencies]
+libsqlite3-sys = "0.29"
```
kw217, line 39: Ooh, this is a nice feature!

kw217 (changes requested):
LGTM, but in previous bumps of libsqlite3-sys you've noted it in CHANGELOG - do you want to do that this time too?

weiznich:
Good catch with the chaneglog entry :pray:  I've pushed https://github.com/diesel-rs/diesel/pull/4093/commits/f56268c4ba4e10470ce4a5a9b2969c43fc4062c8 to address that.

Issue #4092, opened 2024-07-01
Add support for currently unsupported range operators and methods


weiznich:
Diesel currently supports the postgres range types. We do not provide built-in support for [various operators and methods](https://www.postgresql.org/docs/current/functions-range.html) available for these types. This is a tracking issue for adding support for these operators. 

The general strategy for adding support for new operators is as following:

1. Define the operator via [`infix_operator!()`](https://docs.diesel.rs/master/diesel/macro.infix_operator.html). The RangeExtends operator  would be defined as following: `infix_operator!(RangeExtends, " &> ", backend: Pg);` These operators can be defined [here](https://github.com/diesel-rs/diesel/blob/dd5f41dc93c574077646185393cd064c7103000b/diesel/src/pg/expression/operators.rs). If there is already an existing definition, this step could be skipped.
2. Define a new helper type [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/expression/helper_types.rs#L30). Helper types are a tool for diesel users to simplify type definitions. Again if there is already a matching type definition this could be skipped. 
3. Add a new method to the [`PgRangeExpressionMethods` trait](https://github.com/diesel-rs/diesel/blob/99ef8bbdee614b8eb6560b00c4c1ead88fb0e737/diesel/src/pg/expression/expression_methods.rs#L706-L758). This definition should:
    * Use the previously defined helper type
    * Contain a documentation with an example
4. If possible add a test case for the auto_type attribute [here](https://github.com/diesel-rs/diesel/blob/99ef8bbdee614b8eb6560b00c4c1ead88fb0e737/diesel_derives/tests/auto_type.rs#L277-L288). (If not possible add a documentation entry there)
5. Check if compile test output is up to date by running `TRYBUILD=overwrite cargo test` inside of `diesel_compile_tests`. (This is required if you've added a new operator based on `infix_operator!`
5. Submit a PR with the change

The general strategy for adding support for new methods is as following:

[...]

PR #4095, opened 2024-07-02
feat/range contains range

guissalustiano:
Implements ` anyrange @> anyelement -> boolean` from https://github.com/diesel-rs/diesel/issues/4092
Following the helpful steps described by @weiznich 

## API Drawbacks
Currently the `‎PgArrayExpressionMethods::contains` receives `Array<T>`, so the Range API will be different using `contains` for a single element and `contains_range` for the range ([which was already noted here](https://github.com/diesel-rs/diesel/blob/99ef8bbdee614b8eb6560b00c4c1ead88fb0e737/diesel_derives/tests/auto_type.rs#L279))
Maybe it's possible to accept both `Range<T>` and `T` in the same function signature, but this is far from my knowledge, please let me know if it is possible!

This is also my first PR here, please let me know if I can do anything better!

### DIFF
--- a/diesel/src/pg/expression/helper_types.rs
+++ b/diesel/src/pg/expression/helper_types.rs
@@ -62,7 +62,7 @@ pub type RangeContains<Lhs, Rhs> = Grouped<
 /// The return type of [`lhs.contains(rhs)`](super::expression_methods::PgRangeExpressionMethods::contains_range)
 /// for range expressions
 #[cfg(feature = "postgres_backend")]
-pub type RangeContainsRange<Lhs, Rhs> = Contains<Lhs, Rhs>;
+pub type ContainsRange<Lhs, Rhs> = Contains<Lhs, Rhs>;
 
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/helper_types.rs
```diff
@@ -59,6 +59,11 @@ pub type RangeContains<Lhs, Rhs> = Grouped<
     >,
 >;
 
+/// The return type of [`lhs.contains(rhs)`](super::expression_methods::PgRangeExpressionMethods::contains_range)
```
weiznich, line 62:
```suggestion
/// The return type of [`lhs.contains_range(rhs)`](super::expression_methods::PgRangeExpressionMethods::contains_range)
```

weiznich (commented):
Thanks for submitting this PR. The added operator looks good :+1:

I just found a minor type in the documentation of the helper type otherwise this is ready to be merged.

diesel/src/pg/expression/helper_types.rs
guissalustiano: Thanks! I commit your sugestion

PR #4100, opened 2024-07-05
Add is_contained_by to PgRangeExpressionMethods

cole-wildcardsoftware:
This is adding support for `anyrange <@ anyrange -> boolean` for #4092 
Added helper type, method operation, and auto test

This is my first PR, let me know if there's anything I need to add to this or anything I should do different in the future

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -806,6 +806,57 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
     {
         Grouped(Contains::new(self, other.as_expression()))
     }
+
+    /// Creates a PostgreSQL `<@` expression.
+    ///
+    /// This operator returns whether a range is contained by a specific element
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     posts {
+    /// #         id -> Integer,
+    /// #         versions -> Range<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::posts::dsl::*;
+    /// #     use std::collections::Bound;
+    /// #     let conn = &mut establish_connection();
+    /// #     diesel::sql_query("DROP TABLE IF EXISTS posts").execute(conn).unwrap();
+    /// #     diesel::sql_query("CREATE TABLE posts (id SERIAL PRIMARY KEY, versions INT4RANGE NOT NULL)").execute(conn).unwrap();
+    /// #
+    /// diesel::insert_into(posts)
+    ///     .values(versions.eq((Bound::Included(5), Bound::Unbounded)))
+    ///     .execute(conn)?;
+    ///
+    /// let cool_posts = posts.select(id)
+    ///     .filter(versions.is_contained_by((Bound::Included(1), Bound::Unbounded)))
+    ///     .load::<i32>(conn)?;
+    /// assert_eq!(vec![1], cool_posts);
+    ///
+    /// let amazing_posts = posts.select(id)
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -806,6 +806,58 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
...
+    ///
+    /// let amazing_posts = posts.select(id)
+    ///     .filter(versions.is_contained_by((Bound::Included(1), Bound::Included(2))))
+    ///     .load::<i32>(conn)?;
+    /// assert!(amazing_posts.is_empty());
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[allow(clippy::wrong_self_convention)] // This is named after the sql operator
+    fn is_contained_by<T>(self, other: T) -> dsl::IsContainedByRange<Self, T>
```
pksunkara, line 854: This should be named `is_contained_by_range` going by names of the other functions above.

guissalustiano: Hi, I'm new to the project but I was who implemented the above method.
`PgNetExpressionMethods`, `PgJsonbExpressionMethods`, and `PgArrayExpressionMethods` uses `is_contained_by` for `<@` operator.
I think we should only break the overall pattern only in "contains_range", and for the other operations, we should keep the pattern with the overall API. What do you think?

cole-wildcardsoftware: @pksunkara I'm fine doing either. For *all* of these new range operators do we want to follow the *operator*_range naming convention as well

pksunkara: Actually, my bad. I was mistaken that this `is_contained_by` will conflict with `is_contained_by` of `PgJsonbExpressionMethods` but I don't think it will. 

And yes, if that's true, `contains_range` should be named `contains` using deprecation. @weiznich Would like your thoughts here.

guissalustiano: I agree, `contains` should accept a `Range<T>` and we could have something like `contains_one` that's receives `T`

pksunkara: I gave it a deeper look and I think the `is_contained_by` is the correct name for this PR.

diesel/src/pg/expression/helper_types.rs
```diff
@@ -64,6 +64,11 @@ pub type RangeContains<Lhs, Rhs> = Grouped<
 #[cfg(feature = "postgres_backend")]
 pub type ContainsRange<Lhs, Rhs> = Contains<Lhs, Rhs>;
 
+/// The return type of [`lhs.is_contained_by(rhs)`](super::expression_methods::PgRangeExpressionMethods::is_contained_by)
+#[cfg(feature = "postgres_backend")]
+pub type IsContainedByRange<Lhs, Rhs> =
```
pksunkara, line 69: But this should be named `RangeIsContainedBy` though.

weiznich: Ideally this type name should exactly match (in CamelCase) the name of the method, so that `#[auto_type]` works. That means I would rather prefer naming the method `range_is_contained_by` or something like that.

diesel/src/pg/expression/expression_methods.rs
weiznich: `contains` is a bit of outlier here as it exists before `#[auto_type]`. It's one of the few messages that are currently not supported by that macro due to the different return type structure. I would like to avoid that for new methods by just choosing a fitting not already existing name.

diesel/src/pg/expression/helper_types.rs
cole-wildcardsoftware: Would we be able to use the helper type of IsContainedBy that is also being used by by the array method just below this type so that method names remain consistent? Or do we want to just have it be the range_is_contained_by method name and helper type

weiznich: Yes that would be possible and preferable in this case (and any other case where the type definition matches). That would need an update of the documentation down below there to mention the range methods as well.

weiznich (approved):
Thanks :heart:

PR #4103, opened 2024-07-05
add lower(Range) function

guissalustiano:
This is adding support for `lower(anyrange) -> any`, part of https://github.com/diesel-rs/diesel/issues/4092
Added method operation and auto test.

This function has the same name as `lower(text)`, maybe we want a specific name to not overlap in the future, like `lower_range`.

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -104,5 +104,5 @@ define_sql_function! {
     /// # }
     /// ```
     #[cfg(feature = "postgres_backend")]
-    fn lower<T: RangeHelper<Inner: SingleValue> + SingleValue>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
+    fn lower<T: RangeHelper>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
 }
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -62,3 +63,46 @@ define_sql_function! {
...
+    ///     ]).execute(conn)?;
+    ///
+    /// let cool_posts = posts.select(lower(versions))
+    ///     .load::<Option<i32>>(conn)?;
+    /// assert_eq!(vec![Some(5), None], cool_posts);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "postgres_backend")]
+    fn lower<T: RangeHelper<Inner: SingleValue> + SingleValue>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
```
guissalustiano, line 107: From what I could understand, `define_sql_function` does not accept `where` clause, so I used rust [1.79 associated item bounds syntax](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position) to make it inline, but this breaks the MNV

weiznich: I've pushed https://github.com/diesel-rs/diesel/pull/4103/commits/42681bd8d8b36f36c014f2c310ae158ac6cbb0bb to fix this.

guissalustiano: Thanks!

weiznich (approved):
Thanks :heart:

PR #4107, opened 2024-07-08
Implements `int4range`

guissalustiano:
This is adding support for `int4range(Nullable<Integer>, Nullable<Integer>, RangeBoundEnum) -> int4range` for https://github.com/diesel-rs/diesel/issues/4092

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -106,3 +106,51 @@ define_sql_function! {
     #[cfg(feature = "postgres_backend")]
     fn lower<T: RangeHelper>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
 }
+
+define_sql_function! {
+    /// Returns range of integer.
+    /// if the range is empty or has no lower bound, it returns NULL.
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     posts {
+    /// #         id -> Integer,
+    /// #         versions -> Range<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::posts::dsl::*;
+    /// #     use std::collections::Bound;
+    /// #     let conn = &mut establish_connection();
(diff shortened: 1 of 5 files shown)


diesel/src/pg/types/mod.rs
```diff
@@ -161,6 +161,29 @@ pub mod sql_types {
...
+    #[allow(clippy::enum_variant_names)]
+    pub enum RangeBound {
+        /// postgres '[]'
+        LowerBoundInclusiveUpperBoundInclusive,
+        /// postgres '[)'
+        LowerBoundInclusiveUpperBoundExclusive,
+        /// postgres '(]'
+        LowerBoundExclusiveUpperBoundInclusive,
+        /// postgres '()'
+        LowerBoundExclusiveUpperBoundExclusive,
```
guissalustiano, line 184: Do you have any suggestions for this enum's names? Clippy didn't like that everyone starts with the same prefix, but I think `InclusiveExclusive` a little confunsing

diesel_derives/tests/auto_type.rs
```diff
@@ -290,8 +290,12 @@ fn test_pg_range_expression_methods() -> _ {
     // function. We could likely support it by
     // renaming the function to `.range_contains()` (or something similar)
     // .contains(42_i32)
+    //.select(
     // this kind of free standing functions is not supported by auto_type yet
-    //.select(lower(pg_extras::range))
+    //lower(pg_extras::range),
+    // The auto_trait also dind't like this one
```
weiznich, line 296:
```suggestion
    // The auto_trait also didn't like this one
```

diesel/src/pg/types/mod.rs
weiznich: I think these names are fine. After all code completion is relatively good these days so it shouldn't matter to use longer names.

diesel/src/pg/expression/functions.rs
```diff
@@ -106,3 +106,51 @@ define_sql_function! {
     #[cfg(feature = "postgres_backend")]
     fn lower<T: RangeHelper>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
 }
+
+define_sql_function! {
+    /// Returns range of integer.
+    /// if the range is empty or has no lower bound, it returns NULL.
```
weiznich, line 112: This doesn't sound correct?

weiznich (commented):
Looks good beside the minor documentation issues

diesel/src/pg/expression/functions.rs
guissalustiano: I copied and past and forgot to change this line, sorry

PR #4112, opened 2024-07-09
implements to_sql `diesel::Range` to `std::Range*`

guissalustiano:
With this PR we can now use `1..` instead `(Bound::Inclded(1), Bound::Unbounded)` for all `std::ops::Range.*<T>`.

I only didn't implement the `RangeFull` because it does not have a type so I didn't know how to keep the contraint in `T`.

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -135,6 +135,11 @@ pub mod sql_types {
     /// ### [`ToSql`] impls
     ///
     /// - [`(Bound<T>, Bound<T>)`][bound] for any `T` which implements `ToSql<ST>`.
+    /// - [`Range<T>`][std::range] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeInclusive<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeFrom<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeTo<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeToInclusive<T>`] for any `T` which implements `ToSql<ST>`.
     ///
     /// ### [`FromSql`] impls
     ///
@@ -143,6 +148,11 @@ pub mod sql_types {
     /// [`ToSql`]: crate::serialize::ToSql
     /// [`FromSql`]: crate::deserialize::FromSql
     /// [bound]: std::collections::Bound
+    /// [std::range]: std::ops::Range
+    /// [`RangeInclusive<T>`]: std::ops::RangeInclusive
+    /// [`RangeFrom<T>`]: std::ops::RangeFrom
+    /// [`RangeTo<T>`]: std::ops::RangeTo
+    /// [`RangeToInclusive<T>`]: std::ops::RangeToInclusive
     /// [`Range`]: https://www.postgresql.org/docs/current/rangetypes.html
     #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
     #[cfg(feature = "postgres_backend")]
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -135,11 +135,11 @@ pub mod sql_types {
     /// ### [`ToSql`] impls
     ///
     /// - [`(Bound<T>, Bound<T>)`][bound] for any `T` which implements `ToSql<ST>`.
-    /// - [`Range<T>`][std::range] for any `T` which implements `ToSql<ST>`.
-    /// - [`RangeInclusive<T>`] for any `T` which implements `ToSql<ST>`.
-    /// - [`RangeFrom<T>`] for any `T` which implements `ToSql<ST>`.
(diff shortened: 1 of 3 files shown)


guissalustiano:
May I replace the test with this notation?
Like here
https://github.com/diesel-rs/diesel/blob/18af06f3a798f7f1a3e8298882d1d457c118ad1d/diesel/src/pg/expression/functions.rs#L96-L97

diesel/src/pg/types/mod.rs
```diff
@@ -135,6 +135,11 @@ pub mod sql_types {
     /// ### [`ToSql`] impls
     ///
     /// - [`(Bound<T>, Bound<T>)`][bound] for any `T` which implements `ToSql<ST>`.
+    /// - [`Range<T>`][std::range] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeInclusive<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeFrom<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeTo<T>`] for any `T` which implements `ToSql<ST>`.
+    /// - [`RangeToInclusive<T>`] for any `T` which implements `ToSql<ST>`.
```
weiznich, line 142: Might be worth to point out the syntactic sugar for this
```suggestion
    /// - [`Range<T>`][std::range] (aka `start..end`) for any `T` which implements `ToSql<ST>`.
    /// - [`RangeInclusive<T>`] (aka `start..=end`) for any `T` which implements `ToSql<ST>`.
    /// - [`RangeFrom<T>`] (aka `start..`) for any `T` which implements `ToSql<ST>`.
    /// - [`RangeTo<T>`] (aka `..end`) for any `T` which implements `ToSql<ST>`.
    /// - [`RangeToInclusive<T>`] (aka `..=end`) for any `T` which implements `ToSql<ST>`.
```

diesel/src/pg/types/ranges.rs
```diff
@@ -24,41 +24,51 @@ bitflags::bitflags! {
...
-    fn as_expression(self) -> Self::Expression {
-        SqlBound::new(self)
-    }
+macro_rules! range_as_expression {
+    ($ty:ty; $sql_type:ty) => {
+        #[cfg(feature = "postgres_backend")]
+        // this simplifies the macro implementation
+        // as some macro calls use this lifetime
+        #[allow(clippy::extra_unused_lifetimes)]
+        impl<'a, 'b, ST: 'static, T> AsExpression<$sql_type> for $ty {
```
weiznich, line 33: Can you point out where `'b` is used?

weiznich (approved):
Looks good for me beside the minor notes about documentation and the one question.

I had some concerns about this due to https://github.com/rust-lang/rfcs/blob/master/text/3550-new-range.md but the text explicitly states that after that we only need to add a few new impls for the new range types and it should continue to work on all editions.

diesel/src/pg/types/ranges.rs
guissalustiano: Well notice, I forgot that there. Thanks!

guissalustiano:
> I had some concerns about this due to https://github.com/rust-lang/rfcs/blob/master/text/3550-new-range.md but the text explicitly states that after that we only need to add a few new impls for the new range types and it should continue to work on all editions.
From what I understood we will only need to add the macro implementations to the legacy ranges.

The best implementation would be to depend only on `RangeBound`, but [when I tried](https://github.com/diesel-rs/diesel/commit/af1065ee8) I received a conflicting implementations error:
```
error[E0119]: conflicting implementations of trait `ToSql<types::sql_types::Range<_>, pg::backend::Pg>` for type `&_`
   --> diesel/src/pg/types/ranges.rs:190:1
    |
190 | / impl<ST, T, RB> ToSql<Range<ST>, Pg> for RB
191 | | where
192 | |     ST: 'static,
193 | |     RB: RangeBounds<T>,
194 | |     T: ToSql<ST, Pg>,
    | |_____________________^ conflicting implementation for `&_`
    |
   ::: diesel/src/serialize.rs:283:1
    |
283 | / impl<'a, A, T, DB> ToSql<A, DB> for &'a T
284 | | where
285 | |     DB: Backend,
286 | |     T: ToSql<A, DB> + ?Sized,
    | |_____________________________- first implementation here
    |
    = note: downstream crates may implement trait `std::ops::RangeBounds<_>` for type `&_`
error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
   --> diesel/src/pg/types/ranges.rs:190:10
    |
190 | impl<ST, T, RB> ToSql<Range<ST>, Pg> for RB
    |          ^ unconstrained type parameter
```
And I don't know how to fix it

weiznich (approved):

weiznich:
> From what I understood we will only need to add the macro implementations to the legacy ranges.
>
> The best implementation would be to depend only on RangeBound, but [when I tried](https://github.com/diesel-rs/diesel/commit/af1065ee8) I received a conflicting implementations error:

I think it's not possible to fix that. So going for the concrete impl is the best solution that's possible.

PR #4115, opened 2024-07-12
Fixed a potential panic in SQLite row iterators

weiznich:
This commit fixes a potential panic in sqlite row iterator if the first iteration step returns an error. We return now a `None` value there instead of panicing as the iterator "ends" at that point.

### DIFF
--- a/diesel/src/sqlite/connection/statement_iterator.rs
+++ b/diesel/src/sqlite/connection/statement_iterator.rs
@@ -92,7 +92,7 @@ impl<'stmt, 'query> Iterator for StatementIterator<'stmt, 'query> {
     fn next(&mut self) -> Option<Self::Item> {
         use PrivateStatementIterator::{NotStarted, Started};
         match &mut self.inner {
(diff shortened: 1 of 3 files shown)


diesel/src/sqlite/connection/statement_iterator.rs
```diff
@@ -161,12 +161,12 @@ impl<'stmt, 'query> Iterator for StatementIterator<'stmt, 'query> {
...
-                "You've reached an impossible internal state. \
-                             If you ever see this error message please open \
-                             an issue at https://github.com/diesel-rs/diesel \
-                             providing example code how to trigger this error."
-            ),
+            NotStarted(_s) => {
+                // we likely got an error while executing the other
+                // `NotStarted` branch above. In this case we just want to stop
+                // iterating here
+                None
```
kw217, line 168: This seems reasonable to me. Either we've already returned some error or we've already returned `None`. In either case, returning `None` forever after is good.

kw217 (approved):
LGTM.

PR #4116, opened 2024-07-12
Add space in "DISTINCT ON (...) "

Ten0:
For prettier query string.
This is homogeneous to what we do with commas and joins.


Ten0:
Currently we generate:
```sql
SELECT DISTINCT ON ("table"."id")"table"."id", "table"."field" FROM ...
```

pksunkara (approved):

PR #4119, opened 2024-07-13
Implement Pgrange overlaps_with

vijairamg:
Implementing `anyrange && anyrange -> boolean` as part of #4092

### DIFF
--- a/diesel/src/pg/expression/helper_types.rs
+++ b/diesel/src/pg/expression/helper_types.rs
@@ -37,6 +37,7 @@ pub type IsDistinctFrom<Lhs, Rhs> =
     Grouped<super::operators::IsDistinctFrom<Lhs, AsExpr<Rhs, Lhs>>>;
 
 /// The return type of [`lhs.overlaps_with(rhs)`](super::expression_methods::PgArrayExpressionMethods::overlaps_with)
+/// and [`lhs.overlaps_with(rhs)`](super::expression_methods::PgRangeExpressionMethods::overlaps_with)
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/helper_types.rs
```diff
@@ -59,6 +59,10 @@ pub type RangeContains<Lhs, Rhs> = Grouped<
     >,
 >;
 
+/// The return type of [`lhs.overlaps_with(rhs)`](super::expression_methods::PgRangeExpressionMethods::overlaps_with)
+/// for range expressions
+#[cfg(feature = "postgres_backend")]
+pub type RangeOverlaps<Lhs, Rhs> = OverlapsWith<Lhs, Rhs>;
```
weiznich, line 65: I believe it's not necessary to have this type def as it matches the typedef of the array operators in line 41. We can just adjust the doc comment there and mention that this is also the return type of `PgRangeExperssionMethods::overlaps_with(rhs)`.

weiznich (changes requested):
Thanks for submitting this PR and sorry for taking some time for doing a review.

The change looks mostly good beside the minor note about the new type def. It can be just replaced by the existing one.

diesel/src/pg/expression/helper_types.rs
vijairamg: I have removed the new type declaration and added the comment.

weiznich (approved):

weiznich:
Thanks for the update. It looks good now.

PR #4124, opened 2024-07-18
add enum with string and mysql backend doc

Ludea:
add some example to create custom enum with String type


diesel/src/serialize.rs
```diff
@@ -207,6 +207,39 @@ where
 /// }
 /// ```
 ///
+/// Example with String and custom type created by diesel cli
```
weiznich, line 210:
```suggestion
/// Example of creating a custom type mapping based on a MySQL [enum type](https://dev.mysql.com/doc/refman/8.0/en/enum.html)
///
/// This is designed to reuse the SQL type definition generated by diesel-cli
```

diesel/src/serialize.rs
```diff
@@ -207,6 +207,39 @@ where
...
+/// # use diesel::sql_types::*;
+/// # use diesel::serialize::{self, ToSql, Output, IsNull};
+/// # use std::io::Write;
+/// #
+/// pub mod sql_types {
+///    #[derive(diesel::sql_types::SqlType)]
+///    #[diesel(mysql_type(name = "Enum"))]
+///    pub struct PostEnum; //<- generated by diesel cli
+/// }
+/// #[derive(Debug, AsExpression, PartialEq, Clone)]
```
weiznich, line 224:
```suggestion

/// #[derive(Debug, AsExpression, PartialEq, Clone)]
```

diesel/src/serialize.rs
```diff
@@ -207,6 +207,39 @@ where
...
+/// pub enum Post {
+///    FirstValue,
+///    SecondValue,
+/// }
+///
+/// # #[cfg(feature = "mysql")]
+/// impl ToSql<sql_types::PostEnum, diesel::mysql::Mysql> for Post {
+///    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, diesel::mysql::Mysql>) -> serialize::Result {
+///        match *self {
+///            Post::FirstValue => out.write_all(b"one")?,
```
weiznich, line 235:
```suggestion
///            // these string values need to match the labels used in your
///            // enum definition in SQL. So this expects that you defined the 
///            /// relevat enum type as`ENUM('one', 'two')` in your `CREATE TABLE` statement
///            Post::FirstValue => out.write_all(b"one")?,
```

weiznich (approved):
Thanks

PR #4129, opened 2024-07-25
Add support for downcasting Instrumentation

Ten0:
This is useful to instrument additional events related to the connection.

My use-case is that I want to check that `StartQuery` s were consistent with `FinishQuery` s when releasing the connection to the Pool (which I already have a type wrapper of `r2d2::PooledConnection` with its `Drop` implementation for), and I also want to clean up the instrumentation object at the same time.

The `DynInstrumentation` type is useful because without it we actually did tend to return `&mut Option<Box<dyn Instrumentation>> as &mut dyn Instrumentation` from `connection.instrumentation()`, so downcasting would have to be done in these two steps by the user, which is counter-intuitive and doesn't seem ideal inside diesel itself either (because it induces extra dyn dispatch levels).

An alternate implementation would be to just make `()` implement `Instrumentation` and use Box<dyn Instrumentation> always (Box of ZST doesn't allocate). This means we would always dyn dispatch even if instrumentation is disabled, but maybe the complexity/cost ratio is good, because probably the cost is very low given the events we currently instrument.


diesel/src/connection/instrumentation.rs
```diff
@@ -242,10 +242,11 @@ impl<'a> InstrumentationEvent<'a> {
 /// More complex usages and integrations with frameworks like
 /// `tracing` and `log` are supposed to be part of their own
 /// crates.
-pub trait Instrumentation: Send + 'static {
+pub trait Instrumentation: Downcast + Send + 'static {
```
weiznich, line 245: I'm not sure if that's a breaking change or not. 
On the one hand it adds a new super trait without a wild card impl for all types on the other hand `Downcast` is implemented for all `T: Any` and any is implemented by the compiler for all types that are `'static` (we already have that bound) so it's likely not breaking?

diesel/src/connection/instrumentation.rs
```diff
@@ -242,10 +242,11 @@ impl<'a> InstrumentationEvent<'a> {
...
     /// The function that is invoced for each event
     fn on_connection_event(&mut self, event: InstrumentationEvent<'_>);
 }
+downcast_rs::impl_downcast!(Instrumentation);
```
weiznich, line 249: We should use that for `BoxableConnection` as well if we add this dependency.

(I can address that in a followup PR)

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+)]
+/// An optional dyn instrumentation.
+/// For ease of use, this type implements [`Deref`] and [`DerefMut`] to `&dyn Instrumentation`,
+/// falling back to a no-op implementation if no instrumentation is set.
+pub(crate) struct DynInstrumentation {
+    /// zst
+    no_instrumentation: NoInstrumentation,
+    inner: Option<Box<dyn Instrumentation>>,
+}
+impl Deref for DynInstrumentation {
```
weiznich, line 331:
```suggestion

impl Deref for DynInstrumentation {
```

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+    type Target = dyn Instrumentation;
+
+    fn deref(&self) -> &Self::Target {
+        self.inner.as_deref().unwrap_or(&self.no_instrumentation)
+    }
+}
+impl DerefMut for DynInstrumentation {
```
weiznich, line 338:
```suggestion

impl DerefMut for DynInstrumentation {
```

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.inner
+            .as_deref_mut()
+            .unwrap_or(&mut self.no_instrumentation)
+    }
+}
+impl DynInstrumentation {
```
weiznich, line 345:
```suggestion

impl DynInstrumentation {
```

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
+    pub(crate) fn on_connection_event(&mut self, event: InstrumentationEvent<'_>) {
+        if let Some(inner) = self.inner.as_deref_mut() {
+            inner.on_connection_event(event)
+        }
+    }
+}
+impl<I: Instrumentation> From<I> for DynInstrumentation {
```
weiznich, line 375:
```suggestion

impl<I: Instrumentation> From<I> for DynInstrumentation {
```

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+    fn from(instrumentation: I) -> Self {
+        Self {
+            inner: Some(unpack_instrumentation(Box::new(instrumentation))),
+            no_instrumentation: NoInstrumentation,
+        }
+    }
+}
+
+struct NoInstrumentation;
+impl Instrumentation for NoInstrumentation {
```
weiznich, line 385:
```suggestion

impl Instrumentation for NoInstrumentation {
```

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
+    }
+
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
+    pub(crate) fn on_connection_event(&mut self, event: InstrumentationEvent<'_>) {
+        if let Some(inner) = self.inner.as_deref_mut() {
+            inner.on_connection_event(event)
+        }
+    }
```
weiznich, line 373: Shouldn't we just implement `Instrumentation` here instead?

diesel/src/connection/instrumentation.rs
```diff
@@ -313,3 +316,86 @@ where
...
     }
 }
+
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
+/// An optional dyn instrumentation.
+/// For ease of use, this type implements [`Deref`] and [`DerefMut`] to `&dyn Instrumentation`,
+/// falling back to a no-op implementation if no instrumentation is set.
+pub(crate) struct DynInstrumentation {
```
weiznich, line 326: Can we add the following part of your PR description as comment here?

> The DynInstrumentation type is useful because without it we actually did tend to return &mut Option<Box<dyn Instrumentation>> as &mut dyn Instrumentation from connection.instrumentation(), so downcasting would have to be done in these two steps by the user, which is counter-intuitive and doesn't seem ideal inside diesel itself either (because it induces extra dyn dispatch levels).

That provides the necessary context on why we use this abstraction instead of just `Option<Box<dyn Instrumentation>>` as before.

Also how about naming this type `OptionalInstrumentation` instead?

weiznich (approved):
Thanks for working on this. This seems to be a usefull addition. I have some conceptual questions that I would like to clarify and a number of stylistic requests.

diesel/src/connection/instrumentation.rs
Ten0: The risk is that this would allow `&mut DynInstrumentation` to be cast to `&mut dyn Instrumentation` directly instead of converting via the `Deref` impl.

diesel/src/connection/instrumentation.rs
Ten0: > Also how about naming this type OptionalInstrumentation instead?

I have no strong preference for either naming. Just so that the reasoning for the current naming is brought to the table: initially I named it `OptionalDynInstrumentation` for completeness but then I thought it was pretty verbose and maybe unnecessarily so, considering that fundamentally, no instrumentation is also an instrumentation, so I opted to keep in the name the part that would let readers know "hey, this is essentially `Box<dyn Instrumentation>` for all intents and purposes".

diesel/src/connection/instrumentation.rs
Ten0: > on the other hand Downcast is implemented for all T: Any and any is implemented by the compiler for all types that are 'static (we already have that bound)

Yes, that's why I thought that it was very unlikely to break anyone 😀

Ten0:
Thanks for the extremely quick review! 😊🙏

diesel/src/connection/instrumentation.rs
weiznich: That sounds reasonable. 
Can you add that as comment onto the function as well?

diesel/src/connection/instrumentation.rs
weiznich: That sounds reasonable :+1:

diesel/src/connection/instrumentation.rs
Ten0: f8d2d9d

PR #4130, opened 2024-07-25
Fix new Rust unused warnings

Ten0:
New compiler warnings seem to indicate that this wasn't used.
Instead the `CombinationClause` type holds both right hand side and left side of combinations, and implements `Query` directly.


Ten0:
Ah that error seems wrong:
`User` is constructed here: https://github.com/Ten0/diesel/blob/c7ab84fc71fc637cb0cbd1f93e5730bcae6eb189/examples/postgres/all_about_inserts/src/lib.rs#L273

weiznich:
I'm not sure what's going on with the compile tests update for you. It doesn't produce that output locally for me. I just pushed the changes produced locally, hopefully that is also what the CI produces.

I also fixed the new clippy warnings + bumped the toolchain version to 1.80

diesel_compile_tests/tests/fail/only_only_on_table.stderr
```diff
@@ -74,4 +74,4 @@ note: the trait `Table` must be implemented
    = note: the following trait defines an item `only`, perhaps you need to implement it:
            candidate #1: `diesel::dsl::OnlyDsl`
 
-error: internal compiler error: compiler/rustc_infer/src/infer/at.rs:400:21: relating different kinds: diesel::pg::Pg '?2
+error: internal compiler error: compiler/rustc_infer/src/infer/at.rs:364:21: relating different kinds: diesel::pg::Pg '?2
```
weiznich, line 77: https://github.com/rust-lang/rust/issues/126129

Ten0:
Thanks a lot!

weiznich (approved):

PR #4133, opened 2024-07-26
Implement lhs not extends to the right of rhs

ccarral:
Implements postgres `anyrange &< anyrange -> boolean` range operation as part of #4092


weiznich (approved):
Thanks :pray:

PR #4135, opened 2024-07-27
Implement pgrange union 

vijairamg:
Implements `anyrange + anyrange -> anyrange`, `anyrange * anyrange -> anyrange` and `anyrange - anyrange -> anyrange` part of #4092

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -1074,6 +1074,150 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
     {
         Grouped(ContainsNet::new(self, other.as_expression()))
     }
+
+    /// Creates a PostgreSQL ` + ` expression.
+    ///
+    /// This operator unions two ranges and returns the union.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     posts {
+    /// #         id -> Integer,
+    /// #         versions -> Range<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use self::posts::dsl::*;
+    /// #     use std::collections::Bound;
+    /// #     let conn = &mut establish_connection();
+    /// #     diesel::sql_query("DROP TABLE IF EXISTS posts").execute(conn).unwrap();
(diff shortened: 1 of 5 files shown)


diesel/src/pg/expression/operators.rs
```diff
@@ -118,3 +116,49 @@ where
...
+    }
+}
+
+impl<L, R, ST> crate::expression::Expression for UnionRange<L, R>
+where
+    L: crate::expression::Expression<SqlType = Range<ST>>,
+    R: crate::expression::Expression<SqlType = Range<ST>>,
+    ST: SqlType + TypedExpressionType,
+{
+    type SqlType = ST;
```
weiznich, line 142: That's the problematic line that's causing the error.

You specify here that `range_a.union_range(range_b)` return the inner type of the range instead. You want `type SqlType = Range<ST>` instead, which just says that the operator returns a range as well.

That written: You should likely use the `infix_operator!()` macro instead to generate all this code for you. It should be much easier that way.

vijairamg: Thanks for the pointer.
My bad! Should have noticed the `__diesel_infix_operator!`.

vijairamg:
Let me know if any of the names need to be changed.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1074,6 +1074,150 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
...
+    where
+        Self::SqlType: SqlType,
+        T: AsExpression<Self::SqlType>,
+    {
+        Grouped(DifferenceRange::new(self, other.as_expression()))
+    }
+
+    /// Creates a PostgreSQL ` * ` expression.
+    ///
+    /// This operator takes two ranges and returns the difference.
```
weiznich, line 1176:
```suggestion
    /// This operator takes two ranges and returns the intersection.
```

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1074,6 +1074,150 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
...
+    where
+        Self::SqlType: SqlType,
+        T: AsExpression<Self::SqlType>,
+    {
+        Grouped(UnionsRange::new(self, other.as_expression()))
+    }
+
+    /// Creates a PostgreSQL ` - ` expression.
+    ///
+    /// This operator takes two ranges and returns the difference.
```
weiznich, line 1128:
```suggestion
    /// This operator takes two ranges and returns the difference.
    /// 
    /// The second range must not be contained in the first in such a way that the 
    /// difference would not be a single range.
```

weiznich (approved):
Looks good beside the two minor documentation fixes. Thanks for working on this :+1:

PR #4138, opened 2024-07-31
impl pg array append

guissalustiano:
- **impl array_append**
- **allow non nullable range** - can be discarded if we don't want the polymorphism
- **typo** - the CI pointed out the typo and I think it isn't worth opening a new PR for that.

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -705,23 +705,23 @@ define_sql_function! {
     /// #     use diesel::dsl::array_append;
     /// #     use diesel::sql_types::{Nullable, Integer, Array};
     /// #     let connection = &mut establish_connection();
-    /// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![1, 2], 3))
+    /// let ints = diesel::select(array_append::<Array<_>, Integer, _, _>(vec![1, 2], 3))
     ///     .get_result::<Vec<i32>>(connection)?;
     /// assert_eq!(vec![1, 2, 3], ints);
     ///
-    /// let ints = diesel::select(array_append::<Nullable<Integer>, Array<_>, _, _>(vec![Some(1), Some(2)], None::<i32>))
+    /// let ints = diesel::select(array_append::<Array<_>, Nullable<Integer>, _, _>(vec![Some(1), Some(2)], None::<i32>))
     ///     .get_result::<Vec<Option<i32>>>(connection)?;
     /// assert_eq!(vec![Some(1), Some(2), None], ints);
     ///
-    /// let ints = diesel::select(array_append::<Integer, Nullable<Array<_>>, _, _>(None::<Vec<i32>>, 3))
+    /// let ints = diesel::select(array_append::<Nullable<Array<_>>, Integer, _, _>(None::<Vec<i32>>, 3))
     ///     .get_result::<Vec<i32>>(connection)?;
     /// assert_eq!(vec![3], ints);
     ///
-    /// let ints = diesel::select(array_append::<Nullable<Integer>, Nullable<Array<_>>, _, _>(None::<Vec<i32>>, None::<i32>))
+    /// let ints = diesel::select(array_append::<Nullable<Array<_>>, Nullable<Integer>, _, _>(None::<Vec<i32>>, None::<i32>))
     ///     .get_result::<Vec<Option<i32>>>(connection)?;
     /// assert_eq!(vec![None], ints);
     /// #     Ok(())
     /// # }
     /// ```
-    fn array_append<T: SingleValue, Arr: ArrayOrNullableArray<Inner=T> + SingleValue>(a: Arr, e: T) -> Array<T>;
+    fn array_append<Arr: ArrayOrNullableArray<Inner=T> + SingleValue, T: SingleValue>(a: Arr, e: T) -> Array<T>;
 }
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -705,11 +705,11 @@ define_sql_function! {
     /// #     use diesel::dsl::array_append;
     /// #     use diesel::sql_types::{Integer, Array};
     /// #     let connection = &mut establish_connection();
-    /// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![Some(1), Some(2)], Some(3)))
-    ///     .get_result::<Vec<Option<i32>>>(connection)?;
-    /// assert_eq!(vec![Some(1), Some(2), Some(3)], ints);
+    /// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![1, 2], 3))
+    ///     .get_result::<Vec<i32>>(connection)?;
+    /// assert_eq!(vec![1, 2, 3], ints);
     /// #     Ok(())
     /// # }
     /// ```
-    fn array_append<T: SingleValue, Arr: ArrayOrNullableArray<Inner=Nullable<T>> + SingleValue>(a: Arr, e: Nullable<T>) -> Array<Nullable<T>>;
+    fn array_append<T: SingleValue, Arr: ArrayOrNullableArray<Inner=T> + SingleValue>(a: Arr, e: T) -> Array<T>;
 }
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -512,9 +512,11 @@ define_sql_function! {
     /// # }
     /// #
     /// # fn main() {
+    /// #     #[cfg(feature = "numeric")]
     /// #     run_test().unwrap();
     /// # }
     /// #
+    /// # #[cfg(feature = "numeric")]
     /// # fn run_test() -> QueryResult<()> {
     /// #     use self::posts::dsl::*;
     /// #     use std::collections::Bound;
@@ -560,9 +562,11 @@ define_sql_function! {
     /// # }
     /// #
     /// # fn main() {
+    /// #     #[cfg(feature = "time")]
     /// #     run_test().unwrap();
     /// # }
     /// #
+    /// # #[cfg(feature = "time")]
     /// # fn run_test() -> QueryResult<()> {
     /// #     use self::posts::dsl::*;
     /// #     use std::collections::Bound;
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -687,3 +688,28 @@ define_sql_function! {
...
+    /// #     use diesel::dsl::array_append;
+    /// #     use diesel::sql_types::{Integer, Array};
+    /// #     let connection = &mut establish_connection();
+    /// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![Some(1), Some(2)], Some(3)))
+    ///     .get_result::<Vec<Option<i32>>>(connection)?;
+    /// assert_eq!(vec![Some(1), Some(2), Some(3)], ints);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "postgres_backend")]
```
Ten0, line 713: Shouldn't this be outside of the macro call?

diesel/src/pg/expression/functions.rs
```diff
@@ -687,3 +688,28 @@ define_sql_function! {
...
+    fn array_append<T: SingleValue, Arr: ArrayOrNullableArray<Inner=Nullable<T>> + SingleValue>(a: Arr, e: Nullable<T>) -> Array<Nullable<T>>;
```
Ten0, line 714: It doesn't look like there should be Nullable here : the fact that pg arrays are always Nullable is already expressed to the Rust type system by the fact that we export them to the `schema.rs` as `Array<Nullable<T>>`, and if the schema was patched to express that the type of the array is in fact `Array<T>` we should still be able to use this function by providing an expression of sql type T.

Ten0 (changes requested):

diesel/src/pg/expression/functions.rs
Ten0: It looks like this can be achieved by having `Nullable<T>` as ElementType for ArrayOrNullableArray

diesel/src/pg/expression/functions.rs
guissalustiano: I keep the pattern from here
https://github.com/diesel-rs/diesel/blob/444bab7eb1c054fbcde123c45d9fd671ebf79973/diesel/src/pg/expression/functions.rs#L9-L13

I double-check with `cargo expand`, and the output follows. The docs are in the correct position, and the feature is in all generated code but duplicated.
```rust
# cfg inside
❯ cargo expand -F postgres pg::expression::functions::array_append
    Checking diesel v2.2.0 (/.../diesel/diesel)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.01s

/// Append an element to the end of an array.
/// # Example
///
/// ```rust
/// # include!("../../doctest_setup.rs");
/// #
/// # fn main() {
/// #     run_test().unwrap();
/// # }
/// #
/// # fn run_test() -> QueryResult<()> {
/// #     use diesel::dsl::array_append;
/// #     use diesel::sql_types::{Integer, Array};
/// #     let connection = &mut establish_connection();
/// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![Some(1), Some(2)], Some(3)))
///     .get_result::<Vec<Option<i32>>>(connection)?;
/// assert_eq!(vec![Some(1), Some(2), Some(3)], ints);
/// #     Ok(())
/// # }
/// ```
pub fn array_append<
    T: SingleValue,
    Arr: ArrayOrNullableArray<Inner = Nullable<T>> + SingleValue,
    a,
    e,
>(a: a, e: e) -> array_append<T, Arr, a, e>
where
    a: ::diesel::expression::AsExpression<Arr>,
    e: ::diesel::expression::AsExpression<Nullable<T>>,
{
    array_append_utils::array_append {
        a: a.as_expression(),
        e: e.as_expression(),
        T: ::std::marker::PhantomData,
        Arr: ::std::marker::PhantomData,
    }
}
#[allow(non_camel_case_types, non_snake_case)]
///The return type of [`array_append()`](super::fn_name)
pub type array_append<T, Arr, a, e> = array_append_utils::array_append<
    T,
    Arr,
    <a as ::diesel::expression::AsExpression<Arr>>::Expression,
    <e as ::diesel::expression::AsExpression<Nullable<T>>>::Expression,
>;
```

And with the `cfg` outside the macro, it removes the duplicated `cfg` in the generated code.
```rust
# cfg outside
❯ cargo expand -F postgres pg::expression::functions::array_append
    Blocking waiting for file lock on build directory
    Checking diesel v2.2.0 (/home/guiss/projetos/open_source/diesel/diesel)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 22.55s

/// Append an element to the end of an array.
/// # Example
///
/// ```rust
/// # include!("../../doctest_setup.rs");
/// #
/// # fn main() {
/// #     run_test().unwrap();
/// # }
/// #
/// # fn run_test() -> QueryResult<()> {
/// #     use diesel::dsl::array_append;
/// #     use diesel::sql_types::{Integer, Array};
/// #     let connection = &mut establish_connection();
/// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![Some(1), Some(2)], Some(3)))
///     .get_result::<Vec<Option<i32>>>(connection)?;
/// assert_eq!(vec![Some(1), Some(2), Some(3)], ints);
/// #     Ok(())
/// # }
/// ```
pub fn array_append<
    T: SingleValue,
    Arr: ArrayOrNullableArray<Inner = Nullable<T>> + SingleValue,
    a,
    e,
>(a: a, e: e) -> array_append<T, Arr, a, e>
where
    a: ::diesel::expression::AsExpression<Arr>,
    e: ::diesel::expression::AsExpression<Nullable<T>>,
{
    array_append_utils::array_append {
        a: a.as_expression(),
        e: e.as_expression(),
        T: ::std::marker::PhantomData,
        Arr: ::std::marker::PhantomData,
    }
}
#[allow(non_camel_case_types, non_snake_case)]
///The return type of [`array_append()`](super::fn_name)
pub type array_append<T, Arr, a, e> = array_append_utils::array_append<
    T,
    Arr,
    <a as ::diesel::expression::AsExpression<Arr>>::Expression,
    <e as ::diesel::expression::AsExpression<Nullable<T>>>::Expression,
>;
```

You are right, thanks!
I can open a PR later to fix the others

diesel/src/pg/expression/functions.rs
```diff
@@ -687,3 +688,28 @@ define_sql_function! {
...
+    /// #     use diesel::dsl::array_append;
+    /// #     use diesel::sql_types::{Integer, Array};
+    /// #     let connection = &mut establish_connection();
+    /// let ints = diesel::select(array_append::<Integer, Array<_>, _, _>(vec![1, 2], 3))
+    ///     .get_result::<Vec<i32>>(connection)?;
+    /// assert_eq!(vec![1, 2, 3], ints);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn array_append<T: SingleValue, Arr: ArrayOrNullableArray<Inner=T> + SingleValue>(a: Arr, e: T) -> Array<T>;
```
Ten0, line 714: We should have the following two properties (unless I'm mistaken on how postgresql behaves) 

1. The SQL type of the resulting expression is Nullable depending on whether Arr is Nullable (assuming postgres returns NULL if passing a NULL array) 
2. The SQL type of the resulting expression is *not* Nullable depending on whether T is Nullable (assuming that postgres would append a NULL to the array, not make the whole resulting expression NULL) 

Can you write tests that showcase that both these things work as expected? 

Thanks 😊🙏

It is fine if the 
4.

guissalustiano: Thanks for the review!

The pg behavior for this functions with NULL is listed bellow:
 1.  `SELECT array_append(ARRAY[1,2], 3)` returns `ARRAY[1,2,3]`
 1.  `SELECT array_append(ARRAY[1,2], NULL)` returns `ARRAY[1, 2, NULL]`
 1. `SELECT array_append(NULL, 3)` returns `ARRAY[3]`
 1. `SELECT array_append(NULL, NULL)` returns `ARRAY[NULL]`
 
 > 1. The SQL type of the resulting expression is Nullable depending on whether Arr is Nullable (assuming postgres returns NULL if passing a NULL array)
 
 If passing NULL, postgres apparently handles it as an empty array, so I can't think of a way that this function returns NULL.
 
 > 2. The SQL type of the resulting expression is not Nullable depending on whether T is Nullable (assuming that postgres would append a NULL to the array, not make the whole resulting expression NULL)
 
 I see, thanks for the clarification. Do you know how to do that? That's why I put Nullable in everything before.
 
 > Can you add tests that showcase that both these things work as expected?
 
 Sure! Should I create doc tests or normal ones?
 
 Thanks!

Ten0: Ah I thought that we may have to involve the nullability propagation helpers traits here to construct the return type, which would maybe have involved either declaring it without the macro or [adding a feature to do so in the macro](https://github.com/diesel-rs/diesel/blob/4763e7c9c968e20d7dff7e23ba027fd1de30c37d/diesel_derives/src/sql_function.rs#L124), however given what you're saying (3 and 4) it seems that we only ever want to return an array, regardless of original nullability of the array, so it seems that the current signature is correct. 😊
(Which is furtunate because that turns out to be significantly simpler.) 

It's fine as a first version if a user needs the input array to already have its elements be Nullable to be able to append potentially null values to it. That may be incomplete (although I doubt there would be many use cases), but most importantly it's correct 😊. 
What wouldn't have been fine is if it could return nulls but that wasn't expressed in the signature (1), or if by fixing (1) we accidentally made (2) wrong (by considering the arguments in the same way, in a mode where if any argument is Nullable resulting value is Nullable, which IIRC is already our default behavior for operators but apparently not for functions - what I meant is in that case we couldn't just "enable" that behavior for functions as well).

OK so overall, current implementation looks good, we can probably just add tests to make sure that we get the correct type for all 4 scenarios, with original type of the array already expressed. 

I have no strong opinion on where each test for the 4 scenarios you mentioned should live, but maybe another reviewer will. By default unless too verbose I'd probably just have them all in the one doctest.

guissalustiano: Thanks! I added the test as a docstring

weiznich (approved):
Looks good now. Thanks for submitting

(I adjusted the generic argument order to match the function argument order to make that more consistent and to possible have `#[auto_type]` support there as well)

Ten0 (approved):

Ten0:
Ah I thought there was only one failing but there were 2... Probably that will fail again then...
It's only on Ubuntu so I would suspect an ubuntu-related package upgrade...

Ten0:
Ah I thought there was only one failing but there were 2... Probably that's not random will fail again then...
It's only on Ubuntu so I would suspect an ubuntu-related package upgrade...
That's the difference between the two runs during PG install:
image
Nothing striking...
We'd have to get the segfault's backtrace. In any case it seems very unlikely that it's related to this PR so we can probably merge this anyway.

PR #4142, opened 2024-08-02
Update bench (postgres) : Added `hair_color` filter to `complex_query_by_name`

nurmohammed840:


weiznich (approved):
Thanks for opening this PR. It seems like that was just missed to be updated at some point.
I pushed another commit that also updates the mysql/rusqlite benchmark to fix the same issue there.

PR #4151, opened 2024-08-08
Add the last missing range operators

### LINKED ISSUE #4092: Add support for currently unsupported range operators and methods

weiznich:
This commit adds the last missing range operators.

Additionally it fixes the `#[auto_type]` support for the existing range functions. It also goes over the documentation again and fixes minor inconsistencies there. Finally a change log entry is added.

Fixes #4092

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -11,7 +11,7 @@ use crate::expression::grouped::Grouped;
 use crate::expression::operators::{Asc, Concat, Desc, Like, NotLike};
 use crate::expression::{AsExpression, Expression, IntoSql, TypedExpressionType};
 use crate::pg::expression::expression_methods::private::BinaryOrNullableBinary;
-use crate::sql_types::{Array, Inet, Integer, SqlType, Text, VarChar};
+use crate::sql_types::{Array, Inet, Integer, Range, SqlType, Text, VarChar};
 use crate::EscapeExpressionMethods;
 
 /// PostgreSQL specific methods which are present on all expressions.
@@ -71,6 +71,89 @@ pub trait PgExpressionMethods: Expression + Sized {
         T: AsExpression<Self::SqlType>,
     {
         Grouped(IsDistinctFrom::new(self, other.as_expression()))
+    }
+
+    /// Creates a PostgreSQL `<@` expression.
+    ///
+    /// This operator returns true whether a element is contained by a range
+    ///
+    /// This operator evaluates to true for the following cases:
+    ///
+    /// ```text
+    /// self:     |
+    /// other:  [-----]
+    /// ```
+    ///
+    /// ```text
+    /// self:  |
+    /// other: [----]
+    /// ```
+    ///
+    /// ```text
+    /// self:       |
+    /// other:  [----]
+    /// ```
+    ///
+    /// This operator evaluates to false for the following cases:
+    ///
+    /// ```text
+    /// self:           |
+    /// other:  [-----]
+    /// ```
+    ///
+    /// ```text
+    /// self:  |
+    /// other:   [----]
+    /// ```
+    ///
+    /// ```text
+    /// self:       |
+    /// other: [----)
+    /// ```
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # table! {
+    /// #     posts {
+    /// #         id -> Integer,
+    /// #         versions -> Range<Integer>,
+    /// #     }
+    /// # }
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use std::collections::Bound;
+    /// #     use diesel::dsl::int4range;
+    /// #     use diesel::sql_types::Integer;
+    /// #
+    /// #     let conn = &mut establish_connection();
+    /// #
+    ///       let my_range = int4range(1, 5, diesel::sql_types::RangeBound::LowerBoundInclusiveUpperBoundExclusive);
+    ///
+    ///       let (first, second) = diesel::select((
+    ///           4.into_sql::<Integer>().is_contained_by_range(my_range),
+    ///           10.into_sql::<Integer>().is_contained_by_range(my_range)
+    ///       )).get_result::<(bool, bool)>(conn)?;
+    ///
+    ///       assert_eq!(first, true);
+    ///       assert_eq!(second, false);
+    /// #
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[allow(clippy::wrong_self_convention)] // This is named after the sql operator
+    fn is_contained_by_range<T>(self, other: T) -> dsl::IsContainedByRange<Self, T>
+    where
+        Self::SqlType: SqlType,
+        T: AsExpression<Range<Self::SqlType>>,
+    {
+        Grouped(IsContainedBy::new(self, other.as_expression()))
     }
 }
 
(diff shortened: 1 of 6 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -916,7 +965,11 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
 
     /// Creates a PostgreSQL `&<` expression.
     ///
-    /// This operator returns whether a range doesn't extend to the right of another
+    /// This operator returns true whether the argument range extend to the right of the current range
+    ///
+    /// Postgresql defines "extends" as does not have a lower bound smaller than the lower bound of the
+    /// self range. That means the right hand side range can overlap parts of the left hand side
+    /// range or be on the right side of the left hand side range
```
Mingun, line 972: A small ASCII art would be useful here and in other range operators

weiznich: That's a great idea, I updated the PR to include those examples as well.

guissalustiano (approved):
Thanks to finish this! I'm not a official review but becase I implement some of this features I think I can help with a review.
That's sad, everthing looks good for me, I only can't find the ansi art that you mention above

weiznich:
Reviews are always welcome even if you are not part of the review team. If you want it's also possible to join that team. In the end it's just a group of people that get pinged for such reviews. 

ASCII art refers to this little diagrams added to all operators: 
https://github.com/diesel-rs/diesel/pull/4151/files#diff-bf88b4900340a435caafd7c947e7b9a95e61ff158ef37596ab3262b3c754c45fR82-R84

biwecka:
Hey, really excited to see range operator support in diesel! Is there anything planned when they’ll make it into a stable release? If there’s anything still outstanding I’d be happy to help push it over the line. Thanks!

guissalustiano:
Hey @biwecka, weiznich answer about that [here](https://github.com/diesel-rs/diesel/discussions/4624)

weiznich:
I would like to land at least #4685,#4688 and #4690 before cutting a new minor release. 

#4600 might also be interesting to land before cutting the release, but that depends on the feedback on that PR. 

You can help getting there faster by reviewing these PR, and by helping fixing the remaining issue mentioned in #4624

PR #4158, opened 2024-08-11
Remove call to deprecated array::IntoIter::new

ADSteele916:
Since Diesel's MSRV is now 1.78, the call to `std::array::IntoIter::new` and the`#[allow(deprecated)]` attribute are no longer necessary.


weiznich (approved):
Thanks :+1:

PR #4159, opened 2024-08-12
Add support for the postgres multirange type

guissalustiano:
This PR implements the Postgres multirange types


guissalustiano:
Tests fail on Windows because the minimum version for multirange is [postgres 14](https://www.postgresql.org/about/news/postgresql-14-released-2318/), released on 09/2021, and [windows install postgres 12](https://github.com/diesel-rs/diesel/blob/c4822088da5b1cb51164cec2bfe3ca0864e32e70/.github/workflows/ci.yml#L163)

The other failure is related to the protest
```
[quickcheck] TEST FAILED (runtime error). Arguments: ([(-210866803201, 0, 0, 0)])
Error: "Query failed: DatabaseError(Unknown, \"timestamp out of range\") -> value: [(Included(-4713-11-23T23:59:59), Excluded(1970-01-01T00:00:00))]"
```
I'm not sure how the protest for `Range<Timestamp>` works for this, but we need to reduce the test domain on these tests (or just test as `vec![(Bound::, Bound::)`.


And 
```
[quickcheck] TEST FAILED. Arguments: ([(-211785610425, 0, -210004501573, 0), (-1, 0, 0, 0)])
```
I don't have any idea why it fails, but looks like a db error again

weiznich:
For the windows CI error: Sounds like we just need to install a newer Postgres version there. It's fine to include that as part of the PR. 

For the proptest error: I'm not at my computer but I will have a look in the next days and see what von be done there.

diesel/src/pg/types/multirange.rs
```diff
@@ -0,0 +1,111 @@
...
+where
+    T: ToSql<ST, Pg>,
+    [(Bound<T>, Bound<T>)]: ToSql<Multirange<ST>, Pg>,
+{
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        ToSql::<Multirange<ST>, Pg>::to_sql(self.as_slice(), out)
+    }
+}
+
+// TODO: nullable impl
```
weiznich, line 111: Is that todo still relevant?

diesel/src/pg/types/multirange.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        out.write_u32::<NetworkEndian>(self.len() as u32)?;
+
+        let mut buffer = Vec::new();
+        for value in self {
+            {
+                let mut inner_buffer = Output::new(ByteWrapper(&mut buffer), out.metadata_lookup());
+                ToSql::<Range<ST>, Pg>::to_sql(&value, &mut inner_buffer)?;
+            }
+            out.write_u32::<NetworkEndian>(buffer.len() as u32)?;
```
weiznich, line 91: It might be worth to convert that into a failable cast via `TryInto`, just to make sure that we don't overflow the u32 here. It might be an edge case, but better be save than sorry. (I really should check the other variable sized `ToSql` impls for such issues, at some point)

weiznich (changes requested):
Thanks for opening this PR. That's really helpful work :pray: 

I have some small note about the `ToSql` implementation + a minor question. Otherwise I would like to see an `Changelog.md` entry for this, as this is a relevant new feature. Otherwise this should be fine to merge.

diesel/src/pg/types/multirange.rs
guissalustiano: Ops, yes! Thanks for notice

diesel/src/pg/types/multirange.rs
guissalustiano: I agree, I copied from the Range implementation it should have been implemented before `TryFrom` was stable.

guissalustiano: [I fix this in range too](https://github.com/diesel-rs/diesel/pull/4159/commits/f75e5d856eef3d8b60e2d2f366215a22ea78816d)****

weiznich: Thanks. The array `ToSql` impl does seem to suffer from the same problem: https://github.com/diesel-rs/diesel/blob/c4822088da5b1cb51164cec2bfe3ca0864e32e70/diesel/src/pg/types/array.rs#L104

weiznich (approved):

diesel/src/pg/types/multirange.rs
guissalustiano: Thanks, I open a PR for that: https://github.com/diesel-rs/diesel/pull/4166/files

PR #4161, opened 2024-08-12
Selecting custom caching strategy

fraillt:
Hello,
Following #4058 discussion (about statement caches), I want to propose a possible solution.
Since rebuilding statement cache on error is not an option (at least for PostgreSQL). The only convenient option that I see, is to give control to end user, to decide how he wants to use statement cache (of course user can implement it's own connection type, but that is not convenient option, as most likely user also needs to implement r2d2 traits as well).

There's the core idea:
* decision (whether to cache or not) is maid by `StatementCacheStrategy` trait. To implementations are already provided: `WithCacheStrategy` (default), `WithoutCacheStrategy`.
* user can choose what strategy to use by setting `set_cache_strategy` globally per connection type (because traits are generic over DB and Statement).
* there's no breaking changes for end users (maybe there are breaking changes if users already use `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature flags).

Notable changes:
* introduced `ConnectionStatementCache` that is implemented by all connections which will use statement cache. This trait has two purposes: 1) exposes functionality for end user to set custom caching strategy, 2) `StatementCache` use this, to actually get current caching strategy. Another option was to implement this directly on `Connection`, but that would require adding new `type Statement` associated type parameter, and I didn't wanted to deal with `MultiConnection` macro.
* removed `len` from `StatementCache`. It was used by PostgreSQL implementation to generate statement cache names, but this might easily break if user provides it's own implementation for statement cache. Instead, `StatementCache` implementation provide a `cache_counter: Option<u64>` that is incremented every time if statement will be cached, or None otherwise.
* introduced `IntrospectCachingStrategy` for tests only (behind `cfg(test)`), that wraps actual `StatementCacheStrategy` and records outcome. This is used in tests to check if tests related to statement caching behave correctly. (previously `len` field from `StatementCache` was utilized, but this is more precise).


weiznich:
Thanks for opening this PR and working on this issue.

That written: I'm not fond of the chosen approach. I personally do not see why rebuilding the statement isn't an option. Yes it might fail for PostgreSQL in a transaction, but on the other hand there is so many other things that might file and render a transaction unusable that I don't feel that this really matters. If you care about such cases you must implement a transaction retry logic anyway.

There are several things to consider with the proposed solution, that I personally not a fan of:
* It introduces global state, which is something that we want to avoid as best as possible
* It exposes types that were previously considered implementation details. By doing that they become part of the stable API, which in turn would make it really hard to change them later. They were private on propose.
* I'm not sure of the performance implications of this kind of change. Especially for the SQLite backend the additional dyn dispatch could come with a significant cost.

I'm sorry if that sounds like I want to dismiss your work, but I just have different ideas in mind on how to approach that problem. I personally would have preferred to discuss the design of this feature first before posting a PR, as that would likely result in much less frustration on both sides.

fraillt:
Since this is not a huge feature, I like to start discussion from some point. Real implementations usually have some subtle nuances and you also see full scope, these things are hard to discuss up-front.
And please, don't feel frustrated, - it's ok that you don't like it :)

One important thing, regarding transactions. YMMV, but for us transactions fail only for network issues or breaking-scheme-change issues. Breaking-scheme-change we solve by properly managing migrations in non-breaking change manner, the rest (network issues) we simply propagate to upstream, as technical problems (which might be retried later or not).
Implementing transaction retry, just for this one type of error, would be very painful for us, because that would mean, that we need to update all functions that is written in 4+ years from `FnOnce` to `Fn`, while disabling cache would make some breaking-scheme-change into non-breaking-scheme-change and make life significantly simpler :)

Before discussing further points, I think that we need to agree on one fundamental point first:
- do you consider an option, where user can choose custom caching strategy?

weiznich:
I'm open to allow configuring the cache, but I would like to make that option connection specific rather than global. So just `PgConnection::set_cache_size(CacheSize)` with cache size having 3 variants (`Unbound` (default), `Size(NonZeroUsize)` and `Disabled`). This likely needs some adjustments on the cache hashmap to drop values for the `Size` variant, otherwise I expect this to be a rather simple change.

That written: I would prefer to have the automatic retry for statements nevertheless as at least I do want to use the statement cache whenever possible as it gives a quite large performance boost.

fraillt:
Let me think out lout :).
Ideally, we would like to satisfy all of these constrains for a service:
* service must be running while we run migrations.
* service should run efficiently, e.g. utilize statement cache.
* service uses connection pools to handle many concurrent requests

To achieve that, migration process could be implemented like this:
* disable caching at runtime (we'll probably send some sort of trigger to a service, we need to somehow disable caching for all subsequent queries)
* run migrations
* enable caching at runtime again.

In order to implement caching behavior at runtime, we would wrap r2d2 connection pool into our own and every time we get connection from the pool, we would `set_cache_size`. If underlying implementation is efficient, it should be no-op if we set the same value as it was before.

I probably it's a bit more work on end user side (to wrap connection), but actually that would solve our problem :)

So I guess we'll come back to you with updated revision ;)

weiznich:
That sounds like a good summary, thanks for writing it up. I think it would be also fine to have an additional function that just empties the current query cache.

fraillt:
Some comments:
* regarding cache clearing, I think it's not necessary, as you can disable & enable and achieve cache clearing this way.
* you raised concern about dynamic dispatch performance. I did some testing locally (on my laptop `13th Gen Intel® Core™ i7-1355U × 12`). Overhead is around 2.1ms (+/- 0.1ms) per 1M calls, (or ~2ns per call)). So I think we can ignore it.
* I had to expose some fields `pub(crate) statement_cache` in order to proper testing using `IntrospectCachingStrategy`.
* I did not implemented `Size(NonZeroUsize)` caching strategy, to be honest I'm not sure what would be the use case...

Happy reviewing and waiting for your comments :)

diesel/src/sqlite/connection/mod.rs
```diff
@@ -351,7 +357,15 @@ impl SqliteConnection {
             &source,
             &Sqlite,
             &[],
-            |sql, is_cached| Statement::prepare(raw_connection, sql, is_cached),
+            |sql, counter| {
```
weiznich, line 360: I would prefer forwarding the cache information explicitly here instead of relaying on the workaround using the optional counter. It should be fine to just have it as third function argument.

diesel/src/test_helpers.rs
```diff
@@ -60,7 +60,13 @@ pub fn pg_connection() -> PgConnection {
 
 #[cfg(feature = "postgres")]
 pub fn pg_connection_no_transaction() -> PgConnection {
-    PgConnection::establish(&pg_database_url()).unwrap()
+    use crate::connection::statement_cache::strategy::{
```
weiznich, line 63: I think I would prefer to keep the tests for this functionality local to the `strategy.rs` file. That means:

* Don't setup a custom strategy for the global tests
* Have a few tests to check all cases (cache vs not cache, cachable statement vs not cacheable statement) in that module. 

That should make the changeset a bit smaller and the tests more focused on the actual desired behavior.

weiznich (changes requested):
Thanks for the update. This is now much closer to what I would expect in terms of public API. I've left a bunch of mostly minor comments. The only major thing I would like to see changed is how this feature is tested. Instead of having the tests "everywhere" I would like to focus them on only this feature inside of the `strategy` module.

fraillt:
I think I have addressed all your comments.
I have removed `IntrospectCachingStrategy` because using `Instrumentation` is enough to test caching.

diesel/src/connection/statement_cache/strategy.rs
```diff
@@ -0,0 +1,392 @@
...
+        assert_eq!(0, count_cache_calls(connection));
+    }
+
+    #[test]
+    fn inserts_from_select_are_cached() {
+        let connection = &mut connection();
+        connection.begin_test_transaction().unwrap();
+
+        crate::sql_query(
+            "CREATE TABLE IF NOT EXISTS users(id INTEGER PRIMARY KEY, name TEXT NOT NULL);",
```
weiznich, line 241: Can we use `CREATE TEMPORARY TABLE` statements here instead? That prevents possible conflicts.

diesel/src/connection/statement_cache/strategy.rs
```diff
@@ -0,0 +1,392 @@
...
+            .downcast_ref::<RecordCacheEvents>()
+            .unwrap()
+            .list
+            .len()
+    }
+}
+
+#[cfg(test)]
+#[cfg(feature = "postgres")]
+mod tests_pg {
```
weiznich, line 145: Would it be possible to add another test that shows that disabling the cache has an effect?

diesel/src/connection/statement_cache/strategy.rs
```diff
@@ -0,0 +1,392 @@
...
+    use crate::{
+        dsl::sql,
+        expression::functions::define_sql_function,
+        insertable::Insertable,
+        macros::table,
+        pg::Pg,
+        sql_types::{Integer, VarChar},
```
weiznich, line 152: We don't have such nested imports in diesel otherwise. I would prefer to keep the style consistent.

weiznich:
Thanks for the update. I really like the solution to test this via the instrumentation tooling. 
It looks nearly ready now. I added a number of minor comments about stylistic things and changes that are not required anymore. The only larger point that still remains is adding at least one test that checks the disabled cache behaviour.

weiznich (approved):

fraillt:
Hi, @weiznich 
Thanks for merging this!
I was on vacation for a while and didn't managed to address your comments on time...
Should I create a new pull request to address your comments, or it's not worth it?

weiznich:
@fraillt No worries, holiday is important and you shouldn't response to anything there. I already fixed the last few comments in https://github.com/diesel-rs/diesel/pull/4161/commits/bda514b4f88110a64d31f76bada2e81c599477a9, so there is nothing that needs to be done here.

PR #4168, opened 2024-08-14
Convert primary keys to serialize_as type

Aethelflaed:
I've got a newtype on oxydized_money::Currency to implement ToSql/FromSql/...

I recently added a table with a composite primary key which included a currency column, and I'm expecting to be able to use it the same way I used another id.

```rust
#[derive(Debug, Queryable, Selectable, Identifiable)]
#[diesel(table_name = monthly_stats)]
#[diesel(primary_key(year, month, currency))]
#[diesel(check_for_backend(diesel::sqlite::Sqlite))]
pub struct MonthlyStats {
    pub year: i32,
    pub month: i32,
    #[diesel(deserialize_as = db::Decimal)]
    pub amount: Decimal,
    #[diesel(deserialize_as = db::Currency, serialize_as = db::Currency)]
    pub currency: Currency,
}
```

Everything seems fine, until I try to compile the following instruction:

```rust
        diesel::update(&*self)
            .set(monthly_stats::amount.eq(db::Decimal::from(self.amount)))
            .execute(conn)?;
```

> error[E0277]: the trait bound `SelectStatement<FromClause<monthly_stats::table>>: FilterDsl<expression::grouped::Grouped<expression::operators::And<expression::grouped::Grouped<expression::operators::Eq<monthly_stats::columns::year, expression::bound::Bound<diesel::sql_types::Integer, &i32>>>, expression::grouped::Grouped<expression::operators::And<expression::grouped::Grouped<expression::operators::Eq<monthly_stats::columns::month, expression::bound::Bound<diesel::sql_types::Integer, &i32>>>, expression::grouped::Grouped<expression::operators::Eq<monthly_stats::columns::currency, &oxydized_money::Currency>>>>>>>` is not satisfied

Or as explained a few lines later:
> note: required by a bound in `update`
> `  pub fn update<T: IntoUpdateTarget>(source: T) -> UpdateStatement<T::Table, T::WhereClause> {`
>  the trait `diesel::Expression` is not implemented for `oxydized_money::Currency`, which is required by `&MonthlyStats: IntoUpdateTarget`

I can fix that by implementing `Identifiable` and `HasTable` manually, or by compiling against the provided patch

It works because my type is `Copy`, but at the same time this isn't uncommon for id types.

When it's not copy, I get something like:

> cannot move out of `self.field` which is behind a shared reference
> move occurs because `self.field` has type `MyType`, which does not implement the `Copy` trait

Plus some advice on implementing `Clone`, if it's not already `Clone`.

I could maybe find a way of ensuring the type is `Copy` if you think it's better that way, but if we go this way we might as well relax the `AsChangeset` & `Insertable` restriction on `serialize_as` for the same reasons, so it would be a bigger change.

Tell me what you think

### DIFF
--- a/diesel_derives/src/identifiable.rs
+++ b/diesel_derives/src/identifiable.rs
@@ -31,7 +31,7 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
         } = &model.find_column(pk)?;
         if let Some(AttributeSpanWrapper { item: ty, .. }) = serialize_as.as_ref() {
             field_ty.push(quote!(#ty));
-            field_name.push(quote!(::std::convert::Into::<#ty>::into(self.#name)));
+            field_name.push(quote!(::std::convert::Into::<#ty>::into(self.#name.clone())));
         } else {
             field_ty.push(quote!(&'ident #ty));
             field_name.push(quote!(&self.#name));
--- a/diesel_derives/src/identifiable.rs
+++ b/diesel_derives/src/identifiable.rs
@@ -4,6 +4,8 @@ use syn::parse_quote;
 use syn::DeriveInput;
 use syn::Result;
 
+use crate::attrs::AttributeSpanWrapper;
+use crate::field::Field;
 use crate::model::Model;
 use crate::util::wrap_in_dummy_mod;
 
@@ -21,9 +23,19 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
     let mut field_ty = Vec::new();
     let mut field_name = Vec::new();
     for pk in model.primary_key_names.iter() {
-        let f = model.find_column(pk)?;
-        field_ty.push(&f.ty);
-        field_name.push(&f.name);
+        let Field {
+            ty,
+            name,
+            serialize_as,
+            ..
+        } = &model.find_column(pk)?;
+        if let Some(AttributeSpanWrapper { item: ty, .. }) = serialize_as.as_ref() {
+            field_ty.push(quote!(#ty));
+            field_name.push(quote!(::std::convert::Into::<#ty>::into(self.#name)));
+        } else {
+            field_ty.push(quote!(&'ident #ty));
+            field_name.push(quote!(&self.#name));
+        }
     }
 
     Ok(wrap_in_dummy_mod(quote! {
@@ -42,20 +54,20 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
         impl #ref_generics Identifiable for &'ident #struct_name #ty_generics
         #where_clause
         {
-            type Id = (#(&'ident #field_ty),*);
+            type Id = (#(#field_ty),*);
 
             fn id(self) -> Self::Id {
-                (#(&self.#field_name),*)
+                (#(#field_name),*)
             }
         }
 
         impl #ref_generics Identifiable for &'_ &'ident #struct_name #ty_generics
             #where_clause
         {
-            type Id = (#(&'ident #field_ty),*);
+            type Id = (#(#field_ty),*);
 
             fn id(self) -> Self::Id {
-                (#(&self.#field_name),*)
+                (#(#field_name),*)
             }
         }
     }))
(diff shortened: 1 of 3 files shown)


diesel_derives/src/identifiable.rs
```diff
@@ -21,9 +23,19 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
...
-        field_name.push(&f.name);
+        let Field {
+            ty,
+            name,
+            serialize_as,
+            ..
+        } = &model.find_column(pk)?;
+        if let Some(AttributeSpanWrapper { item: ty, .. }) = serialize_as.as_ref() {
+            field_ty.push(quote!(#ty));
+            field_name.push(quote!(::std::convert::Into::<#ty>::into(self.#name)));
```
weiznich, line 34: I wonder if we can change this to 
```suggestion
            field_name.push(quote!(::std::convert::Into::<#ty>::into(self.#name.clone())));
```

This would at least relax the `Copy` requirement and `Clone` is automatically implemented for any `Copy` type. The other alternative would be to require a `From<&Key>` impl instead.

diesel_derives/src/identifiable.rs
```diff
@@ -21,9 +23,19 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
...
-        field_ty.push(&f.ty);
-        field_name.push(&f.name);
+        let Field {
+            ty,
+            name,
+            serialize_as,
+            ..
+        } = &model.find_column(pk)?;
+        if let Some(AttributeSpanWrapper { item: ty, .. }) = serialize_as.as_ref() {
+            field_ty.push(quote!(#ty));
```
weiznich, line 33: I believe that might be seen as breaking change, although one could also argue that this is a bug fix. Pinging @diesel-rs/contributors to get some more opinions on this.

diesel_derives/tests/identifiable.rs
```diff
@@ -150,3 +150,28 @@ fn derive_identifiable_with_composite_pk() {
...
+            MyI32(value)
+        }
+    }
+
+    #[derive(Identifiable)]
+    struct Foo {
+        #[diesel(serialize_as = MyI32)]
+        id: i32,
+        #[allow(dead_code)]
+        foo: i32,
```
weiznich, line 170: That field can just be removed, right?

weiznich (commented):
Thanks for submitting this PR. I think this is a reasonable change, although I have some concerns around backward compatibility here. Depending on what the other members of the contributor team decide we might or might not move forward with this change.

If we move forward we likely want to have a changelog entry for this.

diesel_derives/src/identifiable.rs
Aethelflaed: Shifting the requirement to `Clone` instead will silently clone objects that might have a non-trivial `Clone` implementation (e.g. `Vec`).

I'm not closed to the idea if you think it's best, but I'll wait for other people to review, as per your other comment (I'm not sure it pinged anyone though)

Aethelflaed: Fair enough, and done. Do you want me to write a test-case for that?

weiznich: That would be great to have.

diesel_derives/src/identifiable.rs
kw217: (I'm a recent Diesel reviewer, so don't give my opinion too much weight)
There are two changes here, right?

- One seems to me clearly a bugfix: for a composite primary key, we miss the `&'ident` / `&self` for all but the first element, leading to potential nonsense. We should just fix that.
- The other is less clear: we apply `serialize_as` when we didn't before. Clearly we "should" apply it, but could people be depending on this in some way? I don't have enough experience to know.

Aethelflaed: As far as I know, composite primary key already works (as shown in the tests), so this is not a fix. The `&'ident` / `&self` part was repeated for every element already, as per [`quote!`'s interpolation](https://docs.rs/quote/1.0.9/quote/macro.quote.html#interpolation).

Here I changed `field_ty` and `field_name` to be `Vec<TokenStream>` directly, so I can push what will be needed by the `impl` directly.

weiznich (commented):
I think we go with the current variant. I would like to see an Changelog entry for this before merging otherwise the code + test cases look ok.

Aethelflaed:
> I think we go with the current variant. I would like to see an Changelog entry for this before merging otherwise the code + test cases look ok.

I've added a changelog entry and rebased the PR

weiznich (approved):
Thanks for the update

PR #4170, opened 2024-08-15
Enable some numeric cast releated clippy lints and fix them in the code base

weiznich:
* clippy::cast_possible_wrap
* clippy::cast_possible_truncation
* clippy::cast_sign_loss

These lints can point to serious problems if you hit one of the edge cases in low level unsafe/byte shuffling code

This is a reaction to
https://media.defcon.org/DEF%20CON%2032/DEF%20CON%2032%20presentations/DEF%20CON%2032%20-%20Paul%20Gerste%20-%20SQL%20Injection%20Isn't%20Dead%20Smuggling%20Queries%20at%20the%20Protocol%20Level.pdf

It fixes several places that could be possibly exploited by specially crafted values.

### DIFF
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -244,7 +244,10 @@
     clippy::enum_glob_use,
     clippy::if_not_else,
     clippy::items_after_statements,
-    clippy::used_underscore_binding
+    clippy::used_underscore_binding,
+    clippy::cast_possible_wrap,
+    clippy::cast_possible_truncation,
+    clippy::cast_sign_loss
 )]
 #![deny(unsafe_code)]
 #![cfg_attr(test, allow(clippy::map_unwrap_or, clippy::unwrap_used))]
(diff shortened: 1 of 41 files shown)


weiznich:
I double checked the sqlite benchmarks locally and this doesn't seem to have an impact on performance.

Sytten:
Also for reference, SQLx is also working on it.
https://github.com/launchbadge/sqlx/issues/3440

weiznich:
@diesel/core I would like to issue a patch release next Friday (23.8) with this change and a few others (marked as `maybe-backport`). Are there any objections?

Ten0 (approved):
That's really nice, thanks a lot for working on this!
I'd like to propose the following additional commit (pushed it, feel free to revert) that explicits a few more types to guarantee match between the `expects` and the actual target types in case of change of signature of the underlying called functions.
(Arguably most of them are unlikely to change.)

weiznich:
> I'd like to propose the following additional commit that explicits a few more types to guarantee match between the expects and the actual target types in case of change of signature of the underlying called functions.
(Arguably most of them are unlikely to change.)
(pushed it, feel free to revert)

That's a good idea. Thanks for pushing this change.

Sytten:
@weiznich If you want to add a regression test here is what sqlx did https://github.com/launchbadge/sqlx/pull/3441/commits/f9e51769e905cbc1be54122b20f12b025ee7eaf0

Issue #4153, opened 2024-08-09
Add support for currently unsupported array functions


weiznich:
Diesel currently supports the postgres array types. We do not provide built-in support for [various operators and methods](https://www.postgresql.org/docs/current/functions-array.html) available for these types. This is a tracking issue for adding support for these operators. 

The general strategy for adding support for new methods is as following:

1. Define the operator via [`define_sql_function!()`](https://docs.diesel.rs/master/diesel/expression/functions/macro.define_sql_function.html). These operators can be defined [here](https://github.com/diesel-rs/diesel/blob/861893c28673841595805586c059bced5f314163/diesel/src/pg/expression/functions.rs#L698-L733). See the linked definition of `array_append` for an example. If there is already an existing definition, this step could be skipped. This function should have a short documentation snippet with en example
2. If the function is generic, add a helper type definition [here](https://github.com/diesel-rs/diesel/blob/6c1cb1d42eb2b3000502d67249dc3cdd76066dd7/diesel/src/pg/expression/helper_types.rs#L351-L354). Again see the linked definition for an example for the `array_append` function.
3. Add a test for  `#[auto_type]` support for the newly added function [here](https://github.com/diesel-rs/diesel/blob/6c1cb1d42eb2b3000502d67249dc3cdd76066dd7/diesel_derives/tests/auto_type.rs#L413)
6. Check if compile test output is up to date by running `TRYBUILD=overwrite cargo test` inside of `diesel_compile_tests`. 
5. Submit a PR with the change

Method list:

- [x] `array_append`: ` ( anycompatiblearray, anycompatible ) → anycompatiblearray`: Appends an element to the end of an array (same as the anycompatiblearray || anycompatible operator).
- [ ] `array_cat`: `( anycompatiblearray, anycompatiblearray ) → anycompatiblearray`: Concatenates two arrays (same as the anycompatiblearray || anycompatiblearray operator). *
- [ ] `array_dims`: `( anyarray ) → text`: Returns a text representation of the array's dimensions. *
- [ ] `array_fill`: `( anyelement, integer[] [, integer[] ] ) → anyarray`: Returns an array filled with copies of the given value, having dimensions of the lengths specified by the second argument. The optional third argument supplies lower-bound values for each dimension (which default to all 1). **
- [ ] `array_lenght`: ` ( anyarray, integer ) → integer`: Returns the length of the requested array dimension. (Produces NULL instead of 0 for empty or missing array dimensions.) *
[...]

PR #4178, opened 2024-08-17
Implement array_ndims function

danila-b:
Adding support for `array_ndims` array [postgresql ](https://www.postgresql.org/docs/current/functions-array.html) function.
Connected issue: https://github.com/diesel-rs/diesel/issues/4153

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1380,14 +1380,10 @@ define_sql_function! {
     /// #     use diesel::sql_types::{Nullable, Array, Integer};
     /// #     let connection = &mut establish_connection();
     ///
-    /// let dims = diesel::select(array_ndims::<Array<_>, _>(vec![1, 2]))
+    ///  // diesel currently only supports 1D arrays
+    /// let dims = diesel::select(array_ndims::<Array<Integer>, _>(vec![1, 2]))
     ///     .get_result::<i32>(connection)?;
     /// assert_eq!(1, dims);
-    ///
-    /// let dims = diesel::select(array_ndims::<Array<_>, _>(vec![vec![1, 2], vec![3, 4]]))
-    ///     .get_result::<i32>(connection)?;
-    /// assert_eq!(2, dims);
-    ///
     ///
     /// #     Ok(())
     /// # }
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1361,3 +1361,36 @@ define_sql_function! {
         elem: E,
     ) -> Nullable<Array<Integer>>;
 }
+
+#[cfg(feature = "postgres_backend")]
+define_sql_function! {
+    /// Returns the number of dimensions of the array
+    ///
+    /// # Example
+    ///
+    /// ```rust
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/helper_types.rs
```diff
@@ -352,3 +352,8 @@ pub type range_merge<R1, R2> = super::functions::range_merge<SqlTypeOf<R1>, SqlT
 #[allow(non_camel_case_types)]
 #[cfg(feature = "postgres_backend")]
 pub type array_append<A, E> = super::functions::array_append<SqlTypeOf<A>, SqlTypeOf<E>, A, E>;
+
+/// Return type of [`array_ndims(array)`](super::functions::array_ndims())
+#[allow(non_camel_case_types)]
+#[cfg(feature = "postgres_backend")]
+pub type array_ndims<A> = super::functions::array_ndims<SqlTypeOf<A>, A>;
```
danila-b, line 359: Is this even required for functions that do not have a generic return type and always return `Integer`? 
I tried to remove it but without this generic arguments in `auto_types `explodes.

diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,36 @@ define_sql_function! {
...
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::array_ndims;
+    /// #     use diesel::sql_types::{Nullable, Array, Integer};
+    /// #     let connection = &mut establish_connection();
+    ///
+    /// let dims = diesel::select(array_ndims::<Array<_>, _>(vec![1, 2]))
```
danila-b, line 770: Not sure why, but it fails to infer the types for `Array` here.

danila-b (commented):
The PR is still WIP but I would appreciate any pointers if I am doing something completely wrong from the beginning,

diesel/src/pg/expression/functions.rs
guissalustiano: It works on `array_append` because array append has a second argument which is defined
```suggestion
    /// let dims = diesel::select(array_ndims::<Array<Integer>, _>(vec![1, 2]))
```

guissalustiano (changes requested):
Thanks for open the PR!

The code looks good! I add a comment about the inference error. 
Could you a test with a null input to guarantee that it does not break the type signature?

diesel/src/pg/expression/helper_types.rs
weiznich: It's required for the generic arguments. `define_sql_function!` generates a function signature as follows:

```
fn array_ndims<A, Arr>(arr: Arr) -> Integer
where Arr: AsExpression<A>, A: ArrayOrNullableArray + SingleValue>;
```

The `AsExpression` part allows us to automatically convert rust values (i.e. `vec![1,2,3]`) to sql side binds and at the same time it allows us to accept sql side expressions with the same type at the same location (i.e `table::array_column`).  On the other hand the `#[auto_type]` macro expects exactly as many generic arguments as there are function parameters. The type def in then used to map that.

PR #4179, opened 2024-08-17
extend pg lower function to support Multirange

guissalustiano:
Extending the work of https://github.com/diesel-rs/diesel/pull/4159, this PR updates one Range function which also support multi-range.
This is a small PR to check the format, once approved I'm going to implement for the other methods.
I also add more example testing nullability and simplify the example removing the table and only use the select.

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -4,6 +4,7 @@ use super::expression_methods::InetOrCidr;
 use super::expression_methods::RangeHelper;
 use crate::expression::functions::define_sql_function;
 use crate::pg::expression::expression_methods::ArrayOrNullableArray;
+use crate::pg::expression::expression_methods::MultirangeOrRangeMaybeNullable;
 use crate::sql_types::*;
 
 define_sql_function! {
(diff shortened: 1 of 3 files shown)


guissalustiano:
The autotype test is broken
```
error[E0107]: type alias takes 2 generic arguments but 1 generic argument was supplied
   --> diesel_derives/tests/auto_type.rs:395:9
    |
395 |         lower(pg_extras::range),
    |         ^^^^^ ---------------- supplied 1 generic argument
    |         |
    |         expected 2 generic arguments
    |
note: type alias defined here, with 2 generic parameters: `R`, `T`
   --> /home/guiss/oss/diesel/diesel/src/pg/expression/helper_types.rs:314:10
    |
314 | pub type lower<R, T> = super::functions::lower<SqlTypeOf<R>, SqlTypeOf<T>, R>;
    |          ^^^^^ -  -
help: add missing generic argument
    |
395 |         lower(pg_extras::range, T),
```
I tried to follow the `array_append` example but it didn't work. Could someone help-me with that?

diesel/src/pg/expression/functions.rs
```diff
@@ -75,39 +76,32 @@ define_sql_function! {
...
+    /// assert_eq!(None, int);
+    ///
+    /// let int = diesel::select(lower::<Multirange<_>, Integer, _>(vec![(Bound::Included(5), Bound::Included(7))])).get_result::<Option<i32>>(connection)?;
+    /// assert_eq!(Some(5), int);
     /// #     Ok(())
     /// # }
     /// ```
     #[cfg(feature = "postgres_backend")]
-    fn lower<T: RangeHelper>(range: T) -> Nullable<<T as RangeHelper>::Inner>;
+    fn lower<Rang: MultirangeOrRangeMaybeNullable<Inner=T> + SingleValue, T: SingleValue>(range: Rang) -> Nullable<T>;
```
weiznich, line 104:
```suggestion
    fn lower<Rang: MultirangeOrRangeMaybeNullable + SingleValue>(range: Rang) -> Nullable<Rang::Inner>;
```

We should be able to remove the `T` type here in favor of using the associated type as return type directly. That should hopefully also allow the `type lower` to be unchanged and remain compatible with `#[auto_type]`

guissalustiano: Thanks 🙏🏽 !

PR #4180, opened 2024-08-17
Added support for missing operations on Citext

liamw1:
Just a small extension of #4021. This allows the use of like, not_like, and ordering operations on Citext columns.


weiznich (approved):
Thanks 👍

PR #4181, opened 2024-08-18
Ignore newlines while comparing schemas

### LINKED ISSUE #4172: diesel-cli with --locked-schema fails when schema.rs has CRLF newlines
### Versions

- **Diesel:** 2.2.2 (also works on master branch)

## Problem Description

If you have a schema.rs file that uses CRLF line endings, any diesel-cli command that could change `schema.rs` but was supplied with `--locked-schema` will fail, even if the line endings are the only thing that change from what diesel-cli would output vs what's on disk. This happens as diesel-cli directly compares the generated output with the on-disk file byte-for-byte.

Particularly on windows, git might change any LF line endings to CRLF for compatability reasons, so a user that clones a repository with a `schema.rs` file committed will see an error when they execute `diesel database reset --locked-schema`, even if the schema is correct.

No error, as the generated schema matches in everything but line endings.

An error.

No

### Steps to reproduce

 - clone diesel
 - change line endings in `examples/postgres/advanced-blog-cli/src/schema.rs` to CRLF
 - execute `cargo xtask run-tests --no-integration-tests --no-doc-tests`

### Workaround

Git can be told not to use CRLF line endings, changing this setting and all files in the repository can be done like this (WARNING: Will delete your local changes!):
```sh
git config --global core.autocrlf false
git rm -rf --cached .
git reset --hard HEAD
```

schitcrafter:
Fixes #4172

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -276,13 +276,12 @@ fn regenerate_schema_if_file_specified(matches: &ArgMatches) -> Result<(), crate
             }
 
             if matches.get_flag("LOCKED_SCHEMA") {
-                let mut buf = Vec::new();
-                print_schema::run_print_schema(&mut connection, config, &mut buf)?;
+                let schema = print_schema::output_schema(&mut connection, config)?;
 
-                let old_buf = std::fs::read(path)
+                let old_buf = std::fs::read_to_string(path)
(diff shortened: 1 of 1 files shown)


weiznich (approved):
Thanks :+1:

PR #4182, opened 2024-08-18
implements std::ops::Range to Multirange toSql

guissalustiano:
Implements #4112 for #4159

Unfortunately It doesn't allow diferente types in the same vec, e.g. `vec![..0, 4..7, 9..]`

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -178,8 +178,8 @@ pub mod sql_types {
     ///
     /// ### [`ToSql`] impls
     ///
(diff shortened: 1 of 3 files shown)


weiznich (commented):
Thanks, looks good for me :+1:

> Unfortunately It doesn't allow diferente types in the same vec, e.g. vec![..0, 4..7, 9..]

We might be able to somewhat workaround that by allowing `Box<dyn RangeBounds<T>>` and/or `&dyn RangeBounds<T>`  as types there as well. That should allow at least to write `vec![Box::new(..0), Box::new(4..7), Box::new(9..)]` to represent these different types.

PR #4184, opened 2024-08-19
Adds --no-default-migration option diesel_cli

lgrosz:
Adds --no-default-migration option diesel_cli to disable the generation of the default migration on setup.

This also tweaks the behavior of the db subcommands so they never generate the migration.


diesel_cli/src/main.rs
```diff
@@ -167,7 +167,7 @@ fn run_database_command(matches: &ArgMatches) -> Result<(), crate::errors::Error
     {
         ("setup", args) => {
             let migrations_dir = self::migrations::migrations_dir(matches)?;
-            database::setup_database(args, &migrations_dir)?;
+            database::setup_database(args, &migrations_dir, false)?;
```
weiznich, line 170: This changes the current default behavior. We should also use the `--no-default-migration` flag here instead.

diesel_cli/src/database.rs
```diff
@@ -157,17 +157,20 @@ pub fn reset_database(
     migrations_dir: &Path,
 ) -> Result<(), crate::errors::Error> {
     drop_database(&database_url(args)?)?;
-    setup_database(args, migrations_dir)
+    setup_database(args, migrations_dir, false)
```
weiznich, line 160: This changes the current default behavior. We should also use the --no-default-migration flag here instead.

weiznich (changes requested):
Thanks for working on this. I've added two comments about the reset and setup database command, which should also use the new flag instead of changing the default behaviour.

lgrosz:
> Thanks for working on this. I've added two comments about the reset and setup database command, which should also use the new flag instead of changing the default behaviour.

Sounds good. I can make those changes. Out of curiosity, would it make sense to be creating any migrations during any of the `diesel database` sub-commands? I was under the impression those commands would setup the database with the existing migrations while `diesel setup` would be used to "setup" the project, encompassing creating the default migration.

Either way, I'm aware such a behavior change wouldn't belong in a quick patch like this.

diesel_cli/src/database.rs
lgrosz: Done.

diesel_cli/src/main.rs
lgrosz: Done.

lgrosz:
I added the argument to the `database setup` and `database reset` commands, and passed the args appropriately.

I noticed that I added a parameter to `database::setup_database`, but maybe didn't need to? However.. when I apply this patch to resolve that, all the cli tests fail.. I'm not sure if it's just my environment or not:

```
From a9847a4fdb920979afe6364dbda9a71dbb342c5d Mon Sep 17 00:00:00 2001
From: Logan Grosz <logan.grosz@gmail.com>
Date: Wed, 21 Aug 2024 22:48:46 -0600
Subject: [PATCH] Use arg matches over extra parameter in database setup

---
 diesel_cli/src/database.rs | 7 ++++---
 diesel_cli/src/main.rs     | 4 ++--
 2 files changed, 6 insertions(+), 5 deletions(-)

diff --git a/diesel_cli/src/database.rs b/diesel_cli/src/database.rs
index 85c3fe9a2..2287b59cb 100644
--- a/diesel_cli/src/database.rs
+++ b/diesel_cli/src/database.rs
@@ -157,20 +157,21 @@ pub fn reset_database(
     migrations_dir: &Path,
 ) -> Result<(), crate::errors::Error> {
     drop_database(&database_url(args)?)?;
-    setup_database(args, migrations_dir, !args.get_flag("NO_DEFAULT_MIGRATION"))
+    setup_database(args, migrations_dir)
 }

 pub fn setup_database(
     args: &ArgMatches,
     migrations_dir: &Path,
-    default_migrations: bool,
 ) -> Result<(), crate::errors::Error> {
     let database_url = database_url(args)?;

-    create_database_if_needed(&database_url)?;
+    let default_migrations = !args.get_flag("NO_DEFAULT_MIGRATION");
+
     if default_migrations {
         create_default_migration_if_needed(&database_url, migrations_dir)?;
     }
+
     create_schema_table_and_run_migrations_if_needed(&database_url, migrations_dir)?;
     Ok(())
 }
diff --git a/diesel_cli/src/main.rs b/diesel_cli/src/main.rs
index a16a7e2db..77767ae06 100644
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -99,7 +99,7 @@ fn run_setup_command(matches: &ArgMatches) -> Result<(), crate::errors::Error> {
     let migrations_dir = create_migrations_dir(matches)?;
     create_config_file(matches, &migrations_dir)?;

-    database::setup_database(matches, &migrations_dir, !matches.get_flag("NO_DEFAULT_MIGRATION"))?;
+    database::setup_database(matches, &migrations_dir)?;
     Ok(())
 }

@@ -167,7 +167,7 @@ fn run_database_command(matches: &ArgMatches) -> Result<(), crate::errors::Error
     {
         ("setup", args) => {
             let migrations_dir = self::migrations::migrations_dir(matches)?;
-            database::setup_database(args, &migrations_dir, !args.get_flag("NO_DEFAULT_MIGRATION"))?;
+            database::setup_database(args, &migrations_dir)?;
             regenerate_schema_if_file_specified(matches)?;
         }
         ("reset", args) => {
--
2.45.1
```

lgrosz:
Sorry for all the updates.. I'm done now.

weiznich:
> I was under the impression those commands would setup the database with the existing migrations while diesel setup would be used to "setup" the project, encompassing creating the default migration.

`diesel setup` setups your project, which not only includes the database, but also the local folder structure. That includes the default migrations + a default `diesel.toml` file.

weiznich (approved):

PR #4185, opened 2024-08-19
ci: fix macos mariadb instalation

guissalustiano:
I notice that MySQL is broken on macOS, this PR solve it using updating the mariaDB version


weiznich (approved):
Thanks 👍

Issue #4152, opened 2024-08-09
Provide a `r2d2::CustomizeConnection` for testing


weiznich:
The [r2d2](https://docs.diesel.rs/2.2.x/r2d2/index.html) pooling crate provides hooks to customize connections on checkout from the pool. Diesel should provide a struct that implements [CustomizeConnection](https://docs.diesel.rs/2.2.x/r2d2/trait.CustomizeConnection.html#) by calling [`Connection::begin_test_transaction`](https://docs.diesel.rs/2.2.x/diesel/connection/trait.Connection.html#method.begin_test_transaction) to simplify the pooling setup for integration tests. In addition we might want to extend the documentation in the diesel r2d2 module to include information about how to setup the pool for integration tests.

This involves the following steps:

1. Add a new public `TestCustomizer<C>` type to the [`r2d2` module](https://github.com/diesel-rs/diesel/blob/861893c28673841595805586c059bced5f314163/diesel/src/r2d2.rs)
2. Implement [CustomizeConnection](https://docs.diesel.rs/2.2.x/r2d2/trait.CustomizeConnection.html#) for this type by calling [`Connection::begin_test_transaction`](https://docs.diesel.rs/2.2.x/diesel/connection/trait.Connection.html#method.begin_test_transaction)  in the `on_acquire` function 
3. Add some documentation to the new type
4. Extend the module level documentation [here](https://github.com/diesel-rs/diesel/blob/861893c28673841595805586c059bced5f314163/diesel/src/r2d2.rs#L93) to explain how to setup the pool for tests:
    * Set pool size to 1
    * Set the connection customizer
    * -> Run Tests in parallel as changes are not submitted to the database, but local to each test due to the "dangling" transaction.
5. Add an entry to the `Changelog.md` file in the repository root
6. Submit the PR

kpagacz:
I would love to take on this issue. It would be my first in `diesel`, so I would be grateful for some guidance. 

Here's where I am at:

* I followed the contributor's guide and set up `diesel` locally. It passed all the tests on my machine, so I looked at the places where the changes should be made.
* I understand that the new type is going to modify the instance of `C: Connection` while implementing the function `on_acquire` of [`CustomizeConnection`](https://docs.diesel.rs/2.2.x/r2d2/trait.CustomizeConnection.html#)
* Do you want anything to happen to the connection when calling `on_release` or do you want it left as the default (which is no-op)

weiznich:
Thanks for working on this. 

[...]

PR #4186, opened 2024-08-19
feat: added `diesel::r2d2::TestCustomizer`

kpagacz:
Closes https://github.com/diesel-rs/diesel/issues/4152

Let me know if you need me to make any improvements. I am unsure about the Error mapping. The problem I solved there was:
* the `Pool::builder` requires a connection customizer's error to be `ConnectionManager::Error`, which is `diesel::r2d2::Error`
* the return value from `begin_test_transaction()` is `QueryResult`, which error is `diesel::result::Error`

Hence the mapping, but I wonder if this is even correct for diesel. Let me know if you need me to make any adjustments here and with the docs!


weiznich (approved):
Thanks :+1:

PR #4187, opened 2024-08-20
Implement `array_prepend` and `array_remove`

edwinlzs:
Implementing `array_prepend` and `array_remove` functions for Postgres under #4153

using specs https://www.postgresql.org/docs/current/functions-array.html

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -831,3 +831,37 @@ define_sql_function! {
     /// ```
     fn array_prepend<T: SingleValue, Arr: ArrayOrNullableArray<Inner=T> + SingleValue>(e: T, a: Arr) -> Array<T>;
 }
+
+#[cfg(feature = "postgres_backend")]
+define_sql_function! {
+    /// Removes all elements equal to the given value from the array
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::array_remove;
+    /// #     use diesel::sql_types::{Nullable, Integer, Array};
+    /// #     let connection = &mut establish_connection();
+    /// let ints = diesel::select(array_remove::<Array<_>, Integer, _, _>(vec![1, 2, 3, 2], 2))
+    ///     .get_result::<Option<Vec<i32>>>(connection)?;
+    /// assert_eq!(Some(vec![1, 3]), ints);
+
+    /// let ints = diesel::select(array_remove::<Array<_>, Nullable<Integer>, _, _>(vec![None, Some(1), Some(2), None, Some(4)], None::<i32>))
+    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
+    /// assert_eq!(Some(vec![Some(1), Some(2), Some(4)]), ints);
+
+    /// let ints = diesel::select(array_remove::<Nullable<Array<_>>, Nullable<Integer>, _, _>(None::<Vec<i32>>, None::<i32>))
+    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
+    /// assert_eq!(None, ints);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn array_remove<Arr: ArrayOrNullableArray<Inner=T> + SingleValue, T: SingleValue>(a: Arr, e: T) -> Nullable<Array<T>>;
+}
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,75 @@ define_sql_function! {
...
+    ///     .get_result::<Vec<i32>>(connection)?;
+    /// assert_eq!(vec![3], ints);
+    ///
+    /// let ints = diesel::select(array_prepend::<Nullable<Integer>, Nullable<Array<_>>, _, _>(None::<i32>, None::<Vec<i32>>))
+    ///     .get_result::<Vec<Option<i32>>>(connection)?;
+    /// assert_eq!(vec![None], ints);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn array_prepend<T: SingleValue, Arr: ArrayOrNullableArray<Inner=T> + SingleValue>(e: T, a: Arr) -> Array<T>;
```
edwinlzs, line 787: Flipped the order of the generics and params (as compared to `array_append`) since the underlying Postgres function needs the element first before the array

diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,75 @@ define_sql_function! {
...
+    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
+    /// assert_eq!(Some(vec![Some(1), Some(2), Some(4)]), ints);
+
+    /// let ints = diesel::select(array_remove::<Nullable<Array<_>>, Nullable<Integer>, _, _>(None::<Vec<i32>>, None::<i32>))
+    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
+    /// assert_eq!(None, ints);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn array_remove<Arr: ArrayOrNullableArray<Inner=T> + SingleValue, T: SingleValue>(a: Arr, e: T) -> Nullable<Array<T>>;
```
edwinlzs, line 821: I think it might be odd that I specified the return type as `Nullable<Array<T>>` instead of simply `Array<T>`.

This is because I found that `array_remove` returns a wholly `null` value if both arguments are `null`, unlike `array_append` which returns an `Array<null>`. Therefore, I wrapped the return type in `Nullable` to avoid getting an unexpected null deserialization error if the arguments both happen to be `null`.

Added a screenshot illustrating this point! Let me know if there are any issues with this

array_remove vs array_append

edwinlzs:
I'm still new at contributing to the open source space, do let me know if there's anything that could be done better (e.g. if I should split implementations into 2 separate PRs instead of one)

diesel/src/pg/expression/functions.rs
guissalustiano, line 821: I think most of the time we would use this function with a not null array, so it would be nice if we can remove the `Nullable<_>` from the return time when isn't need.  
```suggestion
    fn array_remove<Arr: ArrayOrNullableArray<Inner=T> + SingleValue, T: SingleValue>(a: Arr, e: T) -> Arr;
```

diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,75 @@ define_sql_function! {
...
+    /// #     use diesel::sql_types::{Nullable, Integer, Array};
+    /// #     let connection = &mut establish_connection();
+    /// let ints = diesel::select(array_remove::<Array<_>, Integer, _, _>(vec![1, 2, 3, 2], 2))
+    ///     .get_result::<Option<Vec<i32>>>(connection)?;
+    /// assert_eq!(Some(vec![1, 3]), ints);
+
+    /// let ints = diesel::select(array_remove::<Array<_>, Nullable<Integer>, _, _>(vec![None, Some(1), Some(2), None, Some(4)], None::<i32>))
+    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
+    /// assert_eq!(Some(vec![Some(1), Some(2), Some(4)]), ints);
+
```
guissalustiano, line 814: You need comment this break lines to them appears in the doc 
```suggestion
    ///
    /// let ints = diesel::select(array_remove::<Array<_>, Nullable<Integer>, _, _>(vec![None, Some(1), Some(2), None, Some(4)], None::<i32>))
    ///     .get_result::<Option<Vec<Option<i32>>>>(connection)?;
    /// assert_eq!(Some(vec![Some(1), Some(2), Some(4)]), ints);
    ///
```

guissalustiano (approved):
Thanks for submitting this PR!

The change looks mostly good beside the minor improvement in the return type of `array_remove`.

diesel/src/pg/expression/functions.rs
guissalustiano: Double checking the behaviour on postgres:
- `SELECT array_remove(ARRAY[1, 2, 3], NULL);` returns `ARRAY[1, 2, 3]`

guissalustiano: I start the review before [your explanation](https://github.com/diesel-rs/diesel/pull/4187#discussion_r1723833738) :sweat_smile:, but I think it answers it

edwinlzs:
@guissalustiano Thanks for reviewing this PR! I've added a commit with fixes for the issues you highlighted :)

weiznich:
Thanks for contributing :+1: 

I've rebased this PR to resolve merge conflicts.

Issue #1186, opened 2017-09-21
More code reviewers wanted!


sgrif:
With the push to 1.0, the current pool of reviewers is starting to run low on bandwidth. We're looking for more people to help get involved with code review. You don't need to be familiar with Diesel's internals to help. Most people who have been on the Diesel core team got started by doing code review. If you're unsure what a piece of code does or why it's being done a certain way, just ask! (Make sure that the commit message doesn't answer that question already though). Don't feel like you can't ask a question just because other reviewers might already know the answer. We're happy to help you get up to speed.

juliusdelta:
Nice. Is the best way to get started to just review and add comments/questions to PRs?

sgrif:
Yup. If you'd like I can add you to the github group which gets pinged for code review

juliusdelta:
Sure that’d be great!

toddWannaCode:
I am really interested in this but I am just getting started with diesel.

killercup:
@toddWannaCode that's absolutely fine! Most PRs are quite isolated in scope so you don't need to know everything beforehand. :) Feel free to have a look at and review PRs to get a feel for it, and if you like it, ping me (or another core team member) and we'll add you to the reviewers team.

leshow:
I'd be interested in jumpiing into some code reviews, if you folks are still looking for people.

killercup:
Absolutely! Want me to add you to the Reviewers team?

Evan Cameron <notifications@github.com> schrieb am So. 24. Sep. 2017 um
23:22:

toddWannaCode:
Hey I would be like to be a member of the Reviewers team. Thanks. @killercup

killercup:
@toddWannaCode, welcome to the league of extraordinary reviewers! :)

leshow:
@killercup Sure, sounds good.

killercup:
@leshow, welcome to the ~~Four~~ Nine Horsemen of the Code Review!

ecasilla:
Count me in also. I would love to help anyway I can.

killercup:
@ecasilla, thank you for joining the Order of Magnitude!

ecasilla:
@killercup How will the mentoring work exactly is there a gitter for the core team?

sgrif:
@ecasilla You can ask questions either in the public gitter or as comments on the PRs you're trying to review (feel free to ask questions after things have been closed/merged as well)

gsquire:
[...]

PR #4190, opened 2024-08-22
implement array_dims

valkrypton:
Added support for `array_dims` postgres function under [#4153](https://github.com/diesel-rs/diesel/issues/4153)
This is my first PR ever so I would love some poitners!


valkrypton:
I can't fix the doc test i've written. Help would be appreciated

diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,34 @@ define_sql_function! {
...
+    /// #   use diesel::sql_types::{Nullable,Array,Integer};
+    /// #   let connection = &mut establish_connection();
+    ///
+    /// let dims = diesel::select(array_dims::<Array<_>,_,_>(vec![1,2]))
+    ///     .get_result::<String>(connection).unwrap();
+    /// assert!(String::from("[1:2]").eq(&dims));
+    ///
+    /// let dims = diesel::select(array_dims::<Array<_>,_,_>(vec![vec![1,2,3],vec![4,5,6]]))
+    /// .get_result::<String>(connection).unwrap();
+    /// assert!(String::from("[1:2][1:3]").eq(&dims));
```
guissalustiano, line 775: The crazy error here say that diesel didn't know how to transform `Vec<Vec<_>>` to 'Array<_>` becase Diesel does not support multidimensional arrays yet. So you can remove this test for now.

diesel/src/pg/expression/functions.rs
```diff
@@ -748,3 +748,34 @@ define_sql_function! {
...
+    /// #
+    /// # fn main(){
+    /// #    run_test().unwrap();
+    /// # }
+    /// # fn run_test()->QueryResult<()>{
+    /// #   use diesel::dsl::array_dims;
+    /// #   use diesel::sql_types::{Nullable,Array,Integer};
+    /// #   let connection = &mut establish_connection();
+    ///
+    /// let dims = diesel::select(array_dims::<Array<_>,_,_>(vec![1,2]))
```
guissalustiano, line 769:
```suggestion
    /// let dims = diesel::select(array_dims::<Array<Integer>,_,_>(vec![1,2]))
```
The error says that rust don't know with implementation use, this `i32` could be `Int4`, `Int8` or something else, to fix you need to specify the type here. The `array_append` doesn't need this because it Array inner type is the same of the second argument, and the second argument is specified

guissalustiano (changes requested):
Thanks for open the PR!

I let some comment about the errors. Could you also add a test with a None input to check it behavior? Thanks!

weiznich (approved):

weiznich:
Thanks for working on this. The PR looks now good :+1:

diesel/src/pg/expression/functions.rs
valkrypton: Thanks for the help!

valkrypton:
@weiznich I also resolved the conflicts and merged the branches, is that a problem?

weiznich:
@valkrypton I's fine, I just tend to resolve conflicts on accepted merge conflicts on my own to not burden contributors with the merge order.

valkrypton:
@weiznich I would love to contribute more to the repo, how do you recommend I start learning about Diesel internals?

weiznich:
@valkrypton Well the best way to learn about internals is to contribute and ask questions as soon as you hit a problem. It is also usful to join the reviewers teams (see #1186) and read code written by others. That is often an excellent place to learn more about how others approach certain problems and you often get explanations there on how certain things are implemented.

PR #4193, opened 2024-08-22
Extend range functions to support nullable and multirange

guissalustiano:
This PR extends the range functions to suportt multirange and nullable types, folowing the behavior of the[ postgres functions](https://www.postgresql.org/docs/current/functions-range.html#RANGE-FUNCTIONS-TABLE)

Please review after #4182 and #4179 be merged

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -186,39 +186,32 @@ define_sql_function! {
     /// ```rust
     /// # include!("../../doctest_setup.rs");
     /// #
-    /// # table! {
-    /// #     posts {
-    /// #         id -> Integer,
-    /// #         versions -> Range<Integer>,
-    /// #     }
-    /// # }
-    /// #
     /// # fn main() {
     /// #     run_test().unwrap();
     /// # }
     /// #
     /// # fn run_test() -> QueryResult<()> {
-    /// #     use self::posts::dsl::*;
+    /// # use diesel::pg::sql_types::{Range, Multirange};
+    /// # use diesel::dsl::lower_inc;
     /// #     use std::collections::Bound;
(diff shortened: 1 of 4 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -377,40 +328,63 @@ define_sql_function! {
...
-    /// assert_eq!(vec![(Bound::Included(5), Bound::Unbounded)], cool_posts);
+    /// let int = diesel::select(multirange_merge::<Nullable<Multirange<Integer>>, _>(None::<Vec<std::ops::Range<i32>>>)).get_result::<Option<(Bound<i32>, Bound<i32>)>>(connection)?;
+    /// assert_eq!(None, int);
     /// #     Ok(())
     /// # }
     /// ```
     #[cfg(feature = "postgres_backend")]
-    fn range_merge<T1: RangeHelper, T2: RangeHelper<Inner = T1::Inner>>(lhs: T1, rhs: T2) -> Range<T1::Inner>;
+    #[sql_name = "range_merge"]
+    fn multirange_merge<R: MultirangeOrNullableMultirange + SingleValue>(multirange: R) -> R::Range;
```
weiznich, line 387: We need to add a helper type definition for this new function + a test variant in the `#[auto_type]` test case module.

weiznich (changes requested):
Thanks for unifying these functions. The change looks mostly fine, but I think we miss a helper type for `multirange_merge` + a test case for `#[auto_type]` for this function.

diesel/src/pg/expression/functions.rs
guissalustiano: Thanks! I made the changes

weiznich (approved):
Thanks :+1:

PR #4197, opened 2024-08-23
Update cargo-dist to 0.21.1

weiznich:


kw217:
LGTM - definitely an improvement (e.g., Ubuntu version is now pinned).

PR #4198, opened 2024-08-23
Fix the tidy command

weiznich:
We did not return a non-zero exit code before if the format or typos check failed. This commit fixes that problem and fixes any broken formating that happened in the mean-time.


guissalustiano (approved):
That's really nice to new developer experience, thanks for work on that!

PR #4203, opened 2024-08-26
implement array_fill

valkrypton:
implemented array_fill under [#4153](https://github.com/diesel-rs/diesel/issues/4153)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1096,3 +1096,82 @@ define_sql_function! {
     /// ```
     fn array_length<Arr: ArrayOrNullableArray + SingleValue>(array: Arr, dimension: Integer) -> Nullable<Integer>;
 }
+
+#[cfg(feature = "postgres_backend")]
+define_sql_function! {
+    /// Returns an array initialized with supplied value and dimensions,
+    /// optionally with lower bounds other than 1
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main(){
+    /// #    run_test().unwrap();
+    /// # }
+    /// # fn run_test()->QueryResult<()>{
+    /// #   use diesel::dsl::array_fill;
+    /// #   use diesel::sql_types::{Nullable,Array,Integer,Text};
+    /// #   let connection = &mut establish_connection();
+    ///
+    /// let array = diesel::select(array_fill::<Integer,_,_>(2,vec![2]))
+    /// .get_result::<Vec<i32>>(connection)?;
+    /// assert_eq!(vec![2,2],array);
+    ///
+    /// let array = diesel::select(array_fill::<Text,_,_>(String::from("abc"),vec![3]))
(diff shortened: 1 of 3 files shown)


valkrypton:
not sure what arguments should i pass in auto_type.rs function call

diesel/src/pg/expression/helper_types.rs
valkrypton, line 1: I fixed it by looking at `array_to_string` definition and changing the definition in helper_type.rs according to, but I am not sure why I have to write a generic argument for an argument with a concrete type

diesel/src/pg/expression/helper_types.rs
```diff
@@ -407,10 +407,10 @@ pub type cardinality<A> = super::functions::cardinality<SqlTypeOf<A>, A>;
...
 #[cfg(feature = "postgres_backend")]
-pub type array_fill<E> = super::functions::array_fill<SqlTypeOf<E>, E, Array<Integer>>;
+pub type array_fill<E, A> = super::functions::array_fill<SqlTypeOf<E>, E, A>;
 
 /// Return type of [`array_fill_with_lower_bound(value,array,array)`](super::functions::array_fill_with_lower_bound())
 #[allow(non_camel_case_types)]
 #[cfg(feature = "postgres_backend")]
-pub type array_fill_with_lower_bound<E> =
-    super::functions::array_fill_with_lower_bound<SqlTypeOf<E>, E, Array<Integer>, Array<Integer>>;
+pub type array_fill_with_lower_bound<E, A1, A2> =
```
valkrypton, line 415: Not sure why i have to write generics here for arguments that have concrete types, but this works. An explanation would be appreciated!

diesel/src/pg/expression/functions.rs
```diff
@@ -996,3 +996,83 @@ define_sql_function! {
...
+    /// .get_result::<Vec<Option<i32>>>(connection)?;
+    /// assert_eq!(vec![Some(4),Some(4),Some(4)],array);
+    ///
+    /// let array = diesel::select(array_fill::<Nullable<Integer>,_,_>(None::<i32>,vec![3]))
+    /// .get_result::<Vec<Option<i32>>>(connection)?;
+    /// assert_eq!(vec![None::<i32>,None::<i32>,None::<i32>],array);
+    /// # Ok(())
+    /// # }
+    ///
+    fn array_fill<E:SingleValue>(value:E,dim:Array<Integer>) -> Array<E>;
```
weiznich, line 1037:
```suggestion
    fn array_fill<E:SingleValue>(value: E, dim: Array<Integer>) -> Array<E>;
```

diesel/src/pg/expression/helper_types.rs
weiznich: `define_sql_function!` generates a function signature as follows for the `array_fill` function:

```rust
fn array_fill<E: SingleValue, value: T1, dim: T2>(value: T1, dim: T2) -> _
where T1: AsExpression<E>, 
            T2: AsExpression<Array<Integer>>
```

The additional `AsExpression` bounds enable diesel to accept a typed SQL expression (like for example a column) or a rust side value (for example `5`) as argument in this position.

PR #4209, opened 2024-08-27
Add embed support for AsChangeSet

### LINKED ISSUE #1917: Insert Embedded Struct for AsChangeset
Will diesel support #[diesel(embed)] on AsChangeset structs? Would be nice for reuse between Insertable and "update-able" structs.

vnghia:
This is a copy verbatimly from `Insertable`. Since the interaction between `treat_none_as_default_value` and `embed` isn't allowed in `Insertable` so it is not allowed here as well.

Closes #1917

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 ### Added
 
+* Added embedded struct support for `AsChangeset` via `#[diesel(embed)]`
 * Support for libsqlite3-sys 0.30.0
 * Add support for built-in PostgreSQL range operators and functions
 * Support for postgres multirange type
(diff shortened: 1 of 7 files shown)


weiznich:
Thanks for opening this PR. 
Could you also add a [test like that one](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/tests/insertable.rs#L469) to demonstrate that it works as expected?

Additionally we should document this attribute [here](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/lib.rs#L96-L106). See the relevant documentation on the [`Insertable` derive for an example](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/lib.rs#L296-L332)

Finally that's a great candidate for adding an entry to our `Changelog.md` file.

vnghia:
I think it's ready. Thank you for your pointers. It makes everything become much easier

weiznich (approved):
Thanks, looks good now :+1:

(I just pushed a minor addition that checks as well that you cannot mix embed + serialize_as)

PR #4210, opened 2024-08-27
print diff if print-schema fails due to mismatch with --locked-schema flag

### LINKED ISSUE #4208: Print a diff if `diesel print-schema --locked` fails due to a mismatch
We want to improve the user experience of diesel-cli. The `diesel print-schema` command provides a `--locked` schema flag that allows users to check if there `schema.rs` file is up to data. It's designed to be used in CI setups. It currently only provides the feedback that the schema file would change, but there is no output what exactly would change. To improve the user experience there we would like to print a diff between the current version of the schema stored in the `schema.rs` file and what `schema.rs` file would be constructed from the database.

The relevant comparison code is here: https://github.com/diesel-rs/diesel/blob/09d5e1050a40d698b10af98bd10495c98ccfaa7f/diesel_cli/src/main.rs#L284-L288

Implementing this feature should be as simple as:

1. Add a dependency on [similar-assert](https://docs.rs/similar-asserts/latest/) to the `Cargo.toml` file of `diesel-cli`
2. Construct a instance of [`similar_assert::SimpleDiff`](https://docs.rs/similar-asserts/latest/similar_asserts/struct.SimpleDiff.html) from the old and new version of the schema
3. Print the diff via `println!`

aznszn:
using similar-asserts to print a diff if print-schema fails with the --locked-schema flag provided due to a mismatch

fixes #4208

### DIFF
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -283,9 +283,13 @@ fn regenerate_schema_if_file_specified(matches: &ArgMatches) -> Result<(), crate
                     .map_err(|e| crate::errors::Error::IoError(e, Some(path.to_owned())))?;
 
                 if schema.lines().ne(old_buf.lines()) {
+                    // it's fine to leak here, we will
+                    // exit the application anyway soon
+                    let label = path.file_name().expect("We have a file name here");
+                    let label = label.to_string_lossy().into_owned().leak();
                     println!(
                         "{}",
-                        SimpleDiff::from_str(&schema, &old_buf, "new schema", "schema in file")
+                        SimpleDiff::from_str(&old_buf, &schema, label, "new schema")
                     );
                     return Err(crate::errors::Error::SchemaWouldChange(
                         path.display().to_string(),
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for working on this. The change is mostly what I imagined. I just tweaked the output a bit (changed new/old + the label).

For reference the output now looks like this:

image

weiznich (approved):

diesel_cli/src/main.rs
```diff
@@ -282,6 +283,14 @@ fn regenerate_schema_if_file_specified(matches: &ArgMatches) -> Result<(), crate
                     .map_err(|e| crate::errors::Error::IoError(e, Some(path.to_owned())))?;
 
                 if schema.lines().ne(old_buf.lines()) {
+                    // it's fine to leak here, we will
+                    // exit the application anyway soon
+                    let label = path.file_name().expect("We have a file name here");
+                    let label = label.to_string_lossy().into_owned().leak();
```
kw217, line 289: The `leak()` here is probably not necessary, but as the comment says it's fine.

weiznich: https://docs.rs/similar-asserts/latest/similar_asserts/struct.SimpleDiff.html#method.from_str unfortunately requires that the labels are `&'static str`, so yes that `leak()` is strictly speaking not required if we would use a static label instead.

kw217: Ah, d'oh, I missed that the labels had a different lifetime from the values. This has actually been relaxed in main of similar-asserts but not yet released. I've added a comment to the relevant issue: https://github.com/mitsuhiko/similar-asserts/pull/9#issuecomment-2327210092

PR #4224, opened 2024-09-02
Fix a potential null pointer deref in the Sqlite backend

### LINKED ISSUE #4223: Undefined behavior when using `diesel::sql_query` with an empty string for Sqlite
### Versions

- **Rust:** rustc 1.80.1
- **Diesel:** v2.2.3
- **Database:** sqlite 3.44.1
- **Operating System** Linux

### Feature Flags

- **diesel:** sqlite

## Problem Description

This code has UB:
```rust
use diesel::{Connection, RunQueryDsl, SqliteConnection};

fn main() {
    let mut conn = SqliteConnection::establish("tmp.sqlite").unwrap();
    diesel::sql_query("").execute(&mut conn).unwrap();
}
```

Output with `cargo run`:
```
thread 'main' panicked at library/core/src/panicking.rs:219:5:
unsafe precondition(s) violated: NonNull::new_unchecked requires that the pointer is non-null
stack backtrace:
   0: rust_begin_unwind
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/std/src/panicking.rs:652:5
   1: core::panicking::panic_nounwind_fmt::runtime
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/panicking.rs:110:18
   2: core::panicking::panic_nounwind_fmt
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/panicking.rs:120:5
   3: core::panicking::panic_nounwind
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/panicking.rs:219:5
   4: core::ptr::non_null::NonNull<T>::new_unchecked::precondition_check
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/ub_checks.rs:68:21
   5: core::ptr::non_null::NonNull<T>::new_unchecked
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/ub_checks.rs:75:17
   6: diesel::sqlite::connection::stmt::Statement::prepare::{{closure}}
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/sqlite/connection/stmt.rs:53:43
   7: core::result::Result<T,E>::map
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/result.rs:771:25
   8: diesel::sqlite::connection::stmt::Statement::prepare
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/sqlite/connection/stmt.rs:51:9
   9: diesel::sqlite::connection::SqliteConnection::prepared_query::{{closure}}
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/sqlite/connection/mod.rs:356:30
  10: diesel::connection::statement_cache::StatementCache<DB,Statement>::cached_statement_non_generic
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/connection/statement_cache.rs:222:20
  11: diesel::connection::statement_cache::StatementCache<DB,Statement>::cached_statement
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/connection/statement_cache.rs:196:9
  12: diesel::sqlite::connection::SqliteConnection::prepared_query
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/sqlite/connection/mod.rs:352:31
  13: <diesel::sqlite::connection::SqliteConnection as diesel::connection::Connection>::execute_returning_count
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/sqlite/connection/mod.rs:187:29
  14: <T as diesel::query_dsl::load_dsl::ExecuteDsl<Conn,DB>>::execute
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/query_dsl/load_dsl.rs:90:9
  15: diesel::query_dsl::RunQueryDsl::execute
             at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/diesel-2.2.3/src/query_dsl/mod.rs:1433:9
  16: poc_diesel::main
             at ./src/main.rs:5:5
  17: core::ops::function::FnOnce::call_once
             at /rustc/3f5fd8dd41153bc5fdca9427e9e05be2c767ba23/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread caused non-unwinding panic. aborting.
[1]    6997 abort (core dumped)
```

It seems the problem is linked to the [`sqlite3_prepare_v3`](https://github.com/diesel-rs/diesel/blob/08911c8420f4e2ff732a3b263a5a4f3df0f745d2/diesel/src/sqlite/connection/stmt.rs#L37) call (see [sqlite doc](https://www.sqlite.org/c3ref/prepare.html)):
> If the input text contains no SQL (if the input is an empty string or a comment) then *ppStmt is set to NULL.

## Checklist

weiznich:
The documentation of `sqlite3_prepare_v3` states that:

> *ppStmt is left pointing to a compiled prepared statement that can be
executed using sqlite3_step(). If there is an error, *ppStmt is set to NULL. If the input text contains no SQL (if the input is an empty string or a comment) then *ppStmt is set to NULL. The calling procedure is responsible for deleting the compiled SQL statement using sqlite3_finalize() after it has finished with it. ppStmt may not be NULL.

We already guard against the error case there before constructing the `NonNull` pointer, we do not guard against empty statements yet. This commit fixes that particular issue + adds a test to check if the fix works as expected.

This fixes #4223

### DIFF
--- a/diesel/src/sqlite/connection/stmt.rs
+++ b/diesel/src/sqlite/connection/stmt.rs
@@ -48,11 +48,14 @@ impl Statement {
             )
         };
 
-        ensure_sqlite_ok(prepare_result, raw_connection.internal_connection.as_ptr()).map(|_| {
-            Statement {
-                inner_statement: unsafe { NonNull::new_unchecked(stmt) },
-            }
-        })
+        ensure_sqlite_ok(prepare_result, raw_connection.internal_connection.as_ptr())?;
+
+        // sqlite3_prepare_v3 returns a null pointer for empty statements. This includes
+        // empty or only whitespace strings or any other non-op query string like a comment
+        let inner_statement = NonNull::new(stmt).ok_or_else(|| {
+            crate::result::Error::QueryBuilderError(Box::new(crate::result::EmptyQuery))
+        })?;
+        Ok(Statement { inner_statement })
     }
 
     // The caller of this function has to ensure that:
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -456,6 +456,21 @@ impl fmt::Display for EmptyChangeset {
 
 impl StdError for EmptyChangeset {}
 
+/// Expected when you try to execute an empty query
+#[derive(Debug, Clone, Copy)]
+pub struct EmptyQuery;
+
+impl fmt::Display for EmptyQuery {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Detected an empty query. These are not supported by your database system"
+        )
+    }
+}
+
+impl StdError for EmptyQuery {}
+
 /// An error occurred while deserializing a field
 #[derive(Debug)]
 #[non_exhaustive]
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -952,4 +952,21 @@ mod tests {
         .unwrap();
         assert_eq!(res, Some(Vec::new()));
     }
+
+    #[test]
+    fn correctly_handle_empty_query() {
+        let check_empty_query_error = |r: crate::QueryResult<usize>| {
+            assert!(r.is_err());
+            let err = r.unwrap_err();
+            assert!(
+                matches!(err, crate::result::Error::QueryBuilderError(ref b) if b.is::<crate::result::EmptyQuery>()),
+                "Expected a query builder error, but got {err}"
+            );
+        };
+        let connection = &mut SqliteConnection::establish(":memory:").unwrap();
+        check_empty_query_error(crate::sql_query("").execute(connection));
+        check_empty_query_error(crate::sql_query("   ").execute(connection));
+        check_empty_query_error(crate::sql_query("\n\t").execute(connection));
+        check_empty_query_error(crate::sql_query("-- SELECT 1;").execute(connection));
+    }
 }


wowinter13 (approved):

weiznich:
cc @diesel-rs/core I would like to release that as patch release somewhen this week.

diesel/src/sqlite/connection/stmt.rs
```diff
@@ -25,6 +25,12 @@ impl Statement {
         sql: &str,
         is_cached: PrepareForCache,
     ) -> QueryResult<Self> {
+        if sql.is_empty() {
+            return Err(crate::result::Error::QueryBuilderError(Box::new(
+                crate::result::EmptyQuery,
+            )));
+        }
+
```
stormshield-kg, line 33: Should we add a check after the call to `sqlite3_prepare_v3()` that `stmt` is not null before calling `NonNull::new_unchecked()` ?

weiznich: Given the documentation of `sqlite3_prepare_v3` that shouldn't be necessary as we already cover both variants that are documented to return null. That written: It might still be worth to add that additional check to be sure.

Ten0:
It looks like that doesn't guard against strings that are considered empty by sqlite, e.g. whitespace and comments-only.
It looks like it would be safer to replace the initial check on the Diesel side by a match on `NonNull::new(stmt)`, returning `EmptyQuery` if at that point it's still a null pointer, instead of using `new_unchecked`.

diesel/src/sqlite/connection/stmt.rs
stormshield-kg: The above check is not enough, since the pointer will be null if the query string contains only whitespaces or comments. We should also add a test for this.

stormshield-kg:
> It looks like that doesn't guard against strings that are considered empty by sqlite, e.g. whitespace and comments-only. It looks like it would be safer to replace the initial check on the Diesel side by a match on `NonNull::new(stmt)` after making the sqlite C call, returning `EmptyQuery` if at that point it's still a null pointer, instead of using `new_unchecked`.

Yes, but it should be done only if the call returned `ffi::SQLITE_OK`.

weiznich:
You are both correct, I misread that statment as literal empty string. Note to myself: Don't fix this kind of issues just before needing to leave.

I've now pushed a fixed version by using `NonNull::new` instead of the `new_unchecked` variant. I also also added tests for the other variants.

Ten0 (approved):

Issue #4206, opened 2024-08-26
array methods breaks index method


guissalustiano:
In latest diesel (not publish yet), we are [adding array methods](#4153) and it breaks the `index` inline method.

The follow code generate a wrong SQL:
```rust
diesel::select(array_append::<Array<_>, Integer, _, _>(vec![1, 2], 3).index().get_result::<i32>(connection).unwrap();
``` 
Generates the query:
```sql
SELECT array_append($1, $2)[$3] -- binds: [[1, 2], 3, 1]
```

And throw the error:
```rust
called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "syntax error at or near \"[\"")
```

weiznich:
Thanks for filling this. From playing a bit around on the psql command line, it seems like `array_append('{1, 2}', 3)[2]` is just not considered valid by postgresql. We can fix that by introducing an additional parenthesis there as `(array_append('{1, 2}', 3))[2]` is considered valid and gives the expected result.

Fixing this should be as simple as:

1. Change the generated SQL here (https://github.com/diesel-rs/diesel/blob/09d5e1050a40d698b10af98bd10495c98ccfaa7f/diesel/src/pg/expression/operators.rs#L110) to add the additional parenthesis around the left side (the linked line).
2. Add the provided example as test case in diesel-tests to make sure we have a regression test for this

valkrypton:
i'd like to give this a try

weiznich:
Closed by #4227

PR #4227, opened 2024-09-03
fix index method

valkrypton:
fixed array method breaking index method under [#4206](https://github.com/diesel-rs/diesel/issues/4206)


guissalustiano (approved):
Thank for working on that!

The changes looks good to me.

Issue #4216, opened 2024-08-30
Add support for currently unsupported postgres json/jsonb functions


weiznich:
Diesel currently supports the postgres array types. We do not provide built-in support for [various methods](https://www.postgresql.org/docs/current/functions-json.htmll) available for these types. This is a tracking issue for adding support for these methods. 

The general strategy for adding support for new methods is as following:

1. Define the operator via [`define_sql_function!()`](https://docs.diesel.rs/master/diesel/expression/functions/macro.define_sql_function.html). These operators can be defined [here](https://github.com/diesel-rs/diesel/blob/861893c28673841595805586c059bced5f314163/diesel/src/pg/expression/functions.rs#L698-L733). See the linked definition of `array_append` for an example. If there is already an existing definition, this step could be skipped. This function should have a short documentation snippet with an example (See the linked postgres documentation for examples for all of the function, please also add variants with null values, etc). If there is a `json` and a `jsonb` variant, please implement both in a single PR.
2. If the function is generic, add a helper type definition [here](https://github.com/diesel-rs/diesel/blob/6c1cb1d42eb2b3000502d67249dc3cdd76066dd7/diesel/src/pg/expression/helper_types.rs#L351-L354). Again see the linked definition for an example for the `array_append` function.
3. Add a test for  `#[auto_type]` support for the newly added function [here](https://github.com/diesel-rs/diesel/blob/6c1cb1d42eb2b3000502d67249dc3cdd76066dd7/diesel_derives/tests/auto_type.rs#L413)
5. Submit a PR with the change

Method list:

- [ ] `to_json ( anyelement ) → json` Converts any SQL value to json. *
- [ ] `to_jsonb ( anyelement ) → jsonb` Converts any SQL value to jsonb. *
- [ ] `array_to_json ( anyarray [, boolean ] ) → json` Converts an SQL array to a JSON array.*
- [ ] `json_array ( [ { value_expression [ FORMAT JSON ] } [, ...] ] [ { NULL | ABSENT } ON NULL ] [ RETURNING data_type [ FORMAT JSON [ ENCODING UTF8 ] ] ])`  Constructs a JSON array from either a series of value_expression parameters or from the results of query_expression ***
- [ ] `json_array ( [ query_expression ] [ RETURNING data_type [ FORMAT JSON [ ENCODING UTF8 ] ] ])` Constructs a JSON array from either a series of value_expression parameters or from the results of query_expression ***
[...]

PR #4229, opened 2024-09-03
added `to_json` function

aznszn:
added the `to_json` function under #4216


aznszn:
The last test I am a bit unsure about since trying to parse the returned into JSON into serde_json::Value results in an error

diesel/src/pg/expression/functions.rs
```diff
@@ -1423,3 +1423,56 @@ define_sql_function! {
...
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::to_json;
+    /// #     use serde_json::Value;
+    /// #     use diesel::sql_types::{Integer, Array, Json, Text, Nullable};
+    /// #     let connection = &mut establish_connection();
+    /// let result = diesel::select(to_json::<Integer, _>(1))
+    ///     .get_result::<Value>(connection)?;
+    ///
+    /// let expected: Value = serde_json::from_str("1").unwrap();
```
weiznich, line 1448: can we use the `serde_json::json!` macro instead to construct the expected values?

diesel/src/pg/expression/functions.rs
```diff
@@ -1423,3 +1423,56 @@ define_sql_function! {
...
+    ///
+    /// let result = diesel::select(to_json::<Nullable<Array<Text>>, _>(None::<Vec<String>>))
+    ///     .get_result::<Value>(connection);
+    ///
+    /// assert!(result.is_err());
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn to_json<E: SingleValue>(e: E) -> Json;
```
weiznich, line 1477: This function seems to return `null` if you pass in a null values. We want to have a doc test demonstrating that this is correctly propagated.

weiznich (changes requested):
Thanks for working on this :pray: 

I added two comments about potential improvements, if you need help with the nullabilty handling just let me know.

diesel/src/pg/expression/functions.rs
aznszn: done

diesel/src/pg/expression/functions.rs
aznszn: ive changed the last test a bit, let me know if its okay

weiznich (approved):

PR #4230, opened 2024-09-03
implement json_object

valkrypton:
added support for `json_object` under [#4216](https://github.com/diesel-rs/diesel/issues/4216)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -5,9 +5,9 @@ use crate::expression::functions::define_sql_function;
 use crate::pg::expression::expression_methods::ArrayOrNullableArray;
 use crate::pg::expression::expression_methods::MaybeNullableValue;
 use crate::pg::expression::expression_methods::MultirangeOrNullableMultirange;
-use crate::pg::expression::expression_methods::TextArrayOrNullableTextArray;
 use crate::pg::expression::expression_methods::MultirangeOrRangeMaybeNullable;
 use crate::pg::expression::expression_methods::RangeOrNullableRange;
+use crate::pg::expression::expression_methods::TextArrayOrNullableTextArray;
 use crate::sql_types::*;
 
 define_sql_function! {
@@ -1588,5 +1588,7 @@ define_sql_function! {
     /// #     Ok(())
     /// # }
     /// ```
-    fn json_object<Arr:TextArrayOrNullableTextArray + MaybeNullableValue<Json>>(text_array: Arr) -> Arr::Out;
+    fn json_object<Arr: TextArrayOrNullableTextArray + MaybeNullableValue<Json>>(
+        text_array: Arr,
+    ) -> Arr::Out;
 }
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1546,3 +1546,38 @@ define_sql_function! {
     /// ```
     fn to_jsonb<E: MaybeNullableValue<Jsonb>>(e: E) -> E::Out;
 }
+
+define_sql_function! {
+    /// Builds a JSON object out of a text array. The array must have an even number of members,
+    /// in which case they are taken as alternating key/value pairs
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::json_object;
+    /// #     use diesel::sql_types::{Json};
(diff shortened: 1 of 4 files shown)


valkrypton:
@weiznich  what value should i pass it in `auto_type.rs` as `pg_extras::array` is of type `Array<Integer>`?

weiznich:
@valkrypton I would suggest to just define a new column with the type `Array<Text>` in the pg_extras table.

diesel/src/pg/expression/functions.rs
```diff
@@ -1423,3 +1423,38 @@ define_sql_function! {
...
+    /// #     use diesel::dsl::json_object;
+    /// #     use diesel::sql_types::{Json};
+    /// #     use serde_json::Value;
+    /// #     let connection = &mut establish_connection();
+    /// let json = diesel::select(json_object(vec!["hello","world"])).get_result::<Value>(connection)?;
+    /// let expected:Value = serde_json::from_str(r#"{"hello":"world"}"#).unwrap();
+    /// assert_eq!(expected,json);
+    ///
+    /// let json = diesel::select(json_object(vec!["hello","world","John","Doe"])).get_result::<Value>(connection)?;
+    /// let expected:Value = serde_json::from_str(r#"{"hello":"world","John":"Doe"}"#).unwrap();
```
guissalustiano, line 1450: Could we use the serde_json::json! macro instead to construct the expected values?

diesel/src/pg/expression/functions.rs
```diff
@@ -1423,3 +1423,38 @@ define_sql_function! {
...
+    /// assert_eq!(expected,json);
+    ///
+    /// let json = diesel::select(json_object(vec!["hello","world","John"])).get_result::<Value>(connection);
+    /// assert!(json.is_err());
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn json_object(text_array:Array<Text>)->Json;
```
guissalustiano, line 1459: Array are always generated with nullable elements in diesel which can make hard to use this function. I think this functions should be able to accept `Nullable<Array<Nullable<Text>>`, maybe using a helper trait to make simple to call it directly.

diesel/src/pg/expression/functions.rs
```diff
@@ -1423,3 +1423,38 @@ define_sql_function! {
     /// ```
     fn array_upper<Arr: ArrayOrNullableArray + SingleValue>(array: Arr, dimension: Integer) -> Nullable<Integer>;
 }
+
+define_sql_function! {
```
guissalustiano, line 1427: This function also support a form `jsonb_object(keys text[], values text[])`, whitch takes keys and values pairwise from two separate arrays. Could you implement that too?

guissalustiano (changes requested):
Thanks for open the PR!

I have some comments about the type signature and also request the other variant of this function

diesel/src/pg/expression/functions.rs
valkrypton: yes sure ill implement that too

valkrypton:
@guissalustiano can you take a look at the type signature now?

diesel/src/pg/expression/functions.rs
weiznich: I would just accept both variants here. That might require an additional helper trait that is implemented for both `Array<Text>` and `Array<Nullable<Text>>`.  Additionally what's about `Nullable<Array<_>>` here?

guissalustiano:
Sure!
The signature `fn json_object(text_array:Nullable<Array<Text>>)->Json;`  is broken because if you call `SELECT json_object(NULL)` it returns `NULL`, but the function signatures don't tell you that.

A signature like `fn json_object(text_array:Nullable<Array<Text>>)->Nullable<Json>;` would fix that, but is annoying to have this extra `Nullable<_>` when you use that with non-nullable inputs. On [`to_json`](https://github.com/diesel-rs/diesel/pull/4229/files#diff-bf88b4900340a435caafd7c947e7b9a95e61ff158ef37596ab3262b3c754c45fR3702-R3714) we use a helper trait to handle this. 

We also would like to support an `Nullable<Array<Nullable<Text>>>` because diesel always generates columns as `Array<Nullable<_>>`, and so not supporting that would mean that `json_object` couldn't be used with a table, only literal values.

valkrypton:
Thanks for the detailed explanation, i appreciate it.
The pr got messed up when i rebased this branch onto the updated master, i had no idea the previous commits would show up here as well. I tried to fix but couldn't figure it out, so if you could help me I'd appreciate it.

weiznich (approved):
Thanks for the update. The signature looks good now.

I just rebased this PR again to resolve the merge conflicts, but the history already looked right before that. So either it was a github issue before or you already fixed it :pray: 

I also added a `#[diagnostic::on_unimplemented]` attribute to the new helper trait to improve the compiler error message if some user passes in a wrong typed value.

PR #4231, opened 2024-09-03
make `skip_from` public for 3rd party backends

insipx:


kw217 (approved):
This seems entirely reasonable to me. The code is correct, nothing else in the `skip_from` logic is SQLite-specific as far as I can see, and it's guarded by the here-be-dragons declaration.

PR #4242, opened 2024-09-07
fix: array_ndims must handle null inputs

guissalustiano:


weiznich:
Thanks for opening this PR. I think I rather would want to go with the solution discussed in https://github.com/diesel-rs/diesel/discussions/4241.

guissalustiano:
Thanks for answering! I should open a PR implementing that soon. I opened this PR before your answer because this method was already broken.

PR #4245, opened 2024-09-08
WIP Implement Offset and Limit Dsl on CombinationClause

Aethelflaed:
Implement `limit` and `offset` on combination clauses (e.g. unions)


diesel/src/query_builder/combination_clause.rs
```diff
@@ -32,41 +56,88 @@ impl<Combinator, Rule, Source, Rhs> CombinationClause<Combinator, Rule, Source,
...
+            },
         }
     }
 }
 
-impl<Combinator, Rule, Source, Rhs> Query for CombinationClause<Combinator, Rule, Source, Rhs>
+impl<Combinator, Rule, Source, Rhs, LOf>
+    CombinationClause<Combinator, Rule, Source, Rhs, LOf>
+{
+    pub(crate) fn new_full(
```
Aethelflaed, line 70: I could remove this method and change `new` instead, but then I would need to propagate the LimitOffsetClause declaration to each call site, which happens in several files, so I'm not entirely sure about that and I'm looking for feedback on it.

The method name is subpar :/

weiznich: Given that we only use this function inside of this file I just wouldn't add it at all. Instead use the direct struct construction syntax at the calling sides:

```rust
 CombinationClause{
            combinator: self.combinator,
            duplicate_rule: self.duplicate_rule,
            source: self.source.0,
            rhs: self.rhs.0,
            limit_offest: LimitOffsetClause {
                limit_clause: self.limit_offset.limit_clause,
                offset_clause,
            },
        }
```

(I think that even looks clearer)

diesel_tests/tests/combination.rs
```diff
@@ -117,6 +117,65 @@ fn except() {
...
+    let data: Vec<User> = users
+        .filter(id.le(tess.id))
+        .union(users.filter(id.ge(tess.id)))
+        .limit(2)
+        .load(conn)
+        .unwrap();
+
+    // Cannot determine any `expected_data` to match against because the database might return
+    // different rows
+    // The positional_order_by fix is not applicable, because ORDER should appear before LIMIT
```
weiznich, line 145: Could you elaborate why this is not possible?

weiznich (commented):
Thanks for working on this. I think this is the correct way to implement this feature. I've left a few comments about details, otherwise this looks mostly fine.

We likely want to have a changelog entry for this as well.

diesel_tests/tests/combination.rs
Aethelflaed: I've added a more extensive comment, tell me if that's enough or if you need more

weiznich: The explanation sounds fine to me, thanks for adding it. :+1:

That written: This indicates that we likely should change how `positional_order_by` is applied to the type. It likely should behave similar to the newly implemented functions instead of being stacked on top of the existing query. 

(That's something for a followup PR)

PR #4247, opened 2024-09-09
Fixed mysql not supporting offset() without a limit()

Aethelflaed:
While working on https://github.com/diesel-rs/diesel/pull/4245, I realised I needed a `limit()` call before calling `offset()` for mysql, and thought that was a normal behaviour.

Then I saw in the 2.0.0 changelog that there should have been a workaround for that already.

It seems it wasn't entirely tested and working, so here are the missing pieces


weiznich:
Thanks for opening this PR

The PR that introduced this behavior was https://github.com/diesel-rs/diesel/pull/2029. Back then we explicitly decided to not support `offset` without `limit` for mysql wherever possible as that "workaround" is really an ugly hack that should be made explicit by the calling side. Unfortunately we can only guarantee that for unboxed statements, not for boxed ones, therefore the "workaround" there. So no, this is no missing piece but an explicit design decision.

Aethelflaed:
> The PR that introduced this behavior was #2029. Back then we explicitly decided to not support `offset` without `limit` for mysql wherever possible as that "workaround" is really an ugly hack that should be made explicit by the calling side. Unfortunately we can only guarantee that for unboxed statements, not for boxed ones, therefore the "workaround" there. So no, this is no missing piece but an explicit design decision.

I'll look if it's possible to add a warning

Aethelflaed:
It's still not possible to add a `deprecated` message on a trait implementation item so nothing changed since #2029.

cc @weiznich

PR #4250, opened 2024-09-11
`json_typeof` and `jsonb_typeof` function

anna-ahmed19:
Added the `json_typeof` and `jsonb_typeof` functions under [Issue#4216](https://github.com/diesel-rs/diesel/issues/4216)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1654,8 +1654,8 @@ define_sql_function! {
     /// # #[cfg(feature = "serde_json")]
     /// # fn run_test() -> QueryResult<()> {
     /// #     use diesel::dsl::json_typeof;
-    /// #     use serde_json::json;
-    /// #     use diesel::sql_types::Json;
+    /// #     use serde_json::{json, Value};
+    /// #     use diesel::sql_types::{Json, Nullable};
     /// #     let connection = &mut establish_connection();
     /// let result = diesel::select(json_typeof::<Json, _>(json!({"a": "b", "c": 1})))
     ///     .get_result::<String>(connection)?;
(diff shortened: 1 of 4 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -1634,3 +1636,113 @@ define_sql_function! {
...
+    /// let result = diesel::select(json_typeof::<Json, _>(json!(true)))
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!("boolean".to_string(), result);
+    ///
+    /// let result = diesel::select(json_typeof::<Json, _>(json!(null)))
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!("null".to_string(), result);
+    /// #     Ok(())
```
guissalustiano, line 1689: Could you add a test with None input?
It is something like this:
```
    /// let result = diesel::select(json_typeof::<Nullable<Json>, _>(None::<Value>))
    ///     .get_result::<String>(connection)?;
    ///
    /// assert_eq!(<something>, result);
```

diesel/src/pg/expression/functions.rs
```diff
@@ -1634,3 +1636,113 @@ define_sql_function! {
...
+    ///
+    /// let result = diesel::select(jsonb_typeof::<Jsonb, _>(json!(true)))
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!("boolean".to_string(), result);
+    ///
+    /// let result = diesel::select(jsonb_typeof::<Jsonb, _>(json!(null)))
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!("null".to_string(), result);
```
guissalustiano, line 1743: Could you also add a None test here?

guissalustiano (changes requested):
Thanks for opening the PR!

The changes look good, I only would like to see some tests about the null behavior, in past cases, Postgres returns `NULL` when you pass `NULL` to this kind of functions.
If this is case, the function signature is wrong, It should also cover the NULL input return.

To fix that we can use 
```rust
fn json_typeof<E: JsonOrNullableJson + SingleValue + MaybeNullableValue<Text>>(e: E) -> E::Out;
```
The `E: MaybeNulllableValue<Text>` says that the output values has the same nullability as the input value. That means Nullable<Json<T>> evaluates to Nullable<Text> and Json<T> evaluates to Text via the Arr::Out return type

anna-ahmed19:
Thank you @guissalustiano for the suggestions. 
I have added `MaybeNullableValue<Text>` and the `None` tests. 
Let me know any further changes.

guissalustiano (approved):
Thanks!

weiznich:
Thanks for submitting this PR :heart:

PR #4251, opened 2024-09-11
Rewrite positional order similarly to OrderDsl and add order clause to combinations (UNION, ...)

Aethelflaed:
Here's my work so far for https://github.com/diesel-rs/diesel/discussions/4246

`CombinationClause` gets an `order` field with its generic type defaulting to `NoOrderClause`.

I changed `PositionalOrderDsl` to mimic the pattern of the other `*Dsl` traits, which I implemented on `CombinationClause`.

The generic `Expr` passed to `PositionalOrderDsl` is further constraint by a new trait called `PositionalOrderExpr`, which in a way replaces the `positional_order_dsl::Order` trait


diesel/src/query_dsl/positional_order_dsl.rs
```diff
@@ -1,61 +1,40 @@
...
-    }
+impl Expression for OrderColumn {
+    type SqlType = crate::sql_types::Integer;
 }
 
-#[derive(Debug, Clone, Copy, QueryId)]
-pub struct OrderColumn(u32);
-
+// XXX: Do I need to add DieselReserveSpecialization here?
 impl<DB: Backend> QueryFragment<DB> for OrderColumn {
```
Aethelflaed, line 38: Do I need to add DieselReserveSpecialization here?

diesel_tests/tests/combination.rs
```diff
@@ -26,7 +26,7 @@ fn union() {
     let data: Vec<_> = users
         .filter(id.le(tess.id))
         .union(users.filter(id.ge(tess.id)))
-        .positional_order_by(2) // name is the second column
+        .positional_order_by(OrderColumn::from(2)) // name is the second column
```
Aethelflaed, line 29: The main drawback of the current implementation is the need to explicitly call `OrderColumn::from`.

On the other hand,  `1u32.desc()` didn't work either before, so explicit typing was required in some cases anyway.

We could add a helper method, e.g. something like `column_index(1)`. I need feedback on that

weiznich: As this is currently an internal API it's not worth to spend too much time on optimizing this, otherwise we could adjust the trait bounds on the `positional_order_by` methods with an additional `AsPositionalOrderExpr` that is implemented for:

* All types that implement `PositionalOrderExpr`
* `u32`

and that transforms `u32` to `OrderColumn`.

diesel/src/pg/expression/operators.rs
```diff
@@ -71,6 +71,10 @@ __diesel_infix_operator!(
     backend: Pg
 );
 
+use crate::query_dsl::positional_order_dsl::PositionalOrderExpr;
```
weiznich, line 74: Can we move this use statement to the top of the file?

weiznich (commented):
Thanks for opening this PR.

The implementation looks good. I have some minor stylistic remarks and I would like to see at least the `// TODO` line removed.

diesel_tests/tests/combination.rs
Aethelflaed: I've taken a shot at that (except I named it IntoPositionalOrderExpr, as I'd need to return a new OrderColumn), can you tell me what you think?

diesel_tests/tests/combination.rs
```diff
@@ -57,7 +57,7 @@ fn union_all() {
     let data: Vec<_> = users
         .filter(id.le(tess.id))
         .union_all(users.filter(id.ge(tess.id)))
-        .positional_order_by(2) // name is the second column
+        .positional_order_by(OrderColumn::from(2)) // name is the second column
```
weiznich, line 60: That shouldn't be required anymore now that you've implemented `IntoPositionalOrderExpr`. Given that I would prefer to roll back these changes (`OrderColumn::from()`) here.

weiznich (commented):
Thanks for the update. Given that you already started to implement the `IntoPositionalOrderExpr` solution I would prefer to use that in the tests instead of explicitly constructing a `OrderColumn` there.

diesel_tests/tests/combination.rs
Aethelflaed: sorry, I wanted to implement IntoPositionalOrderExpr for tuples too but then got busy!

PR #4252, opened 2024-09-11
added support for `json_array_length` and `jsonb_array_length` functions

zaira-bibi:
Added support for `json_array_length` and `jsonb_array_length` functions under #4216.

Help needed:
1. I can't figure out the correct signature for both of the functions. I have implemented a `JsonOrNullableJson` trait but it does not tell that the functions are for json arrays only.
2. Under the current signature, I can't seem to test for nullability as `json!(null)` gives a scalar.

I'd really appreciate some guidance on this. TIA.


zaira-bibi:
@weiznich I would really appreciate some help on the issues mentioned in the PR description. Thanks!

diesel/src/pg/expression/functions.rs
```diff
@@ -1634,3 +1636,83 @@ define_sql_function! {
...
+    ///
+    /// assert_eq!(result, 0);
+    ///
+    ///
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+
+    fn json_array_length<E: JsonOrNullableJson + SingleValue>(e: E) -> Integer;
```
weiznich, line 1677: Sorry for taking a few days to write this answer, this somehow got lost between all the other PR's :disappointed: 

This is already mostly the correct signature. We cannot restrict this function to only accept json arrays at compile time as that's a purely runtime concept, so it's fine to accept `JsonOrNullableJson` here. What needs to be done is to make the nullability of the return value dependend on the input nullability, which means if you put in a `Nullable<Json>` we want to return a `Nullable<Integer>` as the function returns `null` for `null` inputs.

You can do that by changing the function signature as follows:

```suggestion
    fn json_array_length<E: JsonOrNullableJson + MaybeNullableValue<Integer>>(e: E) -> E::Out;
```

As for the test: You should be able to test for null values like this:

```
let result = diesel::select(json_array_length::<Nullable<Json>, _>(None::<serde_json::Value>)).get_result::<Option<i32>(conn);
```

(The same applies to the jsonb variant of this function)
```

zaira-bibi: Thank you so much for the explanation! I have pushed in a new commit for the required changed. Please let me know if anything else needs to be changed/fixed. Thanks once again!

weiznich (approved):
Looks good now, thanks :pray:

PR #4253, opened 2024-09-11
implement json_object with separate keys and values arguments

valkrypton:
added support for `json_object(keys[],values[])` under [#4216](https://github.com/diesel-rs/diesel/issues/4216)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1656,6 +1656,51 @@ define_sql_function! {
 
 #[cfg(feature = "postgres_backend")]
 define_sql_function! {
+    /// This form of json_object takes keys and values pairwise from two separate arrays.
+    /// In all other respects it is identical to the one-argument form.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::json_object_with_keys_and_values;
+    /// #     use diesel::sql_types::{Array, Json, Nullable, Text};
+    /// #     use serde_json::Value;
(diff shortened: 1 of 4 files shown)


guissalustiano (approved):
Thanks for open the PR!
The changes look good to me.

diesel/src/pg/expression/functions.rs
```diff
@@ -1634,3 +1636,48 @@ define_sql_function! {
...
+    /// let json = diesel::select(json_object_with_keys_and_values::<Array<Text>, Array<Text>, _, _>(
+    ///             vec!["hello","John"],empty))
+    ///             .get_result::<Value>(connection);
+    /// assert!(json.is_err());
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "json_object"]
+    fn json_object_with_keys_and_values<K: TextArrayOrNullableTextArray + MaybeNullableValue<Json>, V: TextArrayOrNullableTextArray + MaybeNullableValue<Json>>(
```
weiznich, line 1680: Unfortunately that's still not 100% correct as it only couples the nullability of the return value onto the first argument, but this function seems to return also `Null` if you pass in `Null` as second argument. (I'm sorry that null value handling is such a mess. I also feel that postgres doesn't document this in a good fashion at all.)

I cannot say without further experimentation how to solve this. I will update the comment as soon as I know more.

valkrypton: I did write a doc test where the 2nd argument is `null` and it returns `null` as expected, but you're right that the output is coupled with the first argument only. I couldn't think of a way to couple it with both arguments

valkrypton:
@weiznich i made a small change to the signature, take a look now

weiznich:
I've pushed https://github.com/diesel-rs/diesel/pull/4253/commits/b488df40b6ddbcbbf0e35225badc90513cabcc4e which introduces a new helper trait that essentially evaluates to `Nullable<Out>` if one of the input types in nullable, that should fix the issue.

PR #4254, opened 2024-09-11
`jsonb_pretty` function

anna-ahmed19:
Added th `jsonb_pretty` function under [Issue#4216](https://github.com/diesel-rs/diesel/issues/4216)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -1653,8 +1653,8 @@ define_sql_function! {
     /// # #[cfg(feature = "serde_json")]
     /// # fn run_test() -> QueryResult<()> {
     /// #     use diesel::dsl::jsonb_pretty;
-    /// #     use serde_json::json;
-    /// #     use diesel::sql_types::Jsonb;
+    /// #     use serde_json::{json, Value};
+    /// #     use diesel::sql_types::{Jsonb, Nullable};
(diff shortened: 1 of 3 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -1634,3 +1635,75 @@ define_sql_function! {
...
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!(r#"null"#, result);
+    ///
+    /// let result = diesel::select(jsonb_pretty::<Jsonb, _>(json!({})))
+    ///     .get_result::<String>(connection)?;
+    ///
+    /// assert_eq!(r#"{
+    /// }"#, result);
+    /// #     Ok(())
```
guissalustiano, line 1705: Could you add a test case with a None input?

guissalustiano (changes requested):
Thanks for opening the PR!

The changes look good, I only would like to see some tests with NULL input (I explain the reason [in yours other PR](https://github.com/diesel-rs/diesel/pull/4250#pullrequestreview-2296751838).

anna-ahmed19:
Thank you for the suggestion.
I have added `MaybeNullableValue<Text>` and the `None` test.
Let me know if this is alright now.

guissalustiano (approved):
Thanks!

PR #4255, opened 2024-09-11
added support for `json_strip_nulls` and `jsonb_strip_nulls`

zaira-bibi:
Added support for functions `json_strip_nulls` and `jsonb_strip_nulls` under #4216.


weiznich (changes requested):
Thanks for opening this PR.

I added a few comments about the signatures of these functions. Additionally it would be great to have a test for `json_strip_null(None)` to verify that the null value handling work.

zaira-bibi:
@weiznich Thank you sm for the review! I have pushed in a commit for the requested changes. Please let me know if it's alright or if there are any further changes.

weiznich (approved):
Looks good, thanks for the update :pray:

PR #4256, opened 2024-09-12
implement jsonb_object (both variants)

valkrypton:
added support for `jsonb_object(text[])` and `jsonb_object(keys[], values[])` under [#4216](https://github.com/diesel-rs/diesel/issues/4216)

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -2078,3 +2078,96 @@ define_sql_function! {
 
     fn jsonb_array_length<E: JsonbOrNullableJsonb + MaybeNullableValue<Integer>>(jsonb: E) -> E::Out;
 }
+
+#[cfg(feature = "postgres_backend")]
+define_sql_function! {
+    /// Builds a JSON object out of a text array. The array must have an even number of members,
+    /// in which case they are taken as alternating key/value pairs. This function also has a form that
+    /// that takes keys and values as separate text array arguments.
+    /// See [jsonb_object_with_keys_and_values]
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::jsonb_object;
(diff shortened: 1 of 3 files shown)


weiznich (changes requested):
Thanks for opening this PR. The general definitions look fine, but I think we might need some small tweaks for correctly handling null values in all the cases.

diesel/src/pg/expression/functions.rs
```diff
@@ -1992,3 +1992,104 @@ define_sql_function! {
...
+    /// let expected:Value = serde_json::json!({"hello": "world","John": "Doe"});
+    /// assert_eq!(expected, jsonb);
+    ///
+    /// let jsonb = diesel::select(jsonb_object::<Array<Text>, _>(vec!["hello","world","John"]))
+    ///                 .get_result::<Value>(connection);
+    /// assert!(jsonb.is_err());
+    ///
+    /// let jsonb = diesel::select(jsonb_object::<Nullable<Array<Text>>, _>(None::<Vec<String>>))
+    ///                 .get_result::<Option<Value>>(connection);
+    /// assert!(jsonb.is_err());
```
weiznich, line 2033: That does seem to be not correct, this should return `None` and no error.

weiznich (changes requested):
Thanks for the update. I locally checked how `jsonb_object(null)` works and it seems to return a `null` value. Therefore I'm not sure why the doc tests checks for an error instead, that does not seem to be correct.

diesel/src/pg/expression/functions.rs
valkrypton: If I replace the current statement with this: 

```
let jsonb = diesel::select(jsonb_object::<Nullable<Array<Text>>, _>(None::<Vec<String>>))
           .get_result::<Option<Value>>(connection)?;
``` 

The test fails because of this error:

```
called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "current transaction is aborted, commands ignored until end of transaction block")
```

I tried with different variations of the argument type, but couldn't get this error to resolve.

weiznich: That's likely caused by the error in the statement beforehand. Either get a new connection here or reorder the test cases to first have those without error and then these that return an error.

weiznich (approved):
Thanks for the update, looks good now :heart:

PR #4257, opened 2024-09-12
added `jsonb_insert`

aznszn:
added the `jsonb_insert` function.

Having issues with coupling the output with the three arguments because the output is null if any of the inputs are null and the second argument is a text array whereas the first and second are jsonb.

I could make the output `Nullable<Json>` but that reduced ergonomics of the function.

I would appreciate some guidance on how to move forward, perhaps a new trait?

Once this is reolved I will add the other variant of this function


aznszn:
@weiznich

diesel/src/pg/expression/functions.rs
```diff
@@ -2078,3 +2078,100 @@ define_sql_function! {
...
+    ///
+    /// let result = diesel::select(jsonb_insert::<Nullable<Jsonb>, Array<Text>, _, _, _>(json!([1, 2, 3]), vec!["1"], Option::<Value>::None))
+    ///                 .get_result::<Option<Value>>(connection)?;
+    /// assert_eq!(None, result);
+    ///
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn jsonb_insert<J: JsonbOrNullableJsonb + SingleValue, P: TextArrayOrNullableTextArray + CombinedNullableValue<J, Jsonb>>(target: J, path: P, value: J) -> P::Out;
```
weiznich, line 2124: We need to have separate generic types for the first and last argument here to allow using different types in these places. That allows us to use a column in one place and an expression in another place while calling this function. That means we won't be able to use `CombineNullableValue` as it is, we might need another helper trait there or some more experimentation how to combine three or four values with the existing trait.

weiznich (changes requested):
Thanks for opening this PR. The tests look mostly fine, but the function signatures might need more tweaking to express all possibilities correctly.

PR #4259, opened 2024-09-12
Fix unused qualifications

Sindbad-Walter:
The unused_qualifications lint gets triggered even in the macros, an [issue](https://github.com/rust-lang/rust/issues/130277) was  openned on rust to update it but as for now this should prevent the lint from triggerring on macros.

Added #[allow(unused_qualifications)] directly in wrap_in_dummy_mod.
Removed #[allow(unused_qualifications)] from Insertable, reverting https://github.com/diesel-rs/diesel/pull/3353 as it will be taken care by the one on wrap_in_dummy_mod.

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -93,9 +93,10 @@ where
 }
 
 pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
-    #[allow(unused_qualifications)] // Can remove if https://github.com/rust-lang/rust/issues/130277 gets done
+    // #[allow(unused_qualifications)] can be removed if https://github.com/rust-lang/rust/issues/130277 gets done
     quote! {
         #[allow(unused_imports)]
+        #[allow(unused_qualifications)]
         const _: () = {
             // This import is not actually redundant. When using diesel_derives
             // inside of diesel, `diesel` doesn't exist as an extern crate, and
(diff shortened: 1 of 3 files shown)


diesel_derives/src/util.rs
```diff
@@ -95,6 +95,7 @@ where
 pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
     quote! {
         #[allow(unused_imports)]
+        #[allow(unused_qualifications)] // Can remove if https://github.com/rust-lang/rust/issues/130277 gets done
```
Ten0, line 98: it's probably best to not put the comment in `quote!`, because I would expect that to get the comment to be expanded as well, but we probably really only want the comment within Diesel, not in the expanded code.

Sindbad-Walter: [8b8c43e](https://github.com/diesel-rs/diesel/pull/4259/commits/8b8c43ee9471bd85f9dbaf9c1e4dc4dd67c59493)

Sindbad-Walter: [5c95a6b](https://github.com/diesel-rs/diesel/pull/4259/commits/5c95a6bd510d1194746cd0c89278d1af65dd5653)

weiznich (approved):
Thanks :pray:

PR #4260, opened 2024-09-12
use MaybeNullableValue on pg functions

guissalustiano:
As disscussed on https://github.com/diesel-rs/diesel/discussions/4241, this PR uses MaybeNullableValue in all other pg function where this is possible.

My first list was wrong, I add some extra tests in other functions showing how they return null without a nullable value.


weiznich (approved):
Looks good, thanks for submitting :heart:

Issue #4240, opened 2024-09-06
Add support for currently unsupported Postgres multirange functions


guissalustiano:
Diesel is on path to supports the Postgres multirange type. This is a tracking issue for adding support for the type and it operators.

Type:
- [X] Multirange type on diesel - https://github.com/diesel-rs/diesel/pull/4159
- [X] Generate multirange type on Diesel CLI - https://github.com/diesel-rs/diesel/pull/4183

[Operator list:](https://www.postgresql.org/docs/current/functions-range.html#MULTIRANGE-OPERATORS-TABLE)
- [ ] `anymultirange @> anymultirange → boolean` - Does the first multirange contain the second?
- [ ] `anymultirange @> anyrange → boolean` - Does the multirange contain the range?
- [ ] `anymultirange @> anyelement → boolean` - Does the multirange contain the element?
- [ ] `anyrange @> anymultirange → boolean` - Does the range contain the multirange?
- [ ] `anymultirange <@ anymultirange → boolean` - Is the first multirange contained by the second?
- [ ] `anymultirange <@ anyrange → boolean` - Is the multirange contained by the range?
- [ ] `anyrange <@ anymultirange → boolean` - Is the range contained by the multirange?
- [ ] `anyelement <@ anymultirange → boolean` - Is the element contained by the multirange?
- [ ] `anymultirange && anymultirange → boolean` - Do the multirange overlap, that is, have any elements in common?
- [ ] `anymultirange && anyrange → boolean` - Does the multirange overlap the range?
- [ ] `anyrange && anymultirange → boolean` - Does the range overlap the multirange?
- [ ] `anymultirange << anymultirange → boolean` - Is the first multirange strictly left of the second?
- [ ] `anymultirange << anyrange → boolean` - Is the multirange strictly left of the range?
- [ ] `anyrange << anymultirange → boolean` - Is the range strictly left of the multirange?
- [ ] `anymultirange >> anymultirange → boolean` - Is the first multirange strictly right of the second?
- [ ] `anymultirange >> anyrange → boolean` - Is the multirange strictly right of the range?
- [ ] `anyrange >> anymultirange → boolean` - Is the range strictly right of the multirange?
- [ ] `anymultirange &< anymultirange → boolean` - Does the first multirange not extend to the right of the second?
- [ ] `anymultirange &< anyrange → boolean` - Does the multirange not extend to the right of the range?
[...]

PR #4238, opened 2024-09-06
Try to improve the first user experience with our examples

weiznich:
* Fix a minor issue in one of the sqlite migrations
* Always use bundled libs by default there, as that should "just work" on most platforms
* Minor fix to the crate level documentation to point to the bundled features for pq-sys and mysqlclient-sys as well

PR #4261, opened 2024-09-13
add some multirange operations support

guissalustiano:
Add multirange contains support under #4240

It is a small test to check the format before moving everything else

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -896,37 +896,30 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
     /// ```rust
     /// # include!("../../doctest_setup.rs");
     /// #
-    /// # table! {
-    /// #     posts {
-    /// #         id -> Integer,
-    /// #         versions -> Range<Integer>,
-    /// #     }
-    /// # }
-    /// #
     /// # fn main() {
     /// #     run_test().unwrap();
     /// # }
     /// #
     /// # fn run_test() -> QueryResult<()> {
-    /// #     use self::posts::dsl::*;
-    /// #     use std::collections::Bound;
+    /// #     use diesel::sql_types::{Integer, Range, Multirange};
     /// #     let conn = &mut establish_connection();
-    /// #     diesel::sql_query("DROP TABLE IF EXISTS posts").execute(conn).unwrap();
-    /// #     diesel::sql_query("CREATE TABLE posts (id SERIAL PRIMARY KEY, versions INT4RANGE NOT NULL)").execute(conn).unwrap();
-    /// #
-    /// diesel::insert_into(posts)
-    ///     .values(versions.eq((Bound::Included(5), Bound::Unbounded)))
-    ///     .execute(conn)?;
+    /// assert!(diesel::select(
+    ///     (1..5).into_sql::<Range<Integer>>().contains_range(1..5)
+    ///     ).first::<bool>(conn).unwrap());
     ///
-    /// let cool_posts = posts.select(id)
-    ///     .filter(versions.contains_range((Bound::Included(10), Bound::Included(50))))
-    ///     .load::<i32>(conn)?;
-    /// assert_eq!(vec![1], cool_posts);
+    /// assert!(diesel::select(
+    ///     (1..5).into_sql::<Range<Integer>>().contains_range(1..5)
+    ///     ).first::<bool>(conn).unwrap());
+    /// assert!(!diesel::select(
+    ///     (1..5).into_sql::<Range<Integer>>().contains_range(3..7)
+    ///     ).first::<bool>(conn).unwrap());
     ///
-    /// let amazing_posts = posts.select(id)
-    ///     .filter(versions.contains_range((Bound::Included(2), Bound::Included(7))))
-    ///     .load::<i32>(conn)?;
-    /// assert!(amazing_posts.is_empty());
+    /// assert!(diesel::select(
+    ///     vec![1..5].into_sql::<Multirange<Integer>>().contains_range(vec![1..5])
+    ///     ).first::<bool>(conn).unwrap());
+    /// assert!(!diesel::select(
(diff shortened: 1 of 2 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1779,7 +1790,7 @@ pub trait PgRangeExpressionMethods: Expression + Sized {
 impl<T> PgRangeExpressionMethods for T
 where
     T: Expression,
-    T::SqlType: RangeOrNullableRange,
+    T::SqlType: MultirangeOrRangeMaybeNullable,
```
guissalustiano, line 1793: I can't test without flipping this, and this allows all the range operations to accept multirange without proper test. What do the reviewers think? Can I merge the PR like that and add tests in the future in other small PRs, or should I add a multirange test for all the range methods?

weiznich: The correct way to handle this would be to split up this trait into two:

* The `PgRangeExpressionMethods` contains methods that are only available for `Range` but not for `MultiRange`. That is then only implemented for `T::SqlType: RangeOrNullableRange`
* The `PgRangeAndMultirangeMethods` contains methods that are available for both
* 

(If there are methods that are only available for `MultiRange` we would just add a third trait for that.

guissalustiano: I did a fast check, and it looks like all the range operations are also multirange operations. If this is the case, do you still want to rename it to `PgRangeAndMultirangeMethods`?

I will add the tests for each of the supported methods in this PR

weiznich: In this case it should be fine to keep the old name.

guissalustiano:
Hi @weiznich, sorry for keeping this PR open for so long.
I would like to use select to write some tests. It's simpler to write instead create a table, insert table and write some filters.
But I'm having some problems with how to use the range extension with literal value, the compiler doesn't know how to use the proper trait.

I want to write something like https://github.com/diesel-rs/diesel/pull/4261/commits/063297d7e33aaade96c17da397919e123d623115

weiznich:
No worries for taking some time on this. 

> I would like to use select to write some tests. It's simpler to write instead create a table, insert table and write some filters.
> But I'm having some problems with how to use the range extension with literal value, the compiler doesn't know how to use the proper trait.

It's a good idea to skip the table setup as that should make the tests easier. As for the problem: I think something like this should work:

```rust
diesel::select((1..5).into_sql::<Range<Integer>>().contains(4)).load::<bool>(conn).unwrap();
```

The `into_sql()` call explicitly converts the rust side value into a SQL side expression (bind in this case), which then allows the DSL to work as expected.

guissalustiano:
Hi @weiznich, thanks for the patient. I think it adds a test to all the new behavior.
With this PR we can use the operations between Range and Range, or Multirange and Multirange.
I would like to also support operation between Multirange and Range (and vice versa), for that I would need to change the `T` here to receives `RangeOrMultirange`, but I don't know how to do that. Could you help-me with that? (We can merge this PR and a open another one for that)
https://github.com/diesel-rs/diesel/blob/381be195688db339fe2927e49bc818ab86754dd9/diesel/src/pg/expression/expression_methods.rs#L950-L956

weiznich (approved):
Looks good, thanks :pray:

weiznich:
> I would like to also support operation between Multirange and Range (and vice versa), for that, I would need to change the T here to receive RangeOrMultirange, but I don't know how to do that. Could you help-me with that? (We can merge this PR and an open another one for that)
>
> ```rust
>  fn contains_range<T>(self, other: T) -> dsl::ContainsRange<Self, T> 
> where 
>     Self::SqlType: SqlType, 
>     T: AsExpression<Self::SqlType>, 
> ```

Technically this should be possible by introducing another generic parameter on the function signature above. The `AsExpression<Self::SqlType>` part essentially says: Give me something that can by turned in an SQL expression with the same SQL type (Range/Multirange) than the left hand side. We could now have a `Rhs` generic type there that just says, we expect some `T` that can be turned into the `Rhs` type, and expect that `Rhs` type to be either `Range` or `MultiRange`. Now the problem with that approach is that I'm not sure if that plays well with type inference. I would expect that the compiler than cannot infer the right type of `Rhs` on it's own, which would require that users always need to annotate these calls, which is not that great. I'm not sure on the last point, as I've not tried it yet, so it might be that rustc is clever enough to see that there is often just one type that fulfills this and does not require explict type annotations in that case.

guissalustiano:
> > I would like to also support operation between Multirange and Range (and vice versa), for that, I would need to change the T here to receive RangeOrMultirange, but I don't know how to do that. Could you help-me with that? (We can merge this PR and an open another one for that)
> > ```rust
> >  fn contains_range<T>(self, other: T) -> dsl::ContainsRange<Self, T> 
> > where 
> >     Self::SqlType: SqlType, 
> >     T: AsExpression<Self::SqlType>, 
> > ```
> 
> Technically this should be possible by introducing another generic parameter on the function signature above. The `AsExpression<Self::SqlType>` part essentially says: Give me something that can by turned in an SQL expression with the same SQL type (Range/Multirange) than the left hand side. We could now have a `Rhs` generic type there that just says, we expect some `T` that can be turned into the `Rhs` type, and expect that `Rhs` type to be either `Range` or `MultiRange`. Now the problem with that approach is that I'm not sure if that plays well with type inference. I would expect that the compiler than cannot infer the right type of `Rhs` on it's own, which would require that users always need to annotate these calls, which is not that great. I'm not sure on the last point, as I've not tried it yet, so it might be that rustc is clever enough to see that there is often just one type that fulfills this and does not require explict type annotations in that case.

Thanks so much for the explanation. Can we merge this PR?
I'll open another PR to explore this approach.

weiznich:
> Can we merge this PR?

I will do that as soon as I fixed the CI. That turns out to be more problematic this time for whatever reason :( 
Follow https://github.com/diesel-rs/diesel/pull/4238 for updates.

guissalustiano:
Oh thanks!

PR #4262, opened 2024-09-13
improve range_merge type signature with new CombinedNullableValue

guissalustiano:
This PR uses `CombinedNullableValue` introduced in https://github.com/diesel-rs/diesel/commit/b488df40b6ddbcbbf0e35225badc90513cabcc4e on range_merge

Looking at the other signatures, I couldn't find any other place where we can use this new type.


weiznich (approved):
Thanks, looks good :+1:

PR #4265, opened 2024-09-16
Fix span in proc macro

Sindbad-Walter:
Follow up of https://github.com/rust-lang/rust/issues/130277 that was initially handled in https://github.com/diesel-rs/diesel/pull/4259. The rust team pointed out that the macro was not correctly setting the span and proposed a fix, being the use of Span::mixed_site instead of coping directly the span of the field.

The #[allow(unused_qualifications)] is still here in the wrap_in_dummy_mod as it is hard to make sure the lint is correctly done in every proc macro.

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -93,10 +93,7 @@ where
 }
 
 pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
-    // #[allow(unused_qualifications)] can be removed if https://github.com/rust-lang/rust/issues/130277 gets done
     quote! {
-        #[allow(unused_imports)]
-        #[allow(unused_qualifications)]
         const _: () = {
             // This import is not actually redundant. When using diesel_derives
             // inside of diesel, `diesel` doesn't exist as an extern crate, and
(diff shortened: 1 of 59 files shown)


diesel_derives/src/util.rs
```diff
@@ -93,7 +93,8 @@ where
 }
 
 pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
-    // #[allow(unused_qualifications)] can be removed if https://github.com/rust-lang/rust/issues/130277 gets done
+    // allow(unused_qualifications) is here as it hard to unsure the span is correctly set. Should stay until it is
```
weiznich, line 96: I would rather prefer auditing the derive crate to ensure all usages of spans are encoded correctly. Yes that's a bit of work but it should fix all these issues at once. 
The alternative is to remove this allow and start searching again if something pops up in some code.

weiznich (commented):
Thanks for opening this PR :pray: 

The changes itself look fine to me, although I personally would prefer to address all potential other locations that somewhat use a span at once.

(Also as minor note: Would it be possible to squash all those commits?)

diesel_derives/src/util.rs
Ten0: That seems like a lot of work for something that's seldom going to be noticed considering the warning is not enabled by default. It looks like it's going to be hard to maintain if we don't have it in our CI.

To make that practical to fix and maintain, it looks like [`syn`'s `Spanned` trait](https://docs.rs/syn/latest/syn/spanned/trait.Spanned.html) should have a `span_mixed` method that constructs this.

PR #4269, opened 2024-09-18
Bump similar_asserts to 1.6

weiznich:
This allows us to remove a call to `String::leak` as they relaxed the life time requirements on the one method we use.

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -461,6 +461,7 @@ mod tests {
     use std::sync::mpsc;
     use std::sync::Arc;
     use std::thread;
+    use std::time::Duration;
 
     use crate::r2d2::*;
     use crate::test_helpers::*;
@@ -636,6 +637,9 @@ mod tests {
         assert_eq!(release_count.load(Ordering::Relaxed), 2);
         assert_eq!(checkin_count.load(Ordering::Relaxed), 3);
         assert_eq!(checkout_count.load(Ordering::Relaxed), 3);
+        // this is required to workaround a segfault while shutting down
+        // the pool
+        std::thread::sleep(Duration::from_millis(100));
     }
 
     #[cfg(feature = "postgres")]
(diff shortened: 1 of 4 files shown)


diesel_cli/src/main.rs
```diff
@@ -286,10 +286,10 @@ fn regenerate_schema_if_file_specified(matches: &ArgMatches) -> Result<(), crate
                     // it's fine to leak here, we will
                     // exit the application anyway soon
                     let label = path.file_name().expect("We have a file name here");
-                    let label = label.to_string_lossy().into_owned().leak();
+                    let label = label.to_string_lossy();
```
Mingun, line 289: Comment above obsoleted by this.

weiznich: Thanks for pointing out :heart:. It's fixed in https://github.com/diesel-rs/diesel/pull/4269/commits/325ffc541252e40fafb367553add9ed8c9bda3cd now.

diesel/src/r2d2.rs
```diff
@@ -636,6 +637,9 @@ mod tests {
         assert_eq!(release_count.load(Ordering::Relaxed), 2);
         assert_eq!(checkin_count.load(Ordering::Relaxed), 3);
         assert_eq!(checkout_count.load(Ordering::Relaxed), 3);
+        // this is required to workaround a segfault while shutting down
+        // the pool
+        std::thread::sleep(Duration::from_millis(100));
```
kw217, line 642: This is unfortunate. Do we know what we're waiting for? A timed delay like this has the potential to be unreliable, e.g., if the test system is slow for some reason. If possible it would be best to wait in a loop, testing for whatever condition is needed to avoid the segfault.

kw217 (approved):
Looks good, but it would be better if we could test for a condition rather than blindly wait 100ms. Fine if this is the pragmatic solution for now, but we should improve if we can.

diesel/src/r2d2.rs
weiznich: You are correct that this is not great. Unfortunately I do not know what exactly we are waiting for, just that waiting seems to resolve the segfaults. My guess is that there is something happening in parallel here that causes the segfault. R2d2 is acquiring a new connection, while the pool is shutting down it might just be that openssl cannot handle creating/dropping a connection in a fast turnaround? Or that it returns early but does stuff on it's own in the background, which in turn leads to collisions while shutting down the connection?
Given that I did not found a way to reproduce the problem locally it's relatively hard for me to debug it. I just went for the first thing that seem to resolve the issue to unblock the other PR's for now. If someone has a better solution I'm more than happy to accept that.

PR #4271, opened 2024-09-19
Expose last_insert_id from Ok packet on MySQL

mcronce:
Took a whack at implementing what we discussed in https://github.com/diesel-rs/diesel/discussions/3605 (finally); I'm sure I need to add some documentation at minimum.

Notably, I copied most of the body of `execute_returning_id()` from `execute_returning_count()`, not sure if that's how you want to handle that or not

Might have overabstracted; I implemented it using traits for extensibility.

I'll have a follow-up in `diesel_async` once this is merged.

### DIFF
--- a/diesel/src/mysql/connection/mod.rs
+++ b/diesel/src/mysql/connection/mod.rs
@@ -211,6 +211,34 @@ impl Connection for MysqlConnection {
     }
 }
 
+impl ConnectionWithReturningId for MysqlConnection {
+    type ReturnedId = u64;
+
+    fn execute_returning_id<T>(&mut self, source: &T) -> QueryResult<Self::ReturnedId>
+    where
+        T: QueryFragment<Self::Backend> + QueryId,
+    {
+        #[allow(unsafe_code)] // call to unsafe function
+        update_transaction_manager_status(
+            prepared_query(
+                &source,
+                &mut self.statement_cache,
+                &mut self.raw_connection,
+                &mut *self.instrumentation,
+            )
+            .and_then(|stmt| {
+                // we have not called result yet, so calling `execute` is
+                // fine
+                let stmt_use = unsafe { stmt.execute() }?;
+                Ok(unsafe { stmt_use.insert_id() })
+            }),
+            &mut self.transaction_state,
+            &mut self.instrumentation,
+            &crate::debug_query(source),
+        )
+    }
+}
+
 #[inline(always)]
 fn update_transaction_manager_status<T>(
     query_result: QueryResult<T>,
--- a/diesel/src/mysql/connection/stmt/mod.rs
+++ b/diesel/src/mysql/connection/stmt/mod.rs
@@ -160,6 +160,11 @@ impl<'a> StatementUse<'a> {
             .map_err(|e| Error::DeserializationError(Box::new(e)))
     }
 
+    /// SAFETY:  This should only be called for INSERT queries.
+    pub(in crate::mysql::connection) unsafe fn insert_id(&self) -> u64 {
+        ffi::mysql_stmt_insert_id(self.inner.stmt.as_ptr())
+    }
+
     /// This function should be called after `execute` only
     /// otherwise it's not guaranteed to return a valid result
     pub(in crate::mysql::connection) unsafe fn result_size(&mut self) -> QueryResult<usize> {
--- a/diesel/src/query_dsl/load_dsl.rs
+++ b/diesel/src/query_dsl/load_dsl.rs
@@ -1,7 +1,9 @@
 use self::private::LoadIter;
 use super::RunQueryDsl;
 use crate::backend::Backend;
-use crate::connection::{Connection, DefaultLoadingMode, LoadConnection};
+use crate::connection::{
+    Connection, ConnectionWithReturningId, DefaultLoadingMode, LoadConnection,
+};
 use crate::deserialize::FromSqlRow;
 use crate::expression::QueryMetadata;
 use crate::query_builder::{AsQuery, QueryFragment, QueryId};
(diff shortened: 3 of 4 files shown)


mcronce:
I don't _think_ that that test failure has anything to do with my changes?

diesel/src/mysql/connection/mod.rs
```diff
@@ -211,6 +211,34 @@ impl Connection for MysqlConnection {
...
+                &source,
+                &mut self.statement_cache,
+                &mut self.raw_connection,
+                &mut *self.instrumentation,
+            )
+            .and_then(|stmt| {
+                // we have not called result yet, so calling `execute` is
+                // fine
+                let stmt_use = unsafe { stmt.execute() }?;
+                Ok(unsafe { stmt_use.insert_id() })
```
weiznich, line 233: There should be a `Safety: explain how the invariants of the called function are uphold` comment here.

Additionally as far as I can see the `insert_id()` function states that it is only safe to call this function on insert statements, while we can call it on arbitrary statements here. That sounds like the current implementation is unsound.

In the end we want to replace the `T: QueryFragement<Self::Backend>` bound above with essentially `InsertStatement` to restrict the allowed statements to only insert statements.

diesel/src/query_dsl/load_dsl.rs
```diff
@@ -91,6 +93,34 @@ where
...
+/// it's in a trait to avoid a backwards-incompatible change in the future if such
+/// functionality becomes available in other back-ends.
+///
+/// [`MySQL`]: crate::mysql::Mysql
+pub trait ExecuteInsertWithReturningId<
+    Conn: ConnectionWithReturningId<Backend = DB>,
+    DB: Backend = <Conn as Connection>::Backend,
+>: Sized
+{
+    /// Execute this command
```
weiznich, line 109: There should be a doc test/example on this function to show how this works

weiznich (changes requested):
Thanks for working on this. I've added a few comments around the Safety remarks of the new unsafe function. I think we do need additional restrictions/checks there to make sure the result is defined.

As for the implementation: I personally think we don't need to introduce these two new traits there, as we could just implement the relevant methods directly on `InsertStatement` and `MysqlConnection`. We don't need to be super generic there as these methods are only relevant for this combination. (That's nothing I would block the merge on, although I would be happy to see that resolved)

diesel/src/mysql/connection/mod.rs
mcronce: Will do

diesel/src/mysql/connection/stmt/mod.rs
```diff
@@ -160,6 +160,11 @@ impl<'a> StatementUse<'a> {
             .map_err(|e| Error::DeserializationError(Box::new(e)))
     }
 
+    /// SAFETY:  This should only be called for INSERT queries.
+    pub(in crate::mysql::connection) unsafe fn insert_id(&self) -> u64 {
+        ffi::mysql_stmt_insert_id(self.inner.stmt.as_ptr())
```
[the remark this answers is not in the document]
mcronce: While that's how the documentation is written, I don't think that the result is UB.  The value is always present in the Ok packet and will always be a valid byte sequence for a u64.  The value absolutely might not make any sense if no `AUTO_INCREMENT` column is present, but that's not the same as the "spooky action at a difference" type of Undefined Behavior.

Looking at what `mysql_common` does, for example, the only check it performs is during packet deserialization; `last_insert_id` is an `Option<u64>` in their representation, and if `last_insert_id` on the wire is `0`, they consider that `None`

I also, FWIW, don't have any idea what kind of check we'd be able to perform in a query context for that

diesel/src/mysql/connection/mod.rs
mcronce: > There should be a `Safety: explain how the invariants of the called function are uphold `comment here.

That's covered by the comment on lines 230-231, right?  This was copied+pasted directly from `execute_returning_count()` above and matches it precisely, the only change being the call to `stmt_use.insert_id()` instead of `stmt_use.affected_rows()`

diesel/src/query_dsl/load_dsl.rs
mcronce: Will do

diesel/src/mysql/connection/stmt/mod.rs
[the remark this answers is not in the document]
weiznich: I'm still wary to relay on something that's explicitly documented as not being defined as that could change at any point in the future. Remember that we want to provide a reasonable level of stability that should be decoupled from our dependencies. So yes, it might not be undefined behavior in terms of programming to read this value, the result still stays undefined and might change at any point. It's something that at least I do not feel comfortable with exposing in a stable API.

What we can easily do is to restrict it to insert statements with `Integer` primary keys at compile time by adding the relevant trait bounds on the target type for `InsertStatement`. What's harder to do is to check if the column is actually marked as `AUTO_INCREMENT`, as that would additional changes to include this information in the generated schema. That's still not impossible, although I think it might be to much trouble in this case.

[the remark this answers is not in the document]
Mingun: Another way to quickly get this feature is expose it under some unstable feature flag, if you not against that.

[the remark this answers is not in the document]
weiznich: I would rather prefer restricting it at least to Integer primary keys, as that isn't much harder than marking it as unstable feature. After all that should just require one or two lines of trait bounds on the relevant method.

LucaCappelletti94:
I am closing this PR as superseeded by the merged #5105

PR #4272, opened 2024-09-20
Auto type binary operators

Ten0:
This adds support for type inference of e.g. `some_timestamptz_column - some_pginterval_expression`.

Also adds support for `update(...)` expression (I found myself needing it for the test).

### DIFF
--- a/dsl_auto_type/src/auto_type/expression_type_inference.rs
+++ b/dsl_auto_type/src/auto_type/expression_type_inference.rs
@@ -302,6 +302,54 @@ impl TypeInferrer<'_> {
                     }
                 }
             }
+            (syn::Expr::Binary(binary_expression), None) => {
+                let op_span = Span::mixed_site().located_at(binary_expression.op.span());
+                let trait_name = match binary_expression.op {
+                    syn::BinOp::Add(_) => "Add",
+                    syn::BinOp::Sub(_) => "Sub",
+                    syn::BinOp::Mul(_) => "Mul",
+                    syn::BinOp::Div(_) => "Div",
+                    syn::BinOp::Rem(_) => "Rem",
+                    syn::BinOp::BitXor(_) => "BitXor",
+                    syn::BinOp::BitAnd(_) => "BitAnd",
+                    syn::BinOp::BitOr(_) => "BitOr",
+                    syn::BinOp::Shl(_) => "Shl",
+                    syn::BinOp::Shr(_) => "Shr",
+                    syn::BinOp::And(_)
+                    | syn::BinOp::Or(_)
+                    | syn::BinOp::Eq(_)
+                    | syn::BinOp::Lt(_)
+                    | syn::BinOp::Le(_)
+                    | syn::BinOp::Ne(_)
+                    | syn::BinOp::Ge(_)
+                    | syn::BinOp::Gt(_) => return Ok(parse_quote!(bool)),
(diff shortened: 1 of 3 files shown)


Ten0:
The ubuntu mysql fails seem unrelated.

weiznich:
Yes the test failures are unrelated. I hadn't the time to investigate them yet. What seems to be really strange is that the test itself passes, but nextest then reports a sigsev

> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 52 filtered out; finished in 0.00s

dsl_auto_type/src/auto_type/expression_type_inference.rs
```diff
@@ -302,6 +302,54 @@ impl TypeInferrer<'_> {
...
+                    syn::BinOp::Div(_) => "Div",
+                    syn::BinOp::Rem(_) => "Rem",
+                    syn::BinOp::And(_) => return Ok(parse_quote!(bool)),
+                    syn::BinOp::Or(_) => return Ok(parse_quote!(bool)),
+                    syn::BinOp::BitXor(_) => "BitXor",
+                    syn::BinOp::BitAnd(_) => "BitAnd",
+                    syn::BinOp::BitOr(_) => "BitOr",
+                    syn::BinOp::Shl(_) => "Shl",
+                    syn::BinOp::Shr(_) => "Shr",
+                    syn::BinOp::Eq(_) => "PartialEq",
```
weiznich, line 320: I think that case won't work. [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) is neither a member of `std::ops` nor has it an associated type `Output`. Maybe we can just mark it as unsupported for now?

weiznich (changes requested):
Overall this change looks good to me. Thanks for submitting.

I think `PartialEq` is not handled correctly yet, we might just need to mark it as unsupported for now?

dsl_auto_type/src/auto_type/expression_type_inference.rs
Ten0: Ah tight this one should have been amon the "always boolean". 👍

dsl_auto_type/src/auto_type/expression_type_inference.rs
```diff
@@ -302,6 +302,54 @@ impl TypeInferrer<'_> {
...
+                    | syn::BinOp::DivAssign(_)
+                    | syn::BinOp::RemAssign(_)
+                    | syn::BinOp::BitXorAssign(_)
+                    | syn::BinOp::BitAndAssign(_)
+                    | syn::BinOp::BitOrAssign(_)
+                    | syn::BinOp::ShlAssign(_)
+                    | syn::BinOp::ShrAssign(_) => return Ok(parse_quote!(())),
+                    _ => {
+                        // This is here because the `BinOp` enum is marked as #[non_exhaustive],
+                        // but in effect we really support all the variants
```
weiznich, line 338:
```suggestion
```
Let's just remove this comment is it's not correct anymore

weiznich (approved):
Thanks for the update, I just added a correction to remove the now wrong comment about the fact that we implement all the ops as `PartialEq` is now missing.

weiznich:
You are correct, I just search for `PartialEq` and missed that it's only `Eq` :facepalm: 
I've reverted the removal, so everything should be fine now.

PR #4277, opened 2024-09-22
Allow specifying diesel path as derive attribute

Elrendio:
Resolves #4276

### DIFF
--- a/diesel_derives/src/util.rs
+++ b/diesel_derives/src/util.rs
@@ -1,8 +1,14 @@
-use proc_macro2::TokenStream;
-use quote::quote;
-use syn::parse::{Parse, ParseStream, Peek, Result};
-use syn::token::Eq;
-use syn::{parenthesized, parse_quote, Data, DeriveInput, GenericArgument, Ident, Type};
+use {
+    proc_macro2::TokenStream,
+    quote::quote,
+    syn::{
+        parenthesized,
+        parse::{Parse, ParseStream, Peek, Result},
+        parse_quote,
+        token::Eq,
+        Data, DeriveInput, GenericArgument, Ident, Type,
+    },
+};
 
 use crate::model::Model;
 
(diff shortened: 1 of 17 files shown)


diesel_derives/src/util.rs
```diff
@@ -92,7 +93,14 @@ where
     content.parse_terminated(T::parse, sep)
 }
 
-pub fn wrap_in_dummy_mod(item: TokenStream) -> TokenStream {
+pub fn wrap_in_dummy_mod(
+    item: TokenStream,
+    diesel_path_override: Option<&syn::Path>,
```
Ten0, line 98: Is there a downside to adding diesel itself to its own prelude instead?
It looks like this would solve the same issue (being able to use diesel as an indirect dependency) without requiring as much more code within diesel_derives, and without requiring more boilerplate in the caller's code.

Ten0:
> Might I also suggest picking any import granularity (I however advise `One`) from the ones that can be automatically formatted by cargo fmt?

Unless I'm mistaken currently it is "module"

diesel_derives/src/util.rs
weiznich: I don't think that would work. The import is currently `use diesel` which means we either need to have a direct dependency called diesel or a module with than name in the consuming crate root. Now assuming that someone want's to build a crate on top of diesel that reexports diesel's derives they would need to require that all consumer crates need to depend on the extension crate and diesel. Reexporting diesel from the extension crate wouldn't help as it needs to be done from the consuming crate.

Ten0: > we either need to have a direct dependency called diesel or a module with than name in the consuming crate root

So it seems that a module won't work anymore.
That [was the case](https://github.com/diesel-rs/diesel/blob/4f48b2b99cdbb161b18ea68d04077a8e6500db4c/diesel/src/lib.rs#L332-L334) at the time [the comment in the existing code was written](https://github.com/diesel-rs/diesel/commit/4f48b2b99cdbb161b18ea68d04077a8e6500db4c), but now we need [`extern crate self as diesel;`](https://github.com/diesel-rs/diesel/blob/f7e9819303877b5a255143c4b32fa1e5745aaad0/diesel/src/lib.rs#L767)

But in addition to direct dependency and `extern crate` in the crate root, we can also resolve things that are explicitly brought in scope, including by a prelude. That is, the following code compiles:

```rust
mod y {
	use crate::prelude::*;

	use x;

	use x::main;
}

mod prelude {
	pub use crate as x;
}
fn main() {}
```

So it looks like adding `pub use diesel;`  to the `diesel::prelude` would lead to every crate importing the derives via the diesel prelude (which is the most natural way to do so) to also get `diesel` in scope, in turn making the code expanded by the derives work, because it would then be able to resolve `diesel`.

This seems to be a simpler way to be able to use diesel as an indirect dependency than what is currently implemented in this PR. (And that is the sole goal of this PR.)

pksunkara: @Elrendio Would you be willing to try proposed approach to see if it works?

PR #4281, opened 2024-09-23
Add support for joins to sub-joins to aliases

Ten0:
Prior to this, we couldn't write the following:
```rust
diesel::alias! {
	schema::t1 as t1_alias: T1,
	schema::t2 as t2_alias: T2,
	schema::t3 as t3_alias: T3
}

t1_alias.inner_join(t2_alias.inner_join(t3_alias)) // <- This used to fail to compile
```

This now works as expected, that is, using the implicit join clause between t1 and t2, with both hitting their respective aliases.

The call path is as follows:
- `Alias<T1>: JoinTo<SelectStatement<t2_alias, joined to t3_alias>>`
- `SelectStatement<FromClause<t2_alias>, joined to t3_alias>: JoinTo<Alias<T1>>`
- `t2_alias: JoinTo<Alias<T1>>` (in macro expansion)
- `Alias<T1>: JoinTo<t2_alias>` (this is the first `impl<T: Table, S> JoinTo<T> for Alias<S>`, and that one performs the remapping of the on clause, so only of the fields of the alias, as otherwise already happens when joining t2_alias to t1_alias directly).

### DIFF
--- a/diesel/src/query_source/aliasing/joins.rs
+++ b/diesel/src/query_source/aliasing/joins.rs
@@ -6,11 +6,12 @@ use super::{Alias, AliasSource, AliasedField};
 use crate::expression::{AppearsOnTable, SelectableExpression};
 use crate::query_builder::AsQuery;
 use crate::query_dsl::InternalJoinDsl;
-use crate::query_source::joins::ToInnerJoin;
 use crate::query_source::joins::{
-    AppendSelection, Inner, Join, JoinOn, JoinTo, LeftOuter, OnClauseWrapper,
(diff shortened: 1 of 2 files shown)


weiznich (changes requested):
Thanks for submitting. The change itself looks fine, but I would appreciate having a test for this case as well. It doesn't need to be something complex, just something that shows that this now compiles.

Ten0:
> Thanks for submitting. The change itself looks fine, but I would appreciate having a test for this case as well. It doesn't need to be something complex, just something that shows that this now compiles.

You're right I was lazy, I only tested it on my own codebase 😁
Done in 721c137baebb9d117f7e8292ff246846b2697384.

weiznich (approved):

PR #4284, opened 2024-09-24
Add SQLite support for serde_json::Value using the Json/Jsonb

JMLX42:
Implements #3918 

- [x] Implement `FromSql<Json, Sqlite> for serde_json::Value`
- [x] Implement `ToSql<Json, Sqlite> for serde_json::Value`
- [x] Implement `FromSql<Jsonb, Sqlite> for serde_json::Value`
- [x] Implement `ToSql<Jsonb, Sqlite> for serde_json::Value`
- [x] Add unit tests for `FromSql<Json, Sqlite>`
- [x] Add unit tests for `ToSql<Json, Sqlite>`
- [x] Add unit tests for `FromSql<Jsonb, Sqlite>`
- [x] Add unit tests for `ToSql<Jsonb, Sqlite>`
- [x] Add unit tests for SQLite's JSONB format reader/writer
- [x] Update the documentation
- [x] Update `CHANGELOG.md`
- ~~[ ] Update the CI (if needed)~~ (update: nope)
- [x] Make sure the CI passes
- [ ] Cleanup the git history


diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,164 @@
...
+    //         .values((
+    //             test_insert_json_into_table_as_text::id.eq(1),
+    //             test_insert_json_into_table_as_text::json.eq(value),
+    //         ))
+    //         .execute(conn)
+    //         .unwrap();
+    // }
+
+    #[test]
+    fn some_json_from_sql() {
```
JMLX42, line 95: @weiznich I am having a lot of difficulties writing the tests. I tried to mimic what's done for Pg or Sqlite/Chrono, without success.

I can't find the SQLite equivalent for:

```rust
#[test]
    fn json_to_sql() {
        let mut buffer = Vec::new();
        let mut bytes = Output::test(ByteWrapper(&mut buffer));
        let test_json = serde_json::Value::Bool(true);
        ToSql::<sql_types::Json, Pg>::to_sql(&test_json, &mut bytes).unwrap();
        assert_eq!(buffer, b"true");
    }
```

The SQLite interface for `Output` is completely different.

Any help would be greatly appreciated.

weiznich (commented):
Thanks for opening this PR. This is already a good starting point, but we need to adjust the (de)serialization of the jsonb type to match the actual underlying SQLite data format. Additionally tests + documentation need some tweaks.

Finally it would be great to have a changelog entry for this change.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+#[cfg(all(feature = "sqlite", feature = "serde_json"))]
+impl FromSql<sql_types::Jsonb, Sqlite> for serde_json::Value {
+    fn from_sql(value: SqliteValue<'_, '_, '_>) -> deserialize::Result<Self> {
+        let bytes = value.read_blob();
+        if bytes[0] != 1 {
+            return Err("Unsupported JSONB encoding version".into());
+        }
+        serde_json::from_slice(&bytes[1..]).map_err(|_| "Invalid Json".into())
```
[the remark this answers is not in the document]
JMLX42: @weiznich I might have misunderstood something, but how does Pg's support for JSONB work then?

https://github.com/diesel-rs/diesel/blob/f7e9819303877b5a255143c4b32fa1e5745aaad0/diesel/src/pg/types/json.rs#L35

As far as I can tell, it reads the first byte then assumes the rest is a valid JSON UTF-8 String. Or am I missing something?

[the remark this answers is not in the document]
JMLX42: I have updated my PR with commits to implement a proper SQLite JSONB reader/writer. Including unit tests.

diesel/src/sqlite/types/json.rs
JMLX42: @weiznich thank you! Here is what I tried:

```rust
#[test]
    fn json_to_sql() {
        let conn = &mut connection();
        let value = json!(true);
        let res = diesel::select(value.into_sql::<Jsonb>().eq(&sql("true")))
            .get_result::<bool>(conn)
            .unwrap();
        assert!(res);
    }
```

But it does not work:

```
error[E0277]: the trait bound `Value: expression::Expression` is not satisfied
   --> diesel/src/sqlite/types/json.rs:521:40
    |
521 |         let res = diesel::select(value.into_sql::<Jsonb>().eq(sql("true")))
    |                                        ^^^^^^^^ the trait `expression::Expression` is not implemented for `Value`, which is required by `Value: expression::AsExpression<Jsonb>`
    |
    = help: the following other types implement trait `expression::Expression`:
              &'a T
              (T0, T1)
              (T0, T1, T2)
              (T0, T1, T2, T3)
              (T0, T1, T2, T3, T4)
              (T0, T1, T2, T3, T4, T5)
              (T0, T1, T2, T3, T4, T5, T6)
              (T0, T1, T2, T3, T4, T5, T6, T7)
            and 122 others
note: required for `Value` to implement `expression::AsExpression<Jsonb>`
```

diesel/src/sql_types/mod.rs
```diff
@@ -395,11 +395,110 @@ pub struct Timestamp;
...
+    doc = "[`serde_json::Value`]: serde_json::value::Value"
+)]
+#[cfg_attr(
+    not(feature = "serde_json"),
+    doc = "[`serde_json::Value`]: https://docs.rs/serde_json/1.0.64/serde_json/value/enum.Value.html"
+)]
+///
+/// # Examples
+///
+/// ```rust
```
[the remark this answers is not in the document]
JMLX42: @weiznich AFAIK this is the part that's not compatible with SQLite:

```
/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
/// #     use diesel::insert_into;
/// #     use self::contacts::dsl::*;
/// #     let connection = &mut connection_no_data();
/// #     diesel::sql_query("CREATE TABLE contacts (
/// #         id SERIAL PRIMARY KEY,
/// #         name VARCHAR NOT NULL,
/// #         address JSONB NOT NULL
/// #     )").execute(connection)?;
```

**But** it doesn't show in the rendered doc example:

```rust
table! {
    contacts {
        id -> Integer,
        name -> VarChar,
        address -> Jsonb,
    }
}

let santas_address: serde_json::Value = serde_json::from_str(r#"{
    "street": "Article Circle Expressway 1",
    "city": "North Pole",
    "postcode": "99705",
    "state": "Alaska"
}"#)?;
let inserted_address = insert_into(contacts)
    .values((name.eq("Claus"), address.eq(&santas_address)))
    .returning(address)
    .get_result::<serde_json::Value>(connection)?;
assert_eq!(santas_address, inserted_address);
```

And I think that code samples works fine for both SQLite and Pg.

diesel/src/sql_types/mod.rs
```diff
@@ -395,11 +395,110 @@ pub struct Timestamp;
...
+///
+/// - [`serde_json::Value`]
+///
+/// ### [`FromSql`] impls
+///
+/// - [`serde_json::Value`]
+///
+/// [`ToSql`]: crate::serialize::ToSql
+/// [`FromSql`]: crate::deserialize::FromSql
+/// [`jsonb`]: https://www.postgresql.org/docs/current/datatype-json.html
```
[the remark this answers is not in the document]
JMLX42: @weiznich done in a557574b012c5749589735ed7f7ab70089510d8b

diesel/src/sql_types/mod.rs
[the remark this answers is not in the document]
weiznich: The main issue here is that this code is run as test for both the sqlite and the postgres backend. That means it will fail due to the not shown part for sqlite. We somehow need to fix that, possibly by conditionally creating the right table structure for both supported backends.

diesel/src/sqlite/types/json.rs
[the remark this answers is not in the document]
weiznich: The test does not work because there are missing parts in the implementation:

* We need to include this module for sqlite as well: https://github.com/diesel-rs/diesel/blob/master/diesel/src/type_impls/mod.rs#L5
* We need to allow mapping to jsonb for sqlite as well: https://github.com/diesel-rs/diesel/blob/master/diesel/src/type_impls/json.rs#L12

diesel/src/sqlite/types/json.rs
JMLX42: @weiznich [This](https://github.com/diesel-rs/diesel/pull/4284#discussion_r1775089572) + `use crate::ExpressionMethods;` fixed the build of [that test](https://github.com/diesel-rs/diesel/pull/4284#discussion_r1774135529).

And the following test passes:

```rust
#[test]
    fn json_to_sql() {
        let conn = &mut connection();
        let value = json!(true);
        let res = diesel::select(value.into_sql::<Json>().eq(&sql("json('true')")))
            .get_result::<bool>(conn);
        assert!(res.unwrap());
    }

    #[test]
    fn jsonb_to_sql() {
        let conn = &mut connection();
        let value = json!(true);
        let res = diesel::select(value.into_sql::<Jsonb>().eq(&sql("jsonb('true')")))
            .get_result::<bool>(conn);
        assert!(res.unwrap());
    }
```

I' ll make them more comprehensive and commit them.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    // Write the ASCII text representation of the float as the payload
+    buffer.extend_from_slice(float_str.as_bytes());
+
+    Ok(IsNull::No)
+}
+
+// Write a JSON string
+fn write_jsonb_string(s: &str, buffer: &mut Vec<u8>) -> serialize::Result {
+    write_jsonb_header(buffer, JSONB_TEXT, s.len())?;
```
weiznich, line 302: It might be meaningful to use `JSONB_TEXTRAW` here as we don't handle any escaping. That seems like the easiest solution, although I'm not sure if that has any downsides in compatibility to other tools. That might require some additional tests to confirm that at least sqlite can handle that.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    let value = match element_type {
+        JSONB_NULL => Ok(serde_json::Value::Null),
+        JSONB_TRUE => Ok(serde_json::Value::Bool(true)),
+        JSONB_FALSE => Ok(serde_json::Value::Bool(false)),
+        JSONB_INT => read_jsonb_int(payload_bytes, payload_size),
+        JSONB_INT5 => Err("INT5 is not supported in this implementation".into()),
+        JSONB_FLOAT => read_jsonb_float(payload_bytes, payload_size),
+        JSONB_FLOAT5 => Err("FLOAT5 is not supported in this implementation".into()),
+        JSONB_TEXT | JSONB_TEXTJ | JSONB_TEXTRAW => read_jsonb_text(payload_bytes, payload_size),
```
weiznich, line 127: I think we would need some additional handling for `JSONB_TEXT` and `JSONB_TESTJ` to handle escaped characters as as far as I remember `serde_json::Value::String` is expected to contain escaped values.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,575 @@
...
+        assert!(res.unwrap());
+    }
+
+    #[test]
+    fn jsonb_to_sql() {
+        let conn = &mut connection();
+        let res = diesel::select(json!(true).into_sql::<Jsonb>().eq(&sql("jsonb('true')")))
+            .get_result::<bool>(conn);
+        assert!(res.unwrap());
+    }
```
weiznich, line 540: This test is a good starting point, but I would like to see some more variants of this for:

* All variants of `serde_json::Value`
* Some edgecases including escaped string values and new lines

weiznich (commented):
Thanks for the update. It looks now much closer to what I had in mind here. I think we still should improve the test situation by adding more tests for jsonb roundtrips. We also might want to have a few tests that load data from the database and check that the expected result comes back. Finally it seems like that the handling of escaped string sequences is still missing from the parser/writer implementation. This should also be covered by tests.

diesel/src/sqlite/types/json.rs
JMLX42: @weiznich done fd7f5658

diesel/src/sqlite/types/json.rs
JMLX42: > Some edgecases including escaped string values and new lines

@weiznich done.

diesel/src/sqlite/types/json.rs
JMLX42: > It might be meaningful to use JSONB_TEXTRAW here as we don't handle any escaping.

@weiznich do you have an actual example of a valid Rust String that would be interpreted as TEXTRAW by SQLite?

JMLX42: The main problem is that I don't think `serde_json::Value` deals only with `String`. So my understanding is that there is no `String` value that would be written as a `TEXTRAW` field in SQLite's JSONB. But I might be wrong.

If I am right, the `TEXTRAW` is out of scope for this MR that revolves solely around `serde_json::Value`.

diesel/src/sql_types/mod.rs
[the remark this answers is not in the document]
JMLX42: @weiznich Done in 3a1a612c

diesel/src/sql_types/mod.rs
```diff
@@ -395,11 +395,173 @@ pub struct Timestamp;
...
+/// let santas_address: serde_json::Value = serde_json::from_str(r#"{
+///     "street": "Article Circle Expressway 1",
+///     "city": "North Pole",
+///     "postcode": "99705",
+///     "state": "Alaska"
+/// }"#)?;
+/// let inserted_address = insert_into(contacts)
+///     .values((name.eq("Claus"), address.eq(&santas_address)))
+///     .returning(address)
+///     .get_result::<serde_json::Value>(connection)?;
```
weiznich, line 549: I think that fails on CI because support for get_result() for insert statements for sqlite is behind a feature flag.

diesel/src/sql_types/mod.rs
```diff
@@ -395,11 +395,173 @@ pub struct Timestamp;
...
+/// table! {
+///     contacts {
+///         id -> Integer,
+///         name -> Text,
+///         address -> Jsonb,
+///     }
+/// }
+///
+/// # #[cfg(all(feature = "sqlite", feature = "serde_json"))]
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
```
weiznich, line 528: I would prefer to not duplicate the whole example. I suggest to just put the create table statement behind `#[cfg(feature = "postgres")]` and `#[cfg(feature = "sqlite")]`, as everything else is the same.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,852 @@
...
+        JSONB_ARRAY => read_jsonb_array(payload_bytes, payload_size),
+        JSONB_OBJECT => read_jsonb_object(payload_bytes, payload_size),
+        _ => Err(format!("Unsupported or reserved JSONB type: {}", element_type).into()),
+    }?;
+
+    Ok((value, total_size))
+}
+
+// Read a JSON integer in canonical format (INT)
+fn read_jsonb_int(bytes: &[u8], payload_size: usize) -> deserialize::Result<serde_json::Value> {
```
weiznich, line 148: bytes should already be payload_size here, right? In that case we don't need to take another subslice below. (The same likely applies to the other read functions as well)

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,852 @@
...
+}
+
+// Helper function to create the correct JsonbHeader based on the payload size
+fn create_jsonb_header(element_type: u8, payload_size: usize) -> Result<Vec<u8>, String> {
+    // Check if payload size exceeds the maximum allowed size
+    if payload_size > 2_147_483_647 {
+        return Err("Payload size exceeds the maximum allowed size of 2GB".into());
+    }
+
+    let mut header = Vec::new();
```
weiznich, line 254: We likely should use `Vec::with_capacity()` here as we know the payload size beforehand?

diesel/src/sqlite/types/json.rs
weiznich: My idea was to not bother with deciding if we can write something as `JSONB_TEXT` or not and always mark the buffer as `TEXTRAW`. As you already implemented the better more complete variant that's also fine and even preferable.

weiznich (changes requested):
Thanks for the update. I feel we are mostly there now. I've added a few minor remarks otherwise I would like to see an entry into the `Changelog.md` file in the repository root for this feature + at least a handful tests that are essentilally:

```rust
let res = diesel::select(sql::<Jsonb>("jsonb('{\"whatever\": 42}')")).get_result::<serde_json::Value>(conn).unwrap();
assert_eq!(res, serde_json::json!(…));
```

to verify that loading values also works

diesel/src/sql_types/mod.rs
JMLX42: @weiznich like this def9cee0 ?

JMLX42:
> Thanks for the update. I feel we are mostly there now. I've added a few minor remarks otherwise I would like to see an entry into the `Changelog.md` file in the repository root for this feature + at least a handful tests that are essentilally:
> 
> ```rust
> let res = diesel::select(sql::<Jsonb>("jsonb('{\"whatever\": 42}')")).get_result::<serde_json::Value>(conn).unwrap();
> assert_eq!(res, serde_json::json!(…));
> ```
> 
> to verify that loading values also works

@weiznich done in 7fe0c670

JMLX42:
> Finally it would be great to have a changelog entry for this change.

@weiznich Done in 76ac6504

JMLX42:
@weiznich I just updated the description with the check list. Please update it if necessary.

weiznich:
The added changes look fine so far, the only thing missing here is making the CI pass. That likely requires small adjustments to when the tests are run. For postgres they should be gated with `#[cfg(all(feature = "postgres", feature = "serde_json"))]` for sqlite they likely require `#[cfg(all(feature = "sqlite", feature = "serde_json", feature = "returning_clauses_for_sqlite_3_35"))]`.

JMLX42:
> for sqlite they likely require `#[cfg(all(feature = "sqlite", feature = "serde_json", feature = "returning_clauses_for_sqlite_3_35"))]`.

@weiznich it gets deeper: since the doc for the `Jsonb` type also needs the `returning_clauses_for_sqlite_3_35`, either:

1. 8657492b We required the `returning_clauses_for_sqlite_3_35` feature for the `Jsonb` type as a whole. Which means we also require it for:
    - the whole `sqlite::types::json` mod use
    - the `type_impl::json` use of `crate::sql_types::Jsonb` and the `SerdeJsonValueProxy` type.
2. Or we de-factorize the doc example to completely separate the SQLite from the Pg code (but you asked me not to [here](https://github.com/diesel-rs/diesel/pull/4284#discussion_r1778595059)).

JMLX42:
> > for sqlite they likely require `#[cfg(all(feature = "sqlite", feature = "serde_json", feature = "returning_clauses_for_sqlite_3_35"))]`.
> 
> @weiznich it gets deeper: since the doc for the `Jsonb` type also needs the `returning_clauses_for_sqlite_3_35`, either:
> 
>     1. [8657492](https://github.com/diesel-rs/diesel/commit/8657492b88ca05d5fcd8cb0f1cb474fca40d26bb) We required the `returning_clauses_for_sqlite_3_35` feature for the `Jsonb` type as a whole. Which means we also require it for:
>        
>        * the whole `sqlite::types::json` mod use
>        * the `type_impl::json` use of `crate::sql_types::Jsonb` and the `SerdeJsonValueProxy` type.
> 
>     2. Or we de-factorize the doc example to completely separate the SQLite from the Pg code (but you asked me not to [here](https://github.com/diesel-rs/diesel/pull/4284#discussion_r1778595059)).

@weiznich Actually I think I found a good middle-ground. I'll let you know when the CI is green.

JMLX42:
@weiznich the CI is green except for this:

https://github.com/diesel-rs/diesel/actions/runs/11113502349/job/30877989999

`serde_json::Number` does not exist before version 0.9.0. Can I upgrade the dependency version requirement or should I find a workaround?

JMLX42:
> `serde_json::Number` does not exist before version 0.9.0. Can I upgrade the dependency version requirement or should I find a workaround?

@weiznich since:

- [`0.9.0` dates from 2017](https://github.com/serde-rs/json/releases/tag/v0.9.0)
- `diesel` relies on Rust edition 2021

I chose to upgrade the lower bound req to 0.9.0: e3b1f584a525612e2628228cf0369ff5a27723e0

JMLX42:
@weiznich CI is all green now. If you are OK with the current state of the PR, I will squash my changes, then rebase. Then you can merge them.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -1,10 +1,14 @@
 //! PostgreSQL specific expression methods
 
+#[cfg(all(feature = "postgres_backend", feature = "serde_json"))]
```
weiznich, line 3: None of these traits should be coupled to the `serde_json` feature. You should be able to use them without enabling that feature, as you could use them purely in the generated SQL.

That applies to essentially all changes in this file.

diesel/src/sql_types/mod.rs
```diff
@@ -395,11 +395,144 @@ pub struct Timestamp;
...
+/// #   feature = "serde_json",
+/// #   any(
+/// #       feature = "postgres_backend",
+/// #       all(feature = "sqlite", feature = "returning_clauses_for_sqlite_3_35"),
+/// #   )
+/// # )))]
+/// # fn main() {}
+/// ```
+#[cfg(all(
+    feature = "serde_json",
```
weiznich, line 528: This types should be independent from the `serde_json` feature. It's fine to gate the doc-test to that feature, but not the type.

weiznich (commented):
Thanks for the update. I'm sorry that I still have found something that needs to be changed, but the `Jsonb` type shouldn't be gated behind the `serde_json` feature. Only the actual `ToSql`/`FromSql` impls for `serde_json::Value` should be gated, as you can use columns/values of those type without actually loading them from the database, e.g by converting them to string or just comparing single fields.

weiznich (approved):
Thanks for fixing and thanks again for working on this :heart: 

I think the next step for json support in the SQLite backend is to add definitions for the [built-in SQL functions](https://sqlite.org/draft/json1.html) to diesel. Especially adding the `json` and `jsonb` functions should make it much easier to actually use this work. If you are interested in working on that as well, just reach out and I will provide more details on how this should be done. That part should be much more straight forward than adding these impls, as we don't need to care about parsing complicated stuff.

diesel/src/sqlite/types/json.rs
```diff
@@ -0,0 +1,1073 @@
+//! Support for JSON and JSONB values under PostgreSQL.
```
sofiaritz, line 1: nit - Should this comment be: `//! Support for JSON and JSONB values under SQLite.`?

Great work! Hoping this lands on the next diesel version :)

sofiaritz: > Can you submit a PR to fix that?

Done: #4307

JMLX42:
@weiznich hello there! I might be mistaking but my understanding is that this is a minor feature addition that does not break compatibility. If that's the case, is there any particular reason why is has not been released yet?

weiznich:
@JMLX42 I generally do not answer questions when I'm ponged like that. Otherwise this is a new feature and new features are part of the next feature (==minor release) after the PR has been merged. There haven't been such a release yet.

JMLX42:
> I generally do not answer questions when I'm pinged like that.

@weiznich thank you for taking the time.

> Otherwise this is a new feature and new features are part of the next feature (==minor release) after the PR has been merged. There haven't been such a release yet.

If I am not mistaken, there has been many minor releases since since has been merged last October.

PR #4286, opened 2024-09-27
working on json_path_exists

valkrypton:
@weiznich I was trying to implement the `jsonb_path_exists` function. One of its argument is of type `Jsonpath`. Currently there is no such type in diesel, so I tried to implement it using the `jsonpath_rust` crate. 

I implemented the `FromSql` trait for it, but I could not figure out how to implement the `ToSql` trait. I'd appreciate if you could take a look and help me with this or maybe i am going completely in the wrong direction with this?

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -2191,3 +2191,8 @@ define_sql_function! {
         values: Arr2
     ) -> Arr2::Out;
 }
+
+#[cfg(feature = "postgres_backend")]
+define_sql_function!{
+    fn jsonb_path_exists<J: JsonbOrNullableJsonb + SingleValue, >(jsonb:J,path:Jsonb);
+}
--- a/diesel/src/pg/types/json_path.rs
+++ b/diesel/src/pg/types/json_path.rs
@@ -1,0 +1,43 @@
+use std::fmt::{Debug, Display};
+use std::io::Write;
+use std::str;
+use std::str::FromStr;
+use crate::{deserialize, serialize, sql_types};
+use crate::deserialize::FromSql;
+use crate::pg::Pg;
+use crate::serialize::{Output, ToSql};
+
+#[cfg(all(feature = "postgres_backend", feature = "serde_json"))]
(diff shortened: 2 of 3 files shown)


weiznich:
Thanks for working on this. The `FromSql` implementation needs to read the binary representation of a json path as defined in the PostgreSQL protocol. Unfortunately the documentation of that format is mostly not existing, at least for anything that's not a trivial type. That means you likely need to perform some experimentation. One possibility would be to just setup a test case that loads a jsonpath value from the database and inspect the output. That should tell us how postgres serializes these data. Hopefully that's just the path as string or something simple like that. The other way is to dig through the postgres source code and find something there. [This file](https://github.com/postgres/postgres/blob/master/src/include/utils/jsonpath.h) might be a starting point.

diesel/src/pg/expression/functions.rs
```diff
@@ -2279,3 +2279,33 @@ define_sql_function! {
...
+    /// #
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::{jsonb_path_exists};
+    /// #     use diesel::sql_types::{Array, Nullable, Text,Jsonb, Jsonpath};
+    /// #     use serde_json::Value;
+    /// #     use jsonpath_rust::JsonPath;
+    /// #     let connection = &mut establish_connection();
+    /// let jsonb:Value = serde_json::json!({"a":[1,2,3,4,5]});
+    /// let json_path = jsonpath_rust::path!("$.a[ ? (@ >= 2 && @ <= 4)]");
+    /// let exists = jsonb_path_exists::<Jsonb,Jsonpath,_,_>(jsonb,json_path).get_result::<bool>(connection)?;
```
weiznich, line 2305: You want to write something like that here:

```suggestion
    /// let exists = diesel::select(jsonb_path_exists::<Jsonb,Jsonpath,_,_>(jsonb,json_path)).get_result::<bool>(connection)?;
```

The underlying issue is that just calling the sql function only creates a query fragment that only represents the function, not a full SELECT statement. By using `diesel::select` you explicitly construct a SELECT statement without from cause, while using whatever you provided as select clause.

PR #4290, opened 2024-09-29
Expose error code for PG backend

jrandolf:
This allows for robust handling of PG errors that are not already enumerated.

### DIFF
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -275,7 +275,7 @@ mod error_codes {
     pub(in crate::pg::connection) const SQLCLIENT_UNABLE_TO_ESTABLISH_SQLCONNECTION: &str = "08001";
     pub(in crate::pg::connection) const SQLSERVER_REJECTED_ESTABLISHMENT_OF_SQLCONNECTION: &str =
         "08004";
-    pub(in crate::pg::connection) const RESTRICT_VIOLATION: &str = "23501";
+    pub(in crate::pg::connection) const RESTRICT_VIOLATION: &str = "23001";
     pub(in crate::pg::connection) const NOT_NULL_VIOLATION: &str = "23502";
     pub(in crate::pg::connection) const FOREIGN_KEY_VIOLATION: &str = "23503";
     pub(in crate::pg::connection) const UNIQUE_VIOLATION: &str = "23505";
--- a/diesel/src/pg/connection/result.rs
+++ b/diesel/src/pg/connection/result.rs
@@ -73,6 +73,12 @@ impl PgResult {
                             DatabaseErrorKind::NotNullViolation
                         }
                         Some(error_codes::CHECK_VIOLATION) => DatabaseErrorKind::CheckViolation,
+                        Some(error_codes::RESTRICT_VIOLATION) => {
+                            DatabaseErrorKind::RestrictViolation
+                        }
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. Could you provide some more details about which SQL Error codes you want to handle outside of diesel and why that's required. 

Additionally: I fear we cannot merge the PR as it is as it contains a breaking change (adding a new required method to an public trait). 

Finally: There should be at least one test for this.

jrandolf:
I'd like to handle exclusion errors, but possibly others if they come up. I'd rather not file a PR for each case though, so it makes more sense IMO to just expose the code. After all, sql state is somewhat standardized AFAIU. Perhaps not the codes themselves, but the concept of the code itself.

I will add a test.

diesel/src/result.rs
```diff
@@ -188,6 +188,12 @@ pub trait DatabaseErrorInformation {
     /// An optional integer indicating an error cursor position as an index into
     /// the original statement string.
     fn statement_position(&self) -> Option<i32>;
+
+    /// An error code indicating the cause of the error.
+    ///
+    /// Currently this method will return `None` for all backends other than
+    /// PostgreSQL.
```
Ten0, line 195: maybe we want to add a link to here:
https://www.postgresql.org/docs/current/errcodes-appendix.html

to the doc?

Ten0:
> I'd like to handle exclusion errors, but possibly others if they come up. I'd rather not file a PR for each case though, so it makes more sense IMO to just expose the code

IMO adding those to [`DatabaseErrorKind`](https://github.com/jrandolf/diesel/blob/3af57c4a3411342136c5e683d208e01d91c16255/diesel/src/pg/connection/result.rs#L61) will result in a more readable and less error-prone code when matching on those errors, and in addition it wouldn't be a breaking change, as documented on `DatabaseErrorKind` (which is also marked as `#[non_exhaustive]`).
Adding the 2 or 3 "[Integrity Constraint Violations](https://www.postgresql.org/docs/current/errcodes-appendix.html)" that are still missing there would seem reasonable, and I would tend to imagine that that represents most of the "other cases that one may want to catch later on".

I'm not opposed to adding the method currently suggested by this PR but the above seems like a reasonable first step given the compromise that there is otherwise between backwards compatibility (previously mentioned breaking change) and correctness (any existing implementation of this trait is unlikely to be as correct as it should be if we add a default impl for this method).

jrandolf:
@Ten0 Alright, I've just added the remaining constraints to the enum. Do you folks want test for these errors or is this sufficient?

diesel/src/pg/connection/result.rs
```diff
@@ -269,10 +275,12 @@ mod error_codes {
     pub(in crate::pg::connection) const SQLCLIENT_UNABLE_TO_ESTABLISH_SQLCONNECTION: &str = "08001";
     pub(in crate::pg::connection) const SQLSERVER_REJECTED_ESTABLISHMENT_OF_SQLCONNECTION: &str =
         "08004";
+    pub(in crate::pg::connection) const RESTRICT_VIOLATION: &str = "23501";
```
Ten0, line 278: [pg doc](https://www.postgresql.org/docs/current/errcodes-appendix.html) says `23001`

Ten0 (approved):
LGTM apart from this

PR #4295, opened 2024-10-04
Fix deserialization of empty PostgreSQL ranges

mpflanzer:
In diesel empty ranges are represented as:

```
(
  Bound::Excluded(T::default()),
  Bound::Excluded(T::default()),
)
```

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -578,3 +578,26 @@ where
 {
     const FIELD_COUNT: usize = <ST as crate::util::TupleSize>::SIZE;
 }
+
+/// A helper trait for giving a type a useful default value.
+///
+/// This is needed for types that can be used as range to represent the empty range as
+/// (Bound::Excluded(DEFAULT), Bound::Excluded(DEFAULT)).
+/// When possible, implementations of this trait should fall back to using std::default::Default.
+#[allow(dead_code)]
+pub(crate) trait Defaultable {
+    /// Returns the "default value" for a type.
+    fn default_value() -> Self;
+}
+
+// We cannot have this impl because rustc
+// then complains in third party crates that
+// diesel may implement `Default`in the future.
+// If we get negative trait impls at some point in time
+// it should be possible to make this work.
+//// Defaultable for types that has Default
+//impl<T: Default> Defaultable for T {
+//    fn default_value() -> Self {
+//        T::default()
+//    }
+//}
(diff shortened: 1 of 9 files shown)


diesel/src/deserialize.rs
```diff
@@ -578,3 +578,25 @@ where
 {
     const FIELD_COUNT: usize = <ST as crate::util::TupleSize>::SIZE;
 }
+
+/// A helper trait for giving a type a useful default value.
+///
+/// This is needed for types that can be used as range to represent the empty range as
+/// (Bound::Excluded(DEFAULT), Bound::Excluded(DEFAULT)).
+/// When possible, implementations of this trait should fall back to using std::default::Default.
+pub trait Defaultable {
```
weiznich, line 587: We do not want to make this a public trait for now, as this is just a workaround. I would rather prefer to go the way to keep this trait private and have the wild card impl below instead. In that way other crates can still support it by implementing `Default`.

weiznich (commented):
Thanks for submitting this PR and sorry for taking that long to response. I was occupied with other work :disappointed: 

I think the implemented approach is mostly reasonable beside the fact that I do not want to add this new trait to the public API. I would rather prefer to keep it private.

diesel/src/deserialize.rs
mpflanzer: The wild card is what I tried first but couldn't get it working due to the conflicting implementations error, e.g. `"conflicting implementations of trait MyDefault for type OffsetDateTime ... upstream crates may add a new impl of trait std::default::Default for type OffsetDateTime in future versions"`. But indeed long term that would be nicer.

Is it ok if I just mark the `Defaultable` trait as `pub(crate)`? Or do you have another idea in mind?

weiznich: It's fine to just use `pub(crate)` then. I read the comment as: This is not possible because it will cause errors in crates depending on diesel and not as this is not possible because in causes errors in diesel itself.

mpflanzer: Sorry for the confusion. Reading it again it's indeed a bit misleading. I just copied it from here: https://github.com/diesel-rs/diesel/blob/a3ea5242567b1a44f3b606d9b552bdd23065e001/diesel/src/deserialize.rs#L553

I updated the MR to have the trait private to the crate.

weiznich (approved):

diesel/src/deserialize.rs
weiznich: Thanks, looks good now

PR #4303, opened 2024-10-10
Add the row_to_json Method for diesel postgres

PratikFandade:
Adding the support as specified in #4216 

Not sure if everything correct, still new to rust, let me know if you need any other changes.

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -3643,8 +3643,7 @@ pub(in crate::pg) mod private {
     impl TextArrayOrNullableTextArray for Nullable<Array<Nullable<Text>>> {}
 
     #[diagnostic::on_unimplemented(
-        message = "`{Self}` is neither `Record<Text>`, `Record<Nullable<Text>>`,\
-                   `Nullable<Record<Text>>` nor `diesel::sql_types::Nullable<Record<Nullable<Text>>>`",
+        message = "`{Self}` is neither `Record<T>` nor `Nullable<Record<T>>`",
         note = "try to provide an expression that produces one of the expected sql types"
     )]
     pub trait RecordOrNullableRecord {}
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -4,7 +4,8 @@ pub(in crate::pg) use self::private::{
     ArrayOrNullableArray, CombinedNullableValue, InetOrCidr, JsonIndex, JsonOrNullableJson,
     JsonOrNullableJsonOrJsonbOrNullableJsonb, JsonRemoveIndex, JsonbOrNullableJsonb,
     MaybeNullableValue, MultirangeOrNullableMultirange, MultirangeOrRangeMaybeNullable,
-    RangeOrMultirange, RangeOrNullableRange, TextArrayOrNullableTextArray, TextOrNullableText,
+    RangeOrMultirange, RangeOrNullableRange, RecordOrNullableRecord, TextArrayOrNullableTextArray,
+    TextOrNullableText,
 };
 use super::date_and_time::{AtTimeZone, DateTimeLike};
 use super::operators::*;
@@ -3302,7 +3303,7 @@ where
 pub(in crate::pg) mod private {
     use crate::sql_types::{
         Array, Binary, Cidr, Inet, Integer, Json, Jsonb, MaybeNullableType, Multirange, Nullable,
-        OneIsNullable, Range, SingleValue, SqlType, Text,
+        OneIsNullable, Range, Record, SingleValue, SqlType, Text,
     };
     use crate::{Expression, IntoSql};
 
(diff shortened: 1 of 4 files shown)


diesel/src/pg/expression/expression_methods.rs
```diff
@@ -3749,4 +3750,16 @@ pub(in crate::pg) mod private {
...
+        message = "`{Self}` is neither `Record<Text>`, `Record<Nullable<Text>>`,\
+                   `Nullable<Record<Text>>` nor `diesel::sql_types::Nullable<Record<Nullable<Text>>>`",
+        note = "try to provide an expression that produces one of the expected sql types"
+    )]
+    pub trait RecordOrNullableRecord {}
+
+    impl RecordOrNullableRecord for Record<Text> {}
+    impl RecordOrNullableRecord for Record<Nullable<Text>> {}
+    impl RecordOrNullableRecord for Nullable<Record<Text>> {}
+    impl RecordOrNullableRecord for Nullable<Record<Nullable<Text>>> {}
```
weiznich, line 3764: `Record` is a somewhat especial type, in so far that it always accepts a tuple as generic argument. So something like `Record<(Integer, Text)>` or `Record<(Text, Date, Integer)>`. 

The right variant here would be to write something like `impl<T> RecordOrNullableRecord for Record<T> {}` and `impl<T> RecordOrNullableRecord for Nullable<Record<T>> {}`

diesel/src/pg/expression/functions.rs
```diff
@@ -2191,3 +2192,53 @@ define_sql_function! {
...
+    ///     });
+    ///
+    ///     // Asserting the result
+    ///     assert_eq!(expected, json_value);
+    ///
+    ///     Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "jsonb_object"]
+    fn row_to_json<R: RecordOrNullableRecord + SingleValue + CombinedNullableValue<Record<Text>, Jsonb>>(record: R) -> R::Out;
```
weiznich, line 2243: I fear this function is somewhat problematic to implement with the current diesel types setup. It doesn't accept anonymous record types (as represented by the diesel `Record` type), but only rows with names. The later is something that's not yet expressible with diesels type setup. 

Given that I suppose this won't work in that way :(

(I also do not have any real suggestion how to make it easily work)

diesel_derives/tests/auto_type.rs
```diff
@@ -52,7 +52,8 @@ table! {
         multirange -> Multirange<Integer>,
         timestamptz -> Timestamptz,
         name -> Text,
-        text_array -> Array<Text>
+        text_array -> Array<Text>,
+        record -> Record,
```
weiznich, line 56: We would need to define the record type here as `Record<(Integer, Text)>` or something like that to fix the compiler error.

weiznich (changes requested):
Thanks for working on this.

I fear I have bad news: The function seems to be impossible to implement with the current diesel type setup. Sorry for that :(

diesel/src/pg/expression/expression_methods.rs
PratikFandade: Thanks @weiznich , I'll make these changes

diesel_derives/tests/auto_type.rs
PratikFandade: Sure, I'll make the changes

diesel/src/pg/expression/functions.rs
PratikFandade: I'll modify the function as pointed by @meeshal to make it correct

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -3749,4 +3750,14 @@ pub(in crate::pg) mod private {
     impl TextArrayOrNullableTextArray for Array<Nullable<Text>> {}
     impl TextArrayOrNullableTextArray for Nullable<Array<Text>> {}
     impl TextArrayOrNullableTextArray for Nullable<Array<Nullable<Text>>> {}
+
+    #[diagnostic::on_unimplemented(
+        message = "`{Self}` is neither `Record<Text>`, `Record<Nullable<Text>>`,\
+                   `Nullable<Record<Text>>` nor `diesel::sql_types::Nullable<Record<Nullable<Text>>>`",
+        note = "try to provide an expression that produces one of the expected sql types"
+    )]
```
Mingun, line 3758: The reference to `Text` fragments here is outdated

diesel/src/pg/expression/functions.rs
```diff
@@ -2234,3 +2235,52 @@ define_sql_function! {
...
+    /// #
+    /// # #[cfg(feature = "serde_json")]
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::sql;
+    /// #     use diesel::sql_types::{Record, Text, Integer};
+    /// #     use serde_json::Value;
+    /// #     let connection = &mut establish_connection();
+    ///
+    ///     // Example SQL query selecting a custom record type
+    ///     let query = diesel::select(sql::<Record<(Text, Integer)>>("ROW('John', 30)::record"));
```
meeshal, line 2261: Imho, `row_to_json` should be used here instead of `sql` so, the returned value could be casted to json right away.

diesel/src/pg/expression/expression_methods.rs
PratikFandade: Opps! My bad, guess I missed to unstash some changes, I'll fix this.

diesel/src/pg/expression/functions.rs
PratikFandade: I'm so sorry, will fix this test

diesel/src/pg/expression/functions.rs
```diff
@@ -2234,3 +2235,52 @@ define_sql_function! {
...
+    ///     });
+    ///
+    ///     // Asserting the result
+    ///     assert_eq!(expected, json_value);
+    ///
+    ///     Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "row_to_json"]
+    fn row_to_json<R: RecordOrNullableRecord + SingleValue + MaybeNullableValue<Json>>(record: R) -> R::Out;
```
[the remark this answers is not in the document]
PratikFandade: Thanks for the insight will make these changes

weiznich (approved):

PR #4306, opened 2024-10-14
add Record support for 1 element tuples

meeshal:
* One element tuples are not interpreted correctly and when they are run against psql they are treated as a single value. In this case using ROW() clearly indicates that 1 element tuple should be interpreted as a row.
* Supporting 1 element tuples is needed in #4216 for implementing row_to_json


weiznich (approved):
Thanks for opening this PR :heart: 

This looks reasonable for me and would likely resolve the issue I've seen in the other PR.

PR #4307, opened 2024-10-14
Fix comment related to JSON and JSONB support under SQLite

sofiaritz:
As mentioned in #4284


weiznich (approved):
Thanks for submitting this fix:pray:

PR #4314, opened 2024-10-19
add is_json and is_not_json expression methods

meeshal:
This PR adds support for is_json and is_not_json expression methods mentioned in #4216


weiznich (approved):
Thanks, this looks good :+1:

PR #4321, opened 2024-10-24
fix: connect to template1 when database is postgres

### LINKED ISSUE #4275: `diesel database reset` errors with `Failed to execute a database query`
### Versions

- **Rust:** `rustc 1.83.0-nightly (506f22b46 2024-09-19)`
- **Diesel:** `2.2.4`
- **Database:** `postgres`
- **Operating System:** Linux

### Feature Flags

- **diesel:** `["postgres"]`

Run `diesel database reset`.

Database is successfully reset.

```
Dropping database: postgres
Failed to execute a database query: cannot drop the currently open database
```

No.

### Steps to reproduce

`.env`
```bash
DATABASE_URL="postgresql://postgres:postgres@localhost:5432/postgres"
```

`diesel.toml`
```toml
[print_schema]
file = "src/schema.rs"
custom_type_derives = ["diesel::query_builder::QueryId", "Clone"]

[migrations_directory]
dir = "migrations"
```

I only have one migration file:

`up.sql`
```sql
CREATE TABLE users (
    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    name TEXT NOT NULL,
);
```
`down.sql`
```sql
DROP TABLE IF EXISTS users;
```

## Checklist

zaira-bibi:
Closes #4275.

**Description**
Diesel explicitly connected to `postgres` to drop any other database, which posed issues when `postgres` itself was to be dropped. 

**Changes**
Added a check for the current database, and connect to `template1` for the case when current database is `postgres`.


diesel_cli/src/database.rs
```diff
@@ -429,6 +429,11 @@ fn change_database_of_url(
         .last()
         .expect("The database url has at least one path segment")
         .to_owned();
+
+    if database.eq("postgres") {
```
Ten0, line 433: I've read the issue and I understand it may be sensible to do this specifically when we are trying to drop a database.
It also sounds likely that at the moment this is the only place at which this function is called.
However the naming of this function  `fn change_database_of_url(` would not let any future code author expect this behavior, so it's likely to introduce a bug.
I think it would be best to put this logic where the function is called, that is in a place where the current code path is about removing the database.

zaira-bibi: Thanks for the review! I get what you mean, will push a commit with the required change.

weiznich (approved):
Thanks for fixing this :heart:

PR #4322, opened 2024-10-25
First draft for supporting window functions and other aggregate function expressions

weiznich:
See
[here](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS) and
[here](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES) for the relevant postgresql documentation

This PR implements mainly the relevant DSL structure for this types. As of today this should support most of the relevant API surface.

This PR is not finished yet, I mainly push this now to get some early feedback on the exposed user API. See the two dummy tests for how to use the new functionality.

What's still missing:

- [ ] Support for other backends
- [ ] Tests, a lot of tests
- [ ] Compile tests to verify that we do not generate invalid SQL
- [ ] Documentation

cc @oeed


oeed:
Looks nice to me! Only observations from my end:

- I see you added a `distinct()` function, e.g. `dsl::count(users::id).distinct()`, is that functionally the same as `count_distinct()`?
- The naming of `filter_aggregate()` is a little more verbose (and potentially a little less discoverable) than `filter()`, I assume that was primarily done to prevent confusion with query filtering?

I haven't needed windowing functions much in the past, so don't have much to add on those.

weiznich:
> I see you added a distinct() function, e.g. dsl::count(users::id).distinct(), is that functionally the same as count_distinct()?

Yes that's the same for `count()`, it just generalizes that to any aggregate function (as allowed by postgres). I plan to deprecate `count_distinct()` as soon as that is ready.

> The naming of filter_aggregate() is a little more verbose (and potentially a little less discoverable) than filter(), I assume that was primarily done to prevent confusion with query filtering?

Rustc had issues resolving the right `filter()` method, so I just went with a different name. I think we should be able to improve the discoverability by adding a `#[doc(alias = "filter")]` there so that it shows up whn you search for `filter` in the API docs?

weiznich:
I pushed an update that brings most of the infrastructure for in place for supporting several backends. This was rather tricky as different backends have different restrictions on which functions are allowed as window functions. It seems like mysql doesn't allow all aggregate functions as window functions for example. For oracle it's even worse, there specific functions cannot be used with frame clauses but work otherwise fine as window functions… For that reason I introduced an attribute  based way to restrict a window function to a specific backend, backend bound or sql dialect. That should be rather flexible and allow us to go forward by just putting the right bounds there.

One rather important thing that still needs to be solved is restricting window functions to be used in order and select clauses only. Otherwise most of the remaining work is:

+ Writing documentation
+ Writing tests
+ Writing compile fail tests
+ Marking relevant existing functions as window functions
+ Adding new relevant window only functions
(+ Figuring out how to deal with these `WITHIN GROUP` postgres specific functions)

diesel_compile_tests/tests/fail/reject_window_functions_in_where_clause.rs
```diff
@@ -0,0 +1,20 @@
...
+        id -> Integer,
+        name -> VarChar,
+    }
+}
+
+fn main() {
+    let mut connection = PgConnection::establish("").unwrap();
+
+    let res = users::table
+        .filter(dsl::count(users::id).over().lt(53))
```
weiznich, line 16: I'm still looking for a good solution to reject this. It's currently accepted as I did not find a good way yet to restrict the bounds on `FilterDsl::filter()` without introducing a breaking change yet.

https://github.com/diesel-rs/diesel/blob/0da4c0599153d08f125893521492a87e2a73b94e/diesel/src/query_builder/select_statement/dsl_impls.rs#L195-L217

Both `Expression` and `NonAggregate` must be usable in a trait object context so we cannot just add an associated const or a something like that there. We also cannot add another associated type there as there is no stable way to define default values for such a type (also it would break existing trait objects?).

We could add a new trait bound there as users cannot express the `WhereAnd` bound anyway (that trait is private). Unfortunately I do not see a way to express: No existing impl is a window function beside those explicitly opting into being one as that still would require us to relay essentially one either `Expression` or `NonAggregate` to allow one or the other as wild card trait bound for the "new" trait and window function also need to implement both traits, so you would run into conflicting impls there.

weiznich: cc @diesel-rs/contributors for ideas.

weiznich: I rebased this on top of #4655 and added https://github.com/diesel-rs/diesel/pull/4322/commits/1411136d9029474d9c6de04247fd9d41bde29b0d to perform a const eval check for rejecting window functions in most cases (== if nothing is boxed) for where clauses.

weiznich:
@diesel-rs/reviewers This PR is now ready to be merged. I'm looking forward to your review here.

PR #4324, opened 2024-10-27
add support for json expression methods

meeshal:
This PR adds support for json expression methods mentioned in #4216 
- is_json_object
- is_not_json_object
- is_json_array
- is_not_json_array
- is_json_scalar
- is_not_json_scalar


meeshal:
@guissalustiano , I moved json expression methods under PgTextExpressionMethods as you suggested. When it comes to `SELECT NULL  ...` I think that is how psql works and null will always will be returned, it applies to other methods as well. I added a test in `is_json_scalar`.

guissalustiano (approved):
Thanks!

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -810,6 +734,326 @@ pub trait PgTextExpressionMethods: Expression + Sized {
...
+    /// assert_eq!(res, true);
+    /// let res = diesel::select(("abc".into_sql::<Text>().is_json_scalar())).get_result::<bool>(conn)?;
+    /// assert_eq!(res, false);
+    /// let res = diesel::select(("{\"products\": [1,2,3]}".into_sql::<Text>().is_json_scalar())).get_result::<bool>(conn)?;
+    /// assert_eq!(res, false);
+    /// let res = diesel::select(("[1,2,3]".into_sql::<Text>().is_json_scalar())).get_result::<bool>(conn)?;
+    /// assert_eq!(res, false);
+    /// let res = diesel::select(("\"abc\"".into_sql::<Text>().is_json_scalar())).get_result::<bool>(conn)?;
+    /// assert_eq!(res, true);
+    /// let res = diesel::select(sql::<Nullable<Text>>("NULL").is_json_scalar().nullable()).get_result::<Option<bool>>(conn)?;
```
weiznich, line 1006: I'm sorry to bring this up now, but this is an issue. In this case `is_json_scalar()` should evaluate to `Nullable<Bool>` instead. It's actually a short coming of the `postfix_operator!()` macro that I forgot about. The `diesel::infix_operator!()` already handles this internally: https://github.com/diesel-rs/diesel/blob/2090344bf4b7259a262ea80311ba2c33b7e66c12/diesel/src/expression/operators.rs#L296-L330

Something like this is also needed for the `postfix_operator!()` macro, although:

* We need to make sure that's not the new default, as that would be a breaking chaneg
* It's likely simpler as we just have one argument instead of two.

We likely just need to introduce an explict opt-in branch to this macro that sets the return type to 

```
$crate::sql_types::is_nullable::MaybeNullable<<T as Expression>::SqlType>
```
after that this test should just compile as

```
diesel::select(sql::<Nullable<Text>>("NULL").is_json_scalar()).get_result::<Option<bool>>(conn)?;
```

weiznich (changes requested):
I'm sorry to took so long to add this here, but the nullability handling is still not correct. I've added a more detailed comment in the relevant location explaining what needs to change there. If you need more help to fix this, just let me know. I can also try to change the macro myself, but that won't happen before next week.

diesel/src/pg/expression/expression_methods.rs
meeshal: I am not an expert when it comes to macros but I am thinking about sth like `postfix_operator!(IsJsonScalar, " IS JSON SCALAR", ConditionalNullability Bool, backend: Pg);` to make it backward compatible. @weiznich , what do you think ?

weiznich: That would be a fine solution :+1:

meeshal: @weiznich , I added ConditionalNullability support to postfix_operator. Please let me know if it makes sense.

PR #4325, opened 2024-10-27
Adding Support for `json_populate_record` and `jsonb_populate_record`

prkbuilds:
Using the traits already defined and have written the test_cases with my understanding of the functions, but open to suggestions and learnings!

### DIFF
--- a/diesel/src/pg/expression/expression_methods.rs
+++ b/diesel/src/pg/expression/expression_methods.rs
@@ -1,11 +1,11 @@
 //! PostgreSQL specific expression methods
 
 pub(in crate::pg) use self::private::{
-    ArrayOrNullableArray, CombinedNullableValue, InetOrCidr, JsonIndex, JsonOrNullableJson,
-    JsonOrNullableJsonOrJsonbOrNullableJsonb, JsonRemoveIndex, JsonbOrNullableJsonb,
-    MaybeNullableValue, MultirangeOrNullableMultirange, MultirangeOrRangeMaybeNullable,
-    RangeOrMultirange, RangeOrNullableRange, RecordOrNullableRecord, TextArrayOrNullableTextArray,
-    TextOrNullableText,
+    ArrayOrNullableArray, CombinedAllNullableValue, CombinedNullableValue, InetOrCidr, JsonIndex,
+    JsonOrNullableJson, JsonOrNullableJsonOrJsonbOrNullableJsonb, JsonRemoveIndex,
+    JsonbOrNullableJsonb, MaybeNullableValue, MultirangeOrNullableMultirange,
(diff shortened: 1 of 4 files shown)


PratikFandade:
just for context the PR is for the issue #4216

weiznich (changes requested):
I'm sorry to take that long to review this PR and I'm also sorry to bring up an issue, but it seems like the nullable bounds are not correct yet. I've added a more detailed comment in the specific location on how this can be fixed.

diesel/src/pg/expression/functions.rs
```diff
@@ -2279,3 +2279,121 @@ define_sql_function! {
...
+    /// assert_eq!(record, ("John".to_string(), 42));
+    ///
+    /// #    Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "jsonb_populate_record"]
+    fn jsonb_populate_record<
+        B: RecordOrNullableRecord + SingleValue,
+        J: JsonbOrNullableJsonb + CombinedNullableValue<Jsonb, B>
+    >(base: B, from_json: J) -> J::Out;
```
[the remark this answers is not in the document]
PratikFandade: Thanks @weiznich ! This is great to know and really informative, I totally understand what needs to be done, I'll implement the required change are add the changes. and thanks for telling a good way to test the nullability, will definitely keep that in mind next time.

PratikFandade:
I'm not sure if the naming convention is the best for the new Nullable Trait implementation. I think this will take care of Nullablility, but let me know if there is anything I might have missed.

diesel/src/pg/expression/expression_methods.rs
```diff
@@ -3726,4 +3726,19 @@ pub(in crate::pg) mod private {
 
     impl<T> RecordOrNullableRecord for Record<T> {}
     impl<T> RecordOrNullableRecord for Nullable<Record<T>> {}
+
+    pub trait CombinedOneNullableValue<O, Out>: SingleValue {
```
weiznich, line 3730: Can we name this `CombinedAllNullableValue` or something like that instead to transport that this only evaluates to `Nullable` if all inputs are `Nullable`?

weiznich (commented):
Thanks for the update. This looks good beside the name of the new trait.

diesel/src/pg/expression/expression_methods.rs
PratikFandade: Hi @weiznich , sure I will change the name accordingly. And yes they allarenullable only returns nullable value when both values are null, completing the logic we need for this trait.

weiznich (approved):
Thanks for the update

PR #4343, opened 2024-11-13
Prepare a diesel 2.2.5 release

weiznich:
This includes the following backports:

* #4101 (the crates.io team asked for that)
* #4342 (The spam problem become more serve in the last days)

I would like to issue a new release as soon as possible to update the docs.


Ten0 (approved):
LGTM :)

We probably also want to add #4345 though :)

(Not sure why CI doesn't pass)

PR #4345, opened 2024-11-15
Fix and/or function argument to compile check bool expression

E-anae:
Currently if you try putting any sort of expression in a and function of diesel it will not be detected at compile time but will raise an error at runtime

### DIFF
--- a/diesel/src/expression_methods/bool_expression_methods.rs
+++ b/diesel/src/expression_methods/bool_expression_methods.rs
@@ -92,6 +92,7 @@ pub trait BoolExpressionMethods: Expression + Sized {
         Self::SqlType: SqlType,
         ST: SqlType + TypedExpressionType,
         T: AsExpression<ST>,
+        <T::Expression as Expression>::SqlType: BoolOrNullableBool,
         Or<Self, T::Expression>: Expression,
     {
         Grouped(Or::new(self, other.as_expression()))
--- a/diesel/src/expression_methods/bool_expression_methods.rs
+++ b/diesel/src/expression_methods/bool_expression_methods.rs
@@ -43,6 +43,7 @@ pub trait BoolExpressionMethods: Expression + Sized {
         Self::SqlType: SqlType,
         ST: SqlType + TypedExpressionType,
         T: AsExpression<ST>,
+        <T::Expression as Expression>::SqlType: BoolOrNullableBool,
         And<Self, T::Expression>: Expression,
     {
         Grouped(And::new(self, other.as_expression()))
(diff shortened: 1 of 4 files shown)


diesel_compile_tests/tests/fail/numeric_ops_require_numeric_column.stderr
```diff
@@ -13,3 +13,6 @@ note: an implementation of `std::ops::Add` might be missing for `columns::name`
    |         ^^^^ must implement `std::ops::Add`
 note: the trait `std::ops::Add` must be implemented
   --> $RUST/core/src/ops/arith.rs
+   |
+   | pub trait Add<Rhs = Self> {
```
Ten0, line 17: ah it looks like this may be due to differing versions of Rust between CI and version that was used to generate the snapshots

diesel/src/expression_methods/bool_expression_methods.rs
```diff
@@ -43,6 +43,7 @@ pub trait BoolExpressionMethods: Expression + Sized {
         Self::SqlType: SqlType,
         ST: SqlType + TypedExpressionType,
         T: AsExpression<ST>,
+        <T::Expression as Expression>::SqlType: BoolOrNullableBool,
```
Ten0, line 46: It seems that this can be written as `ST: BoolOrNullableBool` given the other bounds

Ten0 (approved):

weiznich:
@Ten0 I will add it to the backport PR as soon as I'm to work.

Roardom:
Hi, maybe I'm misunderstanding, but it looks like this restriction is placed on all database backends. The query provided in the test works fine for mysql on 2.2.4. Maybe it should only restrict postgres instead?

weiznich:
@Roardom Thanks for raising that concern. You are right that this kind of query is allowed with mysql, it seems to be just another syntax for `column IS NOT NULL`. As there is currently no easy way to place this restriction as backend specific restriction I would like to move forward anyway as that's a rather critical bugfix for the postgresql/sqlite backend. I'm open to other opinions here, but I would like to ask for proposed solutions then.

Roardom:
While not a complete solution, it should at least be highlighted in the release notes as a breaking change for users supporting mysql backends. I saw this change and immediately thought of similar queries in my code. But I double checked, and I only use it inside `SUM()`, such as `SUM(NOT active AND visible)` which already isn't permitted in diesel (since boolean doesn't implement foldable requiring raw queries).

> it seems to be just another syntax for `column IS NOT NULL`

I believe `column IS NOT NULL AND column != 0 AND column != ''` is more accurate (`0` and `''` would depend on the type). There might be other variations as well.

PR #4350, opened 2024-11-19
Fixes #4349

### LINKED ISSUE #4349: `.eq_any([["foo"]])` on `SqlType = Array<Text>` causes runtime error on Postgres
### Versions

- **Diesel:** ce6fa0bd93877391665254d9bdb23a9c895902ef
- **Database:** Postgres

### Feature Flags

- **diesel:** postgres

## Problem Description
`array.eq_any(array_of_array)` doesn't work on Postgres, because Postgres doesn't have such a thing as "array of array". It has 2D array but ["2-D array is a different concept"](https://pgsql-general.postgresql.narkive.com/eJuaaIgQ/could-not-find-array-type-for-data-type-character-varying#post2).
This causes an error, so we should probably use the non-specialized ANSI `IN (...)` when dealing with arrays.

```rust
table! {
    posts (id) {
        id -> Int4,
        // ...
        tags -> Array<Text>,
    }
}

#[test]
#[cfg(feature = "postgres")]
fn filter_array_by_in() {
    use crate::schema::posts::dsl::*;

    let connection: &mut PgConnection = &mut connection();
    let tag_combinations_to_look_for: &[&[&str]] = &[&["foo"], &["foo", "bar"], &["baz"]];
    let result: Vec<i32> = posts
        .filter(tags.eq_any(tag_combinations_to_look_for))
        .select(id)
        .load(connection)
        .unwrap();
    assert_eq!(result, &[] as &[i32]);
}
```

```
called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "could not find array type for data type text[]")
```

Either of:
- test passes, serializing as `IN ($1, $2, $3)` with `$1 = ["foo"]`...
- Doesn't compile, requiring an explicit writing (e.g. `.eq_any(`abc`).ansi()`) to tell it to use ANSI serialization

ATM AFAICT there's no way to even tell it to use ANSI serialization because of bounds here:
https://github.com/diesel-rs/diesel/blob/ce6fa0bd93877391665254d9bdb23a9c895902ef/diesel/src/expression/array_comparison.rs#L100-L103
which forbids the use of the ANSI implementation on `Pg` `Backend` (which seems weird since ANSI should always work, no?)

(NB: All of this is useful only to the extent to which indexing will work and this behaves reasonably even if there are 500 different values in there. I recall that we did tend to prefer `= ANY` maybe because of this but I may be mistaken here, was it only for statement caching ?)

### Steps to reproduce
https://github.com/Ten0/diesel/commit/d6c93fd2a8af99df9bc8dc362144e6f599c6030c

or dedicated working branch to fixing this:
```
cd diesel
git remote add Ten0 https://github.com/Ten0/diesel.git
git checkout array_eq_any
cd diesel_tests
DATABASE_URL="..." cargo test --no-default-features --features "postgres" -- filter_array_by_in --nocapture
```

weiznich:
This commit fixes an issue that allowed to trigger a runtime error by
passing an array of arrays to `.eq_any()`. We rewrite queries containing
`IN` expressions to `= ANY()` on postgresql as that more
efficient (allows caching + allows binding all values at once). That's
not possible for arrays of arrays as we do not support nested arrays
yet.

The fix introduces an associated constant for the `SqlType` trait that
tracks if a SQL type is a `Array<T>` or not. This information is then
used to conditionally generate the "right" SQL. By defaulting to `false`
while adding this constant we do not break existing code. We use the
derive to set the right value based on the type name.

cc @Ten0 

(I would suggest to include that PR in the patch release as well)

### DIFF
--- a/diesel/src/pg/query_builder/query_fragment_impls.rs
+++ b/diesel/src/pg/query_builder/query_fragment_impls.rs
@@ -66,10 +66,14 @@ where
     U: QueryFragment<Pg> + InExpression,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
-        self.left.walk_ast(out.reborrow())?;
-        out.push_sql(" = ANY(");
-        self.values.walk_ast(out.reborrow())?;
-        out.push_sql(")");
+        if self.values.is_array() {
+            self.walk_ansi_ast(out)?;
+        } else {
+            self.left.walk_ast(out.reborrow())?;
+            out.push_sql(" = ANY(");
+            self.values.walk_ast(out.reborrow())?;
+            out.push_sql(")");
+        }
         Ok(())
     }
 }
@@ -80,10 +84,14 @@ where
     U: QueryFragment<Pg> + InExpression,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
-        self.left.walk_ast(out.reborrow())?;
-        out.push_sql(" != ALL(");
-        self.values.walk_ast(out.reborrow())?;
-        out.push_sql(")");
+        if self.values.is_array() {
+            self.walk_ansi_ast(out)?;
+        } else {
+            self.left.walk_ast(out.reborrow())?;
+            out.push_sql(" != ALL(");
+            self.values.walk_ast(out.reborrow())?;
+            out.push_sql(")");
+        }
         Ok(())
     }
 }
@@ -92,10 +100,15 @@ impl<ST, I> QueryFragment<Pg, PgStyleArrayComparison> for Many<ST, I>
 where
     ST: SingleValue,
     Vec<I>: ToSql<Array<ST>, Pg>,
+    I: ToSql<ST, Pg>,
     Pg: HasSqlType<ST>,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
-        out.push_bind_param::<Array<ST>, Vec<I>>(&self.values)
+        if ST::IS_ARRAY {
+            self.walk_ansi_ast(out)
+        } else {
+            out.push_bind_param::<Array<ST>, Vec<I>>(&self.values)
+        }
     }
 }
 
--- a/diesel_tests/tests/filter_operators.rs
+++ b/diesel_tests/tests/filter_operators.rs
@@ -338,6 +338,21 @@ fn filter_array_by_in() {
     assert_eq!(result, &[] as &[i32]);
 }
 
+#[test]
+#[cfg(feature = "postgres")]
+fn filter_array_by_not_in() {
+    use crate::schema::posts::dsl::*;
+
+    let connection: &mut PgConnection = &mut connection();
+    let tag_combinations_to_look_for: &[&[&str]] = &[&["foo"], &["foo", "bar"], &["baz"]];
+    let result: Vec<i32> = posts
+        .filter(tags.ne_all(tag_combinations_to_look_for))
+        .select(id)
+        .load(connection)
+        .unwrap();
+    assert_eq!(result, &[] as &[i32]);
+}
+
 fn connection_with_3_users() -> TestConnection {
     let mut connection = connection_with_sean_and_tess_in_users_table();
     diesel::sql_query("INSERT INTO users (id, name) VALUES (3, 'Jim')")
--- a/diesel_tests/tests/filter_operators.rs
+++ b/diesel_tests/tests/filter_operators.rs
@@ -323,6 +323,21 @@ fn filter_by_in_explicit_array() {
     );
 }
 
+#[test]
+#[cfg(feature = "postgres")]
+fn filter_array_by_in() {
+    use crate::schema::posts::dsl::*;
+
+    let connection: &mut PgConnection = &mut connection();
+    let tag_combinations_to_look_for: &[&[&str]] = &[&["foo"], &["foo", "bar"], &["baz"]];
+    let result: Vec<i32> = posts
+        .filter(tags.eq_any(tag_combinations_to_look_for))
+        .select(id)
+        .load(connection)
+        .unwrap();
+    assert_eq!(result, &[] as &[i32]);
+}
+
 fn connection_with_3_users() -> TestConnection {
     let mut connection = connection_with_sean_and_tess_in_users_table();
     diesel::sql_query("INSERT INTO users (id, name) VALUES (3, 'Jim')")
--- a/diesel/src/expression/array_comparison.rs
+++ b/diesel/src/expression/array_comparison.rs
@@ -61,11 +61,45 @@ impl<T, U> In<T, U> {
     pub(crate) fn new(left: T, values: U) -> Self {
         In { left, values }
     }
+
+    pub(crate) fn walk_ansi_ast<'b, DB>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()>
+    where
+        DB: Backend,
+        T: QueryFragment<DB>,
+        U: QueryFragment<DB> + InExpression,
+    {
+        if self.values.is_empty() {
+            out.push_sql("1=0");
+        } else {
+            self.left.walk_ast(out.reborrow())?;
+            out.push_sql(" IN (");
+            self.values.walk_ast(out.reborrow())?;
+            out.push_sql(")");
+        }
+        Ok(())
+    }
 }
 
 impl<T, U> NotIn<T, U> {
     pub(crate) fn new(left: T, values: U) -> Self {
         NotIn { left, values }
+    }
+
+    pub(crate) fn walk_ansi_ast<'b, DB>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()>
+    where
+        DB: Backend,
+        T: QueryFragment<DB>,
+        U: QueryFragment<DB> + InExpression,
+    {
+        if self.values.is_empty() {
+            out.push_sql("1=1");
+        } else {
+            self.left.walk_ast(out.reborrow())?;
+            out.push_sql(" NOT IN (");
+            self.values.walk_ast(out.reborrow())?;
+            out.push_sql(")");
+        }
+        Ok(())
     }
 }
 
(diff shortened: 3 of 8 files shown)


Ten0:
So I found this comment:
https://dba.stackexchange.com/a/335502/196014

Which suggests that `= ANY` may actually be not-that-dumb:
> When the length of the array is very large, the other queries simply do a parallel seq scan. This is very fast, because the "Filter where id=ANY(...)" isn't dumb, it uses some kind of fast search like hashing or bisect, it doesn't compare every row with every value of the array.

So it seems that `= ANY ( ... )` may sometimes be more performant than `IN` 

But my experience was contrary to this (it was in fact, dumb, and I would see n² CPU consumption profiles).

So I've performed new tests (see code below), and I can find:
- Sometimes clear stupid bitmap heap scan with O(n) recheck on each line so a total O(n²) query
- Scenarios where it seems to not do n², specifically when no hash index exists

Maybe I should send a message to the PG mailing list pointing that as a bug because the two recheck-on-array scenarios are very inconsistent...

So I don't know what to think... If ANY is supposed to be smart but `IN` isn't when we pass in an array, maybe we want to do the "good worse case" of `VALUES` to not have surprising perf loss for users who don't expect any significant difference of perf between `eq_any(array_of_values)` and `eq_any(array_of_values_that_happen_to_also_be_arrays)`.
But if ANY is known-sometimes-dumb and from the PG perspective it's user's responsibility to use `VALUES` instead of `= ANY(ARRAY)` when relevant, then as stated above the more consistent approach of using `IN` seems to win.

Tests:
```sql
BEGIN;
DROP TABLE IF EXISTS tmp_benchmark;
CREATE TABLE tmp_benchmark(
    id Int8 NOT NULL GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    value text NOT NULL
);
-- Insert 10M values
INSERT INTO tmp_benchmark(value) SELECT DISTINCT (floor(random() * 1000000000000)::int8)::text FROM generate_series(1,1000000);

ALTER TABLE tmp_benchmark ADD CONSTRAINT tmp_benchmark_unique_value_and_corresponding_index UNIQUE(value);

CREATE INDEX tmp_benchmark_benchmark_value_hash ON tmp_benchmark USING HASH (value);

ANALYZE tmp_benchmark;

-- Randomly sample 50k of those values
CREATE TEMPORARY TABLE lookup_test ON COMMIT DROP AS
WITH sub AS (SELECT value FROM tmp_benchmark LIMIT 10000000)
SELECT value FROM sub ORDER BY RANDOM() LIMIT 50000;

ANALYZE lookup_test;

EXPLAIN ANALYZE SELECT * FROM tmp_benchmark WHERE value = ANY (SELECT value FROM lookup_test);
/* Gives:
Hash Semi Join  (cost=1445.00..20996.25 rows=50000 width=20) (actual time=4.595..107.748 rows=50000 loops=1)
  Hash Cond: (tmp_benchmark.value = lookup_test.value)
  ->  Seq Scan on tmp_benchmark  (cost=0.00..16370.00 rows=1000000 width=20) (actual time=0.008..39.863 rows=1000000 loops=1)
  ->  Hash  (cost=820.00..820.00 rows=50000 width=12) (actual time=4.566..4.567 rows=50000 loops=1)
        Buckets: 65536  Batches: 1  Memory Usage: 2704kB
        ->  Seq Scan on lookup_test  (cost=0.00..820.00 rows=50000 width=12) (actual time=0.006..1.836 rows=50000 loops=1)
Planning Time: 0.247 ms
Execution Time: 108.723 ms
 */
EXPLAIN ANALYZE SELECT * FROM tmp_benchmark WHERE value = ANY (CAST((SELECT array_agg(value) FROM lookup_test) AS text[]));
/* Gives:
Bitmap Heap Scan on tmp_benchmark  (cost=985.09..1024.13 rows=10 width=20) (actual time=26.626..14440.230 rows=50000 loops=1)
  Recheck Cond: (value = ANY ($0))
  Rows Removed by Index Recheck: 7
  Heap Blocks: exact=6369
  InitPlan 1 (returns $0)
    ->  Aggregate  (cost=945.00..945.01 rows=1 width=32) (actual time=4.606..4.607 rows=1 loops=1)
          ->  Seq Scan on lookup_test  (cost=0.00..820.00 rows=50000 width=12) (actual time=0.007..1.998 rows=50000 loops=1)
  ->  Bitmap Index Scan on tmp_benchmark_benchmark_value_hash  (cost=0.00..40.08 rows=10 width=0) (actual time=25.848..25.848 rows=50007 loops=1)
        Index Cond: (value = ANY ($0))
Planning Time: 0.098 ms
Execution Time: 14441.520 ms
 */

-- Randomly sample 100k of those values

CREATE TEMPORARY TABLE lookup_test_2 ON COMMIT DROP AS
WITH sub AS (SELECT value FROM tmp_benchmark LIMIT 10000000)
SELECT value FROM sub ORDER BY RANDOM() LIMIT 100000;

EXPLAIN ANALYZE SELECT * FROM tmp_benchmark WHERE value = ANY (SELECT value FROM lookup_test_2);
/* Gives:
Nested Loop  (cost=1555.20..3025.00 rows=78336 width=20) (actual time=15.002..101.907 rows=100000 loops=1)
  ->  HashAggregate  (cost=1555.20..1557.20 rows=200 width=32) (actual time=14.995..26.162 rows=100000 loops=1)
        Group Key: lookup_test_2.value
        Batches: 5  Memory Usage: 6977kB  Disk Usage: 232kB
        ->  Seq Scan on lookup_test_2  (cost=0.00..1359.36 rows=78336 width=32) (actual time=0.006..4.192 rows=100000 loops=1)
  ->  Index Scan using tmp_benchmark_benchmark_value_hash on tmp_benchmark  (cost=0.00..7.88 rows=1 width=20) (actual time=0.001..0.001 rows=1 loops=100000)
        Index Cond: (value = lookup_test_2.value)
        Rows Removed by Index Recheck: 0
Planning Time: 0.148 ms
Execution Time: 103.924 ms
 */
EXPLAIN ANALYZE SELECT * FROM tmp_benchmark WHERE value = ANY (CAST((SELECT array_agg(value) FROM lookup_test_2) AS text[]));
/* Gives:
Bitmap Heap Scan on tmp_benchmark  (cost=1595.29..1634.33 rows=10 width=20) (actual time=44.862..58360.269 rows=100000 loops=1)
  Recheck Cond: (value = ANY ($0))
  Rows Removed by Index Recheck: 16
  Heap Blocks: exact=6370
  InitPlan 1 (returns $0)
    ->  Aggregate  (cost=1555.20..1555.21 rows=1 width=32) (actual time=9.109..9.110 rows=1 loops=1)
          ->  Seq Scan on lookup_test_2  (cost=0.00..1359.36 rows=78336 width=32) (actual time=0.007..4.002 rows=100000 loops=1)
  ->  Bitmap Index Scan on tmp_benchmark_benchmark_value_hash  (cost=0.00..40.08 rows=10 width=0) (actual time=44.242..44.243 rows=100018 loops=1)
        Index Cond: (value = ANY ($0))
Planning Time: 0.089 ms
Execution Time: 58364.590 ms
 */

DROP INDEX tmp_benchmark_benchmark_value_hash;
-- From now on PG is not stupid anymore, even with a 500k sample and bitmap heap scan plan it doesn't do n² anymore

-- On the same 100k sample:
EXPLAIN ANALYZE SELECT * FROM tmp_benchmark WHERE value = ANY (CAST((SELECT array_agg(value) FROM lookup_test_2) AS text[]));
/* Gives:
Bitmap Heap Scan on tmp_benchmark  (cost=1599.54..1638.58 rows=10 width=20) (actual time=236.604..252.188 rows=100000 loops=1)
  Recheck Cond: (value = ANY ($0))
  Heap Blocks: exact=6370
  InitPlan 1 (returns $0)
    ->  Aggregate  (cost=1555.20..1555.21 rows=1 width=32) (actual time=9.091..9.093 rows=1 loops=1)
          ->  Seq Scan on lookup_test_2  (cost=0.00..1359.36 rows=78336 width=32) (actual time=0.009..3.946 rows=100000 loops=1)
  ->  Bitmap Index Scan on tmp_benchmark_unique_value_and_corresponding_index  (cost=0.00..44.33 rows=10 width=0) (actual time=236.169..236.169 rows=100000 loops=1)
        Index Cond: (value = ANY ($0))
Planning Time: 0.161 ms
Execution Time: 254.145 ms
*/

-- So despite doing a bitmap heap scan on the same 100k elements array as above, this one seemingly doesn't n².

ROLLBACK;
```

diesel_tests/tests/filter_operators.rs
```diff
@@ -264,6 +264,36 @@ fn filter_by_in() {
...
 
+#[test]
+#[cfg(feature = "postgres")]
+fn filter_array_by_in() {
+    use crate::schema::posts::dsl::*;
+
+    let connection: &mut PgConnection = &mut connection();
+    let tag_combinations_to_look_for: &[&[&str]] = &[&["foo"], &["foo", "bar"], &["baz"]];
+    let result: Vec<i32> = posts
+        .filter(tags.eq_any(tag_combinations_to_look_for))
```
Ten0, line 275: we probably want to document this behavior in these places:
https://github.com/diesel-rs/diesel/blob/5a2940eeaa935c0cc0a7fc57ffe012f3aaa80f43/diesel/src/expression/array_comparison.rs#L31-L32
https://github.com/diesel-rs/diesel/blob/5a2940eeaa935c0cc0a7fc57ffe012f3aaa80f43/diesel/src/expression/array_comparison.rs#L51-L52
https://github.com/diesel-rs/diesel/blob/5a2940eeaa935c0cc0a7fc57ffe012f3aaa80f43/diesel/src/expression_methods/global_expression_methods.rs#L109-L110

diesel/src/pg/query_builder/query_fragment_impls.rs
```diff
@@ -92,10 +100,15 @@ impl<ST, I> QueryFragment<Pg, PgStyleArrayComparison> for Many<ST, I>
...
     Vec<I>: ToSql<Array<ST>, Pg>,
+    I: ToSql<ST, Pg>,
     Pg: HasSqlType<ST>,
 {
     fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
-        out.push_bind_param::<Array<ST>, Vec<I>>(&self.values)
+        if ST::IS_ARRAY {
+            self.walk_ansi_ast(out)
+        } else {
+            out.push_bind_param::<Array<ST>, Vec<I>>(&self.values)
```
Ten0, line 110: this got me thinking (long-term changes thoughts): maybe `QueryId` should be generic on the backend as well, so that we can properly express that this variant could in fact use static query id.
But maybe that's not worth.

Ten0 (approved):
I realized I also have examples in mind where `= ANY(array)` performs better than `= ANY(subquery)` because it manages to push a filter further into the query plan in that case, so I think that ends up convincing me that we just want to map to SQL as close as possible, so that we should do `IN([1],[2],...)`.

The only thing is we should probably document it, considering that sometimes `= ANY(...)` manages to not do n², and I suspect `IN([1],[2],...)` may not be able to be as efficient in similar scenarios since it seems to expand to just many `OR` clauses...

Ten0:
For completeness: while the "reproducer" above is indeed n² with expressions like `= ANY(ARRAY(subselect))`, I can't seem to reproduce the same n² with an actual `= ANY($1)` anymore, so that in fact probably behaves fine (https://github.com/postgres/postgres/commit/50e17ad281b8d1c1b410c9833955bc80fbad4078), which in turn means that the array case of `array_value IN (arr1,arr2,arr3)` being expanded as `array_value = arr1 OR array_value = arr2` may be:

1. Way slower than people might expect because the OR are rechecked linearly (n²)
2. Crashing faster, because postgres will return a `stack depth limit exceeded` error when there are more than 10k values

I still feel slightly in favor of using `IN($1, $2, ...)` anyway,because we probably want to write the SQL that people are essentially asking for.

PR #4351, opened 2024-11-19
Add a missing type def

weiznich:
This commit adds a missing type-def that allows now to use `table.count()` with `#[auto_type]`

Found and reported by @Turbo87


rienmack (approved):
Looks good! 🚀

PR #4353, opened 2024-11-20
Add support for subselects to the `array(...)` function

Ten0:
As underlined [here](https://github.com/diesel-rs/diesel/pull/4350#pullrequestreview-2447144357), there are cases where `= ANY(array)` does not behave the same as `= ANY(subquery)`, so it may be useful to explicitly convert a subquery into an array.

This PR allows doing this.

It's technically a breaking change, but only to the extent to which people implemented the `AsExpressionList` trait manually, which seems unlikely enough.

### DIFF
--- a/diesel/src/pg/expression/array.rs
+++ b/diesel/src/pg/expression/array.rs
@@ -11,10 +11,10 @@ use crate::query_builder::{
 use crate::sql_types::{self, SqlType};
 use std::marker::PhantomData;
 
-/// Creates an `ARRAY[...]` expression.
+/// Creates an `ARRAY[e1, e2, ...]` or `ARRAY(subselect)` expression.
 ///
 /// The argument should be a tuple of expressions which can be represented by the
-/// same SQL type.
+/// same SQL type, or a subquery.
 ///
 /// # Examples
 ///
--- a/diesel/src/pg/expression/array.rs
+++ b/diesel/src/pg/expression/array.rs
@@ -74,6 +74,18 @@ pub trait IntoArrayExpression<ST: SqlType + TypedExpressionType> {
     fn into_array_expression(self) -> Self::ArrayExpression;
 }
 
+impl<ST, T> IntoArrayExpression<ST> for T
+where
+    T: AsExpression<sql_types::Array<ST>>,
+    ST: SqlType + TypedExpressionType + 'static,
+{
+    type ArrayExpression = <T as AsExpression<sql_types::Array<ST>>>::Expression;
+
+    fn into_array_expression(self) -> Self::ArrayExpression {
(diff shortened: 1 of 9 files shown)


diesel/src/pg/expression/array.rs
```diff
@@ -43,20 +42,63 @@ pub struct ArrayLiteral<T, ST> {
...
+    /// Type of the expression returned by [AsArrayExpression::as_in_expression]
+    type ArrayExpression: Expression<SqlType = sql_types::Array<ST>>;
+
+    /// Construct the diesel query dsl representation of
+    /// the `ARRAY (values)` clause for the given type
+    fn into_array_expression(self) -> Self::ArrayExpression;
+}
+
+/// Implement as a no-op for things that were already arrays (that is, don't wrap in ARRAY[]).
+impl<ST, T> IntoArrayExpression<ST> for T
```
Ten0, line 83: I'm not sure if we should have this impl. Maybe not, if we want to map as close as SQL as possible and forbid writing `array(array_expr)` for something that's actually going to get serialized as just `array_expr`.

diesel/src/pg/expression/array.rs
```diff
@@ -43,20 +42,63 @@ pub struct ArrayLiteral<T, ST> {
...
 {
-    ArrayLiteral {
-        elements: elements.as_expression_list(),
-        _marker: PhantomData,
+    type ArrayExpression = <T as AsExpression<sql_types::Array<ST>>>::Expression;
+
+    fn into_array_expression(self) -> Self::ArrayExpression {
+        <T as AsExpression<sql_types::Array<ST>>>::as_expression(self)
     }
 }
```
Ten0, line 93: I'm not sure if we should have this impl. Maybe not, if we want to map as close as SQL as possible and forbid writing `array(array_expr)` for something that's actually going to get serialized as just `array_expr`.
It seems easier to add later on than to remove.

diesel/src/expression/array_comparison.rs
```diff
@@ -74,15 +72,15 @@ impl<T, U> NotIn<T, U> {
 impl<T, U> Expression for In<T, U>
 where
     T: Expression,
-    U: Expression<SqlType = T::SqlType>,
+    U: InExpression<SqlType = T::SqlType>,
 {
     type SqlType = Bool;
```
Ten0, line 77: TODO: this seems wrong: PG returns null for IN expressions if LHS is null

Ten0: Yes indeed that's already wrong on master as showcased by 6912d7b23d4433ae5dc03afce3367a1de966a69c

Ten0: #4354

weiznich (approved):
Looks good for me :+1:

PR #4357, opened 2024-11-24
add jsonb_set support

accestfromim:
add jsonb_set support function mentioned in issue #4216


accestfromim:
I am not sure if I did it right.  I'd be glad if I can receive your instruction

prkbuilds (commented):
LGTM. Types are correctly defined as well the nullability, Great Work!

prkbuilds (approved):

diesel/src/pg/expression/functions.rs
```diff
@@ -2398,3 +2398,80 @@ define_sql_function! {
...
+    /// let empty:Vec<String> = Vec::new();
+    /// let result = diesel::select(jsonb_set::<Jsonb, Array<Nullable<Text>>, _, _, _>(
+    ///         // cannot be json!(null)
+    ///         json!([]),
+    ///         empty,
+    ///         json!(null)
+    ///     )).get_result::<Value>(connection)?;
+    /// let expected = json!([]);
+    /// assert_eq!(result, expected);
+    ///
```
weiznich, line 2467: I'm sorry to bring that up, but the following example won't infer nullablity correctly:

```rust
    /// let result = diesel::select(jsonb_set::<Jsonb, Nullable<Array<Nullable<Text>>>, _, _, _>(
    ///         // cannot be json!(null)
    ///         json!([]),
    ///         None,
    ///         json!({"foo": 42})
    ///     )).get_result::<Value>(connection)?;
```

This is expected to only accept a `Option<Value>` as return type as the return value will be `NULL` in this case.

weiznich (changes requested):
Thanks for submitting this PR.

I've found an potential issue with nullability inference that should be fixed. You can use the existing [`CombinedNullableValue`](https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/expression/expression_methods.rs#L3936) trait for this.

diesel/src/pg/expression/functions.rs
accestfromim: Thanks for your kind instuction.
I fixed it by doing this:
before:
```
fn jsonb_set<
        E: JsonbOrNullableJsonb + SingleValue,
        Arr: TextArrayOrNullableTextArray + SingleValue
    >(base: E, path: Arr, new_value: E) -> E;
```
after:
```
fn jsonb_set<
        E: JsonbOrNullableJsonb + SingleValue,
        Arr: TextArrayOrNullableTextArray + CombinedNullableValue<E,Jsonb>
    >(base: E, path: Arr, new_value: E) -> Arr::Out;
```
Thanks for your instruction again!

weiznich (approved):
Thanks for the update, looks good now

PR #4358, opened 2024-11-25
Remove gitter from the readme

weiznich:


kw217 (approved):
Ah, I just checked the URL and it was a very unpleasant experience. I agree, removing it is very sensible.
Changes look good.

PR #4361, opened 2024-11-26
Remove more mentions of gitter

weiznich:
I've searched and found more locations. That should hopefully be the last round of changes regarding to gitter.


kw217 (changes requested):
Great! There's one more hiding in `.github/ISSUE_TEMPLATE/config.yml` which should also be removed. I think that's all of them now.

kw217 (changes requested):
It shouldn't say "Complex questions", it should just say "Questions". And the `about` needs to be updated to match, e.g., "Do you have a question? Ask in our forum".

kw217 (approved):
Nice! Thanks.

PR #4364, opened 2024-11-26
Add a skip_update field attribute for #[derive(AsChangeset)] (#860)

Jaltaire:
Implement `skip_update` attribute for `#[derive(AsChangeset)]` as described here: https://github.com/diesel-rs/diesel/issues/860#issuecomment-2114074186.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Added `limit()` and `offset()` DSL to combination clauses such as `UNION`
 * Fixed `#[derive(Identifiable)]` ignoring attribute `#[diesel(serialize_as)]` on primary keys
 * Added embedded struct support for `AsChangeset` via `#[diesel(embed)]`
+* Added a `#[diesel(skip_update)]` attribute for the `AsChangeset` derive to skip updating a field present in the struct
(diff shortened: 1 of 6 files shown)


Jaltaire:
@weiznich, I'm wondering if `skip_update` should be a separate attribute, or if it should be rolled into `skip_insertion` and renamed?

If the purpose of both attributes is to skip the corresponding operation on a generated column, is there ever a case where we would want to `skip_insertion` but **not** `skip_update`, or vice versa? Perhaps there is a scenario in which this is valid of which I am not considering, but if not, I do think this makes for a more error-prone implementation, by requiring these annotations be needlessly duplicated for what should be consistent behavior between inserts and updates. On the other hand, _should_ someone require differing behavior for their particular use case, then we already get that functionality here.

weiznich (changes requested):
Thanks for working on this :heart:.
The change looks reasonable for me. It just implements what was already discussed. 

Before merging I would like to address the following points:

* Add documentation [here](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/src/lib.rs#L101)
* Add a changelog entry
* Add at least one test [here](https://github.com/diesel-rs/diesel/blob/master/diesel_derives/tests/as_changeset.rs)

As for the discussion of a "merged skip" attribute: I would like to go with separate attributes for now, but we might consider something like that in a followup change

PR #4365, opened 2024-11-26
Introduce a `#[declare_sql_function]` attribute macro

weiznich:
This commit introduces a new `#[declare_sql_function]` attribute macro that can be applied to `extern "SQL"` blocks. This is essentially the same as the existing `define_sql_function!` function like macro in terms of functionality.

I see the following advantages of using an attribute macro + an `extern "SQL"` block instead:

* This is closer to rust syntax, so rustfmt will understand that and work correctly inside these blocks
* This allows to put several functions into the same block
* Maybe in the future this also allows to apply attributes to the whole block instead of to each item (that's something for a followup PR)

The downside of this change is that we then have three variants to declare sql functions:

* `sql_function!()` (deprectated)
* `define_sql_function!()` (introduced in 2.2, we might want to deprecate that as well?)
* The new attribute macro

For now this is more a request for comments if that's a desirable change. If we decide to go that way I would like to:

* [x] Deprecating `define_sql_function!()`
* [x] Reexport this new macro from the crate root
* [x] Convert all existing usages in diesel
* [x] Add a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,14 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Support for postgres multirange type
 * Added `diesel::r2d2::TestCustomizer`, which allows users to customize their `diesel::r2d2::Pool`s
 in a way that makes the pools suitable for use in parallel tests.
+* Added support for built-in PostgreSQL range operators and functions
+* Added support for various built-in PostgreSQL array functions
 * Added `Json` and `Jsonb` support for the SQLite backend.
+* Added a `#[diesel::declare_sql_function]` attribute macro to easily define support for 
+  multiple sql functions at once via an `extern "SQL"` block
+
+### Fixed 
+
 * Fixed diesel thinking `a.eq_any(b)` was non-nullable even if `a` and `b` were nullable.
 * Generate `InstrumentationEvent::BeginTransaction` for immediate and exclusive transactions in SQLite
 * Added `wasm32-unknown-unknown` target support for sqlite backend.
--- a/diesel/src/connection/statement_cache/strategy.rs
+++ b/diesel/src/connection/statement_cache/strategy.rs
@@ -149,7 +149,6 @@ mod testing_utils {
 mod tests_pg {
     use crate::connection::CacheSize;
     use crate::dsl::sql;
-    use crate::expression::functions::define_sql_function;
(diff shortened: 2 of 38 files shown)


Ten0:
Currently in the examples and tests we write things such as:
```rust
define_sql_function!(fn nextval(a: sql_types::VarChar) -> sql_types::BigInt);
```
Indeed this feels harder to read than
```rust
#[declare_sql_function]
extern "SQL" {
    fn nextval(a: sql_types::VarChar) -> sql_types::BigInt;
}
```

However I actually almost never use the first writing, I rather use this one:
```rust
define_sql_function! {
    fn nextval(a: sql_types::VarChar) -> sql_types::BigInt;
}
```
and that one feels to me like it's as readable as the proposed syntax.

My first feeling is that the new syntax may feel more magical and more compiler-integrated than it actually is.
It also seems more verbose. (In particular "SQL" as a string literal rubs me the wrong way, maybe because there's already `#[declare_sql_function]` above 😬)

I wouldn't want to prevent anyone from having this writing if they prefer though. :)
It feels like a nice compromise would be to allow the two writings (and not deprecate `define_sql_function`)? (in a similar spirit as we did with `alias`)?

> This is closer to rust syntax, so rustfmt will understand that and work correctly inside these blocks

I recall seeing rustfmt being able to format code that is used in macros, maybe specifically when used through rust-analyzer which is able to understand how items end up being parsed (expression, ..., in particular with `lazy_static!`).
Currently we seem to be parsing [each element manually](https://github.com/diesel-rs/diesel/blob/5a2940eeaa935c0cc0a7fc57ffe012f3aaa80f43/diesel_derives/src/sql_function.rs#L451-L463). I wonder what would be the extent to which we could instead parse it directly as `ForeignItemFn`, which seems to be what the new code does, and that may solve this issue?
(But then maybe that doesn't work and rust-anlayzer/rustfmt uses some magic that's related to spans of expanded macros instead, I'm not sure.)

weiznich:
As for the formatting argument: It seems like you can configure rustfmt to format macro calls: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#skip_macro_invocations

That setting is `false` by default and currently unstable.

Ten0:
I have no strong preference about this.
My gut feeling is that this writing feels more magical than it really is but if you feel strongly about not having two writings that's fine with me 🙂

PR #4370, opened 2024-11-29
Add a dependabot config

weiznich:
This commit introduces a dependabot config to automate dependency updates. Hopefully the `version_strategy: widen` setting works, because that's want we want here

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,5 +7,5 @@ updates:
     schedule:
       interval: "weekly"
     reviewers:
-      - "diesel/reviewer"
+      - "diesel/reviewers"
     versioning-strategy: "widen"


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    allow:
+      dependency-type: "direct"
+    schedule:
+      interval: "weekly"
+    reviewers:
+      - "diesel/reviewer"
```
rienmack, line 10:
```suggestion
      - "diesel/reviewers"
```
question: Should this be diesel/reviewers? 

Looks good to me other than this question! 👍

rienmack (approved):
Looks good! 👍🚀

PR #4386, opened 2024-12-10
Fix typo in code example

BrianDouglasIE:
`dotenvy` should be `dotenv`


weiznich:
Thanks for opening this PR. `dotenvy` is the correct crate name there as the `dotenv` crate is no longer maintained.

BrianDouglasIE:
@weiznich thanks for clearing that up.

Issue #4366, opened 2024-11-27
Add missing sqlite json/jsonb functions and operators


weiznich:
Diesel currently supports the sqlite json/jsonb types. We do not provide built-in support for [various methods](https://sqlite.org/draft/json1.html) available for these types. This is a tracking issue for adding support for these methods. 

The general strategy for adding support for new methods is as following:

1. Define the function via [`define_sql_function!()`](https://docs.diesel.rs/master/diesel/expression/functions/macro.define_sql_function.html). These functions can be defined [here](https://github.com/diesel-rs/diesel/tree/master/diesel/src/sqlite/expression) in a new `functions.rs` module. See the [linked definition of `to_json` for an example from the postgres backend](https://github.com/diesel-rs/diesel/blob/a55f0fef4816602889c51fc2a403cbb20563ec2b/diesel/src/pg/expression/functions.rs#L1567-L1610) as an example. This function should have a short documentation snippet with an example (See the linked sqlite documentation for examples for all of the function, please also add variants with null values, etc). If there is a `json` and a `jsonb` variant, please implement both in a single PR. 
2. If the function is generic, add a helper type definition [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/sqlite/expression/helper_types.rs). Again see the [linked definition for an example for the `to_json` function](https://github.com/diesel-rs/diesel/blob/a55f0fef4816602889c51fc2a403cbb20563ec2b/diesel/src/pg/expression/helper_types.rs#L512-L515) from the postgres backend for an example.
3. Add a test for  `#[auto_type]` support for the newly added function [here](https://github.com/diesel-rs/diesel/blob/6c1cb1d42eb2b3000502d67249dc3cdd76066dd7/diesel_derives/tests/auto_type.rs#L413)
5. Submit a PR with the change

Method list:

- [x] [`json(json)`](https://sqlite.org/draft/json1.html#jmini) *
- [x] [`jsonb(json)`](https://sqlite.org/draft/json1.html#jminib) *
- [ ] [`json_array(value1, value2, …)`](https://sqlite.org/draft/json1.html#jarray) **
- [ ] [`jsonb_array(value1, value2, …)`](https://sqlite.org/draft/json1.html#jarrayb) **
- [x] [`json_array_length(json)`](https://sqlite.org/draft/json1.html#jarraylen) *
- [x] [`json_array_length(json, path)`](https://sqlite.org/draft/json1.html#jarraylen) *
[...]

PR #4388, opened 2024-12-11
Add sqlite functions `json` and `jsonb`

xuehaonan27:
Add SQLite functions `json` and `jsonb` according to issue #4366 .

Following the implementation of PostgreSQL binding, mod declaration statements were added and modified in files such as diesel/src/sqlite/mod.rs and diesel/src/sqlite/expression/mod.rs to facilitate subsequent contributors' coding and testing.

NOTE: testing on sqlite 3.47.2 and dependency libsqlite-sys 0.30.1 failed with no such function: jsonb. Given that sqlite 3.47.2 already had jsonb available, the problem is within the SQLite binding crate. It seems that libsqlite-sys crate does not support jsonb currently, but corresponding ergonomic wrapper rusqlite does and fits well. The `rusqlite` dependency might be removed in the future if libsqlite-sys adds support for JSONB related functions. For now, I add it to make everything work well.

See this for more info: https://github.com/rusqlite/rusqlite/issues/1453

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -25,6 +25,7 @@ byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.20", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
 libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled_bindings"] }
+rusqlite = { version = "0.32.0", features = ["bundled"] }
 mysqlclient-sys = { version = ">=0.2.5, <0.5.0",  optional = true }
 mysqlclient-src = { version = "0.1.0", optional = true }
 pq-sys = { version = ">=0.4.0, <0.7.0", optional = true }
--- a/diesel/src/sqlite/expression/expression_methods.rs
+++ b/diesel/src/sqlite/expression/expression_methods.rs
@@ -1,5 +1,8 @@
 //! Sqlite specific expression methods.
 
+pub(in crate::sqlite) use self::private::{
+    JsonOrNullableJsonOrJsonbOrNullableJsonb, MaybeNullableValue,
+};
 use super::operators::*;
 use crate::dsl;
 use crate::expression::grouped::Grouped;
@@ -82,3 +85,26 @@ pub trait SqliteExpressionMethods: Expression + Sized {
 }
 
 impl<T: Expression> SqliteExpressionMethods for T {}
+
+pub(in crate::sqlite) mod private {
+    use crate::sql_types::{Json, Jsonb, MaybeNullableType, Nullable, SingleValue};
+
+    pub trait JsonOrNullableJsonOrJsonbOrNullableJsonb {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Json {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Nullable<Json> {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Jsonb {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Nullable<Jsonb> {}
+
+    pub trait MaybeNullableValue<T>: SingleValue {
+        type Out: SingleValue;
+    }
+
+    impl<T, O> MaybeNullableValue<O> for T
+    where
+        T: SingleValue,
+        T::IsNull: MaybeNullableType<O>,
+        <T::IsNull as MaybeNullableType<O>>::Out: SingleValue,
+    {
+        type Out = <T::IsNull as MaybeNullableType<O>>::Out;
+    }
+}
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -1,0 +1,114 @@
+//! SQLite specific functions
+use crate::expression::functions::define_sql_function;
+use crate::sql_types::*;
+use crate::sqlite::expression::expression_methods::JsonOrNullableJsonOrJsonbOrNullableJsonb;
+use crate::sqlite::expression::expression_methods::MaybeNullableValue;
+
+#[cfg(feature = "sqlite")]
+define_sql_function! {
+    /// Verifies that its argument is a valid JSON string or JSONB blob and returns a minified
+    /// version of that JSON string with all unnecessary whitespace removed.
+    ///
+    /// # Example
+    ///
+    /// ```rust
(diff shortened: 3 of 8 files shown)


xuehaonan27:
Branch updated by force pushing: `libsqlite3-sys` feature `bundled-bindings` updated to `bundled` to support SQLite JSONB. No longer using `rusqlite` to keep dependency tree clean.

diesel/src/sqlite/expression/functions.rs
```diff
@@ -0,0 +1,114 @@
...
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn json<E: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue + MaybeNullableValue<Text>>(e: E) -> E::Out;
+}
+
+#[cfg(feature = "sqlite")]
+define_sql_function! {
+    ///
```
weiznich, line 68: Documentation is missing here

diesel/src/sqlite/expression/functions.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// let result = diesel::select(json::<Nullable<Jsonb>, _>(None::<Value>))
+    ///     .get_result::<Option<String>>(connection)?;
+    ///
+    /// assert!(result.is_none());
+    ///
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn json<E: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue + MaybeNullableValue<Text>>(e: E) -> E::Out;
```
weiznich, line 63: I think the correct signature here would be `fn json<E: TextOrNullableText + MaybeNullableValue<Json>>(e: E) -> E::Out;` as this function essentially turns text into json, not the other way around.

diesel/src/sqlite/expression/functions.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// let result = diesel::select(jsonb::<Nullable<Jsonb>, _>(None::<Value>))
+    ///     .get_result::<Option<Value>>(connection)?;
+    ///
+    /// assert!(result.is_none());
+    /// println!("jsonb null");
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn jsonb<E: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue + MaybeNullableValue<Jsonb>>(e: E) -> E::Out;
```
weiznich, line 113: I think the correct signature here would be `fn json<E: BinaryOrNullableBinary + MaybeNullableValue<Json>>(e: E) -> E::Out;` as this function essentially turns a blob into json, not the other way around.

diesel/Cargo.toml
```diff
@@ -24,7 +24,7 @@ include = [
 byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.20", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
-libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled_bindings"] }
+libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled"] }
```
weiznich, line 27: I would prefer not to enable the bundled feature here as that breaks anyone that requires dynamically linking libsqlite3.

I would rather prefer to have a `SELECT sqlite_version();` query in the doc tests and just skip the test if the local version is too old.

weiznich (changes requested):
Thanks for working on this. This is really welcome :heart: 

I've left a bunch of comments on details I would like to see solved differently to ensure that this works similar to other parts of diesel and does not break existing diesel users.

diesel/src/sqlite/expression/functions.rs
xuehaonan27: Maybe you want to say `fn **jsonb**<E: BinaryOrNullableBinary + MaybeNullableValue<**Jsonb**>>(e: E) -> E::Out;` ? I tried out and found `MaybeNullableValue<Jsonb>` works but `MaybeNullableValue<Json>` not.

diesel/Cargo.toml
xuehaonan27: In fact, this has nothing to do with the version of `sqlite3`, but with the feature switch of the `libsqlite3-sys` crate that `diesel` depends on. If we continue to use the `bundled_bindings` feature instead of `bundled`, even if the sqlite3 version is higher than 3.45.0 (mine is 3.47.2), it will still report an error of `no such function: jsonb`. This means that we cannot test `jsonb` at all, and (possibly) all JSONB-related functions mentioned in that issue.

xuehaonan27: I have raised an issue under `rusqlite` (the repository where `libsqlite3-sys` is located) and received a relevant answer. https://github.com/rusqlite/rusqlite/issues/1614

diesel/src/sqlite/expression/functions.rs
weiznich: Yes, that should be `jsonb` and not `json`. I just copied the other commend and failed to adjust that part, sorry for this.

diesel/Cargo.toml
weiznich: I've tested that locally and I believe it really just depends on which libsqlite3 (system library, not rust crate) version is linked. If that version is newer than 3.45 it works, if it's older it doesn't work. At least that's the case with my local 3.46 version.

xuehaonan27: Oh you are right, I rechecked my libsqlite and found that it was linked to the system library provided by MacOS instead of the library provided by homebrew. 😭

xuehaonan27:
Change made

diesel/src/sqlite/expression/expression_methods.rs
```diff
@@ -82,3 +85,46 @@ pub trait SqliteExpressionMethods: Expression + Sized {
...
+    use crate::sql_types::{Binary, MaybeNullableType, Nullable, SingleValue, Text};
+
+    /*
+    pub trait JsonOrNullableJsonOrJsonbOrNullableJsonb {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Json {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Nullable<Json> {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Jsonb {}
+    impl JsonOrNullableJsonOrJsonbOrNullableJsonb for Nullable<Jsonb> {}
+    */
+
```
weiznich, line 99:
```suggestion
```

weiznich (approved):
Thanks for fixing

PR #4390, opened 2024-12-15
[Test] Disable Nagle's algorithm

c410-f3r:
Much to my surprise, almost all database clients apply `stream.set_nodelay(true)` by default.

`mysql`: https://github.com/blackbeam/rust-mysql-simple/blob/bdc32fae7f2869c29ca52284dff7bf19beeaca56/src/io/tcp.rs#L102
`sqlx`: https://github.com/launchbadge/sqlx/blob/1678b19a4672fd6a18b4891c53bf0b57638b92a4/sqlx-core/src/net/socket/mod.rs#L201
`tokio-postgres`: https://github.com/sfackler/rust-postgres/blob/6ae17e0f2d174fc4d5592ac7869909d7bdd346d9/tokio-postgres/src/connect_socket.rs#L27

If possible, I would like to see the impact of this parameter in the `wtx` project.

### DIFF
--- a/diesel_bench/benches/wtx.rs
+++ b/diesel_bench/benches/wtx.rs
@@ -259,11 +259,13 @@ async fn connection() -> Executor<wtx::Error, ExecutorBuffer, TcpStream> {
         .expect("DATABASE_URL must be set in order to run tests");
     let uri = UriRef::new(url.as_str());
     let mut rng = Xorshift64::from(simple_seed());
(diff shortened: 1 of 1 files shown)


c410-f3r:
Ping

The changes are temporary, everything can be rolled-back after a couple of runs

weiznich:
Seems like the CI is broken, I need to investigate why, so this might take some time.

PR #4393, opened 2024-12-16
Generate `InstrumentationEvent::BeginTransaction` for custom transaction SQL

### LINKED ISSUE #4392: Missing transaction instrumentation in `AnsiTransactionManager`
### Versions

- **Rust:** 1.83.0
- **Diesel:** 2.2.6
- **Database:** SQLite
- **Operating System** Linux

### Feature Flags

- **diesel:** `r2d2`, `returning_clauses_for_sqlite_3_35`, `sqlite`, `uuid`

## Problem Description

The following problem happens with `SqliteConnection`'s methods `immediate_transaction()` and `exclusive_transaction()`. Under the hood, these use `AnsiTransactionManager::begin_transaction_sql()` instead of the usual `begin_transaction()` (implemented for trait `TransactionManager`).

The latter method's implementation generates instrumentation events of type `InstrumentationEvent::BeginTransaction`:

https://github.com/diesel-rs/diesel/blob/b705023d85e6ef76292a427aa70f21c28a5902ff/diesel/src/connection/transaction_manager.rs#L337-L359

The former, however, is not generating these events:

https://github.com/diesel-rs/diesel/blob/b705023d85e6ef76292a427aa70f21c28a5902ff/diesel/src/connection/transaction_manager.rs#L314-L329

This causes custom instrumentation to no longer uphold the invariant that every `BeginTransaction` is followed by exactly one of either `RollbackTransaction` or `CommitTransaction`.

Trace ongoing transactions using custom instrumentation, including `IMMEDIATE` and `EXCLUSIVE` transactions.

Using `SqliteConnection::immediate_transaction()` or `exclusive_transaction()` over the regular `transaction()` should generate the same instrumentation events. In particular, starting transactions with these two methods should generate a `BeginTransaction` event.

Transactions started with `SqliteConnection::immediate_transaction()` or `exclusive_transaction()` do not generate the expected instrumentation event `BeginTransaction`.

No.

### Steps to reproduce

See above. The problem is independent of any actual database structure or queries inside the transcation.

## Checklist

sgoll:
Fixes #4392

### DIFF
--- a/diesel/src/connection/transaction_manager.rs
+++ b/diesel/src/connection/transaction_manager.rs
@@ -316,15 +316,19 @@ impl AnsiTransactionManager {
         Conn: Connection<TransactionManager = Self>,
     {
         let state = Self::get_transaction_state(conn)?;
-        match state.transaction_depth() {
-            None => {
-                conn.batch_execute(sql)?;
-                Self::get_transaction_state(conn)?
-                    .change_transaction_depth(TransactionDepthChange::IncreaseDepth)?;
-                Ok(())
-            }
-            Some(_depth) => Err(Error::AlreadyInTransaction),
+        if let Some(_depth) = state.transaction_depth() {
+            return Err(Error::AlreadyInTransaction);
         }
+        conn.instrumentation().on_connection_event(
+            super::instrumentation::InstrumentationEvent::BeginTransaction {
+                depth: NonZeroU32::new(1).expect("Transaction depth is too large"),
+            },
+        );
+        conn.batch_execute(sql)?;
+        Self::get_transaction_state(conn)?
+            .change_transaction_depth(TransactionDepthChange::IncreaseDepth)?;
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR, seems like this was just missed while implementing that feature. 

> Please note that I was unsure which branch to base this pull request on. I picked 2.2.x because that seems to include the latest release. However, I can easily rebase my changes on top of master if necessary.

PR's should target the master branch by default as that's where the development is happening. If something is an important fix we might backport is later. That could be the case here.

diesel/src/connection/transaction_manager.rs
```diff
@@ -316,15 +316,22 @@ impl AnsiTransactionManager {
...
-            Some(_depth) => Err(Error::AlreadyInTransaction),
+        if let Some(_depth) = state.transaction_depth() {
+            return Err(Error::AlreadyInTransaction);
         }
+        let instrumentation_depth = NonZeroU32::new(1);
+        // Keep remainder of this method in sync with `begin_transaction()`.
+
+        conn.instrumentation().on_connection_event(
+            super::instrumentation::InstrumentationEvent::BeginTransaction {
+                depth: instrumentation_depth.expect("Transaction depth is too large"),
```
weiznich, line 327:
```suggestion
                depth: instrumentation_depth.expect("We know that 1 is not zero"),
```

weiznich (approved):

PR #4395, opened 2024-12-18
Add SQLite function `json_pretty`

xuehaonan27:
`TextOrNullableTextOrBinaryOrNullableBinary`. SQLite doc says "the first argument is the JSON or JSONB that is to be pretty-printed" so I added such a trait. I don't know whether it's proper.

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -4,6 +4,7 @@ use crate::sql_types::*;
 use crate::sqlite::expression::expression_methods::BinaryOrNullableBinary;
 use crate::sqlite::expression::expression_methods::MaybeNullableValue;
 use crate::sqlite::expression::expression_methods::TextOrNullableText;
+use crate::sqlite::expression::expression_methods::TextOrNullableTextOrBinaryOrNullableBinary;
 
 #[cfg(feature = "sqlite")]
 define_sql_function! {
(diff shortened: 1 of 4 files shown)


xuehaonan27:
Related to Issue #4366

diesel/src/sqlite/expression/functions.rs
```diff
@@ -105,3 +106,110 @@ define_sql_function! {
...
+    ///
+    /// let result = diesel::select(json_pretty::<Nullable<Text>, _>(None::<&str>))
+    ///     .get_result::<Option<String>>(connection)?;
+    ///
+    /// assert!(result.is_none());
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    fn json_pretty<E: TextOrNullableTextOrBinaryOrNullableBinary + MaybeNullableValue<Text>>(e: E) -> E::Out;
```
weiznich, line 214: I think this signature is unfortunately not correct yet. The sqlite documentation says:

> The json_pretty() function works like [json()](https://sqlite.org/draft/json1.html#jmini) except that it adds extra whitespace to make the JSON result easier for humans to read. **The first argument is the JSON or JSONB that is to be pretty-printed.** The optional second argument is a text string that is used for indentation. If the second argument is omitted or is NULL, then indentation is four spaces per level.

which would translate to 

```rust
fn json_pretty<J: JsonOrJsonbOrNullableJson + MaybeNullableValue<Text>>(j: J) -> J::Out;
```

this then would require users to call it that way: `select(json_pretty(json("your_json_text"))` or `select(json_pretty(jsonb(b"your_json_blob"))`

weiznich (commented):
Thanks for opening this PR. I think the signature is not right yet, as pointed out in the comment at the function. It would be great to fix that before merging.

diesel/src/sqlite/expression/functions.rs
weiznich: You are correct we want to use `JsonOrNullableJsonOrJsonbOrNullableJsonb` there (any likely also for most of the other sqlite json functions).

weiznich (approved):
Thanks for the fast update. Looks good now

PR #4396, opened 2024-12-18
Add support for postgre jsonb_set_lax

accestfromim:
#4216 
Though create_if_missing and null_value_treatment is optional, the case without them is already supported in jsonb_set and jsonb_set_create_if_missing. So I just do not treat them as optional field here. Do you think it's ok?

### DIFF
--- a/diesel/src/pg/expression/functions.rs
+++ b/diesel/src/pg/expression/functions.rs
@@ -13,6 +13,7 @@ use crate::pg::expression::expression_methods::MultirangeOrRangeMaybeNullable;
 use crate::pg::expression::expression_methods::RangeOrNullableRange;
 use crate::pg::expression::expression_methods::RecordOrNullableRecord;
 use crate::pg::expression::expression_methods::TextArrayOrNullableTextArray;
+use crate::pg::expression::expression_methods::TextOrNullableText;
 use crate::sql_types::*;
 
 define_sql_function! {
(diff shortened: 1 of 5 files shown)


diesel/src/pg/expression/functions.rs
```diff
@@ -2578,3 +2579,101 @@ define_sql_function! {
...
+    /// All earlier steps in the path must exist, or the target is returned unchanged.
+    /// As with the path oriented operators, negative integers that appear in the path count from the end of JSON arrays.
+    /// If the last path step is an array index that is out of range,
+    ///    the new value is added at the beginning of the array if the index is negative,
+    ///     or at the end of the array if it is positive.
+    ///
+    /// If new_value is not NULL, behaves identically to jsonb_set.
+    ///    Otherwise behaves according to the value of null_value_treatment
+    ///    which must be one of 'raise_exception', 'use_json_null', 'delete_key', or 'return_target'.
+    ///    The default is 'use_json_null'.
```
weiznich, line 2599: I would rather prefer having an enum with all allowed variants instead of allowing a text input here.

That would require something like this:

https://github.com/diesel-rs/diesel/blob/9861bcba0f48f64481bf7a312da2bb2d3ea2f208/diesel/src/pg/types/mod.rs#L211-L232

and 

https://github.com/diesel-rs/diesel/blob/9861bcba0f48f64481bf7a312da2bb2d3ea2f208/diesel/src/pg/types/ranges.rs#L276-L289

with the corresponding enum variants.

Afterwards the function signature needs to change so that it accepts only the SQL type (`RangeBoundEnum` in that example) instead of arbitrary text. (See here for an example usage: https://github.com/diesel-rs/diesel/blob/9861bcba0f48f64481bf7a312da2bb2d3ea2f208/diesel/src/pg/expression/functions.rs#L442)

weiznich (changes requested):
Thanks for opening this PR. The definition looks ok, but I would like to make the invariant for the `null_value_treatment` parameter a bit more robust by only accepting enum values with any of the allowed values there. I've left more detailed instructions for this a comment on the code itself. If you need help implementing that please let me know.

diesel/src/pg/expression/functions.rs
accestfromim: Thank you for your kind instructions.
I impl to_sql in a new mod called json_function_enum, do you think it's ok?
I'm ready to  make more changes if you think it's not fit to make a new mod for it.

weiznich (approved):
Looks good now, thanks for the fast update :+1:

diesel/src/pg/expression/functions.rs
weiznich: No it's good that way, thanks for implementing this change :heart

PR #4398, opened 2024-12-18
Remove unnecessary space from DELETE that causes duplication

### LINKED ISSUE #4397: DELETE command includes duplicate whitespace
### Versions

- **Rust:** 1.83.0
- **Diesel:** 2.2.6
- **Database:** SQLite
- **Operating System** Linux

### Feature Flags

- **diesel:** `r2d2`, `returning_clauses_for_sqlite_3_35`, `sqlite`, `uuid`

## Problem Description

SQL queries generated for DELETE commands include an unnecessary duplicate space between `DELETE` and `FROM`. This can be seen from emitting the corresponding strings in `walk_ast()`:

https://github.com/diesel-rs/diesel/blob/b705023d85e6ef76292a427aa70f21c28a5902ff/diesel/src/query_builder/delete_statement/mod.rs#L261-L267

The trailing space can safely be omitted here because `self.from_clause.walk_ast()` always contributes a leading space itself, leading to duplication, in its own `walk_ast()` method:

https://github.com/diesel-rs/diesel/blob/b705023d85e6ef76292a427aa70f21c28a5902ff/diesel/src/query_builder/from_clause.rs#L128-L131

Since `DELETE ` is always followed by ` FROM `—`self.from_clause` is not optional and must always be emitted—, the trailing space is not necessary, similar to how INSERT commands are prepared without trailing space (because they are always followed by ` INTO `):

https://github.com/diesel-rs/diesel/blob/b705023d85e6ef76292a427aa70f21c28a5902ff/diesel/src/query_builder/insert_statement/mod.rs#L488-L491

Issue clean, beautiful SQL queries.

`DELETE FROM "table"`

`DELETE  FROM "table"`

No.

### Steps to reproduce

Issue a DELETE query as in the following code:

```rust
let num_deleted = diesel::delete(posts.filter(title.like(pattern)))
    .execute(connection)
    .expect("Error deleting posts");
```

## Checklist

sgoll:
Fixed #4397.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ in a way that makes the pools suitable for use in parallel tests.
 * Added `Json` and `Jsonb` support for the SQLite backend.
 * Fixed diesel thinking `a.eq_any(b)` was non-nullable even if `a` and `b` were nullable.
 
+### Fixed
+
+* Use a single space instead of two spaces between `DELETE FROM`.
+
 ## [2.2.2] 2024-07-19
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


guissalustiano (approved):
LGTM, thanks!

PR #4401, opened 2024-12-20
Add SQLite function `json_type`

xuehaonan27:


prkbuilds (approved):
LGTM

LucaCappelletti94 (commented):
Maybe the doctests could get a bit trimmed, made more minimal and cleaned up, but that's about it. Looks okay otherwise.

weiznich:
My bad, it seems like I completely missed this PR back than. Sorry about that.

It seems like the same function was implemented by https://github.com/diesel-rs/diesel/pull/4511 as well and got merged there as well. Given that this is now a duplicate (even if this PR was the first one) I close this PR. Sorry again for missing it.

PR #4410, opened 2024-12-26
feat(cli): generate unique versions

### LINKED ISSUE #3698: When generating multiple migrations at once, subsequent migrations that contain the same time stamp are not run.
### Versions

- **Rust:** 1.71.0-nightly
- **Diesel:** 2.1.0
- **Database:** SQLite3
- **Operating System** PopOS

### Feature Flags

- **diesel:** sqlite, serde_json

## Problem Description
When generating multiple migrations, they may be generated with the same timestamp in the folder name.
Despite having different names following the timestamp, subsequent migrations containing the same timestamp after the first migration are ignored and will not be run.

Creating multiple migrations by chaining commands and being able to run the migrations without having to manually edit the timestamp of subsequent migrations.

Migrations are all run even if they contain the same timestamp.

Only the first migration containing a specific timestamp is run.

No errors were reported throughout the process.

### Steps to reproduce
Generate two migrations at the same time by chaining commands to generate two migrations at the same timestamp.

One could also manually edit the timestamp of the migrations to make two that are equal.

After there are two migrations, delete your database and run
```
diesel migration run
```

Only the first migration will be run and the following migration will be ignored with no error.

## Checklist

- [x ] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [ x] This issue can be reproduced on Rust's stable channel. (Your issue will be
- [ x] This issue can be reproduced without requiring a third party crate

AndreCostaaa:
Closes #3698

This PR adds:

1. Migration folder locking mechanism that can easily be added to other migration commands to make sure we don't have multiple diesel cli instances running migration instances
2. Checks to make sure we never generate two migrations with the same version
3.  Adds a "subversion" that gets added to the timestamp so we can generate multiple migrations per second
    - This only happens when the `--version` argument is not specified
    - Adding a subversion allows us to keep the timestamp intact instead of creating a fake timestamp by adding 1 to it

### DIFF
--- a/diesel_cli/src/errors.rs
+++ b/diesel_cli/src/errors.rs
@@ -64,6 +64,8 @@ pub enum Error {
     NoSchemaKeyFound(String),
     #[error("Failed To Run rustfmt")]
     RustFmtFail(String),
+    #[error("Failed To Acquire Migration Folder Lock")]
+    FailedToAcquireMigrationFolderLock(String),
 }
 
 fn print_optional_path(path: &Option<PathBuf>) -> String {
--- a/diesel_cli/src/errors.rs
+++ b/diesel_cli/src/errors.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::infer_schema_internals::TableName;
 
@@ -72,7 +72,7 @@ pub enum Error {
     DuplicateMigrationVersion(PathBuf, String),
 }
 
-fn print_path(path: &PathBuf) -> String {
+fn print_path(path: &Path) -> String {
     format!("{}", path.display())
 }
 fn print_optional_path(path: &Option<PathBuf>) -> String {
--- a/diesel_cli/src/errors.rs
+++ b/diesel_cli/src/errors.rs
@@ -66,6 +66,10 @@ pub enum Error {
     RustFmtFail(String),
     #[error("Failed To Acquire Migration Folder Lock")]
     FailedToAcquireMigrationFolderLock(String),
+    #[error("Couldn't Create Migration Folder")]
+    MigrationFolderCreationError,
+    #[error("Duplicate Migration Version Already Exists")]
+    DuplicateMigrationVersion,
 }
 
 fn print_optional_path(path: &Option<PathBuf>) -> String {
--- a/diesel_cli/src/errors.rs
+++ b/diesel_cli/src/errors.rs
@@ -18,7 +18,7 @@ pub enum Error {
     ProjectRootNotFound(PathBuf),
     #[error("The --database-url argument must be passed, or the DATABASE_URL environment variable must be set.")]
     DatabaseUrlMissing,
-    #[error("Encountered an IO error: {0} {n}", n=print_optional_path(.1))]
+    #[error("Encountered an IO error: {0} for `{n}`", n=print_optional_path(.1))]
     IoError(#[source] std::io::Error, Option<PathBuf>),
     #[error("Failed to execute a database query: {0}")]
     QueryError(#[from] diesel::result::Error),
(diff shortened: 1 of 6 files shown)


diesel_cli/src/migrations/mod.rs
```diff
@@ -85,6 +86,14 @@ pub(super) fn run_migration_command(matches: &ArgMatches) -> Result<(), crate::e
             println!("{result:?}");
         }
         ("generate", args) => {
+            let dir = migrations_dir(matches)?;
+            let mut lock = RwLock::new(migration_folder_lock(dir)?);
```
weiznich, line 90: I would rather like to clone the `dir` here and not call `migrations_dir` again in line 140. Also I would prefer to use a better fitting name, like for example `migration_dir` as variable name here.

diesel_cli/src/migrations/mod.rs
```diff
@@ -157,6 +169,89 @@ pub(super) fn run_migration_command(matches: &ArgMatches) -> Result<(), crate::e
...
+    }
+}
+
+fn create_migration_dir<'a>(
+    migrations_dir: PathBuf,
+    migration_name: &str,
+    version: Box<dyn Display + 'a>,
+    explicit_version: bool,
+) -> Result<PathBuf, crate::errors::Error> {
+    const MAX_MIGRATIONS_PER_SEC: u16 = 0xFFFF;
```
weiznich, line 197: Rather use 
```suggestion
    const MAX_MIGRATIONS_PER_SEC: u16 = u16::MAX;
```

diesel_cli/src/migrations/mod.rs
```diff
@@ -157,6 +169,89 @@ pub(super) fn run_migration_command(matches: &ArgMatches) -> Result<(), crate::e
...
+            }
+            None
+        })
+        .collect();
+
+    // if there's an explicit version try to use it
+    if explicit_version {
+        let version = format!("{version}");
+        if is_duplicate_version(&version, &migration_folders) {
+            return Err(crate::errors::Error::DuplicateMigrationVersion);
```
weiznich, line 237: It might be meaningful to include the version into the error message

weiznich (commented):
Thanks for working on this. Overall this looks reasonable, I've just have minor remarks about small things.

That written: We should have a changelog entry about this change. Could you add it as fix to the `CHANGELOG.md` file. That entry should state that old migrations just continue to work.

diesel_cli/src/migrations/mod.rs
AndreCostaaa: Agreed. Overlooked the code duplication in lines 89 and 140

diesel_cli/src/migrations/mod.rs
AndreCostaaa: Thanks

diesel_cli/src/migrations/mod.rs
AndreCostaaa: Agreed

diesel_cli/src/errors.rs
```diff
@@ -64,6 +64,12 @@ pub enum Error {
     NoSchemaKeyFound(String),
     #[error("Failed To Run rustfmt")]
     RustFmtFail(String),
+    #[error("Failed to acquire migration folder lock")]
+    FailedToAcquireMigrationFolderLock(String, PathBuf),
```
weiznich, line 68: The pathbuf needs to be included in the actual error message that's defined by the `#[error]` attribute. See the enum variants above for examples. The [thiserror documentation](https://docs.rs/thiserror/latest/thiserror/#details) has more details about how these format specifiers work.

weiznich (commented):
Thanks for the update. It's now almost good, we just need to include the paths information stored in the error enum in the actual error messages.

diesel_cli/src/errors.rs
AndreCostaaa: Hope you don't mind, I refactored the IOError variant aswell so i could more easily reuse the code

AndreCostaaa:
I think CI is failing because of the github actions migration of ubuntu-latest to ubuntu-24.04 [here](https://github.com/actions/runner-images/issues/10636)

weiznich (approved):
Looks good now, thanks for the update. 

I will merge this as soon as I've fixed the unrelated CI error.

PR #4411, opened 2024-12-29
Add wasm32-unknown-unknown target support

Spxg:
I recently created [a rust wrapper crate](https://docs.rs/sqlite-wasm-rs/0.1.0/sqlite_wasm_rs/) for [sqlite-wasm](https://github.com/sqlite/sqlite-wasm) and did some development with the C interface used by diesel.

Now the development is complete, diesel can run successfully on the `wasm32-unknown-unknown` platform. and diesel’s integration tests and unit tests all run successfully，except for a few tests that required `std::fs::*` and doc test (wasm-bindgen-test currently does not support this).
Branch: https://github.com/Spxg/diesel/tree/wasm32-test (There are many changes, not planning to merge them at the moment).

```bash
cargo install wasm-pack
cargo xtask run-tests --wasm
```

You will see output similar to this: https://raw.githubusercontent.com/Spxg/Spxg/refs/heads/master/resources/sqlitest.gif

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -24,7 +24,6 @@ include = [
 byteorder = { version = "1.0", optional = true }
 chrono = { version = "0.4.20", optional = true, default-features = false, features = ["clock", "std"] }
 libc = { version = "0.2.0", optional = true }
-libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled_bindings"] }
 mysqlclient-sys = { version = ">=0.2.5, <0.5.0",  optional = true }
 mysqlclient-src = { version = "0.1.0", optional = true }
 pq-sys = { version = ">=0.4.0, <0.7.0", optional = true }
@@ -50,6 +49,12 @@ downcast-rs = "1.2.1"
 version = "~2.2.0"
 path = "../diesel_derives"
 
+[target.'cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))'.dependencies]
+libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled_bindings"] }
+
+[target.'cfg(all(target_family = "wasm", not(target_os = "wasi")))'.dependencies]
+sqlite-wasm-rs = { version = "0.1.0", optional = true }
+
 [dev-dependencies]
 cfg-if = "1"
 dotenvy = "0.15"
@@ -67,7 +72,7 @@ huge-tables = ["64-column-tables"]
 64-column-tables = ["32-column-tables", "diesel_derives/64-column-tables"]
 128-column-tables = ["64-column-tables", "diesel_derives/128-column-tables"]
 postgres = ["dep:pq-sys", "postgres_backend"]
-sqlite = ["dep:libsqlite3-sys", "diesel_derives/sqlite", "time?/formatting", "time?/parsing"]
+sqlite = ["dep:libsqlite3-sys", "dep:sqlite-wasm-rs", "diesel_derives/sqlite", "time?/formatting", "time?/parsing"]
 mysql = ["dep:mysqlclient-sys", "dep:url", "dep:percent-encoding", "dep:bitflags", "mysql_backend"]
 without-deprecated = ["diesel_derives/without-deprecated"]
 with-deprecated = ["diesel_derives/with-deprecated"]
--- a/diesel/src/lib.rs
+++ b/diesel/src/lib.rs
@@ -768,6 +768,13 @@ pub mod prelude {
     #[cfg(feature = "sqlite")]
     #[doc(inline)]
     pub use crate::sqlite::SqliteConnection;
+    // Before using C-API, you must initialize sqlite.
+    //
+    // Initializing the database is a one-time operation during
+    // the life of the program.
+    #[cfg(feature = "sqlite")]
+    #[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+    pub use sqlite_wasm_rs::init_sqlite;
 }
 
 #[doc(inline)]
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -1,4 +1,8 @@
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 mod bind_collector;
 mod functions;
--- a/diesel/src/sqlite/connection/bind_collector.rs
+++ b/diesel/src/sqlite/connection/bind_collector.rs
@@ -3,6 +3,10 @@ use crate::serialize::{IsNull, Output};
 use crate::sql_types::HasSqlType;
 use crate::sqlite::{Sqlite, SqliteType};
 use crate::QueryResult;
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
+use libsqlite3_sys as ffi;
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 #[derive(Debug, Default)]
 pub struct SqliteBindCollector<'a> {
@@ -127,9 +131,8 @@ impl InternalSqliteBindValue<'_> {
     #[allow(unsafe_code)] // ffi function calls
     pub(in crate::sqlite) fn result_of(
         self,
-        ctx: &mut libsqlite3_sys::sqlite3_context,
+        ctx: &mut ffi::sqlite3_context,
     ) -> Result<(), std::num::TryFromIntError> {
-        use libsqlite3_sys as ffi;
         use std::os::raw as libc;
         // This unsafe block assumes the following invariants:
         //
--- a/diesel/src/sqlite/connection/functions.rs
+++ b/diesel/src/sqlite/connection/functions.rs
@@ -1,4 +1,8 @@
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 use super::raw::RawConnection;
 use super::row::PrivateSqliteRow;
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -1,5 +1,9 @@
 #![allow(unsafe_code)] // ffi calls
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 use std::ffi::{CString, NulError};
 use std::io::{stderr, Write};
@@ -45,8 +49,23 @@ impl RawConnection {
             CString::new(database_url)?
         };
         let flags = ffi::SQLITE_OPEN_READWRITE | ffi::SQLITE_OPEN_CREATE | ffi::SQLITE_OPEN_URI;
+        // Persistent Storage is supported, use opfs vfs.
+        // This support is only available when sqlite is loaded from a
+        // Worker thread, whether it's loaded in its own dedicated worker
+        // or in a worker together with client code.
+        #[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+        let vfs = CString::new("opfs")?;
+
         let connection_status = unsafe {
-            ffi::sqlite3_open_v2(database_url.as_ptr(), &mut conn_pointer, flags, ptr::null())
+            ffi::sqlite3_open_v2(
+                database_url.as_ptr(),
+                &mut conn_pointer,
+                flags,
+                #[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
+                ptr::null(),
+                #[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+                vfs.as_ptr(),
+            )
         };
 
         match connection_status {
--- a/diesel/src/sqlite/connection/serialized_database.rs
+++ b/diesel/src/sqlite/connection/serialized_database.rs
@@ -1,5 +1,9 @@
 #![allow(unsafe_code)]
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 use std::ops::Deref;
 
--- a/diesel/src/sqlite/connection/sqlite_value.rs
+++ b/diesel/src/sqlite/connection/sqlite_value.rs
@@ -1,5 +1,9 @@
 #![allow(unsafe_code)] // ffi calls
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 
 use std::cell::Ref;
 use std::ptr::NonNull;
--- a/diesel/src/sqlite/connection/stmt.rs
+++ b/diesel/src/sqlite/connection/stmt.rs
@@ -8,7 +8,10 @@ use crate::query_builder::{QueryFragment, QueryId};
 use crate::result::Error::DatabaseError;
 use crate::result::*;
 use crate::sqlite::{Sqlite, SqliteType};
+#[cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))]
 use libsqlite3_sys as ffi;
+#[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+use sqlite_wasm_rs::c as ffi;
 use std::cell::OnceCell;
 use std::ffi::{CStr, CString};
 use std::io::{stderr, Write};


insipx:
nice, this seems to be cleaner than re-implementing the backend in a separate crate, and maybe better than storing them on function pointers like in [the discussion](https://github.com/diesel-rs/diesel/discussions/4232)?

How would SQLite updates work with this. Further, it would be most efficient to implement a VFS in rust rather than go to JS and back, would this wrapper crate allow for that from diesel?

Spxg:
> nice, this seems to be cleaner than re-implementing the backend in a [separate crate](https://github.com/xmtp/sqlite-web-rs), and maybe better than storing them on function pointers like in [the discussion](https://github.com/diesel-rs/diesel/discussions/4232)?
> 
> How would SQLite updates work with this. Further, it would be most efficient to implement a VFS in rust rather than go to JS and back, would this wrapper crate allow for that from diesel?

The current implementation of `sqlite-wasm-rs` is actually the same as what the "JS side" in the `sqlite-wasm` project does (using various wrappers to implement calls, including type conversion and wasm memory allocation).

It would be most efficient if a `VFS` could be implemented in rust and registered using `sqlite3_vfs_register`. It can be implemented directly in `libsqlite3-sys`, and ORM frameworks such as `diesel` and `sqlx` do not even need to modify the implementation.

However, I know very little about VFS used by the web, such as OPFS. The opfs used by `sqlite-wasm` is not intended to be exported as an API (https://github.com/sqlite/sqlite/blob/04364cb3cc108a044a0d9dc7162f4d550adb2f99/ext/wasm/api/sqlite3-vfs-opfs.c-pp.js#L54), but the `wa-sqlite` project has other VFS for reference (https://github.com/rhashimoto/wa-sqlite/tree/master/src/examples). 

In short, if there is a VFS solution, I am happy to try it.

diesel/src/lib.rs
```diff
@@ -768,6 +768,13 @@ pub mod prelude {
     #[cfg(feature = "sqlite")]
     #[doc(inline)]
     pub use crate::sqlite::SqliteConnection;
+    // Before using C-API, you must initialize sqlite.
+    //
+    // Initializing the database is a one-time operation during
+    // the life of the program.
+    #[cfg(feature = "sqlite")]
+    #[cfg(all(target_family = "wasm", not(target_os = "wasi")))]
+    pub use sqlite_wasm_rs::init_sqlite;
```
weiznich, line 777: I would rather not expose this function from diesel as this would couple diesel's API stability to that one of `sqlite_wasm_rs`. Ideally we would just call that function internally one, or even better `sqlite_wasm_rs` would provide a singelton helper function to do this. 

We do something similar to this for the mysql backend already: 

https://github.com/diesel-rs/diesel/blob/3f69aa4570c505a94889e3c3de4e83e13bc6a37d/diesel/src/mysql/connection/raw.rs#L251-L277

diesel/Cargo.toml
```diff
@@ -50,6 +49,12 @@ downcast-rs = "1.2.1"
 version = "~2.2.0"
 path = "../diesel_derives"
 
+[target.'cfg(not(all(target_family = "wasm", not(target_os = "wasi"))))'.dependencies]
+libsqlite3-sys = { version = ">=0.17.2, <0.31.0", optional = true, features = ["bundled_bindings"] }
+
+[target.'cfg(all(target_family = "wasm", not(target_os = "wasi")))'.dependencies]
```
weiznich, line 55: I'm wondering whether this predicate essentially boils down to `target = "wasm32-unknown-unknow` in practice. Are you aware of any other target that fulfills this constraints?

weiznich (commented):
Thanks for opening this PR. Overall I'm very much in favor of this approach as it looks really clean in terms of code changes. I've added two questions about some details.

That written: I personally would prefer to run at least the tests in CI to ensure that this does not break in the future.

As for the VFS discussion: I can imaging that diesel exposes the necessary traits to implement a custom VFS. A quick search on crates.io brought up https://docs.rs/sqlite-vfs/latest/sqlite_vfs/, which might serve as basis for such an implementation.

diesel/Cargo.toml
Spxg: > I'm wondering whether this predicate essentially boils down to `target = "wasm32-unknown-unknow` in practice. Are you aware of any other target that fulfills this constraints?

I will change it to `#[cfg(all(target_family = "wasm", target_os = "unknown"))]`, which is recommended by the rustc book:
https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-unknown.html#conditionally-compiling-code

diesel/src/lib.rs
Spxg: > I would rather not expose this function from diesel as this would couple diesel's API stability to that one of `sqlite_wasm_rs`. Ideally we would just call that function internally one, or even better `sqlite_wasm_rs` would provide a singelton helper function to do this.
> 
> We do something similar to this for the mysql backend already:
> 
> https://github.com/diesel-rs/diesel/blob/3f69aa4570c505a94889e3c3de4e83e13bc6a37d/diesel/src/mysql/connection/raw.rs#L251-L277

Yes, the best way is to implement a singleton function call internally. But unlike `mysql_server_init`, `init_sqlite` is asynchronous, and since the `diesel` function is called synchronously, there is no opportunity to execute `init_sqlite`.
Any good suggestions for this?

Spxg:
> That written: I personally would prefer to run at least the tests in CI to ensure that this does not break in the future.

I agree with you, but in order to solve the differences between native and wasm tests, I created the [`#[test_derive::test]`](https://github.com/Spxg/diesel/blob/dc66375512b43e94c8a111f50b1ae1ff2dda996f/diesel_tests/test_derive/src/lib.rs#L8) macro to do this. Is it acceptable to replace `#[test]` with `#[test_derive::test]` in future maintenance?


In addition, some changes are required to run rustdoc tests

insipx:
> It would be most efficient if a VFS could be implemented in rust and registered using sqlite3_vfs_register. It can be implemented directly in libsqlite3-sys, and ORM frameworks such as diesel and sqlx do not even need to modify the implementation.

> However, I know very little about VFS used by the web, such as OPFS. The opfs used by sqlite-wasm is not intended to be exported as an API (https://github.com/sqlite/sqlite/blob/04364cb3cc108a044a0d9dc7162f4d550adb2f99/ext/wasm/api/sqlite3-vfs-opfs.c-pp.js#L54), but the wa-sqlite project has other VFS for reference (https://github.com/rhashimoto/wa-sqlite/tree/master/src/examples).


The VFS Implementation is only interesting for web because of the variety of ways to accomplish a VFS, the immaturity of OPFS vs IndexedDb, as well as the newness of SQLite Wasm build itself. Therefore (and maybe my opinion only) it is useful to expose methods for a flexible VFS implementation rather than forcing a specific implementation lower in the stack. At least, until standards and best practices are formed. 

Off the top, certain VFS implementations may play better with multiple open tabs to the same db, for instance, while other may completely disallow that kind of behavior.


I'm satisfied with Weiznich's suggestion, though, didn't even know about that crate! An API to simply register the vfs is all that would be needed

diesel/src/lib.rs
insipx: if this is exclusively for the web, you can make some assumptions about the environment and spawn the `init` future onto the web executor rather than immediately wait for it. I think this makes sense, and maybe you can even rename `init_sqlite` to `init_sqlite_web_lazy`. The function should return a handle to the spawned future to allow awaiting it to completion.

Diesel can store this future wherever it likes. Then maybe add a new async fn to diesel to finish waiting for the spawned future handle. This should decouple the api and let the user wait for sqlite init.

Ideally we could avoid offloading the `init` to the user, but since diesel is pretty much 100% synchronous not sure how that could work

Spxg: > spawn the `init` future onto the web executor rather than immediately wait for it.

Spawn `init_sqlite` means that `init_sqlite` may not be called before sqlite ffi is called and `wasm_bindgen_future::spawn` relies on `event loop`, execution needs to wait until the main thread is idle.
Anyway this may cause the call to sqlite ffi to fail.

As you said, since `diesel` is now fully synchronous, there is no way to initialize sqlite in diesel.
I think there are only two ways to do it, called by the user:
1. ~export `sqlite_wasm_rs::init_sqlite`~
2. Wrapper an `async` init function in `diesel` and call `sqlite_wasm_rs::init_sqlite`

insipx: A method would be exposed in diesel to allow the user to await the spawned future, so init or SQLite ffi would not fail. It would only fail if the user fails to init sqlite, which is no different then the current implementation. As a user of diesel, the  spawn strategy would not appear any different then exposing `init_sqlite` from `sqlite_wasm`.


The difference between the two is a decoupling between diesel and `sqlite_wasm_rs` and diesel

weiznich: As written above: My main concern for just reexporting this function is the API stability of the `sqlite_wasm_rs` crate. I'm not sure who is in charge of maintaining it, so I'm also not sure how future updates are planed.

For diesel it would be problematic if some of the functions used by diesel change between versions. If that's not planed it's even fine to issue new major version releases of this crate, as diesel could just support a version range then. If one of the functions used by diesel changes between versions it is essentially impossible for diesel to support the new `sqlite_wasm_rs` version without breaking its (diesels) semver promise as this exported function is then part of the public API. The other problematic point is: Even if we don't reexport that function the same public dependency relation ship would exist due to being a essentially a `-sys` crate. After all you need to call this function for the right version of `sqlite_wasm_rs`, right?

weiznich:
> I agree with you, but in order to solve the differences between native and wasm tests, I created the [#[test_derive::test]](https://github.com/Spxg/diesel/blob/dc66375512b43e94c8a111f50b1ae1ff2dda996f/diesel_tests/test_derive/src/lib.rs#L8) macro to do this. Is it acceptable to replace #[test] with #[test_derive::test] in future maintenance?

I would prefer to handle this via `#[cfg_attr(target = "wasm", wasm_bindgen::test)]` and `#[cfg_attr(not(target = "wasm"), test)]` attributes instead if possible.

>  In addition, some changes are required to run rustdoc tests

I think it would be fine to just run unit and integration tests and skip the doc tests on this target.

diesel/src/lib.rs
insipx: FWIW its not impossible to make this sync, although it is more involved:
Looking through `sqlite/sqlite-wasm` code for `sqlite3InitModule`

1.) `sqlite/sqlite-wasm` offers a way to pass an already-instantiated wasm module to `sqlite3InitModule` [here](https://github.com/sqlite/sqlite-wasm/blob/1f3f5cafe8a7a86cc9801069c7bea849c07625d2/sqlite-wasm/jswasm/sqlite3-bundler-friendly.mjs#L406) 
2.) Initializing a WebAssembly module by-default is async, but [it can be done syncronously](https://rustwasm.github.io/wasm-bindgen/examples/synchronous-instantiation.html), and we can do it from rust and pass it to `sqlite/sqlite-wasm` from rust, using the [InitOpts object](https://github.com/Spxg/sqlite-wasm-rs/blob/bdf4f745a9990f482d256036b10984cdfea485bc/src/lib.rs#L53). This InitOpts object is the same exact way [sqlite-web-rs](https://github.com/xmtp/sqlite-web-rs) controls the same params, so I could also try it out there and report back on how it effects things
3.) Create an init that mirrors the `mysql_server_init`

- may require minor modifications to `sqlite/sqlite-wasm` function

Instantiating the WASM module sync should be fine in this case b/c using OPFS needs to be done in a dedicated worker anyway, so it won't block the main thread. 

That should handle the async part of `sqlite3InitModule`, which I think is the bulk of it.




> The other problematic point is: Even if we don't reexport that function the same public dependency relation ship would exist due to being a essentially a -sys crate. After all you need to call this function for the right version of sqlite_wasm_rs, right?

Is this problematic only in the case of using `sqlite_wasm_rs` and `diesel`, or even if there is a `mysql-like` init?

Spxg: > As written above: My main concern for just reexporting this function is the API stability of the `sqlite_wasm_rs` crate. I'm not sure who is in charge of maintaining it, so I'm also not sure how future updates are planed.

`sqlite-wasm-rs` is maintained by me, and its purpose is to wrap `sqlite-wasm` and provide the same `C-Like` API as `libsqlite3-sys`.
1. About the C-Like API, since the SQLite C API is stable, there will be no breaking change in this C-Like API.
2. About `init_sqlite`, since `sqlite.wasm` must be initialized, `init_sqlite` is also stable. Initialization is now an asynchronous `OnceCell`. If there is a synchronous initialization method later, `init_sqlite` can run a noop, and the initialization will be called internally (when calling the C-Like API).

So there will be no breaking change in the future. The only problem is that if diesel use a new SQLite C API when developing new features, but `sqlite-wasm-rs` has not yet been wrapped, it will cause compilation failure. 

Fortunately, the cost of wrapping a new API is not high, and I can support it quickly.

Spxg:
@weiznich I'm ready. I originally expected the test to be merged with this PR, but the changes are too big and some CIs fail unexpectedly. I'm going to open a separate PR to merge the test CI.

The good news is that the test works fine: https://github.com/diesel-rs/diesel/actions/runs/12696215996/job/35389791579?pr=4411

Spxg:
I'm ready for review.

1. Regarding the API stability issue before, I wrapper an `export` module inside `sqlite-wasm-rs` and exposed it to diesel, which provides stable APIs.
2. `td::test` macro is created. If it is a wasm platform test, it is replaced with `wasm_bindgen_test::test`. If the feature is `sqlite`, the `init_sqlite` step will be added before the test.
3. OPFS is no longer provided by default, because the conditions for using opfs are relatively strict. Currently, MemVFS is the default. Which vfs to use is determined by the caller. When using it, the url is passed in `file:xxxx.db?vfs=[opfs|opfs-sahpool|xxxx]`.
4. Currently, the test can be run through` cargo xtask run-tests --wasm` (welcome to experience, but you need to install wasm-pack first)

The test has been successfully run in CI, but some CIs failed unexpectedly. I am not sure if it was caused by my changes.

@weiznich @insipx

weiznich:
> The test has been successfully run in CI, but some CIs failed unexpectedly. I am not sure if it was caused by my changes.

The CI failures are unrelated. They are caused by the new rust release and some changes to the CI runners. I've filled #4428 to address that. You likely need to rebase this PR after that one is merged.


I will do another round of reviews later today.

xtask/src/tests/mod.rs
```diff
@@ -83,6 +86,37 @@ impl TestArgs {
...
+                    .stderr(Stdio::inherit())
+                    .stdout(Stdio::inherit())
+                    .status()
+                    .unwrap();
+                if !out.success() {
+                    eprintln!("Failed to run wasm integration tests");
+                    return false;
+                }
+                return true;
+            } else {
```
weiznich, line 116: Can we emit a helpful error message that says that only sqlite is supported for now?

weiznich (commented):
Thanks for the update. Overall I think we are nearly there. I've noted a few minor organizational things that shouldn't be too hard to fix.

I'm also not happy with the test attribute crate name, sorry about that but I would really like to see a better/longer name there. 

Beside these specific comments I think it would be great to have:

* A short documentation entry in the sqlite connection documentation that states `wasm32` is supported, outlines that you need to call the init function + contains some details about the vfs choices
* A changelog entry as this is a rather large new feature

Spxg:
> I'm also not happy with the test attribute crate name, sorry about that but I would really like to see a better/longer name there.

I totally agree with using a longer and clearer test attribute crate name. I also think diesel_test_helper needs to be in the root of the workspace, and it's a bit strange to put it in the diesel_tests dir, since the diesel crate also uses it.

Thanks for taking the time to review, I will fix these comments tonight (UTC +8)

Spxg:
All comments fixed, additional changes:
1. Added several sqlite-wasm vfs tests
2. Run tests in "Check sqlite wasm" CI directly specify sqlite backend
@weiznich

diesel/src/lib.rs
```diff
@@ -773,6 +778,11 @@ pub mod prelude {
     #[cfg(feature = "sqlite")]
     #[doc(inline)]
     pub use crate::sqlite::SqliteConnection;
+
+    // These exported API from `sqlite-wasm-rs` are stable:
+    #[cfg(feature = "sqlite")]
+    #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+    pub use sqlite_wasm_rs::export::init_sqlite;
```
weiznich, line 785:
```suggestion
    #[doc(inline)]
    pub use sqlite_wasm_rs::export::init_sqlite;
```

xtask/src/tests/mod.rs
```diff
@@ -83,6 +86,40 @@ impl TestArgs {
...
+                eprintln!(
+                    "Only sqlite backend supports wasm for now, the current backend is {backend}"
```
weiznich, line 118:
```suggestion
                    "Only the sqlite backend supports wasm for now, the current backend is {backend}"
```

diesel/src/sqlite/connection/mod.rs
```diff
@@ -164,6 +168,17 @@ impl Connection for SqliteConnection {
...
+    /// The following are notes for use under `wasm32-unknown-unknown` target:
+    ///
+    /// Must initialize sqlite using `diesel::init_sqlite`
+    /// before calling `SqliteConnection::establish`.
+    ///
+    /// Under sqlite-wasm, database is stored in memory by default. sqlite-wasm
+    /// provides some persistent VFS. But they all have some limitations,
+    /// see <https://sqlite.org/wasm/doc/trunk/persistence.md>
+    ///
+    /// VFS can be selected through the `database_url`, such as `file:data.db?vfs=opfs`.
```
weiznich, line 181:
```suggestion
    /// 
    /// ## WASM support
    ///
    /// If you plan to use this connection type on the `wasm32-unknown-unknown` target please
    /// make sure to read the following notes:
    ///
    /// * You must  initialize sqlite using `diesel::init_sqlite` before calling `SqliteConnection::establish`. 
    /// * The database is stored in memory by default. sqlite-wasm
    ///     provides different persistent VFS (Virtual File Systems), but they all have different limitations.
    ///     See <https://sqlite.org/wasm/doc/trunk/persistence.md> for details. Make sure to chose
    ///     an appropriated VFS implementation for your usecase.
    /// * VFS can be selected through the `database_url` via an URL option, such as `file:data.db?vfs=opfs`.
```

weiznich (approved):
Thanks for the update and thanks again for working on this :heart: 

I think it's now ready to be merged. I've adjusted some documentation comments slightly to better fit the over all documentation format. I will just apply the suggestions for that and merge the PR afterwards, so no additional work from your side is required here.

insipx:
> OPFS is no longer provided by default, because the conditions for using opfs are relatively strict. Currently, MemVFS is the default. The type of vfs to use can be specified through the url, like file:xxxx.db?vfs=[opfs|opfs-sahpool|xxxx]

Love it! 

Thanks for your work on this, with the next diesel release it should make [sqlite-web-rs](https://github.com/xmtp/sqlite-web-rs) redundant since everything is provided in mainline diesel, assuming the sqlite-wasm-rs is kept up to date ! 

Excited to start experimenting with different rust-native VFS's!

Spxg:
Thanks for your suggestions on this PR and the `sqlite-web-rs` project ❤️ @weiznich @insipx

WillPapper:
This is huge @Spxg! Thank you for all of your work here. Very helpful for many people!

Spxg:
sqlite-wasm-rs will have its first major update, controlled by `polyfill` features:
1. No wrap sqlite-wasm
2. No need for init_sqlite()
2. Implement a vfs using rust
3. Pure rust implementation

I have basically completed it, and now I can experience sqlite in a "native way"(no wrapper). It works fine now, but needs more time to test. As usual, all tests of diesel have been run successfully.

```toml
[dependencies]
sqlite-wasm-rs = { git = "https://github.com/Spxg/sqlite-wasm-rs", branch = "polyfill" }
```

```rust
use sqlite_wasm_rs::export::*;

async fn test_opfs_sah_vfs_default() {
    // install persistent vfs
    // you can install OPFS-Explorer in your browser to view data
    install_opfs_sahpool(None, true).await.unwrap();

    let filename = cstr("test_opfs_sah_vfs_default.db");
    let mut db = std::ptr::null_mut();
    let ret = unsafe {
        sqlite3_open_v2(
            filename.as_ptr(),
            &mut db as *mut _,
            SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE,
            std::ptr::null_mut(),
        )
    };
    
    assert_eq!(SQLITE_OK, ret);
}
```

@insipx @weiznich @WillPapper

PR #4417, opened 2025-01-02
DebugBinds/Query: Reduce monomorphization overhead

Turbo87:
The output of `cargo llvm-lines --package diesel_tests --test integration_tests --profile test --features postgres` showed the `Debug` and `Display` implementations of the `DebugBinds` and `DebugQuery` structs as two of the main code size reasons, slowing down the compilation of applications. This commit significantly reduces the LLVM IR lines of the two structs, which should hopefully improve compile times a little bit (see commit messages for detailed numbers).

/cc @weiznich 

Related:

- https://matklad.github.io/2021/09/04/fast-rust-builds.html#Keeping-an-Eye-on-Instantiations

### DIFF
--- a/diesel/src/query_builder/debug_query.rs
+++ b/diesel/src/query_builder/debug_query.rs
@@ -25,6 +25,33 @@ impl<'a, T, DB> DebugQuery<'a, T, DB> {
             _marker: PhantomData,
         }
     }
+
+    pub(crate) fn query(&self) -> Result<String, fmt::Error>
+    where
+        DB: Backend + Default,
+        DB::QueryBuilder: Default,
+        T: QueryFragment<DB>,
+    {
+        let mut query_builder = DB::QueryBuilder::default();
+        let backend = DB::default();
+
+        // This is not using the `?` operator to reduce the code size of this
+        // function, which is getting copies a lot due to monomorphization.
+        if QueryFragment::<DB>::to_sql(self.query, &mut query_builder, &backend).is_err() {
+            return Err(fmt::Error);
+        }
+
+        Ok(query_builder.finish())
+    }
+
+    pub(crate) fn binds(&self) -> DebugBinds<'_, T, DB>
+    where
+        DB: Backend + Default,
+        DB::QueryBuilder: Default,
+        T: QueryFragment<DB>,
+    {
+        DebugBinds::<_, DB>::new(self.query)
+    }
 }
 
 impl<T, DB> Display for DebugQuery<'_, T, DB>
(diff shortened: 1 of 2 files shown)


diesel/src/query_builder/debug_query.rs
```diff
@@ -34,12 +61,15 @@ where
...
-        QueryFragment::<DB>::to_sql(self.query, &mut query_builder, &backend)
-            .map_err(|_| fmt::Error)?;
-        let debug_binds = DebugBinds::<_, DB>::new(self.query);
-        write!(f, "{} -- binds: {:?}", query_builder.finish(), debug_binds)
+        // This is not using the `?` operator to reduce the code size of this
+        // function, which is getting copies a lot due to monomorphization.
+        let query = match self.query() {
+            Ok(query) => query,
+            Err(err) => return Err(err),
+        };
```
Turbo87, line 69: I tried to use a `tri!()` macro instead (see https://github.com/serde-rs/serde/blob/v1.0.217/serde/src/lib.rs#L290-L301), but that slightly bumped the numbers upwards again, so I'm not sure whether it's worth it for the readability benefit.

weiznich:
Thanks for opening this PR. Such optimizations are really appreciated. I wonder whether it might be meaningful to use a similar strategy here to what we already do for the statement cache:

https://github.com/diesel-rs/diesel/blob/3f69aa4570c505a94889e3c3de4e83e13bc6a37d/diesel/src/connection/statement_cache/mod.rs#L252-L287

That would mean that instead of storing a reference to the generic query type we would rather just store a reference to a trait object in `DebugQuery` and `DebugBinds`, which should hopefully allow to minimize the number of generated lines of code even further. If you have time and motivation to try that this would be really appreciated, otherwise it shouldn't be a problem to merge this PR as it is.

Turbo87:
> instead of storing a reference to the generic query type we would rather just store a reference to a trait object in `DebugQuery` and `DebugBinds`

yeah, I thought about that as well, but I wasn't sure whether the performance overhead of the `Box` would be acceptable (or how to measure it... 😅)

Turbo87:
hmm, in general this seems to work, but https://github.com/diesel-rs/diesel/blob/3f69aa4570c505a94889e3c3de4e83e13bc6a37d/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs isn't particularly happy about the removal of the generic `T` parameter. any advice on how to deal with this special case?

weiznich:
Thanks for the update. I think removing the generic `T` type is unfortunately not possible, as that would be a breaking change. What we can do is to just put that type into a `PhantomData` and then have all the methods on that generic type call some static method without the generic parameter `T` internally. That should reduce the amount of duplication to the minimum while keeping things stable to the outside.

For a future diesel 3.0 version we might want to remove the `T` type there, but that's something for the future.,

Turbo87:
> just put that type into a `PhantomData`

that didn't work since the code in https://github.com/diesel-rs/diesel/blob/3f69aa4570c505a94889e3c3de4e83e13bc6a37d/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs is accessing `self.query.foo` directly, but I've found another way to shrink it down :)

Turbo87:
total LLVM IR lines as reported by the above command:

- before: 2_024_074
- after: 1_740_806 (-14%)

Turbo87:
> /usr/bin/ld: cannot find -lsqlite3: No such file or directory

that seems unrelated to me changes 😅

weiznich:
>>  /usr/bin/ld: cannot find -lsqlite3: No such file or directory
>
> that seems unrelated to my changes 😅

Yes, seems like the CI image got updated and something is now missing. I need to fix that :disappointed:

PR #4420, opened 2025-01-04
Add support for Postgres notifications

thkoch2001:
This allows to receive notifications through Postgres' LISTEN/NOTIFY mechanism.

https://github.com/diesel-rs/diesel/discussions/4418

I'm a rust newby, so please be frank about any possible errors I've made.

Especially I'm not sure whether I should somehow clone the internal_connection for the NotificationsIterator or if the user must drop the iterator before the connection can be used for other stuff.

I've not yet implemented blocking for new notifications as I don't need it and I think code should only be written when it is needed.


diesel/src/pg/connection/mod.rs
```diff
@@ -652,24 +654,31 @@ mod tests {
...
+            .execute(conn)
+            .unwrap();
+        sql_query("NOTIFY test_notifications, 'second'")
+            .execute(conn)
+            .unwrap();
+
+        let (first, second, third) = {
+            let mut iter = conn.notifications_iter();
+            (iter.next().unwrap(), iter.next().unwrap(), iter.next())
+        };
```
weiznich, line 670: Can we use `let mut notifications = conn.notifications_iter().collect::<Vec<_>>()` here instead and check that this vector contains exactly two elements. We could then try to fetch the next notification via `conn.notifications_iter().next()` to check that there is no notification.

diesel/src/pg/connection/mod.rs
```diff
@@ -547,6 +547,29 @@ impl PgConnection {
             .set_notice_processor(noop_notice_processor);
         Ok(())
     }
+
+    /// See Postgres documentation for SQL commands Notify and Listen
```
weiznich, line 551: It would be good to link the postgres documentation here

It would be even better to have the code from the example as example code in the doc tests.

diesel/src/pg/connection/mod.rs
```diff
@@ -547,6 +547,29 @@ impl PgConnection {
...
+
+pub use raw::PGNotification;
+
+impl Iterator for NotificationsIterator<'_> {
+    type Item = PGNotification;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.conn.pqnotifies()
+    }
 }
```
weiznich, line 573: I would prefer moving all notifications related code into a separate submodule to keep the code clean

diesel/src/pg/connection/raw.rs
```diff
@@ -183,6 +183,41 @@ impl RawConnection {
...
+// Using the same field names as tokio-postgres
+/// See Postgres documentation for SQL Commands NOTIFY and LISTEN
+#[derive(Clone, Debug)]
+pub struct PGNotification {
+    /// process ID of notifying server process
+    pub process_id: i32,
+    /// Name of the notification channel
+    pub channel: String,
+    /// optional data that was submitted with the notification, empty string if no data was submitted
+    pub payload: String,
```
weiznich, line 220: If it's optional I would rather prefer to use a `Option<String>` here.

diesel/src/pg/connection/raw.rs
```diff
@@ -183,6 +183,41 @@ impl RawConnection {
             ))
         }
     }
+
+    pub(super) fn pqnotifies(&self) -> Option<PGNotification> {
+        let conn = self.internal_connection;
+        let _ = unsafe { PQconsumeInput(conn.as_ptr()) };
```
weiznich, line 189: we should check the returned integer here to check if an error occurred. That likely means that the notification iterator returns a `QueryResult<PgNotification>` instead of just a `PgNotification>`.

It should make it also easier to handle the possible UTF-8 errors below.

diesel/src/pg/connection/raw.rs
```diff
@@ -183,6 +183,41 @@ impl RawConnection {
...
+            unsafe { PQfreemem(pgnotify as *mut std::ffi::c_void) };
+            ret
+        }
+    }
+}
+
+// Using the same field names as tokio-postgres
+/// See Postgres documentation for SQL Commands NOTIFY and LISTEN
+#[derive(Clone, Debug)]
+pub struct PGNotification {
```
weiznich, line 214:
```suggestion
pub struct PgNotification {
```

weiznich (changes requested):
Thanks for starting to work on this. This already looks great.

I've added a few comments about mostly stylistic and documentation related things.  Let me know if you need help to address them

Before merging we should also add an `Changelog.md` entry for this as this is a relevant new feature.

diesel/src/pg/connection/mod.rs
thkoch2001: With that I'd appreciate your help or better you doing it, please. It's much faster for you to just move the code where you want it. It's hard for me to balance everything what goes into API design and to predict your appreciation of the different factors.

diesel/src/pg/connection/raw.rs
thkoch2001: ups, sorry. Done.

diesel/src/pg/connection/mod.rs
thkoch2001: Done adding the links to NOTIFY/LISTEN SQL commands documentation. However I couldn't build `cargo doc --no-deps` to test the formatting: "failed to run custom build command for `mysqlclient-src v0.1.2`". I guess I'd have to set up the build for mysql to. I already installed cmake but gave up on missing SASL and LDAP libs. And running rustdoc on just mod.rs fails on missing type information from other files.

I'm not sure what the benefit is of moving the example code here and making it a doctest:
- I don't know how to ensure a db connection during doctest execution
- the example code does not add coverage to the already existing test
- rustdoc has [Scraped examples](https://doc.rust-lang.org/rustdoc/scraped-examples.html) functionality which should include the example in the docs, shouldn't it?

diesel/src/pg/connection/raw.rs
thkoch2001: I'm not sure how I could nicely convert the UTF-8 errors into DatabaseError.

weiznich: I would rather use the [`DeserializationError`](https://docs.diesel.rs/2.2.x/diesel/result/enum.Error.html#variant.DeserializationError) variant for that.

diesel/src/pg/connection/raw.rs
weiznich: In that case this seems to be the only reasonable approach to handle this. It's not great but we cannot do better if Postgres doesn't give us a way to differentiate beten empty and not set. Maybe the doc comment should include that information?

diesel/src/pg/connection/mod.rs
weiznich: I will take care of that in the next days, no worries

diesel/src/pg/connection/mod.rs
weiznich: Hmm, it should be possible to just run the Postgres tests via `cargo xtask run-tests posgres` without needing to care about any MySQL related dependencies. There is also a flag there to only run doc tests. If it doesn't work for you don't worry to much about this, I can take care of moving the example. 

Scraped example unfortunately only work for in crate examples, not for such provided as external crate.

weiznich (approved):

PR #4432, opened 2025-01-11
Add a sqlite-wasm example

Spxg:


Spxg:
A very simple example of using sqlite-wasm, use opfs-sahpool vfs to persist data:

20250112003715_rec_

examples/sqlite/wasm/README.md
```diff
@@ -0,0 +1,18 @@
...
+```
+rustup target add wasm32-unknown-unknown
+# Add wasm32-unknown-unknown toolchain
+
+cargo install wasm-pack
+
+wasm-pack build --web
+# Build wasm
+
+python3 -m http.server 8000
```
weiznich, line 16: Ideally that would work without starting a web server, but I'm not sure if that would be possible. (E.g. by referencing the generated wasm file via a `file://` url?)

examples/sqlite/wasm/README.md
```diff
@@ -0,0 +1,18 @@
...
+# Next, use it on the web page
```
weiznich, line 17: That should maybe just put a clickable link there, e.g. `[on the web page](http://localhost:8000)` or something like that?

examples/sqlite/wasm/migrations/20170124012402_create_posts/down.sql
```diff
@@ -0,0 +1 @@
+-- DROP TABLE posts
```
weiznich, line 1:
```suggestion
DROP TABLE posts
```

weiznich (commented):
Thanks for submitting. Looks good beside these small suggestions to make it a bit easier for new users to follow the instructions.

examples/sqlite/wasm/README.md
Spxg: This is not possible because using `file:://` does not work with `Worker`, and using `opfs` requires the web service to return `COEP` and `COOP` headers.

ps: `opfs` and `opfs-sahpool` are different vfs. The latter does not need `atomics` and `sharedArrayBuffer`, so there is no need to return `COEP` and `COOP` headers.

Spxg:
@weiznich Comments were fixed, and the ability to switch vfs was added, now users can switch vfs and experience their differences

image

weiznich (approved):
Thanks for the update, looks good now.

PR #4439, opened 2025-01-17
fix: setup now sets migrations dir as relative path when possible

### LINKED ISSUE #4436: Setup should not use an absolute path to set the migrations_directory 'dir' field of diesel.toml
### Versions

- **Rust:** 1.83.0
- **Diesel:** 2.2.6
- **Database:** PostgreSQL 17.0

## Problem Description

The `diesel setup` command sets the `migrations_directory` `dir` field of the diesel.toml file to an absolute path, which both leaks personal information (at least the username of the user running the setup command) and breaks it when the local repository has a different path (for instance, when cloned by another team member).

Since it seems to be working perfectly fine with a relative path, I think it would be recommendable to have `diesel setup` set the migrations directory using a path relative to the root of the crate.

## Checklist

the-solid-box:
fixes #4436

### DIFF
--- a/diesel_cli/tests/setup.rs
+++ b/diesel_cli/tests/setup.rs
@@ -308,3 +308,29 @@ fn setup_respects_migrations_dir_from_diesel_toml() {
     assert!(result.is_success(), "Result was unsuccessful {:?}", result);
     assert!(p.has_file("custom_migrations"));
 }
+
+#[test]
+fn setup_writes_migration_dir_as_relative_path_to_config_file() {
+    let p = project("setup_writes_migration_dir_as_relative_path_to_config_file").build();
+
+    let result = p.command("setup").run();
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(p
+        .file_contents("diesel.toml")
+        .contains("dir = \"migrations\""));
+}
+
+#[test]
+fn setup_writes_migration_dir_as_arg_as_relative_path_to_config_file() {
+    let p = project("setup_writes_migration_dir_as_relative_path_to_config_file").build();
+
+    let migrations_dir_arg = p.directory_path().join("foo").display().to_string();
+    let result = p
+        .command("setup")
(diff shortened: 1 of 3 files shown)


the-solid-box:
I do not know where macos differs when it comes to the second test I have included. Since I do not have the means to run tests on macos, and I do not wish to pollute this PR with hypothese-commits on what could be the fix, I'd appreciate if someone can enlighten me :)

diesel_cli/tests/setup.rs
```diff
@@ -308,3 +308,29 @@ fn setup_respects_migrations_dir_from_diesel_toml() {
...
+
+    assert!(result.is_success(), "Result was unsuccessful {:?}", result);
+    assert!(p
+        .file_contents("diesel.toml")
+        .contains("dir = \"migrations\""));
+}
+
+#[test]
+fn setup_writes_migration_dir_as_arg_as_relative_path_to_config_file() {
+    let p = project("setup_writes_migration_dir_as_relative_path_to_config_file").build();
```
AndreCostaaa: Here you're using the same project directory as the other test so you may be causing a race condition in main.c:146 : 

```rust
 if !path.exists() {
 ``` 
I would say that it's probably the reason the CI is failing. Try to change this path

AndreCostaaa: My bad, you changed it in the latest commit. So i guess it's not that

gcavelier:
@the-solid-box, I submitted a PR on your repo to fix the tests on MacOS

https://github.com/the-solid-box/diesel/pull/1

the-solid-box:
> @the-solid-box, I submitted a PR on your repo to fix the tests on MacOS
> 
> [the-solid-box#1](https://github.com/the-solid-box/diesel/pull/1)

Merged, let's see if the CI still complains!

the-solid-box:
Well, looks like windows isn't liking it now... Luckily I do have a windows machine, I'll have a look when I can. Might just write two tests, one for macos, one for windows and linux. Thanks for the help @gcavelier

PR #4442, opened 2025-01-19
Minor rustdoc improvements

thkoch2001:
- The long sentence in statement_cache/mod.rs felt hard to read. I broke it up in two.

- s/a unbound/an unbound/

- I was about to write my own equivalent to BoxedSqlQuery when I found it mentioned in a Github issue. So I thought into_boxed() deserves more advertisement.


weiznich (approved):

diesel/src/pg/connection/raw.rs
```diff
@@ -1,5 +1,5 @@
 #![allow(clippy::too_many_arguments)]
-#![allow(unsafe_code)] // ffi code
+#![allow(unsafe_code)] // ffi Code
```
JohnTitor, line 2: I don't think this diff makes a difference.

weiznich:
> I'm sorry for the unwanted edits. This is very embarrassing. - Thank you for cleaning up after me.

No worries, after all that's one of my tasks as maintainer.

PR #4465, opened 2025-01-30
feat: Add Sqllite json_valid

dhirensr:
Related to https://github.com/diesel-rs/diesel/issues/4366

also I couldn't test the functionality fully, and I've asked a question related to testing in https://github.com/diesel-rs/diesel/discussions/4463

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -421,3 +421,51 @@ extern "SQL" {
         indentation: Nullable<Text>,
     ) -> J::Out;
 }
+
+#[cfg(feature = "sqlite")]
+define_sql_function! {
+    /// Returns  `true`  if the argument is well-formed JSON, or returns  `false`  if is not well-formed.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     #[cfg(feature = "serde_json")]
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # #[cfg(feature = "serde_json")]
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::{sql, json_valid};
+    /// #     use serde_json::{json, Value};
+    /// #     use diesel::sql_types::{Text, Json, Jsonb, Nullable};
+    /// #     let connection = &mut establish_connection();
+    ///
+    /// let version = diesel::select(sql::<Text>("sqlite_version();"))
+    ///         .get_result::<String>(connection)?;
+    ///
+    /// // Querying SQLite version should not fail.
+    /// let version_components: Vec<&str> = version.split('.').collect();
+    /// let major: u32 = version_components[0].parse().unwrap();
+    /// let minor: u32 = version_components[1].parse().unwrap();
+    /// let patch: u32 = version_components[2].parse().unwrap();
+    ///
+    /// if major > 3 || (major == 3 && minor >= 46) {
+    ///     /* Valid sqlite version, do nothing */
+    /// } else {
+    ///     println!("SQLite version is too old, skipping the test.");
+    ///     return Ok(());
+    /// }
+    ///
+    /// let result = diesel::select(json_valid::<Json>(json!({"x":35})))
+    ///     .get_result::<bool>(connection)?;
+    ///
+    /// assert_eq!(1, result);
+    ///
+    /// # }
+    /// ```
+    #[sql_name = "json_valid"]
+    fn json_valid<J: JsonOrNullableJsonOrJsonbOrNullableJsonb + MaybeNullableValue<Bool>>(j: J) -> J::Out;
+}
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -420,10 +420,7 @@ extern "SQL" {
         j: J,
         indentation: Nullable<Text>,
     ) -> J::Out;
-}
 
-#[cfg(feature = "sqlite")]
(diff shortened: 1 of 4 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -422,3 +422,51 @@ define_sql_function! {
...
+    ///
+    /// let result = diesel::select(json_valid::<Json>(json!({"x":35})))
+    ///     .get_result::<Integer>(connection)?;
+    ///
+    /// assert_eq!(1, result);
+    ///
+    /// # }
+    /// ```
+    #[sql_name = "json_valid"]
+    fn json_valid<J: JsonOrNullableJsonOrJsonbOrNullableJsonb + MaybeNullableValue<Text>>(j: J) -> Integer;
```
weiznich, line 471:
```suggestion
    fn json_valid<J: JsonOrNullableJsonOrJsonbOrNullableJsonb + MaybeNullableValue<Bool>>(j: J) -> J::Out;
```

Diesel maps such flag integers to a boolean for sqlite so we use `Bool` as return type. That would usually be `-> Bool`. For this particular function sqlite returns `NULL` if a null value is passed in, which means we want to return `Nullable<Bool>` if the input value is nullable and `Bool` otherwise. The `MaybeNullableValue<Bool>` part takes care of that by transforming the generic parameter into the right variant. That variant then can be accessed via the `Out` associated type.

diesel/src/sqlite/expression/functions.rs
```diff
@@ -422,3 +422,51 @@ define_sql_function! {
...
+    ///
+    /// if major > 3 || (major == 3 && minor >= 46) {
+    ///     /* Valid sqlite version, do nothing */
+    /// } else {
+    ///     println!("SQLite version is too old, skipping the test.");
+    ///     return Ok(());
+    /// }
+    ///
+    /// let result = diesel::select(json_valid::<Json>(json!({"x":35})))
+    ///     .get_result::<Integer>(connection)?;
```
weiznich, line 464:
```suggestion
    ///     .get_result::<bool>(connection)?;
```

diesel/src/sqlite/expression/functions.rs
```diff
@@ -422,3 +422,51 @@ define_sql_function! {
...
+    /// # include!("../../doctest_setup.rs");
+    /// #
+    /// # fn main() {
+    /// #     #[cfg(feature = "serde_json")]
+    /// #     run_test().unwrap();
+    /// # }
+    /// #
+    /// # #[cfg(feature = "serde_json")]
+    /// # fn run_test() -> QueryResult<()> {
+    /// #     use diesel::dsl::{sql, json_pretty_with_indentation};
```
weiznich, line 442:
```suggestion
    /// #     use diesel::dsl::{sql, json_valid};
```

diesel/src/sqlite/expression/functions.rs
```diff
@@ -422,3 +422,51 @@ define_sql_function! {
     #[sql_name = "json_pretty"]
     fn json_pretty_with_indentation<J: JsonOrNullableJsonOrJsonbOrNullableJsonb + MaybeNullableValue<Text>>(j: J, indentation: Nullable<Text>) -> J::Out;
 }
+
+#[cfg(feature = "sqlite")]
+define_sql_function! {
+    /// Returns 1 if the argument is well-formed JSON, or returns 0 if is not well-formed.
```
weiznich, line 428:
```suggestion
    /// Returns `true` if the argument is well-formed JSON, or returns `false` if is not well-formed.
```

diesel/src/sqlite/expression/helper_types.rs
```diff
@@ -27,3 +28,8 @@ pub type json_pretty<E> = super::functions::json_pretty<SqlTypeOf<E>, E>;
 #[cfg(feature = "sqlite")]
 pub type json_pretty_with_indentation<J, I> =
     super::functions::json_pretty_with_indentation<SqlTypeOf<J>, J, I>;
+
+/// Return type of [`json_valid(json)`](super::functions::json_valid())
+#[allow(non_camel_case_types)]
+#[cfg(feature = "sqlite")]
+pub type json_valid<E> = super::functions::json_valid<SqlTypeOf<E>, Integer>;
```
weiznich, line 35:
```suggestion
pub type json_valid<E> = super::functions::json_valid<SqlTypeOf<E>, E>;
```

That needs to be just `E` and not `Integer` as the helper type accepts one generic argument for the SQL type (the `SqlTypeOf<E>` part) and one generic argument for the expression itself (the `E`))

weiznich (changes requested):
Thanks for opening this PR and sorry for taking some days to get back with hints what's wrong here.

The implementation is already on a good way, there are only minor nits that needs to be fixed to run the tests + make this good to merge.

dhirensr:
@weiznich : thank you for reviewing. I can make the changes quickly but facing some errors in testing the change. I tried asking the testing mechanism here : https://github.com/diesel-rs/diesel/discussions/4463 

can you please help? it will be helpful for me in next couple of pull requests.

weiznich:
I'm sorry that it looks like #4463 was not answered. From my point of view the issues outlined there were addressed with the earlier set of comments here.

There was a new set of errors in the relevant doc test, which was essentially caused by the set of comments above being not 100% complete in what needed to be changed. I've now pushed https://github.com/diesel-rs/diesel/pull/4465/commits/6399f89c818430e6faf365c1855e852f3a37f83c to fix these issues, as they were really only minimal syntax errors.

I've also rebased the PR and pushed https://github.com/diesel-rs/diesel/pull/4465/commits/128ffba75937259965fd953f6a0ba74798d2b2fe to pull in recent changes from the master branch.

Thanks again for contributing :heart:

PR #4471, opened 2025-02-06
chore: remove gitter mention

kristof-mattei:
Prevent people from searching for a channel that (no longer) exists!


weiznich:
Thanks

PR #4472, opened 2025-02-07
Improve CompatibleType not implemented error message

Ten0:
I was told "OK I see this error message but my struct already implements Selectable so that's not useful".

As it turns out, this message would sometimes be misunderstood as the `check_for_backend` part only relating to the `QueryableByName` usage, but not to the `Selectable` usage, whereas `check_for_backend` is in fact the key point.

Hopefully this new writing will avoid this. :)

### DIFF
--- a/diesel/src/query_dsl/load_dsl.rs
+++ b/diesel/src/query_dsl/load_dsl.rs
@@ -185,7 +185,7 @@ mod private {
     #[diagnostic::on_unimplemented(
         note = "this is a mismatch between what your query returns and what your type expects the query to return",
         note = "the fields in your struct need to match the fields returned by your query in count, order and type",
-        note = "consider using `#[derive(Selectable)]` or #[derive(QueryableByName)] + `#[diesel(check_for_backend({DB}))]` \n\
+        note = "consider using `#[diesel(check_for_backend({DB}))]` on either `#[derive(Selectable)]` or #[derive(QueryableByName)] \n\
                 on your struct `{U}` and in your query `.select({U}::as_select())` to get a better error message"
     )]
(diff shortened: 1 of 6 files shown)


JohnTitor (commented):
Makes sense 👍

JohnTitor (approved):

diesel/src/query_dsl/load_dsl.rs
```diff
@@ -185,7 +185,7 @@ mod private {
     #[diagnostic::on_unimplemented(
         note = "this is a mismatch between what your query returns and what your type expects the query to return",
         note = "the fields in your struct need to match the fields returned by your query in count, order and type",
-        note = "consider using `#[derive(Selectable)]` or #[derive(QueryableByName)] + `#[diesel(check_for_backend({DB}))]` \n\
+        note = "consider using `#[diesel(check_for_backend({DB}))]` on either `#[derive(Selectable)]` or #[derive(QueryableByName)] \n\
```
weiznich, line 188:
```suggestion
        note = "consider using `#[diesel(check_for_backend({DB}))]` on either `#[derive(Selectable)]` or `#[derive(QueryableByName)]` \n\
```

(Sorry for being pedantic about that)

weiznich (commented):
Looks good for me beside the missing ` to indicate that this is rust code. It just looks strange to use them only partially.

That written: Ideally I would find the time to implement the other part of the `on_unimplemented` attribute, which would allow us to add conditions to that attribute to only apply in certain situations. That would enable us to only talk about `Selectable` or `QueryableByName` if they are relevant. Maybe someday…

PR #4474, opened 2025-02-08
Better implementation using sqlite-wasm-rs

Spxg:
sqlite-wasm-rs has been updated to a major version, providing a better implementation. For details, see <https://github.com/diesel-rs/diesel/discussions/4473>

There are some breaking changes, such as no longer exporting `init_sqlite()`. But I realized that the last merged commit has not been released yet, and I think it is appropriate to merge it now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,8 +318,17 @@ jobs:
       - name: Install wasm-pack
         run: cargo install wasm-pack
 
+      - name: Install emscripten toolchains
+        run: |
+          git clone https://github.com/emscripten-core/emsdk.git
+          cd emsdk
+          ./emsdk install latest
+          ./emsdk activate latest
+
       - name: Test unit and integration tests
-        run: cargo xtask run-tests --wasm sqlite
+        run: |
+          source ./emsdk/emsdk_env.sh
+          cargo xtask run-tests --wasm sqlite
 
   sqlite_bundled:
     name: Check sqlite bundled + Sqlite with asan
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -53,13 +53,13 @@ path = "../diesel_derives"
 libsqlite3-sys = { version = ">=0.17.2, <0.32.0", optional = true, features = ["bundled_bindings"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
-# https://github.com/Spxg/sqlite-wasm-rs#why-provide-precompiled-library
-sqlite-wasm-rs = { version = ">=0.3.0, <0.4.0", optional = true, default-features = false, features = ["precompiled"] }
+sqlite-wasm-rs = { version = ">=0.3.0, <0.4.0", optional = true, default-features = false }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dev-dependencies]
 # Something is dependent on it, we use feature to override it.
 getrandom = { version = "0.2", features = ["js"] }
 wasm-bindgen-test = "0.3.49"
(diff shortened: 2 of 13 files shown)


Spxg:
❤️ Any suggestions?  @weiznich

weiznich:
Thanks for opening this PR. There is generally no need to ping me after having a PR open for a day. If I don't response imitatively that usually means I have no capacity to deal with a PR/question/whatever now. I often answer soon, but sometimes I'm just away on holidays or have other things to do and an answer can take time then. Pinging me doesn't make this faster in any way, often it makes things slower as I have yet another notification to deal with, so please don't do that in the future, at least not after a day. That means I will look into this possibly later this week or next week. If you don't hear something until then feel free to ping me again.

weiznich:
So I had some time to look into this. The change on diesels slide looks ok for me.

I also had a look at the changes in `sqlite-wasm-rs` and I found there a few in my opinion problematic things. Possibly that could be fixed or handled in a better way:

* The crate contains binary data (the `wasm` module itself in `src/wrapper/jswam/sqlite3.wasm` and two static libraries in `library`) See https://diff.rs/sqlite-wasm-rs/0.1.5/0.2.4/ to inspect these files. It's not great to include binary files as they make it really hard to see what's actually included in your final binary. In my opinion it would be desirable to build all three files as part of the build process via the `build.rs` file from source instead.
* There are two sets of bindings in the source code. One in `src/shim/libsqlite3/bindings.rs` and one in `src/wrapper/libsqlite3/bindings.rs`. From a quick look they seem to differ, but it's unclear for my why they differ. Also that might just be handled by a reexport of `libsqlite3-sys`? If the problem there is that `libsqlite3-sys` sets linker arguments you don't want to have it might be [a solution](https://github.com/rusqlite/rusqlite/blob/master/libsqlite3-sys/build.rs#L42) to enable the `in_gecko` feature disable setting linker flags from there. Maybe it's even possible to use the `sqlite3.c` file from there for compilation?

All of that is no hard stopper for merging this, but something I would like to see discussed. Maybe there are hard reasons why all of this is required, in which case it would be great to see that documented somewhere.

weiznich:
Thanks for providing these answers. 

There is a huge problem with including these binary data and link them afterwards: As a downstream consumer you cannot review what is linked into an application. That's a huge security problem and that is something I'm not comfortable with shipping in a released diesel version. I can understand that setting up the emscripten toolchain is complicated, but I personally would consider security to be more important than easy to use. I'm open for a opt-in way to use precompiled versions as long as there is documentation that explains how exactly these binaries can be reproduced and how can be ensured that a binary build according to these instructions can be used to verify that the prebuild binary is the same. The default needs to be building everything from source.

weiznich:
Thanks for the update :heart:. I've gone through the changes and left some comments there. Overall that's a huge step in the right direction. It already mostly addresses my concerns around the prebuild `libsqlite.a` libraries. Given that the `shim` variant is already off by default, it might be acceptable to leave the prebuild wasm binary there.

What I would like to see still changed is the default from using the prebuild binaries to using the bundled feature flags. All the other comments are mostly ideas for making things more robust.

weiznich (changes requested):
Thanks for the update and sorry for taking that long to review. I've been busy with other things.

That's nearly what we want here. The only thing that need to be changed is the enabled features for `sqlite-wasm-rs` As discussed before we shouldn't enable the `precompiled` feature by default.

weiznich:
> Do diesel's test and diesel_tests crate need to specify precompiled by default?

I would prefer to use the `bundled` feature there instead, even if that requires setting up the emscripten tool chain on CI

weiznich (approved):
Thanks

weiznich:
Yes, that just means that #4502 needs to be merged first. I can take care of the rebase afterwards, that shouldn't be a problem.

PR #4478, opened 2025-02-11
Tighten bignum dependency versions

### LINKED ISSUE #4477: Loose dependency ranges on bigint crates may result in a version mismatch
### Versions

- **Diesel:** 2.2.7

### Feature Flags

- **diesel:** `numeric`

## Problem Description

The bigint crate dependencies are given as loose ranges:

```
num-bigint = { version = ">=0.2.0, <0.5.0", optional = true }
bigdecimal = { version = ">=0.0.13, < 0.5.0", optional = true }
```

bigdecimal in turn depends on bigint, but with a more rigid range as is customary for 0.x dependencies in cargo (e.g. bigdecimal 0.4.7 depends on num-bigint ^0.4). Usually there is no problem as there is a single num-bigint version to settle on.

However, in a large workspace that has diesel in the dependency tree, multiple versions of num-bigint or bigdecimal may be pulled by different crates. This can lead to a different version of num-bigint being selected for diesel than the one selected for bigdecimal, resulting in a compilation failure due to missing num-bigint trait implementations for the version required by diesel.

Here's an example: https://github.com/movementlabsxyz/aptos-core/issues/132

## Checklist

mzabaluev:
Fixes #4477

As `bigdecimal` API exposes implementations of traits defined in a particular version of `num-bigint`, it's fragile to have loosely ranged versions on these dependencies. Lock both down to the currently latest ^0.4, that's consistent with how `bigdecimal` 0.4.x depends on `num-bigint`.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -35,10 +35,10 @@ percent-encoding = { version = "2.1.0", optional = true }
 uuid = { version = ">=0.7.0, <2.0.0", optional = true }
 ipnetwork = { version = ">=0.12.2, <0.22.0", optional = true }
 ipnet = { version = "2.5.0", optional = true }
-num-bigint = { version = ">=0.2.0, <0.5.0", optional = true }
+num-bigint = { version = "0.4", optional = true }
 num-traits = { version = "0.2.0", optional = true }
 num-integer = { version = "0.1.39", optional = true }
-bigdecimal = { version = ">=0.0.13, < 0.5.0", optional = true }
+bigdecimal = { version = "0.4", optional = true }
 bitflags = { version = "2.0.0", optional = true }
 r2d2 = { version = ">= 0.8.2, < 0.9.0", optional = true }
 itoa = { version = "1.0.0", optional = true }


weiznich:
Thanks for opening this PR. I'm sorry but I fear we won't be able to accept this PR as this is a breaking change as types from these crates appear in our public API and this removes support for certain versions of these crates.

mzabaluev:
> types from these crates appear in our public API

Doesn't this essentially mean that your public API (its part gated with "numeric") is currently unstable and may break when an unrelated crate in the consuming namespace is updated to pull a particular version of num-bigint that also fits in the range?

weiznich:
> Doesn't this essentially mean that your public API (its part gated with "numeric") is currently unstable and may break when an unrelated crate in the consuming namespace is updated to pull a particular version of num-bigint that also fits in the range?

No that's not what this means and thats also not what semver requires here. The only requirement for a stable API is that the code continues to compile if you use the same version as before. If you somehow change versions on your side that's not a problem in diesel as you always can control which version of a certain crate is used by adjusting your local `Cargo.lock` file, e.g by using `cargo update` to get exactly the right dependency versions there.

mzabaluev:
> If you somehow change versions on your side that's not a problem in diesel as you always can control which version of a certain crate is used by adjusting your local `Cargo.lock` file, e.g by using `cargo update` to get exactly the right dependency versions there.

In https://github.com/movementlabsxyz/aptos-core/issues/132 we were only able to fix this by downgrading bigdecimal to ^0.3, which is not desirable.

> The default behavior depends on cargo, which again is nothing diesel controls.

Cargo gives you what you allow it. Your cargo manifest claims "this crate can be used with any combination of num-bigint and bigdecimal versions in the given ranges". Since bigdecimal actually depends on a narrower version range of num-bigint, this claim is false. #4477 is a valid bug and should not have been closed.

weiznich:
> In https://github.com/movementlabsxyz/aptos-core/issues/132 we were only able to fix this (in the short term) by downgrading bigdecimal to ^0.3, which is not desirable.

I highly doubt that this is the only way to fix that. As your repo seems to be releated to web3/blockchain stuff and I highly disapprove that technology I won't provide further support on fixing that, other than stating: You can fix that by correcting the version used in your `Cargo.lock` file.

> Cargo gives you what you allow it. Your cargo manifest claims "this crate can be used with any combination of num-bigint and bigdecimal versions in the given ranges". 

Given that there is no other way to express this, yes that's what we are saying there. If you want to see that fixed, please work with the cargo team to get us a better way to express what we really want to express there. There is literally nothing in diesel we can do to address this issue without support in cargo itself. Not even changing how `num-bigint` and `bigdecimal` interact would help as that again would require increasing the supported minimal version, which again would require a semver major version bump.

> https://github.com/diesel-rs/diesel/issues/4477 is a valid bug and should not have been closed.

Bugs in our issue tracker should always represent something actionable where people could start working on it. This is not actionable as there is no clear way to fix this, therefore it's closed as won't fix.

As written above: I'm happy to accept a fix that doesn't require a major semver version bump. That's not the case for your presented fix. 

That's likely my last comment on this topic.

PR #4491, opened 2025-02-14
fix(sqlite): fix json_array_length output type

IamTossan:
# fix json_array_length output type

related to 
- https://github.com/diesel-rs/diesel/pull/4479
- https://github.com/diesel-rs/diesel/issues/4366

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -143,14 +143,14 @@ extern "SQL" {
     /// }
     ///
     /// let result = diesel::select(json_array_length::<Json, _>(json!([1,2,3,4])))
-    ///     .get_result::<i32>(connection)?;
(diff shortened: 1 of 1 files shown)


IamTossan:
@weiznich I believe this will complete the feature

weiznich:
For `json_array_lenght` it's not required to unconditionally return a nullable value, as the only way to get a `NULL` return value is to pass in a `NULL` (not json null) value. Otherwise an integer is returned. That means if we know that the input value cannot be null, we also know that the output value cannot be null. That is exactly what the old signature already means.

IamTossan:
> For `json_array_lenght` it's not required to unconditionally return a nullable value, as the only way to get a `NULL` return value is to pass in a `NULL` (not json null) value. Otherwise an integer is returned. That means if we know that the input value cannot be null, we also know that the output value cannot be null. That is exactly what the old signature already means.

I just double-checked, that's correct, my bad

Issue #4333, opened 2024-10-31
allow_tables_to_appear_in_same_query! does not scale with larger number of tables


jeff-hiner:
Adding more tables increases the number of generated `TableNotEqual` trait impls by N squared, since the `allow_tables_to_appear_in_same_query` macro generates pairwise trait impls. Past about 10-15 tables this N squared codegen has a significant impact on compile time.

Making this more difficult, there's no way to prevent the autogenerated `allow_tables_to_appear_in_same_query` block except via patch file. We already have an existing patch file to deal with another schema issue, and it doesn't look like there is a way to specify multiple files in `diesel.toml`.

I'm not sure how to fix this, aside from completely reworking how table checks work. I recognize it's important as part of the type system to not generate queries spanning two disjoint databases, but this is a problem many if not most users won't encounter. Perhaps there's a more scalable way to do the same check with an O(1) trait generation like `BelongsToDatabase<T>` and some clever bounds checking.

### Versions

- Rust: 1.82.0 
- Diesel: 2.2.3
- Database: Postgres
- Operating System: Ubuntu 22 LTS

## Problem Description

We started noticing this more as we've added more tables, and it got significantly worse after changes to the coherence checker in Rust 1.82.

Compounds with https://github.com/rust-lang/rust/issues/132064

### Steps to reproduce

Create a database with a large number of tables, via `diesel cli` and the normal migrations. In our case we have about 70 tables in our database. The 

Generate a library crate with associated autogenerated `schema.rs`, import it and time the build.

Nightly isn't required to exhibit the problem, but `RUSTFLAGS=-Ztime-passes cargo +nightly check` shows the majority of time being spent in the coherence checker. This is consistent with a very large number of trait impls (by my estimate, 70 tables with the Postgres backend will generate approximately 50000 trait impls!)

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [X] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [X] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for opening this issue. 

[...]

PR #4500, opened 2025-02-19
Add print-schema option `allow_tables_in_same_query`

sgoll:
This adds a new config option to `print-schema`, i.e. `allow_tables_in_same_query` (in `diesel.toml`) or `--allow-tables-in-same-query` (in CLI).

Setting this option to `all_tables`, or leaving it unset, remains the current behavior of generating a single invocation of the macro `allow_tables_to_appear_in_same_query!` that lists all tables in the schema.

Setting this option to `fk_related_tables` changes this behavior: instead of a single invocation, several invocations are issued, one for each group of tables related to each other through foreign keys.

For example, given the following tables with foreign keys:

```
users (id)
posts (id, user_id -> users(id))
comments (id, post_id -> posts(id))

sessions (id)
transactions (id, session_id -> sessions(id))

cars (id)
bikes (id)
```

The following macro invocations are generated:

For `all_tables` (default):

```rust
    bikes, cars, comments, posts, sessions, transactions, users);
```

For `fk_related_tables`:

```rust
diesel::allow_tables_to_appear_in_same_query!(comments, posts, users);
diesel::allow_tables_to_appear_in_same_query!(sessions, transactions);
```

See #4333 for rationale.


diesel_cli/src/print_schema.rs
```diff
@@ -594,6 +632,64 @@ impl Display for TableDefinitions<'_> {
...
+        let mut pending = vec![name];
+
+        // Start a depth-first search with the current table name, walking the foreign key relations
+        // in both directions.
+        while let Some(name) = pending.pop() {
+            component.push(name);
+
+            let mut visit = |related_name: &'a TableName| {
+                if !visited.contains(related_name) {
+                    visited.insert(related_name);
```
Ten0, line 669:
```suggestion
                if visited.insert(related_name) {
```

diesel_cli/src/print_schema.rs
```diff
@@ -594,6 +632,64 @@ impl Display for TableDefinitions<'_> {
...
+    fk_constraints: &'a [ForeignKeyConstraint],
+) -> Vec<Vec<&'a TableName>> {
+    let mut visited = BTreeSet::new();
+    let mut components = vec![];
+
+    // Find connected components in table graph. For the intended purpose of this function, we treat
+    // the foreign key relation as being symmetrical, i.e. we are operating on the undirected graph.
+    //
+    // The algorithm is not optimized and suffers from repeated lookups in the foreign key list, but
+    // it should be sufficient for typical table counts from a few dozen up to a few hundred tables.
```
Ten0, line 650: I feel like we probably want to sort the output so as to make it easy to preserve it later if we want to change the algorithm.

diesel_cli/src/print_schema.rs
```diff
@@ -574,16 +595,33 @@ impl Display for TableDefinitions<'_> {
...
                 writeln!(out)?;
-                for table in &self.tables {
-                    if table.name.rust_name == table.name.sql_name {
-                        writeln!(out, "{},", table.name.sql_name)?;
+                for table in tables {
+                    if table.rust_name == table.sql_name {
+                        writeln!(out, "{},", table.sql_name)?;
                     } else {
-                        writeln!(out, "{},", table.name.rust_name)?;
+                        writeln!(out, "{},", table.rust_name)?;
```
Ten0, line 624: I'm aware this was there previously but... isn't that just always printing `rust_name`? 😅

diesel_cli/tests/print_schema/print_schema_allow_tables_in_same_query/diesel.toml
```diff
@@ -0,0 +1,3 @@
+[print_schema]
+file = "src/schema.rs"
+allow_tables_in_same_query = "fk_related_tables"
```
Ten0, line 3: test should probably be named around that it's testing specifically `fk_related_tables`, not `allow_tables_to_appear_in_same_query` itself.

Ten0 (changes requested):
Hey!
Thanks a lot for your contribution! 😊
It looks like this will definitely be useful for people who have very disconnected large graphs.

Unfortunately my 166 tables graph is largely connected so that won't help me personally 😅 - I'm curious, how is your use-case that is so different?

Otherwise from an implementation standpoint this looks good 😊 - although I did find 2-3 things that I would tend to change.

Thanks again for working on this!

Ten0:
> However, since we do not know beforehand which tables will be joined, and the FK-table graph likely has a rather short maximum path length (less than the expected maximum number of table joins, similar to the [small-world experiment](https://en.wikipedia.org/wiki/Small-world_experiment)), this likely would not help reduce the number of table pairs much, if at all.

Haha I gave this strategy some thought yesterday as well while looking at the issue and PR and had reached the exact same conclusion 😵‍💫

I also considered doing static analysis on existing files to figure out what tables were used in the same file/function, which while it may work practically for a lot of cases, might not work well with reusable query fragments that may live across multiple files (often via `#[dsl::auto_type]`). Trying to resolve imports becomes a rather large project that would likely require quite a bit of maintenance, and will always be imperfect if macros are involved, so overall it seems to be a bad complexity/benefit ratio. 😕

diesel_cli/src/print_schema.rs
```diff
@@ -42,6 +43,24 @@ impl Default for DocConfig {
...
+    #[serde(rename = "all_tables")]
+    AllTables,
+}
+
+#[allow(clippy::derivable_impls)] // that's not supported on rust 1.65
+impl Default for AllowTablesInSameQuery {
+    fn default() -> Self {
+        AllowTablesInSameQuery::AllTables
+    }
+}
```
weiznich, line 62: We should be fine to use the derive now, as our minimal supported rust version is currently 1.78

diesel_cli/tests/print_schema/print_schema_allow_tables_in_same_query/diesel.toml
weiznich: I think the suggestion here is to use `print_schema_fk_related_tables` as test name, which should fit the 63 character limit.

weiznich (commented):
Looks good for me beside the minor issues already pointed out.

 As for adding it to a patch release: It's a new feature, we usually don't backport such things to patch releases.

diesel_cli/src/print_schema.rs
Ten0: Looks fine to me to remove it here yes. Thanks! 😊

Ten0: Ah I think the comment was to not just remove the comment but actually use the derive on these types, since now we can

weiznich (approved):

Ten0 (approved):

PR #4503, opened 2025-02-21
Refine stability rules for diesel release

weiznich:
This commit explicitly outlines that we might perform otherwise considered breaking changes in minor releases if that's required by build breaking changes from the rust project itself. A previous example of such a change that nearly required us to make such a breaking change in diesel is the change to default to `resolver = "2"` with the 2021 edition. I rather want to make sure that we don't end up in a similar  situation in the future that requires us to issue a new major diesel version  just to sidestep0breakage from their side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 All user visible changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/), as described
-for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md)
+for Rust libraries in [RFC #1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md). 
+We explicitly reserve the right to perform breaking changes in minor releases to fix build breaking changes 
+from the Rust project. 
 For any named minimal supported Rust version we guarantee that it is possible to build Diesel with the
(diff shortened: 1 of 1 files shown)


weiznich:
To add some context on why I want to change this: https://blog.weiznich.de/blog/cargo-instablity/

Ten0:
To give a worse-case ETA here: I'd like to look a tiny bit further into this, but I'm currently on holidays for the next 15 days, and I don't know that I'll be able to dedicate that time by then.
So if not I'll do that when I'm back at work 😊

weiznich:
No worries, that's something I want to have in place for the diesel 2.3 release, so there is no pressure to merge this soon. 
Enjoy your holidays

pksunkara (approved):

weiznich:
@Ten0 This is a gentle ping, I still would like to move this forward to just have something in place for the worst case that gives us a few more options to handle things reasonable.

Ten0 (approved):
Apologies, I'm finding it hard to dedicate time to this.
I guess the main question is what is defined as a breaking change and what kind of breaking change we ultimately want to allow but I guess this is a question that can be answered practically when considering the relevant potential change.

weiznich:
No worries for taking time to think about this.

I should likely also add that I do not want to ever use this. I mostly see this as last resort option that could be meaningful if it only affects a small part of the user base or are really easy to fix. (E.g. if you write this specific generic code you might need to add a `'static` bound or similar). I still would prefer non-breaking options whenever possible.

PR #4508, opened 2025-02-27
Fix - substitutes DSN exposure with DBMS respective error messages

edonsec:
This PR aims to address some security concerns raised within https://github.com/diesel-rs/diesel/discussions/4504.


diesel_cli/src/errors.rs
```diff
@@ -7,7 +7,7 @@ use crate::infer_schema_internals::TableName;
 pub enum Error {
     #[error("Initializing `.env` file failed: {0}")]
     DotenvError(#[from] dotenvy::Error),
-    #[error("Could not connect to database via `{url}`: {error}")]
+    #[error("Could not connect to database: {error}")]
```
weiznich, line 10: Given that this is a CLI application that expects the URL as explicit argument there is really not much what could be leaked here. So that change should be reverted.

weiznich (changes requested):
Thanks for submitting. The changes to the example project look good to me. I would like to keep the error in `diesel_cli` as it is, as there is really not much possibility to leak anything there, given that the URL is often provided explicitly and the application is designed to be run during development.

diesel_cli/src/errors.rs
edonsec: Thanks for reviewing and fixing the CI error, I have now merged the latest master.

I was curious about this one - but is there a value in exposing this? There would still be concerns if the cli command was say implemented in a CI/CD pipeline and using the environmental DATABASE_URL, where it could also leak the credentials in certain cases.

weiznich: Yes there is value in exposing this. The URL can come from different input sources, like a `.env` file, an environment variable or a command line flag. That means it's not always 100% clear which URL is used. By explicitly printing out the URL, users can easily notice if the wrong URL is used.

And yes it could be exposed in CI situations, but again that environment already has access to this kind of information anyways.

edonsec: Thanks for reviewing weiznich. I have reverted the change to this file and its subsequent test.

I do strongly advise this be reconsidered for a future release. The DSN being exposed is a far bigger risk than a developer being unable to identify which DSN is in effect. 

Perhaps some alternative approaches to consider:
* Sanitizing the DSN to redact the password component
* Having an additional argument in the CLI tool for exposing debug information to limit this behaviour by default.

Happy to prepare additional PR requests for either of the above if desired.

weiznich (approved):

diesel_cli/src/errors.rs
weiznich: As written above: I disagree that this it is security relevant to emit a URL to a database that you **cannot** connect to.

jongiddy: We have diesel migration running in a K8s init container, accessing the DB password through a K8s secret that comes from Vault. No person needs to see the password. However, when there is a database access problem, we see the password in the logs, which exposes it more than necessary.

weiznich: @jongiddy I'm not sure what you are trying to tell us here, as this doesn't add anything to the discussion above. If you don't have any new arguments please don't post to old PR's. As written above: The environment that has these longs has already access to the database url in the first place, eg. via exactly that secret.

PR #4518, opened 2025-03-05
Added optional feature flag `256-column-tables`

LucaCappelletti94:
As per title, this pull request introduces support for tables with up to `256-column-tables` columns.

While reasonable tables usually have way less than 256 columns, in some legacy databases I am dealing with there are tables with over 200 columns. This pull request adds should allow me to use `diesel` to design the data migration from such databases to more well designed ones.

Best,
Luca


weiznich:
I would prefer to rather go with the approach described here: https://github.com/diesel-rs/diesel/discussions/4517#discussioncomment-12422812

PR #4519, opened 2025-03-05
[Bench] Add new MySQL client

c410-f3r:
Adds the new `MySQL` client of `wtx` for evaluating purposes

Also reverts #4390. After a couple of runs, it is possible to confirm that such a thing did increase overall performance.


c410-f3r:
A misunderstanding. Everything should be good now.

Issue #364, opened 2016-07-01
Add support for Uuid to SQLite


flosse:
It would be nice to be able to use `uuid::Uuid` within the structs in combination of SQLite!

sgrif:
As far as I'm aware there is no data type in SQLite that maps to UUID?

flosse:
> As far as I'm aware there is no data type in SQLite that maps to UUID?

I'm sorry. You're totally right!

theduke:
There is no UUID type in sqLite, but you could just store the UUID as a `BYTE` (or `TEXT`).

Automatic serializaton here would be a very welcome feature.

theduke:
No there is not, but the UUID could just be (de)serialized to  `BINARY` or `TEXT`, which would be much nicer than having to convert manually. @sgrif

0xcaff:
I've implemented storing UUIDs in SQLite using `BINARY(128)` using a wrapper type in my package.

Code:
https://github.com/forte-music/core/blob/fc9cd6217708b0dd6ae684df3a53276804479c59/src/models/id.rs#L67-L121

SQL:
https://github.com/forte-music/core/blob/feature/sync-engine/migrations/2018-05-14-200933_setup/up.sql#L1-L12

Usage:
https://github.com/forte-music/core/blob/fc9cd6217708b0dd6ae684df3a53276804479c59/src/models/album.rs#L10-L21

I can make a pull request with an adaptation of this code if anyone is interested.

sameer:
I'm interested @0xcaff, would appreciate it if you have the time to do a PR for this.

0xcaff:
I’d like thoughts from a contributor before making a PR. Thoughts @sgrif?

weiznich:
Our policy for including mappings between certain SQL types an rust types is like following:
* Either a rust type is able to represent all possible values of the matched SQL type and  also the SQL type could represent all values of the rust type
* Or the type is a fundamental type like boolean.
The first condition is not fulfilled because there at `Text` and also `Binary` values that are no valid uuid's (for example all shorter or longer values)
In my opinion also the second point is not fulfilled because uuid is not that fundamental like for example booleans. (If anyone from @diesel/core disagrees, pleas response here :wink:)

Given that (and given that it is quite easy to write a new type wrapper for this outside of diesel) I would say that impl should not be part of diesel itself.

0xcaff:
Playing devils advocate here.

[...]

PR #4521, opened 2025-03-05
Tentatively moved uuid type and implemented blob-based variant for SQLite

LucaCappelletti94:
As per title, I am trying to implement support for UUID with SQLite backend, as I described here: https://github.com/diesel-rs/diesel/discussions/4520

Contrarily to what I described, I have started out with the binary version as I found other code available in diesel that was similar enough and I wanted to start with this first contribution in the smallest and simplest possible way.


weiznich:
Thanks for submitting this PR. There is a lot of related discussion in  https://github.com/diesel-rs/diesel/issues/364 why this is not desirable. I cannot see how any of this is addressed by PR or #4520. It's not that we don't know how to implement this mapping, rather that we feel it's not desirable to have it at all. The variant to have two different feature flags is also not helpful as:

* Features need to be additive, that's not the case here. What happens if both features are enable?
* It adds more features to diesel, which makes it even harder to test things. The current feature count is already barley manageable.

LucaCappelletti94:
Ok, I will try to make a crate adding such SQLite types that do not come with diesel with the necessary wrappers on it.

weiznich:
As demonstrated by for example jiff-diesel it's totally possible to provide support for such types outside of diesel. 

As a more general note: We need to draw the line somewhere which type mappings are supported in diesel and which are not supported as we cannot support every possible third party type.

PR #4526, opened 2025-03-07
FULL OUTER JOIN

oeed:


diesel/src/query_source/aliasing/joins.rs
```diff
@@ -99,6 +101,17 @@ where
...
 }
 
+#[cfg(any(feature = "postgres_backend", feature = "sqlite"))]
+impl<Left, Right, S, C> SelectableExpression<Join<Left, Right, FullOuter>> for AliasedField<S, C>
+where
+    Self: AppearsOnTable<Join<Left, Right, FullOuter>>,
+    Self: SelectableExpression<Left>,
+    Left: AppearsInFromClause<Alias<S>, Count = Never> + QuerySource,
+    Right: AppearsInFromClause<Alias<S>, Count = Never> + QuerySource,
+    // TODO: unsure what specifics are needed here
```
oeed, line 111: I don't really understand how `AppearsInFromClause<Alias<S>, Count = Never>` somehow enforces that nullable columns are required, but this seems to give the required behaviour. Want to double check this is correct.

diesel/src/query_source/joins.rs
```diff
@@ -160,6 +160,43 @@ where
...
+
+    fn from_clause(&self) -> Self::FromClause {
+        self.clone()
+    }
+
+    fn default_selection(&self) -> Self::DefaultSelection {
+        self::private::SkipSelectableExpressionBoundCheckWrapper(
+            self.left
+                .source
+                .default_selection()
```
oeed, line 193: Other impls don't call `.default_selection()` on `Left`, is this correct? Isn't another clear way to call `nullable()` otherwise.

diesel/src/query_source/joins.rs
```diff
@@ -251,6 +288,15 @@ impl<T: Table, Selection> AppendSelection<Selection> for T {
     }
 }
 
+// TODO: not sure if this is the best way, but the existing impls aren't sufficient
+impl<Left: Expression + Clone, Selection> AppendSelection<Selection> for Nullable<Left> {
```
oeed, line 292: The existing `AppendSelection` impls don't allow a `Nullable` left, but is there another type that might be more preferable to implement this on?

diesel/src/query_source/joins.rs
```diff
@@ -407,6 +469,7 @@ impl<Qs, On> QueryDsl for OnClauseWrapper<Qs, On> {}
 /// may be deeply nested, we need to recursively change any appearances of
 /// `LeftOuter` to `Inner` in order to perform this check.
 pub trait ToInnerJoin {
+    // TODO: does anything need to be done for full join here?
```
oeed, line 472: I'm not familiar with `ToInnerJoin`, is there anything to add here?

diesel_derives/src/table.rs
```diff
@@ -690,6 +690,23 @@ fn expand_column_def(column_def: &ColumnDef) -> TokenStream {
     let sql_name = &column_def.sql_name;
     let sql_type = &column_def.tpe;
 
+    let full_join_impl = if cfg!(any(feature = "postgres", feature = "sqlite")) {
```
oeed, line 693: This is separate from `backend_specific_column_impl` as it's both Postgres & SQLite. Not sure if there's different preferred approach though.

diesel/src/query_source/joins.rs
weiznich: That should be fine

diesel/src/query_source/joins.rs
weiznich: Adding that impl should be fine.

diesel/src/query_source/aliasing/joins.rs
weiznich: The simplest way to verify that these are correct is to just write some compile tests that try to select columns without calling `.nullable()` on the left or right side and check if there is a compilation error. If that's the case everything is correct. In fact we do want some test like this one for full joins: https://github.com/diesel-rs/diesel/blob/master/diesel_compile_tests/tests/fail/right_side_of_left_join_requires_nullable.rs

As for how this works:

* `AppearsInFromClause` for joins resolves at some point to a `AppearsInFromClause` for the single table for that column belongs to. That obviously returns that `Count = Once` because that table appears one in the select clause. Now by requiring never we say that a plain `AliasedField`/column cannot appear in the select clause of a left join (full join).

* We enforce that you explicitly need to call `.nullable()` on all columns (even if they are already nullable). That turns the select expression from a plain `AliasField`/column into `Nullable<ThatThing>`, which then uses [this impl here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/expression/nullable.rs#L50-L56), which in turn checks if the inner expression is selectable for the query source if all joins were inner joins.

Hopefully that helps clearing up how that works.

weiznich (commented):
Thanks for working on this. This is great to see :heart:.

I think that's already a solid start implementation wise. I would like to tweak the backend restrictions a bit to make this extendable by third party backends + restrict it really only to postgres + sqlite. Otherwise the implementation is already ready in my opinion.

What is still missing are tests + documentation. Also the compile test coverage could be improved. For tests it would be great to extend the existing join tests [here](https://github.com/diesel-rs/diesel/blob/master/diesel_tests/tests/joins.rs) to have at least:

* A test for a plain full join
* A test for using several full joins in a row
* Mixing full joins with left join and inner join

oeed:
Great, note re docs and tests. 

That's a good point about 3rd party backends. How would you recommend doing so, a trait implemented on the backend? Are there any existing examples of that?

weiznich:
> That's a good point about 3rd party backends. How would you recommend doing so, a trait implemented on the backend? Are there any existing examples of that?

See the instructions here: https://github.com/diesel-rs/diesel/pull/4526#discussion_r1984601480

diesel/src/query_source/joins.rs
```diff
@@ -160,6 +162,42 @@ where
...
+    fn from_clause(&self) -> Self::FromClause {
+        self.clone()
+    }
+
+    fn default_selection(&self) -> Self::DefaultSelection {
+        self::private::SkipSelectableExpressionBoundCheckWrapper(
+            self.left
+                .source
+                .default_selection()
+                .nullable()
```
oeed, line 195: This is actually flawed, and I can't seem to find a good solution. Because `.nullable()` always wraps the selection in `Nullable`, even if already `Nullable`, when using multiple `full_join()` calls you end up with `(Nullable<(Nullable<A>, Nullable<B>)>, Nullable<C>)`, rather than `(Nullable<A>, Nullable<B>, Nullable<C>)`.

One solution could be to make a `IntoNullable` trait on expressions (akin to the `sql_type` version), which would essentially return `Self` if already `Nullable<...>`. However, I've been getting conflicting impl errors between the `T` and `Nullable<T>` blocks and can't seem to find a way to get this to work. Any ideas or other approaches that could be taken? Seemingly the fact it's checking `IsNull` with the indirection from `Expression::SqlType` is preventing Rust from realising they're disjoint as is done in the [sql_type version](https://github.com/diesel-rs/diesel/blob/82df9f08195fad8da634d3e1e8979575985faf8b/diesel/src/sql_types/mod.rs#L614). (still fails if adding `+ SingleValue`)

```rust
/// Convert values into their nullable form, without double wrapping them in `Nullable<...>`.
pub trait IntoNullable<ST> {
    /// The nullable value of this type.
    ///
    /// For all values except `Nullable`, this will be `Nullable<Self>`.
    type Nullable;

    /// Convert this value into its nullable representation.
    ///
    /// For `Nullable<T>`, this remain as `Nullable<T>`, otherwise the value will be wrapped and be `Nullable<Self>`.
    fn into_nullable(self) -> Self::Nullable;
}

impl<T> IntoNullable for T
where
    T: Expression,
    T::SqlType: SqlType<IsNull = sql_types::is_nullable::NotNull>,
{
    type Nullable = Nullable<T>;

    fn into_nullable(self) -> Self::Nullable {
        Nullable(self)
    }
}

impl<T> IntoNullable for Nullable<T>
where
    Nullable<T>: Expression,
    <Nullable<T> as Expression>::SqlType: SqlType, // also fails with SqlType<IsNull = sql_types::is_nullable::IsNullable>
{
    type Nullable = Self;

    fn into_nullable(self) -> Self::Nullable {
        self
    }
}
```

 Then replace `Nullable<Left::DefaultSelection>` with `<Left::DefaultSelection as IntoNullable>::Nullable`.
 
 Alternatively, we could add a `AppendSelection` impl
    for `Nullable<(Nullable<Left>, Nullable<Mid>)>` which unpacks it, but you have the same type issues.

weiznich: I think what could work is using the associated type specialization trick used in other places.

That means we would need to restructure that trait like follows:

```rust
// that shouldn't be part of the public exposed API, i.e. it shouldn't be nameable by 
// users outside of diesel
// Also: Prefer going with adding the `Expression` prefix everywhere 
// as it will otherwise be confused with the sql type trait
// (That already happens from time to time for the types, no 
// need to introduce another case here.)
pub trait IntoNullableExpression<E> {
    type NullableExpression;
    fn into_nullable_expression(e: E) -> Self:::NullableExpression;
}

impl<ST, E> IntoNullableExpression<E> for ST 
where 
    E: Expression<SqlType = ST>,
    ST: SqlType,
    // forward it to the other two impls
    (ST::IsNull, ST): IntoNullableExpression<E>,
{
    type NullableExpression = <(ST::IsNull, ST) as IntoNullableExpression<E>>::NullableExpression;

    fn into_nullable_expression(e: E) -> Self::NullableExpression {
        <(ST::IsNull, ST) as IntoNullableExpression<E>>::into_nullable_expression(e)
    }
}

// these impls cannot overlap with each other as the Self type is clearly
// different. They also shouldn't overlap with the generic one
// above as `SqlType` is only implemented for tuples if all elements
// implement `SqlType` on their own, which is not the case for `is_nullable::NotNull` 
// and `is_nullable::IsNullable`
impl<ST, E> IntoNullableExpression<E> for (is_nullable::NotNull, ST)
where
    E: Expression<SqlType = ST>,
    ST: SqlType<IsNull = is_nullable::NotNull>,
{
    type NullableExpression = Nullable<E>;

    fn into_nullable_expression(e: E) -> Self::NullableExpression {
        Nullable(self)
    }
}

impl<ST, E> IntoNullableExpression<E> for (is_nullable::IsNullable, ST)
where
    E: Expression<SqlType = ST>,
    ST: SqlType<IsNull = is_nullable::IsNullable>,
{
    type NullableExpression = E;

    fn into_nullable_expression(e: E) -> Self::NullableExpression {
        e
    }
}
```

in that way you have to two impl that are clearly not overlapping as they are for different types and a third impl on top of it that forwards to the other impls. You then need to call it slightly different as `<E::SqlType as IntoNullableExpression<E>>::into_nullable_expression(e)` from the calling side, but given that this is only one location + that's not public API I would consider that acceptable.

oeed: Oh clever! Will try update to use that.

oeed: This runs in to an issue unfortunately, as tuples seem to be `IsNullable` if any of the values are `IsNullable`. e.g. `(Integer, Text, Nullable<Text>)::IsNull == IsNullable`. Thus, it keeps `(Integer, Text, Nullable<Text>)` as `(Integer, Text, Nullable<Text>)`, rather than turning it in to `Nullable<(Integer, Text, Nullable<Text>)>`.

[Tuples are using `OneIsNullable`](https://docs.diesel.rs/master/diesel/sql_types/trait.SqlType.html), should they perhaps instead be using `AllAreNullable`? Or if that's correct, is there another way to do this check?

weiznich: No that's correct for tuples. Giving it another thought I noticed that this all is not required. `(Nullable<(Nullable<A>, Nullable<B>)>, Nullable<C>)` is as far as I remember the expected result here. The sql type code handles unwrapping that to something more meaningful if I remember that correctly. In fact that seems to be also the thing we do for `LEFT JOINS`.

oeed: I'm not quite sure I follow sorry, I was under the impression that `AppendSelection` should instead give `(Nullable<A>, Nullable<B>, Nullable<C>)`? In its doc comment it says:

> Used to ensure the sql type of `left.join(mid).join(right)` is `(Left, Mid, Right)` and not `((Left, Mid), Right)`.

The following compiles correctly using left joins:

```rust
    let source: Vec<(User, Option<Post>, Option<Comment>, Option<Like>)> = users::table
        .left_join(posts::table.on(posts::user_id.eq(users::id)))
        .left_join(comments::table.on(posts::id.eq(comments::post_id)))
        .left_join(likes::table.on(comments::id.eq(likes::comment_id)))
        .load(connection)
        .unwrap();
```

weiznich: That means I need to have another look at this to give better suggestions. I'm not sure if I find the time this or next week. If that doesn't happen until then please ping me again.

weiznich: I pushed c718389d which at least fixes compiling all the tests. I'm still not 100% sure if [this impl](https://github.com/diesel-rs/diesel/commit/c718389d9ba91fc6ba0b28fc1cf20be4f7729af7#diff-d0e8a449e498610baa9a5f3086200726d02bd5e1612f5f1d5587c4418baddf0eR344-R358) is correct. It might require an additional step that wraps every element in the tuple returned from the `Join` query source (left hand side) in a `Nullable` wrapper if it's not already nullable. I think this would apply if you have first a inner join and then a full join. If that is still problematic it should be possible to implement such a tuple mapping by having a helper trait for all supported tuple sizes that just applies the `IntoNullableExpression` trait to all tuple elements.

oeed: Thanks for that, I added a test that did confirm that inner join then full join is problematic, so I've added `IntoNullableExpression` for tuples.

It is now essentially working, however, I've needed to add `IntoNullableExpression` for `SkipSelectableExpressionBoundCheckWrapper` as `impl<ST, E> IntoNullableExpression<E> for ST` needs to be where `ST: SingleValue` to avoid conflicting with the tuple impls and `SkipSelectableExpressionBoundCheckWrapper` no longer has an impl. That's conflicting with `impl<ST, E> IntoNullableExpression<E> for ST` though. Looking forward to negative trait bounds stabilising...

I don't have more time right now to try and fix, but will come back in a few more days when I get the time. If you have thoughts on it let me know though.

PR #4529, opened 2025-03-08
sqlite-json_quote(value)

JakobKlocker:
This is not completely done and I'd need some assistance.
Two things are missing, first of all I can't find a value type for Reals/Floats  which I can use with MaybeNullableValue, does this exist?

Secondly, when passing a json as argument a no-op should happen, yet it seems like it gets treated like a normal text and therefore gets quotes around it. I'm not quite sure why this would happen, since I've added the option to call the function with json.

Some help would be appreciated. I've added comments in the code to the last commit, since I wasn't able to add any for this PR.

https://github.com/diesel-rs/diesel/issues/4366#issuecomment-2705003346

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -928,19 +928,25 @@ extern "SQL" {
     ///     println!("SQLite version is too old, skipping the test.");
     ///     return Ok(());
     /// }
-    /// let test_value = 3.14159
-    /// 
-    /// let result = diesel::select(json_quote::<Integer, _>(test_value)
+    /// let test_value = 42;
+    /// let result = diesel::select(json_quote::<Integer, _>(test_value))
     ///     .get_result::<Option<String>>(connection)?;
     ///
-    /// assert_eq!(Some(test_value), result);
+    /// assert_eq!(Some("42".to_string()), result);
     ///
     /// # Ok(())
     /// # }
     /// ```
     #[sql_name = "json_quote"]
     #[cfg(feature = "sqlite")]
-    fn json_quote<J: MaybeNullableValue<Integer> + MaybeNullableValue<Text>>(
-        j: J,
-    ) -> JsonOrNullableJson; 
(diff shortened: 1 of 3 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -892,4 +892,77 @@ extern "SQL" {
...
+    // COMENT----------------------------------------
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "json_quote"]
+    #[cfg(feature = "sqlite")]
+    fn json_quote<J: MaybeNullableValue<Json> + MaybeNullableValue<Integer> + MaybeNullableValue<Text>>(  // Have to add Supoprt for Real/Floates, how? Existing Datatype?
+        j: J
+    ) -> Nullable<Text>;
```
weiznich, line 967: I think we need a custom trait for the input value here. Ideally that would be something that accepts any possible SQL type.

That means something like 

```suggestion
    fn json_quote<J: SqlType + SingleValue>(
        j: J
    ) -> Json;
```

should just work. We return a non-nullable value as the returned value is independently from the input values guaranteed to be a valid json fragment.

weiznich (commented):
Thanks for working on this. I think we should tweak the function signature a bit to match more closely what SQLite actually does. Hopefully that also clears up how to fix the currently broken tests. Let me know if you need more pointers.

JakobKlocker:
> Thanks for working on this. I think we should tweak the function signature a bit to match more closely what SQLite actually does. Hopefully that also clears up how to fix the currently broken tests. Let me know if you need more pointers.

Thanks for the reply, I've added the SupportedSqlTypes trait and return a json now.

One thing I'm unsure about is if I should add every possible type in auto_type.rs.
Let me know if there is anything else I should take care of.

weiznich:
Thanks for the update. I do not think we need an additional trait there as we can literally accept any `SqlType` there as sqlite as far as I know just accepts anything there. So we could drop the additional trait. That would also allow to pass in custom types build on top of/by diesel. For example `Timestamp`  values/expressions. 

As for the testing: You don't need to test every possible type in `auto_type.rs` It's fine to have one or two there and expect that the others should work.

weiznich (approved):

PR #4530, opened 2025-03-08
feat(sqlite): Add `json_extract` and `jsonb_extract` functions

KekmaTime:
This PR covers implementation of [json_extract() function](https://sqlite.org/draft/json1.html#jex)

- `json_extract(X, P)` - Extracts values from JSON data using path expressions.
- `jsonb_extract(X, P)` - Similar to `json_extract` but returns arrays/objects in JSONB format.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -479,12 +479,12 @@ jobs:
           cargo +stable -Z build-std test --manifest-path diesel/Cargo.toml --no-default-features --features "mysql extras __with_asan_tests" --target x86_64-unknown-linux-gnu
 
   minimal_rust_version:
-    name: Check Minimal supported rust version (1.84.0)
+    name: Check Minimal supported rust version (1.86.0)
     runs-on: ubuntu-latest
     needs: [rustfmt_and_clippy]
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.84.0
+      - uses: dtolnay/rust-toolchain@1.86.0
       - uses: dtolnay/rust-toolchain@nightly
       - uses: taiki-e/install-action@cargo-hack
       - uses: taiki-e/install-action@cargo-minimal-versions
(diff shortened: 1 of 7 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -892,4 +892,198 @@ extern "SQL" {
...
     ) -> Nullable<Text>;
+
+    /// The json_extract(X,P1,P2,...) extracts and returns one or more values from the well-formed JSON at X.
+    ///
+    /// If only a single path P1 is provided, then the SQL datatype of the result is NULL for a JSON null,
+    /// INTEGER or REAL for a JSON numeric value, an INTEGER zero for a JSON false value, an INTEGER one for a JSON true value,
+    /// the dequoted text for a JSON string value, and a text representation for JSON object and array values.
+    ///
+    /// If there are multiple path arguments (P1, P2, and so forth) then this routine returns SQLite text
+    /// which is a well-formed JSON array holding the various values.
```
weiznich, line 903: This part is not possible with the function signature below. This refers to a use case where you would call `json_extract(json_col, path_1, path_2)` which is disallowed by the signature.

diesel/src/sqlite/expression/functions.rs
```diff
@@ -892,4 +892,198 @@ extern "SQL" {
         j: J,
         path: Text,
     ) -> Nullable<Text>;
+
+    /// The json_extract(X,P1,P2,...) extracts and returns one or more values from the well-formed JSON at X.
+    ///
+    /// If only a single path P1 is provided, then the SQL datatype of the result is NULL for a JSON null,
+    /// INTEGER or REAL for a JSON numeric value, an INTEGER zero for a JSON false value, an INTEGER one for a JSON true value,
+    /// the dequoted text for a JSON string value, and a text representation for JSON object and array values.
```
weiznich, line 900: That means the signature as expressed below doesn't really match what is returned by the function. Given that is is essentially a runtime determined value I do not have any good recommendation how to make this work with diesels static checking.

I could imaging to expose just several variants of this function, so `json_extract_as_text`, `json_extract_as_integer`, `json_extract_as_float`  and `json_extract_as_json`  for the different return type variants and let the users chose the right one for their usecase. After all you can even convert any sql return value to any other type internally.

weiznich (commented):
Thanks for opening this PR. It seems like this method is not as straighforward to represent in diesel as we would like. I made some suggestions how we could represent it anyway. Maybe you can have a look at updating the PR? Let me know if you need more help there.

The comments about the signatures apply to the `jsonb` variant as well.

PR #4531, opened 2025-03-11
diesel_cli: Fix --diff-schema adds foreign keys incorrectly (small fix)

sk8ersteve:
When creating a schema.rs with tables A and B, where B has a foreign key pointing to table A, the "diesel migration generate --diff-schema" command incorrectly uses the primary key of table B as the referenced column rather than the primary key of table A.

schema.rs
```
diesel::table! {
    books (book_id) {
        book_id -> Nullable<Integer>,
        title -> Text,
    }
}

diesel::table! {
    pages (page_id) {
        page_id -> Nullable<Integer>,
        page_number -> Integer,
        content -> Text,
        foreign_book_id -> Integer,
    }
}

diesel::joinable!(pages -> books (foreign_book_id));

    books,
    pages,
);
```

generated up.sql before proposed change:
```
CREATE TABLE `books`(
	`book_id` INTEGER PRIMARY KEY,
	`title` TEXT NOT NULL
);

CREATE TABLE `pages`(
	`page_id` INTEGER PRIMARY KEY,
	`page_number` INTEGER NOT NULL,
	`content` TEXT NOT NULL,
	`foreign_book_id` INTEGER NOT NULL,
	FOREIGN KEY (`foreign_book_id`) REFERENCES `books`(`page_id`)
);
```

generated up.sql after proposed change:
```
CREATE TABLE `books`(
	`book_id` INTEGER PRIMARY KEY,
	`title` TEXT NOT NULL
);

CREATE TABLE `pages`(
	`page_id` INTEGER PRIMARY KEY,
	`page_number` INTEGER NOT NULL,
	`content` TEXT NOT NULL,
	`foreign_book_id` INTEGER NOT NULL,
	FOREIGN KEY (`foreign_book_id`) REFERENCES `books`(`book_id`)
);
```


sk8ersteve:
I also confirmed that the generated sql generates the same schema.rs code when --diff-schema isn't used. The same is not true when generated a schema.rs with the previously generated sql. This would be a good test to add for --diff-schema.

weiznich:
Thanks for submitting :heart:

Would it be possible to add an test case for this by adding a test function in this file (https://github.com/diesel-rs/diesel/blob/master/diesel_cli/tests/migration_generate.rs) and the relevant files in this directory (https://github.com/diesel-rs/diesel/tree/master/diesel_cli/tests/generate_migrations)

sk8ersteve:
Hey, yeah actually the diff_add_table_with_fk test covers it if I change the primary keys appropriately. I've got the code ready but the postgres tests fail with other issues but at least stops failing on the foreign key test. Haven't gotten a chance to test MySQL yet. Maybe you can pull my change and try it. Otherwise, I'll try at it again tomorrow.

weiznich (approved):
CI seems to be fine with these changes. Also the change itself looks fine for me.

It would be still nice to know which tests failed locally for you. I think we still might have some tests that fail randomly from time to time for whatever reason. It would be good to know about them and fix them at some point. So please submit an issue for that.

PR #4532, opened 2025-03-12
fix: mark the r2d2 connection check query as safe to cache

achanda:
Currently diesel runs this query as a unnamed prepared statement sending the parse and bind in different transactions. This breaks when pgbouncer is used in between diesel and postgres. https://github.com/pgbouncer/pgbouncer/issues/1288

Changing this to a named prepared statement does not affect functionality in any way, and makes this work irrespective of poolers.

### DIFF
--- a/diesel/src/r2d2.rs
+++ b/diesel/src/r2d2.rs
@@ -447,6 +447,11 @@ where
         pass.push_sql("SELECT 1");
         Ok(())
     }
+
+    // Explicitly mark this query as safe to cache, ensuring it gets a name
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for submitting this PR. Can you explain why you believe that this change fixes your problem? I just double checked locally and the default impl already returns true for this particular query, so its unclear for me why exactly you want to change it that way.

achanda:
I tested this locally a bit more and found that this does not solve my problem. Diesel is still sending these `SELECT 1` queries as unnamed prepared statements. I am going to close this PR and explore the code more.

PR #4534, opened 2025-03-12
Fix wtx loading_association bench

weiznich:
It turns out that this benchmark didn't insert the same amount of data as the other benchmarks, so it run faster as it doesn't need to deal with the same amount of data. This commit adds the relevant missing data by inserting the 10k comments to the relevant table.

This commit also adds a few additional asserts to make sure that the number of data is the same between diesel and wtx

cc @c410-f3r

### DIFF
--- a/diesel_bench/benches/diesel_benches.rs
+++ b/diesel_bench/benches/diesel_benches.rs
@@ -102,11 +102,11 @@ pub struct Comment {
     text: String,
 }
 
-#[derive(Debug, Clone, Copy, Insertable)]
+#[derive(Debug, Clone, Insertable)]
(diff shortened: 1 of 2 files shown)


c410-f3r:
Thank you @weiznich!

I tried my best to create equal benchmarks conditions and it is unfortunate that a piece was still missing.

weiznich:
As already written on Reddit: No worries, that's something I should have checked in the initial review. Also it's nothing super critical, something like that can always be fixed later on. 

Otherwise I would like to congratulate on the wtx release, you are doing good work there.

c410-f3r:
You are too kind! Thank you for maintaining and improving Diesel!

Issue #2298, opened 2020-02-12
grouped_by panics if key is not in Parent array


nishtahir:
### Versions

- **Rust:** nightly-2019-12-22
- **Diesel:** version = "1.4.3", features = ["postgres", "r2d2"]
- **Database:** Postgres
- **Operating System:** macOS

### Feature Flags

- **diesel:**

## Problem Description

```
thread '<unnamed>' panicked at 'no entry found for key', src/libcore/option.rs:1188:5
2020-02-12T12:59:33.327706314Z app[web.1]: stack backtrace:
2020-02-12T12:59:33.329244468Z app[web.1]:    0: backtrace::backtrace::libunwind::trace
2020-02-12T12:59:33.329267263Z app[web.1]:              at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
2020-02-12T12:59:33.329297828Z app[web.1]:    1: backtrace::backtrace::trace_unsynchronized
2020-02-12T12:59:33.329303867Z app[web.1]:              at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
2020-02-12T12:59:33.329309380Z app[web.1]:    2: std::sys_common::backtrace::_print_fmt
2020-02-12T12:59:33.329314516Z app[web.1]:              at src/libstd/sys_common/backtrace.rs:84
2020-02-12T12:59:33.329319686Z app[web.1]:    3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
2020-02-12T12:59:33.329325237Z app[web.1]:              at src/libstd/sys_common/backtrace.rs:61
2020-02-12T12:59:33.329330381Z app[web.1]:    4: core::fmt::write
2020-02-12T12:59:33.329335406Z app[web.1]:              at src/libcore/fmt/mod.rs:1057
2020-02-12T12:59:33.329340266Z app[web.1]:    5: std::io::Write::write_fmt
2020-02-12T12:59:33.329345062Z app[web.1]:              at src/libstd/io/mod.rs:1426
2020-02-12T12:59:33.329349777Z app[web.1]:    6: std::sys_common::backtrace::_print
2020-02-12T12:59:33.329354460Z app[web.1]:              at src/libstd/sys_common/backtrace.rs:65
[...]

PR #4536, opened 2025-03-12
Add fallible `GroupedBy` method

Wolveric:
I am making this pull request to partially address my comments on issue #2298 , by adding a fallible alternative to the `GroupedBy::grouped_by` method, in the form of `GroupedBy::try_grouped_by`.

As suggested, the method features a default implementation that calls `GroupedBy::grouped_by`, to avoid breaking changes, but it does not change any implementations of `GroupedBy::grouped_by`, as I don't think collecting ungrouped items is appropriate in all contexts. If this is the preferred approach, I will add it in a future commit, or separate pull-request.

Additionally, I will submit this pull request as a draft for the time being, as the documentation it provides may be insufficient, and it does not include tests outside of the example used in the documentation for `GroupedBy::try_grouped_by`.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -104,6 +104,116 @@ pub trait BelongsTo<Parent> {
 pub trait GroupedBy<'a, Parent>: IntoIterator + Sized {
     /// See the trait documentation.
     fn grouped_by(self, parents: &'a [Parent]) -> Vec<Vec<Self::Item>>;
+
+    /// A fallible alternative to [`GroupedBy::grouped_by`].
+    ///
+    /// If any child record could not be grouped,
+    /// either because of a `NULL` foreign key,
+    /// or a parent record with a matching key could not be found,
+    /// this function should return `Err`,
+    /// with all successfully grouped records, as well as any ungrouped records.
+    ///
+    /// # Errors
+    ///
+    /// If a parent record could not be found for any of the child records,
+    /// this function should return the `TryGroupedByError`.
+    /// Every supplied record should be contained in the returned error,
+    /// either in the `grouped` field, if it was successfully grouped,
+    /// or the `ungrouped` field, if it was not possible to associate
+    /// with a parent record.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # include!("../doctest_setup.rs");
+    /// # use diesel::associations::TryGroupedByError;
+    /// # use schema::{posts, users};
+    /// #
+    /// # #[derive(Identifiable, Queryable, PartialEq, Debug)]
+    /// # pub struct User {
+    /// #     id: i32,
+    /// #     name: String,
+    /// # }
+    /// #
+    /// # #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


diesel/src/associations/belongs_to.rs
```diff
@@ -104,6 +104,116 @@ pub trait BelongsTo<Parent> {
...
+/// it may not be possible to find a parent record it should be grouped by.
+///
+/// When encountering these missing relationships,
+/// it may still be possible to group remaining records,
+/// but would affect the contents of the returned values.
+/// By extracting the contents of this struct, it is still possible
+/// to use these resulting groups, as well as any records
+/// that could not be grouped.
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+pub struct TryGroupedByError<Child> {
```
weiznich, line 210:
```suggestion
#[non_exhaustive]
pub struct TryGroupedByError<Child> {
```

we don't want to allow people to deconstruct this completely.

Maybe we want also to add a constructor to allow adding fields in a backward compatible way later on

diesel/src/associations/belongs_to.rs
```diff
@@ -131,6 +241,36 @@ where
         }
         result
     }
+
+    fn try_grouped_by(
+        self,
+        parents: &'a [Parent],
+    ) -> Result<Vec<Vec<Child>>, TryGroupedByError<Child>> {
+        use std::{collections::HashMap, iter};
```
weiznich, line 249:
```suggestion
        use std::collections::HashMap;
        use std::iter;
```

to keep this consistent with the other imports in diesel.

weiznich (commented):
Thanks for opening this PR. This already looks really good to be. I made some small suggestions to be a bit more conservative about what's considered to be stable API and what not.

Other than that I'm not totally happy with the code duplication between the `grouped_by` and `try_grouped_by` method, maybe there is a better solution for that. It's something I could live with, but I would be more happy if we could avoid it.

diesel/src/associations/belongs_to.rs
Wolveric: Ah, fantastic! I'd never looked closely into the effect of the `#[non_exhaustive]` attribute! I'll definitely add that, alongside a constructor to support external implementors.

Wolveric:
> Other than that I'm not totally happy with the code duplication between the `grouped_by` and `try_grouped_by` method, maybe there is a better solution for that. It's something I could live with, but I would be more happy if we could avoid it.

Fair enough, I don't love it either. There's two options for I thought of in that case:

### Option 1: Have `GroupedBy::grouped_by` call `GroupedBy::try_grouped_by` and unwrap the grouped records

I still don't love this solution, as it may result in the ungrouped records being collected, even if they are subsequently discarded...

### Option 2: Move the initialisation of the `grouped` and `id_indicies` values to a helper functions

This would remove the code duplication, whilst allowing `GroupedBy::grouped_by` method to avoid the work of collecting ungrouped records it will discard.

Additionally, these helper functions could be added to a private module, as using `HashMap` in the function signature would require it being imported outside of the function, and a private module would isolate the imports used by the helper functions.

weiznich (approved):

PR #4537, opened 2025-03-14
fix: use PQsendQueryParams for unnamed prepared statements

achanda:
In the current implementation, unnamed prepared statements are first prepared and then bound to in different transactions. This breaks pgbouncer's prepared statement tracking in transaction mode leading to "unnamed prepade statement not found" errors.

This PR adds a new wrapper for PQsendQueryParams that prepares and binds unnamed statements in the same transaction, helping pgbouncer track those effectively. This wrapper is called while executing unnamed statements.

Thanks to @JelteF for helping debug from pgbouncer's side and for the fix idea.


achanda:
My minimal repro for this problem is here https://github.com/achanda/dieseltest
Clients report errors like these `ERROR r2d2: unnamed prepared statement does not exist`

achanda:
I messed up the commit, I will rewrite everything and push in a new PR.

PR #4539, opened 2025-03-16
fix: use PQsendQueryParams for unnamed prepared statements

achanda:
In the current implementation, unnamed prepared statements are first prepared and then bound to in different transactions. This breaks pgbouncer's prepared statement tracking in transaction mode leading to "unnamed prepade statement not found" errors.

This PR adds a new wrapper for PQsendQueryParams that prepares and binds unnamed statements in the same transaction, helping pgbouncer track those effectively. This wrapper is called while executing unnamed statements.

### DIFF
--- a/diesel/src/pg/connection/stmt/mod.rs
+++ b/diesel/src/pg/connection/stmt/mod.rs
@@ -13,13 +13,8 @@ use crate::result::QueryResult;
 use super::raw::RawConnection;
 
 enum StatementKind {
-    Unnamed {
-        sql: Option<CString>,
-        param_types: Option<Vec<u32>>,
-    },
-    Named {
-        name: CString,
-    },
+    Unnamed { sql: CString, param_types: Vec<u32> },
+    Named { name: CString },
 }
 
 pub(crate) struct Statement {
@@ -52,48 +47,38 @@ impl Statement {
             .try_into()
             .map_err(|e| crate::result::Error::SerializationError(Box::new(e)))?;
 
-        let name = match &self.kind {
-            StatementKind::Named { name } => Some(name),
-            _ => None,
-        };
-
-        if let Some(name) = name {
-            unsafe {
-                // execute the previously prepared statement
-                // in autocommit mode, this will be a new transaction
-                raw_connection.send_query_prepared(
-                    name.as_ptr(),
-                    param_count,
-                    params_pointer.as_ptr(),
-                    param_lengths.as_ptr(),
-                    self.param_formats.as_ptr(),
-                    1,
-                )
-            }?;
-        } else {
-            // execute the unnamed prepared statement using send_query_params
-            // which internally calls PQsendQueryParams, making sure the
-            // prepare and execute happens in a single transaction. This
-            // makes sure these are handled by PgBouncer.
-            // See https://github.com/diesel-rs/diesel/pull/4539
-            if let StatementKind::Unnamed {
-                sql: Some(ref sql),
-                param_types,
-            } = &self.kind
-            {
+        match &self.kind {
+            StatementKind::Named { name } => {
                 unsafe {
-                    raw_connection.send_query_params(
-                        sql.as_ptr(),
+                    // execute the previously prepared statement
+                    // in autocommit mode, this will be a new transaction
+                    raw_connection.send_query_prepared(
+                        name.as_ptr(),
                         param_count,
-                        param_types_to_ptr(param_types.as_ref()),
                         params_pointer.as_ptr(),
                         param_lengths.as_ptr(),
                         self.param_formats.as_ptr(),
                         1,
                     )
-                }?;
+                }?
             }
-        }
+            StatementKind::Unnamed { sql, param_types } => unsafe {
+                // execute the unnamed prepared statement using send_query_params
+                // which internally calls PQsendQueryParams, making sure the
+                // prepare and execute happens in a single transaction. This
+                // makes sure these are handled by PgBouncer.
+                // See https://github.com/diesel-rs/diesel/pull/4539
+                raw_connection.send_query_params(
+                    sql.as_ptr(),
+                    param_count,
+                    param_types.as_ptr(),
+                    params_pointer.as_ptr(),
+                    param_lengths.as_ptr(),
+                    self.param_formats.as_ptr(),
+                    1,
+                )
+            }?,
+        };
 
         if row_by_row {
             raw_connection.enable_row_by_row_mode()?;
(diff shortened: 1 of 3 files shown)


weiznich:
Thanks for opening this PR. I would like to ask for a clear description of the underlying problem before discussing your solution, as there are quite a few things in there that I consider problematic. Would it be possible for you to provide some information about why the existing approach doesn't work with pgbouncer. Ideally without relying on domain specific knowledge and terminology from pgbouncer as I do not use this tool on my own. For example as far as I can see the term "transaction" does not refer to what is commonly referred to as transaction in relation to database systems. 
I think that would help me to understand whether and how we fix that problem in diesel or if we consider that to be a bug in pgbouncer.

JelteF:
PgBouncer maintainer here. I want to be clear upfront: This is a bug in PgBouncer (at least according to current definition of the Postgres protocol). But as [explained in the related PgBouncer issue](https://github.com/pgbouncer/pgbouncer/issues/1288#issuecomment-2709164178) in more detail fixing it in PgBouncer would probably cause significant performance overhead. To give some more context I'll first describe some background and then the problem that is occuring.

### Background
In postgres there are two types of transactions, explicit (BEGIN + COMMIT) and implicit (aka auto-commit). You're using the "[extended protocol](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)" i.e. the Parse+Bind+Describe+Execute messages. In the extended protocol a set of those messages is considered an implicit transaction from the first message until the next Sync message.

PgBouncer makes sure that all messages inside a single transaction from a client are executed on the same connection to a server (both for implicit and explicit transactions). For *named* prepared statements PgBouncer also keeps track which statements each client prepared and recreates them on the server connection automatically when the client tries to use them. For *unnamed* prepared statements PgBouncer doesn't do this, because they are normally used by clients in an extremely short lived manner, so normally only in the same transaction that they were created.

### The problem

Diesel currently sends this series of messages when it executes a query as an unnamed prepared statement:
1. Parse
2. Sync
3. Describe
4. Bind
5. Execute
6. Sync

So this creates two implicit transactions. That means that the creation of the unnamed prepared statement can be done on server connection A, and then the second implicit transaction that uses the prepared statemnt is run on server connection B, where this statement doesn't exist (or probably worse, means something completely different).

By not sending the Sync at step 2 everything would work fine with PgBouncer. The reason this Sync is sent is because Diesel uses `PQprepare` followed by `PQsendQueryPrepared` to execute a query as an unnamed prepared statement. There is also `PQsendQueryParams` which doesn't send that Sync message at step 2. All other clients that I know of that use libpq, use this API to send unnamed prepared statements. So if diesel uses that API, then the bug in PgBouncer would not be hit.

### Additional benefit of this change

Even without considering PgBouncer, it would be beneficial for Diesel to use `PQsendQueryParams` is better than using `PQprepare` followed by `PQsendQueryPrepared`: You reduce the number of roundtrips to the database from 2 to 1 for a query, by sending all messages in one go. (Also you send one fewer Sync message so you shave off a couple of bytes of traffic, but that benefit is almost certainly negligable)

### Summary

PgBouncer has a bug with unnamed prepared statements, but Diesel is the only client (that I know of) that actually hits this bug, because it uses `PQprepare` followed by `PQsendQueryPrepared` instead of `PQsendQueryParams` to execute an unnamed prepared statement.

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -68,31 +88,44 @@ impl Statement {
...
             )
         };
         PgResult::new(internal_result?, raw_connection)?;
 
         Ok(Statement {
-            name,
+            name: name_cstr,
             param_formats: vec![1; param_types.len()],
+            sql: None,
+            param_types: None,
```
weiznich, line 128: Could we refactor this code a bit to better use the following pattern:

```
if let Some(name) = name {
   // do named statement stuff here
} else {
   // do unnamed stuff here
}
```

That would at least initializing the name with nothing.

Also it might be worth to introduce an enum 

```rust
enum StatementKind {
    Unnamed {sql: CString, param_types: Vec<_>},
    Named {name: CString}
}
```

to better group data that belong together. This enum could be used as a second field inside of `Statement` beside the `param_formats` field which is always the same for all statements.

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -41,16 +43,34 @@ impl Statement {
...
-                param_count,
-                params_pointer.as_ptr(),
-                param_lengths.as_ptr(),
-                self.param_formats.as_ptr(),
-                1,
-            )
-        }?;
+
+        if let Some(ref sql) = self.sql {
+            // For regular unnamed statements, use send_query_params
```
weiznich, line 48: It would be great to include the reasons why we should use `send_query_params` here

diesel/src/pg/connection/stmt/mod.rs
```diff
@@ -41,16 +43,34 @@ impl Statement {
...
+                    param_count,
+                    param_types_to_ptr(self.param_types.as_ref()),
+                    params_pointer.as_ptr(),
+                    param_lengths.as_ptr(),
+                    self.param_formats.as_ptr(),
+                    1,
+                )
+            }?;
+        } else {
+            // For named statements, use the existing approach
```
weiznich, line 61: The expression "use the existing approach" looks reasonable now if we look at this change. We won't know what this means in a few month. So better put some longer comment above the if condition, explaining why this is necessary.

weiznich (changes requested):
Thanks for updating this PR and thanks to @JelteF for providing these details. I agree that it is reasonable to address this in diesel.

The code already looks much more reasonable than at the point I wrote my first comment. I've added some details on how we might restructure this to make this a bit more robust. It would be great to get these adjustments.

achanda:
I also tested this using my minimal repro https://github.com/achanda/dieseltest

weiznich:
@achanda I've pushed a cleaned up version. Could you verify that this still resolves your problem. If that's the case this is ready to be merged.

achanda:
Thanks @weiznich checked that this still works without that `unnamed prepared statement not found` error.

achanda:
Thanks for all your help @weiznich 🙏 when is the next planned release? I'd love to be able to roll this out :)

weiznich:
@achanda As this is still a project run in our free time we do not give any release ETA's other than when it's done. I personally would consider this to be a bug fix so it would be eligible to be included in a patch release. You can help preparing such a patch release by opening a PR that backports any [recent PR marked as `maybe backport`](https://github.com/diesel-rs/diesel/pulls?q=is%3Apr+label%3A%22maybe+backport%22+is%3Aclosed) to the 2.2.x branch by cherry picking the relevant commits there + writing entries for all changes to the CHANGELOG.md file. This should include #4527, #4531 and this PR. After that the core team will likely require a few days/weeks to decide on performing this release.

Issue #4543, opened 2025-03-20
Postgres: wrong conversion with ipnet::Ipnet


Zstorm999:
### Versions

- **Rust:**: 1.85
- **Diesel:**: 2.2
- **Database:**: postgres
- **Operating System**: Linux

### Feature Flags

- **diesel:** `postgres_backend`, `ipnet-address`

## Problem Description

When converting ip addresses from `ipnet::IpNet` in the `FromSql` impl, there is an oversight which means ip addresses are not correctly converted. 

The problem only affects ip addresses of the form `192.168.0.1/24` where the bits after the mask are not all 0.
These bits are effectively wiped from the conversion, meaning the former address would be converted to `192.168.0.0/24` (and stored as such in the database).

## Problem cause

As I said before, this problem is due to a wrong conversion, and in particular to using `IpNet::network` instead of `IpNet::addr` in this code:

https://github.com/diesel-rs/diesel/blob/6babd76385ba71c306ae8d19752b4f44e57cb9b1/diesel/src/pg/types/ipnet_address.rs#L111

and this code for ipv6:

https://github.com/diesel-rs/diesel/blob/6babd76385ba71c306ae8d19752b4f44e57cb9b1/diesel/src/pg/types/ipnet_address.rs#L124

There should be a distinction depending on whether we are targeting a `cidr` or `inet` field, but there is none which means all values are effectively treated as `cidr`

## Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [X] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [X] This issue can be reproduced without requiring a third party crate

weiznich:
Fixed by #4544

PR #4544, opened 2025-03-20
Fix conversion `ToSql<Inet, Pg> for Ipnet`

Zstorm999:
This is a fix for #4543. 

During serialization of `ipnet::Ipnet` into the postgres protocol, take into account the targeted type (`inet` or `cidr`) when deciding which bytes in the address should be kept. This avoid erasing part of the address if the netmask is not /32 (in ipv4) or /128 (in ipv6).

I also added a few relevant test cases.


guissalustiano (approved):
Thanks for working on this and provide test cases! It looks good to me

pksunkara (approved):

PR #4548, opened 2025-03-21
Bump dependencies

weiznich:
This commit performs some major dependency updates 

* getrand to 0.3 for diesel_tests
* cargo_metadata to 0.19 for xtask
* downcast-rs to 2.0 for diesel

We can perform these updates without breaking changes as the first two
are only for our internal crates (tests + xtask) and the later one is
only a dependency on master yet.

(Also temporary includes #4547 to let the CI pass)


pksunkara (approved):

pksunkara:
Looks like getrandom is failing the sqlite tests

Spxg:
image

It seems that it is caused by quickcheck using rand 0.8. The good news is that master has been updated to 0.9, but it has not been released yet: https://github.com/BurntSushi/quickcheck/commit/2c2cd21935e65232238595d476489c63463eb8ce

PR #4549, opened 2025-03-21
Make it possible to use copy_from/copy_to with dynamic tables

lutter:
Previously, copy_from and copy_to would only be possible with tables known at compile time. It is now possible to also use them with tables that are only known at runtime. That is achieved by changing the signature of `CopyTarget::walk_target` to take a `self` argument.

A couple things I am not sure about with this PR, but more than happy to address:
1. Since this only changes some internal method signatures, does it need additional tests?
2. Should the docs for `CopyTarget` mention anything about tables only known at runtime, as that's a very niche use case?
3. I had to resort to using an `Rc` - I hope that's not a big issue


diesel/src/pg/query_builder/copy/copy_from.rs
```diff
@@ -372,7 +377,7 @@ where
 #[must_use = "`COPY FROM` statements are only executed when calling `.execute()`."]
 #[cfg(feature = "postgres_backend")]
 pub struct CopyFromQuery<T, Action> {
-    table: T,
+    table: Rc<T>,
```
weiznich, line 380: Using an `Rc` here is unfortunately a breaking change as this turns this struct from something that implements `Send` into something that doesn't implement `Send`.

Fortunately we don't need to explicitly store table inside of `InsertableWrapper`. Instead we can just pass a reference in through the method call chain.

lutter: Thanks for the feedback! I'll incorporate it, but it might take me a little bit to get around to it

PR #4553, opened 2025-03-28
Rewite `GroupedBy::grouped_by` method

Wolveric:
This PR addresses the initial issue raised in issue #2298, by rewriting `GroupedBy::grouped_by`, having the method drop all child rows where a relationship could not be found, rather than panicking.

As this only applies to child rows where the foreign key was `NULL`, or a parent row wasn't found in the `parents` slice, this shouldn't affect existing usage of this method following its documented usage with `BelongingToDsl::belonging_to` function, as it should not produce rows with these conditions.

As some consumers may not expect this behaviour of dropping child rows, this PR also adds a section to the trait documentation, that highlights this behaviour.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -44,6 +44,18 @@ pub trait BelongsTo<Parent> {
 /// In the following relationship, User has many Posts,
 /// so User is the parent and Posts are children.
 ///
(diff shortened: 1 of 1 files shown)


mertwole (approved):
LGTM

weiznich (approved):
Thanks. Looks reasonable to change it that way.

PR #4554, opened 2025-03-29
Sqlite json/jsonb patch

JakobKlocker:
added json_patch & jsonb_patch with the test cases provided by sqlite.
I'm not completely certain if the jsonb test cases are correct, since jsonb_patch returns a jsonb, yet they pass if I compare it to a json. This was done in the existing jsonb function as well, so I guess it should be correct.

https://github.com/diesel-rs/diesel/issues/4366#issuecomment-2729139146

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -999,6 +999,25 @@ extern "SQL" {
     ///
     /// assert_eq!(json!({"a":1, "b":2, "c":3, "d":4}), result);
     ///
+    /// let result = diesel::select(json_patch::<Json, _, _>(json!({"a":[1,2],"b":2}), json!({"a":9})))
+    ///     .get_result::<Value>(connection)?;
+    ///
+    /// assert_eq!(json!({"a":9,"b":2}), result);
+    ///
+    /// let result = diesel::select(json_patch::<Json, _, _>(json!({"a":[1,2],"b":2}), json!({"a":null})))
+    ///     .get_result::<Value>(connection)?;
+    ///
+    /// assert_eq!(json!({"b":2}), result);
+    ///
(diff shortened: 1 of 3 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -959,4 +959,136 @@ extern "SQL" {
...
+    ///     .get_result::<Value>(connection)?;
+    ///
+    /// assert_eq!(json!({"a":{"x":1,"y":9},"b":3,"c":8}), result);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "json_patch"]
+    #[cfg(feature = "sqlite")]
+    fn json_patch<J: JsonOrNullableJson + SingleValue>(j: J, patch: J) -> Json;
```
weiznich, line 1027: The return type is not correct. This function (and also `jsonb_patch`) returns `Null` if either the first or second argument is nullable.

We might want something like 
```suggestion
    fn json_patch<J1: JsonOrNullableJson + SingleValue + CombinedNullableValue<J2, Json>, J2: JsonOrNullableJson + SingleValue>(j: J1, patch: J1) -> J1::Out;
```

with the `CombinedNullableValue` trait defined here:

https://github.com/diesel-rs/diesel/blob/c5cdaa4622309784cd57ec446c62a3e5d309cae0/diesel/src/pg/expression/expression_methods.rs#L3936-L3949 

(That needs to be moved to `diesel/src/expressions/nullable.rs` into a `pub(crate) private` module.

weiznich (changes requested):
Thanks for working on this and sorry for taking that long to review. 

The implementation itself already looks on the right way, but we want to adjust the signatures slightly to better handle Null values as explained in the review comments.

PR #4557, opened 2025-04-03
Change BoxableExpression SelectableExpression constraint to AppearsOnTable

oeed:
I am fairly sure `SelectableExpression` is an unnecessarily strict constraint on `BoxableExpression`. I was unable to use it in a join `.on` as a result. Instead, `AppearsOnTable` is more appropriate I believe.

As I believe this is only intended (and possible to?) use in filters and join conditions (i.e. not selections), it makes sense, as per `AppearsOnTable`.
> [AppearsOnTable] is only used in places where the nullability of a SQL type doesn’t matter (everything except select and returning)

### DIFF
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -1015,7 +1015,7 @@ pub trait BoxableExpression<QS, DB, GB = (), IsAggregate = is_aggregate::No>
 where
     DB: Backend,
     Self: Expression,
-    Self: SelectableExpression<QS>,
+    Self: AppearsOnTable<QS>,
     Self: QueryFragment<DB>,
     Self: Send,
 {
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. Unfortunately it won't be possible to merge this as that is a breaking change. A downstream crate you'll for example have a generic bound like `T: BoxableExpression` somewhere and rely on that implying that `T` implements `SelectableExpression`. 

I thing lifting that restriction would also prevent using a `BoxableExpression` trait object as part of your select clause, which is currently a supported usage of this trait.

oeed:
Ah very good point okay, no problem. I ended up writing my own similar trait anyway as I needed `DynClone`, and as you can fairly easily just copy it in to your own crate it's not a huge issue.

Issue #4565, opened 2025-04-10
Ambiguous associated item compiler error while deriving `FromSqlRow` for enum with `Row` variant.


katyo:
### Versions

- **Rust: 1.86.0**
- **Diesel: 2.2.9**
- **Database: not matter**
- **Operating System**

### Feature Flags

- **diesel:**

### What are you trying to accomplish?

I'm trying derive `FromSqlRow` for custom enum type which has variant named `Row`.

### What is the expected output?

Compiles without errors.

### What is the actual output?

```
error: ambiguous associated item
    --> .../src/types.rs:269:47
     |
2695 |         #[cfg_attr(feature = "diesel", derive(diesel::FromSqlRow, diesel::AsExpression))]
     |                                               ^^^^^^^^^^^^^^^^^^
     |
     = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
     = note: for more information, see issue #57644 <https://github.com/rust-lang/rust/issues/57644>
note: `Row` could refer to the variant defined here
    --> .../src/types.rs:273:13
     |
2739 |             Row,
     |             ^^^
note: `Row` could also refer to the associated type defined here
    --> .../diesel-2.2.9/src/deserialize.rs:262:5
     |
262  |     type Row: FromStaticSqlRow<ST, DB>;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     = note: `#[deny(ambiguous_associated_items)]` on by default
     = note: this error originates in the derive macro `diesel::FromSqlRow` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Are you seeing any additional errors?

No.

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

Seems the problem is here:
https://github.com/diesel-rs/diesel/blob/08f5e646935145a08d16944a6705c12cee97133f/diesel_derives/src/from_sql_row.rs#L39

The ambiguity happens because `Self::Row` is a associated type `Row` of trait and in same time a `Row` variant of `Self` enum.

weiznich:
Fixed by #4567

PR #4567, opened 2025-04-10
diesel_derives: Fix ambiguity problem (FromSqlRow, AsExpression)

katyo:
Related issue #4565

### DIFF
--- a/diesel_derives/src/as_changeset.rs
+++ b/diesel_derives/src/as_changeset.rs
@@ -131,14 +131,14 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
     }
 
     let changeset_owned = quote! {
-        impl #impl_generics AsChangeset for #struct_name #ty_generics
+        impl #impl_generics diesel::query_builder::AsChangeset for #struct_name #ty_generics
         #where_clause
         {
             type Target = #table_name::table;
-            type Changeset = <(#(#direct_field_ty,)*) as AsChangeset>::Changeset;
+            type Changeset = <(#(#direct_field_ty,)*) as diesel::query_builder::AsChangeset>::Changeset;
 
-            fn as_changeset(self) -> Self::Changeset {
-                (#(#direct_field_assign,)*).as_changeset()
+            fn as_changeset(self) -> <Self as diesel::query_builder::AsChangeset>::Changeset {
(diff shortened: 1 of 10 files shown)


weiznich:
Thanks for submitting this PR. That change itself looks good. I've pushed https://github.com/diesel-rs/diesel/pull/4567/commits/419debb1d4bad47037fb6e33aee9c04b6952a5a2 on top to apply the same fix to some other derives as well + make the code a bit more robust against similar problems.

Ten0:
[Rework more derives to avoid similar problems in the future](https://github.com/diesel-rs/diesel/pull/4567/commits/ec72cd9b818e7a757025a9e18b32ae2ecb38ccb8)

is a breaking change for:
```rust
/// To be used like so: #[diesel(deserialize_as = CheckSorted<Vec<Foo>>)]
pub struct CheckSorted<T>(T);
impl<ST, DB, T> Queryable<ST, DB> for CheckSorted<Vec<T>>
where
	DB: diesel::backend::Backend,
	Vec<T>: diesel::deserialize::FromStaticSqlRow<ST, DB>,
	T: Ord,
{
	type Row = Vec<T>;
	fn build(row: Vec<T>) -> diesel::deserialize::Result<Self> {
		if !row.is_sorted() {
			return Err("Expected sorted array".into());
		}
		Ok(CheckSorted(row))
	}
}
impl<T> CheckSorted<T> {
	/// hack based on how #[diesel(deserialize_as)] works (by calling .try_into())
	/// it wouldn't be possible otherwise due to foreign trait impls on foreign types
	pub fn try_into(self) -> diesel::deserialize::Result<T> {
		Ok(self.0)
	}
}
```

Which can't otherwise be implemented as an `Into` or a `From`:
image
image

I recall having written this as `.try_into()` originally specifically to allow this pattern.
Should we revert this part?

weiznich:
That's a tricky one. On the one hand you are totally right that the example is now broken and that use-case is now not possible anymore. On the other hand I'm not sure if I would say that's something that was guranteed before. At least the documentation mostly talks about the traits and no free standing method there, although the documentation itself is also ambiguous. 

That all written: I'm open to change it back as long as it cannot result in any ambiguity that we are aware of. If we change it back we should make sure that both the relevant derive code and the documentation contains hints that `try_into()` can also be a free standing method. Maybe we should even have a test for that then.

PR #4575, opened 2025-04-19
cli: use distinct `DIESEL_LOG` logging filter env variable

mrnossiom:
I use diesel on a lot of Rust projects, and I often set the `RUST_LOG` env var, either manually or automatically with tools like direnv.

This means that I end up seeing diesel-cli logs, even if I absolutely don't care about them.

Rust itself switched from `RUST_LOG` over to `RUSTC_LOG` quite a while ago, with a similar motivation.

Since the `diesel` CLI will be used in Rust development environments, it makes sense to :

- Stop respecting the `RUST_LOG` env var
- use a custom `DIESEL_LOG` for logging configuration instead

I don't know how you feel about it from a breaking change point of view. I feel like this is more breaking for developers that are used to `RUST_LOG` when debugging `diesel-cli`.

Maybe you would prefer `DIESEL_CLI_LOG` instead of `DIESEL_LOG` ?

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ in a way that makes the pools suitable for use in parallel tests.
 * Diesel CLI now ensures that migration versions are always unique. If it fails to generate a unique version, it will return an error. The new version format remains compatible with older Diesel versions.
 * Updated `ipnetwork` to allow version 0.21.
 
+### Changed
+
+* Use distinct `DIESEL_LOG` logging filter env variable instead of the default `RUST_LOG` one (#4575)
+
 ## [2.2.2] 2024-07-19
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


weiznich (changes requested):
Thanks for submitting. I think it's fine to do that given that the rust project itself has done such changes as minor release in the past as well. I would like to add a note about that to our changelog (https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md), otherwise this is ready to go.

mrnossiom:
Done!

Btw, is it normal that the unreleased entry has two "Fixed" sections?

weiznich (approved):

weiznich:
Thanks for teh update

> Btw, is it normal that the unreleased entry has two "Fixed" sections?

That's something that got broken at some point. Needs to be cleaned up before the next minor release.

Issue #4572, opened 2025-04-16
Addressing handling duplicate parents in `GroupedBy` methods


Wolveric:
To clarify: This is not necessarily a bug, but rather a proposal for an addition to documentation, addressing some potentially unintuative behaviour...

## Problem Description

As it stands, the generic implementation of `GroupedBy` uses the following implementation:

```rs
impl<'a, Parent: 'a, Child, Iter> GroupedBy<'a, Parent> for Iter
where
    Iter: IntoIterator<Item = Child>,
    Child: BelongsTo<Parent>,
    &'a Parent: Identifiable,
    Id<&'a Parent>: Borrow<Child::ForeignKey>,
{
    fn grouped_by(self, parents: &'a [Parent]) -> Vec<Vec<Child>> {
        use std::collections::HashMap;
        use std::iter;

        let mut grouped: Vec<_> = iter::repeat_with(Vec::new).take(parents.len()).collect();

        let id_indices: HashMap<_, _> = parents
            .iter()
            .enumerate()
            .map(|(i, u)| (u.id(), i))
            .collect();

        self.into_iter()
            .filter_map(|child| {
                let fk = child.foreign_key()?;
                let i = id_indices.get(fk)?;

                Some((i, child))
            })
            .for_each(|(i, child)| grouped[*i].push(child));

        grouped
    }

// `GroupedBy::try_grouped_by` omitted for brevity. Behaviour still applies.
...
}
```

Provided the IDs of the rows in `parents` are entirely unique, the intended behaviour of this implementation will be to add each index of `parents` as a unique entry to `id_indicies`, which are subsequently looked up using the foreign keys of the child rows, and the associated indexes used to add the child row to to an item in `grouped`.

The unintuative behaviour in question here arises if `parents` contains rows with duplicate IDs. If this occurs, items in `grouped` will have no contents, even if they should have child entries associated with them, because they will instead be added to the index of the last row in `parents`. These results are a knock on effect of `id_indicies` being a `HashMap` here, because whenever a given ID has an existing entry, its value will be replaced with the value of the most recent index.

Whilst this behaviour is technically valid, is handled gracefully, and does not break existing example use-cases, without knowing about this internal implementation, it may be somewhat unintuative... notably, if results were ever inspected, such as reading the results of an API request, consumers may notice they receive duplicates of the parent rows, but the child row data for some of them are empty.

## Proposal

[...]

PR #4581, opened 2025-04-23
Document `GroupedBy` handling duplicate parents

Wolveric:
Whilst graceful, the behaviour of `GroupedBy::grouped_by` and `GroupedBy::try_grouped_by` when encountering duplicate parent rows, is somewhat unintuative.
This PR adds a section to trait documentation, outlining this behaviour, and recommended remediation.

### DIFF
--- a/diesel/src/associations/belongs_to.rs
+++ b/diesel/src/associations/belongs_to.rs
@@ -64,12 +64,6 @@ pub trait BelongsTo<Parent> {
 /// If this is not true, child rows will be grouped against the last row
 /// that produced a given ID.
 ///
-/// This behaviour should typically be a result of `parents` containing
-/// duplicates of one or more rows.
-/// This may happen if parents rows are selected based on a join query,
-/// as a copy of the parent row will be retrieved
-/// for each child row they were joined with.
-///
 /// As a result, it is recommended to add [`QueryDsl::distinct`]
 /// to queries that may exhibit this behaviour,
 /// or use [`Vec::sort`] and [`Vec::dedup`] on the loaded rows,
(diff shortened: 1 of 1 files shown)


diesel/src/associations/belongs_to.rs
```diff
@@ -56,6 +56,25 @@ pub trait BelongsTo<Parent> {
 /// If discarding these rows is undesirable, it may be preferable to use
 /// [`GroupedBy::try_grouped_by`].
 ///
+/// # Handling Duplicate Parent Rows
+///
+/// Both [`GroupedBy::grouped_by`] and [`GroupedBy::try_grouped_by`]
+/// expect all of the elements of `parents` to produce a unique value
+/// when calling [`Identifiable::id`].
+/// If this is not true, child rows will be grouped against the last row
+/// that produced a given ID.
```
weiznich, line 65: I wouldn't document which item gets the child rows so that we can change the implementation of the function later.

Wolveric: Alternatively, the note can say essentially, *"Ensure all of the parent rows are unique, or the child row may not be grouped at the index you expect."*

weiznich (approved):

diesel/src/associations/belongs_to.rs
weiznich: Thanks for the update. The new text looks better. As written in #4572 I'm also open to accept a change that just does the right thing there (duplicates all childs for all duplicated parent instances). I think that might even be the better solution.

Wolveric: The issue in that case, is it would be a breaking change. Currently the implementation doesn't require `Clone`, which that would introduce as a requirement...

Otherwise, that would also require changing the contents of the `HashMap`s, likely as `HashMap<_, Vec<usize>>` or `HashMap<_, LinkedList<usize>>`, which seems a little strange...

weiznich: Ah, right. That bound is missing, so that seems to be the best solution for now.

Wolveric: Perhaps as *another* association trait, if that would be considered valid behaviour... because that would be a significant jump in cost, in cases without duplicate parents...

PR #4593, opened 2025-05-02
Commented .env.sample file for the limitations in PG_EXAMPLE_DATABASE_URL

### LINKED ISSUE #4419: shell completion tests fail with EmptyHost
## Setup

My .env:

    PG_DATABASE_URL=postgresql://thk@/diesel_test?host=/var/run/postgresql
    PG_EXAMPLE_DATABASE_URL=postgresql://thk@/diesel_example?host=/var/run/postgresql

### Versions

- **Diesel:** master branch on January 4th 2024, commit id: 977a2b0f7c9535ee40797eed5d79ce10da7b7b89

## Problem Description

I ran `cargo xtask run-tests --no-doc-tests postgres` and all 3 tests in diesel_cli::tests completion_generation failed at

    diesel_cli/tests/support/project_builder.rs:111:51:
    called `Result::unwrap()` on an `Err` value: EmptyHost

## Cause

line 111 is:

        let mut db_url = url::Url::parse(&var_os).unwrap();

The url crate is used to parse a postgres connection url and I specified the host as a parameter.

I can just changed my .env to:

    PG_DATABASE_URL=postgresql://thk@%2Fvar%2Frun%2Fpostgresql/diesel_test
    PG_EXAMPLE_DATABASE_URL=postgresql://thk@%2Fvar%2Frun%2Fpostgresql/diesel_example

And now the shell completion tests worked.

## Proposal

Either fix the test code or add a comment in .env.sample about this issue.

zaira-bibi:
Closes #4419.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,8 @@
 # The database to use when testing against Postgres.
 PG_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/diesel_test
 # The database to use when running the Postgres examples during testing.
+# Note: This should be a Postgres database url only. If you specify connection parameters (like host) separately 
+# in the configuration, it may cause parsing errors due to limitations in the 'url' crate. 
 PG_EXAMPLE_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/diesel_example
 
 # The database to use when testing against MySQL.
--- a/.env.sample
+++ b/.env.sample
@@ -1,8 +1,16 @@
 # The database to use when testing against Postgres.
 PG_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/diesel_test
 # The database to use when running the Postgres examples during testing.
-# Note: This should be a Postgres database url only. If you specify connection parameters (like host) separately 
-# in the configuration, it may cause parsing errors due to limitations in the 'url' crate. 
+# Note: Specify parameters in the special URL fields, not as query parameters. 
+#
+# For example:
+# INCORRECT: postgresql://username@/dbname?host=/var/run/postgresql
+# CORRECT:   postgresql://username@%2Fvar%2Frun%2Fpostgresql/dbname
+#
+# This is because the underlying 'url' crate strictly parses the URL structure,
+# whereas Postgres's native connection string format supports query parameters.
+#
+# Always URL-encode special characters in paths ('/' becomes '%2F') when using them in connection URLs.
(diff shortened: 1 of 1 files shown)


.env.sample
```diff
@@ -1,6 +1,8 @@
 # The database to use when testing against Postgres.
 PG_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/diesel_test
 # The database to use when running the Postgres examples during testing.
+# Note: This should be a Postgres database url only. If you specify connection parameters (like host) separately 
+# in the configuration, it may cause parsing errors due to limitations in the 'url' crate.
```
Mingun, line 5: I think, this does not explain cause of the original problem at all

zaira-bibi: Should there be more details about the url crate not handling other formats?

Mingun: It seems to me, that the key points are:
- you cannot specify parameters using query parameters (a part after `?`), while in postgres you can (I don't know, just my assumption)
- you should specify parameters in the special URL fields because that is how URL parsing library can handle them (this is implementation detail)
- when you specify parameters, you should escape them as required

With that info it would be clear, why URL is written in that ugly way

zaira-bibi: Thank you so much for the detailed feedback!

Would this be a more appropriate note:

`# Note: Specify parameters in the special URL fields, not as query parameters. For example:
# INCORRECT: postgresql://username@/dbname?host=/var/run/postgresql
# CORRECT:   postgresql://username@%2Fvar%2Frun%2Fpostgresql/dbname
#
# This is because the underlying 'url' crate strictly parses the URL structure,
# whereas Postgres's native connection string format supports query parameters.
# Always URL-encode special characters in paths ('/' becomes '%2F') when using them in connection URLs.`

zaira-bibi: Thank you so much for the detailed feedback!

Would this be a more appropriate note?

```
Note: Specify parameters in the special URL fields, not as query parameters. 

For example:
INCORRECT: postgresql://username@/dbname?host=/var/run/postgresql
CORRECT:   postgresql://username@%2Fvar%2Frun%2Fpostgresql/dbname

This is because the underlying 'url' crate strictly parses the URL structure,
whereas Postgres's native connection string format supports query parameters.

Always URL-encode special characters in paths ('/' becomes '%2F') when using them in connection URLs.
```

Mingun: Yes, much better

zaira-bibi: Thank you! I've committed the change.

PR #4594, opened 2025-05-05
Add domain types configuration

lsunsi:
This PR intends to be the point of discussion of a possible feature discussed in #4592.

### Concept
Domain types are regular types with some added constraints. [Reference](https://www.postgresql.org/docs/current/domains.html).

The current behavior when using them with diesel is that diesel prints the underlying type, which makes sense as a default. As an alternative behavior, this PR proposes that a new type is generated if the user wants to add behavior to the domain type on the application layer.

### Details

I'm not entirely happy with the last commit because I had to use "boxed" in order to have two distinct queries in the same variable, which still having to add "if" on the flag. If you can think of a better way to express the logic in diesel I'd love to learn it and change the implementation.

### Questions

1. Should other databases have support or it? I'm not sure if domain types are a pg-only thing, since the information schema is actually a SQL thing from what I can gather. Anyway, I'd enjoy some guidance here as well.
2. Is a boolean flag enough? The user won't be able to configure which domain types to render, just "all or nothing" behavior. Maybe it'd make sense to allowlist some domain types in the configuration with their postgres name?


a-moreira (approved):

diesel_cli/tests/print_schema/print_schema_domain_types_enabled/postgres/expected.snap
```diff
@@ -0,0 +1,35 @@
...
+/// A module containing custom SQL type definitions
+///
+/// (Automatically generated by Diesel.)
+pub mod sql_types {
+    /// The `posint` SQL type
+    ///
+    /// (Automatically generated by Diesel.)
+    #[derive(diesel::query_builder::QueryId, Clone, diesel::sql_types::SqlType)]
+    #[diesel(postgres_type(name = "posint"))]
+    pub struct Posint;
```
Ten0, line 17: Just laying some thoughts here (but I'm not sure we should do what's suggested below): this loses information of what the underlying type is, which makes it the responsibility of the user to make sure that serialization is appropriate.

We might want to add this data to the type here, such as a new `Domain` derive and `underling_type = "..."` annotation to the `SqlType` derive, that would implement a corresponding `Domain` trait, and allow the user to make use of this type in their implementations of FromSql and ToSql (or even make the relevant implementations conditionally on implementations of some new traits that would define the mapping between input rust types and rust types that have `ToSql/FromSql<UnderlyingType>`) to fully close the door for mistakes here.

If we were to add that later by default, that would be a breaking change for users who already manually define their `FromSql`/`ToSql` implementations for these types.

But I'm not sure if that's ever worth the extra maintenance cost within Diesel itself...
Probably that wouldn't be very consistent with Diesel not supporting enums out-of-the-box, where ideally the CLI would also infer enum variants...

diesel_cli/src/config.rs
```diff
@@ -465,6 +465,8 @@ pub struct PrintSchema {
     pub custom_type_derives: Option<Vec<String>>,
     #[serde(default)]
     pub sqlite_integer_primary_key_is_bigint: Option<bool>,
+    #[serde(default)]
+    pub domain_types: bool,
```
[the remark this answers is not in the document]
lsunsi: I'll take it, 321337694fcc56275a20b96323543d467e4d50ed

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -90,25 +90,47 @@ pub fn get_table_data(
...
+                udt_name,
+                udt_schema.nullable(),
+                __is_nullable,
+                character_maximum_length,
+                col_description(regclass(table), ordinal_position),
+            ))
+            .filter(table_name.eq(&table.sql_name))
+            .filter(table_schema.eq(schema_name))
+            .into_boxed()
+    };
```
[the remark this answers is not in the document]
lsunsi: It worked, today I learned something. Do you think it looks better or worse? 57b1b7d69ce0dbf897fdbb783246db31cdd52fe6

diesel_cli/tests/print_schema/print_schema_domain_types_enabled/postgres/expected.snap
lsunsi: Yeah I fully agree that some information is being lost here and this should be avoided. I'd be willing to try something along the lines you suggested with some help (since I'm a newbie in this project).

And about this breaking compatibility, is there a way in diesel to add features in an "unstable" manner so it can be worked over multiple releases before being stabilized? If so, maybe we should think of this feature in layers, but I'm not sure about it either.

lsunsi:
@Ten0 What do you think of instead of a boolean we have a list of type names to be considered as is? The reason I suggest this is that the implementation would look very similar, while adding more control for the users. Instead of all-or-nothing domain types, the user could allowlist the domains they want.

If it makes sense, we could have "true" representing all, and a list representing allowlist, or something along these lines. I'm not sure about it, but I'd like to hear what you think about it

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -90,25 +90,46 @@ pub fn get_table_data(
...
         Some(ref name) => Cow::Borrowed(name),
         None => Cow::Owned(Pg::default_schema(conn)?),
     };
 
+    let (select_name, select_schema) = if domain_types_enabled {
+        define_sql_function!(#[sql_name = "coalesce"] fn coalesce_maybe<T: SingleValue>(x: Nullable<T>, y: Nullable<T>) -> Nullable<T>);
+        define_sql_function!(#[sql_name = "coalesce"] fn coalesce_strict<T: SingleValue>(x: Nullable<T>, y: T) -> T);
+
+        type BoxedTextExpr = Box<dyn BoxableExpression<columns, Pg, SqlType = Text>>;
+        type BoxedMaybeTextExpr = Box<dyn BoxableExpression<columns, Pg, SqlType = Text>>;
```
Ten0, line 108: Haha I was like, how does this work without the `Nullable<Text>` here?
The answer is it doesn't 😅
```
 error[E0271]: type mismatch resolving `<coalesce_maybe<Text, domain_schema, Nullable<udt_schema>> as Expression>::SqlType == Text`
   --> diesel_cli/src/infer_schema_internals/pg.rs:112:13
    |
112 |             Box::new(coalesce_maybe(domain_schema, udt_schema.nullable())) as BoxedMaybeTextExpr,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Text`, found `Nullable<Text>`
    |
    = note: expected struct `diesel::sql_types::Text`
               found struct `diesel::sql_types::Nullable<diesel::sql_types::Text>`
```

diesel_cli/src/infer_schema_internals/pg.rs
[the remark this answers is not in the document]
Ten0: Yeah that looks better because it's possible that we might add more other things to the query and it's good that these and filter conditions are not repeated.

diesel_cli/src/infer_schema_internals/pg.rs
lsunsi: Yes, I'm sorry! I was in a hurry so I missed it, check this out 8f7a58282e1bdc82b9bec5c1fa45587fd1d4c27c

diesel_cli/tests/print_schema/print_schema_domain_types_enabled/postgres/expected.snap
lsunsi: @Ten0 I'm re-reading your comment and I'm positive this is the right path, but I'm not sure how to begin to propose a solution. Is there any pointer you could give me to similar code I could read up on in order to get up to speed on this implementation?

Ten0: > but I'm not sure how to begin to propose a solution

I understand that this comment is blocked on me giving this issue some more thought. I'll come back to you when I have time to do that 😊🙏

For me the main question is whether we even want to do this here:

> But I'm not sure if that's ever worth the extra maintenance cost within Diesel itself...
Probably that wouldn't be very consistent with Diesel not supporting enums out-of-the-box, where ideally the CLI would also infer enum variants...

Maybe other reviewers will have an opinion.

lsunsi: Of course, I wasn't trying to pressure you or anything @Ten0 , sorry if it came off this way.

Is there any way we could release something without having stability guarantees on it? Maybe a feature a feature or something, maybe this could relieve some pressure and give us some time to think on this.

weiznich: We do currently not have a way to say: This feature is unstable, that by itself would need to be implemented first. That written: I'm not opposed to have something like that as unstable feature that can be explicitly enabled via an `unstable.domain_type = true` flag in the `diesel.toml` configuration.

Finally for the actual discussion: Given that domain types often introduce custom requirements on the underlying types it's hard to see how providing the built-in `FromSql`/`ToSql` impls for the underlying types would be helpful. I would guess that users ~always want to express these requirements in their rust code as well, so having these impls by default might be not helpful for them.

As for the discussion whether this should be a boolean flag or a list of type names: If somewhat possible I would prefer a list of type names or even a list of type name regexes there so that users can have finer control around where thy want domain types and where not.

lsunsi: Hey @weiznich , thanks for the feedback!

I like the idea of prefixing with unstable, seems like a simpler way to avoid this possible improvement. That said, I'll leave this discussion for when we're comfortable with the PR so we know exactly what implementation we are talking about!

About the list of type names, I updated the implementation on [14cbf02](https://github.com/diesel-rs/diesel/pull/4594/commits/14cbf029272c3c28539361bc7b584a8c0f965872) trying to implement a version of this idea. I think it looks way better than the boolean, do you mind checking it out and giving feedback?

I updated the tests too to reflect usage.

lsunsi: Should I do anything special for the `diesel.toml` file to be read on the test? I swear the commit passed the test once but I changed something and can't seem to do it again. haha.

lsunsi: Figured it out [74f763d](https://github.com/diesel-rs/diesel/pull/4594/commits/74f763d72f4e6f18d85a344a75992b93f9c0852d), added the CLI option

weiznich (commented):
Thanks for the update. I think this is nearly ready to be merged, beside the minor comments left.

Otherwise as this is a relevant new feature: Could you add an entry in our `CHANGELOG.md` file for this?

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -90,88 +89,78 @@ pub fn get_table_data(
...
-            row.2,
-            row.3 == "YES",
-            row.4
+            let max_length = row
+                .max_length
                 .map(|n| {
-                    std::convert::TryInto::try_into(n).map_err(|e| {
-                        format!("Max column length can't be converted to u64: {e} (got: {n})")
+                    n.try_into().map_err(|_| {
+                        // format!("Max column length can't be converted to u64: {e} (got: {n})")
```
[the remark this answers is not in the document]
lsunsi: I left it commented out but I couldn't keep it in first implementation and I knew I'd have to come around to it. Trying it now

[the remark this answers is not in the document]
lsunsi: How about like this? I think it's the same of what we already had going on 578f9bda71887da6ad0b71ea9e90af1b3c8894a6

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -326,7 +315,7 @@ mod test {
                 &mut connection,
                 &table_2,
                 &ColumnSorting::OrdinalPosition,
-                false
+                &[]
```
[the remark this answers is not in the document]
lsunsi: How about this? 755750a3a34442d9b040ce9f1ab8cd562415587a

diesel_cli/src/config.rs
```diff
@@ -466,7 +466,7 @@ pub struct PrintSchema {
     #[serde(default)]
     pub sqlite_integer_primary_key_is_bigint: Option<bool>,
     #[serde(default)]
-    pub use_domains_as_column_sql_types: bool,
+    pub domains_as_custom_types: Vec<Regex>,
```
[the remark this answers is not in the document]
lsunsi: Like that? 871d8e0845942f9cfd41db3b7b1fcff288217b32
Should I change the CLI parameter as well?

lsunsi:
@weiznich is this okay for the changelog? [a388b4a](https://github.com/diesel-rs/diesel/pull/4594/commits/a388b4a427a6c71df76f7b2daf2e34ca6b0e8937)

diesel_cli/src/cli.rs
```diff
@@ -335,6 +335,14 @@ pub fn build_cli() -> Command {
                 .number_of_values(1)
                 .help("A list of derives to implement for every automatically generated SqlType in the schema, separated by commas."),
         )
+        .arg(
+            Arg::new("domains-as-custom-types")
```
weiznich, line 339: I would use the pg prefix here as well.

diesel_cli/src/config.rs
```diff
@@ -381,6 +381,15 @@ impl Config {
                 let derives = derives.cloned().collect();
                 config.custom_type_derives = Some(derives);
             }
+
+            if let Some(domains) = matches.get_many::<String>("domains-as-custom-types") {
+                let regexes: Vec<String> = domains.cloned().collect();
```
weiznich, line 386: Technically we don't need that collect, right?

Possibility even the cloned can be skipped given that the regex takes a &str. 

That's a minor nit, it shouldn't matter much, so feel free to ignore.

weiznich (commented):
Thanks for updating. The changelog entry looks good to me. I would like to change the command line flag name as noted, otherwise this is ready from my point of view.

diesel_cli/src/cli.rs
lsunsi: [9ebddb4](https://github.com/diesel-rs/diesel/pull/4594/commits/9ebddb4d7d3052043690a2398dac6192af13d88d)

diesel_cli/src/config.rs
lsunsi: Absolutely wouldn't want to ignore, here we go b213b2e44ae3942e172d109e6a90712804062dc6

diesel_cli/tests/print_schema/print_schema_domain_types_enabled/postgres/expected.snap
Ten0: > But I'm not sure if that's ever worth the extra maintenance cost within Diesel itself...

Let's decide that for now it isn't worth and not add an additional trait. (Discussed with weiznich today) 

Regex seems OK. `pg_domains_as_custom_types` setting name seems OK.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ in a way that makes the pools suitable for use in parallel tests.
 * Added a `#[diesel::declare_sql_function]` attribute macro to easily define support for 
   multiple sql functions at once via an `extern "SQL"` block
 * Support `[print_schema] allow_tables_to_appear_in_same_query_config = "fk_related_tables"` to generate separate `allow_tables_to_appear_in_same_query!` calls containing only tables that are related through foreign keys. (Default: `"all_tables"`.) It is not possible to build queries using two tables that don't appear in the same `allow_tables_to_appear_in_same_query!` call, but that macro generates O(n²) rust code, so this option may be useful to reduce compilation time. ([#4333](https://github.com/diesel-rs/diesel/issues/4333))
+* Add `[print_schema] pg_domains_as_custom_types` parameter to generate custom types for [PostgreSQL domains](https://www.postgresql.org/docs/current/domains.html) that match given regex. (Default: `empty`.) This option allows an application to selectively give special meaning for the serialization/deserialization of these types, avoiding the default behavior of treating the domain as the underlying type. ([#4592](https://github.com/diesel-rs/diesel/discussions/4592))
```
Ten0, line 30: Default: empty : an empty regex matched any type so I would expect all domains to use custom types, which would be a breaking change?
This probably wants to be an option instead, with default None?

lsunsi: I think the default makes sense if the parameter is a list (which it currently is), so I changed the changelog to better reflect that `[e76a0a3](https://github.com/diesel-rs/diesel/pull/4594/commits/e76a0a37b8c6c637c4775eeb04edc397126e897e)`.

In this sense, empty doesn't mean "empty regex"  which matches everything, it means "empty list of regex" which matches nothing and makes the feature non-breaking.

Does that make sense to you @Ten0 and @weiznich ?

It could be argued it should be an optional Regex since the user can use the regex to make a list with "|", but there are already other parameters (at least one) that is a list of regexes.

diesel_cli/tests/print_schema/print_schema_domain_types_enabled/postgres/expected.snap
lsunsi: Deal @Ten0

CHANGELOG.md
Ten0: That makes sense. Isn't "regexes" correct english?
Either way what do you think about writing "Default: `[]`"?

lsunsi: @Ten0 [3169d54](https://github.com/diesel-rs/diesel/pull/4594/commits/3169d54160d61f31d8b19c42757d3048e00eb902) how about this?

Ten0: Ah I missed it was written "any regex in the given list".
Let's write "any of the regexes in the given list, Default: []" just for correct english and I think it will be super clear! 😊

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ in a way that makes the pools suitable for use in parallel tests.
 * Added `wasm32-unknown-unknown` target support for sqlite backend.
 * Add support for the `CAST` operator
 * Support `[print_schema] allow_tables_to_appear_in_same_query_config = "none"` to generate no `allow_tables_to_appear_in_same_query!` calls. (Default: `"all_tables"`.). ([#4333](https://github.com/diesel-rs/diesel/issues/4333))
+* Add `[print_schema] pg_domains_as_custom_types` parameter to generate custom types for [PostgreSQL domains](https://www.postgresql.org/docs/current/domains.html) that matches any regexes in a given list. (Default: `[]`.) This option allows an application to selectively give special meaning for the serialization/deserialization of these types, avoiding the default behavior of treating the domain as the underlying type. ([#4592](https://github.com/diesel-rs/diesel/discussions/4592))
```
weiznich, line 33:
```suggestion
* Add `[print_schema] pg_domains_as_custom_types` parameter to generate custom types for [PostgreSQL domains](https://www.postgresql.org/docs/current/domains.html) that matches any of the regexes in the given list. (Default: `[]`.) This option allows an application to selectively give special meaning for the serialization/deserialization of these types, avoiding the default behavior of treating the domain as the underlying type. ([#4592](https://github.com/diesel-rs/diesel/discussions/4592))
```

weiznich (approved):

PR #4596, opened 2025-05-09
Add `#[diagnostic::do_not_recommend]` to the `FromSql` impl for `String`

weiznich:
This slightly tweaks the error message to not mention `*const str` anymore. All error messages now mention `String` directly instead. I feel that this makes it clearer what's actually wrong there.


guissalustiano (approved):
Nice! Thanks!

PR #4598, opened 2025-05-13
Add print-schema allow table config option: None

lochetti:
Add the `none` option to the `allow_tables_to_appear_in_same_query_config` config. It will generate no allows at the schema file.
Thie pr relates to this [comment](https://github.com/diesel-rs/diesel/issues/4333#issuecomment-2871626144)

### DIFF
--- a/diesel_cli/tests/print_schema.rs
+++ b/diesel_cli/tests/print_schema.rs
@@ -360,7 +360,6 @@ fn print_schema_fk_related_tables() {
 }
 
 #[test]
-#[cfg(feature = "postgres")]
 fn print_schema_allows_none() {
     test_print_schema(
         "print_schema_allows_none",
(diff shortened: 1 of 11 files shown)


diesel_cli/tests/print_schema.rs
```diff
@@ -359,6 +359,15 @@ fn print_schema_fk_related_tables() {
     )
 }
 
+#[test]
+#[cfg(feature = "postgres")]
```
weiznich, line 363: I would prefer running this test for all backends. There is nothing stopping us from doing that.

weiznich (commented):
Thanks for opening this PR. The implementation looks good for me, but I would like to extend the test coverage a bit before merging. Would it be possible to port the tests for the other backends as well?

diesel_cli/tests/print_schema.rs
lochetti: Sure I will extend the test coverage!

lochetti: @weiznich I think I could make the test run for psql, mysql and sqlite. Let me know if something is wrong! Thanks!

weiznich (approved):
Thanks, looks good now

PR #4601, opened 2025-05-15
Add json, Uuid cast operator support

Sagebati:
I added JsonB and Uuid cast from text and to text support for the newish cast operator

I don't know, if cast() must compile check types that can be cast without failure (ex: int to text). If it's the case I can add a assume_cast() or faillible_cast() where the cast can fail.

### DIFF
--- a/diesel/src/expression/cast.rs
+++ b/diesel/src/expression/cast.rs
@@ -160,11 +160,45 @@ pub trait CastsTo<ST> {}
 impl<ST1, ST2> CastsTo<sql_types::Nullable<ST2>> for sql_types::Nullable<ST1> where ST1: CastsTo<ST2>
 {}
 
-impl CastsTo<sql_types::Int4> for sql_types::Int8 {}
-impl CastsTo<sql_types::Int8> for sql_types::Int4 {}
-#[cfg(feature = "postgres_backend")]
-impl CastsTo<sql_types::Text> for sql_types::Uuid {}
-impl CastsTo<sql_types::Text> for sql_types::Int4 {}
-impl CastsTo<sql_types::Text> for sql_types::Int8 {}
-impl CastsTo<sql_types::Text> for sql_types::Jsonb {}
-impl CastsTo<sql_types::Text> for sql_types::Json {}
+macro_rules! casts_impl {
+    (
+        $(
+            $($feature: literal : )? ($to: tt <- $from: tt),
+        )+
+    ) => {
+        $(
+            $(#[cfg(feature = $feature)])?
+            impl CastsTo<sql_types::$to> for sql_types::$from {}
+        )+
+    };
+}
+
+casts_impl!(
+    (Bool <- Int4),
+    (Bool <- Int8),
+    (Bool <- Float4),
+    (Bool <- Float8),
+    (Float4 <- Float8),
+    (Float4 <- Int4),
+    (Float4 <- Int8),
+    (Float8 <- Float4),
+    (Float8 <- Int4),
+    (Float8 <- Int8),
+    (Int8 <- Bool),
+    (Int8 <- Int4),
+    (Int8 <- Float4),
+    (Int8 <- Float8),
+    (Int4 <- Bool),
+    (Int4 <- Int8),
+    (Int4 <- Float4),
+    (Int4 <- Float8),
+    (Text <- Bool),
+    (Text <- Float4),
+    (Text <- Float8),
+    (Text <- Int4),
+    (Text <- Int8),
+    (Text <- Date),
+    (Text <- Json),
+    (Text <- Jsonb),
+    "postgres_backend": (Text <- Uuid),
+);
--- a/diesel/src/expression/cast.rs
+++ b/diesel/src/expression/cast.rs
@@ -158,17 +158,17 @@ where
     }
 }
 
-/// Marker trait: this SQL type (`Self`) can be casted to the target SQL type, but some values can be invalid
-pub trait FaillibleCastsTo<ST> {}
+/// Marker trait: this SQL type (`Self`) can be cast to the target SQL type, but some values can be invalid
+pub trait FallibleCastsTo<ST> {}
 
-impl<ST1, ST2> FaillibleCastsTo<sql_types::Nullable<ST2>> for sql_types::Nullable<ST1> where
+impl<ST1, ST2> FallibleCastsTo<sql_types::Nullable<ST2>> for sql_types::Nullable<ST1> where
     ST1: CastsTo<ST2>
 {
 }
 
-/// Marker trait: this SQL type (`Self`) can be casted to the target SQL type
+/// Marker trait: this SQL type (`Self`) can be cast to the target SQL type
 /// (`ST`) using `CAST(expr AS target_sql_type)`
-pub trait CastsTo<ST>: FaillibleCastsTo<ST> {}
+pub trait CastsTo<ST>: FallibleCastsTo<ST> {}
 
 impl<ST1, ST2> CastsTo<sql_types::Nullable<ST2>> for sql_types::Nullable<ST1> where ST1: CastsTo<ST2>
 {}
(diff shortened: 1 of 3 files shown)


diesel/src/expression/cast.rs
```diff
@@ -121,11 +122,15 @@ type_name! {
...
         Timestamp => "timestamp",
+        "uuid"; Uuid => "uuid",
+        Json => "json",
+        Jsonb => "jsonb",
     }
     diesel::mysql::Mysql: "mysql_backend" {
         Int4 => "integer",
         Int8 => "integer",
         Text => "char",
+
```
guissalustiano, line 133: Could you remove this extra space?

guissalustiano (approved):
Thanks for opening the PR! It looks good to me, just a minor extra space

diesel/src/expression/cast.rs
```diff
@@ -96,10 +96,11 @@ where
 }
 
 macro_rules! type_name {
-    ($($backend: ty: $backend_feature: literal { $($type: ident => $val: literal,)+ })*) => {
+    ($($backend: ty: $backend_feature: literal { $($($feature: literal;)? $type: ident => $val: literal,)+ })*) => {
```
Ten0, line 99: It looks like `?` would feel more intuitive to read than `;` if possible.

diesel/src/expression/cast.rs
```diff
@@ -121,11 +122,15 @@ type_name! {
         Interval => "interval",
         Time => "time",
         Timestamp => "timestamp",
+        "uuid"; Uuid => "uuid",
```
Ten0, line 125: Is the Uuid **sql** type not defined unconditionally on whether the uuid feature is enabled? (If yes then cfg directives below should probably also be removed)

diesel/src/expression/cast.rs
```diff
@@ -157,6 +162,16 @@ impl<ST1, ST2> CastsTo<sql_types::Nullable<ST2>> for sql_types::Nullable<ST1> wh
 {}
 
 impl CastsTo<sql_types::Int8> for sql_types::Int4 {}
+impl CastsTo<sql_types::Int8> for sql_types::Text {}
```
Ten0, line 165: Yes I think I like the name `faillible_cast` you proposed for this kind of conversions: it feels like that would enable using the regular `cast` with more confidence, whereas otherwise every usage would have to be analyzed with great care when reviewing for correctness 🙂

diesel/src/expression/cast.rs
Ten0: Closing because will be removed as per other comment. (I checked and indeed Uuid exists without the feature.)

diesel/src/expression/cast.rs
Sagebati: Good catch, done !

diesel/src/expression/cast.rs
Sagebati: I must misread the doc, thanks, I dropped the uuid feature

diesel/src/expression/cast.rs
Sagebati: I personally prefer assume because the similarity with `assume_not_null()` but faillible is ok

Sagebati:
Took into account the feedback !

- Remove bad formatting 
- Removed the uuid feature
- Added more types to `CastTo
- Added `faillible_cast()` to support, user checked cast
- Added jsonb and json support for sqlite

diesel/src/expression/cast.rs
Ten0: IMO clarity generally wins over similarity.

For `assume_not_null` I'm specifically trying to not convey information about behavior, because behavior may range from errors to wrong values being propagated into applicative code.

In this case however my understanding is that the only thing that can happen is an error at the database level, which is rather "safe" (compared to, say, incorrect values).

If I were to not know what this function does, and I were to read `assume_cast`, first I may not know whether that means that we assume that this is already casted, but even if I interpreted that correctly I wouldn't know what happens if the assumption is wrong, which would likely be something I'd have to look up in the documentation to assess the risk level and amount of care I have to put in making sure this never ever happens.

Generally for me "assume" is a keyword for "pay very special attention to what's written here, because the consequences if it's wrong are undefined", whereas here the consequences are well defined and it may useful for the user to know at a glance what they are.

While similarity may help with finding the function, I believe that error message customization on the trait bound that's required to access the infallible `cast()` version will cover well for the typical user story. That being said I'm not generally opposed to other names (e.g. if you had a strong feeling in favor of `try_cast()`). But `assume` is too intense for this case IMO.

diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -483,6 +483,40 @@ pub trait ExpressionMethods: Expression + Sized {
         cast::Cast::new(self)
     }
 
+    /// Generates a `CAST(expr AS sql_type)` expression, this version does not check the castability, like [`cast()`](Self::cast).
```
Ten0, line 486: Ooh I thought we were going to still check "castability" in the sense that it "may work depending on the value", but not "castability" in the sense that it "will always work from this type to that type".

Sagebati: Maybe I misunderstood the initial concept, there are 3 concepts here:

1. I can always cast from this type T1 to T2 for all values.  => cast()
2. I can always cast from T1 to T2 for some values but some values may fail 
3. Free casting, only if the type is has a type_name => fallible_cast() 


Do you want me go for 2 instead of 3 ? I think it could be better yes

Ten0: Alright if we agree then let's go for version 2 😊

Sagebati: Great ! 

Can Int8 -> Int4 fail ? Or is it just semantically wrong ?

diesel/src/expression/cast.rs
```diff
@@ -149,14 +158,89 @@ where
...
-impl CastsTo<sql_types::Text> for sql_types::Int8 {}
+macro_rules! casts_impl {
+    (
+        $(
+            $($feature: literal : )? ($to: tt <- $from: tt),
+        )+
+    ) => {
+        $(
+            $(#[cfg(feature = $feature)])?
+            impl FaillibleCastsTo<sql_types::$to> for sql_types::$from {}
```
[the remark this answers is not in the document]
Sagebati: I understand,the idea is that if when writing generic code, it's great to only depend on fallible_cast a la TryFrom and From. I can revert

diesel/src/expression/cast.rs
```diff
@@ -149,14 +158,89 @@ where
...
+        )+
+    };
+}
+
+faillible_casts_impl!(
+    (Int4 <- Int8),
+    (Int4 <- Float8),
+    (Int4 <- Text),
+    (Int8 <- Text),
+    (Float4 <- Int8),
```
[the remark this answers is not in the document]
Sagebati: Good catch

diesel/src/expression/cast.rs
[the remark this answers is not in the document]
Ten0: Ok then I'd tend to remove the requirement and supertrait constraint, and we can always add back the missing implementations in a backwards-compatible manner if people ask for them?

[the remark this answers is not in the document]
weiznich: I personally think it's reasonable to have it in the current way to basically mirror what's `From` and `TryFrom` already do for rust types.

diesel/src/expression/cast.rs
```diff
@@ -149,14 +158,89 @@ where
     }
 }
 
+/// Marker trait: this SQL type (`Self`) can be casted to the target SQL type, but some values can be invalid
+pub trait FaillibleCastsTo<ST> {}
```
czotomo, line 162: Shouldn't it be `fallible`?

diesel/src/expression/cast.rs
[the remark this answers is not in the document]
Ten0: Ok

Sagebati:
Renamed Faillible to fallible, I think we are good, aren't we ?

weiznich (approved):

PR #4603, opened 2025-05-16
Setup cargo-semver-checks

weiznich:
This commit sets up a xtask for cargo-semver-checks and integrates it in our CI. It also fixes 2 instances of reported breakage. With that there are still 3 more instances reported. I believe two of them are false positives (The `Downcast` super trait and the `IntoArrayExpression` trait message). The third one is a "real" breakage, but I would nevertheless say that this will likely not affect anyone as I do not see any reasonable way to to do something with a trait object there at all.

I'm still not sure what's the correct way to supress such errors only at specific locations.

Marked as draft as the other instances need to be addressed before merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -502,3 +502,17 @@ jobs:
         run: cargo +1.84.0 minimal-versions check -p diesel_migrations --all-features
       - name: Check diesel_cli
         run: cargo +1.84.0 minimal-versions check -p diesel_cli --features "default sqlite-bundled"
+
+  Semver-Checks:
+    name: Run Cargo Semver-Checks
+    runs-on: ubuntu-latest
+    needs: [rustfmt_and_clippy]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-semver-checks
+      - name: Run cargo semver-checks
+        run: |
+          cargo xtask semver-checks
--- a/diesel/src/expression/mod.rs
+++ b/diesel/src/expression/mod.rs
@@ -1089,12 +1089,31 @@ impl<QS, ST, DB, GB, IsAggregate> ValidGrouping<GB>
     type IsAggregate = IsAggregate;
 }
 
-// Some amount of backwards-compat
-// We used to require `AsExpressionList` on the `array` function.
-// Now we require `IntoArrayExpression` instead, which means something very different.
-// However for most people just checking this bound to call `array`, this won't break.
-// Only people who directly implement `AsExpressionList` would break, but I expect that to be
-// nobody.
-#[doc(hidden)]
+/// Converts a tuple of values into a tuple of Diesel expressions.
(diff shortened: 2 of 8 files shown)


weiznich:
For reference currently the following "breaking" changes are reported:

```

Description:
A publicly-visible method or associated fn is no longer available under its prior name. It may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/inherent_method_missing.ron

Failed in:
  SerializedDatabase::new, previously in file /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.2.0/src/sqlite/connection/serialized_database.rs:16
```
We merged that one deliberately as "bug fix" as this was not supposed to be public and this was unsound.

```
--- failure trait_added_supertrait: non-sealed trait added new supertraits ---

Description:
A non-sealed trait added one or more supertraits, which breaks downstream implementations of the trait
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-tighten
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/trait_added_supertrait.ron

Failed in:
  trait diesel::connection::Instrumentation gained Downcast in file /home/runner/work/diesel/diesel/diesel/src/connection/instrumentation.rs:246
```

I believe that's a false positive for the reasons discussed in https://github.com/diesel-rs/diesel/pull/4129#discussion_r1691143494

```
--- failure trait_no_longer_dyn_compatible: trait no longer dyn compatible ---

Description:
Trait is no longer dyn compatible, which breaks `dyn Trait` usage.
        ref: https://doc.rust-lang.org/stable/reference/items/traits.html#object-safety
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/trait_no_longer_dyn_compatible.ron

Failed in:
  trait SqlOrd in file /home/runner/work/diesel/diesel/diesel/src/sql_types/ord.rs:7

     Summary semver requires new major version: 4 major and 0 minor checks failed
  trait SqlType in file /home/runner/work/diesel/diesel/diesel/src/sql_types/mod.rs:671
  trait Foldable in file /home/runner/work/diesel/diesel/diesel/src/sql_types/fold.rs:4
  trait SingleValue in file /home/runner/work/diesel/diesel/diesel/src/sql_types/mod.rs:655

```

It's correct that this is "breaking" but I would assume that nobody notice this as there as nothing usful what you could do with such  a trait object.

```
--- failure trait_now_doc_hidden: pub trait is now #[doc(hidden)] ---

Description:
A pub trait is now #[doc(hidden)], removing it from the crate's public API.
        ref: https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.41.0/src/lints/trait_now_doc_hidden.ron

```

That seems to be a bug as that item is not public on the old version either. Filled as https://github.com/obi1kenobi/cargo-semver-checks/issues/1267

PR #4610, opened 2025-05-23
Adding extended support for `Box`, `Rc`, and `Arc`

LucaCappelletti94:
This PR addresses discussion #4609

Still have to: 

* [x] Implement also:
    * [x] `impl<T, DB, const N: usize> CanInsertInSingleQuery<DB> for Rc<[T; N]>`
    * [x] `impl<T, DB, const N: usize> CanInsertInSingleQuery<DB> for Arc<[T; N]>`
    * [x] `impl<DB: Backend> Migration<DB> for Rc<dyn Migration<DB> + '_> `
    * [x] `impl<DB: Backend> Migration<DB> for Arc<dyn Migration<DB> + '_> `
    * [x] `impl<T: QueryId + ?Sized> QueryId for Rc<T>`
    * [x] `impl<T: QueryId + ?Sized> QueryId for Arc<T>`
    * [x] `impl<T: ?Sized, DB> QueryFragment<DB> for Rc<T>`
    * [x] `impl<T: ?Sized, DB> QueryFragment<DB> for Arc<T>`
    * [x] `impl<T: Insertable> Insertable for Rc<T>`
    * [x] `impl<T: Insertable> Insertable for Arc<T>`
    * [x] `impl<T: AppearsOnTable> AppearsOnTable for Rc<T>`
    * [x] `impl<T: AppearsOnTable> AppearsOnTable for Arc<T>`
    * [x] `impl<T: SelectableExpression> SelectableExpression for Rc<T>`
    * [x] `impl<T: SelectableExpression> SelectableExpression for Arc<T>`
    * [x] `impl<T: ValidGrouping> ValidGrouping for Rc<T>`
    * [x] `impl<T: ValidGrouping> ValidGrouping for Arc<T>`
    * [x] `impl<T: ?Sized, ST, DB> Queryable<ST, DB> for std::rc::Rc<T>`
    * [x] `impl<T: ?Sized, ST, DB> Queryable<ST, DB> for std::sync::Arc<T>`
    * [x] `impl<T: ?Sized, ST, DB> Queryable<ST, DB> for Box<T>`
    * [x] `impl<impl<T: ?Sized, ST, DB> ToSql<ST, DB> for std::rc::Rc<T>`
    * [x] `impl<impl<T: ?Sized, ST, DB> ToSql<ST, DB> for std::sync::Arc<T>`
    * [x] `impl<impl<T: ?Sized, ST, DB> ToSql<ST, DB> for Box<T>`
    * [x] `impl<T, ST, DB> FromSql<ST, DB> for std::rc::Rc<T>`
    * [x] `impl<T, ST, DB> FromSql<ST, DB> for std::sync::Arc<T>` 
    * [x] `impl<T, ST, DB> FromSql<ST, DB> for Box<T>` 
* [x] Extend test suite for `Rc`
* [x] Extend test suite for `Arc`

### DIFF
--- a/diesel/src/type_impls/primitives.rs
+++ b/diesel/src/type_impls/primitives.rs
@@ -8,6 +8,7 @@ use crate::sql_types::{
 };
 use alloc::borrow::Cow;
 use alloc::borrow::ToOwned;
+use alloc::boxed::Box;
 use alloc::fmt;
 use alloc::string::String;
 use alloc::vec::Vec;
@@ -270,5 +271,174 @@ where
 
     fn as_expression(self) -> Self::Expression {
         Bound::new(&**self)
+    }
+}
+
+impl<T: ?Sized, ST, DB> ToSql<ST, DB> for Box<T>
+where
+    T: ToSql<ST, DB>,
+    DB: Backend,
+    Self: fmt::Debug,
+{
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, DB>) -> serialize::Result {
+        ToSql::<ST, DB>::to_sql(&**self, out)
+    }
+}
+
+impl<T, ST, DB> FromSql<ST, DB> for Box<T>
+where
+    DB: Backend,
+    T: FromSql<ST, DB>,
+{
+    fn from_sql(bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        T::from_sql(bytes).map(Box::from)
+    }
+}
+
+impl<T: ?Sized, ST, DB> Queryable<ST, DB> for Box<T>
+where
+    ST: SingleValue,
+    DB: Backend,
+    Self: FromSql<ST, DB>,
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> deserialize::Result<Self> {
+        Ok(row)
+    }
+}
+
+impl<T: ?Sized, ST, DB> ToSql<ST, DB> for alloc::rc::Rc<T>
+where
+    T: ToSql<ST, DB>,
+    DB: Backend,
+    Self: fmt::Debug,
+{
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, DB>) -> serialize::Result {
+        ToSql::<ST, DB>::to_sql(&**self, out)
+    }
+}
+
+impl<T, ST, DB> FromSql<ST, DB> for alloc::rc::Rc<T>
+where
+    DB: Backend,
+    T: FromSql<ST, DB>,
+{
+    fn from_sql(bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        T::from_sql(bytes).map(alloc::rc::Rc::from)
+    }
+}
+
+impl<T: ?Sized, ST, DB> Queryable<ST, DB> for alloc::rc::Rc<T>
+where
+    ST: SingleValue,
+    DB: Backend,
+    Self: FromSql<ST, DB>,
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> deserialize::Result<Self> {
+        Ok(row)
+    }
+}
+
+impl<T: ?Sized, ST, DB> ToSql<ST, DB> for alloc::sync::Arc<T>
+where
+    T: ToSql<ST, DB>,
+    DB: Backend,
+    Self: fmt::Debug,
+{
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, DB>) -> serialize::Result {
+        ToSql::<ST, DB>::to_sql(&**self, out)
+    }
+}
+
+impl<T, ST, DB> FromSql<ST, DB> for alloc::sync::Arc<T>
+where
+    DB: Backend,
+    T: FromSql<ST, DB>,
+{
+    fn from_sql(bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        T::from_sql(bytes).map(alloc::sync::Arc::from)
+    }
+}
+
+impl<T: ?Sized, ST, DB> Queryable<ST, DB> for alloc::sync::Arc<T>
+where
+    ST: SingleValue,
+    DB: Backend,
+    Self: FromSql<ST, DB>,
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> deserialize::Result<Self> {
+        Ok(row)
+    }
+}
+
+// `FromSql` for the unsized-inner-type cases (`Box<str>`, `Rc<str>`, `Arc<str>`,
+// and the `[u8]` equivalents). The generic `FromSql for Box<T> where T: FromSql`
+// blanket above can't cover these because `str`/`[u8]` are `!Sized` and cannot
+// implement `FromSql` themselves. Instead we go through `FromSqlRef` for `&str` /
+// `&[u8]` and then build the smart pointer with `Box::from` / `Rc::from` / `Arc::from`,
+// which all accept `&str` or `&[u8]` and allocate a sized backing buffer.
+
+impl<ST, DB> FromSql<ST, DB> for Box<str>
+where
+    DB: Backend,
+    for<'a> &'a str: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&str as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(Box::from)
+    }
+}
+
+impl<ST, DB> FromSql<ST, DB> for alloc::rc::Rc<str>
+where
+    DB: Backend,
+    for<'a> &'a str: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&str as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(alloc::rc::Rc::from)
+    }
+}
+
+impl<ST, DB> FromSql<ST, DB> for alloc::sync::Arc<str>
+where
+    DB: Backend,
+    for<'a> &'a str: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&str as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(alloc::sync::Arc::from)
+    }
+}
+
+impl<ST, DB> FromSql<ST, DB> for Box<[u8]>
+where
+    DB: Backend,
+    for<'a> &'a [u8]: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&[u8] as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(Box::from)
+    }
+}
+
+impl<ST, DB> FromSql<ST, DB> for alloc::rc::Rc<[u8]>
+where
+    DB: Backend,
+    for<'a> &'a [u8]: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&[u8] as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(alloc::rc::Rc::from)
+    }
+}
+
+impl<ST, DB> FromSql<ST, DB> for alloc::sync::Arc<[u8]>
+where
+    DB: Backend,
+    for<'a> &'a [u8]: FromSqlRef<'a, ST, DB>,
+{
+    fn from_sql(mut bytes: DB::RawValue<'_>) -> deserialize::Result<Self> {
+        <&[u8] as FromSqlRef<'_, ST, DB>>::from_sql(&mut bytes).map(alloc::sync::Arc::from)
     }
 }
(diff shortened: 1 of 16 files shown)


weiznich:
Thanks for opening this PR. I think we likely want to cover most of [these](https://github.com/search?q=repo%3Adiesel-rs%2Fdiesel+%22for+Box%3C%22&type=code) impls as well. (All but that one for `Instrumentation`)

LucaCappelletti94:
@weiznich I have extended the list of traits to be implemented, please do let me know whether anything is missing.

weiznich:
From the search linked above it's missing the following impls:

* `Insertable for Arc/Rc`
* `Migration` for Arc/Rc`
* `AppearsOnTable for Arc/Rc`
* `SelectableExpression for Arc/Rc`
* `ValidGrouping for Arc/Rc`

Also it lists the instrumentation impl which will not be helpful as the only method there takes a `&mut self` argument which you won't have for Arc/Rc for obvious reasons.

LucaCappelletti94:
> From the search linked above it's missing the following impls:
> 
>     * `Insertable for Arc/Rc`
> 
>     * `Migration for Arc/Rc`
> 
>     * `AppearsOnTable for Arc/Rc`
> 
>     * `SelectableExpression for Arc/Rc`
> 
>     * `ValidGrouping for Arc/Rc`
> 
> 
> Also it lists the instrumentation impl which will not be helpful as the only method there takes a `&mut self` argument which you won't have for Arc/Rc for obvious reasons.

I have updated the PR tasks to include also these traits and removed the `Instrumentation`

LucaCappelletti94:
I am afraid I will require some more instructions regarding how to proceed. As you can see from the [CI run](https://github.com/diesel-rs/diesel/actions/runs/15212136723/job/42788661444?pr=4610), trying to use `std::rc::Rc` or `std::sync::Arc` in a struct leads to the following errors:

```
 the trait `diesel::Expression` is not implemented for `std::string::String`
```

and

```
the trait `diesel::AppearsOnTable<schema::comments::table>` is not implemented for `std::string::String`
```

Which make sense as indeed, those traits are needed for the current trait implementations and those traits are not implemented for String. What should be preferable to do?

weiznich:
> Just to better understand, https://github.com/diesel-rs/diesel/discussions/4609 I was asking exactly regarding the use of Rc in such structs, but maybe I had explained myself poorly. Do you believe it should be viable to do, in theory, or are there any blocking reasons?

That was mostly me skipping over details assuming I know what you want there. 

That written, it's actually good to know that we cannot have both, which means from my point of view this is for now kind of a dead end as I consider both use-cases (using `Rc<dyn BoxableExpression>` and `Rc<String` as field in your struct) as something that would be desirable to support. If we cannot easily support both it's kind of hard to say which one should be supported without making it essentially impossible to support the other one as well. So as long as we do not have a solution for this I would hold of implementing anything to not block any future extensions. I personally do currently not have the time to investigate this in depth, given the many other things I need to handle (like answering all your questions).

> That being said, there is a [blanked implementation](https://github.com/diesel-rs/diesel/blob/db7edfe9db4313dd3f400dcdd0ab2180287ac2eb/diesel/src/expression/mod.rs#L232-L242) adding AsExpression<ST> to all types which already implement Expression<ST> for that ST, and Rc<T> would therefore implement AsExpression if it can implement Expression, which depends on T having implemented Expression - why does not T=String (in this case) implement the Expression trait?

Note that this only implements `AsExpression` if the underlying type is an actual `Expression` `String` or any other rust value type is never considered a SQL expression by diesel, so that wild card impl is not helpful here to support both variants. What might be viable is to generate specific `Rc` and `Arc` variants for the `AsExpression` impls via the derive, but again that would drastically increase the number of impls generated by diesel, which in turn will certainly have an impact on compile times and error messages. It might be worth to quantify that impact, maybe it's not as bad as assumed but without hard numbers I wouldn't like blindly  to go that way.

LucaCappelletti94:
Thank you for your reply - regarding extending the `AsExpression` derive, I can try to add the `Rc` and `Arc` derive impls and report how much the compile time increase for both diesel and the practical case of my project.

I am not sure I understood whether I should try to add support for `impl<ST> AsExpression<ST> for Rc<String> where String: AsExpression<ST>` (and other std/core types), or whether such an implementation would preclude potential future impls of `Rc<dyn BoxableExpression>` and as such should not be added for the time being. I suppose you mentioned extending the `AsExpression` derive exactly because I could then wrap the `String` type into a custom type?

Thank you again for the patience in answering all of these questions, hopefully over time I will learn better the internals of `diesel` and be better able to contribute to it.

diesel_derives/src/lib.rs
```diff
@@ -139,6 +139,10 @@ pub fn derive_as_changeset(input: TokenStream) -> TokenStream {
 /// - `impl AsExpression<Nullable<SqlType>> for YourType`
 /// - `impl AsExpression<SqlType> for &'a YourType`
 /// - `impl AsExpression<Nullable<SqlType>> for &'a YourType`
+/// - `impl AsExpression<SqlType> for std::rc::Rc<YourType>`
+/// - `impl AsExpression<SqlType> for &'a std::rc::Rc<YourType>`
+/// - `impl AsExpression<SqlType> for std::sync::Arc<YourType>`
+/// - `impl AsExpression<SqlType> for &'a std::sync::Arc<YourType>`
```
weiznich, line 145: That should be removed if it's not generated by this derive anymore

diesel/src/type_impls/primitives.rs
```diff
@@ -271,3 +271,119 @@ where
         Bound::new(&**self)
     }
 }
+
+impl<T: ?Sized, ST, DB> ToSql<ST, DB> for std::rc::Rc<T>
```
weiznich, line 275: I would like to add `Box<T>` variants for these impls as well as `Box` is essentially just another smart pointer.

diesel_tests/tests/schema/mod.rs
```diff
@@ -90,6 +90,44 @@ impl Comment {
...
+    Identifiable,
+    Associations,
+    Selectable,
+)]
+#[diesel(belongs_to(Post), table_name = comments)]
+pub struct CommentArc {
+    pub id: i32,
+    pub post_id: i32,
+    pub text: std::sync::Arc<String>,
+}
```
weiznich, line 129: Could we have a simple test here that inserts + loads both variants?

diesel_tests/tests/schema/mod.rs
```diff
@@ -90,6 +90,44 @@ impl Comment {
...
+    Insertable,
+    Identifiable,
+    Associations,
+    Selectable,
+)]
+#[diesel(belongs_to(Post), table_name = comments)]
+pub struct CommentRc {
+    pub id: i32,
+    pub post_id: i32,
+    pub text: std::rc::Rc<str>,
```
weiznich, line 109: I would expect that this causes issues for loading values as we don't have a `ToSql<DB, Text> for str` impl yet. That said it seems reasonable to support this as well, but I would like to keep that off this PR to not expand it's scope any further. So just go with `Rc<String>` for now.

weiznich (commented):
Thanks for updating this PR.

As these impls do not seem to conflict with anything it should be fine to go that way.

There are still a few things I noticed about the changes. 

Additionally I would like to see the following things added to the PR:

- A changelog entry stating that `Rc`+`Arc` are now supported types for serialization + deserialization + `BoxableExpression`
- A testcase that actually verifies inserting + loading data works
- A test case that verifies that `Arc<dyn BoxableExpression>` (and the `Rc` equivalent works)

> That being said, it turns out that methods like get_result and first require Send and since Rc does not implement Send, I get the error  `Rc<str>` cannot be sent between threads safely. At least, these are required when using diesel-async, I am not sure whether the same would apply also for diesel in its sync version.

That's `diesel-async` specific. It should just work with normal diesel, so it's reasonable to keep it here.

> Do you think such Send trait requirements could be removed in any way?

Not without regressing use-cases that require the `Send` bound to be there as there is currently no way to abstract over the fact whether or not a type implements `Send`.

diesel/src/type_impls/primitives.rs
LucaCappelletti94: For all traits described in the first post of the PR, or only for specifically `ToSql`?

LucaCappelletti94: Yeah as expected, I can use `Rc` and `Arc` in the structs but the same does not work for `Box`. Again, it is not clear to me why the blanked covers the `Box` case but does not cover the `Arc` and `Rc`, I believe it has something to do with Rust's auto-dereferencing.

LucaCappelletti94:
I am trying to test and fix the case for `Rc<dyn BoxableExpression>`, but unfortunately this case does not seem to implement Expression: `Rc<dyn BoxableExpression<table, Pg, SqlType = _>>`. Same thing happens for `Arc`.

```rust
#[diesel_test_helper::test]
fn can_rc_query_with_boxable_expression() {
    let connection = &mut connection_with_sean_and_tess_in_users_table();

    let expr: Rc<dyn BoxableExpression<users::table, _, SqlType = _>> =
        Rc::new(users::name.eq("Sean")) as _;

    let data = users::table.into_boxed().filter(expr).load(connection);
    let expected = vec![find_user_by_name("Sean", connection)];
    assert_eq!(Ok(expected), data);

    let expr: Rc<dyn BoxableExpression<users::table, _, SqlType = _>> =
        Rc::new(users::name.eq("Sean")) as _;

    let data = users::table.filter(expr).into_boxed().load(connection);
    let expected = vec![find_user_by_name("Sean", connection)];
    assert_eq!(Ok(expected), data);
}
```

with the full error being:

```rust
error[E0277]: the trait bound `Rc<dyn BoxableExpression<table, Pg, SqlType = _>>: AppearsOnTable<...>` is not satisfied
   --> diesel_tests/tests/boxed_queries.rs:201:42
    |
201 |     let data = users::table.into_boxed().filter(expr).load(connection);
    |                                          ^^^^^^ the trait `diesel::Expression` is not implemented for `Rc<dyn diesel::BoxableExpression<schema::users::table, Pg, SqlType = _>>`
    |
    = help: the following other types implement trait `diesel::Expression`:
              &T
              (T0, T1)
              (T0, T1, T2)
              (T0, T1, T2, T3)
              (T0, T1, T2, T3, T4)
              (T0, T1, T2, T3, T4, T5)
              (T0, T1, T2, T3, T4, T5, T6)
              (T0, T1, T2, T3, T4, T5, T6, T7)
            and 452 others
    = note: required for `Rc<dyn diesel::BoxableExpression<schema::users::table, Pg, SqlType = _>>` to implement `diesel::AppearsOnTable<schema::users::table>`
    = note: required for `BoxedSelectStatement<'_, (Integer, Text, Nullable<Text>), ..., ...>` to implement `FilterDsl<Rc<dyn diesel::BoxableExpression<schema::users::table, Pg, SqlType = _>>>`
    = note: the full name for the type has been written to '/home/luca/github/diesel/target/debug/deps/integration_tests-14ee3c308f18e762.long-type-688333457155762760.txt'
```

weiznich:
So I had a look at the relevant code and it seems like that this PR is missing that impl for `Rc` and `Arc`:

https://github.com/diesel-rs/diesel/blob/15216dcb1d36fd55bc583c4f7a4704bc53983b7b/diesel/src/expression/mod.rs#L166-L168

This impl enables the `Box<dyn BoxableExpression>` pattern, but prevents using the wild card impl. That's that what I wrote initially with it's required to put specific impls in the derive generated output to prevent needing to decide wheter we want to support one (loading/storing values wrapped in a `Arc`/`Rc`) or the other feature (using `Arc<dyn BoxableExpression>` for dynamic clonable query fragments.

The summary is that you need to:

* Add the equivalents of the linked impl for `Arc`/`Rc`
* Remove the two blanked impls
* Add back the generated specific impls for `Arc`/`Rc` (and ideally `Box`) in the `AsExpression` macro + measure the compile time impact of the additional impls

LucaCappelletti94:
Ran the benchmark between this branch and current main.

This time around, I generated a synthetic but plausible schema with 50 tables, each with randomly 6-12 columns, and each with random types from the set of types supported by sqlite & postgres.

Each profile (`dev` and `release`) was built five times from a cold cache (`cargo clean` between runs) and the reported value is the median. Timing comes from `/usr/bin/time -f "%e %U %S %M"` (wall, user CPU, system CPU, peak RSS). The host is an AMD Threadripper PRO 5975WX (64 threads, 1 TB RAM).

| Profile | Branch | Wall | User | Max RSS |
| --- | --- | ---: | ---: | ---: |
| dev | `upstream/main` | 37.67 s | 42.82 s | 1 230 MiB |
| | `rc-arc` | 38.04 s | 43.38 s | 1 238 MiB |
| release | `upstream/main` | 38.70 s | 50.73 s | 1 358 MiB |
| | `rc-arc` | 38.32 s | 49.99 s | 1 361 MiB |

Wall-clock values across all 5 runs per cell (s, sorted), with transient system spike during one run on each branch marked with †. I am unsure what that is.

| Profile | `upstream/main` | `rc-arc` |
| --- | --- | --- |
| dev | 36.71 / 37.00 / **37.67** / 37.88 / 39.02 | 35.89 / 38.03 / **38.04** / 38.45 / 39.16 |
| release | 37.50 / 37.88 / **38.70** / 39.01 / 60.95† | 36.90 / 37.39 / **38.32** / 38.83 / 60.84† |

All four median deltas land between −1.5% and +1.3%, so I would say it does not cause a meaningful change in this setup.

weiznich (approved):
This looks good now, thanks for the update

Would it be possible to squash the commits + add a change log entry for this feature before merging.

I hopefully get around to review at least some of your other PR's in the next weeks as well if nothing important shows up again (like more security issues :()

LucaCappelletti94:
Done!

weiznich (approved):

PR #4616, opened 2025-05-25
implement LoadQuery for batch insert statements for sqlite connection

bwqr:
This is a draft PR requesting for comments. It allows running insert statements with returning clause for sqlite connection. There are a few remaining tasks, which needs a little bit guidance, to consider this PR as ready:

- [x] Find an appropriate Iterator implementation to return from `internal_load` (instead of currently used boxed iteration).
- [x] Implement `LoadQuery` for `Yes` case which requires running multiple insert statements. I can implement this if `No` case looks good.
- [x] Consider adding tests or examples.
- [x] ~Consider using a generic connection with backend `Sqlite` (similar to `ExecuteDsl` implementation). I tried to do this but it caused conflicting implementation error.~ It looks like trait bounds resolution is being improved with newer version of Rust compiler. In the future, it might be possible to use a generic connection bounded by `Sqlite` backend. For now, we can continue using directly `SqliteConnection`.
- [x] Bump MSRV to 1.84 because of compilation errors caused by `conflicting implementation` error. Check [this](https://github.com/diesel-rs/diesel/pull/4616#issuecomment-2915858331) and [this](https://github.com/diesel-rs/diesel/pull/4616#issuecomment-2915895403) messages for more information

With this PR, following example should compile successfully:

```rust
use diesel::prelude::{ExpressionMethods, Queryable};
use diesel::{Connection, RunQueryDsl, SqliteConnection};

diesel::table! {
    apps (id) {
        id -> Int4,
        #[max_length = 8]
        name -> Varchar,
        #[max_length = 8]
        state -> Varchar,
    }
}

#[derive(Queryable)]
struct App {
    id: i32,
    name: String,
    state: String,
}

fn main() {
    let mut conn = SqliteConnection::establish(":memory:").unwrap();

    let query = diesel::insert_into(apps::table).values(vec![
        (apps::name.eq("Good"), apps::state.eq("Running")),
        (apps::name.eq("Bad"), apps::state.eq("Failed")),
    ]);

    query.get_results::<App>(&mut conn);
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ in a way that makes the pools suitable for use in parallel tests.
 * Add support for the `CAST` operator
 * Support `[print_schema] allow_tables_to_appear_in_same_query_config = "none"` to generate no `allow_tables_to_appear_in_same_query!` calls. (Default: `"all_tables"`.). ([#4333](https://github.com/diesel-rs/diesel/issues/4333))
 * Add `[print_schema] pg_domains_as_custom_types` parameter to generate custom types for [PostgreSQL domains](https://www.postgresql.org/docs/current/domains.html) that matches any of the regexes in the given list. (Default: `[]`.) This option allows an application to selectively give special meaning for the serialization/deserialization of these types, avoiding the default behavior of treating the domain as the underlying type. ([#4592](https://github.com/diesel-rs/diesel/discussions/4592))
+* Add support for batch insert and upsert statements with returning for SQLite
 
 ### Fixed 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ in a way that makes the pools suitable for use in parallel tests.
 ### Changed
 
 * Use distinct `DIESEL_LOG` logging filter env variable instead of the default `RUST_LOG` one (#4575)
-* The minimal supported Rust version is now 1.82.0
+* The minimal supported Rust version is now 1.84.0
 
 ## [2.2.2] 2024-07-19
 
--- a/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
+++ b/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
@@ -292,14 +292,17 @@ where
     }
 }
 
-impl<V, T, QId, Op, const STATIC_QUERY_ID: bool> RunQueryDsl<SqliteConnection>
+impl<V, T, QId, Op, O, const STATIC_QUERY_ID: bool> RunQueryDsl<SqliteConnection>
     for (
-        No,
-        InsertStatement<T, BatchInsert<V, T, QId, STATIC_QUERY_ID>, Op>,
+        O,
+        InsertStatement<T, BatchInsert<Vec<ValuesClause<V, T>>, T, QId, STATIC_QUERY_ID>, Op>,
     )
 where
     T: QuerySource,
-    InsertStatement<T, BatchInsert<V, T, QId, STATIC_QUERY_ID>, Op>: RunQueryDsl<SqliteConnection>,
+    V: ContainsDefaultableValue<Out = O>,
+    O: Default,
+    InsertStatement<T, BatchInsert<Vec<ValuesClause<V, T>>, T, QId, STATIC_QUERY_ID>, Op>:
+        RunQueryDsl<SqliteConnection>,
 {
 }
 
@@ -450,6 +453,7 @@ where
         + 'static,
     // Connection bounds
     SqliteConnection: crate::connection::LoadConnection<B>,
+    Self: RunQueryDsl<SqliteConnection>,
 {
     type RowIter<'conn> = Box<dyn Iterator<Item = QueryResult<U>> + 'conn>;
 
--- a/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
+++ b/diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
@@ -273,50 +273,6 @@ where
     }
 }
 
-impl<V, T, QId, C, Op, Target, ConflictOpt, const STATIC_QUERY_ID: bool> ExecuteDsl<C, Sqlite>
-    for (
-        Yes,
-        InsertStatement<
-            T,
-            OnConflictValues<
-                BatchInsert<Vec<ValuesClause<V, T>>, T, QId, STATIC_QUERY_ID>,
-                Target,
-                ConflictOpt,
-            >,
-            Op,
-        >,
-    )
-where
-    C: Connection<Backend = Sqlite>,
-    T: Table + Copy + QueryId + 'static,
-    T::FromClause: QueryFragment<Sqlite>,
-    Op: Copy + QueryId + QueryFragment<Sqlite>,
-    V: InsertValues<Sqlite, T> + CanInsertInSingleQuery<Sqlite> + QueryId,
-{
-    fn execute((Yes, query): Self, conn: &mut C) -> QueryResult<usize> {
-        conn.transaction(|conn| {
-            let mut result = 0;
-            for record in &query.records.values.values {
-                let stmt = InsertStatement {
-                    operator: query.operator,
-                    target: query.target,
-                    records: OnConflictValues {
-                        values: record,
-                        target: query.records.target,
-                        action: query.records.action,
-                        where_clause: query.records.where_clause,
-                    },
-                    returning: query.returning,
-                    into_clause: query.into_clause,
-                };
-
-                result += stmt.execute(conn)?;
-            }
-            Ok(result)
-        })
-    }
-}
-
 impl<'query, V, T, QId, Op, O, U, B, const STATIC_QUERY_ID: bool>
     LoadQuery<'query, SqliteConnection, U, B>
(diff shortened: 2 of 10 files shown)


bwqr:
Thank you for your feedback.

> There seems to be still a failing CI job here: https://github.com/diesel-rs/diesel/actions/runs/15237082999/job/42852408104?pr=4616. I did not look into the details yet, but that's something that needs to be addressed.

For some reason, the same code causes `conflicting implementation for 'InsertStatement<_, BatchInsert<Vec<insert_statement::ValuesClause<_, _>>, _, _, _>, _>'` when it is compiled with rust 1.82 compiler. This error confused me a little bit since other checks pass successfully. I will investigate this by trying to compile the code locally with 1.82.

> I would suggest to return a "private" wrapper type that implements Iterator, that would enable us to change the details later on as users cannot do anything other than iterating with that type. Essentially what we do [here](https://github.com/diesel-rs/diesel/blob/master/diesel/src/query_dsl/load_dsl.rs#L53), which puts the type in a private module so that users cannot name it in their code.

I was hoping to use an already defined iterator instead of defining a new one, such as the `LoadIter` you have linked. However, since `LoadIter` is private to its module, I am not able to use it inside the `insert_with_default_for_sqlite.rs`. Should I define a new iterator similar to `LoadIter` or should we export `LoadIter` with `pub(crate)` to prevent duplication?

I will try to write some tests. Until completing all the tasks, I think converting this PR to draft might be the best.

bwqr:
> I'm not sure if it's related to 1.82. It might also be due to building with --features "sqlite mysql postgres extras", which is not done by the other CI builds. If its related to the rust version we can just bump the MSRV to something that accepts this, this would be fine.

I just checked this issue by compiling this PR with a few rust releases with `--features "sqlite returning_clauses_for_sqlite_3_35"`. Here are the results:
* 1.81: Did not compile due to `extern block cannot be declared unsafe` used in `libsqlite3-sys` crate. I guess this was stabilized in 1.82.
* 1.82: Causes conflicting implementation ❌
* 1.83: Causes conflicting implementation ❌
* 1.84: Compiles fine ✅
* 1.87: Compiles fine ✅

> If LoadIter is fitting it would be good to use it. It's no problem to make it pub(crate), public. We just don't want to have it in our public API.

Nice, I will check whether `LoadIter` is useful for this implementation.

weiznich:
I did a quick search which rustc change did introduce that changed behavior just to make sure that this is not by accident and that pointed to https://github.com/rust-lang/rust/pull/130654, which seems to be reasonable. So I suggest to just bump the minimal supported rust version to 1.84. That needs to be done [here](https://github.com/diesel-rs/diesel/blob/master/Cargo.toml#L39), [here](https://github.com/diesel-rs/diesel/blob/master/.github/workflows/ci.yml#L474-L501) and [here](https://github.com/diesel-rs/diesel/blob/master/CHANGELOG.md?plain=1#L46).

bwqr:
I believe this PR is ready for the review :). I would like to thank you for all your guidance and feedback.

CHANGELOG.md
weiznich, line 1: We might also want to mention the new feature in the Changelog :wink:

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
```diff
@@ -273,6 +273,192 @@ where
...
+        Yes,
+        InsertStatement<T, BatchInsert<Vec<ValuesClause<V, T>>, T, QId, STATIC_QUERY_ID>, Op>,
+    )
+where
+    T: Table + Copy + QueryId + 'static,
+    Op: Copy + QueryId + QueryFragment<Sqlite>,
+    InsertStatement<T, ValuesClause<V, T>, Op>: LoadQuery<'query, SqliteConnection, U, B>,
+    Self: RunQueryDsl<SqliteConnection>,
+{
+    type RowIter<'conn> = std::vec::IntoIter<QueryResult<U>>;
```
weiznich, line 378: Using a nameable type will prevent us from changing this without breaking change. On the other hand I really do not see any other reasonable implementation here so it's likely fine.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
```diff
@@ -273,6 +273,192 @@ where
...
+            for record in query.records.values {
+                let stmt =
+                    InsertStatement::new(query.target, record, query.operator, query.returning);
+
+                let result = stmt
+                    .internal_load(conn)?
+                    .next()
+                    .ok_or(crate::result::Error::NotFound)?;
+
+                results.push(result);
```
weiznich, line 395: We also want to unwrap the inner result to stop the transaction from being comited in the error test. We likely want to have a test case for this (Just insert a reasonable item + a duplicated item and make sure that none are in the database)

```suggestion
                results.push(Ok(result?));
```

(If we would like to be even more correct: We should roll back the transaction for anything but deserialization errors to perfectly mirror the batch variant)

There is a test for batch inserts without returning here: https://github.com/diesel-rs/diesel/blob/efcb1e7e6cc921dbb92edd5fb5a6829c9625da51/diesel_tests/tests/insert.rs#L821 You likely can just copy and adjust that test for insert + returning and upserts. 

Otherwise it sounds like a good idea to quickly go over the tests cfg out for sqlite in that file and check which ones can now be enabled.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
```diff
@@ -273,6 +273,192 @@ where
...
+                    returning: query.returning,
+                    into_clause: query.into_clause,
+                };
+
+                let result = stmt
+                    .internal_load(conn)?
+                    .next()
+                    .ok_or(crate::result::Error::NotFound)?;
+
+                results.push(result);
```
weiznich, line 454: Same as above

weiznich (commented):
Thanks for the update. This looks already really good in terms of implementation. 
I added another bunch of mostly minor comments about some details that need to be adjusted. As soon as those are fixed, this should be fine to be merged.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
bwqr: If I understand your concern correctly, this can cause downstream user to put a bound restricting the type implementing `LoadQuery<RowIter<'conn> = IntoIter<QueryResult<U>>`. As an example, we do not want this to happen right? `impl MyTrait for T where T: LoadQuery<..., RowIter<'_> = IntoIter<...>>`. If that is the case, I would definitely like to change this to an appropriate one.

As a solution to this problem, should we create a thin wrapper struct that wraps the `IntoIter` struct and delagates all `Iterator` trait fns to its inner?

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
```diff
@@ -273,6 +273,192 @@ where
...
+            }
+
+            Ok(results.into_iter())
+        })
+    }
+}
+
+impl<'query, V, T, QId, Op, U, B, Target, ConflictOpt, const STATIC_QUERY_ID: bool>
```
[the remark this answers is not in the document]
bwqr: I think this is not necessary anymore. I just pushed a set of commits where I refactored trait bounds and they do not use `ReturningClause` explicitly. Even more, when users do not enable `returning_clauses_for_sqlite_3_35` feature, they now get a compile time error related to returning clause not supported by backend when they try to use `get_results` (though there are other errors related to other trait bounds). Here is an example error message
```
the trait `QueryFragment<Sqlite, DoesNotSupportReturningClause>` is not implemented for `ReturningClause<...>`
```

But we can put them behind `returning_clauses_for_sqlite_3_35` feature if you think it is necessary still.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
```diff
@@ -400,6 +586,117 @@ where
...
+                target: query.records.target,
+                action: query.records.action,
+                where_clause: query.records.where_clause,
+            },
+            returning: crate::query_builder::returning_clause::ReturningClause(T::all_columns()),
+            into_clause: query.into_clause,
+        };
+
+        Ok(crate::query_dsl::load_dsl::LoadIter {
+            cursor: <_ as crate::connection::LoadConnection<B>>::load(conn, query)?,
```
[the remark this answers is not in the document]
bwqr: Indeed :smiley: , but I found another way to run the query, which eliminated using `LoadIter` explicitly. I reverted changes related to making LoadIter `pub(crate)`.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
weiznich: My concern is not only that users might be able to use it as you describe, but also that they can write something like
```rust
let r: IntoIter<_> = diesel::insert_into(some_table).values(values).internal_load(conn)?;
```

On the other hand as written before: I do not see any other way to implement this reasonably, so we likely never want to change it, so it shouldn't be a problem. I don't think we need to change it, it's just me thinking about that loudly.

diesel/src/query_builder/insert_statement/insert_with_default_for_sqlite.rs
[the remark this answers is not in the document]
weiznich: :+1: That's sounds good to me

PR #4617, opened 2025-05-26
Implement ToSql for Box<str>

guissalustiano:
I'm using Box<str> to use less space on heap and would like to write it direct to the database with diesel


weiznich:
Yes it would be great to have it for the other `Text` like types as well and not only for `Citext`. Also some tests are necessary before accepting this.

weiznich:
#4610 is slightly different than this PR. With what's currently implemented in that other PR it should be possible to load/store values from any `Arc<T>`/`Rc<T>`/`Box<T>` field where the `T` implements `FromSql` or `ToSql`. 

At least for `T == str` and `FromSql` that's not the case yet, as you cannot create a unboxed (or arc/rc) `str` value out of nowhere. So in the end both PR's are somewhat required to address both slightly different uses-cases. That written it totally makes sense to support `Rc<str>` and `Arc<str>` here as well. 

So yes, literally the change in this PR is made obsolete by the other PR but there are similar features that might still be useful to add.

LucaCappelletti94:
I believe the PR can now be closed

LucaCappelletti94:
I am closing this PR as I am fairly confident it is wholly superseeded by #4610

PR #4650, opened 2025-06-16
Added support for multiple having calls (fixes #4649)

### LINKED ISSUE #4649: `having` method overwrites existing clause
### Versions

- **Diesel: master**

## Problem Description

The `having` dsl function appears to overwrite the existing clause, rather than `and`ing, like `filter` does – both on boxed and non-boxed. Perhaps this was intentional, but the docs imply that it is additive: _"Adds to the `HAVING` clause of a query."_, and it's a little confusing as I'd assume it to behave in a similar manner to `filter` does.

See impls:

https://github.com/diesel-rs/diesel/blame/7629365ffbad24019ab3d126433d4c26c7c13c79/diesel/src/query_builder/select_statement/boxed.rs#L520

https://github.com/diesel-rs/diesel/blob/7629365ffbad24019ab3d126433d4c26c7c13c79/diesel/src/query_builder/select_statement/dsl_impls.rs#L686

Dynamically adding multiple `having` clauses to a boxed query.

`having` should `and` with any existing `having` clauses.

Any existing clauses are discarded and overwritten.

### Steps to reproduce

```rust
let data = users::table.inner_join(posts::table)
    .group_by(users::id)
    .having(count(posts::id).gt(1))
    .having(count(posts::title).gt(1))
    .select((users::name, count(posts::id)))
    .load::<(String, i64)>(connection)?;
```

In this case, the id clause will be overwritten, leaving only the title clause.

oeed:
This fixes #4649 by essentially copying the behaviour of where clauses to having clauses. Currently I've just implemented the `and` part, but if this approach is okay I'll add `or_having` too.

The main things to note:

1. This may be somewhat breaking, although only breaking if they were misusing `having` originally. If they have multiple having clauses it'd change the type if not boxed, and behaviour will change to `and`ing them rather than only using the last one. As this use was essentially incorrect, I felt it was acceptable to make this change, rather than the alternate of changing the docs to reflect the non-additive nature.

2. I've reused there `WhereAnd` traits (rather than `HavingAnd`), and it doesn't seems necessary to duplicate what is essentially the same trait and as it's already fairly widely used. Let me know if you disagree.

Will sort clippy issues once `or_having` is done.

### DIFF
--- a/diesel/src/query_builder/having_clause.rs
+++ b/diesel/src/query_builder/having_clause.rs
@@ -1,1 +1,186 @@
-simple_clause!(NoHavingClause, HavingClause, " HAVING ");
+use super::from_clause::AsQuerySource;
+use super::where_clause::{WhereAnd, WhereOr};
+use super::*;
+use crate::backend::DieselReserveSpecialization;
+use crate::expression::grouped::Grouped;
+use crate::expression::operators::{And, Or};
+use crate::expression::*;
+use crate::sql_types::BoolOrNullableBool;
+
+/// Represents that a query has no `HAVING` clause.
+#[derive(Debug, Clone, Copy, QueryId)]
+pub struct NoHavingClause;
+
+impl<DB> QueryFragment<DB> for NoHavingClause
+where
+    DB: Backend + DieselReserveSpecialization,
+{
+    fn walk_ast<'b>(&'b self, _: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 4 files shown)


weiznich:
Thanks for putting the work in this PR. We cannot merge this as this is a breaking change.

In addition as discussed in https://github.com/diesel-rs/diesel/issues/4649#issuecomment-2975151580 I would even consider the current behavior as "expected" behavior, given that there are also good arguments to follow the standard set by other `QueryDsl` methods.

PR #4655, opened 2025-06-20
Replace trybuild with ui_test

weiznich:
ui_test is one of the crates used by rustc for compile_tests. I choose to replace trybuild with ui_test as the former does not perform const evaluation for the tests, which results in certain patterns not beeing testable. It also does not seem to be planned to fix that in trybuild.  See for details https://github.com/dtolnay/trybuild/issues/225

Another thing that I see as advantage of ui_test is that it requires us to annotate which lines emitt errors, which makes it harder to accidently change something from not compiling to compiling by accident.

This is a large diff due to that the output of all tests change slightly. I tried my best to adjust the output via filters us much as possible to match the trybuild output, but there are still some differences. If it turns out to be problematic in the future we can continue to tweak the output.

Finally I also added some docs for the compile tests to the CONTRIBUTING manual.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -158,3 +158,9 @@ Check out [their README](https://github.com/rust-lang/rustfmt) for details.
 `Conn`: Connection
 
 Generally, we prefer to give our types meaningful names. `Lhs` and `Rhs` vs `T` and `U` for a binary expression, for example.
+
+### Compile Tests
+
+Diesel has an extensive suite of compile tests in the `diesel_compile_tests` crate. These test work by having a small test program for each test case and then verifying that the compilation of those tests fail with a specific error message. For that we use the [`ui_test`](https://docs.rs/ui_test/latest/ui_test/) also used by rustc.  
+Running these tests can done by simply running `cargo test` in the `diesel_compile_tests` directory. Adding new tests simply requires adding a new file to `diesel_compile_tests/tests/fail/` containing the source code you want to test.
+You can run these tests with the environment variable `BLESS` set to `1` to update the expected stderr output. You also need to update the inline error annotations in the source code to match on the error message. See the documentation of `ui_test` for how to do that. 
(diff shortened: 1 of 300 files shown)


pksunkara (approved):

pksunkara:
I have approved, but I have the following concerns:

> it requires us to annotate which lines emit errors, which makes it harder to accidentally change something from not compiling to compiling by accident.

If you look at the `stderr` files, the errors have a line number specified anyway. (Note that `21` below)

```
error[E0277]: Cannot select `posts::columns::id` from `users::table`
  --> tests/fail/aggregate_expression_requires_column_from_same_table.rs:21:31
   |
LL |     let source = users::table.select(sum(posts::id));
   |                               ^^^^^^ the trait `SelectableExpression<users::table>` is not implemented for `posts::columns::id`
```

So, what exactly did you mean?

In fact, I actually think the error annotations are just duplicated effort. All the information about compile errors is available in stderr files. Unless, I missed a certain file where this makes sense?

weiznich:
Yes all the information are strictly speaking in the stderr files, so yes it's duplicated effort to also have some annotation in the source. 

The point I wanted to make is that you don't need to look at the stderr files anymore to make sure that an error is emitted at a specific location. For test cases that generate quite a few errors that's at least for me helpful as I don't need to check the rather noisy stderr file in detail to understand if something important (something compiled and it shouldn't or vice versa) changed. 
Now it's up to discussion if that benefit is worth the effort or not. I personally feel yes, as adding the annotation is mostly a one time effort. If they turn out to change a lot we can also simplify the to only contain the relevant key word instead the exact message. 

Finally, that's only a byproduct of the change. The main reason to change is that trybuild just doesn't allow to check for the error in https://github.com/diesel-rs/diesel/pull/4322/commits/1411136d9029474d9c6de04247fd9d41bde29b0d, but ui_test allows that. The only alternative would be compile_test which also requires these annotations. Although we could check if there is an option in ui_test to disable the annotations.

pksunkara:
I have no issues regarding the reason to change to use `ui_test` lib.

> I personally feel yes, as adding the annotation is mostly a one time effort. If they turn out to change a lot we can also simplify the to only contain the relevant key word instead the exact message.

Okay, let's merge this, and we can revisit this later if needed.

PR #4659, opened 2025-06-23
feat(diesel): Add user-defined error type (fixes #4658)

kriswuollett:


weiznich:
Thanks for opening this PR. As written in https://github.com/diesel-rs/diesel/discussions/4658#discussioncomment-13550307 Diesels error type is neither designed nor supposed to be used as you applications error type, so I really see no argument of adding this extension.

kriswuollett:
Right, I can see the point about `From<diesel::result::Error>` now, will reply in the discussion thread.

PR #4665, opened 2025-06-29
Remove unnecessary export module path

Spxg:
The export module was introduced in 0.2, and 0.3 made it compatible, this compatibility will be removed in 0.4.


weiznich (approved):
Thanks :+1:

PR #4671, opened 2025-07-03
Fix print-schema crash with mariadb client library

xanonid:
In certain Linux distributions, the MariaDB client library incorrectly returns a length of 0 for NULL Decimal fields, rather than utilizing the is_null flag as expected.

`diesel print-schema` fails in this case with
`Failed to execute a database query: Error deserializing field 'character_maximum_length': could not convert slice to array`.

This patch introduces an additional check to handle this specific case.

See discussion #4630.

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -410,6 +410,12 @@ impl BindData {
         } else {
             let data = self.bytes?;
             let tpe = (self.tpe, self.flags).into();
+            // On some distributions, the mariadb client library returns length 0 for NULL fields of type DECIMAL
+            // instead of using is_null for unknown reasons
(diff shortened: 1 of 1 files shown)


simonzkl:
Given the recent discussions in https://github.com/diesel-rs/diesel/issues/4699#issuecomment-3143830335 and https://github.com/diesel-rs/diesel/discussions/4806#discussioncomment-14758499

Was this workaround even correct? I could be very wrong, but to my understanding the mariadb c connector now initializes `length` to 0 for all fixed-size buffer types (to avoid potentially uninitialized fields). So isn't this just unconditionally making all ~~decimal~~ longlong values `NULL`?

weiznich:
@simonzkl Thanks for pointing out. Yes this fix is not correct, fortunately it only will produce wrong results with the otherwise broken mariadb lib. We should revert it nevertheless

PR #4675, opened 2025-07-06
Try to fix metrics job

sabify:
What's changed:

- Properly defined `async-diesel` integral dependency via git and without versioning. It always tries to fetch the latest commit to be in sync with the latest `diesel` changes.
- Pinned `wtx` dependency to `v0.34.0` as the new `v0.35.0` uses `rsa v0.10.0-rc1` dependency and causes failed compilation.
- Separated `wtx` `postgres` and `mysql` features.


weiznich (approved):
Thanks ❤️

PR #4695, opened 2025-07-22
Reduce generic code instanciation in one of the postgres connection functions

weiznich:
This commit turns the instrumentation code in
`update_transaction_manager_status` in the `PgConnection` and `MysqlConnection`  implementation
into a non-generic function, which hopefully should reduce the number of times rustc tries to generate that function. The relevant code doesn't depend on the generic paramater at all.

### DIFF
--- a/diesel/src/pg/connection/mod.rs
+++ b/diesel/src/pg/connection/mod.rs
@@ -332,20 +332,30 @@ fn update_transaction_manager_status<T>(
             }
         }
     }
-    non_generic_inner(conn, query_result.is_err());
-    if let Err(ref e) = query_result {
-        conn.instrumentation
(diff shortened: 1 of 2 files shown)


diesel/src/pg/connection/mod.rs
```diff
@@ -332,23 +332,32 @@ fn update_transaction_manager_status<T>(
...
+                    error: None,
+                });
+        }
     }
+
+    non_generic_inner(conn, query_result.is_err());
+    non_generic_instrumentation(query_result.as_ref().map(|_| ()), conn, source, final_call);
     query_result
 }
-
```
pksunkara, line 351: I think this line is mistakenly removed, right? I supposed the format checks would fail.

pksunkara (approved):

PR #4701, opened 2025-07-26
Introduce a testing framework for our proc-macros

weiznich:
This change introduces a testing framework for our proc-macros. We internally already use proc_macro2. To easily check out macro-expansions we now write small test inputs (via `quote::quote!()`), call the corresponding expansion code in a unit test, and finally format the expansion via `rustfmt`. The result in then stored in a insta snapshot file. This hopefully enables iterating faster in on proc-macro changes as it's now easily possible to inspect the output without needing additional steps.

We should extend the base set of tests in this PR in future updates with more attribute variations for each proc macro. 

Finally we can also think about showing at least some of the expansions in our documentation to help users understand what the proc macro does internally.


guissalustiano:
Did you take a look on [prettyplease](https://github.com/dtolnay/prettyplease)? 
> This is a pretty-printer to turn a syn syntax tree into a String of well-formatted source code. In contrast to rustfmt, this library is intended to be suitable for arbitrary generated code.

weiznich:
I changed to formatting to prettyplease, seems like that handles long lines better. I also included the input code in the snapshot files, so that we now easily can generate documentation having both the input and the expanded code. I will work on a follow-up PR that generates the relevant documentation via a build script. 

Finally I moved every test in a separate module per derive/macro as I expected that we expand the tests there soon to cover more options. That's possible a good issue for smaller contributions as it is well defined and can be easily split up into small parts.

PR #4703, opened 2025-07-29
Use `#[doc(fake_variadic)]` to improve documentation for tuple impls

weiznich:
This commit uses `#[doc(fake_variadic)]` to improve the documentation displayed for trait impls for tuples. This is an undocumented rustdoc attribute, on the other hand it's also used by bevy: https://github.com/bevyengine/bevy/pull/14703

Rendered new documentation:

image

Old documentation: https://docs.diesel.rs/2.2.x/diesel/sql_types/trait.SqlType.html#foreign-impls (long blocks of the "same" impl again and again)


pksunkara:
I see `T1::IsNull: OneIsNullable<T0::IsNull>` and so on in the old docs. Is that part unnecessary to be shown in the docs here?

##### Improvements for future

1. Allow "T0<sub>n</sub>" to be set specifically "T0<sub>16</sub>"
2. Use `T` instead of `T0`

weiznich:
> I see T1::IsNull: OneIsNullable<T0::IsNull> and so on in the old docs. Is that part unnecessary to be shown in the docs here?

It's mostly important for the compiler to calculate the right value for the associated type. It shouldn't be important for users as they shouldn't be able to construct any variant where that trait is not satisfied. We only allow a fixed set of types there and all satisfy that bound. 

> Allow "T0n" to be set specifically "T016"

That requires support from rustdoc, so we would need to land a change there. 

> Use T instead of T0

That should be easily possible by just change that identifier in the diesel_for_each_tuple proc macro, I will do that tomorrow.

pksunkara:
> That should be easily possible by just change that identifier in the diesel_for_each_tuple proc macro, I will do that tomorrow.

I was cautious in proposing that because I am not sure how the change looks in other places (IDEs, compiler errors, etc). I will leave it to your discretion.

weiznich:
> I was cautious in proposing that because I am not sure how the change looks in other places (IDEs, compiler errors, etc). I will leave it to your discretion.

It doesn't look too bad in my opinion. See the error messages for an example.

Issue #4705, opened 2025-07-30
Weird Array Behaviour


4A6F6F6E61:
### Versions

- **Rust:** rustc 1.88.0
- **Diesel:** 2.2.12
- **Database:** PostgreSQL
- **Operating System:** CachyOS (Arch Linux)

### Feature Flags

-

## Problem Description
I want to create a table with Arrays, but when I generate the sql,  I get invalid SQL because the [] are not after the sql type but at the end of the line, and diesel puts  `NOT NULL` inbetween the type and brackets.

### What are you trying to accomplish?
Generate valid SQL that represents my schema

### What is the expected output?
```sql
CREATE TABLE "games"(
	"id" INT4 NOT NULL PRIMARY KEY,
	"title" TEXT NOT NULL,
	"prefixes" INT4[] NOT NULL,
	"tags" INT4[] NOT NULL
);
```

### What is the actual output?
```sql
CREATE TABLE "games"(
	"id" INT4 NOT NULL PRIMARY KEY,
	"title" TEXT NOT NULL,
	"prefixes" INT4 NOT NULL[],
	"tags" INT4 NOT NULL[]
);
```

### Are you seeing any additional errors?
err=DatabaseError(Unknown, "syntax error at or near \"[\"")

### Steps to reproduce
1. Create a table with an Array

```rust schema.rs
diesel::table! {
    games (id) {
        id -> Int4,
        title -> Text,
        prefixes -> Array<Int4>,
        tags -> Array<Int4>,
    }
}
```

2. Cenerate the SQL from the schema:

```fish
diesel migration generate --diff-schema create_games
```

3. Run the migrations

```fish
diesel migration run
```

## Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
 
- [x] This issue can be reproduced on Rust's stable channel. (Your issue will be
  closed if this is not the case)
- [x] This issue can be reproduced without requiring a third party crate

PR #4706, opened 2025-07-30
Match syntax generation with what migration run expects

4A6F6F6E61:
Change the SQL generation output so it matches what `diesel migration run` expects.
I opened an issue about this issue: [here](https://github.com/diesel-rs/diesel/issues/4705)

According to [custom_typed.md](https://github.com/diesel-rs/diesel/blob/master/guide_drafts/custom_types.md) and my testing, `diesel migration run` expects the Array modifier `[]` to come immediatly after the type, and not at the end of the line like in most standard SQL versions.

This fix just slightly moves the line that appends the array modifier up.

### DIFF
--- a/diesel_cli/src/migrations/diff_schema.rs
+++ b/diesel_cli/src/migrations/diff_schema.rs
@@ -614,11 +614,11 @@ where
 {
     // TODO: handle schema
     query_builder.push_sql(&format!(" {}", ty.sql_name.to_uppercase()));
-    if let Some(max_length) = ty.max_length {
-        query_builder.push_sql(&format!("({max_length})"));
-    }
     if ty.is_array {
         query_builder.push_sql("[]");
+    }
+    if let Some(max_length) = ty.max_length {
+        query_builder.push_sql(&format!("({max_length})"));
     }
     if !ty.is_nullable {
         query_builder.push_sql(" NOT NULL");


4A6F6F6E61:
But the issue is that if you do this, Diesel seems to stop checking for anything after the Brackets when you run `diesel migration run`, so it modifes the `schema.rs` and sets every Array to be nullable, because it doesn't check `NOT NULL anymore`. So this only fixes a small part of the issue.

4A6F6F6E61:
And it also causes problems with `--diff-schema`, because diesel always seems to think that it has to update the sql, because it thinks it is not on the latest version change, because it doesn't realize that `Array<Int4>` is the same as `INT4[] NOT NULL`

weiznich:
Thanks for opening this PR. Could you also add a test for this case here: https://github.com/diesel-rs/diesel/blob/master/diesel_cli/tests/migration_generate.rs

---

> And it also causes problems with --diff-schema, because diesel always seems to think that it has to update the sql, because it thinks it is not on the latest version change, because it doesn't realive that Array<Int4> is the same as INT4[] NOT NULL

That's expected, as `INT4 [] NOT NULL` is really equivalent to `Array<Nullable<Int4>>` as even a non-nullable array can contain null values.

4A6F6F6E61:
Ah ok, good to know! I will add a test case later.

PR #4714, opened 2025-08-06
Include expanded code for derives in our documentation

weiznich:
This change includes the expanded code for all our derives in our API documentation. This makes it hopefully a bit easier for our users to understand what the macros actually do and also to estimate how much code they generate.

Technically we just take the snapshots from the tests #4701 and use a build script to generate the relevant documentation snippets. The output of the build script is then included into the documentation. It's a bit unfortune that we cannot check in the build script if it's run by rustdoc vs some other compilation feature so I decided to gate this on the `docsrs` cfg flag to not run that step in "normal" builds.

This change also enables the `--generate-link-to-definition` option for rustdoc.

Example Documentation:

image

### DIFF
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -116,6 +116,7 @@ mod valid_grouping;
 ///   `treat_none_as_null` attribute for the current field.
 /// * `#[diesel(skip_update)]`, skips updating this field. Useful for working with
 ///   generated columns.
+#[cfg_attr(docsrs, doc = include_str!(concat!(env!("OUT_DIR"), "/as_changeset.md")))]
 #[cfg_attr(
     all(not(feature = "without-deprecated"), feature = "with-deprecated"),
     proc_macro_derive(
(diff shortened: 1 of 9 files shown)


diesel_derives/src/lib.rs
```diff
@@ -1023,6 +1142,43 @@ fn derive_selectable_inner(input: proc_macro2::TokenStream) -> proc_macro2::Toke
 /// * `#[diesel(mysql_type(name = "TypeName"))]`, specifies support for a mysql type
 ///   with the given name. `TypeName` needs to be one of the possible values
 ///   in `MysqlType`
+///
+#[cfg_attr(docsrs, doc = "# Expanded Code")]
+#[cfg_attr(
+    all(docsrs, feature = "sqlite"),
+    doc = r#"
+### SQLite
```
pksunkara: I can't imagine any other differentiation other than this. So, why don't we:

1. Add the `# Expanded Code` and `<details>` part in the md files.
2. Allow handling of sqlite, postgres etc... there?

That way, we can just `include_str!` and nothing else.

Please let me know if my suggestion doesn't work.

weiznich: You are correct that this is easier. I originally started with that approach, but expected that I need more variants so I went with that cfg_attr based approach. In hindsight it is more meaningful to handle everything in the build script even if we include more variants (for different attribute combinations for example) in the future.

pksunkara (approved):

PR #4718, opened 2025-08-09
feat(sqlite): add `json_patch` and `jsonb_patch` functions

KekmaTime:
Adds SQLite support for the following functions : 
*  `json_patch` 
*  `jsonb_patch`

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -1425,6 +1425,137 @@ extern "SQL" {
         json: J,
         path: Text,
     ) -> Nullable<Jsonb>;
+
+    /// Applies an RFC 7396 MergePatch `patch` to the input JSON `target` and
+    /// returns the patched JSON value.
+    ///
+    /// MergePatch can add, modify, or delete elements of a JSON object. Arrays are
+    /// treated as atomic values: they can only be inserted, replaced, or deleted as a
+    /// whole, not modified element-wise.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -1425,6 +1425,138 @@ extern "SQL" {
...
+    /// # }
+    /// ```
+    #[cfg(feature = "sqlite")]
+    fn json_patch<
+        T: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue,
+        P: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue,
+    >(
+        target: T,
+        patch: P,
+    ) -> Nullable<Json>;
```
weiznich: As far as I can see the return type is only null if one of the inputs is null. 

We need to do something like 


```suggestion
    fn json_patch<
        T: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue,
        P: JsonOrNullableJsonOrJsonbOrNullableJsonb + SingleValue + CombinedNullableValue<T, Json>,
    >(
        target: T,
        patch: P,
    ) -> P::Out
```

This will evaluate to `Nullable<Json>` if one of the types `T` or `P` is nullable and to `Json` if both types are not nullable.

weiznich:
Thanks for the update. Would it be possible to apply the same change for `jsonb_patch` as well?

zacharyftw:
Turns out I had already made the `jsonb_patch` changes in the first correction . My apologies for the mix-up.

PR #4722, opened 2025-08-13
feat(sqlite): add `json_object` and `jsonb_object` functions

KekmaTime:
Adds SQLite support for the following functions :

* `json_object`
* `jsonb_object`

Partially completes https://github.com/diesel-rs/diesel/issues/4366

### DIFF
--- a/diesel/src/sqlite/expression/functions.rs
+++ b/diesel/src/sqlite/expression/functions.rs
@@ -995,6 +995,109 @@ extern "SQL" {
     #[aggregate]
     fn jsonb_group_array<E: SqlType + SingleValue>(elements: E) -> Jsonb;
 
+    /// The `json_object()` SQL function accepts zero or more pairs of arguments and returns a
+    /// well-formed JSON object composed from those pairs. The first argument of each pair is the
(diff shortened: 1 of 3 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -995,6 +995,109 @@ extern "SQL" {
...
+    /// )
+    /// .get_result::<serde_json::Value>(connection)?;
+    /// assert_eq!(json!({"a": 2, "c": {"e": 5}}), result);
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "sqlite")]
+    #[variadic(2)]
+    fn json_object<K: NotBlob, V: NotBlob>(key: K, value: V) -> Json;
```
weiznich: We need to restrict the keys to be not a null value as SQLite doesn't accept a null value there. We should be able to do this via `K: NotBlob<IsNull = is_nullable::NotNull>`. This also applies to `jsonb_object`.

zacharyftw: done 👍🏽

zacharyftw:
@weiznich

PR #4726, opened 2025-08-23
Fix MySQL benchmarks

c410-f3r:
https://github.com/diesel-rs/diesel/actions/runs/17170711667/job/48719768348#step:12:18

> thread 'main' panicked at benches/mysql_benches.rs:32:36:
called `Result::unwrap()` on an `Err` value: DriverError { Could not connect to address `localhost:3306': Connection refused (os error 111) }

Other locations also use `127.0.0.1` instead of `localhost` but I don't know why.

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -61,8 +61,8 @@ jobs:
           sudo systemctl start mysql.service
           sudo apt-get update
           sudo apt-get -y install libmysqlclient-dev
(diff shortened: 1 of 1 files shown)


c410-f3r:
>  error: `accidently` should be `accidentally`
  --> ./CHANGELOG.md:118:13
    |
118 | * Increased accidently decreased limit around element count in  `DISTINCT ON` and `ORDER BY` clauses again as that broke existing code
    |             ^^^^^^^^^^
    |

Doesn't seem related

weiznich (approved):
Thanks, looks good to me

(rebased to pull in the fix for the CI issue)

PR #4743, opened 2025-09-04
Fix sqlite tests in nightly toolchain

Spxg:
`Barrier` internally has now switched to `nonpoison::Mutex` and is no longer `RefUnwindSafe`

### DIFF
--- a/diesel/src/sqlite/connection/row.rs
+++ b/diesel/src/sqlite/connection/row.rs
@@ -357,6 +357,7 @@ mod tests {
         use crate::connection::SimpleConnection;
         use crate::expression_methods::ExpressionMethods;
         use crate::SqliteConnection;
+        use std::panic::AssertUnwindSafe;
         use std::sync::{Arc, Barrier};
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR.

That's an upstream rust regression and should be fixed there. See https://github.com/rust-lang/rust/issues/146087 for details

Spxg:
> That's an upstream rust regression and should be fixed there. See [rust-lang/rust#146087](https://github.com/rust-lang/rust/issues/146087) for details

Looks good

Spxg:
Considering that nightly tests cannot run now, I think it can be used as a temporary fix and revert after the rust fix is completed.

weiznich:
The nightly builds are configured to fail but not to block merging etc. So it's not that critical in my opinion to have a timely fix. It would be great to have it fixed in rust itself before that hits the beta, I will likely start poking people for that next week. Otherwise you also can submit a PR to just revert the problematic change there.

PR #4747, opened 2025-09-05
Fix debug binds for sqlite in diesel-async

weiznich:
We did not emit proper debug information for binds that were moved due to the async connection wrapper. This commit does improve that by emitting what information are still there.

Fixes https://github.com/weiznich/diesel_async/issues/256


JohnTitor (approved):
Looks reasonable.

PR #4751, opened 2025-09-08
Add ToSql Numeric support for SQLite MultiConn

BlackDex:
When trying to convert Vaultwarden to use the MultiConnection feature, I bumped into an issue with Numeric/BigDecimal for SQLite. This PR should fix this issue by adding the `ToSql` impl.

Also added a test for both single and multi connection.

Fixes #4735


BlackDex:
I'm unsure on how to solve the currently broken CI Tests to be honest.
I run all the items described in the Contribution Readme, without any additional failures compared to the master branch as far as i could tell.

If those need to be fixed, please let me know and if possible some hints would be nice 😄 thanks!

weiznich:
You should be able to reproduce the same error with the command shown in line 2 [here](https://github.com/diesel-rs/diesel/actions/runs/17548279944/job/49835095737#step:25:1). The env variables are expanded in the block below. 

From a quick look at the error it seems like all you have to do is to run the test once to get that failure and the accept the changed output via `cargo insta review` (requires cargo insta to be installed). That test checks the expanded code for the MuliConnection derive. I added that a few weeks ago to include the expanded code in the documentation. Seems like I should add a note about that in the documentation.

BlackDex:
Well, it took me a while, bumping into strange errors i didn't see in the CI logs. But it was the `Cargo.lock` which was holding too old version. Had to remove that first. Running those test now again, lets see if that helps.

weiznich:
Thanks for the update. I will take care of fixing the remaining CI errors + cleaning up this change hopefully in the next days.

weiznich (approved):

PR #4753, opened 2025-09-09
Add support for JSON columns in schema inference

3ach:
Running `diesel migration generate --diff-schema <whatever>` doesn't work for SQLite databases + JSON columns. If I manually create migrations, they work fine, so it looks like this was just missed when JSON support was added for SQLite. 

To test this, I applied the patch locally and generated migration.

### DIFF
--- a/diesel_cli/src/infer_schema_internals/sqlite.rs
+++ b/diesel_cli/src/infer_schema_internals/sqlite.rs
@@ -363,6 +363,8 @@ pub fn determine_column_type(
         String::from("Date")
     } else if type_name == "time" {
         String::from("Time")
+    } else if type_name == "json" {
+        String::from("Json")
     } else {
         return Err(crate::errors::Error::UnsupportedType(type_name));
     };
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. We didn't miss this, but decided against implementing this as SQLite usually want's you to call either the `json` or the `jsonb` function to construct values of the relevant types.

3ach:
Sorry, I couldn’t find discussion about this anywhere before I started
working on it. Is there something I missed? The docs are a little unclear
on how to use json and SQLite together.

weiznich:
@3ach As this is a really new feature it might be very well that it's not documented enough yet. We are always happy to receive PR's that extend the documentation.

3ach:
I’m happy to write a PR, but I am not sure what you meant by 'SQLite usually want's you to call either the json or the jsonb function to construct values of the relevant types’ since the JSON types do exist natively in SQLite. I read your first message as implying there was some discussion about whether or not to include this, so it would be helpful if you could point me towards that discussion so I have the context necessary to write a helpful PR. Thanks!

3ach:
I found this discussion: https://github.com/diesel-rs/diesel/discussions/4192 which seems related. Thanks, and sorry for the noise!

PR #4766, opened 2025-09-15
Bug fix and decimal support for casting

Sagebati:
Hi, 

In this PR 

- I corrected an blanket impl that didn't convey they original goal. able to cast Nullable<From> -> Nullable<To>
- Added support to cast from and to for the type Decimal


weiznich (commented):
Thanks for submitting this PR. The change itself looks good to me, but I would prefer having at least some tests that cover the new variants.

Sagebati:
Sure, from what I see, there were no previous tests, where can I write them ?

weiznich:
That would be great :heart: Adding these tests as new module in `diesel_tests` should be fine.

Sagebati:
The test weren't good, I tried to do do a macro_rules to test all the cases, but didn't succeed in it. So I removed the macro method. I tested two cases, is it enough ?

diesel/src/expression/cast.rs
```diff
@@ -190,16 +190,20 @@ macro_rules! casts_impl {
...
     (Float4 <- Int8),
     (Float8 <- Float4),
     (Float8 <- Int4),
     (Float8 <- Int8),
-    (Int8 <- Int4),
-    (Int8 <- Float4),
-    (Int8 <- Float8),
-    (Int4 <- Bool),
-    (Int4 <- Float4),
+    (Decimal <- Int4),
```
weiznich, line 203: I removed the `Decimal <- Bool` cast from this list as that's not supported by postgres

PR #4778, opened 2025-09-21
Generalized optional extension

LucaCappelletti94:
Hi,

In some of the crates where I employ diesel, I wrap the [`diesel::result::Error`](https://docs.diesel.rs/master/diesel/result/enum.Error.html) in other enums. In some cases, down the line, I would like to use the `optional` method from the [`OptionalExtension`](https://docs.diesel.rs/master/diesel/result/trait.OptionalExtension.html) trait to convert the case where the row is not there into an Option. At this time this is not possible as the trait implementation requires for specifically the `diesel::result::Error` Err variant, so I wanted to generalize it to anything that was able to convert any error which may represent such case.

I would rather not introduce a new trait in another crate which provides the same functionality, as `OptionalExtension` seems easily generalisable to handle such a use case.

In order to do this, I have:

1. Added a new trait called `IsOptionalVariant` which provides a method to determine whether a given error is the optional variant
2. Added an associated `Error` type to the `OptionalExtension` trait, so to be able to implement it for any error type which implements `IsOptionalVariant`
3. **[Not necessary for PR, done solely as it seemed cleaner]** Tentatively removed the generic `T` from `OptionalExtension` and replaced it also with an associated type, as the sole use case scenarios for this trait seems to be `Result` enumerations which already have the `T` generic

Best,
Luca

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -260,10 +260,25 @@ pub type QueryResult<T> = Result<T, Error>;
 /// is otherwise a direct mapping to `Result`.
 pub type ConnectionResult<T> = Result<T, ConnectionError>;
 
+/// A trait for error types that can be used with `OptionalExtension`
+/// and are characterized by having a variant that indicates "not found".
+pub trait IsOptionalVariant {
+    /// Returns `true` if the error is the variant that indicates "not found".
+    fn is_optional_variant(&self) -> bool;
+}
+
+impl IsOptionalVariant for Error {
+    fn is_optional_variant(&self) -> bool {
+        matches!(self, Error::NotFound)
+    }
+}
+
 /// See the [method documentation](OptionalExtension::optional).
-pub trait OptionalExtension<T> {
+pub trait OptionalExtension {
     /// The associated error type.
-    type Error;
+    type Error: IsOptionalVariant;
+    /// The item type to wrap in an `Option`.
+    type Item;
 
     /// Converts a `QueryResult<T>` into a `QueryResult<Option<T>>`.
     ///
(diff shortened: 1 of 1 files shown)


weiznich:
Thanks for opening this PR. This sounds like a nice idea to implement.

Unfortunally [`OptionalExtension`](https://docs.diesel.rs/2.3.x/diesel/prelude/trait.OptionalExtension.html) is part of our public API, which means there are quite a lot restrictions on what exactly we can change there without issuing a new major release. Specifically we cannot:

* Add a new associated type
* Remove the generic parameter

We don't plan to release such a major release soon, so this would need to find a other solution.

LucaCappelletti94:
Understood, I was worried that may be a valid concern here. Since the returned error in the trait is a diesel Error, if we cannot modify the trait, the only solution I can see is to create a new trait that may supersede the older trait once a major release is created and they may co-exist before then. Would that be an acceptable trade-of? It is certainly somewhat sub-optimal.

weiznich:
I'm usually open to such solution as long as the impact of such a change is large enough. For this particular case I'm rather skeptical as others still need to implement a trait to use this method on their error types. Given that I tend to not going that duplication way as if you still need to implement a specific trait to use this you can also with minimally more effort do that with the current infrastructure. With the proposed change it would be something like 5 lines of code, without the change it would be something like 12 lines of code. That doesn't sound like a large win to me that would justify duplicating such a trait. If someone from @diesel-rs/reviewers has a different opinion here I would be happy to hear about them to consider that.

LucaCappelletti94:
I am of the same opinion, maybe we should revisit this when an API breaking release is closer. Is there any wish-list kind of document where to add this type of thing?

weiznich:
We have that `diesel 3.0` tag for this to track potential things for a 3.0 release. I've marked this PR as such and would otherwise delay any further discussion until a 3.0 release is planned. To be clear here: We don't plan to issue any breaking changes in the next time, so this could be literally years until that becomes relevant.

To not have sitting that as open in the issue tracker until then I would like to close this PR.

PR #4783, opened 2025-09-23
Artifact attestation + updates for actions

weiznich:
This commit enables githubs artifact attestation for the binaries we build via cargo-dist, updates the checkout action to checkout@v5 and tweaks the dependabot settings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,9 @@
 
 name: Release
 permissions:
+  "attestations": "write"
   "contents": "write"
+  "id-token": "write"
 
 # This task will run whenever you push a git tag that looks like a version
 # like "1.0.0", "v0.1.0-prerelease.1", "my-app/0.1.0", "releases/v1.0.0", etc.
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -144,6 +150,10 @@ jobs:
           # Actually do builds and make zips and whatnot
           dist build ${{ needs.plan.outputs.tag-flag }} --print=linkage --output-format=json ${{ matrix.dist_args }} > dist-manifest.json
           echo "dist ran successfully"
+      - name: Attest
+        uses: actions/attest-build-provenance@v2
```
JohnTitor, line 150: This would fail if a PR comes from a fork, see https://github.com/actions/attest-build-provenance/issues/99.
We have to gate to run this action only on push, or ignore it.

weiznich: These builds are only running for pushs for tags by default. I only enabled them to run on PR's to actually test the feature. seems like I should use a different branch for that.

weiznich: For reference: Succeeding tests builds are here: https://github.com/diesel-rs/diesel/actions/runs/17947046473/job/51036009308?pr=4784
The resulting attestation is there: https://github.com/diesel-rs/diesel/attestations/11029798

weiznich: (I now disabled the pr build mode again so this should only run for pushs of tags which should always be inside this repo)

JohnTitor: Ah seems disabling fixed the issue, thanks!

PR #4785, opened 2025-09-24
Bump `sqlite-wasm-rs` to 0.4.4

Spxg:
This release brings several improvements:
1. No longer requires linking to emscripten's libc, resulting in a smaller and more stable build.
2. No longer requires the `wasm-array-cp` crate, meaning no additional snippet files. https://github.com/Spxg/sqlight/commit/3fc6db9559eac3b809d4a0a0edb85248794842a8#diff-e9e38416882591dc00cb970c3958a2fce5290eb4769c0e1bc5c3b04653b9ab8e
3. Removed several unnecessary dependencies.
4. Optimized the build.rs build process.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -53,13 +53,13 @@ path = "../diesel_derives"
 libsqlite3-sys = { version = ">=0.17.2, <0.36.0", optional = true, features = ["bundled_bindings"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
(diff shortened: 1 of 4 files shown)


weiznich:
Thanks for notifying us for the new release. Given that we released diesel 2.3 a few days back including support for `sqlite-wasm-rs` it's not possible anymore to increase the minimal version of that dependency. That would be a breaking change for any user depending on a specific version being used. This shouldn't be a problem as long as the new version stay compatible with the newer versions.

Spxg:
> This shouldn't be a problem as long as the new version stay compatible with the newer versions.

Agree.

PR #4798, opened 2025-09-30
sql_types::Json: Add documentation how to de/serialize to custom type

matthiasbeyer:
Following https://github.com/diesel-rs/diesel/discussions/4793#discussioncomment-14543542, this adds documentation to the `sql_types::Json` type on how to de/serialize from/to custom types with JSON.

Right now this is in draft mode, to get some feedback whether I added the docs in the right location (I believe so, because that's where I'd expect them) and whether the idea of having a code example at that location was the right way of doing this.

CC @weiznich

### DIFF
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -392,6 +392,12 @@ pub struct Timestamp;
 ///
 /// - [`serde_json::Value`]
 ///
+/// ## Using custom types for JSON deserialization
+///
+/// Using a custom type that can be serialized or deserialized to/from JSON should be done by using
+/// the `serialize_as` and `deserialize_as` attribute macros. For more information have a look at
+/// the [Queryable](crate::deserialize::Queryable) trait documentation.
+///
 /// [`ToSql`]: /serialize/trait.ToSql.html
 /// [`FromSql`]: /deserialize/trait.FromSql.html
 /// [`serde_json::Value`]: /../serde_json/value/enum.Value.html
--- a/diesel/src/sql_types/mod.rs
+++ b/diesel/src/sql_types/mod.rs
@@ -398,6 +398,91 @@ pub struct Timestamp;
 /// the `serialize_as` and `deserialize_as` attribute macros. For more information have a look at
 /// the [Queryable](crate::deserialize::Queryable) trait documentation.
 ///
+/// ```rust
+/// # #![allow(dead_code)]
+/// # include!("../doctest_setup.rs");
+/// #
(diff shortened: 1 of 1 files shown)


diesel/src/sql_types/mod.rs
```diff
@@ -392,6 +392,97 @@ pub struct Timestamp;
...
+/// #     diesel::sql_query("CREATE TABLE contacts (
+/// #         id INT PRIMARY KEY,
+/// #         name TEXT NOT NULL,
+/// #         address BLOB NOT NULL
+/// #     )").execute(connection)?;
+/// #[derive(Queryable, Selectable)]
+/// struct Contact {
+///     pub id: i32,
+///     pub name: String,
+///     #[diesel(deserialize_as = serde_json::Value)]
```
weiznich, line 440: This requires an `impl TryFrom<serde_json::Value> for Address`

diesel/src/sql_types/mod.rs
```diff
@@ -392,6 +392,97 @@ pub struct Timestamp;
...
+///     state: "Alaska".to_string(),
+/// };
+///
+/// #let inserted_address = insert_into(contacts)
+/// #    .values((name.eq("Claus"), address.eq(&santas_address)))
+/// #    .returning(address)
+/// #    .get_result::<serde_json::Value>(connection)?;
+///
+/// let santa = contacts.filter(id.eq(1)).first(connection)?;
+/// assert_eq!(santas_address, santda.address);
```
weiznich, line 473:
```suggestion
/// assert_eq!(santas_address, santa.address);
```

that's likely a typo

diesel/src/sql_types/mod.rs
```diff
@@ -392,6 +392,97 @@ pub struct Timestamp;
...
+/// )?;
+/// let santas_address: Address {
+///     street: "Article Circle Expressway 1".to_string(),
+///     city: "North Pole".to_string(),
+///     postcode: "99705".to_string(),
+///     state: "Alaska".to_string(),
+/// };
+///
+/// #let inserted_address = insert_into(contacts)
+/// #    .values((name.eq("Claus"), address.eq(&santas_address)))
```
weiznich, line 468: For this variant you need to directly convert the address to a `serde_json::Value` as there is no attribute that's applied.

weiznich (commented):
Thanks for opening this PR.

I think that's a good location to document these things and it's also a good idea to include examples. I left a few remarks on what's potentially broken and causes the CI to fail.

PR #4802, opened 2025-10-02
Clarify some pg::Statement::execute error messages

Ten0:
Before this commit, when hitting what would now display as "A bind parameter's serialized size is bigger than fits on an i32", the error would instead display as "out of range integral type conversion attempted", which is significantly more difficult to investigate.


guissalustiano (approved):
Cool! Thanks!

PR #4809, opened 2025-10-10
Fix diesel setup always using absolute migrations dir paths

### LINKED ISSUE #4436: Setup should not use an absolute path to set the migrations_directory 'dir' field of diesel.toml

ADSteele916:
Fixes #4436. The [`dunce`](https://crates.io/crates/dunce) crate is required to avoid unwanted Windows-specific behaviour in [`std::fs::canonicalize`](https://doc.rust-lang.org/stable/std/fs/fn.canonicalize.html).

### DIFF
--- a/diesel_cli/Cargo.toml
+++ b/diesel_cli/Cargo.toml
@@ -40,6 +40,7 @@ chrono = { version = "0.4.20", default-features = false, features = [
 clap = { version = "4.4.14", features = ["cargo", "string"] }
 clap_complete = "4"
 dotenvy = "0.15"
+dunce = "1.0.5"
 heck = "0.5.0"
 serde = { version = "1.0.193", features = ["derive", "std"] }
 toml = { version = "0.9", default-features = false, features = ["parse", "serde"] }
--- a/diesel_cli/src/main.rs
+++ b/diesel_cli/src/main.rs
@@ -145,8 +145,13 @@ fn create_config_file(
     let path = Config::file_path(matches);
     if !path.exists() {
         let source_content = include_str!("default_files/diesel.toml").to_string();
+        let migrations_dir_relative =
+            convert_absolute_path_to_relative(migrations_dir, &find_project_root()?);
         // convert the path to a valid toml string (escaping backslashes on windows)
-        let migrations_dir_toml_string = migrations_dir.display().to_string().replace('\\', "\\\\");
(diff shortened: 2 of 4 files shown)


weiznich:
Thanks for opening this PR. I would rather avoid introducing two new dependencies for this feature, so maybe you can explain a bit why they are required?

ADSteele916:
> Thanks for opening this PR. I would rather avoid introducing two new dependencies for this feature, so maybe you can explain a bit why they are required?

My main motivation for adding them was correcting some improper handling of Windows paths that `setup_writes_migration_dir_as_arg_as_relative_path` revealed. The core problem is that Windows has two formats for paths: MS-DOS-style paths that begin with a drive letter (e.g. `C:\foo\`) and NT-style "Universal Naming Convention" (UNC) paths (e.g. `\\?\C:\foo\`). Rust's `std::fs::canonicalize` will convert any Windows path to a UNC path, which causes problems when you would want to show MS-DOS-style paths whenever possible to users.

`dunce` fixes the UNC issue with quite a bit of platform-specific code. I feel that dependency is justified to avoid rewriting something similar in Diesel, which mostly avoids `#[cfg(windows)]` attributes in core code. `pathdiff` is more of a convenience, but I thought it was preferable to rolling our own version of its `diff_paths` function. Neither crate brings in any transitive dependencies. If pressed, I could replace the call to `pathdiff` by essentially inlining it and removing a couple of sections that don't apply to our case. I'd stand by `dunce` though.

weiznich:
Thanks for clarifying. I think using `dunce` sounds ok. It would be great to replace `pathdiff` with a minimal inline version.

ADSteele916:
> Thanks for clarifying. I think using `dunce` sounds ok. It would be great to replace `pathdiff` with a minimal inline version.

Done. Let me know if you'd like any other changes!

Issue #4787, opened 2025-09-25
`FromSqlRow` derive macro fails when in scope where `Ok` is not the one from the standard library


xavierm02:
### Rust Version

1.89.0

### Diesel Version

master

### Diesel Features

irrelevant

### Database Version

irrelevant

### Operating System Version

irrelevant

### What happened?

Derive macro fails

### What did you expect to happen?

Derive macro works

### Steps to reproduce

```rust
Go to one of the examples (e.g. `examples/sqlite/getting_started_step_1`), and add the code below:


pub enum MyErr<T,U> {
    Ok(T),
    Err(U)
}

use MyErr::Ok;

#[derive(Queryable)]
pub struct A();


This initially happened to me with `use eyre::*` in place of `MyErr`.
```

### Compile time error

```
mismatched types
expected enum `Result<A, Box<(dyn std::error::Error + Send + Sync + 'static)>>`
   found enum `MyErr<A, _>`
```

### Checklist

- [x] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.

- [x] This issue can be reproduced on Rust's stable channel. (Please submit the issue in the Rust issue tracker instead if that's not the case)
- [x] This issue can be reproduced without requiring a third party crate

weiznich:
Thanks for reporting. Seems like we could do a bit better in the generated code. Fundamentally we will never be able to avoid that completely without more support form the language as there is currently no way to refer to specific items in proc-macros in unambiguous ways. We just can make this less likely to happen.

This line needs to be changed
https://github.com/diesel-rs/diesel/blob/a3d30eaa6fd53f2f65cc1240957413b372e73326/diesel_derives/src/from_sql_row.rs#L38
Also you likely need to run at least the diesel_derive tests + accepting the slightly changed output.

We are happy to review a PR for this.

lauk20:
Hello, I would like to give this a go :)

weiznich:
@lauk20 Sure that would be welcome :tada: Just let us know if you need help somewhere.

lauk20:
I've opened a PR, but it seems as if the issue does not stem from the `FromSqlRow` macro based on the example given in the original issue. However, the issue seems to be similar in the `Queryable` macro, which is where I made the change. The compiler error no longer shows for the example once the change has been made and passes all diesel_derive tests.

[...]

PR #4812, opened 2025-10-10
Return explicit diesel::deserialize::Result Ok type in Queryable derive to handle case of shadowed type.

lauk20:
Addresses #4787

The issue in #4787 does not seem to be from `FromSqlRow` but instead from `Queryable`


weiznich:
Thanks for submitting this PR. I've pushed a change that updates another location as well + that also updates the snapshots for the expected output of the proc-macro expansions as they obviously change. Should be fine for merging now.

PR #4815, opened 2025-10-13
Implement window traits for CountStar

stormshield-kg:
This allows using `OVER()` clause with `COUNT(*)`.

The trait `IsAggregateFunction` cannot be implemented since the `aggregate_distinct()` method will generate `COUNT(DISTINCT *)`, which is not valid SQL.


weiznich (approved):
Looks good, thanks for submitting

PR #4818, opened 2025-10-14
Fix MySQL benchmarks (again)

c410-f3r:
At least on my machine it is not possible to connect to MySQL using `localhost`. However, it seems that CLI commands do require `localhost` for whatever the reason.

* `Current`: https://github.com/diesel-rs/diesel/actions/runs/18484639507/job/52665850375
* `Before` (https://github.com/diesel-rs/diesel/pull/4726): https://github.com/diesel-rs/diesel/actions/runs/16586242685/job/46911932778


weiznich (approved):
Thanks

PR #4829, opened 2025-10-21
feat(sqlite): add json_valid_with_flags function

zacharyftw:
Adds SQLite support for the following function :

* [json_valid(json,flags)](https://sqlite.org/draft/json1.html#jvalid)

Partially completes https://github.com/diesel-rs/diesel/issues/4366

### DIFF
--- a/diesel/src/sqlite/types/mod.rs
+++ b/diesel/src/sqlite/types/mod.rs
@@ -279,31 +279,3 @@ impl ToSql<JsonValidFlags, Sqlite> for JsonValidFlag {
         Ok(IsNull::No)
     }
 }
-
-// Allow i32 for backward compatibility
-#[cfg(feature = "sqlite")]
-impl ToSql<JsonValidFlags, Sqlite> for i32 {
-    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Sqlite>) -> serialize::Result {
-        <i32 as ToSql<sql_types::Integer, Sqlite>>::to_sql(self, out)
-    }
-}
-
-// Allow i32 for backward compatibility
-#[cfg(feature = "sqlite")]
-impl AsExpression<JsonValidFlags> for i32 {
-    type Expression = crate::expression::bound::Bound<JsonValidFlags, i32>;
-
-    fn as_expression(self) -> Self::Expression {
-        crate::expression::bound::Bound::new(self)
-    }
-}
-
-// Allow &i32 for backward compatibility
-#[cfg(feature = "sqlite")]
-impl<'a> AsExpression<JsonValidFlags> for &'a i32 {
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/expression/functions.rs
```diff
@@ -663,6 +663,84 @@ extern "SQL" {
...
+    /// assert_eq!(None, result);
+    ///
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[sql_name = "json_valid"]
+    #[cfg(feature = "sqlite")]
+    fn json_valid_with_flags<X: TextOrNullableTextOrBinaryOrNullableBinary + SingleValue>(
+        x: X,
+        flags: Integer,
```
weiznich: I think I would prefer a more strictly typed approach here. That means:

1. Define a new SQL type in https://github.com/diesel-rs/diesel/blob/main/diesel/src/sqlite/types/mod.rs. That should use `#[diesel(sqlite_type(name = "Integer"))`
2. Define an enum with all the variants listed above (marked as `#[non_exhaustive]`
3. Implement `ToSql<YourNewSqlType, Sqlite>` for that enum + derive `AsExpression` for it.
4. Use this new defined sql type here in place of `Integer`

diesel/src/sqlite/expression/functions.rs
```diff
@@ -663,6 +663,84 @@ extern "SQL" {
...
+    ) -> Nullable<Bool>;
```
weiznich: I think we can make the return type not-nullable for not-nullable inports.

That would require:

1. Changing the bound to `X: TextOrNullableTextOrBinaryOrNullableBinary + SingleValue +  MaybeNullableValue<Bool>`
2. Changing the return type to `X::Out`
3. Changing the tests to use `bool` for non-nullable inputs

diesel/src/sqlite/expression/functions.rs
zacharyftw: Done

diesel/src/sqlite/expression/functions.rs
zacharyftw: Fixed !

diesel/src/sqlite/types/mod.rs
```diff
@@ -214,3 +215,95 @@ impl ToSql<sql_types::Double, Sqlite> for f64 {
...
+
+// Allow &i32 for backward compatibility
+#[cfg(feature = "sqlite")]
+impl<'a> AsExpression<JsonValidFlags> for &'a i32 {
+    type Expression = crate::expression::bound::Bound<JsonValidFlags, &'a i32>;
+
+    fn as_expression(self) -> Self::Expression {
+        crate::expression::bound::Bound::new(self)
+    }
+}
```
weiznich: I think these are not required as we don't need to be backward compatible with the old PR version

PR #4834, opened 2025-10-22
Added `Hash` trait derive (and some `Eq` and `PartialEq`) to `PostgreSQL` types

LucaCappelletti94:
Added some trait derive aimed to facilitate re-use of these structs.


weiznich (approved):
Thanks, seems reasonable

PR #4836, opened 2025-10-23
add support for json(b) build array

valkrypton:
Implemented `json_build_array` and `jsonb_build_array` under [#4216](https://github.com/diesel-rs/diesel/issues/4216)


diesel/src/pg/expression/helper_types.rs
```diff
@@ -621,3 +621,25 @@ pub type jsonb_insert<B, J, R> =
...
+/// Return type of [`json_build_array_1(value1, value2)`](super::functions::json_build_array_2)
+#[allow(non_camel_case_types)]
+#[cfg(feature = "postgres_backend")]
+pub type json_build_array_2<V1, V2> =
+    super::functions::json_build_array_2<SqlTypeOf<V1>, SqlTypeOf<V2>, V1, V2>;
+
+/// Return type of [`jsonb_build_array_1(value1)`](super::functions::jsonb_build_array_2)
+#[allow(non_camel_case_types)]
+#[cfg(feature = "postgres_backend")]
+pub type jsonb_build_array_1<V1> = super::functions::jsonb_build_array_1<SqlTypeOf<V1>, V1>;
```
weiznich: This will be problematic if users adjust how many variadic functions are generated.

We have a `#[declare_sql_function(generate_return_type_helpers = true)]` attribute to make the `#[declare_sql_function]` generate these definitions for you now. 

See here https://github.com/diesel-rs/diesel/blob/bf22cc9751cb4027ab70b78febadbe0d5d7f5bee/diesel/src/sqlite/expression/functions.rs#L16 and here https://github.com/diesel-rs/diesel/blob/main/diesel/src/sqlite/expression/mod.rs#L12 here for a usage in the sqlite backend.

This likely requires cleaning up all the other helper types here as well. I you struggle with that let me know and I take care of this.

valkrypton:
I added the macro and re-exported the helper types similar to the sqlite usage. Should I clean the other types in this pr too?

weiznich (approved):
Looks good now. Thanks for the update.

(I've pushed another commit to remove all now unneeded type defs as they are also generated by the macro)

PR #4843, opened 2025-10-30
added boolean argument in the #[variadic] macro to control generation of 0-variadic-variants

valkrypton:
While working on adding `json_extract_path`, I found out that the function requires at least one variadic argument otherwise it throws an error. Therefore, the variant `json_extract_path_0` with zero variadic argument, generated by `#[variadic]` is invalid, and would result in a runtime error, if used.

This PR adds a boolean flag in the macro that control whether the zero-variadic-argument variant should be generated.

### DIFF
--- a/diesel_compile_tests/tests/fail/derive/return_type_helper_errors.rs
+++ b/diesel_compile_tests/tests/fail/derive/return_type_helper_errors.rs
@@ -44,4 +44,16 @@ mod without_return_type_helpers {
     }
 }
 
+mod backward_compatibility_test {
+    use super::*;
+
+    #[declare_sql_function]
+    extern "SQL" {
+        // this should not generate any error to stay backward compatible with
+        // diesel 2.3
+        #[variadic(1)]
+        fn f<A: SingleValue>(a: <A as TypeWrapper>::Type);
+    }
+}
+
 fn main() {}


diesel_derives/src/lib.rs
```diff
@@ -2385,12 +2385,45 @@ const AUTO_TYPE_DEFAULT_FUNCTION_TYPE_CASE: dsl_auto_type::Case = dsl_auto_type:
...
+/// is generated. By default (omitted or `false`), the 0-argument variant is included. Set it to
+/// `true` to skip generating the 0-argument variant for functions that require at least one
+/// variadic argument.
+///
+/// Example:
+///
+/// ```ignore
+/// #[declare_sql_function]
+/// extern "SQL" {
+///     #[variadic(2, true)]
```
weiznich: Would it be possible to use a named argument for here instead of just a boolean flag? Maybe something like `#[variadic(last_arguments = 2, skip_zero_argument_variant = true)]`?

weiznich (commented):
Thanks for opening this PR. The implementation looks fine to me, but I would like to use a named argument instead of just a boolean flag for this to make a bit clearer what this is supposed to represent.

Also you likely need to run the compilation tests + regenerate the output for at least one test. You can do that by executing `BLESS=1 cargo test` in the `diesel_compile_tests` directory + commit the changes there.

diesel_derives/src/lib.rs
Mingun: Perhaps more consistent and easier to understand will be to implement `#[variadic(num_arguments = 1..2)]`?

valkrypton:
What changes do I need to make in compile tests?

valkrypton:
Nvm fixed it

valkrypton:
Ok no worries, i'll make it backwards compatible

PR #4844, opened 2025-10-30
feat(sqlite): add -> and ->> JSON operators

zacharyftw:
Add support for SQLite's [->](https://sqlite.org/draft/json1.html#jptr) and [->>](https://sqlite.org/draft/json1.html#jptr) JSON extraction operators.

The [->](https://sqlite.org/draft/json1.html#jptr) operator extracts a JSON subcomponent and returns it as a TEXT JSON representation (always returns Json type, never Jsonb).

The [->>](https://sqlite.org/draft/json1.html#jptr)  operator extracts a JSON subcomponent and returns it as an SQL TEXT value (unwrapped).

partially completes #4366

### DIFF
--- a/diesel/src/expression/helper_types.rs
+++ b/diesel/src/expression/helper_types.rs
@@ -6,6 +6,8 @@ use super::grouped::Grouped;
 use super::select_by::SelectBy;
 use super::{AsExpression, Expression};
 use crate::expression;
+#[cfg(any(feature = "postgres_backend", feature = "sqlite"))]
+use crate::expression_methods::JsonIndex;
 use crate::expression_methods::PreferredBoolSqlType;
 use crate::sql_types;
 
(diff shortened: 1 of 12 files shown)


zacharyftw:
@weiznich CI is failing on deprecated ```Command::cargo_bin()``` in example tests (unrelated to my changes). The new ``cargo_bin!`` macro doesn't work with clippy. ```#[allow(deprecated)]``` also didn't help. How should I fix this example tests to pass CI?

weiznich:
@KekmaTime I've opened https://github.com/diesel-rs/diesel/pull/4845 to address this. Let's merge that PR first and then just rebase this one on top.

weiznich:
Seems like this collides with the same operators as defined in the postgres backend if both backends are enabled. I think we should be able to workaround that by rather move the operators + the methods from the extension trait in a central location instead of duplicating them for both backend.

We would need to move:

https://github.com/diesel-rs/diesel/blob/143b254c5b2cc49e2d7d52da69f68eec67a46d86/diesel/src/pg/expression/operators.rs#L46-L52

To https://github.com/diesel-rs/diesel/blob/main/diesel/src/expression/operators.rs#L610

Also we need to move the trait methods to a new trait in a new `json_trait_methods` file here:

https://github.com/diesel-rs/diesel/tree/main/diesel/src/expression_methods

zacharyftw:
Ack, will update 👍

weiznich:
Thanks for the update. This looks good now.

I just rebased the PR to fix the CI issues.

zacharyftw:
@weiznich

weiznich:
I rebased the branch to solve the merge conflict and also pushed some cleanup in https://github.com/diesel-rs/diesel/pull/4844/commits/8d1d4917ae955773d292142da51e6d4610de0928 to fix various things. (name collision for the operators, return type for one of the sqlite methods, etc)

Issue #4699, opened 2025-07-24
Panic in MySQL backend when deserializing nullable integer from non-NULL value: range end index 4 out of range for slice of length 0


biot023:
### Versions

- **Rust: 1.88.0 (6b00bc388 2025-06-23)**
- **Diesel: { version = "2.2", features = ["mysql", "r2d2", "chrono", "64-column-tables", "numeric"] }**
- **Database: MySQL 8.0 (from DockerHub: `mysql-8.0`**
- **Operating System: Ubuntu**

### Feature Flags

- **diesel: [mysql, r2d2, chrono, 64-column-tables, numeric]**

## Problem Description

I'm using Diesel 2.2.12 with MySQL 8.0.42. When loading a Vec of my model (Queryable derive, 12 fields including nullable i32) from a SELECT query, it panics in src/mysql/value.rs:69:76 with "range end index 4 out of range for slice of length 0".

The query is a simple SELECT * FROM basket_items WHERE basket_id = ? ORDER BY created_at ASC.

Fetching single rows by ID works; fetching multiple rows fails. Isolated to one row with normal data (no unusual NULLs or values).

### What are you trying to accomplish?

Querying a multiple rows

### What is the expected output?

A Vec<BasketItem> from the query.

### What is the actual output?

thread 'main' panicked at /home/mycomputer/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.2.12/src/mysql/value.rs:69:76:
range end index 4 out of range for slice of length 0
stack backtrace:
0: __rustc::rust_begin_unwind
at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
1: core::panicking::panic_fmt
at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
2: core::slice::index::slice_end_index_len_fail::do_panic::runtime
at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panic.rs:218:21
3: core::slice::index::slice_end_index_len_fail::do_panic
at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/intrinsics/mod.rs:3196:9
4: core::slice::index::slice_end_index_len_fail
at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panic.rs:223:9
[...]

PR #4851, opened 2025-11-05
fix(mysql): restore bind buffer length for newer mariadb versions

simonzkl:
Works around the issue with newer mariadb client versions that initialise the mysql bind buffer, overwriting the `length` for fixed-sized types.

- Upstream issue: https://jira.mariadb.org/browse/CONC-782
- Previous discussion: https://github.com/diesel-rs/diesel/discussions/4806
- Originally reported in: https://github.com/diesel-rs/diesel/issues/4699

### DIFF
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -379,6 +379,9 @@ impl BindData {
     fn from_tpe_and_flags((tpe, flags): (ffi::enum_field_types, Flags)) -> Self {
         // newer mysqlclient versions do not accept a zero sized buffer
         let len = known_buffer_size_for_ffi_type(tpe).unwrap_or(1);
+        // it's important to initialize the data with zeros here
+        // to make sure we don't expose any uninitialized memory if the underlying library
+        // (again…) skrews up something
         let (ptr, length, capacity) = bind_buffer(vec![0; len]);
 
         Self {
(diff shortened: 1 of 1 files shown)


weiznich:
I think the better solution would be to replace the `self.lenght` value here: https://github.com/diesel-rs/diesel/blob/d51ee0d5dacc1f53c785f4c6e87298d2ffdaa873/diesel/src/mysql/connection/bind.rs#L419

By conditionally checking if the type has a statically known size via `known_buffer_size_for_ffi_type`: https://github.com/diesel-rs/diesel/blob/d51ee0d5dacc1f53c785f4c6e87298d2ffdaa873/diesel/src/mysql/connection/bind.rs#L739

The tricky bit is to reason about whether the buffer is sized that many bytes (which likely can be checked looking at the `capacity` field of the `BindData` struct) and if all possible byte patterns are valid for the target type.

simonzkl:
Yeah I was just hacking around. That sounds good though, I'll give it a shot.

diesel/src/mysql/connection/bind.rs
```diff
@@ -406,6 +406,11 @@ impl BindData {
         } else {
             let data = self.bytes?;
             let tpe = (self.tpe, self.flags).into();
+
+            // FIXME: Could be unsafe. Need to make sure the buffer is large enough.
+            let length = known_buffer_size_for_ffi_type(self.tpe)
+                .unwrap_or(self.length.try_into().expect("Usize is at least 32 bit"));
```
simonzkl: > The tricky bit is to reason about whether the buffer is sized that many bytes (which likely can be checked looking at the capacity field of the BindData struct) and if all possible byte patterns are valid for the target type.

I don't really understand where this would be an issue. The way I see it we already set the correct capacity for fixed-sized types when we first initialise `BindData`. MariaDB then resets the length, but the underlying buffer/capacity should stay untouched, no? And dynamically-sized types should already be handled as the breaking mariadb change only affected fixed-sized types.

Though I think it would be safer to set the correct `self.length` right after calling the mysql ffi functions. It's hard to tell which other functions in the module rely on `self.length` being correct.

weiznich: At least I would expect some checks that the length is smaller or at least equal to the capacity. It should be always the case, but given what the mariadb team did here I would rather not trust that the library behaves in any particular way. Better have an assert for this.

The other question is still: Is any type that we eventually construct from this byte buffer able to represent any possible byte pattern? From a short look it seems like this is the case, but that needs to be verified.

---

> Though I think it would be safer to set the correct self.length right after calling the mysql ffi functions. It's hard to tell which other functions in the module rely on self.length being correct.

Diesel always consumes these data via `MysqlValue` and that's the only location where we actually construct this type. Trying to fix the length directly after calling into libmsqlclient would be rather messy as we have quite a few call sides and we always deal with a list of binds. This place seems like a better solution.

simonzkl:
I can confirm that this fixes my integration tests replicating the issue.

diesel/src/mysql/connection/bind.rs
simonzkl: Oh I see you went ahead and did this yourself. I appreciate the help!

weiznich (approved):
Let's merge it then

PR #4853, opened 2025-11-06
decimal casts

guissalustiano:
This PR adds tests for the cast feature and support to some `int2` and `decimal` casts


guissalustiano:
ops, my branch was not updated

Issue #4856, opened 2025-11-07
Missing `AsChangeset` impl on `&'update Option<T>`


Tibo-Ulens:
### Rust Version

1.93.0

### Diesel Version

2.3.3

### Diesel Features

postgres

### Database Version

PostgreSQL 18.0

### Operating System Version

Arch Linux x86_64 kernel 6.17.7-arch1-1

### What happened?

When using an embedded `Option` field in a struct that derives `AsChangeset` diesel throws a compiler error saying `AsChangeset` is not implemented for `&'update Option<the_struct_in_question>`

eg.

schema
```rust
diesel::table! {
	example (id) {
		id -> Int4,
		one -> Text,
		two -> Text,
		three -> Text,
	}
}
```
model
```rust
#[derive(AsChangeset, Clone, Debug)]
#[diesel(table_name = crate::schema::example)]
#[diesel(check_for_backend(Pg))]
pub struct ExampleUpdate {
	pub one: Option<String>,
	#[diesel(embed)]
	pub subset: Option<ExampleSubSet>,
}

#[derive(AsChangeset, Clone, Debug)]
#[diesel(table_name = crate::schema::example)]
#[diesel(check_for_backend(Pg))]
pub struct ExampleSubSet {
	pub two: String,
	pub three: String,
}
```
will result in
```
the trait bound `&'update std::option::Option<ExampleSubSet>: diesel::AsChangeset` is not satisfied
```
When annotating the non-subset field with `#[diesel(serialize_as = String)]`, thus forcing diesel to use the non-borrowing version of `AsChangeset`, the error disappears:
```rust
#[derive(AsChangeset, Clone, Debug)]
#[diesel(table_name = crate::schema::example)]
#[diesel(check_for_backend(Pg))]
pub struct ExampleUpdate {
	#[diesel(serialize_as = String)]
	pub one: Option<String>,
	#[diesel(embed)]
	pub subset: Option<ExampleSubSet>,
}

#[derive(AsChangeset, Clone, Debug)]
#[diesel(table_name = crate::schema::example)]
#[diesel(check_for_backend(Pg))]
pub struct ExampleSubSet {
	pub two: String,
	pub three: String,
}
```

### What did you expect to happen?

I would expect a `Option<T>` field which is annotated with `#[diesel(embed)]` to implement `AsChangeset`

[...]

PR #4857, opened 2025-11-07
Implement AsChangeset on &Option<T>

Tibo-Ulens:
Implement `AsChangeset` on `&Option<T>` where `&T: AsChangeset`, this allows the `AsChangeset` derive to properly generate its borrowing variant when using `#[diesel(embed)]` on an `Option` field.

See #4856 for details


weiznich:
Thanks for the quick fix :heart:

Would it be possible to add some test here as well: https://github.com/diesel-rs/diesel/blob/main/diesel_derives/tests/as_changeset.rs

(Essentially just copy your example in there to make sure that it compiles now)

Issue #4840, opened 2025-10-28
Improve the documentation of our derives


weiznich:
https://github.com/diesel-rs/diesel/pull/4714 introduced expanded code examples for the derives provided by diesel CLI in our API documentation. See for example [here](https://docs.diesel.rs/main/diesel/prelude/derive.AsChangeset.html#input)

We would like to extend this documentation to include examples for all derive attributes supported by the various derives.

Steps to implement this:

1. Go to `diesel_derives/src/tests/$derive.rs` and add a new test example there as new test case. Eg. mirroring this test case: https://github.com/diesel-rs/diesel/blob/7a43396582924220cc45a12b2a7a1dd076150021/diesel_derives/src/tests/as_changeset.rs#L6-L19
 Make sure that the example includes the relevant attribute
2. Run `cargo xtask run-tests all --no-doc-tests --no-example-schema-check -- -E 'package(diesel_derives) & test(#tests::*)'` and accept the newly added snapshots via `cargo insta`
3. Verify that the output is the same for all backends. If this is not the case make sure to add different variants like done here https://github.com/diesel-rs/diesel/blob/7a43396582924220cc45a12b2a7a1dd076150021/diesel_derives/src/tests/has_query.rs#L13-L27
4. Modify the build script here to include the new snapshot in the documentation: https://github.com/diesel-rs/diesel/blob/main/diesel_derives/build.rs#L116
5. Verify via `RUSTFLAGS="--cfg diesel_docsrs" RUSTDOCFLAGS="--cfg diesel_docsrs" cargo +nightly doc -p diesel --open` + opening the documentation the derive you are working on that the code is shown there

We still miss the following attributes:

# `#[derive(AsChangeset)]`

## Struct 
- [x] `#[diesel(treat_none_as_null = true)]`
- [x] `#[diesel(table_name = path::to::table)]]`
- [x] `#[diesel(primary_key(id1, id2))]]

## Field
- [x] `#[diesel(column_name = some_column_name)]`
- [x] `#[diesel(embed)]`
- [x] `#[diesel(serialize_as = SomeType)]`
- [x] `#[diesel(treat_none_as_null = true/false)]`
- [x] `#[diesel(skip_update)]`

# `#[derive(AsExpression)]`

## Type

- [ ] `#[diesel(not_sized)]`

# `#[derive(Associations)]`

## Type

- [x] `#[diesel(table_name = path::to::table)]`

## Field

[...]

PR #4861, opened 2025-11-12
improve documentation of changeset derive

valkrypton:
Working on [#4840](https://github.com/diesel-rs/diesel/issues/4840). I have added the new test for the attribute but it's failing. @weiznich any idea why is that?

### DIFF
--- a/diesel_derives/src/tests/as_changeset.rs
+++ b/diesel_derives/src/tests/as_changeset.rs
@@ -34,3 +34,38 @@ pub(crate) fn as_changeset_treat_none_as_null_1() {
         "as_changeset_treat_none_as_null_1",
     );
 }
+
+#[test]
+pub(crate) fn as_changeset_table_name_1() {
+    let input = quote::quote! {
+        #[diesel(table_name = crate::schema::admin_users)]
+        struct User {
+            id: i32,
+            name: String
+        }
+    };
+    expand_with(
+        &crate::derive_as_changeset_inner as &dyn Fn(_) -> _,
+        input,
+        derive(syn::parse_quote!(#[derive(AsChangeset)])),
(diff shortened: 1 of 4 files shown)


weiznich:
Hi there. Thanks for working on this. I just noticed that the build.rs structure made it quite hard to add new examples, so I restructured it a bit to make this easier in the future. It should now work.

valkrypton:
@weiznich you can review this now

diesel_derives/src/tests/mod.rs
```diff
@@ -96,6 +96,7 @@ fn expand_with<Fn: ProcMacroFn>(
     let file = syn::parse2(out).unwrap();
 
     let out = prettyplease::unparse(&file);
+    println!("snapshot\n{out}");
```
weiznich:
```suggestion
```

That doesn't need to be here

diesel_derives/src/tests/as_changeset.rs
```diff
@@ -17,3 +17,55 @@ pub(crate) fn as_changeset_1() {
...
+        input,
+        derive(syn::parse_quote!(#[derive(AsChangeset)])),
+        "as_changeset_treat_none_as_null_1",
+    );
+}
+
+#[test]
+pub(crate) fn as_changeset_table_name_1() {
+    let input = quote::quote! {
+        #[diesel(table_name = crate::schema::users)]
```
weiznich:
```suggestion
        #[diesel(table_name = crate::schema::some_other_user_table)]
```

Could we use a different table name than the "default" `users` here to demonstrate what's changed by this option?

(That requires updating the snapshots)

weiznich (approved):
Perfect, thanks :heart:

PR #4872, opened 2025-11-17
implements json(b)_extract_path_text

valkrypton:
implement `json(b)_extract_path_text` from #4216


diesel/src/pg/expression/functions.rs
```diff
@@ -3127,6 +3127,126 @@ extern "SQL" {
...
+    ///
+    /// #     Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "postgres_backend")]
+    #[sql_name = "json_extract_path_text"]
+    #[variadic(last_arguments = 1, skip_zero_argument_variant = true)]
+    fn json_extract_path_text<J: JsonOrNullableJson + SingleValue, T: SingleValue>(
+        json: J,
+        text: T,
```
weiznich: This needs to be 
```suggestion
    fn json_extract_path_text<J: JsonOrNullableJson + SingleValue, T: TextOrNullableText>(
        json: J,
        text: T,
```

as the second argument needs to be an expression of the type `Text`. Other expression types are not supported there.

(This applies to the jsonb variant as well)

PR #4879, opened 2025-11-19
Support casting between cidr, inet and text for postgres backend

stormshield-kg:
This PR adds casting support for the `Cidr` and `Inet` type in PostgreSQL. Conversions to `Text` are infallible, conversions from `Text` is fallible and conversion from `Inet` to `Cidr` will truncate the inet value as specified in https://www.postgresql.org/docs/current/functions-net.html.

### DIFF
--- a/diesel/src/expression/cast.rs
+++ b/diesel/src/expression/cast.rs
@@ -125,6 +125,8 @@ type_name! {
         Uuid => "uuid",
         Json => "json",
         Jsonb => "jsonb",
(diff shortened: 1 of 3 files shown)


weiznich:
Thanks for opening this PR.

Would it be possible to add tests for these variants here:

https://github.com/diesel-rs/diesel/blob/main/diesel_tests/tests/cast.rs

weiznich (approved):
Thanks

PR #4880, opened 2025-11-19
improve docs of changeset derives

valkrypton:
Added more documentation for `#[AsChangeset]` derive described in #4840

### DIFF
--- a/diesel_derives/src/tests/as_changeset.rs
+++ b/diesel_derives/src/tests/as_changeset.rs
@@ -69,3 +69,90 @@ pub(crate) fn as_changeset_primary_key_1() {
         "as_changeset_primary_key_1",
     );
 }
+
+#[test]
+pub(crate) fn as_changeset_column_name_1() {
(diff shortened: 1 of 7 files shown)


diesel_derives/src/tests/as_changeset.rs
```diff
@@ -69,3 +69,89 @@ pub(crate) fn as_changeset_primary_key_1() {
...
+        derive(syn::parse_quote!(#[derive(AsChangeset)])),
+        "as_changeset_embed_1",
+    );
+}
+
+#[test]
+pub(crate) fn as_changeset_change_field_to_type_1() {
+    let input = quote::quote! {
+        struct User {
+            #[serialize_as = String]
```
weiznich: As the `id` field will be skipped, this needs to be moved to a different field, otherwise we won't see any change in the output.

Also this needs to be `#[diesel(serialize_as = String)]`.

PR #4887, opened 2025-11-28
Now `sqlite` feature is enabled when `returning_clauses_for_sqlite_3_35` is enabled

LucaCappelletti94:
As per title, now when the `returning_clauses_for_sqlite_3_35` feature is enabled, the `sqlite` feature is enabled in cascade. It would never make sense for `returning_clauses_for_sqlite_3_35` to be enabled when `sqlite` is not enabled.

I have simplified all feature gates of the form like:

```rust
#[all(feature = "sqlite", feature = "returning_clauses_for_sqlite_3_35")]
```

To

```rust
#[feature = "returning_clauses_for_sqlite_3_35"]
```


weiznich (approved):
Thanks

PR #4897, opened 2025-12-07
Bump `sqlite-wasm-rs` to 0.5

Spxg:
Now `sqlite-wasm-rs` will only be compiled with bundled, and the prebuild lib can be utilized by overriding the cargo links configuration.

Additionally, it will like other sys crates without excessive dependencies, as the vfs implementation has been moved to `sqlite-wasm-vfs`. Thanks to this change, can now be compiled with `#![no_std]`.

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -53,13 +53,12 @@ path = "../diesel_derives"
 libsqlite3-sys = { version = ">=0.17.2, <0.36.0", optional = true, features = ["bundled_bindings"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
-sqlite-wasm-rs = { version = "0.4.0", optional = true, default-features = false }
+sqlite-wasm-rs = { version = "0.5.0", optional = true, default-features = false }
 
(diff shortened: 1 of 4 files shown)


weiznich (approved):
Thanks for the update. This looks good to me :tada:

As semi-related note: I wonder whether it would be possible to extract the mem-vfs implementation as well from `sqlite-wasm-rs` and make it "platform-independent". The reasons why I'm asking this is that I'm playing around with diesel+sqlite in a no-std environment. It mostly works but you need to bring your own vfs there. Having some sort of default vfs as crate would be really helpful. For now I just have a local copy of your implementation + some adjustments. I'm happy to submit a PR to sqlite-wasm-rs to extract the code out in a seperate crate + with some platform independent interface. It likely needs something for the creator of the VFS to supply datetime + randomness implementation, so that this can be adjusted to the corresponding environment.

Spxg:
> I'm happy to submit a PR to sqlite-wasm-rs to extract the code out in a seperate crate + with some platform independent interface. 

Welcome! 

> I'm playing around with diesel+sqlite in a no-std environment. 

Amazing work! I didn't expect diesel can be used in a no_std environment.

PR #4900, opened 2025-12-09
Added implementation of `HasTable` for `Option`

LucaCappelletti94:
As described in discussion #4899


LucaCappelletti94:
The failures seem to be coming from the current main

weiznich:
The build failure is releated to a recent rust regression (https://github.com/rust-lang/rust/issues/149845). It needs to be fixed there.

weiznich (approved):

PR #4893, opened 2025-12-05
Another mariadb fix

weiznich:
This commit fixes another issue with libmariadb. In this case again the libmariadb documentation/behaviour is inconsistent. They stated that with the last version they won't set the lenght field for statically sized types. Turns out that even that is not correct, as it is set for the datetime type. Now this on it's own wouldn't be an issue, but libmariadb has a slightly different definition of `MYSQL_TIME` that libmysqlclient. It misses a field in the end. While this is normally not a problem as that field is just zero initialized by diesel, it becomes problematic if libmariadb returns the size of their definition, which results in the assert mentioned in the issue to fail. So it's yet another needless incompatiblity on their side we need to workaround.

Fixes #4889

cc @BlackDex It would be great to know if the fix works for you. As already written in #4889 we should find a better solution for testing and maintaining support for libmariadb in the future. Possibly it would already help if someone sets up at least one CI job that runs tests with libmariadb + mariadb instead of libmysqlclient + mysql and also if someone at least helps looking into issues that come up. To be upfront here: If the situation around the mysql/mariadb backend doesn't change much in the next year or so, I consider issuing a diesel 3.0 release that moves this backend out of tree and marks it as unmaintained by me.

PR #4904, opened 2025-12-12
Prepare 2.3.5

weiznich:
This prepares another diesel 2.3 patch release including the following changes:

* #4897 
* #4896 
* #4893 
* #4887 
* #4884 
* #4907
* #4911

@diesel-rs/core I plan to issue an release next week if there are no concerns


Spxg:
Could you add this PR https://github.com/diesel-rs/diesel/pull/4911 to the list? I think it's a meaningful change.

PR #4906, opened 2025-12-12
Add no_std support to diesel

weiznich:
This commit adds support for using diesel with the SQLite backend
 in a no-std (core + alloc) environment. 

This mainly means fixing a lot of imports to use the relevant items from core or alloc instead. Notable exceptions are:

* There is no hashmap in alloc, so we need to pull in hashbrown for this
* There is no panicing/abort support in core so we need our custom shims for these.
* Global instrumentation setting is not supported yet (cfged away) as it requires lock which are not supported on embedded targets.

This commit also introduces a new `std` default feature for enabling `std` lib support. If disabled the users need to manually enable the `hashbrown` feature to pull in hashbrown for hashmap support. Cargo doesn't support pulling in dependencies for disabled features.

Finally this commit also adds a `example/sqlite/embedded` example project that demonstrates how to use this on a esp32-c6.

This is marked as a draft for now as there are a few upstream things that I would like to resolve properly before merging this:

* [x] Get a new release of libsqlite3-sys with no_std support
* [x] Get a standalone release of the memfs implementation from sqlite-wasm-rs so that we don't need to vendor their implementation here.
* [x] (Resolve the merge conflicts)

### DIFF
--- a/diesel_derives/src/sql_function.rs
+++ b/diesel_derives/src/sql_function.rs
@@ -476,7 +476,7 @@ fn expand_nonvariadic(
         #numeric_derive
         pub struct #fn_name #ty_generics {
             #(pub(in super) #args_iter,)*
-            #(pub(in super) #type_args: ::std::marker::PhantomData<#type_args>,)*
+            #(pub(in super) #type_args: ::core::marker::PhantomData<#type_args>,)*
         }
 
         #[doc = #helper_type_doc]
@@ -674,7 +674,7 @@ fn expand_nonvariadic(
         {
             #internals_module_name::#fn_name {
                 #(#arg_struct_assign,)*
-                #(#type_args: ::std::marker::PhantomData,)*
+                #(#type_args: ::core::marker::PhantomData,)*
             }
         }
 
@@ -765,7 +765,7 @@ fn generate_tokens_for_non_aggregate_functions(
                 f: F,
             ) -> QueryResult<()>
             where
-                F: Fn(#(#arg_name,)*) -> Ret + std::panic::UnwindSafe + Send + 'static,
+                F: Fn(#(#arg_name,)*) -> Ret + ::core::panic::UnwindSafe + Send + 'static,
                 (#(#arg_name,)*): FromSqlRow<(#(#arg_type,)*), Sqlite> +
                     StaticallySizedRow<(#(#arg_type,)*), Sqlite>,
                 Ret: ToSql<#return_type, Sqlite>,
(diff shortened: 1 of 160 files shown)


weiznich:
@LucaCappelletti94 You ask where you could help. This PR could use a review from someone that's not me :wink:

diesel_derives/src/sql_function.rs
```diff
@@ -765,7 +765,7 @@ fn generate_tokens_for_non_aggregate_functions(
                 f: F,
             ) -> QueryResult<()>
             where
-                F: Fn(#(#arg_name,)*) -> Ret + std::panic::UnwindSafe + Send + 'static,
+                F: Fn(#(#arg_name,)*) -> Ret + core::panic::UnwindSafe + Send + 'static,
```
LucaCappelletti94: For consistency this should be `::core` I believe

diesel_derives/src/sql_function.rs
```diff
@@ -791,7 +791,7 @@ fn generate_tokens_for_non_aggregate_functions(
                 mut f: F,
             ) -> QueryResult<()>
             where
-                F: FnMut(#(#arg_name,)*) -> Ret + std::panic::UnwindSafe + Send + 'static,
+                F: FnMut(#(#arg_name,)*) -> Ret + core::panic::UnwindSafe + Send + 'static,
```
LucaCappelletti94: For consistency this should be `::core` I believe

diesel_derives/src/sql_function.rs
```diff
@@ -825,7 +825,7 @@ fn generate_tokens_for_non_aggregate_functions(
                 f: F,
             ) -> QueryResult<()>
             where
-                F: Fn() -> Ret + std::panic::UnwindSafe + Send + 'static,
+                F: Fn() -> Ret + core::panic::UnwindSafe + Send + 'static,
```
LucaCappelletti94: For consistency this should be `::core` I believe

diesel_derives/src/sql_function.rs
```diff
@@ -849,7 +849,7 @@ fn generate_tokens_for_non_aggregate_functions(
                 mut f: F,
             ) -> QueryResult<()>
             where
-                F: FnMut() -> Ret + std::panic::UnwindSafe + Send + 'static,
+                F: FnMut() -> Ret + core::panic::UnwindSafe + Send + 'static,
```
LucaCappelletti94: For consistency this should be `::core` I believe

diesel_derives/src/sql_function.rs
```diff
@@ -949,12 +949,12 @@ fn generate_tokens_for_aggregate_functions(
                         A: SqliteAggregateFunction<#arg_name>
                             + Send
                             + 'static
-                            + std::panic::UnwindSafe
-                            + std::panic::RefUnwindSafe,
+                            + core::panic::UnwindSafe
+                            + core::panic::RefUnwindSafe,
```
LucaCappelletti94: For consistency these should be `::core` I believe

LucaCappelletti94 (changes requested):
Just a few minor `core` -> `::core` changes needed for consistency

PR #4907, opened 2025-12-14
Fix issues with ambiguous glob reexports

petrochenkov:
Fixes https://github.com/rust-lang/rust/issues/149845.

### DIFF
--- a/diesel/src/reexport_ambiguities.rs
+++ b/diesel/src/reexport_ambiguities.rs
@@ -1,144 +1,154 @@
+// This macro exists to workaround
+// https://github.com/rust-lang/rust/issues/149845
+//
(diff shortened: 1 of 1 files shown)


petrochenkov:
The `Compiletests` failures look unrelated to these changes, they appear in other CI runs too.

weiznich:
Thanks for submitting this PR. That's really helpful for fixing the actual issue in diesel :heart:

I pushed another commit to bump the toolchain version + update the compile test output to fix the failing compile tests.

Otherwise I made some minor adjustments to the fix to add some comments + make the structure a bit clearer to future contributors.

This is now ready to be merged.

weiznich (approved):

PR #4908, opened 2025-12-15
Add support for Postgresql array slicing

stormshield-kg:
This PR adds support for `col[1:2]`, `col[1:]` and `col[:2]` syntaxes in expressions and in update statements.


diesel/src/pg/expression/helper_types.rs
```diff
@@ -261,6 +261,19 @@ pub type IsContainedByJsonb<Lhs, Rhs> = IsContainedBy<Lhs, Rhs>;
...
+    super::operators::ArraySlice<Lhs, AsExprOf<Rhs1, Integer>, AsExprOf<Rhs2, Integer>>;
+
+/// The return type of [`lhs.slice_from(rhs)`](super::expression_methods::PgArrayExpressionMethods::slice_from)
+#[cfg(feature = "postgres_backend")]
+pub type SliceFrom<Lhs, Rhs> = super::operators::ArraySliceFrom<Lhs, AsExprOf<Rhs, Integer>>;
+
+/// The return type of [`lhs.slice_to(rhs)`](super::expression_methods::PgArrayExpressionMethods::slice_to)
+#[cfg(feature = "postgres_backend")]
+pub type SliceTo<Lhs, Rhs> = super::operators::ArraySliceTo<Lhs, AsExprOf<Rhs, Integer>>;
+
```
weiznich, line 295: We should have a test that these functions/defs work with `#[auto_type]` in https://github.com/diesel-rs/diesel/blob/main/diesel_derives/tests/auto_type.rs

diesel/src/pg/expression/operators.rs
```diff
@@ -175,6 +168,244 @@ where
...
+
+impl<L, R1, R2, ST> Expression for ArraySlice<L, R1, R2>
+where
+    L: Expression<SqlType = Array<ST>>,
+    R1: Expression<SqlType = Integer>,
+    R2: Expression<SqlType = Integer>,
+    ST: SqlType + TypedExpressionType,
+{
+    type SqlType = Array<ST>;
+}
```
weiznich, line 213: While this definition is technically correct I expect it to be quite restrictive for in practice. 

It disallows using a nullable value as array or as index.

To cover everything postgres allows we want to accept possible nullable values for any of these positions. The return type must then be nullable if any of the arguments are nullable.



```suggestion
impl<L, R1, R2, ST> Expression for ArraySlice<L, R1, R2>
where
impl<L, R1, R2, ST> Expression for ArraySlice<L, R1, R2>
where
    L: Expression, L::SqlType: ArrayOrNullableArray<Inner = ST>,
    R1: Expression, R1::SqlType: IntegerOrNullableInteger,
    R2: Expression, R2::SqlType: IntegerOrNullableInteger,
    <L::SqlType as SqlType>::IsNull: OneIsNullable<<R1::SqlType as SqlType>::IsNull>,
    <<L::SqlType as SqlType>::IsNull as OneIsNullable<<R1::SqlType as SqlType>::IsNull>>::Out: OneIsNullable<<R1::SqlType as SqlType>::IsNull>,
    <<<L::SqlType as SqlType>::IsNull as OneIsNullable<<R1::SqlType as SqlType>::IsNull>>::Out as OneIsNullable<<R1::SqlType as SqlType>::IsNull>>::Out: MaybeNullableType<Array<ST>>,
    ST: SqlType + TypedExpressionType,
{
    type SqlType = <<<<L::SqlType as SqlType>::IsNull as OneIsNullable<<R1::SqlType as SqlType>::IsNull>>::Out as OneIsNullable<<R1::SqlType as SqlType>::IsNull>>::Out as MaybeNullableType<Array<ST>>>::Out;
}
```

I didn't check if all of the `STOrNullableST` helper traits already exist. Possibly you need to add some of them here: https://github.com/diesel-rs/diesel/blob/main/diesel/src/pg/expression/expression_methods.rs#L3720

(This applies ot the other variants as well, although it should be simpler there as we only have two arguments, not three)

weiznich (changes requested):
Thanks for opening this PR. I think that's a really meaningful feature, so thanks for working on it. 

I would like to have a test for `#[auto_type]` to make sure it works there. Additionally I would like to refine the `Expression` impl for the underlying types to more correctly handle null values.

stormshield-kg:
Notes:
* To support null values I had to make separate `*_nullable()` methods: I cannot define an encompassing trait due the generic type in `AsExpression<T>`, since a type may implement both `AsExpression<Integer>` and `AsExpression<Nullable<Integer>>`.
* I added `impl Column for AssumeNotNull<T>` to allow using `slice_from()` and `slice_to()` with nullable array columns in an update statement.

weiznich (approved):
First of all sorry for taking so long to come back to this. Holidays and other stuff came in the way :(

Yes you are right. It doesn't work like I expected, sorry for that
So that's the best solution. Thanks for implementing it.

I've just pushed a minor cleanup commit, otherwise this is ready to be merged.

PR #4911, opened 2025-12-17
Bump `sqlite-wasm-rs` to 0.5.1

Spxg:
Removed emcc requirement, emscripten toolchain no longer needs to be installed.

The previous PR https://github.com/diesel-rs/diesel/pull/4897 has not yet been released, so let's bump the version

### DIFF
--- a/diesel/Cargo.toml
+++ b/diesel/Cargo.toml
@@ -53,7 +53,7 @@ path = "../diesel_derives"
 libsqlite3-sys = { version = ">=0.17.2, <0.36.0", optional = true, features = ["bundled_bindings"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
-sqlite-wasm-rs = { version = "0.5.1", optional = true, default-features = false }
+sqlite-wasm-rs = { version = ">=0.4.0, <0.6.0", optional = true, default-features = false }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dev-dependencies]
 # Something is dependent on it, we use feature to override it.


diesel/Cargo.toml
```diff
@@ -53,7 +53,7 @@ path = "../diesel_derives"
 libsqlite3-sys = { version = ">=0.17.2, <0.36.0", optional = true, features = ["bundled_bindings"] }
 
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
-sqlite-wasm-rs = { version = "0.5.0", optional = true, default-features = false }
+sqlite-wasm-rs = { version = "0.5.1", optional = true, default-features = false }
```
weiznich: Given that `sqlite-wasm-rs` is a sys dependency we need to stay compatible with the version that was released with diesel 2.3.

So this needs to be


```suggestion
sqlite-wasm-rs = { version = ">=0.4.0, <0.6.0", optional = true, default-features = false }
```

For the other locations it's fine to pin it to the new version

weiznich (approved):

weiznich:
BTW: It's great to see the emscripten dependency gone :tada:

One thing I wondered: As far as I understand you need to compile the musl libc thing there as sqlite requires some libc functions, right? For my embedded demo-case I've just used [tinyrlibc](https://lib.rs/crates/tinyrlibc) for that, which implements all minimally necessary functions in mostly rust. It should possibly be a better solution?

Spxg:
> It should possibly be a better solution?

I think it' fine. There are reasons why I use musl C symbols are as follows:

* Symbols implemented in Rust are usually exposed in the final product because of 
```rust
#[no_mangle]
pub extern "C" fn xx()
```
* I'm not entirely confident about the correctness of Rust symbol implementations, so using musl is the best option for me.
* I need to rename those symbols, as symbol conflicts are prone to occur on platforms like wasm, which lacks libc.

PR #4914, opened 2025-12-19
Add DSL traits for generic Upsert operations

LucaCappelletti94:
This PR introduces a set of new traits to `diesel::query_dsl` that mirror existing inherent methods for `ON CONFLICT` and `UPDATE` clauses. These traits enable users to write generic code that abstracts over upsert and update operations, which was previously impossible (to my knowledge) because the methods were only defined as inherent implementations on specific structs with private trait bounds.

## Changes
The following traits have been added and exported via `diesel::query_dsl`:

*   **`OnConflictDsl<Target>`**: Mirrors `InsertStatement::on_conflict`.
*   **`OnConflictDoNothingDsl`**: Mirrors `InsertStatement::on_conflict_do_nothing`.
*   **`DoNothingDsl`**: Mirrors `IncompleteOnConflict::do_nothing`.
*   **`DoUpdateDsl`**: Mirrors `IncompleteOnConflict::do_update`.
*   **`SetUpdateDsl<V>`**: Mirrors `UpdateStatement::set` and `IncompleteDoUpdate::set`.

Implementations have been added for:
*   `InsertStatement`
*   `IncompleteOnConflict`
*   `IncompleteDoUpdate`
*   `UpdateStatement`

These implementations simply delegate to the existing inherent methods, ensuring behavior remains identical while allowing the methods to be used in trait bounds.

## Motivation
Currently, methods like `.on_conflict()`, `.do_update()`, and `.set()` are inherent methods. This makes it difficult to write generic traits or functions that perform upserts across different tables or models, as Rust requires trait bounds to access methods in generic contexts.

By introducing these traits, developers can define generic abstractions for common patterns like "Upsert".

### Example Use Case
With these changes, a [user can define a generic `ModelUpsert` trait that works for any table supporting the necessary operations](https://github.com/LucaCappelletti94/diesel-builders/blob/e6e6e632c75f6237664e520ba85f55c3c863113d/diesel-builders/src/ancestors.rs#L236-L302) and, in my use case, [employ it as such](https://github.com/LucaCappelletti94/diesel-builders/blob/e6e6e632c75f6237664e520ba85f55c3c863113d/diesel-builders/tests/test_base_case.rs#L87-L92).

Abstracted a bit, the impl looks like:

```rust
pub trait ModelUpsert<Conn>: HasTable {
    fn upsert(&self, conn: &mut Conn) -> QueryResult<Self>
    where
        Self: Sized;
}

impl<Conn, M> ModelUpsert<Conn> for M
where
    M: HasTable,
    Conn: LoadConnection,
    // The new traits allow us to express these bounds:
    for<'query> InsertStatement<...>: OnConflictDsl<
        <M::Table as Table>::PrimaryKey,
        Output: DoUpdateDsl<
            Output: SetUpdateDsl<..., Output: LoadQuery<'query, Conn, ...>>
        >
    >,
{
    fn upsert(&self, conn: &mut Conn) -> QueryResult<Self> {
        // We can now call these methods generically:
        diesel::insert_into(table)
            .values(...)
            .on_conflict(table.primary_key())
            .do_update()
            .set(...)
            .get_results(conn)
    }
}
```

## Backward Compatibility
This change is purely additive. The existing inherent methods remain in place (and are used by the trait implementations), so no existing code should break. The new traits are exported in `diesel::query_dsl` for users who opt-in to using them.


weiznich:
I just want to leave a note here that I run out of time today for diesel related work. I will try to have a look at this next year if noone else from @diesel-rs/reviewers reviews this in between.

weiznich (commented):
Instead of duplicating all these methods I would rather just add the relevant types as type defs. That should allow you to write your bounds and it should also allow using this methods with `#[auto_type]`.  For the second use case we want to add some test cases in diesel_derives/tests/auto_type.rs.

LucaCappelletti94:
I am not sure I understand how adding type defs would help remove the newly introduced traits. For all type defs, such as the existing [`ThenOrderBy`](https://docs.rs/diesel/latest/diesel/dsl/type.ThenOrderBy.html), there is an accompanying trait (in this example [`ThenOrderDsl`](https://docs.rs/diesel/latest/diesel/query_dsl/methods/trait.ThenOrderDsl.html)) with an associated type [`Output`](https://docs.rs/diesel/latest/diesel/query_dsl/methods/trait.ThenOrderDsl.html#associatedtype.Output) and a method [`then_order_by`](https://docs.rs/diesel/latest/diesel/query_dsl/methods/trait.ThenOrderDsl.html#tymethod.then_order_by) which is defined as returning the associated type.

I can add the requested type defs, but I fail to understand how this could allow me to remove the trait or even only the method in the trait. I would still need the where constraint to express "this generic object `T` does have the method `on_conflict`".

Suppose I were to create a type `OnConflict` without using an accompanying trait:

```rust
pub type OnConflict<T, U: IntoConflictValueClause, Op, Ret, Target> = IncompleteOnConflict<InsertStatement<T, U::ValueClause, Op, Ret>, ConflictTarget<Target>>;
```

I would get the following errors:

```rust
trait `crate::query_source::impls_which_are_only_here_to_improve_error_messages::upsert::into_conflict_clause::IntoConflictValueClause` exists but is inaccessible
```

```rust
struct `crate::query_source::impls_which_are_only_here_to_improve_error_messages::upsert::on_conflict_target::ConflictTarget` exists but is inaccessible
```

which are the reason for which I created the trait.

As code duplication API/breaking is a concern, we could add a deprecation warning on the original struct methods and when a new major version comes along we could remove them entirely.

weiznich:
Seems like the only relevant trait that's currently missing from the public API is `OnConflictTarget`: https://docs.diesel.rs/2.3.x/src/diesel/query_builder/upsert/on_conflict_target.rs.html#7 via https://docs.diesel.rs/2.3.x/diesel/query_builder/struct.InsertStatement.html#method.on_conflict. Anything else are private types as far as I can see. 

It should be fine to expose `OnConflictTarget` as part of the public API if the trait is sealed so that downstream crates can use it as bound, but not implement it. That combined with the type defs proposed above should allow you to write the relevant top level bounds to express whatever is needed to use upsert statements in generic code.

LucaCappelletti94:
I certainly agree that these traits should not be implementable outside of diesel. Is there already in rust some "standard" sealing super-trait I should reuse?

LucaCappelletti94:
While the [`on_conflict`](https://docs.diesel.rs/2.3.x/diesel/query_builder/struct.InsertStatement.html#method.on_conflict) method itself does have as where clause bound solely [`OnConflictTarget`](https://docs.diesel.rs/2.3.x/src/diesel/query_builder/upsert/on_conflict_target.rs.html#7), the context in which that method is implemented [additionally requires `UndecoratedInsertRecord<T>` and `IntoConflictValueClause`](https://docs.diesel.rs/2.3.x/src/diesel/upsert/on_conflict_extension.rs.html#17). Therefore, to fully define that a given `InsertStatement<T, U, Op, Ret>` has the `on_conflict` method, it needs to be possible to specify [`U: UndecoratedInsertRecord<T> + IntoConflictValueClause,`](https://docs.diesel.rs/2.3.x/src/diesel/upsert/on_conflict_extension.rs.html#17).

Since those are private traits, I still see no other solution than the `OnConflictDsl` trait, which should likely be turned into a sealed trait as well.

weiznich:
I think it's fine to expose these two traits as well as sealed traits. For `IntoConflictValueClause` all the items should be `#[doc(hidden)]` as they shouldn't be part of the public API.

LucaCappelletti94:
I am not sure from where this error is coming from - the change I have just made should have really not caused such a logical change:

```rust
  stdout ───

    running 1 test
    test limit_offset::limit_offset ... FAILED

    failures:

    failures:
        limit_offset::limit_offset

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 571 filtered out; finished in 0.02s
    
  stderr ───

    thread 'limit_offset::limit_offset' (26796) panicked at diesel_tests/tests/limit_offset.rs:50:5:
    assertion `left == right` failed
      left: [("Ruby", None)]
     right: [("Tess", None)]
```

diesel/src/query_builder/update_statement/mod.rs
```diff
@@ -47,6 +47,19 @@ impl<T: QuerySource, U> UpdateStatement<T, U, SetNotCalled> {
...
+    V: changeset::AsChangeset<Target = T>,
+    UpdateStatement<T, U, V::Changeset>: AsQuery,
+{
+    type Output = UpdateStatement<T, U, V::Changeset>;
+
+    fn set(self, values: V) -> Self::Output {
+        self.set(values)
+    }
+}
+
```
weiznich: I think these additional DSL traits can now be removed?

diesel/src/query_dsl/on_conflict_dsl.rs
weiznich, line 1: These as well?

diesel/src/query_dsl/update_dsl.rs
```diff
@@ -0,0 +1,8 @@
+/// The `set` method
```
weiznich: A that one as well

weiznich (changes requested):

LucaCappelletti94:
Yes, I will remove those traits.

I am not sure how to make public `UndecoratedInsertRecord`, since it is currently decorated with:

```rust
#[cfg_attr(
    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes",
    cfg(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes")
)]
pub trait UndecoratedInsertRecord<Table> {}
```

Should I remove that decorator? Maybe depending on the presence/absence of that flag I make the trait sealed or not?

weiznich:
> Maybe depending on the presence/absence of that flag I make the trait sealed or not?

That's the correct way forward.

LucaCappelletti94:
I think I misunderstood something about this syntax:

```rust
#[cfg_attr(
    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes",
    cfg(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes")
)]
pub trait UndecoratedInsertRecord<Table> {}
```

That feature gating has solely the effect to add a feature gate in the doc? As I believe it reads: "when the feature "i-implement-a-third-party-backend-and-opt-into-breaking-changes" is enabled, then add the tautological attribute that checks whether the same feature "i-implement-a-third-party-backend-and-opt-into-breaking-changes" is enabled to include the following code.

If so, there is no need to seal this trait.

diesel/src/query_builder/update_statement/mod.rs
LucaCappelletti94: Done

diesel/src/query_dsl/on_conflict_dsl.rs
LucaCappelletti94: Done

diesel/src/query_dsl/update_dsl.rs
LucaCappelletti94: Done

LucaCappelletti94:
I will now test these edits in [`diesel-builders`](https://github.com/LucaCappelletti94/diesel-builders/) to make sure that the necessary constraints can be expressed, and let you know whether all works nominally.

LucaCappelletti94:
Ok, the next pain point is `ConflictTarget`, which is a non exported public struct and therefore not reusable outside of diesel.

https://github.com/diesel-rs/diesel/blob/85ea4677b177de37588db8f3db9f18254f783216/diesel/src/query_builder/upsert/on_conflict_target.rs#L27-L28

As such, it is not possible to express the constraint relative to the [`on_conflict`](https://docs.diesel.rs/2.1.x/diesel/query_builder/struct.InsertStatement.html#method.on_conflict) method:

https://github.com/diesel-rs/diesel/blob/85ea4677b177de37588db8f3db9f18254f783216/diesel/src/upsert/on_conflict_extension.rs#L311-L313

Should I re-export it? Making a type def for it seems redundant, as to my understanding it would be semantically identical.

weiznich:
> Should I re-export it? Making a type def for it seems redundant, as to my understanding it would be semantically identical.

It's fine to reexport it if the field value is made private

LucaCappelletti94:
Should `ConflictTarget` be instantiate with a `From<T>` or should I make a `pub(crate) fn new()` method?

weiznich:
I would use a `pub(crate) fn new()` method, maybe with a `#[diesel_derives::__diesel_public_if(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes")]` added

LucaCappelletti94:
Ok, assuming that the above works (I will fix the tests), we now need to express that the following type (a bit redacted for brevity), where both `OnConflictValues` and `DoUpdate` are sealed public structs:

```rust
InsertStatement<
   my_table::table,
  OnConflictValues<
       [Values clause],
       diesel::query_builder::upsert::on_conflict_target::ConflictTarget<my_table::columns::id>,
       diesel::query_builder::upsert::on_conflict_actions::DoUpdate<[Assign clause]>
  >
>
```

Full unredacted type:

<details>
```rust
diesel::query_builder::insert_statement::InsertStatement<test_base_case::shared_animals::animals::table, diesel::query_builder::upsert::on_conflict_clause::OnConflictValues<diesel::query_builder::insert_statement::ValuesClause<(diesel::insertable::ColumnInsertValue<test_base_case::shared_animals::animals::columns::id, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>, diesel::insertable::ColumnInsertValue<test_base_case::shared_animals::animals::columns::name, diesel::expression::bound::Bound<diesel::sql_types::Text, alloc::string::String>>, diesel::insertable::ColumnInsertValue<test_base_case::shared_animals::animals::columns::description, diesel::expression::bound::Bound<diesel::sql_types::Nullable<diesel::sql_types::Text>, core::option::Option<alloc::string::String>>>), test_base_case::shared_animals::animals::table>, diesel::query_builder::upsert::on_conflict_target::ConflictTarget<test_base_case::shared_animals::animals::columns::id>, diesel::query_builder::upsert::on_conflict_actions::DoUpdate<(diesel::query_builder::update_statement::changeset::Assign<diesel::query_builder::update_statement::changeset::ColumnWrapperForUpdate<test_base_case::shared_animals::animals::columns::id>, diesel::expression::bound::Bound<diesel::sql_types::Integer, i32>>, diesel::query_builder::update_statement::changeset::Assign<diesel::query_builder::update_statement::changeset::ColumnWrapperForUpdate<test_base_case::shared_animals::animals::columns::name>, diesel::expression::bound::Bound<diesel::sql_types::Text, alloc::string::String>>, diesel::query_builder::update_statement::changeset::Assign<diesel::query_builder::update_statement::changeset::ColumnWrapperForUpdate<test_base_case::shared_animals::animals::columns::description>, diesel::expression::bound::Bound<diesel::sql_types::Nullable<diesel::sql_types::Text>, core::option::Option<alloc::string::String>>>), test_base_case::shared_animals::animals::table>>>
```
</details>

implements, for a certain generic `Conn` and `Model`:

```rust
for<'query> LoadQuery<
        'query,
        Conn,
        Model
    >
```

This constraint was originally expressed with:

```rust
InsertStatement<...>: for<'query>  OnConflictDsl<
        <M::Table as Table>::PrimaryKey,
        Output: DoUpdateDsl<
            Output: SetUpdateDsl<..., Output: LoadQuery<'query, Conn, ...>>
        >
    >,
```

Should I make those type accessible as well, or is there any other solution you would prefer?

LucaCappelletti94:
> Should I make those type accessible as well, or is there any other solution you would prefer?

@weiznich kindly do let me know what would be your preference on this matter when you have time, otherwise the PR will remain stuck there (I am pinging you just in case you had not seen the question in the previous comment).

weiznich:
As written before: I do not want to make these types public. Each of then is constructed by a chain of method calls. There should be just a chain of type defs mirroring that.

So for a hypothetical call chain like this:

```rust
diesel::insert_into(users)
    .values(&same_id_different_name)
    .on_conflict(id)
    .do_nothing()
    .execute(conn)?
```

we want to have the following type defs:

```
// already exists
pub type insert_into<T> = crate::query_builder::IncompleteInsertStatement<T>;
// already exists
pub type Values<I, U> = crate::query_builder::InsertStatement<
        <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Table,
        <U as crate::Insertable<
            <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Table,
        >>::Values,
        <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Op,
    >;
pub type OnConflict<I, C> = IncompleteOnConflict<I, ConflictTarget<C>>;
pub type DoNothing<I> = // some trait type mapping which doesn't need any new method, just associated types like for Values
```

for each of the new types we need to add a test to `diesel_derives/tests/auto_type.rs`.

Based on that you can also write your constraints like that:

```
DoNothing<OnConflict<Values<insert_into<users>, …>, id>: for<'query> LoadQuery<
        'query,
        Conn,
        Model
    > // or whatever
```

LucaCappelletti94:
I believe I may be fairly unfamiliar with this approach, so I will try to proceed with one type at the time and ask for your feedback bit by bit.

Starting with `OnConflict`, did you expect something like:

```rust
/// Represents the return type of
/// [`InsertStatement::on_conflict()`](crate::query_builder::InsertStatement::on_conflict)
pub type OnConflict<I, Target> = crate::upsert::IncompleteOnConflict<
    I,
    crate::query_builder::ConflictTarget<Target>
>;
```

Or like:

```rust
/// Represents the return type of
/// [`InsertStatement::on_conflict()`](crate::query_builder::InsertStatement::on_conflict)
pub type OnConflict<I, Target> = crate::upsert::IncompleteOnConflict<
    crate::query_builder::InsertStatement<
        <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Table,
        <<I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Values as crate::query_builder::IntoConflictValueClause>::ValueClause,
        <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Op,
        <I as crate::query_builder::insert_statement::InsertAutoTypeHelper>::Ret,
    >,
    crate::query_builder::ConflictTarget<Target>,
>;
```

Also, by:

> some trait type mapping which doesn't need any new method

You meant that I should add back a variant of `DoUpdateDsl` and `DoNothingDsl`, with no method duplicated but only an associated type, say, `Output`?

.gitignore
```diff
@@ -3,3 +3,4 @@ Cargo.lock
 !diesel_cli/Cargo.lock
 .env
 *.snap.new
+test.db
```
weiznich: That shouldn't be here in this PR

diesel_derives/tests/auto_type.rs
```diff
@@ -722,3 +722,10 @@ fn window_functions2() -> _ {
         nth_value(users::id, 1_i32).over(),
     )
 }
+
+#[auto_type]
+fn test_upsert_on_conflict() -> _ {
+    insert_into(users::table)
+        .values(users::id.eq(42_i32))
+        .on_conflict(users::id)
```
weiznich, line 739: If that compiles the type definition is correct. The other way to check it is to just use the type definition as return type of the relevant function. That would even be preferable because the documentation then at least links to the correct public type for that function instead of referring to private types like now.

diesel/src/query_builder/upsert/on_conflict_target.rs
```diff
@@ -24,7 +34,27 @@ impl<Table> OnConflictTarget<Table> for NoConflictTarget {}
...
+    ///
+    /// # Arguments
+    ///
+    /// * `target` - The target of the `ON CONFLICT` clause.
+    pub(crate) fn new(target: T) -> Self {
+        ConflictTarget(target)
+    }
+}
+
+impl<T> AsRef<T> for ConflictTarget<T> {
```
weiznich: This shouldn't be there. We don't want to expose the internal here

weiznich (commented):
>I believe I may be fairly unfamiliar with this approach, so I will try to proceed with one type at the time and ask for your feedback bit by bit.

You can relativly easy verify if the type definitions are correct by:

* Replace the return type of the function with the type definition
* Write the `#[auto_type]` test.

>
>Starting with OnConflict, did you expect something like:
> ```rust
> /// Represents the return type of
> /// [`InsertStatement::on_conflict()`](crate::query_builder::InsertStatement::on_conflict)
> pub type OnConflict<I, Target> = crate::upsert::IncompleteOnConflict<
>    I,
>    crate::query_builder::ConflictTarget<Target>
> >;
>```

I think that's the correct variant


>Also, by:
>
> >    some trait type mapping which doesn't need any new method
>
> You meant that I should add back a variant of DoUpdateDsl and DoNothingDsl, with no method duplicated but only an associated type, say, Output?

I mean to have a trait similar to `InsertAutoTypeHelper` for this usage. You essentially already used it in your second approach. That's required if the type definition internally needs to perform a non-trivial mapping. This trait doesn't need to be exposed it's just there to be able to write the right type definition.

LucaCappelletti94:
IDs in macos seem non deterministic btw, curious [825539a](https://github.com/diesel-rs/diesel/pull/4914/commits/825539a72928756e6e29f4634c8d4e203c7eabb4)

diesel_derives/tests/auto_type.rs
LucaCappelletti94: Ok, I have now used the new types as the return types of the methods, other than in the auto type tests

LucaCappelletti94:
I will try whether the current refactoring works in diesel-builders or there is still something limiting the generic use.

LucaCappelletti94:
Unfortunately no, we are still stuck at the `UndecoratedInsertRecord` trait. I must have misunderstood earlier on in https://github.com/diesel-rs/diesel/pull/4914#issuecomment-3756636495

I will try to turn it into a publicly accessible sealed trait as you said in https://github.com/diesel-rs/diesel/pull/4914#issuecomment-3754050425

LucaCappelletti94:
Ah yes, it was already publicly accessible via the relatively cursed `use diesel::internal::derives::insertable::UndecoratedInsertRecord;`, so it is not really sealable because it is actually used in the derive. That is why I hadn't made it sealed back then.

LucaCappelletti94:
Ok so, while still rather cursed, with it using `internal::derives::insertable::UndecoratedInsertRecord` and not being the most readable bound, but this now works and allows for generic upsert operations.

```rust
M: HasTable<Table: TableExt>
        + GetNestedColumns<<<M::Table as Table>::AllColumns as NestTuple>::Nested>,
    Conn: LoadConnection,
    <<M::Table as Table>::AllColumns as NestTuple>::Nested: TupleEqAll<
        EqAll: FlattenNestedTuple<Flattened: Insertable<M::Table> + AsChangeset<Target = M::Table>>,
    >,
    diesel::query_builder::ConflictTarget<<M::Table as Table>::PrimaryKey>:
        diesel::query_builder::OnConflictTarget<M::Table>,
    <FlattenedValuesOf<M> as Insertable<M::Table>>::Values:
        IntoConflictValueClause + UndecoratedInsertRecord<M::Table>,
    Set<
        DoUpdate<
            OnConflict<
                InsertStatement<
                    Self::Table,
                    <FlattenedValuesOf<M> as Insertable<Self::Table>>::Values,
                >,
                <M::Table as Table>::PrimaryKey,
            >,
        >,
        FlattenedValuesOf<M>,
    >: for<'query> LoadQuery<'query, Conn, <Self::Table as TableExt>::Model>,
```

weiznich (approved):
It's ready now :tada: Thanks again for your patience to work on this :heart:

I've just pushed a minor commit to make a last trait private and added another typedef for `on_conflict_do_nothing` so that all methods are now covered.

PR #4916, opened 2025-12-19
DRAFT PR: Feature batch update

be-mevsane:


diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +142,408 @@ where
...
+        write tests
+        write documentation
+*/
+
+// Copied from diesel/diesel/src/insertable.rs
+
+impl<'a, T> AsChangeset for &'a [T]
+where
+    T: AsChangeset,
+    // TODO: [UndecoratedUpdateRecord] could be used to enforce setting a where_clause maybe.
```
weiznich: I think the where clause needs to be generated automatically

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +142,408 @@ where
...
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        if let None = self.values.first() {
+            // TODO: more meaningful return maybe?
+            return Ok(());
+        }
+
+        // TODO: Create proper struct for alias that implements QueryFragment.
+        // Maybe there is already something?
+        const ALIAS: &str = "\"tmp\"";
```
weiznich: It's fine to use a hard coded name for this, but maybe something with less collision potential like `__diesel_internal_temp_values` or similar

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +142,408 @@ where
...
+impl<T, Tab, DB> QueryFragment<DB> for BatchChangeSet<T, Tab>
+where
+    DB: Backend,
+    Tab: Table,
+    T: BatchUpdateTargetAssign<DB, Tab> + BatchUpdateExpr<DB, Tab> + Clone,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        if let None = self.values.first() {
+            // TODO: more meaningful return maybe?
+            return Ok(());
```
weiznich: For the "normal" changesets we return an error if it's empty

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +142,408 @@ where
...
+        //     AS tmp(column_a, column_b, column_c)
+        if let Some(last) = self.values.last() {
+            out.push_sql(" AS ");
+            out.push_sql(ALIAS);
+            out.push_sql("(");
+            last.column_target(out.reborrow())?;
+            out.push_sql(")");
+        }
+
+        // id = primary key -> therefore not available in update batch! See Note below.
```
weiznich: We should get the primary key names via [`Table::primary_key()`](https://docs.diesel.rs/main/diesel/prelude/trait.Table.html#tymethod.primary_key)

We then also need to ensure that 
* we don't list these keys in the assign in the actual update (that is likely already the case?)
* the update struct contains primary key values. Requiring [`Identifiable`](https://docs.diesel.rs/main/diesel/associations/trait.Identifiable.html) on the struct should give us the right information there. We can list these values as first/last values in the `VALUES` clause

weiznich (commented):
Thanks for opening this draft PR and sorry for taking that long to response. There was just too much stuff going on in the last weeks to take the time to really look into this.

Hopefully the comments I made help to understand how to finish this.

Otherwise I think it might also be a good idea to:

* Add at least one compile test to show that this won't compile with Sqlite/MysqlConnection (it should already be the case, I just want to have a test for this for the future)
* Add a changelog entry as this is a relevant new feature

diesel/src/query_builder/update_statement/changeset.rs
be-mevsane: That would imply updates require the PrimaryKey value to be included in the changeset. Making this the default behavior sounds desirable indeed. Are you sure you want to have it restricted to only this usecase though?

I have personal interest to make this dynamic as it opens up more flexibility w.r.t. how devs can utilize batch updates in their projects business logic.

diesel/src/query_builder/update_statement/changeset.rs
be-mevsane: > we don't list these keys in the assign in the actual update (that is likely already the case?)

That's true, [AsChangeset](https://github.com/diesel-rs/diesel/blob/main/diesel_derives/src/as_changeset.rs#L48) already handles that

> the update struct contains primary key values. Requiring Identifiable on the struct should give us the right information there. We can list these values as first/last values in the VALUES clause

Nice suggestion. Identifiable provides the values for the primary key and Table::primary_key() privides the column name(s).

On first glance I was unable to find the QueryFragment-like implementations for Identifiable::Id and Table::PrimaryKey. I'll keep searching but hints are welcome if you already know by heart.

be-mevsane:
Hi again @weiznich ,
tldr:
- moved the actual batch update implementations to its correct file.
- Added Backend specific QueryFragment implementations for postgres and mysql.
- postgres works, mysql could work see notes.
- QueryFragment now differentiates between primary key columns and regular columns and both of their associated values.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,436 @@
...
+{
+    fn walk_ast<'b>(&'b self, mut _out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        // TODO:
+        // UpdateStatement::walk_ast(...) already added " SET " to the sql.
+        // MySql does not allow a temporary table definition or a join right
+        // after the keyword 'SET' and its assignments.
+        // Return Error by default due to missing support.
+        return Err(QueryBuilderError(Box::new(EmptyChangeset)));
+
+        // TODO:
```
be-mevsane: Hi @weiznich ,
Regular values are dependent on the " SET " keyword being pushed to the sql inside UpdateStatement::walk_ast(...) just before the values are converted to Changeset. Moving the responsibility to the values to add this fragment would resolve the issue. `out.push_sql(" SET ");` would then appear in [Assign](https://github.com/diesel-rs/diesel/blob/main/diesel/src/query_builder/update_statement/changeset.rs#L94) instead of [UpdateStatement](https://github.com/diesel-rs/diesel/blob/main/diesel/src/query_builder/update_statement/mod.rs#L206)
In regards to MySql support, this would allow it.

This did not seem to change anything in general behavior, but I am not sure about implications this might cause. Hence why I did not include it in the changes. I wanted your opinion on this first.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,436 @@
...
+//     /// Docs
+//     pub fn new(values: V) -> Self {
+//         Self {
+//             values,
+//             _marker: PhantomData,
+//         }
+//     }
+// }
+#[derive(Debug)]
+pub struct BatchUpdate<I, C, PK, Tab, QId, const STABLE_QUERY_ID: bool> {
```
be-mevsane: Similar to BatchInsert I've added `QId` and `STABLE_QUERY_ID` to BatchUpdate. I did not backtrack if that is actually required, but I just did that to stay consistent. One relevant place I found was in `multiconnection` derive.

Should we keep `QId` and `STABLE_QUERY_ID` ?

diesel/src/query_builder/update_statement/batch_update.rs
be-mevsane: Ahh yikes, some tests in `cargo xtask run-tests` fail due to this...

be-mevsane: Alternatively, moving `out.push_sql(" SET ");` from [UpdateStatement](https://github.com/diesel-rs/diesel/blob/main/diesel/src/query_builder/update_statement/mod.rs#L206) to its own struct that implements QueryFragment differently for single values or Vectory/Arrays might be another approach.

diesel/src/query_builder/update_statement/mod.rs
```diff
@@ -306,3 +313,33 @@ mod private {
...
+///
+/// - [SetClause::Immediate]
+/// will add `SET` right after the [UpdateStatement::from_clause] `QueryFragment`. \
+/// `Update users SET ... ;`
+/// - [SetClause::Delegated]
+/// hands over the control of adding `SET` to [UpdateStatement::values] `QueryFragment`. \
+/// `Update users ... SET ... ; ` will then be permitted.
+/// Batch update for mysql requires this behavior.
+#[derive(Clone, Copy, Debug)]
+pub enum SetClause {
```
be-mevsane, line 392: Hi again, @weiznich this approach worked and passed both commands at least locally:
- `cargo xtask run-tests -- postgres`
- `cargo xtask run-tests -- mysql`

I will look for an approach for SqLite as well.

Further TODO's:
- proper cleanup
- improve docs where required
- fix github action complaints
- add Changelog entry with explained restrictions

diesel/src/query_builder/update_statement/batch_update.rs
weiznich: I think it shouldn't be required. That only was needed for emulating support on the sqlite backend for batch inserts in a particular way. I don't think we need that here.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,438 @@
...
+    type Table = Tab;
+
+    fn walk_ast<'b>(
+        &'b self,
+        mut out: AstPass<'_, 'b, DB>,
+        _sep: &'_ str,
+        ambiguous: bool,
+        alias: &'_ str,
+    ) -> QueryResult<()> {
+        match ambiguous {
```
weiznich: Just to clarify this: The values of `ambiguous` and `alias` are always the same between calls for the same aschangeset struct at different runtime locations. If that is not the case we might need to mark this kind of query to always require a dynamic prepared statement cache id.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,438 @@
...
+    DB: Backend,
+    Self: QueryFragment<DB, DB::BatchUpdateSupport>,
+{
+    fn walk_ast<'b>(&'b self, pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        <Self as QueryFragment<DB, DB::BatchUpdateSupport>>::walk_ast(self, pass)
+    }
+}
+
+/// TODO:
+/// Does it make sense to put this to pg/query_builder
```
weiznich: If that's possible without a major rework that would be appreciated

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,438 @@
...
+            false,
+            Self::ALIAS,
+        )?;
+
+        Ok(())
+    }
+}
+
+/// TODO:
+/// Does it make sense to put this to mysql/query_builder
```
weiznich: Similar to the postgres case that would be great.

diesel/src/query_builder/update_statement/batch_update.rs
weiznich: I think the easiest way to solve this is to move the call to `out.push_sql(" SET ");` from the `UpdateStatement` implementation into the various value clause implementations. I don't think we need an additional type for that.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,438 @@
...
+        //      UNION ALL SELECT 2, 5, 10
+        //      UNION ALL SELECT 3, 15, 30
+        // ) AS tmp
+        // ON
+        //      tab.column_b = tmp.column_b
+        // SET
+        //      tab.column_a = tmp.column_a,
+        //      tab.column_c = tmp.column_c;
+
+        // Option 2: mysql 8.x or above
```
weiznich: It should be fine to go with that variant as that is already implemented. If really someone shows up that needs support for older versions of mysql we still can change the generated SQL. Or rather accept a PR from them to do that. It would be still meaningful for that to keep that comment here.

diesel/src/query_builder/update_statement/batch_update.rs
```diff
@@ -0,0 +1,438 @@
...
+        //      tab.column_c = tmp.column_c
+        out.push_sql(" SET ");
+        BatchColumnAssign::walk_ast(&first.1, out.reborrow(), ", ", false, Self::ALIAS)?;
+
+        Ok(())
+    }
+}
+
+/// TODO:
+/// Check out if there is a supported batch update syntax for sqlite.
```
weiznich: It's also fine to just not support it for now.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +150,138 @@ where
...
+
+// Identifiable takes ownership if not implemented on a reference.
+// Following implementations:
+// - impl<U> AsChangeset for Vec<U>
+// - impl<U, const N: usize> AsChangeset for [U; N]
+// - impl<U, const N: usize> AsChangeset for Box<[U; N]>
+//
+// result in compile error:
+// the parameter type `U` may not live long enough
+// ...so that the reference type `&'a U` does not outlive the data it points at
```
weiznich: It might be possible to address this with a higher ranked lifetime, but it also might make that code hit edge cases in rustc.

Given that I think it's reasonable to just start with supporting only references for now and add support for owned types later on.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +150,138 @@ where
...
+}
+
+impl<'a, U, I, C, PK, const N: usize> AsChangeset for &'a [U; N]
+where
+    U: AsChangeset + HasTable<Table = U::Target>,
+    U::Target: Table<PrimaryKey = PK>,
+    &'a U: AsChangeset<Target = U::Target, Changeset = C> + Identifiable<Table = U::Target, Id = I>,
+{
+    type Target = U::Target;
+    type Changeset = BatchUpdate<I, C, PK, U::Target, (), false>;
```
weiznich: The `STABLE_QUERY_ID` generic can be set to `true` here as we always get the same query for this type as the size of the slice cannot change at runtime.

diesel/src/query_builder/update_statement/changeset.rs
```diff
@@ -140,3 +150,138 @@ where
...
+// Identifiable takes ownership if not implemented on a reference.
+// Following implementations:
+// - impl<U> AsChangeset for Vec<U>
+// - impl<U, const N: usize> AsChangeset for [U; N]
+// - impl<U, const N: usize> AsChangeset for Box<[U; N]>
+//
+// result in compile error:
+// the parameter type `U` may not live long enough
+// ...so that the reference type `&'a U` does not outlive the data it points at
+impl<'a, U, I, C, PK> AsChangeset for &'a [U]
```
weiznich, line 181: At some point we need to call `AstPass::unsafe_to_cache_prepared()` for the `&[U]` and `&Vec<U>` variant as both might generate an unbound number of prepared statements due to different sizes of the values vector/slice. 

For `&[U;N]` that's not required as that has a statically known size at compile time. In fact that's the reason why this specialization exists at all.

diesel/src/query_builder/update_statement/mod.rs
```diff
@@ -57,6 +60,7 @@ impl<T: QuerySource, U> UpdateStatement<T, U, SetNotCalled> {
 pub struct UpdateStatement<T: QuerySource, U, V = SetNotCalled, Ret = NoReturningClause> {
     from_clause: T::FromClause,
     where_clause: U,
+    set_clause: SetClause,
```
weiznich, line 68: As written before: I would rather always defer calling `pass.push_sql("SET");` to the "values" wrappers instead. Seems easier and doesn't need that much setup.

diesel/src/backend.rs
```diff
@@ -470,6 +487,39 @@ pub(crate) mod sql_dialect {
...
+        /// In this case diesel will emulate batch update support
+        /// by updating each row on its own
+        #[derive(Debug, Copy, Clone)]
+        pub struct DoesNotSupportBatchUpdate;
+
+        /// Indicates that this backend supports postgres style
+        /// batch update statements to update multiple rows using one
+        /// update statement
+        #[derive(Debug, Copy, Clone)]
+        pub struct PostgresLikeBatchUpdateSupport;
```
weiznich: Can we move that to the postgres backend directory?

diesel/src/backend.rs
```diff
@@ -470,6 +487,39 @@ pub(crate) mod sql_dialect {
...
+
+        /// Indicates that this backend supports mysql style
+        /// batch update statements to update multiple rows using one
+        /// update statement
+        #[derive(Debug, Copy, Clone)]
+        pub struct MySqlLikeBatchUpdateSupport;
```
weiznich: Can we move that to the mysql backend directory?

diesel/src/backend.rs
```diff
@@ -470,6 +487,39 @@ pub(crate) mod sql_dialect {
...
+
+        impl SupportsBatchUpdate for PostgresLikeBatchUpdateSupport {}
```
weiznich: Either we also need an implementation for the mysql variant or we don't need that trait at all?

weiznich (commented):
Thanks for keeping up the good work. I think we are nearly there. The overall implementation approach looks good to me (beside the handling of the `SET` clause…), there are still a good number of minor style things I would like to see addressed.

Other than that we would like to have some documentation for this, possibly by just having at least one example at `diesel::update` for this behaviour. Finally this is in my opinion a feature that justifies a CHANGELOG entry, so it would be great to have it listed there as well.

diesel/src/query_builder/update_statement/batch_update.rs
be-mevsane: Alright, removed.

diesel/src/query_builder/update_statement/batch_update.rs
be-mevsane: > The values of ambiguous and alias are always the same between calls for the same aschangeset struct at different runtime locations.

Thats holds true.

diesel/src/query_builder/update_statement/batch_update.rs
be-mevsane: Sure, done.

diesel/src/query_builder/update_statement/batch_update.rs
be-mevsane: Sure, done.

diesel/src/query_builder/update_statement/changeset.rs
be-mevsane: Added.

diesel/src/backend.rs
be-mevsane: Done.

diesel/src/backend.rs
be-mevsane: Done.

diesel/src/backend.rs
be-mevsane: Updated all such that `SupportsBatchUpdate` is now required for all that want to support BatchUpdate.

diesel/src/query_builder/update_statement/mod.rs
be-mevsane: Assume `struct User { id: i32, name: String, hair_color: String, type: String }` with `id`as the PK:

Calling `AsChangeset::as_changest(&User)` will generate a tuple of three components `(Assign<name, Expr>, Assign<hair_color, Expr>, Assign<type, Expr>)` where the tuple and each Assign implements QueryFragment. There is currently no values "wrapper" other than that tuple generated for regular updates.

Therefore calling `pass.push_sql("SET");` has to be done at the QueryFragment level for the tuple then. Otherwise, that keyword will repeat for each assigned value, in this case 3 times.

That would be here: [QueryFragment implementation ](https://github.com/diesel-rs/diesel/blob/main/diesel/src/type_impls/tuples.rs#L105). This is used globally for other implementations, that expect to just add a comma between the individual elements/fragments.

Since the regular update case does not provide a "values" wrapper similar to the BatchUpdate case I was only able to achieve this through an additional setup.

Note: Adding a values "wrapper" for the regular case is an option here. This would probably introduce more refactoring for the default behavior and for the BatchUpdate as well, since BatchUpdate was implemented by respecting that the `Assign` struct contains individual fields of an update struct rather than all relevant fields for that update struct.

weiznich:
I just wanted to leave a note here that I'm aware of the update and I started looking into the changes, but I unfortunately did not finish doing that yet. Hopefully I'm able to do it next week.

weiznich:
It took me quite a long time to get back to this. Sorry about that. I finally found the to look into the remaining blockers. As some of them had a quite unclear solution I sat down and did all of the required cleanup from my side.

The commit beside the merges does:

* Add support for sqlite
* Add support for owned container types
* Refactor the internal trait structure to not require that many traits
* Add additional bounds to disallow `filter`/`find` calls for the update statement
* Add more tests

From my perspective this is now ready, but I would appreciate a review from @diesel-rs/reviewers, especially for my additional changes as they are non-trivial.

diesel/Cargo.toml
```diff
@@ -71,7 +71,7 @@ tempfile = "3.10.1"
 diesel_test_helper = { path = "../diesel_test_helper" }
 
 [features]
-default = ["with-deprecated", "32-column-tables", "std"]
```
LucaCappelletti94, line 74: Why are you removing `32-column-tables`?

LucaCappelletti94 (changes requested):

diesel/Cargo.toml
weiznich: That shouldn't be there. I just tend to remove this feature for local development as it helps with compile times. (Pushed a fix for this)

weiznich (approved):

PR #4919, opened 2025-12-26
fix(sqlite): classify ON DELETE RESTRICT violations as ForeignKeyViolation

### LINKED ISSUE #4299: DatabaseError caused by foreign key `ON DELETE RESTRICT` is not classified as `ForeignKeyViolation` using Sqlite
### Versions

- **Rust:** rustc 1.81.0 (eeb90cda1 2024-09-04) (rustc 1.83.0-nightly (fb4aebddd 2024-09-30) originally tested )
- **Diesel:** diesel v2.2.4
- **Database:** 3.46.0 2024-05-23 13:25:27 96c92aba00c8375bc32fafcdf12429c58bd8aabfcadab6683e35bbb9cdebf19e (64-bit)
- **Operating System** NixOS 24.11.20241006.c31898a (Vicuna)

### Feature Flags

- **diesel:** "sqlite", "returning_clauses_for_sqlite_3_35", "r2d2"

## Problem Description
Diesel using sqlite doesn't recognize foreign key violations caused by deleting a row that is referenced by a row in a different table which is a foreign key with `ON DELETE RESTRICT`. It returns a `DatabaseErrorKind::Unknown` instead of a `DatabaseErrorKind::ForeignKeyViolation`.

I believe a `DatabaseErrorKind::ForeignKeyViolation` would be appropriate because the error arises from the foreign key constraint and the error message says "FOREIGN KEY constraint failed".

Check if the error return from `diesel::delete(...).execute(...)` is caused by a foreign key violation, or some general database error, in order to properly handle the error.

For now I'm thinking of just checking for the error message, but I'm not sure if that is stable or might change at some point.

```
DatabaseError(
    ForeignKeyViolation,
    "FOREIGN KEY constraint failed",
)
```

```
DatabaseError(
    Unknown,
    "FOREIGN KEY constraint failed",
)
```

No

### Steps to reproduce

1. Create a table with a foreign key with `ON DELETE RESTRICT`
2. Violate that restriction using `diesel::delete`.

Reproduction: https://github.com/dav-wolff/diesel-foreign-key-violation

## Checklist

majiayu000:
# Description

SQLite returns `SQLITE_CONSTRAINT_TRIGGER` for `ON DELETE RESTRICT` violations instead of `SQLITE_CONSTRAINT_FOREIGNKEY`. This caused Diesel to incorrectly classify these errors as `DatabaseErrorKind::Unknown` instead of `DatabaseErrorKind::ForeignKeyViolation`.

## Solution

When `SQLITE_CONSTRAINT_TRIGGER` is returned, check if the error message contains "FOREIGN KEY constraint failed". If so, classify it as `ForeignKeyViolation`. This distinguishes actual foreign key violations from user-defined trigger failures.

## Changes

- `diesel/src/sqlite/connection/stmt.rs`: Added pattern matching for `SQLITE_CONSTRAINT_TRIGGER` with message check
- `diesel_tests/tests/errors.rs`: Added test case `foreign_key_restrict_violation_detected`

## Test

```rust
// Create tables with ON DELETE RESTRICT
// Insert parent and child records
// Attempt to delete parent → should return ForeignKeyViolation
```

Fixes #4299

### DIFF
--- a/diesel/src/sqlite/connection/stmt.rs
+++ b/diesel/src/sqlite/connection/stmt.rs
@@ -212,16 +212,25 @@ pub(super) fn ensure_sqlite_ok(
 
 fn last_error(raw_connection: *mut ffi::sqlite3) -> Error {
     let error_message = last_error_message(raw_connection);
-    let error_information = Box::new(error_message);
-    let error_kind = match last_error_code(raw_connection) {
+    let error_code = last_error_code(raw_connection);
+    let error_kind = match error_code {
         ffi::SQLITE_CONSTRAINT_UNIQUE | ffi::SQLITE_CONSTRAINT_PRIMARYKEY => {
             DatabaseErrorKind::UniqueViolation
         }
         ffi::SQLITE_CONSTRAINT_FOREIGNKEY => DatabaseErrorKind::ForeignKeyViolation,
+        // SQLITE_CONSTRAINT_TRIGGER is returned for ON DELETE RESTRICT violations,
+        // which are actually foreign key violations. We check the error message
+        // to distinguish from user-defined trigger failures.
+        ffi::SQLITE_CONSTRAINT_TRIGGER
+            if error_message.contains("FOREIGN KEY constraint failed") =>
+        {
+            DatabaseErrorKind::ForeignKeyViolation
+        }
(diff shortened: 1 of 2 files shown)


weiznich (approved):
Thanks, looks good to me

majiayu000:
Happy to help!

PR #4927, opened 2026-01-08
Added helper traits for `NamedField` and `DynamicRow`

LucaCappelletti94:
This PR implements standard Rust traits for `diesel_dynamic_schema` types to improve usability and allow common patterns like mutation, collection, and transparent field access.

For `DynamicSelectClause`, I've added `Extend` and `FromIterator` implementations to allow building clauses dynamically from iterators. `IntoIterator` and `AsRef` have also been implemented to facilitate inspection and composition.

`DynamicRow` now supports mutable access via `IndexMut` (by index or name) and `iter_mut()`. `IntoIterator` implementations for owned, borrowed, and mutable borrowed `DynamicRow`s enable flexible iteration. Additionally, `From` and `Into` conversions for `Vec` and `AsRef` are added to simplify data extraction and construction.

Finally, `NamedField` has been enhanced with `Deref` and `DerefMut` implementations targeting the inner value, making it act transparently like the value it holds. `AsRef`, `Borrow`, and standard derives like `Clone`, `Hash`, and `Eq` have also been added.

New tests covering these implementations were added to dynamic_values.rs and verified with the `diesel-dynamic-schema` test suite.

### DIFF
--- a/diesel_dynamic_schema/src/dynamic_select.rs
+++ b/diesel_dynamic_schema/src/dynamic_select.rs
@@ -3,6 +3,7 @@ use diesel::expression::{is_aggregate, NonAggregate, ValidGrouping};
 use diesel::query_builder::{AstPass, QueryFragment, QueryId};
 use diesel::sql_types::Untyped;
 use diesel::{AppearsOnTable, Expression, QueryResult, SelectableExpression};
+use std::iter::FromIterator;
 use std::marker::PhantomData;
 
 /// Represents a dynamically sized select clause
@@ -39,6 +40,18 @@ impl<'a, DB, QS> DynamicSelectClause<'a, DB, QS> {
         DB: Backend,
     {
         self.selects.push(Box::new(field))
+    }
+
+    /// Add multiple fields to the dynamically sized select clause
+    pub fn add_fields<I, F>(&mut self, fields: I)
+    where
+        I: IntoIterator<Item = F>,
+        F: QueryFragment<DB> + SelectableExpression<QS> + NonAggregate + Send + 'a,
+        DB: Backend,
+    {
+        for field in fields {
+            self.add_field(field);
+        }
     }
 }
 
@@ -80,5 +93,17 @@ impl<'a, DB, QS> AsRef<[Box<dyn QueryFragment<DB> + Send + 'a>]>
 {
     fn as_ref(&self) -> &[Box<dyn QueryFragment<DB> + Send + 'a>] {
         &self.selects
+    }
+}
+
+impl<'a, DB, QS, F> FromIterator<F> for DynamicSelectClause<'a, DB, QS>
+where
+    F: QueryFragment<DB> + SelectableExpression<QS> + NonAggregate + Send + 'a,
+    DB: Backend,
+{
+    fn from_iter<I: IntoIterator<Item = F>>(iter: I) -> Self {
+        let mut select_clause = DynamicSelectClause::new();
+        select_clause.add_fields(iter);
+        select_clause
     }
 }
--- a/diesel_dynamic_schema/src/dynamic_select.rs
+++ b/diesel_dynamic_schema/src/dynamic_select.rs
@@ -53,6 +53,16 @@ impl<'a, DB, QS> DynamicSelectClause<'a, DB, QS> {
             self.add_field(field);
         }
     }
+
+    /// Returns the number of fields in the select clause
+    pub fn len(&self) -> usize {
+        self.selects.len()
+    }
+
+    /// Returns whether the select clause is empty
+    pub fn is_empty(&self) -> bool {
+        self.selects.is_empty()
+    }
 }
 
 impl<DB, QS> AppearsOnTable<QS> for DynamicSelectClause<'_, DB, QS> where Self: Expression {}
@@ -88,14 +98,6 @@ impl<DB, QS> ValidGrouping<()> for DynamicSelectClause<'_, DB, QS> {
     type IsAggregate = is_aggregate::No;
 }
 
-impl<'a, DB, QS> AsRef<[Box<dyn QueryFragment<DB> + Send + 'a>]>
-    for DynamicSelectClause<'a, DB, QS>
-{
-    fn as_ref(&self) -> &[Box<dyn QueryFragment<DB> + Send + 'a>] {
-        &self.selects
-    }
-}
-
 impl<'a, DB, QS, F> FromIterator<F> for DynamicSelectClause<'a, DB, QS>
 where
     F: QueryFragment<DB> + SelectableExpression<QS> + NonAggregate + Send + 'a,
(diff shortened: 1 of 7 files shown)


diesel_dynamic_schema/src/dynamic_select.rs
```diff
@@ -74,3 +87,54 @@ where
...
     type IsAggregate = is_aggregate::No;
 }
+
+impl<'a, DB, QS> AsRef<[Box<dyn QueryFragment<DB> + Send + 'a>]>
+    for DynamicSelectClause<'a, DB, QS>
+{
+    fn as_ref(&self) -> &[Box<dyn QueryFragment<DB> + Send + 'a>] {
+        &self.selects
+    }
+}
```
weiznich: I would like to avoid exposing this internal representation as it makes harder to change it later, e.g adding more trait bounds to the trait object or similar changes.

diesel_dynamic_schema/src/dynamic_select.rs
```diff
@@ -74,3 +87,54 @@ where
...
+    type Item = &'b (dyn QueryFragment<DB> + Send + 'a);
+    type IntoIter = std::iter::Map<
+        std::slice::Iter<'b, Box<dyn QueryFragment<DB> + Send + 'a>>,
+        fn(&'b Box<dyn QueryFragment<DB> + Send + 'a>) -> &'b (dyn QueryFragment<DB> + Send + 'a),
+    >;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.selects.iter().map(|b| &**b)
+    }
+}
```
weiznich: Same as with the `AsRef` impl, although that one is not that critical as we could just casting to the right trait object here in the future. After all both the `QueryFragment` and `Send` bound is definitively required. For that suggestion it would be required to somewhat turn these `IntoIter` types into opaque types, maybe just even `Box<dyn Iterator<Item = Self::Item>>`?

diesel_dynamic_schema/src/dynamic_value.rs
```diff
@@ -186,22 +187,72 @@ impl diesel::expression::QueryMetadata<Any> for diesel::mysql::Mysql {
...
+        &self.value
+    }
+}
+
+impl<I> DerefMut for NamedField<I> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.value
+    }
+}
+
```
weiznich: I would like to avoid these impl as they are rather atypical for non-smart pointers.

diesel_dynamic_schema/src/dynamic_value.rs
```diff
@@ -231,6 +289,16 @@ impl<I> DynamicRow<I> {
         self.values.is_empty()
     }
 
+    /// Returns an iterator over the values of the row
+    pub fn iter(&self) -> std::slice::Iter<'_, I> {
```
weiznich: Rather use `impl Iterator<Item =& I> + use<'_>` to make the type opaque

diesel_dynamic_schema/src/dynamic_value.rs
```diff
@@ -231,6 +289,16 @@ impl<I> DynamicRow<I> {
...
+        self.values.iter()
+    }
+
+    /// Returns a mutable iterator over the values of the row
+    pub fn iter_mut(&mut self) -> std::slice::IterMut<'_, I> {
```
weiznich: Rather use `impl Iterator<Item =&mut  I> + use<'_>` to make the type opaque

diesel_dynamic_schema/src/dynamic_value.rs
```diff
@@ -221,6 +272,13 @@ impl<I> DynamicRow<I> {
         self.values.get(index)
     }
 
+    /// Get the mutable field value at the provided row index
+    ///
+    /// Returns `None` if the index is outside the bounds of the row
+    pub fn get_mut(&mut self, index: usize) -> Option<&mut I> {
```
weiznich, line 251: Should we also have a `fn get_ref(&self, index: usize) -> Option<&I>` variant here?

diesel_dynamic_schema/src/dynamic_value.rs
```diff
@@ -264,6 +332,18 @@ impl<I> DynamicRow<NamedField<I>> {
             .find(|f| f.name == name.as_ref())
             .map(|f| &f.value)
     }
+
+    /// Get the mutable field value by the provided field name
+    ///
+    /// Returns `None` if the field with the specified name is not found.
+    /// If there are multiple fields with the same name, the behaviour
+    /// of this function is unspecified.
+    pub fn get_mut_by_name<S: AsRef<str>>(&mut self, name: S) -> Option<&mut I> {
```
weiznich, line 314: Should we also have a fn get_ref_by_name(&self, name: impl AsRef<str>) -> Option<&I> variant here?

diesel_dynamic_schema/tests/dynamic_values.rs
```diff
@@ -67,6 +67,108 @@ impl FromSql<Any, diesel::mysql::Mysql> for MyDynamicValue {
...
+        DynamicSelectClause::new();
+    select_clause.add_field(name);
+
+    // Extend
+    let hair = users.column::<Untyped, _>("hair_color");
+    let fields = vec![hair];
+    select_clause.extend(fields);
+
+    // IntoIterator
+    assert_eq!(select_clause.into_iter().count(), 2);
```
weiznich: We can also add a `.len()` function for the container as well. (Maybe also `is_empty()`?)

weiznich (commented):
Overall these are mostly meaningful additions. I'm happy to see them merged. I have some minor remarks on making things more maintainable in the future, otherwise this should be ready to go.

Might be something to add to the changelog as well

(Also the collation stuff should likely have a changelog entry if I think about it)

diesel_dynamic_schema/src/dynamic_select.rs
LucaCappelletti94: Removed

diesel_dynamic_schema/src/dynamic_value.rs
LucaCappelletti94: Removed

diesel_dynamic_schema/src/dynamic_select.rs
LucaCappelletti94: I removed these impl to avoid causing any future issue

diesel_dynamic_schema/src/dynamic_value.rs
LucaCappelletti94: Done

diesel_dynamic_schema/src/dynamic_value.rs
LucaCappelletti94: Done

diesel_dynamic_schema/src/dynamic_value.rs
LucaCappelletti94: There is already `get` which returns `Option<&I>`, should I rename it?

diesel_dynamic_schema/src/dynamic_value.rs
LucaCappelletti94: There is already `get_by_name`, should I rename it?

diesel_dynamic_schema/tests/dynamic_values.rs
LucaCappelletti94: Added both.

weiznich (approved):

PR #4930, opened 2026-01-08
Added a first version of `UntypedExpressionMethods`

LucaCappelletti94:
This PR introduces the `UntypedExpressionMethods` trait, enabling the use of `is_null`, `is_not_null`, `asc`, and `desc` methods on expressions where the SQL type is `Untyped`.

Previously, these methods were provided solely by `ExpressionMethods`, which restricts usage to expressions where `SqlType` implements `SingleValue`. Since `Untyped` does not implement `SingleValue`, users could not use these helper methods on untyped expressions (e.g., in certain dynamic query scenarios common in [diesel_dynamic_schema](https://docs.rs/diesel-dynamic-schema/latest/diesel_dynamic_schema/)).

While it might be preferable to split the `ExpressionMethods` trait into two traits to reduce code redoundancy, it would change the APIs and is therefore undesirable not being nearing a major release.

## Changes
- **Added `UntypedExpressionMethods` trait**: Defined in global_expression_methods.rs with the following methods:
  - `is_null()`
  - `is_not_null()`
  - `asc()`
  - `desc()`
- **Implemented Trait**: Added a blanket implementation for any `Expression` where `SqlType = Untyped`.
- **Exported Trait**: Exposed `UntypedExpressionMethods` in mod.rs so it is available by default.
- **Updated Tests**: Updated existing compile-fail tests (`cannot_use_expression_methods_on_tuples`) to reflect that `UntypedExpressionMethods` is now considered during trait resolution.

## Motivation
This change improves ergonomics for users working with `Untyped` expressions, allowing them to perform standard null checks and ordering without needing to manually construct the expression objects or cast types unnecessarily.

See discussion #4929

### DIFF
--- a/diesel/src/expression_methods/global_expression_methods.rs
+++ b/diesel/src/expression_methods/global_expression_methods.rs
@@ -3,7 +3,7 @@ use crate::expression::array_comparison::{AsInExpression, In, NotIn};
 use crate::expression::grouped::Grouped;
 use crate::expression::operators::*;
 use crate::expression::{assume_not_null, cast, nullable, AsExpression, Expression};
-use crate::sql_types::{SingleValue, SqlType};
+use crate::sql_types::{SingleValue, SqlType, Untyped};
 
 /// Methods present on all expressions, except tuples
 pub trait ExpressionMethods: Expression + Sized {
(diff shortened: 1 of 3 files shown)


diesel/src/expression_methods/global_expression_methods.rs
```diff
@@ -589,6 +589,137 @@ where
 {
 }
 
+/// Methods present on untyped expressions.
```
weiznich, line 592: Maybe mention that this is mostly useful for `diesel_dynamic_schema` in the documentation?

weiznich (approved):
Looks reasonable to me, but I would like to improve the docs a bit more.

diesel/src/expression_methods/global_expression_methods.rs
LucaCappelletti94: Done

weiznich (approved):

PR #4932, opened 2026-01-09
Added `collate` method to `TextExpressionMethods`

LucaCappelletti94:
This PR adds support for the SQL `COLLATE` clause to `TextExpressionMethods`. This allows users to specify collation for text columns directly in their queries using `.collate("collation_name")`.

## Changes

-   **Added `Collate` operator**: Implemented in `diesel/src/expression/operators.rs`. It wraps an expression and a collation string, generating `expression COLLATE collation` SQL. Accepted collation argument is generic over `ToString`.
-   **Added `Collate` helper type**: Added type alias in `diesel/src/expression/helper_types.rs` for easier usage.
-   **Extended `TextExpressionMethods`**: Added `collate<S: ToString>(self, collation: S)` method in `diesel/src/expression_methods/text_expression_methods.rs`. This changes the APIs of the method, but since it is solely additive and the method added has a default implementation I believe it should be fine and don't break pre-existinging non-failure code.

## Usage Example

```rust
let names = users
    .select(name)
    .filter(name.collate("NOCASE").eq("sean")) // Generates: name COLLATE NOCASE = 'sean'
    .load::<String>(connection)?;
```

See discussion #4931


weiznich:
I just want to leave a note here: I really appreciate all the work you are putting in here. Unfortunately I do currently not have the capacity to review all the changes timely, especially without only focusing on your requests/PR's. The later is obviously something I cannot do as others also have the same rights for support as you.
I will try to review the changes one after another, but that will take time. Hopefully I get much of what's already open reviewed until February, but don't take that as promise. Depending on what else comes up it might also take longer.

LucaCappelletti94:
> I just want to leave a note here: I really appreciate all the work you are putting in here. Unfortunately I do currently not have the capacity to review all the changes timely, especially without only focusing on your requests/PR's. The later is obviously something I cannot do as others also have the same rights for support as you. I will try to review the changes one after another, but that will take time. Hopefully I get much of what's already open reviewed until February, but don't take that as promise. Depending on what else comes up it might also take longer.

That's ok, at this time for my projects I am using my own branch. I just try my best to do PRs any time I find myself needing some features that would most likely benefit not only my own projects. Often times this approach makes me realize some bad engineering choice I made, and generally makes the resulting code better.

Of course, I will not be able to publish dependent crates until the updated `diesel` is published, but I prefer the current slow and steady approach. Better a solid software in a month than a bash-script today.

weiznich (commented):
I did not have a in depth look at the various in invariants that might relevant here, but one thing I would like to change here is using typed collations instead of strings. So essentially define a zero sized struct for each collation, implement the relevant traits (QueryFragment, etc) so that each collation is only allowed for a specific backend that supports it and likely have another trait to restrict the argument of the collate function to these types. Likely all of that could use an internal macro to not repeat everything all the time. 

That should make it more robust and also extensible for downstream users that want to have custom collations. 


Also as unrelated note: You don't need to rebase/merge to keep up with main. It's sufficient that there are no conflicts

LucaCappelletti94:
Ok, with the idea that if there is some unsupported or user-defined collation, the user will either have to use the raw sql or create an appropriate zero-sized struct, correct?

weiznich:
We could have a custom `struct CustomCollation(String);` for that use case as well.

LucaCappelletti94:
I believe using `struct CustomCollation(String);` may facilitate the user to commit errors, especially if we are not validating that `String` - and I have no idea regarding how to validate the collation. That being said, it may be a good fit for the context of [`diesel_dynamic_schema`](https://docs.rs/diesel-dynamic-schema/latest/diesel_dynamic_schema/)?

I searched for a while and composed the following list of common collations we may want to support as a struct type by default. There are of course many more, especially if we consider the locale collations such as `de_DE.utf8`, which are generally of the form `{locale}.{charset}`. We could possibly create a macro that generates all collation structs for such [ISO language codes](https://docs.oracle.com/cd/E13214_01/wli/docs92/xref/xqisocodes.html), or possibly more reasonably, do so in a downhill crate.

It may be desirable to add a `collate_{type of collate}()` methods for some of the most commonly used ones.

| Proposed Struct Name            | Databases / Name             | Behavior (short)                                                | Docs |
|-----------------|------------------------------|------------------------------------------------------------------|------|
| `BinaryCollation`        | MySQL `BINARY`, SQLite       | Byte-wise compare, case-sensitive, locale-free                   | https://dev.mysql.com/doc/refman/8.4/en/charset-mysql.html |
| `NoCaseCollation`        | SQLite `NOCASE`             | ASCII-only case-insensitive, others byte-wise                    | https://sqlite.org/datatype3.html#collating_sequences |
| `PosixCollation`             | PostgreSQL `C`, `POSIX`      | Byte-wise, only ASCII A–Z treated as letters                     | https://www.postgresql.org/docs/current/collation.html#COLLATION-STANDARD-COLLATIONS |
| `RTrimCollation`         | SQLite `RTRIM`              | Ignores trailing spaces                                          | https://sqlite.org/datatype3.html#collating_sequences |

weiznich:
Thanks for putting together that list. I think that's a good starting point. I would like to propose organizing it a bit differently. Instead of naming each collation `*Collation`, what's about having a `collation` module instead and allow the user to choose to either import them directly or write `collation::Default` (or whatever).

> I believe using struct CustomCollation(String); may facilitate the user to commit errors, especially if we are not validating that String - and I have no idea regarding how to validate the collation.

I think it might be fine to have this in diesel itself, maybe as `struct Custom(&'static str)` to restrict a possibility to pass in user provided input there? Otherwise there are already other apis like `sql_query` or `dsl::sql` that just take unverified content and put it literally into the SQL query. There we just explicitly note that this is the case and trust that the user does the right thing. After all it's part of the API then.

> There are of course many more, especially if we consider the locale collations such as de_DE.utf8, which are generally of the form {locale}.{charset}. We could possibly create a macro that generates all collation structs for such [ISO language codes](https://docs.oracle.com/cd/E13214_01/wli/docs92/xref/xqisocodes.html), or possibly more reasonably, do so in a downhill crate.

Raising the language collations here is a good point. I wonder if an large enum would be more appropriated for them? But yes, for now that's something I would just push to downstream crates to handle. They can define whatever collation they want to have or just use the `Custom` type for that. Should be simple enough for them.

LucaCappelletti94:
I have reduced the collations to a set with which I am familiar with, as the other ones are rather under documented and I do not understand how to use them.

diesel/src/collation/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[cfg(feature = "postgres")]
+pub struct C;
+
+#[cfg(feature = "postgres")]
+impl fmt::Display for C {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\"C\"")
+    }
+}
```
weiznich, line 70: I'm sorry to bring up yet another request, but I had something different in mind here. The current implementation would allow to pass a `collation::Binary` to a postgres query if you enable both the `sqlite` and `postgres` feature at the same time. This is something I would like to avoid. 

Instead of the current implementation I would prefer using more of diesels query builder to exactly prevent this situation

Something like that would be closer to what I have in mind:


```suggestion
pub trait Collation {}

/// The `BINARY` collation.
///
/// This collation is binary, case-sensitive, and locale-free.
/// It is supported by SQLite.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default, QueryId)]
#[cfg(feature = "sqlite")]
pub struct Binary;

impl Collation for Binary {}

#[cfg(feature = "sqlite")]
impl QueryFragment<Sqlite> for Binary {
    fn walk_ast<'b>(&'b self, mut pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
        pass.push_sql("BINARY");
        Ok(())
    }
}
// same impl for Mysql


/// The `C` collation.
///
/// This collation is byte-wise, with only ASCII A-Z treated as letters.
/// It is supported by PostgreSQL.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default, QueryId)]
#[cfg(feature = "postgres")]
pub struct C;

impl Collation for C {}

#[cfg(feature = "postgres")]
impl QueryFragment<Sqlite> for C {
    fn walk_ast<'b>(&'b self, mut pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
        pass.push_sql("BINARY");
        Ok(())
    }
}
```

then the `TextExpressionMethod` crate should have a single `collate` method that's defined like this:

```
fn collate<C>(self, collation: C) -> dsl::Collate<Self, C> where C: Collation {
    Grouped(Collate::new(self, collation))
}
```

This allows us to use static query ids (and therefore prepared statement caching) for all collations but `Custom`. Also it makes everything more strictly typed and harder to mix up collations between backends in multi-backend code-bases.

LucaCappelletti94: Ok

LucaCappelletti94: Done

diesel/src/collation/mod.rs
```diff
@@ -3,7 +3,12 @@
 //! These types are used as arguments for
 //! [`TextExpressionMethods::collate`](crate::expression_methods::TextExpressionMethods::collate).
 
-use std::fmt;
+use crate::backend::Backend;
+use crate::query_builder::{AstPass, QueryFragment, QueryId};
+use crate::result::QueryResult;
+
+/// Trait to identify a valid collation.
+pub trait Collation: QueryId + Copy + Send + Sync + 'static {}
```
weiznich: I would rather not have the `Copy` bound here. That might be a future compatibility hazard so I rather would not guarantee it

diesel/src/collation/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+
+/// The `C` collation.
+///
+/// This collation is byte-wise, with only ASCII A-Z treated as letters.
+/// It is supported by PostgreSQL.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[cfg(feature = "postgres")]
+pub struct C;
+
+#[cfg(feature = "postgres")]
```
weiznich:
```suggestion
#[cfg(feature = "postgres_backend")]
```

otherwise it's not useable with diesel-async

(The same for the other locations and for "mysql" -> "mysql_backend")

weiznich (commented):
Thanks for the update. It's nearly ready now, just some minor nits.

diesel/src/collation/mod.rs
LucaCappelletti94: Done, replaced with `Clone`

diesel/src/collation/mod.rs
LucaCappelletti94: I actually ended up not adding any collations for `mysql` as, being some 10 years that I do not use it, I am really unfamiliar with which collations work in that dialect. I changed all "postgres" to "postgres_backend".

PR #4933, opened 2026-01-10
Add benchmarks of zero-mysql and zero-postgres

elbaro:
This PR adds the initial benchmark code of `zero-mysql` and `zero-postgres` to diesel-bench.

- Add zero-mysql (sync, async)
- Add zero-postgres (sync, async)
- Fix Cargo.toml indentation to 4 spaces as specified in .editorconfig
- Include futures-util in the feature tokio_postgres
- Explicitly disable the silent upgrade of tcp socket to unix socket (`prefer_socket`, `upgrade_to_unix_socket`)

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -71,13 +71,13 @@ jobs:
           toolchain: stable
       - name: Run Benchmarks (Postgres)
         if: matrix.backend == 'postgres'
-        run: cargo +stable bench --no-fail-fast --manifest-path diesel_bench/Cargo.toml --no-default-features --features "postgres sqlx-bench sqlx/postgres rand_chacha rust_postgres tokio_postgres futures futures-util sea-orm sea-orm/sqlx-postgres criterion/async_tokio serde diesel-async diesel-async/postgres wtx wtx/postgres"
+        run: cargo +stable bench --no-fail-fast --manifest-path diesel_bench/Cargo.toml --no-default-features --features "postgres sqlx-bench sqlx/postgres rand_chacha rust_postgres tokio_postgres futures futures-util sea-orm sea-orm/sqlx-postgres criterion/async_tokio serde diesel-async diesel-async/postgres wtx wtx/postgres zero-postgres"
       - name: Run Benchmarks (Sqlite)
         if: matrix.backend == 'sqlite'
         run: cargo +stable bench --no-fail-fast --manifest-path diesel_bench/Cargo.toml --no-default-features --features "sqlite sqlx-bench sqlx/sqlite tokio rusqlite futures sea-orm sea-orm/sqlx-sqlite criterion/async_tokio diesel-async diesel-async/sqlite"
       - name: Run Benchmarks (Mysql)
         if: matrix.backend == 'mysql'
-        run: cargo +stable bench --no-fail-fast --manifest-path diesel_bench/Cargo.toml --no-default-features --features "mysql sqlx-bench sqlx/mysql tokio rustorm rustorm/with-mysql rand_chacha rustorm_dao rust_mysql futures sea-orm sea-orm/sqlx-mysql criterion/async_tokio serde diesel-async diesel-async/mysql wtx wtx/mysql"
+        run: cargo +stable bench --no-fail-fast --manifest-path diesel_bench/Cargo.toml --no-default-features --features "mysql sqlx-bench sqlx/mysql tokio rustorm rustorm/with-mysql rand_chacha rustorm_dao rust_mysql futures sea-orm sea-orm/sqlx-mysql criterion/async_tokio serde diesel-async diesel-async/mysql wtx wtx/mysql zero-mysql"
       - name: Push metrics
         env:
           SSH_AUTH_SOCK: /tmp/ssh_agent.sock


diesel_bench/benches/mysql_benches.rs
```diff
@@ -28,7 +28,8 @@ fn connection() -> Conn {
     let connection_url = dotenvy::var("MYSQL_DATABASE_URL")
         .or_else(|_| dotenvy::var("DATABASE_URL"))
         .expect("DATABASE_URL must be set in order to run tests");
-    let opts = Opts::from_url(&connection_url).unwrap();
+    let opts = OptsBuilder::from_opts(Opts::from_url(&connection_url).unwrap())
+        .prefer_socket(false);
```
weiznich, line 32: What is the reason for this change?

diesel_bench/benches/zero_mysql_async_benches.rs
```diff
@@ -0,0 +1,329 @@
...
+    size: usize,
+    hair_color_init: impl Fn(usize) -> Option<&'static str>,
+) {
+    let mut query = String::from("INSERT INTO users (name, hair_color) VALUES ");
+    for x in 0..size {
+        if x > 0 {
+            query.push(',');
+        }
+        let hair_color = match hair_color_init(x) {
+            Some(c) => format!("'{}'", c),
```
weiznich: It's required to use prepared statements like all the other backends

diesel_bench/benches/zero_mysql_async_benches.rs
```diff
@@ -0,0 +1,329 @@
...
+        let stmt = conn
+            .prepare("SELECT id, name, hair_color FROM users")
+            .await
+            .unwrap();
+        (conn, stmt)
+    });
+
+    b.iter(|| {
+        runtime.block_on(async {
+            let mut users = Vec::with_capacity(size);
```
weiznich: This shouldn't use the "know" benchmark size. You can preallocate the vector, but you need to get the size value from the query itself otherwise this wouldn't be a fair comparison with other implementations.

(Applies to all other functions as well)

diesel_bench/benches/zero_mysql_benches.rs
```diff
@@ -0,0 +1,289 @@
+use crate::Bencher;
```
weiznich, line 1: Here mostly the same comments as for the async mysql benchmark apply.

diesel_bench/benches/zero_postgres_async_benches.rs
```diff
@@ -0,0 +1,300 @@
+use crate::Bencher;
```
weiznich, line 1: Here mostly the same comments as for the async mysql benchmark apply. 

The other thing I wondered is if it might be meaningful to have both benchmarks share the same file as a lot of code is the same

diesel_bench/benches/zero_postgres_benches.rs
```diff
@@ -0,0 +1,255 @@
+use crate::Bencher;
```
weiznich, line 1: Here mostly the same comments as for the async mysql benchmark apply. 

The other thing I wondered is if it might be meaningful to have both benchmarks share the same file as a lot of code is the same

diesel_bench/benches/mysql_benches.rs
elbaro: When I benchmarked Python libraries, I suffered from this hidden socket upgrade.

Many libraries have their own upgrade mechanism, for example the python library `mysqliclient` upgrades the connection to UDS if DATABASE_URL contains `localhost`, but does not upgrade if DATABASE_URL contains `127.0.0.1`. The benchmark result via unix socket is much faster than TCP.

`mysql` and `mysql_async` have this `Opts.prefer_socket` which is convenient (I adopted it in my libs as well) but in benchmarks, this can make people compare TCP results against UDS results.

diesel_bench/benches/zero_postgres_async_benches.rs
elbaro: I think we can extract many parts from all benches in a separate PR.

This INSERT statement string builder is observed in many backends as well:

```rust
    let mut insert_query = String::from("INSERT INTO posts(title, user_id, body) VALUES ");
    for (user_id,) in &user_ids {
       insert_query.push('(?, ?)');
        ...
```

elbaro:
> a potential diesel_async connection implementation

This is certainly possible, behind the experimental feature flag until the zero-mysql reaches v1.0.

One thing I miss in this benchmark is the use of [COM_STMT_BULK_EXECUTE](https://mariadb.com/docs/server/reference/clientserver-protocol/3-binary-protocol-prepared-statements/com_stmt_bulk_execute), which is unfortunately MariaDB-only feature.
For bulk INSERTs, this makes a huge difference: https://github.com/elbaro/pyro-mysql/raw/main/report/INSERT.svg

diesel_bench/benches/mysql_benches.rs
weiznich: Sounds reasonable. We still need to be aware of this as libmysqlclient as used by diesel itself also has this behaviour.

PR #4938, opened 2026-01-14
feat(diesel): constify `PgInterval` methods

paolobarbolini:
The `PgInterval` methods are quite trivial. This constifies them so that they can be used in consts.


JohnTitor (approved):
Makes sense!

PR #4943, opened 2026-01-20
Made `Inner` and `JoinWithImplicitOnClause` accessible

LucaCappelletti94:
This PR addresses discussion #4942:

* Exports [`JoinWithImplicitOnClause`](https://docs.rs/diesel/latest/src/diesel/query_dsl/join_dsl.rs.html#29-33) as a sealed trait
* Exports the [`joins::Inner`](https://docs.rs/diesel/latest/src/diesel/query_source/joins.rs.html#286) marker struct

This is done in order to specify trait constraints necessary to use the [`inner_join`](https://docs.rs/diesel/latest/diesel/query_dsl/trait.QueryDsl.html#method.inner_join) method on a generic type `T`, which requires: `T: JoinWithImplicitOnClause<Rhs, Inner>`.

### DIFF
--- a/diesel/src/query_dsl/join_dsl.rs
+++ b/diesel/src/query_dsl/join_dsl.rs
@@ -32,8 +32,10 @@ mod private {
 /// `JoinDsl` support trait to emulate associated type constructors and grab
 /// the known on clause from the associations API
 pub trait JoinWithImplicitOnClause<Rhs, Kind>: private::Sealed<Rhs, Kind> {
+    #[doc(hidden)]
     type Output;
 
+    #[doc(hidden)]
     fn join_with_implicit_on_clause(self, rhs: Rhs, kind: Kind) -> Self::Output;
 }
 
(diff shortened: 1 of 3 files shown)


LucaCappelletti94:
With this PR, it is now possible to write, using nested tuples, variadic joins [as shown here](https://github.com/LucaCappelletti94/diesel-builders/blob/e48812c488996866d5dad892ae7ff2ac03a1632a/diesel-builders/src/load_nested_query_builder/nested_inner_join.rs).

With those, in turn, I was able to write variadic select queries of different kinds, [as shown here](https://github.com/LucaCappelletti94/diesel-builders/blob/e48812c488996866d5dad892ae7ff2ac03a1632a/diesel-builders/src/load_nested_query_builder.rs).

weiznich:
Given https://github.com/diesel-rs/diesel/issues/3223 I would rather not like to expose these types as this restricts any future possibility to change the type level representation of these joins later on to fix that issue at some point.

PR #4946, opened 2026-01-23
Add more migration sources

weiznich:
This PR adds more migration sources to diesel_migrations with the goal to make it easier to write different kinds of migrations in multi-crate deployments.

It adds a CombinedMigrations source that is essentially just a list of sources. The resulting list of migrations is just all migrations from all sources. That is useful for multi-crate projects as then each sub-project can have their own set of migrations (by using files or embedded migrations etc) and then combine them into a final list using this migration source.

As a second variant a RustMigrationSource is added. The goal there is to make it much easier to register rust based migrations by having some sort of typed interface for this. This migration source needs to know the connection type, not only the backend. This information is then used to internally receive the right connection type from the migration connection for the user. The user has the ability to provide different kinds of rust structs/functions all "migration. This includes closures, functions, a migration builder (that in turn accepts clousures but allows more fine grained configuration) or even custom types (via the provided trait).

# Rust migration source example:

```rust
use diesel::prelude::*;
use diesel_migrations::MigrationHarness;
use diesel_migrations::RustMigration;
use diesel_migrations::RustMigrationSource;
use diesel_migrations::TypedMigration;

fn migration_function(conn: &mut SqliteConnection) -> QueryResult<()> {
    diesel::sql_query("SELECT 'EXECUTE YOUR MIGRATION HERE'").execute(conn)?;
    Ok(())
}

struct CustomMigration(&'static str);

impl TypedMigration<SqliteConnection> for CustomMigration {
    fn up(&self, conn: &mut SqliteConnection) -> QueryResult<()> {
        diesel::sql_query("SELECT 'YOUR MIGRATION', ?")
            .bind::<diesel::sql_types::Text, _>(self.0)
            .execute(conn)?;
        Ok(())
    }
}

let mut rust_migrations = RustMigrationSource::<SqliteConnection>::new();

// register migrations callback
rust_migrations.add_migration(
    "2026-01-30-121720_callback",
    |conn: &mut SqliteConnection| {
        diesel::sql_query("SELECT 'EXECUTE YOUR MIGRATION HERE'").execute(conn)?;
        Ok(())
    },
);

// register a migration as function
rust_migrations.add_migration("2026-01-30-122420_function", migration_function);

// register a RustMigration
let migration = RustMigration::new(|conn| {
    diesel::sql_query("SELECT 'EXECUTE YOUR MIGRATION HERE'").execute(conn)?;
    Ok(())
})
.with_down(|conn| {
    diesel::sql_query("SELECT 'REVERT YOUR MIGRATION HERE'").execute(conn)?;
    Ok(())
});
rust_migrations.add_migration("2026-01-30-122520_rust_migration", migration);

// register a custom migration type
rust_migrations.add_migration(
    "2026-01-30-122920_custom_type",
    CustomMigration("your custom args"),
);

// run the migrations
let mut conn = SqliteConnection::establish(&connection_url)?;
conn.run_pending_migrations(rust_migrations)?;
```

# Combined migration source example

```rust


use diesel::prelude::*;
use diesel_migrations::EmbeddedMigrations;
use diesel_migrations::CombinedMigrationSource;
use crate::diesel_migrations::MigrationHarness;
use migrations_macros::embed_migrations;

pub const PG_MIGRATIONS: EmbeddedMigrations = embed_migrations!("../migrations/postgres");
pub const SQLITE_MIGRATIONS: EmbeddedMigrations = embed_migrations!("../migrations/sqlite");

// Create a new empty combined source
let mut combined_sources = CombinedMigrationSource::default();

// It's not particular meaningful to combine PostgreSQL and SQLite like this,
// but that reasonable demonstrates the API
combined_sources.add_source(PG_MIGRATIONS);
combined_sources.add_source(SQLITE_MIGRATIONS);

// run the migrations
let mut connection = SqliteConnection::establish(&connection_url)?;
let res = connection.run_pending_migrations(combined_sources);

```
---

This is currently marked as draft as it is still missing some more tests + drastically more documentation on how to use this new sources. I'm nevertheless interested in feedback on the design and the exposed interface. One question that I would like to discuss is whether we want something more typed for the migration name for the `RustMigrationSource` than the string interface that all the other sources use. Technically we could also use at least two arguments there: a required version (string?) + an optional name (as `Option<String>`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ an up to date version of the specific dependency. We check this by using the uns
 Increasing the minimal supported Rust version will always be coupled at least with a minor release.
 
 ## Unreleased
+### Added
+* Diesel-Migrations now contains a migration source that easily allows you to register Rust based migrations
+* Diesel-Migrations now contains a migration source that allows you to combine migrations from several different sources
 
 ### Added
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,6 +152,17 @@ To run rustfmt tests locally:
 You can also use rustfmt to make corrections or highlight issues in your editor.
 Check out [their README](https://github.com/rust-lang/rustfmt) for details.
 
+### Usage of LLMs and AI agents
+
+The Diesel project doesn't completely disallow to usage of LLMs and AI agents for contributions. There are still a number of restrictions and rules you as a PR author are asked to follow.
+
+Most importantly you as the author of the PR are responsible for carefully reviewing the generated code to make sure that:
+
+* It is correct to your best knowledge
+* It does not contain any copyrighted code that is incompatible with the license used by Diesel
+
+Furthermore you are asked to disclose the usage of any such tools in your PR description. 
+
 ### Common Abbreviations
 
 `ST`: Sql Type. Basically always has the `NativeSqlType` constraint
--- a/diesel/src/connection/instrumentation.rs
+++ b/diesel/src/connection/instrumentation.rs
@@ -1,7 +1,7 @@
 use alloc::boxed::Box;
 use core::fmt::{Debug, Display};
 use core::num::NonZeroU32;
-use core::ops::{Deref, DerefMut};
+use core::ops::DerefMut;
 use downcast_rs::Downcast;
 
 #[cfg(feature = "std")]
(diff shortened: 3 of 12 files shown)


isosphere:
@weiznich, I think you forgot to bump the version for diesel_migrations & publish the updated package; it was changed here.

weiznich:
@isosphere Thats expected. We generally do not bump versions and cut releases for every new feature. That only happens once in a while as new minor release. See the [2.4](https://github.com/orgs/diesel-rs/projects/1) project for a list of things that are expected to be resolved before the next minor release. Your help in doing that is certainly welcome

PR #4952, opened 2026-01-27
Implementing `diesel::result::DatabaseErrorInformation` for `core::convert::Infallible`.

LucaCappelletti94:
This PR adds an implementation of `diesel::result::DatabaseErrorInformation` for `std::convert::Infallible`.

The motivation and detailed rationale are discussed in: https://github.com/diesel-rs/diesel/discussions/4951

In short, this allows generic error types to use `Infallible` as a marker for uninhabited error branches, which is currently impossible to express due to orphan rules.

### DIFF
--- a/diesel/src/result.rs
+++ b/diesel/src/result.rs
@@ -226,7 +226,7 @@ impl DatabaseErrorInformation for String {
     }
 }
 
-impl DatabaseErrorInformation for std::convert::Infallible {
+impl DatabaseErrorInformation for core::convert::Infallible {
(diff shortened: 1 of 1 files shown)


prkbuilds (approved):
LGTM, but considering it's a gap in the runtime which was present and needs to be addressed, maybe @weiznich can give more insight on this and can this hinder any other process?

weiznich (approved):
I think that's reasonable to have. After all that impl only exists at compile time not at runtime as zero sized types get optimized away

PR #4945, opened 2026-01-23
Bump everything to edition 2024

weiznich:
This just bumps the rust edition, most of the changes are just automatic format and fix changes.

It also fixes #4941 on the side

PR #4953, opened 2026-01-27
Fix MySQL CI failure in #4945 (foreign key syntax)

LucaCappelletti94:
This PR builds on top of #4945 attempts to fix the CI failure observed in the MySQL test suite, specifically for the `print_schema_allows_none` test case.

The test `print_schema_allows_none` failed because the generated schema output for MySQL was missing `diesel::joinable!` statements that were present in the expected snapshot.

The SQL schema for the test used inline `REFERENCES` syntax:

```sql
user_id INTEGER REFERENCES users(id)
```

As noted in the [MySQL 8.0 Documentation (Section 1.6.2.3)](https://dev.mysql.com/doc/refman/8.0/en/ansi-diff-foreign-keys.html):

> MySQL parses but ignores “inline REFERENCES specifications” (as defined in the SQL standard) where the references are defined as part of the column specification. MySQL accepts REFERENCES clauses only when specified as part of a separate FOREIGN KEY specification.

Because MySQL silently ignored this syntax, **no foreign key constraints were actually created** in the database.

Diesel CLI infers relationships by querying the `information_schema`. specifically in mysql.rs. The function `load_foreign_key_constraints` relies on finding records where `constraint_type` is explicit:

https://github.com/diesel-rs/diesel/blob/70d1fc8a519ed69d5783d98592fe8d4731b82aef/diesel_cli/src/infer_schema_internals/mysql.rs#L165-L173

Since the inline syntax creates no entry in `information_schema.table_constraints`, Diesel correctly reported 0 associations, causing the snapshot mismatch against expected.snap.

I have updated [diesel_cli/tests/print_schema/print_schema_allows_none/mysql/schema.sql](https://github.com/LucaCappelletti94/diesel/blob/63f335f51ebe161cf75f87e96c57b289b56dfc55/diesel_cli/tests/print_schema/print_schema_allows_none/mysql/schema.sql) to use the explicit `FOREIGN KEY` syntax required by InnoDB:

```diff
- user_id INTEGER REFERENCES users(id)
+ user_id INTEGER, FOREIGN KEY (user_id) REFERENCES users(id)
```

This ensures the constraints are created in the database, allowing the introspection query to succeed and the `joinable!` macros to be generated as expected.


weiznich (approved):
Thanks for submitting this. That's helpful.

Turns out that is yet another painful difference between Mysql and Mariadb (which handles these constraints correctly …)

I've just pushed another minor fix on top of this, but this is now ready to be merged.

PR #4954, opened 2026-01-28
[UNSOUND] SQLite: linker-safe extension loading via `sqlite3_db_config`

LucaCappelletti94:
**Until we make this a clear comment in the code, let this be a warning: at this time, linking and loading SQLite extensions is not viable, and there are better ways than `load_extension`, like https://github.com/diesel-rs/diesel/pull/4993 I suggest that unless anything pretty major changed in Rust or the SQLite configurations that are commonly available on devices, that you leave `load_extension` support in the bin**

This PR provided a basic generic method for loading SQLite extensions in Diesel.

Extension paths are handled via the `SqliteExtension` trait as an associated `CStr` const and a single generic method on `SqliteConnection`, making it virtually impossible to feed user provided strings into the method, as it is rather dangerous being an entry point to load just about any library at runtime.

I have deliberately chosen **not** to include built-in marker structs for specific extensions (like `uuid` or `spellfix`) in the core library for now, as resolving generic paths for extensions across different operating systems remains an open issue. Different user configurations may have different paths to the compiled extensions.

In the PR I used `sqlite3_db_config(SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION)` instead of `sqlite3_enable_load_extension`. This enables extension loading only for the C-API, keeping the `load_extension()` SQL function disabled, [which is a known security problem in SQLite apparently](https://sqlite.org/c3ref/enable_load_extension.html)

The PR does not expose `enable_load_extension` and `disable_load_extension` in the public API. State management is handled internally by `load_extension`, which ensures the configuration is enabled only for the exact duration of the load operation, making it impossible for the user to leave `enable_load_extension` enabled.

```rust
use diesel::sqlite::{SqliteConnection, SqliteExtension};
use std::ffi::CStr;

// Define your extension type
struct SqliteUUIDExtension;

impl SqliteExtension for SqliteUUIDExtension {
    // The filename of the shared library (e.g., uuid.so or uuid.dll)
    const FILENAME: &'static CStr = c"uuid"; 
}

fn establish_connection(url: &str) -> SqliteConnection {
    let mut conn = SqliteConnection::establish(url).unwrap();
    
    // Safely load the extension
    // If the environment doesn't allow it, this returns a proper Result error
    conn.load_extension::<SqliteUUIDExtension>()
        .expect("Failed to load UUID extension");
        
    conn
}
```

## Dynamic Loading
Previously, in #2180, the `sqlite3_load_extension` symbol was stripped from the ABI when SQLite is compiled with `-DSQLITE_OMIT_LOAD_EXTENSION`. This prevented Diesel from linking against system libraries on many platforms.

This PR uses dynamic symbol resolution at runtime:
* **Unix**: Uses `libc::dlsym` (with `RTLD_DEFAULT`) to find the symbol in the loaded process image.
* **Windows**: Uses `GetModuleHandle` and `GetProcAddress` to locate the symbol in `sqlite3.dll` or `winsqlite3.dll`.

If the symbol is missing (restricted build), the method gracefully returns a `DatabaseError` mimicking SQLite's "no such function" error, rather than crashing or preventing compilation.

## Infrastructure & Verification

In the PR I have introduced tests to prove this works in both "restricted" and "full" environments on both **Linux** and **Windows**.

### New CI Job: `sqlite_restricted_build`
* **Script**: `test_sqlite_no_extensions.(sh|ps1)`
* **Action**: Compiles SQLite from source with `-DSQLITE_OMIT_LOAD_EXTENSION`.
* **Verification**: Ensures Diesel compiles and runs tests. Use of `load_extension` returns a handled error ("no such function") instead of a segmentation fault or link error.

### New CI Job: `sqlite_extension_build`
* **Script**: `test_sqlite_with_extensions.(sh|ps1)`
* **Action**:
    1. Compiles SQLite from source with extensions enabled.
    2. Manually compiles common extensions (`uuid.c`, `extension-functions.c`, etc.) into `.so` / `.dll` files.
    3. Sets up `LD_LIBRARY_PATH` / `PATH`.
* **Verification**: Checks that extensions are successfully loaded and usable in SQL queries.

**I tried my best for the windows support, but I have no windows machine locally and my windows development experience stopped in high school, so feel free to help out**

Closes issue #1867


diesel_tests/tests/load_extension.rs
```diff
@@ -0,0 +1,91 @@
...
+        assert_eq!(&format!("{}", &result), "not authorized");
+    } else {
+        assert_eq!(&format!("{}", &result), "no such function: load_extension");
+    }
+
+    // enable extension loading
+    conn.enable_load_extension().unwrap();
+
+    // try loading module without enabling extension loading
+    let result = sql_query("SELECT load_extension('/tmp/foo');")
```
Mingun: Copy-paste of comment from previous check

LucaCappelletti94: Hi @Mingun, this PR is a draft and does not work at this time. Please refrain from reviewing it, as I am making extensive changes to it.

LucaCappelletti94: @Mingun a tentatively first working version is ready for review when you happen to have time. Thank you.

.github/workflows/ci.yml
```diff
@@ -555,3 +582,77 @@ jobs:
...
+          key: restricted-sqlite-${{ matrix.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@f7ccc83f9ed1e5b9c81d8a67d7ad1a747e22a561 # master
+        with:
+          toolchain: stable
+
+      - name: Setup MSVC (Windows)
+        if: runner.os == 'Windows'
+        uses: ilammy/msvc-dev-cmd@v1
```
weiznich, line 611: Needs to be pinned

.github/workflows/ci.yml
```diff
@@ -555,3 +582,77 @@ jobs:
...
+          key: extension-sqlite-${{ matrix.os }}-cargo-${{ hashFiles('**/Cargo.toml') }}
+
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@f7ccc83f9ed1e5b9c81d8a67d7ad1a747e22a561 # master
+        with:
+          toolchain: stable
+
+      - name: Setup MSVC (Windows)
+        if: runner.os == 'Windows'
+        uses: ilammy/msvc-dev-cmd@v1
```
weiznich, line 648: Needs to be pinned

.github/workflows/ci.yml
```diff
@@ -555,3 +582,77 @@ jobs:
...
           cargo xtask semver-checks
+
+  sqlite_restricted_build:
+    name: SQLite Restricted Build
+    runs-on: ${{ matrix.os }}
+    needs: [rustfmt_and_clippy]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-2025]
```
weiznich, line 593: We also should check this on macos, possibly even both the aarch64 and x86_64 variant

.github/workflows/ci.yml
```diff
@@ -555,3 +582,77 @@ jobs:
...
+        run: .github/scripts/test_sqlite_no_extensions.ps1
+
+  sqlite_extension_build:
+    name: SQLite Extension Build
+    runs-on: ${{ matrix.os }}
+    needs: [rustfmt_and_clippy]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-2025]
```
weiznich, line 630: We also should check this on macos, possibly even both the aarch64 and x86_64 variant

diesel/src/sqlite/connection/mod.rs
```diff
@@ -374,6 +405,75 @@ impl SqliteConnection {
...
+    ///     Users cannot execute arbitrary code via SQL injection attacks.
+    /// 3.  **Type-Safe Allowlist**: The generic parameter `E` must implement the [`SqliteExtension`] trait.
+    ///     This trait acts as a compile-time allowlist (or registry) of known, trusted extensions.
+    ///     Since the extension filename is defined as a constant in the trait implementation, no
+    ///     arbitrary runtime strings are passed to the loading mechanism.
+    ///
+    /// # Linking and Portability
+    ///
+    /// This method uses dynamic symbol lookup (using `dlsym` on Unix-like systems and `GetProcAddress` on Windows)
+    /// to locate the `sqlite3_load_extension` helper function within the loaded SQLite library.
```
weiznich, line 437: From my previous tries to implement this feature, this strategy has the huge disadvantage that it won't work in a lot of otherwise valid cases:

* musl doesn't have dlsym and dlopen 
* the symbols are likely optimized away for builds using a static linked libsqlite3 version

Likely the proper way to implement this is to have some sort of guard/check in libsqlite3-sys that sets a `DEP_LIBSQLITE3-SYS_LOAD_EXTENSION_ENABLED` flag for downstream build scripts to consume. This can then be used diesel to set a internal cfg to correctly link this function or provide a stub that returns an error. This will require some detection of that symbol in libsqlite3-sys at compile time.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts (search for `DEP_<links>` for the documentation on this particular cargo feature)

The other alternative would be to use some sort of weak linkage for these symbols, but at least the last time I checked that it was not possible in stable rust. (seems to be still not stable: https://github.com/rust-lang/rust/issues/29603)

A third alternative would be to somewhat abstract that decision away and require the user to solve how to find the relevant symbols. By that I mean having another generic argument for the linking strategy and provide two (sealed) implementations: One for static linkage and one for dlsym/dlopen. The linker should be able to remove the unused variants so we shouldn't end up in situations where an unexpected symbol ends up in the final build.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -261,6 +262,193 @@ impl RawConnection {
...
+
+    pub(super) fn load_extension<E: SqliteExtension>(&self) -> QueryResult<()> {
+        unsafe {
+            use std::mem;
+
+            // Implementation Detail: Dynamic Loading of sqlite3_load_extension
+            //
+            // We cannot link directly to `sqlite3_load_extension` because it is not exported
+            // in SQLite builds compiled with `-DSQLITE_OMIT_LOAD_EXTENSION` (common on Android,
+            // iOS, and embedded systems). Direct linking would cause build failures on these targets.
```
weiznich, line 274: Most notably macos as well

diesel/src/sqlite/connection/raw.rs
```diff
@@ -261,6 +262,193 @@ impl RawConnection {
...
+            //    we propagate the error message returned by SQLite.
+
+            // On Unix systems, we search for symbols in the already loaded libraries (RTLD_DEFAULT).
+            // This relies on libsqlite3 being loaded in the process address space.
+            #[cfg(unix)]
+            let symbol =
+                { ::libc::dlsym(::libc::RTLD_DEFAULT, c"sqlite3_load_extension".as_ptr()) };
+
+            // On Windows, looking up symbols is more involved as we need a module handle.
+            #[cfg(target_os = "windows")]
```
weiznich, line 298: We should hopefully never end up with a case where `cfg(unix)` and `target_os = "windows"` are both true, but I rather would like to be safe and make sure we don't have both variants in here via an if/else if variant

.github/scripts/test_sqlite_no_extensions.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [ ! -d "${SQLITE_SOURCE}" ]; then
+    echo "Extracting SQLite..."
+    tar xzf "${SQLITE_TAR}"
+fi
+
+cd "${SQLITE_SOURCE}"
+
+# Configure with extension loading disabled
+if [ ! -f "Makefile" ]; then
```
weiznich, line 38: Shouldn't we always run `./configure` just to make sure we don't end up in a situation using the other build?

.github/scripts/test_sqlite_no_extensions.sh
weiznich, line 1: I think much of this script can be shared with the `test_sqlite_with_extension.sh` variant. After all only the build flags + the extension building differs. The download, extraction and other building flow should be the same.

Also: I understand why this is required, but having this adds quite a lot of work for future maintenance as we now have 4 additional scripts to deal with.

.github/scripts/test_sqlite_with_extensions.sh
```diff
@@ -0,0 +1,114 @@
...
+if [ -f "extension-functions.c" ]; then
+    gcc -g -fPIC -shared extension-functions.c -o "${INSTALL_DIR}/lib/extension-functions.so" -lm
+fi
+
+# ----------------------------------------------------------------
+# Create symlinks for common variations of library names
+# ----------------------------------------------------------------
+# Some systems/loaders might look for .so, others might want different names.
+# Ensure 'spellfix1' is findable.
+cd "${INSTALL_DIR}/lib"
```
weiznich, line 82: It's not really clear for me where these `spellfix` libraries come from and for what they are required.

.github/scripts/test_sqlite_with_extensions.sh
```diff
@@ -0,0 +1,114 @@
...
+echo "Verifying custom sqlite3 version..."
+"${INSTALL_DIR}/bin/sqlite3" --version
+
+echo "----------------------------------------------------------------"
+echo "Running compilation check..."
+
+# Run cargo test
+# We set DIESEL_TEST_SQLITE_EXTENSIONS_DISABLED=0 to tell the test suite 
+# that we expect successful loading (or "not authorized" if disabled at runtime but present)
+export DIESEL_TEST_SQLITE_EXTENSIONS_DISABLED=0
```
weiznich, line 100: I would rather not have the test suite to change results dependent on environment variables like this. That makes it hard for contributors to understand what's going on and why things suddenly fail on CI

.github/scripts/test_sqlite_with_extensions.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# Clean first to ensure we link against the new library
+cargo clean
+
+echo "Running tests against standard SQLite..."
+# Note: --nocapture allows us to see output
+if cargo test --package diesel_tests --test integration_tests load_extension --features sqlite -- --nocapture; then
```
weiznich, line 107: Calling the tests like that should rather be part of our xtask crate, That would also allow using the same "script" for more than one platform

weiznich (commented):
Thanks for picking the work up on this. This is still a tricky topic and I'm still not sure if the proposed solution is the correct one given the disadvantages it has. 

Otherwise I added quite a lot of "minor" things that would need to be changed if this approach is chosen. Before fixing these I would prefer deciding on the actual solution first.

LucaCappelletti94:
> Thanks for picking the work up on this. This is still a tricky topic and I'm still not sure if the proposed solution is the correct one given the disadvantages it has.

Yes, objectively I started out hoping that with the `sqlite3_db_config` it was going to be an "easy" PR because of how [it was documented in SQLite](https://sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenableloadextension), but it ended up as a messy dynamic linking thing. I am also really not satisfied with the current state.

I hope I have somewhat pushed the boulder higher up the hill, but I definitely agree the current solution does not really cut it as a solid solution.

weiznich:
Yes that's essentially what I concluded years ago as I stopped working on this particular feature. There just doesn't seem to be a good solution. I'm also not sure if it's really a relevant feature to have or if diesel might be fine without it given these constraints.

LucaCappelletti94:
One possibility that remains open is to allow the user access to the underlying raw connection, maybe with an unsafe method, and then they can do all their unsafe heart wishes with it I guess.

weiznich:
Yes, that would be a possibility. I avoided doing that in the past for the following reasons:

* It's obviously highly unsafe as you mess with the internal connection state. Alone documenting what you can and cannot do would be a major effort.
* It exposes the `libsqlite3_sys::sqlite3` type via the public API. That type comes from a third party crate with a version below 1.0 so it would be a potential stability hazard.  It's not a show stopper but another stability hazard. On the other hand we could also just return a `*mut c_void` pointer and expect the user to cast it to their sqlite type :thinking:

LucaCappelletti94:
For the specific case of [`sqlite-wasm-rs`](https://github.com/Spxg/sqlite-wasm-rs), I have now extension loading working with the currently published diesel crate.

See example with uuid4/7 here: https://github.com/LucaCappelletti94/sqlite-wasm-uuid-rs/blob/main/test-diesel/src/lib.rs

weiznich:
> See example with uuid4/7 here: https://github.com/LucaCappelletti94/sqlite-wasm-uuid-rs/blob/main/test-diesel/src/lib.rs

For this particular use-case I would like to point to the interface for registering functions with diesel: https://docs.diesel.rs/2.3.x/diesel/expression/functions/attr.declare_sql_function.html#use-with-sqlite

(That doesn't require an extension)

LucaCappelletti94:
> For this particular use-case I would like to point to the interface for registering functions with diesel: https://docs.diesel.rs/2.3.x/diesel/expression/functions/attr.declare_sql_function.html#use-with-sqlite
> 
> (That doesn't require an extension)

In my use case, I need to use the `uuid` function in schema because it involves TRIGGERs. Does the `declare_sql_function` work only in queries or does it make available generally in the SQL?

LucaCappelletti94:
I have added a line in the README of the extension crate stating: 

> Do note that if you are using [`diesel`](https://docs.diesel.rs/2.3.x/diesel/index.html) and only need to use these functions inside a query, and not in schema definitions (e.g., as default values), you can avoid using this extension altogether by simply using [`declare_sql_function`](https://docs.diesel.rs/2.3.x/diesel/expression/functions/attr.declare_sql_function.html#use-with-sqlite) to map the Rust functions you need.

I hope that is correct

LucaCappelletti94:
> > See example with uuid4/7 here: https://github.com/LucaCappelletti94/sqlite-wasm-uuid-rs/blob/main/test-diesel/src/lib.rs
> 
> For this particular use-case I would like to point to the interface for registering functions with diesel: https://docs.diesel.rs/2.3.x/diesel/expression/functions/attr.declare_sql_function.html#use-with-sqlite
> 
> (That doesn't require an extension)

Ok, for all my use cases it turns out that this was quite enough.

LucaCappelletti94:
@weiznich since this PR is most likely not going anywhere, should I close it or switch it to draft?

weiznich:
@LucaCappelletti94 It might be meaningful to change it on such a way that the conclusion of this is in the source code as comment. This would hopefully stop the next person trying the same things again after you and me already spend quite a bit of time figuring out that it won't work.

LucaCappelletti94:
I will make a different PR adding that comment in the code as a post-mortem so that this one can be closed and remain as it is, an historical artifact of the attempt, if you agree.

PR #4961, opened 2026-02-03
Add support for feature-gated columns in `table!` macro

LucaCappelletti94:
This PR adds support for placing `#[cfg(...)]` attributes on columns within the `table!` macro. This allows users to conditionally include columns based on feature flags, which is useful when certain columns are variable, like when the underlying schema changes with certain versions and we want to support several versions (today's PR triggering example for me was Postgres [18](https://www.postgresql.org/docs/18/view-pg-backend-memory-contexts.html) and [17](https://www.postgresql.org/docs/17/view-pg-backend-memory-contexts.html)).

Resolves discussion https://github.com/diesel-rs/diesel/discussions/4922

## Example Usage

```rust
table! {
    users {
        id -> Integer,
        name -> Text,
        #[cfg(feature = "chrono")]
        created_at -> Timestamp,
        #[cfg(feature = "uuid")]
        user_uuid -> Uuid,
    }
}
```

## Implementation Details

Rust doesn't allow `#[cfg]` attributes on tuple type elements directly. This means we can't simply write:

```rust
// This doesn't work!
pub type SqlType = (
    Integer,
    Text,
    #[cfg(feature = "chrono")]
    Timestamp,
);
```

For `n` distinct cfg groups, we generate `2^n` variants of all aggregate types, each guarded by the appropriate combination of `#[cfg(all(...))]` conditions:

```rust
#[cfg(all(not(feature = "chrono"), not(feature = "uuid")))]
pub const all_columns: (id, name) = (id, name);

#[cfg(all(feature = "chrono", not(feature = "uuid")))]
pub const all_columns: (id, name, created_at) = (id, name, created_at);

#[cfg(all(not(feature = "chrono"), feature = "uuid"))]
pub const all_columns: (id, name, user_uuid) = (id, name, user_uuid);

#[cfg(all(feature = "chrono", feature = "uuid"))]
pub const all_columns: (id, name, created_at, user_uuid) = (id, name, created_at, user_uuid);
```

The following are generated with combinatorial variants:

- `all_columns` constant
- `SqlType` type alias  
- `impl diesel::Table for table` (specifically `AllColumns` associated type)
- `impl ValidGrouping<__GB> for star`

All trait implementations for individual columns (e.g., `Expression`, `QueryFragment`, `SelectableExpression`, `AppearsOnTable`, operator impls) are wrapped with the column's `#[cfg(...)]` attributes so they only exist when the feature is enabled.

The `IsContainedInGroupBy` cross-product impls between column pairs include cfg attributes from both columns.

## New Helper Functions

Added and documented:

- `CfgGroup` - Groups columns by their cfg condition
- `collect_cfg_groups()` - Organizes columns into cfg groups
- `generate_combined_cfg_condition()` - Creates `#[cfg(all(...))]` with enabled/disabled conditions
- `generate_aggregate_variants()` - Generates 2^n variants of `all_columns`, `SqlType`, `ValidGrouping`
- `generate_kind_specific_impls_variants()` - Generates 2^n variants of `Table`/`QueryRelation` impls

## Tests Added

- Snapshot tests for single and multiple feature-gated columns
- Type-checking tests that verify the generated code contains:
  - Correct cfg guards for all combinations
  - Correct column tuples in `all_columns` for each variant
  - Correct types in `SqlType` for each variant
  - Correct `AllColumns` associated type for each variant
  - Correct `ValidGrouping` bounds for `star`

## Limitations

With many distinct cfg groups, (feature-gated) code size grows exponentially (2^n). This is an inherent limitation of Rust's cfg system. In practice, most tables will have at most 1-2 distinct feature groups, keeping the generated code manageable.


LucaCappelletti94:
This PR should be ready for review. 

The Windows 2025 + PostgreSQL CI failure is unrelated to these changes - it's a linker error (`LNK1181: cannot open input file 'libpq.lib'`) indicating the PostgreSQL client library wasn't found on the CI runner. The PR changes only modify proc-macro code generation and don't affect native library linking. This is likely to be a transient CI infrastructure issue.

LucaCappelletti94:
I have just slammed against the case of [`pg_catalog.pg_auth_members (oid)`](https://www.postgresql.org/docs/18/catalog-pg-auth-members.html) vs [`pg_catalog.pg_auth_members (roleid, member)`](https://www.postgresql.org/docs/15/catalog-pg-auth-members.html).

I guess that while it is somewhat reasonable to feature gate the single columns, I suppose that if the primary key changes such as in this case, one really needs to feature gate the entire table.

Do you have any better idea?

diesel_derives/src/table.rs
```diff
@@ -5,6 +5,467 @@ use syn::parse_quote;
...
+    } else {
+        quote::quote! {
+            #[cfg(all(#(#conditions),*))]
+        }
+    }
+}
+
+/// Generates all combinatorial variants of aggregate types for feature-gated columns.
+///
+/// This function generates 2^n variants of `all_columns`, `AllColumns`, and `SqlType`
```
weiznich: I would rather like to avoid generating an exponential amount of instances to any user input. The cost of compiling large schemas is already large enough, so I rather do not want to introduce another feature to make this even worse. Either we find a different solution to the underlying problem or we need to accept that certain things require a bit more code (e.g. require the user to put a cfg on the table itself and not only on a specific column.)

For reference there is an open RFC to possible allow putting `#[cfg]` on tuple types like required to solve this in a better way: https://github.com/rust-lang/rfcs/pull/3532 There are no concerns listed there and there seems to be a slight push to accept that (at least a few months back). So it might be worth to rather push for accepting this RFC and then just implementing the relevant feature.
Other than that, given that this is already allowed for tuple expressions it should be rather simple to get the lang team to also accept it for types. Also it shouldn't be too hard to stabilize this, as there is not much that could go wrong there as far as I can tell.

diesel_derives/src/table.rs
```diff
@@ -5,6 +5,467 @@ use syn::parse_quote;
...
+/// A `TokenStream` containing the combined `#[cfg(all(...))]` attribute, or empty if no groups
+fn generate_combined_cfg_condition(groups: &[CfgGroup<'_>], enabled_mask: usize) -> TokenStream {
+    if groups.is_empty() {
+        return TokenStream::new();
+    }
+
+    let mut conditions: Vec<TokenStream> = Vec::new();
+
+    for (i, group) in groups.iter().enumerate() {
+        let is_enabled = (enabled_mask & (1 << i)) != 0;
```
weiznich: It's probably not worth to use a bitmask here compared to just a `Vec<boo>`. The later is better readable and the code shouldn't be that performance critical to need to safe that few bytes of memory.

diesel_derives/src/table.rs
```diff
@@ -5,6 +5,467 @@ use syn::parse_quote;
...
+                    type AllColumns = (#(#column_names,)*);
+
+                    fn primary_key(&self) -> Self::PrimaryKey {
+                        #primary_key
+                    }
+
+                    fn all_columns() -> Self::AllColumns {
+                        (#(#column_names,)*)
+                    }
+                }
```
weiznich: There is no reason to duplicate al of that. Only the two associated types need to be put behind a cfg flag. For tuple expressions it's allowed to put a field behind a cfg flag.

weiznich (changes requested):
I'm not really in favour introducing a feature that allows users to generate significantly more code in the expansion of a table macro like this.

I've left a few remarks how that could be stripped down + what I would rather like to prefer doing instead even if that's blocked on future language development.

This is not a full review yet, just a dump of related thoughs

LucaCappelletti94:
To illustrate concretely how I am using this, here are a few `table!` macro calls which without this new functionality were rather unmaintainable:

* https://github.com/earth-metabolome-initiative/pg_diesel/blob/main/src/schema/pg_catalog/pg_database.rs
* https://github.com/earth-metabolome-initiative/pg_diesel/blob/main/src/schema/pg_catalog/pg_stat_bgwriter.rs
* https://github.com/earth-metabolome-initiative/pg_diesel/blob/main/src/schema/pg_catalog/pg_replication_slots.rs

I didn't notice a particular increase to the (already rather long) compile times, and this schema I think is rather large for my standards.

While I agree that it is always desirable to avoid increase of compile time imposed by default in general, here I feel it is more of a pay-as-you-use-it type of feature. Tables without feature gates should not experience any added cost, and the code that a person needs to edit is exponentially less (before it was 2^n where n is the number of feature gates).

Of course, it is very regrettable that the language at this time does not allow for a cleaner solution.

I will apply the suggested changes and see if I can further trim it down.

diesel_derives/src/table.rs
LucaCappelletti94: I waited a while for the RFC to be merged, which would change some of the PR and be much preferable, but it seems stuck in limbo. I pinged the developers tagged to judge the RFC but none replied. I will now do the requested changes to this PR, and hopefully we can switch to the cleaner option when it finally becomes concrete. I agree that exponential code generation is undesirable, but it is also de-facto what you find yourself obliged to do when you want to write these combinatorial schemas. The only change made by this PR is, in the end, just code cleanliness and is paid only by who actually uses feature gates on the columns.

diesel_derives/src/table.rs
LucaCappelletti94: Done!

diesel_derives/src/table.rs
LucaCappelletti94: Done!

diesel_derives/src/tests/snapshots/diesel_derives__tests__table_with_column_feature_gate (postgres).snap
```diff
@@ -0,0 +1,1264 @@
...
+        pub fn star(&self) -> star {
+            star
+        }
+    }
+    #[cfg(all(not(feature = "chrono")))]
+    #[doc = concat!("The SQL type of all of the columns on this ", "table")]
+    pub type SqlType = (Integer, Text);
+    #[cfg(all(feature = "chrono"))]
+    #[doc = concat!("The SQL type of all of the columns on this ", "table")]
+    pub type SqlType = (Integer, Text, Timestamp);
```
weiznich: There is no need to generate different variants here. We can just express this as 


```suggestion
    #[doc = concat!("The SQL type of all of the columns on this ", "table")]
    pub type SqlType = <AllColumns as diesel::Expression>::SqlType;
```

instead

diesel_derives/src/table.rs
```diff
@@ -544,12 +779,7 @@ fn expand(input: TableDecl, kind: QuerySourceMacroKind) -> TokenStream {
                 /// being used that way
                 pub struct star;
 
-                impl<__GB> diesel::expression::ValidGrouping<__GB> for star
-                where
-                    (#(#column_names,)*): diesel::expression::ValidGrouping<__GB>,
-                {
-                    type IsAggregate = <(#(#column_names,)*) as diesel::expression::ValidGrouping<__GB>>::IsAggregate;
-                }
+                #valid_grouping_star
```
weiznich: It shouldn't be necessary to construct this dynamically, as we can use the `AllColumns` type here

diesel_derives/src/table.rs
```diff
@@ -377,8 +611,9 @@ fn expand(input: TableDecl, kind: QuerySourceMacroKind) -> TokenStream {
                 }
             }
 
-            #[doc = concat!("The SQL type of all of the columns on this ", #kind_name)]
-            pub type SqlType = (#(#column_ty,)*);
+            #sql_type_variants
```
weiznich:
```suggestion
            #[doc = concat!("The SQL type of all of the columns on this ", #kind_name)]
            pub type SqlType = <AllColumns as diesel::Expression>::SqlType;
```

diesel_derives/src/table.rs
```diff
@@ -4,6 +4,298 @@ use syn::{Ident, parse_quote};
...
+    }
+}
+
+fn generate_kind_specific_impls<'a>(
+    non_gated_columns: &[&'a ColumnDef],
+    cfg_groups: &[CfgGroup<'a>],
+    primary_key: &Option<TokenStream>,
+    kind: QuerySourceMacroKind,
+) -> TokenStream {
+    let tuple_expr = all_columns_tuple_expr(non_gated_columns, cfg_groups);
```
weiznich: Maybe rather use the `all_columns` constant instead?

diesel_derives/src/table.rs
```diff
@@ -4,6 +4,298 @@ use syn::{Ident, parse_quote};
...
+
+    let valid_grouping_star = quote::quote! {
+        impl<__GB> diesel::expression::ValidGrouping<__GB> for star
+        where
+            super::AllColumns: diesel::expression::ValidGrouping<__GB>,
+        {
+            type IsAggregate =
+                <super::AllColumns as diesel::expression::ValidGrouping<__GB>>::IsAggregate;
+        }
+    };
```
weiznich:
```suggestion
```

diesel_derives/src/table.rs
```diff
@@ -4,6 +4,298 @@ use syn::{Ident, parse_quote};
...
+                    column_types.push(&col.tpe);
+                }
+            }
+        }
+
+        sql_type_variants.push(quote::quote! {
+            #cfg_condition
+            #[doc = concat!("The SQL type of all of the columns on this ", #kind_name)]
+            pub type SqlType = (#(#column_types,)*);
+        });
```
weiznich:
```suggestion

```

diesel_derives/src/table.rs
```diff
@@ -4,6 +4,298 @@ use syn::{Ident, parse_quote};
...
+    debug_assert_eq!(groups.len(), enabled_flags.len());
+
+    if groups.is_empty() {
+        return TokenStream::new();
+    }
+
+    let mut conditions: Vec<TokenStream> = Vec::new();
+
+    for (i, group) in groups.iter().enumerate() {
+        let is_enabled = enabled_flags[i];
```
weiznich: Rather use `zip` with the `enabled_flags` slice?

diesel_derives/src/table.rs
```diff
@@ -4,6 +4,298 @@ use syn::{Ident, parse_quote};
...
+    columns: Vec<&'a ColumnDef>,
+}
+
+/// Yields `2^n` flag vectors of length `n`. `n == 0` yields one empty vector,
+/// so callers can handle the no-cfg-groups case without a special branch.
+fn cfg_combinations(n: usize) -> impl Iterator<Item = Vec<bool>> {
+    (0..1usize << n).map(move |mask| (0..n).map(|i| (mask >> i) & 1 == 1).collect())
+}
+
+fn generate_combined_cfg_condition(groups: &[CfgGroup<'_>], enabled_flags: &[bool]) -> TokenStream {
```
weiznich, line 29: It might be good to have some small unit tests for these two functions

weiznich (changes requested):
I spend some time thinking again about this.

I came to the conclusion that it is reasonable to merge this even if the RFC is still up in the air. If the RFC gets accepted and implemented we can switch the implementation easily over to the new syntax, if it doesn't that is the best solution, so we might just start using this "now".

I still have a bunch of comments how to generate less code by reusing constants and some other nits.

Otherwise I also would like to see this feature mentioned in the `table!` documentation (in `diesel_derives/src/libs.rs`) and also in the Changelog.

diesel_derives/src/tests/snapshots/diesel_derives__tests__table_with_column_feature_gate (postgres).snap
LucaCappelletti94: `SqlType` is now a single `pub type SqlType = <AllColumns as diesel::Expression>::SqlType;`, so it is no longer enumerated per feature combination. Only the `AllColumns` type alias still has one variant per combination.

diesel_derives/src/table.rs
LucaCappelletti94: The `ValidGrouping` impl for `star` is now written inline and projects through `super::AllColumns`, so it is a single static impl rather than a token field.

diesel_derives/src/table.rs
LucaCappelletti94: The generated `all_columns()` bodies for both `Table` and `QueryRelation` now return the `all_columns` constant instead of rebuilding the tuple expression.

diesel_derives/src/table.rs
LucaCappelletti94: Added a `helper_tests` module covering `cfg_combinations` (zero groups yields a single empty vector, and full enumeration for n = 2) and `generate_combined_cfg_condition` (no groups, single attribute negation, and the multiple attribute case). The last one also pins the fix described in the summary.

weiznich (approved):
Looks good now, thanks for the update.

I will take care of the CI failures.

Issue #4958, opened 2026-01-30
Add support for multi-dimensional Postgres Arrays


weiznich:
### Backend

PostgreSQL

### What do you want to add to diesel?

Hi, is there a plan to add support for multi-dimensional arrays in diesel? I took a bit of a deep dive into diesels array implementation today as well as the Postgres byte encoding for array types and I thought I might give it a go, starting with 2D-arrays first.

Since Postgres defines MAXDIM=6 for arrays, I guess it would technically be possible to add support for all possible dimensions, but I feel like adding 2D support first would be a simpler approach.

I might totally have a blind spot here and underestimate how hard this is or miss some side effects this would cause. But after my research today I'll probably give it a go just out of curiosity anyways. If you are interested in the results, let me know.

### Implementation notes

Implement `FromSql<Array<ST>, Pg> for Vec<Vec<T>>` for arrays with ndim=2 and the equivalent `ToSql` trait.
In case of a 2D array, the array header would include two additional `i32` fields before the actual array payload, containing `num_elements` and `lower_bound` for the 2nd dimension. With that information one can loop over the array values accordingly and parse them into a 2D vector of size ndim1 x ndim2.

I haven't looked at the ToSql implementation as closely yet, but I assume taking a slice of a slice one has to construct the correct array header and then flatten the values.

### Additional details

I'd keep the current assumption of `lower_bound=1` for all dimensions.

### Checklist

- [X] I have already looked over the [issue tracker](https://github.com/diesel-rs/diesel/issues) and the [discussion forum](https://github.com/diesel-rs/diesel/discussions) for similar possible closed issues.
</div>

valkrypton:
Hey I'd like to give this a try. Can i make a draft PR and then we'll go from there?

weiznich:
@HoodyJazz already expressed interest in working on this issue in https://github.com/diesel-rs/diesel/discussions/4957

I should have assigned them earlier.

RakshitaReddy21:
Hi @weiznich,Is this still open for contribution?

PR #4962, opened 2026-02-04
Feature: support decoding of two-dimensional arrays

HoodyJazz:
Related Issue" https://github.com/diesel-rs/diesel/issues/4958

### Summary
Adds initial support for decoding 2-dimensional Postgres arrays into `NdArray`.

### Scope
- Decode only (FromSql)
- 2D arrays only

### Testing
- Added decoding tests for multiple element types and shapes
- Added tests for arrays containing NULL elements as well as the whole array being NULL
- Tested with `cargo xtask run-tests postgres`

### Comments
This is a very minimal first version of this implementation, as I thought smaller increments might be the way to go. I see a number of possible next steps here:
- adding serialization `ToSql`
- adding dynamic decoding of arrays with more than two dimensions
- adding helper functions for `NdArray`, such as a `get(&self, idx: &[usize])` to fetch elements with `array.get([i,j])`
- adding a type wrapper `pub struct Vec2<T>(pub Vec<Vec<Option<T>>>)` and a `into_vec2()` method
- implement additional tests

I already started implementing some of those things, but thought it would be helpful to get feedback first. I'm also not sure that my tests are exhaustive yet.

### DIFF
--- a/diesel/src/pg/types/array.rs
+++ b/diesel/src/pg/types/array.rs
@@ -15,7 +15,15 @@ use crate::sql_types::{Array, HasSqlType, Nullable};
 /// representation with the dimension information encoded in the header. This struct represents a
 /// multi-dimensional array with elements of type `T` as opposed to Vec<T> which can be used for 1d-arrays.
 pub struct NdArray<T> {
+    /// A list that describes how many values for each dimension are returned
     pub dims: Vec<usize>,
+    /// The actual data flattened to a single array
+    ///
+    /// This array contains values ordered by the left most dimension
+    /// which means there will be dim[0] values for the first element of the second dimension
+    /// followed by dim[0] values for the second element of the second dimensions
+    /// and so up to dim[1] times. Afterwards that number of values is repeated for dim[2],
+    /// and so for all dimensions in the dimension field above
     pub data: Vec<T>,
 }
 
@@ -87,26 +95,21 @@ where
             });
         }
 
-        if num_dimensions == 1 {
-            return Err("trying to deserialize one-dimensional postgres array into NdArray<T>, use Vec<T> instead".into());
-        }
-
         let num_dims: usize = num_dimensions
             .try_into()
             .map_err(|_| "number of dimensions must be positive")?;
 
-        let mut dims = Vec::with_capacity(num_dims);
-        let mut num_elements: i32;
+        let dims = (0..num_dims)
+            .map(|_| {
+                let num_elements = bytes.read_i32::<NetworkEndian>()?;
+                let _lower_bound = bytes.read_i32::<NetworkEndian>()?;
 
-        for _ in 0..num_dims {
-            num_elements = bytes.read_i32::<NetworkEndian>()?;
-            let _lower_bound = bytes.read_i32::<NetworkEndian>()?;
-
-            let dim: usize = num_elements
-                .try_into()
-                .map_err(|_| "array dimension length must be positive")?;
-            dims.push(dim);
-        }
+                let dim: usize = num_elements
+                    .try_into()
+                    .map_err(|_| "array dimension length must be positive")?;
+                Ok(dim)
+            })
+            .collect::<deserialize::Result<Vec<_>>>()?;
 
         let data = (0..dims.iter().product::<usize>())
             .map(|_| -> deserialize::Result<T> {
--- a/diesel_tests/tests/types.rs
+++ b/diesel_tests/tests/types.rs
@@ -1861,7 +1861,7 @@ fn citext_fields() {
 
 #[diesel_test_helper::test]
 #[cfg(feature = "postgres")]
-fn deserialize_wrong_dimension_pg_array_into_ndarray_gives_error() {
+fn deserialize_1d_array_to_ndarray() {
     use diesel::data_types::NdArray;
     use diesel::sql_types::{Array, Bool};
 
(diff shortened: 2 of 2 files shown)


diesel/src/pg/types/array.rs
```diff
@@ -58,6 +69,68 @@ where
...
+            return Err("trying to deserialize one-dimensional postgres array into NdArray<T>, use Vec<T> instead".into());
+        }
+
+        if num_dimensions != 2 {
+            return Err(
+                "currently only two-dimensional arrays are supported for NdArray<T>".into(),
+            );
+        }
+
+        // This is very hardcoded for two-dimensional arrays for now but can easily be adapted to be dynamic later
```
weiznich: I would welcome generalizing this to more dimensions before landing this PR. From what I can see here that shouldn't be too hard as we only need to:

* read the elements_per_dimension + bound tuple for each dimension in a loop
* convert the elements_per_dimension in the same loop
* Produce the ndims vector out of that.
* Use the [https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.product] iterator extension to get the actual number of elements to read
* Read the elements as already done in a flatten manner

diesel/src/pg/types/array.rs
```diff
@@ -2,12 +2,23 @@ use byteorder::{NetworkEndian, ReadBytesExt, WriteBytesExt};
...
 use crate::serialize::{self, IsNull, Output, ToSql};
 use crate::sql_types::{Array, HasSqlType, Nullable};
 
+#[cfg(feature = "postgres_backend")]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, AsExpression, FromSqlRow)]
+#[diesel(sql_type = Array<T>)]
+/// Postgres allows multi-dimensional arrays of at most 6 dimensions. Internally they are stored as a flattened
+/// representation with the dimension information encoded in the header. This struct represents a
+/// multi-dimensional array with elements of type `T` as opposed to Vec<T> which can be used for 1d-arrays.
+pub struct NdArray<T> {
```
weiznich, line 17: As a future extension outside of this PR we can provide some helper methods to just turn this datastructure into a 2/3/4d vector. 

(That doesn't need to be part of the initial Pr, I just try to sketch up a possible API)

diesel/src/pg/types/array.rs
HoodyJazz: I'm happy to do that, it might juts take a bit since I don't have much time right now to work on this. But latest next week I do.

diesel/src/pg/types/array.rs
HoodyJazz: Perfect, I already have some code for that as well.

diesel/src/pg/types/array.rs
HoodyJazz: Just pushed this

diesel/src/pg/types/array.rs
HoodyJazz: btw, when I ran `cargo xtask tidy` before pushing earlier, I didn't get any of the problems that popped up in the CI later (https://github.com/diesel-rs/diesel/actions/runs/21865907707/job/63106954736?pr=4962), but got `All checks were successful. Ready to submit the code!`. This also happened the first time I submitted the PR. Just want to let you know, in case this isn't intentional.

HoodyJazz: Hi again,
I pushed some changes implementing Wrapper types to a different branch. It's not ready for full review yet (some commented out code still there), but I thought maybe you're interested, because potentially it could already be included in this PR as well.
https://github.com/HoodyJazz/diesel/blob/feature/add-narray-wrappers/diesel/src/pg/types/array.rs
**Features:**
- `get(&self, indices: &[usize])` helper method for NdArray [here](https://github.com/HoodyJazz/diesel/blob/feature/add-narray-wrappers/diesel/src/pg/types/array.rs#L23C8-L23C37)
- [Wrapper types](https://github.com/HoodyJazz/diesel/blob/feature/add-narray-wrappers/diesel/src/pg/types/array.rs#L253) for Vec2, ..., Vec6 including the following conversion functions for Vec2 and Vec3:
    - `TryFrom` for VecN <-> NdArray
    - FromSql for VecN for direct querying into the wrapper from postgres
    - `Vec<Vec<T>>` and `Vec<Vec<Vec<T>>>` To and From NdArray
- tests for everything at the bottom

A few thoughts I have:
- As I was implementing this, I started thinking that use of more than 3 dimensions is probably super rare so maybe providing wrappers up to Vec3 is enough and beyond that people could still use NdArray with the get() function.

- the reshaping of NdArray into VecN can't really be done very dynamically, at least I haven't found a way to do this. You can still find my commented out [recursive approach](https://github.com/HoodyJazz/diesel/blob/feature/add-narray-wrappers/diesel/src/pg/types/array.rs#L377) to generalize it for N dimensions, but that doesn't work because the return type of the function changes with every call, making the resulting vector more nested. This isn't really a problem, just means the implementations of VecN -> NdArray would become very hard-coded and long (N - 1 nested for-loops) and it feels kind of ugly to do this. Maybe there's some rustic way to do this using macros or so that I'm unaware of, since I'm very new to rust.

Cheers :)

weiznich: Sorry for taking that long to response, but I was rather busy with attending and preparing RustNation. That burned away quite a lot of my time reserved for working on Diesel.

For the `cargo xtask tidy` failure: Do you have any details on the commit that happens on. That's not supposed to happen.

The proposals for the wrapper types + getter look fine to me. I  think using `TryInto` there is reasonable. We might want to have some documentation examples explaining the design otherwise this is fine for me in terms of API. In terms of implementation we might want to see if we can use some sort of macro to define the `TryFrom` impls to not have to repeat them 5 times.

HoodyJazz: No problem at all, I hope you had a good time at RustNation, Thanks for the cleanup, documentation, and merge!

Regarding the `tidy` failure, I'm not 100% sure anymore, but it might have happened on [this commit](https://github.com/HoodyJazz/diesel/commit/22e3e20e7a303f11a3c0fb56b4c02a76299bef3c) and I'm pretty sure happened after I initially created the PR, [here](https://github.com/HoodyJazz/diesel/commit/178260fd018a473400f4bb07aac154f55c65e785) with this [subsequent commit](https://github.com/HoodyJazz/diesel/commit/b504fcf43c1f96d48df1f92f0559dec61e4f1928) to fix the problems that occurred in the CI but not locally.

PR #4966, opened 2026-02-07
Drafted first version of `WithRawConnection`, with extensive documentation

LucaCappelletti94:
This PR introduces the `WithRawConnection` trait for `SqliteConnection`, providing scoped access to the underlying `sqlite3*` handle. This enables users to leverage advanced SQLite C APIs, such as the [Session Extension](https://www.sqlite.org/sessionintro.html), [Backup API](https://www.sqlite.org/backup.html), and [Pre-update Hooks](https://www.sqlite.org/c3ref/update_hook.html), without requiring Diesel to wrap these optional, ABI-variable features.

**Related Discussion:** [#4965 - Exposing SQLite's Variable ABI: Scoped Raw Connection Access](https://github.com/diesel-rs/diesel/discussions/4965)

The new trait is **extensively** documented, and is tested using only stable ABI methods. Ideally, we would want to test also with the session ABI etc, but the point of this trait is exactly that we cannot have those maybe-linkable ABI in the crate in the first place.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -573,6 +573,90 @@ impl SqliteConnection {
         self.raw_connection.deserialize(data)
     }
 
+    /// Provides temporary access to the raw SQLite database connection handle.
+    ///
+    /// This method provides a way to access the underlying `sqlite3` pointer,
+    /// enabling direct use of the SQLite C API for advanced features that
+    /// Diesel does not wrap, such as the [session extension](https://www.sqlite.org/sessionintro.html),
+    /// [hooks](https://www.sqlite.org/c3ref/update_hook.html), or other advanced APIs.
+    ///
+    /// # Why Diesel Doesn't Wrap These APIs
+    ///
+    /// Certain SQLite features, such as the session extension, are **optional** and only
+    /// available when SQLite is compiled with specific flags (e.g., `-DSQLITE_ENABLE_SESSION`
+    /// and `-DSQLITE_ENABLE_PREUPDATE_HOOK` for sessions). These compile-time options determine
+    /// whether the corresponding C API functions exist in the SQLite library's ABI.
+    ///
+    /// Because Diesel must work with any SQLite library at runtime—including system-provided
+    /// libraries that may lack these optional features—it **cannot safely provide wrappers**
+    /// for APIs that may or may not exist. Doing so would either:
+    ///
+    /// - Cause **linker errors** at compile time if the user's `libsqlite3-sys` wasn't compiled
+    ///   with the required flags, or
+    /// - Cause **undefined behavior** at runtime if Diesel called functions that don't exist
+    ///   in the linked library.
+    ///
+    /// While feature gates could theoretically solve this problem, Diesel already has an
+    /// extensive API surface with many existing feature combinations. Each new feature gate
+    /// adds a **combinatorial explosion** of test configurations that must be validated,
+    /// making the library increasingly difficult to maintain. Therefore, exposing the raw
+    /// connection is the preferred approach for niche SQLite features.
+    ///
+    /// By exposing the raw connection handle, Diesel allows users who **know** they have
+    /// access to a properly configured SQLite build to use these advanced features directly
+    /// through their own FFI bindings.
+    ///
+    /// # Safety
+    ///
+    /// This method is marked `unsafe` because improper use of the raw connection handle
+    /// can lead to undefined behavior. The caller must ensure that:
+    ///
+    /// - The connection handle is **not closed** during the callback.
+    /// - The connection handle is **not stored** beyond the callback's scope.
+    /// - Any state changes made through the raw handle are **compatible** with
+    ///   Diesel's expectations about the connection state.
+    /// - Concurrent access rules are respected (SQLite connections are not thread-safe
+    ///   unless using serialized threading mode).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use diesel::sqlite::SqliteConnection;
+    /// use diesel::Connection;
+    ///
+    /// let mut conn = SqliteConnection::establish(":memory:").unwrap();
+    ///
+    /// // SAFETY: We do not close or store the connection handle,
+    /// // and we do not modify state that Diesel depends on.
+    /// let version = unsafe {
+    ///     conn.with_raw_connection(|_raw_conn| {
+    ///         // Use raw_conn with SQLite C API functions from your own
+    ///         // `libsqlite3-sys` (native) or `sqlite-wasm-rs` (WASM) dependency.
+    ///         // For example: ffi::sqlite3session_create(raw_conn, ...)
+    ///         42 // return a value
+    ///     })
+    /// };
+    /// assert_eq!(version, 42);
+    /// ```
+    ///
+    /// # Platform Notes
+    ///
+    /// This method works identically on both native and WASM targets. However,
+    /// you must depend on the appropriate FFI crate for your target:
+    ///
+    /// - **Native**: Add `libsqlite3-sys` as a dependency
+    /// - **WASM** (`wasm32-unknown-unknown`): Add `sqlite-wasm-rs` as a dependency
+    ///
+    /// Both crates expose a compatible `sqlite3` type that can be used with the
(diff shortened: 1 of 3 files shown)


LucaCappelletti94:
Apparently now PR titles are frozen, I cannot update the title from the current draft one.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -607,6 +607,108 @@ impl SqliteConnection {
...
+/// This trait works identically on both native and WASM targets. However,
+/// you must depend on the appropriate FFI crate for your target:
+///
+/// - **Native**: Add `libsqlite3-sys` as a dependency
+/// - **WASM** (`wasm32-unknown-unknown`): Add `sqlite-wasm-rs` as a dependency
+///
+/// Both crates expose a compatible `sqlite3` type that can be used with the
+/// pointer returned by this method.
+#[allow(unsafe_code)]
+pub trait WithRawConnection {
```
weiznich: Instead of having a trait for this, I would rather turn this in a method on `SqliteConnection`. That safes some code and after all this will only be implemented by `SqliteConnection` at all.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -607,6 +607,108 @@ impl SqliteConnection {
...
+/// # Safety
+///
+/// The `with_raw_connection` method is marked `unsafe` because improper use
+/// of the raw connection handle can lead to undefined behavior. The caller
+/// must ensure that:
+///
+/// - The connection handle is **not closed** during the callback.
+/// - The connection handle is **not stored** beyond the callback's scope.
+/// - Any state changes made through the raw handle are **compatible** with
+///   Diesel's expectations about the connection state.
```
weiznich: While this part is technically correct, it doesn't really say much about the expectations here. 

In the end we need to disallow anything that could mess up the internal prepared statement cache or transaction manager. So essentially no statements/function calls that invalidate the state of prepared statements in any way and no calls that might change the transaction state.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -622,6 +724,275 @@ mod tests {
...
+    fn with_raw_connection_provides_valid_handle() {
+        let connection = &mut connection();
+
+        // SAFETY: We only read the SQLite version, which is a safe operation
+        // that doesn't modify connection state.
+        let version = unsafe {
+            connection.with_raw_connection(|raw_conn| {
+                assert!(!raw_conn.is_null());
+                // Use sqlite3_libversion to verify we have a valid connection
+                let version_ptr = ffi::sqlite3_libversion();
```
weiznich: This doesn't actually use the connection

diesel/src/sqlite/connection/mod.rs
```diff
@@ -622,6 +724,275 @@ mod tests {
...
+        crate::sql_query("CREATE TABLE test_table (id INTEGER PRIMARY KEY, value TEXT)")
+            .execute(connection)
+            .unwrap();
+        crate::sql_query("INSERT INTO test_table (value) VALUES ('hello')")
+            .execute(connection)
+            .unwrap();
+
+        // SAFETY: We only read the last insert rowid, which is a read-only operation.
+        let last_rowid = unsafe {
+            connection.with_raw_connection(|raw_conn| ffi::sqlite3_last_insert_rowid(raw_conn))
```
weiznich, line 748: Side note: We can certainly expose this function from the normal API as well.

weiznich (changes requested):
I'm open to accept this change, but I rather would not like to introduce a new trait for this functionality as this can be a method on `SqliteConnection` without any disadvantages.

Otherwise the documentation could be bit more specific on what diesel actually expects to be not touched. I left a comment about that as well.

Finally there should be a changelog entry for this feature.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: The rowid one you mean? I put it here https://github.com/diesel-rs/diesel/pull/4969

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: I will add richer tests

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: I went for a trait just picturing whether we could ever have more, in the future, SQLite connection objects. I am not sure whether that could be the case, but with this trait in dependant crates I could write `C: WithRawConnection` type of thing. Also, it could be conceivable that such `WithRawConnection` trait should be generalized to other connection types, of course making the raw connection it contains an associated type.

weiznich: If we need that in the future we always can add this trait. Currently it's unnecessary and only makes things more complicated

diesel/src/sqlite/connection/mod.rs
weiznich: Thinking a bit more about this particular test: It's fine to just remove this one.

diesel/src/sqlite/connection/mod.rs
weiznich: No, I mean exposing the `last_inserted_rowid` function
(But as written before: It's not necessary to implement it directly, I just wanted to mention that this would be acceptable to have)

LucaCappelletti94: Ok!

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Removed the comment

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Turned `WithRawConnection` into method

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Removed

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Should I consider this conversation resolved as we may want to add this method in a separate PR? I will open an issue or discussion about it so to not forget it that is okay.

LucaCappelletti94:
I think that test that is currently failing is just a fluke

LucaCappelletti94:
I have drafted a first version of the session extension integration for diesel in another crate: https://github.com/LucaCappelletti94/diesel-sqlite-session

Once this PR gets all the way to crates, I should be able to publish it. I have already tested it in WASM and against Rustqlite on the same methods, and performs pretty decently.

LucaCappelletti94:
Improves the `with_raw_connection` method on `SqliteConnection` by strengthening its safety documentation, fixing tests, and adding a changelog entry. The `# Safety` section now explicitly warns against modifying transaction state and disturbing Diesel's cached prepared statements. The doc example was updated to actually use the raw pointer instead of ignoring it. A tautological assertion in the filename test was replaced with a meaningful check, two redundant tests were removed, and two new tests were added: one verifying the connection remains usable after a panic inside the callback, and another checking filename retrieval on a file-based database.

weiznich (approved):

weiznich:
Related note: I'm quite slow to go through all the PR currently, so expect another month or so until I'm done with all of your PR's. I know that this is disappointing, but there is just not more time from my side and I cannot spend all of that time on your PR's alone.

LucaCappelletti94:
> Related note: I'm quite slow to go through all the PR currently, so expect another month or so until I'm done with all of your PR's. I know that this is disappointing, but there is just not more time from my side and I cannot spend all of that time on your PR's alone.

As already stated some other times, not to worry, that also gives me time to refine them as I am actually using them in my projects. I do not exclude that I have made implementation choices that turn out to not be particularly ergonomic. I hope the diagrams I made are helpful.

As said other times, I just prefer to make PRs when a feature seems in my opinion to be a better fit (or only fit) for diesel rather than other projects, and adding it to the main repo should guarantee that in the years to come it should ideally be maintained or at the very least not be broken by refactoring. The main project I am working on, Earth Metabolome Initiative, likely won't be completed for decades. Slow and steady is better that quick and dirty.

Issue #4955, opened 2026-01-28
Port diesel-cli argument parsing from clap builders to clap derives


weiznich:
Diesel-cli is a relatively old application. It was written at a time when clap did not ship a derive. This now changed on the upstream side (since quite some time). Using the derive allows to more easily change/add arguments and documentation to the CLI tool, therefore we would like to move our old argument parsing code over to use the clap derive macro. Hopefully this also makes handling CLI arguments in our application easier as they are then type safe. This needs to keep the old command line interface

The relevant code is here

https://github.com/diesel-rs/diesel/blob/a7061cc602e9dd69f889236f3c66220239c9313e/diesel_cli/src/cli.rs

Relevant documentation:

* [Clap Builder reference](https://docs.rs/clap/latest/clap/struct.Command.html) (old code)
* [Clap Derive reference](https://docs.rs/clap/latest/clap/_derive/index.html) (new code)

Wesley-Arizio:
I’d love to work on this refactoring and will try to go over it this weekend.

Wesley-Arizio:
@weiznich Hey, as I was working on it, I noticed that there is a lot of helper functions that uses `ArgMatches` in way to get the args so this PR might be a little big/annoying to review.

I'm a bit unsure about changing them but I guess that it's one of the reasons this needs a refactor. 

What do you think? I can try to do some work around and use both: clap derive and builder but then it would need two PRs, in case we need to remove the `ArgMatches` in the future

weiznich:
It's fine to also change the helper functions in the PR that ports over the parsing even if it becomes large. That's the goal of the transition to have everything use the more strongly typed interface.

Wesley-Arizio:
This refactor ended up being more complicated than I expected. There’s still some logic that depends on `ArgMatches`, which I had to keep to preserve the argument order. There are still a few improvements to make, especially around argument parsing and code organization.

I’ve been trying to fix an annoying issue that causes four tests to fail locally, but it looks like a bunch of tests are also failing in CI/CD, so I’ll keep working on it.

First time contributing to open source, still getting the hang of it.

PR #4968, opened 2026-02-08
Port diesel-cli argument parsing to clap derives

Wesley-Arizio:
This PR moves diesel-cli from clap’s builder-style argument parsing to clap derive macros to make the CLI code easier to read, type-safe, and simpler to maintain. While most of the argument definitions now use derives, some of the existing `ArgMatches` logic is still needed to keep the exact argument order and preserve full backwards compatibility.

Issue [#4955](https://github.com/diesel-rs/diesel/issues/4955)

### DIFF
--- a/examples/mysql/getting_started_step_1/tests/step_1.rs
+++ b/examples/mysql/getting_started_step_1/tests/step_1.rs
@@ -34,7 +34,6 @@ fn show_posts() {
     let migrations = diesel_migrations::FileBasedMigrations::find_migrations_directory().unwrap();
     conn.run_pending_migrations(migrations).unwrap();
 
-    #[expect(deprecated, reason = "The suggested method doesn't work in this case")]
     let _ = Command::cargo_bin("show_posts")
         .unwrap()
         .env("MYSQL_DATABASE_URL", db_url.to_string())
--- a/examples/mysql/getting_started_step_2/tests/step_2.rs
+++ b/examples/mysql/getting_started_step_2/tests/step_2.rs
@@ -34,7 +34,6 @@ fn write_post() {
     let migrations = diesel_migrations::FileBasedMigrations::find_migrations_directory().unwrap();
     conn.run_pending_migrations(migrations).unwrap();
 
-    #[expect(deprecated, reason = "The suggested method doesn't work in this case")]
     let _ = Command::cargo_bin("write_post")
         .unwrap()
         .env("MYSQL_DATABASE_URL", db_url.to_string())
@@ -48,7 +47,6 @@ fn write_post() {
                 + " when finished)\n\n\nSaved draft Test Title with id 1\n",
         );
 
-    #[expect(deprecated, reason = "The suggested method doesn't work in this case")]
     let _ = Command::cargo_bin("show_posts")
         .unwrap()
         .env("MYSQL_DATABASE_URL", db_url.to_string())
--- a/examples/mysql/getting_started_step_3/tests/step_3.rs
+++ b/examples/mysql/getting_started_step_3/tests/step_3.rs
@@ -34,7 +34,6 @@ fn publish_post() {
     let migrations = diesel_migrations::FileBasedMigrations::find_migrations_directory().unwrap();
     conn.run_pending_migrations(migrations).unwrap();
 
-    #[expect(deprecated, reason = "The suggested method doesn't work in this case")]
     let _ = Command::cargo_bin("show_posts")
         .unwrap()
         .env("MYSQL_DATABASE_URL", db_url.to_string())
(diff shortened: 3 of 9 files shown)


diesel_cli/src/main.rs
```diff
@@ -327,19 +300,6 @@ fn regenerate_schema_if_file_specified(matches: &ArgMatches) -> Result<(), crate
     Ok(())
 }
 
-fn supported_backends() -> String {
```
weiznich, line 330: We would like to keep outputting the compiled backend in the `--version` output. We can use `#[command(version = some_function())]` to construct the old output:

```
diesel 
 Version: 2.3.2
 Supported Backends: postgres mysql sqlite
```

diesel_cli/src/main.rs
```diff
@@ -66,19 +67,40 @@ fn inner_main() -> Result<(), crate::errors::Error> {
...
-        .expect("Clap should prevent reaching this without subcommand")
-    {
-        ("migration", matches) => self::migrations::run_migration_command(matches)?,
-        ("setup", matches) => run_setup_command(matches)?,
-        ("database", matches) => run_database_command(matches)?,
-        ("completions", matches) => generate_completions_command(matches),
-        ("print-schema", matches) => run_infer_schema(matches)?,
-        _ => unreachable!("The cli parser should prevent reaching here"),
+    let cli = Cli::command();
+    let matches = cli.get_matches();
```
weiznich: I suppose that's an intermediate migration step? Or is there a particular reason why working with the plain matches is sometimes required?

diesel_cli/src/print_schema.rs
```diff
@@ -1,17 +1,156 @@
...
+    /// Include views in the generated schema
+    #[arg(id = "INCLUDE_VIEWS", long = "include-views", action = ArgAction::SetTrue)]
+    pub include_views: bool,
+
+    /// UNSTABLE: Infer nullability for view fields
+    #[arg(id = "EXPERIMENTAL_INFER_NULLABLE_FOR_VIEWS", long = "experimental-infer-nullable-for-views", action = ArgAction::SetFalse)]
+    pub experimental_infer_nullable_for_views: bool,
+
+    /// Only include tables from table-name that matches regexp.
+    #[arg(id = "ONLY_TABLES", long = "only-tables", short = '0', action = ArgAction::SetFalse,  value_parser = clap::value_parser!(bool))]
```
weiznich:
```suggestion
    #[arg(id = "ONLY_TABLES", long = "only-tables", short = 'o', action = ArgAction::SetFalse,  value_parser = clap::value_parser!(bool))]
```

That used a `-o` flag not a `-O`

diesel_cli/src/main.rs
Wesley-Arizio: It is, I tried migrating everything at once but it was hard to debug what exactly was wrong with the tests failing so I kept the matches just to see what was going on.

It seems that the order in which the arguments are passed matters, this is why I had to keep matches there:

https://github.com/diesel-rs/diesel/blob/7dc4312625e5be0b91282e5ef10f47c454248eb9/diesel_cli/src/config.rs#L23

https://github.com/diesel-rs/diesel/blob/7dc4312625e5be0b91282e5ef10f47c454248eb9/diesel_cli/src/config.rs#L78

https://github.com/diesel-rs/diesel/blob/7dc4312625e5be0b91282e5ef10f47c454248eb9/diesel_cli/src/config.rs#L121

Wesley-Arizio: @weiznich To fully remove the ArgMatches we would have to re-formulate how the CLI is structured because it parses into schema blocks to apply the different flags and configurations, but with clap derive this is not so straightforward.

I guess we could create something like this

```rust
#[derive(Debug, clap::Args)]
pub struct PrintSchemaArgs {
    #[arg(
        long = "schema-block",
        action = clap::ArgAction::Append,
        num_args = 1..,
    )]
    pub schema_blocks: Vec<Vec<String>>,
}
```

and use like this:

```sh
diesel print-schema \
  --schema-block default with-docs=true only-tables=users \
  --schema-block analytics schema=analytics
```

But that might be a major change. 
Right now it's working with both, clap derive and ArgMatches

weiznich: Sorry for being silent for that long, but there was much other stuff to care about in the meantime :(

I did poke around a bit and came up with the following solution: https://github.com/diesel-rs/diesel/pull/4968/commits/e1543d919f0de644d80f1ccdef8c203c8c5d8dcb to that problem. Essentially it just adds at a single place a targeted parser for the desired behaviour instead of having `ArgMatches` and the type arguments at the same time.

Otherwise I rebased the PR and added a few snapshot tests for the help output before and after this change, so that we can easily see if everything should be fine. You can find the snapshot changes here: https://github.com/diesel-rs/diesel/pull/4968/commits/daf7a4912d4d5f3ee36b711065fb42e7bcca9f19

I will try do to a final review next week, but otherwise this should be now ready to go. Thanks again for your help here.

Wesley-Arizio: Awesome refactor, I was so focused on removing `ArgMatches` that I didn't think of doing a structure similar to it. 

I'll try a good first issue next time, thanks for the help

PR #4969, opened 2026-02-08
Added SQLite hooks helper methods

LucaCappelletti94:
Exposes SQLite's C-level callback hooks through a safe Rust API on `SqliteConnection`: row-change notification, the transaction hooks, the progress and busy handlers, the authorizer, and execution tracing.

## Motivation

I need a LISTEN/NOTIFY-style setup for SQLite in WASM frontends, where the hooks drive UI updates and forward changes to the server. The diagram shows which hook fires when over a statement's lifecycle. We use it internally, and it might suit the diesel docs.

SQLite Hooks Lifecycle

## API surface

| C API | Rust surface | Purpose | Multiplicity |
|---|---|---|---|
| [`sqlite3_update_hook`](https://www.sqlite.org/c3ref/update_hook.html) | `on_change`, `on_table_change`, `on_insert`, `on_update`, `on_delete` | Row change events | Many per connection (internal dispatcher) |
| [`sqlite3_commit_hook`](https://www.sqlite.org/c3ref/commit_hook.html) | `on_commit` | Pre-commit interception | One (replaces) |
| [`sqlite3_rollback_hook`](https://www.sqlite.org/c3ref/commit_hook.html) | `on_rollback` | Post-rollback notification | One (replaces) |
| [`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html) | `on_wal` | WAL commit notification | One (replaces) |
| [`sqlite3_progress_handler`](https://www.sqlite.org/c3ref/progress_handler.html) | `on_progress` | Long-running query interruption | One (replaces) |
| [`sqlite3_busy_handler`](https://www.sqlite.org/c3ref/busy_handler.html) | `on_busy` | Custom lock-contention handling | One (replaces) |
| [`sqlite3_busy_timeout`](https://www.sqlite.org/c3ref/busy_timeout.html) | `set_busy_timeout` | Simple timeout-based retry | One (clears `on_busy`) |
| [`sqlite3_set_authorizer`](https://www.sqlite.org/c3ref/set_authorizer.html) | `on_authorize` | SQL compilation access control | One (replaces) |
| [`sqlite3_trace_v2`](https://www.sqlite.org/c3ref/trace_v2.html) | `on_trace`, `on_read` | SQL execution tracing | One (replaces) |

Row-change hooks multiplex over the single `sqlite3_update_hook`: an internal dispatcher fans it out to any number of callbacks, each filtered by table name and an op mask. `on_change` matches every table, `on_table_change::<T>(ops, ...)` matches one table, and `on_insert` / `on_update` / `on_delete` are single-op wrappers. Each returns a `ChangeHookId` for removal via `remove_change_hook`, `clear_change_hooks_for::<T>`, or `clear_all_change_hooks`.

## Quick start

```rust
use diesel::prelude::*;
use std::sync::{Arc, Mutex};

diesel::table! {
    users (id) { id -> Integer, name -> Text, }
}

let conn = &mut SqliteConnection::establish(":memory:").unwrap();
let log = Arc::new(Mutex::new(Vec::new()));
let log2 = log.clone();

conn.on_insert::<users::table, _>(move |event| {
    log2.lock().unwrap().push(event.rowid);
});

diesel::insert_into(users::table)
    .values(users::name.eq("Alice"))
    .execute(conn)
    .unwrap();

assert_eq!(*log.lock().unwrap(), vec![1i64]);
```

Runnable examples for every hook are in the rustdoc on each method.

## Design notes

The row-change dispatcher lives behind a lazily allocated box (`OnceCell<Box<RefCell<ChangeHookDispatcher>>>`), and SQLite is handed that stable heap address. Storing it inline would dangle once the freely movable connection is returned, stored in a struct, or moved into a pool, and allocating it lazily means the borrowed connection wrapper used during SQL-function callbacks, which never registers a hook, does not allocate or leak it. A regression test pins the address across a move. The C update hook itself is registered on the first callback and removed with the last.

Single-slot hooks store `Option<Box<dyn Any + Send>>` on `RawConnection`, and the trampoline recovers the concrete `F` by pointer cast. Every trampoline wraps its callback in `catch_unwind` and aborts on panic (matching the existing `run_collation_function`), and `Drop` unregisters all hooks before `sqlite3_close`. Because a callback cannot use the connection, `find_by_rowid::<T, M>(rowid)` reloads the affected row afterward via `WHERE rowid = ?`, which works even with TEXT primary keys.

## New types

- Row changes: `SqliteChangeEvent`, `SqliteChangeOp` (`#[non_exhaustive]`, with `Unknown(i32)`), `SqliteChangeOps` (an INSERT/UPDATE/DELETE/UNKNOWN bitmask), `ChangeHookId`.
- Authorizer: `AuthorizerAction` (documented action codes plus `Unknown(i32)`), `AuthorizerContext`, `AuthorizerDecision`.
- Tracing: `SqliteTraceFlags` (`STMT`, `PROFILE`, `ROW`, `CLOSE`, `ALL`) and `SqliteTraceEvent`, which carries a `readonly` flag from `sqlite3_stmt_readonly`.

## Compatibility and testing

Stable SQLite C APIs only, with no compile-time flags. `sqlite3_trace_v2` needs 3.14.0 and `sqlite3_wal_hook` needs 3.7.0, the rest are 3.0.0. `no_std` compatible. Needs neither `WithRawConnection` nor dynamic symbol lookup, so it is independent of and complementary to #4954 (extension loading) and #4966 (`WithRawConnection`). Every public method has a runnable doctest, each hook has integration tests, and each inherited limitation below has a test pinning the current behavior.

## Limitations (inherited from SQLite)

Row-change hooks fire only for [rowid tables](https://www.sqlite.org/rowidtable.html), and not for system-table changes, implicit `ON CONFLICT REPLACE` deletes, or the [truncate optimization](https://www.sqlite.org/lang_delete.html#truncateopt). Their callbacks cannot use the connection (use `find_by_rowid` afterward).

Single-slot hooks replace the previous callback on re-registration. The WAL hook is overwritten by `PRAGMA wal_autocheckpoint`, `set_busy_timeout` and `on_busy` clear each other, and a panic in any callback aborts the process.

The authorizer runs during `sqlite3_prepare()` rather than execution (it may re-run if a schema change forces recompilation) and is defense in depth, not a sandbox. For tracing, `SQLITE_TRACE_ROW` fires per row (prefer `STMT` and `PROFILE`), and the `readonly` flag misses writes made by UDFs on a separate connection or by virtual tables with side effects. For the progress handler, N around 1000 is suggested (about 1.5us per callback at N=1), and since 3.41.0 it may also fire during `sqlite3_prepare()`.

## Open questions

SQLite delivers only `SQLITE_INSERT`, `SQLITE_UPDATE`, and `SQLITE_DELETE` today, so `SqliteChangeOp::Unknown(i32)` is forward-compatibility only (`SqliteChangeOps::ALL` includes the `UNKNOWN` bit so a catch-all still fires if a future code appears). Do you want to surface `Unknown`, or abort on an unrecognized code?

Separately, this adds several independent hook families, and I am happy to split the row-change hooks from the authorizer and tracing if that reviews more easily.

## Relation to PostgreSQL LISTEN/NOTIFY

Unlike PostgreSQL `NOTIFY` (inter-process, deferred to commit, explicit), these hooks are in-process, synchronous, and fire automatically during `sqlite3_step()`. The nearest analogue is a statement-level trigger that cannot run SQL, and diesel's existing `PgConnection::notifications_iter()` stays separate because libpq exposes a queue to poll while SQLite exposes execution-time callbacks.

### DIFF
--- a/diesel/src/sqlite/connection/authorizer.rs
+++ b/diesel/src/sqlite/connection/authorizer.rs
@@ -90,6 +90,50 @@ pub enum AuthorizerAction {
 }
 
 impl AuthorizerAction {
+    /// Returns `true` if this action modifies the database schema.
+    ///
+    /// Schema-modifying actions include creating, dropping, or altering
+    /// tables, indexes, triggers, views, and virtual tables.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use diesel::sqlite::AuthorizerAction;
+    ///
+    /// assert!(AuthorizerAction::CreateTable.is_schema_modifying());
+    /// assert!(AuthorizerAction::DropIndex.is_schema_modifying());
+    /// assert!(AuthorizerAction::AlterTable.is_schema_modifying());
+    ///
+    /// assert!(!AuthorizerAction::Select.is_schema_modifying());
+    /// assert!(!AuthorizerAction::Insert.is_schema_modifying());
+    /// assert!(!AuthorizerAction::Update.is_schema_modifying());
+    /// assert!(!AuthorizerAction::Delete.is_schema_modifying());
+    /// ```
+    pub fn is_schema_modifying(&self) -> bool {
+        matches!(
+            self,
+            Self::CreateIndex
+                | Self::CreateTable
+                | Self::CreateTempIndex
+                | Self::CreateTempTable
+                | Self::CreateTempTrigger
+                | Self::CreateTempView
+                | Self::CreateTrigger
+                | Self::CreateView
+                | Self::CreateVTable
+                | Self::DropIndex
+                | Self::DropTable
+                | Self::DropTempIndex
+                | Self::DropTempTable
+                | Self::DropTempTrigger
+                | Self::DropTempView
+                | Self::DropTrigger
+                | Self::DropView
+                | Self::DropVTable
+                | Self::AlterTable
+        )
+    }
+
     /// Converts an FFI action code to the Rust enum variant.
     pub(crate) fn from_ffi(code: i32) -> Self {
         match code {
--- a/diesel/src/sqlite/connection/authorizer.rs
+++ b/diesel/src/sqlite/connection/authorizer.rs
@@ -1,0 +1,211 @@
+//! Types for the SQLite authorizer callback.
+//!
+//! See [`sqlite3_set_authorizer`](https://sqlite.org/c3ref/set_authorizer.html)
+//! and the [action code constants](https://sqlite.org/c3ref/c_alter_table.html).
+
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+use sqlite_wasm_rs as ffi;
+
+/// Authorizer action codes.
+///
+/// These correspond to the action codes passed to the authorizer callback
+/// by SQLite. The callback is invoked during SQL statement compilation
+/// (not during execution).
+///
+/// Added in SQLite 3.0.0 (June 2004).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum AuthorizerAction {
+    /// CREATE INDEX
+    CreateIndex,
+    /// CREATE TABLE
+    CreateTable,
+    /// CREATE TEMP INDEX
+    CreateTempIndex,
+    /// CREATE TEMP TABLE
+    CreateTempTable,
+    /// CREATE TEMP TRIGGER
+    CreateTempTrigger,
+    /// CREATE TEMP VIEW
+    CreateTempView,
+    /// CREATE TRIGGER
+    CreateTrigger,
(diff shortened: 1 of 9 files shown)


weiznich (commented):
Generally I welcome the addition of all these API's. This is good work.

It would be great if you could split up this PR into smaller pieces. Right now it adds 6k lines of code, which make it quite large and hard to review in one go. I believe this would be easier to handle if we had one PR per added C-API. I acknowledge that this adds quite a bit of overhead on your side, but I believe this is the fastest way to get something merged.

LucaCappelletti94:
Absolutely understandable, this is one of those PRs that would benefit by doing stacked PRs

weiznich:
It seems like [stacked PR's](https://github.github.com/gh-stack/) are currently in a private preview. I will sign up diesel for the wait list tomorrow, maybe we get accessor that feature soonish.

LucaCappelletti94:
Ok, otherwise I will do the poor man's equivalent and go for a series of git trees

PR #4970, opened 2026-02-11
Enable MySQL "LOAD DATA LOCAL" with url option

TheLQ:
[Enable LOAD DATA LOCAL](https://dev.mysql.com/doc/refman/9.0/en/server-system-variables.html#sysvar_local_infile) in `batch_execute`. mysqlclient handles all the IO. I'm inserting 30 million rows significantly faster now.

With native Diesel support, the [mysql](https://dev.mysql.com/doc/refman/9.5/en/load-data.html) and [postgres](https://www.postgresql.org/docs/current/sql-copy.html) versions get complicated quickly. Also LOAD DATA isn't supported in transactions. This option is enough for me to do it when rarely needed.

# Example

```rust
fn main() {
    let url = "mysql://u:p@localhost/db?local_infile";
    let conn = &mut MysqlConnection::establish(url).unwrap();
    conn.batch_execute(
        "CREATE TEMPORARY TABLE fast_table (\
        id INT NOT NULL PRIMARY KEY,\
        value VARCHAR(20) NOT NULL\
        )",
    )
        .unwrap();

    let load_path = Path::new("mysql-rows.dat");
    write_rows_file(&load_path);
    load_rows_file(conn, &load_path);
}

const ROW_SEP: u8 = 0x1e;
const COL_SEP: u8 = 0x1f;
fn write_rows_file(path: &Path) -> () {
    let mut out_file = File::create(path).unwrap();
    for row in 0..10 {
        let something = "something";
        write!(out_file, "{row}{COL_SEP}{something}{ROW_SEP}").unwrap();
    }
}

fn load_rows_file(conn: &mut MysqlConnection, path: &Path) {
    conn.batch_execute(&format!(
        "LOAD DATA LOCAL INFILE '{}' \
        INTO TABLE `fast_table` \
        FIELDS TERMINATED BY '{COL_SEP}' \
        LINES TERMINATED BY '{ROW_SEP}' \
        (id, value)",
        path.display()
    ))
        .unwrap()
}

```

### DIFF
--- a/diesel/src/mysql/connection/url.rs
+++ b/diesel/src/mysql/connection/url.rs
@@ -54,6 +54,7 @@ pub(super) struct ConnectionOptions {
     ssl_ca: Option<CString>,
     ssl_cert: Option<CString>,
     ssl_key: Option<CString>,
+    local_infile: Option<bool>,
 }
 
 impl ConnectionOptions {
@@ -114,6 +115,17 @@ impl ConnectionOptions {
             _ => None,
         };
 
+        let local_infile = match query_pairs.get("local_infile") {
+            Some(v) => match v.parse() {
+                Ok(v) => Some(v),
+                Err(_) => {
+                    let msg = "unknown local_infile";
+                    return Err(ConnectionError::InvalidConnectionUrl(msg.into()));
+                }
+            },
+            None => None,
+        };
+
         let host = match url.host() {
             Some(Host::Ipv6(host)) => Some(CString::new(host.to_string())?),
             Some(host) if host.to_string() == "localhost" && unix_socket.is_some() => None,
@@ -146,6 +158,7 @@ impl ConnectionOptions {
             ssl_ca,
             ssl_cert,
             ssl_key,
+            local_infile,
         })
     }
 
@@ -191,6 +204,10 @@ impl ConnectionOptions {
 
     pub(super) fn ssl_mode(&self) -> Option<mysql_ssl_mode> {
         self.ssl_mode
+    }
+
+    pub(super) fn local_infile(&self) -> Option<bool> {
+        self.local_infile
     }
 }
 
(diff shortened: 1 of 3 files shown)


diesel/src/mysql/connection/url.rs
```diff
@@ -132,7 +132,10 @@ impl ConnectionOptions {
         };
 
         // this is not present in the database_url, using a default value
-        let client_flags = CapabilityFlags::CLIENT_FOUND_ROWS;
+        let mut client_flags = CapabilityFlags::CLIENT_FOUND_ROWS;
+        if query_pairs.contains_key("local_infile") {
```
JohnTitor: `local_infile` should take binary value so we need handle it as well: https://dev.mysql.com/doc/refman/8.4/en/mysql-command-options.html#option_mysql_local-infile

JohnTitor (commented):
Also I'd tweak `ConnectionOptions` and add tests.

TheLQ:
Initially used mysqlclient's default compiled flag (usually disabled) or explicit enable.

To have explicit enable and disable we need to use mysql_option instead.
Docs for MYSQL_OPT_LOCAL_INFILE https://dev.mysql.com/doc/c-api/9.5/en/mysql-options.html

Added parsing unit test. Example above was successful too.

TheLQ:
Rebased after #4972 so tests pass. Also fixed the doc comment.

ThunderComplex:
I am trying to run tests on the main branch but they don't work and it seems to be because of this. I am getting this strange error:  

```
error[E0425]: cannot find value `MYSQL_OPT_LOCAL_INFILE` in module `mysqlclient_sys::mysql_option`
   --> diesel/src/mysql/connection/raw.rs:260:48
    |
260 |                 mysqlclient_sys::mysql_option::MYSQL_OPT_LOCAL_INFILE,
    |                                                ^^^^^^^^^^^^^^^^^^^^^^ not found in `mysqlclient_sys::mysql_option`

For more information about this error, try `rustc --explain E0425`.
Failed to run migrations
```  

Note that I'm on arch and had mariadb installed to give me the libmysqlclient. I uninstalled mariadb and installed https://aur.archlinux.org/packages/libmysqlclient from the AUR using yay, but that just gives me `Did not find a compatible version of libmysqlclient.`

Will continue to investigate

weiznich:
@ThunderComplex Thanks for reporting, that shouldn't happen. I've filled https://github.com/diesel-rs/diesel/issues/5013 with what's going on there and how to resolve this. Help with fixing that would be welcome.

PR #4974, opened 2026-02-14
Fix zero-mysql benches

elbaro:
This PR fixes zero-mysql benches.

Tested with `cargo bench --bench benchmarks --features "zero-mysql,mysql"`.


weiznich (approved):
Thanks

PR #4975, opened 2026-02-14
Implemented API support for `set_limit` (`sqlite3_limit`) in sqlite

LucaCappelletti94:
This PR exposes SQLite's [`sqlite3_limit`](https://www.sqlite.org/c3ref/limit.html) API through new methods on `SqliteConnection`. It is another PR in the series of: "let's make more of SQLite usable with Diesel in frontend WASM development", but it really applies to many other cases.

## Motivation

SQLite's [security documentation](https://sqlite.org/security.html) explicitly recommends using `sqlite3_limit()` for hardening:

> Reduce the limits that SQLite imposes on inputs. This can help prevent denial of service attacks and other kinds of mischief that can occur as a result of unusually large inputs. You can do this either at compile-time using -DSQLITE_MAX_... options, or at run-time using the sqlite3_limit() interface.

The documentation also notes: *"Most applications can reduce limits dramatically without impacting functionality."*

SQLite's [limits documentation](https://sqlite.org/limits.html) provides specific context:

- **`CompoundSelect`**: Defaults to 500, but *"in practice we almost never see the number of terms in a compound select exceed single digits"*
- **`ExprDepth`**: Can prevent stack exhaustion from deeply nested expressions
- **`SqlLength`**: The security page suggests reducing from 1,000,000,000 to 100,000 for high-security deployments

While users could technically call the FFI directly, this PR provides a safe, ergonomic, and documented API that makes following SQLite's security recommendations straightforward.

## Usage

Apply all recommended security limits in one call:

```rust
use diesel::sqlite::{SqliteConnection, SqliteLimit};
use diesel::Connection;

let mut conn = SqliteConnection::establish(":memory:")?;

// Apply SQLite's recommended high-security limits
conn.set_recommended_security_limits();

// Relax specific limits if needed for your application
conn.set_limit(SqliteLimit::VariableNumber, 999);
```

Or configure individual limits:

```rust
use diesel::sqlite::{SqliteConnection, SqliteLimit};
use diesel::Connection;

let mut conn = SqliteConnection::establish(":memory:")?;

// Limit SQL statement length to 100KB
conn.set_limit(SqliteLimit::SqlLength, 100_000);

// Prevent ReDoS via LIKE patterns
conn.set_limit(SqliteLimit::LikePatternLength, 100);

// Query current limits
let expr_depth = conn.get_limit(SqliteLimit::ExprDepth);
```


LucaCappelletti94:
The compile test failures (`mysql_on_conflict_tests.rs`, `pg_on_conflict_requires_valid_conflict_target.rs`, `update_requires_left_side_of_eq_to_be_a_column.rs`) seem to be pre-existing on main and unrelated to this PR. They're caused by rustc error message formatting changes (e.g., `Column` vs `diesel::Column` in trait bounds).

I will do a quick separate PR to fix them later if that is correct and they are unrelated to this PR.

LucaCappelletti94:
Resolved the error case, it was a naming collision.

LucaCappelletti94:
Made sure it is merge-able, and trimmed down a bit the documentation.

diesel/src/sqlite/connection/limits.rs
```diff
@@ -0,0 +1,108 @@
...
+    TriggerDepth,
+
+    /// Maximum number of auxiliary worker threads that a single prepared
+    /// statement may start.
+    ///
+    /// Default: 0 (varies by SQLite compile-time settings)
+    WorkerThreads,
+}
+
+impl SqliteLimit {
```
weiznich, line 109: Can we add two sets of associated constants to `SqliteLimit`, one for the safe values for each variant and one for the default value for each variant?

So something like that:

```rust
impl SqliteLimit {
    pub const SAFE_SQL_LENGTH_LIMIT: i32 = 100_000;
    pub const DEFAULT_SQL_LENGTH_LIMIT: i32 = 1_000_000;
    // for all other variants as wel
}
```

This would give users an easy way to just use these values without needing to lookup all the values.

Ideally the documentation of the variants in the enum above should then link to the constants and the constants should link back to the right variant.

weiznich (commented):
This looks good, thanks for adding it and keeping it  up to date ❤️ 

I have one suggestion, so that we also can include and expose the relevant default/secure default values which might make using this a bit easier for our users.

diesel/src/sqlite/connection/limits.rs
LucaCappelletti94: Ok

LucaCappelletti94:
Added the consts, and since I realized I had only tested some of the limits I have extended those tests. Currently it is failing because of a flaky contention to read/write the sqlite database, which I have addressed in #5072

weiznich (approved):

PR #4976, opened 2026-02-14
Added support for `sqlite3_db_config` API

LucaCappelletti94:
This PR adds two related features for SQLite security hardening:

1. **[`sqlite3_db_config`](https://www.sqlite.org/c3ref/db_config.html) methods** for per-connection security configuration
2. **`SqliteFunctionBehavior` flags** for controlling custom SQL function behavior

These features are interdependent: when [`SQLITE_DBCONFIG_TRUSTED_SCHEMA`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigtrustedschema) is disabled via `set_trusted_schema(false)`, custom SQL functions registered via `define_sql_function!` cannot be called from schema objects (views, triggers, CHECK constraints, etc.) unless marked with [`SQLITE_INNOCUOUS`](https://www.sqlite.org/c3ref/c_deterministic.html#sqliteinnocuous). Without the ability to specify this flag, the security hardening feature would be incomplete.

### Connection Configuration

New methods on `SqliteConnection` wrapping [`sqlite3_db_config()`](https://www.sqlite.org/c3ref/db_config.html):

```rust
let mut conn = SqliteConnection::establish(":memory:")?;

// Security hardening
conn.set_defensive(true)?;                // SQLITE_DBCONFIG_DEFENSIVE
conn.set_trusted_schema(false)?;          // SQLITE_DBCONFIG_TRUSTED_SCHEMA
conn.set_load_extension_enabled(false)?;  // SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION

// Feature toggles
conn.set_triggers_enabled(false)?;        // SQLITE_DBCONFIG_ENABLE_TRIGGER
conn.set_views_enabled(false)?;           // SQLITE_DBCONFIG_ENABLE_VIEW
conn.set_foreign_keys_enabled(true)?;     // SQLITE_DBCONFIG_ENABLE_FKEY

// SQL compliance
conn.set_double_quoted_string_dml(false)?;  // SQLITE_DBCONFIG_DQS_DML
conn.set_double_quoted_string_ddl(false)?;  // SQLITE_DBCONFIG_DQS_DDL
```

See [SQLITE_DBCONFIG constants](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html) for details on each option.

> **Note:** `set_load_extension_enabled()` only controls whether the `load_extension()` **SQL function** can be used. It does not affect the C API `sqlite3_load_extension()`. If SQLite was compiled with `SQLITE_OMIT_LOAD_EXTENSION`, this setting has no effect.

### Function Behavior Flags

New `SqliteFunctionBehavior` bitflags wrapping [function flags](https://www.sqlite.org/c3ref/c_deterministic.html) passed to `sqlite3_create_function_v2()`:

```rust
use diesel::sqlite::SqliteFunctionBehavior;

// Deterministic function (most common) - SQLITE_DETERMINISTIC
my_func_utils::register_impl(
    conn,
    SqliteFunctionBehavior::DETERMINISTIC,
    |x: String| x.to_uppercase(),
)?;

// Safe for untrusted schemas - SQLITE_INNOCUOUS
// Use with set_trusted_schema(false) for security hardening
my_func_utils::register_impl(
    conn,
    SqliteFunctionBehavior::DETERMINISTIC | SqliteFunctionBehavior::INNOCUOUS,
    |x: String| x.len() as i32,
)?;

// Has side effects, block from schema objects - SQLITE_DIRECTONLY
my_func_utils::register_impl(
    conn,
    SqliteFunctionBehavior::DIRECTONLY,
    |x: String| { log::info!("{}", x); x },
)?;

// Non-deterministic (e.g., random) - no flags
my_func_utils::register_impl(
    conn,
    SqliteFunctionBehavior::empty(),
    || rand::random::<i32>(),
)?;
```

## Breaking Change

The `register_impl` and `register_nondeterministic_impl` functions generated by `#[declare_sql_function]` have been consolidated into a single `register_impl` that takes a `SqliteFunctionBehavior` parameter:

| Before | After |
|--------|-------|
| `register_impl(conn, f)` | `register_impl(conn, SqliteFunctionBehavior::DETERMINISTIC, f)` |
| `register_nondeterministic_impl(conn, f)` | `register_impl(conn, SqliteFunctionBehavior::empty(), f)` |

**Rationale:** The previous API provided no way to specify `SQLITE_INNOCUOUS` or `SQLITE_DIRECTONLY`. This made custom functions incompatible with security-hardened configurations where `set_trusted_schema(false)` is used. Also, I think this function was rather under-documented as I only discovered it while doing this PR, and I had previously erroneously used the `register_impl` to register a `uuid`, which is very much non deterministic.

## Visual Reference

Since these flags are rather complex and I needed some visual help to figure them out, I prepared the following diagrams illustrate how these settings interact. I think they should be integrated with the documentation when they are positioned somewhere more permanent:

**Connection Settings Overview** — Shows all `db_config` options, their defaults, and what each controls (DDL statements, defensive mode, schema objects like views/triggers/foreign keys, etc.):

SQLite db_config Connection Settings

**Function Behavior Flags** — Shows how `SqliteFunctionBehavior` flags control where custom functions can be called from, and the difference between trusted and hardened modes:

SQLite Function Behavior Flags

## SQLite Version Requirements

| Feature | SQLite Version |
|---------|----------------|
| [`set_defensive`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigdefensive) | 3.26.0 (2018-12) |
| [`set_trusted_schema`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigtrustedschema) | 3.31.0 (2020-01) |
| [`set_load_extension_enabled`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenableloadextension) | 3.13.0 (2016-05) |
| [`set_triggers_enabled`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenabletrigger) | 3.8.7 (2014-10) |
| [`set_views_enabled`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenableview) | 3.30.0 (2019-10) |
| [`set_foreign_keys_enabled`](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenablefkey) | 3.8.7 (2014-10) |
| [`DETERMINISTIC`](https://www.sqlite.org/c3ref/c_deterministic.html#sqlitedeterministic) | 3.8.3 (2014-02) |
| [`DIRECTONLY`](https://www.sqlite.org/c3ref/c_deterministic.html#sqlitedirectonly) | 3.31.0 (2020-01) |
| [`INNOCUOUS`](https://www.sqlite.org/c3ref/c_deterministic.html#sqliteinnocuous) | 3.31.0 (2020-01) |


LucaCappelletti94:
Just ensuring that everything remains mergiable.

diesel_derives/src/lib.rs
```diff
@@ -2189,14 +2189,26 @@ const AUTO_TYPE_DEFAULT_FUNCTION_TYPE_CASE: dsl_auto_type::Case = dsl_auto_type:
 ///
 /// On most backends, the implementation of the function is defined in a
 /// migration using `CREATE FUNCTION`. On SQLite, the function is implemented in
-/// Rust instead. You must call `register_impl` or
-/// `register_nondeterministic_impl` (in the generated function's `_internals`
-/// module) with every connection before you can use the function.
+/// Rust instead. You must call `register_impl` (in the generated function's
```
weiznich, line 2192: This also could use the new auto_extension function, so we should link it from here?

diesel_derives/src/sql_function.rs
```diff
@@ -756,44 +756,52 @@ fn generate_tokens_for_non_aggregate_functions(
...
 
                 #[allow(dead_code)]
-                /// Registers an implementation for this function on the given connection
-                ///
-                /// This function must be called for every `SqliteConnection` before
-                /// this SQL function can be used on SQLite. The implementation must be
-                /// deterministic (returns the same result given the same arguments). If
-                /// the function is nondeterministic, call
-                /// `register_nondeterministic_impl` instead.
-                pub fn register_impl<F, Ret, #(#arg_name,)*>(
```
weiznich, line 766: Removing the two existing `register` functions in a breaking change. We cannot ship that without a major release, which we don't plan to ship anytime soon.

Fortunately it's not a critical problem as we could just add a third function here that allows passing the `SqliteFunctionBehaviour` parameter explicitly. The existing two functions can internally call the more general function then.

That written: There is something in the existing API that I don't see in the new API. The existing `register_impl` function takes a `Fn` and not a `FnMut, which at least somewhat enforces at compile time that the behaviour of the callback is deterministic as you cannot easily share and mutate state between your application and the SQLite function implementation. It's obviously not a real safety mechanism, but it is at least a nudge in the right direction. I really would like to keep this in some way as this helps to write more correct code.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -4,6 +4,22 @@ extern crate libsqlite3_sys as ffi;
...
+//
+// These constants were introduced in SQLite 3.49.0 and are only present in
+// `libsqlite3-sys` >= 0.35.0.  Diesel supports `libsqlite3-sys` >= 0.17.2,
+// so the constants may be absent at compile time.  We define them here so
+// that diesel compiles against any supported `libsqlite3-sys` version; if
+// the linked SQLite library is too old to recognise the option, the
+// `sqlite3_db_config()` call will return an error at runtime, which the
+// calling code already handles.
+const SQLITE_DBCONFIG_ENABLE_ATTACH_CREATE: i32 = 1020;
+const SQLITE_DBCONFIG_ENABLE_ATTACH_WRITE: i32 = 1021;
```
weiznich: Maybe move this constants into the `raw` module instead?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -955,13 +969,513 @@ impl SqliteConnection {
...
 
+    /// Enable or disable SQLite defensive mode.
+    ///
+    /// When enabled, defensive mode prevents:
+    /// - Direct writes to shadow tables (used by FTS5, R-Tree, etc.)
+    /// - Dangerous PRAGMA commands like `writable_schema`
+    /// - `sqlite3_deserialize()` from opening unsafe database images
+    /// - Other potentially dangerous operations
+    ///
+    /// # Security Hardening Recipe
```
weiznich: Maybe also link to the secure defaults function added in the limits PR?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -955,13 +969,513 @@ impl SqliteConnection {
...
+            .set_db_config_bool(ffi::SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION, enabled)
+    }
+
+    /// Check if `load_extension()` SQL function is enabled.
+    ///
+    /// See [`set_load_extension_enabled`][Self::set_load_extension_enabled] for details.
+    pub fn is_load_extension_enabled(&self) -> QueryResult<bool> {
+        self.raw_connection
+            .get_db_config_bool(ffi::SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION)
+    }
```
weiznich: It's likely a better idea to expose this kind of functionality in a callback style function. So that you restrict the time `set_load_extension_enabled` to true to the minimal possible amount. So something like:

```rust
fn with_load_extensions_enabled<R, E>(&mut self, f: impl FnOnce(&mut Self) -> Result<R, E>) -> Result<R, E> where E: From<diesel::result::Error>{
     self.set_load_extension_enabled(true)?;
     let r = f(self);
     self.set_load_extension_enabled(false)?;
     r
}

```

diesel/src/sqlite/connection/mod.rs
```diff
@@ -955,13 +969,513 @@ impl SqliteConnection {
...
+    /// ```
+    pub fn set_double_quoted_string_dml(&mut self, enabled: bool) -> QueryResult<()> {
+        self.raw_connection
+            .set_db_config_bool(ffi::SQLITE_DBCONFIG_DQS_DML, enabled)
+    }
+
+    /// Check if double-quoted strings in DML are enabled.
+    ///
+    /// See [`set_double_quoted_string_dml`][Self::set_double_quoted_string_dml] for details.
+    pub fn is_double_quoted_string_dml_enabled(&self) -> QueryResult<bool> {
```
weiznich:
```suggestion
    pub fn are_double_quoted_string_dml_enabled(&self) -> QueryResult<bool> {
```

Not 100% sure, but this sounds like more sound English to me

diesel/src/sqlite/connection/mod.rs
```diff
@@ -955,13 +969,513 @@ impl SqliteConnection {
...
+    /// ```
+    pub fn set_double_quoted_string_ddl(&mut self, enabled: bool) -> QueryResult<()> {
+        self.raw_connection
+            .set_db_config_bool(ffi::SQLITE_DBCONFIG_DQS_DDL, enabled)
+    }
+
+    /// Check if double-quoted strings in DDL are enabled.
+    ///
+    /// See [`set_double_quoted_string_ddl`][Self::set_double_quoted_string_ddl] for details.
+    pub fn is_double_quoted_string_ddl_enabled(&self) -> QueryResult<bool> {
```
weiznich:
```suggestion
    pub fn are_double_quoted_string_ddl_enabled(&self) -> QueryResult<bool> {
```

docs/sqlite_db_config.svg
weiznich, line 1: These two graphics don't seem to be linked anywhere. I rather would put them on the web page than in the crate documentation. Maybe combined with a short guide how to use all the SQLite specific features?

weiznich (changes requested):
Overall these are welcome additions to the API.

My main concern here is the breaking change. We need to avoid that, while also keeping the old "safety" invariant around deterministic/non-deterministic functions. Otherwise I only have minor remarks on the implementation.

The added features should also be added to the changelog.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: I went for `are_double_quoted_strings_ddl_enabled`, with an `s` to the `strings` so it is conjugated. I double checked this with the clankers as I am not a English mother-tongue myself.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Also here I went for `are_double_quoted_strings_dml_enabled`, with an `s` to the `strings` so it is conjugated.

diesel_derives/src/lib.rs
LucaCappelletti94: The "Use with SQLite" section now links to `register_auto_extension` and explains that you can register the implementation automatically for every new connection instead of doing it manually per connection.

diesel_derives/src/sql_function.rs
LucaCappelletti94: Implemented exactly as you suggested. The generated `_utils` module now exposes three functions:

- `register_impl` taking `Fn` (deterministic, the default and the compile-time nudge you wanted to keep),
- `register_nondeterministic_impl` taking `FnMut`,
- `register_impl_with_behavior` taking an explicit `SqliteFunctionBehavior`.

The first two delegate to `register_impl_with_behavior` (with `DETERMINISTIC` and `empty()` respectively), so there is no breaking change and the `Fn` vs `FnMut` distinction is preserved. The same three-function shape is generated for aggregate functions, where `register_impl` keeps its previous no-flag behavior and `register_impl_with_behavior` is the new explicit variant.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Moved both constants into the `raw` module (as `pub(super)`), next to the other db_config plumbing.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: The Security Hardening Recipe now references `set_recommended_security_limits`.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Implemented as `with_load_extension_enabled` with the signature you proposed. It enables extension loading, runs the closure, and disables it again, so the window where it is on is as small as possible. The plain `set_load_extension_enabled` setter is now a private helper used only by the callback, so the only public entry point is the scoped one. There is a test asserting the flag is on inside the closure and off afterwards, and I documented that a panic inside the closure leaves it enabled.

docs/sqlite_db_config.svg
LucaCappelletti94: Agreed. I will move both diagrams to the diesel.rs website repository alongside a short guide covering the SQLite-specific features (db_config hardening, the resource limits, custom-function behavior flags, and auto-extensions), and remove them from the crate here. I will open that website PR and drop the SVGs from this PR once it is ready.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -955,13 +953,304 @@ impl SqliteConnection {
...
+    /// does not expose). Extension loading is off by default, and scoping it to a
+    /// closure keeps the window in which it is enabled as small as possible.
+    ///
+    /// Requires SQLite 3.13.0 or later, otherwise returns an error. Has no effect
+    /// if SQLite was compiled with `SQLITE_OMIT_LOAD_EXTENSION`.
+    ///
+    /// # Panics
+    ///
+    /// If `f` panics, extension loading is not disabled again before the panic
+    /// propagates.
```
weiznich: We should catch the panic, roll back the extension setting and rethrow the panic again in this case

diesel_derives/src/tests/snapshots/diesel_derives__tests__declare_sql_function_1 (sqlite).snap
```diff
@@ -120,25 +122,55 @@ pub(crate) mod lower_utils {
...
+    /// Registers an implementation for this function on the given connection,
+    /// with explicit control over the SQLite behavior flags.
+    ///
+    /// This function must be called for every `SqliteConnection` before
+    /// this SQL function can be used on SQLite. Prefer
+    /// [`register_impl`](self::register_impl) (deterministic) or
+    /// [`register_nondeterministic_impl`](self::register_nondeterministic_impl)
+    /// unless you need to set behavior flags explicitly. See
+    /// [`SqliteFunctionBehavior`] for the available flags.
+    pub fn register_impl_with_behavior<F, Ret, input>(
```
weiznich, line 171: This is an ok solution, although `register_impl_with_behavior` now doesn't have the indication behaviour of using `Fn` for deterministic functions anymore.

I don't think we can do much about that so it's probably fine.

weiznich (commented):
This looks good now with the small comment about the panic behaviour of `with_load_extension_enabled()` resolved.

Removing the images would be still a good idea before merging.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: On std builds the closure now runs inside `catch_unwind`. On a panic the method disables extension loading again, as a best effort so a failing disable cannot mask the panic, and then `resume_unwind`s the original panic. On a normal return it disables as before and propagates any disable error. The no-std build keeps the previous straight-line behavior, since `catch_unwind` is not available there, and the doc and `# Panics` section note that. Added a test asserting the flag is disabled again after a panic.

weiznich (approved):
Looks good now, I just changed the default for `SqliteFunctionBehaviour` to be `empty()` as we don't need to mirror the existing `register_impl`  here.

LucaCappelletti94:
Hopefully mysql won't flake again

weiznich:
If that happens I always can rerun the corresponding job :see_no_evil:

PR #4977, opened 2026-02-14
Add last_insert_rowid method to SqliteConnection and RawConnection

Alexi24601:
Exposes SQLite's [`sqlite3_last_insert_rowid()`](https://www.sqlite.org/c3ref/last_insert_rowid.html) through `SqliteConnection`, following the same pattern as the existing `rows_affected_by_last_query`.

Returns `Option<i64>` — `None` if no INSERT has been performed on the connection, `Some(rowid)` otherwise.

### Example

```rust
use diesel::prelude::*;
use diesel::connection::SimpleConnection;

let conn = &mut SqliteConnection::establish(":memory:").unwrap();
conn.batch_execute("CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL)").unwrap();

assert_eq!(conn.last_insert_rowid(), None);

conn.batch_execute("INSERT INTO users (name) VALUES ('Sean')").unwrap();
assert_eq!(conn.last_insert_rowid(), Some(1));

conn.batch_execute("INSERT INTO users (name) VALUES ('Tess')").unwrap();
assert_eq!(conn.last_insert_rowid(), Some(2));
```

### Notes

- Only affected by successful INSERTs into rowid tables
- Not changed by failed INSERTs, UPDATEs, or DELETEs
- Scoped per-connection

@LucaCappelletti94 made me aware of the need to do this, so i tried my hand at this. It is still among the very first pr i do, i am still learning Rust and how to properly use git.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -35,6 +35,7 @@ use crate::sql_types::{HasSqlType, TypeMetadata};
 use crate::sqlite::Sqlite;
 use alloc::string::String;
 use core::ffi as libc;
+use core::num::NonZeroI64;
 
 /// Connections for the SQLite backend. Unlike other backends, SQLite supported
 /// connection URLs are:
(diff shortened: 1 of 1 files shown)


LucaCappelletti94:
@weiznich LGTM, the only thing that could be improved at all could maybe be to replace the INSERT/DELETE raw SQL in the test submodules with actual ORM calls if I really were to be extra nitpicky

diesel/src/sqlite/connection/mod.rs
```diff
@@ -374,6 +374,47 @@ impl SqliteConnection {
...
+    /// conn.batch_execute("INSERT INTO users (name) VALUES ('Sean')")?;
+    /// let rowid = conn.last_insert_rowid();
+    /// assert_eq!(rowid, Some(1));
+    /// conn.batch_execute("INSERT INTO users (name) VALUES ('Tess')")?;
+    /// let rowid = conn.last_insert_rowid();
+    /// assert_eq!(rowid, Some(2));
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn last_insert_rowid(&self) -> Option<i64> {
```
weiznich:
```suggestion
    pub fn last_insert_rowid(&self) -> Option<NonZeroI64> {
```

Given that this value is never zero it might be meaningful to use the `NonZero` type variant instead.

PR #4981, opened 2026-02-19
Reset search path after migration

### LINKED ISSUE #4729: diesel_migrations fails when search_path is set in migration script
## Setup
Any migration script similar to  the following
up.sql
```
SET search_path='foo';

--alternatively you could use:
--SELECT pg_catalog.set_config('search_path', '', false);
```

### Versions

- **Rust:** 1.90
- **Diesel:** 
```
diesel = { version = "2.2.0", features = ["postgres"] }
diesel_migrations = { version = "2.2.0", features = ["postgres"] }
```
- **Database:** postgres
- **Operating System** linux

### Feature Flags

- **diesel:** `postgres`
 - **diesel_migrations:** `postgres`

## Problem Description
diesel_migrations fails when search_path is set in migration script

Perform a migration where the search_path is set

This allows using shortened names for things like types

```
$ diesel migration run
Running migration 2025-08-27-111711_initial
$
```

```
$ diesel migration run
Running migration 2025-08-27-111711_initial
2025-08-27T18:52:00.647148Z ERROR diesel::database: Failed to execute query query=Query { sql: "INSERT INTO \"__diesel_schema_migrations\" (\"version\") VALUES ($1)", binds: [] } err=DatabaseError(Unknown, "relation \"__diesel_schema_migrations\" does not exist")
Failed to run migrations: relation "__diesel_schema_migrations" does not exist
$
```

n/a

### Steps to reproduce
 - create a migration script containing `SET search_path='foo'` or similar
 - run migration script with `diesel migration run` or any other method (eg. programmatically is expected to perform similarly to CLI)

## Checklist

sim-hash:
fixes #4729

### DIFF
--- a/diesel/src/migration/mod.rs
+++ b/diesel/src/migration/mod.rs
@@ -237,13 +237,15 @@ pub trait MigrationConnection: Connection {
     /// ```
     fn setup(&mut self) -> QueryResult<usize>;
 
-    /// Save the current search_path so it can be restored after a migration.
-    fn save_search_path(&mut self) -> QueryResult<Option<String>> {
+    /// Read the current search_path so it can be restored after a migration.
+    #[doc(hidden)]
+    fn read_search_path(&mut self) -> QueryResult<Option<String>> {
         Ok(None)
     }
 
-    /// Restore the search_path previously saved.
-    fn restore_search_path(&mut self, _search_path: &str) -> QueryResult<()> {
+    /// Set the search_path previously saved.
+    #[doc(hidden)]
+    fn set_search_path(&mut self, _search_path: &str) -> QueryResult<()> {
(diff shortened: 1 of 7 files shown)


vanhalt:
Hello!

I tested this PR and it works as expected. Also ran tests in main and contributor repo.

Since I am new to rust, my main concern would be the code implementation. If there is some guidance in what needs to be checked, let me know. Otherwise, this seems solid to merge.

weiznich (approved):
Sorry for letting this sit for too long, I must have just forgotten about it.

The change itself is very reasonable, although I had some minor concerns about exposing all of that as public API. I just pushed a potential fix + a rebase, so this is now ready to be merged from my point of view.

PR #4985, opened 2026-02-23
chore: exporting copy_from traits and struct fields

sinasab:
I was looking into adding `COPY FROM` statement support for postgres to `diesel_async` and found that some things I needed from this crate didn't appear exported.

In this PR I've done a first pass at exporting them. There's a sample of how the new exports are used in `diesel_async` in the draft PR on that repo [here](https://github.com/weiznich/diesel_async/pull/278). Will plan to keep that as a draft until this one settles.

Wasn't sure if this warranted a changelog entry, lmk if so and I can add something!

Thanks for taking a look! 😁

### DIFF
--- a/diesel/src/pg/query_builder/copy/copy_from.rs
+++ b/diesel/src/pg/query_builder/copy/copy_from.rs
@@ -130,16 +130,25 @@ where
     }
 }
 
+/// Trait for types that can be used as the source of a PostgreSQL `COPY FROM` statement.
+///
+/// Implement this trait to support custom backends (e.g. async runtimes like diesel-async)
+/// that need to execute COPY FROM with their own I/O.
 pub trait CopyFromExpression<T> {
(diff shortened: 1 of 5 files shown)


diesel/src/pg/query_builder/copy/copy_from.rs
```diff
@@ -130,16 +130,25 @@ where
     }
 }
 
+/// Trait for types that can be used as the source of a PostgreSQL `COPY FROM` statement.
+///
+/// Implement this trait to support custom backends (e.g. async runtimes like diesel-async)
+/// that need to execute COPY FROM with their own I/O.
 pub trait CopyFromExpression<T> {
```
weiznich, line 137: To be honest, this rather feels like something we need to at least partially mirror in `diesel-async` to make `fn callback` accept an async read variant (likely just `tokio::io::AsyncRead`?) and return a future.

PR #4990, opened 2026-02-27
Fix metrics CI

elbaro:
1. Fix a mysql socket permission bug

`systemctl start mysql.service` starts mysqld configures the unix socket with a correct permission.
Then, `apt-get update` upgrades mysql, apt-get restarts mysql but the unix socket has a wrong permission.

2. Fix a sqlite bench compile error

3. Update zero-mysql

4. Temporarily skip wtx mysql bench

```
Benchmarking bench_medium_complex_query/wtx/1
Benchmarking bench_medium_complex_query/wtx/1: Warming up for 3.0000 s

thread 'main' (9559) panicked at benches/wtx.rs:222:14:
called `Result::unwrap()` on an `Err` value: MysqlDbError(DbError { error_code: 1835, error_message: "Malformed communication packet.", sql_state: Some(HY000) })
```


elbaro:
Test: https://github.com/elbaro/diesel/actions/runs/22469596049

PR #4991, opened 2026-02-27
Add a security and a LLM/AI policy to the projects

weiznich:
This change adds a security.md file outlining our way to handle security issues. It lists disclosing issues via the githubs private secuirty reporting feature. It also explicitly outlines that only the latest diesel release is expected to receive fixes for security issues and encurages uses to keep their versions up to date. Finally it outlines which parts of the project are in scope of such reports (almost everything) and which are not in scope (mostly external things like rust/cargo and dependencies). For me this just codifies what I've tried to follow so far.

In addition this PR also adds a section to the Contribution.md file outlining the allowed usage of LLM's and AI agents in this projects. It is impossible to forbid the usage, but we can put up some reasonable rules. I tried to follow here what NlNet (who funded a part of Diesel developments in the past and might to it again in the future) puts up as policy for projects: https://nlnet.nl/foundation/policies/generativeAI/ The most notable addition is that I would like to disallow the usage of LLM's/Agents for good first issues and mentoring available issues. For me personally these issues are not for people copying my instructions into an LLM and posting the result as PR. I can do that myself. I rather prefer having these issues for new contributors instead.

Finally this PR puts up a PR template pointing people to the right places + asking them to also add a changelog entry for their changes. That is something we often need to ask for afterwards.

---

That's for now my idea how to handle these topics. It's not set in stone, I would like to hear the opinions of a wider audience on these topics. So let's ping @diesel-rs/core and @diesel-rs/reviewers as well. (Anyone else is also welcome to put a comment or suggestion below)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,16 +152,27 @@ To run rustfmt tests locally:
 You can also use rustfmt to make corrections or highlight issues in your editor.
 Check out [their README](https://github.com/rust-lang/rustfmt) for details.
 
-### Usage of LLMs and AI agents
+### Usage of LLM's/AI agents
 
-The Diesel project doesn't completely disallow to usage of LLMs and AI agents for contributions. There are still a number of restrictions and rules you as a PR author are asked to follow.
+The Diesel project does not strictly forbid the usage of LLM's and AI agents for submitting pull requests. 
+There are still a number of restrictions and rules you as a PR author are asked to follow:
 
-Most importantly you as the author of the PR are responsible for carefully reviewing the generated code to make sure that:
+* You need to make sure that the code you are trying to submit can be licences under the relevant open source License used by Diesel. It cannot contain any code that is incompatible with the licenses used by Diesel.
+* You need to fully understand and review any generated code before submitting a PR. The expectation from the reviewer team is 
+  to discuss these changes with you as a person.
+* You need to ensure that the submitted code satisfies the general requirements of submitting PR's to Diesel. That especially includes the following points:
+    + The Code passes all tests and style checks
+    + The change is as minimal as possible. Huge changes will be just dismissed
+    + You verified that the change actually fixes/implements what it is supposed to fix/implement
+    + The change contains a sufficient amount of documentation to help others following the code
+* The usage of LLM's/AI need to be disclosed (See the linked NlNet resource on how to do that in a meaningful way)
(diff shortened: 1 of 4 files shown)


SECURITY.md
```diff
@@ -0,0 +1,58 @@
+# Security Policy for Stalwart
```
Ten0: Stalwart? (I'm not sure what this means. Is that accidentally copy/pasted from a different project?)

CONTRIBUTING.md
```diff
@@ -152,6 +152,27 @@ To run rustfmt tests locally:
...
+* You need to ensure that the submitted code satisfies the general requirements of submitting PR's to Diesel. That especially includes the following points:
+    + The Code passes all tests and style checks
+    + The change is as minimal as possible. Huge changes will be just dismissed
+    + You verified that the change actually fixes/implements what it is supposed to fix/implement
+    + The change contains a sufficient amount of documentation to help others following the code
+* The usage of LLM's/AI need to be disclosed (See the linked NlNet resource on how to do that in a meaningful way)
+* We ask you to write pull request descriptions and discussion comments on your own. 
+  We are able to ask an AI agent on our own if we feel that might be helpful.
+
+The usage of LLM's/AI for issues marked with label including "mentoring available" or "good first issue" is strongly discouraged. These issues are for onboarding new people to the code base. If we would like to fix them using a LLM/AI agent we could have pasted the instructions to such a system
```
Ten0: I guess that wouldn't disturb me so much that they get help from an LLM to get the code *written* for an issue that happens to be a good first issue, so long as the LLM itself is being driven by someone who actually needs to see the issue solved, and consequently can have opinions on the resulting interface.
What disturbs me most with e.g. https://github.com/diesel-rs/diesel/pull/4964 is it looks like some OpenClaw who was told "you're a developer who loves to add fuzzing to OSS projects, go ahead", but it's not really solving an issue that anyone had, or had an actual human review the code it produced before it was proposed, and as such generates review load for irrelevant content.

SECURITY.md
weiznich: To be honest I just copied the basic structure [from stalwart](https://github.com/stalwartlabs/stalwart/blob/main/SECURITY.md) and made adjustments based on what I thought afterwards. Seems like I missed at least one place 🤦 I should have double checked that.
It obviously need to be corrected.

PR #4992, opened 2026-02-28
 Close the SQLite connection on `sqlite3_open_v2` failure to prevent a resource leak

LucaCappelletti94:
Per the SQLite documentation:

> Whether or not an error occurs when it is opened, resources associated with the database connection handle should be released by passing it to sqlite3_close() when it is no longer required.

In `RawConnection::establish`, when `sqlite3_open_v2()` returned a non-OK status, the code returned an error without closing the handle. SQLite may still allocate a connection handle on failure, causing a resource leak.

We can simply call `sqlite3_close()` on the (possibly allocated) handle before returning the error. No null check is needed because sqlite3_close(NULL) is a documented no-op.

### DIFF
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -64,6 +64,12 @@ impl RawConnection {
             }
             err_code => {
                 let message = super::error_message(err_code);
+                // sqlite3_open_v2() may allocate a database connection handle
+                // even on failure. To avoid a resource leak, it must be released
+                // with sqlite3_close(). Passing a null pointer to sqlite3_close()
+                // is a harmless no-op, so no null check is needed.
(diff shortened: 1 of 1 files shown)


Ten0 (approved):

weiznich:
Thanks for fixing that issue.

We might want to investigate to extend our memory check CI job with the following options in the future to catch similar issues more easily:

* Enable the following SQLite options [compile time options](https://sqlite.org/compile.html):
    + SQLITE_ENABLE_API_ARMOR
    + SQLITE_DEBUG
* Enable the following [additional sanitizer](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html):
     + leak
     + memory
* Actually run (parts of) the test suite using miri

Help to address any of these points would be very welcome

PR #4993, opened 2026-02-28
SQLite auto-extension support

LucaCappelletti94:
Thin wrappers around [`sqlite3_auto_extension`](https://www.sqlite.org/c3ref/auto_extension.html), [`sqlite3_cancel_auto_extension`](https://www.sqlite.org/c3ref/cancel_auto_extension.html), and [`sqlite3_reset_auto_extension`](https://www.sqlite.org/c3ref/reset_auto_extension.html), plus an `AutoExtensionEntryPoint` type alias for the entry point signature.

These let users register statically-linked SQLite extensions (e.g. Spatialite, sqlite-vec) that run automatically for every new connection. `sqlite3_auto_extension` has been part of core SQLite since [3.3.8 (2006)](https://sqlite.org/releaselog/3_3_8.html) and is the standard mechanism for extensions compiled into the binary and the only option on WASM (no `dlopen`).

Users can already call `sqlite3_auto_extension` directly through `libsqlite3-sys` (or `sqlite-wasm-rs` on WASM), but getting the types and safety invariants right is fiddly. These wrappers provide a typed signature, proper error mapping, and documented safety requirements so users don't have to piece it together from the SQLite C docs.

### Design

- Free functions, not methods on `SqliteConnection`: these are process-global, not per-connection. Matches existing patterns like `diesel::pg::on_constraint`. That said, associated functions on `SqliteConnection` (without `&self`) might be more discoverable for users.
- `cancel_auto_extension` is safe: it only compares and removes a pointer from SQLite's internal list.
- Docs include verified source links for non-obvious SQLite behaviors (`SQLITE_OK_LOAD_PERMANENTLY` masking, NULL `p_api` on `SQLITE_OMIT_LOAD_EXTENSION` builds, mutex release timing).


diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,301 @@
...
+///
+/// **Note on `sqlite-loadable-rs` interop:** that crate's generated entry
+/// points use `*mut sqlite3_api_routines` and `c_uint` return type, which
+/// differ from this signature (`*const` and `c_int`). You may need a thin
+/// wrapper function to bridge the types.
+pub type AutoExtensionEntryPoint = unsafe extern "C" fn(
+    db: *mut ffi::sqlite3,
+    pz_err_msg: *mut *mut core::ffi::c_char,
+    p_api: *const ffi::sqlite3_api_routines,
+) -> core::ffi::c_int;
```
weiznich: Can we change the signature to this function to something like:

```suggestion
pub type AutoExtensionEntryPoint = fn(
    db: &mut ffi::sqlite3,
) -> Result<(), String>;
```

It seems like there are several constraints around handling return types and error messages, so it seems to be easier to handle that in wrapping function in diesel itself.

For the first parameter it seems more reasonable to pass in a `&mut` reference, even if the user actually needs to use the pointer anyway as that gives them a few additional guarantees (like that the pointer is not null, etc). This likely makes it easier to write sound code.

The third parameter doesn't seem to have any meaningful use as far as I can tell, therefore we don't need to expose it to the user for now?

Setting up the diesel side callback might be a bit tricky, as we would need to have a custom wrapping function that is essentially generic over the user provided function. I'm not 100% sure if you can write such generic code at all. In the worst case we can always put the user provided function pointer in a global `AtomicPtr` and extract it in the callback from there.

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,301 @@
...
+            ctx: *mut ffi::sqlite3_context,
+            _argc: core::ffi::c_int,
+            _argv: *mut *mut ffi::sqlite3_value,
+        ) {
+            unsafe { ffi::sqlite3_result_int(ctx, 42) };
+        }
+
+        let name = c"test_auto_ext_42";
+        unsafe {
+            ffi::sqlite3_create_function_v2(
```
weiznich: After looking at this I wonder whether it even might be meaningful to construct a full `SqliteConnection` for the callback as that makes registering functions, etc much easier. Sounds like a better API for me

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,301 @@
...
+        fn drop(&mut self) {
+            reset_auto_extension();
+        }
+    }
+
+    // All sub-tests are combined into a single #[test] function because
+    // `sqlite3_auto_extension` is process-global state and cannot be
+    // safely mutated from parallel test threads.
+    #[test]
+    fn auto_extension_lifecycle() {
```
weiznich, line 259: It looks like we should make sure that we never run other tests in parallel if that one is running?

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: As I recall there were some cursed reason for that function signature, I will re-read the code and see whether that can be done.

weiznich (commented):
Overall I'm in favor of exposing this API, but I rather would like to refactor it to a safer less error prone variant. As for the current implementation: That's what you already can get without anything from diesel by going via the exposed sqlite pointer.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: Found the _cursed reason_: the old signature was forced by libsqlite3-sys, which hand-writes its [sqlite3_auto_extension](https://sqlite.org/c3ref/auto_extension.html) binding with the full `(db, pzErrMsg, pApi) -> c_int` type, so it could only be moved, not deleted. It now lives in a private trampoline, monomorphized per callback type with a compile-time zero-sized assert, so no global `AtomicPtr` is needed and SQLite's pointer-based dedup/cancel still work.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: The callback now gets `&mut SqliteConnection` via a new `with_borrowed_connection` (a `ManuallyDrop` wrapper that finalizes prepared statements but never closes the borrowed handle). The raw pointer is still reachable through `with_raw_connection` for C extensions, so this is a strict superset.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: The auto-extension tests take a static Mutex and register only a benign collation, so parallel tests are unaffected. The failing-extension path is covered by a test that calls the trampoline directly without touching the global registry. I did not force full test isolation, since cargo has no clean way to give one test exclusive process access.

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,324 @@
...
+where
+    F: Fn(&mut SqliteConnection) -> QueryResult<()> + 'static,
+{
+    // `_p_api` matters only for runtime-loaded shared libraries. Statically
+    // linked extensions link the SQLite symbols directly, so we ignore it.
+    let result: Result<(), String> =
+        crate::util::std_compat::catch_unwind(core::panic::AssertUnwindSafe(|| {
+            // Sound: `F` is provably zero-sized (asserted in the public
+            // entry points), so it carries no state to reconstruct.
+            let extension: F = unsafe { core::mem::zeroed() };
```
weiznich: I asked a bit around and this is considered to be unsound. A zero sized closure still might contain any of the following:

* A uninhabited type -> constructing them is unsound
* A moved in zero sized other type with a non trivial destructor that is used for soundness  by the caller -> We drop the "same" closure several times here.

https://doc.rust-lang.org/std/mem/fn.conjure_zst.html#safety provides a bit more context here.

The good news is that it is possible and not hard to make this sound.
We essentially need to mirror what [`extern_c`](https://docs.rs/extern-c/latest/extern_c/fn.extern_c.html) is doing.

That boils down to:

* We need a `'static` + `Sync` bound on closure as it might be called from a different thread
* We need to call `std::mem::forget` instead of `drop`  in entry point: entry_point
* We need to construct a reference to the closure here, and not a owned variant. `extern_c` uses the following construct for that:

```rust
let out_of_thin_air: [F; 0] = [];
let extension: F = unsafe {
                    // SAFETY:
                    //   - conceptually, this is `SOME_GLOBAL_MAP.get(f)`
                    //   - in practice, since this is a ZST which we have
                    //     `mem::forget`-ten, we can just produce any kind of
                    //      well-aligned and non-deallocated pointer and say it
                    //      is a valid reference.
                    //      The "end" of an empty array properly fits both
                    //      requirements.
                    &*out_of_thin_air.as_ptr()
                };
```

diesel/src/sqlite/connection/mod.rs
```diff
@@ -772,6 +772,48 @@ impl SqliteConnection {
...
+    pub(crate) unsafe fn with_borrowed_connection<R>(
+        db: core::ptr::NonNull<ffi::sqlite3>,
+        f: impl FnOnce(&mut SqliteConnection) -> R,
+    ) -> R {
+        let mut conn = core::mem::ManuallyDrop::new(SqliteConnection {
+            statement_cache: StatementCache::new(),
+            raw_connection: RawConnection {
+                internal_connection: db,
+            },
+            transaction_state: AnsiTransactionManager::default(),
```
weiznich, line 813: We might want to ensure that we are not in a transaction at the point where we destruct the connection again?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -772,6 +772,48 @@ impl SqliteConnection {
...
+            metadata_lookup: (),
+            instrumentation: DynInstrumentation::none(),
```
weiznich: Might be meaningful to get the global instrumentation provider instead?

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,324 @@
+#![allow(unsafe_code)]
```
weiznich: I rather would like to see this moved to the functions that really need an unsafe block, instead of allowing it for the whole file

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,324 @@
...
+/// receives the [`SqliteConnection`] being opened and returns `Ok(())` to
+/// continue or an error to fail the open. Use it to register SQL functions,
+/// collations, or aggregates through the usual connection API, or to initialize
+/// a statically linked C extension such as Spatialite or sqlite-vec (the only
+/// option on WASM, where there is no `dlopen`) via
+/// [`SqliteConnection::with_raw_connection`].
+///
+/// Call this before opening any connection. Extensions run in registration
+/// order, and the first error aborts the open. The callback must be a plain
+/// `fn` item or a non-capturing closure, because SQLite stores only a bare
```
weiznich: Nit: The `non-capturing` closure part is technically not correct as it might capture other zero sized types. We cannot ensure that it really doesn't capture anything.

diesel/src/sqlite/auto_extension.rs
```diff
@@ -0,0 +1,324 @@
...
+                return Err(String::from(
+                    "auto-extension received a null database handle",
+                ));
+            };
+            // SAFETY: `db` is a valid handle for the duration of this call, and
+            // the borrowed connection does not take ownership of it.
+            unsafe { SqliteConnection::with_borrowed_connection(db, |conn| extension(conn)) }
+                .map_err(|e| e.to_string())
+        }))
+        .unwrap_or_else(|_| Err(String::from("auto-extension panicked")));
```
weiznich: Can we extract the panic message here and include it into the error message as well?

weiznich (commented):
Thanks for the update. Overall this looks much better, but it needs some changes to the closure reconstruction to make this sound.

Otherwise there are a number of minor nits that would be nice to see addressed.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: The trampoline downcasts the panic payload to `&str`/`String` and includes it as `auto-extension panicked: <message>`, with a generic fallback for other payloads. It is std-only since the no_std `catch_unwind` carries no payload. Covered by a test asserting the payload round-trips.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: Fixed, reworded to "a `fn` item or a closure that captures only zero-sized values".

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: Removed the file-level `#![allow(unsafe_code)]` and moved it to per-item `#[allow(unsafe_code)]` on the functions that need it.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: It now uses `DynInstrumentation::default_instrumentation()`.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Added a `debug_assert!` that the transaction depth is `Ok(None)` before the borrowed connection is torn down.

diesel/src/sqlite/auto_extension.rs
LucaCappelletti94: Added `+ Sync + 'static`, switched `entry_point` to `core::mem::forget` instead of `drop`, and the trampoline now reconstructs a `&F` reference rather than an owned value. I used `core::ptr::NonNull::<F>::dangling().as_ref()` instead of the empty-array form, which to my understanding yields the same well-aligned non-null reference to a ZST. Added a regression test that asserts a captured ZST guard's `Drop` never runs.

weiznich (approved):
Looks good now, thanks for the update

PR #4994, opened 2026-03-02
Implement BLOB reading for sqlite

matthiasbeyer:
This patch implements BLOB reading via the sqlite3_blob_read() API.

--

We (@weiznich and I) talked about this on [mastodon](https://social.linux.pizza/@musicmatze/116115172910294034).

This is a first proposal to see what you think of it.

This patch would not have been possible without the tremendous help from @TheNeikos who is mentioned as a co-author in the patch. Thank you very much for your help :heart:

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -10,7 +10,7 @@ mod owned_row;
 mod raw;
 mod row;
 mod serialized_database;
-mod sqlite_blob;
+pub(in crate::sqlite) mod sqlite_blob;
 mod sqlite_value;
 mod statement_iterator;
 mod stmt;
@@ -430,23 +430,24 @@ impl SqliteConnection {
     /// # fn main() {
     /// #     run_test().unwrap();
     /// # }
-    /// # fn run_test() -> QueryResult<()> {
+    /// # fn run_test() -> Result<(), Box<dyn std::error::Error>> {
     /// use std::io::Read;
     /// use diesel::connection::SimpleConnection;
     /// let conn = &mut SqliteConnection::establish(":memory:").unwrap();
     /// conn.batch_execute("CREATE TABLE myblobs (id INTEGER PRIMARY KEY, mydata BLOB)")?;
     /// conn.batch_execute("INSERT INTO myblobs (mydata) VALUES ('abc')")?;
-    /// let mut data = conn.get_blob::<myblobs::mydata>(1).unwrap();
+    /// let mut data = conn.get_read_only_blob(myblobs::mydata, 1)?;
     /// let mut buf = vec![];
-    /// data.read_to_end(&mut buf).unwrap();
+    /// data.read_to_end(&mut buf)?;
     /// assert_eq!(buf, b"abc");
     /// # Ok(())
     /// # }
     /// ```
-    pub fn get_blob<'conn, 'query, U>(
+    pub fn get_read_only_blob<'conn, 'query, U>(
         &'conn self,
-        primary_key: i64,
-    ) -> Result<sqlite_blob::SqliteBlob<'conn>, Error>
+        blob_column: U,
+        row_id: i64,
+    ) -> Result<sqlite_blob::SqliteReadOnlyBlob<'conn>, Error>
     where
         'query: 'conn,
         U: crate::Column,
@@ -454,13 +455,15 @@ impl SqliteConnection {
         <U::Table as nodes::StaticQueryFragment>::Component: HasDatabaseAndTableName,
     {
         use crate::query_builder::nodes::StaticQueryFragment;
+        // this mostly exists for a more natural way to call this function
+        let _ = blob_column;
 
         let database_name = U::Table::STATIC_COMPONENT.database_name().unwrap_or("main");
         let column_name = U::NAME;
         let table_name = U::Table::STATIC_COMPONENT.table_name();
 
         self.raw_connection
-            .blob_open(database_name, table_name, column_name, primary_key)
+            .blob_open(database_name, table_name, column_name, row_id)
     }
 
     fn transaction_sql<T, E, F>(&mut self, f: F, sql: &str) -> Result<T, E>
@@ -1605,13 +1608,13 @@ mod tests {
         )
         .execute(conn);
 
-        let mut data = conn.get_blob::<blobs::data>(1).unwrap();
+        let mut data = conn.get_read_only_blob(blobs::data, 1).unwrap();
         let mut buf = vec![];
         data.read_to_end(&mut buf).unwrap();
 
         assert_eq!(buf, b"abc");
 
-        let mut data2 = conn.get_blob::<blobs::data2>(1).unwrap();
+        let mut data2 = conn.get_read_only_blob(blobs::data2, 1).unwrap();
         let mut buf = vec![];
         data2.read_to_end(&mut buf).unwrap();
 
@@ -1638,7 +1641,7 @@ mod tests {
 
         let _ = crate::sql_query("INSERT INTO blobs (data) VALUES ('abcdefghi')").execute(conn);
 
-        let mut data = conn.get_blob::<blobs::data>(1).unwrap();
+        let mut data = conn.get_read_only_blob(blobs::data, 1).unwrap();
 
         let mut buf = [0; 1];
         assert_eq!(data.read(&mut buf).unwrap(), 1);
@@ -1689,7 +1692,7 @@ mod tests {
 
         let _ = crate::sql_query("INSERT INTO blobs (data) VALUES ('abc')").execute(conn);
 
-        let data = conn.get_blob::<blobs::data>(1).unwrap();
+        let data = conn.get_read_only_blob(blobs::data, 1).unwrap();
         drop(data);
 
         let _ = crate::sql_query("INSERT INTO blobs (data) VALUES ('def')").execute(conn);
@@ -1714,7 +1717,7 @@ mod tests {
         let _ = crate::sql_query("INSERT INTO blobs (data) VALUES ('abc')").execute(conn);
 
         {
-            let mut data = conn.get_blob::<blobs::data>(1).unwrap();
+            let mut data = conn.get_read_only_blob(blobs::data, 1).unwrap();
             let mut buf = vec![];
             data.read_to_end(&mut buf).unwrap();
             assert_eq!(buf, b"abc");
@@ -1723,7 +1726,7 @@ mod tests {
         let res = conn.exclusive_transaction(|conn| {
             crate::sql_query("UPDATE blobs SET data = 'def' WHERE id = 1").execute(conn)?;
 
-            let mut data = conn.get_blob::<blobs::data>(1).unwrap();
+            let mut data = conn.get_read_only_blob(blobs::data, 1).unwrap();
             let mut buf = vec![];
             data.read_to_end(&mut buf).unwrap();
             assert_eq!(buf, b"def");
@@ -1733,7 +1736,7 @@ mod tests {
 
         assert_eq!(res.unwrap_err(), Error::RollbackTransaction);
 
-        let mut data = conn.get_blob::<blobs::data>(1).unwrap();
+        let mut data = conn.get_read_only_blob(blobs::data, 1).unwrap();
         let mut buf = vec![];
         data.read_to_end(&mut buf).unwrap();
         assert_eq!(buf, b"abc");
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -279,26 +279,22 @@ impl RawConnection {
         database_name: &str,
         table_name: &str,
         column_name: &str,
-        pkey: i64,
-    ) -> Result<super::sqlite_blob::SqliteBlob<'conn>, Error> {
+        row_id: i64,
+    ) -> Result<super::sqlite_blob::SqliteReadOnlyBlob<'conn>, Error> {
         let database_name = alloc::ffi::CString::new(database_name)?;
         let column_name = alloc::ffi::CString::new(column_name)?;
         let table_name = alloc::ffi::CString::new(table_name)?;
 
         let mut blob: *mut ffi::sqlite3_blob = core::ptr::null_mut();
 
-        // SAFETY: From the SQLite docs:
-        //
-        // > For the main database file, the database name is "main"
-        //
-        // The other variables are properly initialized
+        // SAFETY: All variables are properly initialized
         let ret = unsafe {
             ffi::sqlite3_blob_open(
                 self.internal_connection.as_ptr(),
                 database_name.as_c_str().as_ptr(),
                 table_name.as_c_str().as_ptr(),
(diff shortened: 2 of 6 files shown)


diesel/src/sqlite/connection/raw.rs
```diff
@@ -273,6 +273,31 @@ impl RawConnection {
...
+
+        let blob = unsafe {
+            core::ptr::NonNull::new_unchecked(blob)
+        };
+
+        let blob_size = unsafe {
+            ffi::sqlite3_blob_bytes(blob.as_ptr())
+        };
+
+        let blob_size = usize::try_from(blob_size).unwrap(); // TODO: What to do?
```
matthiasbeyer: @weiznich We are not entirely sure how diesel should handle this. The source of a possible error here is, if you're on a 32 bit platform and you want to open a blob which is larger than 2^32-1 bytes.
Should we add a new `Error` variant for that case? Should we use an existing one? Should we simply panic?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -414,6 +415,20 @@ impl SqliteConnection {
         NonZeroI64::new(self.raw_connection.last_insert_rowid())
     }
 
+    pub fn get_blob<'conn, 'query, U>(&'conn self, primary_key: i64) -> Result<sqlite_blob::SqliteBlob<'conn>, Error>
```
matthiasbeyer: Documentation missing here, yes.

weiznich:
Thanks for working on this. I (or someone else) will try to have a look at the implementation in the next days/weeks.

cc @LucaCappelletti94 as he has done quite a lot of Sqlite api additions in the last weeks and might also help here

diesel/src/sqlite/connection/raw.rs
TheNeikos: Another alternative is to use u64 all the way, which should be fine on 64bit arches. I'm not sure how much Rust (And specifically diesel) is used on 32bit machines.

TheNeikos:
One thing I'd like to mention wrt. Blobs in SQLITE and the current interface is that one cannot hold Blobs while executing any other command.

This is fine and the easiest way from a technical standpoint. However I feel it does preclude from more interesting applications like having proper-streaming or similar backed by a sqlite database. The docs seem fairly clear on this:

- [`sqlite_close`](https://sqlite.org/c3ref/close.html) specifically mentions BLOB handles:

> If the database connection is associated with unfinalized prepared statements, **BLOB handlers**, and/or unfinished sqlite3_backup objects then sqlite3_close() will leave the database connection open and return [**SQLITE_BUSY**](https://sqlite.org/rescode.html#busy).
(Emphasis mine)

As well as in [`blob_open`](https://sqlite.org/c3ref/blob_open.html):

> If the row that a BLOB handle points to is **modified** by an [UPDATE](https://sqlite.org/lang_update.html), [DELETE](https://sqlite.org/lang_delete.html), or by [ON CONFLICT](https://sqlite.org/lang_conflict.html) side-effects then the BLOB handle is marked as "expired". This is true if any column of the row is changed, even a column other than the one the BLOB handle is open on. Calls to [sqlite3_blob_read()](https://sqlite.org/c3ref/blob_read.html) and [sqlite3_blob_write()](https://sqlite.org/c3ref/blob_write.html) for an **expired BLOB handle fail with a return code of** [SQLITE_ABORT](https://sqlite.org/rescode.html#abort). Changes written into a BLOB prior to the BLOB expiring are not rolled back by the expiration of the BLOB. Such changes will eventually commit if the transaction continues to completion.

This means that the lifetime of a Blob _is_ tied to a database, but not in a way that Rust can properly check. The current interface also means that users cannot move a blob handle from where it was created (due to blob being bound to the lifetime of the connection).

As far as I can tell it is safe to not have the lifetime. However, this would introduce the hazard of panicking when dropping the SQLite connection while having any blobs open (as it would return SQLITE_BUSY on close).

None of these are real 'problems' IMO, just decisions to be made one way or another.

I believe a potential path forward could be something akin to a `detach` method on the blob that would make it `'static` and put the onus on the user to understand what this entails.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -273,6 +273,45 @@ impl RawConnection {
...
+    ) -> Result<super::sqlite_blob::SqliteBlob<'conn>, Error> {
+        let column_name = alloc::ffi::CString::new(column_name)?;
+        let table_name = alloc::ffi::CString::new(table_name)?;
+
+        let mut blob: *mut ffi::sqlite3_blob = core::ptr::null_mut();
+
+        let ret = unsafe {
+            ffi::sqlite3_blob_open(
+                self.internal_connection.as_ptr(),
+                c"main".as_ptr(),
```
weiznich: Given that diesel also allows to specify custom schemas we should likely allow that if a table with a custom schema is specified that's passed as db name here, if that's not the case we use `main` as database name.

That likely requires in `src/sqlite/connection.mod` to abstract over [Identifier](https://github.com/diesel-rs/diesel/blob/main/diesel/src/query_builder/nodes/mod.rs#L35) (currently used) and [InfixNode](https://github.com/diesel-rs/diesel/blob/main/diesel/src/query_builder/nodes/mod.rs#L55) (used for tables with an explicit schema). That likely requires having a small local non-public trait that is implement for both types and gets you the table name and the schema name.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -273,6 +273,45 @@ impl RawConnection {
...
+                column_name.as_c_str().as_ptr(),
+                pkey,
+                0,
+                &mut blob,
+            )
+        };
+
+        Self::process_sql_function_result(ret)?;
+
+        let blob = unsafe { core::ptr::NonNull::new_unchecked(blob) };
```
weiznich, line 312: I would rather like to check for a null pointer here, just to be safe.

diesel/src/sqlite/connection/raw.rs
weiznich: It seems like [this function](https://sqlite.org/c3ref/blob_bytes.html) returns a `int`, which should be a 32 bit integer on most platforms? 
That would mean that the only platform supported by rust this can fail on are 16 bit targets. While I got diesel + sqlite working in an embedded context I'm not aware of anyone doing that in a 16 bit environment, so it might be fine to just return an error for the overflow case.

diesel/src/sqlite/connection/sqlite_blob.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn drop(&mut self) {
+        use crate::util::std_compat::panicking;
+
+        // SAFETY: From the sqlite3_blob_close documentation:
+        //
+        //     If an error occurs while committing the transaction, an error code is returned and
+        //     the transaction rolled back.
+        //
+        // As we are in read-only mode here, this is not an issue
+        let close_result = unsafe { ffi::sqlite3_blob_close(self.blob.as_ptr()) };
```
weiznich: Can we have an explicit close function that returns an error? 

This function should be called from `Drop` as well but it would allow users to explicitly handle these errors if required. (Basically similar to how `std::fs::File` works)

diesel/src/sqlite/connection/sqlite_blob.rs
```diff
@@ -0,0 +1,117 @@
...
+        let offset: i32 = self.read_index.try_into().map_err(to_io_error)?;
+
+        // From the sqlite docs:
+        //
+        //  If offset iOffset is less than N bytes from the end of the BLOB, SQLITE_ERROR is returned and no data is read.
+        //
+        // Thus we need to make sure to not provide a buffer that is too big for the remaining data
+        // from the blob.
+        let read_length: i32 =
+            (i32::try_from(self.blob_size).map_err(to_io_error)? - offset).min(buflen);
```
weiznich: Better use `saturating_sub` here to be safe

diesel/src/sqlite/connection/sqlite_blob.rs
```diff
@@ -0,0 +1,117 @@
...
+                buf.as_mut_ptr() as *mut core::ffi::c_void,
+                read_length,
+                offset,
+            )
+        };
+
+        if ret != ffi::SQLITE_OK {
+            let error_message = crate::sqlite::connection::error_message(ret);
+            return Err(std::io::Error::other(error_message.to_string()));
+        }
```
weiznich, line 117: Also use the check error function from the raw connection module here?

diesel/src/sqlite/connection/sqlite_blob.rs
```diff
@@ -0,0 +1,117 @@
...
+                #[cfg(feature = "std")]
+                eprintln!("Error closing SQLite blob: {error_message}");
+            } else {
+                panic!("Error closing SQLite blob: {error_message}");
+            }
+        }
+    }
+}
+
+impl SqliteBlob<'_> {
```
weiznich: We likely also want to implement `Write`?

That doesn't necessarily need to be done in this pr, but the design should allow it.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -273,6 +273,45 @@ impl RawConnection {
...
+        let mut blob: *mut ffi::sqlite3_blob = core::ptr::null_mut();
+
+        let ret = unsafe {
+            ffi::sqlite3_blob_open(
+                self.internal_connection.as_ptr(),
+                c"main".as_ptr(),
+                table_name.as_c_str().as_ptr(),
+                column_name.as_c_str().as_ptr(),
+                pkey,
+                0,
```
weiznich, line 298: We need a parameter to control this flag. Maybe even as compile time flag (generic parameter?)

Otherwise an `enum` could also be fine. For now that could just have a read variant and be marked as `#[non_exhaustive]`

weiznich (commented):
Thanks for working on this. Overall this looks like a good start already. I've left a bunch of comments around the API design I would like to discuss and potentially address before merging.

I still need to look a the tests to decide if we want to add more cases.

Finally I think that's a feature that deserves a changelog entry, so please at it to the top level `Changelog.md` file.

diesel/src/sqlite/connection/raw.rs
TheNeikos: This does not seem to be about schemas, from `blob_open`:

> Parameter zDb is not the filename that contains the database, but rather the symbolic name of the database. For attached databases, this is the name that appears after the AS keyword in the [ATTACH](https://sqlite.org/lang_attach.html) statement. For the main database file, the database name is "main". For TEMP tables, the database name is "temp".

Does diesel allow attaching extra databases to the same connection? As far as I can tell it does not explicitly.

weiznich: Diesel doesn't explicitly expose an attach database functions, but you certainly can just do a `sql_query("ATTACH '/tmp/foo.db' AS FOO")` and then refer to tables in this database via `foo.table_name`, which exactly mirrors what a table defined as

```rust
mod foo {
    table! {
        foo.table_name {
        // your columns here
        }
    }
}
```

would generate for `foo::table_name::table.load(conn)` would generate (`SELECT * FROM foo.table`).

With that in mind it only seems reasonable to reuse that as "database" name here as well.

diesel/src/sqlite/connection/sqlite_blob.rs
matthiasbeyer: I am not sure how `RawConnection::process_sql_function_result()` (assuming you mean that function) is of help here, as we need to return a `std::io::Error`, so we need to use `std::io::Error::other` here, right? 🤔

diesel/src/sqlite/connection/raw.rs
matthiasbeyer: I see what you mean here. If you don't mind, I would include `Write` functionality in this very same PR later down the road and do that bit of refactoring for a generic parameter (or something like that) then.

But first I would like to get the `Read` functionality in a shape that's `ACK`'d by you or other diesel maintainers!

diesel/src/sqlite/connection/sqlite_blob.rs
matthiasbeyer: Yes, I would like to include that later down the road!

matthiasbeyer:
I pushed a fixup to address a number of your review points, where I did not have any questions! Also included some `SAFETY` comments in the codebase now.

diesel/src/sqlite/connection/raw.rs
matthiasbeyer: We addressed this in the [latest fixup commit](https://github.com/diesel-rs/diesel/pull/4994/changes/f6c536c9ad66612238eebfbc94ede4e868ffddb5).

diesel/src/sqlite/connection/mod.rs
```diff
@@ -414,6 +415,26 @@ impl SqliteConnection {
         NonZeroI64::new(self.raw_connection.last_insert_rowid())
     }
 
+    pub fn get_blob<'conn, 'query, U>(
```
LucaCappelletti94: Undocumented public method, maybe a documentation with doctests would be nice to have?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -745,6 +766,34 @@ fn error_message(err_code: libc::c_int) -> &'static str {
     ffi::code_to_str(err_code)
 }
 
+#[doc(hidden)]
```
weiznich: Minor nit: I usually put these traits in a local `mod private {}` and then reexport them as crate local via `pub(crate) use self::private::HasDatabaseAndTableName;` in the parent scope. That ensures that they are really not usable by downstream crates.

diesel/src/sqlite/connection/raw.rs
weiznich: It's fine to also postpone the write functionality to a later PR.

This comment was mostly about API design.

diesel/src/sqlite/connection/sqlite_blob.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// Close the handle
+    ///
+    /// Even if an error is returned, the handle is still closed (from the sqlite documentation):
+    ///
+    /// > The BLOB handle is closed unconditionally. Even if this routine returns an error code,
+    /// > the handle is still closed.
+    ///
+    /// Using the handle after calling this function is an error if not otherwise stated.
+    pub fn close(&mut self) -> Result<(), crate::result::Error> {
```
weiznich: Maybe rather expose this publically as consuming method (so `pub fn clone(self)`) and only internally have a `fn close_intern(&mut self)` method that can be shared between `Drop` and this function?

Seems like an easy way to remove this potential error path

diesel/src/sqlite/connection/sqlite_blob.rs
weiznich: Ah right, consider that to be resolved then

matthiasbeyer:
I finally came around to clean up some of your suggestions!

matthiasbeyer:
Don't you want me to clean up the fixup commits? :eyes:

weiznich:
> Don't you want me to clean up the fixup commits? 👀

Thanks for the reminder. That's done now

PR #5004, opened 2026-03-22
Improve `#[derive(AsChangeset)]` error messages

### LINKED ISSUE #5001: Cannot derive `AsChangeset` when a `#[diesel(embed)]`-ded field does not implement `&T: AsChangeset`
### Rust Version

1.94.0

### Diesel Version

2.3.7

### Diesel Features

sqlite returning_clauses_for_sqlite_3_35 uuid chrono

### Database Version

Sqlite 3.43.2

### Operating System Version

macOS 15.7.3 (Sequoia) arm64

### What happened?

See the reproduction example below.

To me, it looks like, if `#[diesel(serialize_as)]` or `#[diesel(deserialize_as)]` is present in the struct definition (in this example, `MyTable`), the derive macro `#[derive(AsChangeset)]` always tries to implement `AsChangeset` for `&'update MyTable`.

However, if the struct contains a field with `#[diesel(embed)]` and its type (I've called it `Embed`) does not implement `&'update Embed: AsChangeset`, this leads to a compilation error.

The error goes away in both the following cases:
- if I comment the line marked with `(1)`: `&'update Embed: AsChangeset` is now derived and the implementation of `&'update MyTable: AsChangeset` works;
- if I uncomment the line marked with `(2)`: `#[derive(AsChangeset)]` detects the use of `#[diesel(serialize_as)]`/`#[diesel(deserialize_as)]` and doesn't try to implement `&'update MyTable: AsChangeset`.

I think `#[derive(AsChangeset)]` should not try and implement `&'update MyTable: AsChangeset` if at least one field has a type that does not implement the trait by reference (especially `#[diesel(embed)]`ded fields).
However, it should implement `MyTable: AsChangeset` correctly.

### Additional details

_No response_

### Steps to reproduce

```rust
use diesel::{
    deserialize::{FromSql, FromSqlRow},
    expression::AsExpression,
    prelude::*,
    serialize::ToSql,
    sql_types::Text,
    sqlite::Sqlite,
};

#[derive(Debug, Clone, PartialEq, Eq, AsExpression, FromSqlRow)]
#[diesel(sql_type = Text)]
struct Foo(String);
impl From<String> for Foo {
    fn from(value: String) -> Self {
        Foo(value)
    }
}
impl Into<String> for Foo {
    fn into(self) -> String {
        self.0
    }
}
impl FromSql<Text, Sqlite> for Foo {
    fn from_sql(
        bytes: <Sqlite as diesel::backend::Backend>::RawValue<'_>,
    ) -> diesel::deserialize::Result<Self> {
        <String as FromSql<Text, Sqlite>>::from_sql(bytes).map(Self)
    }
}
impl ToSql<Text, Sqlite> for Foo {
    fn to_sql<'b>(
        &'b self,
        out: &mut diesel::serialize::Output<'b, '_, Sqlite>,
    ) -> diesel::serialize::Result {
        ToSql::<Text, Sqlite>::to_sql(&self.0, out)
    }
}

diesel::table! {
    my_table(id) {
        id -> Text,
        foo -> Text,
        bar -> Text,
    }
}

#[derive(Queryable, Selectable, AsChangeset)]
#[diesel(table_name = my_table)]
struct Embed {
    #[diesel(serialize_as = String, deserialize_as = String)] // <-- (1)
    foo: Foo,
}

#[derive(HasQuery, AsChangeset)]
#[diesel(table_name = my_table)]
struct MyTable {
    id: String,
    //#[diesel(serialize_as = String, deserialize_as = String)] // <-- (2)
    bar: Foo,
    #[diesel(embed)]
    embed: Embed,
}
```

### Compile time error

```
error[E0277]: the trait bound `&'update Embed: diesel::AsChangeset` is not satisfied
  --> src/lib/test.rs:54:20
   |
54 | #[derive(HasQuery, AsChangeset)]
   |                    ^^^^^^^^^^^ the trait `diesel::AsChangeset` is not implemented for `&'update Embed`
   |
help: the trait `diesel::AsChangeset` is implemented for `Embed`
  --> src/lib/test.rs:47:33
   |
47 | #[derive(Queryable, Selectable, AsChangeset)]
   |                                 ^^^^^^^^^^^
   = note: required for `(Grouped<Eq<bar, Bound<Text, &Foo>>>, &Embed)` to implement `diesel::AsChangeset`
   = note: the full name for the type has been written to '…'
   = note: this error originates in the derive macro `AsChangeset` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Checklist

weiznich:
This commit seeks to improve error messages emitted for field type
mismatches in `#[derive(AsChangeset)]`. Before this change the error
only pointed at the derive itself, with this change we now have at least
one error (in the begining of the error block) to point at a specific
broken field. This is by no means perfect, but it's the only thing that
I managed to get out here.

On the side it also fixes #5001 by adding the relevant bounds to prevent
the borrowed `AsChangeset` impl to error out if an inner embedded
borrowed `AsChangeset` impl doesn't exist due to the usage of
`#[diesel(serialize_as)]`

Finally this contains minor diagnostic fixes like hiding some of the
`AsExpression` impls for double references as users do not really need
to know about them.

See [here](https://github.com/diesel-rs/diesel/pull/5004/changes/0a5011957474a41b8d9d7e77461f59cf0099633a#diff-8628a4912f214ad401b01db6451ebb7dad97d6549343908bd7b91df69b7f28e5) for a diff between the old and new error message

### DIFF
--- a/diesel_derives/src/as_changeset.rs
+++ b/diesel_derives/src/as_changeset.rs
@@ -1,3 +1,5 @@
+use std::collections::{HashMap, HashSet};
+
 use proc_macro2::{Span, TokenStream};
 use quote::{quote, quote_spanned};
 use syn::spanned::Spanned as _;
@@ -45,6 +47,12 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
     let mut ref_field_ty = Vec::with_capacity(fields_for_update.len());
     let mut ref_field_assign = Vec::with_capacity(fields_for_update.len());
 
+    // Explicit trait bounds to improve error messages
+    let mut field_ty_bounds = Vec::with_capacity(model.fields().len());
+    let mut borrowed_field_ty_bounds = Vec::with_capacity(model.fields().len());
+    let mut field_ty_bounds_guard = HashMap::new();
+    let mut borrowed_field_ty_bounds_guard = HashMap::new();
+
     for field in fields_for_update {
         // skip this field while generating the update
         if field.skip_update() {
@@ -84,6 +92,14 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
                     field,
                     table_name,
                     ty,
+                    treat_none_as_null,
+                )?);
+                field_ty_bounds.push(generate_field_bound(
+                    field,
+                    table_name,
+                    Some(ty),
+                    None,
+                    &mut field_ty_bounds_guard,
                     treat_none_as_null,
                 )?);
 
@@ -126,11 +142,41 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
                     Some(quote!(&)),
                     treat_none_as_null,
                 )?);
+
+                field_ty_bounds.push(generate_field_bound(
+                    field,
+                    table_name,
+                    None,
+                    None,
+                    &mut field_ty_bounds_guard,
+                    treat_none_as_null,
+                )?);
+
+                borrowed_field_ty_bounds.push(generate_field_bound(
+                    field,
+                    table_name,
+                    None,
+                    Some(parse_quote!('update)),
+                    &mut borrowed_field_ty_bounds_guard,
+                    treat_none_as_null,
+                )?);
             }
         }
     }
 
+    let field_ty_bounds = field_ty_bounds
+        .into_iter()
+        .filter_map(|(type_to_check, bound)| {
+            super::insertable::filter_bounds(&field_ty_bounds_guard, type_to_check, bound)
+        });
+
+    let tpe_lifetimes = item.generics.lifetimes();
+
     let changeset_owned = quote! {
+        fn _check_owned<#(#tpe_lifetimes,)*>()
+        where #(#field_ty_bounds,)*
+        {}
+
         impl #impl_generics diesel::query_builder::AsChangeset for #struct_name #ty_generics
         #where_clause
         {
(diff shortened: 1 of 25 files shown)


diesel_derives/src/as_changeset.rs
```diff
@@ -100,6 +117,7 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
                 direct_field_assign.push(field_changeset_expr_embed(field, None));
                 ref_field_ty.push(field_changeset_ty_embed(field, Some(quote!(&'update))));
                 ref_field_assign.push(field_changeset_expr_embed(field, Some(quote!(&))));
+                embedded_types.push(&field.ty);
```
ayarotsky: `embedded_types` does not seem to be used anywhere.

diesel_derives/src/as_changeset.rs
```diff
@@ -147,10 +195,38 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
...
+            borrowed_field_ty_bounds
+                .into_iter()
+                .filter_map(|(type_to_check, bound)| {
+                    super::insertable::filter_bounds(
+                        &borrowed_field_ty_bounds_guard,
+                        type_to_check,
+                        bound,
+                    )
+                });
+        let tpe_lifetimes = item.generics.lifetimes().map(|lp| {
```
ayarotsky, line 206: It redefines the definition at 175. Was it intentional?

diesel_derives/src/as_changeset.rs
```diff
@@ -147,10 +195,38 @@ pub fn derive(item: DeriveInput) -> Result<TokenStream> {
...
+            let lt = &lp.lifetime;
+            let bound = lp.bounds.iter();
+            if lp.bounds.is_empty() {
+                quote!(#lt)
+            } else {
+                quote!(#lt: #(#bound +)*)
+            }
+        });
+        let tpe_lifetimes2 = item.generics.lifetimes().map(|lp| &lp.lifetime);
```
ayarotsky: What do you think if we name it `lifetime_idents` to better distinguish it from `tpe_lifetimes`?

diesel_derives/src/as_changeset.rs
weiznich: Yes, the definition variable defined in line 174 is an iterator, that is then consumed by the `quote` macro in between.

diesel_derives/src/as_changeset.rs
weiznich: I just removed it entirely in favor of adding the relevant bound in the `tpe_lifetimes` iterator above as well. That should be cleaner.

PR #5010, opened 2026-03-29
fix: add where bounds for embed fields in borrowed AsChangeset impl

### LINKED ISSUE #5001: Cannot derive `AsChangeset` when a `#[diesel(embed)]`-ded field does not implement `&T: AsChangeset`

MavenRain:
When a struct with #[diesel(embed)] derives AsChangeset, the borrowed impl now requires &'update EmbedType: AsChangeset in its where clause. This prevents a hard compile error (E0277) when the embedded type does not have a borrowed AsChangeset impl (e.g. because it uses serialize_as).

Closes #5001

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,42 @@
+name: Docs
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - run: cargo doc --no-deps --all-features --workspace
+      - name: Create redirect to first crate
+        run: |
+          FIRST=$(find target/doc -maxdepth 2 -name index.html \
+            | sed 's|target/doc/\([^/]*\)/index.html|\1|' \
+            | grep -v '^target' \
+            | sort | head -1)
+          echo "<meta http-equiv=\"refresh\" content=\"0; url=${FIRST}/index.html\">" > target/doc/index.html
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: target/doc
+  deploy:
+    needs: build
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - id: deployment
+        uses: actions/deploy-pages@v4
--- a/diesel_derives/src/tests/as_changeset.rs
+++ b/diesel_derives/src/tests/as_changeset.rs
@@ -156,3 +156,21 @@ pub(crate) fn as_changeset_treat_skip_update_1() {
         "as_changeset_treat_skip_update_1",
     );
 }
+
+#[test]
+pub(crate) fn as_changeset_embed_with_where_clause_1() {
+    let input = quote::quote! {
+        struct User {
+            id: i32,
+            bar: String,
+            #[diesel(embed)]
+            embed: Embed
+        }
+    };
+    expand_with(
+        &crate::derive_as_changeset_inner as &dyn Fn(_) -> _,
+        input,
+        derive(syn::parse_quote!(#[derive(AsChangeset)])),
+        "as_changeset_embed_with_where_clause_1",
+    );
+}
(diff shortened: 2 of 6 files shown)


weiznich:
@MavenRain Thanks for working on this. Would it be possible to address the CI errors? They seem to related to your changes

.github/workflows/docs.yml
weiznich, line 1: Why is this file added? This doesn't seem useful for me

It shouldn't be there

diesel_derives/src/tests/as_changeset.rs
```diff
@@ -156,3 +156,21 @@ pub(crate) fn as_changeset_treat_skip_update_1() {
         "as_changeset_treat_skip_update_1",
     );
 }
+
+#[test]
```
weiznich, line 160: Given that we already have a test above that verifies the code generated by the `#[diesel(embed)]` attribute, this shouldn't be needed.

weiznich (changes requested):
Thanks for the update.

There is a github actions CI file that does seem to be totally unrelated to this PR added here. This needs to be removed.

Also I would like to add the example from the issue as complete compile test in `diesel_derives/tests/as_changeset.rs`.

Could you update the PR to fix that?

weiznich:
Already fixed by #5004

PR #5011, opened 2026-04-01
Document migration transaction behaviour at the crate root

### LINKED ISSUE #4855: Document transaction behaviour of migrations
### Rust Version

1.90.0

### Diesel Version

2.3.2

### Diesel Features

"postgres", "postgres_backend", "uuid", "chrono", "r2d2", "numeric"

### Database Version

PostgreSQL 17.4

### Operating System Version

Arch Linux

### What happened?

The current behavior about if/when migrations are run inside transactions is not documented.
From a search through the code it seems like they run in transactions by default, but there might be cases where that default is not used.

It would be nice if the behavior about if/when migrations are run inside transactions would be documented on https://docs.rs/diesel_migrations/latest/diesel_migrations/ and/or https://diesel.rs/guides/getting-started 

### Additional details

For more details see: https://github.com/diesel-rs/diesel/discussions/4854

### Steps to reproduce

```rust

```

### Compile time error

_No response_

### Checklist

ayarotsky:
The existing transaction documentation in `FileBasedMigrations` was not linked from anywhere discoverable. This adds a `## Transactions` section to the `diesel_migrations` crate root doc, which is the landing page on docs.rs, summarising the default behaviour and linking to `FileBasedMigrations`, `RustMigration::without_transaction`, and `TypedMigration::run_in_transaction` for each opt-out path.

`MigrationHarness::run_migration` and `revert_migration` also get a brief note explaining that the default implementation wraps the operation in a transaction when `run_in_transaction = true`.

Closes #4855

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Diesel-Migrations now contains a migration source that easily allows you to register Rust based migrations
 * Diesel-Migrations now contains a migration source that allows you to combine migrations from several different sources
 * Added `SqliteConnection::with_raw_connection` to provide safe, callback-based access to the raw `*mut sqlite3` handle for advanced SQLite C APIs (session extension, hooks, etc.)
+* Added documentation for migration transaction behaviour at the crate root
 
 ### Changed
 


ayarotsky:
@weiznich Let me know if you think it'd be valuable to mention this behavior on https://diesel.rs/guides/getting-started.

PR #5015, opened 2026-04-02
Fix non-deterministic test failures on PostgreSQL

### LINKED ISSUE #2949: order_by_column test is not deterministic
I'm in my own branch (`only`) which builds from b2c58897c3.
I have ran `cargo test --features postgres` many times. Only once, I got one failed test (that my code could not possibly affect I think?):

```
failures:

---- order::order_by_column stdout ----
thread 'order::order_by_column' panicked at 'assertion failed: `(left == right)`
  left: `[User { id: 275, name: "Aaron", hair_color: None }, User { id: 275, name: "Jim", hair_color: None }, User { id: 273, name: "Sean", hair_color: None }, User { id: 274, name: "Tess", hair_color: None }]`,
 right: `[User { id: 55, name: "Aaron", hair_color: None }, User { id: 275, name: "Jim", hair_color: None }, User { id: 273, name: "Sean", hair_color: None }, User { id: 274, name: "Tess", hair_color: None }]`', diesel_tests/tests/order.rs:40:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: integration_tests::order::order_by_column
             at ./tests/order.rs:40:5
   3: integration_tests::order::order_by_column::{{closure}}
             at ./tests/order.rs:5:1
   4: core::ops::function::FnOnce::call_once
             at /home/erlend/.rustup/toolchains/1.51.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    order::order_by_column

test result: FAILED. 363 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 41.36s
```

In any case, this test passed right after it failed without any changes to the code.

ayarotsky:
PostgreSQL does not guarantee row order without an explicit ORDER BY clause — "the actual order will depend on the scan and join plan types and the order on disk, but it must not be relied on" [1].

Tests that loaded rows unordered and then indexed into the result (data[0], data[1], ...) were silently assuming insertion order, which caused intermittent failures when PostgreSQL returned rows in a different order.

Add ORDER BY id to the initial loads in order_by_descending_column, except, filter_by_like, and filter_by_ilike to guarantee stable ordering.

[1] https://www.postgresql.org/docs/current/queries-order.html

Closes #2949

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ### Fixed
 
 * Raise a compile-time error when mixing aggregate and non-aggregate expressions in an `ORDER BY` clause without a `GROUP BY` clause
+* Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
 
 ### Changed
 
--- a/diesel_tests/tests/combination.rs
+++ b/diesel_tests/tests/combination.rs
@@ -102,7 +102,7 @@ fn except() {
         NewUser::new("Jim", None),
     ];
     insert_into(users).values(&data).execute(conn).unwrap();
-    let data = users.load::<User>(conn).unwrap();
+    let data = users.order(id).load::<User>(conn).unwrap();
     let sean = &data[0];
     let tess = &data[1];
     let _jim = &data[2];
--- a/diesel_tests/tests/copy.rs
+++ b/diesel_tests/tests/copy.rs
@@ -183,6 +183,7 @@ fn copy_from_from_insertable_struct() {
     assert_eq!(user_count, 2);
     let users = users::table
         .select((users::name, users::hair_color))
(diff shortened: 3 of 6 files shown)


weiznich (approved):
Thanks for working on this, this is highly appreciated as that makes our CI more robust

PR #5021, opened 2026-04-06
Add doc example for #[diesel(table_name)] in Insertable derive

gr1zer1:
Adds a test example for `#[diesel(table_name)]` attribute in `#[derive(Insertable)]` 
as requested in #4840

### DIFF
--- a/diesel_derives/build.rs
+++ b/diesel_derives/build.rs
@@ -231,7 +231,13 @@ fn main() {
         ),
         (
             "insertable",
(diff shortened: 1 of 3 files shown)


weiznich (approved):
Thanks for working on this. This change looks good, I just adjusted the table name to use something that doesn't match the default generated name to make it a bit clearer what changes in between using and not using this attribute.

gr1zer1:
Thanks

PR #5029, opened 2026-04-10
Add SQLite json_extract and jsonb_extract functions

KonstantinosVasilopoulos:
Declaration of the `json_extract` and `jsonb_extract` sqlite functions. The following new functions were created for the sqlite crate.

The decision to create one function per return type was taken after [discussing this in the Diesel Github forum](https://github.com/diesel-rs/diesel/discussions/5025).

- json_extract_integer
- json_extract_json
- json_extract_string
- jsonb_extract_integer
- jsonb_extract_json
- jsonb_extract_string
 
Partially completes issue #4366


diesel/src/sqlite/expression/functions.rs
```diff
@@ -333,6 +333,442 @@ extern "SQL" {
...
+    ///
+    /// Returns `NULL` if the path does not exist in the JSON document, or if the value at that
+    /// path is a JSON `null`.
+    ///
+    /// To extract other types, use:
+    /// - [`json_extract_integer`](json_extract_integer_1()) for integer values
+    /// - [`json_extract_double`](json_extract_double_1()) for floating-point values
+    /// - [`json_extract_json`](json_extract_json_1()) for JSON objects or arrays
+    ///
+    /// This function requires at least SQLite 3.38 or newer.
```
ayarotsky: Please double-check this. If I'm not mistaken `json_extract` has been available since SQLite 3.9.0

diesel/src/sqlite/expression/functions.rs
```diff
@@ -333,6 +333,442 @@ extern "SQL" {
...
+    ///     .get_result::<Option<Value>>(connection)?;
+    /// assert_eq!(Some(json!({"f": 7})), result);
+    ///
+    /// let json = json!({"a": 2, "c": [4, 5, {"f": 7}]});
+    /// let result = diesel::select(json_extract_json_1::<Json, _, _>(json, "$.x"))
+    ///     .get_result::<Option<Value>>(connection)?;
+    /// assert_eq!(None, result);
+    ///
+    /// let json = json!({"a": 2, "c": [4, 5], "f": 7});
+    /// let result = diesel::select(json_extract_json_2::<Json, _, _, _>(json, "$.c", "$.a"))
```
ayarotsky: The `_2` variants of the scalar-return functions generate SQL like `json_extract(json, p1, p2)`. SQLite's `json_extract` only returns a native `scalar (INTEGER, TEXT, REAL)` when given exactly one path. With two or more paths, it always returns a JSON array, regardless of the value types.

So `json_extract_integer_2` would produce this SQL:
```sql
SELECT json_extract('{"a":1,"b":2}', '$.a', '$.b')
-- returns: '[1,2]'   ← a JSON text string, not an integer
```

Diesel will then try to deserialize the string `'[1,2]'` as `Option<i32>`, which will fail at runtime.

ayarotsky (commented):
1. Please add a CHANGELOG entry.
2. Please add integration tests not only covering happy path, but also cases like type mismatch, null json, null path, etc.

diesel/src/sqlite/expression/functions.rs
KonstantinosVasilopoulos: You are right. Just confirmed that `json_extract` and `jsonb_extract` were both added in [SQLite 3.9.0](https://sqlite.org/releaselog/3_9_0.html) under the the json1 extension module.

diesel/src/sqlite/expression/functions.rs
KonstantinosVasilopoulos: I will resolve this by making json_extract_string, json_extract_integer and json_extract_double non-variadic. Only json_extract_json will remain variadic. Same for jsonb_extract variants. Please advise if you believe that another approach would be better.

KonstantinosVasilopoulos:
@ayarotsky I have resolved the issue you highlighted. Please review and let me know if everything is okay now.

weiznich (approved):
Looks good now. Thanks for your contribution

PR #5032, opened 2026-04-15
Add regression test for renamed primary key column

### LINKED ISSUE #436: Renaming the primary key of a table blows everything up
A project with the following migration would fail to compile when using `infer_schema!`

``` sql
CREATE TABLE users (non_standard_pk SERIAL PRIMARY KEY);
ALTER TABLE users RENAME COLUMN non_standard_pk TO id;
```

While the column has been renamed, the `pg_attribute` entry that the index is pointing to continues to have the old name. This causes our generated code to explode when we do `type PrimaryKey = users::non_standard_pk`. I haven't looked too deeply into this. My guess is that there is some indicator on the `pg_attribute` entry which will indicate that it's no longer being directly used, and the new entry is somehow referencing the old one.

ayarotsky:
The bug was fixed in 49337801c (Feb 2017) when schema inference was rewritten to use `information_schema.key_column_usage` instead of `pg_attribute`. The old `pg_attribute`-based implementation retained stale entries after a column rename, causing `infer_schema!` to emit the old column name.

Closes #436.

### DIFF
--- a/diesel_cli/src/infer_schema_internals/information_schema.rs
+++ b/diesel_cli/src/infer_schema_internals/information_schema.rs
@@ -417,4 +417,32 @@ mod tests {
             get_primary_keys(&mut connection, &table_2).unwrap()
         );
     }
+
+    /// Regression test for https://github.com/diesel-rs/diesel/issues/436
+    ///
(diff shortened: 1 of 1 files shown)


weiznich (approved):
Thanks ❤️

PR #5034, opened 2026-04-16
Fix reject aggregate select with non aggregate order by

### LINKED ISSUE #3815: Runtime Error for count() with order_by clause combined
### Versions

- **Rust:**: 1.72
- **Diesel:**:  2.x.x
- **Database:**: Postgres
- **Operating System**: Windows/Linux/Mac

### Feature Flags

- **diesel:**: "chrono", "r2d2"

## Problem Description
I am creating a hateoas API. A hateoas API contains besides the data also some metadata containing the amount of elements that are contained with the query (different to the current elements). 
To not duplicate code I just clone the normal query and add a .count() to it. 
Unfortunately it seems like an order by and a count operator are incompatible and generate incompatible SQL.

I try to create an API for the Etherpad website that allows different sorting by downloads/name etc. 

The total count query should output the total number of items in this query so that I can lazy load other elements if needed

thread 'actix-rt|system:0|arbiter:0' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(Unknown, "column \"datas.downloads\" must appear in the GROUP BY clause or be used in an aggregate function")', src\entities\plugin.rs:162:88
stack backtrace:

No. I had a look at the SQL statements and it is definetely faulty.

### Steps to reproduce

1. `git clone https://github.com/SamTV12345/etherpad-plugin-data-extractor.git`
2. git checkout bug/failing-query
3. curl localhost:8001/api/plugins
4. See the error message

If you remove every order_by clause for the count query it works flawlessly. I am unfortunately unsure why this happens. Group by SQL statements shouldn't have anything to do with order by clauses

## Checklist

- [ x] This issue can be reproduced without requiring a third party crate

ayarotsky:
Reject aggregate SELECT combined with non-aggregate ORDER BY at compile time

Calling `.count()` or `.select(aggregate)` on a query that already has a non-aggregate `.order_by()` clause previously compiled but generated invalid SQL, rejected by PostgreSQL at runtime with:
```
column "x" must appear in the GROUP BY clause or be used in an aggregate function
```

Fix this by implementing `ValidGrouping` for `NoOrderClause` and `OrderClause<Expr>`, then adding the same MixedAggregates` cross-check to `SelectDsl` that already exists in `OrderDsl`. The reverse order  (`.select(aggregate).order_by(non_aggregate)`) was already a compile error; this closes the symmetric gap.

Closes #3815.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ### Fixed
 
 * Raise a compile-time error when mixing aggregate and non-aggregate expressions in an `ORDER BY` clause without a `GROUP BY` clause
+* Calling `.count()` or `.select(aggregate_expr)` on a query that already has a non-aggregate `.order_by()` clause now raises a compile-time error instead of generating invalid SQL that would be rejected by the database at runtime (fixes [#3815](https://github.com/diesel-rs/diesel/issues/3815))
 * Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
 
 ### Changed
--- a/diesel/src/query_builder/order_clause.rs
+++ b/diesel/src/query_builder/order_clause.rs
@@ -1,5 +1,7 @@
 use alloc::boxed::Box;
 
+use crate::expression::{ValidGrouping, is_aggregate};
+
 simple_clause!(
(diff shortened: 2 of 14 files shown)


weiznich (approved):
Looks good, thanks for taking care of this. 

I will address the CI failure later today, that likely needs #5035 to be merged and this PR to be rebased on top of it.

PR #5040, opened 2026-04-20
Send bpchar OID for CHAR(N) bind parameters on PostgreSQL

### LINKED ISSUE #1912: Queries on indexed char(n) column in PostgreSQL use sequential scan
### Versions

- **Rust:** rustc 1.30.1 (1433507eb 2018-11-07)
- **Diesel:** 1.3.3
- **Database:** postgresql-9.6 **9.6.10-0+deb9u1** OR postgresl-11 **11.1-1.pgdg90+1** (similar results)
- **Operating System** Debian 9.5

### Feature Flags

- **diesel:** postgres

## Problem Description

My table has a primary key of type `char(11)`.  When querying this table, Diesel sends over a parameterized query and a value of type `text` instead of `bpchar`.  This forces a slow sequential scan instead of doing an index-only scan.  Running a similar query from psql is fast.

Find an indexed `char(n)` quickly

results in tens of milliseconds

results in 1+ second

No

### Steps to reproduce

I think all you need to observe this is an indexed `char(n)` column, a `.find(` or `.filter(id.eq(` in Diesel, and `auto_explain` on the server side to confirm sequential scans.  But I have a complete test case set up to demonstrate this:

For the sake of completeness, I did this on a new Debian 9.5:

```
apt-get update
apt-get install --no-install-recommends libpq-dev curl git gcc libc-dev
adduser user
su user
curl https://sh.rustup.rs -sSf | sh
cargo install diesel_cli --no-default-features --features postgres
```

then

```
su postgres
createuser -S -D -R -P diesel_bpchar
createdb -O diesel_bpchar diesel_bpchar
```

```
git clone https://github.com/ludios/diesel_bpchar
cd diesel_bpchar
cargo build

echo "DATABASE_URL=postgres://diesel_bpchar:PASS@localhost:5432/diesel_bpchar" > .env
diesel setup
diesel migration run

for i in {00000000001..0005000000}; do echo $i >> /tmp/ids; done
chmod a+r /tmp/ids
psql postgres://diesel_bpchar:PASS@localhost:5432/diesel_bpchar -c "COPY videos FROM STDIN" < /tmp/ids
```

What should be a quick index lookup instead takes 1 second:

```
# time ./target/debug/find_video 00005000000
Found 1 videos
./target/debug/find_video 00005000000  0.01s user 0.00s system 1% cpu 0.998 total
```

With these postgresql server settings, you can see what is happening on the server side:

```
shared_preload_libraries = 'auto_explain'       # (change requires restart)

auto_explain.log_min_duration = 0
auto_explain.log_analyze = true
auto_explain.log_verbose = true

log_statement = 'all'           # none, ddl, mod, all
```

(note: `auto_explain` is in the `postgresql-contrib` package on Debian.)

When Diesel sends over its parameterized query, PostgreSQL logs a sequential scan:

```
2018-11-10 17:35:00.414 UTC [29344] diesel_bpchar@diesel_bpchar LOG:  execute __diesel_stmt_0: SELECT "videos"."id" FROM "videos" WHERE "videos"."id" = $1
2018-11-10 17:35:00.414 UTC [29344] diesel_bpchar@diesel_bpchar DETAIL:  parameters: $1 = '00005000000'
2018-11-10 17:35:01.390 UTC [29344] diesel_bpchar@diesel_bpchar LOG:  duration: 976.167 ms  plan:
    Query Text: SELECT "videos"."id" FROM "videos" WHERE "videos"."id" = $1
    Seq Scan on public.videos  (cost=0.00..102028.00 rows=25000 width=12) (actual time=976.154..976.154 rows=1 loops=1)
      Output: id
      Filter: ((videos.id)::text = '00005000000'::text)
      Rows Removed by Filter: 4999999
```

Run a similar query from `psql` and PostgreSQL logs an `Index Only Scan`:

```
# time psql postgres://diesel_bpchar:PASS@localhost:5432/diesel_bpchar -c "SELECT videos.id FROM videos WHERE videos.id = '00005000000';"
     id
-------------
 00005000000
(1 row)

psql postgres://diesel_bpchar:PASS@localhost:5432/diesel_bpchar   0.03s user 0.01s system 91% cpu 0.041 total
```

```
2018-11-10 17:39:55.150 UTC [31473] diesel_bpchar@diesel_bpchar LOG:  statement: SELECT videos.id FROM videos WHERE videos.id = '00005000000';
2018-11-10 17:39:55.150 UTC [31473] diesel_bpchar@diesel_bpchar LOG:  duration: 0.033 ms  plan:
        Query Text: SELECT videos.id FROM videos WHERE videos.id = '00005000000';
        Index Only Scan using videos_pkey on public.videos  (cost=0.43..8.45 rows=1 width=12) (actual time=0.028..0.028 rows=1 loops=1)
          Output: id
          Index Cond: (videos.id = '00005000000'::bpchar)
          Heap Fetches: 1
```

## Checklist

ayarotsky:
`Bpchar` was a hidden type alias for `Varchar`. When Diesel sent a bind parameter for a `CHAR(N)` / `bpchar` column, it used OID 25 (text), causing PostgreSQL to cast the column to text for comparison. This prevented index scans, turning indexed lookups into sequential scans.

Fix by promoting `Bpchar` to a distinct SQL type with OID 1042 (the PostgreSQL bpchar OID). Bind parameters for `CHAR(N)` columns are now annotated with the correct OID, allowing PostgreSQL to compare them directly against the column without a cast and enabling index scans.

Closes #1912

### DIFF
--- a/diesel/src/pg/types/mod.rs
+++ b/diesel/src/pg/types/mod.rs
@@ -337,7 +337,10 @@ pub mod sql_types {
     pub type Bytea = crate::sql_types::Binary;
 
     #[doc(hidden)]
-    pub type Bpchar = crate::sql_types::VarChar;
+    #[cfg(feature = "postgres_backend")]
+    #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
+    #[diesel(postgres_type(oid = 1042, array_oid = 1014))]
+    pub struct Bpchar;
 
     /// The PostgreSQL [Money](https://www.postgresql.org/docs/current/static/datatype-money.html) type.
     ///
--- a/diesel/src/pg/types/primitives.rs
+++ b/diesel/src/pg/types/primitives.rs
@@ -101,6 +101,27 @@ impl FromSql<sql_types::Citext, Pg> for String {
     }
 }
 
+#[cfg(feature = "postgres_backend")]
+impl ToSql<crate::pg::sql_types::Bpchar, Pg> for str {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        <str as ToSql<sql_types::Text, Pg>>::to_sql(self, out)
+    }
+}
+
+#[cfg(feature = "postgres_backend")]
+impl ToSql<crate::pg::sql_types::Bpchar, Pg> for String {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        <String as ToSql<sql_types::Text, Pg>>::to_sql(self, out)
+    }
+}
+
+#[cfg(feature = "postgres_backend")]
+impl<'a> FromSqlRef<'a, crate::pg::sql_types::Bpchar, Pg> for &'a str {
+    fn from_sql(value: &'a mut PgValue<'_>) -> deserialize::Result<Self> {
+        Ok(core::str::from_utf8(value.as_bytes())?)
+    }
+}
+
 /// The returned pointer is *only* valid for the lifetime to the argument of
 /// `from_sql`. This impl is intended for uses where you want to write a new
 /// impl in terms of `Vec<u8>`, but don't want to allocate. We have to return a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 ### Fixed
 
+* `Bpchar` is now a distinct PostgreSQL SQL type (previously a hidden alias for `Varchar`). Binds on `CHAR(N)` / `BPCHAR` columns are now sent with OID 1042, allowing PostgreSQL to use the column's index instead of casting it to text.
 * Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
 
 ### Changed
--- a/diesel/src/type_impls/primitives.rs
+++ b/diesel/src/type_impls/primitives.rs
@@ -95,6 +95,7 @@ mod foreign_impls {
     #[cfg_attr(feature = "__sqlite-shared", diesel(sql_type = crate::sql_types::Time))]
     #[cfg_attr(feature = "__sqlite-shared", diesel(sql_type = crate::sql_types::Timestamp))]
     #[cfg_attr(feature = "postgres_backend", diesel(sql_type = crate::sql_types::Citext))]
+    #[cfg_attr(feature = "postgres_backend", diesel(sql_type = crate::pg::sql_types::Bpchar))]
     struct StringProxy(String);
 
     #[derive(AsExpression)]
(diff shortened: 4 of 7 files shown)


diesel/src/pg/types/record.rs
```diff
@@ -133,6 +134,14 @@ macro_rules! tuple_impls {
 
 diesel_derives::__diesel_for_each_tuple!(tuple_impls);
 
+thread_local! {
```
weiznich: I would rather not carry state via a side channel like that. This looks like it will cause issues down the road.

Maybe we can pass in that information via the `MetadataLookup` struct instead?

Maybe there is another solution to use the correct oid at calling side?

weiznich (changes requested):
Thanks for working on this. The overall approach of passing a zero oid to the server and let the server infer the correct type seems reasonable, although I'm still not 100% sure if there might be cases where this breaks (e.g. if the server infers that the bind must be an `INT` instead for whatever reason).

That written: The code around handling the `ROW` case seems rather brittle, I would like to avoid such side-channel handling if somewhat possible. Also the CI seems to indicate that there are more cases where this kind of workaround might not work?

diesel/src/pg/types/record.rs
ayarotsky: Thank you @weiznich. I'm on the fence with this fix myself. Please allow me to try a few more ideas before I give up.

diesel/src/pg/types/mod.rs
```diff
@@ -337,7 +337,10 @@ pub mod sql_types {
     pub type Bytea = crate::sql_types::Binary;
 
     #[doc(hidden)]
-    pub type Bpchar = crate::sql_types::VarChar;
+    #[cfg(feature = "postgres_backend")]
+    #[derive(Debug, Clone, Copy, Default, QueryId, SqlType)]
+    #[diesel(postgres_type(oid = 1042, array_oid = 1014))]
+    pub struct Bpchar;
```
weiznich, line 343: Technically that's a breaking change as it breaks expressions like

```
some_table_bpchar_col.eq(some_table::text_col)
```

That written: I think we might get away with changing that given that is seems unlikely that users write such code.

diesel/src/pg/types/primitives.rs
```diff
@@ -101,6 +101,28 @@ impl FromSql<sql_types::Citext, Pg> for String {
...
+
+#[cfg(feature = "postgres_backend")]
+impl ToSql<crate::pg::sql_types::Bpchar, Pg> for String {
+    fn to_sql<'b>(&'b self, out: &mut Output<'b, '_, Pg>) -> serialize::Result {
+        <String as ToSql<sql_types::Text, Pg>>::to_sql(self, out)
+    }
+}
+
+#[cfg(feature = "postgres_backend")]
+impl FromSql<crate::pg::sql_types::Bpchar, Pg> for String {
```
weiznich: Should we have a `FromSqlRef` impl for `&str` as well?

What's about `ToSql` and `FromSql` for `Cow<str>`?

weiznich (commented):
Overall that seems a good solution, although I have some remarks about additional `ToSql`/`FromSql` impls.

Thanks for pushing this.

By chance: Are you at RustWeek in Utrecht? If that's the case it might be nice to meet there.

ayarotsky:
> By chance: Are you at RustWeek in Utrecht? If that's the case it might be nice to meet there.

@weiznich Unfortunately, no. But I hope we can meet at one of the events next year.

diesel/src/pg/types/primitives.rs
ayarotsky: > Should we have a FromSqlRef impl for &str as well?

Done.

> What's about ToSql and FromSql for Cow<str>?

I think `type_impls/primitives.rs:212` already covers it.

weiznich (approved):

PR #5042, opened 2026-04-24
Fix several UB instances and other probematic behaviour

weiznich:
This commit fixes a larger number of undefined behaviour usage, API misuses and otherwise security related bugs. Most of these issues where detected using LLM's (mostly Github copilot, various models), all of them were verified and fixed manually. The following list describes each issue and the relevant fix in detail:

# Unsound construction of a rust String with non-utf8 bytes in `SqliteValue::read_text`

The `SqliteValue::read_text` function (and any way to deserialize a `String` value from `Sqlite` constructed strings via the unsafe `str::from_utf8_unchecked` function, relying on the sqlite documentation stating that `sqlite3_value_text` returns a UTF-8 string. It turns out that this is not guranteed if you call this function on a byte column, which in turn can result in the construction of a rust String/str containing non-utf-8 bytes, which in turn is violating the safety contract of these types. This is fixed in two ways: Whereever possible we now return UTF-8 error if we encounter such a case. For `SqliteValue::read_text` that wouldn't be possible without an API breaking change, so I opted to use `String::from_ut8_lossy` there combined with caching the allocated result inside of `SqliteValue`.
Any user that might load values from a field with blob storage as `String` via diesel can be affected from this if the data contained non-utf8 data.

# Invalid assumption about aggregator alignment in the implementation of custom SQLite aggregate functions

We relied on the `sqlite3_aggregate_context` function to allocate an instance of the user provided aggregator type for the implementation of custom SQLite aggregate functions. While the code already noted correctly that the `sqlite3_aggregate_context` function doesn't give any gurantee about alignment it did fail to understand that already creating a reference to unaligned data is undefined behaviour in Rust. This is now corrected by not relying on `sqlite3_aggregate_context`, but rather use the user data pointer instead and allocating the corresponding value in Rust with the correct alignment. 
Any user implementing `SqliteAggregateContext` for a type with a custom aligment would have been affected by this.

# Potential memory leaks while registering custom SQLite functions and collations

We discovered that while registering custom SQLite functions and collations we first create a raw pointer for the callback by leaking the allocation, then convert the function name and arguments to the corresponding representation by using failable methods and finally passing everything to the relevant SQLite API function. This can result in a memory leak as the allocation is not freed if any of the conversion steps fails. The fix just reordered the operations to perform the failable steps first and creating the corresponding raw pointer only as final step.
Any user that tries to registered a SQLite function containing a null byte in the name or with more than `i32::MAX` arguments would be affected by this bug.

# Reading of padding while serializing date/time related types in the Mysql backend

The Mysql backend provides a `MysqlTime` type to describe the binary format of the underlying byte buffer. This type is marked as `#[repr(C)]`, but contains padding bytes. The existing serialization code just casted a given struct instance to an slice of bytes via a pointer cast. This touches the padding bytes, which is undefined behaviour in rust. The bytes were then passed directly into libmysqlclient, which in turn reconstructed the `MysqlTime` struct, containing the same padding bytes. To fix this we wrote a manual serialization routine that copies the data values manually to the byte buffer without touching the byte buffers.
Any user serializing date/time values using the Mysql backend is affected by this, although this doesn't have any practical impact yet.

# Debug/Display implementations of batch inserts for the SQLite backend relied on unstable type layout

The Display/Debug implementation for batch insert queries for the SQLite backend performed a pointer cast between `#[repr(Rust)]` types relying on unspecified type layout there. This is corrected in favour of a pointer cast between `#[repr(transparent)]` types with a specified semantic.
This affects any user relying on the Display/Debug implementation for this kind of query, although in practice the layout of the involved types was the same with the currently supported rust compiler versions.

# Invalid call order of `sqlite3_value_blob`/`sqlite3_value_text` and `sqlite3_value_bytes`

The deserialization layer in the SQLite backend relyed on an invalid calling order of these function, while the SQLite documentation explicitly documented a to call these functions the other way around to prevent internally changing the result. This is fixed by changing the order as suggested by the documentation.
This potentially affects any user that deserializes `Vec<u8>` and `String` values using the SQLite backend, although you also would need to trigger an internal data conversion in SQLite to be affected by this bug. At the time of writing we are not aware of any practical way to cause a corruption there.

# Unsound pointer usage in `FromSql` implementations for `String`/`Vec<u8>`
for third party backends providing a `FromSql<Text/Binary, > for *const str/u8` implementation

A `FromSql` implementation for `String` and `Vec<u8>` provided by diesel relies on a raw pointer provided by another `FromSql` implementation. As the relevant functions are safe functions and a custom implementation might return an invalid pointer this could lead to data corruption or other crashes in these impls. Any in tree backend provided a correct pointer there, but a third party backend is not required to do that.
As fixs this mechanism is replaced by a custom `FromSqlRef` trait that allows to operate on references instead. Custom backends relying on the existance of these `FromSql` impls for pointers must be updated to use the new trait instead.
This affects any user of a third party backend that might return an invalid pointer.

# Option injection in the PostgreSQL `COPY FROM/TO` interface

The provided interface for `COPY FROM/TO` in the PostgreSQL backend allows to configure various options as String and char values. Due to how this is implementid in the PostgreSQL wire protocol it is not possible to send these options as bind parameters, but they need to be directly embedded in the query string. These functios allowed a malicous  user to include quote characters `'`, which in turn would have allowed to  modify the query construction. Fortunally the scope of this injection is restricted to the very same `COPY FROM/TO` statement, so the worst impact such an injection can have is changing options for the same statement, which would result in the acceptance/return of an unexpected data format. As a fix we now correctly escape any included quote character.
Any user using the `COPY FROM/TO` interface in combination with user controlled options for any configuration method accepting a `String` or `char` is affected by this issue. 

Beside these specific issues this commit also contains various smaller  improvements to make several unsafe code instances more robust:

* We improved some pointer handling in the mysql bind code
* For the sqlite function argument deserialization we removed the usage of `ManuallyDrop` in favour of a simpler solution based on slices
* We added/improved various SAFETY comments to describe the relevant invariants in more details
* For the PostgreSQL backend we refactored the `RawResult` related  pointer handling to not expose the pointer outside of the raw module anymore. This wraps all necessary unsafe code to interact with this type in one place, which makes it easier to verify and uphold any necessary invariants
* We replaced several other instances of `str::from_utf8_unchecked` with their checked equivalent to make sure not to rely again on invalid assumptions.

### DIFF
--- a/diesel/src/deserialize.rs
+++ b/diesel/src/deserialize.rs
@@ -474,6 +474,16 @@ pub trait FromSql<A, DB: Backend>: Sized {
     }
 }
 
+/// A helper trait for deserializing data in a borrowed variant
+///
+/// This is mostly useful for getting a `&str` or `&[u8]` in your code
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
+pub(crate) trait FromSqlRef<'a, A, DB: Backend>: Sized {
+    fn from_sql(bytes: DB::RawValue<'a>) -> Result<Self>;
+}
+
 /// Deserialize a database row into a rust data structure
 ///
 /// Diesel provides wild card implementations of this trait for all types
--- a/diesel/src/mysql/connection/bind.rs
+++ b/diesel/src/mysql/connection/bind.rs
@@ -1,4 +1,5 @@
 #![allow(unsafe_code)] // module uses ffi
+use core::cmp;
 use core::ffi as libc;
 use core::mem::MaybeUninit;
 use core::mem::{self, ManuallyDrop};
@@ -178,6 +179,13 @@ pub(super) struct BindData {
 // instead of just copying the pointer
 impl Clone for BindData {
     fn clone(&self) -> Self {
+        // we just make sure here that we never create a
+        // slice larger than the allocation
+        let length = cmp::min(
+            self.capacity,
+            self.length.try_into().expect("usize fits the length"),
+        );
+
         let (ptr, len, capacity) = if let Some(ptr) = self.bytes {
             let slice = unsafe {
                 // We know that this points to a slice and the pointer is not null at this
@@ -187,10 +195,7 @@ impl Clone for BindData {
                 // written. At the time of writing this comment, the `BindData::bind_for_truncated_data`
                 // function is only called by `Binds::populate_dynamic_buffers` which ensures the corresponding
                 // invariant.
-                core::slice::from_raw_parts(
-                    ptr.as_ptr(),
-                    self.length.try_into().expect("usize is at least 32bit"),
-                )
+                core::slice::from_raw_parts(ptr.as_ptr(), length)
             };
             bind_buffer(slice.to_owned())
         } else {
@@ -521,10 +526,9 @@ impl BindData {
     /// this function is unsafe unless the binds are immediately rebound.
     unsafe fn bind_for_truncated_data(&mut self) -> Option<(ffi::MYSQL_BIND, usize)> {
         if self.is_truncated() {
-            if let Some(bytes) = self.bytes {
+            if let Some(bytes) = self.bytes.take() {
                 let mut bytes =
                     unsafe { Vec::from_raw_parts(bytes.as_ptr(), self.capacity, self.capacity) };
-                self.bytes = None;
 
                 let offset = self.capacity;
                 let length = usize::try_from(self.length).expect("Usize is at least 32 bit");
--- a/diesel/src/mysql/types/date_and_time/mod.rs
+++ b/diesel/src/mysql/types/date_and_time/mod.rs
@@ -1,5 +1,6 @@
 use core::ffi as libc;
-use core::{mem, slice};
+use core::mem;
+use core::ptr;
 use std::io::Write;
 
 use crate::deserialize::{self, FromSql, FromSqlRow};
@@ -53,7 +54,7 @@ pub struct MysqlTime {
 impl MysqlTime {
     /// Construct a new instance of [MysqlTime]
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub const fn new(
         year: libc::c_uint,
         month: libc::c_uint,
         day: libc::c_uint,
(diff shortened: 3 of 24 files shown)


Mingun:
I suggest to not put so big refactoring in one commit next time (especially just before the release) and give some time to reviewers to review your changes (especially when you requested such review). In fact, your behavior in this PR seems suspicious to me, since no such actions have been noticed before. Has the your account been compromised?

weiznich:
@Mingun 
I see you point regarding the large commit, it's unfortunately an artifact of how this fix was created. I got aware of all the problems at the same time and worked on the fixes in parallel. Several fixes also depend on each other, which made it harder to split it up afterwards.
As for the review: Again I can see your point but I rather wanted to get out the fix for these issues as soon as possible rather than waiting days/weeks for a review. That was especially important for me given that I already know that I have not much time to do any Diesel stuff for the next 2 weeks. So the decision there was either fix the issues as fast as possible and skip the review (and also making the commits as nice as possible) or let the issues sit for at least 14 more days. For me it was more important to fix the issues, as you and others can always put reviews on such changes afterwards. If you feel that a specific change is unwanted I'm happy to provide more details on the reasoning there.

Finally in terms of review: It's not that none looked at the changes, I coordinated this with some colleagues from my day job that are not part of the any diesel team. There is unfortunately no public track record of this as this coordination happened in private channels.

It's hard to provide a proof that the account has not been compromised. I can only write that Yes in fact that PR has been created by me and not by someone else. In fact it's also signed by my GPG key (you can verify it's the same key I always used to sign commits), so it any attacker would have need to not only gain access to my github account, but also to my GPG key and to my crates.io token (for publishing the release).

EDIT: I also would highlight that it's perfectly fine to be suspicious here. Its good to see that people actually look at what I'm doing as it often feels like there is noone double checking my work. I'm more than aware that I make mistakes and shouldn't trust my first pass solution anywhere.

Ten0:
The process here makes sense to me: leaving unadressed ~vulnerabilities public for a long time seems dangerous.
That seems sane to post-review though.

weiznich:
For reference: I also requested a rustsec advisory for the first issue in the PR description as that one looks like it could cause issues in "common" cases: https://github.com/rustsec/advisory-db/pull/2821

@diesel-rs/contributors Please let me know if you feel that we should request an advisory for any of the other points as well.

PR #5044, opened 2026-04-26
Remove MySQL time serializer debug output

cookesan:
## Summary
- remove an accidental `dbg!` wrapper from the MySQL `MysqlTime` serializer

## Details
The serializer only needs the pointer returned by `out_ptr.offset(offset)`. The debug wrapper prints that pointer to stderr whenever this path is used.

This does not change which pointer is passed to `ptr::copy`; it removes only the stderr side effect.

Follow-up to https://github.com/diesel-rs/diesel/discussions/5043.

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo check -p diesel --no-default-features --features mysql_backend`
- `cargo check -p diesel --no-default-features --features mysql`
- `cargo check -p diesel --no-default-features --features 'mysql chrono time'`
- `cargo test -p diesel --no-default-features --features mysql --lib`
- `cargo test -p diesel --no-default-features --features 'mysql chrono time' --lib`

The MySQL library test commands were run against a local MySQL 8.4 database.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 ## Unreleased
 
+## [2.3.9] 2026-04-30
+
+* Removed a `dbg!` statement from the Mysql backend that caused unwanted output
+
 ## [2.3.8] 2026-04-24
 
 * Added support for libsqlite3-sys 0.37.0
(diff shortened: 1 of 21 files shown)


weiznich:
I added some commits on top of this to prepare an release and also to make sure that this won't happen again.

cc @diesel-rs/core, as I would like to cut this release later this week

weiznich (approved):

PR #5046, opened 2026-04-28
Fix `#[derive(AsChangeset)]` regression for structs with type or const generics

apastrana6:
Fixes a regression introduced in 2.3.8 by https://github.com/diesel-rs/diesel/pull/5004. The new `_check_owned` / `_check_borrowed` diagnostic helpers forward only lifetime params, so any `AsChangeset` struct with a type or const generic referenced in a field type fails to compile with `error[E0425]: cannot find type 'T' in this scope`.
  

Fix: forward all generic params, not just lifetimes.

  ## Reproduction

  Minimal repro on 2.3.8:

  ```rust
  use diesel::prelude::*;

  diesel::table! {
      users { id -> Integer, name -> Text }
  }

  #[derive(diesel::expression::AsExpression, diesel::deserialize::FromSqlRow)]
  #[diesel(sql_type = diesel::sql_types::Text)]
  struct Wrapper<T: 'static>(String, std::marker::PhantomData<T>);

  #[derive(AsChangeset)]
  #[diesel(table_name = users)]
  struct UserForm<T: 'static> {
      name: Wrapper<T>,
  }
  ```

  Fails with:

  ```
  error[E0425]: cannot find type `T` in this scope
     |
     |     name: Wrapper<T>,
     |                   ^ not found in this scope
  ```

  Compiles with this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ### Fixed
 
 * Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
+* Fix a regression in `#[derive(AsChangeset)]` introduced in 2.3.8 where structs with a type or const generic parameter referenced in a field type failed to compile with `error[E0425]: cannot find type 'T' in this scope`. The diagnostic helper functions added to improve `AsChangeset` error messages now forward all generic parameters of the input struct, not only lifetimes.
 
 ### Changed
 
--- a/diesel_derives/tests/as_changeset.rs
+++ b/diesel_derives/tests/as_changeset.rs
@@ -275,6 +275,104 @@ fn with_lifetime_constraints() {
     assert_eq!(Ok(expected), actual);
 }
 
+// Regression test for the 2.3.8 `_check_owned` / `_check_borrowed` helpers
+// dropping type and const generics. Prior to the fix, the generated
+// diagnostic helper fns forwarded only lifetime params from the struct, so
+// a `where`-clause bound on a field type that mentioned a type param like
+// `T` failed to compile with `error[E0425]: cannot find type 'T' in this
+// scope`. Mirrors the lohi-lib failure shape: a generic wrapper type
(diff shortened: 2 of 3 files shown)


pksunkara (approved):

apastrana6:
@pksunkara thanks for the review! do you know when a new version will be released including this change?

diesel_derives/tests/as_changeset.rs
```diff
@@ -275,6 +275,91 @@ fn with_lifetime_constraints() {
     assert_eq!(Ok(expected), actual);
 }
 
+// Regression test: the diagnostic helper fns must forward type and const
+// generics, not only lifetimes.
```
Mingun, line 279: Why "const generics"? As your test demonstrate, any generics is affected

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ### Fixed
 
 * Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
+* Fix a regression in `#[derive(AsChangeset)]` introduced in 2.3.8 where structs with a type or const generic parameter referenced in a field type failed to compile with `error[E0425]: cannot find type 'T' in this scope`. The diagnostic helper functions added to improve `AsChangeset` error messages now forward all generic parameters of the input struct, not only lifetimes.
```
Mingun, line 25: Same here

diesel_derives/tests/as_changeset.rs
apastrana6: Just meant as opposed to lifetimes, is there some other way you would prefer this to be described?

weiznich: The comment is fine in my opinion. It states "**type** and const generics" and not only const generics are effected. The test is also a reasonable regression test as handling const generics different than normal type generics requires additional effort.

weiznich:
> do you know when a new version will be released including this change?

I backported the change to #5044 to be included in the next patch release. We generally do not give any ETA's for releases, but I'm optimistic to find some time this week to issue that release as well.

PR #5049, opened 2026-05-06
Add support for PostgreSQL's RETURNING old.column

Ten0:
PostgreSQL 18 introduces the `RETURNING old.col_name` syntax to fetch the previous value of a column in `UPDATE` or `INSERT ... ON CONFLICT DO UPDATE` statements.
This is a useful feature to avoid DB round trips on update.
This PR implements this feature via `diesel::pg::returning::old`, similarly to `diesel::upsert::excluded`.

Line diff count may look scary but most of it is really test and snapshot updates.

This is implemented by making `QS` a `ReturningQuerySource<Table>` when considering whether `F: SelectableExpression<QS>` for returning, instead of just checking whether `F: SelectableExpression<Table>`.
(Same for `AppearsInFromClause`.)
This allows leveraging existing generic recursive `SelectableExpression` implementations to keep allowing everything else that was already allowed (functions, tuples...) along with the new syntax.

**Warnings:**
- This implies this minor breaking change:
https://github.com/Ten0/diesel/blob/62fcb8a2f0283dabea9b0ca9b96e4a679dc46537/diesel/src/query_builder/update_statement/mod.rs#L245
(where this used to be only `SelectableExpression<T>`) which may imply breakage only when users were writing very generic code.
In the past we have often considered similar "technically a breaking change" to be very unlikely to cause issues for many people, such that they would be worth it.
- As highlighted by the compile tests, error messages for selecting a column of the wrong table from a returning expression are slightly worsened. I have experimented with adding an extra `ValidInReturningOf` trait to improve error messages [`71d599e`](https://github.com/diesel-rs/diesel/pull/5049/changes/71d599e031ecf39eb5d3388ffbe2beb6a49a39cb) , but couldn't find a way to do it fully generically without running into conflicting impls, so users would have had to implement it manually on custom types, and overall it drastically worsened the complexity of the implementation, so I ended up reverting it.
- I have opted to not squash commits to keep these experimentations for reference, but that probably means that if we merge this we want to "squash-and-merge" it, not just merge it, to avoid bloating the tree with non descriptive commits.

Most of the code is effectively *written* by Opus 4.6/7 but design is mine and I told it what to write (and hopefully I reviewed accurately enough that I didn't miss mistakes 😅).

### DIFF
--- a/diesel/src/pg/mod.rs
+++ b/diesel/src/pg/mod.rs
@@ -12,6 +12,7 @@ pub(crate) mod backend;
 pub(crate) mod connection;
 mod metadata_lookup;
 pub(crate) mod query_builder;
+pub mod returning;
 pub(crate) mod serialize;
 mod transaction;
 mod value;
--- a/diesel/src/pg/returning/mod.rs
+++ b/diesel/src/pg/returning/mod.rs
@@ -1,0 +1,12 @@
+//! PostgreSQL-specific `RETURNING` clause helpers.
+//!
+//! This module exposes [`old()`], the wrapper used to refer to the
+//! pre-modification value of a column in a PostgreSQL `RETURNING` clause —
+//! the `RETURNING old.col` syntax introduced in PostgreSQL 18.
+
+// Naming: we suffix with `_impl` to avoid name conflicts with the `old` re-export below.
+mod old_impl;
+
+pub use self::old_impl::old;
+
+pub(crate) use self::old_impl::return_type_helpers_reexported;
--- a/diesel/src/pg/returning/old_impl.rs
+++ b/diesel/src/pg/returning/old_impl.rs
@@ -1,0 +1,175 @@
+//! `RETURNING old.col` support for PostgreSQL 18 and later.
+
+use crate::backend::{Backend, sql_dialect};
+use crate::expression::{
+    AppearsOnTable, Expression, SelectableExpression, ValidGrouping, is_aggregate,
+};
+use crate::query_builder::returning::{
+    InsertStmtWithOnConflictDoUpdate, ReturningQuerySource, UpdateStmt,
+};
+use crate::query_builder::{AstPass, QueryFragment, QueryId};
+use crate::query_source::Column;
+use crate::result::QueryResult;
+
+/// Wraps a column to refer to its pre-modification value in the `RETURNING`
+/// clause of a PostgreSQL `UPDATE` or `INSERT ... ON CONFLICT ... DO UPDATE`
+/// statement.
+///
+/// This is the type returned by [`old()`](old()).
+#[derive(Debug, Clone, Copy)]
+pub struct Old<C> {
+    _column: C,
+}
+
+impl<C> Old<C> {
+    pub(crate) fn new(c: C) -> Self {
+        Old { _column: c }
+    }
+}
+
+/// Refer to the pre-modification value of `col` in a PostgreSQL `RETURNING`
+/// clause.
+///
+/// This corresponds to the SQL `RETURNING old.col` syntax introduced in
+/// PostgreSQL 18.
+///
+/// # Requires PostgreSQL 18 or newer
+///
+/// Diesel emits `old.col` in the SQL it sends to the database. Earlier
+/// versions of PostgreSQL will reject the query at execution time.
+///
+/// # Statement compatibility
+///
+/// `old(col)` is valid inside the `RETURNING` clause of:
+///
+/// * an `UPDATE` statement, where it has the same Rust SQL type as `col`
+///   (since every returned row necessarily came from a pre-existing row).
+/// * an `INSERT ... ON CONFLICT ... DO UPDATE` statement, **but only when
+///   wrapped in [`.nullable()`]**: rows that were freshly inserted (rather
+///   than updated) have no `old` row, and `old.col` is `NULL` for them, so
+///   for type-safe deserialization you must opt into a nullable Rust SQL
+///   type. Writing `old(col)` directly (without `.nullable()`) in this
+///   context is rejected at compile time.
+///
+/// Use of `old(col)` in plain `INSERT` (without `ON CONFLICT ... DO UPDATE`)
+/// or `DELETE` `RETURNING` is rejected at compile time, because it is not useful
+/// there. (Note that `ON CONFLICT DO NOTHING` never returns untouched rows.)
+///
+/// [`.nullable()`]: crate::NullableExpressionMethods::nullable
+///
+/// # Example
+///
+/// ```rust
+/// # include!("../../doctest_setup.rs");
+/// #
+/// # #[cfg(feature = "postgres")]
+/// # fn main() {
+/// #     use schema::users::dsl::*;
+/// #     use diesel::pg::returning::old;
+/// #     let connection = &mut establish_connection();
+/// #     // `RETURNING old.col` requires PostgreSQL 18+
+/// #     let pg_version: i32 = diesel::dsl::sql::<diesel::sql_types::Integer>(
+/// #         "SELECT current_setting('server_version_num')::int",
+/// #     ).get_result(connection).unwrap();
+/// #     if pg_version < 180000 { return; }
+/// let was_and_now = diesel::update(users.find(1))
+///     .set(name.eq("Updated"))
+///     .returning((old(name), name))
+///     .get_result::<(String, String)>(connection);
+/// assert_eq!(Ok(("Sean".to_string(), "Updated".to_string())), was_and_now);
+/// # }
+/// # #[cfg(not(feature = "postgres"))]
+/// # fn main() {}
+/// ```
+pub fn old<C: Column>(col: C) -> Old<C> {
+    Old::new(col)
+}
+
+impl<C> QueryId for Old<C> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
+
+impl<C> Expression for Old<C>
+where
+    C: Column + Expression,
+{
+    type SqlType = <C as Expression>::SqlType;
+}
+
+impl<C> ValidGrouping<()> for Old<C>
+where
+    C: Column,
+{
+    type IsAggregate = is_aggregate::No;
+}
+
+// `Old<C>` is selectable on a `RETURNING` clause whose statement-kind marker
+// is `UpdateStmt`. It is deliberately *not* selectable on
+// `ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, _>` directly — only
+// `Nullable<Old<C>>` is, via the existing `Nullable` machinery and the
+// `ToInnerJoin` mapping in `returning_query_source` (which makes
+// `InsertStmtWithOnConflictDoUpdate`'s inner-join "fall back" to `UpdateStmt`).
+// That's how we force users to write `old(col).nullable()` in an
+// `INSERT ... ON CONFLICT ... DO UPDATE RETURNING` and reject `old(col)`
+// alone at compile time.
+impl<C> AppearsOnTable<ReturningQuerySource<UpdateStmt, C::Table>> for Old<C>
+where
+    C: Column,
+    Self: Expression,
+{
+}
+
+impl<C> AppearsOnTable<ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, C::Table>> for Old<C>
+where
+    C: Column,
+    Self: Expression,
+{
+}
+
+// We intentionally did not add implementations for use of `old(col)` in plain `INSERT`
+// (without `ON CONFLICT ... DO UPDATE`) or `DELETE` `RETURNING`, because it is not useful
+// there as one can just use `RETURNING column_name`. (`ON CONFLICT DO NOTHING` never returns
+// untouched rows, so allowing `ON CONFLICT DO NOTHING ... RETURNING` might be misleading to
+// users on that regard - I, the author, have seen bugs caused by not being aware of this
+// PG behavior.)
+
+impl<C> SelectableExpression<ReturningQuerySource<UpdateStmt, C::Table>> for Old<C>
+where
+    C: Column,
+    Self: AppearsOnTable<ReturningQuerySource<UpdateStmt, C::Table>>,
+{
+}
+
+impl<C, DB> QueryFragment<DB> for Old<C>
+where
+    DB: Backend,
+    Self: QueryFragment<DB, DB::ReturningClause>,
+{
+    fn walk_ast<'b>(&'b self, pass: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        <Self as QueryFragment<DB, DB::ReturningClause>>::walk_ast(self, pass)
+    }
+}
+
+impl<C, DB> QueryFragment<DB, sql_dialect::returning_clause::PgLikeReturningClause> for Old<C>
+where
+    DB: Backend<ReturningClause = sql_dialect::returning_clause::PgLikeReturningClause>,
+    C: Column,
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, DB>) -> QueryResult<()> {
+        out.push_sql("old.");
+        out.push_identifier(C::NAME)?;
+        Ok(())
+    }
+}
+
+pub use return_type_helpers_reexported::*;
+
+pub(crate) mod return_type_helpers_reexported {
+    use super::Old;
+
+    /// The return type of [`old(col)`](super::old()).
+    #[allow(non_camel_case_types)]
+    pub type old<C> = Old<C>;
+}
--- a/diesel/src/query_builder/delete_statement/mod.rs
+++ b/diesel/src/query_builder/delete_statement/mod.rs
@@ -1,7 +1,9 @@
 use crate::backend::DieselReserveSpecialization;
 use crate::dsl::{Filter, IntoBoxed, OrFilter};
-use crate::expression::{AppearsOnTable, SelectableExpression};
-use crate::query_builder::returning_clause::*;
+use crate::expression::{AppearsOnTable, Expression, SelectableExpression};
+use crate::query_builder::returning::{
+    DeleteStmt, NoReturningClause, ReturningClause, ReturningQuerySource,
+};
 use crate::query_builder::where_clause::*;
 use crate::query_builder::*;
 use crate::query_dsl::RunQueryDsl;
@@ -269,8 +271,8 @@ where
 impl<T, U> AsQuery for DeleteStatement<T, U, NoReturningClause>
 where
     T: Table,
-    T::AllColumns: SelectableExpression<T>,
     DeleteStatement<T, U, ReturningClause<T::AllColumns>>: Query,
+    T::AllColumns: SelectableExpression<ReturningQuerySource<DeleteStmt, T>>,
 {
     type SqlType = <Self::Query as Query>::SqlType;
     type Query = DeleteStatement<T, U, ReturningClause<T::AllColumns>>;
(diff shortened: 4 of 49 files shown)


diesel/src/pg/returning/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! PostgreSQL-specific `RETURNING` clause helpers.
+//!
+//! This module exposes [`old`] / [`Old`], the wrapper used to refer to the
+//! pre-modification value of a column in a PostgreSQL `RETURNING` clause —
+//! the `RETURNING old.col` syntax introduced in PostgreSQL 18.
+
+mod old;
+
+pub use self::old::{Old, old};
```
Ten0: we should add helper types for compatibility with auto_type

diesel/src/query_builder/returning/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+//!   SQL. This submodule is only re-exported when the
+//!   `i-implement-a-third-party-backend-and-opt-into-breaking-changes`
+//!   feature is enabled.
+//!
+//! [`ReturningQuerySource`]: returning_query_source::ReturningQuerySource
+//! [`InsertStmtKind`]: returning_query_source::InsertStmtKind
+//! [`ReturningClause<E>`]: returning_clause::ReturningClause
+//! [`NoReturningClause`]: returning_clause::NoReturningClause
+
+pub mod returning_query_source;
```
Ten0: reconsider whether visibility of this, current module, and inner types, are most appropriate
this should be homogeneous with how visibility is otherwise generally handled in Diesel

diesel/src/query_builder/delete_statement/mod.rs
```diff
@@ -314,7 +316,6 @@ impl<T: QuerySource, U> DeleteStatement<T, U, NoReturningClause> {
     /// ```
     pub fn returning<E>(self, returns: E) -> DeleteStatement<T, U, ReturningClause<E>>
     where
-        E: SelectableExpression<T>,
```
Ten0, line 317: note for review (nothing to do): removed to be homogeneous with `returning` function on update & insert, but maybe we want to add it everywhere instead.
We are covered for this by `DeleteStatement<T, U, ReturningClause<E>>: Query` which implies this (or the new version of this) anyway, so that removes dupication.
I expect the error message is unchanged.

diesel/src/query_builder/returning/returning_query_source.rs
```diff
@@ -0,0 +1,191 @@
...
+/// which `Nullable<Old<C>>` is a valid `RETURNING` element) but reject plain
+/// `INSERT` (where the marker is [`InsertStmt`], for which `Old<C>` does not
+/// implement `SelectableExpression`).
+///
+/// The trait is sealed in spirit — it only has impls for the values shapes
+/// `diesel` itself produces — but it is exposed publicly under the
+/// `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature
+/// so third-party backends that introduce new values shapes can add their own
+/// impls.
+pub trait InsertStmtKind {
```
Ten0, line 84: note for review (nothing to do): We need this trait to extract whether we have an on conflict do update on the insert statements, because instead of being stored as a separate field there, it's stored as part of the values clause.
At first glance this looks weird that it isn't just a separate field, like on e.g. update.
I wonder whether it might be ok to change that (possibly put it in QS instead, having QS be ReturningQuerySource (or `MaybeReturningQuerySource<NoReturningClause/ReturningClause<xxx>, CurrentQS>` to avoid conflicting impls)), which in turn may remove the need for this extraction trait.
Although even with this, maybe the extraction layer ends up being the simplest abstraction to factor between update statements and insert statements with/without on conflict.

diesel/src/query_builder/returning/returning_query_source.rs
```diff
@@ -0,0 +1,191 @@
...
+/// ... DO UPDATE` (where the marker is [`InsertOnConflictDoUpdateStmt`], for
+/// which `Nullable<Old<C>>` is a valid `RETURNING` element) but reject plain
+/// `INSERT` (where the marker is [`InsertStmt`], for which `Old<C>` does not
+/// implement `SelectableExpression`).
+///
+/// The trait is sealed in spirit — it only has impls for the values shapes
+/// `diesel` itself produces — but it is exposed publicly under the
+/// `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature
+/// so third-party backends that introduce new values shapes can add their own
+/// impls.
```
Ten0, line 83: double check that this is indeed true at the end, after "reconsider visibility" comment

diesel/src/pg/returning/mod.rs
Ten0: 991d6b4f16a1d652c1b18f7c415dcb4a46ea2653

diesel/src/query_builder/returning/mod.rs
Ten0: 65af2dbedb4d16cd3ba2ad4f0317d31e30717218

diesel_derives/src/table.rs
```diff
@@ -890,6 +889,16 @@ fn expand_column_def(
         impl diesel::SelectableExpression<super::#query_source_ident> for #column_name {
         }
 
+        impl<__StmtKind>
```
Ten0, line 892: note for review (nothing to do): unfortunately this can't be implemented generically for `T: Column` without running into conflicting impls, so we have to implement it explicitly like is the case for the regular `SelectableExpression<Table> for Column`.
I expect that users don't typically manually implement `SelectableExpression<ConcreteType> for Xxx`, esp. not to use it a returning clause later on, so that shouldn't cause much breakage.

diesel/src/pg/mod.rs
```diff
@@ -12,6 +12,7 @@ pub(crate) mod backend;
 pub(crate) mod connection;
 mod metadata_lookup;
 pub(crate) mod query_builder;
+pub mod returning;
```
Ten0, line 15: note for review (nothing to do): I opted to put `old` in `pg::returning::old` because for now we don't support MariaDB's version, and MariaDB's version doesn't work on Mysql, and other backends don't support it, so it feels like we might want to export an additional mariadb-specific query fragment if we export this, rather than factoring it and making it always available for Mysql-like backends.
So I opted to being conservative here and scoped it to `pg`.
Worse case scenario if we change our mind on this we can always move it later and add deprecated doc-hidden re-exports here to avoid breaking changes, so that's not too big a commitment.

diesel/src/pg/returning/old_impl.rs
```diff
@@ -0,0 +1,168 @@
...
+impl<C> ValidGrouping<()> for Old<C>
+where
+    C: Column,
+{
+    type IsAggregate = is_aggregate::No;
+}
+
+// `Old<C>` is selectable on a `RETURNING` clause whose statement-kind marker
+// is `UpdateStmt`. It is deliberately *not* selectable on
+// `ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, _>` directly — only
```
weiznich, line 110: This comment doesn't match the actual code

The constraint mentioned here is important, we shouldn't allow selecting non nullable values for insert (on conflict) statements

diesel/src/pg/returning/old_impl.rs
```diff
@@ -0,0 +1,168 @@
...
+// `INSERT ... ON CONFLICT ... DO UPDATE RETURNING` and reject `old(col)`
+// alone at compile time.
+impl<C> AppearsOnTable<ReturningQuerySource<UpdateStmt, C::Table>> for Old<C>
+where
+    C: Column,
+    Self: Expression,
+{
+}
+
+impl<C> AppearsOnTable<ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, C::Table>> for Old<C>
```
weiznich, line 124:
```suggestion
impl<C> AppearsOnTable<ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, C::Table>> for Nullable<Old<C>>
```

At least the comment above indicates that.

The other code and the compile test indicates that it works in a slightly different way, so maybe the comment needs to be adjusted to match the behaviour?

diesel/src/query_builder/returning/returning_query_source.rs
```diff
@@ -0,0 +1,132 @@
...
+//!
+//! See [`ReturningQuerySource`] for the high-level picture.
+
+use crate::query_source::joins::ToInnerJoin;
+use crate::query_source::{AppearsInFromClause, Once};
+use core::marker::PhantomData;
+
+/// Statement-kind marker
+#[derive(Debug, Copy, Clone)]
+pub struct InsertStmtWithoutOnConflictDoUpdate;
```
weiznich, line 14: I'm not sure if it's meaningful to differentiate between `INSERT` statements with and without `ON CONFLICT` clauses, as both can return null values for `old` values if the row doesn't exist. For plain `INSERT` statements you just get always a null value, but that doesn't matter that much from type system perspective.  So yes it's not meaningful but still valid to use this with plain insert statements, but that's then a users choice?

diesel/src/pg/returning/old_impl.rs
```diff
@@ -0,0 +1,168 @@
...
+// `Old<C>` is selectable on a `RETURNING` clause whose statement-kind marker
+// is `UpdateStmt`. It is deliberately *not* selectable on
+// `ReturningQuerySource<InsertStmtWithOnConflictDoUpdate, _>` directly — only
+// `Nullable<Old<C>>` is, via the existing `Nullable` machinery and the
+// `ToInnerJoin` mapping in `returning_query_source` (which makes
+// `InsertStmtWithOnConflictDoUpdate`'s inner-join "fall back" to `UpdateStmt`).
+// That's how we force users to write `old(col).nullable()` in an
+// `INSERT ... ON CONFLICT ... DO UPDATE RETURNING` and reject `old(col)`
+// alone at compile time.
+impl<C> AppearsOnTable<ReturningQuerySource<UpdateStmt, C::Table>> for Old<C>
```
weiznich, line 117: Do we miss impls for `DeleteStmt`?

diesel/src/query_builder/mod.rs
```diff
@@ -95,9 +94,10 @@ pub(crate) use self::insert_statement::{UndecoratedInsertRecord, ValuesClause};
 #[doc(inline)]
 pub use self::insert_statement::{DefaultValues, InsertOrIgnore, Replace};
 
+#[cfg(not(feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"))]
```
weiznich, line 97: There is the `__diesel_public_if` proc macro for this

diesel_compile_tests/tests/fail/custom_returning_requires_selectable_expression.rs
```diff
@@ -30,14 +30,16 @@ fn main() {
     let stmt = update(users.filter(id.eq(1)))
         .set(name.eq("Bill"))
         .returning(bad::age);
-    //~^ ERROR: cannot select `bad::columns::age` from `users::table`
+    //~^ ERROR: cannot select `bad::columns::age` from `ReturningQuerySource<UpdateStmt, table>`
```
weiznich, line 33: We might try to use `#[diagnostic::do_not_recommend]` on the relevant impl to improve the error message? If that doesn't work out that's still "good enough"

diesel_compile_tests/tests/fail/old_in_insert_on_conflict_do_update_requires_nullable.rs
```diff
@@ -0,0 +1,53 @@
...
+        .on_conflict(users::id)
+        .do_update()
+        .set(users::name.eq(""))
+        .returning(UpsertOldNew::as_select())
+        //~^ ERROR: cannot select `returning::old_impl::Old<columns::name>` from `ReturningQuerySource<..., ...>`
+        .get_result::<UpsertOldNew>(&mut connection)
+        //~^ ERROR: cannot select `returning::old_impl::Old<columns::name>` from `ReturningQuerySource<..., ...>`
+        .unwrap();
+
+    // The plain tuple version mirrors the same constraint: writing
```
weiznich, line 41: Can we also add a passing variant here that correctly uses `.nullable()`?

diesel_derives/src/table.rs
weiznich: I wouldn't care much about the breakage here as this is mostly required for the new feature anyway, right? In that case it's in my opinion not unexpected that you might need to implement additional traits anyway.

diesel/src/pg/mod.rs
weiznich: That's reasonable :+1:

diesel/src/pg/returning/old_impl.rs
weiznich: Given that some of this happens in a different file, it also might be useful to move everything into one location?

weiznich (commented):
Overall I like the implementation. This looks reasonable to add

I still have some smaller questions and  minor requests to improve certain details.

diesel_derives/src/table.rs
Ten0: > I wouldn't care much about the breakage here as this is mostly required for the new feature anyway, right?

I don't see a way around it so I'd say it's "required for the new feature" but if people had written custom expressions, implemented `SelectableExpression<ConcreteType>` on them and then used them in return clauses, this would break them and they'd need to manually implement that as well.
(Which sounds reasonable to me because unlikely enough.)

diesel/src/query_builder/returning/returning_query_source.rs
Ten0: That makes sense... The idea behind this is that in the past we've had bugs where we attempted to use returning when doing "on conflict do nothing", which actually doesn't return the untouched rows, and I was hoping this would actually help prevent these bugs for users.
("Why is the interface like this? Ah that's why. (documented)")

diesel/src/pg/returning/old_impl.rs
Ten0: I opted to not have them: for returning from a delete just listing the field names gives old values, so writing `old.` seems less idiomatic, so I opted to not add it to leave "one standard way to write things". But I have no strong opinion here, I'm open to adding it as well if you think it's best. 🙂

diesel_compile_tests/tests/fail/custom_returning_requires_selectable_expression.rs
Ten0: I'm not sure what `impl` you're referring to.

diesel_compile_tests/tests/fail/old_in_insert_on_conflict_do_update_requires_nullable.rs
Ten0: 774b935231f

diesel/src/pg/returning/old_impl.rs
Ten0: We implement this is `AppearsOnTable for Old here`, not `SelectableExpression`, which will in turn implement `SelectableExpression<Nullable<Old>>` via the `ToInnerJoin` "regular" `ToInnerJoin` transformation.
Isn't this the idiomatic way to implement it?

diesel/src/pg/returning/old_impl.rs
Ten0: Maybe I'm missing something but it seems like [returning/old_impl.rs](https://github.com/diesel-rs/diesel/pull/5049/files/adb0dc6822d5c365ce8a7742ed017a852f4d8c11#diff-441cc5dd9c3c11731e4b4eaccbd7c67e174ef92e512473181e365531f6be7d5a) showcases that it works as comment describes?

(Maybe linked to https://github.com/diesel-rs/diesel/pull/5049#discussion_r3254685818)

diesel/src/query_builder/mod.rs
Ten0: Doesn't seem to work for module declarations:
```
error[E0658]: file modules in proc macro input are unstable
   --> diesel/src/query_builder/mod.rs:100:1
    |
100 | pub(crate) mod returning;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54727 <https://github.com/rust-lang/rust/issues/54727> for more information

For more information about this error, try `rustc --explain E0658`.
```

diesel/src/pg/returning/old_impl.rs
weiznich: No that's fine. I think my main problem here is/was that parts of this are in this file, while other parts are in a different file. It might be a good idea to just put a few more comments about that in the code. 

The implementation is sound as far as I can tell.

diesel/src/query_builder/returning/returning_query_source.rs
weiznich: If you have feelings about this, it should also be fine to disallow non-meaningful variants. We should just explicitly document that and maybe have a compile fail test for that.

diesel/src/pg/returning/old_impl.rs
weiznich: It's fine to not have them, but I would like to have a comment with the reason then. Otherwise I fear someone will wonder about that in a few month/years.

diesel/src/query_builder/mod.rs
weiznich: That's sad. I will put it on my list to poke compiler people about this.

diesel_compile_tests/tests/fail/custom_returning_requires_selectable_expression.rs
weiznich: After reading the error message again, I noticed that this is a misconception. Never mid the comment then.

Fixing that would require work on `#[diagnostic::on_unimplemented]`

diesel_derives/src/table.rs
weiznich: Let's just go with and see if anyone complains then.

diesel/src/pg/returning/old_impl.rs
weiznich: After reading over it again: I think the actual code is fine. I'm likely just confused by the comment in combination with having some impls in this file and some impls in the general returning file. Might be worth to put at least a note in the comment saying that parts are over in the other file?

weiznich:
Would it also be possible to squash all commits and resolve the merge conflict? Otherwise its mostly fine, so feel free to merge it after doing that.

diesel/src/pg/returning/old_impl.rs
Ten0: Added this: https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/query_builder/returning/returning_query_source.rs#L5-L6

diesel/src/pg/returning/old_impl.rs
Ten0: Added this: https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/query_builder/returning/returning_query_source.rs#L5-L6

diesel/src/query_builder/returning/returning_query_source.rs
Ten0: We had this:
https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/pg/returning/old_impl.rs#L54-L56

And I added this:
https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/pg/returning/old_impl.rs#L131-L136

And a compile test:
[`7b83eb0` (this PR)](https://github.com/diesel-rs/diesel/pull/5049/commits/7b83eb035143968af70df005f4226904e8ac6e04)

diesel/src/pg/returning/old_impl.rs
Ten0: We had this:
https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/pg/returning/old_impl.rs#L54-L56

And I added this:
https://github.com/Ten0/diesel/blob/538fa0aa28e4c2c768bd302ff1a61fed5fef05e1/diesel/src/pg/returning/old_impl.rs#L131-L136

weiznich (approved):
Looks good now.

For transparency: I just squashed all the commits and used the PR description as commit message. That's hopefully much more understandable in the future than 50 WIP commits 🙈

Ten0:
Yes that's what I wanted to do but I was going to use the "squash and merge" button on the PR rather than squashing beforehand, so that history would be preserved here.

weiznich:
Sorry for the trouble. I just wanted to merge the PR without letting it sit and rot for too long. Will try to remember next time that this button is exists.

PR #5050, opened 2026-05-09
Allow casting to REAL in sqlite

clonejo:
https://www.sqlite.org/datatype3.html#affinity_name_examples

  - i assume this is too minor for the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,13 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ## Unreleased
 
 ### Added
+
 * Diesel-Migrations now contains a migration source that easily allows you to register Rust based migrations
 * Diesel-Migrations now contains a migration source that allows you to combine migrations from several different sources
 * Added `SqliteConnection::with_raw_connection` to provide safe, callback-based access to the raw `*mut sqlite3` handle for advanced SQLite C APIs (session extension, hooks, etc.)
 * Added `json_extract` and `jsonb_extract` SQL function support for the SQLite backend
 * Added `SqliteConnection::get_read_only_blob` method to stream blob's from a SQLite database to Rust via `std::io::Read`
+* Added support for casting to `REAL` in SQLite.
 
 ### Fixed
 
--- a/diesel/src/expression/cast.rs
+++ b/diesel/src/expression/cast.rs
@@ -139,6 +139,7 @@ type_name! {
     diesel::sqlite::Sqlite: "sqlite" {
         Int4 => "integer",
         Int8 => "bigint",
+        Double => "real",
         Text => "text",
         Json => "json",
         Jsonb => "jsonb",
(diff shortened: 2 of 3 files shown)


diesel/src/expression/cast.rs
```diff
@@ -139,6 +139,7 @@ type_name! {
     diesel::sqlite::Sqlite: "sqlite" {
         Int4 => "integer",
         Int8 => "bigint",
```
clonejo, line 141: Casting to `Int4` is misleading, sqlite only has 8-byte signed integers, called `INTEGER`, so casting to `Int4` is the same as casting to `Int8`.

I don't think there is any special handling when casting: https://www.sqlite.org/lang_expr.html#castexpr

ayarotsky (commented):
The change looks good. We need two more things before we merge it:
1. Please extend `diesel_tests/tests/cast.rs` with new tests to verify this change.
2. Add a changelog entry.

diesel_tests/tests/cast.rs
```diff
@@ -149,14 +149,14 @@ mod infallible_cast {
     test_infallible_cast!(int8_to_float4, 3 => i64, BigInt => Float, 3.0 => f32);
     #[cfg(feature = "postgres")]
     test_infallible_cast!(int4_to_float8, 3 => i32, Integer => Double, 3.0 => f64);
```
ayarotsky, line 151: Please enable this test for `sqlite` as well.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Added `SqliteConnection::with_raw_connection` to provide safe, callback-based access to the raw `*mut sqlite3` handle for advanced SQLite C APIs (session extension, hooks, etc.)
 * Added `json_extract` and `jsonb_extract` SQL function support for the SQLite backend
 * Added `SqliteConnection::get_read_only_blob` method to stream blob's from a SQLite database to Rust via `std::io::Read`
+* Allow casting to `REAL` in SQLite.
```
ayarotsky: Let's keep the chanelog format consistent and rephrase this to:
```
Added support for casting to `REAL` in SQLite.
```

ayarotsky (commented):
A couple of small changes, and we're good to merge.

diesel_tests/tests/cast.rs
clonejo: added. not sure how useful it is, since int4 in sqlite just aliases to int8.

ayarotsky (approved):
Thank you for your contribution. Looks great.

diesel_tests/tests/cast.rs
weiznich: Yes `int4` is essentially the same as `int8` on the SQLite backend from the database point of view. From diesel's point of view they are not the same, so it is still meaningful to have this conversation here, especially if you want to write code targeting more than one database backend at the same time

diesel/src/expression/cast.rs
weiznich: As written in the other comment: 

From SQLite's type system perspective you are correct. From diesel's side they are still "different" types. This especially matters if you write database code targeting more than one database backend at the same time. So it's meaningful to keep it that way.

PR #5058, opened 2026-05-25
QueryableByName is now nostd.

oolonek:
Make `#[derive(QueryableByName)]` no-std compatible.

Codex was used to draft and test this small change.

### DIFF
--- a/diesel_compile_tests/tests/fail/derive/queryable_by_name_no_std.rs
+++ b/diesel_compile_tests/tests/fail/derive/queryable_by_name_no_std.rs
@@ -1,0 +1,20 @@
+//@check-pass
+//@compile-flags: --crate-type lib
+#![no_std]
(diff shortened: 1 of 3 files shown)


LucaCappelletti94 (approved):

LucaCappelletti94:
I did this PR as a small Rust example with my colleague @oolonek, I am pretty sure all is good

pksunkara:
We haven't made any other thing no-std compatible. Why this? Is this even what we want to do for diesel?

LucaCappelletti94:
Diesel is no-std compatible, for only the SQLite backend. This was a struggler that had remained behind.

LucaCappelletti94:
See https://github.com/diesel-rs/diesel/pull/4906 for reference

pksunkara (approved):

PR #5059, opened 2026-05-25
Fix order_by/then_order_by with GROUP BY rejecting valid aggregate expressions

### LINKED ISSUE #5057: OrderDsl is not supported for aggregated statement even with group by.
### Rust Version

1.95.0

### Diesel Version

2.3.8

### Diesel Features

postgres

### Database Version

PostgresSQL 18

### Operating System Version

Fedora Linux 44

### What happened?

A regression between version 2.3.7 and 2.3.8. Following example compiles in diesel 2.3.7 while it does not in diesel 2.3.8. I think this has something to do with the statement `Raise a compile-time error when mixing aggregate and non-aggregate expressions in an ORDER BY clause without a GROUP BY clause` located inside 2.3.8 changelog.

The same statement that compiles in diesel 2.3.7 and is accepted by postgres should still compile in 2.3.8 or 2.3.9 (latest).

### Additional details

_No response_

### Steps to reproduce

```rust
use diesel::prelude::*;

/// Cargo.toml
/// diesel = { version = "=2.3.8", features = ["postgres"] }

/// 
/// create table parent( id serial primary key not null, capacity int not null default 10 );
/// 
/// create table child( id serial primary key not null, parent_id int not null references parent );
/// 

diesel::table! {
    parent (id) {
        id -> Int4,
        capacity -> Int4,
    }
}

diesel::table! {
    child (id) {
        id -> Int4,
        parent_id -> Int4,
    }
}

diesel::joinable!(child -> parent (parent_id));
diesel::allow_tables_to_appear_in_same_query!(parent, child);

fn main() {
    let query = parent::table
        .left_join(child::table)
        .group_by(parent::id)
        .order_by(diesel::dsl::count(child::id).asc())
        .select(parent::id);

    println!("{}", diesel::debug_query::<diesel::pg::Pg, _>(&query));
}
```

The produced sql statement is below and accepted by postgres.

```sql
SELECT "parent"."id" FROM ("parent" LEFT OUTER JOIN "child" ON ("child"."parent_id" = "parent"."id")) GROUP BY "parent"."id" ORDER BY count("child"."id") ASC
```

### Compile time error

```rust
error[E0277]: the trait bound `SelectStatement<FromClause<JoinOn<Join<table, table, LeftOuter>, ...>>, ..., ..., ..., ..., ..., ...>: OrderDsl<_>` is not satisfied
   --> src/main.rs:24:10
    |
 24 |         .order_by(diesel::dsl::count(child::id).asc())
    |          ^^^^^^^^ the trait `OrderDsl<_>` is not implemented for `SelectStatement<FromClause<JoinOn<Join<table, table, LeftOuter>, Grouped<...>>>, ..., ..., ..., ..., ..., ...>`
    |
help: the trait `OrderDsl<Expr>` is implemented for `query_builder::select_statement::SelectStatement<query_builder::from_clause::FromClause<F>, S, D, W, O, LOf, G, H, LC>`
   --> .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-2.3.8/src/query_builder/select_statement/dsl_impls.rs:270:1
    |
270 | / impl<ST, F, S, D, W, O, LOf, G, H, LC, Expr> OrderDsl<Expr>
271 | |     for SelectStatement<FromClause<F>, S, D, W, O, LOf, G, H, LC>
272 | | where
273 | |     F: QuerySource,
...   |
283 | |     <S::Selection as ValidGrouping<G::Expressions>>::IsAggregate:
284 | |         MixedAggregates<<Expr as ValidGrouping<G::Expressions>>::IsAggregate>,
    | |______________________________________________________________________________^
    = note: the full name for the type has been written to '/target/debug/deps/diesel_bug_report-73e760db963c88fc.long-type-14113715080190992066.txt'

error: could not compile `diesel-bug-report` (bin "diesel-bug-report") due to 1 previous error
```

### Checklist

ayarotsky:
Since `v2.3.8`, calling `order_by()` or `then_order_by()` with an aggregate expression on a query that has a `group_by()` clause incorrectly failed to compile, even though the resulting SQL is valid. The check introduced in `v2.3.8` validated the `SELECT` clause against the `ORDER BY` expression at `order_by()` call time, which blocked valid patterns like:

```rust
table.group_by(col).order_by(max(other_col)).select(...)
```

The fix splits the `OrderDsl` and `ThenOrderDsl` impls into two variants: one for `NoGroupByClause` (preserves the existing `MixedAggregates` check) and one for `GroupByClause<GB>` (only requires the `ORDER BY` expression to satisfy `ValidGrouping<GB>`, i.e. be a grouped column or aggregate).

Closes https://github.com/diesel-rs/diesel/issues/5057

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * `Bpchar` is now a distinct PostgreSQL SQL type (previously a hidden alias for `Varchar`). Binds on `CHAR(N)` / `BPCHAR` columns are now sent with OID 1042, allowing PostgreSQL to use the column's index instead of casting it to text.
 * Fix non-deterministic test failures on PostgreSQL caused by loading rows without `ORDER BY` and assuming insertion order
 * `diesel_derives` does now correctly handle feature flag unification in mixed build/target dependency situations
+* Fix a regression introduced in 2.3.8 where `order_by()` and `then_order_by()` with an aggregate expression incorrectly failed to compile when a `group_by()` clause was present
 
 ### Changed
 
--- a/diesel/src/query_builder/select_statement/dsl_impls.rs
+++ b/diesel/src/query_builder/select_statement/dsl_impls.rs
@@ -378,7 +378,7 @@ where
     }
 }
 
-// With GROUP BY: only validate that the new ORDER BY term is valid for the group.
+// With GROUP BY: Validate the whole order expression against the given group by expression
 impl<ST, F, S, D, W, O, LOf, GB, H, LC, Expr> ThenOrderDsl<Expr>
     for SelectStatement<FromClause<F>, S, D, W, OrderClause<O>, LOf, GroupByClause<GB>, H, LC>
 where
@@ -387,7 +387,7 @@ where
     Self: SelectQuery<SqlType = ST>,
     SelectStatement<FromClause<F>, S, D, W, OrderClause<(O, Expr)>, LOf, GroupByClause<GB>, H, LC>:
         SelectQuery<SqlType = ST>,
-    Expr: ValidGrouping<GB>,
+    (O, Expr): ValidGrouping<GB>,
 {
     type Output = SelectStatement<
         FromClause<F>,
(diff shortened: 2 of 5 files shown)


ayarotsky:
@weiznich @pksunkara what do you think?

weiznich (approved):
Thanks for working on this fix. This looks good to me. I did some minor followup adjustments + a rebase.

PR #5062, opened 2026-05-27
Add SQLite json_insert and jsonb_insert functions

eyupcanakman:
Adds the SQLite `json_insert` and `jsonb_insert` functions. SQLite stores newly inserted string text and object keys with element type `JSONB_TEXTRAW`, which the diesel jsonb reader previously rejected, so this also routes `TEXTRAW` through `read_jsonb_text` per [#4366 (comment)](https://github.com/diesel-rs/diesel/issues/4366#issuecomment-4535037694). Partially completes #4366.

I used Claude Code (claude-opus-4-7) as a coding assistant for this change. I verified and tested it myself.


weiznich (approved):
Thanks for submitting this PR. The added functions look good, although I slightly adjusted the return type to reflect the fact that this never will return a null value as long as we don't pass as SQL null value in.

PR #5063, opened 2026-05-28
Remove any feature flag related code generation in diesel_derives

weiznich:
Feature flags and proc-macros do not work together in a good way due to feature flag unification for proc-macros happening for sub-dependencies of build and target dependencies. This bit us several times in the past, first with the resolver changes done with the 2018 edition.

During Rustweek David Tolnay pointed out to me that you could use this macro based workaround to side step this issue. The trick here is to have a macro in the "parent" crate that is feature flaged and only coditionally expands the tokens to what we want depend on if the feature flag is on or not. This is mostly straightforward to implement in diesel_derives for the real code generation. There are still a few pain points:

* Our test/snapshot/expanded code in the docs setup would break if we just do this, as the macros are not expanded there. To fix this I wrote some minor post-processing step for these tests to just "expand" the macros by just removing the macro call, and keep the inner tokens for the "right" feature flags.
* The error reporting for tables/views with too many columns get worse as we cannot detect that in the proc-macro anymore and emit a compile_error! there. We now use an assert in a const block for this, which gives at least some message pointing in the right direction. This is still not as good as the old message as it only shows up on `cargo build` and not on `cargo check` and it is only emitted after the trait errors due to the missing impls. It also has no access to the actual number of columns currently supported at that point.
 I'm unfortunally not aware of a good
way to improve this, beside eventually parsing your `Cargo.toml` file from `diesel-cli` and emit the message on generation side there as well?

After these changes none of the feature flags in `diesel_derives` have an affect on the generated code anymore. They are still there to guide the tests, but for the "real" code they are unused.

Finally this PR introduce another feature feature flag called `custom-count-column-table` that if enabled allows the user to set a `DIESEL_MAX_COLUMN_COUNT` environment variable to configure the maximal number of supported columns generated by the `diesel_derives::for_each_tuple!` macro. This enables users to configure a larger number of columns than 128 and also to use something in between the existing features to have a better control over their compile times.


diesel_derives/src/multiconnection.rs
JohnTitor, line 1: I guess you'd also need to change https://github.com/weiznich/diesel/blob/558d01186d552d87231c2a4cc8cd2c6896c846f9/diesel_derives/src/multiconnection.rs#L1536

diesel/src/lib.rs
```diff
@@ -197,6 +197,12 @@
 //!   `32-column-tables` feature. Enabling this feature will increase your compile times.
 //! - `128-column-tables`: This feature enables support for tables with up to 128 columns. It implies the
 //!   `64-column-tables` feature. Enabling this feature will increase your compile times significantly.
+//! - `custom-count-column-tables`: This features allows to customize the number of colums
```
JohnTitor:
```suggestion
//! - `custom-count-column-tables`: This feature allows to customize the number of columns
```

JohnTitor (approved):
Haven't reviewed precisely but I love the idea here. Great work!

PR #5065, opened 2026-05-29
Fix some expressions breaking in RETURNING old.xxx

Ten0:
3e94a1d33e7624c698e7c6c9db9bc7bbaf812adb missed a no-breaking-change test for some expressions that might actually be used by users: specifically subselects, and boxed expressions. 
This commit fixes backwards compatibility for these two use-cases, introducing tests for them.
This is essentially achieved by adding more idiomatic implementations of `AppearsInFromClause<ReturningQuerySource<...>> for ReturningQuerySource<...>` where before we only forwarded `AppearsInFromClause<Table> for AppearsInFromClause<ReturningQuerySource<..., Table>>`, which in turn allows propagating the `ReturningQuerySource<...>` QS to subqueries as usual, in such a way that the `Count` of `AppearsInFromClause` will be correct.

Also adds a compile test that "returning old.xxx" is rejected outside returning clauses.


weiznich:
Likely needs to have https://github.com/diesel-rs/diesel/pull/5064 merged first to update the compile tests output

diesel/src/query_source/mod.rs
```diff
@@ -119,7 +119,7 @@ where
     }
 }
 
-/// Determines how many times `Self` appears in `QS`
+/// Determines how many times `QS` appears in `Self`
```
Ten0, line 122: I'm reasonably confident about most things in this PR, the main thing I'd like confirmation about is this change: [`a85dbd7` (this PR)](https://github.com/diesel-rs/diesel/pull/5065/changes/a85dbd72aa047dd34a723bbced2e6c302193cc67).
(I would squash the rest and keep this commit separate prior to merging.)

diesel_compile_tests/tests/fail/returning_clause_boxed_expression.rs
```diff
@@ -0,0 +1,30 @@
...
+table! {
+    users {
+        id -> Integer,
+        name -> VarChar,
+    }
+}
+
+// Boxed expression in RETURNING — fails because the boxed expression is
+// parameterized over the table (`users::table`) rather than
+// `ReturningQuerySource<UpdateStmt, users::table>`.
```
Ten0: I have not yet found a fix for this. Currently that means #5049 is actually breaking on that regard, which I had not pointed out initially in that other PR.
If we decide the feature is worth the breaking change because little enough users affected, we should probably at least add a note to the changelog.

weiznich: I will give it a try next week, maybe I can come up with something.

Ten0: This makes it work: ee7fb06e07951af3f754ef35fa365c8c7b4e03ac

diesel/src/query_builder/returning/returning_query_source.rs
```diff
@@ -131,3 +152,27 @@ impl<V, Target, Changeset, Tab, WhereClause> InsertStmtKind
...
+
+/// Make `Box<dyn BoxableExpression<table>>` be `SelectableExpression` in returning clauses.
+/// This is necessary for backwards-compatibility.
+///
+/// Unfortunately we cannot implement this directly for `dyn BoxableExpression`
+/// (and rely on our existing generic impls for `Box<T>`) because the compiler
+/// complains that there is already an automatic implementation of `SelectableExpression` for
+/// `dyn BoxableExpression`, even though the `QS` type is different.
+/// As a workaround, we implement it for `Box<dyn BoxableExpression<table>>`, which should
+/// cover for most users.
```
Ten0, line 165: Not sure if it's worth implementing those also for Rc/Arc/&_... seems like a lot of boilerplate and I'm unsure to what extent users actually use this.
https://github.com/diesel-rs/diesel/blob/c5d681ebe8de20ce14196293942a4e55bda0030a/diesel/src/expression/mod.rs#L376-L402

weiznich: The `Arc`/`Rc` impls for `dyn BoxableExpression` only exist on main so far, so not having them is not e breaking change

weiznich (approved):

diesel/src/query_source/mod.rs
weiznich: This looks right, keeping the commit separate sounds like a good idea

weiznich:
This looks good to me. I pushed a minor fix for the `nostd` build. Feel free to merge it as soon as you squashed the commits.

PR #5069, opened 2026-06-02
Improve documentation for upsert `filter_target`

### LINKED ISSUE #3791: Documentation: Improve documentation for upsert `.filter_target()`
### Versions

- **Rust: 1.70**
- **Diesel: 2.1.0**
- **Database: Postgres (cockroach)**
- **Operating System: Mac OS 13**

### Feature Flags

- **diesel: `["chrono", "uuid", "serde_json"]`**

## Problem Description

Unless I missed something, I was not able to find an example for using `.filter_target()` in the inline documentation. This corresponds to `ON CONFLICT ... WHERE index_predicate DO UPDATE SET ...`, which is different from the `WHERE` clause after `DO UPDATE`. It took me a bit of digging to figure out that diesel actually supports this. 

It would be great if the we can add an example for this use case. 

Using the `ON CONFLICT ... WHERE index_predicate` syntax with partial unique indexes. 

N/A

N/A

N/A

### Steps to reproduce

Documentation for normal on conflict do update with extensive examples.
https://docs.diesel.rs/2.1.x/diesel/upsert/struct.IncompleteOnConflict.html#examples

`filter_target` function is grouped with the trait 
https://docs.diesel.rs/2.1.x/diesel/upsert/struct.IncompleteOnConflict.html#method.filter_target

## Checklist

- [x] (N/A, inline doc) This issue can be reproduced on Rust's stable channel. (Your issue will be
- [x] (N/A, inline doc) This issue can be reproduced without requiring a third party crate

## Potential solution

I propose that all the examples can be moved to the `upsert` module or the `IncompleteOnConflict` struct, which will make more sense if we were to add an example using `filter_target()`

PS: There should be a separate issue template documentation related issues.

ayarotsky:

Closes #3791

### DIFF
--- a/diesel/src/query_builder/upsert/on_conflict_target_decorations.rs
+++ b/diesel/src/query_builder/upsert/on_conflict_target_decorations.rs
@@ -32,17 +32,36 @@ pub trait DecoratableTarget<P> {
     ///
     /// # Example
     ///
-    /// ```rust,ignore
+    /// ```rust
+    /// # include!("../../upsert/on_conflict_docs_setup.rs");
+    /// # #[cfg(feature = "postgres")]
+    /// # fn main() -> diesel::QueryResult<()> {
+    /// #     use self::users::dsl::*;
+    /// #     let conn = &mut establish_connection();
+    /// #     diesel::sql_query("CREATE UNIQUE INDEX users_name_idx ON users (name)")
+    /// #         .execute(conn)?;
     /// diesel::insert_into(users)
-    ///     .values(&new_user)
-    ///     .on_conflict(name)
-    ///     .filter_target(id.gt(5))
-    ///     .do_nothing()
+    ///     .values(name.eq("Sam"))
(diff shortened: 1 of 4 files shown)


diesel/src/query_builder/upsert/on_conflict_target_decorations.rs
```diff
@@ -10,12 +10,41 @@ pub trait UndecoratedConflictTarget {}
...
+    ///
+    /// This generates `ON CONFLICT (target) WHERE predicate DO ...` SQL.
+    /// PostgreSQL selects unique indexes whose `WHERE` clause is implied by
+    /// the predicate; an exact match is not required.
+    ///
+    /// Calling `.filter_target()` multiple times combines predicates with `AND`.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
```
pksunkara: Please make this a proper doctest.

PR #5072, opened 2026-06-05
Use an in-memory SQLite database in migration doctests to avoid lock races

LucaCappelletti94:
The `RustMigrationSource` and `CombinedMigrationSource` doctests both run migrations against the shared `SQLITE_DATABASE_URL` file. When they run concurrently, one can fail with `DatabaseError(Unknown, "database is locked")`. This recently happened on the `(beta, sqlite, ubuntu-latest)` job in [a run for #4975](https://github.com/diesel-rs/diesel/actions/runs/27028692179/job/79775677160) and in [a merge queue run for #4981](https://github.com/diesel-rs/diesel/actions/runs/27015995452/job/79731459600).

This PR switches the sqlite branch of both doctests to a private in-memory database, making the race impossible. It also keeps the partially applied migrations of the `CombinedMigrationSource` example out of the shared test database.

### DIFF
--- a/diesel_migrations/src/rust_migrations.rs
+++ b/diesel_migrations/src/rust_migrations.rs
@@ -40,8 +40,9 @@ use crate::MigrationError;
 ///
 /// # #[cfg(feature = "postgres")]
 /// # let connection_url = database_url_from_env("PG_DATABASE_URL");
+/// # // An in-memory database avoids lock races between concurrently running doctests.
 /// # #[cfg(feature = "sqlite")]
-/// # let connection_url = database_url_from_env("SQLITE_DATABASE_URL");
+/// # let connection_url = String::from(":memory:");
 /// # #[cfg(feature = "mysql")]
 /// # let connection_url = database_url_from_env("MYSQL_DATABASE_URL");
 /// # #[cfg(feature = "postgres")]
--- a/diesel_migrations/src/rust_migrations.rs
+++ b/diesel_migrations/src/rust_migrations.rs
@@ -40,7 +40,8 @@ use crate::MigrationError;
 ///
 /// # #[cfg(feature = "postgres")]
 /// # let connection_url = database_url_from_env("PG_DATABASE_URL");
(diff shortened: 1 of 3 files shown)


diesel_migrations/src/rust_migrations.rs
```diff
@@ -40,8 +40,9 @@ use crate::MigrationError;
 ///
 /// # #[cfg(feature = "postgres")]
 /// # let connection_url = database_url_from_env("PG_DATABASE_URL");
+/// # // An in-memory database avoids lock races between concurrently running doctests.
 /// # #[cfg(feature = "sqlite")]
-/// # let connection_url = database_url_from_env("SQLITE_DATABASE_URL");
+/// # let connection_url = String::from(":memory:");
```
ayarotsky, line 46: It seems that `migrations_macros/src/lib.rs:59` should also be changed.

diesel_migrations/src/rust_migrations.rs
```diff
@@ -40,8 +40,9 @@ use crate::MigrationError;
 ///
 /// # #[cfg(feature = "postgres")]
 /// # let connection_url = database_url_from_env("PG_DATABASE_URL");
+/// # // An in-memory database avoids lock races between concurrently running doctests.
```
ayarotsky: The comment you provided at `diesel_migrations/src/combined_migrations.rs` is more complete. Please replicate it here as well.

ayarotsky (commented):
Looks great. A few small changes and we're good to go.

diesel_migrations/src/rust_migrations.rs
LucaCappelletti94: Oh ok, I actively didn't want to redound it

LucaCappelletti94:
I believe that the failure may be another flakiness, but I am unsure what is causing that one.

weiznich:
Seems to be a runner related failure. There is no much we can do about that outside of rerunning the job

PR #5074, opened 2026-06-09
Fix invalid step-level permissions in doc workflow

LucaCappelletti94:
The "Publish Docs" workflow fails to start on every push that evaluates it (reported as a startup failure: "This run likely failed because of a workflow file issue"), including the merge queue on `main`.

The cause is a step-level `permissions:` block on the "Publish documentation" step, added in 6745d0fc5. `permissions:` is only valid at the [workflow level or the job level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions), not on a step, so the whole workflow file is invalid and no run can start.

This moves the `contents: write` grant up to the `publish_docs` job (the job that pushes to `gh-pages`), keeping the workflow-level `contents: read` default.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -14,10 +14,12 @@ concurrency:
 
 name: Publish Docs
 jobs:
-  publish_docs:
+  build_docs:
     if: github.repository == 'diesel-rs/diesel'
-    name: Publish Docs
+    name: Build Docs
     runs-on: ubuntu-latest
+    permissions:
+      contents: "read"
     steps:
       - name: Checkout sources
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
@@ -27,11 +29,6 @@ jobs:
         uses: Swatinem/rust-cache@f51f967e158417aafa338a1e46ab22095f07aa01 # v2.8.2
         with:
           key: cargo-doc-cargo-${{ hashFiles('**/Cargo.toml') }}
-      - name: Get the branch name
-        id: current_branch
-        shell: bash
-        run: |
-          echo "branch=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_OUTPUT
       - name: Install rust toolchain
         run: rustup update nightly
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/doc.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     if: github.repository == 'diesel-rs/diesel'
     name: Publish Docs
     runs-on: ubuntu-latest
+    permissions:
```
weiznich, line 21: That somehow defeats the purpose of putting the permission on the step level. We rather only want to allow that single step to have publish rights at all.

I will do a fix for that and the metrics job hopefully later today

LucaCappelletti94: Problem is GitHub Actions does not support permissions at the step level at all, and that invalid key is what made the whole workflow fail to start. You can only scope the token per job, not per step. So I have now tentatively split it into two jobs, where `build_docs` builds the docs with read-only access and uploads them as an artifact, then `publish_docs` downloads that and deploys with write access. Now only the small deploy job can write. Pushed it here so you do not have to redo it.

.github/workflows/doc.yml
```diff
@@ -43,11 +40,37 @@ jobs:
...
+    name: Publish Docs
+    needs: build_docs
+    runs-on: ubuntu-latest
+    permissions:
+      contents: "write"
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
```
weiznich, line 61: Given that we just download the artifact below we don't really need the checkout here, right?

LucaCappelletti94: The artifact gives us the docs content, but the deploy still needs the checkout because the action pushes to gh-pages through the local git repo and, to my understanding, never clones on its own. It just runs [`git remote add origin` plus a worktree in the workspace](https://github.com/JamesIves/github-pages-deploy-action/blob/9d877eea73427180ae43cf98e8914934fe157a1a/src/git.ts#L128-L142), so with no checkout there is no repo to push from. [Their docs say the checkout is required](https://github.com/JamesIves/github-pages-deploy-action/blob/9d877eea73427180ae43cf98e8914934fe157a1a/README.md#L33), and [their own artifact based two job example keeps the checkout in the deploy job](https://github.com/JamesIves/github-pages-deploy-action/blob/9d877eea73427180ae43cf98e8914934fe157a1a/README.md#L205-L250). `persist-credentials: false` is the pairing they recommend with it.

weiznich: Ah right, thanks for writing this down.

weiznich (approved):

PR #5076, opened 2026-06-11
Fix flaky MySQL setup on the Windows CI runner

LucaCappelletti94:
The Windows MySQL job intermittently fails in setup because it waits a fixed `sleep 15` for `mysqld` to start before creating the test databases, and on the runner the server occasionally needs longer, so the client races it and the step aborts. This replaces the fixed sleep with a `mysqladmin ping` readiness loop that proceeds as soon as the server accepts connections.


ayarotsky (approved):
Looks good 🎉

PR #5077, opened 2026-06-12
Add SqliteConnection::on_commit wrapping sqlite3_commit_hook

LucaCappelletti94:
Part of splitting #4969 into one PR per SQLite C-API.

Adds `SqliteConnection::on_commit` and `SqliteConnection::remove_commit_hook`, wrapping [`sqlite3_commit_hook`](https://www.sqlite.org/c3ref/commit_hook.html). The callback runs just before a transaction commits and returns `bool`: `false` lets the commit proceed, `true` converts it into a rollback. Only one commit hook is active at a time, and re-registering replaces the previous one. The callback cannot use the connection, and a panic in it aborts the process (matching the existing collation callback behavior).

```rust
conn.on_commit(|| {
    println!("about to commit");
    false // allow the commit; return true to convert it into a rollback
});
```

Adding the hook field makes `RawConnection` no longer `UnwindSafe`, so this introduces a private `RawConnection::from_ptr` constructor and wraps the borrowed connection in `AssertUnwindSafe` inside `run_custom_function` (which never registers a hook). The remaining hook PRs in the split reuse this same plumbing.

### DIFF
--- a/diesel/src/sqlite/connection/hooks.rs
+++ b/diesel/src/sqlite/connection/hooks.rs
@@ -1,0 +1,202 @@
+use super::SqliteConnection;
+
+pub(super) use super::CommitDecision;
+
+impl SqliteConnection {
+    /// Registers a callback invoked when a transaction is about to be
+    /// committed.
+    ///
+    /// The callback returns a [`CommitDecision`]: `Proceed` lets the commit
+    /// complete, `Rollback` converts it into a rollback.
+    ///
+    /// Only one commit hook can be active at a time per connection.
+    /// Registering a new one replaces the previous.
+    ///
+    /// The callback must not use the database connection. Panics in the
+    /// callback abort the process.
+    ///
+    /// See: [`sqlite3_commit_hook`](https://www.sqlite.org/c3ref/commit_hook.html)
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use diesel::prelude::*;
+    /// use diesel::sqlite::{SqliteConnection, CommitDecision};
+    /// use std::sync::{Arc, Mutex};
+    ///
+    /// diesel::table! {
+    ///     users (id) {
+    ///         id -> Integer,
+    ///         name -> Text,
+    ///     }
+    /// }
+    ///
+    /// # use diesel::connection::SimpleConnection;
+    /// # let conn = &mut SqliteConnection::establish(":memory:").unwrap();
+    /// # conn.batch_execute("CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL)").unwrap();
+    /// let commits = Arc::new(Mutex::new(0u32));
+    /// let commits2 = commits.clone();
+    ///
+    /// conn.on_commit(move || {
+    ///     *commits2.lock().unwrap() += 1;
+    ///     CommitDecision::Proceed
+    /// });
+    ///
+    /// conn.immediate_transaction(|conn| {
+    ///     diesel::insert_into(users::table)
+    ///         .values(users::name.eq("Alice"))
+    ///         .execute(conn)?;
+    ///     Ok::<_, diesel::result::Error>(())
+    /// }).unwrap();
+    ///
+    /// assert_eq!(*commits.lock().unwrap(), 1);
+    /// ```
+    pub fn on_commit<F>(&mut self, hook: F)
+    where
+        F: FnMut() -> CommitDecision + Send + 'static,
+    {
+        self.raw_connection.set_commit_hook(hook);
+    }
+
+    /// Removes the commit hook. Subsequent commits will not invoke any
+    /// callback.
+    ///
+    /// See [`on_commit`](Self::on_commit) for usage example.
+    pub fn remove_commit_hook(&mut self) {
+        self.raw_connection.remove_commit_hook();
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::connection::Connection;
+    use crate::query_dsl::RunQueryDsl;
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicU32, Ordering};
+
+    fn connection() -> SqliteConnection {
+        SqliteConnection::establish(":memory:").unwrap()
+    }
+
+    #[derive(crate::QueryableByName)]
+    struct CountResult {
+        #[diesel(sql_type = crate::sql_types::BigInt)]
+        c: i64,
+    }
+
+    #[diesel_test_helper::test]
+    fn on_commit_fires_on_commit() {
+        let conn = &mut connection();
+
+        let count = Arc::new(AtomicU32::new(0));
+        let c2 = count.clone();
+
+        conn.on_commit(move || {
+            c2.fetch_add(1, Ordering::Relaxed);
+            CommitDecision::Proceed
+        });
+
+        conn.immediate_transaction(|conn| {
+            crate::sql_query("CREATE TABLE t1 (id INTEGER PRIMARY KEY)")
+                .execute(conn)
+                .unwrap();
+            Ok::<_, crate::result::Error>(())
+        })
+        .unwrap();
+
+        assert_eq!(count.load(Ordering::Relaxed), 1);
+    }
+
+    #[diesel_test_helper::test]
+    fn on_commit_returning_true_forces_rollback() {
+        let conn = &mut connection();
+
+        crate::sql_query("CREATE TABLE t_commit (id INTEGER PRIMARY KEY)")
+            .execute(conn)
+            .unwrap();
+
+        conn.on_commit(|| CommitDecision::Rollback);
+
+        // The transaction will attempt to commit, but the hook will convert
+        // it to a rollback. diesel's AnsiTransactionManager will see the
+        // failure from the COMMIT statement (sqlite returns an error when
+        // the commit hook returns non-zero and the commit is aborted).
+        let result = conn.immediate_transaction(|conn| {
+            crate::sql_query("INSERT INTO t_commit (id) VALUES (1)")
+                .execute(conn)
+                .unwrap();
+            Ok::<_, crate::result::Error>(())
+        });
+
+        // The transaction should have been rolled back.
+        assert!(result.is_err());
+
+        // Remove the hook so subsequent queries don't fail.
+        conn.remove_commit_hook();
+
+        // Verify the row was not persisted.
+        let cnt: i64 = crate::sql_query("SELECT COUNT(*) as c FROM t_commit")
+            .get_result::<CountResult>(conn)
+            .unwrap()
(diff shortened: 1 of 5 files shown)


diesel/src/sqlite/connection/mod.rs
```diff
@@ -678,6 +678,70 @@ impl SqliteConnection {
...
+    /// # use diesel::connection::SimpleConnection;
+    /// # let conn = &mut SqliteConnection::establish(":memory:").unwrap();
+    /// # conn.batch_execute("CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL)").unwrap();
+    /// let commits = Arc::new(Mutex::new(0u32));
+    /// let commits2 = commits.clone();
+    ///
+    /// conn.on_commit(move || {
+    ///     *commits2.lock().unwrap() += 1;
+    ///     false // allow the commit to proceed
+    /// });
```
Mingun: Probably, it is better to use
```rust
enum OnCommitResult {
  Continue,
  Rollback,
}
```
so you don't forgot that the boolean mean.

LucaCappelletti94: Reasonable, done.

LucaCappelletti94:
Rolling dice to see whether flakiness reappears after merger

diesel/src/sqlite/connection/mod.rs
```diff
@@ -3218,4 +3289,128 @@ mod tests {
...
+        #[diesel(sql_type = crate::sql_types::BigInt)]
+        c: i64,
+    }
+
+    #[diesel_test_helper::test]
+    fn on_commit_fires_on_commit() {
+        use std::sync::{Arc, Mutex};
+        let conn = &mut connection();
+
+        let count = Arc::new(Mutex::new(0u32));
```
weiznich: This can use a `AtomicU32` instead of a mutex

Given that this is a `FnMut` a plain `u32` could be sufficient.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -678,6 +688,69 @@ impl SqliteConnection {
         self.raw_connection.deserialize(&self.serialized_data)
     }
 
+    /// Registers a callback invoked when a transaction is about to be
```
weiznich: I would prefer moving all the hook function in a `hooks.rs` file, containing the relevant tests, types and function definitions (we can just have another `impl SqliteConnection` block there). That keeps this file smaller

diesel/src/sqlite/connection/mod.rs
```diff
@@ -3218,4 +3289,128 @@ mod tests {
...
+        assert_eq!(cnt, 0);
+    }
+
+    #[diesel_test_helper::test]
+    fn replacing_commit_hook_drops_old() {
+        use std::sync::{Arc, Mutex};
+        let conn = &mut connection();
+
+        let old_count = Arc::new(Mutex::new(0u32));
+        let new_count = Arc::new(Mutex::new(0u32));
```
weiznich: Again use `AtomicU32` instead

diesel/src/sqlite/connection/raw.rs
```diff
@@ -50,9 +51,20 @@ macro_rules! assert_fail {
 #[allow(missing_debug_implementations, missing_copy_implementations)]
 pub(super) struct RawConnection {
     pub(super) internal_connection: NonNull<ffi::sqlite3>,
+    /// Type-erased boxed closure kept alive while the commit hook is registered.
+    commit_hook: Option<Box<dyn Any + Send>>,
```
weiznich: No need for so drastic type erasing here, we just could use a `Box<dyn FnMut() -> CommitDecision + Send` here, that also will erase the type.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -374,12 +387,42 @@ impl RawConnection {
...
+        F: FnMut() -> super::CommitDecision + Send + 'static,
+    {
+        let mut boxed: Box<F> = Box::new(hook);
+        let ptr = &raw mut *boxed as *mut libc::c_void;
+        unsafe {
+            ffi::sqlite3_commit_hook(
+                self.internal_connection.as_ptr(),
+                Some(commit_hook_trampoline::<F>),
+                ptr,
+            );
```
weiznich, line 412: Needs a safety comment

diesel/src/sqlite/connection/raw.rs
```diff
@@ -374,12 +387,42 @@ impl RawConnection {
...
+        }
+        // The old Box (if any) is dropped here after SQLite has already
+        // switched to the new callback, preventing use-after-free.
+        self.commit_hook = Some(boxed);
+    }
+
+    /// Removes the commit hook.
+    pub(super) fn remove_commit_hook(&mut self) {
+        unsafe {
+            ffi::sqlite3_commit_hook(self.internal_connection.as_ptr(), None, ptr::null_mut());
```
weiznich, line 431: Needs a safety comment

weiznich (changes requested):
Thanks for splitting up the large PR. This is much easier to handle.

I have some minor suggestions mostly about where to place stuff, otherwise this looks ready to be merged.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -47,12 +48,30 @@ macro_rules! assert_fail {
...
+/// single-sized struct so it can be safely passed across the FFI boundary.
+struct CommitHookBox {
+    hook: Box<dyn FnMut() -> CommitDecision + Send>,
+}
+
 #[allow(missing_debug_implementations, missing_copy_implementations)]
 pub(super) struct RawConnection {
     pub(super) internal_connection: NonNull<ffi::sqlite3>,
+    /// Boxed closure kept alive while the commit hook is registered.
+    commit_hook: Option<Box<CommitHookBox>>,
```
Mingun: Now you have two levels of indirection (compared to the first implementation): `Box` in the `Box`. It seems that this is not oversight, because two boxes is obvious in `set_commit_hook`. Why two boxes is needed? Is that means that your first implementation was incorrect in some cases?

LucaCappelletti94: Copy pasted the Box, I am a bit in medias res. Will ping when it is back to reviewable state, sorry for the inconvenience.

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: `RawConnection` now stores `Option<Box<dyn FnMut() -> CommitDecision + Send>>` directly, so the closure type is erased by the trait object itself and there is no `dyn Any` round trip.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Moved

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Switched to `Arc<AtomicU32>` with `fetch_add`/`load`. A plain `u32` captured by the closure would not work here because the test reads the counter after the hook has run, so it needs shared access rather than the closure owning it.

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/hooks.rs
```diff
@@ -0,0 +1,202 @@
...
+    ///         .values(users::name.eq("Alice"))
+    ///         .execute(conn)?;
+    ///     Ok::<_, diesel::result::Error>(())
+    /// }).unwrap();
+    ///
+    /// assert_eq!(*commits.lock().unwrap(), 1);
+    /// ```
+    pub fn on_commit<F>(&mut self, hook: F)
+    where
+        F: FnMut() -> CommitDecision + Send + 'static,
```
weiznich, line 56: I tried to double check the SQLite documentation but I was not able to understand whether or not the provided callback can be called by several threads at once. I would guess no, given that SQLite only allows one writer at the time, but I rather would be sure about that. If it's not guaranteed anywhere we might need to add a `Sync` bound?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -470,6 +529,9 @@ extern "C" fn run_custom_function<F, Ret, RetSqlType>(
     // We need this to move the reference into the catch_unwind part
     // this is sound as `F` itself and the stored string is `UnwindSafe`
     let callback = core::panic::AssertUnwindSafe(&mut data_ptr.callback);
+    // conn holds a `Box<dyn Any + Send>` hook field which is not UnwindSafe.
```
weiznich: This comment seems to be not up to date anymore

weiznich (approved):
One minor note about a comment, and one question about the guarentees given but SQLite, otherwise this is now ready to be merged.

Thanks again for working on this :heart:

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Updated

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: No `Sync` needed. SQLite invokes the commit hook synchronously inside `sqlite3_step()` on the calling thread, never from another thread, and on our side every commit  goes through `&mut SqliteConnection`, so the hook only runs while the caller holds an exclusive borrow of the connection that owns the closure. It can therefore never  be called concurrently. `Send` is required so the connection (which is `Send`, `!Sync`) stays `Send`, but `Sync` would only matter for concurrent shared access that cannot  happen here, so adding it would just reject valid callbacks like `Rc`-capturing ones.

weiznich: Sounds good, thanks for providing these details :+1:

Mingun: So, don't you think it would be helpful to add this explanation to the code? Maybe even to the doc comment so the users known this nuance (as I understand, that is documented SQLite behavior, and not an implementation detail of diesel).

LucaCappelletti94: Ok, let's do this as well, it is reasonable.

PR #5080, opened 2026-06-19
Bound apt-get lock wait in CI

LucaCappelletti94:
Linux CI jobs can hang for the full 6-hour runner limit when `apt-get` blocks waiting for the dpkg/apt lock held by the runner's background `unattended-upgrades` (seen on a recent #5077 run, where the MSRV job's dependency install sat silent for 6 hours before being cancelled). Setting `DPkg::Lock::Timeout=300` on every `apt-get` call bounds that wait to 5 minutes, so each call either acquires the lock and proceeds or fails fast instead of hanging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,8 +92,8 @@ jobs:
       - name: Install postgres (Linux)
         if: runner.os == 'Linux' && matrix.backend == 'postgres'
         run: |
-          sudo apt-get update
-          sudo apt-get install -y libpq-dev postgresql
+          sudo apt-get -o DPkg::Lock::Timeout=300 update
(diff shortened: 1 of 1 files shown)


weiznich (approved):

weiznich:
Thanks :heart:

You seem to have a talent for hitting all the wired CI edge-cases :see_no_evil:

LucaCappelletti94:
My friends who hunt bugs for a living tell me I have the brown touch. I have recently found an [NVIDIA compiler (or maybe architecture) bug](https://github.com/LucaCappelletti94/ptxas-uniform-counter-miscompile) just by trying to write a CUDA kernel to do mass spectrometry at scale in GPU for example. If things continue this way, [I may start praying to the Omnissiah](https://www.youtube.com/shorts/3BXz_Qir2Og).

PR #5082, opened 2026-06-19
Rework `#[derive(Enum)]`

weiznich:
This change aims to make `#[derive(Enum)]` much more usable for all the different ways to map a Rust enum to a database field. Instead of directly generating (de)serialization code for native database backends, we now rely on the SQL type to specify how enums should be mapped for a given SQL type.

We provide three differnt strategies for this mapping for various SQL types:

* `IntMapping`, which maps the discriminant values of the enum variants to integers at database level. This is something that works for all database backend. We support this mapping for the various signed and unsigned integer types. We require here that the users annotates all variants with explicit discriminant values to make their code a bit more robust against accidently changes those discriminants.
* `StringMapping`, which maps the name of the variant as text at database level. This is also supported for all backends
* `EnumTypeMapping`, which directly maps the variants to matching database level enum variants. This is only supported on the Mysql and PostgreSQL backend and closely mirrors the "old" implementation

The mapping is determined only based on the sql type provided to `#[derive(Enum)]`. We implement support for this mapping for the built-in integer and text types. For enum types we extend `#[derive(SqlType)]` to have an additional `#[diesel(enum_type)]` attribute to generate the right code for us.

Internally the derive now translates the enum into a runtime accessible structur and passes these information at runtime down to the selected strategy. This gives all implementations the necssary flexibility, without having the need to find out which strategy is used in the proc macro itself. The strategy is only selected later at type system level, which should make this much more robust and extensible.

We also add a `#[diesel(rename_all = "case")] enum level attribute to `#[derive(Enum)]`, which allows to rename all variants at database level in the specified case variant. This closely mirrors `#[serde(rename_all)]`.

Additionally we add a `#[diesel(rename = "newName")]` variant level attribute to only rename a single variant to a specific name. Again this closely mirrors `#[serde(rename)]`.

Finally we add tests, compile tests and documentation for almost all of the added functionality.

---

I opened this a draft PR for now as I still need to figure out how to implement the one remaining TODO item in the multiconnection implementation of `IntMapping`. I currently run into borrow errors there as the temporary value doesn't live long enough :(


diesel_derives/tests/enum_.rs
```diff
@@ -1,17 +1,25 @@
...
+use diesel::query_builder::{Query, QueryFragment, QueryId};
+use diesel::serialize::ToSql;
 use diesel_derives::Enum;
 
 #[derive(Debug, Clone, Copy, Enum, PartialEq)]
 #[cfg_attr(feature = "postgres", diesel(sql_type = schema::sql_types::Color))]
 #[cfg_attr(feature = "mysql", diesel(sql_type = schema::sql_types::CarsPaintColorEnum))]
+#[diesel(sql_type = diesel::sql_types::Text)]
+#[diesel(sql_type = diesel::sql_types::Integer)]
 enum Color {
```
pksunkara, line 16: I am confused. Where is the `derive(SqlType)` here?

diesel_derives/tests/schema.rs
```diff
@@ -20,11 +20,13 @@ pub mod sql_types {
     #[cfg(feature = "postgres")]
     #[derive(diesel::query_builder::QueryId, Clone, diesel::sql_types::SqlType)]
     #[diesel(postgres_type(name = "color"))]
+    #[diesel(enum_type)]
```
pksunkara, line 23: I wanted us to automatically infer this based on the sql type without needing the user to specify it. (I think we are only doing this `enum`s right?)

But I am quite confused about how the sql types are defined. I see 2 ways currently:

1. `diesel(postgres_type(name = "Color"))`
2. `diesel(sql_type = Color)`

weiznich: So there are always 2 types involved. 

The "SQL Type" that derives `#[derive(SqlType)]` and is a zero sized marker type that cannot hold any value. This is used by the compile time checking that verifies that queries are valid and the mapping is meaningful. These types need an annotation like `#[diesel(postgres_type(name = "Color")]` to specify which type they represent at database level.

The "Rust Type" that derives `#[derive(Enum)]` in this case. That's the type that is actually used in the rust code. These types need at least one `#[diesel(sql_type = Color)]` attribute to specify which sql type they are compatible with.

In general this split exists as there is a n:m mapping between SQL types and Rust types. E.g. You could map `Timestamp` to several different time types from `time` and `chrono`. You also can map e.g. a `String` to multiple database types like an enum or a `TEXT` field.

Now `#[derive(Enum)]` covers 3 use-cases:

* Mapping to Integer columns -> The sql side type is already defined in Diesel itself, we only need to take care of the conversion, there is no need for defining another sql type here, we just use `#[diesel(sql_type = Integer)]` for the enum derive
* Mapping to Text columns -> The sql side type is already defined in Diesel itself, we only need to take care of the conversion, there is no need to define another sql type here, we just use `#[diesel(sql_type = Text)]` for the enum derive
* Mapping to a Database level enum -> These are conceptually different types at SQL level, therefore we need to have a distinct SQL type for them, so we have a zero sized struct that derives `SqlType`. This is currently already generated by `diesel print-schema`, although it is still missing the `#[diesel(enum_type)]` attribute. This SQL type is then referenced by the `#[derive(Enum)]` `#[diesel(sql_type = …)]` attribute. That's the case for the `Color` example above. 

For the database side enum type case, the end goal is to generate both the complete SQL type (including `#[diesel(enum_type)]` and the Rust side enum via `diesel print-schema`, so that users don't need to care about these details in most cases. I decided to not put the CLI part in this PR as well, as it is already quite large, but I plan to have a followup PR for this soon after this one is merged.

Hopefully that clears up how this is supposed do work.

pksunkara (approved):

diesel/src/types/enum_.rs
```diff
@@ -0,0 +1,343 @@
...
+// than that one of i128 and also the alignment of any
+// of these types is smaller or equal to i128
+#[allow(unsafe_code)]
+unsafe impl IntegerMappingHelper for i64 {}
+#[allow(unsafe_code)]
+unsafe impl IntegerMappingHelper for i32 {}
+#[allow(unsafe_code)]
+unsafe impl IntegerMappingHelper for i16 {}
+#[allow(unsafe_code)]
+unsafe impl IntegerMappingHelper for i8 {}
```
weiznich, line 261: cc @diesel-rs/reviewers I'm not that happy with this hack, as it requires quite a bit unsafe code. This is the best option I came up to support mapping enums to integers on `#[derive(MultiConnection)]` connections. We need to have a borrowed value there derived from the discriminate, so we cannot just do an in place conversion and use the owned value.

I just want to verify with you that this code is sane. It is running under miri and passing tests there for little and big endian.

pksunkara: Looks okay to me. But if I am going to be honest, I would rather us not have int mapping at all in the first place. String/Enum mapping would solve most of the use cases, right?

weiznich: For new code `String`/`Enum` mapping would solve most use-cases, as the actual representation is mostly not that relevant. For existing code that's a lot harder. For example [active record](https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html) seems to use an integer mapping internally.

pksunkara: Makes sense. Feel free to merge whenever you are ready unless you want others to review the unsafe code.

PR #5083, opened 2026-06-19
Add `SqliteConnection::on_rollback` wrapping `sqlite3_rollback_hook`

LucaCappelletti94:
Part of splitting #4969 into one PR per SQLite C-API, following the commit hook in #5077.

Adds `SqliteConnection::on_rollback` and `SqliteConnection::remove_rollback_hook`, wrapping [`sqlite3_rollback_hook`](https://www.sqlite.org/c3ref/commit_hook.html). The callback fires after a transaction is rolled back. Only one rollback hook is active at a time, and re-registering replaces the previous one. The callback cannot use the connection, and a panic in it aborts the process.

```rust
conn.on_rollback(|| {
    println!("transaction rolled back");
});
```

### DIFF
--- a/diesel/src/sqlite/connection/hooks.rs
+++ b/diesel/src/sqlite/connection/hooks.rs
@@ -65,6 +65,62 @@ impl SqliteConnection {
     pub fn remove_commit_hook(&mut self) {
         self.raw_connection.remove_commit_hook();
     }
+
+    /// Registers a callback invoked after a transaction is rolled back.
+    ///
+    /// This is **not** invoked for the implicit rollback that occurs when
+    /// the connection is closed. It **is** invoked when a commit hook forces
+    /// a rollback by returning [`CommitDecision::Rollback`].
+    ///
+    /// Only one rollback hook can be active at a time per connection.
+    /// Registering a new one replaces the previous.
+    ///
+    /// The callback must not use the database connection. It is invoked
+    /// synchronously on the thread driving the connection, so it is never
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/connection/hooks.rs
```diff
@@ -65,6 +65,61 @@ impl SqliteConnection {
...
+    ///
+    /// ```rust
+    /// use diesel::prelude::*;
+    /// use diesel::sqlite::SqliteConnection;
+    /// use std::sync::{Arc, Mutex};
+    ///
+    /// # use diesel::connection::SimpleConnection;
+    /// # let conn = &mut SqliteConnection::establish(":memory:").unwrap();
+    /// # conn.batch_execute("CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT NOT NULL)").unwrap();
+    /// let rollbacks = Arc::new(Mutex::new(0u32));
```
weiznich: That should use `Arc::new(AtomicU32::new(0))` instead

weiznich (approved):
Looks good beside that one usage of `Arc<Mutex<u32>>`

LucaCappelletti94:
> Looks good beside that one usage of `Arc<Mutex<u32>>`

Will atomize that shortly

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: Done

weiznich (approved):

PR #5085, opened 2026-06-21
Add SqliteConnection::on_wal wrapping sqlite3_wal_hook

LucaCappelletti94:
Part of splitting #4969 into one PR per SQLite C-API, following the commit hook in #5077. It is independent of the parallel rollback hook in #5083, so the two can merge in either order.

Adds `SqliteConnection::on_wal` and `SqliteConnection::remove_wal_hook`, wrapping [`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html). The callback fires after each commit of a transaction in WAL mode and receives the database name and the current WAL page count. Only one WAL hook is active at a time, and re-registering replaces the previous one. The callback cannot use the connection, and a panic in it aborts the process.

```rust
conn.on_wal(|db_name, n_pages| {
    println!("WAL for {db_name}: {n_pages} pages");
});
```

### DIFF
--- a/diesel/src/sqlite/connection/hooks.rs
+++ b/diesel/src/sqlite/connection/hooks.rs
@@ -186,6 +186,55 @@ impl SqliteConnection {
     pub fn remove_progress_handler(&mut self) {
         self.raw_connection.remove_progress_handler();
     }
+
+    /// Registers a callback invoked after each commit of a transaction in
+    /// [WAL mode](https://www.sqlite.org/wal.html). It receives a borrowed
+    /// connection, the database name (`"main"`, `"temp"`, or an `ATTACH`
+    /// alias), and the current WAL page count.
+    ///
+    /// The hook fires after the commit completes and the write-lock is
+    /// released, so the callback may read, write, or
+    /// [checkpoint](https://www.sqlite.org/wal.html#ckpt) through `conn`,
+    /// provided it leaves no open transaction. A write that commits inside the
+    /// callback re-fires the hook re-entrantly, so a callback that writes
+    /// unconditionally will recurse until the stack overflows. Guard against
+    /// that yourself if the callback writes.
+    ///
+    /// Only one WAL hook is active at a time, and re-registering replaces it.
+    /// `PRAGMA wal_autocheckpoint` installs its own WAL hook and overwrites
+    /// this one. A panic in the callback aborts the process.
+    ///
+    /// See: [`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html)
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// # use diesel::prelude::*;
+    /// # use diesel::connection::SimpleConnection;
+    /// # use diesel::sqlite::SqliteConnection;
+    /// # let conn = &mut SqliteConnection::establish("test.db").unwrap();
+    /// # conn.batch_execute("PRAGMA journal_mode = WAL;").unwrap();
+    /// conn.on_wal(|conn, db_name, n_pages| {
+    ///     println!("WAL for {db_name}: {n_pages} pages");
+    ///     if n_pages > 1000 {
+    ///         // The connection may be used here, e.g. to force a checkpoint.
+    ///         let _ = conn.batch_execute("PRAGMA wal_checkpoint(TRUNCATE);");
+    ///     }
+    /// });
+    /// ```
+    pub fn on_wal<F>(&mut self, hook: F)
+    where
+        F: Fn(&mut SqliteConnection, &str, u32) + Send + 'static,
+    {
+        self.raw_connection.set_wal_hook(hook);
+    }
+
+    /// Removes the WAL hook. Subsequent commits will not invoke any callback.
+    ///
+    /// See [`on_wal`](Self::on_wal) for usage example.
+    pub fn remove_wal_hook(&mut self) {
+        self.raw_connection.remove_wal_hook();
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/connection/raw.rs
```diff
@@ -833,3 +887,34 @@ where
...
 }
+
+/// C trampoline for `sqlite3_wal_hook`.
+///
+/// # Safety
+///
+/// `user_data` must point to a live `F` stored in `RawConnection::wal_hook`.
+unsafe extern "C" fn wal_hook_trampoline<F>(
+    user_data: *mut libc::c_void,
+    _db: *mut ffi::sqlite3,
```
weiznich: Is there a reason not to expose the database connection to the callback? For me it looks like having access to the connection might be very important.

The documentation on SQLite's side also states:

> The callback is invoked by SQLite after the commit has taken place and the associated write-lock on the database released, so the implementation may read, write or [checkpoint](https://www.sqlite.org/wal.html#ckpt) the database as required.

So it seems to be totally fine to just interact with the database inside of the callback?

diesel/src/sqlite/connection/raw.rs
```diff
@@ -833,3 +887,34 @@ where
...
+    F: FnMut(&str, i32),
+{
+    let result = crate::util::std_compat::catch_unwind(core::panic::AssertUnwindSafe(|| {
+        // SAFETY: `user_data` points to a live `F` in `RawConnection::wal_hook`.
+        let f = unsafe { &mut *(user_data as *mut F) };
+        // SAFETY: `db_name` is a valid C string provided by SQLite. Schema
+        // names are UTF-8, but use a lossy conversion to avoid aborting the
+        // process on a pathological name.
+        let db_name = unsafe { CStr::from_ptr(db_name) }.to_string_lossy();
+        f(&db_name, n_pages);
```
weiznich: I wonder whether it's more meaningful to convert `n_pages` to a `u32` as I would expect that this number is always positive?

diesel/src/sqlite/connection/hooks.rs
```diff
@@ -65,6 +65,55 @@ impl SqliteConnection {
...
+    /// - `db_name`: the database name (`"main"`, `"temp"`, or an `ATTACH` alias).
+    /// - `n_pages`: the number of pages currently in the WAL file.
+    ///
+    /// Useful for triggering custom checkpoint logic via
+    /// [`PRAGMA wal_checkpoint`](https://www.sqlite.org/pragma.html#pragma_wal_checkpoint).
+    ///
+    /// Only one WAL hook can be active at a time per connection. Registering a
+    /// new one replaces the previous.
+    ///
+    /// The callback must not use the database connection. It is invoked
```
weiznich, line 83: The documentation on sqlite's side indicate that this is not the case. It states:

> The callback is invoked by SQLite after the commit has taken place and the associated write-lock on the database released, so the implementation may read, write or [checkpoint](https://www.sqlite.org/wal.html#ckpt) the database as required.

(Seems like a copy paste thing?)

diesel/src/sqlite/connection/hooks.rs
```diff
@@ -199,4 +248,166 @@ mod tests {
...
+
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+    #[diesel_test_helper::test]
+    fn on_wal_fires_in_wal_mode() {
+        let (conn, _dir) = &mut wal_connection();
+
+        // Enable WAL mode.
+        crate::sql_query("PRAGMA journal_mode=WAL")
+            .execute(conn)
+            .unwrap();
```
weiznich: Maybe move that into `wal_connection()` as well? We always want to call it for such a connection

diesel/src/sqlite/connection/raw.rs
```diff
@@ -833,3 +887,34 @@ where
...
+where
+    F: FnMut(&str, i32),
+{
+    let result = crate::util::std_compat::catch_unwind(core::panic::AssertUnwindSafe(|| {
+        // SAFETY: `user_data` points to a live `F` in `RawConnection::wal_hook`.
+        let f = unsafe { &mut *(user_data as *mut F) };
+        // SAFETY: `db_name` is a valid C string provided by SQLite. Schema
+        // names are UTF-8, but use a lossy conversion to avoid aborting the
+        // process on a pathological name.
+        let db_name = unsafe { CStr::from_ptr(db_name) }.to_string_lossy();
```
weiznich: It might still be worth to check if `db_name` is null and return an empty string in this case. It's not expensive, so better be defensive here.

weiznich (changes requested):
I would like to expose access to the underlying connection as well as the documentation indicates that this is safe.

We also might want to change the size parameter to `u32` even if sqlite passes it as signed integer. It doesn't seam meaningful to have negative values there.

diesel/src/sqlite/connection/raw.rs
```diff
@@ -54,8 +56,12 @@ pub(super) struct RawConnection {
...
     /// Boxed closure kept alive while the commit hook is registered.
     commit_hook: Option<Box<dyn FnMut() -> CommitDecision + Send>>,
-    /// Boxed closure kept alive while the WAL hook is registered.
-    wal_hook: Option<Box<dyn FnMut(&str, i32) + Send>>,
+    /// Boxed closure kept alive while the rollback hook is registered.
+    rollback_hook: Option<Box<dyn FnMut() + Send>>,
+    /// Boxed closure kept alive while the WAL hook is registered. The `RefCell`
+    /// lets the trampoline take a shared `&RefCell<F>`, so a re-entrant call is
+    /// skipped via `try_borrow_mut` instead of aliasing the closure.
+    wal_hook: Option<Box<RefCell<dyn FnMut(&mut SqliteConnection, &str, u32) + Send>>>,
```
weiznich: Is there any reason to go with this `RefCell` approach instead of just turning this into a `dyn Fn(…)` and let the caller handle this if necessary?

LucaCappelletti94: With `FnMut` the caller can move an owned, possibly non-`Sync` resource into the closure and call its `&mut self` methods. With `Fn` each such piece of state has to be wrapped in a `Cell`, `RefCell`, or `Mutex` by the caller, and the matching `.borrow_mut()`/`.get()`/`.lock()` boilerplate, so I thought better figuring out the general case so less difficult code needs to exist downhill.

Which would you prefer?

weiznich: I think I would prefer the `Fn` variant, as this removes a bit complexity from the implementation and enables the use-case of have re-entrant calls (although I'm not sure if anyone has a real use-case for this).

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: The callback now receives `&mut SqliteConnection`. Simply most of these are not compatible with using the connection, and I have honestly forgotten about this one being a-ok with it.

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: I went for `u32::try_from(n_pages).unwrap_or(0)`.

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: Copy-paste from the commit hook, fixed.

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: `wal_connection()` now runs `PRAGMA journal_mode=WAL` itself, so the tests no longer repeat it.

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Switched to `Fn(&mut SqliteConnection, &str, u32) + Send`.

weiznich (approved):
Looks good now, thanks for fixing it up.

weiznich (approved):

LucaCappelletti94:
Can we attempt again the merge?

weiznich:
I will try it again tomorrow, I just need some time to look over the code again. Not that I expect that there is something new in there, but each push can include new unwanted changes

weiznich (approved):

PR #5088, opened 2026-06-22
Add `SqliteConnection::on_progress` wrapping `sqlite3_progress_handler`

LucaCappelletti94:
Part of splitting #4969 into one PR per SQLite C-API, following the commit hook in #5077. It is independent of the parallel rollback (#5083) and WAL (#5085) hooks, so they can merge in any order.

Adds `SqliteConnection::on_progress` and `SqliteConnection::remove_progress_handler`, wrapping [`sqlite3_progress_handler`](https://www.sqlite.org/c3ref/progress_handler.html). The callback fires roughly every `n` virtual-machine instructions while a query runs and returns a `ProgressDecision` (`Continue` or `Interrupt`), mirroring the `CommitDecision` enum from #5077. Only one handler is active at a time, and re-registering replaces the previous one. The callback cannot use the connection, and a panic in it aborts the process.

```rust
// fire every ~1000 VM instructions; return Interrupt to abort the query
conn.on_progress(1000, || ProgressDecision::Continue);
```

### DIFF
--- a/diesel/src/sqlite/connection/hooks.rs
+++ b/diesel/src/sqlite/connection/hooks.rs
@@ -1,6 +1,7 @@
 use super::SqliteConnection;
+use core::num::NonZeroU32;
 
-pub(super) use super::CommitDecision;
+pub(super) use super::{CommitDecision, ProgressDecision};
 
 impl SqliteConnection {
     /// Registers a callback invoked when a transaction is about to be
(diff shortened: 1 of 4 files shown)


diesel/src/sqlite/connection/hooks.rs
```diff
@@ -65,6 +65,63 @@ impl SqliteConnection {
...
+    ///         ProgressDecision::Interrupt
+    ///     } else {
+    ///         ProgressDecision::Continue
+    ///     }
+    /// });
+    ///
+    /// // Later: remove the handler.
+    /// conn.remove_progress_handler();
+    /// ```
+    pub fn on_progress<F>(&mut self, n: i32, hook: F)
```
weiznich: Would it be meaningful to use a `NonZeroI32` here to require removing the handler to explicitly call `remove_progress_handler`?

weiznich (approved):
Looks good beside the one note about the type of the statement count

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: Used `NonZeroU32` instead of `NonZeroI32` as SQLite also disables on negatives, so `NonZeroU32` rules those out too and leaves `remove_progress_handler` as the only way to disable.

weiznich (approved):

weiznich:
Thanks, that looks good now.

PR #5089, opened 2026-06-23
Add SqliteConnection::on_busy and set_busy_timeout wrapping the sqlite3 busy APIs

LucaCappelletti94:
Part of splitting #4969 into one PR per SQLite C-API, following the commit hook in #5077. It is independent of the parallel rollback (#5083), WAL (#5085), and progress (#5088) hooks, so they can merge in any order.

Adds `SqliteConnection::on_busy`, `remove_busy_handler`, and `set_busy_timeout`, wrapping [`sqlite3_busy_handler`](https://www.sqlite.org/c3ref/busy_handler.html) and [`sqlite3_busy_timeout`](https://www.sqlite.org/c3ref/busy_timeout.html). `on_busy` runs a custom callback on lock contention that receives the retry count and returns a `BusyDecision` (`Retry` or `GiveUp`), mirroring the `CommitDecision` enum from #5077, while `set_busy_timeout` installs the built-in retry-until-timeout handler. Setting either one clears the other. The callback cannot use the connection, and a panic in it aborts the process.

```rust
conn.on_busy(|retry_count| {
    if retry_count < 5 { BusyDecision::Retry } else { BusyDecision::GiveUp }
});

// or the simpler built-in handler:
conn.set_busy_timeout(5000);
```

The handler only fires under real cross-connection lock contention, which is a known source of flaky CI, so this is covered by doctests rather than a behavioral test.

### DIFF
--- a/diesel/src/sqlite/connection/raw.rs
+++ b/diesel/src/sqlite/connection/raw.rs
@@ -10,7 +10,7 @@ use super::functions::{build_sql_function_args, process_sql_function_result};
 use super::limits::SqliteLimit;
 use super::serialized_database::SerializedDatabase;
 use super::stmt::ensure_sqlite_ok;
-use super::{CommitDecision, ProgressDecision};
+use super::{BusyDecision, CommitDecision, ProgressDecision};
 use super::{Sqlite, SqliteAggregateFunction};
 use crate::deserialize::FromSqlRow;
 use crate::result::Error::DatabaseError;
@@ -62,6 +62,8 @@ pub(super) struct RawConnection {
     progress_hook: Option<Box<dyn FnMut() -> ProgressDecision + Send>>,
     /// Boxed closure kept alive while the WAL hook is registered.
     wal_hook: Option<Box<dyn Fn(&mut SqliteConnection, &str, u32) + Send>>,
+    /// Boxed closure kept alive while the busy handler is registered.
+    busy_handler: Option<Box<dyn FnMut(i32) -> BusyDecision + Send>>,
 }
 
 impl RawConnection {
@@ -74,6 +76,7 @@ impl RawConnection {
             rollback_hook: None,
             progress_hook: None,
             wal_hook: None,
+            busy_handler: None,
         }
     }
 
@@ -99,6 +102,7 @@ impl RawConnection {
                     rollback_hook: None,
                     progress_hook: None,
                     wal_hook: None,
+                    busy_handler: None,
                 })
             }
             err_code => {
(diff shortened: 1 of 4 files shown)


weiznich (changes requested):
The API looks good, but we should likely have some sort of test for this, at least something basic that verifies that the callback is called and has some effect?

LucaCappelletti94:
> The API looks good, but we should likely have some sort of test for this, at least something basic that verifies that the callback is called and has some effect?

Basically most tests I could think of were flaky, and other than test it "in practice", I could not really put something together that was consistent in causing this type of hook to fire. That being said, it was at least some time ago, so I can give it a try again.

weiznich:
I think combining [`SqliteConnection::immediate_transaction` ](https://docs.diesel.rs/main/diesel/sqlite/struct.SqliteConnection.html#method.immediate_transaction) + some barriers should result in a rather robust test. 

Basically what I'm thinking of is opening 2 connections to the same file, then use a barrier to start one immediate transaction before the other one, then use either a barrier (for a custom busy handler) or a timeout (for busy_timeout) inside of the first transaction to trigger the busy handler, finally check the result.

weiznich (approved):

diesel/src/sqlite/connection/raw.rs
```diff
@@ -602,6 +606,72 @@ impl RawConnection {
...
+    /// Only one busy handler can be active at a time. Setting this clears any
+    /// busy timeout previously set with `set_busy_timeout`.
+    ///
+    /// # Safety
+    ///
+    /// The `ptr` is derived from `&raw mut *boxed` and points to the heap
+    /// allocation of the closure. The `busy_handler_trampoline` function
+    /// matches the signature expected by `sqlite3_busy_handler`. The pointer
+    /// remains valid because we store `boxed` in `self.busy_handler` below,
+    /// keeping it alive for the lifetime of this `RawConnection`.
```
weiznich, line 621: This is currently not true. we never move the boxed closure to `self.busy_handler`, therefore this is currently not sound.

weiznich (changes requested):

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: The move does happen: `set_busy_handler` ends with `self.busy_handler = Some(boxed);` as its last statement, right after the `sqlite3_busy_handler` call, so the boxed closure is kept in the field and its heap allocation stays at the address handed to SQLite. It is the same box-then-register-then-store shape as `set_commit_hook`.

I will still reword the comment to say the box is held until the handler is replaced or removed rather than "for the lifetime", as it is a bit imprecise. Did I miss anything?

weiznich: No that's fine. For whatever reason I missed that line.

weiznich (approved):

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Heatwave? Me I am almost fully fried.

weiznich: More likely doing too many things at the same time. The heat is thankfully gone here for a few days already

Issue #4759, opened 2025-09-12
Automatically infer enums via `print-schema` from the database


weiznich:
We would like to improve the way Diesel supports enums.

## Backends

- PostgreSQL
- MySQL

# Idea

Diesel ships a CLI tool that inspects the database and generates a `schema.rs` file based on that database. For any "custom" type that we encounter we already generate a custom type mapping via our `#[derive(SqlTye)]`. This includes enum types. We would like to extend this functionality to not only generate the SQL side type, but also a matching Rust enum definition if the user has opted into this behavior. This would be only supported for the following narrow use-cases:

- [Postgresql Enums](https://www.postgresql.org/docs/current/datatype-enum.html)
- [Mysql Enums](https://dev.mysql.com/doc/refman/8.4/en/enum.html)

For both variants we would like to generate the matching field less rust enum if the user enabled a corresponding opt-in setting for diesel-cli.

To make this happen we need to:

1. Implement a custom derive implementing `FromSql` and `ToSql` for the generated enum for this narrow use-case
2. Extend the schema inference code in [diesel-cli](https://github.com/diesel-rs/diesel/tree/master/diesel_cli/src/infer_schema_internals) to also include information about the fields of these enums
3. Generate the enums + the relevant derive as part of the [print-schema](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/print_schema.rs) process
4. Make sure to also handle this in [generate-migration](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/migrations/diff_schema.rs).

Each of these steps should be implemented as separate PR to keep the changes as small as possible. Each of the steps should have documentation and some tests to verify that everything works as expected.

demhadais:
I'd love to try to tackle this. Also I want to hijack this comment and just say that diesel is a phenomenal crate and the constant support, fast issue responses, and general maintenance does not go unnoticed. This is a shining example of good open-source software. Anyways, I'll be trying to show my appreciation by contributing to this crate as much as I can!

weiznich:
Thanks for the kind words :heart:
Just let me know if you need help to implement this feature. I'm happy to provide more pointers for the specific steps.

olorikendrick:
@weiznich I would love to tackle this if still available

weiznich:
[...]

PR #5093, opened 2026-06-26
Generate matching rust enums via Diesel-cli

weiznich:
This commit adds support for generating matching rust enums via diesel-cli. We now query the database to load information about the custom type. If the custom type is an enum we query all variants and generate a matching rust enum. 

This commit also adds the following options to diesel-cli:

* `generate_rust_enum_definitions` to enable/disable this behaviour. The new default for this setting is `true`, so enums are generated by default
* `custom_enum_derives` to give the user the ability to add custom additional derives to the generated enums. By default we only add those derives that are strictly necessary, but with this setting it's possible to add arbitary derives as well.

This commit also adds several tests and adjusts the output of existing tests.

Fixes most of #4759, the generate-migration part is still missing.


diesel_cli/src/infer_schema_internals/data_structures.rs
```diff
@@ -288,3 +293,20 @@ impl SupportedQueryRelationStructures {
...
+}
+
+impl EnumVariant {
+    pub fn rust_name(&self) -> String {
+        let n = if super::inference::is_reserved_name(&self.sql_name) {
+            Cow::Owned(format!("{}_", self.sql_name))
+        } else {
+            Cow::Borrowed(&self.sql_name)
+        };
+        n.to_pascal_case()
```
pksunkara, line 310: There's no way in our architecture to let users configure this, right? For ex: I might want the `Language` enum variants to be `DE`, `EN` etc..

diesel_cli/src/config.rs
```diff
@@ -484,13 +504,21 @@ pub struct PrintSchema {
     pub include_views: bool,
     #[serde(default)]
     pub experimental_infer_nullable_for_views: bool,
+    #[serde(default)]
+    pub custom_enum_derives: Option<Vec<String>>,
+    #[serde(default)]
+    pub generate_rust_enum_definitions: Option<bool>,
```
pksunkara, line 519: Why not use `bool` directly here? Since the default is `true` and not `null`.

I see the same issue for `generate_missing_sql_type_definitions`. Not sure what the conventions here are.

diesel_cli/src/config.rs
```diff
@@ -533,6 +560,20 @@ impl PrintSchema {
         }
     }
 
+    pub fn custom_rust_types_derives(&self) -> Vec<String> {
+        let mut derives = self
+            .custom_enum_derives
+            .as_ref()
+            .map_or(Vec::new(), |derives| derives.to_vec());
+        if !derives.iter().any(|item| item == "diesel::types::Enum") {
```
pksunkara: nit: Make a set and simply push them instead of checking?

pksunkara (approved):
No blockers, feel free to merge after addressing the comments.

diesel_cli/src/infer_schema_internals/data_structures.rs
weiznich: We do not have an option for that in the configuration or CLI flags. That written I think it's certainly possible to add this, although I would prefer keeping that as followup task for someone that needs this feature.

diesel_cli/src/config.rs
weiznich: The main reason is that setting default that is not literally `Default::default()` (so `false` for bools) with serde is requires rather a lot of code, as you need to define a function for that. It's not that this is a hard requirement, it's just that way currently. If you feel better with the other variant I will just change it.

pksunkara: Ah, serde defaults. Missed that.

PR #5095, opened 2026-06-26
Add `SqliteConnection::on_authorize` wrapping `sqlite3_set_authorizer`

LucaCappelletti94:
This adds `on_authorize` and `remove_authorizer` (and the `AuthorizerAction`, `AuthorizerContext`, and `AuthorizerDecision` types) to register an authorizer that allows, denies, or ignores SQL actions at compile time. It continues the per-API split of #4969.

The callback is `FnMut` and must not use the connection, like the commit and rollback hooks, so it needs no `RefCell`. Action codes map to a `#[non_exhaustive]` enum with an `Unknown(i32)` arm, and the trampoline maps null or non-UTF-8 arguments to `None` instead of aborting.

It also reaches into the shared statement cache: `remove_authorizer` clears diesel's prepared-statement cache so a statement compiled under an `IGNORE` authorizer does not keep its baked-in `NULL` after removal. That is the part I would appreciate a careful review of, since it is the only change outside the SQLite hook code.

### DIFF
--- a/diesel/src/connection/statement_cache/mod.rs
+++ b/diesel/src/connection/statement_cache/mod.rs
@@ -183,6 +183,15 @@ where
         }
     }
 
+    /// Removes all cached statements so subsequent queries are re-prepared,
+    /// while keeping the configured caching strategy.
+    // Currently used only by the SQLite authorizer, so it is compiled only for
+    // backends that provide that caller.
+    #[cfg(feature = "__sqlite-shared")]
+    pub(crate) fn clear(&mut self) {
+        self.cache.clear();
+    }
+
     /// Setting custom caching strategy. It is used in tests, to verify caching logic
     #[allow(dead_code)]
     pub(crate) fn set_strategy<Strategy>(&mut self, s: Strategy)
--- a/diesel/src/sqlite/connection/authorizer.rs
+++ b/diesel/src/sqlite/connection/authorizer.rs
@@ -1,0 +1,794 @@
+//! Types for the SQLite authorizer callback.
+//!
+//! See [`sqlite3_set_authorizer`](https://sqlite.org/c3ref/set_authorizer.html)
+//! and the [action code constants](https://sqlite.org/c3ref/c_alter_table.html).
+//!
+//! The authorizer callback receives an [`AuthorizerContext`]. It is an enum
+//! with one variant per SQLite action code, and every variant carries the
+//! arguments for that action under descriptive field names, so there is no
+//! need to consult a table of what each positional argument means.
+//!
+//! Two fields recur across variants. `database` is the name of the database
+//! (`"main"`, `"temp"`, or an `ATTACH` alias) an object lives in, when SQLite
+//! reports it. `accessor` is the name of the inner-most trigger or view
+//! responsible for the access, or `None` when the access is directly from
+//! top-level SQL. A string argument is `None` when SQLite passes `NULL` or a
+//! value that is not valid UTF-8.
+
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+extern crate libsqlite3_sys as ffi;
+
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+use sqlite_wasm_rs as ffi;
+
+/// Authorizing a `CREATE INDEX` statement.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct CreateIndex<'a> {
+    /// Name of the index being created.
+    pub index: Option<&'a str>,
+    /// Name of the table the index is created on.
+    pub table: Option<&'a str>,
+    /// Database the object lives in (`"main"`, `"temp"`, or an `ATTACH` alias).
+    pub database: Option<&'a str>,
+    /// Inner-most trigger or view responsible for the access, if any.
+    pub accessor: Option<&'a str>,
+}
+
+/// Authorizing a `CREATE TABLE` statement.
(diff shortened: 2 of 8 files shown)


diesel/src/connection/statement_cache/mod.rs
```diff
@@ -183,6 +183,15 @@ where
         }
     }
 
+    /// Removes all cached statements so subsequent queries are re-prepared,
+    /// while keeping the configured caching strategy.
+    // Currently used only by the SQLite authorizer, so it is dead code for
+    // backends compiled without that caller.
+    #[allow(dead_code)]
```
weiznich: Maybe rather use `#[cfg(feature = "__sqlite_shared")]` to conditionally remove it if not used?

diesel/src/sqlite/connection/authorizer.rs
```diff
@@ -0,0 +1,256 @@
...
+/// | `Attach` | Filename | - | - | - |
+/// | `Detach` | Database name | - | - | - |
+/// | `AlterTable` | Database name | Table name | - | - |
+/// | `Reindex` | Index name | - | db | - |
+/// | `Analyze` | Table name | - | db | - |
+/// | `CreateVTable` | Table name | Module name | db | - |
+/// | `DropVTable` | Table name | Module name | db | - |
+/// | `Function` | - | Function name | - | - |
+/// | `Savepoint` | Operation | Name | - | - |
+/// | `Recursive` | - | - | - | - |
```
weiznich: Given this table I wonder whether it's a good idea to just change `AuthorizerContext` from a struct to an enum which contains variants (from `AuthorizerAction`) with named fields according to their variant. This would require a bit more work on the mapping side in diesel, but make the API much easier to use. You wouldn't need to lookup which argument is what in this table anymore, you just would get it out of the received data.

The enum itself and all variants should likely be `#[non_exhaustive]` to make sure we can add fields later.

diesel/src/sqlite/connection/hooks.rs
```diff
@@ -235,6 +236,72 @@ impl SqliteConnection {
...
+    /// The callback is invoked during `sqlite3_prepare()` (statement
+    /// compilation) to control access to database objects. It receives
+    /// an [`AuthorizerContext`] describing the operation and returns an
+    /// [`AuthorizerDecision`].
+    ///
+    /// The authorizer is consulted only at statement preparation, so treat it
+    /// as defense-in-depth rather than a complete sandbox. SQLite re-checks
+    /// already-prepared statements when an authorizer is installed. Removing the
+    /// authorizer clears diesel's statement cache so statements prepared while
+    /// it was active are re-prepared without it.
```
weiznich, line 336: This means we need to clear the statement cache for each change of the authorizer, not only for the removal. So `self.cache.clear` also needs to be called by `on_authorize` itself

weiznich (commented):
Overall that's a good addition.

The statement cache changes are ok, although I have some comments on the details.

Otherwise I would like to suggest changing `AuthorizerContext` to carry the relevant information in a better format.

diesel/src/connection/statement_cache/mod.rs
LucaCappelletti94: Doner

diesel/src/sqlite/connection/authorizer.rs
LucaCappelletti94: `AuthorizerContext` is now a `#[non_exhaustive]` enum, one variant per action, each wrapping a dedicated `#[non_exhaustive]` struct with named fields, as I believe it should be the most practical version

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: I put the clear in `remove_authorizer` only, and I think `on_authorize` does not need one, because the two paths behave differently inside SQLite.

Installing a non-null authorizer expires every prepared statement on the connection (`sqlite3_set_authorizer` runs `sqlite3ExpirePreparedStatements(db, 1)` under `if( db->xAuth )`). Since we prepare with `sqlite3_prepare_v3`, those expired statements are transparently re-prepared under the new authorizer on their next step, so the cache stays correct on its own. I have double checked this just now in case I misremembered, but at least on Linux that is the case. I am not 100% certain whether it applies to all SQLites and whether people can somehow shut it down with some magical compile flag.

Removal takes the `xAuth == NULL` branch, which does not expire anything. A statement prepared while the authorizer was active can have the decision compiled into it (an `Ignore` on a column read bakes a `NULL` into the statement), and on removal nothing invalidates that cached statement, so it keeps returning the old result. That is why the manual clear is needed there.

We can add the clear to `on_authorize` too if you would rather keep it symmetric, but it would discard the whole cache on every call for no correctness gain as far as I can tell.

weiznich: I rather wouldn't rely on this behaviour as it doesn't seem to be documented, so rather lets be safe and just call `cache.clear()` then run into some future incompatibility. After all, I don't expect that users will change their authorize all the time, so this performance impact should be fine.

(Will push that fix myself, so no action needed from your side)

weiznich (approved):
Looks good now, I just push two minor fixes on top of this and then it can be merged.

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: It might be a false memory, but I recall it was documented. I will need to check, but yeah it does not seem that expensive to just be doubly sure the cache gets invalidated.

PR #5096, opened 2026-06-29
Add `SqliteConnection::on_trace` wrapping `sqlite3_trace_v2`

LucaCappelletti94:
This adds `on_trace` and `remove_trace` (and the `SqliteTraceEvent` and `SqliteTraceFlags` types) to register a callback for SQL execution tracing, wrapping `sqlite3_trace_v2`. It continues the per-API split of #4969.

Like the commit and rollback hooks the callback is `FnMut` and must not use the connection, so it is non-reentrant and needs no `RefCell`. Non-UTF-8 SQL is decoded lossily instead of aborting, matching the wal hook, and `SqliteTraceFlags` uses the `bitflags!` macro bound to the `ffi::SQLITE_TRACE_*` constants, like the existing `SqliteFunctionBehavior`.

One deliberate omission worth flagging: I did not expose the `SQLITE_TRACE_CLOSE` event. diesel unregisters the trace in `RawConnection::drop` before `sqlite3_close`, so a close event could never reach the callback, and the only way to make it reachable (closing with the callback still live) is fragile. So `SqliteTraceFlags` covers `STMT`, `PROFILE`, and `ROW` only.

### DIFF
--- a/diesel/src/sqlite/connection/hooks.rs
+++ b/diesel/src/sqlite/connection/hooks.rs
@@ -3,6 +3,7 @@ use core::num::NonZeroU32;
 
 pub(super) use super::authorizer::{AuthorizerContext, AuthorizerDecision};
 pub(super) use super::{BusyDecision, CommitDecision, ProgressDecision};
+use super::{SqliteTraceEvent, SqliteTraceFlags};
 
 impl SqliteConnection {
     /// Registers a callback invoked when a transaction is about to be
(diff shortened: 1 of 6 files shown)


diesel/src/sqlite/connection/trace.rs
```diff
@@ -0,0 +1,77 @@
...
+#[allow(clippy::unnecessary_cast)]
+pub(crate) const TRACE_ROW: u32 = ffi::SQLITE_TRACE_ROW as u32;
+
+bitflags::bitflags! {
+    /// Trace event mask selecting which events the callback receives.
+    ///
+    /// Added in SQLite 3.14.0 (2016-08-08). Combine flags with `|`, and
+    /// `SqliteTraceFlags::all()` selects every event.
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+    pub struct SqliteTraceFlags: u32 {
```
weiznich, line 32: Is there any reason not to have `SQLITE_TRACE_CLOSE`?

Edit: Seems like this is because:

```
            // Unknown or unhandled events are ignored. CLOSE is intentionally
            // not handled: diesel removes the trace before closing the
            // connection, so it never fires.
``` 

Might be worth to document this here as well?

diesel/src/sqlite/connection/trace.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Trace events delivered to the trace callback.
+///
+/// The callback receives one of these events based on the mask
+/// registered with [`on_trace`](crate::sqlite::SqliteConnection::on_trace).
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum SqliteTraceEvent<'a> {
+    /// A prepared statement is beginning to execute.
+    Statement {
```
weiznich, line 54: Can we mark the variants as `#[non_exhaustive]` as well? This would allow us to expose the currently missing information (stmt) in the future if we come up with an use-case and interface

diesel/src/sqlite/connection/raw.rs
```diff
@@ -1101,3 +1156,86 @@ where
...
+///
+/// `user_data` must point to a live `F` stored in `RawConnection::trace_hook`.
+unsafe extern "C" fn trace_trampoline<F>(
+    event_code: libc::c_uint,
+    user_data: *mut libc::c_void,
+    p: *mut libc::c_void,
+    x: *mut libc::c_void,
+) -> libc::c_int
+where
+    F: FnMut(SqliteTraceEvent<'_>),
```
weiznich: For completeness sake we want `+Send + 'static' here as well

diesel/src/sqlite/connection/hooks.rs
```diff
@@ -782,4 +835,82 @@ mod tests {
...
+            "at least one Profile event should have fired"
+        );
+    }
+
+    #[diesel_test_helper::test]
+    fn remove_trace_stops_events() {
+        use std::sync::Mutex;
+
+        let conn = &mut connection();
+        let count: Arc<Mutex<usize>> = Arc::new(Mutex::new(0));
```
weiznich:
```suggestion
        let count: Arc<AtomicUsize> = Arc::new(AtomicUsize::new(0));
```

weiznich (changes requested):
The API looks good, but I have a list of minor nits that I would like to see addressed

diesel/src/sqlite/connection/trace.rs
LucaCappelletti94: Yes, basically close can never fire

diesel/src/sqlite/connection/trace.rs
LucaCappelletti94: Ok

diesel/src/sqlite/connection/trace.rs
LucaCappelletti94: Added the note to the `SqliteTraceFlags` doc

diesel/src/sqlite/connection/trace.rs
LucaCappelletti94: All three variants (`Statement`, `Profile`, `Row`) are now `#[non_exhaustive]`

diesel/src/sqlite/connection/hooks.rs
LucaCappelletti94: DOne

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: And done

weiznich (approved):
Looks good now

PR #5098, opened 2026-06-30
Add `SqliteConnection::on_update` wrapping `sqlite3_update_hook`

LucaCappelletti94:
This adds `SqliteConnection::on_update` and `remove_update_hook`, wrapping `sqlite3_update_hook` to deliver row-change notifications (insert, update, delete). It continues the per-API split of #4969, after the commit, rollback, and progress hooks, and reuses their single-slot boxed-callback machinery and generic trampoline unchanged, so the FFI here should look familiar.

Because SQLite permits only one update hook per connection, `on_update` takes a single `SqliteUpdateRouter` rather than registering multiple callbacks. The router is composed with `on::<T>(ops, cb)` for typed per-table routes (the table is a `table!` marker, matched by name) and `on_any(ops, cb)` for a table-agnostic route, and every matching route fires for a given change:

```rust
use diesel::sqlite::{SqliteChangeOps, SqliteUpdateRouter};

conn.on_update(
    SqliteUpdateRouter::new()
        .on::<users::table, _>(SqliteChangeOps::INSERT, |change| {
            log_new_user(change.rowid);
        })
        .on::<posts::table, _>(SqliteChangeOps::INSERT | SqliteChangeOps::DELETE, |change| {
            reindex(change.op, change.rowid);
        }),
);
```

A plain catch-all is a single `on_any` route, and inside any callback `SqliteChangeEvent::is_from::<T>()` and `rowid_in::<T>()` match a marker without a string, which is also how a `diesel_dynamic_schema` user (who has no `table!` marker) filters on `change.table_name`:

```rust
conn.on_update(
    SqliteUpdateRouter::new().on_any(SqliteChangeOps::ALL, |change| {
        if let Some(rowid) = change.rowid_in::<users::table>() {
            invalidate_user(rowid);
        }
    }),
);

conn.remove_update_hook();
```

Keeping the dispatch table an explicit value the caller builds, rather than connection-owned subscription state, means there are no per-hook handles to manage: to change routing you install a different router or call `remove_update_hook`. The callback follows the same contract as the commit hook, so it must not use the connection, and a panic aborts the process.

I left out a `find_by_rowid` helper that an earlier draft carried for loading the changed row by its `rowid`. It is independent of the hook, and its public signature needs more thought, so it belongs in its own PR.

**This is the largest and last of all PRs in this series.**


diesel/src/sqlite/connection/raw.rs
```diff
@@ -407,6 +412,64 @@ impl RawConnection {
...
+    /// `sqlite3_update_hook` (the heap address of the boxed callback), or
+    /// `None` when no hook is installed. Used to assert the pointer stays
+    /// stable across a connection move, the property that prevents a
+    /// regression to a non-boxed, move-relocated callback.
+    #[cfg(test)]
+    pub(super) fn update_hook_user_data(&self) -> Option<*const ()> {
+        self.update_hook
+            .as_ref()
+            .map(|boxed| &raw const **boxed as *const ())
+    }
```
weiznich: Gives that we don't have this for the other callbacks it might be fine to remove this function?

diesel/src/sqlite/connection/update_hook.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// Routes changes on table `T` matching `ops` to `callback`.
+    ///
+    /// `T` is a [`table!`](macro@crate::table)-generated table type (for
+    /// example `users::table`); its name is read at build time via
+    /// [`StaticQueryFragment`]. Only the table name is matched, not the
+    /// database, so a same-named table in an `ATTACH`-ed database also matches.
+    /// `T` must be an unqualified table: a schema-qualified `table!` type does
+    /// not satisfy the bound, so match those on `table_name` via
+    /// [`on_any`](Self::on_any) instead.
+    pub fn on<T, F>(mut self, ops: SqliteChangeOps, callback: F) -> Self
```
weiznich: I would rather prefer taking the `T` as normal argument as well. That closely mirrors what other DSL functions do as well and feels a bit more natural to type. Also this would in theory enable passing a dynamic_schema table as well.

diesel/src/sqlite/connection/update_hook.rs
```diff
@@ -0,0 +1,480 @@
...
+    where
+        T: StaticQueryFragment<Component = Identifier<'static>>,
```
weiznich: It might be useful to also allow accessing tables with a schema here, as that could be used for the `attached` database name. Similar to what is done for blobs: https://github.com/diesel-rs/diesel/blob/main/diesel/src/sqlite/connection/mod.rs#L491

diesel/src/sqlite/connection/update_hook.rs
```diff
@@ -0,0 +1,480 @@
...
+        F: FnMut(SqliteChangeEvent<'_>) + Send + 'static,
+    {
+        self.routes.push(Route {
+            table: Some(T::STATIC_COMPONENT.0),
```
weiznich: We should include the "schema" here

weiznich (commented):
These are some preliminary comments. The overall design looks good to me, I have some comments about some details with registering tables and handling schema, but that's nothing fundamental.

I still need to look through all the tests, so this is no final review yet. That said: I would like to thank you for writing all these tests, having a good coverage always help to confidently ship such features.

diesel_dynamic_schema/tests/change_hooks.rs
```diff
@@ -0,0 +1,123 @@
...
+        .unwrap();
+
+    let items = table("items");
+    let name = items.column::<Text, _>("name");
+
+    let events: Arc<Mutex<Vec<(SqliteChangeOp, String, i64)>>> = Arc::new(Mutex::new(Vec::new()));
+    let events_hook = events.clone();
+
+    conn.on_update(
+        SqliteUpdateRouter::new().on_any(SqliteChangeOps::ALL, move |ev| {
```
weiznich, line 40: By turning the table in `fn on()` from a generic argument into a normal function argument we should also be able to enable passing a `table("items")` so that this also can be used with a dynamically typed table type.

weiznich (commented):
The tests look good to me

diesel/src/sqlite/connection/raw.rs
LucaCappelletti94: Yes, it was a refuse of a previous version of this PR that was more complex. Removed.

diesel/src/sqlite/connection/update_hook.rs
LucaCappelletti94: `on` now takes the table by value: `on(users::table, ops, callback)`, bound on `StaticQueryFragment` for static `table!` types. Dynamic tables go through the new `on_dynamic`, otherwise we would need to add a trait derive in the `table!` macro, see the other reply.

diesel/src/sqlite/connection/update_hook.rs
LucaCappelletti94: `on` is now bound on `T::Component: HasDatabaseAndTableName`, the same trait `get_read_only_blob` uses, so a schema-qualified `table!` type is accepted.

diesel/src/sqlite/connection/update_hook.rs
LucaCappelletti94: `Route` now carries the database next to the table name, filled from `STATIC_COMPONENT.database_name()`, and `matches` compares it against `event.db_name`. `None` means any database, so unqualified tables behave exactly as before, and a schema-qualified table fires only for its own database. Added `router_on_matches_schema_qualified_table`, which attaches a second database and asserts a schema-qualified route fires only for the attached copy, not the same-named table in `main`.

diesel_dynamic_schema/tests/change_hooks.rs
LucaCappelletti94: Added through a new method `SqliteUpdateRouter::on_dynamic` which allows users to pass a runtime table straight in using `router.on_dynamic(table("items"), ops, cb)`. It is bound on a new public trait `DynamicChangeTable`, which `diesel_dynamic_schema` implements for its `Table` behind its `sqlite` feature.

One `on` accepting both a static `table!` type and a dynamic `Table` would need a single trait implemented for both, which due to the lack of a sound specialization forces use to emit the trait from the `table!` macro for every table. I do not think the cost of adding that extra code generation is worth the (I believe) limited use cases for the dynamic tables + on update sqlite hooks.

`Route` now stores `Cow<'static, str>` for the table and schema, borrowed and zero-cost for static tables via `on`, owned for the dynamic ones via `on_dynamic`.

weiznich (approved):
I pushed https://github.com/diesel-rs/diesel/pull/5098/changes/dbb0e9bcb03659ddebedc4dd853e5a5cb914e050 to refactor the table name handling to support both dynamic and static tables through exactly the same code path. This removes quite a bit redundancy  and also makes the implementation a bit more clearer. Hopefully that's fine for you. (It still would be great to have a second person looking at this change).


Also I did a quick search in the SQLite documentation and the following hooks seem to be currently not covered:

* https://sqlite.org/c3ref/autovacuum_pages.html
* https://sqlite.org/c3ref/collation_needed.html
* https://sqlite.org/c3ref/preupdate_blobwrite.html

(To be clear here: That's fine, I just was curious what is still "missing")

LucaCappelletti94:
I was working just now on preupdate, which I was thinking to add in https://github.com/LucaCappelletti94/diesel-sqlite-session since session always requires the preupdate compile flag and it may not be guaranteed to be always there

I need to check the other ones, they just seemed of relatively little interest. Let's see if OCD completion compels to collect all the figurines.

LucaCappelletti94:
> It still would be great to have a second person looking at this change

Yeah objectively it is cleaner and the obvious solution, for some reason I was mentally stuck in the wrong "no we can't do it this way cos of blanket and orphan rule" space. Green for me.

PR #5099, opened 2026-06-30
add diesel_async attribute to multiconnection macro

jusax23:
Adds and implements a diesel_async attribute to multiconnection macro.
This change depends on: https://github.com/diesel-rs/diesel_async/pull/295

Prior Discussion: https://github.com/diesel-rs/diesel_async/discussions/108

I ported the test pointed out in the discussion in https://github.com/jusax23/diesel_async/tree/dev-async_multi_connection-test
This test in diesel_async is depended on this macro change. 
Should I add this test to the existing pr in diesel_async, or open a new one after it is merged?

I added a feature `async_pool`, so diesel_async can control the implementation of `PoolableConnection`.

---

### DIFF
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -1730,6 +1730,24 @@ fn view_proc_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream 
 /// corresponding enum manually by first establishing the connection via the inner type and then
 /// wrap the result into the enum.
 ///
+/// ## Required Container Attributes
+/// This macro supports generating code for [`diesel-async`](https://docs.rs/diesel-async/latest/diesel_async/).
+/// By adding the `#[diesel_async]` attribute, the macro instead derives
+/// implements `diesel_async::AsyncConnection` and related traits
+/// for an enum of connections to different databases.
+/// This way each tuple field type must implement `diesel_async::AsyncConnection`.
+/// It is possible to use `diesel-async`'s `SyncConnectionWrapper` as connection type.
+/// The implementation of `diesel_async::AsyncConnection::establish` works similar to the sync variant.
+///
+/// The usual stability guarantees of `diesel` do not apply here.
+/// When this attribute is used the generated code is only compatible to the `diesel-async` version supported,
+/// by this `diesel` version.
+///
+/// Note: The associated types `LoadFuture` and `ExecuteFuture` in the implementation
+/// of the trait [`AsyncConnectionCore`](https://docs.rs/diesel-async/latest/diesel_async/trait.AsyncConnectionCore.html)
+/// must be bound by `'conn`.
+///
+///
 /// # Example
 /// ```
 /// # extern crate diesel;
@@ -1890,7 +1908,7 @@ fn view_proc_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream 
 /// ```
 ///
 #[cfg_attr(diesel_docsrs, doc = include_str!(concat!(env!("OUT_DIR"), "/multiconnection.md")))]
-#[proc_macro_derive(MultiConnection)]
+#[proc_macro_derive(MultiConnection, attributes(diesel_async))]
 pub fn derive_multiconnection(input: TokenStream) -> TokenStream {
     derive_multiconnection_inner(input.into()).into()
 }
--- a/diesel_derives/src/lib.rs
+++ b/diesel_derives/src/lib.rs
@@ -1730,7 +1730,8 @@ fn view_proc_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream 
 /// corresponding enum manually by first establishing the connection via the inner type and then
 /// wrap the result into the enum.
 ///
-/// ## Required Container Attributes
+/// ## Optional Container Attributes
+///
 /// This macro supports generating code for [`diesel-async`](https://docs.rs/diesel-async/latest/diesel_async/).
 /// By adding the `#[diesel_async]` attribute, the macro instead derives
 /// implements `diesel_async::AsyncConnection` and related traits
--- a/diesel_derives/src/multiconnection.rs
+++ b/diesel_derives/src/multiconnection.rs
@@ -7,6 +7,7 @@ struct ConnectionVariant<'a> {
 }
 
 pub fn derive(item: DeriveInput) -> TokenStream {
+    let helper = MultiHelper::from_derive_input(&item);
     if let syn::Data::Enum(e) = item.data {
         let connection_types = e
             .variants
@@ -19,11 +20,11 @@ pub fn derive(item: DeriveInput) -> TokenStream {
                 _ => panic!("Only enums with one field per variant are supported"),
             })
             .collect::<Vec<_>>();
-        let backend = generate_backend(&connection_types);
-        let query_builder = generate_querybuilder(&connection_types);
-        let bind_collector = generate_bind_collector(&connection_types);
-        let row = generate_row(&connection_types);
-        let connection = generate_connection_impl(&connection_types, &item.ident);
+        let backend = generate_backend(&connection_types, &helper);
+        let query_builder = generate_querybuilder(&connection_types, &helper);
+        let bind_collector = generate_bind_collector(&connection_types, &helper);
+        let row = generate_row(&connection_types, &helper);
+        let connection = generate_connection_impl(&connection_types, &item.ident, &helper);
 
         quote::quote! {
             mod multi_connection_impl {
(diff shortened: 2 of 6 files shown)


diesel_derives/Cargo.toml
```diff
@@ -48,6 +48,7 @@ r2d2 = []
 chrono = []
 time = []
 numeric = []
+async_pool = []
```
weiznich: Adding new feature flags to `diesel_derives` is something we need to avoid at all cost. Feature flags are broken for proc-macro crate, we recently moved away all the code generation from being depended on feature flags. See https://github.com/diesel-rs/diesel/pull/5063 for details. A similar approach needs to be taken here, which likely means:

* diesel-async needs to export a conditional expand macro for this feature as `#[doc(hidden)]` item
* The expanded code needs to use that macro in the generated code to conditionally generate the right code.

diesel/Cargo.toml
```diff
@@ -109,6 +109,7 @@ __with_asan_tests = [
 ]
 std = ["serde_json?/std"]
 hashbrown = ["dep:hashbrown"]
+async_pool = ["diesel_derives/async_pool"]
```
weiznich: This shouldn't be required as well

diesel_derives/src/tests/snapshots/diesel_derives__tests__multiconnection_1.snap
```diff
@@ -1122,7 +1128,7 @@ mod multi_connection_impl {
         }
         impl<'a> diesel::query_builder::BindCollector<'a, MultiBackend>
         for MultiBindCollector<'a> {
-            type Buffer = multi_connection_impl::bind_collector::BindValue<'a>;
```
weiznich, line 1125: I would rather keep the fully qualified name here, as proc-macro code sometimes is used in wired places, so rather make it as robust as possible.

diesel_derives/src/tests/snapshots/diesel_derives__tests__multiconnection_1.snap
```diff
@@ -1133,7 +1139,7 @@ mod multi_connection_impl {
                 U: diesel::serialize::ToSql<T, MultiBackend> + ?Sized + 'a,
             {
                 let out = {
-                    let out = multi_connection_impl::bind_collector::BindValue::default();
+                    let out = BindValue::default();
```
weiznich: Keep the qualified path here as well

diesel_derives/src/tests/multiconnection.rs
```diff
@@ -15,3 +15,22 @@ pub(crate) fn multiconnection_1() {
...
 }
+
+#[cfg(all(feature = "r2d2", feature = "chrono", feature = "time"))]
+#[test]
+pub(crate) fn multiconnection_2() {
+    let input = quote::quote! {
+        #[diesel_async]
+        enum DbConnection {
+            Pg(PgConnection),
+            Sqlite(diesel::SqliteConnection),
```
weiznich: Can we use the real `diesel-async` connection types here? That's only cosmetic, but helps a bit to judge the generated code.

diesel_derives/src/lib.rs
```diff
@@ -1730,6 +1730,17 @@ fn view_proc_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream
...
+/// This macro supports generating code for `diesel-async`.
+/// By adding the `#[diesel_async]` attribute, the macro instead derives
+/// implements `diesel_async::AsyncConnection` and related traits
+/// for an enum of connections to different databases.
+/// This way all each tuple field type must implement `diesel_async::AsyncConnection`.
+/// It is possible to use `diesel-async`'s `SyncConnectionWrapper` as connection type.
+/// The implementation of `diesel_async::AsyncConnection::establish` works similar to the sync variant.
+///
+/// Note: The associated types `LoadFuture` and `ExecuteFuture` in the implementation
+/// of the trait `AsyncConnectionCore` must be bound by `'conn`.
```
weiznich: We usually document such attributes in a "Required Container Attributes" section

Also it might be worth to turn `diesel-async` in a docs.rs link to diesel-async.

Finally this needs to document that the usual diesel stability guarantee doesn't apply here. When this attribute is used the generated code is only compatible to whatever diesel-async version is supported by this diesel version.

diesel_derives/src/lib.rs
```diff
@@ -1890,14 +1901,19 @@ fn view_proc_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream
...
 
 fn derive_multiconnection_inner(input: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
     syn::parse2(input)
-        .map(multiconnection::derive)
+        .map(|input: syn::DeriveInput| {
+            let is_async = input.attrs.iter().any(
+                |a| matches!(&a.meta, syn::Meta::Path(ident) if ident.is_ident("diesel_async")),
+            );
+            multiconnection::derive(input, is_async)
+        })
```
weiznich: This kind of attribute parsing should be done inside of `multiconnection::derive`

diesel_derives/src/multiconnection.rs
```diff
@@ -6,8 +6,10 @@ struct ConnectionVariant<'a> {
     name: &'a syn::Ident,
 }
 
-pub fn derive(item: DeriveInput) -> TokenStream {
+pub fn derive(item: DeriveInput, is_async: bool) -> TokenStream {
     if let syn::Data::Enum(e) = item.data {
+        let helper = MultiHelper::new(is_async);
```
weiznich: Maybe even pass the attributes in this function and perform the parsing inside of the helper construction.

diesel_derives/src/multiconnection.rs
```diff
@@ -65,20 +67,111 @@ pub fn derive(item: DeriveInput) -> TokenStream {
...
+            quote::quote! { .await }
+        } else {
+            quote::quote! {}
+        };
+
+        let async_token = if is_async {
+            quote::quote! { async }
+        } else {
+            quote::quote! {}
+        };
```
weiznich, line 138: These should be `Option<TokenStream>` values

diesel_derives/src/multiconnection.rs
```diff
@@ -87,51 +180,152 @@ fn generate_connection_impl(
...
-                conn.execute_returning_count(&query)
+                conn.execute_returning_count(#ref_token query)
             }
         }
     });
 
-    let load_impl = connection_types.iter().map(|c| {
-        let variant_ident = c.name;
-        let ty = &c.ty;
+    let (impl_execute_returning_count, impl_execute_returning_count_async) = if is_async {
```
weiznich: This should likely not be a if else with two return values, where each branch only generate one

diesel_derives/src/multiconnection.rs
```diff
@@ -87,51 +180,152 @@ fn generate_connection_impl(
...
+            let variant_ident = c.name;
+            let ty = &c.ty;
+            quote::quote! {
+                Self::#variant_ident(conn) => {
+                    let query = SerializedQuery {
+                        inner: source.as_query(),
+                        backend: MultiBackend::#variant_ident(Default::default()),
+                        query_builder: super::query_builder::MultiQueryBuilder::#variant_ident(Default::default()),
+                        p: std::marker::PhantomData::<#ty>,
+                    };
```
weiznich: Would it be possible to deduplicate this part with the sync counter part below?

diesel_derives/src/multiconnection.rs
```diff
@@ -280,22 +499,43 @@ fn generate_connection_impl(
...
+        quote::quote! { diesel_async::SimpleAsyncConnection }
+    } else {
+        quote::quote! { diesel::connection::SimpleConnection }
+    };
+
+    let transaction_manager = if is_async {
+        quote::quote! { diesel_async::TransactionManager }
+    } else {
+        quote::quote! { diesel::connection::TransactionManager }
+    };
```
weiznich: That should be part of the helper instead

diesel_derives/src/multiconnection.rs
```diff
@@ -280,22 +499,43 @@ fn generate_connection_impl(
...
+
+    let simple_impls = if is_async {
+        if cfg!(feature = "async_pool") {
```
weiznich: This feature flag needs to be replaced by a macro call in the generated code

weiznich (commented):
Thanks for opening this Pr.

Overall that's the right way to go. I have quite a few comments about the details of the implementation. We can move forward with this as soon as they are addressed.

diesel_derives/src/multiconnection.rs
jusax23: Yes that would be possible. The only difference is the `.as_query()`. 
I did not do this, because i do not think the code is better readable that way.
If you disagree I will dedpulicate this.

jusax23:
Hi, I've modified the commit and hopefully resolved all the issues (except one).
I've added a commit to the diesel_async pr, to add extend_poll macro.

weiznich (approved):
Looks good now, with the minor fixes applied on top

PR #5105, opened 2026-07-04
Add `InsertStatement::execute_returning_id` for MySQL

LucaCappelletti94:
Revives #4271 by @mcronce.

Reading the `AUTO_INCREMENT` key of a freshly inserted row on MySQL or MariaDB normally costs an extra `SELECT LAST_INSERT_ID()` round trip. This adds `insert_into(table).values(..).execute_returning_id(conn)`, which returns the generated key straight from the client library (`mysql_stmt_insert_id`) after executing the insert, for both `MysqlConnection` and `MariadbConnection`.

The method only exists where it can mean something. `diesel print-schema` now detects `AUTO_INCREMENT` columns from `information_schema` and marks them in the generated `schema.rs` with a new `#[auto_increment]` column attribute:

```rust
diesel::table! {
    users (id) {
        #[auto_increment]
        id -> Integer,
        name -> Text,
    }
}
```

The `table!` macro turns that attribute into an implementation of a sealed `AutoIncrementTable` trait and the method requires it, so calling it on a table without a generated key is a compile error instead of a silent `None`. The attribute is limited to one column per table (the column does not have to be the primary key) and rejected in `view!`.

```rust
use core::num::NonZeroU64;

let new_id: Option<NonZeroU64> = diesel::insert_into(users::table)
    .values(users::name.eq("Sean"))
    .execute_returning_id(conn)?;
```

The method also accepts only single-row inserts, since the client library reports a single id per statement. Batch inserts, `INSERT ... SELECT`, and unmarked tables all fail to compile with tailored messages, for example:

```text
error[E0277]: `BatchInsert<Vec<ValuesClause<...>>, ...>` may insert zero or several rows
  = note: the database client reports one generated id per statement,
          so this operation supports only single-row inserts
```

Upserts and `default_values()` keep working. The return type stays `Option<NonZeroU64>` because a skipped `INSERT IGNORE` row still produces no id at runtime, while an explicitly supplied key and the row updated by `ON DUPLICATE KEY UPDATE` are reported back.

`migration generate --diff-schema` also emits `AUTO_INCREMENT` in generated `CREATE TABLE` statements for these two backends, so a schema carrying the attribute round-trips: the generated migration builds a database from which print-schema regenerates the same attribute.


diesel/src/mysql/connection/stmt/mod.rs
```diff
@@ -170,6 +170,14 @@ impl StatementUse<'_> {
             .map_err(|e| Error::DeserializationError(Box::new(e)))
     }
 
+    /// The executed statement's `mysql_stmt_insert_id`. Must be called after
+    /// `execute`. The C API documents this as undefined when the statement
```
weiznich: Is this c undefined behaviour? Or is this: We return an integer but we might not return a meaningful value "undefined behaviour"? The first variant would be very bad, the second one is acceptable, but we should clarify that in the doc comment in this case.

diesel/src/mysql/query_builder/insert_returning_id.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// `mysql_insert_id` never reports, such as a `Uuid`.
+    pub trait AutoIncrementInteger {}
+}
+use self::sealed::AutoIncrementInteger;
+
+impl AutoIncrementInteger for crate::sql_types::TinyInt {}
+impl AutoIncrementInteger for crate::sql_types::SmallInt {}
+impl AutoIncrementInteger for crate::sql_types::Integer {}
+impl AutoIncrementInteger for crate::sql_types::BigInt {}
+impl<T> AutoIncrementInteger for crate::sql_types::Unsigned<T> where T: AutoIncrementInteger {}
```
weiznich: You could declare these without an auto_increment marker as primary key. So we either just accept that (in this case the trait should be named differently) or we need to track that information as part of the generated schema.rs file. That would require reading it from the database, generating some attribute marker, adding support of that attribute marker to `table!` and finally implementing the right trait for the column type.

weiznich (changes requested):
I still think it's worth to add such an API, but I have questions about the invariants we should uphold here

diesel/src/mysql/connection/stmt/mod.rs
LucaCappelletti94: Not C undefined behaviour. `mysql_stmt_insert_id` returns an already initialised struct field, so it is your second variant, a well defined integer that may not be meaningful. The doc comment says that explicitly now. The value can go stale because the client copies the connection's last insert id into the statement after every execute, but an `INSERT` always gets a fresh one.

diesel/src/mysql/query_builder/insert_returning_id.rs
LucaCappelletti94: I have been stuck on this part of the PR as `MySQL` allows for ONE auto incrementing column in a table, which does not needs to be the primary key. Now, I agree that it might be a bit daft to have a `uuid` primary key and a `u64` auto increment table, I am not sure why anybody would want that, but if we bind this feature only on the primary key (why would leave the method reasonably ergonomic) we also make it impossible to retrieve the value from the corner case I described above.

I am unsure how to proceed on this matter, as the only "general" solutions I see are:

1) As it is now, in the current tentative refactoring, in which I just return a `Option<NonZeroU64>` with no compile time validation
2) Adding the attribute you were describing above, but to be applied specifically on the column of interest, and that information is then used to implement a trait `TableWithAutoIncrementingColumn` with an associated type `AutoIncrementingColumn` which is the marked column, and then we use this for the method compile time validation
3) Adding the attribute you were describing above on the table, and choose to not support the case where a table has an auto-incrementing column that is not the primary key

I am conflicted between 1 and 2, as 1 does not give compile time guarantees but it is much less code and does not come with a pletora of additional edits, and 2 as I would really like to force this method to only be available on tables for which it should be available, but will come with a non trivial amount of additions just about everywhere.

diesel/src/mysql/query_builder/insert_returning_id.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// use core::num::NonZeroU64;
+    /// let conn = &mut establish_connection();
+    /// let new_id = diesel::insert_into(users)
+    ///     .values(name.eq("Ruby"))
+    ///     .execute_returning_id(conn)?;
+    /// // Two users (ids 1 and 2) are seeded, so Ruby's generated id is 3.
+    /// assert_eq!(new_id, NonZeroU64::new(3));
+    /// # Ok(())
+    /// # }
+    /// ```
```
weiznich: Can restrict this to only work with single inserts. This looks like potential problem with batch inserts as we only return a single id and not a list of ids.

diesel/src/mysql/query_builder/insert_returning_id.rs
weiznich: I'm open to merge this PR as currently implement, although if you would like to have the challenge to implement variant 2 I'm not rejecting that either. 
From a technical point of view variant 2 is more robust, although as you have written it would require much more changes.

LucaCappelletti94: I went for option 2, after rebasing it to support also the newly added MariaDB backend

diesel/src/mysql/query_builder/insert_returning_id.rs
LucaCappelletti94: The insert values now have to implement a `SingleRowInsertValues` marker, implemented for single-row values clauses, `default_values()`, and upserts wrapping those. Batch inserts and `INSERT ... SELECT` fail to compile with a `#[diagnostic::on_unimplemented]` message explaining that the client library reports one generated id per statement. The former multi-row runtime test is now a compile-fail test.

weiznich (approved):
This looks really good. Thanks for the update

PR #5106, opened 2026-07-07
Expose collected SQLite bind values under the third-party backend feature

LucaCappelletti94:
The SQLite bind collector could be populated but not read back, because `SqliteBindCollectorData.binds` and the `OwnedSqliteBindValue` enum were private. Postgres already allows this through `RawBytesBindCollector<Pg>` under `i-implement-a-third-party-backend-and-opt-into-breaking-changes`, so a third-party backend can render a typed query to placeholder SQL and recover the ordered bind values that go with it. SQLite had no equivalent.

This makes `SqliteBindCollector` (and its `new`), `SqliteBindCollectorData` with a public `binds` field, and `OwnedSqliteBindValue` public under that same feature, via `__diesel_public_if` and feature-gated re-exports. The enum stays exhaustive since its variants cover SQLite's storage classes.

One more note: initially, I had also added tests in the external testing crate actually using this, but I saw that there was no test for the equivalent Pg object, so I decided against it as it would require to run tests with the `i-implement-a-third-party-backend-and-opt-into-breaking-changes` feature. I could go for the other direction, and add add tests for both SQLite and Pg, as you prefer.

### DIFF
--- a/diesel/src/sqlite/connection/bind_collector.rs
+++ b/diesel/src/sqlite/connection/bind_collector.rs
@@ -11,14 +11,73 @@ use libsqlite3_sys as ffi;
 #[cfg(all(target_family = "wasm", target_os = "unknown"))]
 use sqlite_wasm_rs as ffi;
 
+/// The [`BindCollector`] used by the SQLite backend.
+///
+/// You only interact with this when adding a third-party SQLite backend.
 #[derive(Debug, Default)]
 pub struct SqliteBindCollector<'a> {
-    pub(in crate::sqlite) binds: Vec<(InternalSqliteBindValue<'a>, SqliteType)>,
+    pub(in crate::sqlite) binds: Vec<(SqliteBindValueRef<'a>, SqliteType)>,
 }
 
-impl SqliteBindCollector<'_> {
+impl<'a> SqliteBindCollector<'a> {
+    /// Construct an empty `SqliteBindCollector`
+    #[diesel_derives::__diesel_public_if(
+        feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+    )]
     pub(in crate::sqlite) fn new() -> Self {
         Self { binds: Vec::new() }
+    }
+
+    /// Iterate over the collected bind values and their SQLite storage classes,
+    /// in positional order.
+    ///
+    /// Each yielded tuple carries a reference to the [`SqliteBindValueRef`] as
+    /// it lives inside the collector, so borrowed string and blob variants are
+    /// exposed without an intervening copy. If the caller needs a
+    /// [`Send`] snapshot instead, use [`MoveableBindCollector::moveable`]
+    /// and inspect [`SqliteBindCollectorData`].
+    ///
+    /// # Example
+    ///
+    /// A third-party backend can render a typed Diesel query into
+    /// placeholder SQL and recover the ordered bind values:
+    ///
+    /// ```rust
+    /// use diesel::expression::IntoSql;
+    /// use diesel::query_builder::{QueryBuilder, QueryFragment};
+    /// use diesel::sql_types::{BigInt, Integer, Text};
+    /// use diesel::sqlite::{
+    ///     Sqlite, SqliteBindCollector, SqliteBindValueRef, SqliteQueryBuilder, SqliteType,
+    /// };
+    ///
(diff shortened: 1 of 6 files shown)


hiSandog:
For a third-party-backend API, an external crate compile test would be more convincing than only internal tests. It should enable `i-implement-a-third-party-backend-and-opt-into-breaking-changes`, build a SQLite query, and assert both placeholder SQL and ordered bind values are recoverable without reaching into private modules.

LucaCappelletti94:
I agree, that is what I wrote in the last paragraph of the PR post. Should I add those?

diesel/src/sqlite/connection/bind_collector.rs
```diff
@@ -211,13 +218,25 @@ impl<'a> BindCollector<'a, Sqlite> for SqliteBindCollector<'a> {
...
 }
 
+/// An owned value bound to a SQLite prepared statement.
+///
+/// The readable counterpart to the values a [`SqliteBindCollector`] holds.
 #[derive(Debug, Clone)]
+#[diesel_derives::__diesel_public_if(
+    feature = "i-implement-a-third-party-backend-and-opt-into-breaking-changes"
+)]
 enum OwnedSqliteBindValue {
```
weiznich, line 291: That's not the type used to collect binds, that's the type used by `SqliteBindCollectorData`. I think we rather want to expose `InternalSqliteBindValue` (maybe as `SqliteBindValue`?)

weiznich (changes requested):
It's fine to expose that behind the unstable feature flag.

That written I still see no way to actually access the collected bind data in the public api.

This would need another helper function (or two) on `SqliteBindCollector` that expose an iterator of bind values and bind types

(It's also fine to duplicate that for SqliteBindCollectorData, although that's strictly speaking not required)

LucaCappelletti94:
Since I am already using this PR in a crate downhill, I am certain that the moveable path is already reachable from an outside crate. Here is a minimal repro that only turns on `i-implement-a-third-party-backend-and-opt-into-breaking-changes` this compiles and runs, using only items re-exported from `diesel::sqlite`.

```rust
use diesel::query_builder::{MoveableBindCollector, QueryBuilder, QueryFragment};
use diesel::sqlite::{
    OwnedSqliteBindValue, Sqlite, SqliteBindCollector, SqliteBindCollectorData,
    SqliteQueryBuilder, SqliteType,
};

let mut qb = SqliteQueryBuilder::new();
q.to_sql(&mut qb, &Sqlite)?;
let sql = qb.finish();

let mut c = SqliteBindCollector::new();
q.collect_binds(&mut c, &mut (), &Sqlite)?;
let data: SqliteBindCollectorData = c.moveable();
for (bind, ty) in &data.binds {
    match bind {
        OwnedSqliteBindValue::I32(v) => ...,
        OwnedSqliteBindValue::String(s) => ...,
        ...
    }
}
```

That said, having to go through `moveable()` and `OwnedSqliteBindValue` clones every borrowed string and blob for a caller that only wants to walk the live collector once, and `OwnedSqliteBindValue` is not what actually rides through `push_bound_value`.

Since there is already a `SqliteBindValue` in the context of the `ToSql` impls, I went for renaming `InternalSqliteBindValue` into `SqliteBindValueRef` and `SqliteBindCollector::binds` to iterate them.

To actually add an executed test, I have added a pair of doctests that basically runs the above snippet, and enabled `i-implement-a-third-party-backend-and-opt-into-breaking-changes` in xtask.

xtask/src/tests/mod.rs
```diff
@@ -227,7 +227,9 @@ impl TestArgs {
                 .arg("-F")
                 .arg(format!("diesel-dynamic-schema/{backend}"))
                 .arg("-F")
-                .arg(format!("diesel_migrations/{backend}"));
```
weiznich, line 230: We shouldn't change that as we also want to run tests without that flag enabled. 

The CI setup covers different variants here: https://github.com/diesel-rs/diesel/blob/main/.github/workflows/ci.yml#L242

LucaCappelletti94: Ok, so I should drop that, correct? The doctest as it is is fine?

weiznich: Yes please drop it

The doc test is a great location to test this.

LucaCappelletti94: Done

weiznich (approved):

PR #5108, opened 2026-07-08
Raise minimum libsqlite3-sys to 0.24.0

LucaCappelletti94:
Prerequisite for wrapping `sqlite3_autovacuum_pages`, mentioned by @weiznich in a [comment in PR #5098](https://github.com/diesel-rs/diesel/pull/5098#pullrequestreview-4647852559). That function was added in SQLite `3.37.0` (2021-11) and is absent from the bundled amalgamation of every `libsqlite3-sys` older than `0.24.0` (the first release to ship SQLite `3.38.0`). Bumping the floor here lets the follow-up `on_autovacuum` PR use the `ffi::` alias like every other hook wrapper in the crate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 ### Changed
 
 * The minimal supported Rust version is now 1.88.0
+* The minimum supported version of `libsqlite3-sys` is now `0.24.0`, up from `0.17.2`. `0.24.0` is the first release whose bundled SQLite amalgamation (3.38.0) exposes `sqlite3_autovacuum_pages`, which future hook wrappers rely on. Users who pinned to an older `libsqlite3-sys` will need to update.
 * Add support for no-std environments using the SQLite backend
 * Improved documentation and added examples for `filter_target` on `IncompleteOnConflict`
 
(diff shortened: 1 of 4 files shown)


weiznich (changes requested):
I fear that's not possible to change as changing the minimal supported libsqlite3-sys version is a public visible change due to the status of libsqlite3-sys being a sys crate and only one version of a sys crate could exist in a single dependency tree. So a hypothetical uses could have pinned libsqlite3-sys to 0.17.2 and that would break to build with this version bump. So this would require a major version bump, which we don't have planned yet.

I think it's fine to just not have `sqlite3_autovacuum_pages` support then.

LucaCappelletti94:
Ok, then `sqlite3_autovacuum_pages` gets parked until some future date when we decide to bump libsqlite3. Maybe we should register this among the "things that may get unlocked upon a major version bump"?

weiznich:
Yes that would be meaningful. I've already put a few `TODO: diesel 3.0` comments in the source code for similar minor things. Might be meaningful to have that for this function as well.

LucaCappelletti94:
It would be nice if there was a way to specify "future PR" or some such in GitHub, maybe the next best thing could be a tag for the PR?

weiznich:
I can tag it with the `Diesel 3.0` label we have. Hopefully it doesn't get lost

PR #5113, opened 2026-07-13
feat: Optionally run migrations without generating schemas

har7an:
for use in situations where both steps should run strictly separated. This allows, for example, to keep a schema file path configured in `diesel.toml` without actually overwriting it during database/schema generation in favor of writing the schema to a user-defined location. This simplifies diffing outputs and working with schemas in read-only environments.

### DIFF
--- a/diesel_cli/src/migrations/mod.rs
+++ b/diesel_cli/src/migrations/mod.rs
@@ -33,7 +33,11 @@ pub enum MigrationFormat {
 #[derive(Debug, Subcommand)]
 pub enum MigrationCommand {
     /// Runs all pending migrations.
-    Run,
+    Run {
+        /// Do not regenerate `schema.rs` during migration.
+        #[arg(long = "no-schema", action = ArgAction::SetTrue)]
+        no_schema: bool,
+    },
 
     /// Reverts the specified migrations.
     Revert {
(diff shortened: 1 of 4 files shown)


har7an:
Note: I tried getting the setup running on my machine but after about an hour I gave up on getting the tests going. Though now I see I need to fix some of the snapshots so I'll try to get back to it. Sorry for the noise!

har7an:
The code is finished and ready for review.

Sorry once more for the initial noise!

diesel_cli/src/migrations/mod.rs
```diff
@@ -200,12 +204,16 @@ pub(super) fn run_migration_command(
...
 ) -> Result<(), crate::errors::Error> {
     match args.command {
-        MigrationCommand::Run => {
+        MigrationCommand::Run { no_schema } => {
             let (mut conn, dir) =
                 conn_and_migration_dir(migration_dir, database_url.clone(), config_file.clone())?;
 
             run_migrations_with_output(&mut conn, dir)?;
-            regenerate_schema_if_file_specified(config_file, database_url, locked_schema)?;
+            if no_schema {
```
LucaCappelletti94: Why not just write `if !no_schema` instead of an empty branch?

LucaCappelletti94 (changes requested):

diesel_cli/src/migrations/mod.rs
har7an: I don't generally like the cognitive complexity of double negations (`!` and `no_`). In this case I felt like renaming the CLI flag to rephrase `no_` wasn't exactly a good idea either, since it's pretty common in many tools for CLI flags that disable something to start with `--no-` (think of `git`, for example).

Would you prefer I changed this?

har7an: Sorry, I just realized after sending that you *did* in fact request I change this. I'm not used to GitHubs UI for this.

Let me know whether you'd like me to rebase these changes before merging, or if you're going to squash the changes anyway (Your contributing guide doesn't mention which you prefer).

weiznich (approved):

diesel_cli/src/migrations/mod.rs
```diff
@@ -33,7 +33,11 @@ pub enum MigrationFormat {
 #[derive(Debug, Subcommand)]
 pub enum MigrationCommand {
     /// Runs all pending migrations.
-    Run,
+    Run {
+        /// Do not regenerate `schema.rs` while running the  migrations.
```
pksunkara: Unintentional extra space before `migrations`

PR #5119, opened 2026-07-21
Emit `#[doc(hidden)]` for undocumented `table!` and `view!` items

LucaCappelletti94:
Under `#![forbid(missing_docs)]` a bare `diesel::table! { ... }` fails to compile as the generated schema module and each column struct carry only the caller's own doc comments, so an undocumented schema leaves them as undocumented public items.

The generator now marks the schema module and the column structs `#[doc(hidden)]`, but only when the caller wrote no doc comment for that item, reusing the `#[doc(hidden)]` approach the macro already applies to its own internals rather than fabricating filler text. Caller-written docs are left exactly as-is and still render.

Arguably, a user could just document the table macro call, but I thought this little ease of life thing may still be desirable to have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * The minimal supported Rust version is now 1.88.0
 * Add support for no-std environments using the SQLite backend
 * Improved documentation and added examples for `filter_target` on `IncompleteOnConflict`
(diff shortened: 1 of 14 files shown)


Mingun:
It seems to me that you may want to get those errors to catch undocumented items in the `table!` macros. So it is better to stick to the explicit approach here.

weiznich:
I agree with @Mingun here: It's a bad idea to generate `#[doc(hidden)]` items here just to suppress the warning. After all the items are not documented so the warning is correct. The correct fix for that is either to document the items by putting doc comments on everything, or generating them via having comments at database level or to suppress the warning at module level by e.g putting `#[allow(missing_docs)]` on `mod schema;`

LucaCappelletti94:
Ok!

PR #5120, opened 2026-07-21
fix await position in load_query builder

jusax23:
I noticed that the await in the load_query builder was outside an async block, causing the diesel_async version to not compile.

### DIFF
--- a/diesel_derives/src/multiconnection.rs
+++ b/diesel_derives/src/multiconnection.rs
@@ -239,7 +239,7 @@ fn generate_connection_impl(
                 query_builder: super::query_builder::MultiQueryBuilder::#variant_ident(Default::default()),
                 p: std::marker::PhantomData::<#ty>,
             };
-            let r = <#ty as #conn_load>::load(conn, query)#await_token?;
+            let r = <#ty as #conn_load>::load(conn, query);
         }
     };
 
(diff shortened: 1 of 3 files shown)


weiznich:
@gmarcha We are not interested in any automated reviews. This kind of tools are not welcome here.

gmarcha:
> @gmarcha We are not interested in any automated reviews. This kind of tools are not welcome here.

Sorry for the noise, removing it.

LucaCappelletti94 (commented):
It makes the  non-async generated code look a bit quirky with the `r?`, but other than that looks fine.

The only odd thing I noticed is that in the two snap tests we are including only postgres and sqlite, and there is no mysql. I do not know whether this is intended or it is a gap.

I have not verified the main claim of the PR that without this code diesel-async would fail compilation, and I find it is curious that the `*2.snap` file did not fail to compile previously if that was the case.

pksunkara (approved):

weiznich:
> The only odd thing I noticed is that in the two snap tests we are including only postgres and sqlite, and there is no mysql. I do not know whether this is intended or it is a gap.
>
> I have not verified the main claim of the PR that without this code diesel-async would fail compilation, and I find it is curious that the *2.snap file did not fail to compile previously if that was the case.

These snap tests do only generate the relevant code, they do not perform any check if the code compiles. There are two use-cases for this: Using the expanded code in the documentation and manually checking if the expanded code looks right. For both it's sufficient to have two backends there.

Actual tests if the generated code is valid are performed in the integration tests for everything that's not diesel-async. For Diesel-async there is a in-flight PR (https://github.com/diesel-rs/diesel_async/pull/295) that adds the relevant methods and tests there.

PR #5123, opened 2026-07-23
diesel_dynamic_schema: support adding heterogeneous SQL-typed fields to DynamicSelectClause

DhineshKrishnan1206:
DynamicSelectClause::add_fields requires every field to be the same type, so columns with different SQL types can't be collected into one vector and added at once. This PR adds IntoBoxedField/DynamicSelectClause::box_field to box fields of different types into a common Box<dyn QueryFragment<...>>, plus add_boxed_fields to add them in bulk.

I used Gemini for the tests/docs — and reviewed everything before opening the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 
 * Diesel-Migrations now contains a migration source that easily allows you to register Rust based migrations
 * Diesel-Migrations now contains a migration source that allows you to combine migrations from several different sources
+* Added `DynamicSelectClause::add_boxed_fields`, `DynamicSelectClause::box_field`, and the `IntoBoxedField` trait to `diesel_dynamic_schema`, allowing columns of different SQL types to be collected into a single `Vec` and added to a dynamic select clause in one call
 * Added `SqliteConnection::with_raw_connection` to provide safe, callback-based access to the raw `*mut sqlite3` handle for advanced SQLite C APIs (session extension, hooks, etc.)
 * Added `SqliteConnection::on_commit` and `SqliteConnection::remove_commit_hook` to register a callback invoked when a transaction is about to be committed, wrapping `sqlite3_commit_hook`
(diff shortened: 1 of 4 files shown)


weiznich:
Thanks for opening this PR. I would like to kindly point to our contributions documentation, [especially the section about LLM usage](https://github.com/diesel-rs/diesel/blob/main/CONTRIBUTING.md#usage-of-llmsai-agents). Would it be possible to edit your PR description to include the relevant information?

DhineshKrishnan1206:
> Thanks for opening this PR. I would like to kindly point to our contributions documentation, [especially the section about LLM usage](https://github.com/diesel-rs/diesel/blob/main/CONTRIBUTING.md#usage-of-llmsai-agents). Would it be possible to edit your PR description to include the relevant information?

Thanks for the pointer — I've updated the PR description to include the AI usage disclosure.

weiznich:
So I had some time looking at the change and thinking about the API and somehow this change doesn't feel right to me. Let me try to explain why that's the case.

If you select a `DynamicSelectClause` you get something of the SQL type `Untyped`, which essentially communicates to diesel that the type of this select expression is not know at compile time. You then can use `#[derive(QueryableByName)]` or `DynamicRow` to get access to the result values.

Now this implies that it's not really meaningful to add column types from diesel-dynamic-schema with specific SQL types into the `DynamicSelectClause` container. It will ignore them anyway. Therefore you can just construct your column types with `Untyped` as SQL type as well and then use the existing `add_fields` method or even the `FromIterator` impl as all the column types have the same type. This is also demonstrated in this test:

https://github.com/diesel-rs/diesel/blob/main/diesel_dynamic_schema/tests/dynamic_values.rs#L105

Also the added methods do not really add anything meaningful to the API. You already can just do a `Box::new(whatever_column) as Box<dyn QueryFragment<…>>` to get the same thing with the existing methods.

Now what would be a meaningful addition in my opinion is the following:

* Add documentation about what I wrote above, so ideally several code examples for `DynamicSelectClause`
* Maybe add a `into_untyped()` function to `diesel_dynamic_schema::Column` so that you can easily erase the type there

PR #5128, opened 2026-07-24
Add `attach_database` and `detach_database` helpers for `SqliteConnection`

LucaCappelletti94:
This PR adds `attach_database` and `detach_database` that pass the path and the schema name as bound parameters, so as to provide a standardized and, I believe, safer way to attach DBs in SQLite than raw strings.

```rust
conn.attach_database("data/reports.db", "reports")?;
// query reports.<table> alongside the main schema, then:
conn.detach_database("reports")?;
```


diesel/src/sqlite/connection/mod.rs
```diff
@@ -1223,6 +1223,59 @@ impl SqliteConnection {
             .get_db_config_bool(raw::SQLITE_DBCONFIG_ENABLE_ATTACH_WRITE)
     }
 
+    /// Attach the database file at `path` under `schema_name`.
+    ///
+    /// Runs [`ATTACH DATABASE ? AS ?`](https://www.sqlite.org/lang_attach.html) with
+    /// both operands bound as parameters, so no escaping or quoting is needed and a
+    /// `file:` path is not treated as a [URI](https://sqlite.org/uri.html) unless the
+    /// connection enabled URI filenames.
```
weiznich: This feature flag is enabled by default by diesel: 

https://github.com/diesel-rs/diesel/blob/main/diesel/src/sqlite/connection/raw.rs#L114

So it's meaningful to reflect that in the documentation here?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -3232,6 +3285,227 @@ mod tests {
         let _ = std::fs::remove_file(&path);
     }
 
+    // Build a temp path with an arbitrary file name so we can exercise names
+    // that would break a hand-assembled ATTACH statement.
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+    fn temp_db_path_named(file_name: &str) -> std::path::PathBuf {
+        let mut path = std::env::temp_dir();
```
weiznich: I would rather use the tempfile crate here to also cleanup after test runs? Or just use additionally `:memory:` databases?

diesel/src/sqlite/connection/mod.rs
```diff
@@ -3232,6 +3285,227 @@ mod tests {
...
+        let conn = &mut connection();
+
+        // A space or quote in the schema name would need identifier quoting in a
+        // hand-assembled statement. Bound as a parameter it is taken verbatim. The
+        // reference below stays raw SQL: such a schema is not expressible via `table!`.
+        let schema = "weird 'schema";
+        conn.attach_database(":memory:", schema).unwrap();
+
+        conn.batch_execute(
+            "CREATE TABLE \"weird 'schema\".t (id INTEGER PRIMARY KEY);
```
weiznich: Use a raw string `r#"…"#` for this instead of escaping the `"` quotes

weiznich (commented):
That seems like a meaningful addition. I have only some minor nits about the documentation and the tests, otherwise this looks good to me

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1412,6 +1407,49 @@ fn error_message(err_code: libc::c_int) -> &'static str {
...
+        out.push_bind_param::<crate::sql_types::Text, _>(self.schema_name)?;
+        Ok(())
+    }
+}
+
+impl QueryId for DetachDatabase<'_> {
+    type QueryId = ();
+
+    const HAS_STATIC_QUERY_ID: bool = false;
+}
```
weiznich: It's fine to replace this with `#[derive(QueryId)]` on both `AttachDatabase<'_>` and `DetachDatabase<'_>`. Both produce always the same query string as any user provided value is using bind values.

weiznich (changes requested):
There is this one minor thing about the `QueryId` impl otherwise this looks good to me

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: By adding the very reasonable derive you proposed, it manifested a dormant error in the derive. Fixing it in another PR.

weiznich (approved):

PR #5129, opened 2026-07-24
Add `page_count` and `freelist_count` readers to `SqliteConnection`

LucaCappelletti94:
Reading a SQLite database's size and reclaimable slack currently means hand-rolling a `sql_query` row type per pragma. This adds `SqliteConnection::page_count` and `SqliteConnection::freelist_count`.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -1449,6 +1449,61 @@ impl SqliteConnection {
         .map(|_| ())
     }
 
+    /// Total number of pages in a database, via `PRAGMA page_count`.
+    ///
+    /// `schema` selects an attached database by name, `None` reads `main`.
+    /// Multiply by the page size for the size the database accounts for.
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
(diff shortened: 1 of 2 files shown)


LucaCappelletti94:
I tried to stack these PRs, since some utilities is shared across them (expecially the others relative to vacuum) but it seems like the branch of origin needs to be from the `diesel` repo, so it does not work.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1317,6 +1317,66 @@ impl SqliteConnection {
...
+    fn read_pragma_count(
+        &mut self,
+        pragma: &'static str,
+        schema: Option<&str>,
+    ) -> QueryResult<i64> {
+        use crate::query_dsl::RunQueryDsl;
+
+        let schema = schema.unwrap_or("main");
+        let query = alloc::format!("PRAGMA \"{}\".{pragma}", schema.replace('"', "\"\""));
+        crate::dsl::sql::<crate::sql_types::BigInt>(&query).get_result(self)
```
weiznich: Same as with the vaccum PR: Use the query builder for such constructions. Maybe it can be even shared between these two PR's

LucaCappelletti94: Yeah these PRs are all extremely similar, I will apply your notes across them

weiznich (approved):

PR #5130, opened 2026-07-24
Add typed `auto_vacuum` mode control to `SqliteConnection`

LucaCappelletti94:
Adds `SqliteConnection::auto_vacuum` and `set_auto_vacuum` over a typed `AutoVacuumMode` enum, each taking an optional schema to target an attached database, replacing hand-written `PRAGMA auto_vacuum` strings.


diesel/src/sqlite/connection/mod.rs
```diff
@@ -377,6 +377,52 @@ pub enum BusyDecision {
...
+/// The mode is stored in the database file, not the connection. [`Full`] and
+/// [`Incremental`] can be switched between at any time, but changing from or to
+/// [`None`] only takes effect on a database with no tables yet, or after a
+/// subsequent `VACUUM` rewrites the file.
+///
+/// [`None`]: AutoVacuumMode::None
+/// [`Full`]: AutoVacuumMode::Full
+/// [`Incremental`]: AutoVacuumMode::Incremental
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AutoVacuumMode {
```
weiznich, line 398: I rather would like to add `#[non_exhaustive]` to all enums that users don't need to match on to let the door open to add additional variants in the future, especially for such where we don't control the variants on our own.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1317,6 +1363,58 @@ impl SqliteConnection {
...
+    /// use diesel::sqlite::AutoVacuumMode;
+    /// let conn = &mut SqliteConnection::establish(":memory:").unwrap();
+    /// assert_eq!(conn.auto_vacuum(None)?, AutoVacuumMode::None);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    pub fn auto_vacuum(&mut self, schema: Option<&str>) -> QueryResult<AutoVacuumMode> {
+        use crate::query_dsl::RunQueryDsl;
+
+        let query = alloc::format!("PRAGMA {}.auto_vacuum", quote_schema(schema));
```
weiznich: I rather would prefer having an explicitly build query here. So either manually by using `SqliteQueryBuilder` or even better by having:

```rust

#[derive(QueryId)]
struct PragmaAutoVacuum<'a> {
    schema: Option<&'a str>
}

impl QueryFragment<Sqlite> for PragmaAutoVacuum {
  // impl with using `push_identifier` for the schema, etc
}

impl RunQueryDsl for PragmaAutoVaccum {}

```

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1317,6 +1363,58 @@ impl SqliteConnection {
...
+    /// #     Ok(())
+    /// # }
+    /// ```
+    pub fn set_auto_vacuum(
+        &mut self,
+        schema: Option<&str>,
+        mode: AutoVacuumMode,
+    ) -> QueryResult<()> {
+        let query = alloc::format!("PRAGMA {}.auto_vacuum = {mode}", quote_schema(schema));
+        self.batch_execute(&query)
```
weiznich: Same as above + maybe even use `#[derive(Enum)]` for the `AutoVaccumMode` instead of the display impl.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1359,6 +1457,13 @@ fn error_message(err_code: libc::c_int) -> &'static str {
     ffi::code_to_str(err_code)
 }
 
+// Quotes a schema name for interpolation into a `PRAGMA`, which accepts no bound
+// parameters. The name becomes a double-quoted identifier with interior quotes
+// doubled, `None` targets `main`.
+fn quote_schema(schema: Option<&str>) -> String {
+    alloc::format!("\"{}\"", schema.unwrap_or("main").replace('"', "\"\""))
+}
```
weiznich: That shouldn't be required if we use the existing query builder infrastructure instead.

diesel/src/sqlite/connection/mod.rs
```diff
@@ -1317,6 +1363,58 @@ impl SqliteConnection {
...
+    /// assert_eq!(conn.auto_vacuum(None)?, AutoVacuumMode::None);
+    /// #     Ok(())
+    /// # }
+    /// ```
+    pub fn auto_vacuum(&mut self, schema: Option<&str>) -> QueryResult<AutoVacuumMode> {
+        use crate::query_dsl::RunQueryDsl;
+
+        let query = alloc::format!("PRAGMA {}.auto_vacuum", quote_schema(schema));
+        let code = crate::dsl::sql::<crate::sql_types::BigInt>(&query).get_result(self)?;
+        AutoVacuumMode::from_code(code)
```
weiznich: We likely can also use `#[derive(Enum)]` here as well.

weiznich (changes requested):
This seems a reasonable addition to the API as well, although I would like to change quite a few details about the implementation to not reimplement already existing parts of the query building infrastructure.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Removed

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Generic as `Pragma<'a, ST>` so #5129 can share it.

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Done

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Done

weiznich (approved):

LucaCappelletti94:
@weiznich I think we may attempt merging again? The last attempt failed because of the missing `ToString`. Once this is in main, I will refactor and de-duplicate #5129

PR #5137, opened 2026-07-31
Resolve the getting-started test binaries at compile time

LucaCappelletti94:
The step 1 and step 2 tests asked `assert_cmd` for `show_posts` and `write_post`, but those packages build `show_posts_step_1`, `show_posts_step_2` and `write_post_step_2`. With no matching `CARGO_BIN_EXE_*` set, `assert_cmd` fell back to searching the shared target directory, found step 3's binaries there and ran those instead.

Nightly stopped resolving that fallback, which is what finally turned it into a failure.

Therefore, I switched to `cargo_bin_cmd!`, which resolves the target through `env!` at compile time. A name that does not match a binary in the same package is now a build error instead of a silent substitution.

That needs `assert_cmd` 2.1, and the sqlite step 2 expectation changes to what step 2 actually prints.

**LLM disclosure: I launched Fable against this because it was way too cursed**

### DIFF
--- a/examples/mysql/getting_started_step_1/Cargo.toml
+++ b/examples/mysql/getting_started_step_1/Cargo.toml
@@ -11,7 +11,7 @@ openssl-sys = { workspace = true, features = ["vendored"] }
 dotenvy = "0.15"
 
 [dev-dependencies]
-assert_cmd = "2.0.14"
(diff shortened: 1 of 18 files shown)


weiznich (approved):
Thanks, that is helpful

PR #5138, opened 2026-07-31
Rebuild the generic arguments of `QueryId` derive

LucaCappelletti94:
`#[derive(QueryId)]` rebuilt the type's generic arguments grouped by category, lifetimes then types then consts, rather than in the order they were declared. That breaks three ways

* A lifetime is carried over verbatim, so `struct Query<'a>` yields `type QueryId = Query<'a>` and fails the `Any` bound on `QueryId::QueryId`.
* A bounded lifetime is emitted together with its bounds, so `struct Query<'a: 'b, 'b>` produces `Query<'a: 'b, 'b>` in type position and does not parse at all.
* And a const declared before a type parameter comes back swapped, so `struct Query<const N: usize, T>` is rejected with `E0747`.

The arguments are now built by walking the parameters in order and mapping each to its argument form. Lifetimes become `'static`, since they cannot affect the generated SQL and the associated type exists only as a cache key, so two values of one type holding different lifetimes correctly share a prepared statement.

### DIFF
--- a/diesel_derives/src/query_id.rs
+++ b/diesel_derives/src/query_id.rs
@@ -38,7 +38,20 @@ pub fn derive(mut item: DeriveInput) -> TokenStream {
     });
 
     let struct_name = &item.ident;
-    let lifetimes = item.generics.lifetimes();
+    // Arguments must follow the declaration order of the parameters, so walk them in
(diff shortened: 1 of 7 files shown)


weiznich (approved):

LucaCappelletti94:
@weiznich green for merge I believe

PR #5139, opened 2026-07-31
Add Mariadb as a Backend

jraffeiner:
## Description
Introduces a dedicated `mariadb` backend to Diesel. 

## Implementation Details
- Added `diesel/src/mariadb/` module
- Configured connection URL parsing to recognize `mariadb://` schemes
- Ensured full compatibility with `diesel_migrations` & `diesel_cli`
- Added `mariadb` feature flag to `diesel/Cargo.toml` (no default)
- Added QueryImpls, derives etc. to ensure feature parity with Mysql-Backend
- Added `mariadb_type(_)` to derive with fallback to `mysql_type(_)` -> Mariadb supports all types Supported by Mysql + it's own additional types
- Tried to reuse as much as possible from Mysql
- Added private Trait `MysqlLikeBackend` for shared `FromSql` impls
- Changed `MysqlConnection` to `MysqlLikeConnection<Mysql>`
- Added Type aliases for `MysqlLikeConnection<Mariadb>` and `MysqlLikeConnection<Mysql>`

## Testing & Validation
- Copied Mysql-Tests as Mariadb-Tests
- Added mariadb docker image to docker compose for testing
- Added mariadb environment variables for testing

## Outstanding TODOs
- Enable Uuid support for Mariadb (Should the SQL-Type Uuid be moved from PG-Types to another location?)
- Enable INSERT ... RETURNING support for Mariadb (still got some errors during testing)
- Other Mariadb specific features?


diesel/src/mysql_like/connection/stmt/iterator.rs
```diff
@@ -1,23 +1,28 @@
...
 use crate::connection::statement_cache::MaybeCached;
-use crate::mysql::{Mysql, MysqlType};
+use crate::mysql::{Mysql, MysqlLikeBackend, MysqlType};
 use crate::result::QueryResult;
 use crate::row::*;
 
 #[allow(missing_debug_implementations)]
-pub struct StatementIterator<'a> {
+#[expect(private_bounds)]
+pub struct StatementIterator<'a, B: MysqlLikeBackend> {
```
weiznich: It would be great to move everything that's shared between both backend to a `mysql_like` module, so that the `mysql` and the `mariadb` module only contain the specific code.

diesel/src/mysql/connection/url.rs
```diff
@@ -64,7 +64,7 @@ impl ConnectionOptions {
             Err(_) => return Err(connection_url_error()),
         };
 
-        if url.scheme() != "mysql" {
+        if url.scheme() != "mysql" && url.scheme() != "mariadb" {
```
weiznich: We should somehow provide this via something that's injected from the actual connection, so that only `mariadb://` is allowed for the mariadb connection and only `mysql://` is allowed for the mysql connection type

diesel/src/mariadb/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+/// `mariadb://[user[:password]@]host/database_name[?unix_socket=socket-path&ssl_mode=SSL_MODE*&ssl_ca=/etc/ssl/certs/ca-certificates.crt&ssl_cert=/etc/ssl/certs/client-cert.crt&ssl_key=/etc/ssl/certs/client-key.crt]`
+///
+///* `host` can be an IP address or a hostname. If it is set to `localhost`, a connection
+///   will be attempted through the socket at `/tmp/mariadb.sock`. If you want to connect to
+///   a local server via TCP (e.g. docker containers), use `0.0.0.0` or `127.0.0.1` instead.
+/// * `unix_socket` expects the path to the unix socket
+/// * `ssl_ca` accepts a path to the system's certificate roots
+/// * `ssl_cert` accepts a path to the client's certificate file
+/// * `ssl_key` accepts a path to the client's private key file
+/// * `ssl_mode` expects a value defined for MySQL client command option `--ssl-mode`
```
weiznich: It seems like this is currently broken with mariadb: https://github.com/diesel-rs/diesel/discussions/5111

diesel_cli/src/infer_schema_internals/mariadb.rs
weiznich, line 1: It would be good to not duplicate this file but rather reuse the existing mysql implementation for most queries.

diesel_cli/src/migrations/diff_schema/enum_type_query_fragments.rs
```diff
@@ -45,6 +45,26 @@ impl QueryFragment<diesel::mysql::Mysql> for EnumType<'_> {
     }
 }
 
+#[cfg(feature = "mariadb")]
```
weiznich, line 40: Ideally we could reuse the mysql impls here as well, but that's not super critical

diesel_cli/src/print_schema.rs
```diff
@@ -512,6 +512,19 @@ fn mysql_diesel_types() -> HashSet<&'static str> {
...
+#[cfg(feature = "mariadb")]
+fn mariadb_diesel_types() -> HashSet<&'static str> {
+    let mut types = HashSet::new();
+    common_diesel_types(&mut types);
+
+    types.insert("TinyInt");
+    types.insert("Tinyint");
+    types.insert("Datetime");
+    types.insert("Json");
+    types.insert("Uuid");
```
weiznich: That is not supported yet, so maybe do not hardcode that yet?

diesel_cli/src/print_schema.rs
```diff
@@ -1184,6 +1214,95 @@ impl Display for CustomTypesForTablesForDisplay<'_> {
                     writeln!(out, "pub struct {};", custom_type.rust_name)?;
                 }
 
+                writeln!(f, "}}\n")?;
+                rust_enum_module(f, rust_types, self.custom_types.generate_rust_enums)?;
+                Ok(())
+            }
+            #[cfg(feature = "mariadb")]
+            Backend::Mariadb => {
```
weiznich: If somehow possible I rather would like to share that code with the mysql branch above. It's sufficiently long and complicated that it otherwise becomes problematic.

diesel_derives/src/tests/snapshots/diesel_derives__tests__sql_type_1 (mysql).snap
```diff
@@ -16,4 +16,9 @@ const _: () = {
             diesel::mysql::MysqlType::Long
         }
     }
+    diesel::internal::derives::sql_type::expand_mariadb! {
```
weiznich: This requires a bit of special handling code to also expand this macro. See here: https://github.com/diesel-rs/diesel/blob/main/diesel_derives/src/tests/mod.rs#L117

weiznich (commented):
Thanks for working on this. It's really appropriated to see some movement on that topic :heart:

I did a high level pass over the changes and put some general comments here and there. The overall approach is close to what I described in #1882, so that looks good. I had hoped that we could avoid a bit more code duplication but it seems like that's sometimes not possible?

There is one remaining topic that's not covered by the PR yet: Automatic testing. We need to have a CI setup that covers this backend as well, so there is also a bit work required there.

If there are anything we can help you with just let me know and I try to get around.

diesel/src/mysql_like/connection/stmt/iterator.rs
jraffeiner: i moved the shared types and impls to a `myslq_like` module,referenced and reexported the types so that the public api should remain unchanged

diesel/src/mysql/connection/url.rs
jraffeiner: i added the Backend as a type parameter on the parse function, the trait `MysqlLikeBackend` provides the scheme constant against which it will be checked. "mysql" for `Mysql` and "mariadb" for `Mariadb`

diesel_cli/src/migrations/diff_schema/enum_type_query_fragments.rs
jraffeiner: I tried to use the trait bound `B: MysqlLikeBackend` here, but then the compiler complains: multiple implementations for `Pg`, upstream crate could implement `MysqlLikeBackend` for `Pg`.

diesel_cli/src/print_schema.rs
jraffeiner: I could intend to implement the support for it within this PR. I only need to know if i should move the declaration of the `sql_type::Uuid` out of the `pg` module directly to the `sql_type` module (or some other place) or leave it there and make it reachable via `#cfg[...]`

diesel_derives/src/tests/snapshots/diesel_derives__tests__sql_type_1 (mysql).snap
jraffeiner: i added the necessary changes to the visitor. It should work now as expected.

diesel_cli/src/print_schema.rs
weiznich: I would like to keep this PR focused on adding the connection and the backend with as minimal changes as possible. It's already a large change and therefore harder to review. 
If that one landed it is certainly welcome to add support for all the additional features supported by MariaDB in smaller more targeted PR's.

diesel/src/mariadb/backend.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl TypeMetadata for Mariadb {
+    type TypeMetadata = MariadbType;
+
+    type MetadataLookup = ();
+}
+
+impl SqlDialect for Mariadb {
+    type ReturningClause = sql_dialect::returning_clause::PgLikeReturningClause;
```
LucaCappelletti94: MariaDB's supports `RETURNING` on `DELETE`, `INSERT` and `REPLACE`, but **not** `UPDATE`

With `PgLikeReturningClause`, `UPDATE ... RETURNING` compiles, and it really should not.

See: https://jira.mariadb.org/browse/MDEV-5092

LucaCappelletti94 (changes requested):

diesel_cli/src/infer_schema_internals/mariadb.rs
jraffeiner: i moved as much as possible into a shared module `mysql_like` now only the query for the foreign key constraints, and the tests live in the `mysql` and `mariadb` modules. To enable this i made the trait `MysqlLikeBackend` public

diesel_cli/src/print_schema.rs
jraffeiner: i extracted it into a function which both backends can use

diesel/src/mariadb/backend.rs
jraffeiner: After some further testing i decided to to leave the `ReturningClause` as `DoesNotSupportReturningClause` and then add that feature in a follow up PR

diesel/src/mysql_like/query_builder/limit_offset.rs
```diff
@@ -44,13 +46,13 @@ impl QueryFragment<Mysql> for BoxedLimitOffsetClause<'_, Mysql> {
                 limit.walk_ast(out.reborrow())?;
             }
             (None, Some(offset)) => {
-                // Mysql requires a limit clause in front of any offset clause
+                // B requires a limit clause in front of any offset clause
                 // The documentation proposes the following:
                 // > To retrieve all rows from a certain offset up to the end of the
                 // > result set, you can use some large number for the second parameter.
-                // https://dev.mysql.com/doc/refman/8.0/en/select.html
+                // https://dev.B.com/doc/refman/8.0/en/select.html
```
LucaCappelletti94: I think this is a general replace error

diesel/src/mysql_like/query_builder/query_fragment_impls.rs
```diff
@@ -123,7 +129,7 @@ where
 }
 
 /// A marker type signaling that the given `ON CONFLICT` clause
-/// uses mysql's `ON DUPLICATE KEY` syntax that triggers on
+/// uses B's `ON DUPLICATE KEY` syntax that triggers on
```
LucaCappelletti94: Also here it looks like some general replacement error

LucaCappelletti94 (changes requested):
Just a preliminary round, I found what I believe are replace-all type of errors. Please do correct them, and check whether I missed any other of them.

jraffeiner:
I took a look at some failing CI-Tests, and saw that with the trait based implementation of `QueryFragment` it degrades the compiler diagnostics i.e.: `QueryFragment<...>` is not implemented for `Sqlite` turns into `MysqlLikeBackend` is not implemented for `Sqlite`

Should i remove the trait based implementation und replace it with identical implementation for each backend `Mysql` & `Mariadb` or is there another solution for this?
And if yes keep these implementations in the common `mysql_like` module or move them into the corresponding `mysql` and `mariadb` modules?

As for the automatic testing should i cover it in this PR or in a follow up PR?

weiznich:
> I took a look at some failing CI-Tests, and saw that with the trait based implementation of QueryFragment it degrades the compiler diagnostics i.e.: QueryFragment<...> is not implemented for Sqlite turns into MysqlLikeBackend is not implemented for Sqlite

You might be able to fix that by adding `#[diagnostic::do_not_recommend]` to the relevant impls. That needs a bit playing around to get it right.

diesel_cli/src/migrations/diff_schema/enum_type_query_fragments.rs
jraffeiner: I extracted the impls into functions and call the functions in the impls

diesel/src/mysql_like/query_builder/limit_offset.rs
jraffeiner: these were the only replace-all errors fixed them

diesel/src/expression/functions/window_functions.rs
```diff
@@ -367,7 +391,7 @@ extern "SQL" {
     /// # include!("../../doctest_setup.rs");
     /// # use diesel::dsl::*;
     /// #
-    /// # #[cfg(not(feature = "mysql"))] // mariadb doesn't support this variant
+    /// # #[cfg(not(any(feature = "mysql", feature = "mariadb")))] // mariadb doesn't support this variant
```
weiznich: This might require some additional testing. The comment indicates that this variant might be supported by mysql but not by mariadb. If that's the case we should change the attributes?

diesel/src/expression/functions/window_functions.rs
```diff
@@ -570,7 +606,7 @@ extern "SQL" {
...
```
weiznich: Same here as well

diesel/src/mariadb/backend.rs
```diff
@@ -0,0 +1,71 @@
+use crate::{
+    backend::{Backend, DieselReserveSpecialization, SqlDialect, TrustedBackend, sql_dialect},
+    mariadb::{MariadbQueryBuilder, MariadbValue},
+    mysql_like::MysqlLikeBackend,
+    query_builder::bind_collector::RawBytesBindCollector,
+    sql_types::TypeMetadata,
+};
```
weiznich: Diesel uses non-nested grouping of the imports. I would strongly prefer keeping that for new modules as well.

(Unfortunately that cannot enforced by rustfmt as it requires a nightly feature…)

diesel/src/mariadb/connection.rs
```diff
@@ -0,0 +1,155 @@
...
+/// `mariadb://[user[:password]@]host/database_name[?unix_socket=socket-path&ssl_mode=SSL_MODE*&ssl_ca=/etc/ssl/certs/ca-certificates.crt&ssl_cert=/etc/ssl/certs/client-cert.crt&ssl_key=/etc/ssl/certs/client-key.crt]`
+///
+///* `host` can be an IP address or a hostname. If it is set to `localhost`, a connection
+///  will be attempted through the socket at `/tmp/mysql.sock`. If you want to connect to
+///  a local server via TCP (e.g. docker containers), use `0.0.0.0` or `127.0.0.1` instead.
+/// * `unix_socket` expects the path to the unix socket
+/// * `ssl_ca` accepts a path to the system's certificate roots
+/// * `ssl_cert` accepts a path to the client's certificate file
+/// * `ssl_key` accepts a path to the client's private key file
+/// * `ssl_mode` expects a value defined for MySQL client command option `--ssl-mode`
```
weiznich: I would prefer removing these options for now for mariadb. As written before, they are currently broken and need a slightly different setup. We can fix that in a followup PR as soon as the basics are in place.

It's fine to just drop it from the documentation without changing the underlying implementation

diesel/src/mariadb/mod.rs
```diff
@@ -0,0 +1,27 @@
+//! Provides types and functions related to working with MariaDB
+//!
+//! Much of this module is re-exported from database agnostic locations.
+//! However, if you are writing code specifically to extend Diesel on
+//! MariaDB, you may need to work with this module directly.
+pub(crate) mod backend;
+
+//pub(crate) mod query_builder;
```
weiznich: Commented out code should be removed

diesel/src/mysql/connection.rs
```diff
@@ -0,0 +1,155 @@
+use crate::{mysql::Mysql, mysql_like::MysqlLikeConnection};
```
weiznich: Another nested import

diesel/src/mysql/mod.rs
```diff
@@ -7,28 +7,14 @@
...
-pub mod data_types {
-    #[doc(inline)]
-    pub use super::types::date_and_time::{MysqlTime, MysqlTimestampType};
-}
-
-/// MySQL specific sql types
-pub mod sql_types {
-    #[doc(inline)]
-    pub use super::types::{Datetime, Unsigned};
-}
```
weiznich, line 34: Removing these and just reexporting them from `diesel::mysql` is a breaking change. That needs to be changed back

diesel/src/mysql_like/connection/stmt/iterator.rs
```diff
@@ -223,14 +233,14 @@ impl<'a> Field<'a, Mysql> for MysqlField<'a> {
...
-    fn value(&self) -> Option<<Mysql as Backend>::RawValue<'_>> {
+    fn value(&self) -> Option<<B as Backend>::RawValue<'_>> {
         match &*self.binds {
             PrivateMysqlRow::Copied(b) | PrivateMysqlRow::Direct(b) => b[self.idx].value(),
         }
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "mysql"))]
```
weiznich: Is there any reason to only run these tests for mysql? If so please add a comment in the code why that's the case

diesel/src/mysql_like/connection/bind.rs
```diff
@@ -822,12 +822,12 @@ fn known_buffer_size_for_ffi_type(tpe: ffi::enum_field_types) -> Option<usize> {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "mysql"))]
```
weiznich: Is there any reason to only run these tests for mysql. It feels like we should also run them for mariadb as that is a source of previous incompatibilities.

diesel/src/mysql_like/connection/url.rs
```diff
@@ -224,24 +225,25 @@ fn connection_url_error() -> ConnectionError {
...
-        assert!(ConnectionOptions::parse("file:///tmp/mysql.sock").is_err());
-        assert!(ConnectionOptions::parse("socket:///tmp/mysql.sock").is_err());
-        assert!(ConnectionOptions::parse("mysql://localhost?database=somedb").is_err());
-        assert!(ConnectionOptions::parse("mysql://localhost").is_ok());
+        assert!(ConnectionOptions::parse::<Mysql>("postgres://localhost").is_err());
+        assert!(ConnectionOptions::parse::<Mysql>("http://localhost").is_err());
+        assert!(ConnectionOptions::parse::<Mysql>("file:///tmp/mysql.sock").is_err());
+        assert!(ConnectionOptions::parse::<Mysql>("socket:///tmp/mysql.sock").is_err());
+        assert!(ConnectionOptions::parse::<Mysql>("mysql://localhost?database=somedb").is_err());
+        assert!(ConnectionOptions::parse::<Mysql>("mysql://localhost").is_ok());
```
weiznich: Maybe add a few mariadb cases as well, showing that e.g parsing `mariadb://` fails for the mysql backend and vice versa?

diesel/src/mysql_like/query_builder/mod.rs
```diff
@@ -8,21 +10,30 @@ pub use self::query_fragment_impls::DuplicatedKeys;
 mod limit_offset;
 mod query_fragment_impls;
 
-/// The MySQL query builder
+/// The MySQL-Like query builder
 #[allow(missing_debug_implementations)]
-#[derive(Default)]
-pub struct MysqlQueryBuilder {
```
weiznich, line 15: That's a public type: https://docs.diesel.rs/main/diesel/mysql/struct.MysqlQueryBuilder.html

So just adding a new generic type is a breaking change. We need to make sure:

* The generic type has a default type (So `B: Backend = Mysql`) for the case that the `mysql` feature is enabled
* Both `default()` and `new()` can be called without adding a generic parameter for the case that the `mysql` feature is enabled. (Likely meaning that they are only implemented for `MysqlQueryBuilder<Mysql>`)

diesel/src/mysql_like/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+mod value;
+
+use core::hash::Hash;
+
+use crate::{
+    backend::{Backend, DieselReserveSpecialization},
+    mysql_like::query_builder::MysqlLikeQueryBuilder,
+    query_builder::bind_collector::RawBytesBindCollector,
+    sql_types::TypeMetadata,
+};
```
weiznich: Another nested import

diesel/src/query_builder/combination_clause.rs
```diff
@@ -390,6 +390,28 @@ mod postgres {
...
+            self.inner.walk_ast(out.reborrow())?;
+            out.push_sql(")");
+            Ok(())
+        }
+    }
+
+    impl<B: MysqlLikeBackend> SupportsCombinationClause<Union, Distinct> for B {}
+    impl<B: MysqlLikeBackend> SupportsCombinationClause<Union, All> for B {}
+}
+/*
```
weiznich: Commented out code that should be removed

diesel/src/lib.rs
```diff
@@ -334,8 +334,14 @@ pub mod sql_types;
 pub mod migration;
 pub mod row;
 
+#[cfg(any(feature = "mysql_backend", feature = "mariadb_backend"))]
+pub mod mysql_like;
```
weiznich, line 345: This shouldn't be public as far as I can tell? Only the reexports from the `mysql` and `mariadb` module needs to be public

diesel/Cargo.toml
```diff
@@ -84,6 +84,7 @@ postgres = ["dep:pq-sys", "postgres_backend"]
 sqlite-no-std = ["__sqlite-shared", "hashbrown"]
 sqlite = ["__sqlite-shared", "std"]
 mysql = ["dep:mysqlclient-sys", "dep:url", "dep:percent-encoding", "dep:bitflags", "mysql_backend"]
+mariadb = ["dep:mysqlclient-sys", "dep:url", "dep:percent-encoding", "dep:bitflags", "mariadb_backend"]
```
weiznich: Maybe introduce a shared `__mysql_like_shared` feature for the shared feature flags instead?

Otherwise we also need documentation in `lib.rs` for any new public feature

diesel_cli/src/infer_schema_internals/mysql.rs
```diff
@@ -210,94 +69,13 @@ pub fn load_foreign_key_constraints(
...
-        )
-    } else {
-        None
-    }
-}
+#[cfg(test)]
+use super::mysql_like::{
+    determine_column_type, determine_type_name, determine_unsigned, get_enum_variants,
+    get_table_comment, get_table_data,
+};
```
weiznich, line 76: Nested imports

diesel_cli/Cargo.toml
```diff
@@ -80,13 +80,15 @@ regex = "1.3.9"
...
-default = ["postgres", "sqlite", "mysql"]
+default = ["postgres", "sqlite", "mysql", "mariadb"]
 postgres = ["diesel/postgres", "uses_information_schema"]
 sqlite = ["diesel/sqlite", "diesel/std"]
 mysql = ["diesel/mysql", "uses_information_schema"]
+mariadb = ["diesel/mariadb", "uses_information_schema"]
 sqlite-bundled = ["sqlite", "dep:libsqlite3-sys", "libsqlite3-sys/bundled"]
 postgres-bundled = ["postgres", "dep:pq-sys", "pq-sys/bundled", "dep:openssl-sys"]
 mysql-bundled = ["mysql", "dep:mysqlclient-sys", "mysqlclient-sys/bundled", "dep:openssl-sys"]
+mariadb-bundled = ["mariadb", "dep:mysqlclient-sys", "mysqlclient-sys/bundled", "dep:openssl-sys"]
```
weiznich: I would rather not support this particular feature yet as this statically links `libmysqlclient` instead of `libmariadb`

To be clear: This "works", but I'm not sure if that's what users would expect here

diesel_derives/src/tests/sql_type.rs
```diff
@@ -37,6 +41,27 @@ pub(crate) fn sql_type_1() {
...
+pub(crate) fn sql_type_mysql_type_for_mariadb() {
+    let input = quote::quote! {#[diesel(mysql_type(name = "Long"))]};
+
+    let input = quote::quote! {
+        #input
+        struct Integer;
+    };
+
+    // When `mariadb_type` is absent, but `mysql_type` is present, we should still generate the same code for MariaDB.
+    let name = "sql_type_1 (mariadb)";
```
weiznich: I would rather prefer not doing this as Mysql might add support for types not supported by Mariadb. I rather would prefer having these attributes everywhere to be explicit.

diesel_tests/tests/errors.rs
```diff
@@ -225,7 +225,7 @@ fn not_null_constraints_correct_column_name() {
 }
 
 #[diesel_test_helper::test]
-#[cfg(not(feature = "mysql"))]
+#[cfg(not(any(feature = "mysql", feature = "mariadb")))] //ERROR_CODES Unterscheiden sich -> MariaDB 4022
```
weiznich: Would be good to have an English comment here.

weiznich (changes requested):
Thanks for all the work you already put into this. This looks like a good step forward.

I sat down and went over everything in detail. Most of the comments are rather minor nits. It still would be great to have them addressed. I hope that I flagged all the nested imports, otherwise please apply this to other locations as well.

The largest remaining blocker is to get a CI setup to automatically run the mariadb tests against a mariadb instance in CI.

diesel/src/mysql_like/query_builder/mod.rs
jraffeiner: I renamed the type to MysqlLikeQueryBuilder<...> and implemented the Default trait for it.

And reexported the type `MysqlLikeQueryBuilder<Mysql>`  under the type alias `MysqlQueryBuilder` from the `mysql` module 

So this shouldn't be a breaking change.

diesel/src/lib.rs
jraffeiner: It's needed public to get access to the trait `MysqlLikeBackend` and the generic type `MysqlLikeConnection<...>` to allow a generic implementation for both in external crates i.e. diesel_cli

diesel/src/expression/functions/window_functions.rs
jraffeiner: i tested these with both backends and adjusted the attributes
mysql succeeds, mariadb fails

diesel/src/mariadb/backend.rs
jraffeiner: i unnested them all

diesel/src/mariadb/connection.rs
jraffeiner: i droped it from the documentation, but this is a client difference, it depends on  which client library is used, independently from the backend... 
The possible fix mentioned in https://github.com/diesel-rs/diesel/discussions/5111 seems like a viable solution...

diesel/src/mariadb/mod.rs
jraffeiner: removed the commented out code, seems i overlooked these locations before

diesel/src/mysql/mod.rs
jraffeiner: reexported the modules from `diesel::mysql` and `diesel::mariadb`

diesel/src/mysql_like/connection/stmt/iterator.rs
jraffeiner: that was just me being lazy, i changed the test so they run for both

diesel/src/mysql_like/connection/url.rs
jraffeiner: changed the tests so they run for both and added the same schema test for mariadb

diesel/Cargo.toml
jraffeiner: i added the documentation for the features `mariadb` and `mariadb_backend`, and moved the shared feature flags into the shared `__mysql_like_shared`

diesel_cli/src/infer_schema_internals/mysql.rs
jraffeiner: these are grouped? should i split them into a row per item?

diesel_cli/Cargo.toml
jraffeiner: i removed the feature

diesel_derives/src/tests/sql_type.rs
jraffeiner: i added `#[diesel(mariadb_type(..))]` to all the places where `mysql_type` is used and removed the fall back to mysql_type from the derive for mariadb_type

diesel_tests/tests/errors.rs
jraffeiner: i added a map_error_number function for `Mariadb` and `Mysql` backend, and made the statement generic over the backend, so that the error number can be mapped differently between both backends, as the comment mentioned

jraffeiner:
I tried to add CI for Mariadb hopefully up to standard, and for that added minimal support for the `mariadb` backend to `diesel_bench`.

jraffeiner:
Setting up the CI for mariadb backend, i think i hit https://github.com/diesel-rs/diesel/discussions/5091 
I found two possible solutions for this.
-Solution 1: as implemented save the length, before the `stmt_bind_result` call and set the length values to it afterwards, as mentioned in https://github.com/diesel-rs/diesel/discussions/5091#discussioncomment-17443319
-Solution 2: mark the binds as invalid after `populate_dynamic_buffers`, check if the binds are valid before the `mysql_stmt_fetch` call, and rebind them with `stmt_bind_result` if theyre marked as invalid

It looks like `libmariadb` expects `stmt_bind_result` to only be called before `mysql_stmt_fetch`.

I tested both approaches locally with both libraries and it seems to work with both libs `libmysqlclient` and `libmaria`

diesel/src/mysql_like/connection/bind.rs
```diff
@@ -87,8 +90,21 @@ impl OutputBinds {
                 }
             }
         }
-
-        unsafe { self.with_mysql_binds(|bind_ptr| stmt.bind_result(bind_ptr)) }
+        /*
```
weiznich: I would rather go with the solution described here: https://github.com/diesel-rs/diesel/pull/5147#issuecomment-5182190387

That doesn't require an additional allocation, just an additional integer field in the bind data struct, so it should be a bit cheaper

diesel/src/mysql_like/query_builder/batch_update.rs
```diff
@@ -7,27 +8,24 @@ use crate::result::EmptyChangeset;
 use crate::result::Error::QueryBuilderError;
 use crate::{QueryResult, Table};
 
-impl<C, T> BatchAssignHelper<crate::mysql::Mysql> for Assign<ColumnWrapperForUpdate<C>, T>
+impl<B: MysqlLikeBackend, C, T> BatchAssignHelper<B> for Assign<ColumnWrapperForUpdate<C>, T>
```
weiznich: We generally use `DB` as generic identifier for the backend generic. It would be great to change that here and in other similar locations as well

diesel/src/mysql_like/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    Self: TypeMetadata<TypeMetadata = MysqlType, MetadataLookup = ()>,
+    Self: Backend<QueryBuilder = MysqlLikeQueryBuilder<Self>>,
+    Self: Hash + Eq + Default,
+    Self: DieselReserveSpecialization,
+    Self: MapErrorNumber,
+    Self: 'static,
+{
+    /// The scheme used in the connection URL for this backend.
+    /// "mysql" for MySQL, "mariadb" for MariaDB.
+    const SCHEME: &'static str;
```
weiznich, line 60: Make this `#[doc(hidden)]`. It doesn't need to be part of the public API

diesel/src/lib.rs
```diff
@@ -177,6 +177,8 @@
 //!   mysqlclient-sys = { version = "0.5", features = ["bundled"] }
 //!   openssl-sys = { version = "0.9.100", features = ["vendored"] }
 //!   ```
+//! - `mariadb`: This feature enables the diesel mariadb backend. This feature implies `mariadb_backend`.
+//!   Enabling this feature requires a compatible copy of `libmysqlclient` for your target architecture.
```
weiznich:
```suggestion
//!   Enabling this feature requires a compatible copy of `libmysqlclient` or `libmariadb` for your target architecture.
```

diesel/src/mariadb/connection.rs
weiznich: Yes, this is more a client issue. It's just that chances are very high that people that use mariadb also installed libmariadb instead of libmysqlclient, especially given that most linux distributions just ship mariadb and not mysql anymore, even if you install mysql.

diesel/src/mysql/mod.rs
weiznich: Should be fine now, but I want to do a `cargo semver-check` run in the next days to really make sure we don't break anything. (At least as far as that tool can detect such issues)

diesel/src/lib.rs
weiznich: After giving this another thought I believe you are correct here with making this module public. It's valueable for downstream users to have this. I still want to verify what exactly is exposed in the next days.

diesel_cli/src/infer_schema_internals/mysql.rs
weiznich: Seems like I saw something different there. My mistake

weiznich (commented):
Thanks for your work here. This already looks quite good. We are nearly there. The CI setup looks good to me.

I just have some stylistic requests and a different approach how to workaround #5091. Hopefully that's not too much to ask for

diesel/src/mysql_like/connection/bind.rs
jraffeiner: What do you think of the following approach:

- Add a bool `binds_are_invalid` to `OutputBinds`
- Mark the binds as invalid in `populate_dynamic_buffers` with it -> set `binds_are_invalid`
- Check `binds_are_invalid` before calling `mysql_stmt_fetch`
- And if `binds_are_invalid` rebind the binds with `mysql_stmt_bind_result` and reset the bool

Because it seems [CONC-821](https://jira.mariadb.org/projects/CONC/issues/CONC-821?filter=allopenissues) was only referencing the Time types(because they're length isn't set in `mysql_stmt_fetch`), so it appears they don't consider it wrong setting the length of dynamic sized types to 0 in `mysql_stmt_bind_result`. And reading through the Api docs of both Mysql and Mariadb, there is no mention of the length being preserved through a `mysql_stmt_bind_result` call. (They could even decide to zero out the buffers and it wouldn't break any Api promises).
The only guarantee about the length field that i found is, that it is guaranteed to be set to the correct value after `mysql_stmt_fetch`. (this didn't uphold for Time types -> [CONC-821](https://jira.mariadb.org/projects/CONC/issues/CONC-821?filter=allopenissues))

diesel/src/mysql/mod.rs
jraffeiner: `cargo semver-checks` is likely to give a false positive https://github.com/obi1kenobi/cargo-semver-checks/issues/611

diesel/src/mysql_like/mod.rs
jraffeiner: added the `#[doc(hidden)]` attribute

diesel/src/mysql_like/query_builder/batch_update.rs
jraffeiner: Replaced B, with DB in backend generics

diesel/src/mysql_like/connection/bind.rs
weiznich: I spend some time looking into this and it seems like this is the better solution. Let's go with that for now, we nevertheless want to look for a more robust solution for connecting to a mysql/mariadb database in the future. Maybe using a pure rust connection or whatever.

weiznich (approved):
The changes look now fine.

I still will do a verification that this doesn't break the public API, also I will likely push a few minor additional things on top (changelog, readme, etc)

jraffeiner:
Thanks, for reviewing it, and i'm happy to be able to contribute to this crate

weiznich:
I've pushed some minor cleanup, if the CI passes this time I would like to merge this if you don't have any objections.

> Thanks, for reviewing it, and i'm happy to be able to contribute to this crate

That would be very welcome. We are in the need of a regular contributor that actually uses the Mysql implementation in "production"

jraffeiner:
I'm happy to see this getting merged. 
Is it fine if I open a draft, as soon as this is merged to add support for the `RETURNING` clause to the mariadb backend?
Because that will require a bit of work, especially if we want to support the returning clause only for `INSERT`, `DELETE` and `REPLACE`.
Since Mariadb added support for `UPDATE ... RETURNING` only recently in 13.0, which has no stable release yet.

weiznich:
> Is it fine if I open a draft, as soon as this is merged to add support for the RETURNING clause to the mariadb backend?

Yes, it's always fine to open draft PR's to coordinate new features. The worst thing that could happen is that I or someone else from the core team says: We don't want to have this feature. (That's definitely not the case for adding support for `RETURNING` for mariadb)

> Since Mariadb added support for UPDATE ... RETURNING only recently in 13.0, which has no stable release yet.

Depending on their release timeline it might be fine to just accept that it requires 13.0 to use `update` + `returning`. We should document this then, but it's not a problem on it's own that certain constructs require newer server versions that others.

PR #5145, opened 2026-08-03
Add `SqliteConnection::incremental_vacuum`

LucaCappelletti94:
`PRAGMA incremental_vacuum` hands freelist pages back to the file without the full rewrite `VACUUM` does. It needs a helper because it is easy to run wrongly: SQLite frees one page per step of the statement, so it only empties the freelist when run to completion. Diesel's prepared statement path steps once, since `StatementUse::run` calls `step` a single time, so one page comes back, the call returns success, and the rest of the freelist silently stays. This uses `batch_execute`, which runs `sqlite3_exec` and steps to the end. Two of the added tests fail if that is swapped back.

A pragma takes no bind parameters, so the schema goes through `SqliteQueryBuilder` and `push_identifier` rather than being interpolated.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -1504,6 +1504,55 @@ impl SqliteConnection {
         query.get_result(self)
     }
 
+    /// Shrink a database by releasing freelist pages, without the full rewrite
+    /// [`VACUUM`](https://www.sqlite.org/lang_vacuum.html) performs.
+    ///
+    /// `schema` selects an attached database by name, `None` targets `main`. `pages`
+    /// bounds how many pages are reclaimed. As SQLite specifies, `None` or a value
+    /// below one clears the whole freelist, as does a bound larger than it. Only
+    /// databases in [`AutoVacuumMode::Incremental`] have anything to reclaim, on any
+    /// other mode this succeeds and does nothing.
+    ///
+    /// ```rust
+    /// # include!("../../doctest_setup.rs");
+    /// # fn main() {
+    /// #     run_test().unwrap();
+    /// # }
+    /// # fn run_test() -> QueryResult<()> {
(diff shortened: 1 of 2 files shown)


diesel/src/sqlite/connection/mod.rs
```diff
@@ -4028,4 +4077,146 @@ mod tests {
         assert!(conn.page_count(Some("nope")).is_err());
         assert!(conn.freelist_count(Some("nope")).is_err());
     }
+
```
weiznich, line 4080: Semi related note: This file is now again becoming really large. We should find a way to split it up to stay under 2k lines of code. Maybe by moving all the pragmas to their own file as well?

That's nothing that needs to be done as part of this PR, but a followup with just the moves would be very welcome

diesel/src/sqlite/connection/mod.rs
```diff
@@ -4028,4 +4077,146 @@ mod tests {
...
+            .values(rows)
+            .execute(conn)
+            .unwrap();
+        crate::delete(pragma_probe::table).execute(conn).unwrap();
+    }
+
+    // The same, in an attached schema.
+    fn grow_then_empty_aux_freelist(conn: &mut SqliteConnection) {
+        conn.batch_execute(
+            "CREATE TABLE aux.aux_pragma_probe (id INTEGER PRIMARY KEY, payload TEXT NOT NULL)",
```
weiznich: Might be worth to put that into a constant as well?

weiznich (approved):
Looks good to me beside the minor nit about the potential table definition that could be shared by several tests

diesel/src/sqlite/connection/mod.rs
LucaCappelletti94: Always happy to modularize further. Will do after the last few PRs that should land in this document get merged, if that is okay.

PR #5147, opened 2026-08-03
Preserve output bind lengths across the post-truncation rebind in MySQL backend

stevenwcarter:
I suddenly had 26 failing tests for one of my projects after upgrading my local version of the mariadb libs using Diesel+MySQL. The root cause appears to be MariaDB's `mysql_stmt_bind_result` zeroing `*length` at bind time, which destroys the lengths which were just fetched, resulting in empty strings being returned for the first result from my queries. In my research, I found https://jira.mariadb.org/projects/CONC/issues/CONC-821?filter=allopenissues which confirms this is a long-standing issue and was resolved ~3 weeks ago in their master branch.

Fixes #5091 

Note that the existing CI cannot exercise the breaking behavior, since it is only using `libmysqlclient-dev` on Ubuntu using Oracle's code, not the mariadb libraries.

While I understand the discussion about not wanting an interim fix for a MariaDB-specific breakage, the fix seems small and well-contained. Additionally, it would allow versions of Diesel with this fix in place to support several existing versions of the mariadb libs for users who are on slower-releasing distros with older versions of mariadb which don't already carry the fix.

- [x] I checked for similar changes and made sure to reference them

Disclaimer: AI was used to research this issue, identify the root-cause, and write the change (Fable model). I have run the relevant tests and included a test-case which failed on my local MariaDB libs before the change and was green afterward. All code has been reviewed and verified by a human.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Tighten requirements for `SqliteConnection::deserialize_readonly_database` to closely match the upstream requirements
 * `diesel print-schema` now generates `joinable!` and `allow_tables_to_appear_in_same_query!` for PostgreSQL foreign keys across multiple configured schemas
 * Fixed `FromSql`/`ToSql` for `chrono::NaiveTime` on MySQL dropping the fractional-seconds component of `TIME` values, so a `TIME(N)` column now round-trips its microseconds like `DATETIME` / `TIMESTAMP` already did.
+* Fixed the MySQL backend returning empty strings for every variable-length column in the first row of a result set when diesel is linked against MariaDB Connector/C (whose `mysql_stmt_bind_result` zeroes the caller's length values at bind time, destroying the lengths of the row that was just fetched during diesel's post-truncation rebind)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


weiznich:
Thanks for submitting this PR. As already written in #5901: I would rather prefer seeing that fixed by the upstream lib than in diesel. This is a rather hacky way to work around that issue and their change turns the interface into something that's rather unpredictable. You really should get an information about the buffer size back from these function calls.

---
That written: Even if we end up working around that issue in diesel, this is in my opinion not the correct way to fix the issue. The correct way would be:

* Add a new field to https://github.com/diesel-rs/diesel/blob/8cbfd04d1e8be6a0c70ba2d31b216e10565784ae/diesel/src/mysql/connection/bind.rs#L179 to track the length outside of libmariadb/libmysql via an `Option` field
* For types with a statically known size initialize the field with the statically known size.
* For dynamically sized types set the size here: https://github.com/diesel-rs/diesel/blob/8cbfd04d1e8be6a0c70ba2d31b216e10565784ae/diesel/src/mysql/connection/bind.rs#L542
* Finally prefer that "outside" length value to the `length` field for all cases it is set here: https://github.com/diesel-rs/diesel/blob/8cbfd04d1e8be6a0c70ba2d31b216e10565784ae/diesel/src/mysql/connection/bind.rs#L542. If it was not initialized yet fall back tho the `length` field.

stevenwcarter:
Sure, I understand completely. I'm using this patch locally until the mariadb release filters through the repos, which will hopefully be soon. Thanks for taking the time to look.

PR #5148, opened 2026-08-05
add initial type mapping support for sqlite only

arthmis:
- [ ] I included a changelog entry for relevant new features or changes

This adds basic type mapping support to sqlite. This is an implementation of this proposal https://github.com/diesel-rs/diesel/discussions/4113#discussioncomment-10021444. The issue was referenced in these conversations https://github.com/diesel-rs/diesel/discussions/4192.

This only provides support for Sqlite as a starting point because, one I needed it, and two Sqlite is a lot more permissive with "types" so this would be a good starting point. Also wanted to gather feedback before making anymore broader changes to the other backends. This supports configuring output schema type for specific table, column, or schema or combination of them.

One major question I had was if it was worth it to add a CLI flag for this feature? Or if it is a requirement that there is a CLI flag for all configuration options. I opted to only have this work if using `diesel.toml` because it's easier to write the table of fields that way.

### DIFF
--- a/diesel_cli/src/config.rs
+++ b/diesel_cli/src/config.rs
@@ -521,6 +521,17 @@ pub struct PrintSchema {
     pub custom_enum_derives: Option<BTreeSet<String>>,
     #[serde(default)]
     pub generate_rust_enum_definitions: Option<bool>,
+    #[serde(default)]
+    pub type_map: Option<Vec<SchemaTypeMap>>,
+}
+
+#[derive(Default, Deserialize, Clone, Debug)]
+pub struct SchemaTypeMap {
+    pub database_typename: String,
+    pub schema_typename: String,
+    pub table_name: Option<String>,
+    pub schema_name: Option<String>,
+    pub column_name: Option<String>,
 }
 
 impl PrintSchema {
--- a/diesel_cli/src/infer_schema_internals/sqlite.rs
+++ b/diesel_cli/src/infer_schema_internals/sqlite.rs
@@ -367,6 +367,48 @@ pub fn get_primary_keys(
     Ok(collected)
 }
 
+pub fn map_column_type(
+    type_name: &str,
+    table: &TableName,
+    attr: &ColumnInformation,
+    config: &PrintSchema,
+) -> Option<ColumnType> {
+    let Some(type_mappings) = &config.type_map else {
+        return None;
+    };
+
+    let Some(type_map) = type_mappings.iter().find(|map| {
+        map.database_typename.to_lowercase() == type_name
+            && map
+                .table_name
+                .as_deref()
+                .is_none_or(|t| t == table.sql_name)
+            && map
+                .schema_name
+                .as_deref()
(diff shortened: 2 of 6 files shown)


weiznich:
Thanks for working on this. I did not have the capacity to review the changes this week, hopefully I find some time next week.

> One major question I had was if it was worth it to add a CLI flag for this feature? Or if it is a requirement that there is a CLI flag for all configuration options. 

If it's meaningful possible to support this as CLI flags as well, I would prefer having the CLI flag.

arthmis:
> Thanks for working on this. I did not have the capacity to review the changes this week, hopefully I find some time next week.

No problem!

> If it's meaningful possible to support this as CLI flags as well, I would prefer having the CLI flag.

I'll see what I can do there then.

diesel_cli/src/infer_schema_internals/sqlite.rs
```diff
@@ -367,6 +367,48 @@ pub fn get_primary_keys(
...
+                .schema_name
+                .as_deref()
+                .is_none_or(|s| Some(s) == table.schema.as_deref())
+            && map
+                .column_name
+                .as_deref()
+                .is_none_or(|c| c == attr.column_name)
+    }) else {
+        return None;
+    };
```
weiznich, line 396: I would suggest to have this rather as function on `SchemaTypeMap` instead so that it can be shared between all backends.

Something like:

```
impl SchemaTypeMap {
    fn map_type(&self, db_type: &str, table_name: &str, schema_name: Option<&str>, column_name: &str) -> &str {…}
}
```

diesel_cli/src/config.rs
```diff
@@ -521,6 +521,17 @@ pub struct PrintSchema {
...
+}
+
+#[derive(Default, Deserialize, Clone, Debug)]
+pub struct SchemaTypeMap {
+    pub database_typename: String,
+    pub schema_typename: String,
+    pub table_name: Option<String>,
+    pub schema_name: Option<String>,
+    pub column_name: Option<String>,
 }
```
weiznich, line 535: Matching on these is currently not implemented, would be great to have that, although that's not a requirement for the initial PR to land.

If you decide not to implement this, I would rather drop these fields for now. We always can add them later.

.zed/debug.json
weiznich, line 1: This file shouldn't be there

weiznich (changes requested):
The basic approach looks right to me. I would strongly prefer having this for all supported backends. I would be fine if the initial implementation only supports renaming all instances of a type, as we always can extend this later to be more sophisticated.

diesel_cli/tests/print_schema.rs
```diff
@@ -619,6 +619,60 @@ fn print_schema_rust_injection() {
     )
 }
 
+#[test]
+#[cfg(feature = "sqlite")]
+fn print_schema_type_mapping() {
```
weiznich, line 624: We do want to have a `generate_migration` test as well, as we need to reverse that mapping there

PR #5150, opened 2026-08-07
Add `SqliteConnection::wal_checkpoint`

LucaCappelletti94:
Diesel's own SQLite examples run WAL checkpoints as strings, which throws away the pragma's result, i.e. whether a busy reader blocked the checkpoint and how many log frames were moved back into the database file. Those two values are the difference between a WAL that stays bounded and one that silently grows because a long-lived reader pins it, and reading them today needs a hand-rolled `sql_query` row type.

This adds `SqliteConnection::wal_checkpoint(schema, mode)` with a `WalCheckpointMode` enum and a `WalCheckpointOutcome` result. Frame counts come back as `None` on a database not in WAL mode (SQLite reports `-1` there), so the call is safe to issue unconditionally.

A `None` schema **does not** fall back to main, because SQLite defines the unqualified pragma to checkpoint every attached database.

I went for `PRAGMA` instead of `sqlite3_wal_checkpoint_v2`, I could not find any reason to go for either option in particular so I just picked the one I have been using most often for these use cases, but if you have any particular opinion on the matter let me know.

In SQLite 3.51.0 (November 2025) added a fifth checkpoint mode, `NOOP`, which reports the WAL state without checkpointing anything. It seemed way too recent, so I left it out for now.

### DIFF
--- a/diesel/src/sqlite/connection/mod.rs
+++ b/diesel/src/sqlite/connection/mod.rs
@@ -166,6 +166,7 @@ use core::num::NonZeroI64;
 /// # fn run_test() -> QueryResult<()> {
 /// #     use schema::users;
 /// use diesel::connection::SimpleConnection;
+/// use diesel::sqlite::WalCheckpointMode;
 /// let conn = &mut establish_connection();
 /// // see https://fractaledmind.github.io/2023/09/07/enhancing-rails-sqlite-fine-tuning/
 /// // sleep if the database is busy, this corresponds to up to 2 seconds sleeping time.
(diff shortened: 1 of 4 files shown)


diesel/src/sqlite/connection/mod.rs
```diff
@@ -405,6 +406,43 @@ pub enum AutoVacuumMode {
...
+    Passive,
+    /// Wait until there is no writer and every reader reads from the most
+    /// recent snapshot, then checkpoint every frame.
+    Full,
+    /// Like [`Full`](Self::Full), then wait until no reader uses the WAL, so
+    /// the next writer restarts the log.
+    Restart,
+    /// Like [`Restart`](Self::Restart), then truncate the WAL file to zero
+    /// bytes.
+    Truncate,
```
weiznich, line 425: I think it's fine to support `NOOP` here as well, as users can choose on their own if they want to use that mode or not. Just make sure that the documentation states that this requires a recent sqlite version

LucaCappelletti94: Done

weiznich (approved):
Looks good, thanks for submitting.

(I will take care of the merge conflict soon)

PR #5154, opened 2026-08-17
Add lockfile auto-delete

watsom27:
From #4800

Auto delete `.diesel_lock` file via a `LockFile` wrapper with a `Drop` impl.

- [ ] I included a changelog entry for relevant new features or changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * The minimal supported Rust version is now 1.88.0
 * Add support for no-std environments using the SQLite backend
 * Improved documentation and added examples for `filter_target` on `IncompleteOnConflict`
+* .diesel_lock_ file now auto deletes after use
 
 ## [2.3.12] 2026-08-07
 
--- a/diesel_cli/src/migrations/lock_file.rs
+++ b/diesel_cli/src/migrations/lock_file.rs
@@ -1,6 +1,5 @@
 use std::{
     fs::{File, remove_file},
-    os::fd::AsFd,
     path::PathBuf,
 };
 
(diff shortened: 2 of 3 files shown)


LucaCappelletti94 (changes requested):
Deleting `.diesel_lock` in `Drop` is unsafe under contention. A waiting process can acquire the lock on the unlinked inode while another process recreates the path and acquires a separate lock, allowing concurrent migration generation.

Also, `let _ = lock.write()?;` drops the guard immediately, so the migration creation is not actually protected. I would suggest to keep the lock file persistent and empty and bind the guard to a named variable for the full critical section.

Mingun:
It seems to me that deleting lock file on failure is a bad practice. I believe, all other software using lock files does that another way -- the file is deleted only on successful termination. That allow you to see that something was going wrong and at the same time you may manually delete the lock file if you are sure that consequences are eliminated or unimportant.

watsom27:
Given that the lockfile looks to only be "used" by the `migration generate` command, what contention are we actually protecting against? Someone running the same command in two terminals at once? 

Also given that they lock file isn't actually being used as @LucaCappelletti94 points out, perhaps it is not needed at all and could be removed?

Locking was added in https://github.com/diesel-rs/diesel/pull/4410

LucaCappelletti94:
@watsom27 I didn't say the lock file isn't being used, and I wasn't suggesting it could be removed. My point was that it is used incorrectly. `let _` does not bind, so the write guard returned by `lock.write()` is dropped at the end of that statement and the lock is released before the migration is actually generated, and therefore we should bind the guard to a named variable (for example `let _guard = ...`) so it stays held for the whole generation step.

Furthermore, what I had proposed was to keep the lock file persistent instead of deleting it on drop, because unlinking lets one process lock the old inode while another recreates the path, which hands the lock to two processes at once.

watsom27:
@LucaCappelletti94 
> I wasn't suggesting it could be removed

I know, but I was 😄 

What I am trying to get at is "what exactly does this lockfile protect against"?

Given that: its not implemented correctly right now (and therefore the current version of Diesel is, in effect, acting as if there is no lock at all), there (at first glance) doesn't seem to be any open issues regarding a race condition here, and the race conditions seems to be only triggerable by manually running `migration generate` in two terminals at once; I am questioning if the lock is needed at all.

Perhaps the locking system was an AI hallucination in the original PR?

If there is a genuine need for the lockfile then by all means, i'll close the PR. I'm just trying to get my head around why its needed, and also why it should stick around in my filesystem.

weiznich:
The main use case for the lock was to prevent concurrent runs of the migration generate command. Yes that's an edge case that shouldn't happen but we decided to rather ensure this back then when the locking mechanism was implemented. Now it's a bit unfortunate that this never worked, I should have reviewed it better 😞 
That said, I rather would like to see it fixed than removed. That als includes the concurrent runs with delete scenario. 
I also would like to see at least one test ensuring that the lock file is actually removed for the simple non concurrent case.

Issue #3754, opened 2023-08-18
Raw "SELECT ?" sql query results in a panic for the mysql backend


weiznich:
### Versions

- **Rust:** 1.71
- **Diesel:** 2.1.0
- **Database:** mysql
- **Operating System:** fedora 

### Feature Flags

- **diesel:** mysql

## Problem Description

Raw select queries with binds as select clause fail at runtime with the following error:

```
thread 'multiconnection::nullable_type_checks' panicked at 'internal error: entered unreachable code: We ensure at the call side that we do not hit this type here. If you ever see this error, something has gone very wrong. Please open an issue at the diesel github repo in this case', diesel/src/mysql/connection/bind.rs:685:50
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:64:14
   2: diesel::mysql::connection::bind::<impl core::convert::From<(mysqlclient_sys::enum_field_types,diesel::mysql::connection::bind::Flags)> for diesel::mysql::backend::MysqlType>::from
             at /home/weiznich/Documents/rust/diesel/diesel/src/mysql/connection/bind.rs:685:50
   3: <T as core::convert::Into<U>>::into
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/convert/mod.rs:726:9
   4: diesel::mysql::connection::bind::BindData::value
             at /home/weiznich/Documents/rust/diesel/diesel/src/mysql/connection/bind.rs:410:23
   5: <diesel::mysql::connection::stmt::iterator::MysqlField as diesel::row::Field<diesel::mysql::backend::Mysql>>::value
             at /home/weiznich/Documents/rust/diesel/diesel/src/mysql/connection/stmt/iterator.rs:223:72
```

### What are you trying to accomplish?

```
let mut conn = MysqlConnection::establish("…").unwrap();

#[derive(QueryableByName)]
struct Res {
    #[diesel(sql_type = sql_types::Integer)]
    bar: i32
}

diesel::sql_query("SELECT ? as bar").bind::<sql_types::Integer, i32>(1).get_result::<Res>(&mut conn);

```

### What is the expected output?

The query runs successful 

### What is the actual output?

[...]

PR #5155, opened 2026-08-17
Add Support for Returning in Mariadb

jraffeiner:
## Description
Adds support for `RETURNING` to the `mariadb` backend.

This is a simple and naive implemention for it.

## Implementation Details
- Added type `MariadbReturningClause`
- Implemented `SupportsReturningClause` for it
- Implemented `QueryFragment` for it
- support for the case when metadata is only provided after the execute

## Testing & Validation
- Enabled tests using `INSERT ... RETURNING` and `DELETE ... RETURNING` for `mariadb` backend
- Enabled tests using `UPDATE ... RETURNING` for `mariadb` backend, these tests will check with a query if the server version >= 13.0, since it is only supported since 13.0

## Outstanding TODOs
- Mariadb support `OLD_VALUE(column)` syntax for `UPDATE ... RETURNING`
    - Should i reuse the `diesel::pg::returning::old` for it? with an alias as `old_value`?
    - or should i add a new `diesel::mariadb::returning::old_value` for it? 
- At the moment there is a comment on the `mariadb` backend noting, that `UPDATE ... RETURNING` needs version >= 13.0
    - is that the right place?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,9 @@ jobs:
           - rust: "nightly"
             backend: "mariadb"
             os: "ubuntu-latest"
+          - rust: "stable"
+            backend: "mariadb-13"
+            os: "windows-2025"
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout sources
@@ -283,6 +286,24 @@ jobs:
           echo "C:\Program Files\MariaDB 12.3\lib" >> $GITHUB_PATH
           echo "C:\Program Files\MariaDB 12.3\bin" >> $GITHUB_PATH
           dir "C:\Program Files\MariaDB 12.3\lib"
+
+      - name: Install mariadb-13 (Windows)
+        if: runner.os == 'Windows' && matrix.backend == 'mariadb-13'
+        shell: bash
+        run: |
+          choco install mariadb --pre -y
+          "C:\\Program Files\\MariaDB 13.0\\bin\\mariadb" -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'root'@'localhost';" -u root
+          echo "OPENSSL_RUST_USE_NASM=0" >> $GITHUB_ENV
+          echo OPENSSL_SRC_PERL=C:/Strawberry/perl/bin/perl >> $GITHUB_ENV
+          echo "MARIADB_DATABASE_URL=mariadb://root@127.0.0.1/diesel_test" >> $GITHUB_ENV
+          echo "MARIADB_EXAMPLE_DATABASE_URL=mariadb://root@127.0.0.1/diesel_example" >> $GITHUB_ENV
+          echo "MARIADB_UNIT_TEST_DATABASE_URL=mariadb://root@127.0.0.1/diesel_unit_test" >> $GITHUB_ENV
+          echo "MYSQLCLIENT_LIB_DIR=C:\Program Files\MariaDB 13.0\lib" >> $GITHUB_ENV
+          echo "MYSQLCLIENT_LIBNAME=libmariadb" >> $GITHUB_ENV
+          echo "MYSQLCLIENT_VERSION=3.4.10" >> $GITHUB_ENV
+          echo "C:\Program Files\MariaDB 13.0\lib" >> $GITHUB_PATH
+          echo "C:\Program Files\MariaDB 13.0\bin" >> $GITHUB_PATH
+          dir "C:\Program Files\MariaDB 13.0\lib"
 
       - name: Rust version check
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
             os: "ubuntu-latest"
           - rust: "stable"
             backend: "mariadb-13"
-            os: "windows-2025"
+            os: "ubuntu-latest"
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout sources
@@ -131,6 +131,10 @@ jobs:
         if: runner.os == 'Linux' && matrix.backend == 'mariadb'
         run: |
           sudo apt-get -o DPkg::Lock::Timeout=300 update
+          curl -LsSO https://r.mariadb.com/downloads/mariadb_repo_setup
+          chmod +x mariadb_repo_setup
+          sudo ./mariadb_repo_setup
+          sudo apt-get -o DPkg::Lock::Timeout=300 update
           sudo apt-get -o DPkg::Lock::Timeout=300 install -y libmariadb-dev libmariadb-dev-compat mariadb-server mariadb-client
           sudo systemctl start mariadb.service
           sudo mariadb -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'root'@'localhost' IDENTIFIED BY 'root';" -uroot
@@ -138,6 +142,19 @@ jobs:
           echo "MARIADB_EXAMPLE_DATABASE_URL=maridb://root:root@127.0.0.1/diesel_example" >> $GITHUB_ENV
           echo "MARIADB_UNIT_TEST_DATABASE_URL=mariadb://root:root@127.0.0.1/diesel_unit_test" >> $GITHUB_ENV
           mariadb_config --cc_version
+          
+      - name: Install mariadb-13 (Linux)
+        if: runner.os == 'Linux' && matrix.backend == 'mariadb-13'
+        run: |
+          sudo apt-get -o DPkg::Lock::Timeout=300 update
+          sudo apt-get -o DPkg::Lock::Timeout=300 install -y libmariadb-dev libmariadb-dev-compat mariadb-server mariadb-client
+          sudo systemctl start mariadb.service
+          sudo mariadb -e "create database diesel_test; create database diesel_unit_test; grant all on \`diesel_%\`.* to 'root'@'localhost' IDENTIFIED BY 'root';" -uroot
+          echo "MARIADB_DATABASE_URL=mariadb://root:root@127.0.0.1/diesel_test" >> $GITHUB_ENV
+          echo "MARIADB_EXAMPLE_DATABASE_URL=maridb://root:root@127.0.0.1/diesel_example" >> $GITHUB_ENV
+          echo "MARIADB_UNIT_TEST_DATABASE_URL=mariadb://root:root@127.0.0.1/diesel_unit_test" >> $GITHUB_ENV
+          mariadb_config --cc_version
+          mariadb -version
 
       - name: Install postgres (MacOS)
         if: runner.os == 'macOS' && matrix.backend == 'postgres'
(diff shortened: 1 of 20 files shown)


jraffeiner:
This can maybe close https://github.com/diesel-rs/diesel/issues/2535, or should there be some additional Documentation?

diesel/src/mariadb/backend.rs
```diff
@@ -9,6 +10,8 @@ use crate::result::DatabaseErrorKind;
 use crate::sql_types::TypeMetadata;
 
 /// The MariaDB backend
+///
+/// the `UPDATE ... RETURNING` clause is only supported by Mariadb >= 13.0
```
LucaCappelletti94: 13 is extremely recent, just from March 2026, and the previous versions are to be supported (in theory) until 2029, maybe we could add a generic `version` to the MariaDB backend, with marker structs for the various versions, which allow to define specifically which property is supported by which version? Otherwise the current approach would enable at compile time queries that previous MariaDB should not support.

diesel/src/mysql_like/connection/stmt/mod.rs
```diff
@@ -76,14 +76,11 @@ impl<DB: MysqlLikeBackend> Statement<DB> {
             .into_owned()
     }
 
-    pub(super) fn metadata(&self) -> QueryResult<StatementMetadata> {
-        use crate::result::Error::DeserializationError;
-
+    //we return a Option here because mariadb sometimes doesn't provide the statement metadata before the execute
```
LucaCappelletti94: I would suggest to use normal docstrings

LucaCappelletti94 (changes requested):

diesel/src/mariadb/backend.rs
jraffeiner: True, but that would apply to the other backends too. 
And i'm not sure where to start to limit the implemention of the `RETURNING` clause to `INSERT`, `DELETE` and `REPLACE`.
Do i need to remove the `SupportsReturningClause` impl from `MariadbReturningClause`?
Would i then need to impl `QueryFragment` explicitly for `INSERT`,`DELETE` and `REPLACE` with the returning clause?

weiznich: We do not guarantee this level of compile time checking for other features. It's totally possible to write queries that uses newer features for the PostgreSQL and SQLite backend es well. The only difference there is that it's often not such a often used feature like `RETURNING` clauses, but just some sql functions or copy options. Nevertheless it might happen. 
For SQLite we had a similar problem some years ago when they introduced general support for returning clauses. We put it behind a off by default crate feature back then. I somewhat consider that a mistake now given that most users now use a SQLite version that supports this feature, so it's a speed bump as those users now need to enable it. We cannot remove this feature flag, although we maybe could enable it by default. 

Given that the MariaDB limitation ist "just" it doesn't work for updates for older versions it sounds ok for me to just document it and otherwise say: The relevant  version will be the new baseline in a few years.

weiznich:
> Mariadb support OLD_VALUE(column) syntax for UPDATE ... RETURNING

The relevant Postgres feature was added by https://github.com/diesel-rs/diesel/pull/5049 recently, so it's not in a release yet. Given that it sounds reasonable to just move it to a shared location. I think that PR already includes some idea for that. 

Let's do this in a follow up PR, as smaller PR's are much simpler and faster to review

diesel/src/mariadb/backend.rs
jraffeiner: So its fine as is, or should it be documented somewhere else?

weiznich: I would rather move it to the documentation of `diesel::update`

diesel/src/mysql_like/connection/stmt/iterator.rs
```diff
@@ -23,12 +23,21 @@ impl<'a, DB: MysqlLikeBackend> StatementIterator<'a, DB> {
         stmt: MaybeCached<'a, Statement<DB>>,
         types: &[Option<MysqlType>],
     ) -> QueryResult<Self> {
-        let metadata = stmt.metadata()?;
+        let (metadata, output_binds, mut stmt) = if let Some(metadata) = stmt.metadata()? {
```
LucaCappelletti94: `Statement::metadata` returning an `Option` moves the "No metadata exists" failure from before `mysql_stmt_execute` to after it. `sql_query("UPDATE ...").get_result::<T>(conn)`, where the caller meant `.execute`, used to fail with nothing run. Now the statement executes (and commits under autocommit) before returning the same error.

Only MariaDB needs the execute-first fallback because missing metadata at prepare time is ambiguous there between "no result set" and "`RETURNING` result set with deferred metadata". MySQL always reports metadata at prepare time when a result set exists, so on MySQL the fallback only fires in the mistake case.

I believe we could gate the fallback with an associated `const METADATA_MAY_REQUIRE_EXECUTE: bool` on `MysqlLikeBackend` (`false` for `Mysql`, `true` for `Mariadb`), so the branch compiles out on MySQL.

LucaCappelletti94 (changes requested):

diesel/src/mysql_like/connection/stmt/iterator.rs
weiznich: It's unfortunately not as simple as that as this also happens for mysql: https://github.com/diesel-rs/diesel/issues/3754

jraffeiner: Honestly https://github.com/diesel-rs/diesel/issues/3754 is a bit of a different case, at least for mysql the prepare does report a metadata but of another type: `MYSQL_TYPE_NULL`, after the execute the metadata changes to the correct type.

I have a fix for https://github.com/diesel-rs/diesel/issues/3754 prepared. I can include that fix in this PR, if you want. As it touches only this part of the code.

I'm open to implement it, however you prefer. With a bool to execute the statement, without metadata only if the backend is `mariadb`, or as it is for both backends.

LucaCappelletti94: Ok, it was more cursed than I had understood (and hoped). I believe that the fallback as written may not cover #3754, since there the prepare-time metadata exists but reports `MYSQL_TYPE_NULL` for the placeholder column, so that query would then take the fast path and panic. I suppose that covering it would require reading the metadata after an unconditional execute.

LucaCappelletti94: @jraffeiner the two things seem somewhat similar to me but distinct enough that two different PRs may be the best approach to reduce the cognitive load (even only to really illustrate how they cover different things)

diesel_tests/tests/schema/mod.rs
```diff
@@ -554,3 +554,21 @@ pub fn pg_server_supports_returning_old(connection: &mut TestConnection) -> bool
...
         >= 180000
 }
+
+/// `UPDATE ... RETURNING` was introduced in Mariadb 13.0; on older servers
+/// the query will be rejected at execution time, so we just skip.
+///
+/// Returns `true` if the connected Mariadb server is version 13.0 or newer.
+/// On non-mariadb backends this always returns `false`.
+#[cfg(feature = "mariadb")]
+pub fn mariadb_server_supports_update_returning(connection: &mut TestConnection) -> bool {
```
LucaCappelletti94, line 564: At this time, since the CI only ever runs < 13, this might as well be a `return false`. Maybe we should add a MariaDB 13 job so the feature is actually covered in CI?

diesel_tests/tests/insert.rs
```diff
@@ -425,11 +425,7 @@ fn insert_record_attached_database_using_returning_clause() {
 }
 
 #[diesel_test_helper::test]
-#[cfg(not(any(
-    not(feature = "returning_clauses_for_sqlite_3_35"),
-    feature = "mysql",
-    feature = "mariadb"
-)))]
+#[cfg(not(any(not(feature = "returning_clauses_for_sqlite_3_35"), feature = "mysql")))]
```
LucaCappelletti94: I believe this test is meant to run for MariaDB, but:

* `not(feature = "returning_clauses_for_sqlite_3_35")` is true for MariaDB
* any(true, false) is true
* not true is false

so the test is currently not running for MariaDB

diesel/src/mysql_like/mod.rs
```diff
@@ -56,8 +56,15 @@ where
 {
     #[doc(hidden)]
     /// The scheme used in the connection URL for this backend.
-    /// "mysql" for MySQL, "mariadb" for MariaDB.
+    /// mysql for MySQL, "mariadb" for MariaDB.
```
LucaCappelletti94: Stray edit dropped the quotes I believe

LucaCappelletti94 (changes requested):

diesel_tests/tests/insert.rs
jraffeiner: This test probably should be run on `postgres`  too...
I probably should change the cfg to the same configuration as in `diesel_tests/tests/delete.rs`

diesel_tests/tests/schema/mod.rs
jraffeiner: Would it be enough to add one platform with v13 in CI? Or multiple platforms?

jraffeiner: I added one job for Windows running Mariadb 13.0, since that was the simplest platform to configure for it

diesel/src/mysql_like/mod.rs
jraffeiner: readded the dropped quotes

.github/workflows/ci.yml
```diff
@@ -70,6 +70,9 @@ jobs:
           - rust: "nightly"
             backend: "mariadb"
             os: "ubuntu-latest"
+          - rust: "stable"
+            backend: "mariadb-13"
+            os: "windows-2025"
```
weiznich: I rather would prefer running this job on one of the ubuntu runners. The windows CI job takes like 15 minutes, while the ubuntu CI job completes in 8 minutes.

That written: We don't perform this sort of version range testing for other backends yet, so I personally would also be fine with just not doing it to save some CI time

diesel/src/mysql_like/connection/stmt/iterator.rs
weiznich: To be honest: I'm rather unhappy with this solution as it makes now the behaviour diverge between the MariaDB backend and the Mysql backend for no particular good reason. Thinking a bit more about this it rather feels like it is totally reasonable to first execute the statement and then fetch the actual metadata instead of the current solution.

Maybe we can just adjust `execute_statement` to first call `mysql_stmt_execute`, then fetch the metadata, then create the `OutputBinds` and finally call `mysql_stmt_store_result`?

For the case @LucaCappelletti94 pointed out earlier: Yes that changes, but I would argue that this code is already broken (it already returns an error) and we don't guarantee a particular behaviour there.

diesel_tests/tests/schema/mod.rs
LucaCappelletti94: Do note that in practice the windows VM in the CI is the slowest of them all, if you want a fast crash upon edit the linux one might be a much better candidate. I am not sure how difficult it would be to set it up in CI though.

weiznich (changes requested):
Sorry for just having the time only now to look into the various points in detail and raising different opinions after you adjusted the code already several times. 

I still would like to make sure the behaviour there is somewhat consistent and sane.

diesel/src/mysql_like/connection/stmt/iterator.rs
jraffeiner: > 
> Maybe we can just adjust `execute_statement` to first call `mysql_stmt_execute`, then fetch the metadata, then create the `OutputBinds` and finally call `mysql_stmt_store_result`?
> 
```       
          unsafe {
                ffi::mysql_stmt_execute(self.stmt.as_ptr());
          }
        
        self.did_an_error_occur()?;
        let ret = StatementUse { inner: self };
        let metadata = ret.metadata()?;
        let mut binds = OutputBinds::from_output_types(types, &metadata).map_err(DeserializationError)?;
        binds.bind_results(&ret)?;
        unsafe {
            ffi::mysql_stmt_store_result(ret.inner.stmt.as_ptr());
        }
        ret.inner.did_an_error_occur()?;
        Ok((metadata, binds, ret))
```

I tried it like this and it works for the most part.

But it seems we hit some memory bug within `libmysqlclient`.
The two tests `diesel_cli::tests print_schema::print_schema_allows_none` and `diesel_cli::tests print_schema::print_schema_fk_related_tables` start to fail most of the time on some form of `glibc` error. ie:
```
Fatal glibc error: malloc.c:4512 (_int_malloc): assertion failed: (unsigned long) (size) >= (unsigned long) (nb)
```
or
```
corrupted size vs. prev_size
```
and a segmentation fault on test `diesel_tests::integration_tests order::order_by_column`.

With `libmariadb` there seems to be no issue, as all tests pass. 

This approach is very similar to my fix for #3754, so i'll include that for now, it looks cleaner then the actual implementation too.

.github/workflows/ci.yml
jraffeiner: I moved it to ubuntu

jraffeiner: It seems there is no reliable way to get mariadb-13 on ubuntu, so i removed it from ci.

weiznich (commented):
Looks good to me

weiznich (approved):

PR #5160, opened 2026-08-21
fix #661

### LINKED ISSUE #661: MySQL: Figure out how to handle tests that modify schema
Our test suite for Postgres and SQLite assumes we can modify the db schema inside a transaction. This does not work on MySQL (it commits the transaction when you try to modify the schema).

We need to figure out a way to have those tests run on MySQL without introducing a bunch of overhead.

(To find affected tests, grep for `FIXME: Figure out how to handle tests that modify schema`)

jraffeiner:
This fixes #661 and enables tests for `mysql` and `mariadb` which where blocked by it.

### DIFF
--- a/diesel_tests/tests/insert.rs
+++ b/diesel_tests/tests/insert.rs
@@ -669,19 +669,15 @@ fn insert_in_nullable_with_non_null_default() {
 fn insert_returning_count_returns_number_of_rows_inserted() {
     use crate::schema::users::table as users;
     let connection = &mut connection();
-    #[cfg(any(feature = "mysql", feature = "mariadb"))]
-    let query = "CREATE TEMPORARY TABLE users (
+    diesel::sql_query(
+        "CREATE TEMPORARY TABLE users (
         id SERIAL PRIMARY KEY,
         name VARCHAR(50) NOT NULL,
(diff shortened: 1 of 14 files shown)


diesel_tests/tests/insert.rs
```diff
@@ -674,20 +666,22 @@ fn insert_in_nullable_with_non_null_default() {
...
 
 #[diesel_test_helper::test]
-#[cfg(not(any(feature = "mysql", feature = "mariadb")))] // FIXME: Figure out how to handle tests that modify schema
 fn insert_returning_count_returns_number_of_rows_inserted() {
     use crate::schema::users::table as users;
     let connection = &mut connection();
-    drop_table_cascade(connection, "users");
-    diesel::sql_query(
-        "CREATE TABLE users (
+    #[cfg(any(feature = "mysql", feature = "mariadb"))]
```
weiznich: It's fine to use the same table definition for all backends here. The differences do not matter that much.

weiznich (approved):
Thanks for this cleanup, this looks good beside the one minor nit about the duplication of that query

diesel_tests/tests/insert.rs
jraffeiner: I removed the duplicated statement, and used the table definition for `mysql` and `mariadb` for all backends

PR #5161, opened 2026-08-21
 Decode MySQL and MariaDB unsigned values according to the column's signedness

LucaCappelletti94:
`MysqlValue::numeric_value` folds each unsigned type code into its signed neighbour, and `BindData::for_output` prefers the server's metadata over the declared SQL type for every numeric pair, so a column declared `Integer` over a real `SMALLINT UNSIGNED` holding `40000` arrives as `-25536`. The `Tiny` arm additionally skips the length check its five siblings have, so an unsigned `TINYINT` above `127` fails and an empty buffer panics.

Now, each type code now gets its own arm and an error is returned in the case of overflows.


LucaCappelletti94:
I investigated the failure and it seems like a bad patch uphill, hopefully they will yank it soon.

diesel/src/mysql_like/types/mod.rs
```diff
@@ -268,3 +301,172 @@ impl<DB: MysqlLikeBackend> HasSqlType<Unsigned<BigInt>> for DB {
...
 pub struct Datetime;
+
+#[cfg(all(test, any(feature = "mysql", feature = "mariadb")))]
+mod tests {
+    use super::*;
+
+    #[cfg(feature = "mysql")]
+    type DB = crate::mysql::Mysql;
+    // Not an `any(..)` arm, so that enabling both backends stays a single alias.
+    #[cfg(all(feature = "mariadb", not(feature = "mysql")))]
```
weiznich: There is a `compile_error!` call in the crate root that prevents running tests with more than one backend enabled, so strictly speaking that's not necessary.

diesel/src/mysql_like/value.rs
```diff
@@ -106,21 +106,42 @@ impl<'a> MysqlValue<'a> {
...
+                self.too_short_buffer(1, "UnsignedTiny")?;
+                NumericRepresentation::UnsignedTiny(self.raw[0])
+            }
+            MysqlType::Short => {
                 self.too_short_buffer(2, "Short")?;
                 NumericRepresentation::Small(i16::from_ne_bytes((&self.raw[..2]).try_into()?))
             }
-            MysqlType::UnsignedLong | MysqlType::Long => {
+            MysqlType::UnsignedShort => {
+                self.too_short_buffer(2, "UnsignedShort")?;
```
weiznich: Looking at this it feels like `too_short_buffer` should just take a generic argument and then get the name and the size of the expected type via `std::mem::size_of::<T>` and `std::any::type_name::<T>` instead. That sounds a bit more robust.

diesel/src/mysql_like/value.rs
```diff
@@ -253,4 +282,75 @@ fn invalid_reads() {
...
+            .numeric_value()
+            .is_err()
+    );
+}
+
+#[test]
+fn numeric_value_keeps_signedness() {
+    use super::NumericRepresentation as N;
+
+    assert!(matches!(
```
weiznich, line 317: I think there is a `assert_matches!` macro in the standard library now

weiznich (commented):
This looks reasonable to me.

I just have a bunch of minor remarks

diesel/src/mysql_like/value.rs
LucaCappelletti94: It is stable since 1.96.0 (https://doc.rust-lang.org/std/macro.assert_matches.html), but the workspace minimum is 1.88.0.

diesel/src/mysql_like/value.rs
LucaCappelletti94: I went for the hopefully not too over-engineerred private sealed trait, so each arm is simply a `self.read()?`. It also now unifies the trailing-bytes handling, which differed between the narrow arms and the wide ones.

diesel/src/mysql_like/types/mod.rs
LucaCappelletti94: I could not find that `compile_error`, the only relevant one is the final `else` of the `cfg_if` that picks a test connection, which fires when *no* backend is enabled, and its own message suggests enabling several:

https://github.com/diesel-rs/diesel/blob/713424915ce91dd24b6df30d9cdb5d640163a7e3/diesel/src/test_helpers.rs#L63-L70

That being said, enabling multiple backends currently fails with several compile errors, but I believe that we can make the unit tests work also with several backends.

weiznich (approved):

PR #5163, opened 2026-08-25
Make `diesel-dynamic-schema` `no-std` compatible

LucaCappelletti94:
With this PR, also `diesel-dynamic-schema` can has `no_std` SQLite support.

<details>
image
</details>


weiznich (approved):
That's reasonable to add. Thanks for submitting

PR #5167, opened 2026-08-26
Use `PostgreSQL` array element OIDs when decoding elements

LucaCappelletti94:
Diesel was building each array element `PgValue` from the outer array lookup, so custom `FromSql` code that calls `PgValue::get_oid()` saw the array `OID`, for example `1007` for `int4[]`, instead of the element OID, for example `23` for `int4`.

This uses the element OID from the PostgreSQL array header for each decoded element, while keeping the existing fallback when the header OID is zero. The regression test covers both `Vec<T>` and `NdArray<T>`.

Found out while working on #5164


weiznich (approved):
Looks good to me

PR #5169, opened 2026-08-27
Add integration tests for SQLite json -> and ->> operators

zaveshaa:
Issue #4366 asked for integration coverage of the SQLite json `->` and `->>` operators. The operators exist but only doc-tests cover them.

These integration tests add:
- `->` (`retrieve_as_object_sqlite`) with a JSON path and an integer index, on `Json` and `Jsonb`
- `->>` (`retrieve_as_text`) with a JSON path and an array index, plus the NULL case for a missing path

Relates to #4366.

### DIFF
--- a/diesel_tests/tests/expressions/json_operators.rs
+++ b/diesel_tests/tests/expressions/json_operators.rs
@@ -1,6 +1,10 @@
+#[cfg(feature = "sqlite")]
 use crate::schema::connection;
+#[cfg(feature = "sqlite")]
 use diesel::dsl::sql;
+#[cfg(feature = "sqlite")]
 use diesel::prelude::*;
+#[cfg(feature = "sqlite")]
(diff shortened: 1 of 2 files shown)


weiznich:
Thanks for opening this PR. Could you explain the motivation to add these tests? For me it looks like that this behaviour is already tested by the doc-tests, so the tests look rather duplicated to me?

zaveshaa:
Good point. The doc-tests cover the basic behaviour of `->` and `->>` on `Json`, but they run in-memory and don't hit a real database with the JSON path/index variants. Issue #4366 asked for integration coverage since the doc-tests only exercise the simplest case.

That said, if you think the doc-tests already cover this well enough, I'm fine dropping the PR — no point adding redundant tests. Let me know which way you'd rather go.

PR #5170, opened 2026-08-27
Support negative `MySQL` and `MariaDB` `TIME` values

LucaCappelletti94:
`MySQL` and `MariaDB` use `TIME` for signed durations, but Diesel rejects every negative `MYSQL_TIME` before `MysqlTime` or custom decoders can inspect it.

This PR only adds support to the sign only for SQL Time, while `chrono::NaiveTime` and `time::Time` continue to return a deserialization error because neither type can represent a signed duration. Negative `Date`, `Datetime`, and `Timestamp` values remain rejected at the raw-value boundary.

**For this PR I use GPT Sol.**


LucaCappelletti94:
Filed https://github.com/Alorel/rust-indexed-db/pull/85 uphill so that the audit CI failure may be fixed, but I see that the last commit in that repo was last year. We'll see.

PR #5172, opened 2026-08-28
Fix stackoverflow while decoding jsonb in the sqlite backend

weiznich:
This commit rewrites our sqlite jsonb parser to not use a recursive approach. It now works iterative to avoid potential stackoverflows on deeply nested structures. 

We add also several tests to verify we don't introduce potential stackoverflows here again in the future.

I consider this a security related issue, but I don't think it's serve enough to don't publish this fix before releasing it. I also think we don't need an advisory for this, is it still requires a lot to happen in users code to trigger this. Especially the user would need to allow external actors to supply unchecked json, why on it's own is already an issue. The alternative is that they try to read databases from untrusted sources. I consider both scenarios as not very likely.

This issue was originally discovered by the RustFoundation based
scanning infrastructure created by their AI Security Engineer.


diesel/src/sqlite/types/json.rs
```diff
@@ -79,19 +82,128 @@ mod jsonb {
...
+                    }
+                    keys
+                } else {
+                    Vec::new()
+                };
+                if let Some((payload_bytes, payload_size)) = additional_data {
+                    byte_stack.push((payload_bytes, Some(payload_size), Some((keys, None))));
+                }
+            }
+            if let Some(payload_size) = target_payload_size
```
LucaCappelletti94: I was working on this too downhill, found via fuzzer that an object payload (`[0x2C, 0x17, b'a']`, i.e. object whose payload is a single `TEXT` `"a"`) ending on a pending key is now accepted silently and decodes to `Ok(({}, 3))` here, while main rejects it with `Err("Empty JSONB data")`.

The key parked in the `Option<String>` is never checked when the payload runs out. I imagine it should raise an error when parent still holds `Some(key)`.

diesel/src/sqlite/types/json.rs
```diff
@@ -79,19 +82,128 @@ mod jsonb {
...
+                    _ => Err(
+                        alloc::format!("Unsupported or reserved JSONB type: {element_type}").into(),
+                    ),
+                }?;
+                debug_assert!(
+                    (out.is_none() && parent.is_none()) || (out.is_some() && parent.is_some())
+                );
+                let out = out.get_or_insert_with(|| value.clone());
+
+                let keys = if let Some((parent_key, key)) = &mut parent {
```
LucaCappelletti94: This makes decoding quadratic in nesting depth, as every element clones the full path and re-walks it from the root, turning the crash into a hanging.

Depth | Blob size | Time
-- | -- | --
10k | — | 0.48 s
20k | — | 1.86 s
40k | 156 KB | 7.3 s

I think it could be refactored to be linear, walking the bytes sequential with a stack.

diesel/src/sqlite/types/json.rs
```diff
@@ -79,19 +82,128 @@ mod jsonb {
...
+    }
+
+    #[derive(Debug, Clone)]
+    enum JsonIndex {
+        Array(usize),
+        Object(String),
+    }
+
     // Helper function to read a JSONB value from the byte stream
     pub(super) fn read_jsonb_value(
```
LucaCappelletti94, line 97: Should we cap nesting depth instead of (or in addition to) decoding unboundedly? [`serde_json` refuses depth > 128](https://docs.rs/serde_json/latest/src/serde_json/de.rs.html#63) and SQLite's own JSON functions cap at [`JSON_MAX_DEPTH = 1000`](https://www.sqlite.org/compile.html#_options_to_set_default_parameter_values)

LucaCappelletti94 (changes requested):
I believe that the quadratic -> linear change is rather important as it was choking my fuzzer, the stronger recursion depth limit to 1k I think is reason given SQLite defaults (or even less). I will run the fuzzer more after the quadratic -> linear refactor, as right now it just gives me timeouts.

diesel/src/sqlite/types/json.rs
weiznich: I think that should be fixed now, the implementation is now a single linear scan over the input buffer

diesel/src/sqlite/types/json.rs
weiznich: There is a hard 2gb max blobsize limit for jsonb blobs. Hopefully that's + the last changes are enough to have this working without any depth limit. We only have one thing that's dependent on the depth now, which is the `stack` vector, which contains 2 ptr sized values per depth level. I would expect this to be fine.

If we still want to limit the max nesting depth, we would just need to check the length of that vector and abort after a certain depth is reached.

diesel/src/sqlite/types/json.rs
weiznich: I've added tests for this case as well.

---

If you have a fuzzer for any of the deserialization code (independently of the backend): Please consider upstreaming that as well. It is something that's on my list, but I did not have the capacity to cover it yet.

weiznich:
I pushed another update that also rewrites the writer to use an iterative approach as it also run into stackoverflows on certain platforms.

I also fixed another issue related to a potential numeric overflow while parsing a jsonb header for a malformed payload. This was reported by [aisle](https://aisle.com/)

cc @LucaCappelletti94, would be great if you could have another look

diesel/src/sqlite/types/json.rs
```diff
@@ -63,6 +66,9 @@ impl ToSql<sql_types::Jsonb, Sqlite> for serde_json::Value {
 mod jsonb {
     extern crate serde_json;
 
+    use core::error::Error;
+    use std::io::Write;
```
LucaCappelletti94: I believe this might break `no-std` build

diesel/src/sqlite/types/json.rs
```diff
@@ -1100,4 +1254,136 @@ mod tests {
...
+
+    #[diesel_test_helper::test]
+    fn object_key_without_value_results_in_error() {
+        let mut value = Vec::new();
+        value.extend(create_jsonb_header(JSONB_OBJECT, 2).unwrap());
+        value.extend(create_jsonb_header(JSONB_TEXT, 1).unwrap());
+        value.push(b'a');
+        let res = read_jsonb_value(&value);
+        assert!(res.is_err(), "{:?}", res.unwrap());
+    }
```
LucaCappelletti94, line 1374: The parser currently checks a child's length against the whole input, while I believe it should check against the end of the current parent. Here is a red test:

```suggestion
    }

    #[diesel_test_helper::test]
    fn nested_container_cannot_cross_parent_boundary() {
        let res = read_jsonb_value(&[0x3B, 0x1B, 0x1B, JSONB_NULL]);
        assert!(res.is_err(), "{:?}", res.unwrap());
    }
```

diesel/src/sqlite/types/json.rs
```diff
@@ -357,13 +434,36 @@ mod jsonb {
...
+                JsonValuePtr::Array {
+                    serialized_buffer, ..
+                }
+                | JsonValuePtr::Object {
+                    serialized_buffer, ..
+                },
+            ) => serialized_buffer,
+        };
+        write_jsonb_header(buffer, tpe, serialized_buffer.len())?;
+        buffer.extend(serialized_buffer);
```
LucaCappelletti94, line 464: Each completed nested buffer is copied into its parent, therefore a chain of depth `d` therefore copies about `1 + 2 + ... + d` bytes.

Maybe it could be worth to compute encoded sizes in an iterative first pass, then write every header and value once into the final buffer, using the depth limit as a separate safety bound.

diesel/src/sqlite/types/json.rs
```diff
@@ -1100,4 +1254,127 @@ mod tests {
...
+
+        #[cfg(all(target_family = "wasm", target_os = "none"))]
+        {
+            // on wasm we struggle with the recursion based
+            // overflows while dropping the values
+            // so we just leak them
+            core::mem::forget(res);
+            core::mem::forget(value);
+        }
+    }
```
LucaCappelletti94, line 1305: At `2000` levels, loading succeeds but freeing the nested object overflows the Wasm stack. The current CI fails here.

The test tries to keep the value in memory on Wasm, but its target check misses the CI target. Fixing only that check would leave applications exposed to the same failure.

SQLite defaults to `1000` levels, though `SQLITE_JSON_MAX_DEPTH` can change it. I tested a `1000`-level object on Diesel's `wasm32-unknown-unknown` runner, and that depth works, so I suggest to switch to `1000`.

```suggestion
    #[diesel_test_helper::test]
    fn depth_1000_is_safe_to_drop() {
        let wrap = |encoded: Vec<u8>| {
            let mut payload = create_jsonb_header(JSONB_TEXT, 1).unwrap();
            payload.push(b'k');
            payload.extend(encoded);
            let mut outer = create_jsonb_header(JSONB_OBJECT, payload.len()).unwrap();
            outer.extend(payload);
            outer
        };
        let mut encoded = vec![JSONB_NULL];
        for _ in 0..1000 {
            encoded = wrap(encoded);
        }
        drop(read_jsonb_value(&encoded).unwrap());
        assert!(read_jsonb_value(&wrap(encoded)).is_err());
    }
```

LucaCappelletti94 (changes requested):

diesel/src/sqlite/types/json.rs
weiznich: I have thought about that as well, but given that this is more a bug fix than an improvement and the old code did also allocate these intermediate buffers I went with the more straight forward solution here for now.

This is certainly something that can be improved in the future, although it's not totally trivial to write this without recursion and without code duplication. In fact the remaining `Write` thing above is an artifact of a quick try I gave that approach.

diesel/src/sqlite/types/json.rs
weiznich: I rather made the recursive depth a platform dependent constant to make sure that the value triggers the original issue on each platform (beside miri)

diesel/src/sqlite/types/json.rs
weiznich: Added the test and added a fix

diesel/src/sqlite/types/json.rs
LucaCappelletti94: `RECURSION_DEPTH` only limits the tests, so if `read_jsonb_value` still accepts arbitrary nesting, so a deeply nested value can decode successfully and then overflow the wasm stack when dropped. Maybe we should add a check also in the actual reader?

diesel/src/sqlite/types/json.rs
LucaCappelletti94: Ok, I do not have a better idea for the implementation at this time. Maybe we could leave a TODO and place it in a non-urgent open issue for beginners?

diesel/src/sqlite/types/json.rs
weiznich: Yes, that "only" fixes the diesel tests. The stackoverflow on drop is a `serde_json` problem from my point of view. I quickly tried to fix it there by implementing `Drop` for value and iteratively dropping values there, but that seems to be a breaking change. I still consider filling an issue for this anyway.

Finally it's not impossible for users to handle that on their own as demonstrated by the code example in https://docs.rs/serde_stacker/0.1.14/serde_stacker/

diesel/src/sqlite/types/json.rs
LucaCappelletti94: Ok on upstreaming the fuzzer, will clean it up and make a PR for it.

LucaCappelletti94: Ok so, before I was fuzzing diesel a bit indirectly via fuzzing my downhill crates, but now that I have started fuzzing directly diesel (will do a PR with the fuzzer soon) it turned blood bathy soon enough. Within the time it took from starting the new fuzzer and me opening GitHub to start working on the PR post, the damn thing already found 5-ish bugs. They look distinct, but I will have to investigate.

weiznich: I feared that, but its still better to discover such issues now than later.

You don't need to fix all of them yourself, just submit issues so that others can help with fixing.

LucaCappelletti94: I am spending time setting up (for the first time, I never did this) https://google.github.io/clusterfuzzlite/ since I believe we might get some free Google fuzzing given the size of diesel.

diesel/src/sqlite/types/json.rs
weiznich: Filled https://github.com/diesel-rs/diesel/issues/5192 for that

PR #5173, opened 2026-08-28
Fix code injections in diesel-cli

weiznich:
This commit adds various tests for potential code injection paths via database identifiers and fixes a potential issue with injections in type names.

I don't consider this a critical security issue, but merly something that should be fixed, given that the user always needs to actually compile the generated schema.

This issue was originally discovered by the RustFoundation based scanning infrastructure created by their AI Security Engineer.


CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ Increasing the minimal supported Rust version will always be coupled at least wi
 * Tighten requirements for `SqliteConnection::deserialize_readonly_database` to closely match the upstream requirements
 * `diesel print-schema` now generates `joinable!` and `allow_tables_to_appear_in_same_query!` for PostgreSQL foreign keys across multiple configured schemas
 * Fixed several Tests using schema modifications for `mysql` and `mariadb`
+* Fixed potential code injections in the generated schema.rs file based on malicous database identifiers
```
LucaCappelletti94:
```suggestion
* Fixed potential code injections in the generated schema.rs file caused by malicious database identifiers
```

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -115,23 +116,28 @@ pub fn determine_column_type(
...
-fn regclass(table: &TableName) -> Regclass<AsExprOf<String, sql_types::Text>> {
+fn regclass<'a, QS>(
+    table: &'a TableName,
+) -> Regclass<Box<dyn BoxableExpression<QS, Pg, SqlType = sql_types::Text> + 'a>>
+where
+{
     let table_name = match table.schema {
-        Some(ref schema_name) => format!("\"{}\".\"{}\"", schema_name, table.sql_name),
-        None => format!("\"{}\"", table.sql_name),
+        Some(ref schema_name) => Box::new(Box::new(
```
LucaCappelletti94: Why the `Box::new(Box::new(..))`? Isn't one Box enough?

diesel_cli/src/infer_schema_internals/pg.rs
```diff
@@ -115,23 +116,28 @@ pub fn determine_column_type(
 
 diesel::postfix_operator!(Regclass, "::regclass", sql_types::Oid, backend: Pg);
 
-fn regclass(table: &TableName) -> Regclass<AsExprOf<String, sql_types::Text>> {
+fn regclass<'a, QS>(
+    table: &'a TableName,
+) -> Regclass<Box<dyn BoxableExpression<QS, Pg, SqlType = sql_types::Text> + 'a>>
+where
+{
     let table_name = match table.schema {
```
LucaCappelletti94, line 124: I believe that quoting gives the same guarantee with the original signature, since doubling embedded quotes is what `quote_ident` does and `::regclass` reads it identically:

```rust
fn quoted(ident: &str) -> String { format!("\"{}\"", ident.replace('"', "\"\"")) }
```

Both branches stay a `String`, so the `Box<dyn ...>`, the unconstrained `QS`, and the empty where should be removable.

diesel_cli/tests/print_schema/print_schema_rust_injection/postgres/expected.snap
```diff
@@ -4,41 +4,75 @@ description: "Test: print_schema_rust_injection"
...
     use diesel::sql_types::*;
-    use super::sql_types::XIncludeEnvPModYDieselPostgresTypeNam;
+    use super::sql_types::XIncludeEnvPModYDieselPostgresTypeNa;
 
-    tests (id) {
+    /// Representation of the `test
+    /// " include!(env!("P"))` table.
+    ///
+    /// (Automatically generated by Diesel.)
+    #[sql_name = r#"test
```
LucaCappelletti94, line 67: This looks broken?

LucaCappelletti94 (changes requested):
I believe there are no testes to verify that a table named `a"."b` stops resolving to `a.b`, which is one of the pre-existing failures I believe?

diesel_cli/src/infer_schema_internals/pg.rs
weiznich: I would rather not deal with this kind of replacement if possible. It seems safer to me to just defer to the database given that this is possible here

diesel_cli/tests/print_schema/print_schema_rust_injection/postgres/expected.snap
weiznich: It's a multiline raw string literal. That's valid rust code and correctly represents the corresponding table name. The test harness does check if the code code compiles successfully

